PR #429, opened 2019-11-04
Update reject_application in vendor-api-0.8.0 file to use rejection schema

RobertM905:
### Context

Currently the reject application request body in the open-api spec file does not use the rejection schema. 

Update the open-api spec to make the reject application endpoint use the rejection schema. 
This is linked to the tech-docs PR here: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training-tech-docs/pull/108

### Guidance to review

Check the open-api spec to ensure the correct information is being displayed for rejecting applications.

### DIFF
--- a/app/presenters/vendor_api/single_application_presenter.rb
+++ b/app/presenters/vendor_api/single_application_presenter.rb
@@ -143,7 +143,6 @@ module VendorApi
       if application_choice.rejection_reason?
         {
           reason: application_choice.rejection_reason,
-          date: Time.now,
         }
       end
     end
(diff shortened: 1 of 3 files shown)


duncanjbrown:
Looks good, but should we remove the "date" field from `Rejection`? It doesn't make sense for vendors to send this if they also send it in the metadata. (If we need to send a date in future, we could make separate schemas for creating and receiving Rejections)

duncanjbrown:
@RobertM905 thanks, will approve this and https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training-tech-docs/pull/108 and https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training-tech-docs/pull/107 in advance of that change here and in the docs

RobertM905:
Removing date from the Rejection request body schema would also remove date from the api response. Did we want the rejection response schema to include date?

PR #430, opened 2019-11-04
Add ability to delete a degree

chubberlisk:
### Context

Candidates should be able to delete a degree, currently we don't allow this.

This PR adds the ability to delete a degree with some small refactors.

### Screenshots

image

image

image

### Guidance to review

Would recommend going through each commit and trying it out.

### Link to Trello card

[194 - Adding degree information](https://trello.com/c/ADqlESXx/194-adding-degree-information)

### DIFF
--- a/app/components/degrees_review_component.html.erb
+++ b/app/components/degrees_review_component.html.erb
@@ -2,7 +2,7 @@
   <%= render(SummaryCardComponent, rows: degree_form_rows(degree_form)) do %>
     <header class="app-summary-card__header">
       <h3 class="app-summary-card__title">
-        <%= formatted_degree_title(degree_form.qualification_type, degree_form.subject) %>
+        <%= degree_form.title %>
       </h3>
       <div class="app-summary-card__actions">
         <%= link_to t('application_form.degree.delete'), candidate_interface_degrees_destroy_path(degree_form.id), class: 'govuk-link' %>
--- a/app/components/degrees_review_component.html.erb
+++ b/app/components/degrees_review_component.html.erb
@@ -5,7 +5,7 @@
         <%= degree_form.title %>
       </h3>
       <div class="app-summary-card__actions">
-        <%= link_to t('application_form.degree.delete'), candidate_interface_degrees_destroy_path(degree_form.id), class: 'govuk-link' %>
+        <%= link_to t('application_form.degree.delete'), candidate_interface_confirm_degrees_destroy_path(degree_form.id), class: 'govuk-link' %>
       </div>
     </header>
   <% end %>
--- a/app/components/degrees_review_component.html.erb
+++ b/app/components/degrees_review_component.html.erb
@@ -4,6 +4,9 @@
       <h3 class="app-summary-card__title">
         <%= formatted_degree_title(degree_form.qualification_type, degree_form.subject) %>
       </h3>
+      <div class="app-summary-card__actions">
+        <%= link_to t('application_form.degree.delete'), candidate_interface_degrees_destroy_path(degree_form.id), class: 'govuk-link' %>
+      </div>
     </header>
   <% end %>
 <% end %>
--- a/app/controllers/candidate_interface/degrees/destroy_controller.rb
+++ b/app/controllers/candidate_interface/degrees/destroy_controller.rb
@@ -1,0 +1,23 @@
+module CandidateInterface
+  class Degrees::DestroyController < CandidateInterfaceController
+    def confirm_destroy
+      application_form = current_candidate.current_application
+      @degree = DegreesForm.find_by_application(application_form, degrees_params[:id])
+    end
+
+    def destroy
+      current_candidate.current_application
+        .application_qualifications
+        .find(degrees_params[:id])
+        .destroy!
+
+      redirect_to candidate_interface_degrees_review_path
+    end
+
+  private
+
+    def degrees_params
+      params.permit(:id)
+    end
+  end
+end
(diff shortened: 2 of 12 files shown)


app/components/degrees_review_component.html.erb
```diff
@@ -2,8 +2,11 @@
   <%= render(SummaryCardComponent, rows: degree_form_rows(degree_form)) do %>
     <header class="app-summary-card__header">
       <h3 class="app-summary-card__title">
-        <%= formatted_degree_title(degree_form.qualification_type, degree_form.subject) %>
+        <%= degree_form.title %>
```
alexdesi: just to confirm, 
it looks like we do not need any change to show the title, (for example,  `upcase` ? )
is that right?

app/components/degrees_review_component.html.erb
```diff
@@ -2,8 +2,11 @@
...
       </h3>
+      <div class="app-summary-card__actions">
+        <%= link_to t('application_form.degree.delete'), candidate_interface_degrees_destroy_path(degree_form.id), class: 'govuk-link' %>
```
alexdesi: nitpick:
you could use:
`candidate_interface_degrees_destroy_path(degree_form)`
instead of
`candidate_interface_degrees_destroy_path(degree_form.id)`

app/controllers/candidate_interface/degrees/destroy_controller.rb
```diff
@@ -0,0 +1,23 @@
+module CandidateInterface
+  class Degrees::DestroyController < CandidateInterfaceController
+    def confirm_destroy
+      application_from = current_candidate.current_application
```
alexdesi: you mean 
application_f**or**m ?

app/models/candidate_interface/degrees_form.rb
```diff
@@ -15,17 +15,51 @@ class DegreesForm
 
     validate :award_year_is_date, if: :award_year
 
-    def self.build_from_application(application_form)
-      application_form.application_qualifications.degrees.map do |degree|
+    class << self
```
alexdesi: good to use `class << self` 👍

app/controllers/candidate_interface/degrees/destroy_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+      @degree = DegreesForm.find_using_application(application_from, degrees_params[:id])
```
alexdesi: `find_by_application` sounds better than `find_using_application`, doesn't it?
actually `find_using_application_id`
... actually this is not just a find 😱

app/models/candidate_interface/degrees_form.rb
```diff
@@ -15,17 +15,51 @@ class DegreesForm
...
+        when 'first', 'upper_second', 'lower_second', 'third'
+          grade
+        else
+          if predicted_grade
+            'predicted'
+          else
+            'other'
+          end
+        end
+      end
```
alexdesi: it's nice that you used a private method for this logic,
but it sounds it should not be in an obj whose purpose is to create a form for Degree.

It might be better to have it in the model: `degree.get_grade` 
(it is still not the best place, but I think it's more appropriate)

config/routes.rb
```diff
@@ -75,6 +75,9 @@
 
         get '/review' => 'degrees/review#show', as: :degrees_review
         patch '/review' => 'degrees/review#complete', as: :degrees_complete
+
+        get '/delete/:id' => 'degrees/destroy#confirm_destroy', as: :degrees_destroy
```
alexdesi: what about `as: :confirm_degrees_destroy`? is a bit longer but it's more descriptive

app/models/candidate_interface/degrees_form.rb
```diff
@@ -15,17 +15,51 @@ class DegreesForm
...
+            subject: degree.subject,
+            institution_name: degree.institution_name,
+            grade: degree.grade,
+            predicted_grade: degree.predicted_grade,
+            award_year: degree.award_year,
+          )
+        end
+      end
+
+      def find_using_application(application_form, degree_id)
```
alexdesi: what the method does is not in line with the name `find_using_application` (it is just a small part),
in case we are already doing this way, do not consider this comment, but we should plan to refactor this.

spec/components/degrees_review_component_spec.rb
```diff
@@ -65,4 +67,13 @@
     expect(result.css('.app-summary-card__title').text).to include('BA Woof')
     expect(result.css('.app-summary-card__title').text).to include('BA Meow')
   end
+
+  it 'renders component along with a delete link for each degree' do
+    result = render_inline(DegreesReviewComponent, application_form: application_form)
+
+    expect(result.css('.app-summary-card__actions').text).to include(t('application_form.degree.delete'))
+    expect(result.css('.app-summary-card__actions a')[0].attr('href')).to include(
```
alexdesi: maybe you do not need to check the exact URL, since you have a system test that click on that link

spec/components/degrees_review_component_spec.rb
```diff
@@ -65,4 +67,13 @@
...
+      Rails.application.routes.url_helpers.candidate_interface_degrees_destroy_path(2),
```
alexdesi: 🙀

app/components/degrees_review_component.html.erb
chubberlisk: I believe you're right, but in this case it didn't work. It strangely couldn't find the `id`.

app/controllers/candidate_interface/degrees/destroy_controller.rb
chubberlisk: Updated it to `find_by_application`.

PR #436, opened 2019-11-04
Add extensions mechanism to API spec

mnacos:
### Context

In anticipation of vendor integrations with our API, we would like to extend the way the Apply API works to allow vendors to request or suppress sets of optional fields.  This is a kind of change introduced early on, so that we can support a variety of vendor tech stacks, some of which may not be able to handle the sudden appearance of new fields in responses.

The original idea was to allow vendors to pass arrays of fields as part of the request URI (query params), so that the response format is determined by the url and will remain constant as long as they do not update any urls they target in their scripts/programs.

It has also been suggested we should provide similar hide functionality, for the suppression of fields from responses, but this is considered low priority.

After some thought, we have decided to let vendors specify sets of fields instead of individual fields, for operational reasons.

It is important to introduce this change now, before vendors start building their integrations, and update the API docs accordingly.

JSON:API already supports a top-level ```included:```section, "containing  an array of resource objects that are related to the primary data and/or each other". This seems like a natural place for the optional sets of parameters we would like to support (if they can be organised as a "resource").

JSON:API also supports a ```fields:``` section, but "If a client requests a restricted set of fields for a given resource type, an endpoint MUST NOT include additional fields in resource objects of that type in its response." We are proposing not to support this, as showing/hiding individual fields will make vendor onboarding harder (compared with giving them a default set of fields) and lead to support/troubleshooting headaches (the number of actual urls being used, including query params, will explode).

Our current OpenAPI spec already includes mandatory ```first_name``` and ```last_name``` values for each candidate resource. Vendor X has requested ```combined_name``` and ```username``` fields for each candidate and a ```sector:``` attribute for each job listed.

Instead of allowing them to request these fields, we create an extension e.g. ext1, which adds these optional fields within an ```included:``` section, in a json structure mirroring the one contained in ```data:``` if the vendor requests it in the URL query params. Examples below:

Consider ```GET /applications/{application_id}```. This currently returns a big structure in ```data:``` with all associated resources nested in place, with no ids.

```json
{
  "data": {
    "id": "11fc0d3b2f",
    "type": "application",
    "attributes": {
      "status": "application_complete",
      "submitted_at": "2019-06-13T10:44:31Z",
      "updated_at": "2019-06-13T10:44:31Z",
      "personal_statement": "Since retiring from the Police Service in 2007...",
      "candidate": {
        "first_name": "Boris",
        "last_name": "Brown",
        "date_of_birth": "1985-02-02",
        "nationality": [
          "NL"
        ],
        "uk_residency_status": "UK Citizen",
        "english_main_language": true,
        "english_language_qualifications": "I have a degree in English Language and Literature from the University of Munich",
        "other_languages": "I am bilingual in Finnish and English",
        "disability_disclosure": "I am dyslexic"
      },
      "contact_details": {
        "address_line1": "45",
        "address_line2": "Dialstone Lane",
        "address_line3": "Stockport",
        "address_line4": "England",
        "postcode": "SK2 6AA",
        "country": "GB",
        "email": "boris.brown@example.com",
        "phone_number": "07944386555"
      },
      "course": {
        "start_date": "2020-09-10",
        "provider_ucas_code": "2FR",
        "course_ucas_code": "3CVK",
        "site_ucas_code": "K"
      },
      "qualifications": {
        "gcses": [
          {
            "qualification_type": "BA",
            "subject": "History and Politics",
            "grade": "2:1",
            "award_year": "1992",
            "institution_details": "University of Huddersfield",
            "equivalency_details": "string"
          }
        ],
        "degrees": [
          {
            "qualification_type": "BA",
            "subject": "History and Politics",
            "grade": "2:1",
            "award_year": "1992",
            "institution_details": "University of Huddersfield",
            "equivalency_details": "string"
          }
        ],
        "other_qualifications": [
          {
            "qualification_type": "BA",
            "subject": "History and Politics",
            "grade": "2:1",
            "award_year": "1992",
            "institution_details": "University of Huddersfield",
            "equivalency_details": "string"
          }
        ]
      },
      "work_experience": {
        "jobs": [
          {
            "organisation_name": "Cheadle Hulme High School",
            "start_date": "2020-09-10",
            "end_date": "2019-04-01",
            "role": "Whole School Literacy Specialist",
            "description": "I lead, develop and enhance the literacy teaching practice of others...",
            "working_with_children": true,
            "commitment": "string"
          }
        ],
        "volunteering": [
          {
            "organisation_name": "Cheadle Hulme High School",
            "start_date": "2020-09-10",
            "end_date": "2019-04-01",
            "role": "Whole School Literacy Specialist",
            "description": "I lead, develop and enhance the literacy teaching practice of others...",
            "working_with_children": true,
            "commitment": "string"
          }
        ]
      },
      "references": [
        {
          "name": "Julia Wild",
          "email": "julia@example.com",
          "phone_number": "07890 123456",
          "relationship": "Julia was my mentor at Cheadle Hulme High",
          "confirms_safe_to_work_with_children": true,
          "reference": "Boris is committed to the profession of teaching..."
        }
      ],
      "offer": null,
      "withdrawal": null,
      "rejection": null,
      "hesa_itt_data": {
        "sex": "1",
        "disability": "00",
        "ethnicity": "10"
      },
      "further_information": "I am expecting to be called for jury service in the next 12 months"
    }
  }
}
```

The suggestion is that ```GET /applications/{application_id}?include=ext1``` returns:

```json
{
  "data": {
    "id": "11fc0d3b2f",
    "type": "application",
    "attributes": {
      ... // all of the above
    }
  }, // plus
  "included": {
    "ext1": {
        "id": "11fc0d3b2f",
        "type": "application",
        "meta": { "extension": "1" },
        "attributes": {
          "candidate": {
            "combined_name": "Boris Brown",
            "username": "boris.brown",
          },
          "work_experience": {
            "jobs": [
              { "sector": "education" }
            ],
            "volunteering": [
              {}
            ]
          }
        }
      }
    }
  }
}
```

Furthermore, say they _only_ wanted the extra _candidate_ fields, they could request ```GET /applications/{application_id}?include=ext1.candidate```:

```json
{
  "data": {
    "id": "11fc0d3b2f",
    "type": "application",
    "attributes": {
      ... // all of the above
    }
  }, // plus
  "included": {
    "ext1": {
        "id": "11fc0d3b2f",
        "type": "application",
        "meta": { "extension": "1" },
        "attributes": {
          "candidate": {
            "combined_name": "Boris Brown",
            "username": "boris.brown",
          }
        }
      }
    }
  }
}￼
```

Finally, they could request fields from multiple extension versions at once, using a comma-separated list. For example, ```GET /applications/{application_id}?include=ext1.candidate,ext2.candidate``` gives:

```json
{
  "data": {
    "id": "11fc0d3b2f",
    "type": "application",
    "attributes": {
      ... // all of the above
    }
  }, // plus
  "included": {
    "ext1": {
        "id": "11fc0d3b2f",
        "type": "application",
        "meta": { "extension": "1" },
        "attributes": {
          "candidate": {
            "combined_name": "Boris Brown",
            "username": "boris.brown"
          }
        }
      }
    },
    "ext2": {
        "id": "11fc0d3b2f",
        "type": "application",
        "meta": { "extension": "2" },
        "attributes": {
          "candidate": {
            "favourite_colour": "brown"
          }
        }
      }
    }
  ]
}￼
```

To summarise, we propose:

- adding optional sets of fields in an ```included:``` section (this is part of the JSON:API spec)
- vendors can request these by adding an appropriate ```?include=``` param to their request URLs
- the ```?include=``` query parameter value will be a list of one or more comma-separated expressions
- these expressions will be of the format ```extension_id``` or  ```<extension_id>.<section>``` (the dot notation is also part of JSON:API)
- to be valid, ```<extension_id>``` values need to be chosen from a list of published extensions
- ```<extension_id>``` strings need to contain version information (without dots), e.g. ext1
- if multiple extensions are requested, parameter sets are supplied in separate extension blocks
- ```included:``` contains a list of objects with a structure identical to the top-level resource, so that the documents can be merged if the vendors choose to do so
- for one-to-many relationships, i.e. if the nested attributes structure contains one or more lists, the ordering of objects within these lists must be consistent across the entire document (within ```data:``` and the extension blocks). This is because we currently do not supply ids for nested objects. If this is revised, we could drop this requirement.

### Guidance to review

None yet

### Link to Trello card

[1181 - Add "extensions" field to spec and document it](https://trello.com/c/bE6ZOEyi)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -71,3 +71,4 @@ group :development, :test do
   gem 'pry-byebug'
   gem 'pry-rails'
 end
+


stevehook:
This looks like a big shift away from JSON:API. It isn't standard to have nested data (from associations) in the main attributes part of the top level object, normally all the associated data would be under the `included` key - it flattens out the object graph. One of the reasons for that is that it avoids having to duplicate information which happens a lot when returning lists as well as avoiding circular references.

It also seems like it would be extra work to combine all the various parts of the response into a single object. With requests that return lists then it would be even messier.

I guess I don't really understand the vendor requirement properly, it seems to me that if they need a set list of attributes/includes then we could support that in a standard way by listing them in a `fields` param - it's a little more effort but they would only need to do it once.

I'm not JSON:APIs #1 fan but I think that most times it is better than inventing new conventions.

stevehook:
To be honest I didn't realise until just now that we haven't been following the `JSON:API` conventions to do with relationships up to now anyway so your proposal makes a bit more sense to me now but I still think it's a bit odd to have the same resource spread across `attributes` and `included`.

mnacos:
> To be honest I didn't realise until just now that we haven't been following the `JSON:API` conventions to do with relationships up to now anyway so your proposal makes a bit more sense to me now but I still think it's a bit odd to have the same resource spread across `attributes` and `included`.

Yes, well, that's the thing. To use fields properly we'd have to fix the lack of relationships, ids etc. Which is an option, but requires a decision/commitment. The 'include' proposal I suggested is not far from the spirit of JSON:API, if you think about the included extensions as one-to-one associated resources (we would be a different type, e.g. "application-ext1"). We would also then need a relationships section. I've added some more comments on Slack.

duncanjbrown:
Closing this pending the follow-up PR (context on the Trello card https://trello.com/c/bE6ZOEyi/1181-add-extensions-field-to-spec-and-document-it)

PR #438, opened 2019-11-04
Update work experience validations

chubberlisk:
### Context

While product reviewing the ability to add work history, Rachel found that there was some missing date validations and errors.

This PR adds validation for `end_date`, ensuring that it is shows when editing a job and that it is not beyond the current year. Also enables nil handling for end date.

### Guidance to review

Go through commits.

There's probably a nicer way to handle the conditional validations, ideas are welcome!

### Link to Trello card

[192 - Adding work history](https://trello.com/c/G8slDphi/192-adding-work-history)

### DIFF
--- a/app/models/candidate_interface/work_experience_form.rb
+++ b/app/models/candidate_interface/work_experience_form.rb
@@ -78,17 +78,17 @@ module CandidateInterface
 
   private
 
+    def valid_date_or_nil(year, month)
+      date_args = [year, month, 1].map(&:to_i)
+      Date.new(*date_args) if year.present? && Date.valid_date?(*date_args)
+    end
+
     def end_date_blank?
       end_date_year.blank? && end_date_month.blank?
     end
 
     def end_date_valid
       errors.add(:end_date, :invalid) unless end_date
-    end
-
-    def valid_date_or_nil(year, month)
-      date_args = [year, month, 1].map(&:to_i)
-      Date.new(*date_args) if year.present? && Date.valid_date?(*date_args)
     end
 
     def start_date_valid
--- a/app/models/candidate_interface/work_experience_form.rb
+++ b/app/models/candidate_interface/work_experience_form.rb
@@ -34,9 +34,9 @@ module CandidateInterface
         start_date_day: work_experience.start_date.day,
         start_date_month: work_experience.start_date.month,
         start_date_year: work_experience.start_date.year,
-        end_date_day: work_experience.end_date.day,
-        end_date_month: work_experience.end_date.month,
-        end_date_year: work_experience.end_date.year,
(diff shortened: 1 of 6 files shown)


app/models/candidate_interface/work_experience_form.rb
```diff
@@ -32,9 +34,9 @@ def self.build_from_experience(work_experience)
         start_date_day: work_experience.start_date.day,
         start_date_month: work_experience.start_date.month,
         start_date_year: work_experience.start_date.year,
-        end_date_day: work_experience.end_date.day,
-        end_date_month: work_experience.end_date.month,
-        end_date_year: work_experience.end_date.year,
+        end_date_day: work_experience.end_date.nil? ? '' : work_experience.end_date.day,
```
benilovj: The following **should** do the same thing and be a bit more concise?
```suggestion
        end_date_day: work_experience.end_date&.day || ''
```

chubberlisk: Updated in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/438/commits/464651dbb7618b111c5dd43e7d56dc845e82f077!

app/models/candidate_interface/work_experience_form.rb
```diff
@@ -14,7 +14,9 @@ class WorkExperienceForm
     validates :working_with_children, inclusion: { in: %w(true false) }
 
     validate :start_date_valid
-    validate :start_date_after_end_date
+    validate :end_date_valid, if: :end_date_given?
+    validate :end_date_year_before_current_year, if: :end_date_valid?
+    validate :start_date_after_end_date, if: :start_date_and_end_date_valid?
```
duncanjbrown: The current list of date validations:

```
start_date_valid
end_date_valid
end_date_given?
end_date_valid?
end_date_year_before_current_year
start_date_after_end_end_date
```

A couple of observations

- It's odd to me that the first two are named for the valid state and the second two for the invalid state
- We have to have the two `?` functions to make this declarative logic work
- To follow the logic we have to jump around the file quite a lot

Given that these dates are either valid together or not, have you considered having a single function which expressed the whole validation? `validate :start_and_end_dates_valid`? This function could figure out the error state and then attach errors in the appropriate places in one pass at the end.

aldavidson: I've renamed `start_date_after_end_date` to `start_date_before_end_date` so that they all reflect the valid state now

aldavidson: Also found & fixed a few more edge cases:

partially-blank end date
end date year == current year, but month > current month

@duncanjbrown how does this look now?

PR #439, opened 2019-11-04
195 adding year grade to gcse

alexdesi:
### Context

This PR allow to add a new GCSE maths/english/science

- this add Form objects to allow to use different forms for the same qualification object
- the param `subject` is used render different version of the same page, this allow us to avoid building 3 different pages for maths, English and science GCSI

### Guidance to review
Specs are a good start

### Link to Trello card

https://trello.com/c/Yw5CJCaY/195-adding-maths-english-and-science-gcse-information


app/controllers/candidate_interface/gcse_details_controller.rb
```diff
@@ -1,15 +1,83 @@
 module CandidateInterface
   class GcseDetailsController < CandidateInterfaceController
```
chubberlisk: Considering we have `#edit_type`, `#update_type`, `#edit_details` and `#update_details`, it would be an idea create a namespace for our `GCSE` controllers. E.g.

- `Gcse::TypeController`
- `Gcse::DetailsController`
- and maybe a `Gcse::ReviewController`

With both having their own `#edit` and `#update` actions.

app/controllers/candidate_interface/gcse_details_controller.rb
```diff
@@ -1,15 +1,83 @@
...
+      )
     end
 
-    def update
+    def update_type
+      @subject = subject_param
+      @application_qualification = GcseQualificationTypeForm
+                                     .new(qualification_type: (params[:candidate_interface_gcse_qualification_type_form] || {}).fetch(:qualification_type, ''),
+                                          subject: subject_param,
+                                          level: 'gcse')
```
chubberlisk: This line is quite difficult to read, could we refactor this and potentially `params[:candidate_interface_gcse_qualification_type_form]` into a private method?

app/controllers/candidate_interface/gcse_details_controller.rb
```diff
@@ -1,15 +1,83 @@
...
   class GcseDetailsController < CandidateInterfaceController
-    def edit
-      subject = params[:subject]
+    # 1th step - Edit qualification type
+    def edit_type
+      @subject = subject_param
 
-      @application_qualification = ApplicationQualification.create(subject: subject, application_form: current_candidate.current_application, level: 'gcse')
+      @application_qualification = GcseQualificationTypeForm.new(
+        qualification_type: '',
```
chubberlisk: Does this need to be defined?

app/controllers/candidate_interface/gcse_details_controller.rb
```diff
@@ -1,15 +1,83 @@
 module CandidateInterface
   class GcseDetailsController < CandidateInterfaceController
-    def edit
-      subject = params[:subject]
+    # 1th step - Edit qualification type
+    def edit_type
+      @subject = subject_param
```
chubberlisk: Could we create a `before_action` for setting `@subject` as it's used in all methods?

app/controllers/candidate_interface/gcse_details_controller.rb
```diff
@@ -1,15 +1,83 @@
...
+
+    def gcse_qualification_details_form(grade:, award_year:)
+      GcseQualificationDetailsForm.new(
+        grade: grade,
+        award_year: award_year,
+      )
+    end
+
+    def subject_param
+      params[:subject]
```
chubberlisk: I believe here we should be using strong parameters (see https://guides.rubyonrails.org/action_controller_overview.html#strong-parameters), so only allowing the parameters we expect.

app/controllers/candidate_interface/gcse_details_controller.rb
```diff
@@ -1,15 +1,83 @@
...
+        grade: application_qualification.grade,
+        award_year: application_qualification.award_year,
+)
+
+      render :edit_details
+    end
+
+    def update_details
+      @subject = subject_param
+      current_application_qualification = ApplicationQualification.last
```
chubberlisk: This will only work if there is only a single user. If there are multiple concurrent users, anyone's qualification could be brought back. 

Also, this is doesn't scope for GCSEs only. So you could potentially get a degree. I think we should add a scope to `ApplicationQualification` for GCSEs, e.g. `scope :gcses, -> { where level: 'gcse' }` like done for degrees.

app/controllers/candidate_interface/gcse_details_controller.rb
```diff
@@ -1,15 +1,83 @@
...
-      subject = params[:subject]
+    # 1th step - Edit qualification type
+    def edit_type
+      @subject = subject_param
 
-      @application_qualification = ApplicationQualification.create(subject: subject, application_form: current_candidate.current_application, level: 'gcse')
+      @application_qualification = GcseQualificationTypeForm.new(
+        qualification_type: '',
+        subject: subject_param,
+        level: 'gcse',
```
chubberlisk: I believe we can use our enum here, `ApplicationQualification.levels['gcse']`?

app/controllers/candidate_interface/gcse_details_controller.rb
```diff
@@ -1,15 +1,83 @@
...
+    def update_details
+      @subject = subject_param
+      current_application_qualification = ApplicationQualification.last
+
+      details_form = gcse_qualification_details_form(
+        grade: params[:candidate_interface_gcse_qualification_details_form][:grade],
+        award_year: params[:candidate_interface_gcse_qualification_details_form][:award_year],
+      )
+
+      @application_qualification = details_form.save_base(current_application_qualification)
```
chubberlisk: Considering you have created separate form objects i.e. `GcseQualificationDetailsForm` and `GcseQualificationTypeForm`, they can just be named `#save` instead of `#save_base`.

app/controllers/candidate_interface/gcse_details_controller.rb
```diff
@@ -1,15 +1,83 @@
...
+      @subject = subject_param
       @application_qualification = ApplicationQualification.last
 
-      render :show
+      render :review
+    end
+
+  private
+
+    def gcse_qualification_details_form(grade:, award_year:)
```
chubberlisk: This refactoring doesn't add much value in my opinion, as it adds more lines than saves.

app/helpers/application_helper.rb
```diff
@@ -43,8 +43,12 @@ def current_namespace
...
   end
 
-  def heading_for_gcse_edit(subject)
-    t("gcse_details.heading.#{subject}")
+  def heading_for_gcse_edit_type(subject)
+    t("gcse_edit_type.heading.#{subject}")
+  end
+
+  def heading_for_gcse_edit_details(subject)
+    t("gcse_edit_details.heading.#{subject}")
```
chubberlisk: Nice!

app/models/candidate_interface/gcse_qualification_details_form.rb
```diff
@@ -0,0 +1,15 @@
+module CandidateInterface
+  class GcseQualificationDetailsForm
+    include ActiveModel::Model
+
+    attr_accessor :grade, :award_year
+    validates :grade, :award_year, presence: true
```
chubberlisk: I think we can add further validation for `award_year`, you can reuse what was used in degrees.

app/views/candidate_interface/gcse_details/edit_type.html.erb
```diff
@@ -3,18 +3,18 @@
...
       <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl govuk-!-margin-bottom-8">
-        <h1 class="govuk-fieldset__heading"><%= heading_for_gcse_edit(@application_qualification.subject) %></h1>
+        <h1 class="govuk-fieldset__heading"><%= heading_for_gcse_edit_type(@subject) %></h1>
       </legend>
 
       <%= f.govuk_error_summary %>
 
       <%= f.govuk_radio_buttons_fieldset :qualification_type, legend: { size: 'm', text: 'Type of qualification' } do %>
-
         <% select_gcse_qualification_type_options.each do |option| %>
```
chubberlisk: When a candidate chooses `Other UK qualifications`, they should be able to add in a text field their type.

image

Also `Non-UK qualification` isn't on the prototype, is there a reason why we have this?

app/views/candidate_interface/gcse_details/review.html.erb
```diff
@@ -3,7 +3,7 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-     <h1 class="govuk-heading-xl"><%= heading_for_gcse_show(@application_qualification.subject) %></h1>
+     <h1 class="govuk-heading-xl"><%= heading_for_gcse_show(@subject) %></h1>
 
       <section class="app-summary-card govuk-!-margin-bottom-6">
```
chubberlisk: For summary cards we use Rails components, see `app/components/degrees_review_component.rb` and `app/components/degrees_review_component.html.erb` for example.

The benefit of this is that it can be easily reused when a candidate can reviews their whole application before submitting.

config/locales/application_form.yml
```diff
@@ -221,3 +221,15 @@ en:
...
+            grade:
+              blank: Enter the grade
+            award_year:
+              blank: Enter the year
+        candidate_interface/gcse_qualification_type_form:
+          attributes:
+            qualification_type:
+              blank: Enter the type of qualification
+            level:
+              blank: Enter your level
```
chubberlisk: Is this necessary? Level is something we use internally and is not candidate facing.

spec/models/candidate_interface/gcse_qualification_details_form_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      application_form = create(:application_form)
+
+      qualification = ApplicationQualification.create(level: 'gcse', application_form: application_form)
+
+      details_form = CandidateInterface::GcseQualificationDetailsForm.new(grade: 'AB', award_year: '1990')
+
+      details_form.save_base(qualification)
+
+      expect(qualification.reload.grade).to eq('AB')
+      expect(qualification.reload.award_year).to eq('1990')
```
chubberlisk: Do we need to use `reload` here?  Tested and still green if removed.

app/models/candidate_interface/gcse_qualification_details_form.rb
```diff
@@ -0,0 +1,15 @@
...
+
+    def save_base(qualification)
+      return false unless valid?
+
+      qualification.update!(grade: grade, award_year: award_year)
+      qualification
```
chubberlisk: I don't think we need `qualification` at the end here as `#update` returns the updated object.

spec/system/candidate_interface/candidate_entering_gcse_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+    then_i_see_add_grade_and_year_page
+
+    when_i_fill_in_grade_and_year
+    and_i_click_save_and_continue
+
+    then_i_see_the_review_for_maths_gcse
+    and_i_see_correct_grade_and_awarded_year
+  end
+
+  scenario 'Candidate does not provide a qualification level' do
```
chubberlisk: Elsewhere we've created separate files for separate scenarios. Might be a thing we want to do here.

app/views/candidate_interface/gcse_details/review.html.erb
chubberlisk: Also, the summary card should be full-width instead of two-third. So outside of the `<div class="govuk-grid-column-two-thirds">`.

app/views/candidate_interface/gcse_details/edit_details.html.erb
```diff
@@ -0,0 +1,19 @@
...
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with model: @application_qualification, url: candidate_interface_gcse_details_update_details_path, builder: GOVUKDesignSystemFormBuilder::FormBuilder, method: :patch do |f| %>
+      <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl govuk-!-margin-bottom-8">
+        <h1 class="govuk-fieldset__heading"><%= heading_for_gcse_edit_details(@subject) %></h1>
+      </legend>
+
+      <%= f.govuk_error_summary %>
+
+      <%= f.govuk_text_field :grade, label: { text: 'Enter your qualification grade', size: 'm' } %>
```
chubberlisk: In the prototype, we use the width of 10. This can easily be done by passing in `width: 10`.

image

app/views/candidate_interface/gcse_details/edit_details.html.erb
```diff
@@ -0,0 +1,19 @@
...
+      <%= f.govuk_text_field :award_year, label: { text: 'Enter the year you gained your qualification', size: 'm' } %>
```
chubberlisk: In the prototype, we use the width of 4. This can easily be done by passing in `width: 4`.

image

app/views/candidate_interface/gcse_details/edit_type.html.erb
alexdesi: I spoke with Rachael, a new trello card is coming soon for the 'other UK qualification'
I'll remove the Non-UK since we do not need it as you said 👍

app/models/candidate_interface/gcse_qualification_details_form.rb
alexdesi: removed

spec/system/candidate_interface/candidate_entering_gcse_spec.rb
alexdesi: Not sure, this would require massive duplication, 
let's discuss for the following PR

app/components/review_gcse_qualification_component.html.erb
```diff
@@ -0,0 +1,24 @@
+<section class="app-summary-card govuk-!-margin-bottom-6">
```
chubberlisk: Very close! You can refactor this further and use the `SummaryCardComponent` by passing in `rows: gcse_qualifications_rows`. Have a look at `degrees_review_component.html.erb` as an example.

app/components/review_gcse_qualification_component.html.erb
```diff
@@ -0,0 +1,24 @@
...
+      <% gcse_qualification_rows.each do |row| %>
+        <div class="govuk-summary-list__row">
+          <dt class="govuk-summary-list__key">
+            <%= row[:key] %>
+          </dt>
+          <dd class="govuk-summary-list__value">
+            <%= row[:value] %>
+          </dd>
+          <dd class="govuk-summary-list__actions">
+            <%= link_to 'Change', 'edit_qualification_path' %>
```
chubberlisk: As a heads up, you can use the view helper `govuk_link_to` which adds the `govuk-link` CSS class. Not need to do anything here though when you refactor to use `SummaryCardComponent` in your GCSE review component.

app/components/review_gcse_qualification_component.html.erb
chubberlisk: Can we rename this component to be `GcseQualificationReviewComponent` to follow how we've named our other review components?

app/controllers/candidate_interface/gcse_details_controller.rb
chubberlisk: I definitely agree with you, `before_action` should be used with caution! But in this case I think it's the perfect use of it.

A nitpick, but I tend to see `before_action` used like this `before_action :set_subject` and then a private method defined.

app/controllers/candidate_interface/gcse_details_controller.rb
```diff
@@ -1,15 +1,76 @@
...
+
+  private
+
+    def subject_param
+      params.require(:subject)
+    end
+
+    def qualification_type_param
+      (params[:candidate_interface_gcse_qualification_type_form] || {}).fetch(:qualification_type, '')
     end
```
chubberlisk: Would doing something like this but for GCSEs achieve the same thing?

```ruby
def degrees_params
  params.require(:candidate_interface_degrees_form).permit(
    :qualification_type, :subject, :institution_name, :grade, :other_grade,
    :predicted_grade, :award_year
  )
end
```

alexdesi: that does not work because subject is not part of the form (`candidate_interface_gcse_qualification_type_form`), it is part of the URL param,
I'll see if I can make some improvements on the following PR

PR #443, opened 2019-11-06
Add support for async and scheduled jobs

mnacos:
### Context

This service will need to process applications automatically and periodically in order to enforce time-dependent business rules and modify application status/state. Therefore, we need a way to run domain logic code outside the context of web requests, reliably and in a proactive/scheduled way. 

After some discussion, we have agreed to add the following capabilities to the app:

#### Scheduling

This will be achieved via the ```clockwork``` gem. It alllows periodic jobs to be defined in a ruby file within the main Rails app. For example:

```ruby
# config/clock.rb

class Clock
  include Clockwork

  every(1.minute,  'SayHello') { Rails.logger.info "hi!" }
  every(5.minutes, 'ClockworkCheck') { ClockworkCheck.perform_async }
end
```

Clockwork requires the scheduling daemon to be run as a separate process to the Rails server, but it uses the same codebase, so all models/services/workers defined in Rails are available to be triggered. This extra process can be started like this: ```bundle exec clockwork ./config/clock.rb```

**Note** While many Rails server instances may be run in a scaling out/load-balancing scenario, there must only always be ONLY ONE clockwork process, otherwise tasks will be triggered multiple times. Furthermore, in production this process should be supervised and restarted, if needed.

While it is technically possible to perform any kind of processing within the 'clock' process, this is to be avoided, as any application errors could terminate the scheduler and prevent other tasks from running. Good practice suggests the 'clock' process should only be used to trigger/enqueue tasks, which are then processed within the context of a background processing system. That's why config/clock.rb should be a list of times and Sidekiq worker perform_async statements.

#### Background processing

We will use the ```sidekiq``` gem, which is the current standard for Rails background processing. This also needs to be run as a separate process (e.g. ```bundle exec sidekiq -c 5```, where 5 is a concurrency setting), but it also introduces an infrastructure requirement for Redis.

Workers are usually placed within ```app/workers``` and can call any other classes within the Rails app, such as service objects to achieve their goals. 

This PR introduces ```clockwork``` and ```sidekiq``` gems to the Rails app.  For sidekiq to work, ```REDIS_URL``` environment variable needs to be set (e.g. ```REDIS_URL=redis://redis:6379/1```).

There is a new Procfile, defining three different processes:

```
web: bundle exec puma -C config/puma.rb
worker: bundle exec sidekiq -c 5
clock: bundle exec clockwork config/clock.rb
```

'Clock' schedules tasks and 'worker' processes them, while 'web' corresponds to the web server.

Switching from a single process to a group of three, I have had to refactor the logging code, so that log entries can be filtered based on the service they correspond to (web/worker/clock). Service identification is to be done through the ```SERVICE_NAME``` environment variable. Refactoring was also required to move the generation of custom logging attributes (e.g. candidate_id, provider_id) from lograge (which has access to controller methods) to logstash-logger (which doesn't), as it wasn't possible to get lograge to modify the logging format of sidekiq or clockwork. In order to pass the relevant values from a web request context, ```request_store_rails``` was introduced.

image

As you can see in the screenshot, as well as ```service```, Logstash now will receive metadata for sidekiq jobs, such as elapsed time and the worker class.

### Guidance to review

In development, there are two ways to run all new processes (web, worker & clock) at the same time:

1. The docker development environment is pre-configured to bring all three processes up. You can still use ```make setup && make serve``` as before.
2. If you prefer the more traditional local ```rails s``` you have the option to either not run sidekiq and clockwork (may not be relevant to your work) or start all these processes by running ```foreman start```. You'll need a local Redis instance running on your machine (port 6379).

To observe the logging, as configured by ```LogstashLogging```, please set ```LOGSTASH_ENABLE=true```.

There is also some test coverage around:

- The specific schedule defined in ```config/clock.rb```. This is mainly to verify schedule definitions.
- The example worker, ```ClockworkCheck```
- The logstash configurator ```LogstashLogging``` (needed to retro-fit)

### Link to Trello card

[1209 - Run an asynchronous job on the app (required for time-driven state changes)](https://trello.com/c/GT3lqHsg)

### Env vars

Although two new environment variables are used in the app (```REDIS_URL``` and ```SERVICE_TYPE```), these are taken care of automatically in docker-compose.yml for local development work and testing, while these  should be set automatically in Azure pipelines for production enviroments (SERVICE_TYPE, in particular, needs to be different for each process according to its role/start command).


.rubocop.yml
```diff
@@ -14,6 +14,8 @@ AllCops:
     - 'config/initializers/nationalities.rb'
     - 'lib/tasks/cucumber.rake'
     - 'features/support/env.rb'
+    - 'spec/support_specs/clockwork_spec.rb'
+    - 'spec/support/test_helpers/capture_stdout_helper.rb'
```
mnacos: rubocop complains about using $stdout directly, instead of using the rspec matcher .output(...).to_stdout -- using the rspec matcher is not appropriate here because we need to capture STDOUT output and process it (e.g. JSON.parse)

mnacos: the other complaint is about ran_job?, suggesting be_ran_job instead. I don't like how 'be_ran_job' reads, and the clockwork/test gem does not support it, anyway.

Gemfile
```diff
@@ -36,10 +36,15 @@ gem 'actionview-component'
...
 
+# Background processing
+gem 'sidekiq'
+gem 'clockwork'
+
 # Logging
 gem 'lograge'
 gem 'logstash-logger'
 gem 'logstash-event'
+gem 'request_store_rails'
```
mnacos: This is a better way to share vars across various context within a request, compared to Thread.current or request_store. It is needed because LogStashLogger operates outside any of the usual rails contexts, such a request/controller etc.

tpinney:
Do we want to hold fire merging this change in until we can roll the Azure templates into?

mnacos:
> Do we want to hold fire merging this change in until so we can roll the Azure templates into it at the same time?

Yes, we can hold this back until the Azure templates are in. But we can do code review, as I've refactored some application code, logging in particular.

spec/support_specs/logstash_logging_spec.rb
```diff
@@ -0,0 +1,95 @@
+require 'rails_helper'
+
+RSpec.describe LogstashLogging do
+  include TestHelpers::CaptureStdoutHelper
+
+  before {
+    @lograge = Struct.new(:enabled, :base_controller_class, :custom_options, :formatter).new
+    @rails_config = Struct.new(:logger, :lograge).new
+    allow(@rails_config).to receive(:lograge).and_return(@lograge)
```
duncanjbrown: Would using `double` be slightly more expressive here? `rails_config = double("config", lograge: lograge)`

spec/support_specs/logstash_logging_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+  }
```
duncanjbrown: multilines `before` blocks with `{}` are also non-standard — please could you turn them into `do..end`s?

Gemfile
```diff
@@ -61,6 +67,8 @@ group :test do
   gem 'simplecov', require: false
   gem 'cucumber-rails', require: false
   gem 'database_cleaner'
+  gem 'rspec-sidekiq'
```
duncanjbrown: Are we using this anywhere?

In any case, are we able to use the `ActionJob` testing facilities? https://edgeguides.rubyonrails.org/testing.html#testing-jobs

adr/0009-async-and-scheduled-jobs.md
```diff
@@ -0,0 +1,75 @@
+# 9. Async and scheduled jobs
```
duncanjbrown: Thank you!

adr/0009-async-and-scheduled-jobs.md
```diff
@@ -0,0 +1,75 @@
...
+
+Date: 2019-11-07
+
+## Status
+
+Accepted
+
+## Context
+
+This service will need to process applications automatically and periodically in order to enforce time-dependent business rules and modify application status/state. Therefore, we need a way to run domain logic code outside the context of web requests, reliably and in a proactive/scheduled way.
```
duncanjbrown: just `state` instead of `status/state` please, it's confusing enough as it is 🙂

adr/0009-async-and-scheduled-jobs.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Consequences
+
+Some additional work will be required on Azure in order to:
+
+- Run multiple containers as separate services on Azure, all based on the same Docker image and the same set of environment variables, but using different start commands.
+- Add a Redis instance/service to the stack, ideally using an managed/hosted service with the Azure ecosystem.
+- Logging will need to be adapted to distinguish between logs from different services.
+
+These services will need to be supervised by the platform and restarted in case of errors.
```
duncanjbrown: Great ADR 🌟

app/lib/logstash_logging.rb
```diff
@@ -1,49 +1,80 @@
...
+      event['ctx'] = ctx
+    end
+  end
 
-    # Make stdout_logger the standard Rails logger
-    stdout_logger = ActiveSupport::Logger.new(STDOUT)
-    stdout_logger.formatter = config.log_formatter
-    config.logger = ActiveSupport::TaggedLogging.new(stdout_logger)
+  def self.enable(rails_config)
+    # Add domain, params etc. to the logs
```
duncanjbrown: This is gnarly! (the comment is an understatement 😄)

I feel comfortable with the simple constant-to-hash assignments, but the attribution stuff feels like too much to do in this file.

There are a few cross-cutting concerns in this method: 

- mapping `domain`, `service` and `params`
- mapping the user identity (`candidate`, `provider_id` etc)
- mapping the `exception` field (in a different way to the above)
- configuring lograge

In the case of the identity mapping, is there a case for this hash being provided by the controller in the way that `add_candidate_id_to_log` does? Could we simplify the logstash configuration by forcing controllers to implement some interface like `add_user_identity_to_log`, have them set something like

```
RequestLocals.store[:identity] = {candidate_id: foo}
```

and then in the logstash merge that in: `event.merge(RequstLocals.store[:identity])`.

spec/support_specs/logstash_logging_spec.rb
```diff
@@ -0,0 +1,95 @@
+require 'rails_helper'
+
+RSpec.describe LogstashLogging do
+  include TestHelpers::CaptureStdoutHelper
+
+  before {
+    @lograge = Struct.new(:enabled, :base_controller_class, :custom_options, :formatter).new
```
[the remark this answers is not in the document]
tijmenb: Well, since you're asking! I do have many opinions on this.

I prefer to avoid let/ivar/before altoghether, in the case of this test I'd do:

```rb
describe 'ClockworkCheck' do
  it 'logs it has run to the default Rails logger' do
    logger = stub_rails_logger

    ClockworkCheck.new.perform
    
    expect(logger.string).not_to be_blank
  end

  it 'says "clockwork is running..."' do
    logger = stub_rails_logger

    ClockworkCheck.new.perform

    expect(logger.string).to match(/clockwork is running\.\.\./)
  end
end

def stub_rails_logger
  stringio_output = StringIO.new
  stringio_logger = Logger.new(stringio_output)
  allow(Rails).to receive(:logger).and_return(stringio_logger)
  stringio_output
end
```


https://thoughtbot.com/blog/four-phase-test
https://thoughtbot.com/blog/lets-not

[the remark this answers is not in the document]
tijmenb: Well, since you're asking! I do have many opinions on this.

I prefer to avoid let/ivar/before altoghether, in the case of this test I'd do:

```rb
describe 'ClockworkCheck' do
  it 'logs it has run to the default Rails logger' do
    logger = stub_rails_logger

    ClockworkCheck.new.perform
    
    expect(logger.string).not_to be_blank
  end

  it 'says "clockwork is running..."' do
    logger = stub_rails_logger

    ClockworkCheck.new.perform

    expect(logger.string).to match(/clockwork is running\.\.\./)
  end
end

def stub_rails_logger
  stringio_output = StringIO.new
  stringio_logger = Logger.new(stringio_output)
  allow(Rails).to receive(:logger).and_return(stringio_logger)
  stringio_output
end
```


https://thoughtbot.com/blog/four-phase-test
https://thoughtbot.com/blog/lets-not

app/lib/logstash_logging.rb
```diff
@@ -1,5 +1,6 @@
 class LogstashLogging
   DOMAIN_FOR_LOGS = Rails.env.production? ? AzureEnvironment.hostname : Socket.gethostname
+  SERVICE_NAME = ENV['SERVICE_NAME'] # e.g. web, worker or clock
```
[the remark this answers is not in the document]
mnacos: I still think 'service' fits, because all major cloud providers seem to use 'service' when talking about stack components that can be scaled or load-balanced independently. It seems to me that each of these container types should be defined in Azure in a way that lets them scale independently. What are your thoughts @tpinney?

But yes, SERVICE_NAME is a bit misleading. How about SERVICE_TYPE instead?

docker-compose.yml
```diff
@@ -6,21 +6,61 @@ services:
       - db_data:/var/lib/postgresql/data
     environment:
       - POSTGRES_PASSWORD=developmentpassword
+  redis:
+    image: redis
   web:
     build: .
+    image: apply-for-postgraduate-teacher-training
```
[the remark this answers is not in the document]
mnacos: Yes, by building it and naming it, you can refer to it in the other sections, and docker-compose will build it once, not three times.

app/lib/params_logging.rb
```diff
@@ -0,0 +1,10 @@
+module ParamsLogging
+  IGNORE_PARAMS = %w(candidate authenticity_token).freeze
+
+  def add_params_to_request_store
+    params = request.try(:params)
```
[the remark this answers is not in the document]
stevehook: If this is here to guard against `request` being `nil` then you can do `request&.params`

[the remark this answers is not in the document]
mnacos: Being a bit paranoid, I guess, in case the module is included in a sidekiq worker or somewhere outside a web/request context. Will rename accordingly.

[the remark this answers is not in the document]
mnacos: On second thoughts, I'll assume request.params is available and that the module will only be included in controllers. This way I can move ```before_action :add_params_to_request_store``` to the module.

Gemfile
```diff
@@ -71,3 +71,4 @@ group :development, :test do
   gem 'pry-byebug'
   gem 'pry-rails'
 end
+
```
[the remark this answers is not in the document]
mnacos: fixed

app/lib/params_logging.rb
```diff
@@ -0,0 +1,10 @@
...
+    if params
+      RequestLocals.store[:params] = params.except(*IGNORE_PARAMS)
```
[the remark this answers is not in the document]
mnacos: It's not about logstash, it's about logstash-logger, the gem. The way it operates, event objects do not get access to a controller/request context, so it is impossible to pass request-related information, such as query params or headers to be logged directly. The logstash-logger docs suggest [https://github.com/dwbutler/logstash-logger#logging-http-request-data](https://github.com/dwbutler/logstash-logger#logging-http-request-data), but I've decided to use [https://github.com/ElMassimo/request_store_rails](https://github.com/ElMassimo/request_store_rails), which is meant to be better.

So, the pattern here is the controller exposes what they need in something akin to a global variable, albeit restricted to the current request. The logger then checks if this information is present and includes in the log attributes.

I have added IGNORE_PARAMS as a way to omit some params from the log output (e.g. all candidate-related ones, to suppress information from form submissions).

Gemfile
mnacos: I was going to use this but opted for standard have_received matchers at the ```perform_async``` level instead. Will remove

mnacos: About ActionJob, which can work with sidekiq, we'll need a separate decision. Yes, ActiveJob is a more abstract interface and it is part of Rails, but it also limits you to functionality common across all supported queuing systems.

For clockwork-related triggering, I think ActiveJob is overkill, and it'll be easy to switch to ActiveJob if we need to, we'll just grep for ```perform_async``` calls in the code.

mnacos: For testing, if we wanted ActiveJob-like inline execution, Sidekiq supports this (e.g. by adding ```Sidekiq::Testing.inline!``` to spec_helper.rb

I thought about this and it would make sense for some feature tests (when using 'inline' mode, the worker code gets executed immediately at the point of queuing, bypassing any queuing and need for a sidekiq process).

However, for most cases, I think we should be testing a controller action or a scheduled task enqueues a job of the appropriate type (e.g. ClockworkCheck), then test the actual worker itself or its associated service for business logic.

adr/0009-async-and-scheduled-jobs.md
mnacos: yes, could be read as both Rails app or an applicant's application (both work!)

app/controllers/application_controller.rb
```diff
@@ -1 +1,4 @@
-class ApplicationController < ActionController::Base; end
+class ApplicationController < ActionController::Base
+  include ParamsLogging
+  before_action :add_params_to_request_store
```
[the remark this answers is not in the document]
mnacos: I've renamed this to LogRequestParams, to make it clear it should only be included in controllers, where a request context is guaranteed.

A middleware is a great idea and I started exploring it for a bit but I am not sure it's appropriate, as it would be a middleware depending on another middleware and harder to maintain. With the before_action part of the module now including LogRequestParams to a controller is not much work, and there may be cases we want to suppress the logging of query params entirely in some controllers.

Procfile
```diff
@@ -0,0 +1,3 @@
+web: bundle exec puma -C config/puma.rb
+worker: bundle exec sidekiq -c 5
+clock: bundle exec clockwork config/clock.rb
```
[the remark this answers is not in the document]
mnacos: good call, adding...

app/lib/logstash_logging.rb
mnacos: Please previous comment about logstash-logger not having access to a controller context. Lograge does, and in previous iterations I had such methods in the relevant base controllers. With logstash-logger, however, the only communication strategy between the controller and the logger is through ```request_store_rails``` or equivalent.

I have kept lograge because it gives us the one line per request, does away with ActionView logging (which partials are being rendered etc.) and has access to 'exception' in case of errors.

But I've moved all other attribution to logstash-logger, because I couldn't get lograge to affect sidekiq and clockwork logs.

I like your 'identity' event.merge suggestion, will update.

app/lib/logstash_logging.rb
```diff
@@ -1,51 +1,65 @@
...
 
-    if ENV['LOGSTASH_REMOTE'] == 'true'
-      tcp_logger = LogStashLogger.new(type: :tcp,
-                                      host: ENV.fetch('LOGSTASH_HOST'),
-                                      port: ENV.fetch('LOGSTASH_PORT'),
-                                      ssl_enable: ENV.fetch('LOGSTASH_SSL') == 'true')
+    logstash_logger = \
+      if ENV['LOGSTASH_REMOTE'] == 'true'
+        LogStashLogger.new(
+          type: :multi_delegator,
```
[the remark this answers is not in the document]
duncanjbrown: Please could you clarify?

app/lib/logstash_logging.rb
```diff
@@ -1,49 +1,76 @@
...
+    rails_config.lograge.formatter = Lograge::Formatters::Logstash.new
+  end
 
-    if ENV['LOGSTASH_REMOTE'] == 'true'
-      tcp_logger = LogStashLogger.new(type: :tcp,
-                                      host: ENV.fetch('LOGSTASH_HOST'),
-                                      port: ENV.fetch('LOGSTASH_PORT'),
-                                      ssl_enable: ENV.fetch('LOGSTASH_SSL') == 'true')
+  def self.add_identity_fields(event)
+    identity_hash = RequestLocals.fetch(:identity) { nil }
```
duncanjbrown: `RequestLocals` acts like a normal Hash: it returns `nil` for a missing value, eg `RequestLocals[:nonsense] == nil`

app/lib/logstash_logging.rb
```diff
@@ -1,49 +1,76 @@
...
+    identity_hash.each { |key, val| event[key] = val } if identity_hash
```
duncanjbrown: Does `event.merge(identity_hash)` not do the trick?

spec/support_specs/logstash_logging_spec.rb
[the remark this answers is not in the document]
duncanjbrown: I'd prefer to follow @tijmenb's suggestion here unless you have objections?

app/lib/log_request_params.rb
```diff
@@ -0,0 +1,13 @@
+module LogRequestParams
+  IGNORE_PARAMS = %w(candidate authenticity_token).freeze
```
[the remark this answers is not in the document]
mnacos: The original scope of this story was to log/track vendor api requests. We decided to do this by shipping application logs, which are not restricted to vendor api requests. Considering the logs include request params, any form submission at the candidate side of things will be logged, and params may include sensitive data. So, I've added candidate to IGNORE_PARAMS to err on the side of caution.

app/lib/logstash_logging.rb
mnacos: Yes, but a block is required (removing the block gives ```LocalJumpError: no block given (yield)```). I'll change this to ```RequestLocals.fetch(:identity) {}```

app/lib/logstash_logging.rb
mnacos: No, it's not a hash, it's a LogStash::Event object that provides a hash-like interface.

mnacos:
> Great stuff. I think GitHub collapsed some of my comments last time! I like the `add_identity_to_log` approach. Only two questions left:
> 
> * what's going on with `multi_delegator` and could it be mentioned in the commit
> * why do we exclude the param `candidate` from the logs?
> 
> I think the simplest way forward on the spec style discussion is to adopt the status quo as expressed by Tijmen while we're here.

The multi_delegator is logstash-loggers way to define multiple log outputs. We need multiple log targets because the standard way to log from docker containters is to log to stdout, but we would also like to ship application logs to Logit, for dashboards, analysis etc. Multi_delegator assumes the same log format for all targets.

The previous version of this had lograge logging to stdout, then we were adding a logstash logger to the log dispatch chain for shipping applications logs. This approach wasn't working for sidekiq/clockwork, so I've changed the main logger to be the logstash-logger one.

app/lib/logstash_logging.rb
mnacos: https://github.com/ElMassimo/request_store_rails#atomicity

spec/support/test_helpers/logging_helper.rb
```diff
@@ -0,0 +1,33 @@
...
+    def capture_stdout
+      stringio_output = StringIO.new
+      old_stdout = $stdout
+      $stdout = stringio_output
+      yield
+      $stdout = old_stdout
+      stringio_output.string
+    end
+
+    def capture_logstash_output(config)
```
duncanjbrown: nice :+1:

spec/workers/clockwork_check_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+
+RSpec.describe ClockworkCheck do
+  include TestHelpers::LoggingHelper
+  let(:rails_config) { environment_config_double }
+
+  describe 'ClockworkCheck' do
+    let(:output) { capture_logstash_output(rails_config) { ClockworkCheck.new.perform } }
+
+    it 'logs it has run to the default Rails logger' do
+      expect(output).not_to be_blank
```
duncanjbrown: lovely 👌

PR #446, opened 2019-11-06
Reverse the touch between application_choice and application_form

RobertM905:
### Context

Currently when an application_choice is updated by a provider the parent application_form is also updated.
This may cause some odd behaviours for other providers.

Reverse the touch relationship between application_choice and application_form.

### Guidance to review

This is a Draft PR to discuss and get feedback.

### Link to Trello card

[123 - Example Trello card](http://trello.com/123-example-card)

### Env vars

### DIFF
--- a/app/models/application_choice.rb
+++ b/app/models/application_choice.rb
@@ -1,7 +1,7 @@
 class ApplicationChoice < ApplicationRecord
   before_create :set_initial_status
 
-  belongs_to :application_form, touch: true
+  belongs_to :application_form
   belongs_to :course_option
   has_one :course, through: :course_option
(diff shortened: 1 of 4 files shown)


duncanjbrown:
@RobertM905 can we mark this ready?

duncanjbrown:
@tijmenb without this fix providers retrieving applications over the API will see them out of order. We're not 100% sure that we need to update ApplicationChoice when ApplicationForm is updated (this may never happen) but we ought to fix the bug

tijmenb:
Ok, let's go for it. @RobertM905 can you rebase?

PR #447, opened 2019-11-06
Add working day feature

stevehook:
### Context

There are a number of time calculations that depend on working days. e.g. An application is automatically rejected on behalf of a provider if that provider failed to make a decision within 40 working days of receiving it. Working days are defined as all weekdays except bank holidays. So 5 working days is typically 7 or more calendar days or more depending on when the start date is and whether any bank holidays come into play. This PR attempts to lay the foundation for making time calculations that respect working days.

- Add the `business_time` gem for making working day based time calculations.
- Add the `holidays` gem to provide an authoritative list of UK bank holidays that can be used to configure `business_time`.
- Switch the `ApplicationDate` service class over to use working days.
- Also changed the time components of `ApplicationDate#edit_by` and `respond_by` to be the end of the day because this is what the requirements say.
- Set the default timezone for the application to be `London` so we are not using the system time of the server (whatever that might be).
- Replaced a few calls to `Time.now` with `Time.zone.now` and `Time.new` with `Time.zone.local` so that all time calculations are done in local (UK) time.

### Guidance to review

- The timezone config.
- I've changed the 7 day limit on editing a submitted application by a candidate to 5 working days because that's what the business rules spreadsheet specifies. However I've converted the 40 day limit on providers giving feedback directly to 40 working days because I can't find any spec for what that should be. If you know these numbers are wrong please shout, however we are going to need to revisit these later anyway to make them configurable.
- I've assumed that all providers work to the same holiday calendar and that they use standard UK bank holidays as defined in the `holiday` gem. If there are exceptions to this or there are 'extra' holidays that all providers take then we need to rethink things.
- I've added specs for `BusinessTime` which is a pretty odd thing to do for a gem. These tests are really just here to ensure that the two new gems are configured correctly. I'm in two minds about whether to remove these.

### Link to Trello card

[1195 - Working Days](https://trello.com/c/7wnO49pH/1195-add-working-day-feature)

### Env vars

No.


app/models/application_dates.rb
```diff
@@ -16,7 +16,7 @@ def edit_by
   end
 
   def days_remaining_to_edit
-    (edit_by - Time.now.to_date).to_i
+    (edit_by.to_date - Time.now.to_date).to_i
```
duncanjbrown: What is this object if not a date? Should `ApplicationDate` insulate us from this?

config/initializers/business_time.rb
```diff
@@ -1,5 +1,5 @@
-# Holidays.between(Date.civil(2017, 1, 1), 2.years.from_now, :gb_eng, :observed).map do |holiday|
-#   BusinessTime::Config.holidays << holiday[:date]
-# end
+Holidays.between(Date.civil(2017, 1, 1), 2.years.from_now, :gb_eng, :observed).map do |holiday|
```
duncanjbrown: Why go back to 2017?

app/models/application_dates.rb
```diff
@@ -12,7 +12,7 @@ def respond_by
   end
 
   def edit_by
-    7.business_days.after(submitted_at)
+    5.business_days.after(submitted_at)
```
duncanjbrown: Given that specs are entering our world via the spreadsheet and cucumber, does this fix invite us to pull the cucumber spec in? (For subsequent PR?)

spec/services/business_time_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+# rubocop:disable RSpec/RepeatedExample
+RSpec.describe BusinessTime do
+  [
+    Date.new(2019, 11, 4),
+    Date.new(2019, 11, 5),
+    Date.new(2019, 11, 6),
+    Date.new(2019, 11, 7),
+    Date.new(2019, 11, 8),
+  ].each do |date|
```
duncanjbrown: Not sure about this.

- I find this spec hard to follow as I don't know (without looking it up) that these are all weekdays
- I feel ambivalent about testing this as it's `BusinessTime`'s job to handle it? (For the intersection of `BusinessTime` and `Holiday` I see a bit of value)

config/initializers/business_time.rb
stevehook: Fair point - lets make 2019 year zero

spec/services/business_time_spec.rb
stevehook: I'm not too sure about these either. I can perhaps update the descriptions to make them a little clear. But whether we should include these at all is still an open question as far as I'm concerned. I'm thinking that if we have other higher level feature specs that exercise the working days logic and config then we don't need these. Drop them?

app/models/application_dates.rb
stevehook: I think I need to work on this...

But I think `edit_by` (and friends) would need to be a `Time` rather than a `Date` because the cutoff would be midnight (UK time).

stevehook: This opens up a can of worms around timezones. `end_date` is basically `(submitted at time + 5 working days).end_of_day_in_uk` (if that makes sense). I've put some notes on this commit https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/447/commits/f342915711431614a4d001dbd57167a4dcdd9e44 about what I think is the right thing to do here.

spec/services/business_time_spec.rb
stevehook: Removed as discussed.

config/initializers/business_time.rb
```diff
@@ -0,0 +1,5 @@
+Holidays.between(Date.civil(2019, 1, 1), 2.years.from_now, :gb_eng, :observed).map do |holiday|
+  BusinessTime::Config.holidays << holiday[:date]
+end
+
+BusinessTime::Config.end_of_workday = '11:59 pm'
```
benilovj: I think we might need to start the working day at `0:00` as well, otherwise there's some weirdness about applications submitted before 8am (or whatever the default start of the business day is).

spec/components/application_complete_content_component_spec.rb
```diff
@@ -19,7 +19,7 @@ def render_result
   end
 
   it 'renders with correct respond by date' do
-    expect(render_result.text).to include('1 December 2019')
+    expect(render_result.text).to include('17 December 2019')
```
benilovj: It's a bit of a "smell" that we're having to change this spec as a result of changes in the application form business logic. In this spec, we are coupling an actual factory-created application form object into the component, meaning that any changes to the application form logic class breaks the spec.

One proposal would be to make  `application_form` in `render_result` a stub, which stubs out the application form methods that are called by the component. This would also make it much more explicit where the dates (e.g. 17 December 2019) are coming from.

stevehook: @benilovj I've stubbed out the `ApplicationForm` and more importantly `ApplicationDates` so the component test should now have no external implementation dependencies. Let me know if this makes it cleaner or not.

benilovj: Yes, this feels much clearer now, thanks.

PR #449, opened 2019-11-06
Add ability to add and review other qualification

chubberlisk:
### Context

Candidates should be able to add other qualifications, not just degrees and English, Maths and Science GCSEs.

This PR adds the ability to add an other qualification with validation and then review it by introducing:

- the form object `OtherQualificationForm` with validation
- the review component `OtherQualificationsReviewComponent`

image

image

image

image

image

### Guidance to review

The diff is large but chunks of it are just tests and translations. The code written is following patterns we have set out before so my assumption is that it won't be complex to review. Recommend going through commits. 

However, please let me know if my assumption is wrong and I'll happily split this into separate PRs!

### Link to Trello card

[204 - Adding other qualifications](https://trello.com/c/qU3R6EVL/204-adding-other-qualifications)

### DIFF
--- a/app/components/other_qualifications_review_component.html.erb
+++ b/app/components/other_qualifications_review_component.html.erb
@@ -1,0 +1,9 @@
+<% @qualifications.each do |qualification| %>
+  <%= render(SummaryCardComponent, rows: other_qualifications_rows(qualification)) do %>
+    <header class="app-summary-card__header">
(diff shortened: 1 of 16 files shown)


chubberlisk:
Never mind, I'm going to split this PR into two. 🔪 2️⃣ Mah hart.

PR #453, opened 2019-11-07
Simplify rendering of lists in summary card

tijmenb:
### Context

Sometimes the value in a summary card is made up of multiple different elements, like the different parts of an address. We're currently `joining` those in the presenter with a `<br>` in between.

This means that we have to render the value as HTML, which is a security risk. We've mitigated against that by being careful about which things to render as HTML using the `DANGEROUS_html_value` argument. This was introduced in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/394.

This commit changes the approach to avoid all sanitising and passing of unsafe values.

I also considered the following:

```rb
<% if row[:value].is_a?(Array) %>
<%= safe_join(row[:value], '<br>'.html_safe) %>
...
```

But I don't think that is clearer.

### Guidance to review

Code

### Link to Trello card

No card.

### Env vars

- [x] If this PR introduces new environment variables, they have been added to all necessary parts of the Azure config based on the [documentation in the README](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training#azure-hosting-devops-pipeline)

### DIFF
--- a/app/components/summary_card_component.html.erb
+++ b/app/components/summary_card_component.html.erb
@@ -8,10 +8,10 @@
             <%= row[:key] %>
           </dt>
           <dd class="govuk-summary-list__value">
-            <% if row[:DANGEROUS_html_value].nil? %>
-              <%= row[:value] %>
+            <% if row[:value].is_a?(Array) %>
+              <% row[:value].each do |value| %><%= value %><br><% end %>
             <% else %>
-              <%= render inline: row[:DANGEROUS_html_value] %>
+              <%= row[:value] %>
             <% end %>
           </dd>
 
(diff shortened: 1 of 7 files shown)


app/components/summary_card_component.html.erb
```diff
@@ -8,10 +8,10 @@
             <%= row[:key] %>
           </dt>
           <dd class="govuk-summary-list__value">
-            <% if row[:DANGEROUS_html_value].nil? %>
-              <%= row[:value] %>
+            <% if row[:value].is_a?(Array) %>
+              <% row[:value].each do |value| %><%= value %><br><% end %>
```
duncanjbrown: :+1:

PR #457, opened 2019-11-07
Quickfix - Handle the case in which provider 'ABC' exists

stevehook:
`GenerateTestData` hits a uniqueness constraint if you have an existing provider with code `ABC` and a different name.

### Context

This is just a quickfix to address an issue that we found in development.

If a `Provider` with `code` `ABC` exists use that regardless of it's original `name`.

### Guidance to review

- Does the test adequately describe the problem?
- Does it matter if the test provider is given a different name? i.e. is there any code that looks it up by name? (I didn't find any).

### Link to Trello card

No card for this one

### Env vars

No

### DIFF
--- a/app/services/generate_test_data.rb
+++ b/app/services/generate_test_data.rb
@@ -70,10 +70,9 @@ private
   end
 
   def fake_provider
-    Provider.find_or_create_by(
-      name: 'Example Training Provider',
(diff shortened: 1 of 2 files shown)


tijmenb:
Merging to get the number of open PRs down.

PR #470, opened 2019-11-08
Add autofill when adding other qualifications

chubberlisk:
### Context

It's likely that candidates will need to add a number of other qualifications.

This PR adds autofill for `Type of Qualification`, `Institution where you studied` and `Year qualification was awarded` from their last created other relevant qualification when adding a new one.

### Screenshots

image

image

### Guidance to review

Smol edits had to be made on some specs as ordering had changed.

### Link to Trello card

[204 - Adding other qualifications](https://trello.com/c/qU3R6EVL/204-adding-other-qualifications)

### DIFF
--- a/app/controllers/candidate_interface/other_qualifications/base_controller.rb
+++ b/app/controllers/candidate_interface/other_qualifications/base_controller.rb
@@ -1,7 +1,17 @@
 module CandidateInterface
   class OtherQualifications::BaseController < CandidateInterfaceController
     def new
-      @qualification = OtherQualificationForm.new
+      qualifications = OtherQualificationForm.build_all_from_application(current_application)
+
+      @qualification = if qualifications.any?
+                         OtherQualificationForm.new(
+                           qualification_type: qualifications.last.qualification_type,
+                           institution_name: qualifications.last.institution_name,
+                           award_year: qualifications.last.award_year,
+                         )
+                       else
+                         OtherQualificationForm.new
+                       end
     end
 
     def create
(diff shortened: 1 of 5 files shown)


spec/models/candidate_interface/other_qualification_form_spec.rb
```diff
@@ -57,6 +57,7 @@
   describe '.build_all_from_application' do
     let(:application_form) do
       create(:application_form) do |form|
+        form.application_qualifications.create(level: 'other')
```
stevehook: Just wondering what happens if there is a dead heat for `created_at` between this and the one below. It could be avoided by explicitly specifying different `created_at` values.

app/controllers/candidate_interface/other_qualifications/base_controller.rb
```diff
@@ -1,7 +1,17 @@
...
     def new
-      @qualification = OtherQualificationForm.new
+      qualifications = OtherQualificationForm.build_all_from_application(current_application)
+
+      @qualification = if qualifications.any?
+                         OtherQualificationForm.new(
+                           qualification_type: qualifications.last.qualification_type,
+                           institution_name: qualifications.last.institution_name,
+                           award_year: qualifications.last.award_year,
+                         )
```
stevehook: I'm guessing you can't just do `@qualification = qualifications.last` because you want `@qualification` to have empty values for the attributes you are not specifying?

spec/models/candidate_interface/other_qualification_form_spec.rb
chubberlisk: Ah yes! I thought I did that, but between things and my brain not working very well I hadn't done this. Will update, thank you!

app/controllers/candidate_interface/other_qualifications/base_controller.rb
chubberlisk: Yup! We only want to autofill the specified attributes as per the prototype which are type, institution and year.

spec/models/candidate_interface/other_qualification_form_spec.rb
chubberlisk: Updated now as you suggested, with specifying `created_at` values.

PR #471, opened 2019-11-08
Provider Receive Application business rules

stevehook:
### Context

After a candidate has submitted an application the business rules state that two things need to happen before a provider can see the application:

- Two references need to be received
- 5 working days must elapse after application submission

Visibility of the application is controlled by it's state, three states are important here:

- `awaiting_references` is the state immediately after submission
- `application_complete` indicates that references have been received but 5 working days have not yet elapsed
- `awaiting_provider_decision` indicates that both above conditions have been met and that the provider can now see the application

It is possible for an application to go straight from `awaiting_references` to `awaiting_provider_decision` if references are received after 5 working days.

- [x] Cucumber scenario to specify the business rules above
- [x] Calculation and storage of the time at which the 5 working days elapse for each application choice at time of submission. Stored on `ApplicationChoice#edit_by` and calculated ahead of time rather than on the fly so that we can easily find applications that have passed the given date later. Note that setting `ApplicationChoice#edit_by` depends on using `SubmitApplication` service class (it's no good just setting status manually).
- [x] 'Service' class to check for applications that need to be progressed to `awaiting_provider_decision`. This one will be called by Clockwork (in a separate PR).
- [x] Changes to `ReceiveReferences` to factor logic out of state machine and into this service class. Also enforce the 2 reference minimum.

### Guidance to review

- There is an extra cucumber scenario, does it make sense? Are we all happy to use cucumber to specify the high-level business rule?
- Do the changes to `ReceiveReference` make sense?
- Are we happy that we have to use the service classes, e.g. `ReceiveReference` and `SubmitApplication` to get applications into the correct state?

### Link to Trello card

[1202 - Add Provider_Receive_Application](https://trello.com/c/oJ8RQ9mQ/1202-add-providerreceiveapplication)

### Env vars

No new env vars

### DIFF
--- a/app/models/application_form.rb
+++ b/app/models/application_form.rb
@@ -8,12 +8,14 @@ class ApplicationForm < ApplicationRecord
   has_many :application_qualifications
   has_many :references
 
+  MINIMUM_REFERENCES = 2
+
   def submitted?
     application_choices.any? && !application_choices.first.unsubmitted?
   end
 
   def references_complete?
-    references.any? && references.all?(&:complete?)
+    references.select(&:complete?).count >= MINIMUM_REFERENCES
   end
 
   def qualification_in_subject(level, subject)
--- a/app/models/application_form.rb
+++ b/app/models/application_form.rb
@@ -8,14 +8,14 @@ class ApplicationForm < ApplicationRecord
   has_many :application_qualifications
   has_many :references
 
-  MINIMUM_REFERENCES = 2
+  MINIMUM_COMPLETE_REFERENCES = 2
 
   def submitted?
     application_choices.any? && !application_choices.first.unsubmitted?
   end
 
   def references_complete?
-    references.select(&:complete?).count >= MINIMUM_REFERENCES
+    references.select(&:complete?).count == MINIMUM_COMPLETE_REFERENCES
   end
 
   def qualification_in_subject(level, subject)
--- a/app/services/get_application_choices_ready_to_send_to_provider.rb
+++ b/app/services/get_application_choices_ready_to_send_to_provider.rb
@@ -1,10 +1,20 @@
 class GetApplicationChoicesReadyToSendToProvider
   def self.call
-    ApplicationChoice
+    application_choices_past_edit_by(
+      application_choices_with_references_complete(
+        ApplicationChoice.where(status: :application_complete),
+      ),
+    )
+  end
+
+  def self.application_choices_past_edit_by(scope)
+    scope.where('edit_by < ?', Time.zone.now)
+  end
+
+  def self.application_choices_with_references_complete(scope)
+    scope
       .joins(application_form: :references)
       .where('feedback is not null')
-      .where('edit_by < ?', Time.zone.now)
-      .where(status: :application_complete)
       .group('application_choices.id')
       .having('count("references"."feedback") >= 2')
   end
--- a/app/services/get_application_choices_ready_to_send_to_provider.rb
+++ b/app/services/get_application_choices_ready_to_send_to_provider.rb
@@ -1,0 +1,11 @@
+class GetApplicationChoicesReadyToSendToProvider
+  def self.call
+    ApplicationChoice
+      .joins(application_form: :references)
+      .where('feedback is not null')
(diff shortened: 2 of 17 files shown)


features/submitting_references.feature
```diff
@@ -17,3 +17,16 @@ Feature: references
...
+
+  Scenario: an application is sent to a provider after it's received two references and 5 working days elapse after submission
+    Given an application choice has "unsubmitted" status
+    And the candidate has specified "j.moriarty@uni.ac.uk" and 's.skinner@springfield-elementary.edu' as referees
+    And the date is "2019-11-04"
+    And the candidate takes action "submit"
+    And "j.moriarty@uni.ac.uk" provides a reference
+    And "s.skinner@springfield-elementary.edu" provides a reference
+    Then the new application choice status is "application_complete"
+    When the date is "2019-11-11"
```
duncanjbrown: Is it worth adding a line in here like `And the daily application cron has run`, to trigger it? Not sure how granular to be in the description

spec/services/send_applications_to_provider_spec.rb
```diff
@@ -0,0 +1,13 @@
+RSpec.describe SendApplicationsToProvider do
```
duncanjbrown: Could this spec be a cucumber table? Or perhaps delegate the business-rules bit to cucumber, and test whatever else it does assuming it's got an application it definitely ought to send — maybe there's a case for making this a `SendApplicationToProvider`, having it always send what it's given, and pulling the knowledge of which applications to send up into the caller?

spec/services/submit_application_spec.rb
```diff
@@ -6,14 +6,21 @@
       application_form = create(:application_form)
       create(:application_choice, application_form: application_form, status: 'unsubmitted')
       create(:application_choice, application_form: application_form, status: 'unsubmitted')
+      application_dates = instance_double(
+        ApplicationDates,
+        edit_by: Time.zone.local(2019, 11, 11, 11, 59, 0),
+      )
+      allow(ApplicationDates).to receive(:new).and_return(application_dates)
```
duncanjbrown: hm. Could we avoid this stub somehow?

app/models/application_form.rb
```diff
@@ -8,12 +8,14 @@ class ApplicationForm < ApplicationRecord
...
 
+  MINIMUM_REFERENCES = 2
+
   def submitted?
     application_choices.any? && !application_choices.first.unsubmitted?
   end
 
   def references_complete?
-    references.any? && references.all?(&:complete?)
+    references.select(&:complete?).count >= MINIMUM_REFERENCES
```
tijmenb: As discussed in the meeting, candidates will need exactly 2 references.

app/services/get_application_choices_ready_to_send_to_provider.rb
```diff
@@ -0,0 +1,21 @@
+class GetApplicationChoicesReadyToSendToProvider
+  def self.call
+    application_choices_past_edit_by(
```
tijmenb: Personal preference, I like it more procedural:

```rb
scope = ApplicationChoice.where(status: :application_complete)
scope = application_choices_past_edit_by(scope)
scope = application_choices_with_references_complete(scope)
```

WDYT?

app/services/send_applications_to_provider.rb
```diff
@@ -0,0 +1,7 @@
+class SendApplicationsToProvider
```
tijmenb: Can you add a comment here explaining that it's a clockwork job that calls this class periodically?

app/services/get_application_choices_ready_to_send_to_provider.rb
stevehook: Yes I think that's better, if it were Elixir they have a nice syntax for this :)

app/models/application_form.rb
stevehook: Have also added a max length validation for `ApplicationForm#references`

stevehook:
@tijmenb & @duncanjbrown this one is ready for re-review. I added some small changes to integrate Paul's CSV import.

PR #474, opened 2019-11-08
Add terms of use (candidate) page

paulrobertlloyd:
### Context

Adding ~~terms and conditions~~ terms of use content page, linked to from the footer.

In addition to the new page and associated routes and translations:

* updates `app-styled-content` class to cover more HTML that can be generated by markdown, ensures list items are rendered correctly, and renders headings one size larger to account for the addition of another heading level in the terms and conditions content.
* uses the shorter `<name@address.example>` markdown syntax for email addresses in the accessibility statement

### Guidance to review

Visit ~~<http://localhost:3000/terms-conditions>~~ <http://localhost:3000/candidate/terms-of-use>

### Link to Trello card

[322 - Terms and conditions](https://trello.com/c/lt7ielt3/322-terms-and-conditions)


paulrobertlloyd:
Some last-minute changes to content needing to be made before this is ready to merge.

app/views/layouts/application.html.erb
```diff
@@ -96,6 +96,9 @@
               <li class="govuk-footer__inline-list-item">
                 <%= link_to t('layout.accessibility'), accessibility_path, class: 'govuk-footer__link' %>
               </li>
+              <li class="govuk-footer__inline-list-item">
+                <%= link_to t('layout.terms_conditions'), terms_conditions_path, class: 'govuk-footer__link' %>
```
tijmenb: I *think* these T&Cs only apply to the candidate facing bit, the providers apparently have different terms: https://trello.com/c/4zjNLeCu/1148-tcs

paulrobertlloyd: Ah! I didn’t realise providers where using the same template (or app, in fact). Should provider and candidate views use a different template (or something like that)?

paulrobertlloyd:
@tijmenb Not sure if this is the best resolution, but have updated route to use `/candidate/terms-of-use`. I tried moving this into `:candidate_interface` namespace, but that started causing follow-on issues with the content controller. Perhaps requires a separate PR to address having different candidate/provider content pages?

config/routes.rb
```diff
@@ -11,9 +11,11 @@
   root to: redirect('/candidate')
 
   get '/accessibility', to: 'content#accessibility'
+  get '/candidate/terms-of-use', to: 'content#terms_candidate', as: :terms_candidate
```
tijmenb: We'll need to update the link in the signup screen as well.

app/views/layouts/application.html.erb
```diff
@@ -96,6 +96,9 @@
               <li class="govuk-footer__inline-list-item">
                 <%= link_to t('layout.accessibility'), accessibility_path, class: 'govuk-footer__link' %>
               </li>
+              <li class="govuk-footer__inline-list-item">
```
tijmenb: Let's figure something out on Monday about splitting up the layouts for candidates/providers.

config/routes.rb
paulrobertlloyd: Have updated 👍🏻

PR #479, opened 2019-11-11
Return applications for accrediting providers

RobertM905:
### Context

Providers need to be able to retrieve applications for courses that they accredit.

- Update `courses` table to include `accrediting_provider `relationship.
- Update `course` and `application_choice` model to include `accrediting_provider`.
- Update `get_application_choices_for_provider` service to allow a given provider to retrieve their applications for courses they are the main provider and accrediting provider for.
- Update `application_choice` controller `current_user` stub to use the `Provider` model.

Updates to Syncing data from find:
- Update `sync_providers_from_find` service to allow the creation of `accrediting_providers`.
- Update current and create new service specs to handle the retrieval and creation of `accrediting_providers`.
- Update `support_interface` provider system specs to handle above changes.

### Guidance to review

Run spec and ensure the correct application_choice are being retrieved. 

### Link to Trello card

[1194 - Return application for accrediting providers](https://trello.com/c/2j6Lt55j/1194-return-applications-from-multiple-organisations%F0%9F%8F%88-kick-off)

### Env vars

N\A

### DIFF
--- a/app/controllers/provider_interface/application_choices_controller.rb
+++ b/app/controllers/provider_interface/application_choices_controller.rb
@@ -22,8 +22,8 @@ module ProviderInterface
     # by a proper ProviderUser when implementing Signin.
     def current_user
       fake_user_class = Struct.new(:provider)
-      fake_provider_class = Struct.new(:code)
-      fake_user_class.new(fake_provider_class.new('ABC'))
+      fake_provider = Provider.find_or_create_by(code: 'ABC')
+      fake_user_class.new(fake_provider)
     end
   end
 end
--- a/app/controllers/provider_interface/application_choices_controller.rb
+++ b/app/controllers/provider_interface/application_choices_controller.rb
@@ -22,7 +22,7 @@ module ProviderInterface
     # by a proper ProviderUser when implementing Signin.
     def current_user
       fake_user_class = Struct.new(:provider)
-      fake_provider = Provider.find_or_create_by(code: 'ABC')
+      fake_provider = Provider.find_by(code: 'ABC')
       fake_user_class.new(fake_provider)
     end
   end
--- a/db/migrate/20191107110337_add_accrediting_provider_to_courses.rb
+++ b/db/migrate/20191107110337_add_accrediting_provider_to_courses.rb
@@ -1,6 +1,5 @@
 class AddAccreditingProviderToCourses < ActiveRecord::Migration[6.0]
   def change
     add_column :courses, :accrediting_provider_id, :integer
-    add_index :courses, %i[accrediting_provider_id code], unique: true
   end
 end
--- a/db/migrate/20191107110337_add_accrediting_provider_to_courses.rb
+++ b/db/migrate/20191107110337_add_accrediting_provider_to_courses.rb
@@ -1,0 +1,6 @@
+class AddAccreditingProviderToCourses < ActiveRecord::Migration[6.0]
+  def change
+    add_column :courses, :accrediting_provider_id, :integer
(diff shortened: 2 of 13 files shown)


spec/services/get_application_choices_for_provider_spec.rb
```diff
@@ -59,4 +59,39 @@
...
+
+    create_list(
+      :application_choice,
+      4,
+      course_option: course_option_for_provider(provider: alternate_provider),
+      status: 'awaiting_provider_decision',
+    )
+
+    returned_applications = GetApplicationChoicesForProvider.call(provider: current_provider)
+    expect(returned_applications.size).to be(3)
```
duncanjbrown: Would it be tidy to assert on the names explicitly set in the setup?

RobertM905: Updated the service spec to reflect the candidate names associated with the applications returned. 
Also added a system spec to test returning applications from the perspective of the provider UI.

app/controllers/provider_interface/application_choices_controller.rb
```diff
@@ -22,8 +22,8 @@ def show
     # by a proper ProviderUser when implementing Signin.
     def current_user
       fake_user_class = Struct.new(:provider)
-      fake_provider_class = Struct.new(:code)
-      fake_user_class.new(fake_provider_class.new('ABC'))
+      fake_provider = Provider.find_or_create_by(code: 'ABC')
```
tijmenb: I think shouldn't create test data like this anymore now that we're "live", since it could pollute the production database.

db/migrate/20191107110337_add_accrediting_provider_to_courses.rb
```diff
@@ -0,0 +1,6 @@
+class AddAccreditingProviderToCourses < ActiveRecord::Migration[6.0]
+  def change
+    add_column :courses, :accrediting_provider_id, :integer
+    add_index :courses, %i[accrediting_provider_id code], unique: true
```
tijmenb: Why do we need this index? Isn't `code` already unique?

RobertM905: Yup that's right. I must of been on autopilot 👍

app/controllers/provider_interface/application_choices_controller.rb
RobertM905: I feel like test stubbing should not be in this controller at all and we need a nicer what to stub out the current_user in tests.

RobertM905: Updated the stub to only find and not create.  
here: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/479/commits/77c895fac48fc37fc154ac653b0918f6061690ed

I would like to move that stub out of the controller, I can talk to @duncanjbrown about his current PR around DFE Sign In https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/507

duncanjbrown: Propose keeping the change `find_by` for now, so it works in development and breaks in prod, and we'll have Sign-in fixed up in the near future? Will all be behind basic auth anyway

duncanjbrown: (Sign in vars are not configured on prod rn, so if we deployed #507 the effect of clicking "Sign in with DfE Sign-in" is 💥 )

RobertM905: Sounds like a good approach for now. 👍

PR #480, opened 2019-11-11
Add training with a disability

aldavidson:
### Context

Candidates need to have the option to tell providers about any disabilities they may have, so that they can make sure any specific needs they have can be catered for.

Add UI for entering disability info into the application form

Before:
Screen Shot 2019-11-11 at 13 27 09

After:

Screen Shot 2019-11-11 at 13 29 29
Screen Shot 2019-11-11 at 13 29 52

... plus more UI not screenshotted here, for review answers, etc

### Guidance to review

Start (or continue) an application, then from "Your application", click on "Training with a disability"

### Link to Trello card

[324 - Training with a disability](https://trello.com/c/cEV7CtIw/324-training-with-a-disability)

### Env vars

N/A


spec/factories.rb
```diff
@@ -17,6 +17,7 @@
       other_language_details { Faker::Lorem.paragraph_by_chars(number: 200) }
       further_information { Faker::Lorem.paragraph_by_chars(number: 300) }
       uk_residency_status { 'I have the right to study and/or work in the UK' }
+      disclose_disability { true }
```
tijmenb: To avoid tests relying on static data in the factories:
 
```suggestion
      disclose_disability { [true, false].sample }
```

spec/system/candidate_interface/candidate_entering_disability_info.rb
```diff
@@ -0,0 +1,94 @@
+require 'rails_helper'
+
+RSpec.feature 'Entering their disability information' do
+  include CandidateHelper
+
+  scenario 'Candidate submits their disability information' do
+    given_i_am_not_signed_in
+    and_i_visit_the_training_with_a_disability_page
+    then_i_should_see_the_homepage
```
tijmenb: Perhaps we can skip the 3 steps above, the auth is tested in a couple of places already.

app/models/candidate_interface/training_with_a_disability_form.rb
```diff
@@ -0,0 +1,32 @@
...
+      new(
+        disclose_disability: application_form.disclose_disability,
+        disability_disclosure: application_form.disability_disclosure,
+      )
+    end
+
+    def save(application_form)
+      return false unless valid?
+
+      # explicitly null-out the text field if the user said 'No'
```
tijmenb: Perhaps we can avoid some of the confusion here by using different values in the form. 

If we make the form values `yes` and `no`, we explicitly translate them to true/false. I think this is how it works in the personal details form:

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/a22b836482ad2ba579f0711c645d796c815450db/app/models/candidate_interface/personal_details_form.rb#L33

aldavidson: we still have to explicitly handle `nil`, and there's an interdependence between the two values - but I've refactored as you suggested, and it does make it clearer 👍

PR #481, opened 2019-11-11
Import references from CSV

fofr:
### Context

- Importing reference from Google Form responses
- Don't import existing references
- Only import if a reference ID and email match
- Show success or failure for each row in the CSV with a reason if it errored

Screen Shot 2019-11-11 at 16 23 12

Screen Shot 2019-11-11 at 15 35 34

### Guidance to review

First time working on this app in anger, please point out how to do things in a way more consistent with the application if you can.

This is a temporary feature that will be removed once references themselves have been built.

I will need help with the system spec.

[Trello](https://trello.com/c/JcBL1Ejh/1249-epic-reference-thing)

### DIFF
--- a/app/services/import_references_from_csv.rb
+++ b/app/services/import_references_from_csv.rb
@@ -1,9 +1,9 @@
 require 'csv'
 
 class ImportReferencesFromCsv
-  def self.call
+  def self.call(csv_file:)
     outcomes = []
-    CSV.foreach('references.csv') do |row|
+    CSV.foreach(csv_file) do |row|
       # First column in the headers row of an exported Google Form is always Timestamp
       next if row[0] == 'Timestamp'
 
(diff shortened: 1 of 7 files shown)


app/services/import_references_from_csv.rb
```diff
@@ -0,0 +1,59 @@
+require 'csv'
+
+class ImportReferencesFromCsv
+  def self.call(csv_file:)
+    outcomes = []
+    CSV.foreach(csv_file) do |row|
```
duncanjbrown: This code might be more readable using the `headers: true` option for `CSV.open` (https://stackoverflow.com/a/3717622), which means there's no need for numeric indexes in `process_row`.

fofr: @duncanjbrown I started with headers, but switched. There's risk in using headers given the way Google Forms spreadsheets work. The labels for the columns might change. The order of the columns will not (new fields get appended to the end). The column headings can also get quite long.

app/services/import_references_from_csv.rb
```diff
@@ -0,0 +1,59 @@
+require 'csv'
+
+class ImportReferencesFromCsv
```
tijmenb: Can you put this into a `SupportInterface` namespace?

app/views/support_interface/import_references/index.html.erb
```diff
@@ -0,0 +1,39 @@
...
+<% end %>
+
+<h1 class='govuk-heading-xl'>Import references</h1>
+
+<%= form_tag support_interface_import_references_path, method: :post, multipart: true do %>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <p class="govuk-body">
+        You’ll need to <a href="https://docs.google.com/spreadsheets/d/18Lr1SWD20o_sPbyj7gsTwkaJO9jwL-YZTTdQfQOuIiw/export?format=csv" class="govuk-link" rel="download">download a CSV of the latest reference responses</a> from Google Docs before you can upload it here.</p>
+
```
tijmenb: Perhaps we explain to the user here what the importing of the CSV will do? Something that says that we'll add feedback to the references, change the state, etc etc.

app/services/import_references_from_csv.rb
fofr: Update: Our services aren't currently namespaced. Leaving this as-is for now.

PR #487, opened 2019-11-12
Allow candidates to specify a "other" UK qualification type

alexdesi:
### Context

This PR allows specifying a custom qualification type from UK.

- the qualification type are store in the field `qualification_type` (possible values GCSE, GCE_O_LEVEL, SCOTTISH_HIGHER, OTHER_UK)

- when the qualification type is OTHER_UK, then the new field `other_uk_qualification_type` will be populated with the custom qualification, for example, Scottish Baccalaureate, CSE, IGCSE.

### Guidance to review
See the test: Candidate specifies GCSE maths with "Other UK qualification" type

### Link to Trello card

https://trello.com/c/44joKRvx/306-other-uk-qualifications-adding-maths-english-and-science-gcse-information


db/schema.rb
```diff
@@ -57,25 +57,26 @@
     t.text "other_language_details"
     t.date "date_of_birth"
     t.text "further_information"
+    t.datetime "submitted_at"
+    t.string "support_reference", limit: 10
+    t.string "reference", limit: 10
```
tijmenb: I think this doesn't belong here?

alexdesi: thanks, I'll check it!

alexdesi:
👋 @tijmenb I've addressed your comment, this should be good now.

PR #493, opened 2019-11-12
Add checkbox for course confirmation

dankmitchell:
### Context
Course selection UX

Add confirmation checkbox to complete course selection

### Guidance to review
- Add courses
- Check completed checkbox

### Link to Trello card
[331 - Allow the course selection section to be marked complete and enable submission](https://trello.com/c/Ttl1Wr5Z/331-allow-the-course-selection-section-to-be-marked-complete-and-enable-submission)

### After
localhost_3000_candidate_application_courses(Laptop with MDPI screen)

### DIFF
--- a/app/controllers/candidate_interface/course_choices_controller.rb
+++ b/app/controllers/candidate_interface/course_choices_controller.rb
@@ -1,6 +1,7 @@
 module CandidateInterface
   class CourseChoicesController < CandidateInterfaceController
     def index
+      @application_form = current_application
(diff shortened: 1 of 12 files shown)


tijmenb:
This is excellent! The only thing that is missing (might be missing from the design even) is that we should show something different when there's no course chosen yet:

Screenshot 2019-11-13 at 09 58 49

- "Do you want to add another course?" won't make sense without courses
- We shouldn't show the checkbox

PR #494, opened 2019-11-12
Add dynamic validation for qualification grade

alexdesi:
### Context

The Grades for validation need to respect the format specified in the doc:
https://docs.google.com/document/d/1NMUgKmzPSWJn85CVCUjkeArGuy1_WwlVeFZ9QNy6AeA/edit

NOTE: the logic in the doc is complex, for this MVP I **relaxed** the regular expressions:
it tests all the allowed letters/numbers but it does not consider the order.
Atm this tradeoff sounded reasonable to me, due the time restrictions.

Add validations that consider the qualification type. 

### Guidance to review
see the screenshot and the new specs 

### Link to Trello card
https://trello.com/c/hvL9GGWZ/340-verifying-gcses-and-equivalents

### DIFF
--- a/app/models/candidate_interface/gcse_qualification_details_form.rb
+++ b/app/models/candidate_interface/gcse_qualification_details_form.rb
@@ -5,6 +5,7 @@ module CandidateInterface
     attr_accessor :grade, :award_year, :qualification_id
     validates :grade, :award_year, :qualification_id, presence: true
     validate :validate_award_year, if: :award_year
+    validate :validate_grade_format
 
     def self.build_from_qualification(qualification)
       new(
@@ -17,16 +18,39 @@ module CandidateInterface
     def save_base
       return false unless valid?
 
-      qualification = ApplicationQualification.find(qualification_id)
-
       qualification.update(grade: grade, award_year: award_year)
     end
 
   private
-
     def validate_award_year
       valid_award_year = award_year.match(/^[1-9]\d{3}$/)
       errors.add(:award_year, :invalid) unless valid_award_year
+    end
+
+    def validate_grade_format
+      return if new_record? || qualification.qualification_type.nil?
+
+      qualification_rexp = invalid_grades[qualification.qualification_type.to_sym]
+
+      if qualification_rexp && grade.match(qualification_rexp)
+        errors.add(:grade, :invalid)
+      end
+    end
+
+    def qualification
+      @qualification ||= ApplicationQualification.find(qualification_id)
+    end
+
+    def invalid_grades
+      {
+        gcse: /[^1-9A-GU\*]/i,
+        gce_o_level: /[^A-EU]/i,
+        scottish_higher: /[^A-D1-7]/i,
+      }
+    end
+
+    def new_record?
+      qualification_id.nil?
     end
   end
 end
--- a/app/models/candidate_interface/gcse_qualification_details_form.rb
+++ b/app/models/candidate_interface/gcse_qualification_details_form.rb
@@ -2,17 +2,19 @@ module CandidateInterface
   class GcseQualificationDetailsForm
     include ActiveModel::Model
 
-    attr_accessor :grade, :award_year, :qualification_id
-    validates :grade, :award_year, :qualification_id, presence: true
+
+    attr_accessor :grade, :award_year, :qualification
+    validates :grade, :award_yearpresence: true
     validate :validate_award_year, if: :award_year
+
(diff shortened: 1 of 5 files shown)


app/models/candidate_interface/gcse_qualification_details_form.rb
```diff
@@ -17,16 +18,39 @@ def self.build_from_qualification(qualification)
...
       errors.add(:award_year, :invalid) unless valid_award_year
     end
+
+    def validate_grade_format
+      return if new_record? || qualification.qualification_type.nil?
+
+      qualification_rexp = invalid_grades[qualification.qualification_type.to_sym]
+
+      if qualification_rexp && grade.match(qualification_rexp)
+        errors.add(:grade, :invalid)
```
benilovj: Can we please add logging here that would:

- allow us to see which combination of qualification type and grade triggered the error
- allow us to easily grep for it in the logs

app/models/candidate_interface/gcse_qualification_details_form.rb
```diff
@@ -17,16 +18,39 @@ def self.build_from_qualification(qualification)
...
+
+    def qualification
+      @qualification ||= ApplicationQualification.find(qualification_id)
+    end
+
+    def invalid_grades
+      {
+        gcse: /[^1-9A-GU\*]/i,
+        gce_o_level: /[^A-EU]/i,
+        scottish_higher: /[^A-D1-7]/i,
```
benilovj: in light of https://trello.com/c/Pb9QRUPv/345-change-scottish-higher-to-scottish-national, this line will need an update

app/models/candidate_interface/gcse_qualification_details_form.rb
```diff
@@ -17,16 +18,39 @@ def self.build_from_qualification(qualification)
...
+      end
+    end
+
+    def qualification
+      @qualification ||= ApplicationQualification.find(qualification_id)
+    end
+
+    def invalid_grades
+      {
+        gcse: /[^1-9A-GU\*]/i,
```
duncanjbrown: Wouldn't this permit `B*`, `U*` etc?

alexdesi: that's true, I'll see if this fix is not too complex to implement.

app/models/candidate_interface/gcse_qualification_details_form.rb
tijmenb: Let's pull this out into a separate card, we'd want a generic solution: https://trello.com/c/zlfQ2il2/358-keep-track-of-validation-errors-in-the-application-form

app/models/candidate_interface/gcse_qualification_details_form.rb
alexdesi: Here is the PR for that: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/516

spec/models/candidate_interface/gcse_qualification_details_form_spec.rb
```diff
@@ -4,6 +4,30 @@
   describe 'validations' do
     it { is_expected.to validate_presence_of(:grade) }
     it { is_expected.to validate_presence_of(:award_year) }
+
+    describe 'grade format' do
```
tijmenb: We should add a test for each of the different regexes. At the mimimum test that [Juan's examples from the document](https://docs.google.com/document/d/1NMUgKmzPSWJn85CVCUjkeArGuy1_WwlVeFZ9QNy6AeA/edit) pass.

app/models/candidate_interface/gcse_qualification_details_form.rb
alexdesi: That's right, anyway, our first implementation of the validation quite permissive.

Once we have a system to log the wrong attempts (described in https://trello.com/c/zlfQ2il2/358-keep-track-of-validation-errors-in-the-application-form) we'll be able to be more precise on the validations.

alexdesi:
@tijmenb I've just updated this!
As discussed, this PR does not include validation for the option "Other UK qualification", it's not clear what the validation should be for that field.
Let's postpone it, I'll create a trello for it.

spec/models/candidate_interface/gcse_qualification_details_form_spec.rb
alexdesi: Done!

PR #499, opened 2019-11-13
Referee emailer

fofr:
### Context

- Add mailer and template for requesting references from referees.

Personalise email by:
- including candidate name
- generating a link to a prefilled Google form

Referee name is missing at the moment, it's being added to the reference model in a separate PR by @aldavidson.

It’s not tied to any application state changes, I think for now we’d want to manually trigger them. But this means we don’t need to be copying and pasting long Google form URLs into Zendesk, plus other fields, and there’s less chance for human error (like confusing referee name and candidate name).

Based on the existing candidate_mailer.

https://trello.com/c/JcBL1Ejh/1249-build-system-to-request-and-receive-references

### DIFF
--- a/app/mailers/referee_mailer.rb
+++ b/app/mailers/referee_mailer.rb
@@ -1,0 +1,10 @@
+class RefereeMailer < ApplicationMailer
+  def reference_request_email(application_form, reference)
+    @application_form = application_form
+    @reference = reference
+
+    view_mail(GENERIC_NOTIFY_TEMPLATE,
+              to: reference.email_address,
+              subject: t('reference_request.email.subject'))
+  end
+end
--- a/app/mailers/referee_mailer.rb
+++ b/app/mailers/referee_mailer.rb
@@ -9,11 +9,20 @@ class RefereeMailer < ApplicationMailer
               subject: t('reference_request.email.subject', candidate_name: @candidate_name))
   end
 
+private
+
   def google_form_url_for(candidate_name, reference)
-    t('reference_request.google_form_url') + '?' +
+    # `to_query` replaces spaces with `+`, but a Google Form with a prefilled parameter
+    # shows a `+` in the actual form, eg "Jane Doe" becomes "Jane+Doe", so we need to
+    # switch them to %20 without stripping out a possible `+` in an email address
+    t('reference_request.google_form_url') +
+      '?' +
       {
         t('reference_request.email_entry') => reference.email_address,
         t('reference_request.reference_id_entry') => reference.id,
+      }.to_query +
+      '&' +
+      {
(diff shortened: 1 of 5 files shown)


app/mailers/referee_mailer.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def google_form_url_for(candidate_name, reference)
+    t('reference_request.google_form_url') + '?' +
+      {
+        t('reference_request.email_entry') => reference.email_address,
+        t('reference_request.reference_id_entry') => reference.id,
+        t('reference_request.candidate_name_entry') => candidate_name,
+        t('reference_request.referee_name_entry') => 'TODO: Referee name',
+      }.to_query.gsub('+', '%20')
```
duncanjbrown: Any reason not to use `URI.escape`? https://ruby-doc.org/stdlib-2.5.1/libdoc/uri/rdoc/URI/Escape.html#method-i-encode

app/views/referee_mailer/reference_request_email.text.erb
```diff
@@ -0,0 +1,24 @@
...
+
+If we don’t receive your reference within 10 working days from the date of this email, we’ll ask <%= @candidate_name %> to suggest another referee.
+
+# Privacy notice
+
+<%= @candidate_name %> put us in touch with you so that you can provide a reference for them.
+
+The Department for Education and the teacher training provider(s) the candidate has applied to have a ‘legitimate interest’ under data protection legislation to hold your data for the period needed to process the candidate’s application.
+
+We’ll only use your data for this purpose and will delete it when no longer needed for this purpose.
```
duncanjbrown: Maybe the second "for this purpose" is redundant?

app/views/referee_mailer/reference_request_email.text.erb
```diff
@@ -0,0 +1,24 @@
...
+
+We currently use secure G Suite tools to process reference data. Your data will only be accessed by the staff who need it to process reference data.
```
duncanjbrown: Is "G suite" a bit jargony? Do we need that first sentence?

fofr: Agree. This is based on a first draft by Emma F and I'll feedback.

app/mailers/referee_mailer.rb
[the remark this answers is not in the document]
fofr: Done in a8cc2ab

app/views/referee_mailer/reference_request_email.text.erb
fofr: Section removed in ccc8ad282c5e3b3ae2f9c11e14953929d66e7381

PR #500, opened 2019-11-13
Fix the broken build: specs and spec reporting failures

duncanjbrown:
This code relied on an interface which has changed.

At the same time as correcting the call to ReceiveReference, roll in
the service now responsible for publishing applications to providers in
anticipation for simply invoking the cron job which will cause the
transition

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,6 @@ ci.test: ## Run the tests with results formatted for CI
 	docker-compose run web /bin/sh -c 'mkdir $(RESULTS_PATH) && \
 		apk add nodejs yarn && \
 		bundle exec rails assets:precompile && \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -52,6 +52,6 @@ ci.test: ## Run the tests with results formatted for CI
 	docker-compose run web /bin/sh -c 'mkdir $(RESULTS_PATH) && \
 		apk add nodejs yarn && \
 		bundle exec rails assets:precompile && \
-		bundle exec --verbose rspec --format RspecJunitFormatter --out $(RESULTS_PATH)/rspec-results.xml'
```
rizzkhan1: @duncanjbrown You build pipeline will fail as it relies on rspec-results.xml file - see line 99 `https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/azure-pipelines.yml`

PR #501, opened 2019-11-13
Integrate SendApplicationsToProvider with Sidekiq & Clockwork

stevehook:
### Context

This PR configures Clockwork to schedule a nightly job to update the status of applications that have passed their `edit_by` date are are therefore ready to 'send' to provider. It is a follow up to https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/471

- Convert the `SendApplicationsToProvider` service class to a Sidekiq job.
- Configure Clockwork to run the `SendApplicationsToProvider` job every day at 1:00AM (localtime).

Note that we don't actually _send_ anything to a provider yet, we are just updating the status of the `ApplicationChoice` to `awaiting_provider_decision`. Notifications are an upcoming feature.

### Guidance to review

- Does it make sense to run this job daily at 1:00AM? (Note that `edit_by` values are always set to the end of a day so expire at midnight)
- Is the Clockwork config correct?
- The Sidekiq job should retry if it fails. Is there anything else that could stop the job running correctly? If it doesn't run the provider will not see applications on time.

### Link to Trello card

[1202 - Add Provider_Receive_Application](https://trello.com/c/oJ8RQ9mQ/1202-add-providerreceiveapplication)

### Env vars

No new env vars


config/clock.rb
```diff
@@ -7,4 +7,5 @@ class Clock
   include Clockwork
 
   every(5.minutes, 'ClockworkCheck') { ClockworkCheck.perform_async }
+  every(1.day, 'SendApplicationsToProvider', at: '01:00') { SendApplicationsToProvider.perform_async }
```
mnacos: I know we are going for a persisted Redis on production, but my preference would be build some fault-tolerance into the system. How about we code this as if we don't trust that:
- clockwork will be running at exactly 01:00 hours to trigger the job (may be down)
- tasks may disappear from the Redis queue before they are picked up (e.g. sidekiq is down and Redis loses all data or we deliberately empty all queues

I'd recommend scheduling the task to run 3 times a day (for a daily task), while making sure things are processed only once (so adding some tracking to whether a task was performed or not).

tijmenb: Does it make sense to run the task every hour? 

It's already idempotent, so we don't have to worry about only-once I think.

tijmenb:
Would it make sense to trigger the job after all references have come in, so the application is pushed to the provider more quickly?

config/clock.rb
stevehook: If things work as they should then the only time when you should pick up any applications to progress is just after midnight because we round up `edit_by` to the end of the day. Once an hour would be the simplest thing to do I guess.

stevehook:
> Would it make sense to trigger the job after all references have come in, so the application is pushed to the provider more quickly?

Yes. In `ReceiveReference` we already handle the case in which `edit_by` expires before the references are all complete and make the state transition per application. https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/app/services/receive_reference.rb#L27-L28

app/services/send_applications_to_provider.rb
```diff
@@ -1,4 +1,4 @@
-# This service class will be called as a background job, scheduled to run nightly
+# This worker will be scheduled to run nightly
```
duncanjbrown: This comment is now out of date as it runs every hour 🙂

spec/workers/send_applications_to_provider_worker_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'rails_helper'
+
+RSpec.describe SendApplicationsToProviderWorker do
```
duncanjbrown: I question the value of this spec. It exposes the implementation details of the worker without testing either its return value or any side effects

stevehook: thats true, the clockwork spec covers the same ground so this one is pretty much redundant. I'll remove it unless you can think of anything more useful it could test given that all it does is delegate the work to a service class?

app/services/send_applications_to_provider.rb
stevehook: in theory each hourly run except the first one after midnight should be a no-op. The extra runs give us a bit of a safety net in case of Redis or Sidekiq being down.

spec/workers/send_applications_to_provider_worker_spec.rb
duncanjbrown: I think remove it. We also test this whenever we run a spec that invokes the cron

stevehook: it's gone!

PR #506, opened 2019-11-13
Do not use draft courses from Find 

tijmenb:
### Context

We currently [save all of the courses that Find returns](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/app/services/sync_provider_from_find.rb) in the database. It turns out the Find API also returns courses that are still in draft. For example, Gorse has a course "History", which is [returned by the API](https://api2.publish-teacher-training-courses.service.gov.uk/api/v3/recruitment_cycles/2020/providers/1N1/?include=sites,courses.sites) but is not available on the Find site.

As far as I can see, the `findable?` attribute on Find courses determines whether or not it's visible to normal users. This commit copies the `findable?` attribute to our local database, so that (later) we can make sure we only display courses that are also in Find.

I've also considered to stop creating courses that aren't `findable?` in the sync. This has the downside of not removing courses from the system that flip from `findable?` to not `findable?`. We would also have to remove the 5 draft courses that are currently in the Apply database.

On the support page for Gorse provider it looks like this:

image

### Guidance to review

Does this make sense?

### Link to Trello card

https://trello.com/c/96nQagv3

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -15,8 +15,6 @@ class Course < ApplicationRecord
     further_education: 'Further education',
   }, _suffix: :course
 
-  scope :that_candidates_can_apply_to, -> { where(exposed_in_find: true) }
-
   def name_and_code
     "#{name} (#{code})"
   end
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -15,6 +15,7 @@ class Course < ApplicationRecord
     further_education: 'Further education',
   }, _suffix: :course
 
+  scope :that_candidates_can_apply_to, -> { where(exposed_in_find: true) }
 
   def name_and_code
     "#{name} (#{code})"
--- a/app/services/sync_provider_from_find.rb
+++ b/app/services/sync_provider_from_find.rb
@@ -1,5 +1,9 @@
 class SyncProviderFromFind
   def self.call(provider_code:)
+    # Request all providers, courses and sites.
+    #
+    # For the full response, see:
+    # https://api2.publish-teacher-training-courses.service.gov.uk/api/v3/recruitment_cycles/2020/providers/1N1/?include=sites,courses.sites
     find_provider = FindAPI::Provider
       .current_cycle
       .includes(:sites, courses: [:sites])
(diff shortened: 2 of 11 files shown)


app/services/sync_provider_from_find.rb
```diff
@@ -1,5 +1,9 @@
 class SyncProviderFromFind
   def self.call(provider_code:)
+    # Request all providers, courses and sites.
+    #
+    # For the full response, see:
+    # https://api2.publish-teacher-training-courses.service.gov.uk/api/v3/recruitment_cycles/2020/providers/1N1/?include=sites,courses.sites
```
duncanjbrown: :+1:

app/models/course.rb
```diff
@@ -15,6 +15,7 @@ class Course < ApplicationRecord
     further_education: 'Further education',
   }, _suffix: :course
 
+  scope :that_candidates_can_apply_to, -> { where(exposed_in_find: true) }
```
benilovj: There are two attributes on Find courses:

* `findable?` - these courses are listed on Find (with or without vacancies)
* `open_for_applications?` - these courses are listed on Find and can be applied to (i.e. definitely have vacancies)

so the labelling isn't quite correct

duncanjbrown: Would it be confusing for someone to arrive from Find with a `findable?` course and then not be able to summon it using the find-a-course UI on the Apply side?

benilovj: The Apply action isn't available on Find if `open_for_applications?` is false.

tijmenb: My intention was for that scope to gain additional checks, like whether or not the provider is whitelisted to do Dual running (part of https://trello.com/c/5wXHsiOC). That's why the more generic name. 

I've also [added a story to the candidate board](https://trello.com/c/LAh0iGIc/348-only-allow-candidates-to-apply-to-courses-that-have-spaces-available) to show/hide courses based on the availibility - I don't think that was previously planned in.

Does that make sense?

benilovj: Uhh still uneasy with a mismatch between naming and what it actually does – can't it just be renamed at the same time as you broaden the scope?

tijmenb: It's not a mismatch, it accurately describes the situation: you can really apply to all these courses in the current version of Apply.

Removed the scope here: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/506/commits/ac8f42b3643f846a7669fcb3885a559bd46b9d21

benilovj: Can apply ≠ should be able to apply, thanks for taking the scope out.

PR #507, opened 2019-11-13
Put the provider interface behind DfE Sign-in

duncanjbrown:
### Context

This is how providers will authenticate to the service.

Add omniauth and configure DfE Sign-in as a provider on the model of Claim https://github.com/DFE-Digital/dfe-teachers-payment-service/blob/master/config/initializers/omniauth.rb

It is now possible to sign in to the provider interface at `/provider`. You assume the identity of provider `ABC` as this is still hardcoded in the controller. Also, it's impossible to log out 😏. 

### Guidance to review

You will need the DfE Sign in credentials for your `.env` — please slack me if you need them.

### Link to Trello card

### Env vars

- [x] If this PR introduces new environment variables, they have been added to all necessary parts of the Azure config based on the [documentation in the README](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training#azure-hosting-devops-pipeline)

### DIFF
--- a/app/controllers/provider_interface/provider_interface_controller.rb
+++ b/app/controllers/provider_interface/provider_interface_controller.rb
@@ -11,7 +11,7 @@ module ProviderInterface
     # Stub out the current user and their organisation. Will be replaced
     # by a proper ProviderUser when implementing Signin.
     def current_provider_user
-      if session[:provider_user]
+      if session['provider_user']
         fake_user_class = Struct.new(:provider, :email_address)
         fake_provider = Provider.find_by(code: 'ABC')
 
--- a/app/controllers/provider_interface/provider_interface_controller.rb
+++ b/app/controllers/provider_interface/provider_interface_controller.rb
@@ -3,5 +3,27 @@ module ProviderInterface
     include BasicAuthHelper
     before_action :require_basic_auth_for_ui
     layout 'application'
+
+    helper_method :current_provider_user
+
+  private
+
+    # Stub out the current user and their organisation. Will be replaced
+    # by a proper ProviderUser when implementing Signin.
+    def current_provider_user
+      if session[:provider_user]
+        fake_user_class = Struct.new(:provider, :email_address)
+        fake_provider = Provider.find_by(code: 'ABC')
+
+        fake_user_class.new(
+          fake_provider,
+          session['provider_user']['email_address'],
+        )
+      end
+    end
+
+    def authenticate_provider_user!
+      redirect_to provider_interface_sign_in_path unless current_provider_user
+    end
   end
 end
--- a/app/controllers/provider_interface/provider_interface_controller.rb
+++ b/app/controllers/provider_interface/provider_interface_controller.rb
@@ -1,7 +1,7 @@
 module ProviderInterface
   class ProviderInterfaceController < ActionController::Base
     include BasicAuthHelper
-    before_action :require_basic_auth_for_ui
+    before_action :authenticate_provider_user!
     layout 'application'
 
     helper_method :current_provider_user
--- a/config/initializers/omniauth.rb
+++ b/config/initializers/omniauth.rb
@@ -1,0 +1,9 @@
+options = {
+  name: :dfe,
+  callback_path: '/auth/dfe/callback',
+}
+
+class OmniAuth::Strategies::Dfe < OmniAuth::Strategies::OpenIDConnect; end
+
+Rails.application.config.middleware.use OmniAuth::Strategies::Dfe, options
+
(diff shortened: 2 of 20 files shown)


app/controllers/provider_interface/provider_interface_controller.rb
```diff
@@ -1,7 +1,27 @@
...
+    helper_method :current_user
+
+    def current_user
+      if session[:provider_user]
+        fake_user_class = Struct.new(:provider, :email_address)
+        fake_provider_class = Struct.new(:code)
+
+        fake_user_class.new(
+          fake_provider_class.new('ABC'),
+          session[:provider_user]['email_address'],
```
tijmenb: `email_address` is a symbol when setting the session, can we use that there too?

config/initializers/omniauth.rb
```diff
@@ -0,0 +1,28 @@
+OmniAuth.config.logger = Rails.logger
+
+dfe_sign_in_issuer = ENV['DFE_SIGN_IN_ISSUER']
```
tijmenb: Perhaps `DFE_SIGN_IN_URI`?

config/initializers/omniauth.rb
```diff
@@ -0,0 +1,28 @@
...
+dfe_sign_in_redirect_base_url = HostingEnvironment.application_url
+dfe_sign_in_identifier = ENV['DFE_SIGN_IN_CLIENT_ID']
+dfe_sign_in_secret = ENV['DFE_SIGN_IN_SECRET']
+
+dfe_sign_in_issuer_uri = dfe_sign_in_issuer.present? ? URI(dfe_sign_in_issuer) : nil
```
tijmenb: Perhaps this can use the ENV var directly to avoid intermediate variables.

app/controllers/provider_interface/provider_interface_controller.rb
tvararu: I've noticed some weird behaviour around `session` when working on similar code in `manage-courses-frontend`; I believe though it's set as a symbol and you can call it by a symbol within the current request lifecycle, it turns into a string on future requests. Could be mistaken, but I remember this being a bit unintuitive/tricky

duncanjbrown: :+1: I'll see if I can make this consistent

app/controllers/provider_interface/provider_interface_controller.rb
```diff
@@ -1,7 +1,27 @@
 module ProviderInterface
   class ProviderInterfaceController < ActionController::Base
     include BasicAuthHelper
-    before_action :require_basic_auth_for_ui
+    before_action :authenticate_provider_user!
     layout 'application'
+
+    helper_method :current_user
+
+    def current_user
```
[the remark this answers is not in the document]
duncanjbrown: yes to all

config/initializers/omniauth.rb
duncanjbrown: :+1: yes, because this is the boundary

config/initializers/omniauth.rb
duncanjbrown: yes

duncanjbrown:
thanks @tijmenb, will fix this up

stevehook:
> > Looks good.
> > It would be good to get this integrated with the audit trail so that the correct attribution appears in the support interface application history (though not in the scope of this PR). We would need to add an `audit_user` method to `ProviderInterfaceController` and adapt https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/app/components/support_interface/audit_trail_item_component.rb#L15-L23
> > Happy to add a card if it makes sense.
> 
> Good spot, please do add a card 👍

Done https://trello.com/c/OsJEowau/324-integrate-provider-interface-with-audit-trail (not sure it's in the right place - I put it into tech debt on the backlog board)

app/controllers/provider_interface/provider_interface_controller.rb
duncanjbrown: @tvararu is right. When the session contents are parsed Rails turns them into string keys, presumably for a good reason (all symbols are valid strings but not vice versa?) https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/middleware/session/cookie_store.rb#L80. 

I've added a commit to make the assignment use strings only, so it's now consistent.

duncanjbrown:
Feedback should be ✅

app/controllers/provider_interface/provider_interface_controller.rb
duncanjbrown: @tvararu is right. Rails turns session keys into strings on read: https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/middleware/session/cookie_store.rb#L80

I've changed the assignment to a string for consistency

duncanjbrown:
#479 has been merged, need to fix up the specs

PR #508, opened 2019-11-13
Fix bug in page to show course isn't found

tijmenb:
When a course isn't found, we show an error page. However, there's a stray link in the template which means the error page throws up an error.

Fixes https://sentry.io/organizations/dfe-bat/issues/1331456942


duncanjbrown:
If we want to avoid users hitting basic auth, should we change the href to `#`?

PR #510, opened 2019-11-13
Request references from referees when application is submitted

fofr:
### Context

If we manually send reference requests, we need to note their status somehow, either with a manual tracker spreadsheet or with a reference status.

Now we have the Referee emailer from #499, we can send the emails automatically.

- Automatically send reference request emails when an application is submitted
- Add test coverage for all emails being sent
- Includes referee name in emails as it's now available

https://trello.com/c/6zDP7N8t/1262-allow-support-to-send-an-email-to-request-references

### DIFF
--- a/spec/services/submit_application_spec.rb
+++ b/spec/services/submit_application_spec.rb
@@ -16,30 +16,6 @@ RSpec.describe SubmitApplication do
       expect(application_form.support_reference).not_to be_empty
     end
 
-    context 'sending emails' do
-      let(:mailer) { double }
-
-      before { allow(mailer).to receive(:deliver_now) }
-
-      it 'sends an application submitted email' do
-        allow(CandidateMailer).to receive(:submit_application_email).and_return(mailer)
-
-        SubmitApplication.new(application_form).call
-        expect(CandidateMailer).to have_received(:submit_application_email).with(application_form)
-        expect(mailer).to have_received(:deliver_now)
-      end
-
-      it 'sends a reference request email to each referee' do
-        allow(RefereeMailer).to receive(:reference_request_email).and_return(mailer)
-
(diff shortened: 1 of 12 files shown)


spec/services/submit_application_spec.rb
```diff
@@ -2,19 +2,50 @@
...
+      expect(application_form.application_choices[1]).to be_awaiting_references
+      expect(application_form.support_reference).not_to be_empty
+    end
+
+    context 'sending emails' do
+      let(:mailer) { double }
+
+      before { allow(mailer).to receive(:deliver_now) }
+
+      it 'sends an application submitted email' do
```
duncanjbrown: I think this is covered in `candiate_submitting_application_spec.rb`

spec/services/submit_application_spec.rb
```diff
@@ -2,19 +2,50 @@
...
+        allow(CandidateMailer).to receive(:submit_application_email).and_return(mailer)
+
+        SubmitApplication.new(application_form).call
+        expect(CandidateMailer).to have_received(:submit_application_email).with(application_form)
+        expect(mailer).to have_received(:deliver_now)
+      end
+
+      it 'sends a reference request email to each referee' do
```
duncanjbrown: This isn't, though. I wonder if it's better to have it in a referee-focussed system spec rather than here, anyway? (for example, If we change to async email delivery, which is likely, this spec will need to change)

fofr:
Holding off merging this until https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/509 is ready. This PR is dependent on the ability to complete referee details in the candidate form so that `candiate_submitting_application_spec` can be updated.

PR #514, opened 2019-11-14
Convert summary card line breaks into HTML

tijmenb:
### Context

Users may enter line breaks into text fields, but we're not displaying them at the moment.

This turns the line breaks a user entered into a field into HTML breaks and paragraphs, using `simple_format`:

> Returns text transformed into HTML using simple formatting rules. Two or more consecutive newlines(\n\n or \r\n\r\n) are considered a paragraph and wrapped in <p> tags. One newline (\n or \r\n) is considered a linebreak and a <br /> tag is appended. This method does not remove the newlines from the text.

https://apidock.com/rails/ActionView/Helpers/TextHelper/simple_format

### Guidance to review

When you enter paragraphs:

Screenshot 2019-11-14 at 09 30 23

They display correctly:

Screenshot 2019-11-14 at 09 30 18

Single values still look the same:

Screenshot 2019-11-14 at 09 32 56

### Link to Trello card

Finishes up https://trello.com/c/Hd0PKANG/278-adding-personal-statement

### DIFF
--- a/app/components/summary_card_component.html.erb
+++ b/app/components/summary_card_component.html.erb
@@ -11,7 +11,7 @@
             <% if row[:value].is_a?(Array) %>
               <% row[:value].each do |value| %><%= value %><br><% end %>
             <% else %>
-              <%= row[:value] %>
+              <%= simple_format row[:value], { class: 'govuk-body' } %>
             <% end %>
           </dd>
 


chubberlisk:
Thinking we probably have a number of places where we use a text area. Could this PR include using `simple_format` in the other places?

PR #517, opened 2019-11-14
Reject by default business rule

stevehook:
### Context

This is another business rule. In a nutshell applications are automatically rejected if a provider doesn't make a decision within a configurable time limit.

- [x] Add Cucumber feature for the high-level specification
- [x] Add an `ApplicationChoice#reject_by_default_by` attribute
- [x] Add a new lookup table and model for 'time limits'
- [x] Implement setting `ApplicationChoice#reject_by_default_by` when application is 'sent' to provider
- [x] Implement worker (triggered to run hourly by Clockwork/Sidekiq) to automatically reject applications that are in the `awaiting_provider_decision` status and have passed their `reject_by_default_by` date
- [x] Revisit `ApplicationDates`. Make consistent with new rule.

### Guidance to review

- Is there anything about the new cucumber feature that we can improve?
- Do the various service classes added here make sense? (Beyond the scope of this PR we may need to consider organising these classes by type or feature).
- I've configured the background job to run every hour to be consistent with the similar `SendApplicationToProvider` job though, in theory, it only needs to run daily.
- I've updated `ApplicationDates` and `ApplicationCompleteContentComponent` so handle situations in which the `reject_by_default_by` date is not available (because application has not been sent to provider yet). Do these make sense?
- I've added a line in `seeds.rb` to seed the `time_limits` table - For now it's just a 40 day default limit. Without this RBDs just don't happen.

### Link to Trello card

[1245 - Implement RBD business rule](https://trello.com/c/YawJFcim/1245-rbd-business-rule)

### Env vars

No new env vars


app/models/time_limit.rb
```diff
@@ -0,0 +1,4 @@
+class TimeLimit < ApplicationRecord
```
tijmenb: As discussed, I'm a little on the fence about having this in the database. Adding new rules in the database will require a PR anyway, so we don't win anything in terms of deployment speed. Have you considered to hard code the values in a way that doesn't require table, but keeps the nice extensibility of `TimeLimitCalculator`?

My preference is that the app code works with an empty database. Since `seeds.rb` doesn't run automatically, does this mean the app will error if we merge this PR as it is currently?

db/migrate/20191113165001_add_application_choices_reject_by_default_at.rb
```diff
@@ -0,0 +1,5 @@
+class AddApplicationChoicesRejectByDefaultAt < ActiveRecord::Migration[6.0]
+  def change
+    add_column :application_choices, :reject_by_default_at, :datetime
```
tijmenb: Since it'll be configurable, perhaps also save the number of days here.

stevehook: Yes could do that, you mean as well as the timestamp?

app/models/time_limit.rb
stevehook: It doesn't error with an empty DB - it just doesn't ever reject by default. To make it work with an empty database we'd need a hard-coded default. If we are going to do that then perhaps we should bring the whole config into code. The interface to `TimeLimitCalculator` can stay as it is even if we put the config into a constant collection of rules, e.g.:

```
  TimeLimit = Struct.new(:from_date, :to_date, :limit)
  RULES = {
    reject_by_default: [
      TimeLimit.new(nil, nil, 40),
      TimeLimit.new(Date.new(2020, 8, 1), Date.new(2020, 9, 30), 30),
      TimeLimit.new(Date.new(2020, 10, 1), Date.new(2020, 12, 31), 20),
    ],
  }.freeze
```

I'm on the fence too, happy to be pushed either way!

Having it in the database makes more sense if we think we are likely to have an 'admin' interface to tweak these rules but you could argue YAGNI.

tijmenb: 👍 Your hard coded config looks good, I'd be happy with that.

tvararu: I also like that approach!

stevehook: :+1: I've removed the DB table/model and added the hard-coded config.

spec/factories.rb
```diff
@@ -162,4 +162,9 @@
     email_address { "#{SecureRandom.hex(5)}@example.com" }
     accept_ts_and_cs { true }
   end
+
+  factory :time_limit do
```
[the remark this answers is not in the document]
stevehook: this is gone now

app/services/send_application_to_provider.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+private
+
+  def set_reject_by_default
+    days = TimeLimitCalculator.new(
+      rule: :reject_by_default,
+      effective_date: Time.zone.now,
+    ).call
+    return unless days
```
tijmenb: Can this ever be nil?

app/services/send_application_to_provider.rb
```diff
@@ -0,0 +1,26 @@
...
+class SendApplicationToProvider
+  attr_accessor :application_choice
+
+  def initialize(application_choice:)
+    self.application_choice = application_choice
+  end
+
+  def call
+    set_reject_by_default
+    ApplicationStateChange.new(application_choice).send_to_provider!
```
tijmenb: Are we relying on the state change to persist the attribute changes? Might be better to explicitly save here, even if it costs an extra query.

app/services/reject_application_by_default.rb
```diff
@@ -0,0 +1,14 @@
+class RejectApplicationByDefault
+  attr_accessor :application_choice
+
+  def initialize(application_choice:)
+    self.application_choice = application_choice
+  end
+
+  def call
+    #TODO: Do we need to include some other flag here to distinguish
```
tijmenb: Good point, let's card this up and remove the comment.

app/services/send_application_to_provider.rb
stevehook: Well I guess not any more. When the config was in the DB this made sense but not now - will remove.

app/services/reject_application_by_default.rb
stevehook: https://trello.com/c/VrqWtwch/1280-add-boolean-attribute-to-distinguish-application-choices-that-have-been-rejected-by-default

PR #521, opened 2019-11-14
Check that all questions are answered before submit

tijmenb:
### Context

This makes sure that the user has filled in all the fields before submitting.

To do:

- [x] Add logic to check whether the form is complete
- [ ] Test behaviour in one of the feature tests
- [ ] Prevent submission of forms if they're not complete
- [ ] Come up with a shortcut for manual testing
- [ ] Implement a design 

### Link to Trello card

https://trello.com/c/TYfYUDKt/351-check-that-all-questions-are-answered-before-submit

### DIFF
--- a/app/presenters/candidate_interface/application_form_presenter.rb
+++ b/app/presenters/candidate_interface/application_form_presenter.rb
@@ -4,6 +4,34 @@ module CandidateInterface
       @application_form = application_form
     end
 
+    def ready_to_submit?
+      [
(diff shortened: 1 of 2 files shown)


stevehook:
Is there a case for using `dry-validations` for this kind of thing? (not sure it would into a presenter - just a thought)

tijmenb:
@tvararu will continue with this PR.

PR #522, opened 2019-11-14
Avoid reloading references in the receive references action

tijmenb:
Update to https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/509.

We are currently reloading the application_form object in `ReceiveReference` because of the way we set up the test data in the specs.

By explicitly updating the `application_form.references` in the tests, the object will already be up to date when passed into the `ReceiveReference` and we won't have to `reload`.

https://trello.com/c/33rFaaqv/200-requesting-references-from-candidate

### DIFF
--- a/spec/services/receive_reference_spec.rb
+++ b/spec/services/receive_reference_spec.rb
@@ -4,8 +4,8 @@ RSpec.describe ReceiveReference do
   it 'updates the reference on an application form with the provided text' do
     application_form = FactoryBot.create(:completed_application_form, references_count: 0)
     application_form.application_choices.each { |choice| choice.update(status: 'awaiting_references') }
-    create(:reference, :unsubmitted, email_address: 'ab@c.com', application_form: application_form)
-    create(:reference, :unsubmitted, email_address: 'xy@z.com', application_form: application_form)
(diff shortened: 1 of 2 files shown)


spec/services/receive_reference_spec.rb
```diff
@@ -4,8 +4,8 @@
   it 'updates the reference on an application form with the provided text' do
     application_form = FactoryBot.create(:completed_application_form, references_count: 0)
     application_form.application_choices.each { |choice| choice.update(status: 'awaiting_references') }
-    create(:reference, :unsubmitted, email_address: 'ab@c.com', application_form: application_form)
-    create(:reference, :unsubmitted, email_address: 'xy@z.com', application_form: application_form)
+    application_form.references << build(:reference, :unsubmitted, email_address: 'ab@c.com')
+    application_form.references << build(:reference, :unsubmitted, email_address: 'xy@z.com')
```
RobertM905: `application_form: application_form` is removed here but kept in the  creates below.
Is this intentional?

tijmenb: Done!

PR #530, opened 2019-11-15
Fix for application_form.references needing explicit reload in tests

aldavidson:
### Context

A [known-issue with FactoryBot's way of creating dependent records](https://github.com/thoughtbot/factory_bot/issues/549) means that because we are validating the length of the `application_form.references` association in the application_form, we have to explicitly reload the association after creating references.

This gives counter-intuitive results, for instance: 
```
FactoryBot.build(:completed_application_form).references.length
=> 0
```

There's no way to _fix_ this, so we have to workaround it. This PR adds an explicit `references.reload` after creating the references in the factory, so that we don't need to do that in tests which use that association. It also tidies up a couple of specs that can be less verbose as a result.

### Guidance to review

run the tests, if they're all green, we're good.

### Link to Trello card

No trello card created for this issue, but it was highlighted after merging PR #509 for [200 - Requesting references from candidate](https://trello.com/c/33rFaaqv/200-requesting-references-from-candidate)

### Env vars

n/a

### DIFF
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -109,7 +109,7 @@ FactoryBot.define do
   factory :course do
     provider
 
-    code { Faker::Alphanumeric.alphanumeric(number: 4).upcase }
+    code { Faker::Alphanumeric.alphanumeric(number: 4, min_alpha: 1).upcase }
     name { Faker::Educator.subject }
     level { 'primary' }
     start_date { Date.new(2020, 9, 1) }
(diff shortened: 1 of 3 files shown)


spec/factories.rb
```diff
@@ -51,6 +51,14 @@
...
         create_list(:application_qualification, evaluator.qualifications_count, application_form: application_form)
         create_list(:reference, evaluator.references_count, application_form: application_form)
+        # The application_form validates the length of this collection when
+        # it is created, which is BEFORE we create the references here.
+        # This then *caches* the association on the  application_form, and means
+        # you have to explicitly reload it to pick up the created references.
+        # We do this here, so we only have to do it in one place, rather than
+        # everywhere we refer to application_form.references in tests.
+        # See https://github.com/thoughtbot/factory_bot/issues/549 for details.
+        application_form.references.reload
```
tvararu: 👍

PR #533, opened 2019-11-15
add audit trail to References

aldavidson:
### Context

all changes to an application form should be auditable

Add the `audited` setup to the Reference model, and a test to make sure that it works.

### Guidance to review

1. Go to an application form. 
2. Add/update/delete a reference.
3. Go to the support interface /support/applications
4. Click on the application's "History" link
5. You should see the audit entries for the reference.

### Link to Trello card

[384 - Adding a referee didn't show up in the audit trail in the support app](https://trello.com/c/MPtdICu2/384-adding-a-referee-didnt-show-up-in-the-audit-trail-in-the-support-app)

### Env vars

N/A

### DIFF
--- a/spec/models/reference_spec.rb
+++ b/spec/models/reference_spec.rb
@@ -71,13 +71,8 @@ RSpec.describe Reference, type: :model do
   describe 'auditing' do
     let(:application_form) { create(:application_form) }
 
-    it 'creates audit entries' do
-      reference = create :reference, application_form: application_form
-      expect(reference.audits.count).to eq 1
-      reference.update!(feedback: 'hello again')
-      expect(reference.audits.count).to eq 2
-    end
-
+    it { should be_audited.associated_with :application_form }
+    
     it 'creates an associated object in each audit record' do
       reference = create :reference, application_form: application_form
       expect(reference.audits.last.associated).to eq reference.application_form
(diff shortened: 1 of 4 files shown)


spec/models/reference_spec.rb
```diff
@@ -67,4 +67,28 @@
       end
     end
   end
+
+  describe 'auditing' do
```
benilovj: [The RSpec matchers](https://www.rubydoc.info/gems/audited/4.4.1/RSpec/Matchers) that come baked into `audited` may slim these specs down?

benilovj: See https://github.com/DFE-Digital/manage-courses-backend/pull/259/files for examples

aldavidson: thanks! I've used those to replace 2 of the 3 cases, I think the third probably still needs testing explicitly

spec/models/reference_spec.rb
```diff
@@ -67,4 +67,23 @@
...
+    let(:application_form) { create(:application_form) }
+
+    it { should be_audited.associated_with :application_form }
```
stevehook:
```
    it { is_expected.to be_audited.associated_with :application_form }
```

aldavidson: 👍 done

PR #535, opened 2019-11-15
Validate reference email addresses are unique within an application_form

aldavidson:
### Context

QA testing found that although the DB constraint prevented a duplicate email address being entered, it wasn't being validated - so the user just saw a 500 error page, rather than an explanatory form validation.

Add a validation to catch this when the user tries to enter a duplicate referee email within an application form (see screenshot)

Screen Shot 2019-11-15 at 14 02 12

### Guidance to review

1. Go to an application form with no referees
2. Add a referee
3. Add a second referee with the same email address
4. You should see a validation error like in the screenshot above

### Link to Trello card

No specific trello card, but it cropped up in testing of [200 - requesting references from candidate](https://trello.com/c/33rFaaqv/200-requesting-references-from-candidate)

### Env vars

N/A

### DIFF
--- a/app/models/reference.rb
+++ b/app/models/reference.rb
@@ -1,6 +1,8 @@
 class Reference < ApplicationRecord
   validates :name, presence: true, length: { minimum: 2, maximum: 200 }
-  validates :email_address, presence: true, length: { maximum: 100 }
+  validates :email_address, presence: true,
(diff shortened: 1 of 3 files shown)


fofr:
Nit: We don't need the "Please" in the validation message.

PR #536, opened 2019-11-15
Add a flash message mechanism

mnacos:
### Context

We need a flash message mechanism for the app.

In Rails, flash messages are a well understood/established method of providing feedback. For clarification:

- We are talking about the standard Rails flash mechanism e.g. flash[:error] = 'oops'
- The flash message will be a constant part of page content (i.e. no animations appearing/disappearing), good for accessibility
- We'll need three levels of severity to start with: success, info, warning

To use a single FlashMessageComponent for the display of flash messages within the app, referenced within the application layout, so that views do not need to include it explicitly.

### Guidance to review

Set ```flash[:success] = 'something'``` in any controller action and you should see the relevant flash message in your view. Examples:

With ```flash[:success]```:
image

With ```flash[:info]```:
image

With ```flash[:warning]```:
image

### Link to Trello card

[365 - Implement flash message system](https://trello.com/c/zkYK0EVJ)

### Env vars

No new environment variables.


paulrobertlloyd:
The space below the message seems unusually large… but is using the same values as in the prototype. Which would suggest there is some spacing above the page title (or it’s parent container). Will investigate…

paulrobertlloyd:
Two changes:

* rename `_app-banner.scss` to `_banner.scss` (we don’t include the `app-` prefix in file names)
* move banner component to sit within `govuk-main-wrapper`. That fixes the spacing issue.

Mocked up to show a banner on the start page:

Screenshot 2019-11-18 at 12 15 47

mnacos:
> Two changes:
> 
> * rename `_app-banner.scss` to `_banner.scss` (we don’t include the `app-` prefix in file names)
> * move banner component to sit within `govuk-main-wrapper`. That fixes the spacing issue.
> 

Thanks Paul, will merge.

PR #539, opened 2019-11-16
Add cookie policy page

dankmitchell:
### Context
Cookie policy as per https://www.publish-teacher-training-courses.service.gov.uk/cookies

Add cookie policy and some markdown overrides to get tables displaying correctly 😓 

### Guidance to review
`/cookies`

localhost_3000_candidate_cookies(Laptop with MDPI screen)

### Link to Trello card
[84 - Create cookies page](https://trello.com/c/thu6GdXn/84-create-cookies-page)

### Env vars
n/a

### DIFF
--- a/app/lib/govuk/markdown_renderer.rb
+++ b/app/lib/govuk/markdown_renderer.rb
@@ -1,0 +1,36 @@
+module Govuk
+  class MarkdownRenderer < ::Redcarpet::Render::Safe
+    def table(header, body)
+      <<~HTML
+        <table class='govuk-table'>
+          <thead class='govuk-table__head'>
+            #{header}
+          </thead>
+          <tbody class='govuk-table__body'>
+            #{body}
+          </tbody>
+        </table>
+      HTML
+    end
+
+    def table_row(content)
+      <<~HTML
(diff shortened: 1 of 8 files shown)


dankmitchell:
@tvararu screenshot attached. The actual values need confirming but this a solid start and should be merged.

We need a cookie message and probably Google Tag Manager

app/lib/govuk/markdown_renderer.rb
```diff
@@ -0,0 +1,36 @@
+module Govuk
+  class MarkdownRenderer < ::Redcarpet::Render::Safe
+    def table(header, body)
```
tijmenb: @dankmitchell  just to be sure, could we have achieved this by styling tables in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/app/frontend/styles/_styled-content.scss?

dankmitchell: Yes but it’s not a mixin so would have involved copying all the SCSS into the app and then any upgrades to govuk-frontend might be slightly messier.

I chose to extend markdown instead, we’ve done the same on FIND

PR #547, opened 2019-11-18
Make sure Sidekiq jobs report correct environment

tijmenb:
### Context

We rely on `CUSTOM_HOSTNAME` to populate the "hosting environment" for reporting to Sentry (https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/478).

Because it's not set in the Sidekiq worker, they are currently reporting development (the fallback). For an example, see https://sentry.io/organizations/dfe-bat/issues/1341153288, which is happening on QA.

I've also added it the the clock process, which probably needs this is as well.

### Guidance to review

@tpinney is this enough to get the environment variable working?

### Link to Trello card

https://trello.com/c/YtFlMRNo/1268-allow-sidekiq-workers-to-access-the-database

### DIFF
--- a/azure/template.json
+++ b/azure/template.json
@@ -550,6 +550,10 @@
                                 "secureValue": "[parameters('secretKeyBase')]"
                             },
                             {
+                                "name": "CUSTOM_HOSTNAME",
+                                "value": "[parameters('customHostName')]"
+                            },
+                            {
                                 "name": "SENTRY_DSN",
                                 "value": "[parameters('sentryDSN')]"
                             },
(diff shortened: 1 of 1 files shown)


tpinney:
Yes that should be sufficient to get the custom hostname working in the two new containers instances.

The only thing to be mindful of is that some of the environments don't have a gov.uk domain name (e.g. pentest) and in those cases that environment variable can't be populated. I'm not sure how your application will handle that scenario.

tpinney:
You could set `CUSTOM_HOSTNAME` to the value of the `environment` parameter instead. That is set in all cases and might be a more appropriate way to differentiate the environments.

PR #548, opened 2019-11-18
Refactor presenters to use SimpleDelegator

fofr:
- Simplify the presenters
- Use delegation to access application_choice and application_form models to DRY up code

### DIFF
--- a/app/presenters/provider_interface/application_choice_presenter.rb
+++ b/app/presenters/provider_interface/application_choice_presenter.rb
@@ -1,24 +1,11 @@
 module ProviderInterface
-  class ApplicationChoicePresenter
-    def initialize(application_choice)
-      @application_choice = application_choice
-      @application_form = application_choice.application_form
-    end
-
-    def id
-      application_choice.id
-    end
-
-    def status
-      application_choice.status
-    end
-
+  class ApplicationChoicePresenter < SimpleDelegator
     def status_tag_text
-      application_choice.status.humanize.titleize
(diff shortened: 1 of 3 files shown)


app/presenters/provider_interface/application_choice_presenter.rb
```diff
@@ -1,24 +1,11 @@
...
-    end
-
-    def status
-      application_choice.status
-    end
-
+  class ApplicationChoicePresenter < SimpleDelegator
     def status_tag_text
-      application_choice.status.humanize.titleize
+      status.humanize.titleize
```
tijmenb: I'm not following how this works, how does it know where `status` is defined?

fofr: It's inherited from `< SimpleDelegator`, it causes method missing to resolve everything to the object given to the constructor - @duncanjbrown

tijmenb:
I'd like to propose a different refactor that should achieve much the same:

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/compare/alternative-refactor

This way we only delegate the methods that we need to sub-objects. The methods can stay private, and we don't have to make modifications to core models.

Thoughts?

duncanjbrown:
We're going to go at this outside-in (https://trello.com/c/lf08PU3R/1233-provider-can-view-an-application) preferring the @tijmenb approach and enhancing the ApplicationChoicePresenter as we go. If we end up with lots of duplication we can look again

PR #552, opened 2019-11-18
Improve hosting environment in Sentry

tijmenb:
This supersedes https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/547. Instead of setting the `CUSTOM_HOSTNAME`, we can also set a custom variable. This is better because the pentest environment doesn't have a `CUSTOM_HOSTNAME `.


tpinney:
From the ARM template point of view everything is fine with me. Not sure I'm qualified to comment on any of the rest of the code changes.

app/lib/hosting_environment.rb
```diff
@@ -35,19 +35,16 @@ def self.phase_banner_text
       'This is a internal environment used by DfE to test deploys'
     when 'development'
       'This is a development version of the Apply service'
+    when 'unknown-environment'
```
tvararu: :alien:

PR #553, opened 2019-11-18
Allow Sidekiq process to talk to Postgres

tijmenb:
### Context

This commit should configure the firewall to allow Postgres access from the Sidekiq worker.

Fixes errors like https://sentry.io/organizations/dfe-bat/issues/1341153288.

Paired with @tpinney.

### Guidance to review

Does this make sense?

### Link to Trello card

https://trello.com/c/YtFlMRNo/1268-allow-sidekiq-workers-to-access-the-database

### DIFF
--- a/azure/template.json
+++ b/azure/template.json
@@ -841,7 +841,7 @@
                         "value": "AZURE_IP-"
                     },
                     "ipAddresses": {
-                        "value": "[reference('app-service').outputs.possibleOutboundIpAddresses.value]"
+                        "value": "[concat(reference('app-service').outputs.possibleOutboundIpAddresses.value, array(reference('container-instances-worker').outputs.containerIPv4Address.value))]"
                     },
                     "serverName": {
                         "value": "[variables('databaseServerName')]"
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -70,3 +70,6 @@ RSpec/InstanceVariable:
 RSpec/PredicateMatcher:
   Exclude:
     - 'spec/support_specs/clockwork_spec.rb'
+
+RSpec/DescribeClass:
+  Enabled: false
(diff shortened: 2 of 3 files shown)


azure/template.json
```diff
@@ -841,7 +841,7 @@
                         "value": "AZURE_IP-"
                     },
                     "ipAddresses": {
-                        "value": "[reference('app-service').outputs.possibleOutboundIpAddresses.value]"
+                      "[concat(reference('app-service').outputs.possibleOutboundIpAddresses.value, array(reference('container-instances-worker').outputs.containerIPv4Address.value))]"
```
tpinney: This line should read `"value": "[concat(reference('app-service').outputs.possibleOutboundIpAddresses.value, array(reference('container-instances-worker').outputs.containerIPv4Address.value))]"`

tpinney:
I also recommend you test the deployment phase from your branch before you commit to master. You can change the `devDeployBranchNameOverride` variable in the DevOps variable group to the name of your branch to get it to deploy from your branch as well as master.

azure/template.json
tijmenb: Fixed, also added a test that would've caught this:

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/553/commits/1f05cae2b5a29c1b82eb95b19707d40e76d64bb2

tijmenb:
Testing & deploying in https://dfe-ssp.visualstudio.com/Become-A-Teacher/_build/results?buildId=30694&view=results.

tijmenb:
Closing this. The container doesn't have an IP address, so this failed when configuring.

@tpinney has enabled "Azure service access" for Postgres, which means that internal services can all talk to Postgres.

PR #555, opened 2019-11-18
Allow provider make an offer via provider UI

RobertM905:
### Context

Providers using the UI need a process for making offers on applications

Adding pages for `new_offer` and `confirm_offer`.
Updating `routes` and `decisions_controller` to use the new pages.
Add system spec to test the happy path of creating an offer.

Screenshot 2019-11-18 at 16 23 16

Screenshot 2019-11-18 at 16 24 26

Screenshot 2019-11-18 at 16 24 37
### Guidance to review

### Link to Trello card

[1256 - Make an offer using provider UI](https://trello.com/c/QP87Ig3k/1256-provider-make-an-offer)

### Env vars

None

### DIFF
--- a/app/controllers/provider_interface/decisions_controller.rb
+++ b/app/controllers/provider_interface/decisions_controller.rb
@@ -18,12 +18,13 @@ module ProviderInterface
     end
 
     def confirm_offer
-      standard_conditions_array = params.dig(:make_an_offer, :standard_conditions)
-      further_conditions_array = [params.dig(:make_an_offer, :first_condition), params.dig(:make_an_offer, :second_condition), params.dig(:make_an_offer, :third_condition), params.dig(:make_an_offer, :further_condition)].reject(&:blank?)
-      complete_conditions_array = [standard_conditions_array, further_conditions_array].compact.reduce([], :|)
+      offer_conditions = [
+        make_an_offer_params[:standard_conditions],
+        make_an_offer_params[:further_conditions]
+      ].flatten.reject(&:blank?)
       @application_offer = MakeAnOffer.new(
         application_choice: @application_choice,
-        offer_conditions: complete_conditions_array,
+        offer_conditions: offer_conditions,
       )
       render action: :new_offer if !@application_offer.valid?
     end
@@ -80,6 +81,10 @@ module ProviderInterface
         .find(params[:application_choice_id])
 
       @presenter = ApplicationChoicePresenter.new(@application_choice)
+    end
+
+    def make_an_offer_params
+      params.require(:make_an_offer)
     end
   end
 end
--- a/app/controllers/provider_interface/decisions_controller.rb
+++ b/app/controllers/provider_interface/decisions_controller.rb
@@ -14,7 +14,36 @@ module ProviderInterface
     end
 
     def new_offer
-      raise 'Not yet implemented'
+      @application_offer = MakeAnOffer.new(application_choice: @application_choice)
+    end
+
+    def confirm_offer
+      standard_conditions_array = params.dig(:make_an_offer, :standard_conditions)
+      further_conditions_array = [params.dig(:make_an_offer, :first_condition), params.dig(:make_an_offer, :second_condition), params.dig(:make_an_offer, :third_condition), params.dig(:make_an_offer, :further_condition)].reject(&:blank?)
+      complete_conditions_array = [standard_conditions_array, further_conditions_array].compact.reduce([], :|)
+      @application_offer = MakeAnOffer.new(
+        application_choice: @application_choice,
+        offer_conditions: complete_conditions_array
+      )
+      render action: :new_offer if !@application_offer.valid?
+    end
+
+    def create_offer
+      offer_conditions_array = JSON.parse(params.dig(:offer_conditions))
+
+      @application_offer = MakeAnOffer.new(
+        application_choice: @application_choice,
+        offer_conditions: offer_conditions_array
+      )
+
+      if @application_offer.save
+        flash[:success] = 'Application status changed to ‘Offer made’'
+        redirect_to provider_interface_application_choice_path(
+          application_choice_id: @application_choice.id,
+        )
+      else
+        render action: :new_offer
+      end
     end
 
     def new_reject
--- a/app/controllers/provider_interface/decisions_controller.rb
+++ b/app/controllers/provider_interface/decisions_controller.rb
@@ -37,7 +37,7 @@ module ProviderInterface
       )
 
       if @application_offer.save
-        flash[:success] = 'Application status changed to ‘Offer made’'
+        flash[:success] = 'Application status changed to ‘Offer’'
         redirect_to provider_interface_application_choice_path(
           application_choice_id: @application_choice.id,
         )
--- a/app/controllers/provider_interface/decisions_controller.rb
+++ b/app/controllers/provider_interface/decisions_controller.rb
@@ -38,7 +38,7 @@ module ProviderInterface
       )
 
       if @application_offer.save
-        flash[:success] = 'Application status changed to ‘Offer’'
+        flash[:success] = 'Application status changed to ‘Offer Made’'
         redirect_to provider_interface_application_choice_path(
           application_choice_id: @application_choice.id,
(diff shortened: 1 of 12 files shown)


spec/system/provider_interface/provider_makes_an_offer_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+    when_i_respond_to_an_application
+    and_i_choose_to_make_an_offer
+    then_i_see_some_application_info
+    when_i_select_standard_reasons
+    and_i_add_optional_further_conditions
+    and_i_click_to_continue
+    then_i_am_asked_to_confirm_the_offer
+    when_i_confirm_the_offer
+    then_i_am_back_to_the_application_page
+    and_i_can_see_the_application_has_has_offer_made
```
duncanjbrown: Typo: `has_has`

app/components/summary_card_component.html.erb
```diff
@@ -9,7 +9,13 @@
           </dt>
           <dd class="govuk-summary-list__value">
             <% if row[:value].is_a?(Array) %>
-              <% row[:value].each do |value| %><%= value %><br><% end %>
+              <% if @bullet == true %>
```
duncanjbrown: :+1: to the feature — it would be good to have a spec too

duncanjbrown: This will need a rebase on https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/574 (which also adds specs)

app/controllers/provider_interface/decisions_controller.rb
```diff
@@ -14,7 +14,36 @@ def submit_response
     end
 
     def new_offer
-      raise 'Not yet implemented'
+      @application_offer = MakeAnOffer.new(application_choice: @application_choice)
+    end
+
+    def confirm_offer
+      standard_conditions_array = params.dig(:make_an_offer, :standard_conditions)
+      further_conditions_array = [params.dig(:make_an_offer, :first_condition), params.dig(:make_an_offer, :second_condition), params.dig(:make_an_offer, :third_condition), params.dig(:make_an_offer, :further_condition)].reject(&:blank?)
```
duncanjbrown: Did you consider having a single "further_conditions" array which is numerically indexed? What's the downside there? We could do this:

which will reduce `dig`ging up here:

```
offer_conditions = [
  make_an_offer_params[:standard_conditions],
  make_an_offer_params[:further_conditions]
].flatten

...

def make_an_offer_params
  params.require(:make_an_offer)
end
```

app/controllers/provider_interface/decisions_controller.rb
```diff
@@ -14,7 +14,36 @@ def submit_response
...
+      complete_conditions_array = [standard_conditions_array, further_conditions_array].compact.reduce([], :|)
```
duncanjbrown: Would `flatten` do the trick here? https://ruby-doc.org/core-2.3.0_preview1/Array.html#method-i-flatten

app/views/provider_interface/decisions/confirm_offer.html.erb
```diff
@@ -0,0 +1,42 @@
...
+
+      <%= render(SummaryCardComponent, rows: [{ key: 'Conditions', value: @application_offer.offer_conditions }], border: false, bullet: true) %>
+      <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
+    </div>
+  </div>
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <p class="govuk-body govuk-!-font-weight-bold">
+        Please make a note of the conditions and recommendations you have set.
```
duncanjbrown: If recommendations are descoped, should we take them out of this sentence?

spec/system/provider_interface/provider_makes_an_offer_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+  include DfeSignInHelpers
+
+  let(:course_option) { course_option_for_provider_code(provider_code: 'ABC') }
+  let(:application_awaiting_provider_decision) {
+    create(:application_choice, status: 'awaiting_provider_decision', course_option: course_option, application_form: create(:application_form, first_name: 'Alice', last_name: 'Wunder'))
+  }
+
+  scenario 'Provider makes an offer' do
+    given_i_am_a_provider_user_authenticated_with_dfe_sign_in
+    when_i_respond_to_an_application
```
duncanjbrown: Please add line breaks before `when`s for readability

spec/system/provider_interface/provider_makes_an_offer_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+    expect(page).to have_content 'Payment of fees'
+    expect(page).to have_content 'A further condition'
+  end
+
+  def when_i_confirm_the_offer
+    click_on 'Confirm offer'
+  end
+
+  def then_i_am_back_to_the_application_page
+    expect(page).to have_current_path(
```
duncanjbrown: If possible, it's better to assert on something the user sees rather than an implementation detail like the URL

spec/system/provider_interface/provider_makes_an_offer_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+      provider_interface_application_choice_path(
+        application_awaiting_provider_decision.id,
+      ),
+    )
+  end
+
+  def and_i_can_see_the_application_has_has_offer_made
+    expect(page).to have_content 'Application status changed to ‘Offer made’'
```
duncanjbrown: This is a bit confusing when it happens because the offer status is "OFFER", not "OFFER MADE"

Screenshot 2019-11-19 at 12 39 17

app/controllers/provider_interface/decisions_controller.rb
```diff
@@ -14,7 +14,36 @@ def submit_response
...
+    def create_offer
+      offer_conditions_array = JSON.parse(params.dig(:offer_conditions))
+
+      @application_offer = MakeAnOffer.new(
+        application_choice: @application_choice,
+        offer_conditions: offer_conditions_array,
+      )
+
+      if @application_offer.save
+        flash[:success] = 'Application status changed to ‘Offer’'
```
fofr: Not keen on this success message. What's the effect of changing the status? Do we need to communicate more?

RobertM905: Changed to use en locale for displaying status. Should sort this out 👍 
updated here: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/555/commits/04095d4763930cf8266f569bf6a7e96b5b17d370

app/components/summary_card_component.html.erb
RobertM905: For now I've rebased and moved the bullet list code out of the summary-card / list components here: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/555/commits/bcabe35887beaaad7588f23cad391cb391c3cc41.

Currently this is the only page that requires a bulleted list in a summary card.
If this changes we can move functionality into the summary-list component.

RobertM905:
> Working on something else, I noticed that https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/spec/services/make_an_offer_spec.rb is no longer accurate. MakeAnOffer shouldn't be valid when still in status `:application_complete`, shouldn't it be valid only once `:awaiting_provider_decision`?

Good catch. I'll make the change here 👍

app/controllers/provider_interface/decisions_controller.rb
```diff
@@ -14,7 +14,37 @@ def submit_response
...
     def new_offer
-      raise 'Not yet implemented'
+      @application_offer = MakeAnOffer.new(application_choice: @application_choice)
+    end
+
+    def confirm_offer
+      offer_conditions = [
+        make_an_offer_params[:standard_conditions],
+        make_an_offer_params[:further_conditions],
+      ].flatten.reject(&:blank?)
```
duncanjbrown: nice!

PR #561, opened 2019-11-18
Update the Application dashboard with course choices and referees

chubberlisk:
### Context

Candidates need to be able to track their application after they submit it. Atm, we have a basic "Your application" page (see below for before screenshot).

This PR updates the "Your application" page to be the "Application Dashboard" page with the ability to view your course choices and referees.

### Before

image

### After

image

### Guidance to review

We added two new components `SubmittedCourseChoicesComponent` and `SubmittedRefereesComponent`, however we'll potentially remove these in our next PR and instead refactor all our current `ReviewComponent`s to handle not showing the change and delete links.

Review commit by commit.

### Link to Trello card

[359 - Build dashboard for candidates after they've submitted](https://trello.com/c/8NgjUhNG/359-epic-build-dashboard-for-candidates-after-theyve-submitted)


app/components/application_complete_content_component.html.erb
```diff
@@ -1,17 +1,12 @@
...
-  respond to you by <%= respond_by_date %>.
-</p>
-
 <% if @dates.form_open_to_editing? %>
   <div class="govuk-inset-text">
     <p class="govuk-body">
-      You have <%= formatted_days_remaining %> (until <%= edit_by_date %>) to withdraw your application and
-      resubmit it to a different course or provider. You’ll be able to edit
-      your application before you resubmit.
+      You have <%= render(ApplicationDateComponent, application_form: @application_form, type: :days_remaining).squish %>
```
alexdesi: Is `squish` necessary? maybe can we move it inside the component,
it is a bit noisy to have `.squish` after renders.

app/components/application_date_component.html.erb
```diff
@@ -0,0 +1,12 @@
...
+  when :submitted_at
+    submitted_at_date
+  when :respond_by
+    respond_by_date
+  when :edit_by
+    edit_by_date
+  when :days_remaining
+    formatted_days_remaining
+  end
+%>
```
alexdesi: It looks like this component is not doing a lot,
it is like a delegator (apart for `formatted_days_remaining`),
I am wondering if simple helper would not be simpler.

app/components/submitted_referees_component.rb
```diff
@@ -0,0 +1,41 @@
...
+    @application_form = application_form
+  end
+
+  def referee_rows(work)
+    [
+      name_row(work),
+      email_row(work),
+      relationship_row(work),
+    ]
+      .compact
```
alexdesi: nitpick: compact looks so so lonely, mayne should we attach it to the `]`?

app/components/application_complete_content_component.html.erb
tvararu: Yes; the component returns a buffer which is `'15 May 2019\n'`. Removing the `\n` is not necessary in this exact case, but is necessary in cases like:

```
Application submitted on 18 November 2019 .
                                         ^~~~~ annoying space!
```

app/components/application_date_component.html.erb
tvararu: Yeah, we thought the same thing when we did it. Helpers and components fill in the same role. We thought we would try it out in this case to see how it works.

It's slightly DRYer than the equivalent helper implementation, which would be like:

```ruby
def submitted_at_date
  dates = ApplicationDates.new(@application_form)
  dates.submitted_at.strftime('%e %B %Y')
end
```

Which is also less explicit because it doesn't require you to pass in an `@application_form`. I like the clear enforcing of passing in arguments that you get with `ActionView::Component`.

tijmenb: I'm leaning that way too, it feels like https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/561/files#r347861143 is telling us that this is not an ideal use case for components.

tvararu: Cool, will refactor to helpers.

app/components/submitted_course_choices_component.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+private
+
+  attr_reader :application_form
+
+  def course_row(course_choice)
+    {
+      key: 'Course',
+      value: "#{course_choice.course.name} (#{course_choice.course.code})",
```
tijmenb:
```suggestion
      value: course_choice.course.name_and_code,
```

app/components/submitted_referees_component.rb
```diff
@@ -0,0 +1,40 @@
...
+  def initialize(application_form:)
+    @application_form = application_form
+  end
+
+  def referee_rows(work)
+    [
+      name_row(work),
+      email_row(work),
+      relationship_row(work),
+    ].compact
```
tijmenb: Nothing will return `false` from the rows, so I think we can drop this.

app/components/application_date_component.html.erb
tvararu: Updated!

PR #562, opened 2019-11-18
Fake DfE Sign-in in development

duncanjbrown:
Use the Omniauth "Developer" strategy to provide a fake DfE Sign in interface so we don't depend on it for development. You can provide a UID and an email of your choosing on Sign-in and the system will act as though those details came back from DSI after a successful authentication.

Double-lock on the feature: it works if and only if an env var is set and we're in development mode. The env var is not added to the Azure config so it will be impossible to enable it in prod.

No test included as Omniauth uses Rack middleware for implementation, the initializer runs before RSpec, and swapping middleware at runtime requires unpleasant contortions. Not a public-facing feature, and if it breaks we'll know about it, so I held off.

### DIFF
--- a/config/initializers/omniauth.rb
+++ b/config/initializers/omniauth.rb
@@ -22,4 +22,12 @@ options = {
   },
 }
 
-Rails.application.config.middleware.use OmniAuth::Strategies::OpenIDConnect, options
+if DfESignIn.bypass?
+  Rails.application.config.middleware.use OmniAuth::Builder do
+    provider :developer,
+             fields: %i[uid email],
+             uid_field: :uid
+  end
+else
+  Rails.application.config.middleware.use OmniAuth::Strategies::OpenIDConnect, options
+end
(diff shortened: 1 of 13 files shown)


config/initializers/omniauth.rb
```diff
@@ -22,4 +22,12 @@
   },
 }
 
-Rails.application.config.middleware.use OmniAuth::Strategies::OpenIDConnect, options
+if DfESignIn.bypass?
+  Rails.application.config.middleware.use OmniAuth::Builder do
```
mnacos: Is this what provides the view/form?

duncanjbrown:
Good question about provider auditing: we talked about it re DSI and it's to come. Have added a placeholder https://trello.com/c/FhGp8ffU/1277-audit-provider-user-actions

config/initializers/omniauth.rb
duncanjbrown: Yes, it's magic 🎩

PR #563, opened 2019-11-19
Fix failing tests on master

tijmenb:
This fixes one consistently failing test and one intermittently failing test on master. 

See commits for details.

### DIFF
--- a/spec/mailers/referee_mailer_spec.rb
+++ b/spec/mailers/referee_mailer_spec.rb
@@ -4,7 +4,7 @@ RSpec.describe RefereeMailer, type: :mailer do
   subject(:mailer) { described_class }
 
   describe 'Send request reference email' do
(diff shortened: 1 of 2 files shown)


spec/mailers/referee_mailer_spec.rb
```diff
@@ -4,7 +4,7 @@
   subject(:mailer) { described_class }
 
   describe 'Send request reference email' do
-    let(:application_form) { create(:completed_application_form) }
+    let(:application_form) { create(:completed_application_form, first_name: 'Harry', last_name: 'Potter') }
```
stevehook: I saw this one fail late yesterday when there was an apostrophe in the generated name - is that the same issue?

stevehook: I saw this one fail late yesterday when there was an apostrophe in the generated name.

PR #564, opened 2019-11-19
Handle internal error in POST requests

tijmenb:
### Context

Currently, when an exception occurs we're not seeing the proper error page, but a blank page instead. This is because we haven't configured the 500 page to show for POST requests (only GET).

Adds request routes for all HTTP methods. I've taken this from https://github.com/alphagov/content-publisher/commit/d97db8f4f9bcfcb9def8d4c4a1217e3ceac99d02, because when in doubt you should always copy @kevindew's code.

### Guidance to review

Write a bug, deploy to QA, observe the error.

### Link to Trello card

https://trello.com/c/YtFlMRNo

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -237,7 +237,11 @@ Rails.application.routes.draw do
   end
 
   get '/check', to: 'healthcheck#show'
-  get '/404', to: 'errors#not_found'
-  get '/500', to: 'errors#internal_server_error'
+
+  scope via: :all do
+    match '/404', to: 'errors#not_found'
+    match '/500', to: 'errors#internal_server_error'
+  end
+
   get '*path', to: 'errors#not_found'
 end


duncanjbrown:
Per the (excellent) original commit message, does this also need a `skip_before_action :verify_authenticity_token` in the `ErrorsController`? Hitting reload on a POSTed error might have unexpected results otherwise.

tijmenb:
@duncanjbrown I wondered about that, but I *think* we're alright because we never use `protect_from_forgery` in the `ErrorsController` or `ApplicationController`. Content Publisher does do that, so it needs it.

duncanjbrown:
> @duncanjbrown I wondered about that, but I _think_ we're alright because we never use `protect_from_forgery` in the `ErrorsController` or `ApplicationController`. Content Publisher does do that, so it needs it.

Is there any reason not to have it on `ApplicationController`?

tijmenb:
@duncanjbrown We don't use ApplicationController for anything, except content and errors.

PR #565, opened 2019-11-19
Do not 500 when Notify errors with a 400

duncanjbrown:
Instead report the error to Sentry and continue as though nothing
happened.

Handles `BadRequestError`s, ie:

- "BadRequestError: Can't send to this recipient using a team-only API key"
- "BadRequestError: Can't send to this recipient when service is in trial mode - see https://www.notifications.service.gov.uk/trial-mode"

https://docs.notifications.service.gov.uk/ruby.html#error-codes

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -1,3 +1,9 @@
 class ApplicationMailer < Mail::Notify::Mailer
+  rescue_from Notifications::Client::BadRequestError, with: :report_notify_error
+
+  def report_notify_error(e)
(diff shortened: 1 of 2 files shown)


duncanjbrown:
We're not sure this is the right approach, closing for now

tijmenb:
So my hesitation is that if something goes wrong, eg with the access token, we won't be sending out signup emails, but we're telling users that we _have_ sent the email. That could be frustrating. What's more, we don't have a way of fixing the issue atm - we can't fix the bug and trigger an email that the user requested (which would be odd anyway).

PR #573, opened 2019-11-19
Add email validation to references

dankmitchell:
### Context
Validate email format for references

Add rails email validation for references

### Guidance to review
Add a new reference with an invalid email address

### Link to Trello card
[392 - No email validation on 'Reference'](https://trello.com/c/Uz2Lohwl/392-no-email-validation-on-reference)

### Env vars
n/a

### DIFF
--- a/app/models/reference.rb
+++ b/app/models/reference.rb
@@ -1,6 +1,7 @@
 class Reference < ApplicationRecord
   validates :name, presence: true, length: { minimum: 2, maximum: 200 }
   validates :email_address, presence: true,
+                            email_address: true,
                             length: { maximum: 100 },
                             uniqueness: { scope: :application_form_id }
   validates :relationship, presence: true, length: { minimum: 2, maximum: 500 }
--- a/app/validators/email_address_validator.rb
+++ b/app/validators/email_address_validator.rb
@@ -1,0 +1,7 @@
+class EmailAddressValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    if value.blank? || !value.match?(/@/)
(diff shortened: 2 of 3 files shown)


spec/system/candidate_interface/candidate_adding_referees_spec.rb
```diff
@@ -28,6 +29,24 @@
     i_see_referees_is_complete
   end
 
+  scenario 'Candidate enters an invalid email' do
```
tijmenb: Perhaps we can remove this scenario. We have a unit test for it, and the accepatance tests work best if there's a single scenario per file (otherwise they become long and hard to read).

spec/models/reference_spec.rb
```diff
@@ -17,6 +17,10 @@
 
     it { is_expected.to validate_presence_of :relationship }
     it { is_expected.to validate_length_of(:relationship).is_at_most(500) }
+
+    it 'is invalid when the email address format is incorrect' do
```
tijmenb: I'm assuming that other tests would fail if the email validation was too strict?

dankmitchell:
@tijmenb had a re-think this PR and stripped it back to match the other custom validators. I think this is enough to merge now.

spec/system/candidate_interface/candidate_adding_referees_spec.rb
dankmitchell: Removed

spec/models/reference_spec.rb
dankmitchell: Removed

PR #587, opened 2019-11-19
Remove the provider question from the eligibility page

tijmenb:
### Context

> Candidates may not know the name of providers, so it might put potential candidates off the service. It would be better to wait until we can build in the course selection feature pre-application. 

- Remove the question
- Add validation

### Guidance to review

See the code.

https://trello.com/c/G4wVhS9J


spec/system/candidate_interface/candidate_eligibility_spec.rb
```diff
@@ -16,10 +18,17 @@ def when_i_click_start_on_the_start_page
     click_on t('application_form.begin_button')
   end
 
+  def and_i_press_continue
+    click_on 'Continue'
+  end
+
+  def then_i_see_validation_errors
+    expect(page).to have_content 'Choose if you are a citizen of the UK, EU or EEA'
```
tvararu: Nit, can merge without

```suggestion
    expect(page).to have_content t('activemodel.errors.models.candidate_interface/eligibility_form.attributes.eligible_citizen.blank')
```

(please double check I typed it in correctly)

PR #589, opened 2019-11-19
Add guidance and hint text for GCSE qualifications

paulrobertlloyd:
### Context

Adds guidance and hint text to GCSE qualification (grade and year) pages. This guidance varies depending on subject and/or qualification type. Also updates the qualification validation message on the previous page, which previously mentioned ‘degree’. 

Here’s an example page:

Screenshot

### Guidance to review

### Link to Trello card

[415 - Add guidance and hint text for GCSE grades/subjects](415-add-guidance-and-hint-text-for-gcse-grades-subjects)

### DIFF
--- a/app/helpers/gcse_qualification_helper.rb
+++ b/app/helpers/gcse_qualification_helper.rb
@@ -8,4 +8,17 @@ module GcseQualificationHelper
   def conditional_radios_for_gcse_qualifications
     { other_uk: [other_uk_qualification_type: 'Enter type of qualification'] }
   end
+
+  def guidance_for_gcse_edit_details(subject, qualification_type)
+    if I18n.exists?("gcse_edit_details.guidance.#{subject}.#{qualification_type}")
+      tag.p(t("gcse_edit_details.guidance.#{subject}.#{qualification_type}"), class: 'govuk-body')
+    end
+  end
+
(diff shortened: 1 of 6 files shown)


app/helpers/gcse_qualification_helper.rb
```diff
@@ -17,6 +17,19 @@ def heading_for_gcse_edit_details(subject)
     t("gcse_edit_details.heading.#{subject}")
   end
 
+  def guidance_for_gcse_edit_details(subject, qualification_type)
+    if I18n.exists?("gcse_edit_details.guidance.#{subject}.#{qualification_type}")
+      tag.p(t("gcse_edit_details.guidance.#{subject}.#{qualification_type}"), class: 'govuk-body')
```
fofr: Nit: Seeing `tag.p` feels like a code smell, should this be in the view?

paulrobertlloyd: As the guidance (and this paragraph) only appears if there’s guidance, if this were in the view, there would need to be a lot of logic there to determine whether to output the HTML or not. As this is a helper function, seems like an okay place for this to live, bit happy to defer to those more familiar with Ruby and conventions used in this app.

tijmenb: It's a view helper, they're definitely fine to generate HTML 👍

PR #590, opened 2019-11-19
Redirect to dashboard if application is submitted

dankmitchell:
### Context
Don't allow completed applications to be edited i.e. going direct to a URL

Redirect all routes to dashboard if the application is `submitted?`

### Guidance to review
Complete an application then manually navigate to a route i.e. `/candidate/application/courses`

### Link to Trello card
[402 - For a submitted application, don't allow users to manually edit their application if they manually navigate to an edit page](https://trello.com/c/SfoKiwYW/402-for-a-submitted-application-dont-allow-users-to-manually-edit-their-application-if-they-manually-navigate-to-an-edit-page?menu=filter&filter=label:Dev)

### Env vars
n/a

### DIFF
--- a/app/controllers/candidate_interface/application_form_controller.rb
+++ b/app/controllers/candidate_interface/application_form_controller.rb
@@ -1,20 +1,18 @@
 module CandidateInterface
   class ApplicationFormController < CandidateInterfaceController
+    before_action :redirect_to_dashboard_if_submitted, only: %i[show review]
+
     def show
       return redirect_to candidate_interface_application_form_path if params[:token]
 
-      if current_application.submitted?
-        @application_form = current_application
-
-        render :complete
-      else
-        @application_form_presenter = CandidateInterface::ApplicationFormPresenter.new(current_application)
-
-        render :show
-      end
+      @application_form_presenter = CandidateInterface::ApplicationFormPresenter.new(current_application)
     end
 
     def review
+      @application_form = current_application
+    end
+
+    def complete
       @application_form = current_application
     end
 
--- a/app/controllers/candidate_interface/candidate_interface_controller.rb
+++ b/app/controllers/candidate_interface/candidate_interface_controller.rb
@@ -10,6 +10,10 @@ module CandidateInterface
 
   private
 
+    def redirect_to_dashboard_if_submitted
+      redirect_to candidate_interface_application_complete_path if current_application.submitted?
+    end
+
     # controller-specific additional info to include in logstash logs
     def add_identity_to_log
       RequestLocals.store[:identity] = { candidate_id: current_candidate&.id }
--- a/app/controllers/candidate_interface/contact_details/address_controller.rb
+++ b/app/controllers/candidate_interface/contact_details/address_controller.rb
@@ -1,5 +1,7 @@
 module CandidateInterface
   class ContactDetails::AddressController < CandidateInterfaceController
+    before_action :redirect_to_dashboard_if_submitted
+
     def edit
       @contact_details_form = ContactDetailsForm.build_from_application(
         current_application,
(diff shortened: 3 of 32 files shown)


app/controllers/candidate_interface/candidate_interface_controller.rb
```diff
@@ -7,9 +7,15 @@ class CandidateInterfaceController < ActionController::Base
     before_action :add_identity_to_log
     layout 'application'
     alias :audit_user :current_candidate
+    before_action :redirect_to_dashboard
 
   private
 
+    def redirect_to_dashboard
+      @application_form = current_application
+      render template: 'candidate_interface/application_form/complete' if current_application.submitted?
```
tvararu: The dashboard will link to other pages, like a new review component, and the pages for accepting/declining/calling up support to make a decision on a course choice. With the line above, a user will be redirected back to the main dashboard page when they go to one of those.

This may also prevent users from reaching candidate-specific static pages e.g. terms of use or stuff accessible from the footer.

I think a callback like this could be the right approach. It feels like we need some kind of namespacing, or whitelisting/blacklisting to determine which pages should redirect to the dashboard post-submission, and which pages should continue working the same (such as static pages). Those could work but I think they may be not very elegant / involve lots of brittle repetition.

Another consideration: we *will* actually open up editing eventually, but it needs to be allowed only after the application has been put into a specific state in order to comply with business rules.

Channeling @tijmenb if he has ideas on how to implement whitelisting of routes or some kind of post-submission namespace so that users can't go back to edit pages.

tvararu: Also, probably just a nit as will have been considered, but this should be a `redirect_to` instead of a `render`.

tijmenb: Probably the safest way is to add `before_action :redirect_to_dashboard_if_submitted` to all controllers with a edit interface. We'll just need to make sure we don't forget to add it to everything.

dankmitchell: Thanks @tijmenb @tvararu

dankmitchell: Sorted. More cleanup and test required but making progress

app/controllers/candidate_interface/contact_details/address_controller.rb
```diff
@@ -1,5 +1,7 @@
 module CandidateInterface
   class ContactDetails::AddressController < CandidateInterfaceController
+    before_action :redirect_to_dashboard_if_submitted, only: :edit
```
tijmenb: Would this allow a intrepid user to update their form by doing a manual POST request to `update`?

dankmitchell: Possibly, yes @tijmenb but...

```
- When a user has submitted their application
- And they manually navigate to an edit page, such as /candidate/application/personal-details/
- They should be redirected to the dashboard page
```

tijmenb: Yeah, I don't think the author of the card considered POST requests, sorry about hat! 

I think we should do all controller actions, because you could have people who have a form open in browser window 1, and the submit in 2, and then press submit in window 1. Their change would still be saved.

tvararu: Yep, POSTs should also be in scope for this, sorry for any confusion!

dankmitchell: No probs. It makes the code easier so no problem here

app/controllers/candidate_interface/application_form_controller.rb
```diff
@@ -1,12 +1,14 @@
 module CandidateInterface
   class ApplicationFormController < CandidateInterfaceController
+    before_action :redirect_to_dashboard_if_submitted, only: :review
```
tvararu:
```suggestion
    before_action :redirect_to_dashboard_if_submitted, only: %i[review submit_show submit submit_success]
```

Should we redirect these pages too?

dankmitchell: Infinite redirects unless we refactor the controller actions :)

PR #591, opened 2019-11-19
Refactor components to pass down editable argument to SummaryCard

chubberlisk:
### Context

I added a whole bunch of duplication when adding `editable` argument for components.

This PR refactors component to pass down `editable` to ~`SummaryListComponent`~ `SummaryCardComponent` so it handles whether or not editing links should be shown so we can remove the `if @editable` for each key and value pair for each row in each review component. 😅 

### Guidance to review

Does this make sense? Is this a better solution?

### Link to Trello card

N/A


fofr:
Rather than passing a flag onwards to SummaryListComponent, can we instead massage the input in the SummaryCardComponent?

eg If not editable, maps the rows passed in to remove action, change_path and action_path keys?

This will keep SummaryListComponent simpler, aligned with design system component, and as it's intended for wider use (eg in provider UI).

app/components/summary_card_component.rb
```diff
@@ -13,4 +13,15 @@ def border_css_class
 private
 
   attr_reader :rows
+
+  def determine_rows(rows, editable)
+    rows.map do |row|
+      row.tap do |r|
```
fofr: What's the `tap` doing here?

chubberlisk: `.tap` pretty cool. In this case, it returns the object `r` at the end so we don't have to do an explicit return of it.

See https://apidock.com/ruby/Object/tap.

chubberlisk:
Following a conversation with @fofr, it is unnecessary to have a `deletable` parameter. As if a section is editable, all things can be changes (or deleted).

PR #602, opened 2019-11-20
Remove whitespacing for award year of a degree

chubberlisk:
### Context

If candidates enter " 2020" this is not accepted. We should allow this input and removing the whitespacing before and after if any.

This PR adds a `.strip` when validating and saving the award year in `DegreeForm`.

### Guidance to review

Does it make sense?

### Link to Trello card

[408 - Update saving graduation year in degrees to remove whitespacing](https://trello.com/c/zUZ966Vg/408-update-saving-graduation-year-in-degrees-to-remove-whitespacing)

### DIFF
--- a/spec/system/candidate_interface/candidate_entering_degrees_spec.rb
+++ b/spec/system/candidate_interface/candidate_entering_degrees_spec.rb
@@ -86,7 +86,8 @@ RSpec.feature 'Entering their degrees' do
 
     choose t('application_form.degree.grade.first.label')
 
-    fill_in t('application_form.degree.award_year.label'), with: '2009'
+    year_with_preceding_space = ' 2009'
+    fill_in t('application_form.degree.award_year.label'), with: year_with_preceding_space
   end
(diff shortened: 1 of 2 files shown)


spec/system/candidate_interface/candidate_entering_degrees_spec.rb
```diff
@@ -86,7 +86,7 @@ def when_i_fill_in_my_undergraduate_degree
 
     choose t('application_form.degree.grade.first.label')
 
-    fill_in t('application_form.degree.award_year.label'), with: '2009'
+    fill_in t('application_form.degree.award_year.label'), with: ' 2009'
```
fofr: Does this space need a comment, to avoid future developers deleting?
Or a variable name "year_with_preceding_space"

chubberlisk: Yeah, that makes sense! Will update this.

PR #607, opened 2019-11-20
Allow ProviderUsers to sign out

duncanjbrown:
Only GET, as the benefits of DELETE (HTTP semantics, CSRF protection) aren't critical and they require `rails-ujs` which is a bit heavy for all that

Screenshot 2019-11-20 at 12 25 30

### DIFF
--- a/app/controllers/provider_interface/sessions_controller.rb
+++ b/app/controllers/provider_interface/sessions_controller.rb
@@ -12,6 +12,12 @@ module ProviderInterface
       redirect_to provider_interface_path
     end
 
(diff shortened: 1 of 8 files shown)


duncanjbrown:
@tijmenb I hadn't considered that. I have https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/commit/852b59f5337c3a442be845b95e334cb2464c6265 up my sleeve — shall I add it?

tijmenb:
@duncanjbrown yeah, that looks good!

PR #612, opened 2019-11-20
Revert "Add rails-ujs to support DELETE links"

duncanjbrown:
This reverts commit 1d00041361820fb99d9e7494b7eb2681a8cb85ae.

Following deployment to QA candidate sign in was broken in Safari. Removing `Rails.start()` from the js file seems to fix it. UJS is not a hard requirement, so we are rolling it back to fix the build.


tvararu:
Merging this as it's a priority to fix QA!

PR #613, opened 2019-11-20
Fix template error caused by job with `nil` `end_date`

stevehook:
### Context

The trigger for this change was the following error on `qa`:

https://sentry.io/organizations/dfe-bat/issues/1343572021/?project=1765973&referrer=slack

It happens because the code that evaluates whether a candidate has gaps in work experience doesn't handle missing end dates except on the last job (ordered by start date). That assumption makes sense until you consider overlapping (part-time?) jobs.

- [x] Handle `nil` `end_date` values for jobs that are not the most recently started.
- [x] Handle other cases of overlapping jobs.

### Guidance to review

- I'm making an assumption here that we permit overlapping jobs, including jobs that have not ended yet. Make sense?
- Do the specs cover enough of the various permutations?
- Does the modified logic make sense?

### Link to Trello card

[409 - Investigate ActionView::Template::Error in candidate work experience form](https://trello.com/c/oTji12v5/409-investigate-https-sentryio-organizations-dfe-bat-issues-1343572021-project1765973queryis3aunresolved)

### Env vars

no env vars added

### DIFF
--- a/app/services/check_breaks_in_work_history.rb
+++ b/app/services/check_breaks_in_work_history.rb
@@ -4,16 +4,15 @@ class CheckBreaksInWorkHistory
       jobs = application_form.application_work_experiences.sort_by(&:start_date)
 
       return false if jobs.empty?
-      return break_between_only_job_and_current_date?(jobs.first) if jobs.one?
 
       jobs.each_cons(2).any? do |first_job, next_job|
         month_or_more_break_between?(first_job, next_job)
-      end
+      end || break_between_job_and_current_date?(jobs.last)
     end
 
   private
 
-    def break_between_only_job_and_current_date?(job)
+    def break_between_job_and_current_date?(job)
       if current_role?(job)
         false
       else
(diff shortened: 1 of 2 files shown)


app/services/check_breaks_in_work_history.rb
```diff
@@ -3,34 +3,33 @@ class << self
...
-      return break_between_only_job_and_current_date?(jobs.first) if jobs.one?
+      previous_end_date = nil
+      jobs.each do |job|
+        return true if month_or_more_break_between_dates?(previous_end_date, job.start_date)
 
-      jobs.each_cons(2).any? do |first_job, next_job|
-        month_or_more_break_between?(first_job, next_job)
+        return false if less_than_month_to_current_date?(job.end_date)
+
+        previous_end_date = [previous_end_date, job.end_date].compact.max
```
tvararu: Using the new helper methods, this reads really well! :+1:

app/services/check_breaks_in_work_history.rb
```diff
@@ -3,34 +3,33 @@ class << self
     def call(application_form)
       jobs = application_form.application_work_experiences.sort_by(&:start_date)
 
-      return false if jobs.empty?
-      return break_between_only_job_and_current_date?(jobs.first) if jobs.one?
+      previous_end_date = nil
```
tvararu:
```suggestion
      latest_end_date = nil
```

Is this a more intuitive name? Minor

PR #617, opened 2019-11-20
Remove OGL licence from footer, update footer styles

paulrobertlloyd:
### Context

Removes OGL licence text from footer (not relevant for this service), and tweaks footer styles to fix spacing and line lengths.

Before:

Screenshot 2019-11-20 at 16 39 18

After:

Screenshot 2019-11-20 at 16 46 26

### DIFF
--- a/app/frontend/styles/_footer.scss
+++ b/app/frontend/styles/_footer.scss
@@ -1,0 +1,13 @@
+.govuk-footer--app {
+  .govuk-footer__meta-custom {
+    max-width: 40rem;
+  }
+
+  .govuk-footer__inline-list {
+    margin-bottom: 0;
+  }
+
+  .govuk-footer__inline-list-item {
+    margin-bottom: 0;
+  }
+}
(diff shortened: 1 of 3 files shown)


app/frontend/styles/application.scss
```diff
@@ -10,12 +10,13 @@ $govuk-font-url-function: frontend-font-url;
...
-@import "_task-list";
-@import "_summary-card";
 @import "_banner";
+@import "_environments";
+@import "_footer";
 @import "_status-box";
+@import "_task-list";
 @import "_styled-content";
-@import "_environments";
+@import "_summary-card";
```
paulrobertlloyd: Alphabetically ordered these imports.

app/frontend/styles/_footer.scss
```diff
@@ -0,0 +1,13 @@
+.govuk-footer--app {
```
tvararu: Is `.app-govuk-footer` closer to what this is, an override in our app for the default footer styling?

paulrobertlloyd: I went with this given BEM convention and that we are modifying an existing component, rather than creating an entirely new one.

tvararu: 👌

paulrobertlloyd: I went with this given BEM convention and that we are modifying an existing component, rather than creating an entirely new one.

PR #627, opened 2019-11-21
Content check of build 4

laurahermionetennant92:
Content check of build 4

### Context

Why are you making this change? What might surprise someone about it?

If there are UI changes, please include Before and After screenshots.

### Guidance to review

### Link to Trello card

[123 - Example Trello card](http://trello.com/123-example-card)

### Env vars

### DIFF
--- a/app/views/candidate_interface/application_form/submit_success.html.erb
+++ b/app/views/candidate_interface/application_form/submit_success.html.erb
@@ -14,17 +14,17 @@
   <div class="govuk-grid-column-two-thirds">
     <h2 class="govuk-heading-m">What happens next</h2>
     <p class="govuk-body">
-      You will receive an email from the Apply for postgraduate teacher training service to confirm the successful submission of your application.
+      The Apply for teacher training service will send you an email to confirm the successful submission of your application.
     </p>
 
     <h2 class="govuk-heading-m">References</h2>
-    <p class="govuk-body">We will also contact your referees to ask for your references. When we receive them, we will send them to your training provider. Your provider then has 10 working days to respond to your application.</p>
+    <p class="govuk-body">We will also contact your referees to ask for your references. When we receive them, we will send them to your training provider. Your provider then has 40 working days to respond to your application.</p>
 
     <h2 class="govuk-heading-m">Interview</h2>
     <p class="govuk-body">If your training provider decides to offer you an interview, they will get in touch directly via email to organise a date and give you all the information you need.</p>
-
-    <p class="govuk-body">
-      <%= govuk_link_to t('submit_application_success.track_your_application'), candidate_interface_application_form_path %>.
+    <h2 class="govuk-heading-m">Track, edit or withdraw</h2>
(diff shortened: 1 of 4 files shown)


app/views/candidate_interface/application_form/submit_success.html.erb
```diff
@@ -14,17 +14,17 @@
   <div class="govuk-grid-column-two-thirds">
     <h2 class="govuk-heading-m">What happens next</h2>
     <p class="govuk-body">
-      You will receive an email from the Apply for postgraduate teacher training service to confirm the successful submission of your application.
+      The Apply for teacher training service will send you an email to confirm the successful submission of your application.
```
fofr: This feels like a change from active voice to passive.
Maybe "We have sent you"? They should get the email almost immediately.

app/views/candidate_interface/application_form/submit_success.html.erb
```diff
@@ -14,17 +14,17 @@
...
     </p>
 
     <h2 class="govuk-heading-m">References</h2>
-    <p class="govuk-body">We will also contact your referees to ask for your references. When we receive them, we will send them to your training provider. Your provider then has 10 working days to respond to your application.</p>
+    <p class="govuk-body">We will also contact your referees to ask for your references. When we receive them, we will send them to your training provider. Your provider then has 40 working days to respond to your application.</p>
```
fofr: Yikes. Good catch.

fofr: @tijmenb Could we derive this number from business rules in code rather than free text?

tvararu:
```suggestion
    <p class="govuk-body">We will also contact your referees to ask for your references. When we receive them, we will send them to your training provider. Your provider then has <%= TimeLimitConfig.limits_for(:reject_by_default).first.limit %> working days to respond to your application.</p>
```

Not sure if there is a cleaner way. 🤔

duncanjbrown: Could be calculated in the controller?

```
    def submit_success
      @support_reference = current_application.support_reference
      @days_to_reject_by_default = TimeLimitCalculator.new(rule: :reject_by_default, effective_date: current_application.submitted_at).call
    end
```

tvararu: @duncanjbrown yes

duncanjbrown: Added a card for this so we can merge the present PR https://trello.com/c/8QUESpBF/533-we-hardcode-40-days-when-telling-candidates-about-reject-by-default

paulrobertlloyd: Filed in https://trello.com/c/8QUESpBF/533-we-hardcode-40-days-when-telling-candidates-about-reject-by-default

PR #628, opened 2019-11-21
Add referee chaser email, update request email

paulrobertlloyd:
### Context

* Updates email to referee requesting a reference
* Adds an email to send to referees after 5 days, reminding them to submit a reference. **This will be manually sent from support console, functionality that is not built yet**

### Link to Trello card

[416 - Email: Initial email to referee](https://trello.com/c/U6p4SPID/)
[417 - Email: Chase email to referee](https://trello.com/c/ffjNEQHc/)

### DIFF
--- a/app/mailers/referee_mailer.rb
+++ b/app/mailers/referee_mailer.rb
@@ -1,12 +1,24 @@
 class RefereeMailer < ApplicationMailer
   def reference_request_email(application_form, reference)
+    @application_form = application_form
     @reference = reference
     @candidate_name = "#{application_form.first_name} #{application_form.last_name}"
     @google_form_url = google_form_url_for(@candidate_name, @reference)
 
     view_mail(GENERIC_NOTIFY_TEMPLATE,
               to: reference.email_address,
-              subject: t('reference_request.email.subject', candidate_name: @candidate_name))
+              subject: t('reference_request.subject.initial', candidate_name: @candidate_name))
+  end
+
+  def reference_request_chaser_email(application_form, reference)
+    @application_form = application_form
+    @reference = reference
+    @candidate_name = "#{application_form.first_name} #{application_form.last_name}"
+    @google_form_url = google_form_url_for(@candidate_name, @reference)
+
+    view_mail(GENERIC_NOTIFY_TEMPLATE,
+              to: reference.email_address,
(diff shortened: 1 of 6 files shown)


app/mailers/referee_mailer.rb
```diff
@@ -1,12 +1,24 @@
 class RefereeMailer < ApplicationMailer
   def reference_request_email(application_form, reference)
+    @application_form = application_form
     @reference = reference
-    @candidate_name = "#{application_form.first_name} #{application_form.last_name}"
+    @candidate_name = "#{@application_form.first_name} #{@application_form.last_name}"
```
fofr: Not sure this line needs to change.

config/locales/reference_request.yml
```diff
@@ -1,7 +1,8 @@
 en:
   reference_request:
-    email:
-      subject: Please give a reference to support the teacher training application of %{candidate_name}
+    subject:
+      initial: Please give a reference to support the teacher training application of %{candidate_name}
```
fofr: Do we need Please here if we dropped it from the opening heading? Subject is quite long.

spec/mailers/referee_mailer_spec.rb
```diff
@@ -12,11 +12,60 @@
...
-      expect(mail.subject).to include(t('reference_request.email.subject', candidate_name: candidate_name))
+      expect(mail.subject).to include(t('reference_request.subject.initial', candidate_name: candidate_name))
     end
 
     it 'sends an email with the correct heading' do
-      expect(mail.body.encoded).to include("give a reference for #{candidate_name}")
+      expect(mail.body.encoded).to include("Give a reference for #{candidate_name}")
+    end
+
+    it 'sends an email with the correct courses listed' do
```
paulrobertlloyd: @tvararu @tijmenb We’ve added a list of selected courses to the reference request email, but @fofr was suggesting we test that this list appears and is correct. (We currently don’t test this in the application submitted email, either). I’ve replicated the code in the test, which is not a great test. How do you suggest I test this?

tvararu: Suggested offline: we can specify the choices explicitly in the setup for this test and then assert explicitly, without an `.each` loop.

paulrobertlloyd: Updated with suggestion. Thanks @tvararu!

PR #632, opened 2019-11-21
Add qualifications to api response

RobertM905:
### Context

Currently the qualifications are hardcoded in the SingleApplicationPresenter.

Update SingleApplicationPresenter to remove hardcoded qualifications and use application_qualifications from the application_form.

Update presenter spec to assert on application_qualifications.

Update `vendor-api-0.8.0` spec to include `awarding_body` field **(This may need be to be discussed)**

### Guidance to review

Run specs and ensure that the qualifications are displayed correctly.
### Link to Trello card

[1250 - Return real data within the api response](https://trello.com/c/S43gJ9xx/1250-return-real-data-within-the-api-response)

### Env vars

N/A


app/presenters/vendor_api/single_application_presenter.rb
```diff
@@ -186,5 +147,31 @@ def experience_to_hash(experience)
...
+        degrees: application_form.application_qualifications.degrees.map do |degree| qualification_to_hash(degree) end,
+        other_qualifications: application_form.application_qualifications.other.map do |other| qualification_to_hash(other) end,
+      }
+    end
+
+    def qualification_to_hash(qualification)
+      {
+        qualification_type: qualification.qualification_type,
+        subject: qualification.subject,
+        grade: "#{qualification.grade}#{' (Predicted)' if qualification.predicted_grade}",
```
stevehook: should this evaluate to `nil` if `qualification_grade` is `nil` (I think it will evaluate to empty string)? Would make it more consistent with other fields like `institution_details`

app/presenters/vendor_api/single_application_presenter.rb
```diff
@@ -186,5 +147,31 @@ def experience_to_hash(experience)
...
+        award_year: qualification.award_year,
```
stevehook: the DB column is nullable but the JSON schema field is not (thought that wasn't part of this PR). Same applies to `qualification_type` and `subject`.

duncanjbrown: good catch — one for the backlog?

spec/presenters/vendor_api/single_application_presenter_spec.rb
```diff
@@ -13,4 +13,51 @@
       expect(single_application_presenter.as_json[:attributes][:candidate][:nationality]).to eq(%w[GB US])
     end
   end
+
+  describe 'qualifications' do
+    it 'returns qualifications in the correct format' do
```
duncanjbrown: I'm not convinced this spec does anything that the end-to-end spec doesn't

duncanjbrown:
@stevehook as @RobertM905 is off, would you mind making any necessary changes for approval directly on this branch?

app/presenters/vendor_api/single_application_presenter.rb
RobertM905: Do we expect `qualification_grade` to be nil when the application is viewable to providers?

PR #633, opened 2019-11-21
Fix English and other language details fields

stevehook:
### Context

This was reported as a bug - see Trello card linked below. The candidate interface was storing details about other languages spoken in the `english_language_details` attribute and details about English qualifications in the `other_language_details` attribute. They should be the other way round.

- [x] Use `english_language_details` to store details about English qualifications for candidates whose first language is not English.
- [x] Use `other_language_details` to store details about other languages for candidates whose first language is English.
- [x] Check the Provider interface (manual testing as well as automated tests)
- [x] Check the Vendor API (via tests)
- [x] Manually check the candidate interface

english-language

### Guidance to review

- Are there any areas that I've missed?
- There is an assumption here that we don't have an live applications in production that could be wrong so there is no data fixup migration etc.

### Link to Trello card

[1284 - English language details fields are named the wrong way around](https://trello.com/c/NKzxKYR6/1284-english-language-details-fields-are-named-the-wrong-way-around)

### Env vars

- no env vars

### DIFF
--- a/app/components/provider_interface/language_skills_component.rb
+++ b/app/components/provider_interface/language_skills_component.rb
@@ -34,14 +34,14 @@ module ProviderInterface
     def english_main_language_details_row
       {
         key: t('application_form.personal_details.other_language_details.label'),
-        value: english_language_details.present? ? english_language_details : 'No details given',
+        value: other_language_details.present? ? other_language_details : 'No details given',
       }
     end
 
     def other_language_details_row
       {
         key: t('application_form.personal_details.other_language_details.label'),
-        value: other_language_details.present? ? other_language_details : 'No details given',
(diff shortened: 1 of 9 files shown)


app/components/provider_interface/language_skills_component.rb
```diff
@@ -34,14 +34,14 @@ def language_details_row
     def english_main_language_details_row
       {
         key: t('application_form.personal_details.english_main_language_details.label'),
```
fofr: Looks like these needed updating too?

stevehook: @fofr I must admit I find this a bit confusing but it looks right to me - the label here is _Other languages spoken_ which works if English main language is 'Yes'. I checked it locally and I see this:

image

PR #634, opened 2019-11-21
Render the Apply from Find page from the database

tijmenb:
### Context

We currently make a API call to Find for the course information. Now that we have it in the database this is no longer necessary. Using the data in the database will allow us to build the functionality for dual running.

No visible changes.

### Guidance to review

N/A

### Link to Trello card

https://trello.com/c/5wXHsiOC/203-give-candidates-the-option-to-select-u
cas-or-dfe-apply-routes


app/controllers/candidate_interface/apply_from_find_controller.rb
```diff
@@ -6,18 +6,12 @@ class ApplyFromFindController < CandidateInterfaceController
...
-
-      course = FindAPI::Course.fetch(provider_code, course_code)
-
-      if course.nil?
-        render_not_found
-      else
-        @course = CoursePresenter.new course
-      end
+      provider = Provider.find_by!(code: params.fetch(:providerCode))
+      course = provider.courses.where(exposed_in_find: true).find_by!(code: params.fetch(:courseCode))
```
tvararu: `.where(exposed_in_find: true)` feels like it could be a `.findable` scope, feels like we may reuse it. Minor

PR #639, opened 2019-11-21
See which courses are available on DFE Apply before starting

tijmenb:
### Context

This adds a page where candidates can see what courses are available on DfE Apply.

image

### Guidance to review

The tests are a bit weird - I'll make those better in https://trello.com/c/5wXHsiOC.

### Link to Trello card

https://trello.com/c/POKDlIlJ/283-see-which-courses-are-available-on-dfe-apply-before-starting

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -7,6 +7,8 @@ class Course < ApplicationRecord
   validates :level, presence: true
   validates :code, uniqueness: { scope: :provider_id }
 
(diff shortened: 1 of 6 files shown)


app/models/course.rb
```diff
@@ -7,6 +7,8 @@ class Course < ApplicationRecord
   validates :level, presence: true
   validates :code, uniqueness: { scope: :provider_id }
 
+  scope :visible_to_candidates, -> { where(exposed_in_find: true, open_on_apply: true) }
```
tvararu: 🎉

PR #642, opened 2019-11-21
Show missing sections in candidate review and prevent submission with missing sections

tvararu:
### Context

We need to show candidates which sections still need to be completed, and we need to stop users from submitting when they haven't finished all sections.

Add a component and use it in every review component. Update end to end tests to check for it.

### Guidance to review

Review commit by commit.

Screenshot 2019-11-22 at 20 04 34
Screenshot 2019-11-22 at 20 05 07
Screenshot 2019-11-22 at 21 10 00

TODO (follow-up PRs?):

- Improve content design

### Link to Trello card

[351 - Check that all questions are answered before submit](https://trello.com/c/TYfYUDKt/351-check-that-all-questions-are-answered-before-submit)

### DIFF
--- a/app/components/contact_details_review_component.html.erb
+++ b/app/components/contact_details_review_component.html.erb
@@ -1,1 +1,9 @@
-<%= render(SummaryCardComponent, rows: contact_details_form_rows, editable: @editable) %>
+<% if show_missing_section? %>
+  <%= render(SectionMissingBannerComponent, text: 'Contact details not entered') do %>
+    <%= link_to candidate_interface_contact_details_edit_base_path, class: 'govuk-link' do %>
+      Complete <span class="govuk-visually-hidden">contact details</span> section
+    <% end %>
+  <% end %>
+<% else %>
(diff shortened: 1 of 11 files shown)


spec/system/candidate_interface/candidate_reviewing_application_spec.rb
```diff
@@ -4,9 +4,28 @@
...
+      volunteering
+      degrees
+      maths_gcse
+      english_gcse
+      science_gcse
+      becoming_a_teacher
+      subject_knowledge
+      interview_preferences
+      references
+    ].map { |section| check_and_fill_in_section_for(section) }
```
chubberlisk: Really like the thorough testing but this section here is a bit jarring when reading the scenario. Could we extract a method for this?

tvararu: Discussed offline, this is OK as the trade-off to doing that is that the fields get somewhat obfuscated.

tvararu:
Oops, botched force push, reverting

PR #644, opened 2019-11-21
Do not show all applications to providers not in the permissions mapping 

duncanjbrown:
Draft, existing specs still broken

Due to an `or` condition in `GetApplicationsForProvider`, when a provider is `nil` we return all applications. (or at least all of the ones without an accrediting provider).

- [x] raise an exception if this happens
- [x] show the user a page saying their account is being set up
- [x] send a sentry notification

Couple of spec refactors up front to smooth the road.

Screenshot 2019-11-21 at 16 46 24

### Guidance to review

commit by commit

https://trello.com/c/OTiDNrff/1295-provider-applications-are-not-scoped-on-the-provider-ui-you-can-see-everything

### DIFF
--- a/features/step_definitions/application_steps.rb
+++ b/features/step_definitions/application_steps.rb
@@ -4,7 +4,6 @@ Given(/an application choice has "(.*)" status/) do |original_application_status
   application_form = FactoryBot.create(:application_form)
   @application_choice = FactoryBot.create(
     :application_choice,
-    :single,
     application_form: application_form,
     status: original_application_status.gsub(' ', '_'),
   )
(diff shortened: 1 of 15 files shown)


features/step_definitions/application_steps.rb
```diff
@@ -4,7 +4,6 @@
   application_form = FactoryBot.create(:application_form)
   @application_choice = FactoryBot.create(
     :application_choice,
-    :single,
```
stevehook: :+1:

duncanjbrown:
> Looks all good to me. My only question would be whether it's possible to get into the authenticated state and not associated with a provider state for some other reason than it's in progress? Could a user be authenticated but never have access to a provider?
> 
> I've not done any manual testing by the way because I don't know how to reproduce locally.

Thank you.

That question occurred to me too, but right now I don't think there is another way to get into that state.

For local testing follow https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training#provider-permissions

PR #651, opened 2019-11-22
Add ability to view an offer for a course choice

chubberlisk:
### Context

Currently we have a link which shows when an offer has been made to a candidate but currently you can't see what that offer is.

This PR adds a page for viewing an offer for a course choice. It does so by adding a new component called `OfferReviewComponent` which uses `OfferConditionsReviewComponent`.

### Screenshots (updated 25 Nov, 11:20)

image

### Guidance to review

You can test this by using the provider interface to make an offer and then viewing it on the candidate side.

### Link to Trello card

[Candidates can receive and review an offer from the provider](https://trello.com/c/HhzbUnQO/360-candidates-can-receive-and-review-an-offer-from-the-provider)

### DIFF
--- a/app/views/candidate_interface/course_choices/offer.html.erb
+++ b/app/views/candidate_interface/course_choices/offer.html.erb
@@ -1,0 +1,40 @@
+<% content_for :title, t('page_titles.view_and_respond_to_offer') %>
+<% content_for :before_content, govuk_back_link_to(candidate_interface_application_complete_path, 'Back to application dashboard') %>
+
+<h1 class="govuk-heading-xl govuk-heading-xl">
+  <span class="govuk-caption-xl"><%= @course_choice.provider.name %> - <%= @course_choice.course_option.course.name %> (<%= @course_choice.course_option.course.code %>)</span>
+  <%= t('page_titles.view_and_respond_to_offer') %>
+</h1>
+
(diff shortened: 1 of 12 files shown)


app/views/candidate_interface/course_choices/offer.html.erb
```diff
@@ -0,0 +1,34 @@
...
+    <ul class="govuk-list govuk-list--bullet">
+      <li>you are under no obligation to accept or decline an offer at this stage</li>
+      <li>you can only accept one offer, across both Apply for teacher training and UCAS Teacher Training</li>
+    </ul>
+
+    <h2 class="govuk-heading-xl govuk-heading-m">
+      How do you want to respond to this offer?
+    </h2>
+
+    <p class="govuk-body">You can respond to offer to study:</p>
```
fofr: I don't think this line reads right?

chubberlisk: You're right @fofr! 🤦‍♀ Changed a few words:

image

How's this?

chubberlisk: Updated the content of this with @laurahermionetennant92 to:

image

tvararu: 🙌

PR #661, opened 2019-11-25
Fix bugs, issues and exceptions found in testing

tijmenb:
### Context

This fixes a number of bugs, issues and exceptions that were uncovered in (pen)testing.

### Changes proposed in this pull request

See individual commits.

### Guidance to review

N/A

### Link to Trello card

See individual commits for links to Trello & Sentry.

PR #663, opened 2019-11-25
Small support UI tweaks

fofr:
- Put email back into UI title, but not browser title following #661 
- Use a clearer reference status

### DIFF
--- a/app/views/support_interface/application_forms/show.html.erb
+++ b/app/views/support_interface/application_forms/show.html.erb
@@ -1,5 +1,4 @@
 <% content_for :browser_title, "Application ##{@application_form.id}" %>
-<% content_for :title, "Application ##{@application_form.id} - #{@application_form.first_name} #{@application_form.last_name}" %>
 
 <% content_for :before_content do %>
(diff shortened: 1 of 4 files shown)


app/views/support_interface/application_forms/show.html.erb
```diff
@@ -14,6 +13,11 @@
   </div>
 <% end %>
 
+<h1 class="govuk-heading-xl">
```
tijmenb: Can you add this to `content_for :title` for consistency?

fofr: You can't set `:title` and add a caption AFAIK?

tijmenb: This should work:

```rb
<% content_for :title do %>
  <span class="govuk-caption-xl">Application #<%= @application_form.id %></span>
  <%= @application_form.candidate.email_address %> 
<% end %>
```

fofr: Fixed in 39f29e7

PR #666, opened 2019-11-25
A single place for product names

paulrobertlloyd:
### Context

I found a stray occurrence of the old service name ‘Apply for postgraduate teacher training’, so in replacing that, made all the service names localisable strings and then reference them from the views.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -6,14 +6,14 @@ module ApplicationHelper
 
   def service_name
     case current_namespace
-    when 'provider_interface'
-      'Manage teacher training applications'
     when 'candidate_interface'
-      'Apply for teacher training'
+      t('service_name.apply')
+    when 'provider_interface'
+      t('service_name.manage')
     when 'support_interface'
-      'Support for Apply'
+      t('service_name.support')
     else
-      'Apply for teacher training'
+      t('service_name.apply')
     end
   end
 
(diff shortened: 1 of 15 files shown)


app/helpers/application_helper.rb
```diff
@@ -6,14 +6,14 @@ def browser_title
 
   def service_name
     case current_namespace
-    when 'provider_interface'
```
paulrobertlloyd: As well as referencing localised strings, alphabetised this list of services

app/views/authentication_mailer/sign_in_email.text.erb
```diff
@@ -1,5 +1,5 @@
 # <%= t('authentication.sign_in.email.subject') %>
 
-Click the link below to sign in to the Apply for teacher training service.
+Click the link below to sign in to the <%= t('service_name.apply') %> service.
```
paulrobertlloyd: Emails don’t have access to `service_name` variable, so we call translated string directly.

app/views/provider_interface/account_creation_in_progress.html.erb
```diff
@@ -6,13 +6,13 @@
       Account creation in progress
     </h1>
     <p class="govuk-body">
-    You are seeing this page because you have successfully signed in to the
-    Apply for teacher training service, but your user account has not yet been
-    associated with your training provider.
+      You are seeing this page because you have successfully signed in to the
+      <%= service_name %> service, but your user account has
```
paulrobertlloyd: This previously said ‘Apply for teacher training’, but by referencing `service_name`, should now correctly say ‘Manage teacher training applications`.

PR #667, opened 2019-11-25
Show error when not submitting a course

tijmenb:
Fixes an issue in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/p
ull/660, where the error wasn't shown on the form because a `ActiveRecord::RecordNotFound` found error was triggered by the `user_cant_apply_to_same_course_twice` validation, which shows a 404 page.


spec/system/candidate_interface/candidate_selecting_a_course_spec.rb
```diff
@@ -66,6 +70,15 @@ def and_i_choose_a_provider
     click_button 'Continue'
   end
 
+  def when_submit_without_choosing_a_course
+    click_button 'Continue'
+  end
+
+  def then_i_should_see_an_error
+    puts page.body
```
tvararu:
```suggestion
```

PR #676, opened 2019-11-25
Allow us to open and close the pilot with a feature flag

tijmenb:
### Context

We want to open and close the pilot using a feature flag. This will allow us to 1) launch without doing a deploy (which means we can roll back more easily), 2) stop the pilot when something goes wrong, 3) stop the pilot when there are 15 candidates.

Nov-25-2019 14-38-30

### Guidance to review

Does this make sense?

### Deployment

We need to remove the `BASIC_AUTH_USERNAME` and `BASIC_AUTH_PASSWORD` variables before deploying to production, otherwise the Apply from Find page will be behind auth.

### Link to Trello card

https://trello.com/c/9sTu3ie6/493-controlled-fall-back-page-pilot-feature-flag


spec/system/candidate_interface/candidate_account_spec.rb
```diff
@@ -36,6 +38,10 @@
     then_i_should_be_signed_out
   end
 
+  def given_the_pilot_is_open
+    FeatureFlag.activate('pilot_open')
```
duncanjbrown: Do we need to tear this down afterwards?

app/controllers/support_interface/feature_flags_controller.rb
```diff
@@ -0,0 +1,17 @@
+module SupportInterface
+  class FeatureFlagsController < SupportInterfaceController
+    def index; end
+
+    def activate
+      FeatureFlag.activate(params[:feature_name])
+      flash[:success] = "#{params[:feature_name].humanize} activated"
```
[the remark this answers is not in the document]
tijmenb: Done!

app/views/support_interface/feature_flags/index.html.erb
```diff
@@ -0,0 +1,33 @@
...
+          <%= feature_name.humanize %>
+        </td>
+
+        <td class="govuk-table__cell">
+          <%= FeatureFlag.active?(feature_name) ? 'Yes' : 'No' %>
+        </td>
+
+        <td class="govuk-table__cell">
+          <% if FeatureFlag.active?(feature_name) %>
+            <%= button_to "Deactivate #{feature_name.humanize}", support_interface_deactivate_feature_flag_path(feature_name), class: 'govuk-button' %>
```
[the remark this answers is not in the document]
tijmenb: I'd rather duplicate, makes it less likely we trigger something by accident. More generally it's an accesibility thing, similar to the "Change" links in the review components having hidden extra info like "Change contact details". Don't think it's worth it hiding them in the UI here.

spec/system/candidate_interface/candidate_account_spec.rb
tijmenb: Perhaps, but I thought it added overhead and all of the tests should do the correct setup themselves.

tijmenb:
> I think `candidate_submitting_application_spec.rb` shouldn't really be passing without the flag set but the login_as method supplied by Devise sidesteps the two controllers which do the checking.

I think it's allright. It tests the submitting of an application, which would still be possible when the pilot is closed. With the pilot disabled candidates can still sign in and make changes to the application, but not sign up anymore.

Gemfile
```diff
@@ -43,6 +43,9 @@ gem 'uk_postcode'
 gem 'business_time'
 gem 'holidays'
 
+# feature flags
+gem 'rollout'
```
tvararu: 👍

config/initializers/basic_auth.rb
```diff
@@ -1,6 +1,8 @@
 options = {
   ui_auth: {
-    enabled:  Rails.env.production? || ENV['BASIC_AUTH_ENABLE'] == 'true',
+    # The UI only has basic auth enabled if the environment variables
+    # are set. This allows us to disable auth on production.
+    enabled:  ENV['BASIC_AUTH_USERNAME'] && ENV['BASIC_AUTH_PASSWORD'],
```
tvararu: 👍

PR #683, opened 2019-11-25
Fix size, position and naming of tags

paulrobertlloyd:
### Context

Fixes size and spacing of tag component, so filed and unfilled occupy the same amount of space.

Screenshot 2019-11-25 at 15 26 50

### DIFF
--- a/app/components/task_list_item_component.html.erb
+++ b/app/components/task_list_item_component.html.erb
@@ -1,10 +1,10 @@
 <%= govuk_link_to text, path, class: 'app-task-list__task-name', 'aria-describedby': tag_id %>
 <% if completed %>
-  <strong class="govuk-tag app-task-list__task-completed" id=<%= tag_id %>>
+  <strong class="govuk-tag app-tag app-tag--complete" id=<%= tag_id %>>
     Completed
   </strong>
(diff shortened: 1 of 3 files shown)


app/frontend/styles/_tag.scss
```diff
@@ -37,6 +35,10 @@
   @include tag("turquoise", $filled: false);
 }
 
+.app-tag--incomplete {
```
fofr:
```
.app-tag--incomplete,
.app-tag--primary-unfilled {
  @include tag("blue", $filled: false);
}
```

app/components/task_list_item_component.html.erb
```diff
@@ -1,10 +1,10 @@
 <%= govuk_link_to text, path, class: 'app-task-list__task-name', 'aria-describedby': tag_id %>
 <% if completed %>
-  <strong class="govuk-tag app-task-list__task-completed" id=<%= tag_id %>>
+  <strong class="govuk-tag app-tag app-tag--complete" id=<%= tag_id %>>
```
fofr: We should update this to use the TagComponent, but maybe not in this PR.

PR #689, opened 2019-11-25
Only let ReceiveReference make one kind of state change

duncanjbrown:
Discovered while working on https://trello.com/c/jbybYv2c/423-email-email-to-candidate-when-application-is-sent-to-provider

When this object can make two kinds of state transitions, the ceremony
around each transition gets lost. In this case there is a Slack
notification which happens before a choice gets `send_to_provider!`
in SendApplicationChoiceToProvider, but doesn't happen here.

The cron will pick up this change within the hour and do the transition
properly.


duncanjbrown:
👌  Updated with the state taken out

duncanjbrown:
oh I broke cucumber 🥒

PR #701, opened 2019-11-26
Fix bug in references importer

tijmenb:
### Context

A bug in `ReceiveReference` made it so that `references_complete?` is never true when the references are updated:

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/b4c0b90808f407183186bc449183dcb56d08c4cd/app/services/receive_reference.rb#L20-L29

We think this is caused by the `@application_form.references` objects not being updated after the update. It's really weird because there's a explicit test for this:

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/b4c0b90808f407183186bc449183dcb56d08c4cd/spec/services/receive_reference_spec.rb#L23-L38

I can't explain why this test passes.

Initially I fixed this bug by adding `@application_form.reload` to `ReceiveReference`. This fixed the issue, but wouldn't fix recurrence in other places.

A more dependable fix is to use a new query to get the references that have feedback.

### Guidance to review

N/A

### Link to Trello card

https://trello.com/c/W0hxwS3A

### DIFF
--- a/app/models/application_form.rb
+++ b/app/models/application_form.rb
@@ -22,7 +22,7 @@ class ApplicationForm < ApplicationRecord
   end
 
   def references_complete?
-    references.select(&:complete?).count == MINIMUM_COMPLETE_REFERENCES
+    references.completed.count == MINIMUM_COMPLETE_REFERENCES
   end
 
   def qualification_in_subject(level, subject)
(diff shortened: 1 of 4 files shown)


fofr:
References have always behaved strangely in tests. I think @aldavidson tracked some of it down to a potential naming clash for `.references` on a Rails object.

PR #704, opened 2019-11-26
Update application_form.yml

emmajhf:
### Context

Why are you making this change? What might surprise someone about it?

We need to change the error message a user gets if they don't select a site.

### DIFF
--- a/config/locales/application_form.yml
+++ b/config/locales/application_form.yml
@@ -302,7 +302,7 @@ en:
         candidate_interface/pick_site_form:
           attributes:
             course_option_id:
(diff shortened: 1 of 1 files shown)


config/locales/application_form.yml
```diff
@@ -302,7 +302,7 @@ en:
         candidate_interface/pick_site_form:
           attributes:
             course_option_id:
-              blank: Select a site
+              blank: Select which location you're applying to
```
fofr:
```suggestion
              blank: Select which location you’re applying to
```

PR #705, opened 2019-11-26
Automatically detect weirdness in the system

tijmenb:

This adds a hourly check on the data to make sure the system is in a normal state. The first check would have caught the references bug fixed in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/701.

I envision this check being expanded in the future as we discover more situations that shouldn't happen, but have.

### Link to Trello card

https://trello.com/c/W0hxwS3A/1317-when-edited-by-time-is-updated-the-application-is-still-not-visible-in-the-provider-interface

### DIFF
--- a/app/workers/detect_invariants.rb
+++ b/app/workers/detect_invariants.rb
@@ -1,0 +1,27 @@
+# Detect state that *should* be impossible in the system and report them to Sentry
+class DetectInvariants
+  include Sidekiq::Worker
+
+  def perform
+    detect_application_choices_stuck_in_awaiting_references_state
+  end
+
+  def detect_application_choices_stuck_in_awaiting_references_state
+    # Application choices with completed feedback, but still awaiting references
+    choices_in_wrong_state = begin
+      ApplicationChoice.where(status: 'awaiting_references', application_form: ApplicationForm.includes(:references).select(&:references_complete?))
+    end
+
+    if choices_in_wrong_state.any?
+      message = <<~MSG
+        One or more application choices in `awaiting_references` state, but all feedback is collected:
+
+        #{choices_in_wrong_state.map(&:id).join('\n')}
+      MSG
+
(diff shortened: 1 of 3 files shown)


app/workers/detect_invariants.rb
```diff
@@ -0,0 +1,27 @@
+# Detect state that *should* be impossible in the system and report them to Sentry
+class DetectInvariants
+  include Sidekiq::Worker
+
+  def perform
+    detect_unprocessed_feedbacks
```
duncanjbrown: could this method have a better name? `detect_application_choices_stuck_in_awaiting_references_state`?

app/workers/detect_invariants.rb
```diff
@@ -0,0 +1,27 @@
...
+      One or more application choices in `awaiting_references` state, but all feedback is collected:
+
+      #{choices_in_wrong_state.map(&:id).join('\n')}
+      msg
+
+      Raven.capture_exception(WeirdSituationDetected.new(message))
+    end
+  end
+
+  class WeirdSituationDetected < StandardError; end
```
duncanjbrown: 👌

app/workers/detect_invariants.rb
```diff
@@ -0,0 +1,27 @@
...
+  include Sidekiq::Worker
+
+  def perform
+    detect_unprocessed_feedbacks
+  end
+
+  def detect_unprocessed_feedbacks
+    # Application choices with completed feedback, but still awaiting references
+    choices_in_wrong_state = begin
+      ApplicationChoice.where(status: 'awaiting_references', application_form: ApplicationForm.includes(:references).select(&:references_complete?))
```
duncanjbrown: This query feels a bit heavy but I can't improve on it

tijmenb: Yeah, I'm pretty sure we can do this in one SQL query, but I've forgotten how to SQL in the late eighties.

PR #706, opened 2019-11-26
Strip trailing spaces when submitting referees

tvararu:
### Context

Without this, our emails and various sections in the website can look a bit weird.

Add a `.transform_values` and a test.

### Guidance to review

Should we be doing this in a more generic way and reusing it on all the form components?

### Link to Trello card

[534 - Strip out spaces on referee name](https://trello.com/c/ewaXY108/534-strip-out-spaces-on-referee-name)

### DIFF
--- a/spec/system/candidate_interface/candidate_adding_referees_spec.rb
+++ b/spec/system/candidate_interface/candidate_adding_referees_spec.rb
@@ -90,7 +90,8 @@ RSpec.feature 'Candidate adding referees' do
   end
 
   def and_i_fill_in_all_required_fields
-    fill_in('Full name', with: 'Bill Lumbergh')
+    full_name_with_trailing_space = 'Bill Lumbergh '
+    fill_in('Full name', with: full_name_with_trailing_space)
     fill_in('Email address', with: 'lumbergh@example.com')
     fill_in(t('application_form.referees.relationship.label'), with: 'manager for several years')
   end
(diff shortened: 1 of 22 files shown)


spec/system/candidate_interface/candidate_adding_referees_spec.rb
```diff
@@ -100,7 +101,8 @@ def i_see_both_referees
     expect(page).to have_content('ajptaylor@example.com')
     expect(page).to have_content('Thats my tutor, that is')
 
-    expect(page).to have_content('Bill Lumbergh')
+    full_name_without_trailing_space = "Bill Lumbergh\n"
```
duncanjbrown: Why the newline?

tvararu: I'm testing to see if the trailing space is gone; here's an alternative I could think of:

```ruby
    full_name_with_trailing_space = "Bill Lumbergh "
    expect(page).to_not have_content(full_name_with_trailing_space)
    expect(page).to have_content("Bill Lumbergh")
```

Is there a better way??

tvararu:
Discussed offline, I'm going to expand this to every candidate controller.

tvararu:
Added another commit that makes this apply to every controller.

Tijmen and I considered calling this something like `clean_params` or `strip_trailing_whitespace_values` etc, but decided since we would have to call the method explicitly in every affected method, it's easier if we just call `.transform_values(&:strip)` and be done with it.

PR #712, opened 2019-11-26
Replace tcs in build before launch

stevehook:
### Context

We need to update the copy for provider terms before launch.

Just copy updates in app/views/content/terms_provider.md

There formatting of the numbered headings is not quite the same as the Google doc:

provider-terms1

There is a single quote used in the original doc, I've changed it to an apostrophe for consistency:

provider-terms2

### Guidance to review

Have the copy changes been made correctly and do we need to address either of the issues above?

### Link to Trello card

[1247 - Replace T&Cs in build before launch](https://trello.com/c/nlSgzWgs/1247-replace-tcs-in-build-before-launch)

### Env vars

no env vars

### DIFF
--- a/app/views/content/terms_provider.md
+++ b/app/views/content/terms_provider.md
@@ -47,7 +47,7 @@ Thank you for taking part in trialling Manage teacher training applications. Her
 
     Once a candidate has accepted an offer, you can’t change the terms of the offer without the candidate’s permission. If necessary, you can change conditions up until this point.
 
-    Don't pressure candidates into making a decision or withdraw an offer if the applicant has told you that they’re waiting to hear back from other providers.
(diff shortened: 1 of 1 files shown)


app/views/content/terms_provider.md
```diff
@@ -1,3 +1,5 @@
+## Terms of use for teacher training providers
```
paulrobertlloyd: @stevehook Not sure why we’d need to do this? Can we drop commit `b8c809` (‘Split the title’), as this was fine before, and consistent with how we title the other policy content across the service.

stevehook: @paulrobertlloyd sure can drop it, the reason for the change was just me trying to reproduce what was in the google doc:

image

stevehook: Dropped that commit, now looks like this...

image

paulrobertlloyd: Awesome, thanks!

PR #721, opened 2019-11-26
Calculate and store DBD (Declined By Default) date.

alexdesi:
### Context

Candidates are given 10 business days to accept or decline offers from providers for their application choices. The candidate dashboard will need to display the relevant deadline (not in scope for this PR) and the backend will need to calculate the relevant deadline (in scope) and eventually automatically process all applications to enforce it (not in scope). 

During kickoff, the following decisions were made:

- We'll store a ```:declined_by_default_at``` against ```ApplicationChoice```, as everything else related to provider workflows is stored against this model, not ```ApplicationForm```.
- The 'declined by default date' (DBD) should be the same for all application_choices.
- The DBD is only set for all choices when no more choices are left waiting for a provider decision. Until then it is ```nil```.
- The number of business days we are using to calculate the DBD may change in the future and may, in fact, vary according to different time periods (e.g. summer/winter). We'll use a similar approach to 'Reject by default', which uses ```TimeLimitConfig``` (which may be db-backed in the future).

A ```SetDeclinedByDefault``` service is responsible for setting DBD on all application_choices for an application_form if these conditions are met:

- there is a decision (offer/rejected) for all of the application_choices, i.e. no pending decisions
- at least one of these decisions is an offer (we've decided if all choices are rejected we do nothing)

This service is triggered by provider actions, in particular:

- MakeAnOffer
- RejectApplication
- RejectApplicationByDefault

The reason we are triggering the service even for rejection events is that the timing and order of these provider decisions can vary. For example, imagine you have 3 choices, the first one is approved, the second one is rejected and the third one is rejected by default. Up to the point when the third one is rejected by default, you still have application choices awaiting for a provider decision, so DBD cannot be set. The rule for calculating DBD is 'add dbd_days to the date of the most recent decision'.

### Guidance to review

This is hard one to test manually, as the UI for exposing the DBD is not merged yet. Please look at the relevant specs.

### Link to Trello card

[531 - Calculate and store DBD date](https://trello.com/c/GKbPrlnp)

### DIFF
--- a/app/services/make_an_offer.rb
+++ b/app/services/make_an_offer.rb
@@ -16,17 +16,22 @@ class MakeAnOffer
   def save
     return unless valid?
 
-    ApplicationStateChange.new(@application_choice).make_offer!
-    @application_choice.offer = { 'conditions' => (@offer_conditions || []) }
+    ApplicationStateChange.new(application_choice).make_offer!
+    application_choice.offer = { 'conditions' => (@offer_conditions || []) }
 
-    @application_choice.save!
-    StateChangeNotifier.call(:make_an_offer, application_choice: @application_choice)
+    application_choice.offered_at = Time.now
+    application_choice.save!
+
+    SetDeclineByDefault.new(application_form: application_choice.application_form).call
+    StateChangeNotifier.call(:make_an_offer, application_choice: application_choice)
   rescue Workflow::NoTransitionAllowed => e
     errors.add(:state, e.message)
     false
   end
 
 private
+
+  attr_reader :application_choice
 
   def validate_offer_conditions
     return if @offer_conditions.blank?
--- a/app/services/set_decline_by_default.rb
+++ b/app/services/set_decline_by_default.rb
@@ -1,0 +1,42 @@
+class SetDeclineByDefault
+  def initialize(application_form:)
+    @application_form = application_form
+  end
+
+  def call
+    return false if pending_decisions? || no_offers?
+
+    most_recent_decision_date = [
+      application_choices.maximum(:offered_at),
+      application_choices.maximum(:rejected_at),
+    ].compact.max
+
+    dbd_days = TimeLimitCalculator.new(
+      rule: :decline_by_default,
+      effective_date: most_recent_decision_date,
+    ).call
+
+    dbd_date = dbd_days.business_days.after(most_recent_decision_date).end_of_day
+
+    application_choices.where('status = \'offer\' AND decline_by_default_at IS NULL').update_all(
+      decline_by_default_at: dbd_date,
+      decline_by_default_days: dbd_days,
+    )
+  end
+
+private
+
+  attr_reader :application_form
+
+  def application_choices
+    application_form.application_choices
+  end
+
+  def pending_decisions?
+    application_form.awaiting_provider_decisions?
+  end
+
+  def no_offers?
+    application_choices.where(status: :offer).none?
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -29,6 +29,10 @@ ActiveRecord::Schema.define(version: 2019_11_27_105827) do
     t.datetime "reject_by_default_at"
     t.boolean "rejected_by_default", default: false, null: false
     t.integer "reject_by_default_days"
+    t.datetime "decline_by_default_at"
+    t.integer "decline_by_default_days"
+    t.datetime "offered_at"
+    t.datetime "rejected_at"
     t.index ["application_form_id"], name: "index_application_choices_on_application_form_id"
     t.index ["course_option_id"], name: "index_application_choices_on_course_option_id"
   end
--- a/spec/services/make_an_offer_spec.rb
+++ b/spec/services/make_an_offer_spec.rb
@@ -1,6 +1,20 @@
 require 'rails_helper'
 
 RSpec.describe MakeAnOffer do
+  describe '#save' do
+    it 'sets the offered_at date' do
+      application_choice = create(:application_choice, status: :awaiting_provider_decision)
+
+      MakeAnOffer.new(application_choice: application_choice)
+
+      Timecop.freeze do
+        MakeAnOffer.new(application_choice: application_choice).save
+
+        expect(application_choice.offered_at).to eq(Time.now)
+      end
+    end
+  end
+
   describe 'validation' do
     it 'accepts nil conditions' do
       decision = MakeAnOffer.new(
@@ -36,6 +50,25 @@ RSpec.describe MakeAnOffer do
       )
 
       expect(decision).not_to be_valid
+    end
+  end
+
+  describe 'decline by default' do
+    let(:application_form) { create :application_form }
+
+    let(:application_choice) {
+      create(:application_choice,
+             application_form: application_form,
+             status: 'awaiting_provider_decision',
+             edit_by: 2.business_days.ago)
+    }
+
+    it 'calls SetDeclineByDefault service' do
+      MakeAnOffer.new(application_choice: application_choice).save
+      application_choice.reload
+
+      expect(application_choice.decline_by_default_at).not_to be_nil
+      expect(application_choice.decline_by_default_days).not_to be_nil
     end
   end
 end
--- a/spec/services/reject_application_by_default_spec.rb
+++ b/spec/services/reject_application_by_default_spec.rb
@@ -29,6 +29,22 @@ RSpec.describe RejectApplicationByDefault do
     expect(application_choice.reload.rejected_by_default).to be true
   end
 
+  it 'sets `rejected_at`' do
+    application_choice = create_application
+    described_class.new(application_choice: application_choice).call
+    expect(application_choice.reload.rejected_at).not_to be_nil
+  end
+
+  it 'calls SetDeclineByDefault service' do
+    service_double = instance_double(SetDeclineByDefault)
+    allow(service_double).to receive(:call)
+    allow(SetDeclineByDefault).to receive(:new).and_return(service_double)
+
+    application_choice = create_application
+    described_class.new(application_choice: application_choice).call
+    expect(service_double).to have_received(:call)
+  end
+
   it 'sends a Slack notification' do
     allow(SlackNotificationWorker).to receive(:perform_async)
     application_choice = create_application
--- a/spec/services/set_decline_by_default_spec.rb
+++ b/spec/services/set_decline_by_default_spec.rb
@@ -1,0 +1,164 @@
+require 'rails_helper'
+
+RSpec.describe SetDeclineByDefault do
+  describe '#call' do
+    let(:application_form) { create(:completed_application_form) }
+    let(:choices) { application_form.application_choices }
+    let(:time_limit_calculator) { instance_double('TimeLimitCalculator', call: 10) }
+    let(:now) { Time.zone.local(2019, 11, 26, 12, 0, 0) }
+    let(:call_service) { SetDeclineByDefault.new(application_form: application_form).call }
+
+    before { allow(TimeLimitCalculator).to receive(:new).and_return(time_limit_calculator) }
+
+    around do |example|
+      Timecop.freeze(now) do
+        example.run
+      end
+    end
+
+    def expect_timestamps_to_match_excluding_milliseconds(first, second)
+      expect(first.change(usec: 0)).to eq(second.change(usec: 0))
+    end
+
+    def expect_all_relevant_decline_by_default_at_values_to_be(expected)
+      application_form.application_choices.reload.each do |application_choice|
+        if application_choice.offered_at
+          dbd_at = application_choice.decline_by_default_at
+
+          if !expected.nil?
+            expect_timestamps_to_match_excluding_milliseconds(dbd_at, expected)
+          else
+            expect(dbd_at).to be_nil
+          end
+        end
+      end
+    end
+
+    context 'when all the application choices have an offer' do
+      it 'the DBD is set to 10 business days from the date of the most recent offer' do
+        choices[0].update(status: :offer, offered_at: 1.business_days.before(now))
+        choices[1].update(status: :offer, offered_at: 2.business_days.before(now))
+        choices[2].destroy # this tests that we can handle fewer than 3 choices
+
+        expected_dbd_date = 9.business_days.after(now).end_of_day
+
+        call_service
+
+        expect_all_relevant_decline_by_default_at_values_to_be expected_dbd_date
+      end
+    end
+
+    context 'when no offers or rejections have been made for any of the application choices' do
+      it 'the DBD is not set for any of the application_choices' do
+        choices[0].update(status: :awaiting_provider_decision)
+        choices[1].update(status: :awaiting_provider_decision)
+        choices[2].update(status: :awaiting_provider_decision)
+
+        call_service
+
+        expect_all_relevant_decline_by_default_at_values_to_be nil
+      end
+    end
+
+    context 'when one offer was made, and some decisions are pending' do
+      it 'the DBD is not set for any of the application_choices' do
+        choices[0].update(status: :offer, offered_at: 1.business_days.before(now))
+        choices[1].update(status: :awaiting_provider_decision)
+        choices[2].update(status: :awaiting_provider_decision)
+
+        call_service
+
+        expect_all_relevant_decline_by_default_at_values_to_be nil
+      end
+    end
+
+    context 'when one offer was made, and two decisions are rejected' do
+      it 'the DBD is set to 10 business days from the date of the most recent decision' do
+        choices[0].update(status: :offer, offered_at: 1.business_days.before(now))
+        choices[1].update(status: :rejected, rejected_at: 2.business_days.before(now))
+        choices[2].update(status: :rejected, rejected_at: 3.business_days.before(now))
+
+        expected_dbd_date = 9.business_days.after(now).end_of_day
+
+        call_service
+
+        expect_all_relevant_decline_by_default_at_values_to_be expected_dbd_date
+      end
+    end
+
+    context 'when the most recent decision is a rejection' do
+      it 'the DBD is set to 10 business days from the date of this rejection' do
+        choices[0].update(status: :rejected, rejected_at: 3.business_days.before(now))
+        choices[1].update(status: :offer, offered_at: 2.business_days.before(now))
+        choices[2].update(status: :rejected, rejected_at: 1.business_days.before(now))
+
+        expected_dbd_date = 9.business_days.after(now).end_of_day
+
+        call_service
+
+        expect_all_relevant_decline_by_default_at_values_to_be expected_dbd_date
+      end
+    end
+
+    context 'when all application choices have been rejected' do
+      it 'the DBD is not set for any of the application_choices' do
+        choices[0].update(status: :rejected, rejected_at: 1.business_days.before(now))
+        choices[1].update(status: :rejected, rejected_at: 2.business_days.before(now))
+
+        call_service
+
+        expect_all_relevant_decline_by_default_at_values_to_be nil
+      end
+    end
+
+    context 'when the service is run twice' do
+      it 'the DBD is not set again on choices which already have a DBD' do
+        old_dbd_date = 8.business_days.after(now).end_of_day
+
+        choices[0].update(status: :rejected, rejected_at: 4.business_days.before(now))
+        choices[1].update(status: :rejected, rejected_at: 3.business_days.before(now))
+        choices[2].update(
+          status: :offer,
+          offered_at: 2.business_days.before(now),
+          decline_by_default_at: old_dbd_date,
+          decline_by_default_days: 10,
+        )
+
+        choices[1].update(status: :offer, offered_at: 1.business_days.before(now))
+        new_dbd_date = 9.business_days.after(now).end_of_day
+        call_service
+
+        dbd_for_old_offer = choices[2].reload.decline_by_default_at
+        dbd_for_new_offer = choices[1].reload.decline_by_default_at
+
+        expect_timestamps_to_match_excluding_milliseconds(dbd_for_old_offer, old_dbd_date)
+        expect_timestamps_to_match_excluding_milliseconds(dbd_for_new_offer, new_dbd_date)
+      end
+    end
+
+    it 'the decline_by_default_days is set to 10 days when DBD is present' do
+      choices[0].update(status: :offer, offered_at: 1.business_days.before(now))
+      choices[1].update(status: :offer, offered_at: 2.business_days.before(now))
+      choices[2].update(status: :offer, offered_at: 2.business_days.before(now))
+
+      call_service
+
+      choices.each do |choice|
+        expect(choice.reload.decline_by_default_days).to eq 10
+      end
+    end
+
+    it 'does not set DBD fields on non-offer application_choices (e.g. rejected/withdrawn)' do
+      choices[0].update(status: :offer, offered_at: 1.business_days.before(now))
+      choices[1].update(status: :rejected, rejected_at: 2.business_days.before(now))
+      choices[2].update(status: :withdrawn, offered_at: 3.business_days.before(now))
+
+      call_service
+
+      choices.where.not(status: :offer).each do |choice|
+        expect(choice.reload.decline_by_default_at).to be_nil
+        expect(choice.reload.decline_by_default_days).to be_nil
+      end
+    end
+  end
+end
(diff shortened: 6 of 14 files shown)


app/services/set_decline_by_default.rb
```diff
@@ -0,0 +1,29 @@
+class SetDeclineByDefault
+  def initialize(application_form:)
+    @application_form = application_form
+  end
+
+  def call
+    return false unless application_choices.count == 3
```
stevehook: Make `3` a constant somewhere?

app/services/set_decline_by_default.rb
```diff
@@ -0,0 +1,29 @@
...
+                                            application_choices.maximum(:rejected_at)]
+                                             .compact.max
+
+    dbd_days = TimeLimitCalculator.new(rule: :decline_by_default, effective_date: most_recent_offered_or_rejected_date).call
+
+    dbd_date = dbd_days.business_days.after(most_recent_offered_or_rejected_date).end_of_day
+
+    application_form.application_choices.all? do |application_choice|
+      application_choice.update(decline_by_default_at: dbd_date)
+    end
```
stevehook: Do we need to wrap the operations in this method in a DB transaction or is it already called with a transaction?

app/services/set_decline_by_default.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+private
+
+  attr_reader :application_form
+
+  def application_choices
+    application_form.application_choices.where(status: %i[offer rejected])
+  end
```
stevehook: Maybe rename this method to `offered_or_rejected_application_choices` or something shorter (I got confused above thinking that `application_choices` was all the application choices (regardless of status).

app/services/set_decline_by_default.rb
stevehook: What happens if the candidate only makes 1 or 2 choices? (My understanding may be wrong here - I'm thinking that 3 is the maximum and that candidates don't have to use all their choices)

app/services/make_an_offer.rb
```diff
@@ -39,4 +45,14 @@ def validate_offer_conditions
...
+
+  def set_decline_by_default
+    days = TimeLimitCalculator.new(
+      rule: :decline_by_default,
+      effective_date: Time.zone.now,
+      ).call
+
+    application_choice.decline_by_default_days = days
+    application_choice.decline_by_default_at = days.business_days.from_now.end_of_day
+  end
```
stevehook: Is this method going to be replaced by a call to `SetDeclineByDefault`?

spec/services/set_decline_by_default_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      application_form.application_choices.each do |application_choice|
+        expect(application_choice.decline_by_default_at).to eq(expected_value)
+      end
+    end
+
+    context 'all the application choices have an offer' do
+      it 'the DBD is set to 10 business days from the date of the most recent offer' do
+        choices[0].update(status: :offer, offered_at: 1.business_days.from_now)
+        choices[1].update(status: :offer, offered_at: 2.business_days.from_now)
+        choices[2].update(status: :offer, offered_at: 3.business_days.from_now)
```
stevehook: seems a it odd to have the offer dates in the future, though it's logically correct I guess

spec/services/set_decline_by_default_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+
+        expect(result).to eq true
+
+        # control the dbd_days to be 10 days
```
stevehook: Comment was a it confusing, maybe...

    # assert that the DBD date is set to 10 days after last offer date

app/services/set_decline_by_default.rb
duncanjbrown: This logic also needs to check that the choices have all received decisions — there can be fewer than three

app/services/set_decline_by_default.rb
```diff
@@ -0,0 +1,33 @@
+class SetDeclineByDefault
+  def initialize(application_form:)
+    @application_form = application_form
+  end
+
+  def call
+    return false unless decided_choices.count == 3 && decided_choices.any?(&:offered_at)
```
duncanjbrown: There can be fewer than three choices

app/services/set_decline_by_default.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    dbd_days = TimeLimitCalculator.new(
+      rule: :decline_by_default,
+      effective_date: most_recent_decision_date,
+    ).call
+
+    dbd_date = dbd_days.business_days.after(most_recent_decision_date).end_of_day
+
+    application_form.application_choices.all? do |application_choice|
+      application_choice.update(decline_by_default_at: dbd_date)
```
duncanjbrown: It feels a bit odd to make the update as a side-effect of `all?`. How about `application_form.application_choices.update_all(...)`.

app/services/make_an_offer.rb
alexdesi: Yes! we'll call the service.

app/services/set_decline_by_default.rb
alexdesi: `decided_choices` covers both, offered and rejected.

But we need to fix the `decided_choices.count == 3` part, as you said there can be fewer that 3.

app/services/set_decline_by_default.rb
alexdesi: yes, this smell a bit 👃

app/services/set_decline_by_default.rb
alexdesi: we'll probably use `update_all`

app/services/set_decline_by_default.rb
duncanjbrown: Great :+1:. Is is right that we then go on to set `decline_by_default_at` for all the choices? Some of them may have been rejected, which makes DBD redundant

app/services/set_decline_by_default.rb
alexdesi: We updated this, now do not check the number of choices.
As we understood there is no need to check the Max number of choices,
is that correct?
cc @mnacos

app/services/set_decline_by_default.rb
alexdesi: we are using a different logic, we have now a method called `pending_application_choices`, that is self explaining

spec/services/set_decline_by_default_spec.rb
mnacos: I agree, has been changed

app/services/set_decline_by_default.rb
alexdesi: we think this is not harmful, and it is a bit more futureproof, It would be possible, for example, to change an application from rejected to offered in case of a mistake from a provider.

app/services/set_decline_by_default.rb
mnacos: Yes, we've refactored this to check if there are any application choices awaiting decision, so we don't need any constants and the code works with any number of application choices. The service doesn't need to know anything about the number of application choices allowed elsewhere in the application and shouldn't have any role enforcing these limits.

alexdesi:
Thanks for all the comments @duncanjbrown and @stevehook, 
this should be ready now (hopefully 😉  )

app/services/make_an_offer.rb
```diff
@@ -28,6 +31,8 @@ def save
 
 private
 
+  attr_reader :application_choice
```
tijmenb: Is this needed?

spec/services/make_an_offer_spec.rb
```diff
@@ -38,4 +38,37 @@
...
+      application_form = create :completed_application_form
+      create(
+        :application_choice,
+        application_form: application_form,
+        status: 'awaiting_provider_decision',
+        edit_by: 2.business_days.ago,
+        )
+    end
+
+    it 'sets the `offered_at` date' do
```
tijmenb: Does this `it` belong in the `decline by default` context?

spec/services/make_an_offer_spec.rb
```diff
@@ -38,4 +38,37 @@
...
+      application_choice = create_application
+
+      Timecop.freeze do
+        MakeAnOffer.new(application_choice: application_choice).save
+        expect(application_choice.offered_at).to eq(Time.now)
+      end
+    end
+
+    it 'calls SetDeclineByDefault service' do
```
tijmenb: Instead of the stubbing and instances, can we check if the attributes are updated on the form? This does a lot of testing of internals.

spec/services/reject_application_by_default_spec.rb
```diff
@@ -29,6 +29,22 @@ def create_application
     expect(application_choice.reload.rejected_by_default).to be true
   end
 
+  it 'sets `rejected_at`' do
```
tijmenb: Same as above.

spec/services/set_decline_by_default_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+    around do |example|
+      Timecop.freeze(now) do
+        example.run
+      end
+    end
+
+    def expect_all_decline_by_default_at_values_to_be(expected_value)
+      application_form.application_choices.reload.each do |application_choice|
+        dbd_at = application_choice.decline_by_default_at
+        without_milliseconds = dbd_at && dbd_at.change(usec: 0)
```
tijmenb: Doesn't timecop help us avoid having to do this?

app/services/make_an_offer.rb
alexdesi: We were still using instance variable,  I'm gonna change `@application_choice` -> `application_choice`

mnacos:
> I haven't looked at the DBD logic before, so I'm having trouble understanding why it's being set when the provider rejects an application (either explicitly or manually). Do we have an explanation of this somewhere?
> 
> I think it would be good to have this tested in a system test as well - perhaps something that can be expanded on later when we have the actual implementation. I'm having trouble understanding the logic of `SetDeclineByDefault` and the tests.

I think more context is needed in the PR description, so that someone who hasn't followed the kickoff/discussion can follow the logic.

The main idea is that DBD is only set if there are no pending provider decisions on any application choices. The timing and order of these provider decisions can vary. For example, you can have 3 choices, the first one is approved, the second one is rejected and the third one is rejected by default. Up to the point when the third one is rejected by default, you still have application choices awaiting for a provider decision, so DBD cannot be set. The rule for calculating DBD is 'add dbd_days to the date of the most recent decision'.

spec/services/set_decline_by_default_spec.rb
mnacos: Weirdly, no. The explicit reload, which makes the application_choices visible to the test, fetch the timestamp from the database and the value from the database seems to have different precision/scale than the memory-based calculated Ruby time. We've tried using ```.round```, but ```:end_of_day``` setting .99999999 means that round rounds up to the following day.

spec/services/make_an_offer_spec.rb
alexdesi: Not sure I got this, is not aready in there?

```ruby
describe 'decline by default' do
 ...
  it 'sets the `offered_at` date' do
  end
```

app/services/set_decline_by_default.rb
duncanjbrown: I think there's more to clarify here.

Our understanding is that provider offer/rejection is a binding decision and is not reversible. The UI does not support it.

If we set `decline_by_default_at` on *all* the applicationchoices, that puts the responsibility on the sweep job which comes round and does the declining (that is, the thing which will be implemented in the follow-up PR to this one) to know whether or not to do it. This means the logic for selecting applications to decline by default lives in two places: here and in the sweep. For this reason I would argue that it is better to only set the `dbd_at` and `dbd_days` on choices which will actually be declined.

spec/services/make_an_offer_spec.rb
alexdesi: There are 2 reasons:
- SetDeclineByDefault is already extensively tested on `set_decline_by_default_spec.rb`
- the `MakeAnOffer` spec is already using a lot of stubs (`MakeAnOffer`), changing the style of the test would require rewriting all the other tests

mnacos:
Hi @tijmenb I have updated the PR description to provide more context (@duncanjbrown let me know if I've missed anything important).

mnacos:
> I think it would be good to have this tested in a system test as well - perhaps something that can be expanded on later when we have the actual implementation. I'm having trouble understanding the logic of `SetDeclineByDefault` and the tests.

There is another ticket to integrate the frontend for this with the backend, at which point we can test end-to-end.

spec/services/make_an_offer_spec.rb
tijmenb: I think it shouldn’t be, it’s not really part of decline by default.

spec/services/make_an_offer_spec.rb
tijmenb: You don’t have to test the exact values (which would dupe the existing tests), just make sure the attributes on the form are updated.

app/services/set_decline_by_default.rb
alexdesi: Ok, actually it should not take too much to implement it, thanks for clarify it.

spec/services/make_an_offer_spec.rb
alexdesi: 👍 
I got it! at first, in my mind was kind of related.
refactored

spec/services/make_an_offer_spec.rb
alexdesi: good call @tijmenb, I've changed it!

spec/services/reject_application_by_default_spec.rb
mnacos: I think this belongs here, it is RejectApplicationByDefault's responsibility to set ```rejected_at```, which SetDeclineByDefault then uses to determine whether decisions are pending or what the most recent decision is. SetDeclineByDefault only reads these attributes, doesn't write them. The are written by:

- MakeAnOffer sets ```offerred_at```
- RejectApplication sets ```rejected_at```
- RejectApplicationByDefault sets ```rejected_at``` (as well as ```rejected_by_default```)

app/services/set_decline_by_default.rb
mnacos: @duncanjbrown I've changed the logic so that the DBD fields are only set on offered application_choices, not rejected ones. The sweeper task will just need to look at ```:offered_at``` this way.

tijmenb:
Thanks for clarifying the PR description! I think I get it now - in other words, we want to set the decline by default date when all of the application choices are in a state which requires the user to act on them.

Couple of points:

- We're missing withdrawls in this implementation, because we haven't build that yet. Might be worth documenting somewhere that once we implement that we need to trigger `SetDeclineByDefault` in the service, and update the `pending_application_choices` to include things in withdrawn state.

- Once all applications are rejected/offered/withdrawn, we also want to email the candidate - otherwise they won't know that an offer has been made. Would be good to structure the code so that adding that will be easy (renaming `SetDeclineByDefault` should work, or separating out the code that checks if we need to set decline by default from the actual setting). 

- In the near future we also need to detect whether or not all of the applications are rejected, so that the candidate can be put into "Apply 2" - worth keeping in mind too.

- Should we split the `offer` state in `application_choice` into 2: `offer_made_but_not_sent` and `offer_sent`? Would that make this easier to build?

- I wonder if `decline_by_default_at` really belongs on `ApplicationChoice`. If you [look at the design](https://apply-mvp-prototype.herokuapp.com/applications?phase=dbd&state=has-decisions), it shows "You have 12 days (until 9 December 2019) to respond to any offers". Displaying that data will mean that you have to get `decline_by_default_at` from one of the choices, which feels a bit clunky.

app/services/set_decline_by_default.rb
```diff
@@ -0,0 +1,38 @@
...
+
+private
+
+  attr_reader :application_form
+
+  def application_choices
+    application_form.application_choices
+  end
+
+  def pending_application_choices
```
tijmenb: We're missing withdrawls in this implementation, because we haven't build that yet. Might be worth documenting somewhere that once we implement that we need to trigger `SetDeclineByDefault` in the service, and update the `pending_application_choices` to include things in withdrawn state.

app/services/set_decline_by_default.rb
```diff
@@ -0,0 +1,38 @@
+class SetDeclineByDefault
```
tijmenb: Once all applications are rejected/offered/withdrawn, we also want to email the candidate - otherwise they won't know that an offer has been made. Would be good to structure the code so that adding that will be easy (renaming `SetDeclineByDefault` should work, or separating out the code that checks if we need to set decline by default from the actual setting). 

In the near future we also need to detect whether or not all of the applications are rejected, so that the candidate can be put into "Apply 2" - worth keeping in mind too.

app/services/set_decline_by_default.rb
```diff
@@ -0,0 +1,38 @@
...
+      application_choices.maximum(:rejected_at),
+    ].compact.max
+
+    dbd_days = TimeLimitCalculator.new(
+      rule: :decline_by_default,
+      effective_date: most_recent_decision_date,
+    ).call
+
+    dbd_date = dbd_days.business_days.after(most_recent_decision_date).end_of_day
+
```
tijmenb: Should we split the `offer` state in `application_choice` into 2: `offer_made_but_not_sent` and `offer_sent`? Would that make this easier to build?

db/schema.rb
```diff
@@ -29,6 +29,10 @@
     t.datetime "reject_by_default_at"
     t.boolean "rejected_by_default", default: false, null: false
     t.integer "reject_by_default_days"
+    t.datetime "decline_by_default_at"
```
tijmenb: I wonder if `decline_by_default_at` really belongs on `ApplicationChoice`. If you [look at the design](https://apply-mvp-prototype.herokuapp.com/applications?phase=dbd&state=has-decisions), it shows "You have 12 days (until 9 December 2019) to respond to any offers". Displaying that data will mean that you have to get `decline_by_default_at` from one of the choices, which feels a bit clunky.

app/services/set_decline_by_default.rb
```diff
@@ -0,0 +1,35 @@
...
+    application_form.application_choices.update_all decline_by_default_at: dbd_date, decline_by_default_days: dbd_days
```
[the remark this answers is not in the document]
mnacos: Yes, this has been addressed

app/services/set_decline_by_default.rb
```diff
@@ -0,0 +1,35 @@
+class SetDeclineByDefault
+  def initialize(application_form:)
+    @application_form = application_form
+  end
+
+  def call
+    return false if pending_application_choices.any? || application_choices.none?(&:offered_at)
```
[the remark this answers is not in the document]
mnacos: As already discussed f2f, there are two checks that the service needs to make for this bit of work:
1. whether there are still application choices waiting for a provider decision (determines if DBD is set)
2. whether an application choice is an offer (DBD should be set only on 'offer' choices)

I have decided to keep the timestamps (```offered_at```, ```rejected_at```) but switch to ```status == 'awaiting_provider_decision'``` for check 1 and ```status == 'offer'``` for check 2 because this is better aligned with our state machine workflow.

This is also more consistent with the new logic of setting DBD only on 'offer' choices.

db/schema.rb
mnacos: That's how I felt about this originally, but:
- states and state transitions are all ApplicationChoice-specific already
- and there will be a sweeper task performing the actual 'decline' part of DBD, operating at the ApplicationChoice level (e.g. marking each choice as declined, possibly sending a notification to the provider)

app/services/set_decline_by_default.rb
mnacos: We hadn't considered withdrawals up to now, but funnily enough the refactor I discussed with Steve earlier will solve this neatly. If check 1 (see above) around pending choices is checking that ```status == 'awaiting_provider_decision'```, then any withdrawn applications will be regarded automatically as decided upon/non-pending. If check 2 is based on ```status == 'offer'```, then DBD will not be set on withdrawn applications (if DBD fields have already been set before withdrawal, the sweeper task just needs to make sure it declines only application_choices in 'offer' state and it will be fine).

So, all we need is a note somewhere to call SetDeclineByDefault service from the withdrawal service.

app/services/set_decline_by_default.rb
mnacos: The pending check will now be based on ```status == 'awaiting_provider_decision'```, so perhaps we could have a helper method on ApplicationForm, e.g. ```#awaiting_provider_decisions?```

This could be used by SetDeclineByDefault, but also from PutCandidateIntoApply2 (with an additional check there are no offers).

app/services/set_decline_by_default.rb
mnacos: Depends if these should be treated as independent states. My preference would be to avoid this, it feels like we abusing states for book-keeping purposes, such as knowing if an email has been sent. Unless you mean something different by ```not_sent``` and ```sent``` (e.g. candidate accepting offer while other provider decisions are still pending -- yes this is thing -- but it should be fine with the new checks).

db/schema.rb
tijmenb: > states and state transitions are all ApplicationChoice-specific already

Having the dates on ApplicationChoice is already causing problems, like we see in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/738/files#r351666170. Are we going to have the same problem here?

See also https://trello.com/c/zPsXM1TW.

spec/services/reject_application_by_default_spec.rb
```diff
@@ -29,6 +29,22 @@ def create_application
...
+    application_choice = create_application
+    described_class.new(application_choice: application_choice).call
+    expect(application_choice.reload.rejected_at).not_to be_nil
+  end
+
+  it 'calls SetDeclineByDefault service' do
```
tijmenb: Probably best to do the same as you did in `make_an_offer_spec.rb` here.

app/services/set_decline_by_default.rb
duncanjbrown: I think I agree with Michael, but I would like to check my understanding.

AFAICS the way this works at the moment is that the checks at the top of this procedure mean it only runs once all the choices are ready. In that sense it's ready to become a more generic place, as @tijmenb suggests above in his comment about renaming, to hook actions that need to happen once the batch as a whole is decided on, like emailing the candidate.

I think there's another thing to check here though, which is what happens when this gets the same application_form twice? if the dbd date is already set, it shouldn't change it.

spec/services/set_decline_by_default_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+      end
+    end
+
+    context 'when one offer was made, and two decisions are rejected' do
+      it 'the DBD is set to 10 business days from the date of the most recent decision' do
+        choices[0].update(status: :offer, offered_at: 1.business_days.before(now))
+        choices[1].update(status: :rejected, rejected_at: 2.business_days.before(now))
+        choices[2].update(status: :rejected, rejected_at: 3.business_days.before(now))
+
+        # usec: 0 strips away milliseconds from Ruby Time
```
tijmenb:
```suggestion

```

spec/services/set_decline_by_default_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+      end
+    end
+
+    context 'when the most recent decision is a rejection' do
+      it 'the DBD is set to 10 business days from the date of this rejection' do
+        choices[0].update(status: :rejected, rejected_at: 3.business_days.before(now))
+        choices[1].update(status: :offer, offered_at: 2.business_days.before(now))
+        choices[2].update(status: :rejected, rejected_at: 1.business_days.before(now))
+
+        # usec: 0 strips away milliseconds from Ruby Time
```
tijmenb:
```suggestion
```

db/schema.rb
duncanjbrown: I'd be happy to merge this as-is for now — maybe we should take a step back and review the way we organise this information in the near future? It also came up in #738.

duncanjbrown: Missed your comment @tijmenb 😄

tijmenb: Yeah, good idea.

spec/services/reject_application_by_default_spec.rb
mnacos: ok

mnacos: Ha, can't do that, because calling SetDeclineByDefault on a rejected application has no side-effects for that application (but may trigger DBD to be set on another, if no choices remain awaiting_provider_decision).

PR #723, opened 2019-11-27
Put word count validation messages in locale file and fix messages

fofr:
### Context

Previously we weren’t passing in a custom message, and error messages read quite generically, "Reduce the word count for X".

We had some `too_long` strings in the yml but they were never being used.

Switch to a custom error message, `too_many_words`, add a default message that includes the number of words and update existing error keys so they show correctly for becoming a teacher, subject knowledge and interview preferences.

Also switch referee relationship from being a character length validator to a word length one.

https://trello.com/c/nCljMNvt/562-error-for-word-count-should-include-maximum-word-count

Also relates to https://trello.com/c/dYpc40B1/563-a-list-of-small-problems-to-fix

### DIFF
--- a/app/models/reference.rb
+++ b/app/models/reference.rb
@@ -4,7 +4,7 @@ class Reference < ApplicationRecord
                             email_address: true,
                             length: { maximum: 100 },
                             uniqueness: { scope: :application_form_id }
-  validates :relationship, presence: true, length: { minimum: 2, maximum: 500 }
+  validates :relationship, presence: true, word_count: { maximum: 50 }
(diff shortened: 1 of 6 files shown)


fofr:
@RobertM905 I don't think we have any word limit validation on provider UI or API _yet_? So this only affects candidate for now.

We need to work out how we'll show different error messages for API versus provider UI, which I was discussing with @duncanjbrown in relation to https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/compare/better-rejection-errors

PR #725, opened 2019-11-27
Fix incorrect path on Your application page when added volunteering

chubberlisk:
### Context

Currently, when you've added a volunteering role (but don't click that you've completed the selection) you are taken to the `Do you have experience volunteering with young people or in school?`. Instead you should be taken to the review volunteering page.

This PR fixes the incorrect path when you've added a volunteering role by adding in `#volunteering_path` to `ApplicationFormPresenter`.

However, when doing this it was discovered that we have no way to differentiate between when it’s the first time a candidate has arrived at the volunteering form and when they have no volunteering roles. In order to rectify this, a boolean `volunteering_experience` has been added which will be set at the `Do you have experience volunteering with young people or in school?` page using `VolunteeringExperienceForm`.

### Guidance to review

Does this make sense? This was the simplest solution I could think of but feel free to shout if there's a better way.

Also, naming of the new attribute?

### Link to Trello card

[561 - Once you've added 1 volunteering role (but don't click completed), you should see the review when going back to the section](https://trello.com/c/XtjBDYhv/561-once-youve-added-1-volunteering-role-but-dont-click-completed-you-should-see-the-review-when-going-back-to-the-section)


app/presenters/candidate_interface/application_form_presenter.rb
```diff
@@ -88,6 +88,14 @@ def other_qualification_path
       end
     end
 
+    def volunteering_path
+      if volunteering_completed? || volunteering_added? || [true, false].include?(@application_form.volunteering_experience)
```
tvararu: Feels a bit clunky, but can't find a good way to improve

chubberlisk: Agreed! Could extract a private method?

tvararu: Feel free but not blocking 👍

fofr: What can `@application_form.volunteering_experience` be if not true or false, `nil?`

chubberlisk: Yeah, so when it's `nil` then we know the candidate has arrived at the volunteering section for the first time.

chubberlisk: Ended up doing this, means we don't have to update the attribute in `VolunteeringRoleForm` and removes the need for this monstrosity `[true, false].include?(@application_form.volunteering_experience)`.

chubberlisk: Since https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/725#discussion_r351285435, this isn't too bad anymore.

chubberlisk: Nvm, spoke to @tvararu. We're reverting back to original.

PR #730, opened 2019-11-27
Error summary placement

paulrobertlloyd:
### Context

From the accessibility audit, ensures error summary appears before page heading on the following pages:

* Add job page
* Add maths GCSE
* Add english GCSE
* Add science GCSE
* Tell us why you want to be a teacher
* Your subject knowledge
* Interview preferences

I will create a separate PR to fix where we are using `fieldset`s incorrectly (for entire pages)


fofr:
Why does this PR have string changes in locale?

PR #732, opened 2019-11-27
Update phone number validator to include blank error message

RobertM905:
### Context

Currently when a candidate enters a blank phone number they see a `Enter a real phone number` when we want the error message to be `Enter your phone number` for blank entries.

Before

Screenshot 2019-11-27 at 14 55 52

-------------
After

Screenshot 2019-11-27 at 14 56 21

- Update phone number validator to use the `phone_number.blank` translation when the phone number is blank.
- Update phone_number validator spec to reflect the blank error message changes.

### Guidance to review

Run specs and ensure that the correct error message is displayed for blank phone numbers.

### Link to Trello card

[524 - Error message for blank phone number](https://trello.com/c/z0j3sutx/524-code-in-error-message-for-blank-phone-number)

### Env vars

N/A

### DIFF
--- a/app/validators/phone_number_validator.rb
+++ b/app/validators/phone_number_validator.rb
@@ -1,7 +1,9 @@
 class PhoneNumberValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
-    if value.blank? || is_invalid_phone_number_format?(value)
-      record.errors[attribute] << I18n.t('activemodel.errors.models.candidate_interface/contact_details_form.attributes.phone_number.invalid')
+    if value.blank?
+      record.errors.add(attribute, :blank)
+    elsif is_invalid_phone_number_format?(value)
+      record.errors.add(attribute, :invalid)
     end
   end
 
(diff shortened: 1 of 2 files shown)


app/validators/phone_number_validator.rb
```diff
@@ -1,6 +1,8 @@
 class PhoneNumberValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
-    if value.blank? || is_invalid_phone_number_format?(value)
+    if value.blank?
+      record.errors[attribute] << I18n.t('activemodel.errors.models.candidate_interface/contact_details_form.attributes.phone_number.blank')
```
fofr: `record.errors.add(attribute, :blank)`?
Then Rails will defer to the right message automatically?

RobertM905: record.errors.add(attribute, :blank) seems to default to `can't be blank` 🤔 
I'll have a look into it.

RobertM905: It's how the Validation tests are set up. We use a `Validatable` model as a stub, which throws the locales off as it's not `candidate_interface/contact_details_form`.

PR #734, opened 2019-11-27
Disabled pentest deployment environment.

tpinney:
### Context

The pentest environment isn't currently required and is to be removed. Before it can be removed it needs to be removed from the release pipeline code to prevent it being redeployed after its been deleted.

- Commented out the pent test environment code to prevent it running in future releases. I've not removed it entirely in case its required again for the pen test in a couple of weeks.

### Guidance to review

N/A

### Link to Trello card

N/A

### Env vars

### DIFF
--- a/azure-pipelines-release.yml
+++ b/azure-pipelines-release.yml
@@ -139,52 +139,52 @@ stages:
       alertRecipientEmails: '$(alertRecipientEmails)'
 
 
-- stage: deploy_pentest
-  displayName: 'Deploy - Pentest'
-  dependsOn: publish_arm_template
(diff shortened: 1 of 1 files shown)


duncanjbrown:
Does this require extra config to turn off the true/false option for this env in the release pipeline workflow?

tpinney:
The variable in the pipeline will just become ineffective, but its presence is no controlled in the YAML file, that's done from in the DevOps GUI. We can remove it, but it we need it again in a couple of weeks I'd suggest just leaving it there for now since I can't do any harm once the pipeline code is removed.

PR #735, opened 2019-11-27
Remove unnecessary grid row wrapper on GCSE review page

fofr:
Fixes alignment of the SummaryCardComponent

## Before

Screen Shot 2019-11-27 at 15 42 11

### DIFF
--- a/app/views/candidate_interface/gcse/review/show.html.erb
+++ b/app/views/candidate_interface/gcse/review/show.html.erb
@@ -1,10 +1,8 @@
 <% content_for :title, t("gcse_summary.page_titles.#{@subject}") %>
 <% content_for :before_content, govuk_back_link_to(candidate_interface_application_form_path, 'Back to application') %>
 
-<div class="govuk-grid-row">
-  <h1 class="govuk-heading-xl"><%= t("gcse_summary.page_titles.#{@subject}") %></h1>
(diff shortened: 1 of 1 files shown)


app/views/candidate_interface/gcse/review/show.html.erb
```diff
@@ -1,10 +1,8 @@
 <% content_for :title, t("gcse_summary.page_titles.#{@subject}") %>
 <% content_for :before_content, govuk_back_link_to(candidate_interface_application_form_path, 'Back to application') %>
 
-<div class="govuk-grid-row">
```
alexdesi: Hi Paul, why we do not need this here,
I noticed there are other pages where the h1 is inside the `<div class="govuk-grid-row">`,
it sounds like common, but not sure

fofr: @alexdesi Where else are we doing that? We should remove it.

Full width sections don't need to be within a grid. They are constrained by the parent container.

fofr: @alexdesi I've checked all uses of `govuk-grid-row` and everything else looks ok.

When `govuk-grid-row` is used, the next child must be a grid column, like `govuk-grid-column-two-thirds`

PR #736, opened 2019-11-27
Audit provider user actions

stevehook:
### Context

We want to be able to see which user caused each change in the audit trail (Application History page in the support interface). This currently works for candidates and vendor API users. This PR adds support for provider users authenticated via DfE sign-in.

- Set the `username` for that the audited gem uses to attribute audit entries to `<provider-email> (Provider)` in the `ProviderInterfaceController` base controller.
- The existing logic in the view components for application history displays provider audit records like this:

image

### Guidance to review

- At the moment the only info we have about provider users is an email and they are not stored in the database so we are just setting the audit user as a `username` (String) rather than an associated persistent object (like we do with `Candidate`). We could review this in future if we have more info available.
- The request spec is an attempt to cover the logic in `ProviderInterfaceController` that sets up the audited `username`. I didn't want to just dump another assertion into an existing system spec because it's a cross cutting concern rather than main functionality. There could be an easier way to do this though.

### Link to Trello card

[1277 - Audit provider user actions](https://trello.com/c/FhGp8ffU/1277-placeholder-audit-provider-user-actions)

### Env vars

no env vars

### DIFF
--- a/spec/requests/provider_interface/audit_trail_spec.rb
+++ b/spec/requests/provider_interface/audit_trail_spec.rb
@@ -2,31 +2,28 @@ require 'rails_helper'
 
 RSpec.describe 'Provider interface - audit trail', type: :request do
   def create_application
-    application_form = create :application_form
-    application_choice = create(
+    create(
       :application_choice,
-      application_form: application_form,
-      status: 'awaiting_provider_decision',
+      :awaiting_provider_decision,
     )
-    provider = create :provider, code: 'ABC'
-    application_choice.course.update(accrediting_provider: provider)
-    application_choice
   end
 
-  before do
+  def set_provider_permission(application_choice)
+    provider_code = application_choice.course.provider.code
     allow(ProviderUser).to receive(:load_from_session)
       .and_return(
         ProviderUser.new(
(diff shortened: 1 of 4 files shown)


spec/requests/provider_interface/audit_trail_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'rails_helper'
+
+RSpec.describe 'Provider interface - audit trail', type: :request do
+  def create_application
+    application_form = create :application_form
+    application_choice = create(
```
duncanjbrown: nit: We have a factory for this now: `create(:application_choice, :awaiting_provider_decision)`

duncanjbrown:
nit not for this PR, but I noticed the act of rejection and the providing of a rejection reason register as two events in the audit log. 

Screenshot 2019-11-27 at 16 50 13

spec/requests/provider_interface/audit_trail_spec.rb
stevehook: good point, I wasn't too happy with this messy setup

PR #737, opened 2019-11-27
Refactor SectionMissingBannerComponent

paulrobertlloyd:
### Context

Updates link text to match section titles. Also DRYs out `SectionMissingBannerComponent` and adds a test for this component.

### Link to Trello card

[546 - Complete section links do not reflect h3s](https://trello.com/c/GqcnozMG/)

### DIFF
--- a/app/components/becoming_a_teacher_review_component.html.erb
+++ b/app/components/becoming_a_teacher_review_component.html.erb
@@ -1,9 +1,5 @@
 <% if show_missing_banner? %>
-  <%= render(SectionMissingBannerComponent, section: :becoming_a_teacher, text: t('review_application.becoming_a_teacher.incomplete')) do %>
-    <%= link_to candidate_interface_becoming_a_teacher_edit_path, class: 'govuk-link' do %>
-      <%= t('review_application.becoming_a_teacher.link_html') %>
-    <% end %>
-  <% end %>
+  <%= render(SectionMissingBannerComponent, section: :becoming_a_teacher, section_path: candidate_interface_becoming_a_teacher_edit_path, text: t('review_application.becoming_a_teacher.incomplete')) %>
 <% else %>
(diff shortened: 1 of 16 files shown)


app/components/becoming_a_teacher_review_component.html.erb
```diff
@@ -1,9 +1,5 @@
 <% if show_missing_banner? %>
-  <%= render(SectionMissingBannerComponent, section: :becoming_a_teacher, text: t('review_application.becoming_a_teacher.incomplete')) do %>
-    <%= link_to candidate_interface_becoming_a_teacher_edit_path, class: 'govuk-link' do %>
-      <%= t('review_application.becoming_a_teacher.link_html') %>
-    <% end %>
-  <% end %>
+  <%= render(SectionMissingBannerComponent, section: :becoming_a_teacher, section_path: candidate_interface_becoming_a_teacher_edit_path, text: t('review_application.becoming_a_teacher.incomplete')) %>
```
tvararu: This + all of the other ones: do we still need to pass in `text: t('review_application.becoming_a_teacher.incomplete')` if we're passing in `section` above?

paulrobertlloyd: Oooooh!

app/components/section_missing_banner_component.rb
```diff
@@ -1,10 +1,11 @@
 class SectionMissingBannerComponent < ActionView::Component::Base
-  validates :text, :section, presence: true
+  validates :section, :section_path, presence: true
 
-  def initialize(text:, section:)
-    @text = text
+  def initialize(section:, section_path:, text: t("review_application.#{section}.incomplete"))
```
paulrobertlloyd: @tvararu Kept `text` as a parameter, but it defaults to the translated string. This is because we currently provide a somewhat customised string for referees section, and may add more in the future. Make sense?

tvararu: Perfect!

PR #738, opened 2019-11-27
Send an email to the candidate when their application is sent to providers

duncanjbrown:
## Context

We need to let candidates know that their choices have gone to providers and that the RBD clock is ticking.

## Changes proposed in this PR

- make sure we send a form's choices at once, not piecemeal
- send an email to the user when this happens
- delete some specs that were double-testing

More detail in commit messages.

https://trello.com/c/jbybYv2c/423-email-email-to-candidate-when-application-is-sent-to-provider?menu=filter&filter=label:Dev

## Also

This was not as straightforward as it should have been.

This PR changes the behaviour of `SendApplicationsToProvider` so that it runs exactly once per `ApplicationForm`. The driver for this change is that we need to be sure we email the candidate only once about all their application choices going to the provider.

We have created some complexity by duplicating the `application_complete` status across `ApplicationChoices` instead of keeping it in one place. At the moment it is very difficult for these values to come out of sync as they are updated at a stroke by `ReceiveReference`

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/d1da550b9e938ee67bedacd5e2a3683817371e5b/app/services/receive_reference.rb#L25-L29

You can see from this code that the state stored on each `ApplicationChoice` is more or less a proxy for state that also exists on ApplicationForm.

This makes operations which need to work on the application as a whole — like the one in this PR — more complicated than they need to be.

Our original intention with the design was to keep lifecycle concerns on `ApplicationChoice` and keep `ApplicationForm` a dumb information container. I think we might be missing a concept in the system, possibly called `Apply1Application`, which can handle lifecycle concerns which wrap the whole application: choices and form.


app/mailers/candidate_mailer.rb
```diff
@@ -6,4 +6,17 @@ def submit_application_email(application_form)
...
+    @reject_by_default_days = TimeLimitCalculator.new(
+      rule: :reject_by_default,
+      effective_date: Time.zone.now,
+    ).call
+
+    @candidate_name = application_form.first_name
+
+    view_mail(GENERIC_NOTIFY_TEMPLATE,
+              to: application_form.candidate.email_address,
+              subject: 'Your references are in and your application is being considered')
```
tvararu: Translation string in here + test?

app/views/candidate_mailer/application_under_consideration.text.erb
```diff
@@ -0,0 +1,15 @@
...
+
+# Your application is being considered
+
+Your application is now with your teacher training provider(s).
+
+We’ve asked providers to make a final decision within <%= @reject_by_default_days %> working days.
+
+They’ll be in touch with you sooner if they want to arrange an interview.
+
+You can review your applicaiton by signing in to Apply for teacher training:
```
tvararu:
```suggestion
You can review your application by signing in to Apply for teacher training:
```

app/mailers/candidate_mailer.rb
```diff
@@ -6,4 +6,17 @@ def submit_application_email(application_form)
               to: application_form.candidate.email_address,
               subject: t('submit_application_success.email.subject'))
   end
+
+  def application_under_consideration(application_form)
+    @reject_by_default_days = TimeLimitCalculator.new(
```
tijmenb: We can get this from the database:

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/c2d11a372d3f5a9d6f4994e0a87d265be0f3d90d/db/schema.rb#L31

app/views/candidate_mailer/application_under_consideration.text.erb
```diff
@@ -0,0 +1,15 @@
+Dear <%= @candidate_name %>,
+
+# Your application is being considered
+
+Your application is now with your teacher training provider(s).
```
tijmenb: We can use `pluralize` to get rid of the `(s)`:

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/fe5d289ce603ca5ffea3b19a61acf05e4b51e672/app/views/candidate_mailer/submit_application_email.text.erb#L23

app/services/get_application_forms_ready_to_send_to_providers.rb
```diff
@@ -0,0 +1,13 @@
+class GetApplicationFormsReadyToSendToProviders
+  def self.call
+    forms = ApplicationForm
+      .joins(:application_choices)
+      .where('"application_choices"."status" = ?', 'application_complete')
+      .having('max("application_choices"."edit_by") < ?', Time.zone.now)
+      .group(:id)
+
+    forms.select do |f|
+      f.application_choices.all? { |ac| ac.status == 'application_complete' }
```
tijmenb:
```suggestion
      f.application_choices.all? { |application_choice| application_choice.status == 'application_complete' }
```

spec/system/candidate_interface/application_choices_are_delivered_to_providers_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    then_i_should_receive_an_email_saying_my_application_is_under_consideration
+  end
+
+  def given_my_application_is_ready_to_send_to_providers
+    @application_choice = create(:application_choice, :ready_to_send_to_provider)
+  end
+
+  def when_my_application_is_delivered_to_the_provider
+    SendApplicationsToProvider.new.call
```
tijmenb: Can we make this even higher level by running the sidekiq worker?

app/mailers/candidate_mailer.rb
duncanjbrown: Good point. The question then, of course, is "which application_choice's reject_by_default_days is the right one?" This is a property of the batch as a whole. Shall we leave it at `application_form.application_choices.first.reject_by_default_days` for now?

app/views/candidate_mailer/application_under_consideration.text.erb
duncanjbrown: :+1: :+1:

spec/system/candidate_interface/application_choices_are_delivered_to_providers_spec.rb
duncanjbrown: Good idea. It would be good to have an utterly generic interface to the sweeper task so specs like this could just declare 'the cron has run' and call it.

duncanjbrown:
Thanks @tvararu and @tijmenb, fixes on the way

app/views/candidate_mailer/application_under_consideration.text.erb
```diff
@@ -0,0 +1,15 @@
...
+
+We’ve asked providers to make a final decision within <%= @reject_by_default_days %> working days.
```
fofr: I think we should be giving the actual date now we know it rather than "40 working days" which isn't very helpful. cc @paulrobertlloyd

app/views/candidate_mailer/application_under_consideration.text.erb
```diff
@@ -0,0 +1,15 @@
...
+
+They’ll be in touch with you sooner if they want to arrange an interview.
+
+You can review your applicaiton by signing in to Apply for teacher training:
+
+[Sign in](<%= candidate_interface_sign_in_url %>)
```
fofr: Should we show the vanilla URL here, as we do in other emails?

paulrobertlloyd: Yeah, I think so.

duncanjbrown: I thought that was the convention, but then I looked at `submit_application_email.text.erb` where links are inlined. I suppose in that email they're just about part of sentences. 

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/c2d11a372d3f5a9d6f4994e0a87d265be0f3d90d/app/views/candidate_mailer/submit_application_email.text.erb#L21

I'll change it to whatever you suggest.

app/views/candidate_mailer/application_under_consideration.text.erb
paulrobertlloyd: Yup, `40 workings day (1 February 2020)` or something similar.

fofr: `They need to make a final decision before 1 February 2020 (40 working days)`

duncanjbrown:
Got to go to offsite. In the meantime I'll dismiss @tvararu's review so it's clear there are changes still to come

duncanjbrown:
Should all be fixed up now!

PR #741, opened 2019-11-27
Document how to roll back a failed deploy

duncanjbrown:
Add some documentation about how to roll back a failed deploy

https://trello.com/c/ASRjuzva/1258-improve-the-deployment-process


tpinney:
Can we add a note to shut down the staging slot once the swap is complete please @duncanjbrown.

tpinney:
Its possibly also worth noting that this rollback approach will only work if there are no Azure infrastructure changes associated with the deployment. In cases where the ARM template has been changed the swap slots approach may still work, but its equally likely that you'll have to re-run the pipeline for the previous release.

duncanjbrown:
Got your notes @tpinney, will update

PR #746, opened 2019-11-28
Add test helpers for Slack notifications

tijmenb:
### Context

Extracted from https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/744 to avoid that PR from getting too big.

These helpers make it easier to make sure we're sending the correct Slack messages in system tests. Used in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/744 to test new messages when candidates accept & decline offers.

### Guidance to review

Does this make sense?

### Link to Trello card

https://trello.com/c/gg1hElnB/361-candidates-can-accept-an-offer

### DIFF
--- a/spec/support/sidekiq.rb
+++ b/spec/support/sidekiq.rb
@@ -1,0 +1,7 @@
+RSpec.configure do |config|
+  config.around sidekiq: true do |example|
+    Sidekiq::Testing.inline! do
+      example.run
+    end
+  end
+end
--- a/spec/support/slack_notifications.rb
+++ b/spec/support/slack_notifications.rb
@@ -1,0 +1,12 @@
+RSpec.configure do |config|
+  config.before do
+    stub_request(:post, 'https://example.com/slack-webhook').
+      to_return(status: 200, body: '{}')
+  end
+
+  def expect_slack_message_with_text(text)
+    expect(WebMock).to have_requested(:post, 'https://example.com/slack-webhook').with { |req|
+      JSON.parse(req.body)['text'].match(text)
+    }
+  end
+end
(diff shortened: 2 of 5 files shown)


spec/support/slack_notifications.rb
```diff
@@ -0,0 +1,12 @@
+RSpec.configure do |config|
+  config.before do
+    stub_request(:post, "https://example.com/slack-webhook").
+      to_return(status: 200, body: "{}")
+  end
+
+  def expect_slack_message_with_text(text)
+    expect(WebMock).to have_requested(:post, "https://example.com/slack-webhook").with { |req|
```
duncanjbrown: nit (and I know your views on `let` :)):

```
let(:slack_request) { stub_request(...) }
```

and then 

```
expect(slack_request).to_have_been_made.with( blah )
```

spec/support/sidekiq.rb
```diff
@@ -0,0 +1,7 @@
+RSpec.configure do |config|
+  config.around sidekiq: true do |example|
```
duncanjbrown: :+1:

spec/support/slack_notifications.rb
tijmenb: Sorry!

```
       `let!` is not available from within an example (e.g. an `it` block) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc). It is only available on an example group (e.g. a `describe` or `context` block).
```

duncanjbrown: I live to stand corrected

PR #748, opened 2019-11-29
Display user progress stats in the support console

benilovj:
### Context

The product and support team needs a snapshot view of where candidates are in the application process.

Add a table of service stats in the support console:

image

### Guidance to review

- These are the initial stats for right this minute. Subsequent PRs will add:
  - more stats for later stages of the application process
  - some kind of Slack integration so the stats can be automatically posted on a daily basis
- We try not to count DfE users in the stats since they're very unlikely to be lodging real applications, but rather exploring the service

### Link to Trello card

https://trello.com/c/asLhnXMe/398-show-performance-metrics-in-the-support-console

### DIFF
--- a/app/models/performance_statistics.rb
+++ b/app/models/performance_statistics.rb
@@ -1,0 +1,35 @@
+class PerformanceStatistics
+  QUERY = "
+  WITH raw_data AS (
+      SELECT
+          COUNT(c.id) FILTER (WHERE f.id IS NOT NULL) candidate_forms,
+          SUM(CASE WHEN f.id IS NOT NULL AND f.created_at < f.updated_at THEN 1 ELSE 0 END) candidate_started_form_count,
+          SUM(CASE WHEN f.id IS NOT NULL AND f.submitted_at IS NOT NULL THEN 1 ELSE 0 END) candidate_submitted_form_count
+      FROM
+          candidates c
+      LEFT JOIN
+          application_forms f ON f.candidate_id = c.id
+      WHERE
+          c.email_address NOT LIKE '%education.gov.uk%'
+      GROUP BY
+          c.id
+  )
+  SELECT
+      COUNT(*) AS total_non_dfe_sign_ups,
+      SUM(CASE WHEN candidate_forms = 0 THEN 1 ELSE 0 END) AS candidates_signed_up_but_not_signed_in,
+      SUM(CASE WHEN candidate_forms > 0 AND candidate_started_form_count = 0 THEN 1 ELSE 0 END) AS candidates_signed_in_but_not_entered_data,
+      SUM(CASE WHEN candidate_started_form_count > 0 AND candidate_submitted_form_count = 0 THEN 1 ELSE 0 END) AS candidates_with_unsubmitted_forms,
+      SUM(CASE WHEN candidate_submitted_form_count > 0 THEN 1 ELSE 0 END) AS candidates_with_submitted_forms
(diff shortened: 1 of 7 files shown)


benilovj:
Specs coming shortly.

app/views/support_interface/performance/index.html.erb
```diff
@@ -0,0 +1,40 @@
...
+  </thead>
+  <tbody class="govuk-table__body">
+    <tr class="govuk-table__row">
+      <th scope="row" class="govuk-table__header">Total sign ups<br>(excluding DfE users)</th>
+      <th scope="row" class="govuk-table__header">
+        <%= @statistics[:total_non_dfe_sign_ups] %>
+      </th>
+    </tr>
+    <tr class="govuk-table__row">
+      <th scope="row" class="govuk-table__header"># who’ve signed up but not signed in</th>
```
fofr:
```suggestion
      <th scope="row" class="govuk-table__header">Users who signed up but have not signed in</th>
```

app/models/performance_statistics.rb
```diff
@@ -0,0 +1,33 @@
+class PerformanceStatistics
+  QUERY = "
+  WITH raw_data AS (
+      SELECT
+          SUM(CASE WHEN f.id IS NOT NULL THEN 1 ELSE 0 END) candidate_forms,
```
stevehook: I think you can use `COUNT FILTER` for these kind of stats, e.g.

```
COUNT(c.id) FILTER (WHERE f.id IS NOT NULL) AS candidate_forms,
```

stevehook: Maybe with `COUNT(DISTINCT c.id)` you could do without the common table expression and do the query in 'one go'. (I'm not clear about under what circumstances you could have a candidate with more than one application form).

benilovj: Thanks @stevehook, I've incorporated `COUNT FILTER` into the query.

I've also added some model specs which should make it easier to refactor stuff to make the query simpler, although I probably would hold off on that in this PR (my SQL is very rusty at this stage).

app/models/performance_statistics.rb
```diff
@@ -0,0 +1,33 @@
...
+      COUNT(*) AS total_non_dfe_sign_ups,
+      SUM(CASE WHEN candidate_forms = 0 THEN 1 ELSE 0 END) AS candidates_signed_up_but_not_signed_in,
+      SUM(CASE WHEN candidate_forms > 0 AND candidate_started_form_count = 0 THEN 1 ELSE 0 END) AS candidates_signed_in_but_not_entered_data,
+      SUM(CASE WHEN candidate_started_form_count > 0 AND candidate_submitted_form_count = 0 THEN 1 ELSE 0 END) AS candidates_with_unsubmitted_forms,
+      SUM(CASE WHEN candidate_submitted_form_count > 0 THEN 1 ELSE 0 END) AS candidates_with_submitted_forms
+  FROM
+      raw_data".freeze
+
+  def run
+    @results = ActiveRecord::Base.connection.execute(QUERY)[0]
```
stevehook: You could have a memoized method called results then just use `results` in the following method, e.g. 

```
def results
    @results ||= ActiveRecord::Base.connection.execute(QUERY)[0]
end
```

That way the caller doesn't have to worry about calling `run` once.

benilovj: Good idea - done.

PR #753, opened 2019-11-29
Downcase candidate email_address before saving/comparing

mnacos:
### Context

People are able to sign up twice with the same email_address (just different case).

The downcase of email_address before saving to database or comparing for uniqueness.

### Guidance to review

Try to sign up with same email address in different case.

### Link to Trello card

[586 - Emails should be downcased before saving/comparing to the database](https://trello.com/c/OTOi5VAI)

### Env vars

### DIFF
--- a/app/models/candidate.rb
+++ b/app/models/candidate.rb
@@ -2,7 +2,9 @@ class Candidate < ApplicationRecord
   # Only Devise's :timeoutable module is enabled to handle session expiry
   # Custom Warden strategy is used instead see app/warden/magic_link_token.rb
   devise :timeoutable
-  validates :email_address, presence: true, uniqueness: true, length: { maximum: 250 }
+
+  before_validation :downcase_email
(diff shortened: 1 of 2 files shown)


app/models/candidate.rb
```diff
@@ -2,7 +2,9 @@ class Candidate < ApplicationRecord
   # Only Devise's :timeoutable module is enabled to handle session expiry
   # Custom Warden strategy is used instead see app/warden/magic_link_token.rb
   devise :timeoutable
-  validates :email_address, presence: true, uniqueness: true, length: { maximum: 250 }
+
+  before_validation :downcase_email
+  validates :email_address, presence: true, uniqueness: { case_sensitive: false }, length: { maximum: 250 }
```
mnacos: This makes Rails use LOWER() when comparing in the database. For handling existing email addresses in the database that are not lowercase. In theory we could allow any case, as long as the comparison is case-insensitive, but I think it's neater to be consistent.

PR #757, opened 2019-11-29
Asynchronous email sending

stevehook:
### Context

Sending emails should happen in a worker process rather than the main Web applications because (a) sending emails can be a bit slow so we keep response times snappy by offloading this to a worker and (b) if an email service is down at the first attempt we can try again later in a worker by just re-queueing the job.

- [x] Configure ActiveJob to use sidekiq
- [x] Setup a `mailer` queue in Sidekiq (because this is what ActionMailer/ActiveJob use by default)
- [x] Switch all the calls to `deliver_now` to `deliver_later` so that they are queued to Sidekiq rather than executed inline.
- [x] Configure ActiveJob to run jobs immediately (`inline`) in tests
- [x] Update README

### Guidance to review

- Are there any gotchas relating to mailers working asynchronously?
- Have I covered them all?
- Is the change to Azure config right? (All I'm trying to do here is pass an extra command line arg to `sidekiq` for the config file)
- There are no new tests here - it's just reconfiguring existing functionality. Is there any other testing we can do?
- One inconvenience caused by this change is that in a dev environment you can no longer pick the magic sign-up link from the Rails log without Sidekiq and Redis running too. Do we need a way to bypass Sidekiq in dev (perhaps driven by an opt-in env var) so that things jobs run inline for simplicity?

### Link to Trello card

[1311 - Async Email Sending](https://trello.com/c/bwA7j7wa/1311-async-email-sending)

### Env vars

No new env vars.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,12 +40,15 @@ A service for candidates to apply for initial teacher training.
 - Yarn 1.12.x
 - PostgreSQL 9.6
 - Graphviz 2.22+ (`brew install graphviz`) to generate the [domain model diagram](#domain-model)
+- Redis 5.0.x
 
 ## <a name="dev-prerequisites"></a>Prerequisites for development
 
 `docker` and `docker-compose`
 
 ## <a name="dev-env-setup"></a>Setting up the development environment
+
+If you are not planning to run the server processes in docker:
 
 1. Copy `.env.example` to `.env` and fill in the secrets
 1. Run `make setup`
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -40,13 +40,16 @@ A service for candidates to apply for initial teacher training.
...
 - Graphviz 2.22+ (`brew install graphviz`) to generate the [domain model diagram](#domain-model)
+- Redis 5.0.x
 
 ## <a name="dev-prerequisites"></a>Prerequisites for development
 
 `docker` and `docker-compose`
 
 ## <a name="dev-env-setup"></a>Setting up the development environment
 
+If you are not planning to run the server processes in docker:
```
mnacos: ?

stevehook:
> LGTM - re: sign up links in development you could `deliver_now` if Rails.env.development? -- but foreman will also do the trick.

I think we _could_ put a line in `environment/development.rb` that configures all Sidekiq jobs to run inline in dev (same as we already do for test). We could have an env var to switch this on. It's just a question of whether it's worth doing this (worth the added complexity) as a convenience for devs. Like you say foreman or docker-compose solve the problem.

tpinney:
Azure template change is good.

README.md
stevehook: :+1: that sentence doesn't make much sense any longer. Will remove.

PR #762, opened 2019-12-02
Fix flakey vendor application spec

tvararu:
### Context

The GCSEs are not in a consistent order. We could also sort them on the presenter, but that does not feel necessary.


tvararu:
@duncanjbrown sensible! I think we can merge this and if it comes up again I'd give `deepsort` a try.

PR #764, opened 2019-12-02
Show declined by date on Application dashboard.

alexdesi:
### Context
This PR implements the Default By Default date, on the UI side.

See the screenshot.

## Note about the 'ApplicationCompleteContentComponent' class 
I believe the Components class, should contain only code coupled with the front end,
but we all are really not observing this practice, 
as you can see in many `Components` classes we access the DB, with queries or making complex computation,

This makes testing the Components pretty complex, and create duplication.

I'd suggest refactoring this, maybe moving the queries in the appropriate model (that's a pretty easy first step, and maybe later to move again the queries in a Repositories module)

What do you think?

### Link to Trello card

https://trello.com/c/dQcEJNYP/547-update-the-application-dashboard-to-show-calculated-dbd-date

### DIFF
--- a/spec/components/application_complete_content_component_spec.rb
+++ b/spec/components/application_complete_content_component_spec.rb
@@ -2,7 +2,7 @@ require 'rails_helper'
 
 RSpec.describe ApplicationCompleteContentComponent do
   let(:submitted_at) { Time.zone.local(2019, 10, 22, 12, 0, 0) }
-  let(:first_january2019) { Time.zone.local(2020, 1, 1, 12, 0, 0) }
+  let(:first_january_2020) { Time.zone.local(2020, 1, 1, 12, 0, 0) }
 
   around do |example|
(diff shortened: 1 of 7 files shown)


spec/components/application_complete_content_component_spec.rb
```diff
@@ -70,7 +70,7 @@
 
       expect(render_result.text).to include('All your training providers have now reached a decision')
       # TODO: Update hardcoded date with decline by default date
```
chubberlisk:
```suggestion
```

spec/components/application_complete_content_component_spec.rb
```diff
@@ -81,7 +81,7 @@
 
       # TODO: Update hardcoded date with decline by default date
```
chubberlisk:
```suggestion
```

PR #765, opened 2019-12-02
Fix email validation within candidate_interface/sign_up_form

mnacos:
### Context

We introduced case-insensitive email_address validation in the model but the SignUpForm was doing its own email_address validation and the logic wasn't in sync.

Refactor SignUpForm so that all email_address validation happens within the model.

### Guidance to review

Run the tests! Read the code. Try to reproduce the bug.

### Link to Trello card

[595 - Can't sign up/in with uppercase version of existing email](https://trello.com/c/Kq1fmTFm)

### Env vars

No new env vars.

### DIFF
--- a/app/models/candidate_interface/sign_up_form.rb
+++ b/app/models/candidate_interface/sign_up_form.rb
@@ -1,30 +1,35 @@
 module CandidateInterface
   class SignUpForm
     include ActiveModel::Model
+    attr_accessor :email_address, :accept_ts_and_cs, :candidate
 
-    attr_accessor :email_address, :accept_ts_and_cs
     validates :email_address, :accept_ts_and_cs, presence: true
-    validates :email_address, length: { maximum: 250 }
-    validates :email_address, email_address: true
+    validate :candidate_email_address_is_valid
 
+    def initialize(params = {})
+      @email_address = params[:email_address]
+      @accept_ts_and_cs = params[:accept_ts_and_cs]
+      @candidate = Candidate.for_email @email_address
+    end
 
-    def save(candidate)
-      unless candidate.valid?
+    def existing_candidate?
+      candidate.persisted?
+    end
+
+    def save
+      return false if existing_candidate? || !valid?
+
+      candidate.save
+    end
+
+  private
+
+    def candidate_email_address_is_valid
+      if candidate && candidate.invalid?
         candidate.errors[:email_address].each do |error|
           errors.add(:email_address, error)
         end
       end
-
-      # need the explicit empty? check on errors as we're
-      # delegating email address validation to the candidate
-      # and just copying the messages
-      return false unless valid? && errors.empty?
-
-      candidate.update!(email_address: email_address)
-    end
-
-    def self.build_from_candidate(candidate)
-      new(email_address: candidate.email_address)
     end
   end
 end
--- a/app/models/reference.rb
+++ b/app/models/reference.rb
@@ -3,7 +3,7 @@ class Reference < ApplicationRecord
   validates :email_address, presence: true,
                             email_address: true,
                             length: { maximum: 100 },
-                            uniqueness: { scope: :application_form_id }
+                            uniqueness: { case_sensitive: false, scope: :application_form_id }
   validate :email_address_not_own
   validates :relationship, presence: true, word_count: { maximum: 50 }
   validates_presence_of :application_form_id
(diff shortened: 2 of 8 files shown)


spec/system/candidate_interface/candidate_account_spec.rb
```diff
@@ -40,6 +40,14 @@
 
     when_i_signed_in_more_than_a_week_ago
     then_i_should_be_signed_out
+
+    when_i_click_the_signin_link
+    and_i_submit_my_email_address(uppercase: true)
```
tijmenb: These steps should read as English if possible, perhaps this can be:

```suggestion
    and_i_submit_my_email_address_in_uppercase
```

app/models/candidate_interface/sign_up_form.rb
```diff
@@ -1,30 +1,36 @@
...
     validates :email_address, :accept_ts_and_cs, presence: true
-    validates :email_address, length: { maximum: 250 }
-    validates :email_address, email_address: true
+    validate :candidate_email_address_is_valid
 
+    def initialize(params = {})
+      @email_address = params[:email_address]
+      @accept_ts_and_cs = params[:accept_ts_and_cs]
+      @candidate = Candidate.for_email @email_address
+      @candidate ||= Candidate.new email_address: @email_address
```
tijmenb: I'm a bit confused by this. `Candidate.for_email` will only return `nil` if `@email_adress` is nil, so this line will return a `Candidate.new` without an email address?

app/models/vendor_api_user.rb
```diff
@@ -1,7 +1,7 @@
 class VendorApiUser < ApplicationRecord
   belongs_to :vendor_api_token
 
-  validates :email_address, presence: true
+  validates :email_address, presence: true, uniqueness: { case_sensitive: false }
```
tijmenb: Are we sure these should be unique?

app/models/candidate_interface/sign_up_form.rb
mnacos: Removing line 13, it is confusing. It was just fixing a couple of rspec validation errors when email_address is nil. I've dealt with this in a different way now, within candidate_email_address_is_valid.

app/models/vendor_api_user.rb
mnacos: No, will remove for now

spec/system/candidate_interface/candidate_account_spec.rb
```diff
@@ -40,6 +40,14 @@
...
     when_i_signed_in_more_than_a_week_ago
     then_i_should_be_signed_out
+
+    when_i_click_the_signin_link
+    and_i_submit_my_email_address_in_uppercase
+    then_i_receive_an_email_with_a_signin_link
+
+    when_i_visit_the_signup_page
+    and_i_submit_my_email_address_in_uppercase
+    then_i_receive_an_email_with_a_signin_link
```
mnacos: @tvararu I added your system spec but changed the L50 to expect a signin link. Is this spec what you'd expect and does it cover the sign-in bug you described?

tvararu:
Tested this locally and it seems to work really well 👍

spec/system/candidate_interface/candidate_account_spec.rb
tvararu: 👍

app/models/reference.rb
```diff
@@ -3,7 +3,7 @@ class Reference < ApplicationRecord
   validates :email_address, presence: true,
                             email_address: true,
                             length: { maximum: 100 },
-                            uniqueness: { scope: :application_form_id }
+                            uniqueness: { case_sensitive: false, scope: :application_form_id }
```
tvararu: This seems unrelated to the other changes, and doesn't come with a corresponding test, but I think it's fine to sneak it in

mnacos:
> Adding this temporary red review as I want to deploy the current master to prod and then merge this in + do another deploy

Yep that's cool, just make sure you revert this before merging: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/774/files

mnacos:
> Adding this temporary red review as I want to deploy the current master to prod and then merge this in + do another deploy

Cool, I've rebased this against latest master, so it's ready to be merged when you are ready to deploy.

PR #768, opened 2019-12-02
Remove n+1 queries using bullet

duncanjbrown:
## Context

Load testing revealed a very heavy n+1 query behind the Vendor API. Nix it and all of its kind using `bullet`

## Changes proposed

Add `bullet`. The build will now fail if we introduce n+1 queries.

https://trello.com/c/e11QWmbi/1351-vendor-api-application-index-is-very-slow

### DIFF
--- a/app/components/referees_review_component.html.erb
+++ b/app/components/referees_review_component.html.erb
@@ -1,4 +1,4 @@
-<% @application_form.references.each do |referee| %>
+<% @application_form.references.includes(:application_form).each do |referee| %>
   <%= render(SummaryCardComponent, rows: referee_rows(referee), editable: @editable) do %>
     <%= render(SummaryCardHeaderComponent, title: t('application_form.referees.nth_referee')[referee.ordinal], heading_level: @heading_level) do %>
       <% if @editable %>
--- a/app/services/get_application_choices_for_provider.rb
+++ b/app/services/get_application_choices_for_provider.rb
@@ -4,10 +4,18 @@ class GetApplicationChoicesForProvider
   def self.call(provider:)
     raise MissingProvider unless provider.present?
 
-    ApplicationChoice.includes(:course)
+    includes = [
+      :course,
+      :application_form,
+      :provider,
+      :site,
+      application_form: %i[candidate references application_work_experiences application_volunteering_experiences],
+    ]
+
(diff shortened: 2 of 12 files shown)


app/services/get_application_choices_for_provider.rb
```diff
@@ -4,10 +4,10 @@ class GetApplicationChoicesForProvider
   def self.call(provider:)
     raise MissingProvider unless provider.present?
 
-    ApplicationChoice.includes(:course)
+    ApplicationChoice.includes(:course, :application_form, :provider, :site, application_form: [:candidate, :references, :application_work_experiences, :application_volunteering_experiences])
       .where('courses.provider_id' => provider)
-      .or(ApplicationChoice.includes(:course)
-      .where('courses.accrediting_provider_id' => provider))
+      .or(ApplicationChoice.includes(:course, :application_form, :provider, :site, application_form: [:candidate, :references, :application_work_experiences, :application_volunteering_experiences])
```
tijmenb: 😮

duncanjbrown: ikr! Pretty sure this isn't useful in all contexts either...

duncanjbrown:
Updated:

- Deduplicated the big `include` in `GetApplicationChoicesForProvider`
- Added `deepsort` now that we've had to sort twice in `Vendor receives an application` (see #762)

app/components/referees_review_component.html.erb
```diff
@@ -1,4 +1,4 @@
-<% @application_form.references.each do |referee| %>
+<% @application_form.references.includes(:application_form).each do |referee| %>
```
stevehook: This one surprised me. I thought that the reverse association was populated when you load a collection but doesn't seem to be the case. There is something very odd about `ApplicationForm#references` which has tripped us up before.

`application_form.references.last.application_form` lazy loads `application_form` but `application_form.application_choices.last.application_form` does not. Confused.

stevehook: As discussed its the `-> { order('id ASC') }`

duncanjbrown: Removing the `order` does not stop bullet complaining here. Propose leaving it for now and returning to rename `references` to `application_references` in the hope of reducing the territory this problem has to roam about in?

PR #769, opened 2019-12-03
Allow support users to impersonate candidates

tijmenb:
### Context

Now that we've built the first steps in the application process we're building functionality that happens later in the process. For example, we need to test applications that have received an offer to test https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/744. This should make that easier.

This adds a button to the application form that allows a support user to sign in as a candidate. This allows for easier manual testing by people in the team - it avoids us from having to sign in manually.

I've enabled this only on test environments, because user impersonation is a bit perilous when it comes to audibility and non-repudiation (we don't want to accidentally accept an offer on behalf of a real candidate).

impersonation

### Guidance to review

Test that it works.

### Link to Trello card

Part of https://trello.com/c/49yLJZEW


duncanjbrown:
Luxury UX would be if clicking "sign out" when impersonating a candidate returned you to the candidate's page in `/support`

PR #770, opened 2019-12-03
Put the Support Interface behind DfE Sign-in

stevehook:
DO NOT MERGE UNTIL WE HAVE SET UP DfE Sign-in SUPPORT USERS ON QA/STAGING/PRODUCTION

### Context

Currently the support interface is authenticated with basic auth. This is not very secure and means that we cannot grant access to individuals or audit their activity.

- [x] Authenticate users of the support interface with DfE Sign-in. Users can log in with one DfE Sign-in account that could have access to the provider interface, support interface or both.
- [x] Authorise users of the support interface by looking up their email address on a list stored in the database. For authenticated users that are not authorized we render a new 'Not authorized' template (because the one for providers isn't appropriate). (Copy needs to be reviewed).
- [x] Change the DfE Sign-in callback action so that it can serve both interfaces. To do this we now store the 'target' path that the unauthenticated user originally attempted to access in the user session and redirect them here following a successful login. Previously they were redirected to a hard-coded path in the provider interface.
- [x] Handle the situation in which the user went straight to the login page for the support or provider interface (without setting the target path session variable).
- [x] Ensure that we still handle the situation in which a provider can log in but has not yet been granted access.
- [x] Add a rake task to create a support user (or just update it's email address if it already exists). Usage: `bundle exec rails "create_support_user[alice, alice@example.com]"` where `alice` is the DfE Sign-in UID and `alice@example.com` is the email address.
- [x] Update README?
- [ ] Work out a list of email addresses that require access and configure these before we deploy to production. (We can add these using the rake task above or direct DB access).

dfe-sign-in

### Guidance to review

- Are there any loopholes that would allow users to access the support interface unauthenticated or without the correct authorisation?
- Are the tests covering everything they should? e.g. failed login as well as success
- What are the challenges with rolling this out other than having to configure the list of users that should have access? Based on what we discussed at the incident review this should have been broken up somehow and/or feature flagged. Not sure how that would work exactly.
- Content review for the 'Not authorized' page

### Link to Trello card

[1307 - Put support behind DfE Sign-in](https://trello.com/c/DVh2NV9f/1307-put-support-behind-dfe-sign-in)

### Env vars

- no new env vars

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,8 +142,8 @@ GenerateTestData.new(10, Provider.find_by(code: '1N1')).generate
 
 ## <a name="dfe-sign-in"></a>DfE Sign-in
 
-The Provider interface at `/provider` is protected by DfE's SSO provider DfE
-Sign-in.
+The Provider interface at `/provider` and Support interface at
+`/support` are both protected by DfE's SSO provider DfE Sign-in.
 
 ### Environments
 
--- a/app/controllers/provider_interface/sessions_controller.rb
+++ b/app/controllers/provider_interface/sessions_controller.rb
@@ -20,7 +20,7 @@ module ProviderInterface
     end
 
     def authorized_for_support_interface?
-      SupportUser.load_from_session(session)&.authorized?
+      SupportUser.load_from_session(session)
     end
   end
 end
--- a/app/controllers/provider_interface/sessions_controller.rb
+++ b/app/controllers/provider_interface/sessions_controller.rb
@@ -7,13 +7,15 @@ module ProviderInterface
 
     def callback
       dfe_sign_in_session = DfESignIn.parse_auth_hash(request.env['omniauth.auth'])
-      ProviderUser.begin_session!(session, dfe_sign_in_session)
+      DfESignInUser.begin_session!(session, dfe_sign_in_session)
 
-      redirect_to provider_interface_path
+      # TODO: What if the given user doesn't have permission to visit
+      # the provider interface?
+      redirect_to session['post_dfe_sign_in_path'] || provider_interface_path
     end
 
     def destroy
-      ProviderUser.end_session!(session)
+      DfESignInUser.end_session!(session)
 
       redirect_to action: :new
     end
--- a/app/controllers/provider_interface/sessions_controller.rb
+++ b/app/controllers/provider_interface/sessions_controller.rb
@@ -11,7 +11,7 @@ module ProviderInterface
 
       # TODO: What if the given user doesn't have permission to visit
       # the provider interface?
-      redirect_to session['post_dfe_sign_in_path'] || provider_interface_path
+      redirect_to session.delete('post_dfe_sign_in_path') || provider_interface_path
     end
 
     def destroy
--- a/app/controllers/provider_interface/sessions_controller.rb
+++ b/app/controllers/provider_interface/sessions_controller.rb
@@ -1,18 +1,7 @@
 module ProviderInterface
   class SessionsController < ProviderInterfaceController
     skip_before_action :authenticate_provider_user!
-    protect_from_forgery except: :bypass_callback
-
     def new; end
-
-    def callback
-      dfe_sign_in_session = DfESignIn.parse_auth_hash(request.env['omniauth.auth'])
-      DfESignInUser.begin_session!(session, dfe_sign_in_session)
-
-      # TODO: What if the given user doesn't have permission to visit
-      # the provider interface?
-      redirect_to session.delete('post_dfe_sign_in_path') || provider_interface_path
-    end
 
     def destroy
       DfESignInUser.end_session!(session)
@@ -20,6 +9,18 @@ module ProviderInterface
       redirect_to action: :new
     end
 
-    alias :bypass_callback :callback
+  private
+
+    def default_authenticated_path
+      if authorized_for_support_interface?
+        support_interface_path
+      else
+        provider_interface_path
+      end
+    end
+
+    def authorized_for_support_interface?
+      SupportUser.load_from_session(session)&.authorized?
+    end
   end
 end
--- a/app/controllers/support_interface/support_interface_controller.rb
+++ b/app/controllers/support_interface/support_interface_controller.rb
@@ -24,7 +24,7 @@ module SupportInterface
     end
 
     def dfe_sign_in_user
-      ProviderUser.load_from_session(session)
+      DfESignInUser.load_from_session(session)
     end
 
     def authenticate_support_user!
--- a/app/controllers/support_interface/support_interface_controller.rb
+++ b/app/controllers/support_interface/support_interface_controller.rb
@@ -24,7 +24,10 @@ module SupportInterface
     end
 
     def authenticate_support_user!
-      redirect_to support_interface_sign_in_path unless current_support_user
+      return if current_support_user
+
+      session['post_dfe_sign_in_path'] = request.path
+      redirect_to support_interface_sign_in_path
     end
   end
 end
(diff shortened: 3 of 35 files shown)


spec/system/support_interface/authentication_spec.rb
```diff
@@ -24,7 +26,14 @@ def when_i_visit_the_support_interface
     visit support_interface_applications_path
   end
 
-  def and_i_sign_in_via_dfe_sign_in
+  def then_i_should_be_redirected_to_login_page
+    expect(page).to have_current_path support_interface_sign_in_path
+    expect(page).not_to have_link 'Applications'
```
duncanjbrown: It's surely right to hide this stuff on the sign-in page but maybe it's a separate assertion in this spec? Or maybe it doesn't need testing here?

spec/requests/support_interface/comments_spec.rb
```diff
@@ -5,14 +5,20 @@ def create_application_form
...
+    )
+  end
+
+  def set_provider_permission
+    allow(SupportUser).to receive(:load_from_session).and_return(support_user)
+    allow(support_user).to receive(:authorized?).and_return(true)
+  end
+
+  before do
+    set_provider_permission
```
duncanjbrown: Could this method be renamed for the support context?

spec/requests/support_interface/not_authorized_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    before do
+      allow(SupportUser).to receive(:load_from_session)
+        .and_return(
+          SupportUser.new(
+            email_address: 'email@example.com',
+            dfe_sign_in_uid: 'DFE_SIGN_IN_UID',
+          ),
+      )
+
+      # do not grant the user permission to view a provider's applications
```
duncanjbrown: Would it be worth clarifying this comment to say that the permission here is to view/use support?

app/controllers/support_interface/support_interface_controller.rb
```diff
@@ -16,5 +18,25 @@ def protect_with_basic_auth
     def render_404
       render 'errors/not_found', status: :not_found
     end
+
+    def current_support_user
+      SupportUser.load_from_session(session)
```
duncanjbrown: Given that there's nothing to be done if the support user is not authorized, what would be the effect of making an  returning `nil` if the user is unauthorized?. Maybe `load_from_session` itself could return nil? Not sure of the right place to cut it.

app/controllers/provider_interface/sessions_controller.rb
```diff
@@ -9,17 +9,29 @@ def callback
...
 
       redirect_to action: :new
     end
 
     alias :bypass_callback :callback
+
+  private
+
+    def default_authenticated_path
+      if authorized_for_support_interface?
```
duncanjbrown: This feels right: try support first, then provider

app/models/dfe_sign_in_user.rb
```diff
@@ -0,0 +1,25 @@
+module DfESignInUser
+  extend ActiveSupport::Concern
+
+  def self.begin_session!(session, dfe_sign_in_session)
+    session['dfe_user'] = {
```
duncanjbrown: I think this implies the user is from the DfE itself — maybe `dfe_sign_in_user`?

README.md
```diff
@@ -146,6 +146,21 @@ maps to provider code `ABC`. This means that if you have generated [dummy data](
...
+accounts that have access to the Support interface based on their DfE
+Sign-in UID. There is only one privilege level, either you have access
+to everything or nothing.
+
+You can add a new support user using the `create_support_user` rake
+task. You need to supply a DfE Sign-in UID and an email address, e.g.
+
+    $ bundle exec rails "create_support_user[alice, alice@example.com]"
+
+Note that only the UID is used for lookup. The email address serves only
```
duncanjbrown: :+1: label sounds like a good idea

app/views/support_interface/unauthorized.html.erb
```diff
@@ -0,0 +1,19 @@
...
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">
+      Your account is not authorized
+    </h1>
+    <p class="govuk-body">
+      You’re seeing this page because you’ve signed in
+      to the <%= t('service_name.apply') %> service but your account
+      is not authorized to view the support interface.
```
duncanjbrown: Ought we to be a bit more uninformative about what's here?

spec/system/support_interface/authentication_spec.rb
stevehook: Have created a separate `and_i_should_not_see_support_menu` method for these assertions.

app/controllers/support_interface/support_interface_controller.rb
stevehook: This bit does feel messy. The tricky bit is dealing with users that are logged in but don't have authorisation. We can either leave them logged in (in a kind of limbo state) and tell them they don't have authorisation or immediately end their session. I went with the first option mainly because it mirrors what we do in the provider interface but also to avoid people thinking they've mistyped a password or something when it's just a permission thing.

app/views/support_interface/unauthorized.html.erb
stevehook: We can say 'not authorized to view this page'.?

spec/system/support_interface/sign_in_as_candidate_spec.rb
```diff
@@ -10,7 +12,9 @@
   end
 
   def given_i_am_a_support_user
-    page.driver.browser.authorize('test', 'test')
+    support_user_exists_in_dfe_sign_in(email_address: 'user@apply-support.com', dfe_sign_in_uid: 'abc')
```
tijmenb: Perhaps these three lines can go into a `sign_in_as_support_user` helper?

spec/system/support_interface/unauthorized_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    when_i_visit_the_support_interface
+    then_i_should_be_redirected_to_login_page
+
+    when_i_sign_in_via_dfe_sign_in
+
+    then_i_should_be_see_the_not_authorized_page
+    and_i_should_see_my_email_address
+    and_i_should_not_see_support_menu
+
+    when_i_click_sign_out
```
tijmenb: I don't think we have to test sign out in this spec, it's already tested in the other spec.

app/models/support_user.rb
```diff
@@ -0,0 +1,7 @@
+class SupportUser < ActiveRecord::Base
```
tijmenb: Open question - there's an interesting inconsistency between provider and support: 

- support uses both `DfESignInUser` and `SupportUser` (an AR model)
- the provider interface only uses `ProviderUser`, which is a sublass of `DfESignInUser`

How would we make those two things behave the same?

stevehook: It is inconsistent right now. Duncan and I were discussing yesterday. He wants to put 'the list' of authorised provider users into the database too (having a .yml file means redeploying to make changes). He was keen to get started on this yesterday but was waiting on my PR. So it's going to change some more and (I think) become a lot more consistent.

Right now `DfESignInUser` represents an authenticated user whereas `SupportUser` represents an authorised user. On the Provider side it's a bit different because the authorisation check is done at the point the user attempts to query for applications rather than immediately after authentication.

tijmenb:
👏 @steve

PR #771, opened 2019-12-03
make FeatureFlag connect to hardcoded Redis db in test mode

mnacos:
### Context

A previous PR introduced hardcoding the Redis database in Rails test environment so that any Sidekiq tasks generated in the tests do not pollute the local development environment.

This makes rspec tests also connect to a different Redis database.

### Link to Trello card

[591 - Local development with feature flags affected by Redis shared environment](https://trello.com/c/xYFYzeuY)

### Env vars

 No new env vars

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -1,3 +1,4 @@
+REDIS_URL=redis://localhost:6379/9
 FIND_BASE_URL=http://find-test/api/v3/
 SUPPORT_USERNAME=test
 SUPPORT_PASSWORD=test
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -1,15 +1,7 @@
 require 'workers/audit_trail_attribution_middleware'
 
-redis_url = ENV.fetch('REDIS_URL') rescue 'redis://localhost:6379/1'
-redis_url = redis_url.gsub(/\/\d+$/, '/9') if Rails.env.test? # hardcode db to isolate test data
-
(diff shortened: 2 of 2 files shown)


mnacos:
> If we set this here can we delete the extra config added in #763? (Except the `flushdb`)

In theory yes, but I'd rather err on the side of caution. The code in #763 is meant to code with multiple setups, e.g. when REDIS_URL is not set (e.g. during assets:precompile) or when it needs to be different, e.g. local tests are fine with localhost but docker-based local tests are not (REDIS_URL is redis://redis:6379), not to mention Azure-based setups and CI pipelines.

The intention of the code is to set a default if REDIS_URL is not set, but respect the value of REDIS_URL if it is set, just change the database number to 9.

mnacos:
> The intention of the code is to set a default if REDIS_URL is not set, but respect the value of REDIS_URL if it is set, just change the database number to 9.

In fact, perhaps it's better to apply similar logic to FeatureFlag

mnacos:
> > The intention of the code is to set a default if REDIS_URL is not set, but respect the value of REDIS_URL if it is set, just change the database number to 9.
> 
> In fact, perhaps it's better to apply similar logic to FeatureFlag

Ok, after chatting in person we'll push the ```.env.test``` configuration through the system and find out what breaks.

tijmenb:
Don't forget to rebase & squash before merging!

PR #775, opened 2019-12-03
Add conditional showing of Science GCSE on Your application page

chubberlisk:
### Context

A candidate is only required to have a Science GCSE if the level of the course they have chosen is a primary one. Currently, we always show the Science GCSE section on the `Your application` page but this should be conditional on the courses chosen.

This PR adds the showing of Science GCSE on the `Your application` page by adding a method called `science_gcse_needed?` to `ApplicationForm`. This returns `true` if any of courses chosen are primary and `false` for any other level.

We've also added in a feature flag called `conditional_science_gcse` as we'll need to always show the Science GCSE until we add conditional validation when reviewing and submitting an application in our next PR.

### Screenshots

image

### Guidance to review

Upcoming PRs will look at not showing it in the application review and not validating it at submission.

### Link to Trello card

[401 - Primary course conditional](https://trello.com/c/cTiIxGpC/401-primary-course-conditional)

### DIFF
--- a/app/controllers/candidate_interface/application_form_controller.rb
+++ b/app/controllers/candidate_interface/application_form_controller.rb
@@ -6,6 +6,7 @@ module CandidateInterface
       return redirect_to candidate_interface_application_form_path if params[:token]
 
       @application_form_presenter = CandidateInterface::ApplicationFormPresenter.new(current_application)
+      @application_form = current_application
     end
 
     def review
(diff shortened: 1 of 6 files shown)


fofr:
### For reference

This is how `manage-courses-backend` does this for Publish and Find:
https://github.com/DFE-Digital/manage-courses-backend/blob/24ae9c34c11c6479990132ab86374988c3c2c8d6/app/models/course.rb#L314-L328

It caters for the "Further education" case, where no GCSEs are needed. We won't get to that for a while.

The rules are set out in:
https://www.gov.uk/government/publications/initial-teacher-training-criteria/initial-teacher-training-itt-criteria-and-supporting-advice#c11-gcse-standard-equivalent

PR #641, opened 2019-11-21
Content terms of use

laurahermionetennant92:
Content terms of use

### Context

Why are you making this change? What might surprise someone about it?

### Changes proposed in this pull request

If there are UI changes, please include Before and After screenshots.

### Guidance to review

How could someone else check this work? Which parts do you want more feedback on?

### Link to Trello card

[123 - Example Trello card](http://trello.com/123-example-card)

### Env vars

- [ ] If this PR introduces new environment variables, they have been added to all necessary parts of the Azure config based on the [documentation in the README](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training#azure-hosting-devops-pipeline)

PR #778, opened 2019-12-03
Remove "Guidance" from terms of use.

fofr:
Avoid legal questions around combining terms and guidance.

While the terms of use are user friendly, and give a good overview of what you can and can’t do, they are terms and not guidance. 

Where guidance is needed, the service is already doing this in context. If we do need an overview, it probably shouldn't be bundled with the terms.

[Slack message](https://ukgovernmentdfe.slack.com/archives/CAHBLU6ET/p1574356545238400?thread_ts=1574353437.232700&cid=CAHBLU6ET) for further context.

This undoes #641 which I think was merged too early.

Screen Shot 2019-12-03 at 13 46 24


tijmenb:
Merged because Paul is out of the office today.

PR #780, opened 2019-12-03
Use aria for simpler complete section links

paulrobertlloyd:
### Context

We can use `aria-describedby` to simplify how we annotate ‘Complete section’ links, reducing code, localisable strings and code duplication. Nice spot, @adamsilver!

### DIFF
--- a/spec/system/candidate_interface/candidate_reviewing_application_spec.rb
+++ b/spec/system/candidate_interface/candidate_reviewing_application_spec.rb
@@ -145,15 +145,17 @@ RSpec.feature 'Candidate reviews the answers' do
   end
 
   def then_i_should_see_an_incomplete_banner_for(section)
-    expect(page).to have_content "Complete #{t("review_application.#{section}.complete_section_visually_hidden")} section"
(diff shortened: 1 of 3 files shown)


spec/system/candidate_interface/candidate_reviewing_application_spec.rb
```diff
@@ -145,15 +145,17 @@ def check_and_fill_in_section_for(section)
...
 
   def then_i_should_see_an_incomplete_banner_for(section)
-    expect(page).to have_content "Complete #{t("review_application.#{section}.complete_section_visually_hidden")} section"
+    expect(page).to have_selector "[aria-describedby='missing-#{section}']"
   end
 
   def when_i_click_to_complete_section_for(section)
-    click_link "Complete #{t("review_application.#{section}.complete_section_visually_hidden")} section"
+    within "#missing-#{section}-error" do
+      click_link "Complete section"
```
tvararu:
```suggestion
      click_link 'Complete section'
```

PR #783, opened 2019-12-03
Remove return redirect from application_form#show

tvararu:
### Context

Now that we authenticate on a separate route, we don't need to remove the token param on the `application_form#show` route.

Remove some redundant code.

### Guidance to review

Just a minor refactor, there should not be a need to test this.

### Link to Trello card

[552 - Clicking on confirm email link does not sign you out of the previously signed in account](https://trello.com/c/ZJFXK7Tt/552-clicking-on-confirm-email-link-does-not-sign-you-out-of-the-previously-signed-in-account)


tvararu:
TODO: Sort of unrelated to this, but double check if we validate for the token expiring after 1 hour when we sign in with another user when a current user is signed in.

PR #785, opened 2019-12-03
Prevent logging of email addresses during signing-in

mnacos:
### Context

Sign-in POST requests log email addresses.

Change the logging of params to log only ```query_parameters``` and ```path_parameters```, not ```request_parameters```.

Exclude query_parameters which appear in ```config/initializers/filter_parameter_logging.rb```.

Add candidate_id information to the logs when the user signs in.

### Guidance to review

Use ```LOGSTASH_ENABLE=true``` in your local development environment (requires rails restart), perform actions and observe what gets logged.

Here's what I see:
image

### Link to Trello card


app/lib/log_request_params.rb
```diff
@@ -1,6 +1,7 @@
 module LogRequestParams
   IGNORE_PARAMS = [
     /^authenticity_token$/,
```
tijmenb: Replace with checking `filter_parameters` blacklist to filter query params.

tijmenb: And add `token` to `https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/config/initializers/filter_parameter_logging.rb

app/lib/log_request_params.rb
```diff
@@ -11,7 +12,12 @@ def self.included(base)
   end
 
   def add_params_to_request_store
-    keys_to_ignore = IGNORE_PARAMS.map { |regexp| params.keys.grep(regexp) }.flatten.uniq
-    RequestLocals.store[:params] = params.except(*keys_to_ignore)
+    if request.get?
+      keys_to_ignore = IGNORE_PARAMS.map { |regexp| params.keys.grep(regexp) }.flatten.uniq
+      RequestLocals.store[:params] = params.except(*keys_to_ignore)
```
tijmenb: Use `query_params` instead of `params`, so we don't include POST params (and we don't need the conditional).

mnacos: Adding path_parameters, too, which gives us :controller and :action in the params as before, but also params set from the url/path (e.g. course_id from /support/courses/:course_id)

PR #786, opened 2019-12-03
Do not log SQL in production

mnacos:
### Context

Our logs from various environments seem to include SQL statements. This can expose email addresses, as well as magic link tokens etc.

Change to log_level of ActiveRecord::Base to prevent this.

### Guidance to review

You can remove the ```if Rails.env.production?``` guard and switch ```LOGSTASH_ENABLE=true``` in development to see that SQL is being suppressed (e.g. when signing-in).

### Link to Trello card

### DIFF
--- a/config/initializers/active_record_logger.rb
+++ b/config/initializers/active_record_logger.rb
@@ -1,0 +1,1 @@
+ActiveRecord::Base.logger.level = Logger::INFO if Rails.env.production?


duncanjbrown:
Is this necessary now that we have https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/commit/639cde305a0d2d60cfa3c46cc8baa9a7451bca32?

mnacos:
> Is this necessary now that we have [639cde3](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/commit/639cde305a0d2d60cfa3c46cc8baa9a7451bca32)?

Changing the log_level globally seems to have done the trick, judging by the logs since yesterday. Closing

PR #787, opened 2019-12-03
Prevent email addresses in the logs through ActionMailer/Sidekiq

mnacos:
### Context

Now that ActionMailer emails are sent in background jobs, there are log statements exposing email addresses at the time of enqueuing and at the time of executing the job.

image

Pass ```candidate``` to the relevant mailers, rather than ```to```, which is an email address. Rails now will create a GID for the Candidate model, not serialise the object.

### Guidance to review

Set  ```LOGSTASH_ENABLE=true```, run rails/sidekiq/clockwork locally and follow some sign-in/sign-up flows.

### Link to Trello card

### DIFF
--- a/app/mailers/authentication_mailer.rb
+++ b/app/mailers/authentication_mailer.rb
@@ -1,17 +1,17 @@
 class AuthenticationMailer < ApplicationMailer
-  def sign_up_email(to:, token:)
+  def sign_up_email(candidate:, token:)
     @magic_link = "#{candidate_interface_authenticate_url}?token=#{token}"
 
     view_mail(GENERIC_NOTIFY_TEMPLATE,
-              to: to,
+              to: candidate.email_address,
               subject: t('authentication.sign_up.email.subject'))
   end
 
-  def sign_in_email(to:, token:)
+  def sign_in_email(candidate:, token:)
     @magic_link = "#{candidate_interface_authenticate_url}?token=#{token}"
 
     view_mail(GENERIC_NOTIFY_TEMPLATE,
-              to: to,
+              to: candidate.email_address,
               subject: t('authentication.sign_in.email.subject'))
   end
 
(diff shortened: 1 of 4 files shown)


config/initializers/active_job_logger.rb
```diff
@@ -0,0 +1 @@
+ActiveJob::Base.logger = nil if Rails.env.production?
```
duncanjbrown: Is there some way we could suppress the PII without switching off `ActiveJob` logging altogether?

(unrelated: Would `config.active_job.logger` in `production.rb` be more idiomatic?)

mnacos:
> I'd prefer to solve this by making sure that the arguments to sidekiq jobs do not contain PII. We can create a separate job that triggers after creation, which sends the email and slack message. That only needs a `candidate_id` in the args. WDYT?

I agree but we'll need to write our own workers, as ActionMailer jobs pass all arguments to the mailer as arguments for the job. I think we should still merge this in the meantime and when we have these new workers bring back :info log_level for sidekiq.

config/initializers/active_job_logger.rb
mnacos: Not really, because of the way ActionMailer jobs pass arguments to the mailer (see Tijmen's comment/reply). If we switch to Sidekiq workers for mailing we could bring ActiveJob logging back, but then again, we won't be using ActiveJob.

mnacos:
> > I'd prefer to solve this by making sure that the arguments to sidekiq jobs do not contain PII. We can create a separate job that triggers after creation, which sends the email and slack message. That only needs a `candidate_id` in the args. WDYT?
> 
> I agree but we'll need to write our own workers, as ActionMailer jobs pass all arguments to the mailer as arguments for the job. I think we should still merge this in the meantime and when we have these new workers bring back :info log_level for sidekiq.

We've decided to change all mailers to accept a candidate, rather than an email address.

mnacos:
Have updated this to pass ```candidate``` rather than ```email_address``` to the mailers, except during the ```sign_in_without_account_email``` flow.

PR #791, opened 2019-12-04
Bug: Incorrect validation for dob in personal details

alexdesi:
### Context

This prevents to fill in data that:
- do not start with digits 1 or 2
- are not 4 digits long 

Check also for the year correctness, not just for `Date.valid_date?(*date_args)`, in which (for example) 1 or 9999999 would be considered valid years (they should not).

### Link to Trello card

554-incorrect-validation-is-incorrect-for-dob-in-personal-details


spec/models/candidate_interface/personal_details_form_spec.rb
```diff
@@ -109,9 +109,9 @@
     end
 
     it 'returns a date for a valid day/month/year' do
-      personal_details = CandidateInterface::PersonalDetailsForm.new(day: '2', month: '8', year: '802701')
+      personal_details = CandidateInterface::PersonalDetailsForm.new(day: '2', month: '8', year: '1990')
```
tvararu: RIP "Time Machine" by HG Wells reference :(

app/lib/validation_utils.rb
```diff
@@ -0,0 +1,5 @@
+module ValidationUtils
+  def valid_year?(year)
+    year.to_s.match?(/^[1-9]\d{3}$/)
```
tvararu: This module does not have unit tests for itself but it's tested as part of the model specs; based on our conversation earlier in the tech debt discussion this feels fine as we shouldn't be adding redundant tests.

alexdesi: maybe it might be worth add couple of test fot the reg exp

alexdesi: Regexps are quite "susceptible", so at the end, I added some small tests for it,
I did it also to set up a spec, so next time will be easier to add examples on it.

PR #794, opened 2019-12-04
Ensure CUSTOM_HOSTNAME environment variable is always set.

tpinney:
### Context

At the present time the `CUSTOM_HOSTNAME` environment variable is empty if a gov.uk domain and SSL certificate does not exist.

Revised the ARM template to set `CUSTOM_HOSTNAME` to the azurewebsites.net domain where a gov.uk domain is not configured.

### Guidance to review

Probably worth someone with more knowledge of the app than me checking that everything is working as expected in the devops subscription where this new configuration is currently deployed.

https://s106d02-apply-as.azurewebsites.net

### Link to Trello card

N/A

### Env vars

### DIFF
--- a/azure/template.json
+++ b/azure/template.json
@@ -536,7 +536,7 @@
                             },
                             {
                                 "name": "CUSTOM_HOSTNAME",
-                                "value": "[parameters('customHostName')]"
+                                "value": "[if(greater(length(parameters('customHostName')), 0), parameters('customHostName'), split(parameters('authorisedHosts'), ',')[0])]"
                             },
                             {
                                 "name": "SENTRY_DSN",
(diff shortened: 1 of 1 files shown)


tpinney:
> Not sure if this helps with the original problem, which was caused by having a default. Would it be possible instead to make it mandatory to specify `CUSTOM_HOSTNAME`?

Prior to this change `CUSTOM_HOSTNAME` wasn't specified for the container instances which is why you saw the symptom you saw. I believe you changed that so that it is now specified in the template for the container instances.

There is still a problem though. If a gov.uk hostname isn't set in the pipeline variables the `CUSTOM_HOSTNAME` environment variable would exist but would be empty, which would still cause a problem in your code I believe. The pentest and devops environments being example of where we don't have a gov.uk hostname.

This change ensures that if a gov.uk hostname isn't specified it will subsititute in the azurewebsites.net domain automatically to ensure `CUSTOM_HOSTNAME` is always set.

PR #797, opened 2019-12-05
Reload application_choices when sending them to providers

duncanjbrown:
Without this change, running rake generate_test_data will fail because the application choice received by this service is `application_complete` in the database but `awaiting_references` in memory.

I've added a card to the Tech Debt backlog to investigate, because this happens a lot in our codebase.

https://trello.com/c/lQpwnQH2/372-find-out-why-we-need-to-reload-so-much-when-were-working-with-references


duncanjbrown:
Closing as @tijmenb has a branch incoming which makes some other improvements to generate_test_data and includes a fix for this

PR #800, opened 2019-12-05
Allow support users to create test applications 

tijmenb:
### Context

We need better manual testing because we're now building features that are later in the application process. To avoid us having to repetitively create applications in the UI, this adds a button to the support section that creates an application in each known state. 

- Add a button to create test applications
- Improve how we generate the test applications
- Update the support layout so it's clear what state the application is in
- Don't send emails to addresses that end in `@example.com` - this will prevent us from spamming Notify
- Similarly, prevent Slack messages from being sent when generating test applications

### Guidance to review

Test it out locally. Review per commit!

### Link to Trello card

https://trello.com/c/49yLJZEW/1353-improve-how-we-do-manual-testing

### DIFF
--- a/app/views/layouts/_support_header.html.erb
+++ b/app/views/layouts/_support_header.html.erb
@@ -5,4 +5,5 @@
   <%= nav_link 'Providers', support_interface_providers_path, active: 'providers' %>
   <%= nav_link 'Features', support_interface_feature_flags_path, active: 'feature_flags' %>
   <%= nav_link 'Performance', support_interface_performance_path, active: 'performance' %>
+  <%= nav_link 'Tasks', support_interface_tasks_path, active: 'tasks' %>
 </ul>
--- a/app/workers/generate_test_applications.rb
+++ b/app/workers/generate_test_applications.rb
@@ -1,5 +1,7 @@
 class GenerateTestApplications
-  def generate
+  include Sidekiq::Worker
+
+  def perform
     raise 'You can\'t generate test data in production' if HostingEnvironment.production?
 
     (1..11).each do |i|
--- a/app/workers/generate_test_applications.rb
+++ b/app/workers/generate_test_applications.rb
@@ -4,12 +4,20 @@ class GenerateTestApplications
   def perform
     raise 'You can\'t generate test data in production' if HostingEnvironment.production?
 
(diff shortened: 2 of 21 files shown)


app/workers/generate_test_applications.rb
```diff
@@ -4,13 +4,21 @@ class GenerateTestApplications
...
     end
   end
 
 private
 
+  def without_slack_message_sending
+    RequestStore.store[:disable_slack_messages] = true
+    yield
+    RequestStore.store[:disable_slack_messages] = false
+  end
```
tvararu: This speaks to my lack of experience with Rails and maybe threading, but I haven't seen this pattern before.

If I understand correctly, this is setting a global variable, reading it somewhere, and then setting it to something else. I'm not sure I like this approach, as it relies on side-effects, and I have a preference for functional-ish code.

In JS this would look like:

```javascript
global.disableSomething = true;
runMyCode();
global.disableSomething = false;
```

Were there alternatives considered to this approach? 🤔

app/views/layouts/_support_header.html.erb
```diff
@@ -5,4 +5,5 @@
   <%= nav_link 'Providers', support_interface_providers_path, active: 'providers' %>
   <%= nav_link 'Features', support_interface_feature_flags_path, active: 'feature_flags' %>
   <%= nav_link 'Performance', support_interface_performance_path, active: 'performance' %>
+  <%= nav_link 'Tasks', support_interface_tasks_path, active: 'tasks' %>
```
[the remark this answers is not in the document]
tijmenb: Yeah, we should move the "Vendors" thing into this page, it just has one button anyway.

app/workers/generate_test_applications.rb
tijmenb: Agreed, it's not ideal.

@duncanjbrown and I discussed a number of approaches. You could for example pass down a `slack: false` variable from the test generator into the services, or do the same with some kind of dependency injection.

Those options probably impact the readability of the code more than the sub-optimal use of globals.

Note that to avoid some of the inherent problems with globals, we've used https://github.com/steveklabnik/request_store. This means the variable won't affect other running processes.

PR #802, opened 2019-12-05
fix n+1 queries in Vendor API applications index…

aldavidson:
… and add performance tests to prove it

### Context

There is still an N+1 Query problem in the Vendor API's applications index, leading to very slow response times on Azure infrastructure.

Remove the n+1 queries by including the `application_qualifications` relation in the main query, and replacing the references to `application_qualifications.gcses` etc with array select methods on the whole application_qualifications collection. 

This tests much faster on my localhost. Using apache bench with 50 requests against localhost, I get:

Before:

Screen Shot 2019-12-05 at 17 30 26

After:
Screen Shot 2019-12-05 at 17 31 54

### Guidance to review

...needs to be properly load-tested on staging

### Link to Trello card

[API performance: GET /applications still takes 3 seconds to return](https://trello.com/c/EhNeWrgv/1363-api-performance-get-applications-still-takes-3-seconds-to-return)

### Env vars

(none)

### DIFF
--- a/app/presenters/vendor_api/single_application_presenter.rb
+++ b/app/presenters/vendor_api/single_application_presenter.rb
@@ -125,10 +125,19 @@ module VendorApi
 
     def qualifications
       {
-        gcses: application_form.application_qualifications.gcses.map do |gcse| qualification_to_hash(gcse) end,
-        degrees: application_form.application_qualifications.degrees.map do |degree| qualification_to_hash(degree) end,
-        other_qualifications: application_form.application_qualifications.other.map do |other| qualification_to_hash(other) end,
+        gcses: qualifications_of_level('gcse').map { |q| qualification_to_hash(q) },
+        degrees: qualifications_of_level('degree').map { |q| qualification_to_hash(q) },
+        other_qualifications: qualifications_of_level('other').map { |q| qualification_to_hash(q) },
       }
+    end
+
+    def qualifications_of_level(level)
+      # NOTE: we do it this way so that it uses the already-included relation
+      # rather than triggering separate queries, as it does if we use the scopes
+      # .gcses .degrees etc
+      application_form.application_qualifications.select do |q|
+        q.level == level
+      end
     end
 
     def qualification_to_hash(qualification)
--- a/spec/performance/requests/vendor_api/applications_controller_spec.rb
+++ b/spec/performance/requests/vendor_api/applications_controller_spec.rb
@@ -1,0 +1,29 @@
+require 'rails_helper'
+
+RSpec.describe VendorApi::ApplicationsController, type: :request do
+  context 'given two providers with 100 applications in various states' do
+    let(:current_provider) { create(:provider, code: 'BAT') }
+    let(:other_provider) { create(:provider, code: 'DIFFERENT') }
+    let(:api_token) { VendorApiToken.create_with_random_token!(provider: current_provider) }
+
+    before do
+      GenerateTestData.new(100, current_provider).generate
+    end
+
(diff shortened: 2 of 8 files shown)


spec/performance/requests/vendor_api/applications_controller_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'rails_helper'
+
+RSpec.describe VendorApi::ApplicationsController, type: :request do
+  context 'given two providers with 100 applications in various states' do
```
tijmenb: Do you know how long does this test takes to run? If it's too slow we might want to split it out / not run it by default.

app/presenters/vendor_api/single_application_presenter.rb
```diff
@@ -125,12 +125,21 @@ def reference_to_hash(reference)
...
-        gcses: application_form.application_qualifications.gcses.map do |gcse| qualification_to_hash(gcse) end,
-        degrees: application_form.application_qualifications.degrees.map do |degree| qualification_to_hash(degree) end,
-        other_qualifications: application_form.application_qualifications.other.map do |other| qualification_to_hash(other) end,
+        gcses: qualifications_of_level('gcse').map { |q| qualification_to_hash(q) },
+        degrees: qualifications_of_level('degree').map { |q| qualification_to_hash(q) },
+        other_qualifications: qualifications_of_level('other').map { |q| qualification_to_hash(q) },
       }
     end
 
+    def qualifications_of_level(level)
```
tijmenb: Would it make sense to replace the existing scopes in `ApplicationQualification` by this? That would mean the other uses if it in the candidate interface would be fixed as well. I can't think of a situation where it would make it slower.

spec/performance/requests/vendor_api/applications_controller_spec.rb
aldavidson: It takes about 20s to run on my machine, due to the setup - I've excluded all specs under `spec/performance` from the default rake task in the latest commit. You can still run them explicitly with `rake spec:performance` (or `rspec spec/performance`) if you want to.

app/presenters/vendor_api/single_application_presenter.rb
aldavidson: I'm not sure about that, because doing it this way transforms the AR relation into an array - so you can't return a relation to do relation chaining (e.g. `application_qualifications.gcses.where(...)` etc)

This approach makes total sense in our case - looping round a relation to render it as json - but probably wouldn't in others. If there are other parts of the app where this N+1 query issue is a problem, point me at them and i'll happily take a look and see what can be done to make the solution transferrable, but I think this stands alone for now

Rakefile
```diff
@@ -17,7 +17,7 @@ end
 begin
   require 'rspec/core/rake_task'
   RSpec::Core::RakeTask.new(:spec_with_profile) do |t|
-    t.rspec_opts = '--profile'
+    t.rspec_opts = '--profile --exclude-pattern "spec/performance/**/*.rb"'
```
tijmenb: Hmm, we're not running `spec_with_profile` anymore after https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/807, is there another place this could go for default?

app/presenters/vendor_api/single_application_presenter.rb
tijmenb: Cool!

Rakefile
aldavidson: I've updated this after merging in your PR #807

PR #806, opened 2019-12-06
Add :accepted_ts_and_cs_at to Candidate

mnacos:
### Context

For the sign-in with unknown email flow, we are planning to create candidates in the database, send them emails redirecting them to the sign-up form as before and make sure nobody can log in unless they have accepted T&Cs.

This is an initial commit with just the db migration, as requested by Theo.

This PR adds the field as a ```datetime```, so that we are able to ask people to accept new T&Cs if they are revised in the future using date comparisons.

### Guidance to review

Should be straightforward, you either agree with the approach or you don't!

### Link to Trello card

[633 - Add :accepted_ts_and_cs_at to Candidate and change sign_in_without_account_email flow](https://trello.com/c/mK2xd0fi)

### Env vars

No new env vars


mnacos:
Closing for now. May re-open when we do this.

PR #807, opened 2019-12-06
Experiment with using GitHub Actions for CI

tijmenb:
### Context

Our CI builds on Azure currently take around ~11 minutes to complete (and that excludes time queueing):

Screenshot 2019-12-06 at 09 05 49

This implements a [GitHub workflow](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/configuring-a-workflow) to run CI.

It runs in about ~5 minutes. This is because:

1. We don't build the full production containers, but run the tests and linters on GitHubs VMs. This saves a bunch of time setting up the infrastructure.
2. Bundler and Yarn dependencies are cached between runs, so `bundle install` is blazing fast most of the time.
3. GitHub allows multiple jobs to run in parallel. By splitting up linting, integration tests, unit tests and brakeman, we can speed up the overall process. This has the added benefit of quicker feedback while CI runs, and more complete feedback once everything has run: we curently bail out early for linting failures, so you wouldn't know the tests wouldn't pass.

#### Future plans

For now, these tests run in addition to the tests in Azure. If we find that this CI process works well, we could consider a new development workfow. We would rely on GitHub to perform the CI for our Pull Requests - and allow PRs to be merged once they pass the GitHub tests. This would give us a fast feedback cycle. 

Once the code is merged into master, the Azure pipeline would still do it's thing - this would give us the full confidence that the code would work on production. The trade-off we make is that there will be instances where GitHub passes, but Azure pipelines fail, and we have to repair master. This could be a trade-off worth making. 

#### Guidance to review

Per commit - there's some tidying up of the test tasks in this.

### Link to Trello card

https://trello.com/c/OEjmIn3C/327-the-tests-are-becoming-slow-high

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,205 @@
+name: 'Apply CI'
+on:
+  push:
+    branches-ignore:
+      - 'master'
+
+jobs:
+  unit_tests:
+    name: 'Unit tests'
+    runs-on: ubuntu-latest
+
+    services:
+      postgres:
+        image: 'postgres:11.5'
+        ports:
+          - '5432:5432'
+        options: '--health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5'
+      redis:
+        image: redis
+        ports:
+          - '6379:6379'
+        options: '--entrypoint redis-server'
+
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: Set up gem cache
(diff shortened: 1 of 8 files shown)


benilovj:
The arguments in the PR description feel ADR-worthy...

tijmenb:
@benilovj yes, good idea. Once we make a decision whether this makes sense I'll write it up as an ADR1

PR #810, opened 2019-12-06
Add support for error 422 - Unprocessable Entity

alexdesi:
### Context
This PR provides an error in case of unprocessable entity 422

Just standard Rails way using `ErrorsController`

Please read my comment on trello for more: https://trello.com/c/GknT2gs5/596-422s-show-up-as-a-blank-page

### Trello
https://trello.com/c/GknT2gs5/596-422s-show-up-as-a-blank-page


app/views/errors/unprocessable_entity.html.erb
```diff
@@ -0,0 +1,11 @@
+<%= content_for :title, t('page_titles.unprocessable_entity') %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl"><%= t('page_titles.unprocessable_entity') %></h1>
+    <p class="govuk-body">Try again later</p>
+    <p class="govuk-body">
+      If you have provided correct information and still continue to see this error contact the Becoming a Teacher team: <%= bat_contact_mail_to %>.
```
tvararu:
```suggestion
      If you continue to see this error contact the Becoming a Teacher team: <%= bat_contact_mail_to %>.
```

422s should not normally occur to the end user; if they see it they should immediately contact us I think. (This is the content from the 500 page)

config/locales/en.yml
```diff
@@ -17,6 +17,7 @@ en:
     check_your_email: Check your email
     not_found: Page not found
     internal_server_error: Sorry, there is a problem with the service
+    unprocessable_entity: Sorry, an invalid request was made
```
tvararu:
```suggestion
    unprocessable_entity: Sorry, there is a problem with the service
```

Users shouldn't need to know about requests; from their perspective there's a problem.

tvararu:
```suggestion
    unprocessable_entity: Sorry, there is a problem with the service
```

Users shouldn't need to know about requests; from their perspective there's a problem.

PR #814, opened 2019-12-06
Fix Page not found when "Another course"

alexdesi:
### Context
This PR fixes the `Page not found` error when selecting a second-course choice and choosing `Another course`

The error is raised because 

`@pick_course.valid?` was checked in case of course code equal to `"other"`

My quick fix proposal is just inverting the order of the conditionals.

## NOTE
This bug suggests that the controller action `pick_course` is quite brittle and a refactor is much recommended.

If you have any comments on how to refactor this let me know, I am happy to review this code on Monday. 

### Trello

https://trello.com/c/KbNZQgry/629-redirects-to-page-not-found-when-selecting-a-second-course-choice-and-choosing-another-course

### DIFF
--- a/app/controllers/candidate_interface/course_choices_controller.rb
+++ b/app/controllers/candidate_interface/course_choices_controller.rb
@@ -61,10 +61,10 @@ module CandidateInterface
         application_form: current_application,
       )
 
-      if !@pick_course.valid?
-        render :options_for_course
-      elsif @pick_course.other?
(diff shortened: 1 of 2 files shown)


alexdesi:
Hey @tvararu , since the bug involves the controller, I had to change the UI spec (`candidate_selecting_a_course_spec.rb `) in order to test  it properly.

PR #815, opened 2019-12-06
Add audit trail attribution for support users

stevehook:
### Context

We have switched the Support Interface from basic auth over to DfE Sign-in. This means that we know who the 'current user' is in the Support Interface so we can properly attribute changes made there now.

- [x] Link audit trail records to instances of `SupportUser` for all changes made via the Support Interface.
- [x] Label such audit trail records with the email of the support user on the application history page.

image

### Guidance to review

- Are the tests adequate?
- Do the labels in application history make sense?

### Link to Trello card

[1347 - Add attribution to support users for auditing](https://trello.com/c/Fbwpa49B/1347-add-attribution-to-support-users-for-auditing)

### Env vars

- [x] No new env vars

### DIFF
--- a/app/components/support_interface/audit_trail_item_component.rb
+++ b/app/components/support_interface/audit_trail_item_component.rb
@@ -22,7 +22,7 @@ module SupportInterface
       elsif audit.user_type == 'VendorApiUser'
         "#{audit.user.email_address} (Vendor API)"
       elsif audit.user_type == 'SupportUser'
-        "#{audit.user.email_address} (Support User)"
+        "#{audit.user.email_address} (Support user)"
       elsif audit.username.present?
         audit.username
       else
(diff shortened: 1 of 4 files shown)


app/components/support_interface/audit_trail_item_component.rb
```diff
@@ -21,6 +21,8 @@ def audit_entry_user_label
         "#{audit.user.email_address} (Candidate)"
       elsif audit.user_type == 'VendorApiUser'
         "#{audit.user.email_address} (Vendor API)"
+      elsif audit.user_type == 'SupportUser'
+        "#{audit.user.email_address} (Support User)"
```
fofr:
```suggestion
        "#{audit.user.email_address} (Support)"
```

PR #818, opened 2019-12-09
Small tweaks to application support page

fofr:
- Give references and course choices a heading
- Don't show the reject by default row when it would be empty
- Tweak button styles of feature flags page


tijmenb:
Merging because Paul's out today.

PR #823, opened 2019-12-09
Allow vendor API to amend a provider decision

mnacos:
### Context

The application already allows providers to decide on application_choices, offering or rejecting them, and candidates can accept or decline them. Application choices can also be rejected by default, after a set amount of time, if the provider fails to respond (RBD). Finally, provider offers can be declined by default (DBD) if the candidate fails to respond within a time limit.

We'll also need to allow providers to change their mind and

- amend an existing offer (e.g. by changing the conditions) [UI already implemented]
- reject an existing offer [not currently supported in the data model]
- change a rejection to an offer [not currently supported in the data model]

image

Considering all these are decision amendments, it is possible the 'decline by default' (DBD) date has already been set for one or more offers by the time of the amendment, so a decision was needed around what happens to the DBD date. After some initial discussion, we have decided to always extend the DBD date for all offers, based on the most recent event [to confirm: *rejections can also push back the DBD*].

State transitions have been added to allow the following:

image

Tests have been added to confirm:
- [x] provider can change an offer to a rejection
- [x] provider can change a rejection into an offer
- [x] an amended decision resets the 'declined by default' (DBD) date
- [x] provider cannot change an offer after candidate has accepted it
- [x] provider cannot reject an application after candidate has accepted it
- [x] provider cannot change an offer after candidate has declined it

### Guidance to review

Please review:

- spec/requests/vendor_api/post_make_an_offer_spec.rb [changes to api expectations]
- spec/requests/vendor_api/post_reject_application_spec.rb [changes to api expectations]
- spec/services/set_decline_by_default_spec.rb [DBD updates]

### Link to Trello card

[1341 - Allow providers to change their decision via API](https://trello.com/c/yTYvdXBM)

### Env vars

No new env vars have been added.


spec/requests/vendor_api/post_make_an_offer_spec.rb
```diff
@@ -69,6 +71,44 @@
...
+      )
+
+      request_body = { "data": { "conditions": ['DBS Check'] } }
+      post_api_request "/api/v1/applications/#{application_choice.id}/offer", params: request_body
+
+      expect(parsed_response).to be_valid_against_openapi_schema('SingleApplicationResponse')
+      expect(parsed_response['data']['attributes']['status']).to eq('offer')
+    end
+
+    it 'cannot update an offer once it has been accepted by the candidate' do
```
tijmenb: I wonder if we need to test the invalid state transitions here, perhaps they are already tested elsewhere in the system? We should be able to rely on the state machine to avoid duplicated tests.

app/views/api_docs/reference/reference.html.erb
```diff
@@ -29,8 +29,7 @@
 <% @document.paths.each do |path_name, path| %>
   <% %w[get put post delete patch].map do |http_verb| %>
     <% operation = path.node_data[http_verb]
-    next unless operation && !operation.kind_of?(Openapi3Parser::Node::Placeholder)
```
tijmenb: All of these should be solved by https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/827.

spec/requests/vendor_api/post_make_an_offer_spec.rb
mnacos: Althought they share the same model, candidate_interface, provider_interface and vendor_api are different interfaces/products. These tests are there not to test if the transitions are allowed, but to test these actions can be completed through the API.

tijmenb: My point is about these 4:

- `cannot update an offer once it has been accepted by the candidate`
- `cannot update an offer once it has been declined by the candidate`
- `cannot reject application if already rejected`
- `cannot reject an application that a candidate has already accepted`

We already have a test that checks the behaviour if the user attempts an invalid state change:

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/377f338bb81fdc71cf23cc72f3625965f9f70233/spec/requests/vendor_api/post_make_an_offer_spec.rb#L93-L103

It seems like these 4 tests duplicate that? We don't need specific tests for each invalid transition, since that would mean we would have to add a dozen more tests and it'll just duplicate the state machine.

mnacos: Ok

PR #825, opened 2019-12-09
UI to add support users to the database

duncanjbrown:
### Context

To support #770, create a UI to add support users to the db

- we should not merge #770 until this is merged and we have created at least one support user, as without a user in the database it will be impossible to log into support once basic auth is turned off

CRUD for `SupportUsers` in the support interface.

- We don't support deletion as we haven't discussed how this will work
- at the moment all support users will be able to create other users, which is a bit 😬, but we can revisit

### Link to Trello card

https://trello.com/c/Yk89F38A/1371-ui-to-add-support-users-to-the-database

out

### DIFF
--- a/app/views/layouts/_support_header.html.erb
+++ b/app/views/layouts/_support_header.html.erb
@@ -6,4 +6,5 @@
   <%= nav_link 'Features', support_interface_feature_flags_path, active: 'feature_flags' %>
   <%= nav_link 'Performance', support_interface_performance_path, active: 'performance' %>
   <%= nav_link 'Tasks', support_interface_tasks_path, active: 'tasks' %>
+  <%= nav_link 'Support users', support_interface_support_users_path, active: 'users' %>
 </ul>
--- a/app/views/support_interface/support_users/new.html.erb
+++ b/app/views/support_interface/support_users/new.html.erb
@@ -1,0 +1,14 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with model: @support_user, url: support_interface_support_users_path, builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f| %>
+      <%= f.govuk_error_summary %>
+
+      <h1 class='govuk-heading-xl'>Add support user</h1>
+
+      <%= f.govuk_text_field :email_address, label: { text: 'Email address' } %>
+      <%= f.govuk_text_field :dfe_sign_in_uid, label: { text: 'DfE Sign-in UID' } %>
+
+      <%= f.govuk_submit 'Add support user' %>
+    <% end %>
+  </div>
+</div>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -115,6 +115,12 @@ en:
             relationship:
               blank: Tell us about the referee’s relationship to you and how long they’ve known you
               too_many_words: Relationship to referee must be %{count} words or fewer
+        support_user:
+          attributes:
+            email_address:
+              blank: Email address can’t be blank
+            dfe_sign_in_uid:
+              blank: DfE Sign-in UID can’t be blank
   errors:
     messages:
       too_many_words: Must be %{count} words or fewer
--- a/spec/system/support_interface/managing_support_users_spec.rb
+++ b/spec/system/support_interface/managing_support_users_spec.rb
@@ -1,0 +1,33 @@
+require 'rails_helper'
+
+RSpec.feature 'Managing support users' do
+  scenario 'creating a new support user' do
+    given_i_am_a_support_user
+    and_a_support_user_exists_in_the_database
+
+    when_i_visit_the_support_console
+    and_i_click_the_manange_support_users_link
+
+    then_i_should_see_the_list_of_support_users
+  end
+
+  def given_i_am_a_support_user
+    page.driver.browser.authorize('test', 'test')
+  end
+
+  def and_a_support_user_exists_in_the_database
+    create(:support_user, email_address: 'person@education.gov.uk')
+  end
+
+  def when_i_visit_the_support_console
+    visit '/support'
+  end
+
+  def and_i_click_the_manange_support_users_link
+    click_link 'Support users'
+  end
+
+  def then_i_should_see_the_list_of_support_users
+    expect(page).to have_content('Support users')
+  end
+end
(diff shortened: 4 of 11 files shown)


tvararu:
> we should not merge #770 until this is merged and we have created at least one support user, as without a user in the database it will be impossible to log into support once basic auth is turned off

This sounds annoying when we create a new environment - is it worth exploring if we can make the current basic auth a progressive fallback or explicitly togglable via feature flag? My thinking is:

- When the db is first created, use basic auth to access support console
- Create your support users
- Pull the plug 🔌 by toggling on the feature flag: disables the basic auth mechanism, DSI only moving forward

duncanjbrown:
> > we should not merge #770 until this is merged and we have created at least one support user, as without a user in the database it will be impossible to log into support once basic auth is turned off
> 
> This sounds annoying when we create a new environment - is it worth exploring if we can make the current basic auth a progressive fallback or explicitly togglable via feature flag? My thinking is:
> 
> * When the db is first created, use basic auth to access support console
> * Create your support users
> * Pull the plug 🔌 by toggling on the feature flag: disables the basic auth mechanism, DSI only moving forward

Good point. One alternative would be to connect directly to the db and create a user, but that's not terribly slick. Whatever we do would need to go on top of #770 and #815 — shall we revisit once those are merged?

tvararu:
Discussed offline: because leaving in basic auth is icky, and we shouldn't have a need to create new environments too often, we think it's best if we go with the DB insert method. We should document this.

spec/system/support_interface/managing_support_users_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    then_i_should_see_the_list_of_support_users
+    and_i_should_see_the_user_i_created
+  end
+
+  def given_i_am_a_support_user
+    page.driver.browser.authorize('test', 'test')
+  end
+
+  def and_a_support_user_exists_in_the_database
+    create(:support_user, email_address: 'person@education.gov.uk')
```
tijmenb: Nooooo
```suggestion
    create(:support_user, email_address: 'person@example.com')
```

spec/system/support_interface/managing_support_users_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def when_i_visit_the_support_console
+    visit '/support'
```
tijmenb:
```suggestion
    visit support_interface_path
```

(iirc)

spec/system/support_interface/managing_support_users_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def and_i_click_the_manange_support_users_link
+    click_link 'Support users'
+  end
+
+  def and_i_enter_the_users_email_and_dsi_uid
+    fill_in 'support_user[email_address]', with: 'harrison@bergeron.com'
```
tijmenb:
```suggestion
    fill_in 'support_user[email_address]', with: 'harrison@example.com'
```

app/views/support_interface/support_users/new.html.erb
```diff
@@ -0,0 +1,14 @@
+<% content_for :title, 'Add support user' %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with model: @support_user, url: support_interface_support_users_path, builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f| %>
+      <%= f.govuk_error_summary %>
```
fofr: Error summaries should always be above the `h1`.

spec/system/support_interface/managing_support_users_spec.rb
duncanjbrown: Ah — am I breaking with convention?

app/views/layouts/_support_header.html.erb
```diff
@@ -6,4 +6,5 @@
   <%= nav_link 'Features', support_interface_feature_flags_path, active: 'feature_flags' %>
   <%= nav_link 'Performance', support_interface_performance_path, active: 'performance' %>
   <%= nav_link 'Tasks', support_interface_tasks_path, active: 'tasks' %>
+  <%= nav_link 'Support users', support_interface_support_users_path, active: 'users' %>
```
fofr: This is wrapping now.

I wonder if we should move some of these behind an "Admin" link?

Admin page would link to:
- Features
- Tasks
- Performance
- Support users

(probably not in this PR)

duncanjbrown: I discussed with @tvararu — we'd like to revisit this. Will loop you in @fofr when that conversation happens

tvararu: Actually Paul's suggestion sounds good, it was what I was thinking about too 🤔 if it's not too much, we can do it as part of this?

config/locales/en.yml
```diff
@@ -115,6 +115,12 @@ en:
             relationship:
               blank: Tell us about the referee’s relationship to you and how long they’ve known you
               too_many_words: Relationship to referee must be %{count} words or fewer
+        support_user:
+          attributes:
+            email_address:
+              blank: Email address can’t be blank
+            dfe_sign_in_uid:
+              blank: DFE Sign-in UID can’t be blank
```
stevehook: 'DfE' for consistency?

app/views/support_interface/support_users/new.html.erb
```diff
@@ -0,0 +1,14 @@
...
+
+      <%= f.govuk_text_field :email_address, label: { text: 'Email address' } %>
+      <%= f.govuk_text_field :dfe_sign_in_uid, label: { text: 'DFE Sign-in UID' } %>
```
stevehook: 'DfE'?

config/locales/en.yml
duncanjbrown: 👌

app/views/layouts/_support_header.html.erb
duncanjbrown: I'll do another PR after this

duncanjbrown:
Thanks everyone!

- made all the emails in specs `@example.com` (🙏  @tijmenb)
- included instructions for creating a support user in prod, and while I'm at it add some generic instructions for connecting to live databases (🙏  @tvararu and @tijmenb)
- pushed the error summary above the `h1` (🙏 @fofr)
- correctly capitalised DfE Sign-in (🙏 @stevehook)

If I may, I'll punt the menu reorg (@fofr and @tvararu) until the next PR as this is a useful change in itself and no members of the public will see the overflowing menu

tijmenb:
Merging because I need this UI for a follow up PR!

PR #826, opened 2019-12-09
Minor API fixes

stevehook:
### Context

A number of issues were raised after manual testing by @fofr . This PR addresses some of the bug fixes (doc fixed to follow)

- [X] Course codes should be 4 chars, not 3 - `maxLength` added to OpenAPI spec and values adjusted in test data generator
- [X] Site code should be single char, not 3 - `maxLength` added to OpenAPI spec and values adjusted in test data generator. Requires a bit more care in how we generate 'random' codes as they need to be unique.
- [X] Address line 1 not a good example (it’s “45”) - Moved the street name into line 1 and added a county for line 3 (though not sure Stockport is technically in a county)
- [X] Catch the error when an invalid date is provided for ?since= (we 500 rn) - added validation to ensure this error is caught and causes status 422
- [X] Provide a sensible default value for test-data/regenerate (0 is current and it wipes everything, docs say 100 by default) - happens because `nil.to_i -> 0` so have made a small change to ensure default is 100.
- [X] API: Formatting on state errors is bad (mispunctuated, reveals implementation details) - Trapping the invalid transition exception and returning a more bland error message `The application is not ready for that action` - please shout if you have better wording

### Guidance to review

- Commit by commit
- Is there a better way to specify fixed length attributes? (OpenAPI only seems to let you specify a `maxLength`)
- Is there a better wording for the error message above?

### Link to Trello card

[1176 - Minor API fixes following user testing with Paul H](https://trello.com/c/MVlw347D/1176-minor-api-fixes-following-user-testing-with-paul-h)

### Env vars

- [x] No env vars

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -122,6 +122,10 @@ en:
               blank: Email address can’t be blank
             dfe_sign_in_uid:
               blank: DfE Sign-in UID can’t be blank
+        application_choice:
+          attributes:
+            status:
+              invalid_transition: The application is not ready for that action
   errors:
     messages:
       too_many_words: Must be %{count} words or fewer
--- a/config/vendor-api-0.8.0.yml
+++ b/config/vendor-api-0.8.0.yml
@@ -451,7 +451,7 @@ components:
           type: string
           description: The candidate’s address line 3
           maxLength: 50
-          example: Cheshire
+          example: Greater Manchester
         address_line4:
           type: string
           description: The candidate’s address line 4
--- a/config/vendor-api-0.8.0.yml
+++ b/config/vendor-api-0.8.0.yml
@@ -440,17 +440,17 @@ components:
           type: string
           description: The candidate’s address line 1
           maxLength: 50
-          example: "45"
+          example: "45 Dialstone Lane"
         address_line2:
           type: string
           description: The candidate’s address line 2
           maxLength: 50
-          example: Dialstone Lane
+          example: Stockport
         address_line3:
           type: string
           description: The candidate’s address line 3
           maxLength: 50
-          example: Stockport
+          example: Cheshire
         address_line4:
           type: string
           description: The candidate’s address line 4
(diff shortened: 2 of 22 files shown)


fofr:
> Site code should be single char, not 3 - maxLength added to OpenAPI spec and values adjusted in test data generator. Requires a bit more care in how we generate 'random' codes as they need to be unique.
I think we should discuss what Site code is used for and if it's needed at all. The single char comes from UCAS, who limit a provider to 36 locations. We know providers sometimes need more than that when setting up their locations.

duncanjbrown:
> > Site code should be single char, not 3 - maxLength added to OpenAPI spec and values adjusted in test data generator. Requires a bit more care in how we generate 'random' codes as they need to be unique.
> 
> I think we should discuss what Site code is used for and if it's needed at all. The single char comes from UCAS, who limit a provider to 36 locations. We know providers sometimes need more than that when setting up their locations.

AFAIK we need site code as it's part of the identify of a course option, which is the thing that is applied to. Until we have an id that encompasses that intersection of course, site, date and provider I don't think we can dispense with it.

I may be missing something.

config/locales/en.yml
```diff
@@ -115,6 +115,10 @@ en:
             relationship:
               blank: Tell us about the referee’s relationship to you and how long they’ve known you
               too_many_words: Relationship to referee must be %{count} words or fewer
+        application_choice:
+          attributes:
+            status:
+              invalid_transition: The application is not ready for that action
```
duncanjbrown: I think this is fine. We should include the flowchart (or the relevant bit) in the docs, maybe?

config/vendor-api-0.8.0.yml
```diff
@@ -475,17 +475,17 @@ components:
...
           description: The candidate’s address line 1
           maxLength: 50
-          example: "45"
+          example: "45 Dialstone Lane"
         address_line2:
           type: string
           description: The candidate’s address line 2
           maxLength: 50
-          example: Dialstone Lane
+          example: Stockport
```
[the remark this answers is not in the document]
stevehook: 👍 Perhaps I'll shorten it to `Greater Manchester`.

config/locales/en.yml
stevehook: I'll add an item to the docs checklist for adding the flowchart and look at that in the follow up PR

PR #827, opened 2019-12-09
Redesign the API reference

tijmenb:
### Context

This is a bit of a redesign of the API reference.

The page now has a table of contents:

image

This uses the [contents list component](https://components.publishing.service.gov.uk/component-guide/contents_list) from GOV.UK.

image

An endpoint has a title, description, and possible reponses hidden in a details element.

The details will link to the reference for the object being returned.

image

Objects now show attributes in a definition list, and have simpler layout. They correctly link through to other data types. 

### Guidance to review

Spin it up locally and see if it makes sense.

### Link to Trello card

https://trello.com/c/KENwAoIR/1357-merge-tech-docs-into-main-application


duncanjbrown:
💯 This is so much better! I find folded responses much more skimmable and the syntax highlighting makes them a lot easier to read.

A few notes:

- The lack of wrapping means the response exceeds the background, but not harmfully:

Screenshot 2019-12-09 at 17 05 13

- Backticks are broken

Screenshot 2019-12-09 at 17 02 56

- We have some stray %-encoding under `POST /applications/{application_id}/reject` but not elsewhere, interestingly. I'm not sure this was introduced here, but here's a note of it anyway:

Screenshot 2019-12-09 at 17 09 22

- I find monospaced text on the left hand side of attribute tables more readable when normal instead of bold (taking the 700 weight off `govuk-summary-list__key`), but defer to designers

- We need to make another pass to fix all the links
  - on the front page, all the `What this API is for` links and the `release notes` link
  - all the links in the usage scenarios
  - I propose removing the release notes for now as they're a source of broken-link-related confusion?

PR #828, opened 2019-12-09
Fix master linting

mnacos:
Fixes linting errors in master.


mnacos:
Not sure why these were not picked up during the PR merge?

app/views/api_docs/reference/reference.html.erb
```diff
@@ -1,4 +1,4 @@
-<h1 class="govuk-heading-xl"><%= t("page_titles.api_docs.reference") %></h1>
+<h1 class='govuk-heading-xl'><%= t('page_titles.api_docs.reference') %></h1>
```
stevehook: Shouldn't it be double quotes for HTML attributes?

fofr: Aye, always double quotes for HTML attributes.

PR #835, opened 2019-12-10
Remove "ucas" from thing-codes and allow more sites to exist

tijmenb:
### Context

We call provider/course/site codes "ucas_provider_code", "ucas_site_code", etc - this is incorrect because they are now minted and managed by the DfE Find service, and not by UCAS anymore. 

This also means that we don't have to enforce the 1-character code for the site, which limits the number of sites per provider to 36. This limit has been hit already by some providers.

Discussed and 🔨 decided with @fofr, @misaka and @benilovj.

- Remove `ucas_` from attribute names in the API
- Up the limit for sites to 5. This is so that we easily can differentiate them from other codes (1 char for an old-style site code, 3 for a provider, 4 for course, 5 for new-style sites).

### Guidance to review

Nothing in particular.

### Link to Trello card

https://trello.com/c/yKbdofEs/1376-remove-ucas-from-thing-codes-and-allow-more-sites-to-exist

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,8 +218,8 @@ by invoking `bin/webpack`. If all is well, there is a chance that
 ## <a name="documentation"></a>Documentation
 
 ### <a name="documentation-nomenclature"></a>Nomenclature
-- **Course** consists of a UCAS provider code and a UCAS course code. In our system, this is represented by the `ucas_provider_code` and `ucas_course_code` on the `ApplicationChoice` model
-- **Course Choice** is the course plus a training location code, in our system represented by `ucas_provider_code`, `ucas_course_code`, `ucas_location_code`
+- **Course** consists of a UCAS provider code and a UCAS course code. In our system, this is represented by the `provider_code` and `course_code` on the `ApplicationChoice` model
+- **Course Choice** is the course plus a training location code, in our system represented by `provider_code`, `course_code`, `location_code`
 
 ### <a name="documentation-domain-model"></a>Domain Model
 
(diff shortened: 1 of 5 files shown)


config/vendor-api-0.8.0.yml
```diff
@@ -482,30 +482,30 @@ components:
...
-          description: The course’s UCAS code
+          description: The course’s code
           example: 3CVK
           maxLength: 4
-        site_ucas_code:
+        site_code:
           type: string
-          description: The site’s UCAS code
+          description: The site’s code
           example: K
```
stevehook: I guess it makes sense to change the example too

tijmenb:
@stevehook I think that's alright actually! The existing codes won't change for now, so the examples are okay. I like the limiting of sites in the test generation, less than 10 is probably more realistic.

README.md
```diff
@@ -218,8 +218,8 @@ by invoking `bin/webpack`. If all is well, there is a chance that
 ## <a name="documentation"></a>Documentation
 
 ### <a name="documentation-nomenclature"></a>Nomenclature
-- **Course** consists of a UCAS provider code and a UCAS course code. In our system, this is represented by the `ucas_provider_code` and `ucas_course_code` on the `ApplicationChoice` model
-- **Course Choice** is the course plus a training location code, in our system represented by `ucas_provider_code`, `ucas_course_code`, `ucas_location_code`
+- **Course** consists of a UCAS provider code and a UCAS course code. In our system, this is represented by the `provider_code` and `course_code` on the `ApplicationChoice` model
```
fofr: > UCAS provider code and a UCAS course code
Probably should change too?

PR #837, opened 2019-12-10
Add another DSI user to test with

tvararu:
We need a user for DAC and on the pentest environment.


tvararu:
Tested this locally, it works in the `Test` DfE Signin environment.

PR #842, opened 2019-12-10
Remove Back link on Ucas page, and update wording.

alexdesi:
### Context

The Ucas page content is not suitable for users that want to apply to a non-opted in course ("other course").

- Update the wording.

- Since the "back" link should be dynamic because that page is accessed from different paths, remove the link for now.

### Link to Trello card

https://trello.com/c/GkvNXpDd/646-change-content-for-page-users-are-sent-to-when-they-want-to-apply-to-a-non-opted-in-course

### DIFF
--- a/app/views/candidate_interface/course_choices/ucas.html.erb
+++ b/app/views/candidate_interface/course_choices/ucas.html.erb
@@ -7,7 +7,7 @@
       <%= t('page_titles.not_eligible_yet') %>
     </h1>
 
-    <p class="govuk-body"><%= service_name %> is still in development, so for now, we have to limit candidate numbers.</p>
+    <p class="govuk-body"><%= service_name %> is still in development, so for now, we have to limit candidate numbers and the range of courses on offer.</p>
     <p class="govuk-body govuk-!-font-weight-bold">You can apply for all teacher training courses and providers using UCAS.</p>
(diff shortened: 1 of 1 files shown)


fofr:
We need a contextual back link.

fofr:
Please make sure the work to put in a working back link is captured in Trello.

app/views/candidate_interface/course_choices/ucas.html.erb
```diff
@@ -1,13 +1,12 @@
 <% content_for :title, t('page_titles.not_eligible_yet') %>
-<% content_for :before_content, govuk_back_link_to(candidate_interface_course_choices_provider_path) %>
```
fofr: ```javascript:history.back()```
Is this better than nothing? 🤔

tvararu: 👍

alexdesi: Is not supposed to work without JS too?

alexdesi: I am going to revert back the code, the main point of this was to update the wording.

alexdesi:
@fofr I created a new trello for it: https://trello.com/c/EnwFLp1M/647-back-link-on-apply-on-ucas-page

PR #845, opened 2019-12-10
Allow vendors to provide a course option as part of an /offer

tvararu:
### Context

To implement the `Provider_Change_CourseDetails` business rule, we need to allow vendors to pass in an alternate course option when they make an offer to a candidate's application.

- [x] Add offer_course_option_id to ApplicationChoice via migration
- [x] Allow vendors to pass course to /offer
- [x] Write a test for passing in `course` to ensure it updates the model in the right way
- [x] Write a test to check that the `offer_course_option` must be part of the same provider as the `course_option`; implement the validation to make this test pass
- [x] Update vendor API documentation

### Guidance to review

Review commit by commit.

### After (API)

Endpoint:

Screenshot 2019-12-12 at 17 57 35

Course schema:

Screenshot 2019-12-12 at 17 57 46

### Link to Trello card

[1342 - API/UI: Allow providers to offer a different course (Provider_Change_CourseDetails) 🏈](https://trello.com/c/4sPHd0Ci)


duncanjbrown:
Looks good so far. Maybe they don’t need a `course_option_id` — could they send an ordinary Course object? (https://www.apply-for-teacher-training.education.gov.uk/api-docs/reference/#course-object)

app/controllers/vendor_api/decisions_controller.rb
```diff
@@ -6,6 +6,7 @@ def make_offer
       decision = MakeAnOffer.new(
         application_choice: application_choice,
         offer_conditions: params.dig(:data, :conditions),
+        offer_course_option_id: params.dig(:data, :offer_course_option_id),
```
tijmenb: What would you think of making the API accept a Course object instead of the `course_option_id`?

https://qa.apply-for-teacher-training.education.gov.uk/api-docs/reference#course-object

I imagine the API looking something like this:

```rb
describe 'making an offer for another course' do
  it 'returns the updated application' do
    application_choice = create_application_choice_for_currently_authenticated_provider(
      status: 'awaiting_provider_decision',
    )

    other_course_option = course_option_for_provider(provider: currently_authenticated_provider)

    post_api_request '/api/v1/applications/#{application_choice.id}/offer', params: {
      'data' => {
        'conditions' => [],
        'course' => {
          'provider_code' => other_course_option.course.provider.code,
          'course_code' => other_course_option.course.code,
          'site_code' => other_course_option.site.code,
        }
      },
    }

    expect(parsed_response['data']['attributes']['offer']).to eq(
      'conditions' => [],
      'course' => {
        'provider_code' => other_course_option.course.provider.code,
        'course_code' => other_course_option.course.code,
        'site_code' => other_course_option.site.code,
      },
    )
  end
end
```

db/migrate/20191210171030_add_offer_course_option_id_to_application_choice.rb
```diff
@@ -0,0 +1,5 @@
+class AddOfferCourseOptionIdToApplicationChoice < ActiveRecord::Migration[6.0]
+  def change
+    add_column :application_choices, :offer_course_option_id, :integer
```
tijmenb: Perhaps `offered_course_option_id`? Perhaps this can use `add_reference` to generate foreign key constraints.

app/services/make_an_offer.rb
```diff
@@ -8,14 +8,16 @@ class MakeAnOffer
...
+  def initialize(application_choice:, offer_conditions: nil, offer_course_option_id: application_choice.course_option.id)
     @application_choice = application_choice
     @offer_conditions = offer_conditions
+    @offer_course_option_id = offer_course_option_id
   end
 
   def save
     return unless valid?
 
+    application_choice.update!(offer_course_option_id: @offer_course_option_id)
```
tijmenb: Could move down to `application_choice.offer` so that it saves in the same query?

app/controllers/vendor_api/decisions_controller.rb
tijmenb: (What Duncan said)

app/services/make_an_offer.rb
tvararu: I thought about doing this; I'll go for it then.

When we were kicking this off, we mentioned querying to measure which choices have had an alternate offer come in from a provider. We can store that information inside of the JSON blob, it will just make it slightly trickier to retrieve. That being said, it's not impossible, so let's try it.

db/migrate/20191210171030_add_offer_course_option_id_to_application_choice.rb
tvararu: Migration is probably going away if we store it in JSON 👍

+1 for the FK suggestion if not.

app/services/make_an_offer.rb
tijmenb: Sorry I meant something different! Having it as a FK like here is definitely the way to go.

I just meant:

```diff
- application_choice.update!(offer_course_option_id: @offer_course_option_id)
ApplicationStateChange.new(application_choice).make_offer!
+ application_choice.offer_course_option_id = @offer_course_option_id
application_choice.offer = { 'conditions' => (@offer_conditions || []) }

application_choice.offered_at = Time.now
application_choice.save!
```

tvararu: Ah yes, not using `update!` because we `save!` later anyway, doh! 👍

tvararu:
Pushed some more commits, I think in principle the implementation is covered now. What's left is validation and updating the docs / making existing tests pass.

app/presenters/vendor_api/single_application_presenter.rb
```diff
@@ -163,5 +163,24 @@ def institution_details(qualification)
...
+                              provider_code: offered_option.course.provider.code,
+                              course_code: offered_option.course.code,
+                              site_code: offered_option.site.code,
+                           },
+                         }
+                       end
+
+      {
+        conditions: application_choice.offer && application_choice.offer['conditions'],
+      }.merge(offered_course)
```
duncanjbrown: - Maybe the `#course` method could be refactored to take a `course` param and could be reused here? (Hopefully this will keep us up to date with incoming changes to full/part time and cycle year too)
- given we're going on the basis that `offered_option` will always be populated if the offer is present (which seems right), when `offered_option` is `nil` here (ie no offer has been received) then I think it doesn't make sense to include a hash with only an empty `conditions` key because that implies an unconditional offer

app/controllers/vendor_api/decisions_controller.rb
```diff
@@ -51,6 +52,17 @@ def respond_to_decision(decision)
       end
     end
 
+    def offered_course_option
+      course_data = params.dig(:data, :course)
+      return nil if course_data.nil?
+
+      provider = Provider.find_by(code: course_data[:provider_code])
+      course = provider.courses.find_by(code: course_data[:course_code])
+      site = provider.sites.find_by(code: course_data[:site_code])
```
tvararu: This should potentially search by `study_mode` as well, depending on if https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/851 is merged.

app/presenters/vendor_api/single_application_presenter.rb
tvararu: Great suggestion 👍

tvararu: Committed, tests updated!

mnacos:
Looks good. My only concern is where would vendor software find the ```course_option_id```, since they are not exposed in the API. Should we add an endpoint or add a ticket to create an endpoint down the line?

db/schema.rb
```diff
@@ -170,8 +171,8 @@
     t.integer "accrediting_provider_id"
     t.boolean "exposed_in_find"
     t.boolean "open_on_apply", default: false, null: false
-    t.string "study_mode", limit: 1, default: "F", null: false
     t.integer "recruitment_cycle_year", null: false
+    t.string "study_mode", limit: 1, default: "F", null: false
```
mnacos: How do we resolve this? The db migration for study_mode has an earlier timestamp than the one for recruitment_cycle_year, but the recruitment_cycle_year code was merged first. Any new contributor will have study_mode first, recruitment_cycle_year second, do you think you could rollback the migrations and re-apply?

spec/lib/openapi_spec.rb
```diff
@@ -13,7 +13,7 @@
         required_properties = schema.required.to_a
         actual_properties = schema.properties.map { |property_name, _| property_name }
 
-        expect(required_properties).to match_array(actual_properties)
+        expect(actual_properties).to include(*required_properties)
```
mnacos: nice!

app/services/make_an_offer.rb
```diff
@@ -8,15 +8,17 @@ class MakeAnOffer
 
   validate :validate_offer_conditions
```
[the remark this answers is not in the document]
mnacos: And you've already done that :)

app/services/make_an_offer.rb
```diff
@@ -37,6 +39,14 @@ def save
   attr_reader :application_choice
 
   def validate_offer_conditions
+    unless @offered_course_option.nil?
+      current_provider = @application_choice.course_option.course.provider
+      offered_course_provider = @offered_course_option.course.provider
+      providers_dont_match = current_provider != offered_course_provider
```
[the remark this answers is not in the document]
tvararu: Correct, I say we cross that bridge when it comes? 👍

db/schema.rb
tvararu: Good spot, yes!

app/services/make_an_offer.rb
```diff
@@ -36,6 +47,48 @@ def save
...
+      return
+    end
+
+    @site = @provider.sites.find_by(code: @site_code)
+    if @site.nil?
+      errors.add(:offered_course, "site #{@site_code} does not exist or does not belong to provider #{@provider_code}")
+      return
+    end
+
+    @course_option = CourseOption.find_by(course: @course, site: @site, study_mode: @study_mode)
```
tvararu: @mnacos is this `study_mode: @study_mode` what you suggested? (I just added it)

config/vendor-api-v1.yml
```diff
@@ -547,7 +549,7 @@ components:
           example: full_time
     Offer:
       type: object
-      additionalProperties: false
+      additionalProperties: true
```
duncanjbrown: We've preferred to use `nullable: true` on the optional fields elsewhere for 2 reasons I can think of:

- it's better UX as clients know exactly what to expect
- our specs are more robust, as if something unexpected ends up here then the spec will fail

Would it be better in any case to always include the offer if present? Less for clients to think about

app/presenters/vendor_api/single_application_presenter.rb
```diff
@@ -78,13 +78,13 @@ def nationalities
       ].map { |n| NATIONALITIES_BY_NAME[n] }.compact
     end
 
-    def course
+    def course_info_for(application_choice_or_course_option)
```
duncanjbrown: What would you think of naming this parameter `course_option`? It's a duck. If we don't like that (i.e. depending on two types having this interface) we could read `course_option` off `application_choice` before passing it in?

config/vendor-api-v1.yml
tvararu: ok! 👍

app/presenters/vendor_api/single_application_presenter.rb
tvararu: Done :+1:, it also uncovered N+1 queries that were being not picked up before

config/vendor-api-v1.yml
tvararu: Check the last commit, have done this

tvararu: Check the last commit, have done this

app/services/make_an_offer.rb
```diff
@@ -36,6 +47,48 @@ def save
 
   attr_reader :application_choice
 
+  def offered_course_option
+    return nil if @course_data.nil?
```
duncanjbrown: We can return the original option if no override was supplied

app/services/make_an_offer.rb
```diff
@@ -36,6 +47,48 @@ def save
...
+
+    @course_option
+  end
+
+  def validate_course_data
+    return if @course_data.nil?
```
duncanjbrown: idea:

1. load C_o in the constructor using a method which returns a validatable thing
1. ask here whether that thing is valid, if not assign its errors to ourself

spec/lib/openapi_spec.rb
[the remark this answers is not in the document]
tvararu: Discussed offline, due to the `MakeOffer` schema we can't use the stricter version anymore.

tvararu:
Discussed offline, but as this PR is already very big and unwieldy / hard to rebase, we think it's pragmatic to merge it and open up the suggested refactor of the instance variables in a follow-up PR.

PR #849, opened 2019-12-11
Minor API documentation fixes

stevehook:
### Context

This is a follow-up to https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/826. A number of issues were raised after manual testing by @fofr . This PR addresses the remaining fixes to API docs.

- [x] Work experience commitment example and description needed (already done)
- [x] Qualification examples are missing the "institution_name" and "equivalency" examples (already done)
- [x] We should include an example of an ISO8601 date in the since param (PH actually caused an error with this when he tried it) - (done in earlier PR)
- [x] reference content should not be marked "optional" bc we only send apps with references - changed the wording to emphasise that content is required to send apps to providers
- [x] Clarify that tokens can only be passed via header, not via URL param - added a bit of extra copy to clarify

I've not tried to add the state diagram (suggestion from @duncanjbrown) in this PR because there isn't an obvious place to put it in the docs (without causing a confusion). This one is open to debate!

### Guidance to review

- Do the copy changes make sense?

### Link to Trello card

[1176 - Minor API fixes following user testing with Paul H](https://trello.com/c/MVlw347D/1176-minor-api-fixes-following-user-testing-with-paul-h)

### Env vars

- [x] No env vars


config/vendor-api-0.8.0.yml
```diff
@@ -597,7 +597,7 @@ components:
           example: Julia was my mentor at Cheadle Hulme High
         reference:
           type: string
-          description: "Optional. The reference content provided by the referee, once it is available"
+          description: "Required to send the application to the provider. The reference content provided by the referee, once it is available"
```
duncanjbrown: Do we need this qualifier at all (or the `once is it available` clause) if providers will only see responses with references complete?

tijmenb:
```suggestion
          description: "The reference content provided by the referee."
```

config/vendor-api-0.8.0.yml
```diff
@@ -597,7 +597,7 @@ components:
...
           example: Boris is committed to the profession of teaching...
           nullable: true
```
tijmenb:
```suggestion
```

stevehook:
> Perhaps the state diagram could go in the general introduction? On the basis that it's a useful high-level summary of how the thing works. We might need to prune it a bit.

@duncanjbrown I've added a section to the api-docs home page called 'Application Lifecycle' with a copy of the state diagram. Wording is all up for debate. I've not tried to trim the diagram (it's just automatically copied from the `generate_state_diagram` rake task into the `public` directory).

duncanjbrown:
> > Perhaps the state diagram could go in the general introduction? On the basis that it's a useful high-level summary of how the thing works. We might need to prune it a bit.
> 
> @duncanjbrown I've added a section to the api-docs home page called 'Application Lifecycle' with a copy of the state diagram. Wording is all up for debate. I've not tried to trim the diagram (it's just automatically copied from the `generate_state_diagram` rake task into the `public` directory).

:+1: I think this is useful

PR #854, opened 2019-12-11
hardcode study_mode in application API responses

mnacos:
### Context

We'll be adding a ```study_mode``` attribute to CourseOption. More details here:
https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/851

This PR just hardcodes the value in the API responses until the full solution is built.

Add a ```study_mode``` attribute to the ```course``` section of applications, as returned by the API.

### Guidance to review

Run ```rspec spec/presenters/vendor_api/single_application_presenter_spec.rb```.

### Link to Trello card

[1345 - API: add fake full or part time to course selection](https://trello.com/c/qx0OfPLx)

### DIFF
--- a/config/vendor-api-0.8.0.yml
+++ b/config/vendor-api-0.8.0.yml
@@ -515,6 +515,10 @@ components:
           description: The site’s code
           example: K
           maxLength: 5
+        study_mode:
+          type: string
+          description: Can be 'full_time' or 'part_time'
+          example: full_time
     Offer:
       type: object
       additionalProperties: false
--- a/config/vendor-api-0.8.0.yml
+++ b/config/vendor-api-0.8.0.yml
@@ -493,6 +493,7 @@ components:
         - provider_code
         - course_code
         - site_code
+        - study_mode
       properties:
         start_date:
           type: string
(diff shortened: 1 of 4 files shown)


spec/presenters/vendor_api/single_application_presenter_spec.rb
```diff
@@ -19,6 +19,17 @@
     end
   end
 
+  describe '#course' do
```
tijmenb: There's this system test that we can use to test the API output end-to-end:

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/spec/system/vendor_api/vendor_receives_application_spec.rb#L52

@duncanjbrown added that to avoid us having to test every attribute separately.

config/vendor-api-0.8.0.yml
```diff
@@ -493,6 +493,7 @@ components:
         - provider_code
         - course_code
         - site_code
+        - study_mode
```
tijmenb: This also needs the property defined below.

spec/presenters/vendor_api/single_application_presenter_spec.rb
mnacos: Great, thanks

PR #860, opened 2019-12-12
Spike into implementing Apply 2

tijmenb:
### Context

We need to implement "Apply 2". This is when the first application was unsuccesful, and the candidate can now apply to 1 course at the time.

Things that we want from this system:

- [x] Candidates don't have to fill in stuff they've previously filled in
- [x] Candidates can change everything about their applications
- [x] Number of concurrent applications is limited to 1
- [ ] Guidance changes
- [ ] It's exposed in the API
- [ ] It doesn't mess with our reporting
- [ ] The code is easy to follow
- [ ] Users can't edit references

Spikey code. The flow looks like this:

1. When the application is unsuccessful, the user gets a button to "try again"
2. This button duplicates the application, minus the choices
3. The user can then edit the application and make 1 new choice
4. Application is submitted as normal

thing

### Guidance to review

Can we think of other ways of implementing this?

### Link to Trello card

https://trello.com/c/GJb6DGs8

### DIFF
--- a/app/controllers/candidate_interface/application_form_controller.rb
+++ b/app/controllers/candidate_interface/application_form_controller.rb
@@ -8,6 +8,11 @@ module CandidateInterface
       @application_form = current_application
     end
 
+    def apply_2
+      DuplicateApplication.new(current_application).duplicate
+      redirect_to candidate_interface_application_form_path
+    end
+
     def review
       redirect_to candidate_interface_application_complete_path if current_application.submitted?
       @application_form = current_application
--- a/app/services/duplicate_application.rb
+++ b/app/services/duplicate_application.rb
@@ -1,0 +1,43 @@
+class DuplicateApplication
+  attr_reader :original_application_form
+
+  def initialize(original_application_form)
+    @original_application_form = original_application_form
+  end
+
+  def duplicate
+    attrs = original_application_form.attributes.except(*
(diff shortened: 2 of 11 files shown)


app/controllers/candidate_interface/application_form_controller.rb
```diff
@@ -7,6 +7,11 @@ def show
       @application_form = current_application
     end
 
+    def apply_2
+      DuplicateApplication.new(current_application).duplicate
+      redirect_to candidate_interface_application_form_path
+    end
```
tvararu: This probably needs some checking to prevent double submission / back button from generating 2+ new applications. For example, we could guard against duplicating if there already is an unsubmitted application.

app/services/duplicate_application.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def duplicate
+    attrs = original_application_form.attributes.except(*
+      %w[id created_at updated_at submitted_at course_choices_completed phase]
+    ).merge(
+      phase: 'apply_2'
+    )
+
+    new_application_form = ApplicationForm.create!(attrs)
```
tvararu: 👍 This is easy to follow for me, and feels robust

tijmenb:
Closing until we pick up Apply 2 again!

PR #861, opened 2019-12-12
Replace `Course#start_date` with `recruitment_cycle_year`

stevehook:
### Context

We currently sync `Course#start_date` from Find but don't record the `recruitment_cycle_year` which would be very helpful for filtering courses for the current recruitment cycle.

This PR adds `Course#recruitment_cycle_year` but leaves `start_date` as is. To remove `start_date` we'll need to do that as a separate deploy.

- [X] Add `Course#recruitment_cycle_year` attribute
- [X] Sync `Course#recruitment_cycle_year` from Find
- [X] Return `Course#recruitment_cycle_year` from the Vendor API
- [X] Remove references to `Course#start_date` (so that the column can be dropped in a future PR)

### Guidance to review

- Are there any other interfaces (other than the Vendor API) in which we need to expose `recruitment_cycle_year`?
- Are there any tests that I've missed?
- Are we sure we want to drop `Course#start_date`? (we could keep both attributes)

### Link to Trello card

[1377 - Replace `course.start_date` with `course.recruitment_cycle_year`](https://trello.com/c/sVZuvPq7/1377-replace-coursestartdate-with-courserecruitmentcycleyear)

### Env vars

- [X] No new env vars


config/vendor-api-v1.yml
```diff
@@ -497,17 +497,16 @@ components:
       type: object
       additionalProperties: false
       required:
-        - start_date
+        - recruitment_cycle_year
```
tijmenb: We'll need to update the release notes too!

db/migrate/20191211120822_add_courses_recruitment_cycle_year.rb
```diff
@@ -0,0 +1,7 @@
+class AddCoursesRecruitmentCycleYear < ActiveRecord::Migration[6.0]
+  def change
+    add_column :courses, :recruitment_cycle_year, :integer
+    Course.update_all recruitment_cycle_year: 2020
+    change_column :courses, :recruitment_cycle_year, :integer, null: false
```
tijmenb: FYI you can replace above 2 lines with `change_column_null(:courses, :recruitment_cycle_year, 2020)`

https://apidock.com/rails/ActiveRecord/ConnectionAdapters/SchemaStatements/change_column_null

config/vendor-api-v1.yml
stevehook: :+1: Have updated alpha release notes. Only issue with this is that I have to guess when the release will go out (I've guessed tomorrow).

db/migrate/20191211120822_add_courses_recruitment_cycle_year.rb
stevehook: Thanks, much tidier. More importantly this migration wasn't reversible - I guess we should be making that a requirement going forward.

config/vendor-api-v1.yml
tijmenb: Good guess :)

PR #864, opened 2019-12-12
Drop `courses.start_date` database column

stevehook:
### Context

This is a follow up to https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/861 to just clean up the now redundant `courses.start_date` database column.

Migration to drop `start_date` column. (It's reversible in so much as you can rollback to recreate the column but it won't get populated).

### Guidance to review

- Do we really need to do this?

### Link to Trello card

[1377 - Replace `course.start_date` with `course.recruitment_cycle_year`](https://trello.com/c/sVZuvPq7/1377-replace-coursestartdate-with-courserecruitmentcycleyear)

### Env vars

- [X] No new env vars


duncanjbrown:
Rerunning the pipeline here as it looks like a transient network error

PR #867, opened 2019-12-12
Simplify environment variables - Phase 1

tpinney:
### Context

Environment variables for the app service and two container instances are currently specified individually for each instance. With this arrangement it is possible to missing environment variables that are required on one instance but not another resulting in a failure deploying. This change ensure that a common set of variables are specified for all instances in one place.

- The app service and container instances components of the ARM template have been split into a new nested template called `containers.json`
- The existing `template.json` now references this new nested template and this is where the environment variables are specified for all instances deployed by the nested template.
- The environment variables value types permitted are "Value" or "secureValue" with the latter being used for secrets such as passwords and access keys etc.
- Updated the README file to reflect the subtle change in the process with the environment variables.

### Guidance to review

- This pipeline shows a complete successful deployment of the new configuration: https://dfe-ssp.visualstudio.com/Become-A-Teacher/_build/results?buildId=35026&view=results
- It would be worth someone spending some time to check that nothing is broken in QA once this change has been merged because the DevOps environment is not properly configured at the app level for me to be sure.

### Link to Trello card

[1180 - Simplify environment variables in pipeline](https://trello.com/c/6htI39B2/1180-simplify-environment-variables-in-pipeline)

### Env vars

### DIFF
--- a/azure/template.json
+++ b/azure/template.json
@@ -48,7 +48,7 @@
         "appServicePlanSize": {
             "type": "string",
             "metadata": {
-                "description": "The App Serivce Plan instance size."
+                "description": "The App Service Plan instance size."
             }
         },
         "appServicePlanInstances": {
@@ -774,7 +774,7 @@
                         "value": "AZURE_IP-"
                     },
                     "ipAddresses": {
-                        "value": "[reference('app-service-and-containers').outputs.appSerivceIpAddresses.value]"
+                        "value": "[reference('app-service-and-containers').outputs.appServiceIpAddresses.value]"
                     },
                     "serverName": {
                         "value": "[variables('databaseServerName')]"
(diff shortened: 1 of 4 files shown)


azure/template.json
```diff
@@ -1034,7 +774,7 @@
                         "value": "AZURE_IP-"
                     },
                     "ipAddresses": {
-                        "value": "[reference('app-service').outputs.possibleOutboundIpAddresses.value]"
+                        "value": "[reference('app-service-and-containers').outputs.appSerivceIpAddresses.value]"
```
stevehook: Looks like there is a typo here.

tpinney: That is indeed a typo, but one I carried through in the code so the functionality still worked, but I have now submitted a correction for it.

PR #869, opened 2019-12-12
Do not assert on autogenerated ID in spec

duncanjbrown:
We don't decide what IDs things get, postgres does. This caused
intermittent spec failures.

There are plenty of other specs that explicitly set IDs. These may start
flaking too, but unlike this spec, they start to break once the explicit
ID is removed, eg /spec/components/volunteering_review_component_spec.rb:71


duncanjbrown:
I broke another one 💯

PR #870, opened 2019-12-12
Add Referee page, with authentication token

alexdesi:
## Context

This PR is the first slice of replacing the google form to a Referee page. This includes:

- Create the token for the referee (stored on Reference model)
- Create a new page to allow the referee to enter the comments https://apply-beta-prototype.herokuapp.com/reference/comments
- Include the token in the email and check the token is correct when trying to access the reference view

- new Token on `Reference` model
- update email with new Token
- Add new view/controller in a new namespace
- Added new feature Flag `reference_form`

## Guidance to review
Tests are a good starting point

## Link to Trello card

https://trello.com/c/RrlqYDMz/585-referee-can-provide-a-reference-via-dfe-apply-service-%F0%9F%8F%88

### DIFF
--- a/app/controllers/referee_interface/reference_controller.rb
+++ b/app/controllers/referee_interface/reference_controller.rb
@@ -2,11 +2,37 @@ module RefereeInterface
   class ReferenceController < ActionController::Base
     layout 'application'
 
-    def comments
+    def feedback
       reference = Reference.find_by(token: params[:token])
 
       if reference.present?
+        @reference_form = RefereeInterface::ReferenceForm.build_from_reference(reference)
         @application = reference.application_form
+      else
+        render_404
+      end
+    end
+
+    def submit_feedback
+      reference = Reference.find_by(token: params[:token])
+      reference.feedback = params[:reference][:comments]
+      @application = reference.application_form
+
+      @reference_form = RefereeInterface::ReferenceForm.build_from_reference(reference)
+
+      if @reference_form.save(reference)
+        redirect_to referee_interface_confirmation_path
+      else
+        render :feedback
+      end
+    end
+
+    def finish; end
+
+    def confirmation
+      @reference = Reference.find_by(token: params[:token])
+      if @reference.present?
+        render :confirmation
       else
         render_404
       end
--- a/app/controllers/referee_interface/reference_controller.rb
+++ b/app/controllers/referee_interface/reference_controller.rb
@@ -3,8 +3,6 @@ module RefereeInterface
     layout 'application'
 
     def feedback
-      reference = Reference.find_by(token: params[:token])
-
       if reference.present?
         @reference_form = RefereeInterface::ReferenceForm.build_from_reference(reference)
         @application = reference.application_form
@@ -14,7 +12,6 @@ module RefereeInterface
     end
 
     def submit_feedback
-      reference = Reference.find_by(token: params[:token])
       reference.feedback = params[:reference][:comments]
       @application = reference.application_form
 
@@ -30,9 +27,18 @@ module RefereeInterface
     def finish; end
 
     def confirmation
-      @reference = Reference.find_by(token: params[:token])
-      if @reference.present?
+      if reference.present?
         render :confirmation
+      else
+        render_404
+      end
+    end
+
+    def decline
+      if reference.present?
+        @application = reference.application_form
+        #TODO: add logic to record this and notify candidate
+        render :decline
       else
         render_404
       end
@@ -42,6 +48,10 @@ module RefereeInterface
 
     def render_404
       render 'errors/not_found', status: :not_found
+    end
+
+    def reference
+      @reference ||= Reference.find_by(token: params[:token])
     end
   end
 end
--- a/app/controllers/referee_interface/reference_controller.rb
+++ b/app/controllers/referee_interface/reference_controller.rb
@@ -1,0 +1,21 @@
+module RefereeInterface
+  class ReferenceController < ActionController::Base
+    layout 'application'
+
+    def comments
+      reference = Reference.find_by(token: params[:token])
+
+      if reference.present?
+        @application = reference.application_form
+      else
+        render_404
+      end
+    end
+
+  private
+
+    def render_404
+      render 'errors/not_found', status: :not_found
+    end
+  end
+end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -36,4 +36,8 @@ module ApplicationHelper
   def current_namespace
     params[:controller].split('/').first
   end
+
+  def full_name(application_form)
+    "#{application_form.first_name} #{application_form.last_name}"
+  end
 end
--- a/app/models/referee_interface/reference_form.rb
+++ b/app/models/referee_interface/reference_form.rb
@@ -1,0 +1,24 @@
+module RefereeInterface
+  class ReferenceForm
+    include ActiveModel::Model
+
+    attr_accessor :feedback
+
+    validates :feedback,
+              word_count: { maximum: 300 },
+              presence: true
+
+    def self.build_from_reference(reference)
+      new(
+        feedback: reference.feedback,
+      )
+    end
+
+    def save(reference)
+      return false unless valid?
+
+      reference.update!(feedback: feedback)
+      true
+    end
+  end
+end
--- a/app/models/reference.rb
+++ b/app/models/reference.rb
@@ -29,4 +29,9 @@ class Reference < ApplicationRecord
 
     errors.add(:email_address, :own) if email_address == candidate_email_address
   end
+
+  def update_token
+    raw_token, _encrypted_token = Devise.token_generator.generate(Reference, :token)
+    update(token: raw_token)
+  end
 end
--- a/app/services/submit_application.rb
+++ b/app/services/submit_application.rb
@@ -21,7 +21,9 @@ class SubmitApplication
 private
 
   def send_reference_request_email_to_referees(application_form)
-    application_form.references.each do |reference|
+    application_form.references.includes(:application_form).each do |reference|
+      reference.update_token
+
       RefereeMailer.reference_request_email(application_form, reference).deliver_later
     end
   end
@@ -30,11 +32,6 @@ private
     application_choices.each do |application_choice|
       application_choice.edit_by = ApplicationDates.new(application_form).edit_by
       ApplicationStateChange.new(application_choice).submit!
-    end
-
-    application_form.references.includes(:application_form).each do |ref|
-      raw_token, _encrypted_token = Devise.token_generator.generate(Reference, :token)
-      ref.update(token: raw_token)
     end
   end
 end
--- a/app/services/submit_application.rb
+++ b/app/services/submit_application.rb
@@ -32,10 +32,9 @@ private
       ApplicationStateChange.new(application_choice).submit!
     end
 
-    @application_form.reload.references.each do |reference|
+    application_form.references.includes(:application_form).each do |ref|
       _raw_token, encrypted_token = Devise.token_generator.generate(Reference, :token)
-
-      reference.update(token: encrypted_token)
+      ref.update(token: encrypted_token)
     end
   end
 end
--- a/app/services/submit_application.rb
+++ b/app/services/submit_application.rb
@@ -33,8 +33,8 @@ private
     end
 
     application_form.references.includes(:application_form).each do |ref|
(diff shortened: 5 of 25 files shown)


alexdesi:
Screenshot 2019-12-12 at 17 00 30

app/services/submit_application.rb
```diff
@@ -31,5 +31,10 @@ def submit_application
       application_choice.edit_by = ApplicationDates.new(application_form).edit_by
       ApplicationStateChange.new(application_choice).submit!
     end
+
+    application_form.references.includes(:application_form).each do |ref|
```
tijmenb: Perhaps we should do this in the same loop as the mail is sent, because the ordering is important.

alexdesi: Yes, it's nice to have the update at the same time the email is sent.

I've just one small concern: `send_reference_request_email_to_referees` will do 2 different things, instead of just one. (but I think it's fine moving the update_token code in the model, have a look)

app/controllers/referee_interface/reference_controller.rb
```diff
@@ -0,0 +1,47 @@
+module RefereeInterface
+  class ReferenceController < ActionController::Base
+    layout 'application'
```
tijmenb: We'll need to add some things here that the other controllers have as well:

- `include LogQueryParams`
- An implementation of `before_action :add_identity_to_log` so that we know in the logs & Sentry who triggered a log or error

See the candidate bit for an example:

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/d4010b0622ac6d86239210ca8197b72acbcc941a/app/controllers/candidate_interface/candidate_interface_controller.rb#L3-L7

app/controllers/referee_interface/reference_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    def feedback
+      reference = Reference.find_by(token: params[:token])
```
tijmenb: If you do `Reference.find_by!` it'll trigger an `RecordNotFound` exception that you can catch with a `rescue_from`.

app/controllers/referee_interface/reference_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+      if reference.present?
+        @reference_form = RefereeInterface::ReferenceForm.build_from_reference(reference)
+        @application = reference.application_form
+      else
+        render_404
+      end
+    end
+
+    def submit_feedback
+      reference = Reference.find_by(token: params[:token])
```
tijmenb: This doesn't check if the token exists before doing the work, using `find_by!` would help. 

I also wonder what happens if the token is empty, since it'll match the `Reference` that currently don't have a `token` set.

app/helpers/application_helper.rb
```diff
@@ -36,4 +36,8 @@ def service_link
   def current_namespace
     params[:controller].split('/').first
   end
+
+  def full_name(application_form)
```
tijmenb: This probably warrants to be put in `ApplicationForm`.

app/views/referee_interface/reference/finish.html.erb
```diff
@@ -0,0 +1,7 @@
+<main class="govuk-main-wrapper govuk-main-wrapper--auto-spacing" id="main-content" role="main">
+  <h1 class="govuk-heading-xl">Thank you</h1>
+    <div class="govuk-grid-row">
```
tijmenb: Minor: some weird indentation here.

app/views/referee_mailer/reference_request_email.text.erb
```diff
@@ -1,20 +1,10 @@
...
-<% end %>
-
-Please use the link below to give a complete reference as soon as possible.
-
-<%= @google_form_url %>
-
-If we don’t hear from you within 5 working days we’ll send you a reminder.
-
-If you won’t give <%= @candidate_name %> a reference, please let us know as soon as possible by emailing: [becomingateacher@digital.education.gov.uk](mailto:becomingateacher@digital.education.gov.uk)
+<%= @reference_link %>
```
tijmenb: Do we not need to add the paragraph describing what to do when they don't want the reference here?

config/environments/test.rb
```diff
@@ -39,7 +39,7 @@
   config.action_mailer.default_options = {
     from: 'mail@example.com',
   }
-  config.action_mailer.default_url_options = { host: 'localhost', port: 3000 }
+  config.action_mailer.default_url_options = { host: 'www.example.com', port: 3000 }
```
tijmenb: What's the reason behind this change?

config/routes.rb
```diff
@@ -211,6 +211,13 @@
     end
   end
 
+  namespace :referee_interface, path: '/reference' do
+    get '/finish' => 'reference#finish', as: :reference_finish
+    get '/:token/confirmation' => 'reference#confirmation', as: :confirmation
+    get '/:token' => 'reference#feedback', as: :reference_feedback
+    patch '/:token/confirmation' => 'reference#submit_feedback', as: :submit_feedback
+  end
```
tijmenb: By default we would log the full URL, including the token. If we'd added the token in the query param it would be automatically filtered out.

app/views/referee_interface/reference/feedback.html.erb
```diff
@@ -0,0 +1,49 @@
...
+            <p class="govuk-body">For example, can the candidate meet deadlines and manage their time?</p>
+          </li>
+        </ul>
+
+        <div class="govuk-character-count" data-module="govuk-character-count" data-maxwords="300">
+          <div class="govuk-form-group">
+            <label class="govuk-label govuk-label--m govuk-!-margin-top-6" for="reference-comments">
+              Your reference
+            </label>
+            <textarea class="govuk-textarea govuk-js-character-count" id="reference-comments" name="reference[comments]" rows="15" aria-describedby="reference-comments-info"></textarea><span id="reference-comments-info" class="govuk-hint govuk-character-count__message" aria-live="polite">You have 300 words remaining</span>
```
tijmenb: Any reason not to use the formbuilder here?

app/models/reference.rb
```diff
@@ -29,4 +29,9 @@ def email_address_not_own
 
     errors.add(:email_address, :own) if email_address == candidate_email_address
   end
+
+  def update_token
+    raw_token, _encrypted_token = Devise.token_generator.generate(Reference, :token)
+    update(token: raw_token)
```
tijmenb: I would've expected the `raw_token` to be sent to the user, and only the encrypted token saved in the database. 

This is how `VendorApiToken` and `Candidate` tokens work - it prevents the tokens from being used if they ever leak. Open to discussion here!

app/views/referee_interface/reference/confirmation.html.erb
```diff
@@ -0,0 +1,11 @@
+<main class="govuk-main-wrapper govuk-main-wrapper--auto-spacing" id="main-content" role="main">
+
+  <%= form_with model: @reference, url: referee_interface_reference_finish_path, method: :get do |f| %>
```
tijmenb: This doesn't need to be a form I think?

app/views/referee_interface/reference/feedback.html.erb
```diff
@@ -0,0 +1,49 @@
+<% content_for :title, t('page_titles.application') %>
+
+<main class="govuk-main-wrapper govuk-main-wrapper--auto-spacing" id="main-content" role="main">
+
+  <%= form_with model: @reference_form, url: referee_interface_submit_feedback_path, method: :patch do |f| %>
```
tijmenb: I'm surprised the token is automatically passed into this URL, didn't know Rails did that!

app/models/reference.rb
tijmenb: `update!` to avoid silent failures (we're not checking the return value).

```suggestion
    update!(token: raw_token)
```

app/controllers/referee_interface/reference_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+
+      if reference.present?
+        @reference_form = RefereeInterface::ReferenceForm.build_from_reference(reference)
+        @application = reference.application_form
+      else
+        render_404
+      end
+    end
+
+    def submit_feedback
```
tijmenb: We're going to have to check if the feedback has already been given, otherwise we're allowing edits forever, which might lead to weirdness.

app/models/referee_interface/reference_form.rb
```diff
@@ -0,0 +1,24 @@
+module RefereeInterface
+  class ReferenceForm
```
tijmenb: We should use (a version of / refactored version of) `ReceiveReference` instead of this new form object. 

In the flow, this makes sure that when all references are complete the state is changed and we send the applications to the providers.

app/views/referee_mailer/reference_request_email.text.erb
C-gyorfi: Yes, I resolved this in 4354366

app/controllers/referee_interface/reference_controller.rb
alexdesi: Good point, I've added in my todo list.

app/helpers/application_helper.rb
alexdesi: it sounds good (possible we'll need a Container for that view, so I might put full_name in it)

app/models/reference.rb
alexdesi: Yes, I should have added a big `TODO` there, it needs to be secured.

app/views/referee_interface/reference/confirmation.html.erb
alexdesi: That's because the radio button for the consent to be contacted is coming soon,
so it'll need to be a form.

app/views/referee_interface/reference/feedback.html.erb
alexdesi: you made me doubt of it :D, I think it is working,  I'll double check

app/views/referee_interface/reference/feedback.html.erb
alexdesi: not really, I'll fix it!

app/models/reference.rb
alexdesi: Acctually we added that iten to the https://trello.com/c/B4tokVQU/§, because we were aware this PR was getting long.

app/views/referee_mailer/reference_request_email.text.erb
alexdesi: Acctually we added that iten to the https://trello.com/c/B4tokVQU/§, because we were aware this PR was getting long.

config/environments/test.rb
alexdesi: That is only of tests, I thought that having to use `localhost:3000` inside the test was a bit weird.
But not strong opinion on that, possible I should keep that out, since this PR is already quite long.

app/controllers/referee_interface/reference_controller.rb
alexdesi: good point! 
I think we need all of them but `before_action :authenticate_candidate! ` since in this case, the user is the referee

alexdesi: For the  `before_action :authenticate_referee!`
I am gonna use the same password as `ENV.fetch('BASIC_AUTH_PASSWORD'))`

I suppose it makes sense since who will test this in staging/production is the same user that creates the application form.

alexdesi: I am not sure the data to log with `add_identity_to_log`.
Probably just the `id` of the `referee`.  (like the support controllers, candidate controller, ...)

app/models/referee_interface/reference_form.rb
alexdesi: As we discussed, it's fine to have a bit of duplication.

app/controllers/referee_interface/reference_controller.rb
alexdesi: I've added a condition on the `reference_controller#feedback` to check if the feedback already exists.

app/models/reference.rb
alexdesi: Done

alexdesi: done

app/views/referee_interface/reference/feedback.html.erb
alexdesi: anyway, now it uses a query string `?token=XXXXXX`

PR #872, opened 2019-12-12
Use database for provider users

duncanjbrown:
After this change ProviderUsers will be granted access via a database association instead of via the YAML file.

- push the YAML-backed behaviour into a `LegacyProviderUser`
- dispatch the creation of a (proper, database-backed) `ProviderUser` or a `LegacyProviderUser` using a feature flag in the `#load_from_session` method on `ProviderUser` (which all callers use)
- create a new `#provider` method on `ProviderUser` which looks up associated providers via a HABTM association

This PR will not implement support for multiple providers, though the HABTM makes it  very possible.

A follow-up PR will adding UI for managing ProviderUsers.

Migration should look something like this:

1. deploy this PR
1. implement UI for managing users
1. for each env, use the UI to add the provider users and then switch on the feature flag
1. remove the feature flag and the mapping

https://trello.com/c/f0LD8ZaP/1380-use-database-for-provider-users


db/migrate/20191212113810_create_provider_users.rb
```diff
@@ -0,0 +1,10 @@
+class CreateProviderUsers < ActiveRecord::Migration[6.0]
+  def change
+    create_table :provider_users do |t|
+      t.string :email_address
+      t.string :dfe_sign_in_uid
```
tijmenb: Should these be `null: false`?

spec/factories.rb
```diff
@@ -210,4 +210,9 @@
     dfe_sign_in_uid { SecureRandom.hex(5) }
     email_address { "#{Faker::Name.first_name}@example.com" }
   end
+
+  factory :provider_user do
+    dfe_sign_in_uid { SecureRandom.hex(5) }
```
tijmenb:
```suggestion
    dfe_sign_in_uid { SecureRandom.uuid }
```

PR #876, opened 2019-12-13
Ensure that qualifications and experiences are audited

stevehook:
## Context

We need to see activity in the audit trail (History page in the Support Interface) for applications and all of the child objects. This includes volunteering experiences, work experience and qualifications.

Tag the relevant models with the `audited` method so that they are all related to their parent application form.

image

## Guidance to review

- Are tests sufficient? I've added a few model specs that are somewhat repetitive - they only really ensure that we don't mess up the configuration for these three models in future.

Has been tested manually in my dev env.

## Link to Trello card

https://trello.com/c/IN5xB5qu/1390-qualifications-work-history-and-volunteering-roles-arent-showing-up-in-the-audit-trail

## Things to check

### DIFF
--- a/spec/models/application_qualification_spec.rb
+++ b/spec/models/application_qualification_spec.rb
@@ -30,4 +30,20 @@ RSpec.describe ApplicationQualification, type: :model do
         .to raise_error ArgumentError, "'invalid_level' is not a valid level"
     end
   end
+
+  describe 'auditing', with_audited: true do
+    it 'creates audit entries' do
+      application_form = create :application_form
+      application_qualification = create :application_qualification, application_form: application_form
+      expect(application_qualification.audits.count).to eq 1
+      application_qualification.update!(subject: 'Rocket Surgery')
+      expect(application_qualification.audits.count).to eq 2
+    end
+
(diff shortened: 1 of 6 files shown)


spec/models/application_qualification_spec.rb
```diff
@@ -30,4 +30,20 @@
...
     end
   end
+
+  describe 'auditing', with_audited: true do
+    it 'creates audit entries' do
+      application_form = create :application_form
+      application_qualification = create :application_qualification, application_form: application_form
+      expect(application_qualification.audits.count).to eq 1
+      application_qualification.update!(subject: 'Rocket Surgery')
+      expect(application_qualification.audits.count).to eq 2
```
benilovj: May be a personal preference, but would this read better if it were:

```ruby
expect { application_qualification.update!(subject: 'Rocket Surgery' }
  .to change { application_qualification.audits.count }.from(1).to(2)
```

or even

```ruby
expect { application_qualification.update!(subject: 'Rocket Surgery' }
  .to change { application_qualification.audits.count }.by(1)
```

PR #878, opened 2019-12-13
Fix bug where users can sign in with an expired token

tvararu:
## Context

If an existing session is in place, we follow a different code path which does not check for the `token_not_expired?` method. This allows users to sign in using an expired token.

Hoist the `token_not_expired?` logic inside `FindCandidateByToken` so it's consistently applied wherever it's used.

## Guidance to review

Start by understanding the test; the rest of the implementation can be updated as needed, it feels clunky to me but I decided to open it for others to comment anyway.

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/sign_in_controller.rb
+++ b/app/controllers/candidate_interface/sign_in_controller.rb
@@ -23,10 +23,11 @@ module CandidateInterface
     end
 
     def authenticate
-      user = FindCandidateByToken.call(raw_token: params[:token])
-      if user
-        sign_in(user, scope: :candidate)
-        add_identity_to_log user.id
+      candidate = FindCandidateByToken.call(raw_token: params[:token])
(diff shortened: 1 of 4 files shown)


app/controllers/candidate_interface/sign_in_controller.rb
```diff
@@ -23,10 +23,11 @@ def create
...
     def authenticate
-      user = FindCandidateByToken.call(raw_token: params[:token])
-      if user
-        sign_in(user, scope: :candidate)
-        add_identity_to_log user.id
+      candidate = FindCandidateByToken.call(raw_token: params[:token])
+
+      if candidate
+        sign_in(candidate, scope: :candidate)
+        add_identity_to_log candidate.id
```
tvararu: This is just `s/user/candidate/` for consistency.

PR #883, opened 2019-12-16
Use a string for the CourseOption#study_mode enum

tijmenb:
## Context

Follow up to https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/851. In other models we’ve been using strings in enums instead of Rails’ default integers.

This has two advantages:

- Accidental changes in the `enum` will not change the meaning of the data (for example, changing it to `enum study_mode: %i[part_time full_time]` would silently change the data).
- The data in the database makes sense without having access to the code. This makes debugging a little easier.

Use a string for the column. Also adds the possible values to the docs.

## Guidance to review

Read.

## Link to Trello card

https://trello.com/c/4XttCLgg/1189-support-full-time-and-part-time-courses

## Things to check


config/vendor-api-v1.yml
```diff
@@ -545,6 +545,9 @@ components:
           type: string
           description: Can be 'full_time' or 'part_time'
```
duncanjbrown: This description is redundant now, I think?

PR #895, opened 2019-12-17
Support interface candidate list improvements

benilovj:
## Context

Usability improvements to the support interface.

Within the support interface:

- Make the candidates view the default one (and not the application forms view)
- Replace the applications view with the candidates view in the support nav
- Surface the status of the latest application form on the candidates list, including the case when the candidate has never actually logged into the service
- When support agents click on a candidate with only one application, they are redirected to that application's page

image

### DIFF
--- a/spec/system/support_interface/see_candidates_spec.rb
+++ b/spec/system/support_interface/see_candidates_spec.rb
@@ -3,7 +3,7 @@ require 'rails_helper'
 RSpec.feature 'See candidates' do
   include DfESignInHelpers
 
-  scenario 'Support agent visits the list of candidates' do
+  scenario 'Support agent visits the list of candidates and selects one who has never signed in' do
     given_i_am_a_support_user
     and_there_are_candidates_in_the_system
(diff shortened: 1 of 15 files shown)


spec/system/support_interface/see_candidates_spec.rb
```diff
@@ -3,16 +3,25 @@
...
     and_there_are_candidates_in_the_system
     and_i_visit_the_support_candidate_page
     then_i_should_see_the_candidates
 
-    when_i_click_on_a_candidate
+    when_i_click_on_a_candidate_with_no_applications
     then_i_see_the_candidate_details
   end
 
+  scenario 'Support agent selects a candidate with one application' do
```
tijmenb: The goal is to have a single scenario per file (we should still document why - speed & readability), can this be merged with the other scenario?

benilovj: Fair enough – fixed and force pushed.

PR #901, opened 2019-12-17
Hide HESA ITT data from API responses unless status is enrolled

mnacos:
## Context

The API docs state HESA ITT information "... will only be returned once the application status is enrolled."

[https://www.apply-for-teacher-training.education.gov.uk/api-docs/reference#hesaittdata-object](https://www.apply-for-teacher-training.education.gov.uk/api-docs/reference#hesaittdata-object)

This is not currently the case.

Hide the HESA ITT section from API application data responses unless status == 'enrolled'.

## Guidance to review

See changes to ```config/vendor-api-v1.yml```.

```hesa_itt_data``` is now an optional field, but if it supplied it must contain ```sex```, ```disability``` and ```ethnicity```, as before.

## Link to Trello card

[Implement HESA ITT data appearing in API responses only when enrolled](https://trello.com/c/ZDPdiiO3)


mnacos:
I've opened another (cleaner) PR for this.

PR #903, opened 2019-12-17
Hide HESA ITT data from API responses unless status is enrolled

mnacos:
## Context

The API docs state HESA ITT information "... will only be returned once the application status is enrolled."

[https://www.apply-for-teacher-training.education.gov.uk/api-docs/reference#hesaittdata-object](https://www.apply-for-teacher-training.education.gov.uk/api-docs/reference#hesaittdata-object)

This is not currently the case.

Hide the HESA ITT section from API application data responses unless status == 'enrolled'.

## Guidance to review

See changes to ```config/vendor-api-v1.yml```.

```hesa_itt_data``` is now an optional field, but if it supplied it must contain ```sex```, ```disability``` and ```ethnicity```, as before.

## Link to Trello card

[Implement HESA ITT data appearing in API responses only when enrolled](https://trello.com/c/ZDPdiiO3)

### DIFF
--- a/config/vendor-api-v1.yml
+++ b/config/vendor-api-v1.yml
@@ -331,7 +331,6 @@ components:
         - updated_at
         - reject_by_default_at
         - withdrawal
-        - hesa_itt_data
         - further_information
         - work_experience
       properties:
--- a/spec/presenters/vendor_api/single_application_presenter_spec.rb
+++ b/spec/presenters/vendor_api/single_application_presenter_spec.rb
@@ -28,6 +28,25 @@ RSpec.describe VendorApi::SingleApplicationPresenter do
     end
   end
 
+  describe 'attributes.hesa_itt_data' do
+    let(:application_choice) do
+      application_form = create(:completed_application_form, :with_completed_references)
+      create(:application_choice, status: 'awaiting_provider_decision', application_form: application_form)
+    end
+
+    it 'is hidden by default' do
+      response = VendorApi::SingleApplicationPresenter.new(application_choice).as_json
+      expect(response.dig(:attributes, :hesa_itt_data)).to be_nil
+    end
+
+    it 'becomes available once application status is \'enrolled\'' do
+      application_choice.update(status: 'enrolled')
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: config/vendor-api-v1.yml [hunk 1]
```diff
@@ -314,7 +314,7 @@ components:
           $ref: "#/components/schemas/ApplicationAttributes"
     ApplicationAttributes:
       type: object
-      additionalProperties: false
+      additionalProperties: true
```
tijmenb: I think this isn't necessary because `hesa_itt_data` is allowed because it's in the `properties`.

spec/presenters/vendor_api/single_application_presenter_spec.rb
```diff
@@ -28,6 +28,27 @@
     end
   end
 
+  describe 'attributes.hesa_itt_data' do
+    let(:application_form) do
+      create(:completed_application_form, :with_completed_references)
+    end
```
tijmenb: Minor: this can be inlined into the `application_choice` let, we don't need it elsewhere.

spec/presenters/vendor_api/single_application_presenter_spec.rb
```diff
@@ -28,6 +28,27 @@
...
+
+    it 'is hidden by default' do
+      response = VendorApi::SingleApplicationPresenter.new(application_choice).as_json
+      expect(response.dig(:attributes, :hesa_itt_data)).to be_nil
+    end
+
+    it 'becomes available once application status is \'enrolled\'' do
+      application_choice.update_attributes(status: 'enrolled')
+      response = VendorApi::SingleApplicationPresenter.new(application_choice).as_json
+      expect(response.dig(:attributes, :hesa_itt_data)).not_to be_nil
```
tijmenb: Perhaps we can add a check to see if the response conforms to the schema here, I think it wouldn't be tested otherwise.

```
expect(response.to_json).to be_valid_against_openapi_schema('Application')
```

PR #906, opened 2019-12-17
Upgrade to rubocop-govuk, add Rails linting

tijmenb:
## Context

govuk-lint has been deprecated, and replaced by rubocop-govuk. That also includes Rails cops, so this PR also enables those.

There are a ton of Rails violations in `.rubocop_todo.yml` which I don't want to fix now. 

## Guidance to review

Check if the code still makes sense.


.rubocop_todo.yml
```diff
@@ -0,0 +1,108 @@
...
+    - 'db/migrate/20191203201235_add_application_choices_declined_by_default.rb'
+    - 'db/migrate/20191204101615_add_address_to_site.rb'
+
+# Offense count: 2
+# Configuration parameters: Include.
+# Include: db/migrate/*.rb
+Rails/CreateTableWithTimestamps:
+  Exclude:
+    - 'db/migrate/20191022145559_install_audited.rb'
+    - 'db/migrate/20191212113810_create_provider_users.rb'
```
tijmenb: This caught an actual issue that we'll fix!

PR #908, opened 2019-12-17
Add Timestamps to ProviderUsers

duncanjbrown:
Add timestamps, because we missed them out the first time.

This will break if you have local providerusers, as the column is `null: false`.


db/migrate/20191217155606_add_timestamps_to_provider_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddTimestampsToProviderUsers < ActiveRecord::Migration[6.0]
+  def change
+    change_table :provider_users, &:timestamps
```
tijmenb: 🤦‍♂ Oh Rubocop

PR #910, opened 2019-12-17
Start renaming references to application references

tijmenb:
## Context

There’s a Rails method called `references` on associations:

https://apidock.com/rails/ActiveRecord/QueryMethods/references

Over the last months we’ve seen weird, unexplained, mysterious and peculiar behaviour where associations for references aren’t reloading. We’ve had to add a bunch of `reload` statements to the tests to make them pass.

It could very well be unrelated to the naming collisions - another culprit might be the default scope on references.

In any case, this renames the Reference model to ApplicationReference, which avoids the collision and make it consistent with the other models like ApplicationWorkExperience.

This commit does the least possible. It doesn’t rename the database table, and doesn’t rename the factory.

## Guidance to review

Read through it.

## Link to Trello card

https://trello.com/c/lQpwnQH2/372-find-out-why-we-need-to-reload-so-much-when-were-working-with-references

### DIFF
--- a/app/components/referees_review_component.html.erb
+++ b/app/components/referees_review_component.html.erb
@@ -1,4 +1,4 @@
-<% @application_form.references.includes(:application_form).each do |referee| %>
+<% @application_form.application_references.includes(:application_form).each do |referee| %>
   <%= render(SummaryCardComponent, rows: referee_rows(referee), editable: @editable) do %>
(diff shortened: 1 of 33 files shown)


tvararu:
Yes!

duncanjbrown:
What happens if you take out some `reload`s?

tijmenb:
@duncanjbrown I will try that in a separate PR!

PR #913, opened 2019-12-17
Fix intermittent spec failure

duncanjbrown:
## Context

We saw an intermittent spec failure at `spec/system/candidate_interface/candidate_edits_application_after_submission_spec.rb:6`. In this spec we used a factory which set `edit_by` on the `application_choice`. This value was ignored, though, and the view would calculate the editability of the application form depending on the original submission date.

In fixing the view to look at the stored value of `edit_by`, tidy up the responsibilities of `ApplicationDates` and `TimeLimitCalculator`. The former gets useful dates off an `application_form` record. The latter figures out significant dates according to business rules.

Before this change `ApplicationDates` was doing some work that properly belonged in `TimeLimitCalculator`. Things are now a bit neater.

## Guidance to review

Commit by commit

### DIFF
--- a/app/components/application_complete_content_component.rb
+++ b/app/components/application_complete_content_component.rb
@@ -28,6 +28,10 @@ class ApplicationCompleteContentComponent < ActionView::Component::Base
     @dates.decline_by_default_at.strftime('%-e %B %Y')
   end
 
+  def edit_by_date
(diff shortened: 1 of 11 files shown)


duncanjbrown:
Cucumber is failing because we're using `TimeLimitCalculator` more than before. Fix coming

PR #914, opened 2019-12-17
Bypass DfE Signin when running in `development` hosting environment

tvararu:
## Context

This allows us to bypass DfE Signin even when the app is set to Rails.env.production? to debug other things.

Change the call from`Rails.env.development?` to `!HostingEnrivonment.production?`.

## Guidance to review

Should be fine as we initialize `HOSTING_ENVIRONMENT` this in `.env.development` locally.

## Things to check

### DIFF
--- a/app/lib/dfe_sign_in.rb
+++ b/app/lib/dfe_sign_in.rb
@@ -9,6 +9,6 @@ module DfESignIn
   end
 
   def self.bypass?
-    Rails.env.development? && ENV['BYPASS_DFE_SIGN_IN'] == 'true'
+    !HostingEnvironment.production? && ENV['BYPASS_DFE_SIGN_IN'] == 'true'
   end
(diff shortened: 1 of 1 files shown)


app/lib/dfe_sign_in.rb
```diff
@@ -9,6 +9,6 @@ def self.parse_auth_hash(openid_auth_hash)
   end
 
   def self.bypass?
-    Rails.env.development? && ENV['BYPASS_DFE_SIGN_IN'] == 'true'
+    ENV['HOSTING_ENVIRONMENT_NAME'] == 'development' && ENV['BYPASS_DFE_SIGN_IN'] == 'true'
   end
```
duncanjbrown: I think there's a method on `HostingEnvironment` which reads `ENV` for us

PR #920, opened 2019-12-18
Add first steps to allow referees to provide feedback in the app

alexdesi:
## Context

This PR allows a referee to leave a reference to the candidate (1st part)
- send a new email with a tokenized URL
- check if the referee is allowed to see the page

## Link to Trello card

https://trello.com/c/B4tokVQU/§


app/views/referee_mailer/reference_request_by_google_form_email.text.erb
```diff
@@ -0,0 +1,21 @@
+Dear <%= @reference.name %>,
+
+# Give a reference for <%= @candidate_name %>
+
+<%= @candidate_name %> put us in touch with you to get a reference for their teacher training application. They applied to:
+
+<% @application_form.application_choices.each do |application_choice| %>
+* <%= application_choice.provider.name %> - <%= application_choice.course.name %>
```
benilovj: Do we want to have a method on `ApplicationChoice` that returns a human-readable ASCII description of the choice? That way this description can be evolved in one place... /cc @fofr

app/controllers/referee_interface/reference_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+      RequestLocals.store[:identity] = { reference_id: reference.id }
+      Raven.user_context(reference_id: reference.id)
+    end
+
+    def reference
+      @reference ||= ApplicationReference.find_by_unhashed_token(params[:token])
+    end
+
+    def check_referee_has_valid_token
+      render_404 unless reference
```
fofr: 404 or 401?

app/views/referee_mailer/reference_request_by_google_form_email.text.erb
tijmenb: Let's pick that up in a separate PR, the choice description is not consistent across the emails anyway.

app/controllers/referee_interface/reference_controller.rb
tijmenb: 404

PR #922, opened 2019-12-18
Enable Heroku review apps

tvararu:
## Context

Previously: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/915

This enables [Heroku PR review apps](https://devcenter.heroku.com/articles/github-integration-review-apps). Whenever someone opens a PR, a new Heroku application is spun up with a fresh database and workers. We can use this to review and test features!

👇 Check out the review app below this PR description! 👇 

The review application runs in `Rails.env.production?`, but is meant to be local dev-y feeling, so DFE Signin is bypassed and the database is seeded with dev data.

You can access the support section by using the default local dev support user: `dev-support` / `support@example.com`.

Candidate emails work if you have a DfE email and you're signed up with Notify.

Add an `app.json` which is used by Heroku to provision the right addons, dynos, and run the setup task for migrations. Populate `AUTHORISED_HOSTS` using `HEROKU_APP_NAME`.

- [ ] Add some docs to the README about this

## Guidance to review

<details>
<summary>I thought maybe we need to worry about our environment variables being stolen.</summary>
<p>There is one issue with this setup, and it relates to the GOVUK Notify key. I've set this to a `team-and-whitelist` type key, but because review apps get deployed for every pull request, a bad actor could steal this (alongside the other environment variables) by sending us a PR with something that scrapes our vars. The other environment variables are not sensitive.</p>

<p>We would know if this happens, as opening a PR is a highly visible action. The key is also restricted such that it can't be used for too nefarious a purpose. So I don't think it's a high risk, but it would be nice to somehow restrict these apps only to GitHub organisation members.</p>
</details>

But nevermind:

> With automatic creation, Heroku creates a review app as soon as a pull request is opened on your app’s connected GitHub repo. For security and billing reasons, **Heroku does not automatically create review apps for pull requests to public repos that are sent from forks**. You can still create review apps for these pull requests manually.

https://devcenter.heroku.com/articles/github-integration-review-apps

## Things to check

### DIFF
--- a/app.json
+++ b/app.json
@@ -3,6 +3,7 @@
   "scripts": {
     "postdeploy": "bundle exec rake db:schema:load && bundle exec rake setup_local_dev_data"
   },
+  "addons": ["heroku-postgresql:hobby-dev", "heroku-redis:hobby-dev"],
   "buildpacks": [
     {
       "url": "heroku/ruby"
--- a/app.json
+++ b/app.json
@@ -3,6 +3,20 @@
   "scripts": {
     "postdeploy": "bundle exec rake db:schema:load && bundle exec rake setup_local_dev_data"
   },
+  "formation": {
+    "web": {
+      "quantity": 1,
+      "size": "hobby"
+    },
+    "clock": {
+      "quantity": 1,
+      "size": "hobby"
+    },
+    "worker": {
+      "quantity": 1,
+      "size": "hobby"
+    }
+  },
   "addons": ["heroku-postgresql:hobby-dev", "heroku-redis:hobby-dev"],
   "buildpacks": [
     {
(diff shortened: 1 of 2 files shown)


app.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "apply-for-teacher-training",
+  "scripts": {
+    "postdeploy": "heroku config:set AUTHORISED_HOSTS=${HEROKU_APP_NAME}.herokuapp.com --app ${HEROKU_APP_NAME} && bundle exec rake db:migrate && bundle exec rake setup_local_dev_data"
```
duncanjbrown: Maybe `schema:load` would be more intentional?

fofr:
This will make researching design changes on branches a lot easier.

PR #931, opened 2019-12-18
Update resubmit button on application edit page

RobertM905:
## Context

When candidates go to edit their submitted application the submit button suggests that the application will be saved and submitted again then, this is not the case because editing the application after submission edits the application in place.
This PR aims to address this.

This PR includes:
- Update the submit button on application edit page to return to application complete page when the application is amendable and being edited.
- Update system specs to include test to assert that the submit button has been changed for submitted applications.

## Guidance to review

- Would like guidance on the copy for the new `go back to dashboard button title and text`.
- Check specs to ensure that I haven't missed any edge cases when editing an application.

## Screenshots
Before
Screenshot 2019-12-18 at 16 32 21

After
Screenshot 2019-12-18 at 16 32 07

## Link to Trello card

https://trello.com/c/IP7g0jBr/583-amending-an-application-after-submission

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/application_form_controller.rb
+++ b/app/controllers/candidate_interface/application_form_controller.rb
@@ -3,12 +3,14 @@ module CandidateInterface
     before_action :redirect_to_dashboard_if_not_amendable, only: %i[show review]
     before_action :redirect_to_application_form_unless_submitted, only: %i[review_submitted complete submit_success]
 
+
     def show
       @application_form_presenter = CandidateInterface::ApplicationFormPresenter.new(current_application)
       @application_form = current_application
     end
 
     def review
+      redirect_to candidate_interface_application_complete_path if current_application.submitted?
       @application_form = current_application
     end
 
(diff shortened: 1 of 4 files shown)


app/controllers/candidate_interface/application_form_controller.rb
```diff
@@ -3,12 +3,14 @@ class ApplicationFormController < CandidateInterfaceController
...
     before_action :redirect_to_application_form_unless_submitted, only: %i[review_submitted complete submit_success]
 
+
     def show
       @application_form_presenter = CandidateInterface::ApplicationFormPresenter.new(current_application)
       @application_form = current_application
     end
 
     def review
+      refirect_to_application_dashboard_if_amendable
```
C-gyorfi: I don't think you need this line since you included in ` before_action :redirect_to_dashboard_if_not_amendable, only: %i[show review]`

RobertM905: Moved this into an inline to avoid confusion. 👍

PR #933, opened 2019-12-18
Allow bypassing DfE Signin unless we're on the Production hosting env

tvararu:
## Context

Like https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/914, but that one was failing for devs locally which have `BYPASS_DFE_SIGN_IN` set to `true` in `.env`. We can set it to `false` explicitly by setting it in `.env.test`.


tvararu:
The tests still don't seem to pass locally. ?_?

PR #934, opened 2019-12-18
Add forms to allow referee to give feedback to the candidate

alexdesi:
## Context
This PR adds the following pages (in order):

1.  `/reference/feedback`
2. `/reference/confirmation`
3. `reference/finish`

All these pages are protected by token,
the token is passed using a query param `?token=XXXXXXX`

## Guidance to review
see screeshots

Note there is a separate PR for referee decline reference request and contacts for research purposes https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/905

## Link to Trello card

https://trello.com/c/RrlqYDMz/585-referee-can-provide-a-reference-via-dfe-apply-service-%F0%9F%8F%88


fofr:
@alexdesi Can you share some screenshots?

app/views/referee_interface/reference/confirmation.html.erb
```diff
@@ -0,0 +1,8 @@
+<main class="govuk-main-wrapper govuk-main-wrapper--auto-spacing" id="main-content" role="main">
```
fofr: This page needs a `<title>` set.

app/views/referee_interface/reference/feedback.html.erb
```diff
@@ -1,5 +1,42 @@
 <% content_for :title, t('page_titles.application') %>
```
fofr: I think this needs a unique title too.

app/views/referee_interface/reference/confirmation.html.erb
```diff
@@ -0,0 +1,8 @@
...
+  <div class="govuk-panel govuk-panel--confirmation govuk-!-margin-bottom-8">
+    <h1 class="govuk-panel__title">
+      Your reference for <%= @reference.application_form.full_name %> has been&nbsp;submitted
+    </h1>
+  </div>
+  <%= govuk_button_link_to 'Finish', referee_interface_reference_finish_path(token: @token_param) %>
```
fofr: Looks like this is missing the last form/question (https://trello-attachments.s3.amazonaws.com/5de11ac9d5eb9c63c567e3c1/760x383/af823baf411efa48bc0e55394e5e8cf5/Screenshot_2019-12-11_at_10.49.44.png) – is that being worked on separately?

C-gyorfi: Yes there is another PR for this https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/905

alexdesi: that's true @fofr, we know, we just split it into a different trello (to avoid huge PR):
https://trello.com/c/B4tokVQU/667-referee-form-can-we-contact-you-for-research-purposes

alexdesi:
I've realised that the list of the courses is missing from the feedback page.
I'm adding it!

app/controllers/referee_interface/reference_controller.rb
```diff
@@ -3,19 +3,42 @@ class ReferenceController < ActionController::Base
     include LogQueryParams
     before_action :add_identity_to_log
     before_action :check_referee_has_valid_token
+    before_action :set_token_param
```
tvararu: Is this necessary? We call `params[:token]` in other parts of the controller.

We could do:

```ruby
def token_param
  params[:token]
end
```

And use that?

alexdesi: I think it is necessary since we use it in the views too:
For example:

```
  <%= govuk_button_link_to 'Finish', referee_interface_reference_finish_path(token: @token_param) %>

```

alexdesi:
Screenshot 2019-12-20 at 12 17 57
Screenshot 2019-12-20 at 12 17 11
Screenshot 2019-12-20 at 12 16 53
Screenshot 2019-12-20 at 12 16 06

alexdesi:
@fofr I've just added few screenshot.
PS. a new PR that complete the referee flow is coming soon 😏

fofr:
A couple of follow up fixes in #961

PR #938, opened 2019-12-19
Fix interview preferences pre-population of 'None' in text area

chubberlisk:
## Context

When product reviewing the changing the interview preferences section to a question, Matteo found that  when a candidate:

1. selects `No` for any interview preferences
2. clicks `Save and continue`
3. clicks `Change`
4. selects `Yes`

the text area is pre-populated with `None`.

This PR updates the `CandidateInterface::InterviewPreferencesForm` to save an empty string if `No` is selected. Then in order to display `None` when a candidate reviews their interview preferences, `InterviewPreferencesReviewComponent` is updated to display `None` when the value of `interview_preferences` is an empty string.

## Before screenshots

image

image

## After screenshots

image

image

## Guidance to review

Does this makes sense? Is there a better way to deal with this?

## Link to Trello card

https://trello.com/c/htjXI6yz/465-update-interview-preferences-to-a-question

## Things to check

### DIFF
--- a/app/components/interview_preferences_review_component.rb
+++ b/app/components/interview_preferences_review_component.rb
@@ -23,9 +23,11 @@ private
   attr_reader :application_form
 
   def interview_preferences_form_row
+    preferences = @interview_preferences_form.interview_preferences.presence || t('application_form.personal_statement.interview_preferences.no_value')
+
     {
       key: t('application_form.personal_statement.interview_preferences.key'),
-      value: @interview_preferences_form.interview_preferences,
+      value: preferences,
       action: t('application_form.personal_statement.interview_preferences.change_action'),
       change_path: Rails.application.routes.url_helpers.candidate_interface_interview_preferences_edit_path,
     }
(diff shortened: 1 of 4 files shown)


app/components/interview_preferences_review_component.rb
```diff
@@ -23,9 +23,11 @@ def show_missing_banner?
   attr_reader :application_form
 
   def interview_preferences_form_row
+    preferences = @interview_preferences_form.interview_preferences.presence || t('application_form.personal_statement.interview_preferences.no_value')
```
tvararu: What's `.presence` here? 🤔

chubberlisk: Made the change as rubocop told me off for doing:

```ruby
preferences = @interview_preferences_form.interview_preferences.blank? ? t('application_form.personal_statement.interview_preferences.no_value') : @interview_preferences_form.interview_preferences
```

And suggested to do that instead.

So:

```ruby
''.presence
=> nil
'ship it'.presence
=> 'ship it'
''.presence || 'none'
=> 'none'
```

tvararu: Ok 👍

PR #948, opened 2019-12-19
Record `accepted_at` and `declined_at`

duncanjbrown:
## Context

While working on the date widget at the top of applications in the Provider Interface (#945), we noticed that 

- `declined_at` was nil, even though `DeclineOffer` had been called on an application
- we weren't storing `accepted_at` whereas all other dates are

- Add `accepted_at` to `ApplicationChoices`
- Set it in `AcceptOffer`
- Set `declined_at` in `DeclineOffer`
- Use `DeclineOffer` instead of manually calling `ApplicationStateChange`

## Link to Trello card

https://trello.com/c/Php2IR0H/1409-solve-incongruities-between-provider-and-candidate-interfaces
## Things to check

- [ ] This code doesn't rely on migrations in the same Pull Request — ⚠️  **IT DOES** ⚠️  but there are no offers in the system rn and no applications in an offerable state, so there is a very low risk of an offer being accepted, which is the code path which depends on the migration


app/services/accept_offer.rb
```diff
@@ -6,9 +6,10 @@ def initialize(application_choice:)
   def save!
     ActiveRecord::Base.transaction do
       ApplicationStateChange.new(@application_choice).accept!
+      @application_choice.update(accepted_at: Time.zone.now)
 
       other_application_choices_with_offers.each do |other_application_choice|
-        ApplicationStateChange.new(other_application_choice).decline!
+        DeclineOffer.new(application_choice: other_application_choice).save!
```
mnacos: Much better! This, however, may trigger multiple Slack notifications for the automatic declines. Are we ok with that?

duncanjbrown: I decided we were, but open to being convinced the other way. We could always flag it if it became annoying

duncanjbrown: Oh wait we have the block which silences Slack... let me find it

duncanjbrown:
@mnacos updated to not send slack notifications on the declines

PR #950, opened 2019-12-19
Use the WithdrawApplication service to withdraw test applications

tijmenb:
See commits for details!

### DIFF
--- a/app/workers/generate_test_applications.rb
+++ b/app/workers/generate_test_applications.rb
@@ -109,8 +109,7 @@ private
       end
 
       if application_index == 11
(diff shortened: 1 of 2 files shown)


lib/tasks/reset.rake
```diff
@@ -0,0 +1,2 @@
+desc 'Recreate the database and set up all of the test data'
+task reset: %i[db:reset db:migrate setup_local_dev_data generate_test_applications]
```
duncanjbrown: `local_dev_data` calls `GenerateTestApplications` already

tijmenb: It does not!

tijmenb: Narrator: _it did_

PR #954, opened 2019-12-19
Allow bypassing DfE Signin when running in production mode

tvararu:
But still check that the hosting environment is not one of our live ones.

Was previously https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/914 which caused issues with running tests locally.

This is needed to allow bypassing DSI on Heroku review apps, which do run in `Rails.env.production?`, but have autogenerated hostnames and as such can not be configured to work with DSI because they expect a static and whitelisted callback URL.


tvararu:
@tijmenb good idea! Closing and will reopen with that idea instead. 👍

PR #956, opened 2019-12-19
Add review environment for Heroku and bypass DSI

tvararu:
## Context

Previously: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/954.

This allows us to bypass DSI on the Heroku review apps.

I chose to name it `review` instead of `heroku` so that it can be used even if we move away from Heroku.

Add an environment and update the DSI bypass conditional.

## Guidance to review

:ship:

Screenshot 2019-12-19 at 16 43 06

## Things to check

### DIFF
--- a/app/lib/dfe_sign_in.rb
+++ b/app/lib/dfe_sign_in.rb
@@ -1,5 +1,5 @@
 module DfESignIn
   def self.bypass?
-    Rails.env.development? && ENV['BYPASS_DFE_SIGN_IN'] == 'true'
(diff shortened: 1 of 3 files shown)


app/lib/dfe_sign_in.rb
```diff
@@ -1,5 +1,5 @@
 module DfESignIn
   def self.bypass?
-    Rails.env.development? && ENV['BYPASS_DFE_SIGN_IN'] == 'true'
+    (HostingEnvironment.review? || Rails.env.development?) && ENV['BYPASS_DFE_SIGN_IN'] == 'true'
```
duncanjbrown: ❤️

PR #957, opened 2019-12-19
Pre-approve provider emails before first sign-in

mnacos:
### Context

It is possible for any user with a DfE login to get an authenticated sessions within our service. Whether they can use the provider interface depends on the existence of a database-backed ProviderUser with the same `uid`.

`uid` values, however, are not known in advance, when a Provider is onboarded. The current process consists of emailing providers asking for their users to get DfE sign-in accounts, log into Apply, see the authorisation error, at which point we can grab their `uid` and add them to the system.

image
By creating provisional user records with no `uid` but with an email address, if such email addresses are known in advance, we could pre-approve these users so that when they get their DfE sign-in accounts and log in they can start using the provider interface straight away.

We have decided to add these pre-approved/provisional users as `provider_users` with a blank `uid` instead of creating a different table. This will allow us to re-use any UI components developed to support the many-to-many relationship between users and providers. There may be a checkbox or notice attached about leaving `uid` blank within the form.

This PR makes `uid` an optional field for `ProviderUser` by:

- removing the ```NOT NULL``` constraint for `dfe_sign_in_uid` from `provider_users`
- removing the unique index from `dfe_sign_in_uid on` `provider_users` (would only allow one NULL)
- removing `validates_presence_of` for `:uid` in ProviderUserForm
- allowing `nil` as a value for `uid` in ProviderUserForm while preventing the empty string

Furthermore, a unique index is added to `email_address` on `provider_users`. ProviderUser's _#load_from_session_ will now match the DfE sign-in session on `uid` as before, but if that fails it will attempt to match on email and set the `nil` UID if the email matches.

##### Before:

image

##### After:

image

### Guidance to review

Use http://localhost:3000/support/users/provider/new to create a new `ProviderUser` leaving the `uid` blank but adding a previously unused email address. Log out, then sign in to the provider interface using this new email address. You should NOT see:

image

... but be able to access the provider_interface straight away.

### Link to Trello card

[When users arrive from DSI, look them up by email as well as UID](https://trello.com/c/XO8ID26w)


tvararu:
This sounds like a good idea! 👍

db/schema.rb
```diff
@@ -184,10 +184,11 @@
 
   create_table "provider_users", force: :cascade do |t|
     t.string "email_address", null: false
-    t.string "dfe_sign_in_uid", null: false
+    t.string "dfe_sign_in_uid"
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
-    t.index ["dfe_sign_in_uid"], name: "index_provider_users_on_dfe_sign_in_uid", unique: true
+    t.index ["dfe_sign_in_uid"], name: "index_provider_users_on_dfe_sign_in_uid"
```
duncanjbrown: Sad to lose the index. Ought we to constrain DSI UIDs to be unique in the model now?

app/models/provider_user.rb
```diff
@@ -9,8 +9,15 @@ def self.load_from_session(session)
...
     return unless dfe_sign_in_user
 
-    ProviderUser.find_by(
-      dfe_sign_in_uid: dfe_sign_in_user.dfe_sign_in_uid,
-    )
+    approved_user = ProviderUser.find_by dfe_sign_in_uid: dfe_sign_in_user.dfe_sign_in_uid
+    approved_user || onboard_provider_user!(dfe_sign_in_user)
+  end
+
+  def self.onboard_provider_user!(dsi_user)
```
duncanjbrown: We're under `ProviderUser` — maybe just `onboard!`?

app/models/provider_user.rb
```diff
@@ -9,8 +9,15 @@ def self.load_from_session(session)
...
+    provider_user = ProviderUser.find_by email_address: dsi_user.email_address
+    if provider_user && provider_user.dfe_sign_in_uid.nil?
+      provider_user.update(dfe_sign_in_uid: dsi_user.dfe_sign_in_uid)
```
duncanjbrown: `update!`... just in case?

app/controllers/support_interface/provider_users_controller.rb
```diff
@@ -10,6 +10,7 @@ def new
 
     def create
       @form = ProviderUserForm.new(provider_user_params.merge(available_providers: Provider.all))
+      @form.dfe_sign_in_uid = nil if @form.dfe_sign_in_uid.blank?
```
duncanjbrown: It might be safer to change the getter on `ProviderUserForm` to return nil if this is blank

app/views/support_interface/provider_users/new.html.erb
```diff
@@ -23,6 +23,9 @@
 
       <%= f.govuk_text_field :email_address, label: { text: 'Email address' } %>
       <%= f.govuk_text_field :dfe_sign_in_uid, label: { text: 'DfE Sign-in UID' } %>
```
duncanjbrown: The `hint_text:` option might be a good place for the new copy?

https://govuk-form-builder.netlify.com/form-elements/text-field/

app/controllers/support_interface/provider_users_controller.rb
mnacos: I've added validation to ProviderUser to allow nil but ensure the uid is not set to whitespace. L13 above now merely makes sure the resulting model is valid.

spec/system/provider_interface/first_login_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'rails_helper'
+
+RSpec.feature 'See applications' do
+  include CourseOptionHelpers
+  include DfESignInHelpers
+
+  let(:pre_approved_provider_user) do
+    provider = create(:provider, code: 'ABC')
+    create(:provider_user, providers: [provider], dfe_sign_in_uid: nil)
+  end
```
tijmenb: FYI: prefer to use instance variables in the integration tests, so we don't clutter the tests with `let` and `before` blocks.

PR #958, opened 2019-12-19
Move candidate-specific components into namespace

tijmenb:
## Context

We currently have a large number of candidate-specific components in the main namespace, and a few in specific namespaces.

This cleanly separates components that can be reused across namespaces from the ones that are specific.

Not all “review” components could be moved into the Candidate Interface, because they’re used in the Provider interface too - the VolunteeringReviewComponent and WorkHistoryReviewComponent. These should be refactored to be either candidate-specific or more generic.

## Guidance to review

Open up the folder on your machine, and see if it looks more logical.

## Link to Trello card

https://trello.com/c/Php2IR0H/1409-solve-incongruities-between-provider-and-candidate-interfaces


app/components/work_history_review_component.rb
```diff
@@ -1,3 +1,4 @@
+# Used in Candidate and Provider interface
```
tvararu:
```suggestion
# Used in Candidate and Provider interface
```

Rubocop 🤦‍♂

PR #960, opened 2019-12-20
Add e-mail format validations

C-gyorfi:
## Context

At the moment we only specify that an email address must have @. This means a lot of other known errors are able to pass through undetected. This is an issue because a user won't realise they have submitted an invalid email address.

This PR updates the email format validation using a regular expression that allows emails that are allowed by Notify https://github.com/alphagov/notifications-utils/blob/31391091d2351e3e6bd306585e67007bcdb31911/tests/test_recipient_validation.py#L104-L152
- Update the maximum characters from 250 to 100. The max length in the [API docs](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/config/vendor-api-v1.yml#L539) is 100 chars, and iirc this is a hard constraint from vendors

- Update email validator regex according to reproduce [validation of Notify](https://github.com/alphagov/notifications-utils/blob/31391091d2351e3e6bd306585e67007bcdb31911/tests/test_recipient_validation.py#L104-L152)
## Guidance to review

I would suggest having a look at the unit tests and try to come up with other valid/invalid e-mails to check the right behaviour. Please note the validator allows `two-dots..in-local@domain.com`, please advise if this is acceptable or any further changes. 

## Link to Trello card

https://trello.com/c/4lfXwREg/655-revisit-email-validation-%F0%9F%8F%88

## Things to check

### DIFF
--- a/app/models/candidate.rb
+++ b/app/models/candidate.rb
@@ -6,7 +6,7 @@ class Candidate < ApplicationRecord
   before_validation :downcase_email
   validates :email_address, presence: true,
                             uniqueness: { case_sensitive: false },
-                            length: { maximum: 250 },
+                            length: { maximum: 100 },
                             email_address: true
 
   has_many :application_forms
--- a/app/validators/email_address_validator.rb
+++ b/app/validators/email_address_validator.rb
@@ -1,6 +1,15 @@
 class EmailAddressValidator < ActiveModel::EachValidator
+  ALPHANUMERIC = 'a-zA-Z0-9àèìòùÀÈÌÒÙáéíóúýÁÉÍÓÚÝâêîôûÂÊÎÔÛãñõÃÑÕäëïöüÿÄËÏÖÜŸçÇßØøÅåÆæœ' + # Number and letters including accented characters
+    '\u3000-\u303F\u3040-\u309F\u30A0-\u30FF\uFF00-\uFFEF\u4E00-\u9FAF\u2605-\u2606\u2190-\u2195\u203B'.freeze # Chinese/Japanese/Korean characters
+  EMAIL_REGEX = %r{\A[#{ALPHANUMERIC}.!\#$%&'*+\/=?^_`{|}~-] # Local part
+                  +@[#{ALPHANUMERIC}](?:[#{ALPHANUMERIC}-] # Domain name
+                  {0,61}[#{ALPHANUMERIC}])?(?:\. # Allow periods in domain name
+                  [#{ALPHANUMERIC}](?:[#{ALPHANUMERIC}-]{0,61}[#{ALPHANUMERIC}])?)*\.
+                  [#{ALPHANUMERIC}](?:[#{ALPHANUMERIC}-]{0,61} # # End of domain
+                  [#{ALPHANUMERIC}])\z}x.freeze
+
   def validate_each(record, attribute, value)
-    if value.blank? || !value.match?(/@/)
+    if value.blank? || !value.match?(EMAIL_REGEX)
       record.errors[attribute] << I18n.t('activerecord.errors.models.candidate.attributes.email_address.invalid')
     end
   end
--- a/spec/validators/email_address_validator_spec.rb
+++ b/spec/validators/email_address_validator_spec.rb
@@ -9,19 +9,82 @@ RSpec.describe EmailAddressValidator do
     end
   end
 
+  context 'when an email address has a valid format' do
+    let(:model) { Validatable.new }
+    let(:valid_emails) do
+      [
+        'email@domain.com',
+        'email@domain.COM',
+        'firstname.lastname@domain.com',
+        'firstname.o\'lastname@domain.com',
+        'email@subdomain.domain.com',
+        'firstname+lastname@domain.com',
+        '1234567890@domain.com',
+        'email@domain-one.com',
+        '_______@domain.com',
+        'email@domain.name',
+        'email@domain.superlongtld',
+        'email@domain.co.jp',
+        'firstname-lastname@domain.com',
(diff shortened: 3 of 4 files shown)


app/models/candidate.rb
```diff
@@ -6,9 +6,9 @@ class Candidate < ApplicationRecord
   before_validation :downcase_email
   validates :email_address, presence: true,
                             uniqueness: { case_sensitive: false },
-                            length: { maximum: 250 },
+                            length: { maximum: 256 },
```
fofr: Will this need to be represented in the API docs too?

C-gyorfi: I cannot find anything about this in the API docs, so I'm not sure. Do you have an idea? @fofr

duncanjbrown: The max length in the API docs is 100 chars, and iirc this is a hard constraint from vendors

C-gyorfi: Oh, then, I think we should change this to `100`, thanks @duncanjbrown

app/validators/email_address_validator.rb
```diff
@@ -1,6 +1,9 @@
 class EmailAddressValidator < ActiveModel::EachValidator
+  ACCENTED = 'àèìòùÀÈÌÒÙáéíóúýÁÉÍÓÚÝâêîôûÂÊÎÔÛãñõÃÑÕäëïöüÿÄËÏÖÜŸçÇßØøÅåÆæœ'.freeze
+  EMAIL_REGEX = /\A[a-zA-Z#{ACCENTED}0-9.!\#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z#{ACCENTED}0-9](?:[a-zA-Z#{ACCENTED}0-9-]{0,61}[a-zA-Z#{ACCENTED}0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z#{ACCENTED}0-9-]{0,61}[a-zA-Z#{ACCENTED}0-9])?)*\.[a-zA-Z0-9](?:[a-zA-Z#{ACCENTED}0-9-]{0,61}[a-zA-Z#{ACCENTED}0-9])\z/i.freeze
```
alexdesi: this is unreadable 😄 
please, could you make this over multiple lines with some short comments?
see this as example:
https://stackoverflow.com/questions/3762183/split-ruby-regex-over-multiple-lines

spec/validators/email_address_validator_spec.rb
```diff
@@ -9,19 +9,83 @@
...
+      'brackets(in)local@domain.com',
+    ]
     end
 
     it 'returns invalid' do
-      expect(model).not_to be_valid
+      invalid_emails.each do |email|
+        model.email_address = email
+        model.validate(:no_context)
+        expect(model).not_to be_valid
```
alexdesi: This test sounds generic, for example, the email could be invalid because too long, not because the format is wrong.

C-gyorfi: I leave it as it is since it is in a context block `when an email address is in an invalid format` let me know if you are happy with this.

C-gyorfi: This validator does not check the length, because the model has present and length validation

app/models/candidate.rb
C-gyorfi: Resolved in ff060ae

app/validators/email_address_validator.rb
C-gyorfi: I've added some comments and added support for foreign/accented characters 3e090a8

app/models/candidate.rb
```diff
@@ -6,7 +6,7 @@ class Candidate < ApplicationRecord
   before_validation :downcase_email
   validates :email_address, presence: true,
                             uniqueness: { case_sensitive: false },
-                            length: { maximum: 250 },
+                            length: { maximum: 100 },
```
duncanjbrown: :+1:

spec/validators/email_address_validator_spec.rb
alexdesi: 👍

app/validators/email_address_validator.rb
```diff
@@ -1,6 +1,15 @@
 class EmailAddressValidator < ActiveModel::EachValidator
+  ALPHANUMERIC = 'a-zA-Z0-9àèìòùÀÈÌÒÙáéíóúýÁÉÍÓÚÝâêîôûÂÊÎÔÛãñõÃÑÕäëïöüÿÄËÏÖÜŸçÇßØøÅåÆæœ' + # Number and letters including accented characters
+    '\u3000-\u303F\u3040-\u309F\u30A0-\u30FF\uFF00-\uFFEF\u4E00-\u9FAF\u2605-\u2606\u2190-\u2195\u203B'.freeze # Chinese/Japanese/Korean characters
+  EMAIL_REGEX = %r{\A[#{ALPHANUMERIC}.!\#$%&'*+\/=?^_`{|}~-] # Local part
+                  +@[#{ALPHANUMERIC}](?:[#{ALPHANUMERIC}-] # Domain name
+                  {0,61}[#{ALPHANUMERIC}])?(?:\. # Allow periods in domain name
+                  [#{ALPHANUMERIC}](?:[#{ALPHANUMERIC}-]{0,61}[#{ALPHANUMERIC}])?)*\.
+                  [#{ALPHANUMERIC}](?:[#{ALPHANUMERIC}-]{0,61} # # End of domain
+                  [#{ALPHANUMERIC}])\z}x.freeze
```
alexdesi: 👍 nice one, now it's simpler to read!

tvararu:
👍 Excellent work!

PR #962, opened 2019-12-20
Fix incorrect path when signing in as candidate from Support

chubberlisk:
## Context

When signing in as a candidate that has an application that has passed the amend period (no longer editable) using the Support Interface, the support user is taken to the 'Edit your application' page instead they should be taken to the 'Application dashboard' page.

This PR fixes this by wrapping the path around the `edit_application` feature flag and redirecting to the dashboard if it's not submitted or it's amendable.

However, an addition change was required in `GenerateTestApplications`. From the commit description:

>After an application has been submitted, a candidate has 5 working
days to edit their application. This is calculated at submission and
used to work out whether an application is in amend period.

>However, when we generate test applications we are able to skip this
period to send applications to a provider as the service we use
does not check the `edit_by` date.

>Therefore, it is necessary to update the `edit_by` to now, to
make it come up as no longer amendable.

## Link to Trello card

https://trello.com/c/E7Rt3O3q/682-signing-into-a-completed-application-directs-you-to-edit-application-page

## Things to check

### DIFF
--- a/app/workers/generate_test_applications.rb
+++ b/app/workers/generate_test_applications.rb
@@ -71,6 +71,9 @@ private
         # This is only supposed to happen after 7 days, but SendApplicationToProvider
         # doesn't check the `edit_by` date of the ApplicationChoice
         SendApplicationToProvider.new(application_choice: application_choice).call
+
+        # This is so the application is no longer amendable
+        application_choice.update(edit_by: Time.zone.now)
       end
 
       return if application_index == 4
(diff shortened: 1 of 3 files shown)


app/workers/generate_test_applications.rb
```diff
@@ -71,6 +71,9 @@ def create_an_application(application_index)
         # This is only supposed to happen after 7 days, but SendApplicationToProvider
         # doesn't check the `edit_by` date of the ApplicationChoice
         SendApplicationToProvider.new(application_choice: application_choice).call
+
+        # This is so the application is no longer amendable
+        application_choice.update(edit_by: Time.zone.now)
       end
```
tvararu: 👍

PR #963, opened 2019-12-20
Success message gets focused and document title prefix

adamsilver:
## Context

Addresses ticket 683, the accessibility fail from DAC.

Before (not focused):
image

After (focused):
image

### DIFF
--- a/app/components/flash_message_component.html.erb
+++ b/app/components/flash_message_component.html.erb
@@ -16,7 +16,7 @@
       </div>
     </div>
   <% elsif flash[:success] %>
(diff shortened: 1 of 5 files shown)


app/components/flash_message_component.html.erb
```diff
@@ -16,7 +16,7 @@
       </div>
     </div>
   <% elsif flash[:success] %>
-    <div class="app-banner app-banner--success" aria-labelledby="success-message">
+    <div class="app-banner app-banner--success" aria-labelledby="success-message" data-module="govuk-error-summary" tabindex="-1">
```
fofr: Does this also need `role=alert`?

PR #970, opened 2019-12-23
Make sure navigation works as expected on small/mobile screens

adamsilver:
## Context

The navigation didn't work on mobile across support, candidate and provider and failed accessibility.

Before: image

After: image

## Guidance to review

It would be good if we could turn the header and navigation into a component. For this commit I've had to scatter the same button across a number of templates.

## Link to Trello card

https://trello.com/c/Mq9xM59d/686-cant-access-navigation-on-mobile

### DIFF
--- a/app/views/layouts/_support_header.html.erb
+++ b/app/views/layouts/_support_header.html.erb
@@ -1,3 +1,4 @@
+<button type="button" class="govuk-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
 <ul id="navigation" class="govuk-header__navigation " aria-label="Top Level Navigation">
   <% if current_support_user %>
     <%= nav_link 'Candidates', support_interface_candidates_path, active: %w[candidates import_references] %>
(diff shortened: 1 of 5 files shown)


duncanjbrown:
Rather than adding it to each header, could this work outside the `case` statement in `application.html.erb`?

dankmitchell:
Agree with @duncanjbrown

adamsilver:
@duncanjbrown @dankmitchell I would have liked to have been more DRY but the button should only be there if there are navigation items.

See the original component in GOV.UK Frontend here:
https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/header/template.njk#L62

That's partly why it would be great if this was abstracted out into a component with the same ‘dumb’ display logic like the main GOV.UK Frontend component above.

duncanjbrown:
Ah, I see! Agree a component would be better.  In the meantime, happy to approve this if you add the html to `_api_docs_header.html` (currently missing)

app/views/layouts/_support_header.html.erb
```diff
@@ -1,3 +1,4 @@
+<button type="button" class="govuk-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
```
duncanjbrown: nit: the sign-in page for support includes this header, so there's an empty dropdown

adamsilver: Will raise a separate ticket.

PR #978, opened 2019-12-24
Disassociate long hint text from certain questions

adamsilver:
## Context

The long guidance is read out in screeen readers every time it's focused. The advice from DAC is to put it above the field below a separate h1.

## Why do you want to be a teacher?

Before:
image

After:
image

## Subject knowledge

Before: 
image

After:
image

## Link to Trello card

https://trello.com/c/fCwP6f9N/689-inappropriate-use-of-aria-why-do-you-want-to-be-a-teacher


duncanjbrown:
In the past Paul H has reminded me to put the error summary above the top `h1`, per https://design-system.service.gov.uk/components/error-summary/. Could we move it up?

Screenshot 2019-12-24 at 09 33 32

adamsilver:
Yes, nice spot!

PR #992, opened 2019-12-30
Add defensive code for nil dates in StatusBoxComponent

stevehook:
## Context

The provider interface application detail page can error if certain dates, such as `enrolled_at`, are `nil`. 

Add some defensive code to the `StatusBoxComponent` to ensure that missing date values cannot cause errors.

## Guidance to review

`ApplicationChoice#enrolled_at` should in theory be present whenever the `status` value is `enrolled` however there is nothing to stop it being `nil` in either database constraints or model validations. You could take the view that `nil` values here are a symptom of a bigger issue but as far as I can see this has only happened in a development environment.

Setting the `ApplicationChoice#enrolled_at` value does seem to be correctly handled already by https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/app/services/confirm_enrolment.rb.

## Link to Trello card

https://trello.com/c/Q9V4eft5/1414-application-details-view-triggers-error-for-some-applications-with-a-missing-date

## Things to check

### DIFF
--- a/app/components/provider_interface/status_box_component.rb
+++ b/app/components/provider_interface/status_box_component.rb
@@ -51,7 +51,7 @@ module ProviderInterface
   private
 
     def format_date(date)
-      date.to_s(:govuk_date)
+      date&.to_s(:govuk_date) || ''
     end
   end
 end
--- a/spec/components/provider_interface/status_box_component_spec.rb
+++ b/spec/components/provider_interface/status_box_component_spec.rb
@@ -66,7 +66,17 @@ RSpec.describe ProviderInterface::StatusBoxComponent do
   end
 
   it 'outputs a date for applications in the enrolled_at state' do
-    application_choice = make_choice(status: 'enrolled', enrolled_at: Time.zone.now)
+    now = Time.zone.now
+    application_choice = make_choice(status: 'enrolled', enrolled_at: now)
+
+    result = render_inline(described_class, application_choice: application_choice)
+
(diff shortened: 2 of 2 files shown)


stevehook:
@tijmenb agree that hiding issues with logic is a bad policy. I didn't get to the bottom of why we had invalid data but if you go through the normal `ConfirmEnrolment` service to set the state then the `enrolled_at` date gets set. Maybe we need to do a bit more digging to work out how we ended up with applications in the invalid state.

I must admit I'm never quite sure how to enforce rules like these, in theory any code can set the state without setting the `enrolled_at`, it's all fine if all such changes go through `ConfirmEnrolment` but we have no way to _enforce_ that. I'm not sure that we should but our options are:

1. Rails Validation rules for states that require some other value being present.
2. A rake task (or similar) to validate (and if possible fixup) existing data.

Of the environments that I have access to `d01` has some invalid data but `d02` does not.

stevehook:
Best guess on how we got invalid date is that the migration that added the new timestamp columns doesn't have any way to set the `enrolled_at` date for enrolled applications. It can't possibly know when the enrolment happened but I suppose it could set the value to the `updated_at`.

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/90055e8d0611b1b2536ca2401f91100210c3fac9/db/migrate/20191219160331_add_dates_to_application_choice.rb

PR #994, opened 2019-12-31
Allow candidates to see courses not open on Apply

tvararu:
## Context

If a candidate chooses one of these courses, they will be directed to the "Apply on UCAS" page. We also remove the "Another course" option which would take them to the same page.

This is a prerequisite to putting all the course options into an autocomplete component.

Update the scope to look for courses which are just exposed in find, without `available_on_apply: true`. Remove the "Another course" option, and update relevant tests.

## Guidance to review

Does the naming of the new helpers make sense? Should we be concerned that the `courses.visible_to_candidates` scope is now incorrectly named, and should we update and use that instead?

Later edit: have pushed another commit that does this.

Later later edit: have also added an `applyable` scope.

## Link to Trello card

https://trello.com/c/uQaeQEm2/619-users-can-select-a-course-from-a-drop-down

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/course_choices_controller.rb
+++ b/app/controllers/candidate_interface/course_choices_controller.rb
@@ -61,10 +61,10 @@ module CandidateInterface
         application_form: current_application,
       )
 
-      if @pick_course.other?
-        redirect_to candidate_interface_course_choices_on_ucas_path
-      elsif !@pick_course.valid?
+      if !@pick_course.valid?
         render :options_for_course
+      elsif !@pick_course.applyable?
+        redirect_to candidate_interface_course_choices_on_ucas_path
       elsif @pick_course.single_site?
         course_id = Course.find_by(code: course_code)
         course_option = CourseOption.where(course_id: course_id).first
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -7,7 +7,7 @@ class Course < ApplicationRecord
   validates :level, presence: true
   validates :code, uniqueness: { scope: :provider_id }
 
-  scope :visible_to_candidates, -> { where(exposed_in_find: true, open_on_apply: true) }
+  scope :visible_to_candidates, -> { where(exposed_in_find: true) }
 
   CODE_LENGTH = 4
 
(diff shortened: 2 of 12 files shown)


app/controllers/candidate_interface/course_choices_controller.rb
```diff
@@ -61,10 +61,10 @@ def pick_course
         application_form: current_application,
       )
 
-      if @pick_course.other?
-        redirect_to candidate_interface_course_choices_on_ucas_path
-      elsif !@pick_course.valid?
+      if !@pick_course.valid?
```
stevehook: Could a selected course be valid but not applyable here? (I couldn't see a validation that checks that a course is `open_on_apply`).

app/models/course.rb
```diff
@@ -7,7 +7,7 @@ class Course < ApplicationRecord
   validates :level, presence: true
   validates :code, uniqueness: { scope: :provider_id }
 
-  scope :visible_to_candidates, -> { where(exposed_in_find: true, open_on_apply: true) }
+  scope :visible_to_candidates, -> { where(exposed_in_find: true) }
```
stevehook: I think the name is still ok here. The distinction between _visible_ and _applyable_ seems clear enough to me.

spec/system/candidate_interface/candidate_selecting_a_course_spec.rb
```diff
@@ -32,8 +32,6 @@
     when_i_click_on_add_another_course
     and_i_choose_that_i_know_where_i_want_to_apply
     and_i_choose_another_provider
-    and_i_choose_another_course
-    then_i_see_the_not_ready_for_you_yet_page
 
     when_i_click_back
```
stevehook: Does clicking back still make sense here? (having not selected a non-applyable course)

app/controllers/candidate_interface/course_choices_controller.rb
tvararu: Yes. If a course is valid, it will fail this conditional and move to the next one that checks if it's `applyable?`. I think that's fine, right?

app/models/course.rb
tvararu: 👍

spec/system/candidate_interface/candidate_selecting_a_course_spec.rb
tvararu: Good spot!

PR #997, opened 2020-01-02
Fix error labels for condition fields when making an offer

stevehook:
## Context

Offers can include up to 4 'further conditions' selected by filling a text area as well as 2 'standard conditions' selected via a checkbox. The further conditions have a validation rule limiting input to 255 characters. If this is exceeded the client doesn't warn the user, the server then throws a validation error but doesn't render the error message next to the offending text area and discards the value that the user entered. We want to preserve the user input and render error messages in the right place.

* Model each further condition attribute on the `MakeAnOffer` service class a named field rather than an array of non-blank values so that any errors can be mapped back to the form fields.
* Fixup the view to use the new named fields.
* Add a character counter component to each of the fields so that users get a visual warning when they approach or exceed the limit.

## Guidance to review

* Is there a simpler way to do this? I am not all that happy about having `further_conditions0` - `further_conditions3` referenced in so many places.
* Is there any more testing needed here?

## Link to Trello card

https://trello.com/c/SOODVP0D/1316-labels-arent-assigned-with-inputs-for-condition-fields-when-making-an-offer-errors-are-wrong

## Things to check

### DIFF
--- a/app/services/make_an_offer.rb
+++ b/app/services/make_an_offer.rb
@@ -1,5 +1,6 @@
 class MakeAnOffer
-  attr_accessor :offer_conditions
+  attr_accessor :standard_conditions
+  attr_accessor :further_conditions
 
   include ActiveModel::Validations
 
@@ -7,11 +8,12 @@ class MakeAnOffer
   MAX_CONDITION_LENGTH = 255
 
   validate :validate_course_data
-  validate :validate_offer_conditions
+  validate :validate_further_conditions
 
-  def initialize(application_choice:, offer_conditions: nil, course_data: nil)
+  def initialize(application_choice:, standard_conditions: nil, further_conditions: nil, course_data: nil)
     @application_choice = application_choice
-    @offer_conditions = offer_conditions
+    @standard_conditions = standard_conditions
+    @further_conditions = further_conditions
     @course_data = course_data
   end
 
@@ -20,7 +22,7 @@ class MakeAnOffer
 
     ApplicationStateChange.new(application_choice).make_offer!
     application_choice.offered_course_option = offered_course_option
-    application_choice.offer = { 'conditions' => (@offer_conditions || []) }
+    application_choice.offer = { 'conditions' => offer_conditions }
 
     application_choice.offered_at = Time.zone.now
     application_choice.save!
@@ -87,15 +89,22 @@ private
     errors.add(:offered_course, "does not belong to provider #{current_provider.code}, it belongs to #{provider.code}") if providers_dont_match
   end
 
-  def validate_offer_conditions
-    return if @offer_conditions.blank?
+  def validate_further_conditions
+    return if @further_conditions.blank?
 
-    unless @offer_conditions.is_a?(Array)
-      errors.add(:offer_conditions, 'must be an array')
+    unless @further_conditions.is_a?(Array)
+      errors.add(:further_conditions, 'must be an array')
       return
     end
 
-    errors.add(:offer_conditions, "has over #{MAX_CONDITIONS_COUNT} elements") if @offer_conditions.count > MAX_CONDITIONS_COUNT
-    errors.add(:offer_conditions, "has a condition over #{MAX_CONDITION_LENGTH} chars in length") if @offer_conditions.any? { |c| c.length > MAX_CONDITION_LENGTH }
+    errors.add(:further_conditions, "has over #{MAX_CONDITIONS_COUNT} elements") if @further_conditions.count > MAX_CONDITIONS_COUNT
+    errors.add(:further_conditions, "has a condition over #{MAX_CONDITION_LENGTH} chars in length") if @further_conditions.any? { |c| c.length > MAX_CONDITION_LENGTH }
+  end
+
+  def offer_conditions
+    [
+      standard_conditions,
+      further_conditions,
+    ].flatten.reject(&:blank?)
   end
 end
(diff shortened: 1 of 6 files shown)


app/services/make_an_offer.rb
```diff
@@ -87,15 +114,23 @@ def validate_course_data
...
-  def validate_offer_conditions
-    return if @offer_conditions.blank?
+  def validate_further_conditions
+    return if further_conditions.blank?
 
-    unless @offer_conditions.is_a?(Array)
-      errors.add(:offer_conditions, 'must be an array')
-      return
+    further_conditions.each_with_index do |value, index|
+      errors.add("further_conditions#{index}", "has a condition over #{MAX_CONDITION_LENGTH} chars in length") if value && value.length > MAX_CONDITION_LENGTH
```
adamsilver: Error message should follow the GOV.UK Design System guidelines:

[whatever it is] must be 255 characters or fewer

Example: Third condition must be 255 characters or fewer

adamsilver: The errors should relate to the individual field, not the fieldset in this case.

There could be different conditions with different error messages.

duncanjbrown:
I've just noticed we don't have API/Provider interface parity here. The API permits up to 20 conditions but the UI only 4.

app/views/provider_interface/decisions/new_offer.html.erb
```diff
@@ -31,10 +31,10 @@
         <legend class="govuk-fieldset__legend govuk-fieldset__legend--m">Further conditions (optional)</legend>
         <p class="govuk-body">Outline any further conditions (for example, studying a subject knowledge enhancement course) and give deadlines for completing them.</p>
 
-        <%= f.govuk_text_area :further_conditions, label: { text: 'First condition', size: 's' }, id: 'first_condition', rows: 3, multiple: true %>
-        <%= f.govuk_text_area :further_conditions, label: { text: 'Second condition', size: 's' }, id: 'second_condition', rows: 3, multiple: true %>
-        <%= f.govuk_text_area :further_conditions, label: { text: 'Third condition', size: 's' }, id: 'third_condition', rows: 3, multiple: true %>
-        <%= f.govuk_text_area :further_conditions, label: { text: 'Further condition', size: 's' }, id: 'further_condition', rows: 3, multiple: true %>
+        <%= f.govuk_text_area :further_conditions0, label: { text: 'First condition', size: 's' }, rows: 3, max_chars: 255, threshold: 75 %>
```
adamsilver: I don't think we should be adding a countdown unless we know one is useful from research.

Is 255 characters just the default setting or is it limited to this for a reason?

I think we should trust providers to say what they need to say. That might mean we up the limit to avoid constrainting providers too much and avoid getting the error at all.

app/services/make_an_offer.rb
```diff
@@ -87,15 +114,23 @@ def validate_course_data
...
-      return
+    further_conditions.each_with_index do |value, index|
+      errors.add("further_conditions#{index}", "must be #{MAX_CONDITION_LENGTH} characters or fewer") if value && value.length > MAX_CONDITION_LENGTH
     end
+  end
+
+  def validate_conditions_array
+    return if offer_conditions.is_a?(Array)
+
+    errors.add(:offer_conditions, 'must be an array')
```
duncanjbrown: This isn't a useful error message for the user. How does it happen? Should it be an exception?

app/services/make_an_offer.rb
```diff
@@ -87,15 +114,23 @@ def validate_course_data
...
-  def validate_offer_conditions
-    return if @offer_conditions.blank?
+  def validate_further_conditions
+    return if further_conditions.blank?
 
-    unless @offer_conditions.is_a?(Array)
-      errors.add(:offer_conditions, 'must be an array')
-      return
+    further_conditions.each_with_index do |value, index|
+      errors.add("further_conditions#{index}", "must be #{MAX_CONDITION_LENGTH} characters or fewer") if value && value.length > MAX_CONDITION_LENGTH
```
adamsilver: Does this render out as, for example:

"First condition must be 255 charactes or fewer"

?

stevehook: the inline error message does read that way yes - I've added a screengrab gif

app/views/provider_interface/decisions/new_offer.html.erb
stevehook: I'm not sure where the limit comes from. Technically there is nothing stopping us storing arbitrary length condition text values.

app/services/make_an_offer.rb
duncanjbrown: The error at the top is not as good as it could be, esp when more than one field is over

Screenshot 2020-01-02 at 11 54 27

duncanjbrown:
Implementation LGTM until we have to change it — I'm happy when @adamsilver is

app/services/make_an_offer.rb
adamsilver: The red text in the top and next to the field should read:

"First condition must be 255 characters or fewer".

The user shouldn't have to rely on the label text above.

app/services/make_an_offer.rb
stevehook: Removed

app/services/make_an_offer.rb
stevehook: @duncanjbrown & @adamsilver I've included the attribute labels in the error messages now:

image

image

PR #1001, opened 2020-01-02
Add survey emails for a candidate

chubberlisk:
## Context

We want to be able to send surveys to candidates and referees in the first phase of the pilot (November/December) from the Support console.

This PR is the first of many to come. It just adds in the two survey emails for candidates in the `CandidateMailer`: the initial one and the chase up email.

## Guidance to review

Are adding these to the `CandidateMailer` the right move? Or is it better to create a `SurveyMailer`? Was planning on adding the survey emails for a referee to the `RefereeMailer` but considering that the emails for a candidate and referee are very similar, I'm not too sure.

## Link to Trello card

https://trello.com/c/KpaTL5K5/543-survey-emails-to-referee-and-candidate-%F0%9F%8F%88

## Things to check

### DIFF
--- a/app/mailers/candidate_mailer.rb
+++ b/app/mailers/candidate_mailer.rb
@@ -35,6 +35,14 @@ class CandidateMailer < ApplicationMailer
 
     view_mail(GENERIC_NOTIFY_TEMPLATE,
               to: application_form.candidate.email_address,
-              subject: t('survey_emails.subject'))
+              subject: t('survey_emails.subject.initial'))
+  end
+
+  def survey_chaser_email(application_form)
+    @candidate_name = application_form.first_name
+
+    view_mail(GENERIC_NOTIFY_TEMPLATE,
+              to: application_form.candidate.email_address,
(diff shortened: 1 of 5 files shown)


C-gyorfi:
> Was planning on adding the survey emails for a referee to the `RefereeMailer` but considering that the emails for a candidate and referee are very similar, I'm not too sure.

If the survey emails are very similar in both `RefereeMailer` and `CandidateMailer` you might want to create a `SurveyMailer` and you could just `include` these in both Mailers 🤔 Otherwise looks like sensible changes and this can be refactored later, I think it safe to merge for now

chubberlisk:
> > Was planning on adding the survey emails for a referee to the `RefereeMailer` but considering that the emails for a candidate and referee are very similar, I'm not too sure.
> 
> If the survey emails are very similar in both `RefereeMailer` and `CandidateMailer` you might want to create a `SurveyMailer` and you could just `include` these in both Mailers 🤔 Otherwise looks like sensible changes and this can be refactored later, I think it safe to merge for now

Will do this!

PR #1003, opened 2020-01-02
Allow candidates to view an incorrect date format

C-gyorfi:
## Context

When a user doesn't complete the DOB field properly but does include text within the day/month/year field, it doesn't keep the data once the validation error message pops up. 

This PR adds an else condition when to `PersonalDetailsForm#date_of_birth` method to return an Open struct when the date is invalid.

### Before
image

### After
image

## Guidance to review

Try to fill in the date_of_birth field with an invalid or partially completed date, you should be able to see the invalid date along with the errors

## Link to Trello card

https://trello.com/c/zzxq2zKG/444-personal-details-doesnt-retain-month-input-value

## Things to check

### DIFF
--- a/app/models/candidate_interface/personal_details_form.rb
+++ b/app/models/candidate_interface/personal_details_form.rb
@@ -70,15 +70,17 @@ module CandidateInterface
 
       if valid_year?(year) && Date.valid_date?(*date_args)
         Date.new(*date_args)
+      else
+        Struct.new(:day, :month, :year).new(day, month, year)
       end
     end
 
     def date_of_birth_valid
-      errors.add(:date_of_birth, :invalid) if date_of_birth.nil?
+      errors.add(:date_of_birth, :invalid) unless date_of_birth.is_a?(Date)
     end
 
     def date_of_birth_not_in_future
-      errors.add(:date_of_birth, :future) if date_of_birth.present? && date_of_birth > Date.today
+      errors.add(:date_of_birth, :future) if date_of_birth.is_a?(Date) && date_of_birth > Date.today
     end
 
     def english_main_language?
--- a/spec/models/candidate_interface/personal_details_form_spec.rb
+++ b/spec/models/candidate_interface/personal_details_form_spec.rb
@@ -96,16 +96,20 @@ RSpec.describe CandidateInterface::PersonalDetailsForm, type: :model do
   end
 
   describe '#date_of_birth' do
-    it 'returns nil for nil day/month/year' do
+    it 'return a nil for nil day/month/year' do
       personal_details = CandidateInterface::PersonalDetailsForm.new(day: nil, month: nil, year: nil)
 
-      expect(personal_details.date_of_birth).to eq(nil)
+      expect(personal_details.date_of_birth.day).to be_nil
+      expect(personal_details.date_of_birth.month).to be_nil
+      expect(personal_details.date_of_birth.year).to be_nil
     end
 
-    it 'returns nil for invalid day/month/year' do
+    it 'can return an invalid date object for invalid day/month/year' do
       personal_details = CandidateInterface::PersonalDetailsForm.new(day: 99, month: 99, year: 99)
 
-      expect(personal_details.date_of_birth).to eq(nil)
+      expect(personal_details.date_of_birth.day).to eq(99)
+      expect(personal_details.date_of_birth.month).to eq(99)
+      expect(personal_details.date_of_birth.year).to eq(99)
     end
 
     it 'returns a date for a valid day/month/year' do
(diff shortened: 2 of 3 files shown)


app/models/candidate_interface/personal_details_form.rb
```diff
@@ -70,15 +70,17 @@ def date_of_birth
...
       if valid_year?(year) && Date.valid_date?(*date_args)
         Date.new(*date_args)
+      else
+        OpenStruct.new(day: day, month: month, year: year)
       end
     end
 
     def date_of_birth_valid
-      errors.add(:date_of_birth, :invalid) if date_of_birth.nil?
+      errors.add(:date_of_birth, :invalid) if date_of_birth.class == OpenStruct
```
duncanjbrown: My instinct is that it would be better to treat the OpenStruct here as a duck and ask the date whether any of its date, month or year fields are nil. Then we're not coupled to it

app/models/candidate_interface/personal_details_form.rb
```diff
@@ -70,15 +70,17 @@ def date_of_birth
...
     end
 
     def date_of_birth_not_in_future
-      errors.add(:date_of_birth, :future) if date_of_birth.present? && date_of_birth > Date.today
+      errors.add(:date_of_birth, :future) if date_of_birth.class == Date && date_of_birth > Date.today
```
duncanjbrown: Would it be possible to avoid this type check?

spec/models/candidate_interface/personal_details_form_spec.rb
```diff
@@ -96,16 +96,16 @@
   end
 
   describe '#date_of_birth' do
-    it 'returns nil for nil day/month/year' do
+    it 'returns an OpenStruct with nil values for nil day/month/year' do
       personal_details = CandidateInterface::PersonalDetailsForm.new(day: nil, month: nil, year: nil)
 
-      expect(personal_details.date_of_birth).to eq(nil)
+      expect(personal_details.date_of_birth).to eq(OpenStruct.new(day: nil, month: nil, year: nil))
```
duncanjbrown: These specs are coupled to OpenStruct too. Could they use its public interface (`#month`, `#day` etc) instead?

app/models/candidate_interface/personal_details_form.rb
```diff
@@ -70,15 +70,17 @@ def date_of_birth
 
       if valid_year?(year) && Date.valid_date?(*date_args)
         Date.new(*date_args)
+      else
+        OpenStruct.new(day: day, month: month, year: year)
```
duncanjbrown: How about a plain `Struct`? The keys never change

app/models/candidate_interface/personal_details_form.rb
tvararu: Yes!

app/models/candidate_interface/personal_details_form.rb
C-gyorfi: It causes `unknown keywords: day, month, year` error

duncanjbrown: `Struct` and `OpenStruct` have different interfaces. The equivalent `Struct` code would be 

```
Struct.new(:day, :month, :year).new(day, month, year)
```

C-gyorfi: Thanks, I changed this in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1003/commits/54c327762f421c982e7db498761bf5a4faae9107

spec/models/candidate_interface/personal_details_form_spec.rb
```diff
@@ -96,16 +96,20 @@
...
       personal_details = CandidateInterface::PersonalDetailsForm.new(day: nil, month: nil, year: nil)
 
-      expect(personal_details.date_of_birth).to eq(nil)
+      expect(personal_details.date_of_birth.day).to be_nil
+      expect(personal_details.date_of_birth.month).to be_nil
+      expect(personal_details.date_of_birth.year).to be_nil
     end
 
-    it 'returns nil for invalid day/month/year' do
+    it 'can return an invalid date object for invalid day/month/year' do
```
duncanjbrown: 👌

spec/models/candidate_interface/personal_details_form_spec.rb
C-gyorfi: Resolved in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1003/commits/54c327762f421c982e7db498761bf5a4faae9107

app/models/candidate_interface/personal_details_form.rb
C-gyorfi: I agree, however, a date can be invalid even if the fields are not nil. I changed this to `unless date_of_birth.is_a?(Date)` since we expect a valid date to be a Date, therefore coupling to that type does not feel harmful. Otherwise, I could create a helper method that can validate the date using `Date.valid_date?`. What do you think? @duncanjbrown

duncanjbrown: :+1: I think this is good!

PR #1009, opened 2020-01-03
Fix QA deployment

tpinney:
## Context

PR#1000 caused a failure of the deployment ARM template when merged yesterday. This issue was never observed and couldn't be reproduced in the DevOps environment.

Properties are being evaluated in the `availability-test-alerts` resource when the condition on the resource should prevent this. The resource conditions have been tested and match correctly between QA and DevOps, but QA is still attempting to parse the resource properties unnecessarily resulting in the failure observed. The DevOps environment is parsing correctly. Why this is happening is still unknown, but a workaround to prevent the error observed as been added. 

An additional `if` has been placed around the offending property in the resource to prevent it being evaluated even when the condition should prevent the `availability-test-alerts` resource from deploying.

## Guidance to review

Successful deployments can be observed in the QA environment in the DevOps GUI showing the fix has worked.
https://dfe-ssp.visualstudio.com/Become-A-Teacher/_build?definitionId=49&_a=summary&repositoryFilter=37&branchFilter=7626

## Link to Trello card

N/A - Bugfix

## Things to check

### DIFF
--- a/azure/template.json
+++ b/azure/template.json
@@ -783,7 +783,7 @@
                         "value": "[concat(variables('appServiceName'), '-at-alert-', split(parameters('customAvailabilityMonitors')[copyIndex('customAlerts')], ':')[0])]"
                     },
                     "actionGroupId": {
-                        "value": "[reference('action-groups', '2019-03-01').outputs.actionGroupResourceId.value]"
+                        "value": "[if(greater(length(parameters('alertRecipientEmails')), 0), reference('action-groups', '2019-03-01').outputs.actionGroupResourceId.value, '')]"
                     },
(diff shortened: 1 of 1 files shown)


tpinney:
> Makes sense to me. I'm don't know Powershell (if that's what it is) so just wonder what happens when `parameters('alertRecipientEmails')` is null/undefined - does `length` return `0`?.

The Azure Resource Manager (ARM) templates are just plain JSON. The `alertRecipientEmails` variable has a default value of an empty array in the parameters section at the top of the file, so if no value is passed in from the azure pipeline code, it will simply use the default.

PR #1018, opened 2020-01-06
Refactor status tag class names

adamsilver:
## Context

Changed class names to colours as they are used to mean different things in different contexts.

## Link to Trello card

https://trello.com/c/UfpbyXvb/1430-refactor-class-names-for-status-tags

### DIFF
--- a/app/components/tag_component.rb
+++ b/app/components/tag_component.rb
@@ -8,24 +8,7 @@ private
 
   attr_reader :text
 
-  def css_classes(type)
-    tag_css_class = case type
-                    when :danger
-                      'app-tag--danger'
-                    when :secondary
-                      'app-tag--secondary'
-                    when :info
-                      'app-tag--info'
-                    when :info_unfilled
-                      'app-tag--info-unfilled'
-                    when :warning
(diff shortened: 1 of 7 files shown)


app/components/tag_component.rb
```diff
@@ -10,20 +10,20 @@ def initialize(text:, type:)
...
+                    when :green
+                      'app-tag--green'
+                    when :orange
+                      'app-tag--orange'
+                    when :purple
+                      'app-tag--purple'
                     else
-                      'app-tag--primary'
+                      'app-tag--blue'
                     end
```
tijmenb: Since we're using colour names in the Perhaps we can simplify this by replacing everything with `app-tag--#{color}`.

adamsilver: Thanks @tijmenb.

I was also thinking that the else statement should be killed. I think if you want it to be ‘blue’ then pass that in.

Then the mapping is more explicit further up the stream. What d'ya think?

tijmenb: Yes great idea, go for it.

PR #1020, opened 2020-01-06
Show an accurate message for provider users who haven't signed in

duncanjbrown:
## Context

We currently show "Never signed in" on the list of provider users if we have no `last_signed_in_at` value for the provider. This is the case when a provider has not signed in since records began on 24/12/2019 at ~2:45pm (Azure deployment https://dev.azure.com/dfe-ssp/Become-A-Teacher/_build/results?buildId=36746&view=results).

This PR makes that message a bit more useful

- if they were created before sign-in records began on 24/12/2019, say we don't know
- if they were created after records began but haven't signed in, say
with certainty that they've never signed in

Use the `created_at` date of the `ProviderUser` to decide which message to show in the event there's no `last_signed_in_at` date.

## Link to Trello card

https://trello.com/c/9olWppuF/1431-dont-incorrectly-say-a-provider-never-signed-in-if-they-could-have-signed-in-before-records-began

## Things to check

### DIFF
--- a/app/components/support_interface/provider_users_table_component.html.erb
+++ b/app/components/support_interface/provider_users_table_component.html.erb
@@ -18,7 +18,11 @@
           <% if row[:last_signed_in_at] %>
             <td class="govuk-table__cell"><%= row[:last_signed_in_at] %></td>
           <% else %>
-            <td class="govuk-table__cell"><em>Never signed in</em></td>
+            <% if row[:created_at] < Date.new(2019, 12, 25) %>
+              <td class="govuk-table__cell"><em>Unknown (records began 24 December 2019)</em></td>
+            <% else %>
+              <td class="govuk-table__cell"><em>Never signed in</em></td>
+            <% end %>
           <%end %>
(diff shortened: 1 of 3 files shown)


app/components/support_interface/provider_users_table_component.html.erb
```diff
@@ -18,7 +18,11 @@
           <% if row[:last_signed_in_at] %>
             <td class="govuk-table__cell"><%= row[:last_signed_in_at] %></td>
           <% else %>
-            <td class="govuk-table__cell"><em>Never signed in</em></td>
+            <% if row[:created_at] < Date.new(2019, 12, 25) %>
+              <td class="govuk-table__cell"><em>Unknown (records began 24/12/2019)</em></td>
```
fofr:
```suggestion
              <td class="govuk-table__cell"><em>Unknown (records began 24 December 2019)</em></td>
```

duncanjbrown:
Thanks @fofr, fixed up

PR #1023, opened 2020-01-06
Provider user accepts data sharing agreement for provider

mnacos:
### Context

Provider users must be able to sign the data sharing agreement when they first access the platform. This will assist/streamline the onboarding of SCITTs to the service.

This new flow in the app will replace the current DocuSign-based manual process. The data sharing agreement will be presented within the app and clicking on a simple checkbox will be all that's needed to accept the agreement. The identity of the person agreeing, as well as the timestamp of the acceptance, will be recorded in the database and exposed in the provider interface.

Requirements:

- Each ~~user~~ provider accepts the data sharing agreement only once
- The first user from each `Provider` must select they 'agree to the data sharing agreement'
- No-one linked to a particular provider can access the provider interface until the relevant data sharing agreement has been accepted.

The data sharing agreement will be presented to the user as a web page. The organisation name will need to appear in a couple of places in the form and will be automatically populated by the app. The provider may need access to the data sharing agreement at any point after it has been signed, so an appropriate route within the app will be provided.

Even though provider users go through DfE Sign-in, there is a ProviderUser model we can store values against. However, the relationship between `Provider` and `ProviderUser` is many-to-many and acceptance of the data sharing agreement must be enforced at the `Provider` level. For now, we'll assume any association between a `Provider` and and a `ProviderUser` gives this user the the authority to accept the data sharing agreement on the provider's behalf.

When a `ProviderUser` logs in, the relevant controller will check if any of their associated providers hasn't accepted the data sharing agreement. If this is the case, the controller will redirect the user to a data sharing agreement flow. Subsequent users for the same provider will not be redirected to this flow, provided the first user has completed the flow.

The data sharing agreement flow is triggered with this logic:

- The current user's list of associated providers is found
- We check if any of these providers have not accepted the data sharing agreement
- We trigger the data sharing agreement flow for the first of these providers
- Once the agreement has been accepted we redirect the user to the data sharing agreement flow for the next provider that requires it or the `provider_interface`.

In the database, agreement acceptance will be stored in a new table, `provider_agreements`. The table includes `provider_id`, `provider_user_id`, `agreement_type` and `accepted_at` fields.

Information about who has signed the data sharing agreement and when is visible in the support interface.

### Link to Trello card

### DIFF
--- a/db/migrate/20200106111239_create_provider_agreements.rb
+++ b/db/migrate/20200106111239_create_provider_agreements.rb
@@ -1,0 +1,11 @@
+class CreateProviderAgreements < ActiveRecord::Migration[6.0]
+  def change
+    create_table :provider_agreements do |t|
+      t.references :provider, null: false, foreign_key: true
+      t.references :provider_user, null: false, foreign_key: true
+      t.string :agreement_type
+      t.datetime :accepted_at
+      t.timestamps
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_12_23_162115) do
+ActiveRecord::Schema.define(version: 2020_01_06_111239) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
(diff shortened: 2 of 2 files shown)


duncanjbrown:
Just to confirm the behaviour of:

> The current user's list of associated providers is found

Say there are providers `A` and `B`, where `A` accredits `B`. Is it true that:

When I sign in as a user from provider `B`. I accept terms for `B` only.
I sign in as a user from provider `A`. I accept terms for `A` then `B` as I will be able to see `B`'s incoming applications as well as `A`s? (assuming nobody from `B` has already accepted for `B`)

PR #1028, opened 2020-01-06
Provider user accepts data sharing agreement for provider

mnacos:
### Context

Provider users must be able to sign the data sharing agreement when they first access the platform. This will assist/streamline the onboarding of SCITTs to the service.

This new flow in the app will replace the current DocuSign-based manual process. The data sharing agreement will be presented within the app and clicking on a simple checkbox will be all that's needed to accept the agreement. The identity of the person agreeing, as well as the timestamp of the acceptance, will be recorded in the database and exposed in the provider interface.

Requirements:

- Each ~~user~~ provider accepts the data sharing agreement only once
- The first user from each `Provider` must select they 'agree to the data sharing agreement'
- No-one linked to a particular provider can access the provider interface until the relevant data sharing agreement has been accepted.

The data sharing agreement will be presented to the user as a web page. The organisation name will need to appear in a couple of places in the form and will be automatically populated by the app.

Even though provider users go through DfE Sign-in, there is a ProviderUser model we can store values against. However, the relationship between `Provider` and `ProviderUser` is many-to-many and acceptance of the data sharing agreement must be enforced at the `Provider` level. For now, we'll assume any association between a `Provider` and and a `ProviderUser` gives this user the the authority to accept the data sharing agreement on the provider's behalf.

When a `ProviderUser` logs in, the relevant controller will check if any of their associated providers hasn't accepted the data sharing agreement. If this is the case, the controller will redirect the user to a data sharing agreement flow. Subsequent users for the same provider will not be redirected to this flow, provided the first user has completed the flow.

The data sharing agreement flow is triggered with this logic:

- The current user's list of associated providers is found
- We check if any of these providers have not accepted the data sharing agreement
- We trigger the data sharing agreement flow for the first of these providers
- Once the agreement has been accepted we redirect the user to the data sharing agreement flow for the next provider that requires it or the `provider_interface`.

In the database, agreement acceptance will be stored in a new table, `provider_agreements`. The table includes `provider_id`, `provider_user_id`, `agreement_type` and `accepted_at` fields.

Information about who has signed the data sharing agreement and when is visible in the support interface.

#### Provider Interface

The provider user in this screen capture has been associated with two providers, none of which have a data sharing agreement in place. Therefore, they need to accept two data sharing agreements before given access to the applications interface.

optimised

#### Support Interface

Without agreement:
image

With agreement:
image

### Link to Trello card

### DIFF
--- a/app/controllers/provider_interface/provider_agreements_controller.rb
+++ b/app/controllers/provider_interface/provider_agreements_controller.rb
@@ -1,0 +1,42 @@
+module ProviderInterface
+  class ProviderAgreementsController < ProviderInterfaceController
+    skip_before_action :check_data_sharing_agreements
+
+    def new_data_sharing_agreement
+      @provider_agreement = GetPendingDataSharingAgreementsForProviderUser.call(provider_user: current_provider_user).first
+      if @provider_agreement
+        render :data_sharing_agreement
+      else
+        redirect_to provider_interface_path
+      end
+    end
+
+    def create_data_sharing_agreement
+      @provider_agreement = ProviderAgreement.new(provider_agreement_params.merge(provider_user: current_provider_user))
+      if @provider_agreement.save
+        redirect_to provider_interface_path
+      else
+        render :data_sharing_agreement
+      end
+    end
+
+    def show_data_sharing_agreement
+      @provider_agreement = ProviderAgreement.find_by_id params[:id]
+      if @provider_agreement
+        render :data_sharing_agreement
+      else
+        redirect_to provider_interface_path
+      end
+    end
+
+  private
+
+    def provider_agreement_params
+      params.require(:provider_agreement).permit(
+        :accept_agreement,
+        :agreement_type,
+        :provider_id,
+      )
+    end
+  end
+end
--- a/app/controllers/provider_interface/provider_interface_controller.rb
+++ b/app/controllers/provider_interface/provider_interface_controller.rb
@@ -4,6 +4,7 @@ module ProviderInterface
     before_action :authenticate_provider_user!
     around_action :set_audit_username
     before_action :add_identity_to_log
+    before_action :check_data_sharing_agreements
 
     layout 'application'
 
@@ -57,6 +58,12 @@ module ProviderInterface
 
       RequestLocals.store[:identity] = { dfe_sign_in_uid: current_provider_user.dfe_sign_in_uid }
       Raven.user_context(dfe_sign_in_uid: current_provider_user.dfe_sign_in_uid)
+    end
+
+    def check_data_sharing_agreements
+      if GetPendingDataSharingAgreementsForProviderUser.call(provider_user: current_provider_user).any?
+        redirect_to provider_interface_new_data_sharing_agreement_path
+      end
     end
   end
 end
--- a/app/models/provider_agreement.rb
+++ b/app/models/provider_agreement.rb
@@ -1,0 +1,26 @@
+class ProviderAgreement < ActiveRecord::Base
+  belongs_to :provider
+  belongs_to :provider_user
+  attr_accessor :accept_agreement
+
+  validates :accept_agreement, :agreement_type, :provider, :provider_user, presence: true
+  validate :provider_is_associated_with_the_user
+  before_create :set_accepted_at
+
+  scope :data_sharing_agreements, -> { where(agreement_type: :data_sharing_agreement) }
+  scope :for_provider, ->(specific_provider) { where(provider: specific_provider) }
+
+private
+
+  def provider_is_associated_with_the_user
+    if provider && provider_user
+      unless provider.provider_users.pluck(:id).include? provider_user.id
+        errors.add(:provider, 'Provider/user mismatch')
+      end
+    end
+  end
+
+  def set_accepted_at
+    self.accepted_at ||= Time.zone.now
+  end
+end
--- a/app/services/get_pending_data_sharing_agreements_for_provider_user.rb
+++ b/app/services/get_pending_data_sharing_agreements_for_provider_user.rb
@@ -1,0 +1,9 @@
+class GetPendingDataSharingAgreementsForProviderUser
+  def self.call(provider_user:)
+    providers = provider_user.providers
+    no_dsa = providers.where.not(id: ProviderAgreement.data_sharing_agreements.for_provider(providers).select(:provider_id))
+    no_dsa.map do |provider|
+      ProviderAgreement.new(agreement_type: :data_sharing_agreement, provider: provider, provider_user: provider_user)
+    end
+  end
+end
--- a/app/views/provider_interface/provider_agreements/data_sharing_agreement.html.erb
+++ b/app/views/provider_interface/provider_agreements/data_sharing_agreement.html.erb
@@ -1,0 +1,427 @@
+<% page_name = 'data_sharing_agreement' %>
+<%= content_for :title, t("page_titles.#{page_name}") %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl"><%= t("page_titles.#{page_name}") %></h1>
+
+    <div class="app-styled-content">
+
+      <p>between the Department of Education (DfE) and <em><%= @provider_agreement.provider.name %></em>, ‘the provider’.</p>
+
+      <h2>What we ask of you when you use and share data</h2>
+
+      <p>
+        Our data sharing agreement sets out how you should use and look after the data we share with you.
+        Please read the data sharing agreement below in full to check you can meet your responsibilities.
+        Once you have read it, please scroll to the bottom of the page to accept it.
+      </p>
+
+      <hr>
+
+      <h2>Data sharing agreement for Apply for teacher training</h2>
+
+      <p>
+        This agreement applies to personal data shared between the Department for Education (DfE) and
+        <em><%= @organisation_name %></em>,
+        ‘the provider’, as part of Apply for teacher training.
+      </p>
+
+      <h3 class="app-contents-list__title">What’s in this data sharing agreement?</h3>
+
+      <ol class="app-contents-list__list">
+        <li class="app-contents-list__list-item app-contents-list__list-item--parent">
+          <a class="app-contents-list__link" href="#section-1">Section 1: Introduction</a>
+          <p style="padding: 12px 0 0 10px;">
+            <a class="app-contents-list__link" href="#section-1-1">1.1 Background</a><br>
+          </p>
+        </li>
+
+        <li class="app-contents-list__list-item app-contents-list__list-item--parent">
+          <a class="app-contents-list__link" href="#section-2">
+            Section 2: What personal data DfE and the provider process/share and why
+          </a>
+          <p style="padding: 12px 0 0 10px;">
+            <a class="app-contents-list__link" href="#section-2-1">2.1 What data DfE and the provider will process/share</a><br>
+            <a class="app-contents-list__link" href="#section-2-2">2.2 What the provider can use personal data for</a><br>
+            <a class="app-contents-list__link" href="#section-2-3">2.3 What DfE can use personal data for</a>
+          </p>
+        </li>
+
+        <li class="app-contents-list__list-item app-contents-list__list-item--parent">
+          <a class="app-contents-list__link" href="#section-3">
+            Section 3: DfE’s legal basis for sharing/processing personal data
+          </a>
+          <p style="padding: 12px 0 0 10px;">
+            <a class="app-contents-list__link" href="#section-3-1">3.1 Lawful conditions for sharing/processing personal data</a><br>
+            <a class="app-contents-list__link" href="#section-3-2">3.2 The right to respect for private and family life</a><br>
+            <a class="app-contents-list__link" href="#section-3-3">3.3 Privacy notices</a>
+          </p>
+        </li>
+
+        <li class="app-contents-list__list-item app-contents-list__list-item--parent">
+          <a class="app-contents-list__link" href="#section-4">
+            Section 4: Data handling
+          </a>
+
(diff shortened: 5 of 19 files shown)


mnacos:
Had to re-create this PR, as it was automatically closed when the db migration was merged. Here is @duncanjbrown 's comment from the previous PR:

> Just to confirm the behaviour of:
> 
> > The current user's list of associated providers is found
> 
> Say there are providers `A` and `B`, where `A` accredits `B`. Is it true that:
> 
> When I sign in as a user from provider `B`. I accept terms for `B` only.
> When I sign in as a user from provider `A`, I accept terms for `A` then `B` as I will be able to see `B`'s incoming applications as well as `A`s? (assuming nobody from `B` has already accepted for `B`)

app/controllers/provider_interface/provider_agreements_controller.rb
```diff
@@ -0,0 +1,25 @@
+module ProviderInterface
+  class ProviderAgreementsController < ProviderInterfaceController
+    before_action :set_page_name
+
+    def new_data_sharing_agreement
```
alexdesi: This controller sounds very RESTy,
we might just want to use `new, create, show` actions,
what do you think?

spec/system/provider_interface/provider_accepts_data_sharing_agreement.rb
```diff
@@ -0,0 +1,77 @@
+require 'rails_helper'
+
+RSpec.feature 'Accept data sharing agreement' do
```
tijmenb: Having a single scenario per test file works best for readability and maintainability - I've raised https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1068 to document and discuss.

app/controllers/provider_interface/provider_agreements_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+      @provider_agreement = GetPendingDataSharingAgreementsForProviderUser.call(provider_user: current_provider_user).first
+      if @provider_agreement
+        render_agreement
+      else
+        redirect_to action: :show_data_sharing_agreement
+      end
+    end
+
+    def create_data_sharing_agreement
+      @provider_agreement = ProviderAgreement.new(provider_agreement_params.merge(provider_user: current_provider_user))
```
tijmenb: Could we use a form object for the validation/saving of the data?

app/controllers/provider_interface/provider_agreements_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+      begin
+        @provider_agreement.save!
+        redirect_to '/provider'
```
tijmenb: This should probably use a path helper.

app/controllers/provider_interface/provider_agreements_controller.rb
```diff
@@ -0,0 +1,55 @@
+module ProviderInterface
+  class ProviderAgreementsController < ProviderInterfaceController
+    before_action :set_page_name
```
tijmenb: `@page_name` seems always the same, so this can go into the view I think?

app/controllers/provider_interface/provider_agreements_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+        redirect_to action: :show_data_sharing_agreement
+      end
+    end
+
+    def create_data_sharing_agreement
+      @provider_agreement = ProviderAgreement.new(provider_agreement_params.merge(provider_user: current_provider_user))
+      begin
+        @provider_agreement.save!
+        redirect_to '/provider'
+      rescue ActiveRecord::RecordInvalid
```
tijmenb: When do we expect this to occur?

app/controllers/provider_interface/provider_agreements_controller.rb
mnacos: It's a nod to RESTful conventions, but the model and the controller are meant to be general purpose (ProviderAgreement instead of DataSharingAgreement), while the routes are very specific (`/provider/data-sharing-agreement/new`). So I've made the controller action names specific, too.

app/controllers/provider_interface/provider_agreements_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+      @page_name = 'data_sharing_agreement'
+    end
+
+    def set_declaration
+      @declaration = \
+        if !@provider_agreement.persisted?
+          "#{@organisation_name} agrees to comply with the data sharing practices outlined in this agreement"
+        else
+          agreement_date = @provider_agreement.accepted_at.strftime("%d %B %Y")
+          "#{@organisation_name} agreed to comply with the data sharing practices outlined in this agreement on #{agreement_date}"
```
tijmenb: In what case will the user be seeing this message? Are we ever showing a signed agreement?

app/controllers/provider_interface/provider_agreements_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+        end
+    end
+
+    def render_agreement
+      @organisation_name = @provider_agreement.provider.name
+      set_declaration
```
tijmenb: Perhaps it would be cleaner to use the `@provider_agreement` var in the views, instead of generating new instance variables here.

mnacos:
> Had to re-create this PR, as it was automatically closed when the db migration was merged. Here is @duncanjbrown 's comment from the previous PR:
> 
> > Just to confirm the behaviour of:
> > > The current user's list of associated providers is found
> > 
> > 
> > Say there are providers `A` and `B`, where `A` accredits `B`. Is it true that:
> > When I sign in as a user from provider `B`. I accept terms for `B` only.
> > When I sign in as a user from provider `A`, I accept terms for `A` then `B` as I will be able to see `B`'s incoming applications as well as `A`s? (assuming nobody from `B` has already accepted for `B`)

I agree with your thinking, but the relevant scoping of relationships between providers is not clear yet. I think we should merge this version now, supporting the existing relationships/functionality ('everyone sees everything as long as there is an association in place') and prepare a separate PR to change who is authorised to accept agreements on behalf of which organisations once there is clarity about that.

app/services/get_pending_data_sharing_agreements_for_provider_user.rb
```diff
@@ -0,0 +1,9 @@
+class GetPendingDataSharingAgreementsForProviderUser
+  def self.call(provider_user:)
+    provider_user.providers.map { |provider|
+      if ProviderAgreement.data_sharing_agreements.for_provider(provider).empty?
+        ProviderAgreement.new(agreement_type: :data_sharing_agreement, provider: provider, provider_user: provider_user)
+      end
+    }.compact
```
tijmenb: This executes a query per associated provider, can we avoid the N+1 somehow?

app/views/provider_interface/provider_agreements/data_sharing_agreement.html.erb
```diff
@@ -0,0 +1,423 @@
...
+
+          <h5>Why is DfE developing this service?</h5>
+          <p>
+            DfE research showed that teacher training candidates and training providers find the current application and selection process difficult.
+            DfE is trying to reduce the number of people dropping out of applying for teacher training by building a new application process.
+            Apply for teacher training will eventually share data with approximately 2,000 teacher training providers.
+            DfE is trialling the service with selected teacher training providers. The trial will help DfE improve the service and inform government policy relating to teacher retention and recruitment.
+          </p>
+
+          <h5>Why do we need this data sharing agreement?</h5>
```
tijmenb: This looks unstyled:

image

app/controllers/provider_interface/provider_agreements_controller.rb
tijmenb: What other agreements are we going to have?

app/views/support_interface/providers/show.html.erb
```diff
@@ -4,6 +4,22 @@
...
+      Data Sharing Agreement
+    </h3>
+
+    <p class="govuk-body">
+      <% if @provider_agreement %>
+        Accepted by <strong>
+        <%= @provider_agreement.provider_user.email_address %>
+        </strong>
+        on
+        <%= @provider_agreement.accepted_at.to_s(:govuk_date) %>.
```
tijmenb:
```suggestion
        <%= @provider_agreement.accepted_at.to_s(:govuk_date_and_time) %>.
```

spec/factories.rb
```diff
@@ -146,6 +146,29 @@
     initialize_with { Provider.find_or_create_by code: code }
     code { Faker::Alphanumeric.alphanumeric(number: 3).upcase }
     name { Faker::Educator.university }
+
+    transient do
+      provider_agreements_count { 1 }
```
tijmenb: Do we need to default to having an agreement here? I think this will generate a lot more data agreements than necessary.

app/views/provider_interface/provider_agreements/data_sharing_agreement.html.erb
mnacos: weirdly, I cannot see this. In the template, all of these headings are `<h5>`.

image

app/controllers/provider_interface/provider_agreements_controller.rb
mnacos: yes

app/controllers/provider_interface/provider_agreements_controller.rb
mnacos: We only have data sharing agreements at the mo but the intention is to replace Docusign when onboarding SCITTs.

app/controllers/provider_interface/provider_agreements_controller.rb
mnacos: yes

app/controllers/provider_interface/provider_agreements_controller.rb
mnacos: I started along these lines but it was introducing too much indirection, considering how simple the operations are. Having these validations in the model has additional advantages, e.g. we could find invalid agreements in the db if a `ProviderUser` loses access to a `Provider`.

app/controllers/provider_interface/provider_agreements_controller.rb
mnacos: If a user modifies the html to supply a `provider_id` they are not associated with.

app/controllers/provider_interface/provider_agreements_controller.rb
mnacos: yes

app/controllers/provider_interface/provider_interface_controller.rb
```diff
@@ -58,5 +59,11 @@ def add_identity_to_log
       RequestLocals.store[:identity] = { dfe_sign_in_uid: current_provider_user.dfe_sign_in_uid }
       Raven.user_context(dfe_sign_in_uid: current_provider_user.dfe_sign_in_uid)
     end
+
+    def check_data_sharing_agreements
+      if !GetPendingDataSharingAgreementsForProviderUser.call(provider_user: current_provider_user).empty?
```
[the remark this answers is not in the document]
mnacos: ok

spec/factories.rb
mnacos: Yes, all the other system specs break otherwise (no agreement = going through agreement flow)

app/services/get_pending_data_sharing_agreements_for_provider_user.rb
mnacos: yep, rewrote the query

app/views/provider_interface/provider_agreements/data_sharing_agreement.html.erb
tijmenb: Weird - `h5` isn't defined, so it shouldn't be styled at the moment:

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/app/frontend/styles/_styled-content.scss#L13

app/services/get_pending_data_sharing_agreements_for_provider_user.rb
```diff
@@ -0,0 +1,7 @@
+class GetPendingDataSharingAgreementsForProviderUser
+  def self.call(provider_user:)
+    provider_user.providers.left_outer_joins(:provider_agreements).where(provider_agreements: {id: nil}).map do |provider|
```
tijmenb: Nice 👍

app/services/get_pending_data_sharing_agreements_for_provider_user.rb
mnacos: Got it down to one query, thanks!

app/services/get_pending_data_sharing_agreements_for_provider_user.rb
mnacos: It's even better now (one sql query for everything)

app/views/provider_interface/provider_agreements/data_sharing_agreement.html.erb
mnacos: Ok, I've added something

app/controllers/provider_interface/provider_agreements_controller.rb
mnacos: yes, we do have a show route, so that we can email it to provider users if they need to see it again

app/models/provider_agreement.rb
```diff
@@ -0,0 +1,26 @@
...
+  before_create :set_accepted_at
+
+  scope :data_sharing_agreements, -> { where(agreement_type: :data_sharing_agreement) }
+  scope :for_provider, ->(specific_provider) { where(provider: specific_provider) }
+
+private
+
+  def provider_is_associated_with_the_user
+    if provider && provider_user
+      unless provider.provider_users.pluck(:id).include? provider_user.id
```
duncanjbrown: Ah, `pluck`! 👍

duncanjbrown:
I agree with @tijmenb’s comments - looks good to me, allowing for those.

PR #1029, opened 2020-01-06
Fix field length in API, add missing "breaks:" field to work experience

duncanjbrown:
## Context

When working through the API fields to check their lengths and validation criteria we noticed some inconsistencies:

- The personal statement field in the API has a 600 char limit. In fact this field is made out of the "Personal statement" and "Subject knowledge" sections of the form, so the total is in fact 1000 words. On top of that we add some text to explain that there are two sections munged into this field.
- The API does not include a reason for a break in work history
- The API does not include a reason for not having a Maths GCSE

- increase the character count on `personal_statement`
- introduce a `breaks` field to `work_experience`

The Maths GCSE issue is left out as this information isn't currently in the Provider UI and we need to decide where it goes.

## Link to Trello card

https://trello.com/c/6vU8nJ7i/1433-personal-statement-field-has-wrong-length-in-api

## Things to check

### DIFF
--- a/app/views/api_docs/pages/release_notes.md
+++ b/app/views/api_docs/pages/release_notes.md
@@ -1,5 +1,15 @@
+### 7th January 2020
+
+- Correct size of
+  [`personal_statement`](/api-docs/reference#applicationattributes-object)
+  field to 11624 chars
+- Introduce `work_history_break_explanation` field to
+  [`work_experience`](/api-docs/reference#workexperiences-object).  This
+  contains the candidate’s explanation for any breaks in work history.
+
 ### v1.0 — 18th December 2019
 
 Initial release of the API.
 
-For a log of pre-release changes, [see the alpha release notes](/api-docs/alpha-release-notes).
+For a log of pre-release changes, [see the alpha release
+notes](/api-docs/alpha-release-notes).
--- a/config/vendor-api-v1.yml
+++ b/config/vendor-api-v1.yml
@@ -753,6 +753,7 @@ components:
       required:
         - jobs
         - volunteering
+        - breaks
       properties:
         jobs:
           type: array
@@ -764,6 +765,12 @@ components:
           description: The candidate’s experience as a volunteer
           items:
             $ref: "#/components/schemas/WorkExperience"
+        breaks:
+          type: string
+          nullable: true
+          description: The candidate’s explanation for any breaks in work history
+          maxLength: 4096 # 400 words
+          example: 'I spent time volunteering overseas...'
 
     WorkExperience:
       type: object
--- a/config/vendor-api-v1.yml
+++ b/config/vendor-api-v1.yml
@@ -377,7 +377,7 @@ components:
           example: "2019-06-13T23:59:59Z"
         personal_statement:
           type: string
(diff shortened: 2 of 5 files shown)


app/views/api_docs/pages/release_notes.md
```diff
@@ -3,3 +3,8 @@
 Initial release of the API.
 
 For a log of pre-release changes, [see the alpha release notes](/api-docs/alpha-release-notes).
+
+#### 7th January 2020
```
tijmenb: Perhaps at this add the top to have reverse chronological order like the alpha release notes.

app/views/api_docs/pages/release_notes.md
```diff
@@ -3,3 +3,8 @@
...
+
+- Correct size of `personal_statement` field to accomodate 1000 words, not 600.
```
tijmenb: Can we link to the properties? Would be good to add the new char limit as well (I don't think we mention "words" anywhere else?)

app/views/api_docs/pages/release_notes.md
```diff
@@ -3,3 +3,8 @@
...
+- Introduce `breaks` field under `work_experience`
```
tijmenb:
```suggestion
- Introduce `breaks` field under `work_experience`. This contains the candidate’s explanation for any breaks in work history.
```

config/vendor-api-v1.yml
```diff
@@ -764,6 +765,12 @@ components:
           description: The candidate’s experience as a volunteer
           items:
             $ref: "#/components/schemas/WorkExperience"
+        breaks:
```
tijmenb: Because the other keys here are arrays and "breaks" is plural, I would've expected this to be an array too. Maybe `work_history_break_explanation` or something?

config/vendor-api-v1.yml
```diff
@@ -764,6 +765,12 @@ components:
...
+          type: string
+          nullable: true
+          description: The candidate’s explanation for any breaks in work history
```
tijmenb:
```suggestion
          description: The candidate’s explanation for any breaks in work history. Will be null if there aren't any breaks in the candidate’s work history. We define a break in work history as more than a month between 2 jobs.
```

app/views/api_docs/pages/release_notes.md
duncanjbrown: Yes

app/views/api_docs/pages/release_notes.md
duncanjbrown: Yes

app/views/api_docs/pages/release_notes.md
duncanjbrown: Yes

config/vendor-api-v1.yml
duncanjbrown: Yes

config/vendor-api-v1.yml
duncanjbrown: Yes²

duncanjbrown:
Yes to all

duncanjbrown:
Should be ready for another look

PR #1032, opened 2020-01-06
Add `aria-describedby` for 'Change' links in work history

chubberlisk:
## Context

From our external accessibility audit it was noted that when we have `Change` links on review sections that consist of multiple entries, e.g. work history, they are not unique and descriptive which makes it difficult for screen readers users to identify the destination or purpose of the link out of context.

Following discussions, to improve the accessibility we decided that adding `aria-describedby` that includes a description of the job and the dates will uniquely identify and describe a job role. See https://ukgovernmentdfe.slack.com/archives/CPACU8GUF/p1578310497004300 for more detail.

As there are a number of other pages that have this issue, work history is the only one that is addressed so that a valid solution can be establish first. Further PRs will be made to fix the other pages.

This PR adds `aria-describedby` for `Change` links in work history.  In order to do this `id`s needed to be created for the value cell of the 'Job' and 'Dates' row of an entry so that they can be referenced in the `aria-describedby`.

image

## Guidance to review

The implementation of this feels weird. Feels like there's a better way to do this. Recommendations are more than welcome! Pliz and tenk u.

## Link to Trello card

https://trello.com/c/XsqlA0hu/698-dac-page-33-add-hidden-content-to-change-links-on-work-history-review-and-other-pages

## Things to check

### DIFF
--- a/app/components/work_history_review_component.rb
+++ b/app/components/work_history_review_component.rb
@@ -58,10 +58,12 @@ private
 
   def job_row(work)
     {
+      id: generate_id(work_id: work.id, attribute: 'job'),
       key: 'Job',
       value: [work.role, work.organisation],
       action: 'job',
       change_path: candidate_interface_work_history_edit_path(work.id),
+      aria_describedby: generate_aria_describedby(work.id),
     }
   end
 
@@ -71,6 +73,7 @@ private
       value: work.commitment.dasherize.humanize,
       action: 'type',
       change_path: candidate_interface_work_history_edit_path(work.id),
+      aria_describedby: generate_aria_describedby(work.id),
     }
   end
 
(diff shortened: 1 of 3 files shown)


app/components/work_history_review_component.rb
```diff
@@ -101,4 +111,16 @@ def formatted_end_date(work)
 
     work.end_date.to_s(:month_and_year)
   end
+
+  def generate_id(work_id:, attribute:)
+    "work-history-#{work_id}-#{attribute}"
+  end
```
tvararu: I'm wondering if this method is really necessary; thinking out loud:

```ruby
generate_id(13, 'dates')
```

```ruby
"work-history-13-dates"
```

We need to duplicate the `generate_id` call anyway in this file, the code it's replacing isn't too complex. There's the argument that it does DRY the `work-history-` bit but ehhh... it's not a huge difference either way.

Feel free to ignore this ramble. 👍

fofr:
This looks great 💯 

Do we need this for volunteering experience section too?
What about references?
Do you have an example screenreader output?

chubberlisk:
> This looks great 💯
> 
> Do we need this for volunteering experience section too?
> What about references?

Yush, we need to do those as well. Just wanted to address one section first so that we could establish an implementation before fixing the others. Will be doing those in a future PR or PRs.

app/components/work_history_review_component.rb
chubberlisk: So I like the idea that generating an `id` is handled by a single method so if we wanted it to change the structure we just need to update that method I believe. Also keeping an eye on the fact some refactoring may be done as other sections have the same problem with the `Change` links.

Am going to leave it how it is, although whilst revising this I did notice I didn't need the variable assignments e.g. `id = generate_id(work_id: work.id, attribute: 'dates')` so updating this.

PR #1037, opened 2020-01-07
SPIKE — DO NOT MERGE. Munge "missing_gcses" into a text field adjacent to "gcse"s

duncanjbrown:
## Context

This information is missing from the API, so we need to add it. However, there's no obvious place to put it.

Existing GCSEs follow the `Qualification` schema, with keys for year, subject, grade, type etc. In the case of a missing GCSE we know only the subject.

Considered implementations:

1. include these GCSEs alongside the others, but nullify all but the `subject` field, and introduce a `missing_explanation` field. (the **nullify** approach)
2. introduce a new data type for a `MissingGCSE` or equivalent and let it appear in the list alongside the other GCSEs (the **mixed list** approach)
3. introduce a new field to the qualifications structure with free text containing the `missing_explanation` for any and all missing GCSEs (the **munge** approach)

Discussed with @stevehook @tijmenb @tvararu.

The **nullify** approach keeps the list consistent, but 
- means consumers have to pick through this data more carefully
- relaxes our OpenAPI spec, which weakens our automated tests
- if a user puts details like qualification level or place of study into the explanation, then this information is in the wrong place

The **mixed list** approach means that we're very explicit about what data we're providing, but consumers still have to pick through this data more carefully

The **munge** approach provides all the information with no guarantees of consistency (it's free text) but isn't easy for machines to read.

On the grounds that machines _probably_ don't need to read this, this PR spikes that approach.

Feedback welcome 🙂 

https://trello.com/c/uzwLb0jR/1436-display-reason-for-missing-gcses-in-provider-ui-and-in-api

### DIFF
--- a/config/vendor-api-v1.yml
+++ b/config/vendor-api-v1.yml
@@ -733,6 +733,7 @@ components:
         - gcses
         - degrees
         - other_qualifications
+        - missing_gcses_explanation
       properties:
         gcses:
           type: array
@@ -746,6 +747,12 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/Qualification"
+        missing_gcses_explanation:
+          type: string
+          nullable: true
+          maxLength: 8192 # Up to 800 words
+          description: If the candidate lacks any required GCSEs, this field will contain their free-text explanation of why this is the case.
+          example: 'Maths GCSE or equivalent: I will take Maths GCSE at my local training provider on 18th August 2020'
 
     WorkExperiences:
       type: object
--- a/config/vendor-api-v1.yml
+++ b/config/vendor-api-v1.yml
@@ -753,6 +753,7 @@ components:
       required:
         - jobs
         - volunteering
+        - breaks
       properties:
         jobs:
           type: array
@@ -764,6 +765,12 @@ components:
           description: The candidate’s experience as a volunteer
           items:
             $ref: "#/components/schemas/WorkExperience"
+        breaks:
+          type: string
+          nullable: true
+          description: The candidate’s explanation for any breaks in work history
+          maxLength: 4096 # 400 words
+          example: 'I spent time volunteering overseas...'
 
     WorkExperience:
       type: object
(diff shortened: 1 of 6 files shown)


config/vendor-api-v1.yml
```diff
@@ -746,6 +747,12 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/Qualification"
+        missing_gcses_explanation:
+          type: string
+          nullable: true
+          maxLength: 717 # 700 words: 3 x 200 (one per required GCSE) + 100 for our boilerplate
```
tvararu: The comment is a bit unclear. I think it should say `700 characters:`; but even then it's set to 717 which isn't explained. Should we also round it up so it looks a bit less arbitrary?

This is a nit, and can be ignored.

config/vendor-api-v1.yml
```diff
@@ -746,6 +747,12 @@ components:
...
+          description: If the candidate lacks any required GCSEs, this field will contain their free-text explanation of why this is the case.
+          example: 'Maths GCSE or equivalent: I will take an exam with equivalencytesting.com on 18th August 2019'
```
tvararu: TIL of this site, looks like a real company. By using it as our example, could it be interpreted as advertising for them? Wondering if that's OK or if they're a household name like UCAS so it wouldn't be noticed by anyone.

This is a nit, and can be ignored.

config/vendor-api-v1.yml
duncanjbrown: This is me being unnecessarily fussy: I've been trying to stick to powers-of-two elsewhere, so this should really be rounded up to the next available power of 2 (ie 1024). Will fix.

config/vendor-api-v1.yml
duncanjbrown: Good question. I got this from the designers. I'll replace it with some likely-sounding but fake domain name

duncanjbrown:
Thanks @tvararu!

config/vendor-api-v1.yml
tvararu: I was also going to suggest 1000 or 1024, whichever is rounder 😉

duncanjbrown:
Closing to re-raise as a proper PR

PR #1042, opened 2020-01-07
1434 fix updating references doesnt update timestamps in support

stevehook:
## Context

There is a discrepancy in the 'last updated at' columns between the application index pages in the Support and Provider interfaces. This is because Support interface displays `ApplicationForm#updated_at` whereas the Provider interface displays `ApplicationChoice#updated_at`.

It makes sense for the two interfaces to display forms and choices respectively so that is not something we want to change. Instead we attempt to keep the dates in sync.

- Add a `touch` attribute to the `ApplicationChoice#application_form` association.
- Add a migration to fixup existing `ApplicationForm#updated_at` values. 

## Guidance to review

- Is it reasonable to use `touch` to keep values in sync? It is still possible to see discrepancies if an application form is updated last.
- Is it reasonable to patch up existing data using a migration? I'm making an assumption here that we don't have too many applications at present (in any environment) for this to be fast enough.

## Link to Trello card

https://trello.com/c/pItvx0YF/1434-updating-references-doesnt-update-the-application-timestamps-in-support

## Things to check


stevehook:
Closing for now - there may be a better way to do this and it misses part of the point of the original card

PR #1052, opened 2020-01-07
Allow support users to add/remove provider users to/from organisations

tijmenb:
## Context

This adds functionality to allow support users to edit provider users, and add/remove any organisations they have access to.

Updated create page with checkboxes:

image

New edit page:

image

## Guidance to review

Try it out!

## Link to Trello card

https://trello.com/c/o1e8dgLA/1443-allow-support-users-to-add-remove-provider-users-to-from-organisations

## Things to check

### DIFF
--- a/app/views/support_interface/provider_users/edit.html.erb
+++ b/app/views/support_interface/provider_users/edit.html.erb
@@ -1,0 +1,35 @@
+<% content_for :before_content do %>
+  <div class="govuk-breadcrumbs">
+    <ol class="govuk-breadcrumbs__list">
+      <li class="govuk-breadcrumbs__list-item">
+        <%= link_to 'Users', support_interface_users_path, class: 'govuk-breadcrumbs__link' %>
+      </li>
+      <li class="govuk-breadcrumbs__list-item" aria-current="page">
+        <%= link_to 'Provider users', support_interface_provider_users_path, class: 'govuk-breadcrumbs__link' %>
+      </li>
+      <li class="govuk-breadcrumbs__list-item" aria-current="page">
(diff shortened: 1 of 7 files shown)


duncanjbrown:
What shall we do about `ProviderUser#provider` and its dependents?

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/app/models/provider_user.rb#L6-L8

Providers are unsorted, so it could be the case that adding an org means that they can no longer see applications for the org to which they were originally assigned. I think everyone who calls that method now needs to call `providers` so required changes will be quite far-reaching.

🏈 ?

app/views/support_interface/provider_users/edit.html.erb
```diff
@@ -0,0 +1,35 @@
...
+
+      <h1 class='govuk-heading-xl'>
+        Edit provider user
+      </h1>
+
+      <%= f.govuk_text_field :email_address, label: { text: 'Email address' } %>
+      <% blank_uid_notice = 'If you leave this empty, it will be automatically set when this email address signs in for the first time.' %>
+      <%= f.govuk_text_field :dfe_sign_in_uid, label: { text: 'DfE Sign-in UID' }, hint_text: blank_uid_notice %>
+
+      <%= f.govuk_collection_check_boxes :provider_ids, @form.available_providers, :id, :name_and_code, legend: { text: 'Provider' } %>
```
stevehook: I think the providers should be sorted alphabetically (here or elsewhere).

tijmenb: Done!

PR #1056, opened 2020-01-08
Update 'Change' links in work history for more context

chubberlisk:
## Context

We've been wanting to solve an accessibility issue that was reported in our external audit about `Change` links in sections that have multiple entries not having enough context when using the links list functionality of a screen reader. For example in the work history review page, when using the links list it would list multiple `Change job`, `Change type` links if a candidate has entered multiple jobs. This makes it difficult to know the purpose of the link when navigating out of context. 

We believed that adding `aria-describedby` would solve this issue (https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1032) but it turns out this isn't the case so we reverted it in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1053. To our knowledge, the only way to allow for more context for a link in the links list is by changing the content of the link.

This PR updates the `Change` links in work history to have more detailed visually hidden text by adding in the job title and name of employer. Also, if a candidate has more than one entries for the same role and employer, then the dates are added.

image

image

image

## Guidance to review

- I wasn't able to take a screenshot of the links list so try it out locally
- Is there a nice Ruby or Rails way to work out uniqueness based on attributes? Ended up with `any_jobs_with_same_role_and_organisation?(work)`
- Had to add ordering when displaying the jobs on the review to prevent flakey tests. Does that make sense?

## Link to Trello card

https://trello.com/c/XsqlA0hu/698-dac-page-33-add-hidden-content-to-change-links-on-work-history-review-and-other-pages

## Things to check

### DIFF
--- a/app/components/work_history_review_component.rb
+++ b/app/components/work_history_review_component.rb
@@ -59,7 +59,7 @@ private
     {
       key: 'Job',
       value: [work.role, work.organisation],
-      action: 'job',
+      action: generate_change_action(work: work, attribute: 'job title'),
       change_path: candidate_interface_work_history_edit_path(work.id),
     }
   end
@@ -68,7 +68,7 @@ private
     {
       key: 'Type',
       value: work.commitment.dasherize.humanize,
-      action: 'type',
+      action: generate_change_action(work: work, attribute: 'type'),
       change_path: candidate_interface_work_history_edit_path(work.id),
     }
   end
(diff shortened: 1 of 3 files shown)


app/components/work_history_review_component.rb
```diff
@@ -100,4 +100,17 @@ def formatted_end_date(work)
 
     work.end_date.to_s(:month_and_year)
   end
+
+  def generate_change_action(work:, attribute:)
+    if any_jobs_with_same_role_and_organisation?(work)
```
tvararu: If we didn't have this `if` branching logic, we could have used a `I18n.t('something.change_action', work: work, attribute: 'job title')` call instead of having to declare (and reuse) a `generate_change_action` method... given this, the current implementation is probably the best. 👍

PR #1059, opened 2020-01-08
Keep track of when an API key was last used

tijmenb:
## Context

This will be useful in determining who is using their API keys and makes it easier to revoke inactive keys.

The `last_used_at` will be updated on every API call. It's displayed in Support here:

image

## Guidance to review

See if it makes sense.

## Link to Trello card

Tangential to https://trello.com/c/jsBILlcB

## Things to check

- [x] ~This code doesn't rely on migrations in the same Pull Request~ It does, but I think it'll be alright since the web server starts after the migrations are run 


duncanjbrown:
:+1: — smol request spec to go with it?

tijmenb:
@duncanjbrown erm yeah

PR #1060, opened 2020-01-08
Use reference_form flag to sent chaser email with reference form link

tvararu:
## Context

We currently still send them a link to the Google form.

Also calculate referee token inside `reference_request_email` which fixes the preview emails, and add some test coverage.

## Guidance to review

First commit is just refactoring for `reference_request_email`.

Ignore whitespace changes.

## Link to Trello card

https://trello.com/c/WSSNWFwl/736-change-google-form-url-in-referee-chaser-email

## Things to check

### DIFF
--- a/spec/mailers/referee_mailer_spec.rb
+++ b/spec/mailers/referee_mailer_spec.rb
@@ -22,40 +22,57 @@ RSpec.describe RefereeMailer, type: :mailer do
     end
     let(:reference) { application_form.application_references.first }
     let(:candidate_name) { "#{application_form.first_name} #{application_form.last_name}" }
-    let(:mail) { mailer.reference_request_email(application_form, reference, 'reference-token') }
+    let(:mail) { mailer.reference_request_email(application_form, reference) }
 
-    before { mail.deliver_now }
+    describe 'with reference_form feature flag on' do
+      before do
+        FeatureFlag.activate('reference_form')
+        mail.deliver_now
+      end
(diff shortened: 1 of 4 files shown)


spec/mailers/referee_mailer_spec.rb
```diff
@@ -1,5 +1,6 @@
 require 'rails_helper'
 
+# rubocop:disable RSpec/NestedGroups
```
duncanjbrown: This cop seems sensible (https://github.com/rubocop-hq/rubocop-rspec/issues/148). But we're only in this position because we've got a feature flag on, and that'll go away. So :+1: to ignoring it.

duncanjbrown:
> Code all looks good. Just one (probably dumb) question.
> 
> When we send a chaser email does it invalidate the link in the original email? (because it updates the token). If so I guess that's fine but what about if we generate a preview message - does that also invalidate the last sent token?

I guess it has to, or we'd need to store the unhashed token somewhere (which we won't)?

spec/mailers/referee_mailer_spec.rb
tvararu: Exactly my thinking 👍

tvararu:
@stevehook 

> If so I guess that's fine but what about if we generate a preview message

That's only done locally I believe; yes, it will invalidate, but I think that's of limited impact if it happens locally, right?

PR #1070, opened 2020-01-09
Referee can refuse to give feedback to the candidate.

alexdesi:
## Context

This PR allows the referee to refuse to give feedback to the candidate.

- A new action `refuse_to_give_feedback` in `ReferenceController`
- Update the email with the link to refuse  to give feedback

## Guidance to review
start from `referee_can_submit_a_reference_for_candidate_spec.rb`

## Link to Trello card

https://trello.com/c/3kuEKaQH/666-update-email-content-for-referee-request

### DIFF
--- a/app/controllers/referee_interface/reference_controller.rb
+++ b/app/controllers/referee_interface/reference_controller.rb
@@ -4,16 +4,13 @@ module RefereeInterface
     before_action :add_identity_to_log
     before_action :check_referee_has_valid_token
     before_action :set_token_param
+    before_action :show_finished_page_if_feedback_provided, except: %i[confirmation confirm_consent]
 
     layout 'application'
 
     def feedback
-      if reference.feedback_requested?
-        @application = reference.application_form
-        @reference_form = ReferenceFeedbackForm.new(reference: reference)
-      else
-        render :finish
-      end
+      @application = reference.application_form
+      @reference_form = ReferenceFeedbackForm.new(reference: reference)
     end
 
     def submit_feedback
@@ -41,7 +38,28 @@ module RefereeInterface
       render :finish
     end
 
+    def refuse_feedback
+      @application = reference.application_form
+      @reference = reference
+    end
+
+    def confirm_feedback_refusal
+      case params.dig(:application_reference, :refuse_to_give_feedback)
+      when 'yes'
+        reference.update!(feedback_status: 'feedback_refused')
+        redirect_to referee_interface_confirmation_path(token: @token_param)
+      when 'no'
+        redirect_to referee_interface_reference_feedback_path(token: params[:token])
+      end
+    end
+
   private
+
+    def show_finished_page_if_feedback_provided
+      return if reference.feedback_requested?
+
+      render :finish
+    end
 
     def add_identity_to_log
       return if reference.blank?
--- a/app/mailers/referee_mailer.rb
+++ b/app/mailers/referee_mailer.rb
@@ -5,8 +5,9 @@ class RefereeMailer < ApplicationMailer
     @candidate_name = application_form.full_name
 
     if FeatureFlag.active?('reference_form')
-      @reference_link = referee_interface_reference_feedback_url(token: reference.update_token!)
-      @decline_reference_link = 'link_to_decline_the_reference' #TODO: implement the flow for Referee decline to give references
+      unhashed_token = reference.update_token!
+      @reference_link = referee_interface_reference_feedback_url(token: unhashed_token)
+      @refuse_to_give_feedback_link = referee_interface_refuse_feedback_url(token: unhashed_token)
       template_name = :reference_request_email
     else
       @reference_link = google_form_url_for(@candidate_name, @reference)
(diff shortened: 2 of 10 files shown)


app/controllers/referee_interface/reference_controller.rb
```diff
@@ -38,7 +38,13 @@ def confirmation; end
     def confirm_consent
       consent_to_be_contacted = params.dig(:application_reference, :consent_to_be_contacted)
 
-      reference.update!(consent_to_be_contacted: consent_to_be_contacted)
+      reference.update!(consent_to_be_contacted: consent_to_be_contacted, refuse_to_give_feedback: false)
+
+      render :finish
+    end
+
+    def refuse_to_give_feedback
```
tijmenb: `GET` requests should never change something, I think we should have a page here that allows the referee to confirm that they don't want to give feedback. Do we have designs for the feature?

app/mailers/referee_mailer.rb
```diff
@@ -5,8 +5,9 @@ def reference_request_email(application_form, reference)
     @candidate_name = application_form.full_name
 
     if FeatureFlag.active?('reference_form')
-      @reference_link = referee_interface_reference_feedback_url(token: reference.update_token!)
-      @decline_reference_link = 'link_to_decline_the_reference' #TODO: implement the flow for Referee decline to give references
+      unhashed_token = reference.update_token!
+      @reference_link = referee_interface_reference_feedback_url(token: unhashed_token)
+      @refuse_to_give_feedback_link = referee_interface_refuse_to_give_feedback_url(token: unhashed_token)
```
tijmenb: Perhaps take the opportunity to move the calls to the URL helpers into the template, and only pass in the `@unhashed_token`.

db/migrate/20200109111923_add_refuse_to_give_feedback_to_reference.rb
```diff
@@ -0,0 +1,5 @@
+class AddRefuseToGiveFeedbackToReference < ActiveRecord::Migration[6.0]
+  def change
+    add_column :references, :refuse_to_give_feedback, :boolean
```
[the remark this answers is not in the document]
alexdesi: Yes, it might be a good idea.
Just, the card did mention anything about:
- Didn't respond in time
- Email failed
also, sometimes you need to combine the various states:
when `Feedback provided`
`consent_to_be_contacted` can be true or false

It does sounds a refactor when will implement the cases **Email failed**
and **Didn't respond in time**

app/controllers/referee_interface/reference_controller.rb
alexdesi: You're right, but no design atm. In the first implementation, I assumed that it redirect to the thank you page.

Yesterday I've put a comment to ask for more info about this on trello: https://trello.com/c/3kuEKaQH/666-update-email-content-for-referee-request

db/migrate/20200109111923_add_refuse_to_give_feedback_to_reference.rb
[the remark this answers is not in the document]
tijmenb: Sorry, my initial comment wasn't very clear!

I've added a `feedback_status` in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1077. We should add a `feedback_refused` status to the enum and use that instead of the boolean.

app/mailers/referee_mailer.rb
tijmenb: Let's do this in another PR.

app/controllers/referee_interface/reference_controller.rb
```diff
@@ -41,8 +38,29 @@ def confirm_consent
       render :finish
     end
 
+    def refuse_feedback
+      @application = reference.application_form
+      @reference = reference
+    end
+
+    def confirm_feedback_refusal
+      case params[:application_reference][:refuse_to_give_feedback]
```
tvararu: Do we want to `params.dig`?

app/mailers/referee_mailer.rb
```diff
@@ -24,8 +25,9 @@ def reference_request_chaser_email(application_form, reference)
     @reference = reference
     @candidate_name = application_form.full_name
 
+    @token = reference.update_token!
```
tvararu: It's not obvious to me why we are turning this to an instance variable, what's up? Not blocking anyway.

spec/system/referee_interface/referee_refuses_to_give_a_reference_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+  def then_i_receive_an_email_with_a_reference_request
+    open_email('terri@example.com')
+  end
+
+  def when_i_click_the_refuse_reference_link_in_the_email
+    current_email.click_link(refuse_feedback_url)
+  end
+
+  def and_i_say_that_i_do_actually_want_to_give_a_reference
+    choose 'I\'d like to provide a reference'
```
tvararu: These can be translation strings if you like, not blocking 👍

app/mailers/referee_mailer.rb
tijmenb: It's used inside the email. I'll look into tidying up the emails in a separate PR.

app/controllers/referee_interface/reference_controller.rb
tijmenb: Done

PR #1082, opened 2020-01-10
Clean up nil study_mode course_options

mnacos:
## Context

During the database migration converting `study_mode` from `int` to `string` (https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/883/files), `CourseOption` records were not converted to 'full_time', instead their value `0` was converted to `'0'`, which the rails model cannot map to the enum values and interprets as `nil`. The `SyncFromFind` tasks then ignore these records when reconciling courses and updating course options, creating records which appear as duplicate sites during the application process.

Add a temporary `CleanUpCourseOptionsWithNoStudyMode` worker we can either run from `rails c` or schedule through clockwork.

## Guidance to review

Does the code in the worker make sense?

## Link to Trello card

[1471 - Fix duplicated data from syncing courses and study_mode change](https://trello.com/c/SBaXvzJj)

## Things to check

### DIFF
--- a/app/workers/clean_up_course_options_with_no_study_mode.rb
+++ b/app/workers/clean_up_course_options_with_no_study_mode.rb
@@ -1,0 +1,23 @@
+class CleanUpCourseOptionsWithNoStudyMode
+  include Sidekiq::Worker
+
+  def perform
+    CourseOption.where('study_mode = \'0\'').each do |nil_co|
+      if nil_co.application_choices.any?
+        equivalent_course_option = CourseOption.where(
+          course: nil_co.course,
+          site: nil_co.site,
+          study_mode: :full_time,
+        ).first
+        if equivalent_course_option
+          nil_co.application_choices.each do |application_choice|
+            application_choice.update!(course_option_id: equivalent_course_option.id)
+          end
+          nil_co.destroy!
+        end
+      else
+        nil_co.destroy!
+      end
+    end
+  end
+end
--- a/app/workers/clean_up_course_options_with_no_study_mode.rb
+++ b/app/workers/clean_up_course_options_with_no_study_mode.rb
@@ -2,22 +2,28 @@ class CleanUpCourseOptionsWithNoStudyMode
   include Sidekiq::Worker
 
   def perform
-    CourseOption.where('study_mode = \'0\'').each do |nil_co|
-      if nil_co.application_choices.any?
-        equivalent_course_option = CourseOption.where(
-          course: nil_co.course,
-          site: nil_co.site,
-          study_mode: :full_time,
-        ).first
-        if equivalent_course_option
-          nil_co.application_choices.each do |application_choice|
-            application_choice.update!(course_option_id: equivalent_course_option.id)
-          end
-          nil_co.destroy!
+    CourseOption.where('study_mode IN (\'0\',\'1\')').find_each do |nil_co|
+      intended_study_mode = case nil_co.study_mode_before_type_cast
+                            when '0'; :full_time
+                            when '1'; :part_time
+                            end
+      equivalent_course_option = find_equivalent_course_option nil_co, intended_study_mode
+      if equivalent_course_option
+        nil_co.application_choices.each do |application_choice|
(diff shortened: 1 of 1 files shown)


app/workers/clean_up_course_options_with_no_study_mode.rb
```diff
@@ -0,0 +1,23 @@
+class CleanUpCourseOptionsWithNoStudyMode
+  include Sidekiq::Worker
+
+  def perform
+    CourseOption.where('study_mode = \'0\'').each do |nil_co|
```
stevehook: One I wanted to double check. Could there be other invalid values in the database? e.g. `study_move == '1'` etc.?

As a sanity check I would be tempted to check for any values that are not `IN ('full_time', 'part_time')`

If there are part_time courses in the system then I guess we need to handle those too, otherwise what you have is fine.

stevehook: I would change `each` to `find_each` when processing a big dataset like this to avoid loading all the models into memory at once but it's probably not big enough to worry about here.

app/workers/clean_up_course_options_with_no_study_mode.rb
```diff
@@ -0,0 +1,23 @@
...
+          course: nil_co.course,
+          site: nil_co.site,
+          study_mode: :full_time,
+        ).first
+        if equivalent_course_option
+          nil_co.application_choices.each do |application_choice|
+            application_choice.update!(course_option_id: equivalent_course_option.id)
+          end
+          nil_co.destroy!
+        end
```
stevehook: Do we need to flag some kind of warning if there is no equivalent?

Or just fix it up so that `nil_co.study_mode` is set to `full_time`

app/workers/clean_up_course_options_with_no_study_mode.rb
```diff
@@ -0,0 +1,23 @@
+class CleanUpCourseOptionsWithNoStudyMode
+  include Sidekiq::Worker
+
+  def perform
+    CourseOption.where('study_mode = \'0\'').each do |nil_co|
+      if nil_co.application_choices.any?
+        equivalent_course_option = CourseOption.where(
```
tvararu: `find_by` instead of `where`?

app/workers/clean_up_course_options_with_no_study_mode.rb
mnacos: Historically there shouldn't be any part time courses, but you're right. I've updated the code to handle 'nil' part time courses, too ('1').

app/workers/clean_up_course_options_with_no_study_mode.rb
mnacos: I like this suggestion, if no duplicate record exists for this course option, then we set it to the original intended value (we have this information based on the raw db value, which can be '0' or '1').

app/workers/clean_up_course_options_with_no_study_mode.rb
mnacos: I think `where` is better here, because I want all of them, plus not sure `find_by` works with sql fragments.

app/workers/clean_up_course_options_with_no_study_mode.rb
```diff
@@ -0,0 +1,29 @@
...
+      end
+    end
+  end
+
+  def find_equivalent_course_option(nil_co, intended_study_mode)
+    CourseOption.where(
+      course: nil_co.course,
+      site: nil_co.site,
+      study_mode: intended_study_mode,
+    ).first
```
tvararu: What I meant in my previous comment: if this is `where.first`, isn't `find_by` equivalent? 🤔

tvararu: Not sure re: SQL fragments!

mnacos: Ah sorry, thought you suggested `find_by` for the loop `where`, not the equivalent one.

PR #1084, opened 2020-01-10
703 DAC Degrees page only ask for one thing per page

C-gyorfi:
## Context

This is the first slice of encompassing all areas of the form where we have reverted to single-page forms instead to make these forms more accessible. 

This PR splits the `Degrees` page into 3 separate pages so that the page ask less information per page. Doing this by adding separate routes, views and controllers for `degree/grade` and `degree/year` pages. 

### Before
image

### After
image

image

image

## Guidance to review

The system tests are good to start with, otherwise visit the degree form and try to `fill-up a new degree` complete each page and then `edit` the fields from the `review page`. 
Please note, there is currently a bug where you can [submit an application without having a degree](https://trello.com/c/vAOxWeBb/735-submitting-an-application-without-a-degree), this bug still exists after this change and will be addressed as a separate slice

## Link to Trello card

https://trello.com/c/QJVmXusT/703-dac-page-51-change-all-pages-to-only-ask-for-one-thing-per-page

## Things to check


spec/models/candidate_interface/degree_form_spec.rb
```diff
@@ -273,38 +249,53 @@
...
 
-      expect(degree.update(application_form)).to eq(true)
-      expect(application_form.application_qualifications.degree.first)
-        .to have_attributes(grade: 'First', predicted_grade: true)
+        expect(degree.update_year(application_form)).to eq(true)
+        expect(application_form.application_qualifications.degree.first)
+          .to have_attributes(award_year: '2000')
+      end
     end
   end
```
tvararu: Missing `describe '#update_grade'` I think

C-gyorfi: Its just above :) https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/d126832d9ff9d263020b4f7dd0d3e6aacf594008/spec/models/candidate_interface/degree_form_spec.rb#L268

tvararu: I checked twice but missed it! 👍

PR #1086, opened 2020-01-10
Sync all providers, courses and sites from Find API

stevehook:
## Context

We pull data about providers, courses and sites from the Find API. We currently have a hard-coded list of 12 providers whose data we sync from the Find API (on a background Sidekiq worker triggered by Clock every 15 minutes).

In order to rollout to more providers we want to sync data from Find for all providers. Initial experiments showed that syncing the whole dataset would take around an hour or more. Longer term we will probably need to do that by splitting the import job into multiple jobs.

However the immediate need is to onboard a small proportion of new providers more easily so this PR avoids the complexity of multiple jobs by importing all of the basic provider information but only importing course data for those that have been opted in to course data syncing.

Migration was done in an earlier PR - https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1076 . The earlier PR is merged so this PR is now ready to merge/deploy as soon as it passes review.

- Add a new button to the support provider show page to toggle the `sync_courses` option. (similar to the existing 'enable courses' button).
- Display a _Not synced with Find_ label on the support providers index page to indicate which providers have not been opted in yet.
- Add a new worker to sync all providers by making an API call to 
- Adapt the existing syncing services to respect the `sync_courses` option.

## Guidance to review

- Is the user interface consistent/sensible? (Not sure about having multiple buttons)
- Is having 1900 odd providers going to cause usability issues anywhere on the system? The full list is rendered in a couple of places in the support interface but it's not terribly slow (check out the provider index page and the new provider user page). The candidate facing pages don't (as far as I know) render a full list of providers (they are filtered according to whether they have an available course or not).
- At the moment syncing takes about 1 minute. That's fine for now and as we roll out smallish numbers of providers over the coming weeks, but when we have to sync all 17K courses for all the providers we'll need to restructure the background jobs.

provider-source-syncing

## Link to Trello card

https://trello.com/c/Pk4LBeNl/1452-pull-all-providers-courses-from-find-%F0%9F%8F%88

## Things to check


app/components/support_interface/provider_sync_courses_toggle_component.rb
```diff
@@ -0,0 +1,29 @@
...
+  class ProviderSyncCoursesToggleComponent < ActionView::Component::Base
+    include ViewHelper
+
+    def initialize(provider:)
+      @provider = provider
+    end
+
+    def toggle_path
+      if provider.sync_courses?
+        support_interface_disable_provider_course_syncing_path(provider)
```
tijmenb: What would be the effect of disabling sync for a provider? If we don't delete existing courses, previously synced ones will get out of date as some point, which might not be obvious. 

Would it be safer to make enabling courses a one-way operation?

app/workers/sync_from_find.rb
```diff
@@ -3,7 +3,7 @@ class SyncFromFind
 
   def perform
     Rails.configuration.providers_to_sync[:codes].each do |code|
-      SyncProviderFromFind.call(provider_code: code)
+      SyncProviderFromFind.call(provider_code: code, sync_courses: true)
```
mnacos: This sets `sync_courses` according to the Rails provider config, which is great, but I can't tell when `SyncFromFind` is actually being called.

app/components/support_interface/provider_sync_courses_toggle_component.rb
stevehook: Yes I think you are right, it should be one way. Will remove the disable syncing option for now.  

I thought about showing the button if and only if there are no associated courses but then you could get into a race condition with a background worker that just started a sync run.

app/workers/sync_from_find.rb
stevehook: I don't think it is being called at the moment. I had left it half thinking that we might want to manually sync the old way but that doesn't really make sense anymore. If there are no objections I think I should remove it completely?

stevehook: Another thing I've just realised is that `setup_local_dev_data` doesn't enable syncing on the test provider so I need to change that too.

app/services/sync_provider_from_find.rb
```diff
@@ -10,22 +14,30 @@ def self.call(provider_code:)
...
+      provider.sync_courses = sync_courses
+      provider.save!
+    end
+    update_provider(provider, provider_name) if provider_name.present?
 
     provider
   end
 
-  def self.create_course(find_course, provider)
+  def self.update_provider(provider, provider_name)
```
duncanjbrown: This method feels quite specialised for its name

spec/system/support_interface/provider_sync_courses_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    given_i_am_a_support_user
+    and_a_provider_exists
+    when_i_visit_the_providers_page
+    then_i_see_that_the_provider_is_not_configured_to_sync_courses
+
+    when_i_click_on_a_provider
+    then_i_see_that_course_syncing_is_off
+
+    when_i_click_on_the_enable_course_syncing_button
+    then_i_see_that_course_syncing_is_on
```
duncanjbrown: Not intended as a PR bomb, but... would it be nice to spec that the courses were indeed synced when the cron ran?

app/workers/sync_from_find.rb
mnacos: Yes please either use or remove.

spec/system/support_interface/providers_spec.rb
```diff
@@ -42,6 +42,9 @@ def when_i_visit_the_providers_page
   def and_providers_are_configured_to_be_synced
     allow(Rails.application.config).to receive(:providers_to_sync)
```
duncanjbrown: Do we need a card to remove the YAML etc?

mnacos: Was just chatting with Steve about this. It sounds like we only currently need this for the local test data generation, so it can remain for now, but we may need a different mechanism to customise sandbox environments, not through a yaml file but through something more pipeline specific.

spec/services/sync_provider_from_find_spec.rb
```diff
@@ -12,7 +12,47 @@
...
+
+      provider = Provider.find_by_code('ABC')
+      expect(provider).to be_present
+      expect(provider.courses).to be_blank
+    end
+  end
+
+  describe 'ingesting an existing provider not configured to sync courses' do
+    before do
+      @existing_provider = create :provider, code: 'ABC', name: 'DER College', sync_courses: false
```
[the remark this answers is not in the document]
stevehook: No it doesn't, removing

app/services/sync_provider_from_find.rb
stevehook: Looks like it should be `update_provider_name`. Changing...

spec/system/support_interface/providers_spec.rb
stevehook: Yes it's still used for the `setup_local_dev_data` rake task. We could have a card to come up with a better way to do that, maybe that would be covered by the sandbox work?

spec/services/sync_provider_from_find_spec.rb
```diff
@@ -12,7 +12,47 @@
...
+        findable: true,
+      )
+
+      SyncProviderFromFind.call(provider_name: 'ABC College', provider_code: 'ABC')
+
+      expect(@existing_provider.reload.courses).to be_blank
+    end
+  end
+
+  describe 'ingesting an existing provider configured to sync courses, sites and course_options' do
```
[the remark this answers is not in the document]
stevehook: Yes `sync_courses` also imports the `sites` and `course_options`

spec/system/support_interface/provider_sync_courses_spec.rb
stevehook: :+1: I've had a go at this - invoked clock from the existing system spec (with a stubbed API response) and can now see the newly import course. Let me know if you had something else in mind!

config/locales/provider_sync_courses.yml
```diff
@@ -0,0 +1,8 @@
+en:
+  provider_sync_courses:
+    button_label:
+      enable: Enable course syncing from Find
+      disable: Disable course syncing from Find
```
duncanjbrown: I think this string can go now that turning things on is one-way?

PR #1092, opened 2020-01-13
Add page for choosing which email to send for a reference on Support

chubberlisk:
## Context

Currently, we can send a chase email to a reference and a candidate from Support but we're going to  need to send more emails to a reference such as requesting a new referee when one hasn't responded.

## Changes proposed in this pull request (Updated: 14:18)

This PR changes the flow a bit for sending a chase email by changing the `Chase referee` link to `Send email` and then adding a page for choosing which email to send between the confirmation page. The additional page is to allow for other emails to be sent from here.

Added in a `send_reference_email_via_support` feature flag so this flow is only shown when it is activated.

### Before

image

### After

image

image

image

## Guidance to review

Nothing in particular. The content likely needs a change but we can iterate on this.

## Link to Trello card

https://trello.com/c/fDuObKQB/719-send-all-standard-not-zendesk-emails-from-support

## Things to check

### DIFF
--- a/app/views/support_interface/send_reference_email/new.html.erb
+++ b/app/views/support_interface/send_reference_email/new.html.erb
@@ -1,0 +1,20 @@
+<% content_for :browser_title, title_with_error_prefix('Send email', @send_reference_email.errors.any?) %>
+<% content_for :before_content, govuk_back_link_to(support_interface_application_form_path(@reference.application_form), 'Back to application') %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with model: @send_reference_email, url: support_interface_send_reference_email_path(@reference), method: :post do |f| %>
+      <%= f.govuk_error_summary %>
+
+      <h1 class="govuk-heading-xl">
+        Send email
+      </h1>
+
+      <%= f.govuk_radio_buttons_fieldset :choice, legend: { text: 'Select the action you want to take on the reference' } do %>
+        <%= f.govuk_radio_button :choice, :chase, label: { text: t('application_form.referees.chase') } %>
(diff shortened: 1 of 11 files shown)


adamsilver:
Nice one @chubberlisk — it's not great that there's just one radio button, but I presume that's because the other options are coming imminently in other tickets?

If it's not a pain, may I suggest the existing design is kept until you add the second and third options etc?

chubberlisk:
> Nice one @chubberlisk — it's not great that there's just one radio button, but I presume that's because the other options are coming imminently in other tickets?
> 
> If it's not a pain, may I suggest the existing design is kept until you add the second and third options etc?

Ah, yes @adamsilver! The other options are coming imminently - the first to be added will be the new referee request after a referee hasn't responded.

Completely makes sense! I can add in a feature flag which we can turn on when another email is added.

app/views/support_interface/send_reference_email/new.html.erb
```diff
@@ -0,0 +1,20 @@
+<% content_for :browser_title, title_with_error_prefix('Send email', @send_reference_email.errors.any?) %>
+<% content_for :before_content, govuk_back_link_to(support_interface_application_form_path(@reference.application_form), 'Back to application') %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with model: @send_reference_email, url: support_interface_send_reference_email_path(@reference), method: :post do |f| %>
+      <%= f.govuk_error_summary %>
+
+      <h1 class="govuk-heading-xl">
+        <%= 'Send email' %>
```
tvararu: Unnecessary erb?

```suggestion
        Send email
```

PR #1093, opened 2020-01-13
Fix edit applications to not redirect loop unsubmitted applications

tvararu:
## Context

Turning on the edit applications feature flag breaks unsubmitted applications.

Add a test and fix the underlying issue.

Solving the bug involved just adding `&& current_application.submitted?` to the if clause, but then I realised both the `if` and `else` branches checked for `current_application.submitted?`. So I realised I could refactor to this.

## Guidance to review

Makes sense?

## Link to Trello card

https://trello.com/c/KZvEIjBk/742-candidate-with-unsubmitted-application-cant-sign-in-when-edit-application-feature-flag-is-turned-on

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/candidate_interface_controller.rb
+++ b/app/controllers/candidate_interface/candidate_interface_controller.rb
@@ -10,11 +10,7 @@ module CandidateInterface
   private
 
     def redirect_to_dashboard_if_not_amendable
-      if FeatureFlag.active?('edit_application')
-        redirect_to candidate_interface_application_complete_path unless current_application.amendable?
-      elsif current_application.submitted?
-        redirect_to candidate_interface_application_complete_path
-      end
+      redirect_to candidate_interface_application_complete_path if current_application.submitted? && !current_application.amendable?
     end
 
     def redirect_to_application_form_unless_submitted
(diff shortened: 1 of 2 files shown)


app/controllers/candidate_interface/candidate_interface_controller.rb
```diff
@@ -10,11 +10,7 @@ class CandidateInterfaceController < ActionController::Base
   private
 
     def redirect_to_dashboard_if_not_amendable
-      if FeatureFlag.active?('edit_application')
-        redirect_to candidate_interface_application_complete_path unless current_application.amendable?
-      elsif current_application.submitted?
-        redirect_to candidate_interface_application_complete_path
-      end
+      redirect_to candidate_interface_application_complete_path if current_application.submitted? && !current_application.amendable?
```
tijmenb: This doesn't have the feature flag anymore, are we sure the logic is still the same?

tvararu: Yes! This is `#amendable?`:

```ruby
  def amendable?
    FeatureFlag.active?('edit_application') && submitted? && within_amend_period?
  end
```

The feature flag was redundant.

PR #1097, opened 2020-01-13
Fix incorrect updated_at timestamp for application forms

tijmenb:
## Context

We currently don't updated the application form's `updated_at` when adding/updating/removing associated models like choices, work experience, references, etc.

Is causes:

- the filtering in support & provider UI to be semi-broken
- the "haven’t started" stat on performance dashboard to be wrong
- the "Last updated" time on the candidate dashboard to be semi-broken
 
Use `touch: true` to update to form whenever an associated model changes.

## Link to Trello card

https://trello.com/c/FceP2pDP/1497-fix-incorrect-updatedat-timestamp-for-application-forms

## Things to check

### DIFF
--- a/app/models/application_choice.rb
+++ b/app/models/application_choice.rb
@@ -1,7 +1,7 @@
 class ApplicationChoice < ApplicationRecord
   before_create :set_initial_status
 
-  belongs_to :application_form
+  belongs_to :application_form, touch: true
   belongs_to :course_option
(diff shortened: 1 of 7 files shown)


tijmenb:
I've removed the data migration because the data change yesterday has meant that almost all application forms have an new updated_at, so it wouldn't fix anything.

PR #1100, opened 2020-01-14
Fix provider authentication email case sensitivity issues

stevehook:
## Context

The main issue here is that if a Provider is added to our system with an email address that differs only by case with their DfE Sign-in email address then they are unable to login.

- Ensure that all email addresses (for Providers and Support users) are stored as lower case.
- Migration to fix any mixed case email addresses that may have already got into the database.
- Migration also adds a unique index to `support_users.email_address` (there is already one on `provider_users.email_address`)
- Convert DfE Sign-in email addresses to lower case before doing the lookup in our system.

## Guidance to review

- Are model callbacks the right way to do the downcasing of `email_address` attributes?
- Note that the fixup migration will fail if there are multiple email addresses in the system already that differ only by case. I think we need to do some manual checks on the production DB to work out whether this is going to be a problem and make necessary changes if needed _before deploying_.

## Link to Trello card

https://trello.com/c/vElEdTfv/1475-provider-authentication-by-email-is-case-sensitive

## Things to check

- [x] This code doesn't rely on migrations in the same Pull Request (There is a migration but it's not a dependency - it's a fixup script)


duncanjbrown:
Looking good so far!

stevehook:
There don't seem to be any duplicates on production:

```
irb(main):011:0> ProviderUser.pluck(:email_address).combination(2).select { |a, b| a.downcase == b.downcase }
=> []
irb(main):012:0> SupportUser.pluck(:email_address).combination(2).select { |a, b| a.downcase == b.downcase }
=> []
```

We have one support user duplicate on `t01` and one provider user duplicate on `d01`. Other environments are clear.

stevehook:
Duplicates have now been removed from `t01` and `d01` so going to make this ready to review.

spec/models/provider_user_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+      provider_user = create :provider_user, dfe_sign_in_uid: nil
+      dsi_user = DfESignInUser.new(
+        email_address: provider_user.email_address,
+        dfe_sign_in_uid: 'ABC123',
+      )
+      ProviderUser.onboard!(dsi_user)
+      expect(provider_user.reload.dfe_sign_in_uid).to eq 'ABC123'
+    end
+
+    it 'sets the DfE Sign-in ID on an existing user with a mixed case DfE Sign-in email' do
```
mnacos: nice

stevehook:
> Looks great! How can we make sure no duplicates appear in the database between now and deploying this to prod? I guess the risk is low, may be worth checking for duplicates again just before we deploy.

Good point, I'll check again.

PR #1101, opened 2020-01-14
Render strings without underscores for Reference feedback status

duncanjbrown:
### Before

Screenshot 2020-01-14 at 11 22 28

### After

Screenshot 2020-01-14 at 11 22 00

### DIFF
--- a/app/components/reference_with_feedback_component.rb
+++ b/app/components/reference_with_feedback_component.rb
@@ -29,7 +29,9 @@ private
   def status_row
     {
       key: 'Feedback status',
-      value: render(TagComponent, text: feedback_status, type: 'blue'),
+      value: render(TagComponent,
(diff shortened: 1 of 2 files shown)


app/components/reference_with_feedback_component.rb
```diff
@@ -29,7 +29,10 @@ def rows
   def status_row
     {
       key: 'Feedback status',
-      value: render(TagComponent, text: feedback_status, type: 'blue'),
+      value: render(TagComponent, {
+        text: ApplicationReference.human_attribute_name("feedback_status.#{feedback_status}"),
```
tijmenb: TIL `human_attribute_name`!

fofr: +1

PR #1105, opened 2020-01-14
Allow support users to add provider users to multiple orgs

tijmenb:
## Context

In https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1052 we added the ability to add provider users to multiple orgs, but we had to disable it temporarily until we had design.

First, provider users can be added to multiple orgs now.

Provider users can then see all of their orgs' applications:

image

When viewing the application, there's a new block with previously missing info, including the provider:

image
  
## Guidance to review

See if it make sense.

## Link to Trello card

https://trello.com/c/o1e8dgLA/1443-allow-support-users-to-add-remove-provider-users-to-from-organisations

## Things to check

### DIFF
--- a/app/views/provider_interface/application_choices/show.html.erb
+++ b/app/views/provider_interface/application_choices/show.html.erb
@@ -14,7 +14,7 @@
 <% end %>
 
 <h1 class="govuk-heading-xl">
-  <span class="govuk-caption-xl">Application for <%= @application_choice.course.name_and_code %></span>
+  <span class="govuk-caption-xl">Application for <%= @application_choice.provider.name_and_code %> - <%= @application_choice.course.name_and_code %></span>
   <%= @application_choice.application_form.full_name %>
 </h1>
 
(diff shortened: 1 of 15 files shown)


app/views/provider_interface/application_choices/show.html.erb
```diff
@@ -28,6 +28,16 @@
...
 
+<h2 class="govuk-heading-l govuk-!-margin-top-8" id="course-info">Course</h2>
+
+<%= render SummaryListComponent, rows: [
+  { key: 'Provider', value: @application_choice.provider.name_and_code },
+  { key: 'Course', value: @application_choice.course.name_and_code },
+  { key: 'Cycle', value: @application_choice.course.recruitment_cycle_year },
+  { key: 'Preferred location', value: @application_choice.site.name_and_code },
+  { key: 'Study mode', value: @application_choice.course_option.study_mode.humanize },
+] %>
```
duncanjbrown: :+1: to this partial duplication of the Course Choice component from the Support UI

app/views/provider_interface/application_choices/show.html.erb
```diff
@@ -14,7 +14,7 @@
 <% end %>
 
 <h1 class="govuk-heading-xl">
-  <span class="govuk-caption-xl">Application for <%= @application_choice.course.name_and_code %></span>
+  <span class="govuk-caption-xl">Application for <%= @application_choice.provider.name_and_code %> - <%= @application_choice.course.name_and_code %></span>
```
duncanjbrown: If this title were a component then it would be possible to share the "only show provider name when user has access to multiple providers" logic here and also take some logic out of that view, which might be nice?

tijmenb: I'm inclined to keep it like it is, to avoid having many permutations of the application page. Things become simpler if all application choice pages look the same.

PR #1106, opened 2020-01-14
Refactor `CandidateMailer#new_referee_request` to handle other two emails

chubberlisk:
## Context

See https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1088, https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1092 and https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1103.

We want to add in two more new referee request emails: one when they've refused and one when a referee's email has bounced.

This PR refactors `CandidateMailer#new_referee_request` to handle the two other emails.

## Guidance to review

😕Worried about the readability of this and wondering if it's better to have this 3 separate methods and not worry about the duplication of content? If so, method name suggestions please!

## Link to Trello card

https://trello.com/c/fDuObKQB/719-send-all-standard-not-zendesk-emails-from-support

## Things to check

### DIFF
--- a/app/views/candidate_mailer/new_referee_request.text.erb
+++ b/app/views/candidate_mailer/new_referee_request.text.erb
@@ -2,8 +2,6 @@ Dear <%= @candidate_name %>,
 
 # Give details of a new referee
 
-<%= t('new_referee_request.not_responded.explanation', referee_name: @referee_name) %>
-
-Reply to this email with the name and email address of a new referee, and tell us how you know them.
+<%= t("new_referee_request.#{@reason}.explanation", referee_name: @referee.name, referee_email: @referee.email_address) %>
 
 We can’t send your application to your teacher training providers without 2 complete references.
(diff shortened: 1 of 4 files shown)


app/views/candidate_mailer/new_referee_request.text.erb
```diff
@@ -2,8 +2,8 @@ Dear <%= @candidate_name %>,
 
 # Give details of a new referee
 
-<%= t('new_referee_request.not_responded.explanation', referee_name: @referee_name) %>
+<%= @explanation %>
 
-Reply to this email with the name and email address of a new referee, and tell us how you know them.
+<%= @reply %>
```
tijmenb: Can `@explanation` and `@reply` be merged? 

Also, I think it's cleaner to call the `t()` code inside the view rather than the mailer class. WDYT?

chubberlisk: Yes to both! Definitely cleaner.

chubberlisk: Refactored now, some much better now. Thank you!

PR #1115, opened 2020-01-15
Add request access form to "Your account is not ready yet" provider view

mnacos:
### Context

It is possible for a user to be signed in through DfE Sign In but not authorised to access the provider console. The main focus of this PR is to provide a form for provider users not yet authorised to use the management console to help them request access and streamline the onboarding process.

Requests for access should appear in the support console, so an minimal interface will be needed for this. New requests should also appear as Slack notifications in the appropriate channels.

This PR adds an `ProviderUserAccessRequest` model to the app. It includes `:dfe_sign_in_uid` and `:email_address` of the `ProviderUser` requesting access.

`ProviderUserAccessRequest` includes a `processed_at` which is also used for determining which access request are still pending. `processed_by_support_user_id` references the support user id who has granted or declined the request. `decision` can be either `:granted` or `:declined`. [Should provider users also be able to provide access?]

If the request is granted, a `ProviderUser` with the corresponding `:dfe_sign_in_uid` and `:email_address` is created on the system and a Slack notification is sent to the relevant support channel. [Which providers does the user belong to?] [Should we also notify the user? needs email template]

### Guidance to review

TBD

### Link to Trello card

[1498 - Add form to the "Your account is not ready yet" to request access](https://trello.com/c/5c5VrDTt)


duncanjbrown:
@mnacos can we close this for now, since the Trello card has been archived?

PR #1116, opened 2020-01-15
Add uppercasing when entering and saving postcode

chubberlisk:
## Context

Postcodes should be uppercased.

This PR adds uppercasing when a candidates enters their postcode to make it easier for text entry on mobile and when we save it.

## Guidance to review

Not sure how I targeted the postcode text field using CSS is okay?

## Link to Trello card

https://trello.com/c/003s4hNf/775-uppercase-postcodes-on-save

## Things to check

### DIFF
--- a/app/frontend/styles/_postcode.scss
+++ b/app/frontend/styles/_postcode.scss
@@ -1,0 +1,3 @@
+.govuk-input[autocomplete="postal-code"] {
+  text-transform: uppercase;
+}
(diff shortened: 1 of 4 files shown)


app/frontend/styles/_postcode.scss
```diff
@@ -0,0 +1,3 @@
+.govuk-input[autocomplete="postal-code"] {
```
tvararu: I would target an `#id` instead 👍

chubberlisk:
> Do we want a data migration to uppercase all the existing records? Can be a separate PR.

Yup, will do in a separate PR!

app/frontend/styles/_postcode.scss
tvararu: I've changed my mind on this. Actually, targeting by the autocomplete attribute is good.

PR #1117, opened 2020-01-15
Fix  unformatted GCSE or equivalent qualification titles

stevehook:
## Context

In the support and provider interfaces the table of qualifications displays unformatted qualification types, e.g. _gcse maths_ rather than _GCSE Maths_.

- Use existing mappings of qualification types to localisable strings.
- Title case the subject.
- Add some specs for the `QualificationTitleComponent`.

image

## Guidance to review

- Nothing particular. However there are a few more edge cases that I've stumbled on and added to the original card, see link below. I think it's still worth the changes here now and dealing with other issues in further PRs if needed.

## Link to Trello card

https://trello.com/c/miNcm05m/1500-gcse-or-equivalent-fields-are-unformatted-in-provider-and-support-interface

## Things to check

### DIFF
--- a/app/components/qualification_title_component.rb
+++ b/app/components/qualification_title_component.rb
@@ -6,9 +6,10 @@ class QualificationTitleComponent < ActionView::Component::Base
   def qualification_type
     if @qualification.level == 'gcse' && @qualification.missing_qualification?
       I18n.t('application_form.gcse.qualification_types_abbreviated.gcse')
+      ApplicationQualification.human_attribute_name('qualification_type.gcse')
     elsif @qualification.level == 'gcse'
-      I18n.t(
-        "application_form.gcse.qualification_types_abbreviated.#{@qualification.qualification_type}",
+      ApplicationQualification.human_attribute_name(
+        "qualification_type.#{@qualification.qualification_type}",
         default: @qualification.qualification_type,
       )
     else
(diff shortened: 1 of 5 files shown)


app/components/qualification_title_component.rb
```diff
@@ -5,7 +5,12 @@ def initialize(qualification:)
 
   def qualification_type
     if @qualification.level == 'gcse' && @qualification.missing_qualification?
-      'GCSE'
+      I18n.t('application_form.gcse.qualification_types_abbreviated.gcse')
+    elsif @qualification.level == 'gcse'
+      I18n.t(
+        "application_form.gcse.qualification_types_abbreviated.#{@qualification.qualification_type}",
```
duncanjbrown: Alternatively, we could set these as i18n strings for ActiveRecord:

```
  activerecord:
    attributes:
      application_qualification/qualification_type:
        gcse: GCSE
        gce_o_level: GCE O Level
        scottish_national_5: Scottish National 5
        other_uk: Other UK
        missing: I don’t have this qualification yet
```

And call like this?

```
      ApplicationQualification.human_attribute_name(
        "qualification_type.#{@qualification.qualification_type}", default: @qualification.qualification_type
      )
```

stevehook: thanks, that looks a lot better, done

stevehook: thanks, that looks a lot better, done

PR #1119, opened 2020-01-15
Add migration to upcase all existing postcodes

chubberlisk:
## Context

In #1116, we added in uppercasing when entering and saving postcodes. But as we've only just added in uppercasing when saving, there exists postcodes that aren't uppercase.

This PR adds a migration to upcase all existing postcodes.

## Guidance to review

Not sure if this is the way to go and if the code is right, help pliz!

## Link to Trello card

https://trello.com/c/003s4hNf/775-uppercase-postcodes-on-save

## Things to check

### DIFF
--- a/db/migrate/20200115124513_upcase_postcodes_on_application_forms.rb
+++ b/db/migrate/20200115124513_upcase_postcodes_on_application_forms.rb
@@ -1,0 +1,14 @@
+class UpcasePostcodesOnApplicationForms < ActiveRecord::Migration[6.0]
+  def up
+    execute <<-SQL
+      UPDATE
+        application_forms
+      SET
+        postcode = UPPER(postcode)
+    SQL
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_01_13_140303) do
+ActiveRecord::Schema.define(version: 2020_01_15_124513) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"


db/migrate/20200115124513_upcase_postcodes_on_application_forms.rb
```diff
@@ -0,0 +1,11 @@
+class UpcasePostcodesOnApplicationForms < ActiveRecord::Migration[6.0]
+  def up
+    ApplicationForm.all.each do |a|
+      a.update!(postcode: a.postcode.upcase) unless a.postcode.nil?
```
tijmenb: If we can't use an SQL query we should use `update_column`, which doesn't update the timestamp / add things to the audit log.

chubberlisk: Ah I've updated to use an SQL query now! Didn't realise the timestamp/audit log thing.

db/migrate/20200115124513_upcase_postcodes_on_application_forms.rb
```diff
@@ -0,0 +1,14 @@
+class UpcasePostcodesOnApplicationForms < ActiveRecord::Migration[6.0]
+  def up
+    execute <<-SQL
+      UPDATE
+        application_forms
+      SET
+        postcode = UPPER(postcode)
```
tijmenb: Does this need a `WHERE NOT NULL` or is that fine?

chubberlisk: Just tested this by doing `ApplicationForm.first.update!(postcode: nil)`, then running the migration. It didn't blow up or complain for me. 

Is this the right way to test?

tvararu: I've tested the SQL on my own machine, where I had 3 records `where postcode is null`. It seems to work!

PR #1121, opened 2020-01-15
Support console shows when a referee declined to provide a reference

C-gyorfi:
## Context

We want to display on the support console when a referee declined to provide a reference to a candidate and/or has declined to be contacted for research activities. 

This PR updates:
 
- A support user is able to see when a referee has declined to provide a reference; this should be a separate row called "Given consent for research?"
- "Feedback status" row should be renamed to "Reference status"
- "Feedback requested" badge content should just be "Requested"
- "Feedback declined" badge content should just be "Declined" and red
- A **Slack message** is posted on the channel "twd_apply_support" when a referee declines to give feedback, for example, "X declined to give feedback for Y's application :sadparrot:"

### After
image

image

## Link to Trello card
https://trello.com/c/5dYeBzfT/674-support-console-show-when-a-referee-declined-to-provide-a-reference-or-to-be-contacted-for-user-research-purposes-%F0%9F%8F%88

### DIFF
--- a/app/components/reference_with_feedback_component.rb
+++ b/app/components/reference_with_feedback_component.rb
@@ -6,6 +6,7 @@ class ReferenceWithFeedbackComponent < ActionView::Component::Base
            :email_address,
            :relationship,
            :feedback_status,
+           :consent_to_be_contacted,
            to: :reference
 
   def initialize(reference:, title: '', show_send_email: false)
@@ -20,6 +21,7 @@ class ReferenceWithFeedbackComponent < ActionView::Component::Base
       name_row,
       email_address_row,
       relationship_row,
+      consent_row,
       feedback_row,
     ].compact
   end
@@ -62,6 +64,23 @@ private
         key: 'Reference',
         value: feedback,
       }
+    end
+  end
+
+  def consent_row
+    if feedback
+      {
+        key: 'Given consent for research?',
+        value: consent_to_be_contacted_present
+      }
+    end
+  end
+
+  def consent_to_be_contacted_present
+    if consent_to_be_contacted.nil? 
+      ' - ' 
+    else
+      consent_to_be_contacted == true ? 'Yes' : 'No'
     end
   end
 
--- a/app/components/reference_with_feedback_component.rb
+++ b/app/components/reference_with_feedback_component.rb
@@ -30,10 +30,10 @@ private
 
   def status_row
     {
-      key: 'Feedback status',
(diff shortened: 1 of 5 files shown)


app/controllers/referee_interface/reference_controller.rb
```diff
@@ -78,5 +81,12 @@ def check_referee_has_valid_token
     def render_404
       render 'errors/not_found', status: :not_found
     end
+
+    def send_slack_notification
+      message = "#{reference.name} declined to give feedback for #{reference.application_form.full_name}'s application :sadparrot:"
```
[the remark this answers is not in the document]
tvararu: Didn't realise the PII thing, I originally wrote the example 👍

app/components/reference_with_feedback_component.rb
```diff
@@ -28,10 +30,10 @@ def rows
 
   def status_row
     {
-      key: 'Feedback status',
+      key: 'Reference status',
       value: render(TagComponent,
-                    text: ApplicationReference.human_attribute_name("feedback_status.#{feedback_status}"),
-                    type: 'blue'),
+                    text: t("reference_status.#{feedback_status}"),
+                    type: color(feedback_status)),
```
tvararu:
```suggestion
                    type: feedback_status == 'feedback_refused' ? 'red' : 'blue',
```

I assume Rubocop doesn't like this? 😅 

Otherwise maybe `feedback_tag_color` is less generic, but this is a nit anyway.

config/locales/reference.yml
```diff
@@ -0,0 +1,6 @@
+en:
+  reference_status:
+    not_requested_yet: 'NOT REQUESTED YET'
+    feedback_requested: 'REQUESTED'
+    feedback_provided: 'FEEDBACK PROVIDED'
+    feedback_refused: 'DECLINED'
```
tvararu: Because we already uppercase using CSS, these should all be title cased, which is better for screen readers in some niche cases:

```yml
      not_requested_yet: 'Not requested yet'
      feedback_requested: 'Requested'
      feedback_provided: 'Feedback provided'
      feedback_refused: 'Declined'
```

app/components/reference_with_feedback_component.rb
C-gyorfi: Yes, Rubocop is not happy 👮 
I think `feedback_tag_color` is a sensible name, changed it!

PR #1124, opened 2020-01-15
 Show an error message if a support user already exists

tijmenb:
##  Context

Currently, when you enter a duplicate email address in the provider user form on support we raise a 500 error.

This changes it to a proper validation error.

## Link to Trello card

https://trello.com/c/4dsmwvAW/1519-show-a-sensible-error-if-a-user-tries-to-add-the-same-provider-email-twice

## Things to check

### DIFF
--- a/app/models/support_interface/provider_user_form.rb
+++ b/app/models/support_interface/provider_user_form.rb
@@ -6,7 +6,7 @@ module SupportInterface
     attr_accessor :provider_ids, :provider_user
     attr_reader :email_address
 
-    validates_presence_of :email_address
+    validates :email_address, presence: true
     validates :provider_ids, presence: true
     validate :email_is_unique
 
--- a/app/models/support_interface/provider_user_form.rb
+++ b/app/models/support_interface/provider_user_form.rb
@@ -3,7 +3,8 @@ module SupportInterface
     include ActiveModel::Model
     include ActiveModel::Validations
 
-    attr_accessor :email_address, :provider_ids, :provider_user
+    attr_accessor :provider_ids, :provider_user
+    attr_reader :email_address
 
     validates_presence_of :email_address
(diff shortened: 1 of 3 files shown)


app/models/support_interface/provider_user_form.rb
```diff
@@ -34,5 +35,15 @@ def self.from_provider_user(provider_user)
         provider_ids: provider_user.provider_ids,
       )
     end
+
+  private
+
+    def email_is_unique
```
stevehook: Do we need to downcase `email_address` here (or earlier) to ensure that we can't duplicates that differ only by case?

tijmenb: Great point - fixed in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1124/commits/416a5772dbb765558cdbf45432ef7a37b7cb2e54.

spec/system/support_interface/managing_provider_users_spec.rb
```diff
@@ -11,7 +11,11 @@
     and_i_click_the_users_link
     and_i_click_the_manange_provider_users_link
     and_i_click_the_add_user_link
-    and_i_enter_the_users_email_and_dsi_uid
+    and_i_enter_an_existing_email
+    and_i_click_add_user
+    then_i_see_an_error
+
+    and_i_enter_the_users_email
```
tvararu:
```suggestion
    when_i_enter_the_users_email
```

Minor

app/models/support_interface/provider_user_form.rb
```diff
@@ -19,6 +21,10 @@ def save
       )
     end
 
+    def email_address=(raw_email_address)
+      @email_address = raw_email_address.downcase
+    end
```
stevehook: nice

PR #1130, opened 2020-01-16
Use MoJ Design System for sub navigation

tijmenb:
I was looking at using [MoJ’s Sub navigation component](https://moj-design-system.herokuapp.com/components/sub-navigation) and wanted to copy-paste it.

But what if we just added MoJ’s Design System and used it directly?

We’ll have direct access to some pretty cool new components, like pagination, search, multi-select, and timeline.

https://moj-design-system.herokuapp.com/components

Of course it feels a bit odd for DfE to use a Design System specifically for MoJ - but the longer I think about it, the more it makes sense. Let’s break down these silos and re-use the good stuff
that we’ve built in government!

I used the sub nav component to test if it works:

image

Thoughts - @adamsilver @fofr @paulrobertlloyd @duncanjbrown @tvararu?

### DIFF
--- a/app/frontend/styles/application.scss
+++ b/app/frontend/styles/application.scss
@@ -10,6 +10,7 @@ $govuk-font-url-function: frontend-font-url;
 $govuk-image-url-function: frontend-image-url;
 
 @import "~govuk-frontend/govuk/all";
+@import "~@ministryofjustice/frontend/moj/components/sub-navigation/sub-navigation";
 @import "_autocomplete";
 @import "_banner";
 @import "_cookie-banner";
--- a/app/views/support_interface/provider_users/index.html.erb
+++ b/app/views/support_interface/provider_users/index.html.erb
@@ -13,6 +13,11 @@
       </li>
     </ol>
   </div>
+
+  <%= render SubNavigationComponent, items: [
+    { name: 'Support users', url: support_interface_support_users_path },
+    { name: 'Provider users', url: support_interface_provider_users_path, current: true },
+  ] %>
 <% end %>
 
 <%= render(SupportInterface::ProviderUsersTableComponent, provider_users: @provider_users) %>
--- a/yarn.lock
+++ b/yarn.lock
@@ -692,6 +692,13 @@
   resolved "https://registry.yarnpkg.com/@csstools/convert-colors/-/convert-colors-1.4.0.tgz#ad495dc41b12e75d588c6db8b9834f08fa131eb7"
   integrity sha512-5a6wqoJV/xEdbRNKVo6I4hO3VjyDq//8q2f9I6PBAvMesJHFauXDorcNCsr9RzvsZnaWi5NYCcfyqP1QeFHFbw==
 
+"@ministryofjustice/frontend@0.0.17-alpha":
+  version "0.0.17-alpha"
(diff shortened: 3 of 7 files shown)


app/frontend/styles/application.scss
```diff
@@ -10,6 +10,7 @@ $govuk-font-url-function: frontend-font-url;
 $govuk-image-url-function: frontend-image-url;
 
 @import "~govuk-frontend/govuk/all";
+@import "~@ministryofjustice/frontend/moj/all";
```
tvararu: There is a downside to doing this: we're importing their full CSS. I don't know how big it is but presumably we only need a small part of it.

Do they allow importing just the CSS for the sub-navigation?

Otherwise, this is fine for now, however I'm going to create a tech debt card to investigate the size of our CSS and introduce something like https://www.purgecss.com to slim it down.

app/views/support_interface/provider_users/index.html.erb
```diff
@@ -13,6 +13,18 @@
...
+    <ul class="moj-sub-navigation__list">
+      <li class="moj-sub-navigation__item">
+        <%= govuk_link_to 'Support users', support_interface_support_users_path, class: 'moj-sub-navigation__link' %>
+      </li>
+
+      <li class="moj-sub-navigation__item">
+        <%= govuk_link_to 'Provider users', support_interface_provider_users_path, class: 'moj-sub-navigation__link', 'aria-current': 'page' %>
+      </li>
+    </ul>
+  </nav>
```
tvararu: Is it worth creating a Component for this which means the MOJ-specific CSS is only referred to in one place in the codebase?

app/frontend/styles/application.scss
tijmenb: Great idea, done & updated!

app/views/support_interface/provider_users/index.html.erb
tijmenb: 👍 done!

app/views/support_interface/provider_users/index.html.erb
```diff
@@ -13,6 +13,11 @@
       </li>
     </ol>
   </div>
+
+  <%= render SubNavigationComponent, items: [
+    { name: 'Support users', url: support_interface_support_users_path },
+    { name: 'Provider users', url: support_interface_provider_users_path, current: true },
+  ] %>
```
tvararu: Nice!

yarn.lock
```diff
@@ -4325,6 +4332,11 @@ mixin-deep@^1.2.0:
   dependencies:
     minimist "0.0.8"
 
+moment@^2.22.2:
+  version "2.24.0"
+  resolved "https://registry.yarnpkg.com/moment/-/moment-2.24.0.tgz#0d055d53f5052aa653c9f6eb68bb5d12bf5c2b5b"
+  integrity sha512-bV7f+6l2QigeBBZSM/6yTNq4P2fNpSWj/0e7jQcy87A8e7o2nAfP/34/2ky5Vw4B9S446EtIhodAzkFCcR4dQg==
+
```
tvararu: Just realised MOJ frontend depends on Moment. It's used in [the timeline component](https://moj-design-system.herokuapp.com/components/timeline) but only on the server-side when Nunjucks templates are used. This is fine as it won't end up in our client-side bundle at any point.

(Context: moment.js is notoriously heavy https://bundlephobia.com/result?p=moment@2.24.0 and a very common contributor to slow frontend performance due to bloated JS bundle sizes)

PR #1131, opened 2020-01-16
Automate DfE Sign-In invitation as part of creating a provider user

mnacos:
## Context

Inviting a provider user is a very manual process which doesn't scale. We would like to automate as many steps of this process as possible, and we are starting with invitations to the DfE Sign-In platform.

DfE Sign-In provides an API. This API requires `sourceId`, as well as `given_name` and `family_name`, alongside `email` (see https://github.com/DFE-Digital/login.dfe.public-api#invite-user). Currently, we do not ask for any names when creating provider users within the Support console. This may have to change because names are required by:

- DfE Sign-In invitations
- our own email templates 

Planning to add a database migration for storing names against `provider_users` as a separate PR.

This PR introduces an `InviteProviderUser` service, which handles saving the form and also inviting users to DfE Sign-In via their API. In the future, it will also send welcome emails etc. Operations are wrapped in a transaction because that's a good thing (tm), but also because the DSI API requires `sourceId`, which is our unique identifier for the user (i.e. their AR id). So the code creates a `ProviderUser`, which is rolled back if the DSI invitation fails.

Two new environment variables are added, `DSI_API_URL` and `DSI_API_SECRET`.
image

## Guidance to review

The form for creating new provider users within the support console now contains first and last name fields. These are now stored against the provider user. Editing an existing provider user is exactly the same as before, with added name-related text fields. The update action is restricted to updating names, as `uid` is set automatically the first time a user logs in, and `email` will soon be updated automatically every time a user sign in.

Slack me if you're reviewing this for the appropriate `DSI_API_URL` and `DSI_API_SECRET` values for your local environment.

You'll also need to switch on the relevant feature flag on the environment you are testing (via the support interface):
image

Invite a provider user and complete the sign-up flow. You'll be redirected back to the provider interface of the environment you are testing.

It is not possible to avoid the additional click on the green sign in button when a user is redirected back to the app (once completing DSI sign-up), because triggering the sign-in flow requires a `POST`.

## Link to Trello card

[When a user is added, call the Signin API to invite the user](https://trello.com/c/H9oC6kxo)

## Things to check

### DIFF
--- a/app/controllers/support_interface/provider_users_controller.rb
+++ b/app/controllers/support_interface/provider_users_controller.rb
@@ -10,8 +10,9 @@ module SupportInterface
 
     def create
       @form = ProviderUserForm.new(provider_user_params)
+      service = InviteProviderUser.new(provider_user_form: @form)
 
-      if @form.save
+      if service.call
         flash[:success] = 'Provider user created'
         redirect_to support_interface_provider_users_path
       else
@@ -40,7 +41,7 @@ module SupportInterface
   private
 
     def provider_user_params
-      params.require(:support_interface_provider_user_form).permit(:email_address, provider_ids: [])
+      params.require(:support_interface_provider_user_form).permit(:email_address, :first_name, :last_name, provider_ids: [])
     end
   end
 end
--- a/app/controllers/support_interface/provider_users_controller.rb
+++ b/app/controllers/support_interface/provider_users_controller.rb
@@ -10,12 +10,19 @@ module SupportInterface
 
     def create
       @form = ProviderUserForm.new(provider_user_params)
-      service = InviteProviderUser.new(provider_user_form: @form)
+      provider_user = @form.build
 
-      if service.call
-        flash[:success] = 'Provider user created'
-        redirect_to support_interface_provider_users_path
-      else
+      if provider_user
+        service = InviteProviderUser.new(provider_user: provider_user)
+        begin
+          service.save_and_invite!
+          flash[:success] = 'Provider user created'
+          redirect_to support_interface_provider_users_path
+        rescue DfeSignInApiError => e # show errors from api
+          e.errors.each { |error| @form.errors.add(:base, error) }
+          render :new
+        end
+      else # show errors from form e.g. email uniqueness
         render :new
       end
     end
@@ -27,8 +34,8 @@ module SupportInterface
 
     def update
       provider_user = ProviderUser.find(params[:id])
-
-      @form = ProviderUserForm.new(provider_user_params.merge(provider_user: provider_user))
+      @form = ProviderUserForm.from_provider_user(provider_user)
+      @form.assign_attributes provider_user_params
 
       if @form.save
         flash[:success] = 'Provider user updated'
--- a/app/models/support_interface/provider_user_form.rb
+++ b/app/models/support_interface/provider_user_form.rb
@@ -10,15 +10,19 @@ module SupportInterface
     validates :provider_ids, presence: true
     validate :email_is_unique
 
-    def save
+    def build
       return unless valid?
 
       @provider_user ||= ProviderUser.new
+      @provider_user.first_name = first_name
+      @provider_user.last_name = last_name
+      @provider_user.email_address = email_address
+      @provider_user.provider_ids = provider_ids
+      @provider_user if @provider_user.valid?
+    end
 
-      @provider_user.update!(
-        email_address: email_address,
-        provider_ids: provider_ids,
-      )
+    def save
+      @provider_user.save! if build
     end
 
     def email_address=(raw_email_address)
@@ -30,12 +34,14 @@ module SupportInterface
     end
 
     def persisted?
-      provider_user && provider_user.persisted?
+      @provider_user && @provider_user.persisted?
     end
 
     def self.from_provider_user(provider_user)
       new(
         provider_user: provider_user,
+        first_name: provider_user.first_name,
+        last_name: provider_user.last_name,
         email_address: provider_user.email_address,
         provider_ids: provider_user.provider_ids,
       )
--- a/app/models/support_interface/provider_user_form.rb
+++ b/app/models/support_interface/provider_user_form.rb
@@ -3,7 +3,7 @@ module SupportInterface
     include ActiveModel::Model
     include ActiveModel::Validations
 
-    attr_accessor :provider_ids, :provider_user
+    attr_accessor :first_name, :last_name, :provider_ids, :provider_user
     attr_reader :email_address
 
     validates :email_address, presence: true
--- a/app/services/invite_provider_user.rb
+++ b/app/services/invite_provider_user.rb
@@ -1,0 +1,61 @@
+require 'jwt'
+
+class InviteProviderUser
+  def initialize(provider_user_form:)
+    @provider_user_form = provider_user_form
+  end
+
+  def call
+    ActiveRecord::Base.transaction do
+      if @provider_user_form.save
+        invite_user_to_dfe_sign_in
+
+        # TODO: send welcome/invitation email
+        true
+      end
+    end
+  rescue DfeSignInApiError => e
+    e.errors.each { |error| @provider_user_form.errors.add(:base, error) }
+    false
+  end
+
+  def dfe_invite_url
+    baseurl = ENV.fetch('DSI_API_URL')
+    if baseurl.present?
+      "#{baseurl}/services/apply/invitations"
+    end
+  end
+
+  def invite_user_to_dfe_sign_in
+    return unless FeatureFlag.active?('send_dfe_sign_in_invitations')
+
+    jwt_payload = { iss: 'apply', aud: 'signin.education.gov.uk' }
+    token = JWT.encode jwt_payload, ENV['DSI_API_SECRET'], 'HS256'
+    auth_string = "Bearer #{token}"
+
+    request_params = {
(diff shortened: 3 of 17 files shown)


app/services/invite_provider_user.rb
```diff
@@ -0,0 +1,60 @@
+require 'jwt'
+
+class InviteProviderUser
+  def initialize(provider_user_form:)
```
tijmenb: I'm not sure about passing in the form here. Perhaps we can only pass in the `ProviderUser` object?

app/services/invite_provider_user.rb
```diff
@@ -0,0 +1,60 @@
...
+    @provider_user_form = provider_user_form
+  end
+
+  def call
+    ActiveRecord::Base.transaction do
+      if @provider_user_form.save
+        response = invite_user_to_dfe_sign_in
+        raise DfeSignInApiError.new(response) if response && !response.status.success?
```
tijmenb: Feels like we're mixing concerns in this method. Would the `raise` be better in the `invite_user_to_dfe_sign_in` method?

app/services/invite_provider_user.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def dfe_invite_url
+    baseurl = ENV['DSI_API_URL']
+    if baseurl.present?
+      "#{ENV['DSI_API_URL']}/services/apply/invitations"
+    end
+  end
+
+  def invite_user_to_dfe_sign_in
+    return unless dfe_invite_url && @provider_user_form.persisted?
```
tijmenb: I don't think we have to check for `persisted?` because it'll always be true?

tijmenb:
> Planning to add a database migration for storing names against provider_users as a separate PR.

For consistency, can we add them to `SupportUser` as well?

app/services/invite_provider_user.rb
mnacos: Yes, that is correct. I've addressed this.

app/services/invite_provider_user.rb
mnacos: Ok. I guess I was enjoying calling `invite_user_to_dfe_sign_in` directly for troubleshooting error cases too much. Will keep this as a public method, as it's good for troubleshooting or implementing a 're-send the invite' button.

app/services/invite_provider_user.rb
mnacos: In this iteration, we don't yet have `first_name`, `last_name` for the provider user. The current plan is to add this migration as a separate PR, together with code matching on uid when people sign-in to reconcile their email address, first name and last name. This way DfE Sign-In remains the source of truth for email and names.

app/controllers/support_interface/provider_users_controller.rb
```diff
@@ -10,8 +10,9 @@ def new
 
     def create
       @form = ProviderUserForm.new(provider_user_params)
+      service = InviteProviderUser.new(provider_user_form: @form)
```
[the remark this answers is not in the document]
mnacos: Hmm `ProviderUserForm#save` is also used for the edit/update cycle. I think I prefer using `InviteProviderUser` service directly for new/create. Besides, I see the service as more general than the form (it will soon send emails, too). If you feel calling the service from the form#save is better, we can do that but will need to add logic to `#save` to determine if the service should be called.

mnacos:
> > Planning to add a database migration for storing names against provider_users as a separate PR.
> 
> For consistency, can we add them to `SupportUser` as well?

Yes, but planning to do this as a separate PR

app/services/invite_provider_user.rb
mnacos: On one hand I agree the less coupling the better, so `form <-> model` and `service <-> model` interactions would be kept separate by adding temporary attributes to the model, however a `service <-> form` interaction has the advantage the service can add API errors directly to the form. Not sure what the right way is here. I don't think `service <-> form <-> model` is too bad.

app/controllers/support_interface/provider_users_controller.rb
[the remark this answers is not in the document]
mnacos: Ok, we've got a way forward after chatting IRL. The controller will ask the form to build a ProviderUser model, which will then be passed to the service. The service will attempt to save this model, if it cannot the errors will be fed back to the user through the form. We are in the process of merging https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1145 in first, so `first_name` and `last_name` will be available on the model.

In the presence of API-related errors, we'll stop handling the Exception within the service and let the controller handle it, in order to feed back these errors to the UI.

app/services/invite_provider_user.rb
mnacos: Progress. See comments on controller above. We'll have `service <-> model`, but the model will be built by the form. The controller will be responsible for communicating any errors to the appropriate objects.

app/controllers/support_interface/provider_users_controller.rb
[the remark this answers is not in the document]
mnacos: This is address by what we decided IRL

app/controllers/support_interface/provider_users_controller.rb
```diff
@@ -10,11 +10,23 @@ def new
...
-      else
+      provider_user = @form.build
+
+      if @form.valid?
+        service = InviteProviderUser.new(provider_user: provider_user)
+        begin
+          if service.save_and_invite!
+            flash[:success] = 'Provider user created'
+            redirect_to support_interface_provider_users_path
+          else # show errors from model
```
duncanjbrown: If the form's declared the model valid, would this ever happen?

app/models/support_interface/provider_user_form.rb
```diff
@@ -30,12 +34,14 @@ def available_providers
...
+      @provider_user && @provider_user.persisted?
     end
 
     def self.from_provider_user(provider_user)
       new(
         provider_user: provider_user,
+        first_name: provider_user.first_name,
+        last_name: provider_user.last_name,
         email_address: provider_user.email_address,
         provider_ids: provider_user.provider_ids,
```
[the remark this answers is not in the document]
mnacos: The new environment variables are set in `.env.test` for running tests, both locally and in CI pipelines, which is why we haven't had any failures yet. Have just added some more configuration.

mnacos:
> Looks great. I have one query about the Azure config files. I don't understand the detail but I think there are more files that need to be edited to declare the new environment variables. Do we not need to worry about the pipeline ones? I guess because these variables are not used in the CI build?
> 
> https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/docs/environment-variables.md

The new environment variables are set in .env.test for running tests, both locally and in CI pipelines, which is why we haven't had any failures yet. Have just added some more configuration.

app/models/support_interface/provider_user_form.rb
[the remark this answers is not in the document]
mnacos: Shorter but less readable, I think.

app/controllers/support_interface/provider_users_controller.rb
```diff
@@ -26,10 +38,13 @@ def edit
 
     def update
       provider_user = ProviderUser.find(params[:id])
+      @form = ProviderUserForm.from_provider_user(provider_user)
 
-      @form = ProviderUserForm.new(provider_user_params.merge(provider_user: provider_user))
+      @form.first_name = provider_user_params[:first_name]
+      @form.last_name = provider_user_params[:last_name]
+      @form.provider_ids = provider_user_params[:provider_ids]
```
duncanjbrown: I think we get `#assign_attributes` along with `ActiveModel::Model`, which might make this a bit more succinct?

app/controllers/support_interface/provider_users_controller.rb
mnacos: The only time this can happen is if the validations from the model and the form diverge. Currently, this could happen with the model's unique validation on `dfe_sign_in_uid`, the likelihood of which is non-existent. But I left  this `else` there as a catch-all for non-exception errors.

spec/services/invite_provider_user_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+      it 'rolls back provider user creation' do
+        InviteProviderUser.new(provider_user: new_provider_user_from_form).save_and_invite! rescue nil
+        expect(ProviderUser.find_by_email_address('test+invite_provider_user@example.com')).to be_nil
+      end
+    end
+  end
+
+  context 'with feature flag off' do
+    let(:http_spy) { class_spy('HTTP') }
```
duncanjbrown: Sorry for the late comment — this couples the test to the implementation a bit. Could we use webmock?

app/controllers/support_interface/provider_users_controller.rb
```diff
@@ -26,10 +34,10 @@ def edit
 
     def update
       provider_user = ProviderUser.find(params[:id])
+      @form = ProviderUserForm.from_provider_user(provider_user)
+      @form.assign_attributes provider_user_params
 
-      @form = ProviderUserForm.new(provider_user_params.merge(provider_user: provider_user))
-
-      if @form.save
+      if @form.save!
```
duncanjbrown: `save!` will raise `RecordInvalid` rather than return false — I don't think we handle that?

mnacos: `save!` is misleading, because I added a guard checking the model is valid before attemptint to save. Renaming to just `save`.

PR #1133, opened 2020-01-16
Small typographical fixes

benilovj:
A couple of places in the interfaces where we probably should be using [smart quotes](https://smartquotesforsmartpeople.com/).


duncanjbrown:
There are more!

provider_interface/provider_agreements/data_sharing_agreement.html.erb line 291
candidate_interface/sign_up/new.html.erb  line 17

duncanjbrown:
There should be a linter for this

benilovj:
Thanks @duncanjbrown, I've updated those two places as well.

PR #1134, opened 2020-01-16
Send welcome email to provider user when they are added

stevehook:
## Context

To streamline the onboarding process a DSI account can be created when provider users are added via the support interface. We also need to send out a welcome email with a link to the provider interface.

- add a new mailer for providers with a method to send the 'account created' email.
- queue a job from the `InviteProviderUser` service to send the email after the new DSI account has been requested
- for convenience a `ProviderUser#full_name` method has been added

## Guidance to review

- does it make sense to queue the mailer from the service?
- have I tested the changes to the service and the mailer itself adequately?
- do we ever have to worry about `ProviderUser#first_name` or `last_name` being nil? (the database column is nullable but they are required fields at signup time I think).

## Link to Trello card

https://trello.com/c/cYX0hWq8/1515-automatically-send-an-email-to-the-provider-user-when-theyre-added

## Things to check

### DIFF
--- a/app/views/provider_mailer/account_created.text.erb
+++ b/app/views/provider_mailer/account_created.text.erb
@@ -1,4 +1,4 @@
-Dear <%= @provider_user_name %>,
+Dear <%= @provider_user.full_name %>,
 
 You've created an account with DfE Sign-in.
 
--- a/app/views/provider_mailer/account_created.text.erb
+++ b/app/views/provider_mailer/account_created.text.erb
@@ -1,13 +1,15 @@
-Dear <%= @provider_user.full_name %>,
+Dear <%= @provider_user.full_name %>
 
-You've created an account with DfE Sign-in.
+Welcome to Manage teacher training applications.
 
-<b>Sign in to Manage teacher training applications.</b>
+You will need to use DfE Sign-in to access your applications.
 
-You can now sign in to the service to start managing applications:
+# Already activated your DfE Sign-in account?
 
-[<%= provider_interface_url %>](<%= provider_interface_url %>)
+Visit [Manage teacher training applications](<%= provider_interface_url %>) to sign in.
 
-You'll need your DfE Sign-in account details to log in each time.
+# Don’t have a DfE Sign-in account?
 
-If you have any questions you can contact us at becomingateacher@digital.education.gov.uk
+Please check your inbox for an email from us inviting you to create a DfE Sign-in account.
+
+You can contact us at [becomingateacher@digital.education.gov.uk](mailto:becomingateacher@digital.education.gov.uk) to request a new email, or for help and support. 
--- a/app/views/provider_mailer/account_created.text.erb
+++ b/app/views/provider_mailer/account_created.text.erb
@@ -6,7 +6,7 @@ You've created an account with DfE Sign-in.
 
 You can now sign in to the service to start managing applications:
 
-[<%= provider_interface_applications_url %>](<%= provider_interface_applications_url %>)
+[<%= provider_interface_url %>](<%= provider_interface_url %>)
 
 You'll need your DfE Sign-in account details to log in each time.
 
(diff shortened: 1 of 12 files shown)


app/views/provider_mailer/account_created.text.erb
```diff
@@ -0,0 +1,13 @@
+Dear <%= @provider_user_name %>,
+
+You've created an account with DfE Sign-in.
+
+<b>Sign in to Manage teacher training applications.</b>
+
+You can now sign in to the service to start managing applications:
+
+[<%= provider_interface_applications_url %>](<%= provider_interface_applications_url %>)
```
mnacos: `provider_interface_applications_url` or just `provider_interface_url`?

spec/services/invite_provider_user_spec.rb
```diff
@@ -31,6 +31,13 @@
       it 'a provider user is created' do
         expect(ProviderUser.find_by_email_address('test+invite_provider_user@example.com')).not_to be_nil
       end
+
+      it 'queues an email' do
+        message_delivery = instance_double(ActionMailer::MessageDelivery, deliver_later: nil)
+        allow(ProviderMailer).to receive(:account_created).and_return(message_delivery)
+        InviteProviderUser.new(provider_user: new_provider_user_from_form).save_and_invite! rescue nil
+        expect(ProviderMailer).to have_received(:account_created)
```
tijmenb: Perhaps instead of this we can add a [test in the system spec](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/spec/system/support_interface/managing_provider_users_spec) that confirms that the email is sent? This feels too much coupled to the internals.

spec/services/invite_provider_user_spec.rb
```diff
@@ -31,6 +31,13 @@
       it 'a provider user is created' do
         expect(ProviderUser.find_by_email_address('test+invite_provider_user@example.com')).not_to be_nil
       end
+
+      it 'queues an email' do
+        message_delivery = instance_double(ActionMailer::MessageDelivery, deliver_later: nil)
+        allow(ProviderMailer).to receive(:account_created).and_return(message_delivery)
```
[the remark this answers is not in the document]
duncanjbrown: I forgot to submit this review! Now I see @tijmenb's suggestion which sounds better

app/views/provider_mailer/account_created.text.erb
```diff
@@ -0,0 +1,13 @@
+Dear <%= @provider_user.full_name %>,
+
+You've created an account with DfE Sign-in.
```
duncanjbrown: This email needs to work in two contexts: either the user has just been invited to Sign-in and they need to open that email before continuing, or they're already set up on Sign-in and they can log in to Apply straight away (we don't know which state the user is in). How about:

```
Your account has been created on Manage teacher training applications.

You will need a DfE Sign-in account to use the service. If you already have a DfE Sign-in account you can proceed to the sign in page for Manage teacher training applications (text is a link). If you don't have a DfE Sign-in account we have sent you an email inviting you to create one.
```

Any ideas @laurahermionetennant92?

spec/services/invite_provider_user_spec.rb
stevehook: Have added the assert to the system spec - it does require the feature flag to make this work.

spec/services/invite_provider_user_spec.rb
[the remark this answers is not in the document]
stevehook: :+1: I've added to the system spec. I changed the asserts for the service to be less implementation dependent I think - I kept them here too because I want to test the positive and negative cases and the system spec only does the former.

app/views/provider_mailer/account_created.text.erb
stevehook: @laurahermionetennant92 has added new copy to the Trello card

stevehook:
This is how the latest email template looks:

image

spec/rails_helper.rb
```diff
@@ -70,4 +70,6 @@
   config.include ActionView::Component::TestHelpers
 
   config.before { Faker::UniqueGenerator.clear }
+
+  config.before { ActionMailer::Base.deliveries.clear }
```
duncanjbrown: :+1:

PR #1135, opened 2020-01-16
Add notify reference to #reference_request_email

C-gyorfi:
## Context

This is the first slice of monitoring bounced back emails. We want to be able to keep track on emails that bounce back, so that we can tell candidates if they provided an incorrect reference email.

This PR adds a `notify reference` to the `reference_request_email` so that we can identify which email bounced back.  The notify reference is generated from `environment name`-`email_type`-`reference_id`

## Link to Trello card

https://trello.com/c/FuEKJx9W/772-bounced-email-monitoring

## Things to check

### DIFF
--- a/app/mailers/referee_mailer.rb
+++ b/app/mailers/referee_mailer.rb
@@ -8,6 +8,7 @@ class RefereeMailer < ApplicationMailer
     view_mail(GENERIC_NOTIFY_TEMPLATE,
               to: reference.email_address,
               subject: t('reference_request.subject.initial', candidate_name: @candidate_name),
+              reference: "#{Rails.env}-reference_request-#{reference.id}",
               template_name: :reference_request_email)
   end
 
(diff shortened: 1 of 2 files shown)


app/mailers/referee_mailer.rb
```diff
@@ -8,6 +8,7 @@ def reference_request_email(application_form, reference)
     view_mail(GENERIC_NOTIFY_TEMPLATE,
               to: reference.email_address,
               subject: t('reference_request.subject.initial', candidate_name: @candidate_name),
+              reference: "#{Rails.env}-reference_request-#{reference.id}",
```
tvararu: Double checked that `mail_notify`'s `view_mail` call does support this and [it does](https://github.com/dxw/mail-notify/blob/master/README.md#with-optional-notify-arguments) 👍

tijmenb: This won't work - on all live environments the `Rails.env` is "production". You probably want `HostingEnvironment.environment_name` ("production", "staging", "sandbox", etc), which is what we want to filter on.

PR #1139, opened 2020-01-16
Providers can change offers to rejections

duncanjbrown:
## Draft for now

Needs a look from designers. There’s an argument that the radio buttons are unnecessary and these alternatives could be links instead

## Context

Providers need to be able to reject applications that already have offers.

Create a new `ChangeDecision` controller in the Provider Interface. Its job is to present users with a set of radio buttons and redirect them to the right place once they pick one.

Screenshot 2020-01-16 at 17 27 40

Screenshot 2020-01-16 at 17 12 02

The copy is mine 😬 

## Link to Trello card

https://trello.com/c/KLy8enLu/1482-ui-reject-offered-applications

## Things to check

### DIFF
--- a/app/components/provider_interface/change_decision_component.html.erb
+++ b/app/components/provider_interface/change_decision_component.html.erb
@@ -1,0 +1,19 @@
+<%= form_with url: provider_interface_application_choice_confirm_change_decision_path, method: :post do |f| %>
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+
+      <p class="govuk-body">
+        <%= choice.current_state %>
+      </p>
+
+      <%= f.govuk_collection_radio_buttons :desired_change, choice.change_options, :value, :label, :description %>
+
+      <%= f.govuk_submit 'Continue' %>
+
+      <p class="govuk-body">
+        <%= govuk_link_to 'Cancel', provider_interface_application_choice_path(application_choice), class: 'govuk-link--no-visited-state' %>
+      </p>
+    </div>
+  </div>
+<% end %>
--- a/app/components/provider_interface/change_decision_component.rb
+++ b/app/components/provider_interface/change_decision_component.rb
@@ -1,0 +1,26 @@
+module ProviderInterface
+  class ChangeDecisionComponent < ActionView::Component::Base
+    include ViewHelper
+
+    ChangeOption = Struct.new(:value, :label, :description)
+    Choice = Struct.new(:current_state, :change_options)
+
+    attr_reader :application_choice
+
+    def initialize(application_choice:)
+      @application_choice = application_choice
+    end
+
+    def choice
+      case application_choice.status
+      when 'offer'
+        Choice.new(
+          'You have made this candidate an offer, but they have not accepted it yet.',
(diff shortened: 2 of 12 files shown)


duncanjbrown:
In the cold light of morning, I think I overdid this. The component is not very nice - a plain old view would probably be better?

tvararu:
@duncanjbrown the two main reasons I would advocate for a component:

- It's being used in a number of different places
- It has many different states which all need to be tested (unit tests are simpler than system tests for this)

We're not taking advantage of either here, so a plain view would suffice yes :)

fofr:
A single option Radio is a little odd. Content works well though.

adamsilver:
## Screen 1
I had the button text initially as ‘Change offer’ but having seen your ‘Change decision’ I think given the button in the initial state says ‘Respond to application’ I would suggest this becomes ‘Change response’.

If we prefer ‘Change decision’ then I think we should change ‘Respond to application’ to ‘Make decision’.

## Screen 2
The h1 should match the button text if we change it (see above)

The h1 should be inside the legend (if it isn't already) and the gap should be a lot smaller between the legend and the hint

Change the hint text to ‘You've made this candidate an offer, but they've not accepted it yet.’

I'm not sure about the radio button's label and hint. It's not quite a reject but rescind isn't all that clear.

## Single radio button

I gather this will be feature flagged, @fofr, so won't go out until we have the flows for the other radio buttons.

laurahermionetennant92:
Hi folks,
I have made some content updates on the is branch. 
Content now works for this single option. When we add other radio button options it will need to be looked at again.

laurahermionetennant92:
Screenshot 2020-01-17 at 14 44 41
At this point, as a provider I have gone through the flow of changing my decision from offer to rejection.
But we want to enable a rejection to go back to an offer if need be – at any time, right?
Because this is different from an offer accepted by a candidate and therefore binding.
This is a rejection which may have been noted by the candidate but is not in any way ‘binding’ (and also potentially the candidate would be thrilled to have it go from rejected to offer)

laurahermionetennant92:
So at this point, we should have a Change decision button on this page… which appears if you’ve rejected, but not if you’ve offered and been accepted
Screenshot 2020-01-17 at 15 11 22

laurahermionetennant92:
And on this page
Screenshot 2020-01-17 at 15 07 45
, we should have a Change decision button on this page… which appears if you’ve rejected, but not if you’ve offered and been accepted

laurahermionetennant92:
At
Screenshot 2020-01-17 at 14 52 45
 this point, we should have a Change decision button on this page… which appears if you’ve rejected, but not if you’ve offered and been accepted

laurahermionetennant92:
Also note to @adamsilver  that we need to look at how the candidate's deets are displayed and keep them consistent all through the work flow – maybe stacked (name, course, provider?)
Screenshot 2020-01-17 at 15 31 47

laurahermionetennant92:
@adamsilver I see what you mean re decision versus response – for some reason can't find this instance to change, can you fix?
Screenshot 2020-01-17 at 15 52 22

duncanjbrown:
Closing this PR pending movement on https://trello.com/c/vdo2FhAz/1487-design-change-state-from-offered-to-rejected

PR #1140, opened 2020-01-17
Skip the 'waiting to be sent' state in sandbox

stevehook:
## Context

When testing the application workflow in a Sandbox environment it's not practical for providers to wait for 5 business days before seeing a complete referenced application. So we need to skip the wait period.

- Assume that in the Sandbox there will be an environment variable called `SANDBOX` set to `true`. Add it to the Azure config files as per instructions.
- Adjust the `SubmitApplication` service to set the `edit_by` date to now so that the normal delay is skipped.
- Adjust the `ReceiveReference` service so that it immediately calls the `SendApplicationToProvider` service iff the references are complete and the `edit_by` time has elapsed so that providers don't have to wait for scheduled jobs to run.
- Encapsulate the business days/end of day calculation in TimeLimitCalculator to reduce and make it easier to customise the absolute time limit (not just a number of working days).

Note that this change will be more useful when we implement a mechanism for automatic reference completion in the Sandbox.

## Guidance to review

- Are the tests adequate?
- Does it make sense to trigger `SendApplicationToProvider` from `ReceiveReference`?
- Is all the Azure config right (Just followed the instructions, I don't understand it!)

## Link to Trello card

## Things to check

### DIFF
--- a/app/lib/hosting_environment.rb
+++ b/app/lib/hosting_environment.rb
@@ -53,4 +53,8 @@ module HostingEnvironment
   def self.production?
     environment_name == 'production'
   end
+
+  def self.sandbox?
+    ENV.fetch('SANDBOX') { 'false' } == 'true'
+  end
 end
--- a/app/lib/hosting_environment.rb
+++ b/app/lib/hosting_environment.rb
@@ -55,6 +55,6 @@ module HostingEnvironment
   end
 
   def self.sandbox?
-    ENV.fetch('SANDBOX') { 'false' } == 'true'
+    ENV.fetch('SANDBOX', 'false') == 'true'
   end
 end
--- a/app/services/receive_reference.rb
+++ b/app/services/receive_reference.rb
@@ -26,7 +26,7 @@ class ReceiveReference
 
       if @application_form.application_references_complete?
         @application_form.application_choices.includes(:course_option).each do |application_choice|
-          ApplicationStateChange.new(application_choice).references_complete!
+          complete_references(application_choice)
         end
       end
     end
@@ -44,6 +44,13 @@ private
   def referee_must_exist_on_application_form
     if @application_form.application_references.where(email_address: @referee_email).empty?
       errors.add(:referee_email, 'does not match any of the provided referees')
+    end
+  end
+
+  def complete_references(application_choice)
+    ApplicationStateChange.new(application_choice).references_complete!
+    if application_choice.edit_by <= Time.zone.now
+      SendApplicationToProvider.new(application_choice: application_choice).call
     end
   end
 end
--- a/app/services/time_limit_calculator.rb
+++ b/app/services/time_limit_calculator.rb
@@ -7,6 +7,19 @@ class TimeLimitCalculator
   end
 
   def call
+    days = calculate_days
+    [days, calculate_time(days)]
+  end
+
+private
+
+  def calculate_time(days)
+    return nil unless days
+
+    days.business_days.after(effective_date).end_of_day
+  end
+
+  def calculate_days
     to_and_from_time_limits.each do |time_limit|
       return time_limit.limit if effective_date <= time_limit.to_date && effective_date >= time_limit.from_date
     end
@@ -18,8 +31,6 @@ class TimeLimitCalculator
     end
     default_time_limit&.limit
   end
-
-private
 
   def time_limits_for_rule
     @time_limits_for_rule ||= TimeLimitConfig.limits_for(rule)
--- a/spec/services/submit_application_spec.rb
+++ b/spec/services/submit_application_spec.rb
@@ -42,7 +42,7 @@ RSpec.describe SubmitApplication do
         end
       end
 
-      it 'sets the edit_by timestamp to now and pushes the status on to `awaiting_provider_decision`' do
+      it 'sets the edit_by timestamp to now' do
         application_form = create_application_form
         now = Time.zone.local(2019, 11, 11, 15, 0, 0)
         Timecop.freeze(now) do
--- a/spec/services/submit_application_spec.rb
+++ b/spec/services/submit_application_spec.rb
@@ -2,20 +2,22 @@ require 'rails_helper'
 
 RSpec.describe SubmitApplication do
   describe 'Submit an application' do
-    let(:application_form) { create(:application_form) }
-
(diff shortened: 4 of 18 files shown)


benilovj:
Maybe a silly question or an approach that you've already discounted in discussion...

To achieve the same effect but with a lot less code changes, could you not look up the sandbox env in `TimeLimitConfig` and then configure the `edit_by` number to 0 for the sandbox env?

stevehook:
> Maybe a silly question or an approach that you've already discounted in discussion...
> 
> To achieve the same effect but with a lot less code changes, could you not look up the sandbox env in `TimeLimitConfig` and then configure the `edit_by` number to 0 for the sandbox env?

Not a silly question! As things stand the `edit_by` would still get set to the end of the day (normally the end of today unless it's not a business day). So the provider would still need to wait until the next morning in most cases. I did wonder about setting it to `-1` but I kind of felt that was a bit too much of a hack.

benilovj:
> > Maybe a silly question or an approach that you've already discounted in discussion...
> > To achieve the same effect but with a lot less code changes, could you not look up the sandbox env in `TimeLimitConfig` and then configure the `edit_by` number to 0 for the sandbox env?
> 
> Not a silly question! As things stand the `edit_by` would still get set to the end of the day (normally the end of today unless it's not a business day). So the provider would still need to wait until the next morning in most cases. I did wonder about setting it to `-1` but I kind of felt that was a bit too much of a hack.
> 
> Just to add I'm ok with doing it that way - we would still need to changes to `ReceiveReference` to make sure the transition is immediate.

Riiiight, yes, that's annoying.

Thinking out loud: it seems like `TimeLimitCalculator` is actually more a day calculator. The actual time seems to be calculated in the services themselves (`SetDeclineByDefault`, `SubmitApplication`, etc) which feels like a leaky abstraction. `TimeLimitCalculator` could be extended to be responsible for time calculations, meaning it encapsulates the end-of-day/business time stuff. That way you could also swap the calculator out with an `InstantCalculator` on Sandbox. However this approach above would probably require more work than you're probably willing to put in, given how far this PR is.

stevehook:
> > > Maybe a silly question or an approach that you've already discounted in discussion...
> > > To achieve the same effect but with a lot less code changes, could you not look up the sandbox env in `TimeLimitConfig` and then configure the `edit_by` number to 0 for the sandbox env?
> > 
> > 
> > Not a silly question! As things stand the `edit_by` would still get set to the end of the day (normally the end of today unless it's not a business day). So the provider would still need to wait until the next morning in most cases. I did wonder about setting it to `-1` but I kind of felt that was a bit too much of a hack.
> > Just to add I'm ok with doing it that way - we would still need to changes to `ReceiveReference` to make sure the transition is immediate.
> 
> Riiiight, yes, that's annoying.
> 
> Thinking out loud: it seems like `TimeLimitCalculator` is actually more a day calculator. The actual time seems to be calculated in the services themselves (`SetDeclineByDefault`, `SubmitApplication`, etc) which feels like a leaky abstraction. `TimeLimitCalculator` could be extended to be responsible for time calculations, meaning it encapsulates the end-of-day/business time stuff. That way you could also swap the calculator out with an `InstantCalculator` on Sandbox. However this approach above would probably require more work than you're probably willing to put in, given how far this PR is.

The `TimeLimitCalculator` was written to just deal in days so that it didn't have to know about business days but now given how it's used everywhere that makes less sense. Encapsulating the business days/end of day calculation would certainly reduce some duplication so I think you are right. I think it's worth a go...

app/services/time_limit_calculator.rb
```diff
@@ -19,8 +32,6 @@ def call
     default_time_limit&.limit
   end
 
-private
```
duncanjbrown: Should this be going?

app/services/receive_reference.rb
```diff
@@ -46,4 +46,11 @@ def referee_must_exist_on_application_form
       errors.add(:referee_email, 'does not match any of the provided referees')
     end
   end
+
+  def complete_references(application_choice)
+    ApplicationStateChange.new(application_choice).references_complete!
+    if application_choice.edit_by <= Time.zone.now
+      SendApplicationToProvider.new(application_choice: application_choice).call
```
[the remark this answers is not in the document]
stevehook: There's an argument for disabling all Slack notifications for the Sandbox environment.

app/services/time_limit_calculator.rb
stevehook: I think it's just moved up

spec/services/submit_application_spec.rb
```diff
@@ -2,20 +2,22 @@
 
 RSpec.describe SubmitApplication do
   describe 'Submit an application' do
-    let(:application_form) { create(:application_form) }
-
-    before do
+    def create_application_form
+      application_form ||= create(:application_form)
```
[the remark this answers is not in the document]
stevehook: that's a mistake

app/services/receive_reference.rb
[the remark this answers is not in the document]
stevehook: If nobody looks at the Slack notifications in `sandbox` I'm happy to wrap this in an `unless HostingEnvrionment.sandbox?` or we can add a more generic `DISABLE_SLACK_NOTIFICATIONS` env var etc.?

app/services/time_limit_calculator.rb
duncanjbrown: My bad 🙂

app/lib/hosting_environment.rb
```diff
@@ -53,4 +53,8 @@ def self.review?
   def self.production?
     environment_name == 'production'
   end
+
+  def self.sandbox?
+    ENV.fetch('SANDBOX') { 'false' } == 'true'
```
duncanjbrown:
```suggestion
    ENV.fetch('SANDBOX', 'false') == 'true'
```

tijmenb:
As discussed on Slack, I think this PR won't work because it relies on `ReceiveReference`. Since https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1080 that's no longer used by the candidate-facing code, and should have been removed in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1108.

PR #1146, opened 2020-01-20
Retain course selection from find

davidgisbey:
## Context

A new users course selection from find is currently not being retained through the sign-up process for a non-signed in user.

This PR contains the initial stages required to implement this feature.
 
- Pass courseCode and providerCode params through the apply, eligibility and sign-up pages/controllers
- Migration to add a column to the Candidate model for storing the selected course id
- Logic to save this id if is a new candidate

**Out of Scope(to be added in follow up PRs)** 

- Users who selected a course with one site will be redirected to the course choices page with their selected course
- Users who selected a course with multiple sites will be redirected to the site selection page for the relevant course
- Users who have already signed up to apply

## Guidance to review

You could test this by running the app locally and signing up with a course form an opted in provider e.g. '1N1' '238T'

Any feedback would be appreciated.

## Link to Trello card

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/sign_up_controller.rb
+++ b/app/controllers/candidate_interface/sign_up_controller.rb
@@ -29,7 +29,13 @@ module CandidateInterface
   private
 
     def candidate_sign_up_form_params
-      params.require(:candidate_interface_sign_up_form).permit(:email_address, :accept_ts_and_cs)
+      params.require(:candidate_interface_sign_up_form).permit(:email_address, :accept_ts_and_cs).merge(course_id: course_id)
+    end
+
+    def course_id
+      @provider = Provider.find_by(code: params[:providerCode])
+      @course = @provider.courses.find_by(code: params[:courseCode]) if @provider.present?
+      @course.id if @course.present?
     end
   end
 end
--- a/app/models/candidate_interface/sign_up_form.rb
+++ b/app/models/candidate_interface/sign_up_form.rb
@@ -1,7 +1,7 @@
 module CandidateInterface
   class SignUpForm
     include ActiveModel::Model
-    attr_accessor :email_address, :accept_ts_and_cs, :candidate
+    attr_accessor :email_address, :accept_ts_and_cs, :candidate, :course_from_find_id
 
     validates :email_address, :accept_ts_and_cs, presence: true
     validates :email_address, length: { maximum: 100 }
@@ -12,6 +12,7 @@ module CandidateInterface
       @email_address = params[:email_address]
       @accept_ts_and_cs = params[:accept_ts_and_cs]
       @candidate = Candidate.for_email @email_address
+      @course_from_find_id = params[:course_from_find_id]
     end
 
     def existing_candidate?
@@ -21,6 +22,7 @@ module CandidateInterface
     def save
       return false if existing_candidate? || !valid?
 
+      candidate.course_from_find_id = course_from_find_id
       candidate.save
     end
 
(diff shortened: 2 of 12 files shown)


app/controllers/candidate_interface/sign_up_controller.rb
```diff
@@ -29,7 +32,12 @@ def show; end
   private
 
     def candidate_sign_up_form_params
-      params.require(:candidate_interface_sign_up_form).permit(:email_address, :accept_ts_and_cs)
+      params.require(:candidate_interface_sign_up_form).permit(:email_address, :accept_ts_and_cs).merge(course_id: @course_id)
+    end
+
+    def build_course
```
tvararu: Maybe this could be named `course_id` and be used in `candidate_sign_up_form_params`?

```ruby
def course_id
  @provider = Provider.find_by(code: params[:providerCode])
  @course = @provider.courses.find_by(code: params[:courseCode])
  if @course.present?
    @course.id
  else
    nil
  end
end
```

(Rubocop will probably complain about this code)

tvararu:
Have also manually tested this on the review app and the params do get persisted throughout the sign up journey.

app/controllers/candidate_interface/sign_up_controller.rb
davidgisbey: Good suggestion. Have actioned 👍

app/controllers/candidate_interface/sign_up_controller.rb
```diff
@@ -29,7 +29,13 @@ def show; end
   private
 
     def candidate_sign_up_form_params
-      params.require(:candidate_interface_sign_up_form).permit(:email_address, :accept_ts_and_cs)
+      params.require(:candidate_interface_sign_up_form).permit(:email_address, :accept_ts_and_cs).merge(course_id: course_id)
+    end
+
+    def course_id
+      @provider = Provider.find_by(code: params[:providerCode])
+      @course = @provider.courses.find_by(code: params[:courseCode]) if @provider.present?
```
tijmenb: These don't need to be instance variables because we're not using them elsewhere.

app/models/candidate_interface/sign_up_form.rb
```diff
@@ -12,6 +12,7 @@ def initialize(params = {})
       @email_address = params[:email_address]
       @accept_ts_and_cs = params[:accept_ts_and_cs]
       @candidate = Candidate.for_email @email_address
+      @course_from_find_id = params[:course_from_find_id]
```
tijmenb: Hmm, does this work? It looks like we're passing in `course_id` instead of `course_from_find_id` from the controller.

db/migrate/20200120120940_add_course_from_find_id_to_candidate_model.rb
```diff
@@ -0,0 +1,9 @@
+class AddCourseFromFindIdToCandidateModel < ActiveRecord::Migration[6.0]
+  def up
+    add_column :candidates, :course_from_find_id, :integer, default: nil
```
tijmenb: You could use https://apidock.com/rails/v6.0.0/ActiveRecord/ConnectionAdapters/SchemaStatements/add_reference here, which is reversible / adds some database constraints.

```rb
add_reference(:candidates, :course_from_find, foreign_key: { to_table: :courses }, foreign_key: true)
```

app/models/candidate_interface/sign_up_form.rb
tijmenb: Ah, you already fixed it https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1165/commits/b5ea07666d50982c990dd208900e80a7a0500e36 👍

PR #1150, opened 2020-01-20
Add endpoint for Notify callback

chubberlisk:
## Context

We want to know whether or not an email to a reference has bounced. Previous PRs include: 

- https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1135 to add a Notify reference (`#{current_environment}-#{email_type}-#{reference_id}`) when we send a request for a reference to a referee
- https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1141 to add in the `GOVUK_NOTIFY_CALLBACK_API_KEY` environment variable

This PR creates an endpoint for Notify to [callback](https://docs.notifications.service.gov.uk/ruby.html#callbacks) to us when an email is sent via the service:

```
POST /notify/callback
```

A token is required and must match the value of the `GOVUK_NOTIFY_CALLBACK_API_KEY` environment variable.

As we're only able to add in a single callback URL on Notify, we have to do some filtering using the Notify reference we added on the environment e.g. `qa`, and email type e.g. `reference_request`.

**IF** the `reference` provided by Notify contains:

1. the same environment name as the current environment
2. the email type of `reference_request`

**AND**

the status of:

3. `permanent-failure`

**THEN**

We update the `feedback_status` of the reference we tried to send an email.

image

## Guidance to review

Would like feedback on:

- the naming of the service `ProcessNotifyCallback`... Feels too generic?
- the HTTP status and responses provided... Do they make sense?
- adding `skip_before_action :verify_authenticity_token` in `NotifyController`... Does this make sense? I had to add this in order for Notify to call us, otherwise I received a `Can't verify CSRF token authenticity` error.
- the fact that the endpoint isn't in a namespace... Should one be created or is it fine to be high-level?

## Link to Trello card

https://trello.com/c/FuEKJx9W/772-bounced-email-monitoring

## Things to check


app/controllers/notify_controller.rb
```diff
@@ -0,0 +1,62 @@
+class NotifyController < ApplicationController
+  skip_before_action :verify_authenticity_token
+
+  def callback
+    return render_unauthorized if unauthorized?
+    return render_unprocessable_entity if reference_or_status_missing?
```
tijmenb: We don't need this if we change the line below to `params.fetch(:reference)` / `params.fetch(:status)`, because that'll cause Rails to do the same thing.

app/controllers/notify_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+
+private
+
+  def unauthorized?
+    authorization_token.nil? || authorization_token != ENV.fetch('GOVUK_NOTIFY_CALLBACK_API_KEY')
+  end
+
+  def authorization_token
+    request.headers['Authorization']&.gsub('Bearer ', '')
+  end
```
tijmenb: Rails actually has a - well-hidden - feature to to this stuff, you can see it in action here:

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/commit/36c71e86d37e4e4fbffee969ee839edc4645edf3

app/services/process_notify_callback.rb
```diff
@@ -0,0 +1,29 @@
+class ProcessNotifyCallback
+  class << self
+    def call(notify_reference:, status:)
+      @environment, @email_type, reference_id = notify_reference.split('-')
+      @status = status
```
tijmenb: (talk about offline).

PR #1151, opened 2020-01-20
Add product page for signed out providers

paulrobertlloyd:
## Context

Rather than show just a sign-in button, we can use the provider landing page to market the service, primarily SCITTs that may be concerned that the Apply service and it’s Manage counterpart would be difficult to use. We want to show them that other providers are adopting it, and it pays to get onboard now to influence the design of the service.

The design of this page is influenced by the design of product pages for GOV.UK Notify, Pay, Verify etc.

`/provider` no longer redirects to `/provider/sign-in` if user is signed out
`/provider` still redirects to `/provider/applications` if user is signed in

## Product page

marketing-page-for-providers

## Guidance to review

- [x] Accessibility (alt text where provided, correct markup where using decorative images) @adamsilver
- [x] Content, typos etc (@emmajhf)
- [x] Ensure routes and controllers work as expected.
- [x] Calls to action in masthead. The ‘sign up’ button currently links to our support email address, is that correct? The ’sign in’ link points to the existing sign in page, but that feels… odd. We can’t use a link as DfE Sign In requires a post request.
- [x] I have refactored application layouts, pulling out portions (header, footer, cookie and phase banners) into partials, and having `application` layout inherit from a new base layout called `layout`. Is this the right approach? This was needed because the product page can’t have `govuk-width-container` be on the parent container (the blue masthead bleeds across the full width of the page).

## Still to do

- [x] Add image of the Apply team (uses placeholder image for now)
- [x] Ensure we have the right provider logos

## Link to Trello card

https://trello.com/c/fTr4UmMF/749


emmajhf:
Hi @paulrobertlloyd. This is looking great. Just one tiny change: please can we have a full stop at the end of the Gorse quote?

duncanjbrown:
Discussed with @paulrobertlloyd yesterday: we need to decide what to do about the Sign up button on this page (it has nowhere to go)

app/controllers/provider_interface/product_page_controller.rb
```diff
@@ -0,0 +1,13 @@
+module ProviderInterface
+  class ProductPageController < ProviderInterfaceController
+    skip_before_action :authenticate_provider_user!, :check_data_sharing_agreements
+
+    layout 'layout'
+
+    def show
+      courses = Course.includes(:provider).visible_to_candidates
```
tijmenb:
```suggestion
      courses = Course.includes(:provider).applyable
```

`applyable` courses are the courses that are enabled on Apply, `visible_to_candidates` are courses that are visible on Find.

paulrobertlloyd: Done!

tijmenb: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1166 renames the scopes to be more descriptive, hopefully that makes it clearer.

paulrobertlloyd:
Code reviewed with @duncanjbrown. Dependency on fixing some routing bugs with signed in providers who don’t yet have access. Duncan’s looking at that now. Once that is merged into master (and we have added team photo), this will be ready to merge.

PR #1154, opened 2020-01-21
Upgrade to Ruby 2.6.5

stevehook:
## Context

This upgrade was prompted by Github actions dropping support for 2.6.3 and a bunch of failing builds. Ruby 2.6.5 has been out since October. 

https://github.com/actions/virtual-environments/issues/281

- Upgrade `Gemfile` to Ruby 2.6.5
- Make Install Ruby step in Github actions config less specific (as suggested in https://github.com/actions/setup-ruby)
- Update `Dockerfile`
- Update `README.md`

## Guidance to review

- is this safe?
- is there any additional config to be done with the deployment?
- how do we test the deployment?

(Note that just making the changes to `.github/workflows/tests.yml` and leaving the Ruby version on 2.6.3 does fix the build issue)

## Link to Trello card

n/a

## Things to check


.github/workflows/tests.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       - name: 'Set up Ruby 2.6'
         uses: actions/setup-ruby@v1
         with:
-          ruby-version: 2.6.3
+          ruby-version: 2.6
```
alexdesi: this is '2.6.5' in our Gemfile

stevehook: Github seem to be recommending this. We could change it to `2.6.5` to make the requirement more strict - we would then get a failing build in the future if and when Github drop `2.6.5` like we did with `2.6.3`. What do you think?

PR #1155, opened 2020-01-21
Add Climate Control to bounced email specs

chubberlisk:
## Context

Specs added in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1135 and https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1150 related to bounced emails passed build but when running tests using Rake, they failed. This was because we assumed that the environment tests would run in would be `test`. However, when doing `bundle exec rake` tests are run in `development`, this seems to be a wider thing so this is sort of a work around but also just better practice.

This PR adds `ClimateControl` to set the environment variable `HOSTING_ENVIRONMENT_NAME` in the `referee_mailer_spec.rb` and `process_notify_callback_spec.rb`.

## Guidance to review

Does this make sense?

## Link to Trello card

N/A

## Things to check

### DIFF
--- a/spec/mailers/referee_mailer_spec.rb
+++ b/spec/mailers/referee_mailer_spec.rb
@@ -29,9 +29,12 @@ RSpec.describe RefereeMailer, type: :mailer do
       expect(body).to include(referee_interface_reference_feedback_url(token: ''))
     end
 
-    it 'sends a request with a notify reference' do
-      mail.deliver_now
-      expect(mail[:reference].value).to eq("test-reference_request-#{reference.id}")
+    it 'sends a request with a Notify reference' do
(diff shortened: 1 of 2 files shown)


duncanjbrown:
:+1: to clarify a bit: the tests are run in `RAILS_ENV=test` but `dotenv-rails` loads the `.env.development` file when `rake` first sets up the environment and and then doesn't override it with `.env.test` when `rspec` happens within that `rake` environment. The behaviour we see is that the value of `HOSTING_ENVIRONMENT_NAME` from `.env.development` has precedence under `be rake` but not under plain `be rspec`.

PR #1157, opened 2020-01-21
Add sending of new referee request email when email bounces

chubberlisk:
## Context

We want to be able to automatically send an email requesting that a candidate provides a new referee or email address if the reference request email bounces. 

In #1150, we added in an endpoint for Notify to let us know when an email has bounced. When this happens, we update the feedback status of that referee to `email_bounced`. This is done in the `ProcessNotifyCallback` service. 

This PR adds the sending of a new referee request email with the reason that the email has bounced when Notify has said there's been a `permanent-failure` on a reference request email by updating the  `ProcessNotifyCallback` service.

## Guidance to review

Does spying on whether or not we have called `SendNewRefereeRequestEmail` make sense?

## Link to Trello card

https://trello.com/c/dfpE95vU/784-bounced-email-referee-strategic

## Things to check


app/services/process_notify_callback.rb
```diff
@@ -7,7 +7,14 @@ def call(notify_reference:, status:)
       return :not_updated unless same_environment? && reference_request_email? && permanent_failure_status?
       return :not_found unless ApplicationReference.exists?(reference_id)
 
-      ApplicationReference.find_by(id: reference_id).update!(feedback_status: 'email_bounced')
+      reference = ApplicationReference.find_by(id: reference_id)
+      reference.update!(feedback_status: 'email_bounced')
```
tijmenb: This makes 2 queries to check/fetch the reference. You could slim this down by using `ApplicationReference.find(reference_id)` and rescuing `ActiveRecord::RecordNotFound` at the end of the method.

app/services/process_notify_callback.rb
```diff
@@ -7,7 +7,14 @@ def call(notify_reference:, status:)
...
+
+      SendNewRefereeRequestEmail.call(
+        application_form: reference.application_form,
+        reference: reference,
+        reason: :email_bounced,
+      )
```
tijmenb: Should we wrap this in a transaction with the update?

spec/services/process_notify_callback_spec.rb
```diff
@@ -51,6 +51,16 @@
 
         expect(process_notify_callback).to eq(:not_found)
       end
+
+      it 'send a new referee request email to the candidate' do
+        allow(SendNewRefereeRequestEmail).to receive(:call)
+          .with(application_form: application_form, reference: reference, reason: :email_bounced)
```
tijmenb: I wonder if we want more of an integration test for this (in system or request specs), to avoid testing the internals of the service here.

chubberlisk: Makes sense! Done in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1157/commits/70bb3012837a2bb8368b9f3656acf6afcea9414b.

app/services/process_notify_callback.rb
chubberlisk: Makes sense! Done in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1157/commits/b99f0893026bbf26f8fa4aa4d71778f8492a4756.

app/services/process_notify_callback.rb
chubberlisk: Didn't know this was a thing. Makes sense! Done in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1157/commits/b99f0893026bbf26f8fa4aa4d71778f8492a4756.

chubberlisk:
@tijmenb I've updated the PR with your comments now. They're currently as separate commits to make it easier to review but I might fix them up if approval.

PR #1158, opened 2020-01-21
Add `sign_up_email_bounced` attribute to Candidate

chubberlisk:
## Context

We want to be able to surface on Support when an email sent to a candidate has been bounced like we've done when a candidate has never signed in.

image

However, in order to do so we'll need something on the candidate to keep track of this. 

## Changes proposed in this pull request (Updated: Tues 21 Jan 16:06)

This PR adds a migration to add ~~`email_bounced`~~ `sign_up_email_bounced` attribute for a `Candidate`.

## Guidance to review

Naming is hard. Any suggestions?

## Link to Trello card

https://trello.com/c/eRWPVFng/783-bounced-email-candidate-strategic

## Things to check


db/migrate/20200121121144_add_email_bounced_to_candidate.rb
```diff
@@ -0,0 +1,5 @@
+class AddEmailBouncedToCandidate < ActiveRecord::Migration[6.0]
+  def change
+    add_column :candidates, :email_bounced, :boolean, default: false, null: false
```
tijmenb: Perhaps:

```suggestion
    add_column :candidates, :signup_email_bounced, :boolean, default: false, null: false
```

db/schema.rb
```diff
@@ -148,7 +148,8 @@
     t.string "magic_link_token"
     t.datetime "magic_link_token_sent_at"
     t.boolean "hide_in_reporting", default: false, null: false
-    t.bigint "course_from_find_id"
```
tijmenb: This change isn't intentional I think?

chubberlisk: 🔍 Noice spot, fixed now!

PR #1159, opened 2020-01-21
Fix skip the 'waiting to be sent' state in sandbox 

stevehook:
## Context

The business logic that was originally in the `ReceiveReference` service class moved to `RefereeInterface::ReferenceFeedbackForm`. While this was happening https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1140/files modified `ReceiveReference` with some extra logic needed to support Sandbox. This logic needs to be moved to `RefereeInterface::ReferenceFeedbackForm`.

We also need to remove the `ReceiveReference` class but this will be covered in a separate PR.

- Add logic to `RefereeInterface::ReferenceFeedbackForm` to move straight to `awaiting_provider_decision` when all references are complete and `edit_by` is in the past (the edit by time limit has expired).

## Guidance to review

- Is the one new test adequate?
- Are we mixing too many concerns into the form object here?

Note that https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1140 already ensures that `edit_by` is set to the current time when operating in the Sandbox.

## Link to Trello card

Original card was:

## Things to check

### DIFF
--- a/app/models/referee_interface/reference_feedback_form.rb
+++ b/app/models/referee_interface/reference_feedback_form.rb
@@ -13,12 +13,10 @@ module RefereeInterface
     def save
       return unless valid?
 
-      ActiveRecord::Base.transaction do
-        ReceiveReference.new(
-          reference: reference,
-          feedback: feedback,
-        ).save
-      end
+      ReceiveReference.new(
+        reference: reference,
+        feedback: feedback,
+      ).save
     end
   end
 end
--- a/app/models/referee_interface/reference_feedback_form.rb
+++ b/app/models/referee_interface/reference_feedback_form.rb
@@ -15,8 +15,7 @@ module RefereeInterface
 
       ActiveRecord::Base.transaction do
         ReceiveReference.new(
-          application_form: reference.application_form,
-          referee_email: reference.email_address,
+          reference: reference,
           feedback: feedback,
         ).save
       end
--- a/app/models/referee_interface/reference_feedback_form.rb
+++ b/app/models/referee_interface/reference_feedback_form.rb
@@ -14,18 +14,12 @@ module RefereeInterface
       return unless valid?
 
       ActiveRecord::Base.transaction do
-        reference.update!(feedback: feedback, feedback_status: 'feedback_provided')
-
-        # If all of the references have been provided we need to change the
-        # state of the application choices
-        if reference.application_form.application_references_complete?
-          reference.application_form.application_choices.includes(:course_option).find_each do |application_choice|
-            ApplicationStateChange.new(application_choice).references_complete!
-          end
-        end
+        ReceiveReference.new(
+          application_form: reference.application_form,
+          referee_email: reference.email_address,
+          feedback: feedback,
+        ).save
       end
-
-      true
     end
   end
 end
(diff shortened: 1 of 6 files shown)


app/models/referee_interface/reference_feedback_form.rb
```diff
@@ -20,12 +20,21 @@ def save
...
       true
     end
+
+  private
+
+    def complete_references(application_choice)
+      ApplicationStateChange.new(application_choice).references_complete!
+      if application_choice.edit_by <= Time.zone.now
+        SendApplicationToProvider.new(application_choice: application_choice).call
+      end
```
tijmenb: Would it be possible to schedule a `SendApplicationsToProviderWorker` instead, so that we don't duplicate the logic?

stevehook: Yes we could do that - it will make the process slightly less immediate from the providers point of view but normally the workers should be pretty quick so I guess it's fine.

stevehook: On second thoughts I think it should stay as is.

If we schedule a worker I'm not sure what we would do with the various specs etc. that invoke `ReceiveReference` to move the status on. We couldn't just drop `ReferenceFeedbackForm` in as a replacement. And we could end up with multiple instances of that worker running in parallel which is likely to cause errors (unless we specifically guard against that somehow).

duncanjbrown:
Late question, sorry: doesn’t the fact that this logic is in the namespace violate our architecture principle that business logic lives in shared services? https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/adr/0005-application-structure.md. Is the correct fix for this to move the logic back into ReceiveReference?

stevehook:
> Late question, sorry: doesn’t the fact that this logic is in the namespace violate our architecture principle that business logic lives in shared services? https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/adr/0005-application-structure.md. Is the correct fix for this to move the logic back into ReceiveReference?

Yes that makes sense. Still about confused about the appropriate division of labour between form objects and services but to me this does belong in a service. The form can call that service.

We discussed (on slack) getting rid of `ReceiveReference` after having moved the logic to the form. There are a few other bits of code that currently uses `ReceiveReference (e.g. specs, cucumber features and test data generator). It makes more sense to continue having these call into a service than a form.

What do you think @tijmenb ?

duncanjbrown:
To add to the list: `GenerateTestApplications`, which also uses `ReceiveReference` (and now breaks if `SANDBOX=true`)

tijmenb:
@stevehook sounds good!

stevehook:
> To add to the list: `GenerateTestApplications`, which also uses `ReceiveReference` (and now breaks if `SANDBOX=true`)

Good spot, I added a fix for that just now so that `SendApplicationToProvider` checks the status before attempting to change it.

spec/models/referee_interface/reference_feedback_form_spec.rb
```diff
@@ -36,5 +36,23 @@
       expect(application_form).not_to be_application_references_complete
       expect(application_form.application_choices).to all(be_awaiting_references)
     end
+
+    it 'progresses the application choices to the "awaiting_provider_decision" status once all references have been received if edit_by expired' do
```
duncanjbrown: Does this spec now duplicate `services/receive_reference_spec.rb`? (not just the bit changed here but the two preceding specs too)

stevehook:
> Great! Sorry for the bitty questions

no problem, keep them coming!

PR #1162, opened 2020-01-21
Update content when referee refuses to give a reference

tvararu:
## Context

Currently we thank users for submitting a reference even if they refuse. This is confusing.

Check the `@reference` in the view and display conditional content.

## Guidance to review

Should we hoist the `@reference.feedback_status == 'feedback_refused'` logic into a `#method?` on the model?

Review without whitespace.

## Before

Screenshot 2020-01-21 at 15 28 03

## After

Screenshot 2020-01-21 at 15 27 27

## Link to Trello card

https://trello.com/c/e2WQISLg/790-update-content-when-user-declines-to-give-a-reference

## Things to check

### DIFF
--- a/app/views/referee_interface/reference/confirmation.html.erb
+++ b/app/views/referee_interface/reference/confirmation.html.erb
@@ -1,19 +1,27 @@
-<% content_for :title, t('page_titles.give_a_reference.confirmation') %>
-<div class="govuk-panel govuk-panel--confirmation govuk-!-margin-bottom-8">
-  <h1 class="govuk-panel__title">
-    Your reference for <%= @reference.application_form.full_name %> has been&nbsp;submitted
-  </h1>
-</div>
+<% if @reference.feedback_refused? %>
+  <% content_for :title, t('page_titles.give_a_reference.finish') %>
+  <h1 class="govuk-heading-xl">Thank you</h1>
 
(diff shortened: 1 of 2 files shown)


app/views/referee_interface/reference/confirmation.html.erb
```diff
@@ -1,19 +1,27 @@
-<% content_for :title, t('page_titles.give_a_reference.confirmation') %>
-<div class="govuk-panel govuk-panel--confirmation govuk-!-margin-bottom-8">
-  <h1 class="govuk-panel__title">
-    Your reference for <%= @reference.application_form.full_name %> has been&nbsp;submitted
-  </h1>
-</div>
+<% if @reference.feedback_status == 'feedback_refused' %>
```
duncanjbrown: It would be nice to be able to say `if @reference.feedback_refused?` but that might require some wrangling of the enum (see bottom of page here: https://api.rubyonrails.org/v5.2.4.1/classes/ActiveRecord/Enum.html).

tvararu: `feedback_status_feedback_refused?` should work, thanks for reminding me of this feature! I think it's preferable to the `==`

tvararu: Oh, it's even simpler, `feedback_refused?` works OOTB!

PR #1165, opened 2020-01-21
A candidates selected course from find is displayed after authentication

davidgisbey:
## Context

In a previous PR #1146, if a user arrives from Find with providerCode and courseCode parameters, they are passed through to the signup page. The course id was then added to **new** candidates via the signup form to the candidate model.

This PR builds on this work by checking for the presence of course_from_find_id and redirecting candidates to the appropriate page for their selected course.

The desired behaviour is:

If course_from_find_id is present on a candidate and the course has one site.
  - It should add an ApplicationChoice with the relevant CourseOption. 
  - Redirect to the candidate_interface_course_choices_review_path 
  - Set the candidates course_from_find_id to nil

elsif course_from_find_id is present on a candidate and the course has multiple sites
  - It should redirect the site selection page for their selected course
  - set the candidates course_from_find_id to nil

else
  - It should redirect to the candidate's application page

## Guidance to review

This needs a  refactor as it's fairly tricky to parse at the moment.

**Out of scope**

- Previously persisted candidates
- Candidates can alter the query string to create an invalid course on the eligibility and sign up pages

## Link to Trello card

## Things to check


spec/system/candidate_interface/candidate_new_user_with_course_params_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+
+    within_fieldset('Did you gain all your qualifications at institutions based in the UK?') do
+      choose 'Yes'
+    end
+
+    click_on 'Continue'
+  end
+
+  def when_i_submit_my_email_address
+    perform_enqueued_jobs do
```
davidgisbey: This had to be added as the magic link uses an async call on sign up. This was causing `open_email(@email)` to return nil. Interestingly this did not occur on other specs which use the same service. Adding the query string seems to be the case 🤷‍♂

tijmenb: You can use `sidekiq: true` to run the jobs inline:

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/c136e7c134859301a7105ece79d974606d403a09/spec/system/candidate_interface/candidate_accepts_offer_spec.rb#L3

davidgisbey: I actually tried this earlier and still got the following error

image

tijmenb: That's awesome - a mystery! Lets investigate tomorrow.

spec/system/candidate_interface/candidate_new_user_with_course_params_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+  scenario 'retaining their course selection through the sign up process' do
+    given_the_pilot_is_open
+    and_confirm_course_choice_from_find_is_activated
+    and_the_course_i_selected_only_has_one_site
+
+    when_i_arrive_from_find_to_a_course_that_is_open_on_apply
+    and_i_click_apply_on_apply
+    then_the_url_should_contain_the_course_code_and_provider_code_param
+    when_i_fill_in_the_eligiblity_form_with_yes
+    then_the_url_should_contain_the_course_code_and_provider_code_param
```
tijmenb: Perhaps we drop the tests for URLs? They are an implementation detail - generally in these tests we shouldn't care about _how_ something happens, just that it does.

spec/system/candidate_interface/candidate_new_user_with_course_params_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+    and_i_should_see_the_site(@site)
+    and_my_find_from_id_course_should_be_set_to_nil
+
+    given_the_course_i_selected_has_multiple_sites
+
+    when_i_submit_my_email_address
+    and_click_on_the_magic_link
+    then_i_should_see_the_course_choices_site_page
+    and_i_should_see_the_site(@site1)
+    and_i_should_see_the_site(@site2)
```
tijmenb: What would you think about combining these into a step like `and_i_see_the_form_to_pick_a_location`?

spec/system/candidate_interface/candidate_new_user_with_course_params_spec.rb
tijmenb: Confirmed this on my machine, I don't really understand why but fine for now!

app/controllers/candidate_interface/sign_in_controller.rb
```diff
@@ -24,11 +24,19 @@ def create
...
     def authenticate
       candidate = FindCandidateByToken.call(raw_token: params[:token])
-
       if candidate
         sign_in(candidate, scope: :candidate)
         add_identity_to_log candidate.id
-        redirect_to candidate_interface_application_form_path
+        course_id = candidate.course_from_find_id
+        service = ExistingCandidateAuthentication.new(candidate: candidate).execute
+        if service == :candidate_has_new_course_added
```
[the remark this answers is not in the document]
tvararu: I like this!

davidgisbey:
Thanks so much for all the feedback. It was super helpful 🙏

PR #1166, opened 2020-01-21
Only show providers/courses that Apply offers

tijmenb:
## Context

The courses page currently shows all courses that are in our system, and aren’t hidden in find (drafts).

We should be showing the courses that candidates can apply to on Apply instead.

Renamed the scopes together with @tvararu to avoid future confusion.

## Guidance to review

Do the scope names make sense?

## Link to Trello card

N/A

## Things to check

### DIFF
--- a/app/models/candidate_interface/pick_course_form.rb
+++ b/app/models/candidate_interface/pick_course_form.rb
@@ -6,13 +6,13 @@ module CandidateInterface
     validates :code, presence: true
     validate :user_cant_apply_to_same_course_twice
 
-    def applyable?
+    def open_on_apply?
       course.open_on_apply?
     end
 
     def available_courses
       @available_courses ||= begin
-        provider.courses.visible_to_candidates.order(:name)
+        provider.courses.exposed_in_find.order(:name)
       end
     end
 
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -7,8 +7,8 @@ class Course < ApplicationRecord
   validates :level, presence: true
   validates :code, uniqueness: { scope: :provider_id }
 
-  scope :visible_to_candidates, -> { where(exposed_in_find: true) }
-  scope :applyable, -> { visible_to_candidates.where(open_on_apply: true) }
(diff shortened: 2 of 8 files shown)


app/models/candidate_interface/pick_course_form.rb
```diff
@@ -7,12 +7,12 @@ class PickCourseForm
     validate :user_cant_apply_to_same_course_twice
 
     def applyable?
-      course.open_on_apply?
+      course.open_on_apply? && course.exposed_in_find?
```
duncanjbrown: This duplicates the logic from the `open_on_apply` scope — is there a way to dry it up?

app/models/course.rb
```diff
@@ -7,8 +7,12 @@ class Course < ApplicationRecord
   validates :level, presence: true
   validates :code, uniqueness: { scope: :provider_id }
 
-  scope :visible_to_candidates, -> { where(exposed_in_find: true) }
-  scope :applyable, -> { visible_to_candidates.where(open_on_apply: true) }
+  # Also filter on `exposed_in_find` as to hide courses that are hidden in Find,
+  # but are enabled by a support user in Apply
+  scope :open_on_apply, -> { where(open_on_apply: true, exposed_in_find: true) }
```
duncanjbrown: It's mildly confusing that we have a scope and a flag called `open_on_apply`, and something is only *open* open if it's matched by the scope

tijmenb: That's why we have the comment! Do you have another naming suggestion?

app/models/candidate_interface/pick_course_form.rb
tijmenb: Nope, they're boolean attributes. I don't see any way to combine those?

app/models/course.rb
duncanjbrown: Is it true to say that something which is only `open_on_apply` is actually open, because if it's not `applyable?` then `CourseChoicesController` will tell them they have to apply on UCAS.

PR #1167, opened 2020-01-21
Add work breaks betweek work sections (1st part)

alexdesi:
## Context
Providers need to understand why the candidate has gaps in their work history

- This provides work break sections between work sections.
- To see the changes, activate the feature flag **work_breaks**

## Guidance to review

candidate_entering_work_history_spec.rb is a good place to start.

## Link to Trello card


alexdesi:
Work break:
Screenshot 2020-01-21 at 17 43 44

spec/system/candidate_interface/candidate_entering_work_history_spec.rb
```diff
@@ -4,55 +4,65 @@
   include CandidateHelper
 
   scenario 'Candidate submits their work history' do
-    given_i_am_signed_in
-    and_i_visit_the_site
+    FeatureFlag.activate('work_breaks')
 
-    when_i_click_on_work_history
-    then_i_should_see_a_list_of_work_lengths
+    Timecop.travel(Time.zone.local(2019, 12, 1)) do
```
tijmenb: Can we avoid wrapping this in a block? You could use `Timecop.travel` without the block, use an `around` statement, or only wrap the code in a step.

app/models/application_work_experience.rb
```diff
@@ -9,4 +9,22 @@ class ApplicationWorkExperience < ApplicationExperience
   }
 
   audited associated_with: :application_form
+
+  def next_work_breaks_in_months
```
tijmenb: Can we extract this from the model? I also suspect it [duplicates some of CheckBreaksInWorkHistory](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/app/services/check_breaks_in_work_history.rb), can you check?

alexdesi: Basically, what `CheckBreaksInWorkHistory` does (I think we need a better name for it)
is to find out if there are breaks between works, just true/false.
What now we need is the entire list of gaps in months.
I am going to extract these methods here in a new service that returns the list of work breaks, at that point `CheckBreaksInWorkHistory` will be redundant.

spec/system/candidate_interface/candidate_entering_work_history_spec.rb
alexdesi: I realized that, in that case, we do not needed to freeze the time.

alexdesi:
Hi guys, I'd like to point out two things:
- since we need the list of all the breaks, I create a new service: `GetBreaksInWorkHistory`
- the existent `CheckBreaksInWorkHistory` will be removed in the following PR.

app/services/get_breaks_in_work_history.rb
```diff
@@ -0,0 +1,32 @@
...
+        months = next_work_breaks_in_months(work)
+        breaks[work.id] = months unless months.zero?
+      end
+
+      breaks
+    end
+
+  private
+
+    def next_work_breaks_in_months(work)
```
tvararu:
```suggestion
    def next_work_break_in_months(work)
```

Just a grammar nit

app/services/get_breaks_in_months.rb
```diff
@@ -0,0 +1,16 @@
+class GetBreaksInMonths
+  class << self
+    def call(start_date, end_date)
+      puts '>>>>'
+      puts start_date
+      puts end_date
```
tvararu: Can be removed

app/components/work_history_review_component.html.erb
```diff
@@ -1,15 +1,19 @@
-<% @application_form.application_work_experiences.order(:start_date).each do |work| %>
-  <%= render(SummaryCardComponent, rows: work_experience_rows(work), editable: @editable) do %>
-    <%= render(SummaryCardHeaderComponent, title: work.role, heading_level: @heading_level, check_icon: work.working_with_children) do %>
+<%@application_form.application_work_experiences.order(:start_date).to_a.append(nil).each_cons(2) do |work1, work2| %>
```
tvararu: Pseudocode for how component could deal with the situation where one of the first work experiences is a current work (it goes into the present, has a nil `end_date`)

```ruby
<% work_experiences = @application_form.application_work_experiences.order(:start_date) %>
<% work_experiences_with_breaks, work_experiences_without_breaks = work_experiences.to_a.partition { |w| w.end_date.nil? }

<% work_experiences_with_breaks.append(work_experiences_without_breaks.first).each_cons(2) do |work1, work2| %>
  <%= render(IndividualWorkExperienceReviewComponent, ...) %>
  <%= render(BreakInWorkHistoryComponent, work1: work1, work2: work2) %>
<% end %>

<% work_experiences_without_breaks.each do |work| %>
  <%= render(IndividualWorkExperienceReviewComponent, ...) %>
<% end %>
```

This can be done as a follow-up PR!

PR #1170, opened 2020-01-22
Auto complete references in Sandbox

stevehook:
## Context

In the Sandbox providers need to test the application flow without the normal delays in the process. Normally an application is not visible to the provider until 2 references have been received.

- [x] Define two hard-coded (for now at least) email addresses that providers can use to send reference requests too. (`refbot1@example.com` and `refbot2@example.com`)
- [x] Auto-complete the special references and trigger next step in lifecycle to move status on to `awaiting_provider_decision`. This is done by adding a check in the `SubmitApplication` service to immediately invoke `ReceiveReference` iff running in Sandbox and using on of the email addresses above.

## Guidance to review

- I moved the logic that auto-completes references to `SubmitApplication` (it was in a new `RequestReference` service but that makes less sense - it's too early in the application lifecycle because we don't request references until after submission). Hope that makes sense.
- If anyone can think of better names than `refbot1@example.com` etc. then happy to change.
- I've not attempted to update any of the copy on the various candidate pages, e.g. it still talks about the 5 day wait period.

## Link to Trello card

https://trello.com/c/DLiqTrop/1537-generate-2-email-addresses-that-when-entered-as-a-referee-will-return-a-reference-automatically

## Things to check


app/services/request_reference.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+private
+
+  def auto_approve_in_sandbox
+    @referee.feedback = I18n.t('new_referee_request.auto_approve_feedback') if sandbox? && email_address_is_a_bot?
+  end
+
+  def sandbox?
+    ENV.fetch('SANDBOX') { 'false' } == 'true'
```
duncanjbrown: `HostingEnvironment.sandbox?`

spec/services/request_reference_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+      expect(action.call).to be true
+      expect(action.referee.persisted?).to be true
+      expect(action.referee.name).to eq 'Bob'
+      expect(action.referee.feedback).to be_nil
+    end
+  end
+
+  context 'in the Sandbox environment' do
+    around do |example|
+      ClimateControl.modify(SANDBOX: 'true') do
```
duncanjbrown: Maybe we should make this into an rspec tag as in `spec/support/sidekiq.rb`?

app/services/request_reference.rb
```diff
@@ -0,0 +1,30 @@
+class RequestReference
+  attr_accessor :referee_params, :referee, :application_form
```
duncanjbrown: Maybe make `:referee` `attr_reader` to show it's an output?

stevehook:
sorry @duncanjbrown - I changed this a fair bit since your earlier review though 2/3 of the comments still apply.

PR #1173, opened 2020-01-22
Refactor the ReceiveReference class

tijmenb:
## Context

We originally intended to deprecate this class, but it turns out it’s still useful to extract logic from ReferenceFeedbackForm into a different class. https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1159 will change the form to use ReceiveReference instead of duplicating the logic.

Refactor the class to remove validation (that's done by the form) and to use an ApplicationReference object directly, instead of relying on an email lookup (which is a holdover from when we imported references using a CSV).

## Guidance to review

Does it make sense?

## Link to Trello card

## Things to check

### DIFF
--- a/app/lib/test_applications.rb
+++ b/app/lib/test_applications.rb
@@ -34,8 +34,7 @@ module TestApplications
 
       application_form.application_references.each do |reference|
         ReceiveReference.new(
-          application_form: application_form,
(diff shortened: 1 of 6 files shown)


duncanjbrown:
🙌

PR #1178, opened 2020-01-22
Test applications no longer have multiple identical course choices

duncanjbrown:
A defect was introduced in #1160 when we turned a loop that called `CourseOption.all.sample` into a call to `FactoryBot.create_list`. This resulted in all test applications having ApplicationChoices associated with the same CourseOption.

Also, instead of sampling a random CourseOption when we create the application — which may result in the candidate applying to the same course at two different sites, not permitted via the UI — ensure we pick distinct courses. If there aren't enough distinct courses, throw an exception.

Screenshot 2020-01-22 at 17 16 39

## Guidance to review

Try running `rake setup_local_dev_data`

## Link to Trello card

Towards https://trello.com/c/VLw7X81Z/1526-improvements-to-the-generatetestdata-api-feature-for-tribal

## Things to check

### DIFF
--- a/app/lib/test_applications.rb
+++ b/app/lib/test_applications.rb
@@ -1,11 +1,23 @@
 module TestApplications
-  def self.create_application(states)
+  def self.create_application(states:)
     first_name = Faker::Name.unique.first_name
     last_name = Faker::Name.unique.last_name
     candidate = FactoryBot.create(
       :candidate,
       email_address: "#{first_name.downcase}.#{last_name.downcase}@example.com",
     )
+
+    courses_to_apply_to = Course.joins(:course_options)
+      .where(open_on_apply: true)
+      .order('RANDOM()')
+      .limit(states.count)
+
+    # it does not make sense to apply to the same course multiple times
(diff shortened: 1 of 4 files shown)


app/lib/test_applications.rb
```diff
@@ -1,12 +1,22 @@
...
-  def self.create_application(states)
+  def self.create_application(states:)
     first_name = Faker::Name.unique.first_name
     last_name = Faker::Name.unique.last_name
     candidate = FactoryBot.create(
       :candidate,
       email_address: "#{first_name.downcase}.#{last_name.downcase}@example.com",
     )
 
+    course_pool = Course.joins(:course_options).where(open_on_apply: true).all
```
tijmenb: I think you could golf this down to:

```suggestion
    courses_to_apply_to = Course.joins(:course_options).where(open_on_apply: true).order('RANDOM()').limit(states.count)
```

duncanjbrown: Yes — more efficient too

PR #1182, opened 2020-01-23
Improvements to API test data generation

duncanjbrown:
## Context

Following conversations with Vendors about test data generation in the Sandbox, we've agreed to make some improvements to this feature of the API.

- deprecate `api/v1/test-data/regnerate`
- add `api/v1/test-data/generate` and `api/v1/test-data/clear`

The new `generate` endpoint 

- no longer wipes all data on each call
- supports a `courses_per_application=` parameter
- returns an array of the generated IDs

## Guidance to review

Use the API to generate some test applications.

## Link to Trello card

https://trello.com/c/VLw7X81Z/1526-improvements-to-the-generatetestdata-api-feature-for-tribal

## Things to check

- [X] API release notes have been updated if necessary — **not doing this (yet)**


duncanjbrown:
Added a commit to feature flag this so we don't commit to it before we have the chance to iterate

config/vendor-api-v1.yml
```diff
@@ -238,6 +239,45 @@ paths:
           $ref: "#/components/responses/OK"
         "401":
           $ref: "#/components/responses/Unauthorized"
+  "/test-data/generate":
```
tijmenb: Do we add this now, since it's not available without the feature flag? (If we do, should we add the feature flag?

spec/requests/vendor_api/post_test_data_spec.rb
```diff
@@ -1,31 +1,101 @@
 require 'rails_helper'
 
-RSpec.describe 'Vendor API - POST /api/v1/test-data/regenerate', type: :request do
```
tijmenb: Would it make sense to have a single file per endpoint?

config/vendor-api-v1.yml
duncanjbrown: Right, this shouldn't be in here until it's ready to release. Will remove for now

spec/requests/vendor_api/post_test_data_spec.rb
duncanjbrown: Yep, I'll break it up

app/controllers/vendor_api/test_data_controller.rb
```diff
@@ -1,17 +1,42 @@
...
       render json: { data: { message: 'OK, regenerated the test data' } }
     end
 
+    def generate
+      application_choices = (1..count_param).flat_map do
+        states = [:awaiting_provider_decision] * courses_per_application_param
+        TestApplications.create_application states: states, courses_to_apply_to: current_provider.courses
+      end
+
+      render json: { data: { ids: application_choices.map { |ac| ac.id.to_s } } }
```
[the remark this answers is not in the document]
duncanjbrown: I'll have a go

duncanjbrown:
Following conversations I'm going to put the endpoints back in the docs and move to an `experimental` section of the API

app/lib/test_applications.rb
```diff
@@ -1,13 +1,22 @@
 module TestApplications
-  def self.create_application(states:)
+  def self.generate_for_provider(provider:, courses_per_application:, count:)
+    1.upto(count).flat_map do
+      create_application(
+        states: [:awaiting_provider_decision] * courses_per_application,
+        courses_to_apply_to: provider.courses,
```
stevehook: Should the list of courses be limited to courses that have course options? I got a crash later for a course that didn't have any - I was using Gorse SCCIT auto-synced which seems to have 5 out of 27 courses without options (my database may be messed up though).

app/controllers/vendor_api/test_data_controller.rb
```diff
@@ -1,17 +1,43 @@
 module VendorApi
   class TestDataController < VendorApiController
     before_action :check_this_is_a_test_environment
+    before_action :feature_flag_new_endpoints, only: %i[generate clear!]
 
     MAX_COUNT = 100
     DEFAULT_COUNT = 100
```
stevehook: I know it's not new but this default always seemed high to me.

app/lib/test_applications.rb
stevehook: When I remove the courses with no options it works but I get 2200 new application choices. Not sure why - that's 100 * the number of courses.

app/controllers/vendor_api/test_data_controller.rb
```diff
@@ -1,17 +1,42 @@
...
+      application_choices = (1..count_param).flat_map do
+        states = [:awaiting_provider_decision] * courses_per_application_param
+        TestApplications.create_application states: states, courses_to_apply_to: current_provider.courses
+      end
+
+      render json: { data: { ids: application_choices.map { |ac| ac.id.to_s } } }
+    end
+
+    def clear!
+      current_provider.application_choices.map(&:delete)
```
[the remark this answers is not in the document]
duncanjbrown: It does! Maybe we need a dedicated deleting facility

duncanjbrown:
Still to do: fix runaway application choice creation bug identified by @stevehook, move new routes under `/experimental` or equiv

app/lib/test_applications.rb
duncanjbrown: Aha — we're mapping over all the courses instead of only enough to generate applications in the relevant states 🤕

app/controllers/vendor_api/test_data_controller.rb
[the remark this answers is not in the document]
tijmenb: See https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1218

stevehook:
@duncanjbrown Just did a quick test locally and looks like `generate` is doing what I would expect now - correct number of applications and not tripping over courses without options.

duncanjbrown:
This is now a bit of a monster 👹. Sorry.

It bundles in a new API feature called `experimental_api_features`. This is a family of endpoints, living under `/api/v1/experimental`, which can be toggled on and off by a feature flag. The flag affects the documentation as well as the app behaviour.

Experimental paths live in `config/vendor-api-experimental.yml` and are merged into the main docs  — including the downloaded YAML spec — if the flag is on. 

If this is really painful to review please ask, and I'll do the work to break it up a bit further 🌈 .

PR #1184, opened 2020-01-23
Add ability to see if candidate sign up email bounced in Support

chubberlisk:
## Context

We want to be able to see in Support if a candidate tried to sign up but the email bounced.

This PR adds the ability to see if candidate sign up email bounced in Support by:

- adding a Notify reference when we send a candidate a sign up email
- adding checking in `ProcessNotifyCallback` to see if an email with `sign_up_email` in the Notify reference and `permanent-failure` status, then updating `sign_up_email_bounced` and `hide_in_reporting` attributes of the candidate to `true`.
- displaying `Sign up email bounced` on the candidates page in Support

### Screenshots

image

## Guidance to review

- ~~Making a draft as want to test using Heroku preview app~~
- Review commit by commit

## Link to Trello card

https://trello.com/c/eRWPVFng/783-bounced-email-candidate-strategic

## Things to check

### DIFF
--- a/app/services/process_notify_callback.rb
+++ b/app/services/process_notify_callback.rb
@@ -1,23 +1,33 @@
 class ProcessNotifyCallback
+  EXPECTED_EMAIL_TYPES = %w[reference_request sign_up_email].freeze
+
   def initialize(notify_reference:, status:)
-    @environment, @email_type, @reference_id = notify_reference.split('-')
+    @environment, @email_type, @id = notify_reference.split('-')
     @status = status
     @not_found = false
   end
 
   def call
-    return unless same_environment? && reference_request_email? && permanent_failure_status?
+    return unless same_environment? && expected_email_type? && permanent_failure_status?
 
-    ActiveRecord::Base.transaction do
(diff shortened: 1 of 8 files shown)


app/services/process_notify_callback.rb
```diff
@@ -1,23 +1,31 @@
...
+        reference.update!(feedback_status: 'email_bounced')
 
-      reference.update!(feedback_status: 'email_bounced')
+        SendNewRefereeRequestEmail.call(
+          application_form: reference.application_form,
+          reference: reference,
+          reason: :email_bounced,
+        )
+      end
+    else
```
tvararu: Maybe instead `elsif sign_up_email?` and then another else that raises "Not implemented" or something to that effect?

chubberlisk: I don't think this is needed because of the guard statement particularly the `expected_email_type?` as this checks that the email type provided in the Notify reference matches one of the ones we expect and currently they are only two so it's going to be either a reference request email or a sign up email.

chubberlisk: Updated to be more explicit.

PR #1186, opened 2020-01-23
Update email and personal details of provider and support users when they log in

mnacos:
### Context

Provider and support users in our database now have name fields in the database that should be updated every time they log in to Apply. This will make DfE Sign-In the primary source of truth for these fields, namely `email_address`, `first_name` and `last_name`. With the exception of the first time they log in (when we match on email address rather than uid, because that's not known in advance), the rest of the time we match on `uid`, making these updates safe.

Add `profile` to the omniauth scope (config/initializers/omniauth.rb), so that the payload we receive in the authentication callback contains `first_name` and `last_name` information (otherwise, the fields are there but the values are `nil`). 

Add `first_name` and `last_name` fields to the development omniauth bypass, so that developers can easily test/develop with arbitrary DfE Sign-In profiles.

Add `first_name` and `last_name` attributes to DfESignInUser and populate these from the session object, so that this information is available for both provider and support users.

Trigger a new method `#update_profile_from_dfe_sign_in` within `DfESignInController#callback`, to update user details in our database at the time of sign in.

The update logic is common to both types of user,  and the method is defined in a separate class (`DsiProfile`).

The update profile logic is:

- update `first_name` and `last_name` if the values provided from DfE Sign-In are not blank
- update `email_address` if we have a `dfe_sign_in_uid` for the user
- do not update `email_address` if we don't

The conditional for `email_address` is necessary because when provider users are first invited, we don't have their uid yet and the first time they log in we match on email address. Matching on uid takes precedence, so after their first log in it is safe to update their email address, if they have changed it on DfE Sign-In. It is, however, possible, they have changed the email address to another that already exists in our database, which will fail validation and throw an error on login.

For this reason, we allow the update to fail without disrupting sign in. Should there be a Slack or equivalent notification?

### Guidance to review

See `spec/lib/sync_with_dsi_profile_spec.rb` for the update logic.

You can also test updates in development mode, by creating provider or support users in the support console, then logging in as those users while providing different names or email addresses. User details are updated on signing in.

image

### Link to Trello card

[1543 - Update email_address and name fields for provider users every time they log in](https://trello.com/c/8lQvOrem)

### Things to check

### DIFF
--- a/app/controllers/dfe_sign_in_controller.rb
+++ b/app/controllers/dfe_sign_in_controller.rb
@@ -4,6 +4,9 @@ class DfESignInController < ActionController::Base
   def callback
     DfESignInUser.begin_session!(session, request.env['omniauth.auth'])
 
+    user = get_support_user || ProviderUser.load_from_session(session)
+    DsiProfile.update_profile_from_dfe_sign_in(dfe_user: DfESignInUser.load_from_session(session), local_user: user) if user
+
     redirect_to session.delete('post_dfe_sign_in_path') || default_authenticated_path
   end
 
@@ -12,14 +15,14 @@ class DfESignInController < ActionController::Base
 private
 
   def default_authenticated_path
-    if authorized_for_support_interface?
+    if get_support_user
       support_interface_path
     else
       provider_interface_path
     end
   end
 
-  def authorized_for_support_interface?
-    SupportUser.load_from_session(session)
+  def get_support_user
+    @user != nil ? @user : @user = (SupportUser.load_from_session(session) || false)
   end
 end
--- a/app/controllers/provider_interface/provider_interface_controller.rb
+++ b/app/controllers/provider_interface/provider_interface_controller.rb
@@ -1,6 +1,7 @@
 module ProviderInterface
   class ProviderInterfaceController < ActionController::Base
     include LogQueryParams
+
     before_action :authenticate_provider_user!
     around_action :set_audit_username
     before_action :add_identity_to_log
--- a/app/lib/dsi_profile.rb
+++ b/app/lib/dsi_profile.rb
@@ -1,0 +1,11 @@
+class DsiProfile
+  def self.update_profile_from_dfe_sign_in(dfe_user:, local_user:)
+    fields_to_update = {}
+    if local_user.dfe_sign_in_uid && dfe_user.email_address.present?
+      fields_to_update[:email_address] = dfe_user.email_address
+    end
+    fields_to_update[:first_name] = dfe_user.first_name if dfe_user.first_name.present?
+    fields_to_update[:last_name] = dfe_user.last_name if dfe_user.last_name.present?
+    local_user.update(fields_to_update) if fields_to_update.present?
+  end
+end
--- a/app/models/dfe_sign_in_user.rb
+++ b/app/models/dfe_sign_in_user.rb
@@ -1,15 +1,20 @@
 class DfESignInUser
   attr_reader :email_address, :dfe_sign_in_uid
+  attr_accessor :first_name, :last_name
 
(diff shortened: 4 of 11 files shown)


app/controllers/dfe_sign_in_controller.rb
```diff
@@ -1,9 +1,13 @@
 class DfESignInController < ActionController::Base
   protect_from_forgery except: :bypass_callback
+  include SyncWithDsiProfile
 
   def callback
     DfESignInUser.begin_session!(session, request.env['omniauth.auth'])
 
+    user = get_support_user || ProviderUser.load_from_session(session)
+    update_profile_from_dfe_sign_in(dfe_user: DfESignInUser.load_from_session(session), local_user: user) if user
```
tijmenb: Perhaps use this directly, so we don't have to `include` the module.

```suggestion
    SyncWithDsiProfile.update_profile_from_dfe_sign_in(dfe_user: DfESignInUser.load_from_session(session), local_user: user) if user
```

Including modules / using concerns tends to become tricky if an app grows because it's hard to ensure there aren't clashes in method names (private methods are also included), and it's not easy to find out where a method is defined.

spec/lib/sync_with_dsi_profile_spec.rb
```diff
@@ -0,0 +1,62 @@
+require 'rails_helper'
+
+RSpec.describe SyncWithDsiProfile do
+  let(:controller) { Class.new.extend(SyncWithDsiProfile) }
```
tijmenb: This becomes a little cleaner if we call the class directly, as per comment above.

app/controllers/dfe_sign_in_controller.rb
```diff
@@ -12,14 +16,19 @@ def callback
...
 
-  def authorized_for_support_interface?
-    SupportUser.load_from_session(session)
+  def get_support_user
+    if @already_tried # using separate var as guard as lookup can return nil, which is then cached
+      @user
+    else
+      @already_tried = true
+      @user = SupportUser.load_from_session(session)
+    end
```
tijmenb: Suggestion:

```rb
@user == nil ? (SupportUser.load_from_session(session) || false) : @user
```

app/controllers/dfe_sign_in_controller.rb
```diff
@@ -12,14 +16,19 @@ def callback
...
       support_interface_path
     else
       provider_interface_path
     end
   end
 
-  def authorized_for_support_interface?
-    SupportUser.load_from_session(session)
+  def get_support_user
+    if @already_tried # using separate var as guard as lookup can return nil, which is then cached
```
stevehook: I think you can use `defined?(@user)` if you want to avoid the extra variable here.

app/controllers/dfe_sign_in_controller.rb
mnacos: Nice, thanks, was looking for a more concise version.

mnacos: Went with:

```ruby
@user != nil ? @user : @user = (SupportUser.load_from_session(session) || false)
```

mnacos:
> Looks great. Tested locally and it works nicely. The only question I have is what happens if a DfE account is a support user and a provider user?

In terms of access, I think what happens is exactly what would happen before. In terms of profile updates, support user wins currently; only this user's details will be updated.

app/controllers/dfe_sign_in_controller.rb
mnacos: Thanks, went for modified version of Tijmen's one liner.

app/controllers/dfe_sign_in_controller.rb
mnacos: Ok

PR #1190, opened 2020-01-24
Add auditing to provider and support users

stevehook:
DO NOT MERGE UNTIL https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1194 is DEPLOYED

## Context

Support and provider users can be created and/or updated by many different people. We want to retain a history of who did what when. e.g. who added Bob as a support user and when did they do it?

- Add auditing to the `ProviderUser` and `SupportUser` models via the `audited` gem.

## Guidance to review

- I'm deliberately skipping `last_signed_in_at` column otherwise you'll see an entry every time the user logs in - this is up for debate if you think it might be useful?
- This PR doesn't include a UI to view the audit trail for users. We expect this to be included in the support interface in future. Any thoughts on what this should look like?

## Link to Trello card

https://trello.com/c/KsueSMF8/1521-add-auditing-to-provider-and-support-users

## Things to check

### DIFF
--- a/app/models/provider_user.rb
+++ b/app/models/provider_user.rb
@@ -5,6 +5,8 @@ class ProviderUser < ActiveRecord::Base
 
   before_save :downcase_email_address
 
+  audited except: [:last_signed_in_at]
+
   def self.load_from_session(session)
     dfe_sign_in_user = DfESignInUser.load_from_session(session)
     return unless dfe_sign_in_user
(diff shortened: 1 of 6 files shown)


app/models/provider_user.rb
```diff
@@ -5,6 +5,8 @@ class ProviderUser < ActiveRecord::Base
 
   before_save :downcase_email_address
 
+  audited except: [:last_signed_in_at]
```
tijmenb: Does this automatically track the `has_and_belongs_to_many :providers` of the provider user?

stevehook: No, good point, I need to add that

stevehook: Not as simple as I thought - needs a separate migration and change to the many-many association - https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1194

PR #1194, opened 2020-01-24
Add primary key and timestamps to `provider_users_providers` (migration)

stevehook:
## Context

This is preparation for the `provider_users_providers` table becoming a fully fledged model rather than just a HABTM join table. We need this to enable auditing of provider user - provider membership.

Required by https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1190

- This is just a migration PR
- I've added `id`, `updated_at`, `created_at` columns
- Each column has a default value - I _think_ this means that it'll be compatible with the existing code (I've not changed the association from a `has_and_belongs_to_many` to `has_many_through`).

## Link to Trello card

https://trello.com/c/KsueSMF8/1521-add-auditing-to-provider-and-support-users

## Things to check


db/migrate/20200124114452_add_primary_key_to_provider_users_providers.rb
```diff
@@ -0,0 +1,9 @@
+class AddPrimaryKeyToProviderUsersProviders < ActiveRecord::Migration[6.0]
+  def change
+    change_table :provider_users_providers, bulk: false do |t|
+      t.primary_key :id
+      t.datetime :created_at, null: false, default: -> { 'current_timestamp' }
+      t.datetime :updated_at, null: false, default: -> { 'current_timestamp' }
```
tijmenb: Would `add_timestamps` work here, or do we need the default to backfill?

https://apidock.com/rails/v2.1.0/ActiveRecord/ConnectionAdapters/SchemaStatements/add_timestamps

db/schema.rb
```diff
@@ -148,7 +148,7 @@
     t.string "magic_link_token"
     t.datetime "magic_link_token_sent_at"
     t.boolean "hide_in_reporting", default: false, null: false
-    t.bigint "course_from_find_id"
+    t.integer "course_from_find_id"
```
tijmenb: Looks unintentional this? I think this has happened before to someone.

stevehook: That might be running different versions of Postgres - need to check that...

db/migrate/20200124114452_add_primary_key_to_provider_users_providers.rb
stevehook: That doesn't seem to work - like you say it doesn't backfill.

db/schema.rb
stevehook: I get the `integer` in `schema.rb` when I run this migration https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/db/migrate/20200120120940_add_course_from_find_id_to_candidate_model.rb

I don't know why it's different for me, maybe because I'm still running Postgres 10 though it makes no sense - `bigint` is supported in older versions. (I think we are still on 9 in production).

Anyway I've repaved my database and then rerun my migration to get around it for now.

PR #1200, opened 2020-01-27
Refactor DSI flow for provider and support interfaces

mnacos:
### Context

Currently, we create user sessions for all DfE Sign-In users who have authenticated via DSI, even if they do not have permission to use the Provider or Support UIs. This leads to controller logic checking both the existence of a `current_provider_user`/`current_support_user` and `dfe_sign_in_user`, which is messy. It would be better to create user session only if corresponding users are found in the database and eliminate most `dfe_sign_in_user` checks.

We still need to provide feedback to users with DfE Sing-In accounts who try to access our service. Our assumption, however, is that we only need to do this at the point of sign in, not on every single page of the provider interface. If, for some reason, they attempt to access a specific provider interface page, e.g. they may have followed a deep link they received via email, they will be simply re-directed to the login page, and logging in will show them the 'you are not authorized' page.

We can also remove the need for a default authenticated path based on user type, using information from the current path instead. This will improve the UX for support users trying to access provider interface pages. [Do we allow support users to impersonate provider users? This will be useful for checking per-user visibility of applications once we implement rules around accredited bodies].

This PR simplifies interactions between `DfESignInUser`, `DfESignInController`, `ProviderInterfaceController` and `SupportInterfaceController`. `DfESignInController` is now responsible for all sign-in logic, including timestamp updates. The other controllers just have to check for the presence of an appropriate user.

This PR removes all requirements to check for presence of a `dfe_sign_in_user` if a current provider or support user is not available.

'You are not authorized pages' now only show during log in attempts. Any other pages simply redirect to the relevant sign in paths.

A side-effect of this PR is switching between support and provider interfaces makes more sense now. Accessing the provider interface after having signed-in as a support user requires you to log in again (as a provider user), instead of showing the 'you are not authorized' page.

Finally, system specs are now more consistent with the new `/provider` micro-site, which allows access to the root page without a session.

### Guidance to review

- Run the tests.
- Try logging in as a known provider user or support user to the provider and support interfaces.
- Try logging in as an unknown/unauthorised provider or support user to the provider and support interfaces.

### Link to Trello card

[1559 - Do not create sessions for DfE Sign in users who don't yet have permission to use the Provider UI](https://trello.com/c/LSsfPPsN)

## Things to check


app/controllers/provider_interface/provider_interface_controller.rb
```diff
@@ -40,16 +40,6 @@ def dfe_sign_in_user
     def authenticate_provider_user!
       return if current_provider_user
```
stevehook: Should `current_provider_user` and `current_support_user` be memoized?

duncanjbrown:
Discussed offline with @stevehook and @mnacos to fix an issue around `post_dfe_sign_in_path` persisting when it shouldn't. There are still a couple of references to `current_dfe_sign_in_user` in `lib/navigation_items.rb` that could probably go now that we don't handle this case

mnacos:
> Discussed offline with @stevehook and @mnacos to fix an issue around `post_dfe_sign_in_path` persisting when it shouldn't. There are still a couple of references to `current_dfe_sign_in_user` in `lib/navigation_items.rb` that could probably go now that we don't handle this case

Good call, I've removed all mentions of the dfe_sign_in_user from navigation items code.

app/controllers/provider_interface/provider_interface_controller.rb
mnacos: Might as well, the controllers inheriting from these controllers could choose to do crazy things with these methods!

app/controllers/dfe_sign_in_controller.rb
```diff
@@ -3,26 +3,53 @@ class DfESignInController < ActionController::Base
...
+      )
+    end
   end
 
   alias :bypass_callback :callback
 
 private
 
+  def get_local_user
+    @target_path && @target_path.match(/^#{support_interface_path}/) ? get_support_user : get_provider_user
```
duncanjbrown: The expression including regex match is repeated. Could it be a named function?

PR #1201, opened 2020-01-27
Document the states & transitions from candidate perspective

tijmenb:
## Context

In https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1191 we documented the states and transitions from the perspective of a provider. 

This adds the states and transitions from the perspective of a candidate. To do this, I've added a new state machine, which is the easiest way of generating a consistent flow diagram. The existing code needed to be refactored to allow us to re-use the code for both flows. 

image

## Guidance to review

Look at it on Heroku.

## Link to Trello card

https://trello.com/c/geyePkVs

## Things to check


app/services/process_state.rb
```diff
@@ -34,6 +85,14 @@ def state
     end
   end
 
+  def self.i18n_namespace
+    'candidate_flow_'
+  end
+
+  def self.state_count(_)
```
tijmenb: Will be implemented with https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1220.

app/views/support_interface/docs/provider_flow.html.erb
```diff
@@ -0,0 +1,32 @@
...
+    </span>
+  </summary>
+  <div class="govuk-details__text">
+    <%= StateDiagram.svg(only_from_state: nil, machine: ApplicationStateChange) %>
+  </div>
+</details>
+
+<style>
+  svg { max-width: 100%; }
+</style>
```
tvararu: I'm OK with this

tijmenb: Pfew

PR #1208, opened 2020-01-27
Refactor apply_from_find_contoller#show logic into a service

davidgisbey:
## Context

There is currently a lot of logic sitting in the apply_from_find_contoller#show action. This has now been refactored into a service object.

Refactor logic from the  apply_from_find_controller into a service.
 
## Guidance to review

You could go through the sign up flow starting on /candidate/apply?providerCode=1N1&courseCode=238T and check that the flow works correctly for a course on apply.

Then go to candidate/apply?providerCode=1JA&courseCode=2ZRD and check a course that is on find, but not apply works.

Then go to candidate/apply?providerCode=123&courseCode=1234 and check it redirects you to the course not found page.

## Link to Trello card

## Things to check


app/services/candidate_interface/validate_course_query_string_params.rb
```diff
@@ -0,0 +1,46 @@
+module CandidateInterface
+  class ValidateCourseQueryStringParams
```
tijmenb: This does more than validating course query params - perhaps rename it to reflect that (`ApplyFromFindPage`?)

app/services/candidate_interface/validate_course_query_string_params.rb
```diff
@@ -0,0 +1,46 @@
...
+    attr_accessor :return_course
+
+    def initialize(provider_code:, course_code:)
+      @provider_code = provider_code
+      @course_code = course_code
+      @can_apply_on_apply = false
+      @course_on_find = false
+      @return_course = nil
```
tijmenb: Perhaps it's easier to follow the logic in `execute` if all values are set explicitly, and we don't rely on defaults.

app/services/candidate_interface/validate_course_query_string_params.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    def execute
```
tijmenb: What would you think of giving this a meaningful name like `determine_whether_course_is_on_find_or_apply` or something in that vein?

app/services/candidate_interface/validate_course_query_string_params.rb
```diff
@@ -0,0 +1,46 @@
...
+      begin
```
tijmenb: You can skip the `begin` and `end` here if you move the rescue block to the end of the method definition.

app/services/candidate_interface/validate_course_query_string_params.rb
```diff
@@ -0,0 +1,46 @@
...
+          @course_on_find = true
+        elsif @course.present?
+          @course_on_find = true
+        end
+      rescue ActiveRecord::RecordNotFound
+        @course = FindAPI::Course.fetch(@provider_code, @course_code)
+        @course_on_find = true if @course.present?
+      end
+
+      @return_course = @course
```
tijmenb: Can we expose `@course` instead of creating a new `@return_course` ivar?

app/services/candidate_interface/validate_course_query_string_params.rb
davidgisbey: Yea good shout. There was a misunderstanding of the scope of what this would do. It used to validate query params in multiple controllers 👍

davidgisbey:
Thanks for all the feedback! I think i've actioned it all.

PR #1210, opened 2020-01-27
Migration to add last_signed_in_at to candidate

davidgisbey:
## Context

There has been discussion about the implementation of flash messages which inform users of their account being created. After some discussion, it was noted that adding a last signed in the column in the candidate would allow more robust data collection.

This PR adds a last_changed_at to a candidate

## Guidance to review

You could run the migration locally and check the candidate has a last signed in at which is set to nil by default.

## Link to Trello card

## Things to check

### DIFF
--- a/db/migrate/20200127171445_add_last_signed_in_at_to_candidate.rb
+++ b/db/migrate/20200127171445_add_last_signed_in_at_to_candidate.rb
@@ -1,0 +1,5 @@
+class AddLastSignedInAtToCandidate < ActiveRecord::Migration[6.0]
+  def change
+    add_column :candidates, :last_signed_in_at, :datetime
+  end
+end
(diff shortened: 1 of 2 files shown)


db/migrate/20200127171445_add_last_signed_in_at_to_candidate.rb
```diff
@@ -0,0 +1,9 @@
+class AddLastSignedInAtToCandidate < ActiveRecord::Migration[6.0]
+  def up
```
tijmenb: `add_column` is a reversible statement, so you can use `change` and remove the `down` method:
 
```suggestion
  def change
```

davidgisbey: Thanks 👍

PR #1211, opened 2020-01-27
Add flash message telling them their account has been created

davidgisbey:
## Context

When a user signs in for the first time they should be told that their account has been created.
 
They should also have their last_sign_in set to the current time.

- Add a flash message when you log in telling you your account has been created
- Update your last_signed_in to the current time.

image

## Guidance to review

Navigate the login flow with a course with one site (/candidate/apply?providerCode=C58&courseCode=3D2X). 
Can you see the flash message?

Navigate the login flow with a course with multiple sites (/candidate/apply?providerCode=2LR&courseCode=3255) 
Can you see the flash message?
## Link to Trello card

## Things to check


app/controllers/candidate_interface/sign_in_controller.rb
```diff
@@ -24,12 +24,17 @@ def create
 
     def authenticate
       candidate = FindCandidateByToken.call(raw_token: params[:token])
+      flash[:success] = 'Your apply for teacher training account has been created' if candidate&.last_signed_in_at.nil?
+
       if candidate
         sign_in(candidate, scope: :candidate)
         add_identity_to_log candidate.id
+        candidate.update(last_signed_in_at: Time.zone.now)
```
tijmenb: To avoid silent fails:

```suggestion
        candidate.update!(last_signed_in_at: Time.zone.now)
```

app/controllers/candidate_interface/sign_in_controller.rb
```diff
@@ -24,12 +24,17 @@ def create
 
     def authenticate
       candidate = FindCandidateByToken.call(raw_token: params[:token])
+      flash[:success] = 'Your apply for teacher training account has been created' if candidate&.last_signed_in_at.nil?
```
tijmenb: This can be moved inside the `if candidate` so we don't have to add the `&`.

davidgisbey:
Just had a look into trackable and there's 0 reason I can see we shouldn't be. I'm going to add it now :shipit:

davidgisbey:
Scratch that after further discussion it would require about 5 additional columns on the candidate to function. It's an all or nothing module.

app/controllers/candidate_interface/sign_in_controller.rb
davidgisbey: thanks 👍

spec/system/candidate_interface/candidate_new_user_with_course_params_spec.rb
```diff
@@ -125,11 +126,19 @@ def and_i_see_the_form_to_pick_a_location
...
+    @candidate = Candidate.find_by!(email_address: @email)
+    expect(@candidate.course_from_find_id).to eq(nil)
   end
 
   def then_i_should_see_the_course_choices_site_page
     expect(page).to have_current_path(candidate_interface_course_choices_site_path(@course_with_multiple_sites.provider.code, @course_with_multiple_sites.code))
   end
+
+  def and_i_should_see_an_account_created_flash_message
+    expect(page).to have_content('Your apply for teacher training account has been created')
```
tvararu: You could hoist this + the duplicate in the controller into a translation file, optional 👍

davidgisbey: Have done. Thanks!

PR #1214, opened 2020-01-28
Add encrypted candidate ID to links in submitted application email

chubberlisk:
## Context

We want to improve the flow for when a candidate clicks on a link in an email we've sent them.

For example, currently when we send them an email to confirm they've submitted their application, that email contains links to the application dashboard and sign in page. If a candidate is not signed in and they click the link, then they'll have to enter their email address and click the `Continue` button. To improve this flow, we want to remove the need of a candidate entering their email address and instead all they need to do is click a button to trigger the sign in email.

This PR adds:

- a service `Encryptor` for encrypting and decrypting a string
- a feature flag called ~~`encrypted_candidate_id`~~ `improved_expired_token_flow`
- a candidate's ID but encrypted to links in their submitted application email

## Screenshots (Updated 28 Jan 15:03)

image

## Guidance to review

Would like feedback/help on:

- whether how we are encrypting the candidate ID is sensible and if there are other or better ways in doing so
- if we should test the service `Encryptor` and how
- how to improve the expectation on when the feature flag is off in the mailer spec (see below)

```
expect(mail.body.encoded).to include(candidate_interface_sign_in_url)
expect(mail.body.encoded).not_to include(candidate_interface_sign_in_url(id: 'example_encrypted_id'))
```

## Link to Trello card

https://trello.com/c/ZyModi4f/811-improve-flow-for-expired-tokens-%F0%9F%8F%88

## Things to check

### DIFF
--- a/app/services/encryptor.rb
+++ b/app/services/encryptor.rb
@@ -1,0 +1,11 @@
+class Encryptor
+  ENCRYPTOR = ActiveSupport::MessageEncryptor.new(Rails.application.secrets.secret_key_base[0..31])
+
+  def self.encrypt(text)
+    ENCRYPTOR.encrypt_and_sign(text)
+  end
+
+  def self.decrypt(text)
+    ENCRYPTOR.decrypt_and_verify(text)
+  end
+end
--- a/app/views/candidate_mailer/submit_application_email.text.erb
+++ b/app/views/candidate_mailer/submit_application_email.text.erb
@@ -12,14 +12,14 @@ Your application reference is <%= @application_form.support_reference %>.
 
 You can track the progress of your <%= 'application'.pluralize(@application_form.application_choices.count) %> here:
 
-<%= candidate_interface_application_form_url %>
+<%= candidate_sign_in_url(@candidate) %>
 
 # Making changes to your application
 
 <% if FeatureFlag.active?('edit_application') %>
   You have 5 working days to edit your application or change your choice of provider, course or training location.
 
-  To make changes, [sign back in to your account](<%= candidate_interface_sign_in_url %>).
+  To make changes, [sign back in to your account](<%= candidate_sign_in_url(@candidate) %>).
 
   You can sign back in to change your name or contact details at any point up until enrolment.
 
@@ -40,7 +40,7 @@ You can track the progress of your <%= 'application'.pluralize(@application_form
 
 You can withdraw your application to your <%= 'course'.pluralize(@application_form.application_choices.count) %> at any point, even after you’ve accepted an offer.
 
-[Sign in to your account](<%= candidate_interface_sign_in_url %>) and click ‘withdraw’ next to the <%= 'course'.pluralize(@application_form.application_choices.count) %> you want to withdraw and we’ll let your training <%= 'provider'.pluralize(@application_form.application_choices.count) %> know.
+[Sign in to your account](<%= candidate_sign_in_url(@candidate) %>) and click ‘withdraw’ next to the <%= 'course'.pluralize(@application_form.application_choices.count) %> you want to withdraw and we’ll let your training <%= 'provider'.pluralize(@application_form.application_choices.count) %> know.
 
 # References
 
--- a/app/helpers/view_helper.rb
+++ b/app/helpers/view_helper.rb
@@ -61,6 +61,16 @@ module ViewHelper
     dates.edit_by.to_s(:govuk_date).strip
   end
 
+  def candidate_sign_in_url(candidate)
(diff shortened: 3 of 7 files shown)


app/services/encryptor.rb
```diff
@@ -0,0 +1,15 @@
...
+    encrypted_data = ENCRYPTOR.encrypt_and_sign(text)
+
+    CGI.escape(encrypted_data)
+  end
+
+  def self.decrypt(text)
+    encrypted_data = CGI.unescape(text)
+
+    ENCRYPTOR.decrypt_and_verify(encrypted_data)
+  end
```
tijmenb: From IRL review: add test to see if `encrypt(x) == decrypt(encrypt(x))`

app/services/encryptor.rb
```diff
@@ -0,0 +1,15 @@
+class Encryptor
+  ENCRYPTOR = ActiveSupport::MessageEncryptor.new(Rails.application.secrets.secret_key_base[0..31])
+
+  def self.encrypt(text)
+    encrypted_data = ENCRYPTOR.encrypt_and_sign(text)
+
+    CGI.escape(encrypted_data)
```
tijmenb: I *think* Rails escapes anything passed into a path helper.

app/services/encryptor.rb
```diff
@@ -0,0 +1,15 @@
+class Encryptor
+  ENCRYPTOR = ActiveSupport::MessageEncryptor.new(Rails.application.secrets.secret_key_base[0..31])
```
tijmenb: We should look into this more - Rails examples seem to do a lot more:

https://api.rubyonrails.org/classes/ActiveSupport/MessageEncryptor.html

I don't understand this yet!

app/views/candidate_mailer/submit_application_email.text.erb
```diff
@@ -12,14 +12,24 @@ Your application reference is <%= @application_form.support_reference %>.
 
 You can track the progress of your <%= 'application'.pluralize(@application_form.application_choices.count) %> here:
 
-<%= candidate_interface_application_form_url %>
+<% if FeatureFlag.active?('encrypted_candidate_id') %>
+  <%= candidate_interface_application_form_url(id: @encrypted_candidate_id) %>
```
tijmenb: Use `candidate_interface_sign_in_url(id: @encrypted_candidate_id)` here instead so we link to a single URL from emails. The effect is the same - if the user is already signed in they'll be redirected to `/candidate/application` anyway, and if they're not signed in they see the "expired" page.

app/views/candidate_mailer/submit_application_email.text.erb
```diff
@@ -12,14 +12,24 @@ Your application reference is <%= @application_form.support_reference %>.
...
+<% else %>
+  <%= candidate_interface_application_form_url %>
+<% end %>
 
 # Making changes to your application
 
+<% if FeatureFlag.active?('encrypted_candidate_id') %>
+  <% sign_in_url = candidate_interface_sign_in_url(id: @encrypted_candidate_id) %>
```
tijmenb: We should extract this into a helper - something like `candidate_sign_in_url(candidate)`.

tijmenb:
```suggestion
  <% sign_in_url = candidate_interface_sign_in_url(u: @encrypted_candidate_id) %>
```

app/services/encryptor.rb
chubberlisk: Done!

app/services/encryptor.rb
chubberlisk: It does!

```ruby
[1] pry(main)> Rails.application.routes.url_helpers.candidate_interface_sign_in_url(u: 'example text')
=> "http://localhost:3000/candidate/sign-in?u=example+text"
[2] pry(main)> Rails.application.routes.url_helpers.candidate_interface_sign_in_url(u: 'example text+--')
=> "http://localhost:3000/candidate/sign-in?u=example+text%2B--"
```

chubberlisk:
@tijmenb I've updated the PR with your feedback, thanks! If you could take another look please. 👀

app/services/encryptor.rb
tvararu: Looks like the only extra feature is specifying a purpose. I'm not sure if it's necessary? 🤔

tvararu: The keys that are generated by the encryptor are quite long though, but not sure if that's something we can customise due to the way that encryption works.

PR #1215, opened 2020-01-28
Migration to add sign in count to user

davidgisbey:
## Context

As we're adding in the trackable from Devise, sign in count seems like an easy free win in terms of data collection

Add a sign_in_count column to the candidate table.

## Guidance to review

Run the migration. Is there a new attribute on candidate?

## Link to Trello card

## Things to check


duncanjbrown:
Good idea! Would adding audit logging for the `last_signed_in_at` column get us this plus a bit more?

PR #1219, opened 2020-01-28
Do not open courses not visible in Find when opening all courses for a provider

duncanjbrown:
## Context

Courses `open_on_apply` should be a subset of courses `exposed_in_find`, at least when we first mark them open. (It’s possible for them to be hidden in Find on a subsequent sync).

Some courses not `exposed_in_find` have eg zero course options and are therefore impossible to apply to anyway. We should not mark these courses `open_on_apply`.

Only mark courses as open on apply if they're also exposed in Find.

## Things to check

### DIFF
--- a/app/controllers/support_interface/providers_controller.rb
+++ b/app/controllers/support_interface/providers_controller.rb
@@ -15,7 +15,7 @@ module SupportInterface
     end
 
     def open_all_courses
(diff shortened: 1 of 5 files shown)


duncanjbrown:
> 👍 I think we should also add `exposed_to_find` to the `open_on_apply` scope, so we have double security against draft courses being surfaced on our end.

Sure, will fix up. We can revisit if we ever start using `open_on_apply` for any other reason

PR #1220, opened 2020-01-28
Add application form reporting to the service performance page

benilovj:
## Context

Before this change, the service performance reporting only displays candidates, not applications, and doesn't show what happens post-submission.

image

* Clearly separate candidate and application-centric reporting
* Port the [`ProcessState` class](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/app/services/process_state.rb) to SQL so it can be used for calculating aggregate statistics

## Guidance to review

* I need some help with the visual design of the page – clearly there are issues with spacing, and I'm not sure that the table is the most effective way of conveying this information
* How could the application form states unit test be made more performant, while still maintaining coverage? It's very slow at the moment but there are lots of nasty `CASE...WHEN` statements in the SQL query to cover

After this PR, I would be keen to break the application form status table further into weekly cohorts, so that we would be able to see how the application statuses develop over time.

## Things to check

### DIFF
--- a/app/models/performance_statistics.rb
+++ b/app/models/performance_statistics.rb
@@ -23,8 +23,59 @@ class PerformanceStatistics
   FROM
       raw_data".freeze
 
+  APPLICATION_FORM_STATUS_COUNTS_QUERY = "
+  WITH raw_data AS (
+      SELECT
+          f.id,
+          COUNT(f.id) FILTER (WHERE f.id IS NOT NULL) application_forms,
+          COUNT(ch.id) FILTER (WHERE f.id IS NOT NULL) application_choices,
+          CASE
+            WHEN ARRAY_AGG(DISTINCT ch.status) = '{NULL}' THEN ARRAY['0', 'unsubmitted']
+            WHEN ARRAY_AGG(DISTINCT ch.status) = '{unsubmitted}' THEN ARRAY['0', 'unsubmitted']
+            WHEN 'awaiting_references' = ANY(ARRAY_AGG(ch.status)) THEN ARRAY['1', 'awaiting_references']
+            WHEN 'application_complete' = ANY(ARRAY_AGG(ch.status)) THEN ARRAY['2', 'waiting_to_be_sent']
+            WHEN 'awaiting_provider_decision' = ANY(ARRAY_AGG(ch.status)) THEN ARRAY['3', 'awaiting_provider_decisions']
+            WHEN 'offer' = ANY(ARRAY_AGG(ch.status)) THEN ARRAY['5', 'awaiting_candidate_response']
+            WHEN 'enrolled' = ANY(ARRAY_AGG(ch.status)) THEN ARRAY['8', 'enrolled']
+            WHEN 'recruited' = ANY(ARRAY_AGG(ch.status)) THEN ARRAY['7', 'recruited']
+            WHEN 'pending_conditions' = ANY(ARRAY_AGG(ch.status)) THEN ARRAY['6', 'pending_conditions']
+            WHEN ARRAY_REMOVE(ARRAY_REMOVE(ARRAY_REMOVE(ARRAY_REMOVE(ARRAY_AGG(DISTINCT ch.status), 'withdrawn'), 'rejected'), 'declined'), 'conditions_not_met') = '{}' THEN ARRAY['4', 'ended_without_success']
+            ELSE ARRAY['9', 'unknown_state']
+          END status
+      FROM
+          candidates c
+      LEFT JOIN
+          application_forms f ON f.candidate_id = c.id
+      LEFT JOIN
+          application_choices ch ON ch.application_form_id = f.id
+      WHERE
+          NOT c.hide_in_reporting
+      GROUP BY
+          f.id
+  )
+  SELECT
+      raw_data.status[2],
+      COUNT(*)
+  FROM
+      raw_data
+  GROUP BY
+      raw_data.status
+  ORDER BY
+      raw_data.status[1]".freeze
+
   def [](key)
     candidate_counts[key.to_s]
+  end
+
+  def application_form_status_counts
+    @application_form_status_counts ||= ActiveRecord::Base
+      .connection
+      .execute(APPLICATION_FORM_STATUS_COUNTS_QUERY)
+      .to_a
+  end
+
+  def application_form_counts_total
+    application_form_status_counts.sum { |row| row['count'].to_i }
   end
 
 private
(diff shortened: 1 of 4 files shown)


spec/models/performance_statistics_spec.rb
```diff
@@ -31,8 +34,39 @@
 
     stats = PerformanceStatistics.new
 
-    expect(stats[:total_non_dfe_sign_ups]).to eq(1)
+    expect(stats[:total_candidate_count]).to eq(1)
     expect(stats[:candidates_with_submitted_forms]).to eq(1)
     expect(stats[:candidates_signed_in_but_not_entered_data]).to eq(0)
   end
+
+  it 'classifies application forms in the same way as ProcessState' do
```
tijmenb: > How could the application form states unit test be made more performant, while still maintaining coverage? It's very slow at the moment but there are lots of nasty CASE...WHEN statements in the SQL query to cover

Here's an alternative way of testing this: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/commit/e47351495ac635cd8935969d908dbe19aa1449b2

It would look a lot like the process state spec: 

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/spec/services/process_state_spec.rb

It's much faster to use factories directly (`TestApplications.create_application` is optmised for realism, not speed). It also allows to individually test the states & counts - I think in the current test there's no test to make sure that we haven't accidentally swapped two state counts.

app/views/support_interface/performance/index.html.erb
```diff
@@ -31,3 +33,34 @@
     </div>
   </div>
 </div>
+
+<h2 class="govuk-heading-m govuk-!-font-size-27">Application forms</h2>
```
fofr:
```suggestion
<h2 class="govuk-heading-m govuk-!-font-size-27 govuk-!-margin-top-4">Application forms</h2>
```

fofr:
Nice!

app/views/support_interface/performance/index.html.erb
```diff
@@ -1,8 +1,10 @@
 <% content_for :title, 'Service performance dashboard' %>
```
tijmenb: Here's an alternative design: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/commit/7c462bddd3e3ef9497a75107fc4fb432b0c09ec0

image

Until we have Apply 2, candidates and application forms are essentially the same. This version removes the duplication of "unsubmitted" applications, and uses the blocks. Once https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1201 is merged we can link to that for the explanation of the state.

app/models/performance_statistics.rb
```diff
@@ -15,21 +15,72 @@ class PerformanceStatistics
...
+  WITH raw_data AS (
+      SELECT
+          f.id,
+          COUNT(f.id) FILTER (WHERE f.id IS NOT NULL) application_forms,
+          COUNT(ch.id) FILTER (WHERE f.id IS NOT NULL) application_choices,
+          CASE
+            WHEN ARRAY_AGG(DISTINCT ch.status) = '{NULL}' THEN ARRAY['0', 'unsubmitted']
+            WHEN ARRAY_AGG(DISTINCT ch.status) = '{unsubmitted}' THEN ARRAY['0', 'unsubmitted']
+            WHEN 'awaiting_references' = ANY(ARRAY_AGG(ch.status)) THEN ARRAY['1', 'awaiting_references']
+            WHEN 'application_complete' = ANY(ARRAY_AGG(ch.status)) THEN ARRAY['2', 'waiting_to_be_sent']
```
duncanjbrown: a couple of humble questions (my sql is very much less good than yours):

- why do we need to check for null statuses? I think `null` would break the state machine?
- if I understand `ARRAY_AGG(ch.status)` right, this creates an array (denoted `{item item}`) of the status of the attached choices. In the case of `unsubmitted`, `awaiting_references` and `application_complete` the choices should move forward in lockstep (see also `TestApplications.create_application`), so is the `ANY` necessary in these cases?

app/views/support_interface/performance/index.html.erb
fofr: Cleaner, but I think harder to interpret at a glance. There's no sense of progression with this one.

fofr: Can we stick with the original design but apply the deduplication?

tijmenb: Isn't it the same sense of progression (left to right) as the existing blocks?

Perhaps "Ended without success" and "Enrolled" should be pulled out because they're end states.

app/models/performance_statistics.rb
benilovj: > why do we need to check for null statuses? I think null would break the state machine?

I believe this is to handle the case where the application form hasn't been created at all. If I remove it, the match against `ProcessState` doesn't work for all applications.

> if I understand ARRAY_AGG(ch.status) right, this creates an array (denoted {item item}) of the status of the attached choices. In the case of unsubmitted, awaiting_references and application_complete the choices should move forward in lockstep (see also TestApplications.create_application), so is the ANY necessary in these cases?

I did a direct port of the Ruby logic in `ProcessState` and would be a bit uncomfortable to have them diverge. I think once we prove that this approach is useful, we probably want to do something like materialising the application form state onto the record itself and updating it each time the application choices change.

duncanjbrown: No further questions. Thank you!

app/views/support_interface/performance/index.html.erb
benilovj: In follow-up work, I'd be keen to further break down the data by cohorts, based on when users signed up. Here's a mock-up:

image

This breakdown would allow us to see:

- how our product improvements are playing out over time – hopefully we should see better conversions as we improve stuff
- see a more accurate conversion rate i.e. more recent applications, where the candidate hasn't had time to submit it yet, won't distort the completion rate for older applications

spec/models/performance_statistics_spec.rb
benilovj: Thanks – I've incorporated your approach in 98c97d8

benilovj:
Thanks all for the feedback:

- I've updated the testing approach as per @tijmenb's suggestion so it should be a bit more performant
- I think the idea to combine the pre- and post-submission states into one report is a good one, but I think I'd like to do that as a follow-up PR rather than this one, since I'm keen to see the breakdown of application states on the production system ASAP. Would that be OK?

PR #1224, opened 2020-01-29
Allow provider users to act as candidates in sandbox environments

mnacos:
### Context

From a vendor:

> It would be really useful to be able to act as a candidate whose data has been generated via /generate_test_data (the alternative is to go through the form multiple times)

It is already possible to masquerade as a candidate, but only via support, and granting all sandbox providers/vendors access to support is not a good idea.

Allow provider users to masquerade as candidates within sandbox environments.

This is what the support page for the candidate already looks like (note the sign-in button):
image

This PR introduces a similar button to the provider interface. As there is no concept of a candidates index within the provider interface, the natural place for such a button is on an application page.

Pending:
- [x] Should masquerading be restricted to candidates who have applied to this particular provider?
- [x] Test coverage

### Guidance to review

Before:
image

After:
image

Clicking on this button:
image

Attempting to impersonate an unrelated candidate:
image

### Link to Trello card

[1527 - How can vendors/providers act as candidates in the Sandbox?](https://trello.com/c/rVEXKbEU)

### Things to check

### DIFF
--- a/app/controllers/provider_interface/candidates_controller.rb
+++ b/app/controllers/provider_interface/candidates_controller.rb
@@ -4,21 +4,34 @@ module ProviderInterface
 
     def impersonate
       candidate = Candidate.find(params[:candidate_id])
-      # bypass_sign_in will not update the database when signing in
-      # https://stackoverflow.com/questions/50405133/devise-bypass-sign-in-without-active-for-authentication-callback#50409127
-      bypass_sign_in(candidate, scope: :candidate)
 
-      flash[:success] = "You are now signed in as candidate #{candidate.email_address}"
+      if verify_provider_association(candidate: candidate, providers: current_provider_user.providers)
+        bypass_sign_in(candidate, scope: :candidate)
 
-      redirect_to candidate_interface_application_form_path
+        flash[:success] = "You are now signed in as candidate #{candidate.email_address}"
+
+        redirect_to candidate_interface_application_form_path
+      else
+        flash[:warning] = 'This candidate is not associated with your account'
+
+        redirect_to provider_interface_applications_path
+      end
     end
 
   private
 
+    def verify_provider_association(candidate:, providers:)
+      provider_ids_from_candidate = candidate.application_forms. \
+                                              map(&:application_choices).flatten. \
+                                              map(&:provider).map(&:id).uniq
+
+      providers.any? { |provider| provider_ids_from_candidate.include? provider.id }
+    end
+
     def disable_on_production
       return unless HostingEnvironment.production?
 
-      render_404
+      head :not_found
     end
   end
 end
--- a/app/controllers/provider_interface/candidates_controller.rb
+++ b/app/controllers/provider_interface/candidates_controller.rb
@@ -1,0 +1,24 @@
+module ProviderInterface
+  class CandidatesController < ProviderInterfaceController
+    before_action :disable_on_production, only: :impersonate
+
+    def impersonate
+      candidate = Candidate.find(params[:candidate_id])
+      # bypass_sign_in will not update the database when signing in
+      # https://stackoverflow.com/questions/50405133/devise-bypass-sign-in-without-active-for-authentication-callback#50409127
+      bypass_sign_in(candidate, scope: :candidate)
+
+      flash[:success] = "You are now signed in as candidate #{candidate.email_address}"
+
+      redirect_to candidate_interface_application_form_path
+    end
+
+  private
+
+    def disable_on_production
+      return unless HostingEnvironment.production?
+
+      render_404
+    end
+  end
+end
--- a/app/views/provider_interface/application_choices/show.html.erb
+++ b/app/views/provider_interface/application_choices/show.html.erb
@@ -29,7 +29,7 @@
 </div>
 
 <% unless HostingEnvironment.production? %>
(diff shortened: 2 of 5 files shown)


stevehook:
One thing we need to test here is whether actions by providers impersonating candidates are correctly attributed in the audit trail. Not sure how it would work out of the box.

app/controllers/provider_interface/candidates_controller.rb
```diff
@@ -0,0 +1,24 @@
+module ProviderInterface
+  class CandidatesController < ProviderInterfaceController
+    before_action :disable_on_production, only: :impersonate
+
+    def impersonate
+      candidate = Candidate.find(params[:candidate_id])
```
tijmenb: Just to note that this allows providers to impersonate any candidate, not just candidates that have applications that they have access to.

mnacos: We've decided to add a simple guard in the controller checking there is an association between provider and candidate.

mnacos:
> One thing we need to test here is whether actions by providers impersonating candidates are correctly attributed in the audit trail. Not sure how it would work out of the box.

So far this functionality is restricted to non-production environments, for testing etc. Considering this, the desired behaviour is probably full impersonation, e.g. if you are acting as a candidate then the audit trail should show that the candidate did it, even though you are also a provider or support user. But it's worth checking this is the case.

app/controllers/provider_interface/candidates_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+
+      if verify_provider_association(candidate: candidate, providers: current_provider_user.providers)
+        bypass_sign_in(candidate, scope: :candidate)
+
+        flash[:success] = "You are now signed in as candidate #{candidate.email_address}"
+
+        redirect_to candidate_interface_application_form_path
+      else
+        flash[:warning] = 'Candidate is not associated with your providers'
```
duncanjbrown: They might only have one provider. How about "This candidate does not have applications you can manage" or equiv?

app/views/provider_interface/application_choices/show.html.erb
```diff
@@ -28,6 +28,10 @@
   </div>
 </div>
 
+<% unless HostingEnvironment.production? %>
+  <%= button_to 'Sign in as this candidate (sandbox mode)', provider_interface_impersonate_candidate_path(@application_choice.application_form.candidate), class: 'govuk-button' %>
```
duncanjbrown: I think we can lose "Sandbox mode" here as it's not the case any more? (It's still available when `SANDBOX` is off)

spec/requests/provider_interface/provider_user_impersonates_candidate_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      expect(response.body).to match(/\/candidate/)
+    end
+
+    it 'redirects back to Provider Interface if candidate is not associated with user\'s providers' do
+      unrelated_application = create(:application_choice, status: 'awaiting_provider_decision')
+
+      post provider_interface_impersonate_candidate_path(unrelated_application.application_form.candidate)
+
+      expect(response).to have_http_status 302
+      expect(response.body).to match(/\/provider/)
```
duncanjbrown: Request spec feels like the right things here :+1:. The assertions on `provider` and `candidate` (at l.28) aren't too decisive. Is there something stronger we could match on to prove this works.

spec/system/provider_interface/provider_acts_as_candidate_on_sandbox_spec.rb
```diff
@@ -0,0 +1,55 @@
+require 'rails_helper'
+
+RSpec.describe 'A Provider can log in as a candidate' do
+  include CourseOptionHelpers
+  include DfESignInHelpers
+
+  scenario 'when viewing an application on non-production environments' do
```
duncanjbrown: Nice spec

app/views/provider_interface/application_choices/show.html.erb
mnacos: I would like to say something to clarify that they shouldn't expect to see this in production. How about 'test mode'?

duncanjbrown: "Test environments only"?

app/controllers/provider_interface/candidates_controller.rb
mnacos: Actually, how about just 'This candidate is not associated with your account'? It doesn't really matter, as you'll probably never see this message. This action is not possible through the UI (you're browsing applications already associated with your account) and to post to an arbitrary ID you'd have to be creative (perhaps use a different tool, play with the authenticity token etc.) So, you deserve a badly-worded error message!

spec/requests/provider_interface/provider_user_impersonates_candidate_spec.rb
mnacos: It's tricky precisely because it's a request spec. I've added an attempt to access the application form in candidate_interface, which gives 200 only if you have an active candidate session.

duncanjbrown: I think that's great :100:

app/controllers/provider_interface/candidates_controller.rb
duncanjbrown: Nice :+1:

PR #1225, opened 2020-01-29
Improve process documentation for vendors

tijmenb:
## Context

We currently have a lifecycle diagram in the API docs.

Screenshot 2020-01-29 at 11 36 33

(https://www.apply-for-teacher-training.education.gov.uk/api-docs)

This adds a simplified version of the diagrams added in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1201 and https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1191.

image

## Guidance to review

- Does it make sense to an API user
- Are the states and descriptions correct?
- Are we missing information relevant to the vendors?

## Link to Trello card

https://trello.com/c/9DjRZ313/1584-improve-process-documentation-for-vendors

## Things to check


jpasamoah:
Nice!  
Just a few copy suggestions: 

Awaiting references: 
Worth mentioning explicitly that references are the same across application choices 

Candidate Enrolled: 
The candidate's HESA equality monitoring information is made available to the provider

We should also possibly mention Apply 1 & 2 as distinct phases.

duncanjbrown:
This is great!

Are the "TBD"s on reject -> offer and on confirm enrolment going to invite awkward questions?
The way we call it the "Withdrawn (end) state" is inconsistent with other end states

Other than that, the only thing this loses from the old graph is that the names of the nodes used to map to the values in the `status:` enum. It would be nice to see them (but surely not a blocker)

tijmenb:
> Other than that, the only thing this loses from the old graph is that the names of the nodes used to map to the values in the status: enum. It would be nice to see them (but surely not a blocker)

Good point, those need to go back.

PR #1226, opened 2020-01-29
Revert "Add encrypted candidate ID to links in submitted application email"

chubberlisk:
Reverts DFE-Digital/apply-for-postgraduate-teacher-training#1214

We're unable to deploy to QA because it seems that secret key base is missing which is used in this code.  Going to revert so we can debug better and get master back up and running.


tijmenb:
Merged because it fixes Heroku deploys as well, which I need for https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1225.

PR #1228, opened 2020-01-29
Automate email chasing referees for a response

davidgisbey:
## Context

We need to automate the chase email sent to candidates and referees once 5 days have passed.

This PR implements sending the email to the referee.

- Adds a service that returns the references to be contacted
- Adds a service that sends the email to the referee and creates a ChaserSent object associated with the reference

## Guidance to review

A follow-up PR sending an email to the candidate will follow this

**Update**

This has now been refactored to use the ChaserSent model as discussed at our dev session. 

## Link to Trello card

## Things to check


app/services/send_chase_email.rb
```diff
@@ -0,0 +1,9 @@
+class SendChaseEmail
+  def perform(reference:)
+    RefereeMailer.reference_request_chaser_email(reference.application_form, reference)
+
+    reference.application_form.application_choices.each do |application_choice|
+      application_choice.update!(status: 'awaiting_references_and_chased')
```
tvararu:
```suggestion
      application_choice.send_chase_email!
```

This is a method that is auto-generated by the `event :send_chase_email` declaration in the state machine, it's better to use it instead of `update!`

app/workers/send_chase_email_to_referees_worker.rb
```diff
@@ -0,0 +1,13 @@
+class SendChaseEmailToRefereesWorker
+  include Sidekiq::Worker
+
+  def perform
+    GetRefereesToChase.new.perform.each do |reference|
+      begin
+        SendChaseEmail.new.perform(refence: reference)
+      rescue StandardError => e
+        Rails.logger.warn "[DBD] ignoring reference #{reference.id}: #{e.message}"
+      end
```
tvararu: Hmm, why is this needed? 🤔 

Why are we prefixing this with `[DBD]`, looks like a copy+paste typo?

Could we rescue on something more specific than `StandardError`?

app/services/application_state_change.rb
```diff
@@ -20,6 +20,11 @@ def initialize(application_choice)
 
     state :awaiting_references do
       event :references_complete, transitions_to: :application_complete
+      event :send_chase_email, transitions_to: :awaiting_references_and_chased
+    end
+
+    state :awaiting_references_and_chased do
```
tvararu: I think the naming of this is passable, what do others think? cc maester @duncanjbrown

duncanjbrown: I think the name is good... but I'm not convinced this is a distinct state on the choice.

This seems like a state of the reference request, not the application choice. And we're opening the door on a world in which `:awaiting_references` is not an honest state because it will miss out some choices which are IRL awaiting references but aren't `:awaiting_references`

app/workers/send_chase_email_to_referees_worker.rb
```diff
@@ -0,0 +1,13 @@
+class SendChaseEmailToRefereesWorker
+  include Sidekiq::Worker
+
+  def perform
+    GetRefereesToChase.new.perform.each do |reference|
+      begin
+        SendChaseEmail.new.perform(refence: reference)
```
davidgisbey: Just realised this needs to be more explicit as there will be multiple chase emails
Perhaps SendRefereeChaseEmail?
Or SendFirstChaseEmailToReferee

app/services/get_referees_to_chase.rb
```diff
@@ -0,0 +1,13 @@
+class GetRefereesToChase
```
davidgisbey: This probably needs to be more explicit too.

app/services/send_chase_email.rb
```diff
@@ -0,0 +1,6 @@
+class SendChaseEmail
+  def perform(reference:)
+    RefereeMailer.reference_request_chaser_email(reference.application_form, reference)
+    ChaserSent.create(chased: reference, chaser_type: :referee_mailer_reference_request_chaser_email)
```
tijmenb:
```suggestion
    ChaserSent.create!(chased: reference, chaser_type: :referee_mailer_reference_request_chaser_email)
```

On second look perhaps `referee_mailer_reference_request_chaser_email` is a little wordy? Perhaps just `reference_request`, since it's clear it's an email and chaser.

app/workers/send_chase_email_to_referees_worker.rb
```diff
@@ -0,0 +1,13 @@
...
+      rescue StandardError => e
```
tijmenb: What's the reason for rescuing errors here?

app/services/get_referees_to_chase.rb
```diff
@@ -0,0 +1,13 @@
...
+  def perform
+    references = ApplicationReference.includes(:application_form).feedback_requested
+
+    references_with_applications_submitted_in_the_last_5_days = references.select do |reference|
+      reference.application_form.submitted_at < Time.zone.now - 5.days
+    end
+
+    references_with_applications_submitted_in_the_last_5_days.select do |reference|
+      reference.chasers_sent.referee_mailer_reference_request_chaser_email.blank?
+    end
```
tijmenb: I think we could reduce this down to a single query. We could start by using a `where` for the `submitted_at` filtering. Let me know if you want a hand.

app/services/get_referees_to_chase.rb
tijmenb: I think this is fine 👍

spec/services/get_referees_to_chase_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      application_form = create(:application_form, submitted_at: 6.days.ago)
+      reference = create(:reference, feedback_status: 'feedback_requested', application_form: application_form)
+      create(:application_choice, application_form: application_form, status: 'awaiting_references')
+
+      service = described_class.new.perform
+
+      expect(service).to eq [reference]
+    end
+
+    it 'only return sapplciation choices which are awaiting references' do
```
tijmenb:
```suggestion
    it 'only returns application choices which are awaiting references' do
```

spec/workers/send_chase_email_to_referees_worker_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'rails_helper'
+
+RSpec.describe SendChaseEmailToRefereesWorker do
```
tijmenb: Can we create a system spec instead to test this?

app/services/get_referees_to_chase.rb
davidgisbey: ok thanks

app/workers/send_chase_email_to_referees_worker.rb
davidgisbey: I'd love to have a better answer than it was on the other workers. I assumed it would help debug if a job failed for whatever reason. I'll remove it

spec/workers/send_chase_email_to_referees_worker_spec.rb
davidgisbey: Sure, but roughly how would that look? The system specs seem to largely be user journey focussed. I'm not 100% sure how to test a job that way.

tijmenb: Roughly:

- Given a candidate submits an application (create an application with references)
- When there's no response from the referees within 5 days (use timecop to call `SendChaseEmailToRefereesWorker` 5 days later) 
- The referee gets an email with a reminder (check email contents)

I think something similar to the [reject by default test](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/spec/system/provider_interface/provider_receives_email_when_a_new_application_has_been_submitted_spec.rb#L7-L13) would be nice. Perhaps you can [use some stuff from the referee test as well](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/spec/system/referee_interface/referee_refuses_to_give_a_reference_spec.rb#L9-L12).

davidgisbey: Have done 👍

app/services/get_referees_to_chase.rb
```diff
@@ -0,0 +1,9 @@
+class GetRefereesToChase
+  def perform
+    ApplicationReference
+    .joins(:application_form)
+    .feedback_requested
+    .where(['application_forms.submitted_at < ?', Time.zone.now - 5.days])
+    .select { |reference| reference.chasers_sent.reference_request.blank? }
```
tijmenb:
```suggestion
    .where.not(id: ChaserSent.reference_request.select(:chased_id))
```

This will generate one query for this call:

```
[3] pry(main)> puts GetRefereesToChase.new.perform.to_sql
SELECT "references".* FROM "references" INNER JOIN "application_forms" ON "application_forms"."id" = "references"."application_form_id" WHERE "references"."feedback_status" = 'feedback_requested' AND (application_forms.submitted_at < '2020-01-30 15:07:56.592003') AND "references"."id" NOT IN (SELECT "chasers_sent"."chased_id" FROM "chasers_sent" WHERE "chasers_sent"."chaser_type" = 'reference_request')
```

app/services/send_chase_email_to_referee_and_candidate.rb
```diff
@@ -1,7 +1,10 @@
 class SendChaseEmailToRefereeAndCandidate
   def self.call(application_form:, reference:)
     RefereeMailer.reference_request_chaser_email(application_form, reference).deliver
+    ChaserSent.create!(chased: reference, chaser_type: :reference_request)
+
     CandidateMailer.reference_chaser_email(application_form, reference).deliver
+    ChaserSent.create!(chased: application_form.candidate, chaser_type: :reference_request)
```
tijmenb: I don't think we need this one, we're not using it right?

spec/services/get_referees_to_chase_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      application_form = create(:application_form, submitted_at: 6.days.ago)
+      reference = create(:reference, feedback_status: 'feedback_requested', application_form: application_form)
+      create(:application_choice, application_form: application_form, status: 'awaiting_references')
+
+      service = described_class.new.perform
+
+      expect(service).to eq [reference]
+    end
+
+    it 'only returns applciation choices which are awaiting references' do
```
tijmenb:
```suggestion
    it 'only returns application choices which are awaiting references' do
```

spec/system/referee_interface/referee_does_not_respond_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+RSpec.feature 'Referee does not respond within 5 days', sidekiq: true do
+  include CandidateHelper
+
+  scenario 'A chase email is sent if a referee does not respond within 5 days' do
+    FeatureFlag.activate('training_with_a_disability')
+
+    given_a_candidate_completed_an_application
+    when_the_candidate_submits_the_application
+    and_i_do_not_respond_within_5_days
+    then_i_am_sent_a_chase_email
```
tijmenb: Perhaps `and_the_referee_does_not_respond_within_5_days` (we haven't explained that we're a referee)

app/services/send_chase_email_to_referee_and_candidate.rb
davidgisbey: Yea you're right 👍

davidgisbey:
Thanks for all the help/feedback!

spec/services/get_referees_to_chase_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+RSpec.describe GetRefereesToChase do
+  describe '#perform' do
+    it 'returns referees that were sent their reference email more than 5 days ago and have not already been chased' do
+      application_form = create(:application_form, submitted_at: 6.days.ago)
+      reference = create(:reference, feedback_status: 'feedback_requested', application_form: application_form)
+      create(:application_choice, application_form: application_form, status: 'awaiting_references')
+
+      service = described_class.call
+
+      expect(service).to eq [reference]
```
C-gyorfi:
```suggestion
      expect(described_class.call).to include reference
```

spec/services/get_referees_to_chase_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'rails_helper'
+
+RSpec.describe GetRefereesToChase do
+  describe '#perform' do
+    it 'returns referees that were sent their reference email more than 5 days ago and have not already been chased' do
+      application_form = create(:application_form, submitted_at: 6.days.ago)
+      reference = create(:reference, feedback_status: 'feedback_requested', application_form: application_form)
+      create(:application_choice, application_form: application_form, status: 'awaiting_references')
+
+      service = described_class.call
```
C-gyorfi:
```suggestion
```
You could just call these in the assertion as suggested below

spec/factories.rb
```diff
@@ -3,7 +3,7 @@
     candidate
     chased_id { candidate.id }
     chased_type { candidate.class }
-    chased_type { :referee_mailer_reference_request_chaser_email }
+    chased_type { :reference_request }
```
C-gyorfi: This is a bit unclear to mee, you defined chased_type twice, was this a mistake?

app/services/get_referees_to_chase.rb
```diff
@@ -0,0 +1,9 @@
+class GetRefereesToChase
+  def self.call
+    ApplicationReference
+    .joins(:application_form)
+    .feedback_requested
+    .where(['application_forms.submitted_at < ?', Time.zone.now - 5.days])
```
tijmenb: This needs to be `reference.created_at` to avoid fun after https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1279.

spec/system/referee_interface/referee_does_not_respond_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    and_an_email_is_sent_to_the_candidate
+  end
+
+  def given_a_candidate_completed_an_application
+    candidate_completes_application_form
+  end
+
+  def when_the_candidate_submits_the_application
+    candidate_submits_application
+    @application.application_references.first.update(feedback_status: :feedback_refused)
```
tijmenb:
```suggestion
    @application.application_references.first.update!(feedback_status: :feedback_refused)
```

spec/system/referee_interface/referee_does_not_respond_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+RSpec.feature 'Referee does not respond within 5 days', sidekiq: true do
+  include CandidateHelper
+
+  scenario 'A chase email is sent if a referee does not respond within 5 days' do
+    FeatureFlag.activate('training_with_a_disability')
+
+    given_a_candidate_completed_an_application
+    when_the_candidate_submits_the_application
+    and_the_referee_does_not_respond_within_5_days
+    then_i_am_sent_a_chase_email
```
tijmenb:
```suggestion
    then_the_referee_am_sent_a_chase_email
```

C-gyorfi:
```suggestion
    then_the_referee_sent_a_chase_email
```

spec/factories.rb
davidgisbey: Yea it was meant to be chaser type. Good spot 👍

PR #1240, opened 2020-01-30
Export application data for data analysis 

tijmenb:
## Context

This should generate a CSV for use by Matteo. He's currently working with a sample export I generated by running this code on prod:

https://docs.google.com/spreadsheets/d/1X53Tq1cCOTa0u1C0VxGxIHX6ogCOgWOtHX6xsbvfth4/edit#gid=1470484932

This adds a export feature that generates a CSV with times that something happened to an application. It can be used to analyse how candidates progress through the system.

image

## Guidance to review

Does it make sense? I haven't added many tests.

## Link to Trello card

https://trello.com/c/dKBwLMPv

## Things to check

### DIFF
--- a/app/controllers/support_interface/performance_controller.rb
+++ b/app/controllers/support_interface/performance_controller.rb
@@ -1,5 +1,21 @@
+require 'csv'
+
 module SupportInterface
   class PerformanceController < SupportInterfaceController
     def index; end
+
+    def application_timings
+      applications = SupportInterface::ApplicationsExport.new.applications
+
+      csv = CSV.generate do |rows|
+        rows << applications.first.keys
+
+        applications.each do |a|
+          rows << a.values
+        end
+      end
+
+      render plain: csv
+    end
   end
 end
--- a/app/services/support_interface/applications_export.rb
+++ b/app/services/support_interface/applications_export.rb
@@ -1,0 +1,69 @@
+module SupportInterface
+  class ApplicationsExport
+    def applications
+      relevant_applications.map do |application_form|
+        output = {
+          support_reference: application_form.support_reference,
+          process_state: ProcessState.new(application_form).state,
+          signed_up_at: application_form.candidate.created_at,
+          first_signed_in_at: application_form.created_at,
+          submitted_form_at: application_form.submitted_at,
+          form_updated_at: application_form.updated_at,
+          courses_last_updated_at: application_form.application_choices.max_by(&:updated_at)&.updated_at,
+          qualifications_last_updated_at: application_form.application_qualifications.max_by(&:updated_at)&.updated_at,
+          work_history_last_updated_at: application_form.application_work_experiences.max_by(&:updated_at)&.updated_at,
+          references_last_updated_at: application_form.application_references.max_by(&:updated_at)&.updated_at,
+        }
+
+        audits = application_form.own_and_associated_audits
+
+        %w[
+          address_line1
+          becoming_a_teacher
+          country
+          course_choices_completed
+          date_of_birth
+          degrees_completed
+          disclose_disability
+          english_main_language
+          first_name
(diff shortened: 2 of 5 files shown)


app/services/support_interface/applications_export.rb
```diff
@@ -0,0 +1,60 @@
...
+
+        output
+      end
+    end
+
+  private
+
+    def last_change_to_form(application_form, column)
+      audit = application_form.own_and_associated_audits.find do |audit|
+        audit.audited_changes.has_key?('start_date')
```
tijmenb:
```suggestion
        audit.audited_changes.has_key?(column)
```

stevehook: yes that makes sense now! Nice way to get a granular last updated at summary.

I guess it would be possible to also get a history of when the application reached each status in the workflow pretty much the same way. What I don't think we can do is get a history of 'actions'.

tijmenb: Yeah, the actions would have to be inferred by looking at before/after `status`.

tijmenb:
Closing until I'll be working on this again.

app/views/support_interface/performance/index.html.erb
```diff
@@ -98,10 +98,21 @@
...
+  <h2 class='govuk-heading-l'>Downloads for analysis</h2>
+
+  <h3 class='govuk-heading-m'>Application timings</h3>
+
+  <p class='govuk-body'>
+    The application timings provides data on when a application
+    form attribute was last updated by the candidate.
+  </p>
+
+  <%= govuk_link_to 'Download application timings (CSV)', support_interface_application_timings_path %>
```
tvararu: The performance page is publicly visible - do we want this to be shown if you're not logged in?

benilovj: I'd rather not show this if you're not logged in

tijmenb: Should we pull out the public dashboard into a separate namespace? I half-forgot that this page was public, it's likely that something like that will happen again.

tvararu: @tijmenb 👍

tijmenb: Done!

app/views/support_interface/performance/index.html.erb
```diff
@@ -5,3 +5,14 @@
...
+<h2 class='govuk-heading-l'>Downloads for analysis</h2>
+
+<h3 class='govuk-heading-m'>Application timings</h3>
+
+<p class='govuk-body'>
+  The application timings provides data on when a application
+  form attribute was last updated by the candidate.
+</p>
+
+<%= govuk_link_to 'Download application timings (CSV)', support_interface_application_timings_path %>
```
duncanjbrown: This looks oddly small on the page — should it be inside `govuk-body`?

Screenshot 2020-02-18 at 15 24 43

app/controllers/support_interface/performance_controller.rb
```diff
@@ -1,5 +1,21 @@
...
+
+      csv = CSV.generate do |rows|
+        rows << applications.first.keys
+
+        applications.each do |a|
+          rows << a.values
+        end
+      end
+
+      render plain: csv
```
duncanjbrown: If we set `Content-Type` clients ought to download it rather than rendering it?

tijmenb: I find it easier to copy-paste into Google Sheets than to import it, so I'd like to keep this for now until Matteo says otherwise.

app/views/support_interface/performance/index.html.erb
tijmenb: Fixed!

PR #1247, opened 2020-01-31
Add chaser emails table

davidgisbey:
## Context

As discussed yesterday. We need a table for chaser_emails (notifications email on the whiteboard).

- A migration to add the chaser_emails table

## Guidance to review
 
I think that's all the columns we need.

## Link to Trello card

## Things to check

### DIFF
--- a/db/migrate/20200131130342_create_chasers_sent.rb
+++ b/db/migrate/20200131130342_create_chasers_sent.rb
@@ -1,0 +1,9 @@
+class CreateChasersSent < ActiveRecord::Migration[6.0]
+  def change
+    create_table :chasers_sents do |t|
+      t.string :type
+      t.integer :application_choice_id
+      t.timestamps
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_01_28_164924) do
+ActiveRecord::Schema.define(version: 2020_01_31_130342) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
@@ -153,6 +153,13 @@ ActiveRecord::Schema.define(version: 2020_01_28_164924) do
     t.datetime "last_signed_in_at"
     t.index ["email_address"], name: "index_candidates_on_email_address", unique: true
     t.index ["magic_link_token"], name: "index_candidates_on_magic_link_token", unique: true
+  end
+
+  create_table "chasers_sents", force: :cascade do |t|
+    t.string "type"
+    t.integer "application_choice_id"
+    t.datetime "created_at", precision: 6, null: false
+    t.datetime "updated_at", precision: 6, null: false
   end
 
   create_table "course_options", force: :cascade do |t|


db/schema.rb
```diff
@@ -237,6 +244,7 @@
     t.string "hashed_sign_in_token"
     t.boolean "consent_to_be_contacted"
     t.string "feedback_status", default: "not_requested_yet", null: false
+    t.jsonb "questionnaire"
```
tvararu: Unrelated to this PR?

db/migrate/20200131101835_create_chaser_emails_table.rb
```diff
@@ -0,0 +1,9 @@
+class CreateChaserEmailsTable < ActiveRecord::Migration[6.0]
+  def change
+    create_table :chaser_emails_tables do |t|
```
mnacos: Don't like `:chaser_emails_tables` much. What will the model be called? I'd probably suggest just `SentChaser` and `:sent_chasers` (to make it clear we are talking about sent notifications, not pending ones). This would allow you to write things like `application_choice.sent_chasers.any?` etc.

db/schema.rb
mnacos: jsonb FTW

db/migrate/20200131101835_create_chaser_emails_table.rb
davidgisbey: Yea that was completely misnamed. It was meant to be chaser_emails, but yea sent_chaseers is way better

db/schema.rb
davidgisbey: 👍

davidgisbey: 👍

db/migrate/20200131130342_create_chasers_sent.rb
```diff
@@ -0,0 +1,9 @@
+class CreateChasersSent < ActiveRecord::Migration[6.0]
+  def change
+    create_table :chasers_sents do |t|
+      t.string :type
```
tijmenb: `type` is used by Rails for Single Table Inheritance, so this might conflict: https://stackoverflow.com/questions/7134559/rails-use-type-column-without-sti

db/migrate/20200131130342_create_chasers_sent.rb
```diff
@@ -0,0 +1,9 @@
...
+      t.integer :application_choice_id
```
tijmenb: I think this needs to be a polymorphic association? So that it can be used for the both the reference chaser and RDB/DBD chasers.

db/migrate/20200131101835_create_chaser_emails_table.rb
tijmenb: Does this mean that the model will be called `ChasersSent`, not `ChaserSent`?

db/migrate/20200131130342_create_chasers_sent.rb
tijmenb: BTW foreign keys are `bigint` in Rails (see schema.rb). These columns might benefit from a `null: false` like other associations.

davidgisbey: okay cool, Just having a reading into polymorphic associations and will make the changes in a new migration 👍

PR #1256, opened 2020-02-03
Speed up system tests

tijmenb:
## Context

I was investigating the performance of our system tests when I noticed a large number of log lines like this:

```
[Webpacker] Everything's up-to-date. Nothing to do
```

Screenshot 2020-02-03 at 09 20 52

I think this is Webpacker trying to compile the assets and then telling us that no CSS/JS changes were detected.

Turns out this system is *super* expensive. Turning the `compile` flag off in test causes a preposterous speedup in the tests:

```
Before: Finished in 2 minutes 36.5 seconds
After: Finished in 57.61 seconds
```

Because `compile: true` does a compile on each request, we have to make sure that our assets are compiled before running the tests, similarly to how production works. We can do this by sprinkling the compile task in relevant places:

```
RAILS_ENV=test rails webpacker:compile
```

I’ve also added a reminder that shows every test run, so that if you run `bundle exec rspec` you won’t get caught out. I think this is a reasonable trade-off for the speed up that we’re seeing.

Note that there’s something I don’t understand in the compilation - the current compilation saves the result in `public/packs-test`, but running our rake task saves it in `public/packs`. But it still works.

## Guidance to review

Not sure.

## Link to Trello card

https://trello.com/c/OEjmIn3C

## Things to check

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -30,7 +30,7 @@ rescue ActiveRecord::PendingMigrationError => e
   exit 1
 end
 
-puts "ℹ️ If you change CSS, JS, or Assets - don't forget to run `RAILS_ENV=test rails webpacker:compile` before your test runs"
+puts "ℹ️ If you change CSS, JS, or Assets - don't forget to run `rake compile_assets` before your test runs"
 
 Timecop.safe_mode = true
 
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -30,6 +30,8 @@ rescue ActiveRecord::PendingMigrationError => e
   exit 1
 end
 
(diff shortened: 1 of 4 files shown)


tvararu:
:100: preposterously faster!

spec/rails_helper.rb
```diff
@@ -30,6 +30,8 @@
   exit 1
 end
 
+puts "ℹ️ If you change CSS, JS, or Assets - don't forget to run `RAILS_ENV=test rails webpacker:compile` before your test runs"
```
C-gyorfi: Since you defined a compile_assets task in the rake file, we could just run `be rake compile_assets` 
```suggestion
puts "ℹ️ If you change CSS, JS, or Assets - don't forget to run `rake compile_assets` before your test runs"
```

PR #1259, opened 2020-02-03
Provider can confirm conditions met or not met for accepted offers

mnacos:
### Context

We need a UI within the provider console to allow providers to confirm whether the conditions set out as part of an offer have been met or not. This is high priority because we now have a user on our system in the accepted state and the relevant provider can't progress their application.

The scope for this work is limited to confirming all conditions have been met, not individual conditions. This means our system cannot be used for tracking the progress of meeting conditions at this stage. Conditions are either met in their entirety or not at all.

A minimal set of views, including confirmation steps, has been designed for this process. As these provider actions represent a different feedback phase for an application, a new controller and set of views will be added to the system.

Add new controller (`ConditionsController`) and associated routes to provider interface.

```
get '/applications/:application_choice_id/conditions' => 'conditions#edit', as: :application_choice_edit_conditions

patch '/applications/:application_choice_id/conditions/confirm' => 'conditions#confirm_update', as: :application_choice_confirm_update_conditions

patch '/applications/:application_choice_id/conditions' => 'conditions#update', as: :application_choice_update_conditions
```

In the future, we are planning to allow marking individual conditions as met/not met, this setup will accommodate this.

Add `ConfirmConditionsForm` to handle the presence validation for `:conditions_met`.

### Guidance to review

Run the specs.

To test manually, you'll need applications at the `pending_conditions` state (they show up as 'ACCEPTED' in the provider interface).

#### -- Starting point
image

#### -- Form
image

#### -- Confirmation page if conditions met
image

#### -- Confirmation page if conditions not met
image

#### -- Flash message when conditions met
image

#### -- Flash message when conditions not met
image

Note: The border to flash messages is blue instead of green, because that's the current definition of an :info level flash message in the app. Both are considered success cases.

### Link to Trello card

[1579 - Providers can confirm conditions met/not met (applies to all conditions)](https://trello.com/c/GsKZRset)

### Things to check

### DIFF
--- a/app/components/summary_list_component.html.erb
+++ b/app/components/summary_list_component.html.erb
@@ -1,9 +1,11 @@
 <dl class="govuk-summary-list">
   <% rows.each do |row| %>
     <div class="govuk-summary-list__row">
-      <dt class="govuk-summary-list__key">
-        <%= row[:key] %>
-      </dt>
+      <% if row[:key] %>
+        <dt class="govuk-summary-list__key">
+          <%= row[:key] %>
+        </dt>
+      <% end %>
       <dd class="govuk-summary-list__value">
         <% if row[:value].is_a?(Array) %>
           <% row[:value].each do |value| %><%= value %><br><% end %>
--- a/app/controllers/provider_interface/conditions_controller.rb
+++ b/app/controllers/provider_interface/conditions_controller.rb
@@ -1,0 +1,42 @@
+module ProviderInterface
+  class ConditionsController < ProviderInterfaceController
+    before_action :set_application_choice
+
+    def edit
+      @conditions_form = ConfirmConditionsForm.new
+    end
+
+    def confirm_update
+      @conditions_form = ConfirmConditionsForm.new(
+        conditions_met: params.dig(:provider_interface_confirm_conditions_form, :conditions_met),
+      )
+
+      render action: :edit unless @conditions_form.valid?
+    end
+
+    def update
+      @conditions_form = ConfirmConditionsForm.new(
+        conditions_met: params.dig(:provider_interface_confirm_conditions_form, :conditions_met),
+      )
+
+      redirect_to action: :edit unless @conditions_form.valid?
+
+      if @conditions_form.conditions_met?
+        ConfirmOfferConditions.new(application_choice: @application_choice).save || raise('ConfirmOfferConditions failure')
+        flash[:success] = 'Conditions successfully marked as met'
+      else
+        ConditionsNotMet.new(application_choice: @application_choice).save || raise('ConditionsNotMet failure')
+        flash[:success] = 'Conditions successfully marked as not met'
+      end
+
+      redirect_to provider_interface_application_choice_path(@application_choice.id)
+    end
+
+  private
+
+    def set_application_choice
+      @application_choice = GetApplicationChoicesForProviders.call(providers: current_provider_user.providers)
+        .find(params[:application_choice_id])
+    end
+  end
+end
--- a/app/views/provider_interface/conditions/confirm_update.html.erb
+++ b/app/views/provider_interface/conditions/confirm_update.html.erb
@@ -1,0 +1,41 @@
+<% content_for :browser_title, 'Confirm conditions' %>
+<% content_for :before_content, govuk_back_link_to(provider_interface_application_choice_path(@application_choice.id)) %>
+
+<%= form_with model: @conditions_form,
+      url: provider_interface_application_choice_update_conditions_path(@application_choice.id),
+      method: :patch do |f| %>
+  <%= f.govuk_error_summary %>
+
+  <h1 class="govuk-heading-xl">
+    <% if @conditions_form.conditions_met? %>
+      Are you sure the candidate met the conditions?
+    <% else %>
+      Are you sure the candidate did not meet the conditions?
+    <% end %>
+  </h1>
+
+  <%= render SummaryListComponent, rows: [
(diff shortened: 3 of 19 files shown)


app/controllers/provider_interface/conditions_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+    def edit
+      @conditions_form = ConfirmConditionsForm.new
+    end
+
+    def confirm_update
+      @conditions_form = ConfirmConditionsForm.new(
+        conditions_met: params.dig(:provider_interface_confirm_conditions_form, :conditions_met),
+      )
+
+      render action: :edit if !@conditions_form.valid?
```
tijmenb:
```suggestion
      render action: :edit unless @conditions_form.valid?
```

(like below)

app/views/provider_interface/conditions/edit.html.erb
```diff
@@ -0,0 +1,41 @@
...
+  <h1 class="govuk-heading-xl">
+    Confirm if the candidate has met all conditions
+  </h1>
+
+  <%= render SummaryListComponent, rows: [
+    { key: 'Candidate name', value: @application_choice.application_form.full_name },
+    { key: 'Course', value: @application_choice.course.name_and_code },
+    { key: 'Preferred location', value: @application_choice.site.name },
+    { key: 'Provider', value: @application_choice.course.provider.name_and_code },
+  ] %>
```
tijmenb: Perhaps we can extract https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/app/views/provider_interface/decisions/new_offer.html.erb#L16-L22 into a component to use here?

app/views/provider_interface/conditions/edit.html.erb
```diff
@@ -0,0 +1,41 @@
...
+  <h2 class="govuk-heading-l govuk-!-margin-top-8">
+    Conditions
+  </h2>
+
+  <% condition_rows = @application_choice.offer['conditions'].map { |condition| { value: condition } } %>
+  <% if condition_rows.empty? %>
+    <%= render SummaryListComponent, rows: [{ value: 'No conditions have been specified' }] %>
+  <% else %>
+    <%= render SummaryListComponent, rows: condition_rows %>
+  <% end %>
```
tijmenb: Could we re-use the code from the offer confirmation page here?

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/app/views/provider_interface/decisions/confirm_offer.html.erb#L19-L35

That means we won't have to change the summary list component.

adamsilver:
Nice work @mnacos just a few bits:

1. If there are no conditions, what do we do? Your screenshots so the user performing this task even when there are no conditions.

2. The flash message should be in a green border (‘success’)

3. We need to style the tags accordingly for these states

4. I think we'll need to make a few content tweaks (as you said on Slack)

app/views/provider_interface/conditions/edit.html.erb
mnacos: Ok, I'll create a `ConditionsComponent`

app/views/provider_interface/conditions/confirm_update.html.erb
```diff
@@ -0,0 +1,46 @@
...
+  <% if condition_rows.empty? %>
+    <%= render SummaryListComponent, rows: [{ value: 'No conditions have been specified' }] %>
+  <% else %>
+    <%= render SummaryListComponent, rows: condition_rows %>
+  <% end %>
+
+  <%= f.hidden_field :conditions_met %>
+
+  <% if @conditions_form.conditions_met? %>
+    <%= f.submit 'Yes I‘m sure – they met the conditions', class: 'govuk-button', data: { module: 'govuk-button' } %>
```
mnacos:
```suggestion
    <%= f.submit 'Yes I’m sure – they met the conditions', class: 'govuk-button', data: { module: 'govuk-button' } %>
```

app/views/provider_interface/conditions/confirm_update.html.erb
```diff
@@ -0,0 +1,46 @@
...
+  <% else %>
+    <%= f.submit 'Yes I‘m sure – they did not meet the conditions', class: 'govuk-button govuk-button--warning', data: { module: 'govuk-button' } %>
```
mnacos:
```suggestion
    <%= f.submit 'Yes I’m sure – they did not meet the conditions', class: 'govuk-button govuk-button--warning', data: { module: 'govuk-button' } %>
```

app/views/provider_interface/conditions/edit.html.erb
mnacos: I've created a component, however it doesn't look like we can have a single component for all cases. I've opened a draft card here: https://trello.com/c/JJZ1yaTg

app/views/provider_interface/conditions/edit.html.erb
mnacos: Problem is this code doesn't match the design Adam has given me. Allowing SummaryListComponent to list rows without keys is a minimal and useful change, no?

app/components/summary_list_component.html.erb
```diff
@@ -1,9 +1,11 @@
 <dl class="govuk-summary-list">
   <% rows.each do |row| %>
     <div class="govuk-summary-list__row">
-      <dt class="govuk-summary-list__key">
-        <%= row[:key] %>
-      </dt>
+      <% if row[:key] %>
```
tijmenb: We should refactor this away, because a definitions list without a `dt` is weird.

PR #1260, opened 2020-02-03
Send emails when application is insta-sent to providers

tijmenb:
## Context

Currently when the 2nd referee provides a reference, AND the candidate submitted more than 5 days ago (so the edit by date has elapsed), we send the application to the provider. We do this by calling `SendApplicationToProvider`, which doesn't send emails to providers and candidates.

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/commit/53d86dae594f535d94d0fc703926ca94a7f7237a does the important bits. The rest are mostly refactorings to make it easy.

## Guidance to review

Does this make sense?

## Link to Trello card

Related to https://trello.com/c/hzgtm77m

## Things to check

### DIFF
--- a/app/models/referee_interface/reference_feedback_form.rb
+++ b/app/models/referee_interface/reference_feedback_form.rb
@@ -16,7 +16,9 @@ module RefereeInterface
       ReceiveReference.new(
         reference: reference,
         feedback: feedback,
-      ).save
+      ).save!
+
+      true
     end
   end
 end
--- a/app/lib/test_applications.rb
+++ b/app/lib/test_applications.rb
@@ -63,7 +63,7 @@ module TestApplications
         ReceiveReference.new(
           reference: reference,
           feedback: 'You are awesome',
-        ).save
+        ).save!
       end
       return if states.include? :application_complete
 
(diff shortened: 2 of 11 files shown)


app/models/referee_interface/reference_feedback_form.rb
```diff
@@ -16,7 +16,9 @@ def save
       ReceiveReference.new(
         reference: reference,
         feedback: feedback,
-      ).save
+      ).save!
+
+      true
```
stevehook: How does the presence validation for `feedback` work with this change?

tijmenb: If the feedback isn't provided and form isn't `valid?` it'll return nil early:

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/686d8f8c6e5141a220cf5fc33108bfe7140d6d6d/app/models/referee_interface/reference_feedback_form.rb#L14

tijmenb:
> My only slight concern is what happens to emails sent out by the TestApplications? They should all be addressed to example.com rather than real people.

Factories only use `@example.com` emails, so that should be OK.

app/models/referee_interface/reference_feedback_form.rb
stevehook: should have spotted that!

PR #1263, opened 2020-02-03
861  Add working_pattern column to ApplicationExperience

C-gyorfi:
## Context
Going forward we want to allow candidates to give details about their working pattern when a work experience is part-time

This PR adds working_pattern column to ApplicationExperience table

## Guidance to review

Does the column name make sense? I assume we don't need to backfill nullable fields, right?

## Link to Trello card

https://trello.com/c/M9ktUBA0/861-add-field-for-capturing-working-patterns-for-part-time-jobs

## Things to check


db/schema.rb
```diff
@@ -148,7 +149,7 @@
     t.string "magic_link_token"
     t.datetime "magic_link_token_sent_at"
     t.boolean "hide_in_reporting", default: false, null: false
-    t.bigint "course_from_find_id"
+    t.integer "course_from_find_id"
```
duncanjbrown: This is interesting. We usually see changes in schema dumps which reorder columns (see submitted_at above) but I don't think we've had a column type change. I don't know if this will cause badthings. What's `psql --version` on your machine?

C-gyorfi: Wow, good catch its `psql (PostgreSQL) 12.1`

C-gyorfi: I had a quick chat with @davidgisbey and actually this meant to be an `integer` the mistery is what caused this to be a `bigint` in the first place? If you look into the migrations you can see its defined as `integer`:
https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/db/migrate/20200120120940_add_course_from_find_id_to_candidate_model.rb#L3

C-gyorfi: I think it should be safe to merge this but I'll wait for @tijmenb 's input!

tijmenb: `schema.rb` is only used by CI, so this should be fine to merge. Hopefully this is the last we see of these changes.

PR #1266, opened 2020-02-03
Hide date of birth from providers

fofr:
Don't show the date of birth to providers, for the purposes of diversity.

Draft PR, waiting on product/policy input.

### DIFF
--- a/app/components/personal_details_component.rb
+++ b/app/components/personal_details_component.rb
@@ -18,7 +18,6 @@ class PersonalDetailsComponent < ActionView::Component::Base
   def rows
     [
       name_row,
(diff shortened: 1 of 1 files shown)


fofr:
Spoke with @benilovj about this – we probably need to be a little smarter than this, as we'll need to show the date of birth in the API, which would mean provider UI users would get different information to API users.

Suggestions include:
- making date of birth available but not showing by default (eg behind a disclosure)
- moving date of birth down the page so it's not a factor that forms part of the first impression

duncanjbrown:
@fofr would you object to closing this until we have a plan?

PR #1267, opened 2020-02-03
Use different components for references in support & provider UI

tijmenb:
##  Context

We currently show the “Given consent for research?” value to the provider, which we shouldn’t. It was introduced in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1121.

This PR solves the problem by separating the components entirely, and making the changes to the provider side. I’ve also removed the status from provider UI, which should always be “feedback provided” (otherwise providers can’t see it). Also removed a bunch of if-statements that are no longer necessary, such as the title and whether or not to show the send email links.

Support UI:

Screenshot 2020-02-03 at 16 57 44

Provider UI:

Screenshot 2020-02-03 at 16 57 56

## Link to Trello card

Via @fofr.

## Things to check


fofr:
Nice 👍

PR #1270, opened 2020-02-03
Refactor calculating of work gaps and add handling for current job covering following breaks

chubberlisk:
## Context

We currently don't handle a scenario where a candidate's current job covers breaks in other jobs.

For example:

- Job 1 => start_date: Jan 2019, end_date: nil
- Job 2 => start_date: Mar 2019, end_date: May 2019
- Job 3 => start_date: Sept 2019, end_date: December 2019

There is a break between Job 2 and Job 3 however, we don't need to show this as Job 1 covers this period.

This PR refactors how we calculate work breaks in order to cover this scenario and to make it easier to test.

It does so by:

- removing responsibility of `BreakInWorkHistoryComponent` and `WorkHistoryReviewComponent` to calculate work breaks
- redirecting the responsibility to a new service called `GetWorkHistoryWithBreaks` which takes in a work history and returns the work history with breaks via an array of hashes with a `type` and `entry` key

This means the only responsibility of the components is to render.

## Screenshot

image

## Guidance to review

Would like feedback on:

- whether there's a simpler way to do this with the current implementation instead of this refactoring
- whether the refactoring makes sense
- whether test coverage is good enough or any edge cases are missing
- whether `GetBreaksInMonths` service should be moved into `GetWorkHistoryWithBreaks`
- whether `GetWorkHistoryWithBreaks` and/or its spec can be further refactored

## Link to Trello card

https://trello.com/c/RgdlCMai/862-multiple-current-work-for-work-experiences

## Things to check

### DIFF
--- a/app/components/work_history_review_component.html.erb
+++ b/app/components/work_history_review_component.html.erb
@@ -1,18 +1,20 @@
-<%@application_form.application_work_experiences.order(:start_date).to_a.append(nil).each_cons(2) do |work1, work2| %>
-  <%= render(SummaryCardComponent, rows: work_experience_rows(work1), editable: @editable) do %>
-    <%= render(SummaryCardHeaderComponent, title: work1.role, heading_level: @heading_level, check_icon: work1.working_with_children) do %>
-      <% if @editable %>
-        <div class="app-summary-card__actions">
-          <%= link_to candidate_interface_work_history_destroy_path(work1.id), class: 'govuk-link' do %>
-            <%= t('application_form.work_history.delete_entry') %><span class="govuk-visually-hidden"><%= generate_action(work: work1) %></span>
-          <% end %>
-        </div>
+<% @work_history_with_breaks.each do |history| %>
+  <% if history[:type] == :job %>
+    <%= render(SummaryCardComponent, rows: work_experience_rows(history[:entry]), editable: @editable) do %>
+      <%= render(SummaryCardHeaderComponent, title: history[:entry].role, heading_level: @heading_level, check_icon: history[:entry].working_with_children) do %>
+        <% if @editable %>
+          <div class="app-summary-card__actions">
+            <%= link_to candidate_interface_work_history_destroy_path(history[:entry].id), class: 'govuk-link' do %>
+              <%= t('application_form.work_history.delete_entry') %><span class="govuk-visually-hidden"><%= generate_action(work: history[:entry]) %></span>
+            <% end %>
+          </div>
+        <% end %>
       <% end %>
     <% end %>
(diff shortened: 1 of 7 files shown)


app/components/work_history_review_component.html.erb
```diff
@@ -1,18 +1,20 @@
...
+            <% end %>
+          </div>
+        <% end %>
       <% end %>
     <% end %>
-  <% end %>
-
-  <% if FeatureFlag.active?('work_breaks') %>
-    <%= render(BreakInWorkHistoryComponent, work1: work1, work2: work2) %>
+  <% elsif history[:type] == :break %>
```
C-gyorfi: Neat refactoring

spec/services/get_work_history_with_breaks_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+        work_history = [job1]
+
+        get_work_history_with_breaks = GetWorkHistoryWithBreaks.new(work_history)
+        work_history_with_breaks = get_work_history_with_breaks.call
+
+        expect(work_history_with_breaks.count).to eq(1)
+        expect(work_history_with_breaks[0]).to eq(type: :job, entry: job1)
+      end
+
+      it 'returns an array containing a hash for a job and a break if one month break' do
```
C-gyorfi: Maybe I'm just tired and I'm missing out, but is there a test case for a break that is less than 1 month, if I remember correctly that would not count as a break right?

chubberlisk: I believe the spec covers this.

```ruby
it 'returns an array containing hashes for jobs if there are no breaks' do
  job1 = build_stubbed(:application_work_experience, start_date: september2019, end_date: october2019)
  job2 = build_stubbed(:application_work_experience, start_date: october2019, end_date: december2019)
  job3 = build_stubbed(:application_work_experience, start_date: january2020, end_date: current_date)
  work_history = [job1, job2, job3]

  get_work_history_with_breaks = GetWorkHistoryWithBreaks.new(work_history)
  work_history_with_breaks = get_work_history_with_breaks.call

  expect(work_history_with_breaks.count).to eq(3)
  expect(work_history_with_breaks[0]).to eq(type: :job, entry: job1)
  expect(work_history_with_breaks[1]).to eq(type: :job, entry: job2)
  expect(work_history_with_breaks[2]).to eq(type: :job, entry: job3)
end
```

`job1.end_date` and `job2.start_date` are on the boundary and then `job2.end_date` and `job3.start_date` is December 2019 and January 2020 which means no break, from my understanding we are only considering whole months between two dates.

PR #1283, opened 2020-02-05
Provider withdraws offered applications

stevehook:
## Context

Providers can withdraw an offer that has not yet been accepted by the candidate. This PR implements the UI and backend logic to permit that.

The database schema changes that this depends on are already in production. https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1272

- New feature flag `provider_change_response` to guard all the functionality in this PR.
- UI using the design https://manage-applications-beta.herokuapp.com/application/103/edit-response to prompt the user for options and withdrawal reasons.
- We use the existing `rejected` status for withdrawn offers.
- Add `ApplicationChoice#offer_withdrawal_reason` and `offer_withdrawn_at` attributes.
- Changes to the way to present statuses in provider interface so that we distinguish between rejected applications that were simply rejected or withdrawn from the offer state.

Still to do:
- [x] Check the API - add attributes specific to withdrawn offers
- [x] Test the candidate front-end
- [x] Fix specs

## Guidance to review

- Does it make sense to have a separate `WithdrawOffer` service (as well as `RejectApplication`)?
- Does it make sense to reuse the `rejected` status for `offer_withdrawn` or would it be better to have an explicit `offer_withdrawn` state? - the implication is that we end up with hoops to jump through when we come to display status to the user. There is now a hacky workaround in the spec for `ApplicationStateChange`.
- Does it still make sense to lever the offer withdrawn information (`offer_withdrawal_reason`) into the `reject` key? Is this too inconsistent with the Provider interface that treats these as separate things?

withdraw-offer

On the candidate interface:

image

## Link to Trello card

https://trello.com/c/KLy8enLu/1482-ui-reject-offered-applications

## Things to check


app/services/set_decline_by_default.rb
```diff
@@ -10,6 +10,7 @@ def call
       application_choices.maximum(:offered_at),
       application_choices.maximum(:rejected_at),
       application_choices.maximum(:withdrawn_at),
+      application_choices.maximum(:offer_withdrawn_at),
```
mnacos: nice

stevehook:
> Haven't tested manually but the code looks good. It also handles side-effects, such as extending DBD periods allowing a candidate to re-assess their options. Was there a decision not to email candidates about the retraction of offers or will this be handled as a separate PR?

I think Tijmen said that Candidate team would be doing all the emails and in any case it'll be a separate PR.

app/services/withdraw_offer.rb
```diff
@@ -0,0 +1,33 @@
...
+    return false unless valid?
+
+    ActiveRecord::Base.transaction do
+      ApplicationStateChange.new(@application_choice).reject!
+      @application_choice.update!(
+        offer_withdrawal_reason: @offer_withdrawal_reason,
+        offer_withdrawn_at: Time.zone.now,
+      )
+      SetDeclineByDefault.new(application_form: @application_choice.application_form).call
+      StateChangeNotifier.call(:withdraw_offer, application_choice: @application_choice)
```
duncanjbrown: Does it make sense for this to be inside the transaction?

app/models/provider_interface/withdraw_offer_form.rb
```diff
@@ -0,0 +1,8 @@
+module ProviderInterface
```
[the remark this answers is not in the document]
stevehook: All the other 'forms' in this directory have the same suffix so it seems to be a convention?

PR #1284, opened 2020-02-05
Allow candidates to have more than 2 references

tijmenb:
## Context

In https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1279 we're going to allow more than 2 references. For https://trello.com/c/hzgtm77m we need to manually add a reference. 

We need to update our assumptions in several places.

Assume more than 2 references can come back.

## Guidance to review

Does this make sense? Are there any other places where we assume that there are only 2 references?

## Link to Trello card

https://trello.com/c/hzgtm77m/863-changes-related-to-application-ct3898

## Things to check

### DIFF
--- a/app/models/application_form.rb
+++ b/app/models/application_form.rb
@@ -29,10 +29,6 @@ class ApplicationForm < ApplicationRecord
     submitted_at.present?
   end
 
-  def application_references_complete?
-    application_references.feedback_provided.count == MINIMUM_COMPLETE_REFERENCES
-  end
-
   def awaiting_provider_decisions?
     application_choices.where(status: :awaiting_provider_decision).any?
   end
--- a/app/models/application_form.rb
+++ b/app/models/application_form.rb
@@ -11,7 +11,6 @@ class ApplicationForm < ApplicationRecord
   has_many :application_references, -> { order('id ASC') }
 
   MINIMUM_COMPLETE_REFERENCES = 2
-  validates_length_of :application_references, maximum: MINIMUM_COMPLETE_REFERENCES
 
   enum phase: {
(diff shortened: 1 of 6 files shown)


app/presenters/candidate_interface/application_form_presenter.rb
```diff
@@ -156,7 +156,7 @@ def volunteering_added?
     end
 
     def all_referees_provided_by_candidate?
-      @application_form.application_references.count == ApplicationForm::MINIMUM_COMPLETE_REFERENCES
+      @application_form.application_references.count >= ApplicationForm::MINIMUM_COMPLETE_REFERENCES
```
stevehook: It might make sense to make a similar change in the `ReceiveReference` service to be on the safe side, though I don't think it would get called 'out of order'. It gets called when the candidate submits but that's only for sandbox and wouldn't cause a problem there anyway.

tijmenb: Updated the comment so that it's clearer we need to only progress applications when the number of received applications change from 1 to 2.

app/models/application_form.rb
```diff
@@ -30,10 +29,6 @@ def submitted?
     submitted_at.present?
   end
 
-  def application_references_complete?
-    application_references.feedback_provided.count == MINIMUM_COMPLETE_REFERENCES
-  end
```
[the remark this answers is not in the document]
tijmenb: This was updated in a previous PR: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/commit/bef0bae55066a02a5557e93826d3fbaad74c2478

PR #1286, opened 2020-02-05
Show refbot1 and refbot2 instructions in a Sandbox-only block

duncanjbrown:
## Context

Dan Williams pointed out that during the demo he'd felt he needed to remember the refbot1 and refbot2 email addresses. We should present this information close to where it's needed by users of the sandbox.

Add another sandbox-only block, this time on the "Details of referee" page

Screenshot 2020-02-05 at 13 58 55

## Guidance to review

Check the copy. Should the field be inside the box?

## Things to check

### DIFF
--- a/app/views/candidate_interface/referees/_form.html.erb
+++ b/app/views/candidate_interface/referees/_form.html.erb
@@ -6,6 +6,18 @@
 </ul>
 
 <%= f.govuk_text_field :name, label: { text: t('application_form.referees.name.label'), size: 'm' } %>
+<% if HostingEnvironment.sandbox_mode? %>
+  <div class="app-status-box app-status-box--sandbox govuk-!-margin-bottom-4">
+    <p class="govuk-body">
+      <strong class="govuk-tag govuk-phase-banner__content__tag app-environment-tag app-environment-tag--sandbox">
+        sandbox feature
+      </strong>
+    </p>
+    <p class="govuk-body govuk-!-margin-bottom-0">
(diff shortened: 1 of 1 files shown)


benilovj:
Could be just me, but the way the styling is applied makes it look like "Email address" is a sandbox feature...

mnacos:
> Could be just me, but the way the styling is applied makes it look like "Email address" is a sandbox feature...

I agree, maybe the email field should stay outside the purple box.

fofr:
I agree about separating email from the box. Otherwise looks good.

duncanjbrown:
Thanks all. Updated

Screenshot 2020-02-05 at 15 52 32

app/views/candidate_interface/referees/_form.html.erb
```diff
@@ -6,6 +6,18 @@
...
 
 <%= f.govuk_text_field :name, label: { text: t('application_form.referees.name.label'), size: 'm' } %>
+<% if HostingEnvironment.sandbox_mode? %>
+  <div class="app-status-box app-status-box--sandbox govuk-!-margin-bottom-4">
+    <p class="govuk-body">
+      <strong class="govuk-tag govuk-phase-banner__content__tag app-environment-tag app-environment-tag--sandbox">
+        sandbox feature
+      </strong>
+    </p>
+    <p class="govuk-body">
```
fofr:
```suggestion
    <p class="govuk-body govuk-!-margin-bottom-0">
```

duncanjbrown: Done

PR #1290, opened 2020-02-05
Email to provider when a decision is needed on an application because it has been submitted for 20 working days

C-gyorfi:
## Context

Currently, a support user must notify the provider when they are halfway through their RBD period. We want to send automatic emails to providers warning them of RBD.

This PR adds:
- GetApplicationFormsWaitingForProviderDecision service to find application choices that halfway through the RBD period
- SendChaseEmailToProvider service to call the provider mailer and send chaser emails
- SendChaseEmailToProvidersWorker so we can schedule the task

## Guidance to review

I would suggest following the commit and the system spec also can be helpful. Before we can merge this PR we need to backfill the chased provider data so we won't double send chasers.

## Link to Trello card

## Things to check

### DIFF
--- a/app/services/get_application_forms_waiting_for_provider_decision.rb
+++ b/app/services/get_application_forms_waiting_for_provider_decision.rb
@@ -1,0 +1,12 @@
+class GetApplicationFormsWaitingForProviderDecision
+  def self.call
+    ApplicationChoice
+      .where(status: :awaiting_provider_decision)
+      .where('reject_by_default_at < ?', chase_provider_time_limit)
+      .where.not(id: ChaserSent.provider_decision_request.select(:chased_id))
+  end
+
+  def self.chase_provider_time_limit
+    TimeLimitCalculator.new(rule: :chase_provider_by, effective_date: Time.zone.now).call.second
+  end
+end
--- a/app/services/send_chase_email_to_provider.rb
+++ b/app/services/send_chase_email_to_provider.rb
@@ -1,0 +1,17 @@
+class SendChaseEmailToProvider
+  def self.call(application_choice:)
+    application_choice.provider.provider_users.each do |provider_user|
+      ProviderMailer.chase_provider_decision(provider_user, application_choice).deliver
+      audit_chase_email(provider_user, application_choice)
+    end
+    ChaserSent.create!(chased: application_choice, chaser_type: :provider_decision_request)
+  end
+
+  def self.audit_chase_email(provider_user, application_choice)
+    course_name_and_code = application_choice.course.name_and_code
+    audit_comment =
+      "Chase emails have been sent to the provider #{provider_user.email_address} because " +
+      "the application for #{course_name_and_code} is close to its RBD date."
+    application_choice.application_form.update!(audit_comment: audit_comment)
+  end
+end
--- a/app/workers/send_chase_email_to_providers_worker.rb
+++ b/app/workers/send_chase_email_to_providers_worker.rb
@@ -1,0 +1,11 @@
+class SendChaseEmailToProvidersWorker
+  include Sidekiq::Worker
+
+  def perform
+    return unless FeatureFlag.active?('automated_provider_chaser')
+
+    GetApplicationFormsWaitingForProviderDecision.call.each do |application_choice|
+      SendChaseEmailToProvider.call(application_choice: application_choice)
+    end
+  end
+end
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -264,4 +264,9 @@ FactoryBot.define do
     first_name { Faker::Name.first_name }
     last_name { Faker::Name.last_name }
   end
+
+  factory :provider_users_provider do
+    provider
+    provider_user
+  end
 end
--- a/spec/services/get_application_forms_waiting_for_provider_decision_spec.rb
+++ b/spec/services/get_application_forms_waiting_for_provider_decision_spec.rb
@@ -1,0 +1,53 @@
+require 'rails_helper'
+
+RSpec.describe GetApplicationFormsWaitingForProviderDecision do
+  let(:current_time) { Time.zone.local(2019, 6, 1, 12, 0, 0) }
+
+  around do |example|
+    Timecop.freeze(current_time) do
+      example.run
+    end
+  end
+
+  def create_application(status:, reject_by_default_at:)
+    create(
+      :application_choice,
+      status: status,
+      reject_by_default_at: reject_by_default_at,
+    )
+  end
+
+  it 'returns an application that has not more than 20 working days till RBD date' do
+    application_choice = create_application(
+      status: 'awaiting_provider_decision',
+      reject_by_default_at: 21.business_days.from_now,
+    )
+
+    Timecop.travel(1.business_days.from_now) do
+      expect(described_class.call).to include application_choice
+    end
+  end
+
+  it 'does not return an application that has more than 20 working days till RBD date' do
+    application_choice = create_application(
+      status: 'awaiting_provider_decision',
+      reject_by_default_at: 22.business_days.from_now,
+    )
+    Timecop.travel(1.business_days.from_now) do
+      expect(described_class.call).not_to include application_choice
+    end
+  end
+
+  it 'does not return an application if it has been chased already' do
+    application_choice = create_application(
+      status: 'awaiting_provider_decision',
+      reject_by_default_at: 10.business_days.from_now,
+    )
+
+    ChaserSent.create!(chased: application_choice, chaser_type: :provider_decision_request)
+
+    Timecop.travel(1.business_days.from_now) do
+      expect(described_class.call).not_to include application_choice
+    end
+  end
+end
--- a/spec/system/provider_interface/provider_receives_email_when_an_application_is_waiting_for_decision_for_20_days_spec.rb
+++ b/spec/system/provider_interface/provider_receives_email_when_an_application_is_waiting_for_decision_for_20_days_spec.rb
@@ -1,0 +1,50 @@
+require 'rails_helper'
+
+RSpec.feature 'An application is waiting for decision for 20 working days' do
+  include CourseOptionHelpers
+  include CandidateHelper
+
+  scenario 'the provider receives a chaser email', sidekiq: true do
+    FeatureFlag.activate('training_with_a_disability')
+    FeatureFlag.activate('automated_provider_chaser')
+
+    given_a_candidate_has_submitted_an_application_form
+    and_an_application_was_sent_to_provider
+    and_i_am_a_provider_user_at_the_course_provider
+
+    when_the_application_is_waiting_for_decision_for_20_working_days
+
+    then_i_should_receive_a_chaser_email_with_a_link_to_the_application
+  end
+
+  def given_a_candidate_has_submitted_an_application_form
+    candidate_completes_application_form
+    candidate_submits_application
+  end
+
+  def and_an_application_was_sent_to_provider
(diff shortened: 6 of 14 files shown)


app/services/get_application_forms_waiting_for_provider_decision.rb
```diff
@@ -0,0 +1,12 @@
+class GetApplicationFormsWaitingForProviderDecision
+  def self.call
+    ApplicationChoice
+      .where(status: :awaiting_provider_decision)
+      .joins(:application_form)
+      .where(['application_forms.submitted_at < ?', 20.business_days.ago])
```
benilovj: The constant 20 is effectively part of the business rules, because it's a counterpart to the 40-day limit. I think the 20 should be defined in [`TimeLimitConfig`](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/app/lib/time_limit_config.rb) and pulled into here.

app/services/send_chase_email_to_provider.rb
```diff
@@ -0,0 +1,5 @@
+class SendChaseEmailToProvider
+  def self.call(application_choice:)
+    ChaserSent.create!(chased: application_choice.provider, chaser_type: :provider_decision_request)
```
tijmenb: This is slightly off I think! We're saving one ChaserSent per `provider`, so this would cause only one `provider_decision_request` to be sent per provider ever. What we want to achieve is that we send only one email to the provider per application choice.

app/services/get_application_forms_waiting_for_provider_decision.rb
tijmenb: From offline convo:

```suggestion
      .where(['application_forms.reject_by_default_at > ?', TimeLimitConfig::REMIND_AFTER_DAYS_WHATEVER.business_days.in_the_future])
```

C-gyorfi: Thanks :)

app/services/get_application_forms_waiting_for_provider_decision.rb
```diff
@@ -0,0 +1,15 @@
+class GetApplicationFormsWaitingForProviderDecision
+  def self.call
+    ApplicationChoice
+      .where(status: :awaiting_provider_decision)
+      .where('reject_by_default_at < ?', chase_provider_time_limit)
+      .select do |choice|
+        choice unless ChaserSent.provider_decision_request.map(&:chased_id)
+                                .include?(choice.id)
```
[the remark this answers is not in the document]
stevehook: You can do something like: `ApplicationChoice.left_outer_joins(:chasers_sent).where(chasers_sent: { id: nil })` if you set up the `ApplicationChoice#chasers_sent` association or use SQL.

app/services/get_application_forms_waiting_for_provider_decision.rb
```diff
@@ -0,0 +1,12 @@
+class GetApplicationFormsWaitingForProviderDecision
+  def self.call
+    ApplicationChoice
+      .where(status: :awaiting_provider_decision)
+      .where('reject_by_default_at < ?', chase_provider_time_limit)
+      .where.not(id: ChaserSent.provider_decision_request.select(:chased_id))
```
stevehook: This could get expensive as the database grows because `ChaserSent.provider_decision_request.map` is loading all the `ChaserSent` activerecord objects for that type (ever). I think you could use `pluck` to avoid instantiating the activerecord objects but I think a better way would be to do an outer join combined with an is null condition to get the DB to do all the heavy lifting.

You can do something like: `ApplicationChoice.left_outer_joins(:chasers_sent).where(chasers_sent: { id: nil })` if you set up the `ApplicationChoice#chasers_sent` association or use SQL.

app/services/get_application_forms_waiting_for_provider_decision.rb
```diff
@@ -0,0 +1,12 @@
...
+  end
+
+  def self.chase_provider_time_limit
+    TimeLimitCalculator.new(rule: :chase_provider_by, effective_date: Time.zone.now).call.second
```
stevehook: `effective_date` here should really be the time at which the clock started (because in theory at least the time limits might be different now). In other words counting back from the RBD date _may_ give different results to counting forward.

I'm not sure what to suggest here because we don't record the time at which the clock started (there is no `ApplicationChoice#sent_to_provider_at`) and I know you've already chatted with @tijmenb about it. I think perhaps this is the best solution we can have as things stand.

An alternative would be to drop the configurable limit (in `TimeLimitConfig`) and just count back `ApplicationChoice. reject_by_default_days / 2` days from `ApplicationChoice. reject_by_default_at`.

app/services/send_chase_email_to_provider.rb
```diff
@@ -0,0 +1,17 @@
...
+      audit_chase_email(provider_user, application_choice)
+    end
+    ChaserSent.create!(chased: application_choice, chaser_type: :provider_decision_request)
+  end
+
+  def self.audit_chase_email(provider_user, application_choice)
+    course_name_and_code = application_choice.course.name_and_code
+    audit_comment =
+      "Chase emails have been sent to the provider #{provider_user.email_address} because " +
+      "the application for #{course_name_and_code} is half way through its RBD period."
```
stevehook: With the `TimeLimitConfig` setup we have it's possible that we could tweak things so that it's not half way.

spec/system/provider_interface/provider_receives_email_when_an_application_is_waiting_for_decision_for_20_days_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    create(:provider_users_provider, provider_id: provider_id, provider_user_id: @provider_user.id)
+  end
+
+  def when_the_application_is_waiting_for_decision_for_20_working_days
+    Timecop.travel(20.business_days.after(Time.zone.now + 5.minutes)) do
+      SendChaseEmailToProvidersWorker.perform_async
+    end
+  end
+
+  def then_i_should_receive_a_chaiser_email_with_a_link_to_the_application
```
stevehook:
```suggestion
  def then_i_should_receive_a_chaser_email_with_a_link_to_the_application
```

app/services/send_chase_emails_to_provider.rb
```diff
@@ -0,0 +1,7 @@
+class SendChaseEmailsToProvider
+  def call
+    GetApplicationFormsWaitingForProviderDecision.call.each do |application_choice|
+      SendChaseEmailToProvider.call(application_choice: application_choice)
+    end
```
tijmenb: Perhaps we can move this code into the worker?

app/workers/send_chase_email_to_providers_worker.rb
```diff
@@ -0,0 +1,7 @@
+class SendChaseEmailToProvidersWorker
+  include Sidekiq::Worker
+
+  def perform
+    SendChaseEmailsToProvider.new.call
```
tijmenb: As discussed offline, we should add a feature flag here so that we can deploy without the backfil.

spec/factories.rb
```diff
@@ -264,4 +264,12 @@
     first_name { Faker::Name.first_name }
     last_name { Faker::Name.last_name }
   end
+
+  factory :provider_users_provider do
+    provider
+    provider_user
+
+    provider_id { provider.id }
+    provider_user_id { provider_user.id }
```
tijmenb: These may be not necessary, the model should have the attributes.

spec/services/get_application_forms_waiting_for_provider_decision_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def create_application(status:, reject_by_default_at:)
+    application_form = create :application_form
+    create(
+      :application_choice,
+      application_form: application_form,
+      status: status,
+      reject_by_default_at: reject_by_default_at,
+    )
+    application_form.reload
```
tijmenb: You could return the `application_choice` so we don't have to do `application_form.application_choices.first` elsewhere.

spec/system/provider_interface/provider_receives_email_when_an_application_is_waiting_for_decision_for_20_days_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  scenario 'the provider receives a chaser email', sidekiq: true do
+    FeatureFlag.activate('training_with_a_disability')
+
+    given_a_candidate_has_submitted_an_application_form
+    and_an_application_was_sent_to_provider
+    and_i_am_a_provider_user_at_the_course_provider
+
+    when_the_application_is_waiting_for_decision_for_20_working_days
+
+    then_i_should_receive_a_chaiser_email_with_a_link_to_the_application
```
tijmenb:
```suggestion
    then_i_should_receive_a_chaser_email_with_a_link_to_the_application
```

app/services/send_chase_email_to_provider.rb
```diff
@@ -0,0 +1,17 @@
+class SendChaseEmailToProvider
+  def self.call(application_choice:)
+    application_choice.provider.provider_users.each do |provider_user|
```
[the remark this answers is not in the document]
C-gyorfi: Yes its not great, but we agreed on a previous provider mailer PR that we will send emails to all provider users for now and will figure out something later

app/services/send_chase_email_to_provider.rb
C-gyorfi: What do you think about `...is close to its RBD date.`?

stevehook: could we say `is less than #{TimeLimitConfig...} days from it's RBD date`? (I'm not a content person!)

app/services/get_application_forms_waiting_for_provider_decision.rb
C-gyorfi: Yes, you are right, the problem with `ApplicationChoice. reject_by_default_days / 2` assume the business rule will be always halfway rough RBD. We agreed with @tijmenb that we won't record a chase_provider_at date but we count backwards from the `reject_by_default_at`. I think as long as this rule changes following the current logic it should be fine. I'm more concerned about the naming and readability.

spec/factories.rb
C-gyorfi: You are right!

app/services/get_application_forms_waiting_for_provider_decision.rb
tijmenb: I still haven't wrapped my head around it. Let's make sure we did the right thing though, how do we test that?

C-gyorfi:
> (the task won't do anything because it's in a method, but that's fine if we copy-paste)

That was my dodgy copy paste job, I fixed that 😃

app/services/get_application_forms_waiting_for_provider_decision.rb
C-gyorfi: I missed your comment @tijmenb but I could change the tests to something like
 ``` it 'returns an application that has not more than 20 working days till RBD date' do
    application_choice = create_application(
      status: 'awaiting_provider_decision',
      reject_by_default_at: TimeLimitConfig.limits_for(:chase_provider_by).first.limit.business_days.from_now,
    )
    expect(described_class.call).to include application_choice
  end```
In this case, you can change the time limit to any number and it would find the application choices that have the number of days before RBD. It would be also good to decouple the 20 days rule from the unit tests, so the rule can change and not breaking the current test. Let me know what you think.

PR #1293, opened 2020-02-06
Add chase_provider_decision_at column to application_chices

C-gyorfi:
## Context

We want to send chaser emails to providers when an application has been submitted for 20 working days
The 20 working days rule part of the business rules because it's a counterpart to the 40-day limit, therefore it would make sense to add it to the schema just like `reject_by_default_at`

This PR adds the `chase_provider_decision_at` column to the `application_chices` table

## Guidance to review

Do we need to backfill this data? I assume no API updates needed, is that right?

## Link to Trello card

## Things to check


db/schema.rb
```diff
@@ -151,7 +152,7 @@
     t.string "magic_link_token"
     t.datetime "magic_link_token_sent_at"
     t.boolean "hide_in_reporting", default: false, null: false
-    t.bigint "course_from_find_id"
+    t.integer "course_from_find_id"
```
stevehook: Is this intentional, guess not?

I think I had exactly the same thing happen once and ended up resetting my dev db. Maybe a postgres version thing.

C-gyorfi: I believe its a version thing, according to the migration files this should be `integer` but a previous migration changed https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/commit/8e3dcb555bfdc2fd9432de7744bd4ff8899b96f9

we discussed this before with @tijmenb and it should be harmless

C-gyorfi: What is your `psql --version` @davidgisbey ?

stevehook: I'm on 10.10 but I've not upgraded for ages

C-gyorfi:
> I think we should use the ChaserSent feature that was added yesterday for this?

Yes, but regardless, I think we need to set a `chase provider` date when we send an application to the provider as suggested by @benilovj on this WIP PR https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1290

tijmenb:
We'd have the date in the ChaserSent table. Not sure what comment you're referring to?

C-gyorfi:
The date in the ChaserSent table is when the chaser sent and the `chase_provider_decision_at` field would be populated when we send the application to the provider(20 business days from that date)

PR #1294, opened 2020-02-06
Backfill referee chasers

davidgisbey:
## Context

We currently have 4 candidates who meet the criteria for our automated chaser that will be implemented in #1228. 

All four have these have already been sent chaser emails.

- A migration that creates ChaserSent's for each reference that currently meets the criteria for the job (feedback requested and created more than 5 days ago).
 
## Guidance to review

Not 100% if a migration is the most appropriate way to implement this.
Any thoughts?

## Link to Trello card

## Things to check


db/schema.rb
```diff
@@ -246,6 +249,7 @@
     t.string "hashed_sign_in_token"
     t.boolean "consent_to_be_contacted"
     t.string "feedback_status", default: "not_requested_yet", null: false
+    t.jsonb "questionnaire"
```
tijmenb: You got some unrelated stuff in here I think.

davidgisbey: Thanks!

PR #1295, opened 2020-02-06
Refactor `GetWorkHistoryWithBreaks` to `WorkHistoryWithBreaks` and its implementation

chubberlisk:
## Context

In https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1270, we covered a scenario which we previously didn't. When discussing with @duncanjbrown, he uncovered another one we didn't cover.

1. Job 1: Jan 2019 to Oct 2019
2. Job 2: Mar 2019 to May 2019
3. Job 3: Sept 2019 to Dec 2019

There is a break between Job 2 and Job 3 of 3 months and a break between Job 3 and now of 1 month. However, we don’t want to display the first break because it’s covered by Job 1.

This PR:

- changes the name of the service from `GetWorkHistoryWithBreaks` to `WorkHistoryWithBreaks`
- completely overhauls the previous implementation in order to cover the above scenario not covered
- returns a `BreakPlaceholder` instead of an `OpenStruct`

To calculate work breaks it:

1. Creates a timeline that includes all the months from the first job to now
2. Removes months in the timeline that are covered by jobs
3. Uses the remaining months to work out when work breaks are by grouping consecutive months together

## Screenshots

image

## Guidance to review

Would like feedback on:

- the readability of it - my biggest concern
- whether it makes sense

## Link to Trello card

## Things to check


app/services/work_history_with_breaks.rb
```diff
@@ -0,0 +1,80 @@
...
+      job_end_date = job.end_date.nil? ? Time.zone.now : job.end_date
+      months_in_job_period = month_range(start_date: job.start_date, end_date: job_end_date)
+
+      months_in_timeline -= months_in_job_period
+    end
+
+    months_in_timeline
+  end
+
+  def work_breaks(break_months_in_timeline)
```
C-gyorfi: There is some complexity here that makes it challenging to read this, but  I'm not sure how could you improve. I found challenging that some of these private methods have similar name e.g `work_break_months` and `work_breaks`.  So `work_break_months` returns a timeline, maybe you could call it `breaks_in_timeline`, but in the end, I'm not sure if it makes any difference

app/components/work_history_review_component.html.erb
```diff
@@ -11,7 +11,7 @@
         <% end %>
       <% end %>
     <% end %>
-  <% elsif history[:type] == :break %>
+  <% elsif history[:type] == :break_placeholder %>
```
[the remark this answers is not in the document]
chubberlisk: We'll need to differentiate between a break without a reason -> `break_placeholder`:

image

and a break with a reason -> `break`:

image

Currently we only show breaks, you can't enter a reason so `break_placeholder` makes more sense atm.

app/services/work_history_with_breaks.rb
chubberlisk: I agree, will have a think about renaming!

app/services/work_history_with_breaks.rb
```diff
@@ -0,0 +1,80 @@
...
+    end
+  end
+
+  def initialize(work_history)
+    @work_history = work_history.sort_by(&:start_date)
+    @work_history_with_breaks = []
+    @current_job = nil
+  end
+
+  def call
```
duncanjbrown: This method name could be something like `timeline` which might help make the included code a bit more readable

app/services/work_history_with_breaks.rb
```diff
@@ -0,0 +1,80 @@
...
+    return [] if @work_history.empty?
+
+    @work_history.each { |job| @work_history_with_breaks << job_entry(job) }
```
duncanjbrown: I think it would be more readable to keep this `@variable` local until we've finished with it, then assign? eg

```
work_history_with_breaks = []
...
@work_history_with_breaks = a_fully_populated(work_history_with_breaks)
```

app/services/work_history_with_breaks.rb
```diff
@@ -0,0 +1,80 @@
...
+
+    months_in_timeline = month_range(
+      start_date: @work_history.first.start_date,
+      end_date: Time.zone.now - 1.month,
+    )
+    break_months_in_timeline = work_break_months(months_in_timeline)
+    breaks = work_breaks(break_months_in_timeline)
```
duncanjbrown: How about method names more like this?

```
breaks = break_into_continuous_periods(
           remove_working_months(
               jobs, timeline_in_months
            )
          )
```

app/services/work_history_with_breaks.rb
```diff
@@ -0,0 +1,80 @@
...
+    @work_history_with_breaks = []
+    @current_job = nil
+  end
+
+  def call
+    return [] if @work_history.empty?
+
+    @work_history.each { |job| @work_history_with_breaks << job_entry(job) }
+
+    months_in_timeline = month_range(
```
duncanjbrown: Maybe `timeline_in_months`?

app/services/work_history_with_breaks.rb
chubberlisk: Agree with `remove_working_months` but not sure about `break_into_continuous_periods`. 🤔

app/services/work_history_with_breaks.rb
chubberlisk: Ended up removing the need for an instance variable but there was no need 😅

app/services/work_history_with_breaks.rb
chubberlisk: Did a bit of tinkering! Another :eyes:?

app/services/work_history_with_breaks.rb
chubberlisk: Ended up renaming to `break_entries` because that's what we return in the end. What do you think?

PR #1296, opened 2020-02-06
Update provider mailer with decision chaser

C-gyorfi:
## Context

We want to send chaser emails to providers when they are halfway through their RBD period.
#### Content design:
image

This PR extends the provider mailer with a chaser email
### Email preview
image

## Link to Trello card

## Things to check

### DIFF
--- a/app/mailers/provider_mailer.rb
+++ b/app/mailers/provider_mailer.rb
@@ -48,4 +48,27 @@ class ProviderMailer < ApplicationMailer
               to: provider_user.email_address,
               subject: t('provider_application_rejected_by_default.email.subject', candidate_name: @application.candidate_name))
   end
+
+  def chase_provider_decision(provider_user, application_choice)
+    @application =
+      Struct.new(
+        :candidate_name,
+        :provider_user_name,
+        :course_name_and_code,
+        :submitted_at,
+        :application_choice,
+        :rbd_date,
+      ).new(
+        application_choice.application_form.full_name,
+        provider_user.full_name,
+        application_choice.course.name_and_code,
+        application_choice.application_form.submitted_at.to_s(:govuk_date).strip,
+        application_choice,
+        application_choice.reject_by_default_at,
+    )
+
+    view_mail(GENERIC_NOTIFY_TEMPLATE,
+              to: provider_user.email_address,
+              subject: t('provider_application_waiting_for_decision.email.subject', candidate_name: @application.candidate_name))
+  end
 end
--- a/app/views/provider_mailer/chase_provider_decision.text.erb
+++ b/app/views/provider_mailer/chase_provider_decision.text.erb
@@ -1,0 +1,15 @@
+Dear <%= @application.provider_user_name || 'colleague' %>
+
+# Only 20 working days left to respond
+
+<%= @application.candidate_name %> submitted an application for <%= @application.course_name_and_code %> on <%= @application.submitted_at %>.
+
+Log in to Manage teacher training applications to respond within 20 working days:
+
+<%= provider_interface_application_choice_url(@application.application_choice) %>
+
+If you do not respond to the application by <%= @application.rbd_date.to_s(:govuk_date) %>, we’ll reject it on your behalf.
+
+# Give feedback or report a problem
+
+Contact us at [becomingateacher@digital.education.gov.uk](mailto:becomingateacher@digital.education.gov.uk).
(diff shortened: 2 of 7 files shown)


spec/mailers/provider_mailer_spec.rb
```diff
@@ -106,4 +106,52 @@
...
+    before do
+      @course_option = course_option_for_provider_code(provider_code: 'ABC')
+      @application_choice = create(:submitted_application_choice,
+                                   course_option: @course_option,
+                                   application_form:
+                                     create(
+                                       :completed_application_form,
+                                    ))
+      @provider_user = @application_choice.provider.provider_users.first
+      @mail = mailer.chase_provider_decision(@provider_user, @application_choice)
```
chubberlisk: Wondering if this can be refactored along with the other `describe` blocks here.

- `@course_option = course_option_for_provider_code(provider_code: 'ABC')` seems common
- I'm not sure we need to use `create` here for `application_choice` because we are passing it into the mailer
- Same for `application_form`, I think we can use `build_stubbed(:application_form, submitted_at: <some date>, first_name: 'Blah', last_name: 'Blah')` as that's the only things we care about from the application form as far as I can see

What do you think?

C-gyorfi: "do you even refactor bro?" 
Good shout, I moved stuff into let blocks to remove the duplication! 5bc0e19

app/mailers/provider_mailer.rb
```diff
@@ -48,4 +48,27 @@ def application_rejected_by_default(provider_user, application_choice)
...
+  def chase_provider_decision(provider_user, application_choice)
+    @application =
+      Struct.new(
+        :candidate_name,
+        :provider_user_name,
+        :course_name_and_code,
+        :submitted_at,
+        :application_choice,
+        :rbd_date,
+      ).new(
```
stevehook: Maybe extract the definition of the struct into a constant, e.g.

```
ChaseProviderDecisionApplication = Struct.new(
        :candidate_name,
        :provider_user_name,
        :course_name_and_code,
        :submitted_at,
        :application_choice,
        :rbd_date,
      )
```

(you can almost certainly think of a better name for it!)

spec/mailers/provider_mailer_spec.rb
```diff
@@ -5,6 +5,17 @@
...
+  let(:course_option) { course_option_for_provider_code(provider_code: 'ABC') }
+  let(:application_choice) do
+    build_stubbed(:submitted_application_choice,
+                  course_option: course_option,
+                  application_form:
+                    build_stubbed(
+                      :completed_application_form,
+                   ))
+  end
+  let(:provider_user) { application_choice.provider.provider_users.first }
```
stevehook: This goes against the conventions we've adopted elsewhere to avoid `let`. I think this example is pretty clear with the `let`s personally. @tijmenb pointed me to https://thoughtbot.com/blog/lets-not (you've probably already seen it!).

stevehook: One way to refactor it without using the `let`s would be to have a plain method that set up the test fixtures with args as needed for any variations in those fixtures.

C-gyorfi: I know, we have plenty of existing unit tests with let blocks, I thought this is more of an issue with the system tests, let me know what do you think @tijmenb

app/views/provider_mailer/chase_provider_decision.text.erb
```diff
@@ -0,0 +1,15 @@
+Dear <%= @application.provider_user_name || 'colleague' %>
+
+# Only 20 working days left to respond
```
benilovj: Similar to https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1290#discussion_r375850392, we probably don't want this number hard-coded

app/views/provider_mailer/chase_provider_decision.text.erb
```diff
@@ -0,0 +1,15 @@
...
+
+<%= @application.candidate_name %> submitted an application for <%= @application.course_name_and_code %> on <%= @application.submitted_at %>.
+
+Log in to Manage teacher training applications to respond within 20 working days:
```
benilovj: As above

app/views/provider_mailer/chase_provider_decision.text.erb
C-gyorfi: That makes sense, resolved in 82b6bef

app/views/provider_mailer/chase_provider_decision.text.erb
C-gyorfi: Resolved in 82b6bef

PR #1297, opened 2020-02-06
Add questionnaire for referees who choose to give a reference

davidgisbey:
## Context

We need a more effective way to ask referees for feedback on their experience of providing a reference. The current process of asking them to consent to a follow-up call hasn't been successful.

We think referees are more likely to participate in research if there is a satisfaction questionnaire built into the service. We need a way to ask referees for their feedback so we know if the guidance is useful and the process if working.

- Add a questionnaire for referees who have submitted a reference

image

image

## Guidance to review

The flow has not changed for users who refuse to give a reference

The flow for users who do give a reference should be. The feedback page, the questionnaire page,  then the thank you page. 

Users who have selected that they would like to be contacted should be told they will be contacted soon.

image

## Link to Trello card

https://trello.com/c/BebZdXWo/865-build-referee-questionnaire-into-reference-form-sent-to-referees

## Things to check


app/models/application_reference.rb
```diff
@@ -14,6 +14,8 @@ class ApplicationReference < ApplicationRecord
 
   audited associated_with: :application_form
 
+  serialize :questionnaire
```
stevehook: I don't think this is needed for JSONB columns in Rails these days. In fact it seems to cause activerecord to serialize YAML into the JSONB column (as a string) so it messes up the data structure.

With:

```
├─[ RECORD 3 ]────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ id                      │ 239                                                                                                                                         │
│ email_address           │ bob@example.com                                                                                                                             │
│ feedback                │ good person                                                                                                                                 │
│ application_form_id     │ 118                                                                                                                                         │
│ created_at              │ 2020-02-06 11:32:46.586137                                                                                                                  │
│ updated_at              │ 2020-02-06 15:46:43.887547                                                                                                                  │
│ name                    │ Bob                                                                                                                                         │
│ relationship            │ good mate                                                                                                                                   │
│ hashed_sign_in_token    │ f89f9d20d3c58a203a5ada74e199868b44cdcca47321007850674766b2459e49                                                                            │
│ consent_to_be_contacted │ t                                                                                                                                           │
│ feedback_status         │ feedback_provided                                                                                                                           │
│ questionnaire           │ "---\nexperience_rating: poor\nexperience_explanation: I was eating my dinner\nguidance_rating: ok\nguidance_explanation: Not much to say r…│
│                         │…eally\nsafe_to_work_with_children: 'false'\nsafe_to_work_with_children_explanation: Because he's a bit eccentric\nconsent_to_be_contacted: …│
│                         │…'true'\nconsent_to_be_contacted_details: feel free 012345678\n"                                                                             │

```

Without:
```
├─[ RECORD 4 ]────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ id                      │ 238                                                                                                                                         │
│ email_address           │ billy@example.com                                                                                                                           │
│ feedback                │ yeah alright                                                                                                                                │
│ application_form_id     │ 118                                                                                                                                         │
│ created_at              │ 2020-02-06 11:32:36.54928                                                                                                                   │
│ updated_at              │ 2020-02-06 15:55:53.339515                                                                                                                  │
│ name                    │ Billy                                                                                                                                       │
│ relationship            │ mate                                                                                                                                        │
│ hashed_sign_in_token    │ f3cfb4808e3ee07d640047165106ebe70aaeb3f517f1ff33601c4368e1648d89                                                                            │
│ consent_to_be_contacted │ t                                                                                                                                           │
│ feedback_status         │ feedback_provided                                                                                                                           │
│ questionnaire           │ {"guidance_rating": "very_poor", "experience_rating": "poor", "guidance_explanation": "Didn't have me glasses on so couldn't read it", "exp…│
│                         │…erience_explanation": "Was watching telly", "consent_to_be_contacted": "true", "safe_to_work_with_children": "true", "consent_to_be_contact…│
│                         │…ed_details": "go on then 0898 1234123", "safe_to_work_with_children_explanation": null}                                                     │

```

config/locales/referee.yml
```diff
@@ -1,7 +1,21 @@
...
+  questionnaire_form:
+    experience: Please rate your experience of giving a reference
+    why_that_rating: Tell us why you chose that rating
+    very_poor: Very poor
+    poor: Poor
+    ok: Ok
+    good: Good
+    very_good: Very Good
+    safe_to_work_with_children: If we asked whether a candidate was safe to work with children, would you feel able to answer?
+    is_safe_to_work_with_children: Yes, i could answer
```
chubberlisk:
```suggestion
    is_safe_to_work_with_children: Yes, I could answer
```

app/views/referee_interface/reference/confirmation.html.erb
```diff
@@ -11,25 +11,84 @@
...
-      <% end %>
+    <div class="govuk-panel govuk-panel--confirmation govuk-!-margin-bottom-8">
+        <h1 class="govuk-panel__title">
+          Your reference for <%= @reference.application_form.full_name %> has been submitted
+        </h1>
     </div>
-  </div>
+
+    <div class="govuk-grid-row">
+        <h2 class="govuk-heading-l">Tell us about your experience of giving a  reference</h2>
```
chubberlisk:
```suggestion
        <h2 class="govuk-heading-l">Tell us about your experience of giving a reference</h2>
```

config/locales/referee.yml
willmcb: Hey I think this should be a capital 'I' in:  "Yes, i could answer" 👍

app/services/return_questionnaire_results_hash.rb
```diff
@@ -0,0 +1,17 @@
+class ReturnQuestionnaireResultsHash
```
duncanjbrown: As I understand it this data's only going to be used for printing the results of the survey, hence the simple JSON blob. Would it be more readable if we used the actual questions we ask as the keys (i18n values may change in the future), and for the values, munge together the answer and the explanation? eg

```
{
  "Please rate your experience of giving a reference" => "Good. I liked it",
  "Please rate our guidance" => "Not bad."
}
```

duncanjbrown:
Looks good! question about the hash we save

config/locales/referee.yml
chubberlisk: Never mind just saw Will's comment (https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1297#discussion_r375934727) 👀

app/services/return_questionnaire_results_hash.rb
stevehook: It may not be a requirement today but if, in future, somebody ever asked for say the % of respondents that said the experience was good or better then this could make it a bit more tricky (due to those i18n keys being volatile).

app/models/application_reference.rb
davidgisbey: Thanks for this! I had no idea.

app/services/return_questionnaire_results_hash.rb
davidgisbey: This is interesting. Perhaps we could have a quick chat after stand up tomorrow!

davidgisbey:
Thanks for all the feedback all!

app/models/application_reference.rb
stevehook: Nor did I really until I saw some funny looking JSON in the database!

app/controllers/referee_interface/reference_controller.rb
```diff
@@ -28,16 +30,16 @@ def submit_feedback
       end
     end
 
-    def confirmation; end
-
-    def confirm_consent
-      consent_to_be_contacted = params.dig(:application_reference, :consent_to_be_contacted)
+    def submit_questionnaire
+      questionnaire_hash = ReturnQuestionnaireResultsHash.call(params: params['application_reference'])
```
chubberlisk: Had a thought about this and realised I think we should use our form object pattern here instead of a service. Not sure if you've come across it yet?

If not, have a look at `app/models/candidate_interface/training_with_a_disability_form.rb` and how it's used in `app/controllers/candidate_interface/training_with_a_disability_controller.rb` as an example. However, look at others as well as they all vary a bit to fit the needs of the form.

What do you think?

davidgisbey: Just looked. It's definitely a well-established pattern!  Will and I are going to have a refactor now.

app/services/return_questionnaire_results_hash.rb
duncanjbrown: I wasn't there — let's talk today!

app/controllers/referee_interface/reference_controller.rb
```diff
@@ -4,7 +4,9 @@ class ReferenceController < ActionController::Base
     before_action :add_identity_to_log
     before_action :check_referee_has_valid_token
     before_action :set_token_param
-    before_action :show_finished_page_if_feedback_provided, except: %i[confirmation confirm_consent]
+    before_action :show_finished_page_if_feedback_provided, except: %i[submit_questionnaire submit_feedback confirmation]
```
duncanjbrown: `only:` might be safer here as we don't need to remember to rule out any future actions we might happen to add

spec/system/referee_interface/referee_can_submit_a_reference_for_candidate_spec.rb
```diff
@@ -96,14 +100,20 @@ def and_i_see_the_list_of_the_courses_the_candidate_applied_to
...
-    choose t('contact_form.consent_to_be_contacted')
+    choose t('questionnaire_form.consent_to_be_contacted')
+    fill_in 'Please let us know when you’re available', with: 'anytime 012345 678900'
   end
 
 private
 
   def get_reference_feedback_url
     matches = current_email.body.match(/(http:\/\/localhost:3000\/reference\?token=[\w-]{20})/)
+    @token = matches.captures.first.split('=').last
```
duncanjbrown: Rails/Rack can help here https://stackoverflow.com/a/9398264

app/controllers/referee_interface/reference_controller.rb
davidgisbey: 🤔 I would have thought that the default behaviour post submitting their reference, is that they are redirected to the finish page. The submit_feedback page should not be there though.

spec/system/referee_interface/referee_can_submit_a_reference_for_candidate_spec.rb
davidgisbey: Thanks

PR #1298, opened 2020-02-06
Revert "Add chase_provider_decision_at column to application_chices"

C-gyorfi:
Reverts DFE-Digital/apply-for-postgraduate-teacher-training#1293


tijmenb:
Sorry I didn't press submit on my comment before, but we can't only revert a migration. We need another migration to remove the column from QA's database.

PR #1303, opened 2020-02-06
Put Google Tag Manager behind a feature flag

benilovj:
## Context

It's useful to be able to disable Google Tag Manager in certain environments.

## Changes proposed in this pull request

- put the Google Tag Manager code behind a feature flag
- add spec coverage for candidates visiting the cookie page
- extend the Markdown rendering mechanism to be able to pre-process the source with ERB first, if `.erb` is in the name
- don't display cookie information about Google Tag Manager if the feature flag is switched off

## Things to check

- [x] This code doesn't rely on migrations in the same Pull Request
- [x] If this code includes a migration adding or changing columns, it also backfills existing records for consistency
- [x] API release notes have been updated if necessary
- [x] New environment variables have been [added to the Azure config](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training#azure-hosting-devops-pipeline)

PR #1304, opened 2020-02-07
Remove Google Tag Manager code and user info

benilovj:
## Context

Alternative to #1303.

Remove GTM code and documentation on the cookies pages

## Things to check


fofr:
As we're only using essential cookies, the old cookie banner should probably go too.

app/views/layouts/base.html.erb
```diff
@@ -41,8 +33,6 @@
       document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');
     </script>
```
tijmenb: Did we miss this tag manager bit?

benilovj: Oops - yes. https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1313

PR #1305, opened 2020-02-07
Show reference status on the candidate dashboard

stevehook:
## Context

Candidates cannot currently see what has happened to their reference request. We need to indicate whether a reference request has been sent and what the outcome is.

- [x] Extend the `RefereesReviewComponent` to output a status value row as per design.
- [x] Handle expired reference requests
- [x] Update system spec?

## Guidance to review

- The card doesn't specify what to display for the `not_requested_yet` status - I've just gone with _Not requested yet_ for now (to be consistent with the Support interface). This status is visible right after you add a reference.
- The only slightly complicated part is going to be the `response_overdue` 'psuedo-status' we need for unanswered references requests (those that have exceeded the time limit).
- I'm calculating whether a given reference is overdue on not using `ApplicationReference#created_at` in line with other code but I think this is wrong - see https://trello.com/c/5AQezpqR/966-work-out-whether-we-need-to-refine-the-way-we-calculate-overdue-reference-requests - I'm not expecting to solve that in this PR

image

## Link to Trello card

https://trello.com/c/LdKYcBrb/829-ref-candidate-dashboard-does-not-show-reference-status

## Things to check

### DIFF
--- a/app/lib/time_limit_config.rb
+++ b/app/lib/time_limit_config.rb
@@ -1,6 +1,4 @@
 class TimeLimitConfig
-  DAYS_BEFORE_WE_ASK_FOR_A_NEW_REFEREE = 10
-
   Rule = Struct.new(:from_date, :to_date, :limit)
 
   RULES = {
--- a/app/lib/time_limit_config.rb
+++ b/app/lib/time_limit_config.rb
@@ -1,4 +1,6 @@
 class TimeLimitConfig
+  DAYS_BEFORE_WE_ASK_FOR_A_NEW_REFEREE = 10
+
   Rule = Struct.new(:from_date, :to_date, :limit)
 
   RULES = {
--- a/app/models/application_reference.rb
+++ b/app/models/application_reference.rb
@@ -66,9 +66,8 @@ class ApplicationReference < ApplicationRecord
     replace_referee_at < Time.zone.now
   end
 
-  # TODO: Revisit this to factor out the time limit to a constant and use a
-  # dedicated `request_sent_at` as the start point.
+  # TODO: Revisit this to use a dedicated `request_sent_at` as the start point?
   def response_overdue?
-    feedback_requested? && created_at < 10.business_days.ago
+    feedback_requested? && created_at < TimeLimitConfig::DAYS_BEFORE_WE_ASK_FOR_A_NEW_REFEREE.business_days.ago
   end
 end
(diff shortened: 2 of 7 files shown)


app/models/application_reference.rb
```diff
@@ -46,4 +46,10 @@ def self.find_by_unhashed_token(unhashed_token)
     hashed_token = Devise.token_generator.digest(ApplicationReference, :hashed_sign_in_token, unhashed_token)
     find_by(hashed_sign_in_token: hashed_token)
   end
+
+  # TODO: Revisit this to factor out the time limit to a constant and use a
+  # dedicated `request_sent_at` as the start point.
+  def response_overdue?
+    feedback_requested? && created_at < 10.business_days.ago
```
C-gyorfi: Is this something we should include in the `TimeLimitConfig` rules?

stevehook: Yes I think this is what Tijmen did in another open PR. I will change it here to be consistent - though I think we might need to revisit the logic.

stevehook:
> I'm not sure references that haven't been sent are supposed to show the status. Can we check with design?

Yes double-checked with @fofr and you are not supposed to see status of unsubmitted references. Will change it...

stevehook:
> > I'm not sure references that haven't been sent are supposed to show the status. Can we check with design?
> 
> Yes double-checked with @fofr and you are not supposed to see status of unsubmitted references. Will change it...

Have hidden the status for un-submitted applications. The value for 'Not requested status' is still there partly because we've got tests that create references that are not requested but associated with an un-submitted application.

app/lib/time_limit_config.rb
```diff
@@ -1,4 +1,6 @@
 class TimeLimitConfig
+  DAYS_BEFORE_WE_ASK_FOR_A_NEW_REFEREE = 10
```
tijmenb: This can be removed now.

app/models/application_reference.rb
```diff
@@ -58,4 +58,8 @@ def replace_referee_at
 
     TimeLimitCalculator.new(rule: :replace_referee_by, effective_date: requested_at).call.second
   end
+
+  def response_overdue?
```
tijmenb: We can use https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/app/models/application_reference.rb#L62-L68 now.

stevehook:
> 👍 needs an update after the latest changes to master.

thanks @tijmenb , all done now

PR #1310, opened 2020-02-10
Drop ChaseProviderDecisionAt From ApplicationChoices table

C-gyorfi:
## Context

As flagged by @tijmenb on https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1298 reverting a migration won't drop the column from the DB. This migration is to drop the redundant column. 

Drop ChaseProviderDecisionAt From ApplicationChoices table from DB as we won't need it

## Things to check


C-gyorfi:
@tijmenb I would need to switch off robocop to be able to do this. What would you suggest?

tijmenb:
The column adding was never run on staging/production, so this would fail on those envs:

```
StandardError: An error has occurred, this and all later migrations canceled:

PG::UndefinedColumn: ERROR:  column "chase_provider_decision_at" of relation "application_choices" does not exist
```

Let's close this PR, since it won't affect QA that much.

PR #1311, opened 2020-02-10
Add authorisation service to enforce make_offer user/course association

mnacos:
### Context

As the service scales and more providers join, we'll be onboarding increasing numbers of provider users. These users will be responsible for managing applications for their associated providers, but we are already researching scenaria which require different levels of access, either on a per user or per provider level.

There is no explicit authorisation framework within the app yet. In most cases, authorisation is either minimal/implicit (checks in controller actions) or missing. For example, provider users can make offers on courses unrelated to any of their providers through the API. Going forward, we'll need a consistent way to express and enforce such rules for the applications workflow.

State transitions within the app are already organised/performed through services. These services currently lack authorisation logic, operating outside any notions of user/actor/client. A weak association of such calls with a user actually already exists, in the form of attribution/auditing, but this is not well suited for authorisation. To identify a good authorisation pattern around services, a number of different options were considered:

- make authorisation part of state transitions within workflow gem (pros: we already handle state transition exceptions / cons: current implementation is intentionally minimal; and how do we pass the user context?)
- pass a user/agent/client context to the service and make it the service's responsibility to check authorisation (pros: easy to follow, works well for both UI and API contexts / cons: adding more responsibility to services)
- pass a user/agent/client context to the service but use meta-programming so that authorisation happens behind the scenes with no changes to service logic (pros: programmers follow existing service guidelines without worrying about authorisation / cons: if we are passing a user context, we might as well be explicit about using it)
- expect controller to authorise actions outside the service and only call services if authorisation steps succeed (pros: no changes, this is the current approach / cons: inconsistent, authorisation logic for same service will be scattered/replicated in multiple controllers)

We have decided to make authorisation the responsibility of the service (second option), but pass a pre-configured authorisation context rather a user, which should work well regardless of the controller context (UI vs API) and provide a consistent authorisation API.

Using an authorisation context rather than a user helps us deal with the fact that api keys are currently not associated with a list of providers but with a single provider, whereas provider users can be associated with multiple providers. Both api keys and provider users are allowed to make offers. The providers association is key for validating this specific use case, i.e. that the calling party has permission to offer a specific course.

Add `ProviderAuthorisation` service, which is configured with an actor context, representing a provider user, support user or api key. `MakeAnOffer` service requires a new keyword `actor:` and uses a `ProviderAuthorisation` object to determine if this user has permission to offer this application choice.

In the calling code:

```ruby
result = MakeAnOffer.new(actor: current_provider_user, application_choice: a).save
```

Within the service:

```ruby
def initialize(actor: actor, ....)
  @auth = ProviderAuthorisation.new(actor: actor)
  ...
end

def save
  @auth.assert_can_make_offer! application_choice: application_choice
  ...
end
```

The default interface to `ProviderAuthorisation` is expected to raise errors if actions are not allowed, because in most situations the user should only be presented with actions they are allowed to perform in the UI. If a service expects authorisation to fail as part of normal operation, this is a special case and errors can be handled in the service code.

For the very specific case of making offers, `#can_make_offer?` checks the list of providers associated with the user (api tokens are currently only associated with one provider each). Passing a `SupportUser` to the `ProviderAuthorisation` context will skip `providers` checks.

### Guidance to review

Run the specs, read the code. It is hard to make an offer for an unrelated course within the UI. Not so hard through the API.

### Link to Trello card

[1623 - Prevent provider users from making offers on courses none of their associated providers run](https://trello.com/c/aYMUGX4C)

### DIFF
--- a/app/services/make_an_offer.rb
+++ b/app/services/make_an_offer.rb
@@ -1,6 +1,7 @@
 class MakeAnOffer
   attr_accessor :standard_conditions
   attr_accessor :further_conditions0, :further_conditions1, :further_conditions2, :further_conditions3
+  attr_accessor :auth
 
   include ActiveModel::Validations
 
@@ -12,12 +13,14 @@ class MakeAnOffer
   validate :validate_further_conditions
 
   def initialize(
+    actor:,
     application_choice:,
     offer_conditions: nil,
     standard_conditions: nil,
     further_conditions: {},
     course_data: nil
   )
+    @auth = ProviderAuthorisation.new(actor: actor)
     @application_choice = application_choice
     @offer_conditions = offer_conditions
     @standard_conditions = standard_conditions
@@ -27,6 +30,8 @@ class MakeAnOffer
 
   def save
     return unless valid?
+
+    @auth.assert_can_make_offer! application_choice: application_choice
 
     ApplicationStateChange.new(application_choice).make_offer!
     application_choice.offered_course_option = offered_course_option
--- a/spec/services/make_an_offer_spec.rb
+++ b/spec/services/make_an_offer_spec.rb
@@ -1,14 +1,16 @@
 require 'rails_helper'
 
 RSpec.describe MakeAnOffer do
+  let(:user) { SupportUser.new }
+
   describe '#save' do
     it 'sets the offered_at date' do
       application_choice = create(:application_choice, status: :awaiting_provider_decision)
 
-      MakeAnOffer.new(application_choice: application_choice)
+      MakeAnOffer.new(actor: user, application_choice: application_choice)
 
       Timecop.freeze do
-        MakeAnOffer.new(application_choice: application_choice).save
+        MakeAnOffer.new(actor: user, application_choice: application_choice).save
 
         expect(application_choice.offered_at).to eq(Time.zone.now)
       end
@@ -18,6 +20,7 @@ RSpec.describe MakeAnOffer do
   describe 'validation' do
     it 'accepts nil conditions' do
       decision = MakeAnOffer.new(
+        actor: user,
         application_choice: build_stubbed(:application_choice, status: :awaiting_provider_decision),
         offer_conditions: nil,
       )
@@ -27,6 +30,7 @@ RSpec.describe MakeAnOffer do
 
     it 'limits the number of conditions to 20' do
       decision = MakeAnOffer.new(
+        actor: user,
         application_choice: build_stubbed(:application_choice, status: :awaiting_provider_decision),
         offer_conditions: Array.new(21) { 'a condition' },
       )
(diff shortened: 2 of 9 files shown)


app/services/authorisation.rb
```diff
@@ -0,0 +1,21 @@
...
+    @actor = actor
+  end
+
+  def can_make_offer?(application_choice:)
+    @actor.is_a?(SupportUser) || @actor.providers.include?(application_choice.course.provider)
+  end
+
+  class AssertError < StandardError; end
+
+  instance_methods.select { |m| m.match PERMISSION_METHOD_REGEXP }.each do |method|
```
duncanjbrown: This is very nifty! 🏎 

It's a bit hard to follow, though. Could we make do with a single `assert_can_make_offer?` method for now, and refactor when we have some duplication?

app/services/make_an_offer.rb
```diff
@@ -12,12 +12,14 @@ class MakeAnOffer
   validate :validate_further_conditions
 
   def initialize(
+    auth:,
```
duncanjbrown: :+1: This feels like the right place for this argument

app/services/authorisation.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def initialize(actor:)
+    @actor = actor
+  end
+
+  def can_make_offer?(application_choice:)
+    @actor.is_a?(SupportUser) || @actor.providers.include?(application_choice.course.provider)
+  end
+
+  class AssertError < StandardError; end
```
duncanjbrown: Maybe this could be a bit more specific — `NotAuthorisedError`?

app/services/authorisation.rb
```diff
@@ -0,0 +1,21 @@
+class Authorisation
```
duncanjbrown: It makes sense to have this in the top namespace as it's not shared, but the name is quite generic for something that only handles auth for provider clients. Could we include `Provider` in the name?

spec/services/make_an_offer_spec.rb
```diff
@@ -1,14 +1,16 @@
 require 'rails_helper'
 
 RSpec.describe MakeAnOffer do
+  let(:auth) { Authorisation.new(actor: SupportUser.new) }
```
duncanjbrown: The policy of "SupportUsers can do everything" is accidentally part of this test. What would you think of something like

`double(assert_can_make_offer!: true)`

?

app/services/make_an_offer.rb
tijmenb: How about passing in an `actor` into this class? Inside the `MakeAnOffer` constructor we can still instantiate the `Authorisation` object.

- This will hide the concept of authorisation into the service
- It will make it easier to instantiate the class
- It would make it easier to pass extra arguments to Authorisation if we need it

mnacos: Ok

spec/services/make_an_offer_spec.rb
mnacos: Yes, something along these lines would be good.

PR #1312, opened 2020-02-10
Add ability to add another job when there is a break

chubberlisk:
## Context

Currently, we can see when a break has occurred in a candidate's work history - https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1167, https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1295. What's missing now is allowing the candidate to enter a reason and being able to add another job in place of the break.

This PR adds the `Add another job` link to a break. When a candidate clicks on the link they are taken to the add job with the start and end date filled in already.

## Screenshots

image

image

## Guidance to review

Is using get params to way to go? Can't think of another way of doing it.

## Link to Trello card

## Things to check


davidgisbey:
All looked good 👍

PR #1315, opened 2020-02-10
Correct reference chasing dates

tijmenb:
##  Context

We currently look at the `created_at` for references when deciding when to sent a chaser. We should look at the time was sent, otherwise we’re going to shorten the allowed period for referees.

This PR depends on https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1316 having been **deployed**.

- Use `requested_at`
- Use business days

## Link to Trello card

https://trello.com/c/5AQezpqR

## Things to check

### DIFF
--- a/app/services/get_referees_to_chase.rb
+++ b/app/services/get_referees_to_chase.rb
@@ -2,7 +2,7 @@ class GetRefereesToChase
   def self.call
     ApplicationReference
       .feedback_requested
-      .where(['requested_at < ?', Time.zone.now - 5.days])
+      .where(['requested_at < ?', 5.business_days.ago])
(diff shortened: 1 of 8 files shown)


app/services/get_referees_to_chase.rb
```diff
@@ -2,7 +2,7 @@ class GetRefereesToChase
   def self.call
     ApplicationReference
       .feedback_requested
-      .where(['created_at < ?', Time.zone.now - 5.days])
+      .where(['requested_at < ?', Time.zone.now - 5.business.days])
```
benilovj: Should this use `TimeLimitConfig`?

tijmenb: Done

PR #1321, opened 2020-02-10
Rename other qualifications section

fofr:
Indicate that this section is for "academic qualifications and other relevant" rather than "other relevant academic and non-academic".

We hope this will nudge users to put in more of their academic qualifications and aligns better with our existing guidance.

https://trello.com/c/j1otdP43/454-change-content-for-other-relevant-qualifications-tactical

### DIFF
--- a/app/components/candidate_interface/other_qualifications_review_component.html.erb
+++ b/app/components/candidate_interface/other_qualifications_review_component.html.erb
@@ -13,5 +13,5 @@
 <% end %>
 
 <% if @qualifications.empty? %>
-  <p class="govuk-body">No other qualifications entered.</p>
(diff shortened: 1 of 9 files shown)


emmajhf:
Hi @fofr, I think this looks good. 

The only bit I had a question about was 'including all your GCSEs, A levels and vocational, practical and creative qualifications'. Should we say 'or equivalent' somewhere to acknowledge that not everyone will have GCSEs and A levels? I also wasn't sure about 'practical and creative qualifications'.

How about: 'including all your GCSEs and A levels (or equivalents), and any other qualifications where you showed skills that might help you as a teacher'?

PR #1323, opened 2020-02-10
Error and catch to stop zero applications creation

willmcb:
Adds Error and catch to stop generation of test users with zero courses per application.

## Context

You were able to generate courses with zero courses per application - which is not a state that would be consistent with the functioning of the production application. 

Adds a check to the module `TestApplications.create_applications(states:, courses_to_apply_to: nil)` that raises an error if the `states:` is an empty array `[]` 

Adds new unit and feature tests to ensure correct behaviour.

Updates API related documentation.

## Guidance to review

Make a request to `/experimental/test-data/generate?count=1&courses_per_application=0` and you should receive an error response.

## Link to Trello card

[1627-Investigate how an application on Sandbox was submitted without any course choices](https://trello.com/c/x7395Ph4)

## Things to check

### DIFF
--- a/app/lib/test_applications.rb
+++ b/app/lib/test_applications.rb
@@ -1,5 +1,6 @@
 module TestApplications
   class NotEnoughCoursesError < RuntimeError; end
+  class ZeroCoursesPerApplicationError < RuntimeError; end
 
   def self.generate_for_provider(provider:, courses_per_application:, count:)
     1.upto(count).flat_map do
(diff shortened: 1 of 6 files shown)


app/lib/test_applications.rb
```diff
@@ -11,6 +12,8 @@ def self.generate_for_provider(provider:, courses_per_application:, count:)
   end
 
   def self.create_application(states:, courses_to_apply_to: nil)
+    raise ZeroCoursesPerApplicationError.new('You can\'t have zero courses per application') unless states.any?
```
stevehook: You could use `blank?` here if you wanted to handle both `nil` and `[]` values with the same error. I don't think this is an issue because we are not passing `nil` to this method from anywhere as far as I can see.

willmcb: Hey Steve thanks for the pointer - I think you're right in this case it shouldn't matter, but it is good to know that this could have be an option in future in similar circumstances 👍

PR #1327, opened 2020-02-11
Add sending of a confirmation email when reference submitted

chubberlisk:
## Context

We want to send a referee a confirmation email when they have submitted a reference. This came about follow a thread on Slack - https://ukgovernmentdfe.slack.com/archives/CP18YJXPY/p1580735397016000.

This PR:

- adds the confirmation email to the `RefereeMailer` with a preview
- adds it to the Support Docs as an email to the "Awaiting references" state
- adds sending of the confirmation email when a reference is submitted

## Screenshots

image

image

## Guidance to review

Would like feedback on:

- whether this email has been placed correctly on the Support Docs flow
- whether the audit comment is clear

## Link to Trello card

https://trello.com/c/OORDnap3/879-%E2%9C%85-ref-confirmation-email-to-referees

## Things to check


app/services/send_reference_confirmation_email.rb
```diff
@@ -0,0 +1,9 @@
+class SendReferenceConfirmationEmail
+  def self.call(application_form:, reference:)
+    RefereeMailer.reference_confirmation_email(application_form, reference).deliver
+
+    audit_comment = I18n.t('reference_confirmation_email.audit_comment', referee_name: reference.name, referee_email: reference.email_address)
+    application_comment = SupportInterface::ApplicationCommentForm.new(comment: audit_comment)
+    application_comment.save(application_form)
```
tijmenb: What would you think of using this approach?

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/99f0f2fd1a01c42457b1b13a2309fe26825ff224/app/services/send_chase_email_to_provider.rb#L11-L16

- Use the object directly so we don't have to use `SupportInterface` code in the top-level
- Perhaps skip i18n for this, might make it easier to read

chubberlisk:
Forgot to feature flag this! Going to do this.

spec/system/referee_interface/referee_can_submit_a_reference_for_candidate_spec.rb
```diff
@@ -42,9 +44,11 @@ def when_the_candidate_submits_the_application
   def then_i_receive_an_email_with_a_magic_link
     open_email('terri@example.com')
 
-    reference_feedback_url = get_reference_feedback_url
+    matches = current_email.body.match(/(http:\/\/localhost:3000\/reference\?token=[\w-]{20})/)
+    @token = Rack::Utils.parse_query(URI(matches.captures.first).query)['token']
```
tvararu: What's this for? 🤔

chubberlisk: Twas part of refactoring `get_reference_feedback_url` method. It's used in:

```ruby
def then_i_see_the_confirmation_page
  expect(page).to have_current_path(referee_interface_confirmation_path(token: @token))
end
```

app/services/send_reference_confirmation_email.rb
chubberlisk: Done!

chubberlisk:
Updated with a feature flag now.

app/controllers/referee_interface/reference_controller.rb
```diff
@@ -23,6 +23,8 @@ def submit_feedback
       )
 
       if @reference_form.save
+        SendReferenceConfirmationEmail.call(application_form: @application, reference: reference)
```
tvararu: The feature flag could go here as well; currently, it’s not clear that if it’s off, this service does a noop

chubberlisk: That's a fair point! Will move it here.

PR #1328, opened 2020-02-11
Refactor chase provider decision flow and TimeLimitCalculator 

C-gyorfi:
## Context

As discussed in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1290 the time limit the provider get chased at is calculated backwards from `rejectet_by_default_at` date. The previous implementation was tied to the 20 days rule. This PR is intended to remove the hardcoded 20 days and use the `TimeLimitConfig` where possible. The time limit is calculated backwards from the RBD date, which was not clear from the previous implementation. 

This PR refactors:
- The flow for `chase provider when an application is waiting for decision`
- The `TimeLimitCalculator` to return a hash and improve readability

## Guidance to review

Does this make sense? 

## Things to check

### DIFF
--- a/app/lib/time_limit_config.rb
+++ b/app/lib/time_limit_config.rb
@@ -11,7 +11,7 @@ class TimeLimitConfig
     edit_by: [
       Rule.new(nil, nil, 5),
     ],
-    chase_provider_by: [
+    chase_provider_before_rbd: [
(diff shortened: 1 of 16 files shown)


stevehook:
:+1:

PR #1332, opened 2020-02-11
Fix support interface application detail page

stevehook:
## Context

This was broken by a merge issue - the `TimeLimitCalculator` interface changed in one branch as two new references to it were introduced in another branch. 

This was found on `qa`. Need to merge this fix before we deploy to `production`. It wasn't covered by system specs so we need to address that as well as fix the issue.

- Change references to `TimeLimitCalculator` to use new interface in `ApplicationReference#chase_referee_at` and `replace_referee_at`.
- Update system spec that opens this page to include a submitted application to reproduce the issue.

## Guidance to review

- is the usage of `TimeLimitCalculator` correct?

## Link to Trello card

n/a

## Things to check

### DIFF
--- a/spec/system/support_interface/see_individual_application_spec.rb
+++ b/spec/system/support_interface/see_individual_application_spec.rb
@@ -31,6 +31,7 @@ RSpec.feature 'See an application' do
 
   def and_there_are_applications_in_the_system
     @completed_application = create(:completed_application_form, references_count: 2, with_gces: true)
+    SubmitApplication.new(@completed_application).call
     @unsubmitted_application = create(:application_form)
     @application_with_reference = create(:completed_application_form, references_count: 2, with_gces: true)
   end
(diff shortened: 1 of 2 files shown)


spec/system/support_interface/see_individual_application_spec.rb
```diff
@@ -31,6 +31,7 @@ def given_i_am_a_support_user
 
   def and_there_are_applications_in_the_system
     @completed_application = create(:completed_application_form, references_count: 2, with_gces: true)
+    SubmitApplication.new(@completed_application).call
```
C-gyorfi: Is this intended to test `chase_referee_at` and `replace_referee_at` ? 👍 for that!

stevehook: It does test them both indirectly. They only get called for submitted applications so the original system specs were not picking the issue up.

We could add some unit tests for these methods.

C-gyorfi: Yes, I agree, but good it is covered now!

PR #1337, opened 2020-02-11
Add section in "Add a job" page to ask the candidate if they want to add another job

chubberlisk:
## Context

When little Ting began working on work history breaks, it seemed like there was no light at the end of the tunnel. But then the Two Pauls came down from Heaven and showed Ting the light.

> It's going to be okay my child. It's doesn't have to be complicated. **_- The Two Pauls_**

See https://trello.com/c/jIWipybM/732-calculate-work-gaps#comment-5e4290ca629e7313c54cd3f1 for the actual story.

This is the first step to actually providing a reason for a work history break.

This PR adds a section of the "Add a job" page which has radio buttons to ask the candidate if they want to add another job. This is hidden under the `work_breaks` feature flag.

- If they choose "Yes, I want to add another job", then they are redirected to the "Add a job" page. Else, they are taken to the review page.
- If the candidate is on the "Add a job" page, then the "Yes, I want to add another job" is preselected. Else if, the candidate is on the "Edit a job" page, then the "No, I’ve completed my work history" is preselected.

## Screenshots

image

## Guidance to review

Would like help on:

- if there is a nicer way to uniquely identify the radio buttons than using the `id`
- if there is a nicer way to prefill the `add_another_job` attribute to yes or no. Thought there might be a way to do it with formbuilder but couldn't find anything in the documentation

## Link to Trello card

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/work_history/edit_controller.rb
+++ b/app/controllers/candidate_interface/work_history/edit_controller.rb
@@ -12,9 +12,10 @@ module CandidateInterface
                                   start_date_year: start_date.year,
                                   end_date_month: end_date.month,
                                   end_date_year: end_date.year,
+                                  add_another_job: true,
                                 )
                               else
-                                WorkExperienceForm.new
+                                WorkExperienceForm.new(add_another_job: true)
                               end
     end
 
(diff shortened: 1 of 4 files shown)


chubberlisk:
Paul H suggested that a content tweak to the radio buttons is probably needed as the last two questions and answers look similar. Will catch up with either Emmas about this tomorrow.

PR #1342, opened 2020-02-12
Refactor out a system spec specifically for new work history breaks

chubberlisk:
## Context

When slowly adding the feature to enable a candidate to enter a reason for each work history break, we've been adding to the existing `spec/system/candidate_interface/candidate_entering_work_history_spec.rb` system spec. This spec has become quite long and potentially confusing as it mixes new features with the soon to be old way of us collecting a work history break (a single text area).

This PR extracts the new parts of the work history breaks into a new system spec to make things clearer. 

This means:

- `spec/system/candidate_interface/candidate_entering_work_history_spec.rb` contains the single text box way of entering work history breaks so without the `work_breaks` feature flag activated
- `spec/system/candidate_interface/candidate_entering_work_history_breaks_spec.rb` contains the new way of entering work history breaks / what we have so far and with the `work_breaks` feature flag activated

## Guidance to review

Is this clearer?

## Link to Trello card

## Things to check

### DIFF
--- a/spec/system/candidate_interface/candidate_enter_work_history_breaks_spec.rb
+++ b/spec/system/candidate_interface/candidate_enter_work_history_breaks_spec.rb
@@ -1,0 +1,101 @@
+require 'rails_helper'
+
+RSpec.feature 'Entering reasons for their work history breaks' do
+  include CandidateHelper
+
+  scenario 'Candidate enters a reason for a work break' do
+    FeatureFlag.activate('work_breaks')
+
+    given_i_am_signed_in
+    and_i_visit_the_site
+
+    when_i_click_on_work_history
+    and_i_choose_i_have_work_for_more_than_5_years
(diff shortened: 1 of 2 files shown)


spec/system/candidate_interface/candidate_enter_work_history_breaks_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+      fill_in 'Month', with: '3'
+      fill_in 'Year', with: '2019'
+    end
+
+    within('[data-qa="end-date"]') do
+      fill_in 'Month', with: '8'
+      fill_in 'Year', with: '2019'
+    end
+
+    fill_in t('details.label', scope: scope), with: 'I taught others to be pros at MS Paint.'
```
tvararu: 👌

PR #1343, opened 2020-02-12
Update content on application is being considered email

davidgisbey:
## Context

This is one of the emails that need content tweaks.

- Tweak the content
- Add tests

## Guidance to review

Check the email preview candidate#application_under_consideration and compare with
https://docs.google.com/document/d/1VH_nxuLTiVCkmKhC4eaBpFC61Hd5_Ol1tKwOul-2uUs/edit#

## Link to Trello card

https://trello.com/c/05nmr4eO/962-checking-emails-in-build

## Things to check

### DIFF
--- a/app/mailers/candidate_mailer.rb
+++ b/app/mailers/candidate_mailer.rb
@@ -15,7 +15,6 @@ class CandidateMailer < ApplicationMailer
       candidate: application_form.candidate,
       candidate_name: application_form.first_name,
       choice_count: application_form.application_choices.count,
-      rbd_date: application_form.application_choices.first.reject_by_default_at,
       rbd_days: application_form.application_choices.first.reject_by_default_days,
     )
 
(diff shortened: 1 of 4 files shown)


app/mailers/candidate_mailer.rb
```diff
@@ -15,8 +15,7 @@ def application_under_consideration(application_form)
       candidate: application_form.candidate,
       candidate_name: application_form.first_name,
       choice_count: application_form.application_choices.count,
-      rbd_date: application_form.application_choices.first.reject_by_default_at,
-      rbd_days: application_form.application_choices.first.reject_by_default_days,
+      rbd_days: TimeLimitConfig.limits_for(:reject_by_default).first.limit,
```
tijmenb: The reason we use the database `reject_by_default_days` here is that it might change at some point, so we need to display the number of days that is true for that specific application.

davidgisbey: Okay great have reverted it 👍

PR #1349, opened 2020-02-12
Add interstitial page to warn users that they need to replace references

C-gyorfi:
## Context

We want to allow candidates to provide an alternative referee if my original referees have not responded *or* have declined. 

This PR adds the interstitial page that warns candidates when they need to replace references.
The interstitial page should differ if one or two replacement referees needed, using the `AdditionalRefereesStartComponent`. 
### After:
image

image

## Guidance to review

For manual testing: 
- Create an application form with referees needed replacement 
- Log in as a candidate and visit the root page (Make sure the feature flag is on) 
- Visit the application page
You should be redirected to the interstitial page

## Link to Trello card

https://trello.com/c/oMo8ZCap/761-enable-candidates-to-add-new-referee

## Things to check

### DIFF
--- a/app/components/candidate_interface/additional_referees_start_component.html.erb
+++ b/app/components/candidate_interface/additional_referees_start_component.html.erb
@@ -1,0 +1,26 @@
+<h1 class="govuk-heading-xl">
+  <%= title %>
+</h1>
+
+<p class="govuk-body">
+  <% if reference_status.number_of_references_that_currently_need_replacing == 1 %>
+    <%= reason_for_replacement(references_that_need_replacement.first) %>.
+  <% else %>
+    Your referees have not given us a reference:
+    <ul class="govuk-list govuk-list--bullet">
+      <% references_that_need_replacement.each do |ref| %>
+        <li class="govuk-body"><%= reason_for_replacement(ref) %></li>
+      <% end %>
+    </ul>
+  <% end %>
+</p>
+
+<p class="govuk-body">We can’t send your application to your teacher training providers without 2 complete references.</p>
+
+<% if reference_status.number_of_references_that_currently_need_replacing == 1 %>
+  <p> <%= govuk_button_link_to 'Add a new referee', candidate_interface_new_additional_referee_path %></p>
+  <p class="govuk-body"><%= govuk_link_to 'Continue without adding a new referee', candidate_interface_application_complete_path %></p>
+<% else %>
+   <%= govuk_button_link_to 'Add new referees', candidate_interface_new_additional_referee_path %>
+  <p class="govuk-body"><%= govuk_link_to 'Continue without adding new referees', candidate_interface_application_complete_path %></p>
+<% end %>
--- a/app/components/candidate_interface/additional_referees_start_component.rb
+++ b/app/components/candidate_interface/additional_referees_start_component.rb
@@ -1,0 +1,42 @@
+module CandidateInterface
+  class AdditionalRefereesStartComponent < ActionView::Component::Base
+    include ViewHelper
+
+    validates :application_form, presence: true
+
+    def initialize(application_form:)
+      @application_form = application_form
+    end
+
+  private
+
+    def reference_status
+      @reference_status ||= ReferenceStatus.new(application_form)
+    end
+
+    def title
+      if reference_status.number_of_references_that_currently_need_replacing == 1
+        'You need to add a new referee'
+      else
+        'You need to add 2 new referees'
+      end
+    end
+
+    def reason_for_replacement(referee)
+      case referee.feedback_status
+      when 'email_bounced'
+        "Our email requesting a reference didn’t reach #{referee.name}"
+      when 'feedback_refused'
+        "#{referee.name} said they won’t give a reference"
+      else
+        "#{referee.name} did not respond to our request"
+      end
+    end
+
+    def references_that_need_replacement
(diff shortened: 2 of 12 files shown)


app/components/candidate_interface/additional_referees_start_component.html.erb
```diff
@@ -0,0 +1,24 @@
...
+    <%= reason_for_replacement(references_that_need_replacement.first) %>.
+  <% else %>
+    Your referees have not given us a reference:
+    <% references_that_need_replacement.each do |ref| %>
+      <li class="govuk-body"><%= reason_for_replacement(ref) %></li>
+    <% end %>
+  <% end %>
+</p>
+
+<p class="govuk-body">We’ can't send your application to your teacher training providers without 2 complete references.</p>
```
tijmenb:
```suggestion
<p class="govuk-body">We can’t send your application to your teacher training providers without 2 complete references.</p>
```

app/controllers/candidate_interface/application_form_controller.rb
```diff
@@ -1,5 +1,6 @@
 module CandidateInterface
   class ApplicationFormController < CandidateInterfaceController
+    before_action :redirect_to_additional_references_path, only: %i[show review]
```
tijmenb: As discussed, we'll try to move this to the signin controller auth method so we don't need to do redirects here.

app/components/candidate_interface/additional_referees_start_component.rb
```diff
@@ -0,0 +1,42 @@
+module CandidateInterface
+  class AdditionalRefereesStartComponent < ActionView::Component::Base
```
tijmenb: 👍 really excellent this component.

app/components/candidate_interface/additional_referees_start_component.rb
```diff
@@ -0,0 +1,42 @@
...
+        'You need to add a new referee'
+      else
+        'You need to add 2 new referees'
+      end
+    end
+
+    def reason_for_replacement(referee)
+      case referee.feedback_status
+      when 'email_bounced'
+        "Our email requesting a reference didn't reach #{referee.name}"
```
tijmenb:
```suggestion
        "Our email requesting a reference didn’t reach #{referee.name}"
```

app/components/candidate_interface/additional_referees_start_component.rb
```diff
@@ -0,0 +1,42 @@
...
+      when 'feedback_refused'
+        "#{referee.name} said they won't give a reference"
```
tijmenb:
```suggestion
        "#{referee.name} said they won’t give a reference"
```

spec/components/candidate_interface/additional_referees_start_component_spec.rb
```diff
@@ -0,0 +1,71 @@
+require 'rails_helper'
+
+RSpec.describe CandidateInterface::AdditionalRefereesStartComponent do
+  let(:application_form) { create(:completed_application_form, references_count: 2, with_gces: true) }
+  let(:first_referee) { application_form.application_references.first }
+  let(:second_referee) { application_form.application_references.second }
+
+  context 'when one referee request failed' do
+    it 'has a page content that requests one new referee' do
+      first_referee.update!(feedback_status: 'email_bounced')
```
tijmenb: You could get rid of the `let`s by creating the required references in the `it` block itself - perhaps cleaner?

```suggestion
      create(:reference, :bounced, application_form: application_form) 
```

spec/system/candidate_interface/candidate_needs_to_provide_two_new_referees_spec.rb
```diff
@@ -4,14 +4,15 @@
...
     FeatureFlag.activate('show_new_referee_needed')
 
     given_i_am_signed_in_as_a_candidate
     and_i_have_submitted_my_application
     and_both_of_my_referees_havent_responded_within_a_reasonable_timeframe
 
-    when_i_visit_the_application_dashboard
-    then_i_see_that_i_need_new_references
+    when_i_visit_the_start_page
+    then_i_see_the_interstitial_page_to_add_new_referees
```
tijmenb: Can we keep the existing tests as well? We still need to make sure we show the warning on the dashboard.

spec/components/candidate_interface/additional_referees_start_component_spec.rb
C-gyorfi: I don't think it makes it cleaner, but I had a go: 2636b37

C-gyorfi: I think in this cases those `let` blocks are acceptable, I need to assign the references to a var anyway because they are referenced in the assertions.  I let you comment on this refactoring, but I would prefer to keep these tests as they are 😬

app/controllers/candidate_interface/application_form_controller.rb
C-gyorfi: I'm a bit confused about how to end-to-end test these without creating too much mess 
currently we are using this in the system tests: 
```
  def given_i_am_signed_in_as_a_candidate
    @candidate = create(:candidate)
    login_as(@candidate)
  end
```
and in this case, we never hit the `sign in controller`, maybe we could pair on this tomorrow if you have some time!

PR #1352, opened 2020-02-12
[Spike] Write an end to end smoke test using Cypress and Mailslurp

tvararu:
## Context

This is a very simple test which uses [Cypress](https://www.cypress.io) and [Mailslurp](https://www.mailslurp.com) to sign up as a new candidate on the live QA environment.

I spiked this because I might raise this point in an upcoming incident review. 🐒 

Add two JS dependencies, some auto-generated Cypress boilerplate.

## Guidance to review

`cypress/integration/candidate.spec.js` is the relevant file.

Cypress is not the only end to end testing framework we can use; I used it because I am familiar with it and it has a good workflow for prototyping something quickly. I was mainly looking to find a good solution to having to open a mailbox in order to complete the sign up process.

Screen Recording 2020-02-12 at 17 53 11

## Link to Trello card

😬 

## Things to check


tvararu:
Closing for now! Can be revisited at a later point, e.g. if we have another incident that slips past all of our failsafes but would have been caught by something that is automatable in this way. Notes:

- Does not need to be implemented with Cypress, I used it because it's very fast for prototyping. There's Ruby bindings for it or we can use standard Selenium with our existing system specs.
- That being said, I really am a big fan of Cypress. It's still my choice for e2e testing.
- Mailslurp is OK, the free tier won't do for our needs. We'd need at least the 39 Euro/month basic tier for 1000 emails per month. There's other competitors as well, but I think they're basically the same.

PR #1359, opened 2020-02-13
Consolidate and iterate performance reporting

benilovj:
## Context

The current performance dashboard makes it hard to see what's happening post-submission (how many successful and unsuccessful candidates there are) at a glance.

Iterate the service performance dashboard:

- consolidate candidate and application form reporting into just candidate reporting, because it's effectively the same thing until Apply 2 is added
- add more focus to how many applications are in pre- and post-submission.
- make it clearer how many applications are in important end-states (offers accepted or unsuccessful)

Before:

image

After:

image

## Guidance to review

- is the visual design (layout and colours) OK?
- for the top-level metrics, is it clear what the number is from the labels?

## Things to check

### DIFF
--- a/app/views/support_interface/performance/index.html.erb
+++ b/app/views/support_interface/performance/index.html.erb
@@ -23,28 +23,48 @@
 </div>
 
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-one-quarter">
-    <div class="app-card">
-      <span class="app-card__count"><%= @statistics[:never_signed_in] %></span>
-      <span>never signed in</span>
+  <div class="govuk-grid-column-one-half">
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-one-third">
+        <div class="app-card">
+          <span class="app-card__secondary-count"><%= @statistics[:never_signed_in] %></span>
+          <span>never signed in</span>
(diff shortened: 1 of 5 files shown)


duncanjbrown:
Great! I like very much seeing each layer as the sum of the layer below, which holds for the first two rows (183 = 54 +129). This isn't the case in the bottom-right.

Should we drop the "Dashboard" from the title? It seems a bit redundant

duncanjbrown:
What @tijmenb said 😂

fofr:
Maybe not orange for submitted. 🤔

benilovj:
Thanks all for the feedback – see the updated screenshot:

image

Changes:

- I've pared down the colours, they can be iterated by and actual designer on a later PR
- I've updated the 3rd row of metrics so it adds up to the 2nd row, and the fonts are smaller
- I've removed the word 'dashboard' from the title

app/views/support_interface/performance/index.html.erb
```diff
@@ -1,50 +1,78 @@
-<% content_for :title, 'Service performance dashboard' %>
+<% content_for :title, 'Service performance' %>
```
fofr: :+1:

PR #1360, opened 2020-02-13
Add ability to enter a reason for a break

chubberlisk:
## Context

Currently, we are only able to show when breaks occurs in a candidate's work history.

This PR adds the ability to enter a reason for a work history break and shows the break on the review work history page.

## Screenshots

image

image

## Guidance to review

Review commit by commit as some refactoring has been done.

Should there be a redirect from the `WorkHistory::BreakController` to check that the feature flag is on? Or it's not necessary?

## Link to Trello card

## Things to check

### DIFF
--- a/app/models/application_work_history_break.rb
+++ b/app/models/application_work_history_break.rb
@@ -2,4 +2,8 @@ class ApplicationWorkHistoryBreak < ApplicationRecord
   belongs_to :application_form, touch: true
 
   audited associated_with: :application_form
+
+  def length
+    (end_date.year * 12 + end_date.month) - (start_date.year * 12 + start_date.month) - 1
+  end
 end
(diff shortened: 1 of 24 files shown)


chubberlisk:
I've just noticed that:

1. Enter a job with a break between May 2019 to December 2019
2. Enter a reason for the break between December 2019 and February 2020
3. Delete job

Then the stored break doesn't show, it should.

This will be fixed in the next PR which will add the ability to delete a break.

app/models/application_work_history_break.rb
```diff
@@ -2,4 +2,8 @@ class ApplicationWorkHistoryBreak < ApplicationRecord
   belongs_to :application_form, touch: true
 
   audited associated_with: :application_form
+
+  def length
+    (end_date.year * 12 + end_date.month) - (start_date.year * 12 + start_date.month) - 1
```
tvararu: Nice

PR #1361, opened 2020-02-13
Edit cookies page

fofr:
We aren't collecting information about how they use the site. We can remove this statement.


tvararu:
What about the GA cookie?

fofr:
Thanks for pointing this out @aliuk2012

PR #1362, opened 2020-02-13
Add sandbox mode to the feature flags API

duncanjbrown:
Sandbox mode isn't really a feature flag but it does affect the behaviour of environments where it's turned on.

## Context

We forgot to turn off sandbox mode in staging after our testing party,

Advertise it as a feature flag over the feature_flags API (even though it isn't one)

## Things to check

### DIFF
--- a/app/controllers/integrations/feature_flags_controller.rb
+++ b/app/controllers/integrations/feature_flags_controller.rb
@@ -10,6 +10,11 @@ module Integrations
         features
       end
 
+      feature_flags['sandbox_mode'] = {
+        name: 'Sandbox mode',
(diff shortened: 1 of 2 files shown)


app/controllers/integrations/feature_flags_controller.rb
```diff
@@ -10,6 +10,11 @@ def index
         features
       end
 
+      feature_flags['sandbox_mode'] = {
+        name: 'Sandbox mode',
+        active: HostingEnvironment.sandbox_mode?,
+      }
```
tijmenb: This will make the screen report the feature flag as wrong. Can we put it in another key and display it separately?

PR #1363, opened 2020-02-13
Add migration for `providers.region_code`

stevehook:
## Context

We want to be able to group Providers by region to make large lists easier to navigate.

- Just a migration for now to add column for `Provider#region_code`

## Guidance to review

- There is still some debate about how we will actually populate this new attribute. The Find API seems to only have values for about a quarter of providers. Is there another way to group providers that we do have data for?
- I don't think there is any need to backfill this data yet and it would have to be done partly from the Find API via the existing syncing mechanism, part manually (not sure who by).

## Link to Trello card

https://trello.com/c/luWz8uHE/975-currently-available-providers-page

## Things to check


db/schema.rb
```diff
@@ -161,7 +161,7 @@
     t.string "magic_link_token"
     t.datetime "magic_link_token_sent_at"
     t.boolean "hide_in_reporting", default: false, null: false
-    t.integer "course_from_find_id"
+    t.bigint "course_from_find_id"
```
tvararu: This shouldn't break anything right?

db/schema.rb
```diff
@@ -233,8 +233,8 @@
   create_table "provider_users_providers", force: :cascade do |t|
     t.bigint "provider_id", null: false
     t.bigint "provider_user_id", null: false
-    t.datetime "created_at", default: -> { "now()" }, null: false
-    t.datetime "updated_at", default: -> { "now()" }, null: false
+    t.datetime "created_at", default: -> { "CURRENT_TIMESTAMP" }, null: false
+    t.datetime "updated_at", default: -> { "CURRENT_TIMESTAMP" }, null: false
```
tvararu: Interesting. I have no reason to believe this won't work, but peculiar that it came up as part of this?

stevehook: this keeps flip-flopping as different people update the schema. I believe its a Postgres version thing. It's harmless but very irritating. I think I'm going to try upgrading Postgres. I'm on version 10 which is a bit behind some (though it's ahead of production).

tvararu: 👍 Don't bother. I figured it must be something like that; it's not worth discussing or upheaving your setup, we can roll with it!

db/schema.rb
duncanjbrown: Consensus is not, though we don't completely understand it. See also https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1263#discussion_r374214067

db/schema.rb
stevehook: rollback, checkout `schema.rb`, db:reset and re-migrate seems to 'fix it'

PR #1366, opened 2020-02-13
Document Apply 2 in the process diagram

tijmenb:
## Context

It would be useful to be able to explain Apply 2 using this diagram.

image

## Guidance to review

Should it point to "Form started but unsubmitted" instead? @benilovj 

## Things to check

### DIFF
--- a/config/locales/application_states.yml
+++ b/config/locales/application_states.yml
@@ -363,3 +363,10 @@ en:
       name: Enrol
       by: provider
       description: ''
+
+    ended_without_success-start_apply_2:
+      name: Start in Apply 2 (not built)
+      by: candidate
+      description: This feature isn't built yet, but in the future candidates
+        will be able to go into a second phase called "Apply 2", where they can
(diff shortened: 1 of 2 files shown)


benilovj:
> Should it point to "Form started but unsubmitted" instead?

Yes I believe so, because we're probably gonna pre-populate the Apply 2 form...

config/locales/application_states.yml
```diff
@@ -363,3 +363,10 @@ en:
       name: Enrol
       by: provider
       description: ''
+
+    ended_without_success-start_apply_2:
+      name: Start in Apply 2 (not built)
+      by: candidate
+      description: This feature isn't built yet, but in the future candidates
+        will be able to go into a second phase called "Apply 2", where they can
+        apply to 1 course at the time.
```
benilovj:
```suggestion
        apply to 1 course at a time.
```

tijmenb:
Updated!

PR #1372, opened 2020-02-14
Sync Provider#region_code from Find API

stevehook:
## Context

We want to be able to group Providers by region to make large lists easier to navigate.

- This PR adds region code to the existing Provider data synchronisation mechanism so that it will be pulled from the Find API by the scheduled background job
- Migration is already deployed. https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1363
- Fix an crash in the synchronisation that doesn't handle nil values for certain Site address fields.

## Guidance to review

- Currently Find only have region_code values for about 25% of their Providers. So this data won't be useful to us until they/we have fixed that. Does it make sense to add this now?

## Link to Trello card

https://trello.com/c/luWz8uHE/975-currently-available-providers-page

## Things to check

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -8,6 +8,18 @@ class Provider < ApplicationRecord
   has_many :provider_users, through: :provider_users_providers
   has_many :provider_agreements
 
+  enum region_code: {
+    east_midlands: 'east_midlands',
+    eastern: 'eastern',
+    london: 'london',
+    north_east: 'north_east',
+    north_west: 'north_west',
+    south_east: 'south_east',
+    south_west: 'south_west',
+    west_midlands: 'west_midlands',
+    yorkshire_and_the_humber: 'yorkshire_and_the_humber',
+  }
+
   def name_and_code
     "#{name} (#{code})"
   end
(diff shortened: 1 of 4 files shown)


app/models/provider.rb
```diff
@@ -8,6 +8,18 @@ class Provider < ApplicationRecord
   has_many :provider_users, through: :provider_users_providers
   has_many :provider_agreements
 
+  enum region_code: {
```
benilovj: For safety, is it worth matching what's in Publish? https://github.com/DFE-Digital/teacher-training-api/blob/master/app/models/concerns/region_code.rb

app/services/sync_provider_from_find.rb
```diff
@@ -37,6 +37,12 @@ def self.update_provider_name(provider, provider_name)
     provider.save!
   end
 
+  def self.update_provider(provider, find_provider)
+    provider.region_code = find_provider.region_code.strip if find_provider.region_code
```
tijmenb: Will this crash when calling `strip` on `nil`?

app/models/provider.rb
stevehook: Good point, I think we prefer to use strings in the DB but we can match the same keys.

app/services/sync_provider_from_find.rb
stevehook: I don't think it will due to the `if find_provider.region_code`

davidgisbey:
Looks good to me 👌

PR #1377, opened 2020-02-14
Disable NestedGroups cop in spec

davidgisbey:
## Context

I personally find this cop really annoying in spec files. Thoughts?

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -39,6 +39,10 @@ RSpec/MultipleExpectations:
 RSpec/ContextWording:
   Enabled: false
 
+  RSpec/NestedGroups:
+    Exclude:
+      - 'spec/**/*'
+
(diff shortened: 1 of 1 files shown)


duncanjbrown:
For what it's worth, I like this cop. I think there can be good reasons for disabling it on an ad hoc basis, eg when we have temporary extra context like a feature flag.

tijmenb:
Can you point me to a place where we violate this cop?

davidgisbey:
Ah sorry, I altered it this morning after it was agreed that providers would have to give feedback on rejection. It was something along the lines of:

`RSpec.describe CandidateMailer, type: :mailer 
   describe 'rejection emails' do
      context 'All application choices have been rejected email' do
         context 'when the provider gives feedback
         end
       end
   end 
 end`

PR #1386, opened 2020-02-17
Periodically refresh the service performance screen

tijmenb:
## Context

The service performance dashboard is displayed on our TV, but isn't periodically updated so the stats never change.

This adds a 60 second refresh when in "large screen mode". This mode also hides the page furniture and make the screen full-width, so it's nicer on the TV.

image

## Guidance to review

Forgive the inlining of style and javascript - I think this is actually the cleanest and most discoverable way of doing it. 

## Link to Trello card

🚌

## Things to check

### DIFF
--- a/app/views/support_interface/performance/index.html.erb
+++ b/app/views/support_interface/performance/index.html.erb
@@ -102,4 +102,6 @@
   header, footer, .govuk-phase-banner { display: none; }
   .govuk-width-container { max-width: 99%; }
 </style>
+
+<meta http-equiv="refresh" content="60">
 <% end %>
--- a/app/views/support_interface/performance/index.html.erb
+++ b/app/views/support_interface/performance/index.html.erb
@@ -1,5 +1,9 @@
 <% content_for :title, 'Service performance' %>
 
+<p class='govuk-body'>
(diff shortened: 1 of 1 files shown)


app/views/support_interface/performance/index.html.erb
```diff
@@ -92,3 +96,16 @@
...
   </div>
 </div>
+
+<% if params[:screen] %>
+<style media="screen">
+  header, footer, .govuk-phase-banner { display: none; }
+  .govuk-width-container { max-width: 99%; }
+</style>
+
+<script type="text/javascript">
```
fofr: Could use `<meta http-equiv="refresh" content="30">` instead

app/views/support_interface/performance/index.html.erb
```diff
@@ -92,3 +96,12 @@
...
   </div>
 </div>
+
+<% if params[:screen] %>
+<style media="screen">
+  header, footer, .govuk-phase-banner { display: none; }
+  .govuk-width-container { max-width: 99%; }
+</style>
+
+<meta http-equiv="refresh" content="60">
```
tvararu: TIL

PR #1392, opened 2020-02-17
Send an email to the candidate when a reference comes in

tijmenb:
## Context

When referees provide a reference we should tell the candidate.

Add an email and send it when the reference comes in. This is done in a slightly different way from the referee confirmation, which happens inside the controller. We'll refactor that later.

## Guidance to review

I haven't added unit tests to the candidatemailer, because it seems sufficiently testedby the integration test. What do people think?

## Link to Trello card

https://trello.com/c/RP9Njbix

## Things to check

### DIFF
--- a/app/mailers/candidate_mailer.rb
+++ b/app/mailers/candidate_mailer.rb
@@ -85,6 +85,15 @@ class CandidateMailer < ApplicationMailer
     new_offer(application_choice, :decisions_pending)
   end
 
+  def reference_received(reference)
+    @reference = reference
+    @candidate_name = reference.application_form.first_name
+
+    view_mail(GENERIC_NOTIFY_TEMPLATE,
+              to: reference.application_form.candidate.email_address,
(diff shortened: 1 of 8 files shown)


app/mailers/candidate_mailer.rb
```diff
@@ -85,6 +85,15 @@ def new_offer_decisions_pending(application_choice)
     new_offer(application_choice, :decisions_pending)
   end
 
+  def reference_received(reference)
```
C-gyorfi: Is there any test around this?

tijmenb: Yeah, in the integration test: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1392/files#diff-228a1e3080215c6cca421e46eeb296aa

Do you think it's enough?

C-gyorfi: Oh yeah, its a pretty simple mail!

PR #1393, opened 2020-02-17
Copy feature flags from production when we setup_local_dev_data

duncanjbrown:
## Context

Feature flags are always changing in production and we should keep our dev envs up to date

By default, `setup_local_dev_data` syncs local feature flags with prod.

## Guidance to review

Does this make sense? Is prod the right place to sync from?

### DIFF
--- a/lib/tasks/local_dev.rake
+++ b/lib/tasks/local_dev.rake
@@ -1,5 +1,5 @@
 desc 'Set up your local development environment with data from Find'
-task setup_local_dev_data: :environment do
+task setup_local_dev_data: %i[environment copy_feature_flags_from_production] do
   puts 'Syncing data from Find...'
   ENV['DEV_PROVIDERS_TO_SYNC'].split(',').each do |code|
     SyncProviderFromFind.call(provider_code: code, sync_courses: true)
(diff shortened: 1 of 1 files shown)


lib/tasks/local_dev.rake
```diff
@@ -26,3 +24,18 @@ task setup_local_dev_data: :environment do
...
+task copy_feature_flags_from_production: :environment do
+  flags = JSON.parse(HTTP.get('https://www.apply-for-teacher-training.education.gov.uk/integrations/feature-flags'))
+
+  flags_to_set = flags['feature_flags'].reduce([]) do |acc, (flag_name, flag_data)|
+    acc.push(flag_name) if flag_data['active']
+    acc
+  end
+
+  FeatureFlag::FEATURES.each { |f| FeatureFlag.deactivate(f) }
+  (FeatureFlag::FEATURES & flags_to_set).each { |f| FeatureFlag.activate(f) }
```
tijmenb: Perhaps to avoid the `flags_to_set`:

```suggestion
  FeatureFlag::FEATURES.each { |f| FeatureFlag.activate(f) if flags['features'][f]['active'] }
```

duncanjbrown: better yet!

fofr:
Love this 💯

PR #1399, opened 2020-02-17
Add candidate chaser to candidate mailer and GetApplicationChoicesWaitingForCandidateDecision service

C-gyorfi:
## Context

Currently, a support user must warn the candidate when they only have 5 working days to make a decision. We want to send automated emails to the candidate when an offer is close to decline by default date.

This is the first slice of this service, the PR adds:
- The time limit to the `time_limit_config`
- A new service `GetApplicationChoicesWaitingForCandidateDecision` that gets all course choices that are exceeded the time limit before `declined_by_default` date and groups these by `application form id`
- A mailer to send emails to candidates when there are one or multiple offers that needs a decision 

### After:
image
image

## Guidance to review

There are email previews, I fed them with OpenStruct to avoid using `create` and polluting the db 

## Link to Trello card

https://trello.com/c/g0dJmjJc/842-email-%F0%9F%98%AC-5-day-reminder-to-make-a-decision-after-all-responses-are-in-to-candidate

## Things to check

### DIFF
--- a/app/mailers/candidate_mailer.rb
+++ b/app/mailers/candidate_mailer.rb
@@ -84,6 +84,14 @@ class CandidateMailer < ApplicationMailer
     email_for_candidate(reference.application_form)
   end
 
+  def chase_candidate_decision(application_form)
+    @dbd_date = application_form.application_choices.first.decline_by_default_at.to_s(:govuk_date).strip
+    @application_choices = application_form.application_choices.select { |choice| choice if choice.offer? }
+
+    subject_pluralisation = @application_choices.count > 1 ? 'plural' : 'singular'
+    email_for_candidate(application_form, subject: I18n.t!("chase_candidate_decision_email.subject_#{subject_pluralisation}"))
+  end
+
 private
 
   def application_rejected(application_choice, template_name)
--- a/app/mailers/candidate_mailer.rb
+++ b/app/mailers/candidate_mailer.rb
@@ -86,7 +86,7 @@ class CandidateMailer < ApplicationMailer
 
   def chase_candidate_decision(application_form)
     @dbd_date = application_form.application_choices.first.decline_by_default_at.to_s(:govuk_date).strip
-    @application_choices = application_form.application_choices.select { |choice| choice if choice.offer? }
+    @application_choices = application_form.application_choices.offer
 
     subject_pluralisation = @application_choices.count > 1 ? 'plural' : 'singular'
     email_for_candidate(application_form, subject: I18n.t!("chase_candidate_decision_email.subject_#{subject_pluralisation}"))
--- a/app/mailers/previews/candidate_mailer_preview.rb
+++ b/app/mailers/previews/candidate_mailer_preview.rb
@@ -41,6 +41,44 @@ class CandidateMailerPreview < ActionMailer::Preview
     CandidateMailer.new_referee_request(reference, reason: :email_bounced)
   end
 
+  def chase_candidate_decision_with_one_offer
+    application_choice = OpenStruct.new(
+      provider: OpenStruct.new(name: 'Provider Name'),
+      decline_by_default_at: Time.zone.now,
+      application_form: application_form,
+      course: OpenStruct.new(name: 'Course Name'),
+    )
+
+    CandidateMailer.chase_candidate_decision(application_form_with_chooices([application_choice]))
+  end
+
+  def chase_candidate_decision_with_multiple_offers
+    application_choice = OpenStruct.new(
+      provider: OpenStruct.new(name: 'Wen University'),
+      decline_by_default_at: Time.zone.now,
+      application_form: application_form,
+      course: OpenStruct.new(name: 'Math with Dance'),
+    )
+
+    application_choice_two = OpenStruct.new(
+      provider: OpenStruct.new(name: 'Ting University'),
+      decline_by_default_at: Time.zone.now,
+      application_form: application_form,
+      course: OpenStruct.new(name: 'Dance'),
+    )
+
+    application_choice_three = OpenStruct.new(
+      provider: OpenStruct.new(name: 'Wang College'),
+      decline_by_default_at: Time.zone.now,
+      application_form: application_form,
+      course: OpenStruct.new(name: 'Moar Dance'),
+    )
+
+    CandidateMailer.chase_candidate_decision(application_form_with_chooices(
+                                               [application_choice, application_choice_two, application_choice_three],
+))
+  end
+
   def new_offer_single_offer
     course_option = FactoryBot.build_stubbed(:course_option)
     application_choice = application_form.application_choices.build(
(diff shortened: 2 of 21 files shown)


app/mailers/previews/candidate_mailer_preview.rb
```diff
@@ -43,6 +43,48 @@ def new_referee_request_with_email_bounced
...
+    application_form = OpenStruct.new(
+      candidate: OpenStruct.new(email_address: 'candidate@email.com'),
+      first_name: 'Ting',
+    )
+    application_choice = OpenStruct.new(
+      provider: OpenStruct.new(name: 'Provider Name'),
+      decline_by_default_at: Time.zone.now,
+      application_form: application_form,
+      course: OpenStruct.new(name: 'Course Name'),
+    )
```
tijmenb: Is there a reason not to use FactoryBot like the other previews?

app/services/get_application_choices_waiting_for_candidate_decision.rb
```diff
@@ -0,0 +1,14 @@
+class GetApplicationChoicesWaitingForCandidateDecision
+  def self.call
+    ApplicationChoice
```
tijmenb: As discussed, something like this to get the application forms that need chasing:

```
ApplicationForm.where(
  id: ApplicationChoice
    .where(status: :offer)
    .where('decline_by_default_at < ?', chase_candidate_time_limit)
    .select(:application_form_id)
).where.not(id: ChaserSent.candidate_decision_request.select(:chased_id))
```

app/mailers/previews/candidate_mailer_preview.rb
C-gyorfi: Yes, `build` does not allow me to create application choices with courses and providers unless I use `create`, which would pollute the db

stevehook: A lot of the previews are using `create`. I think the best solution here is to have a transaction rollback mechanism like the one Jake tried to implement a couple of weeks ago. I'd like to try and get that working.

app/mailers/candidate_mailer.rb
```diff
@@ -91,6 +91,19 @@ def new_offer_decisions_pending(application_choice)
     new_offer(application_choice, :decisions_pending)
   end
 
+  def chase_candidate_decision(application_choices)
+    @application_form = application_choices.first.application_form
+    @dbd_date = application_choices.first.decline_by_default_at.to_s(:govuk_date).strip
+    @application_choices = application_choices
+
+    email_address = application_choices.first.application_form.candidate.email_address
+    subject_pluralisation = application_choices.count > 1 ? 'plural' : 'singular'
```
chubberlisk: There's a Rails method, `pluralize`, which can help to avoid the need for this and two subjects in the localisation.

See https://apidock.com/rails/ActionView/Helpers/TextHelper/pluralize.

C-gyorfi: Yes I'm aware of that, but since its a subject referenced in a translation I'm not sure how much if it would be any better?
``` I18n.t!("chase_candidate_decision_email.subject", application: pluralize(@application_choices.count, 'application'))``` I'll try to replace to this 👍

Comment (C-gyorfi) [reply app/mailers/candidate_mailer.rb]: I won't as discussed with @chubberlisk

File: app/mailers/candidate_mailer.rb [hunk 4]
```diff
@@ -84,6 +84,14 @@ def reference_received(reference)
     email_for_candidate(reference.application_form)
   end
 
+  def chase_candidate_decision(application_form)
+    @dbd_date = application_form.application_choices.first.decline_by_default_at.to_s(:govuk_date).strip
+    @application_choices = application_form.application_choices.select { |choice| choice if choice.offer? }
```
tijmenb: This can be:

```suggestion
    @application_choices = application_form.application_choices.select(&:offer?)
```
or use the magic scope added by the enum:

```suggestion
    @application_choices = application_form.application_choices.offer
```

app/services/feature_flag.rb
```diff
@@ -16,6 +16,7 @@ class FeatureFlag
     automated_referee_replacement
     candidate_rejected_by_provider_email
     notify_candidate_of_new_reference
+    automated_candidate_chaser
```
tijmenb:
```suggestion
    automated_decline_by_default_candidate_chaser
```

app/services/get_application_choices_waiting_for_candidate_decision.rb
```diff
@@ -0,0 +1,14 @@
...
+    ApplicationForm.where(
+      id: ApplicationChoice
+        .where(status: :offer)
+        .where('decline_by_default_at < ?', chase_candidate_time_limit)
+        .select(:application_form_id),
+        ).where.not(id: ChaserSent.candidate_decision_request.select(:chased_id))
+  end
+
+  def self.chase_candidate_time_limit
+    TimeLimitCalculator.new(rule: :chase_candidate_before_dbd, effective_date: Time.zone.now).call[:time_in_future]
```
tijmenb: Not for this PR - we should make this clearer!

app/services/send_chase_email_to_candidate.rb
```diff
@@ -0,0 +1,11 @@
+class SendChaseEmailToCandidate
+  def self.call(application_form:)
+    CandidateMailer.chase_candidate_decision(application_form).deliver
+    ChaserSent.create!(chased: application_form, chaser_type: :candidate_decision_request)
+
+    audit_comment =
+      "Chase emails have been sent to candidate (#{application_form.candidate.email_address}) because " +
```
tijmenb:
```suggestion
      "Chase email has been sent to candidate because " +
```

app/services/get_application_choices_waiting_for_candidate_decision.rb
```diff
@@ -0,0 +1,14 @@
+class GetApplicationChoicesWaitingForCandidateDecision
```
tijmenb:
```suggestion
class GetApplicationFormsForDeclineByDefaultReminder
```

spec/system/candidate_interface/candidate_receives_email_when_an_application_declined_by_default_date_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    FeatureFlag.activate('automated_candidate_chaser')
+  end
+
+  def when_i_have_an_offer_waiting_for_my_decision
+    @application_form = create(:completed_application_form)
+    create(:application_choice, status: :offer, application_form: @application_form, decline_by_default_at: Time.zone.now + 10.days)
+  end
+
+  def and_the_time_limit_before_decline_by_default_date_has_been_exceeded
+    time_limit_in_buiness_days = TimeLimitConfig.limits_for(:chase_candidate_before_dbd).first.limit
```
tijmenb:
```suggestion
    time_limit_in_business_days = TimeLimitConfig.limits_for(:chase_candidate_before_dbd).first.limit
```

PR #1403, opened 2020-02-18
Launch expired sign-in link feature

stevehook:
## Context

This is a follow up to PR https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1233

We want to make it simpler for users that attempt to use an expired token to log in. Since we know who they are we can automatically send an email with a new token.

The original implementation passes a plain user id to an expired page that renders the users email address. This PR ensures that the plain user id replaced by a token and the email is no longer rendered.

- [x] Replace the candidate `id` param sent to the expired page (`CandidateInterface::SignInController#expired`) with an encrypted id (`u`).
- [x] Remove the `email_address` hidden field from the expired page and replace with the encrypted id (`u`).

Note that these features remain feature flagged and is off on production at this time.

## Guidance to review

- Are the tests sufficient? - I didn't think there was any need to add anything new because the existing system specs cover all the code I touched.
- Does this feature introduce any holes in the authentication process?

## Link to Trello card

https://trello.com/c/LkdIGwhl/887-launch-expired-sign-in-link-feature-%F0%9F%8F%88

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/sign_in_controller.rb
+++ b/app/controllers/candidate_interface/sign_in_controller.rb
@@ -33,19 +33,18 @@ module CandidateInterface
         rescue ActiveSupport::MessageEncryptor::InvalidMessage
           redirect_to candidate_interface_sign_in_path && return
         end
-      else
-        @candidate = Candidate.for_email candidate_params[:email_address]
       end
 
-      if @candidate&.persisted?
+      @candidate ||= Candidate.for_email candidate_params[:email_address]
+
+      if @candidate.persisted?
         MagicLinkSignIn.call(candidate: @candidate)
         add_identity_to_log @candidate.id
         redirect_to candidate_interface_check_email_sign_in_path
-      elsif @candidate&.valid?
+      elsif @candidate.valid?
         AuthenticationMailer.sign_in_without_account_email(to: @candidate.email_address).deliver_now
         redirect_to candidate_interface_check_email_sign_in_path
       else
-        @candidate = Candidate.new
         render :new
       end
     end
--- a/app/controllers/candidate_interface/sign_in_controller.rb
+++ b/app/controllers/candidate_interface/sign_in_controller.rb
@@ -26,16 +26,26 @@ module CandidateInterface
     end
 
     def create
-      @candidate = Candidate.for_email candidate_params[:email_address]
+      if params[:u] && FeatureFlag.active?('improved_expired_token_flow')
+        begin
+          candidate_id = Encryptor.decrypt(params[:u])
+          @candidate = Candidate.find(candidate_id)
+        rescue ActiveSupport::MessageEncryptor::InvalidMessage
+          redirect_to candidate_interface_sign_in_path && return
+        end
+      else
+        @candidate = Candidate.for_email candidate_params[:email_address]
+      end
 
-      if @candidate.persisted?
+      if @candidate&.persisted?
         MagicLinkSignIn.call(candidate: @candidate)
         add_identity_to_log @candidate.id
         redirect_to candidate_interface_check_email_sign_in_path
-      elsif @candidate.valid?
+      elsif @candidate&.valid?
         AuthenticationMailer.sign_in_without_account_email(to: @candidate.email_address).deliver_now
         redirect_to candidate_interface_check_email_sign_in_path
       else
+        @candidate = Candidate.new
         render :new
       end
     end
--- a/app/controllers/candidate_interface/sign_in_controller.rb
+++ b/app/controllers/candidate_interface/sign_in_controller.rb
@@ -5,13 +5,7 @@ module CandidateInterface
 
     def new
       if FeatureFlag.active?('improved_expired_token_flow') && params[:u]
-        begin
-          Encryptor.decrypt(params[:u])
-
-          redirect_to candidate_interface_expired_sign_in_path(u: params[:u])
-        rescue ActiveSupport::MessageEncryptor::InvalidMessage
-          redirect_to candidate_interface_sign_in_path
-        end
+        redirect_to candidate_interface_expired_sign_in_path(u: params[:u])
       else
         @candidate = Candidate.new
       end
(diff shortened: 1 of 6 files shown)


app/controllers/candidate_interface/sign_in_controller.rb
```diff
@@ -6,9 +6,9 @@ class SignInController < CandidateInterfaceController
     def new
       if FeatureFlag.active?('improved_expired_token_flow') && params[:u]
         begin
-          decrypted_candidate_id = Encryptor.decrypt(params[:u])
+          Encryptor.decrypt(params[:u])
 
-          redirect_to candidate_interface_expired_sign_in_path(id: decrypted_candidate_id)
+          redirect_to candidate_interface_expired_sign_in_path(u: params[:u])
         rescue ActiveSupport::MessageEncryptor::InvalidMessage
           redirect_to candidate_interface_sign_in_path
```
tijmenb: Perhaps we can remove this? We already check it in the expire method.

app/controllers/candidate_interface/sign_in_controller.rb
```diff
@@ -26,7 +26,16 @@ def interstitial
...
     def create
-      @candidate = Candidate.for_email candidate_params[:email_address]
+      if params[:u] && FeatureFlag.active?('improved_expired_token_flow')
+        begin
+          candidate_id = Encryptor.decrypt(params[:u])
+          @candidate = Candidate.find(candidate_id)
+        rescue ActiveSupport::MessageEncryptor::InvalidMessage
+          redirect_to candidate_interface_sign_in_path && return
+        end
+      end
```
tijmenb: Let's create a new action for this case and use that in the form.

app/controllers/candidate_interface/sign_in_controller.rb
```diff
@@ -80,9 +90,18 @@ def authenticate
 
     def expired
       raise unless FeatureFlag.active?('improved_expired_token_flow')
-      return redirect_to candidate_interface_sign_in_path unless params[:id]
 
-      @candidate = Candidate.find(params[:id])
+      if params[:u].blank?
```
tijmenb: `params.fetch(:u)`?

stevehook: I've left as it was otherwise it changes the outcome from a redirect to an error.

app/views/candidate_interface/sign_in/expired.html.erb
```diff
@@ -2,15 +2,15 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <%= form_with model: @candidate, url: candidate_interface_sign_in_path, method: :post do |f| %>
+    <%= form_with model: @candidate, url: candidate_interface_create_expired_sign_in_path, method: :post do |f| %>
```
tijmenb: Does this need to be a form with `@candidate`? Seems like we don't need the `@candidate` at all here?

app/controllers/candidate_interface/sign_in_controller.rb
```diff
@@ -40,6 +34,26 @@ def create
...
+        begin
+          candidate_id = Encryptor.decrypt(params[:u])
+          @candidate = Candidate.find(candidate_id)
+        rescue ActiveSupport::MessageEncryptor::InvalidMessage
+          redirect_to candidate_interface_sign_in_path
+          return
+        end
+      end
+
+      if @candidate&.persisted?
```
tijmenb: This will always be true, since `Candidate.find` would raise a RecordNotFound exception.

app/controllers/candidate_interface/sign_in_controller.rb
```diff
@@ -40,6 +34,26 @@ def create
       end
     end
 
+    def create_expired
```
tijmenb: Perhaps move this down below the expired method? Not sure about the name but can't think of something else.

stevehook: I'm not sure about the name either. `create_from_expired_token`?

tijmenb: `create_for_known_user`?

app/controllers/candidate_interface/sign_in_controller.rb
```diff
@@ -80,9 +75,38 @@ def authenticate
...
+
+      begin
+        candidate_id = Encryptor.decrypt(params[:u])
+        @candidate = Candidate.find(candidate_id)
+      rescue ActiveSupport::MessageEncryptor::InvalidMessage
+        redirect_to candidate_interface_sign_in_path
+      end
+    end
+
+    def create_from_expired_token
```
tijmenb: I think this can be simplified a bit:

```rb
def create_from_expired_token
  render_404 unless FeatureFlag.active?('improved_expired_token_flow')

  candidate_id = Encryptor.decrypt(params.fetch(:u))
  candidate = Candidate.find(candidate_id)
  MagicLinkSignIn.call(candidate: candidate)
  redirect_to candidate_interface_check_email_sign_in_path
rescue ActiveSupport::MessageEncryptor::InvalidMessage
  render 'errors/not_found', status: :forbidden
end
```

- Rely on `fetch` to check for token so that we display a BadRequest error page if it's not there
- Render a forbidden status code when the token is incorrect
- Rely on `Candidate.find` to raise an exception if the user id isn't found
- We don't need the instance variable anymore

app/controllers/candidate_interface/sign_in_controller.rb
```diff
@@ -80,9 +75,38 @@ def authenticate
...
 
-      @candidate = Candidate.find(params[:id])
+      if params[:u].blank?
+        redirect_to candidate_interface_sign_in_path
+        return
+      end
+
+      begin
+        candidate_id = Encryptor.decrypt(params[:u])
+        @candidate = Candidate.find(candidate_id)
```
tijmenb: I think we don't need to decrypt & find the candidate here, we're not using it for anything?

duncanjbrown:
Catching exceptions from `ActiveSupport::MessageEncryptor` breaks the (minor) encapsulation of `Encryptor` and makes the logic a little bit harder to follow than a plain `if` statement. What would you think about trapping the exception inside `Encryptor` and returning `false` if it didn't work?

stevehook:
> Catching exceptions from `ActiveSupport::MessageEncryptor` breaks the (minor) encapsulation in `Encryptor` and makes the logic a little bit harder to follow than a plain `if` statement. What would you think about trapping the exception inside `Encryptor` and returning `false` if it didn't work?

Wasn't sure about this but having tried it I like it. Good suggestion. Thanks!

PR #1407, opened 2020-02-18
Update API to handle individual work history breaks

chubberlisk:
## Context

On the candidate interface, we've added a new feature (which is currently under a feature flag called `work_breaks`) in which we show breaks that occur in their work history so that a candidate can enter a reason for each individual one. See screenshot below (FYI it shouldn't show the "Tell us about any breaks in your history", this is a bug).

Currently, we have a single text attribute called `work_history_breaks` on an `ApplicationForm` and so the API simply spits this out when presenting an application. But with the new feature we added a new table called `ApplicationWorkHistoryBreaks` which means that an `ApplicationForm` can have many `ApplicationWorkHistoryBreaks`.

image

This PR updates the API to return a candidate's work history breaks based on whether or not there is a value for `work_history_breaks` text attribute.

If there is, then use that. Else if there are values for `application_work_history_breaks`, then munge all `application_work_history_breaks` together. Else, return an empty string.

E.g. a munged response will look like this:

```
February 2019 to April 2019: I was watching TV.

September 2019 to December 2019: I was playing games.
```

## Guidance to review

- Are there redundant tests? The code doesn't look at the `work_breaks` feature flag but is it still good to have the tests?
- Are we happy with this response?
- Am I missing anything else?

## Link to Trello card

## Things to check

### DIFF
--- a/app/presenters/vendor_api/single_application_presenter.rb
+++ b/app/presenters/vendor_api/single_application_presenter.rb
@@ -45,7 +45,7 @@ module VendorApi
           work_experience: {
             jobs: work_experience_jobs,
             volunteering: work_experience_volunteering,
-            work_history_break_explanation: application_form.work_history_breaks,
+            work_history_break_explanation: work_history_breaks,
           },
           offer: offer,
           rejection: get_rejection,
@@ -214,6 +214,25 @@ module VendorApi
         disability: '00',
         ethnicity: '10',
       }
+    end
+
+    def work_history_breaks
+      # With the new feature of adding individual work history breaks, `application_form.work_history_breaks`
+      # is a legacy column. So we'll need to check if an application form has this value first.
+      if application_form.work_history_breaks
+        application_form.work_history_breaks
+      elsif application_form.application_work_history_breaks.any?
+        breaks = application_form.application_work_history_breaks.map do |work_break|
+          start_date = work_break.start_date.to_s(:month_and_year)
+          end_date = work_break.end_date.to_s(:month_and_year)
+
+          "#{start_date} to #{end_date}: #{work_break.reason}"
+        end
+
+        breaks.join("\n\n")
+      else
+        ''
+      end
     end
   end
 end
--- a/spec/presenters/vendor_api/single_application_presenter_spec.rb
+++ b/spec/presenters/vendor_api/single_application_presenter_spec.rb
@@ -71,6 +71,135 @@ RSpec.describe VendorApi::SingleApplicationPresenter do
     end
   end
 
+  describe 'attributes.work_history_break_explanation' do
+    let(:february2019) { Time.zone.local(2019, 2, 1) }
+    let(:april2019) { Time.zone.local(2019, 4, 1) }
+    let(:september2019) { Time.zone.local(2019, 9, 1) }
+    let(:december2019) { Time.zone.local(2019, 12, 1) }
+
+    context 'when the work breaks feature flag is on and work history breaks field has a value' do
+      it 'returns the work_history_breaks attribute of an application' do
+        FeatureFlag.activate('work_breaks')
+        breaks = []
+        application_form = build_stubbed(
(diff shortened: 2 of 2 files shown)


app/presenters/vendor_api/single_application_presenter.rb
```diff
@@ -215,5 +215,20 @@ def hesa_itt_data
...
     end
+
+    def work_history_breaks
+      if FeatureFlag.active?('work_breaks')
+        breaks = application_form.application_work_history_breaks.inject('') do |explanation, work_break|
+          start_date = work_break.start_date.to_s(:month_and_year)
+          end_date = work_break.end_date.to_s(:month_and_year)
+
+          explanation << "#{start_date} to #{end_date}: #{work_break.reason}\n\n"
+        end
```
tvararu: Alternatively:

```ruby
breaks = application_form.application_work_history_breaks.map do |work_break|
  start_date = work_break.start_date.to_s(:month_and_year)
  end_date = work_break.end_date.to_s(:month_and_year)

  "#{start_date} to #{end_date}: #{work_break.reason}"
end.join('\n\n')
```

No more need for `.strip` after!

chubberlisk:
Realised that this is a lil more complicated than initially thought, currently this PR doesn't consider if a candidate has a value in `work_history_breaks` while the `work_breaks` feature flag is on.

Talked this through with Duncan and drew up a table: https://trello-attachments.s3.amazonaws.com/5cb88ec4be601441218204a0/5e149b5f6ab09630e5d6f2c9/174eb5e5d0b5a17d9cd7c270d0024b2e/IMG_20200218_122509.jpg.

Will update this PR now.

app/presenters/vendor_api/single_application_presenter.rb
chubberlisk: Rubocop shouted at me for chaining a method call at the end of `do..end`. So did dis instead:

```ruby
breaks = application_form.application_work_history_breaks.map do |work_break|
  start_date = work_break.start_date.to_s(:month_and_year)
  end_date = work_break.end_date.to_s(:month_and_year)

  "#{start_date} to #{end_date}: #{work_break.reason}"
end

breaks.join("\n\n")
```

tvararu: Silly rubocop!

chubberlisk:
@duncanjbrown I've updated the PR now as well as the PR description! Please could you have a look for me. Thanks!

app/presenters/vendor_api/single_application_presenter.rb
```diff
@@ -215,5 +215,22 @@ def hesa_itt_data
         ethnicity: '10',
       }
     end
+
+    def work_history_breaks
+      if application_form.work_history_breaks
```
duncanjbrown: It might be worth adding a comment here explaining why we check this

spec/presenters/vendor_api/single_application_presenter_spec.rb
```diff
@@ -71,6 +71,135 @@
...
+        application_choice = build_stubbed(:application_choice, status: 'awaiting_provider_decision', application_form: application_form)
+
+        response = VendorApi::SingleApplicationPresenter.new(application_choice).as_json
+
+        expect(response.to_json).to be_valid_against_openapi_schema('Application')
+        expect(response[:attributes][:work_experience][:work_history_break_explanation]).to eq('I was sleeping.')
+      end
+    end
+
+    context 'when the work breaks feature flag is off and work history breaks field is empty' do
```
duncanjbrown: Great specs. I think these three could be made a bit more distinct, maybe:

- `when the work breaks feature flag is off and no breaks have been entered`
- `when the work breaks feature flag is off and the work history breaks field has a value`
- `when the work breaks feature flag is off and breaks have been entered`

ie when there are breaks, say what sort they are rather than what sort they aren't

PR #1409, opened 2020-02-18
Move service performance dashboard into integrations

tijmenb:
## Context

This moves the dashboard to `/integrations/performance-dashboard`, to make it clearer it’s a special page that is open to the public. The service performance page now just links to it. This will make it easier to add non-public stuff to the support section in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1240, and prevents us making things public accidentally.

See commit.

## Link to Trello card

https://trello.com/c/dKBwLMPv

## Things to check


config/routes.rb
```diff
@@ -324,6 +324,7 @@
   namespace :integrations, path: '/integrations' do
     post '/notify/callback' => 'notify#callback'
     get '/feature-flags' => 'feature_flags#index'
+    get '/performance-dashboard' => 'performance_dashboard#dashboard', as: :performance
```
benilovj: Do we want `dashboard` in the URL?

app/controllers/support_interface/performance_controller.rb
```diff
@@ -1,9 +1,6 @@
 module SupportInterface
   class PerformanceController < SupportInterfaceController
-    skip_before_action :authenticate_support_user!, only: :index
-
     def index
```
stevehook: Would it make sense just to redirect to the new location?

config/routes.rb
tijmenb: Yeah, I think that makes the purpose clearer.

app/controllers/support_interface/performance_controller.rb
tijmenb: I'd like to add more stuff the the page in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1240, that's why we're keeping it and linking.

PR #1313, opened 2020-02-10
Remove GTM code left behind after PR #1304

benilovj:
## Context

PR #1304 didn't quite get rid of all Google Tag Manager code.

## Changes proposed in this pull request

- remove (hopefully) the last bits of GTM

## Things to check

- [x] This code doesn't rely on migrations in the same Pull Request
- [x] If this code includes a migration adding or changing columns, it also backfills existing records for consistency
- [x] API release notes have been updated if necessary
- [x] New environment variables have been [added to the Azure config](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training#azure-hosting-devops-pipeline)

PR #1410, opened 2020-02-18
Refactor bug caused by SendCandidateRejectionEmail

davidgisbey:
#1313  Context

This SendCandidateRejectionEmail service. is currently triggering sentry errors on QA.

This is due to an else statement failing due to invalid data on QA

- Refactor the conditional logic to use an elsif statement

## Link to Trello card

https://trello.com/c/22e7C80p/

## Things to check

### DIFF
--- a/app/services/send_candidate_rejection_email.rb
+++ b/app/services/send_candidate_rejection_email.rb
@@ -7,16 +7,19 @@ class SendCandidateRejectionEmail
 
   def call
     candidate_application_choices = application_choice.application_form.application_choices
-    number_of_pending_decisions = candidate_application_choices.select(&:awaiting_provider_decision?).count
+    number_of_pending_decisions = candidate_application_choices.awaiting_provider_decision.count
+    number_of_offers = candidate_application_choices.offer.count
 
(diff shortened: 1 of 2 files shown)


app/services/send_candidate_rejection_email.rb
```diff
@@ -7,13 +7,14 @@ def initialize(application_choice:)
...
 
     if candidate_application_choices.all?(&:rejected?)
       CandidateMailer.send(:application_rejected_all_rejected, application_choice).deliver_later
     elsif number_of_pending_decisions.positive?
       CandidateMailer.send(:application_rejected_awaiting_decisions, application_choice).deliver_later
-    else
+    elsif number_of_offers.positive?
       CandidateMailer.send(:application_rejected_offers_made, application_choice).deliver_later
     end
     add_audit_comment(application_choice)
```
stevehook: Does it makes sense to add the audit entry when no email has been sent?

davidgisbey: Good point. Will add a test too.

PR #1412, opened 2020-02-18
Decline by default per application form

tijmenb:
## Context

Currently we do decline by default per application choice.

But there can be multiple choices per application form that expire at exactly the same time (they always do, since the `decline_by_default_at` for all offered choices is set when the last offer comes in). This means candidates would get multiple emails at the same time if we added email notifications.

This aggregates the application forms that need to be declined by default, which means we can start sending a single email. Rewrites the worker spec to be more system-like.

## Link to Trello card

Prerequisite for https://trello.com/c/DtPNK4vW

## Things to check


spec/services/get_application_forms_ready_to_decline_by_default_spec.rb
```diff
@@ -1,33 +1,32 @@
...
     )
+
+    application_choice.application_form
   end
 
   it 'returns application choices with decline_by_default_at in the past' do
-    application1 = create_application(
+    application1 = create_application_form(
       status: 'offer',
       decline_by_default_at: 1.business_days.ago,
```
malcolmbaig: While we're here, suggest ordering these applications by their decline_by_default_at date - makes it a bit easier to follow when we get to the assertions in this scenario.

malcolmbaig:
As discussed IRL, this part of the commit message would benefit from some additional explanation: `...there can be multiple choices per application form that expire at
exactly the same time (they always do)`.

malcolmbaig:
Minor feedback, otherwise LGTM 👍

spec/services/get_application_forms_ready_to_decline_by_default_spec.rb
tijmenb: Refactored to give them meaningful names.

tijmenb:
@malcolmbaig updated PR / commit message!

app/workers/decline_offers_by_default_worker.rb
```diff
@@ -2,11 +2,9 @@ class DeclineOffersByDefaultWorker
   include Sidekiq::Worker
 
   def perform
-    GetApplicationChoicesReadyToDeclineByDefault.call.each do |choice|
-      begin
-        DeclineOfferByDefault.new(application_choice: choice).call
-      rescue StandardError => e
-        Rails.logger.warn "[DBD] ignoring application_choice #{choice.id}: #{e.message}"
+    GetApplicationFormsReadyToDeclineByDefault.call.each do |application_form|
```
stevehook: Nothing to do with this PR but I'm wondering whether we should be using `find_each` here instead of `each` to avoid loading too many model objects into memory at once.

tijmenb:
> I did wonder whether there were edge cases, e.g. when an offer is withdrawn, in which you could end up with different DBD dates on application choices for the same form but looking at the code that doesn't seem to be possible. (the service that sets the DBD operates on a form).

Indeed, it *shouldn't* be the case. We should think about moving the dbd date to the form, to avoid uncertainty.

app/workers/decline_offers_by_default_worker.rb
tijmenb: Yeah good shout, let's keep that in mind.

PR #1413, opened 2020-02-18
Open Apply up to search engines

fofr:
"They tried Googling Apply multiple times but couldn’t find the link."
Candidates have had trouble finding Apply again so they can return to their application.

If we make Apply visible on Google, candidates will be able to find us again and more easily sign in.

- Remove noindex meta tag
- Update robots.txt but disallow /support URLs

https://trello.com/c/Ma1IoQjs/841-sign-in-strategic-review-returning-candidates

### DIFF
--- a/app/views/layouts/base.html.erb
+++ b/app/views/layouts/base.html.erb
@@ -6,7 +6,6 @@
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
 
-    <meta name="robots" content="noindex, nofollow">
     <meta name="viewport" content="width=device-width, initial-scale=1, viewport-fit=cover">
     <meta name="theme-color" content="#0b0c0c">
 
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,1 +1,2 @@
-# See http://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
+User-agent: *
+Disallow: /support


app/views/layouts/base.html.erb
```diff
@@ -6,7 +6,6 @@
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
 
-    <meta name="robots" content="noindex, nofollow">
```
tijmenb: This should only happen for `HostingEnvironment.production?` to avoid staging/QA/sandbox from being indexed.

We should also exclude the support section here, if that's the plan.

public/robots.txt
```diff
@@ -1 +1,2 @@
-# See http://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
+User-agent: *
+Disallow: /support
```
tijmenb: As I understand it, the problem with robots.txt is that if Google finds a link to /support somewhere, it will _index_ the page, but never visit it, so it won't know that the robots `<meta>` doesn't allow indexing. It will therefore keep it in index.

This is what leads to service domains being displayed without description:

Screenshot 2020-02-18 at 14 48 34

fofr:
We can't become discoverable on google (and can't have a GOV.UK page) until we've passed a service assessment. Or we make an exception and can demonstrate the need.

PR #1415, opened 2020-02-18
Update content and collapse courses on providers page

paulrobertlloyd:
## Context

The number of providers and courses shown on ‘Training providers available through Apply for teacher training’ is getting quite long, making it harder to see which providers are opted in. We will group providers by region once that functionality is available, but for now, we can collapse the list of courses each provider is accepting application for on Apply.

This all updated the content to add some context around dual-running and which service you should used based on the information on this page.

Screenshot 2020-02-18 at 16 14 06

## Link to Trello card

https://trello.com/c/IMkev1Ek/1041-use-accordion-components-on-the-candidate-providers-list-page

### DIFF
--- a/app/views/candidate_interface/content/providers.html.erb
+++ b/app/views/candidate_interface/content/providers.html.erb
@@ -32,7 +32,7 @@
         <div class="govuk-details__text">
           <ul class="govuk-list govuk-list--bullet">
             <% courses.sort_by(&:name).each do |course| %>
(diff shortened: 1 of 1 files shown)


duncanjbrown:
Is it right that the links on this page go to eg https://apply-for-te-parsable-p-hgzzel.herokuapp.com/candidate/apply?courseCode=238T&providerCode=1N1 rather than https://apply-for-te-parsable-p-hgzzel.herokuapp.com/candidate/eligibility?courseCode=238T&providerCode=1N1 ?

paulrobertlloyd:
@duncanjbrown Good point, I hadn’t changed this, but I think eligibility rather than apply is the better option.

PR #1426, opened 2020-02-19
Use database IDs instead of Course and Provider codes

malcolmbaig:
## Context

In the candidate interface, there are several places where we use codes
to identify Courses and Providers. Certain URL paths contain these
codes, they subsequently end up in the params for the related actions,
and we end up using them in database queries.

The issue with this is that course codes are unique for a provider, but
not unique globally. In cases where a course code is shared between two
providers, this can cause the system to accidentally choose the wrong
course.

Update route definitions and all related logic so that we use database ids instead of codes to identify Course and Providers.

## Guidance to review

See individual commits for detail.

## Link to Trello card

https://trello.com/c/TU8oMtpf

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/course_choices_controller.rb
+++ b/app/controllers/candidate_interface/course_choices_controller.rb
@@ -67,7 +67,7 @@ module CandidateInterface
       elsif @pick_course.single_site?
         course_option = CourseOption.where(course_id: @pick_course.course.id).first
 
-        pick_site_for_course(course_code, course_option.id)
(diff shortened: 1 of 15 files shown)


tvararu:
Screenshot 2020-02-19 at 11 50 42

Testing this on the review app, I noticed that the autocompletes aren't firing anymore. Maybe the IDs have changed?

fofr:
(FYI a provider/course code combination is always unique)

PR #1428, opened 2020-02-19
Get ready to launcharoo work history breaks!

chubberlisk:
## Context

Currently, we only show individual breaks when the `work_breaks` feature flag is on. However, to actually turn this on we have to consider the fact that we will turn on this feature while there are applications in progress i.e. there may be applications that have already used the "old" way of entering work history breaks (the capture-all text box). This means we must - Improvise. Adapt. Overcome!

This PR does some smol but important things:

- updates the existing breaks component to show the break length in years and months (this was done for the placeholder before but I forgot about existing breaks) and does so by extracting a method into `ViewHelper` and reusing it in `BreakInWorkHistoryComponent`
- redirects to the review work history page if the `work_breaks` feature flag is off when attempting to go to a page to create, edit, etc. an individual break

And then, makes the new feature "backwards-compatible" by taking into consideration whether the feature flag is on and whether there is a value already for `work_history_breaks` (the attribute that stores the capture-all work history breaks in comparison to `application_work_history_breaks` an array of `ApplicationWorkHistoryBreak`s).

It also ensures that change and delete links don't show in Support.

## Guidance to review

- Have a look at `spec/components/work_history_review_component_spec.rb` does the specs make sense? Would highly recommend you also test through the UI make it easier to visualise.
- Does the redirect make sense?
- Is there anything I am missing?

## Link to Trello card

## Things to check

### DIFF
--- a/app/components/work_history_review_component.html.erb
+++ b/app/components/work_history_review_component.html.erb
@@ -12,13 +12,11 @@
       <% end %>
     <% end %>
   <% elsif entry.is_a?(WorkHistoryWithBreaks::BreakPlaceholder) %>
-    <% if FeatureFlag.active?('work_breaks') %>
+    <% if show_break_placeholders? %>
       <%= render(BreakPlaceholderInWorkHistoryComponent.new(work_break: entry)) %>
     <% end %>
   <% elsif entry.is_a?(ApplicationWorkHistoryBreak) %>
-    <% if FeatureFlag.active?('work_breaks') %>
-      <%= render(BreakInWorkHistoryComponent.new(work_break: entry)) %>
-    <% end %>
+    <%= render(BreakInWorkHistoryComponent.new(work_break: entry, editable: @editable)) %>
   <% end %>
 <% end %>
 
@@ -26,7 +24,7 @@
   <%= render(SummaryCardComponent.new(rows: no_work_experience_rows, editable: @editable)) %>
 <% end %>
 
-<% if breaks_in_work_history? %>
+<% if show_consolidated_work_history_breaks? %>
   <%= render(SummaryCardComponent.new(rows: break_in_work_history_rows, editable: @editable)) %>
 <% end %>
 
--- a/app/helpers/view_helper.rb
+++ b/app/helpers/view_helper.rb
@@ -77,6 +77,16 @@ module ViewHelper
     end
   end
 
+  def format_months_to_years_and_months(number_of_months)
+    duration_parts = ActiveSupport::Duration.build(number_of_months.months).parts
+
+    if duration_parts[:years].positive?
+      "#{pluralize(duration_parts[:years], 'year')} and #{pluralize(duration_parts[:months], 'month')}"
+    else
+      pluralize(number_of_months, 'month')
+    end
+  end
+
 private
 
   def prepend_css_class(css_class, current_class)
--- a/spec/components/work_history_review_component_spec.rb
+++ b/spec/components/work_history_review_component_spec.rb
@@ -1,6 +1,8 @@
 require 'rails_helper'
 
 RSpec.describe WorkHistoryReviewComponent do
+  let(:february2019) { Time.zone.local(2019, 2, 1) }
+  let(:april2019) { Time.zone.local(2019, 4, 1) }
   let(:data) do
     {
       role: 'Teaching Assistant',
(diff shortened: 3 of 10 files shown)


tvararu:
After (review app):

Screenshot 2020-02-19 at 13 19 02

spec/components/work_history_review_component_spec.rb
```diff
@@ -97,39 +99,143 @@
...
+      end
+    end
+
+    context 'when the work breaks feature flag is on and work history breaks has a value' do
+      it 'renders component with capture all work breaks and without individual breaks' do
+        FeatureFlag.activate('work_breaks')
+        application_form.work_history_breaks = 'I was sleeping.'
+        application_form.application_work_history_breaks = []
+
+        result = render_inline(described_class.new(application_form: application_form))
```
tvararu: You could create a helper method for this inside of this spec; allow passing in a few arguments which will let you reuse it for all the specs. Would be DRY-er, WDYT?

```ruby
result = render(feature_active: true, explanation: 'I was sleeping')

# ...

def render(feature_active: false, explanation: nil, breaks: [])
  if feature_active
    FeatureFlag.activate('work_breaks')
  else
    FeatureFlag.deactivate('work_breaks')
  end
  application_form.work_history_breaks = explanation
  application_form.application_work_history_breaks = breaks
 
  result = render_inline(described_class.new(application_form: application_form))
end
```

app/components/work_history_review_component.html.erb
```diff
@@ -12,21 +12,19 @@
       <% end %>
     <% end %>
   <% elsif entry.is_a?(WorkHistoryWithBreaks::BreakPlaceholder) %>
-    <% if FeatureFlag.active?('work_breaks') %>
+    <% if FeatureFlag.active?('work_breaks') && @application_form.work_history_breaks.blank? && @editable %>
```
duncanjbrown: Could this logic go into a named method on the component?

app/components/work_history_review_component.html.erb
```diff
@@ -12,21 +12,19 @@
...
+    <%= render(BreakInWorkHistoryComponent.new(work_break: entry, editable: @editable)) %>
   <% end %>
 <% end %>
 
 <% if !show_missing_banner? && @application_form.application_work_experiences.empty? %>
   <%= render(SummaryCardComponent.new(rows: no_work_experience_rows, editable: @editable)) %>
 <% end %>
 
-<% if breaks_in_work_history? %>
+<% if breaks_in_work_history? && (@application_form.work_history_breaks || !FeatureFlag.active?('work_breaks')) %>
```
duncanjbrown: And this one?

spec/components/work_history_review_component_spec.rb
```diff
@@ -97,39 +99,143 @@
...
     end
 
-    context 'when work history breaks are editable' do
+    around do |example|
+      Timecop.freeze(Time.zone.local(2019, 10, 1)) do
+        example.run
+      end
+    end
+
+    context 'when capture-all work history breaks are editable' do
```
duncanjbrown: Does this mean the "old" data? I'm not sure from the name

app/helpers/view_helper.rb
```diff
@@ -77,6 +77,17 @@ def candidate_sign_in_url(candidate)
...
+  def format_months_to_years_and_months(number_of_months)
+    if number_of_months <= 12
+      pluralize(number_of_months, 'month')
+    else
+      years = number_of_months / 12
+      months = number_of_months % 12
+
+      "#{pluralize(years, 'year')} and #{pluralize(months, 'month')}"
+    end
+  end
```
duncanjbrown: How about `ActiveSupport::Duration`?

```
ActiveSupport::Duration.build(13.months).parts
=> {:years=>1, :months=>1}

 ActiveSupport::Duration.build(6.months).parts
=> {:months=>6}
```

spec/components/work_history_review_component_spec.rb
chubberlisk: Yep, trying to refer to this:

image

Apparently, "**_consolidated_** work history breaks" is the way to go according to Paul H and Emma. 😅 Do you think that clearer?

duncanjbrown: Sounds good!

PR #1432, opened 2020-02-19
Use accordion component for providers

paulrobertlloyd:
## Context

The accordion component is better suited for hiding courses… like everyone else said it was 🙄

accordion

### DIFF
--- a/app/controllers/candidate_interface/content_controller.rb
+++ b/app/controllers/candidate_interface/content_controller.rb
@@ -19,11 +19,14 @@ module CandidateInterface
     end
 
     def providers
+      provider_courses = Struct.new(:provider_name, :courses)
+
       @courses_by_provider = Course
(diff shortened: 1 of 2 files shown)


app/controllers/candidate_interface/content_controller.rb
```diff
@@ -18,12 +18,15 @@ def terms_candidate
       render_content_page :terms_candidate
     end
 
+    ProviderCourses = Struct.new(:provider_name, :courses)
```
tvararu:
```suggestion
    provider_courses = Struct.new(:provider_name, :courses)
```

`UpperCamelCased` names are usually classes in Ruby!

tvararu: Also, this should be moved inside of `def providers`

PR #1433, opened 2020-02-19
Add an MVP email log

tijmenb:
## Context

There's been a lot of work around emails, and it has come up repeatedly in discussions that we need better visibility into email sending. This is an MVP to do that.

Every time we send an email, add a new `Email` record. We expose the emails as a list in the support section:

image

We save the to address, subject, full body and the notify reference that some emails pass along (like [some auth mailer emails](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/app/mailers/authentication_mailer.rb#L8)). The reference will allow us to match receipts from the Notify callback to see which ones succeeded.

It's not linked from the menu yet because I don't know where it belongs - we'll see if this even works.

## Guidance to review

- Does this make sense?
- Anything else we should capture?
- We also persist the unhashed sign in tokens in the database, negating our entire strategy of encrypting stuff in the database. Is that a problem?
 
## Link to Trello card

None

## Things to check


tvararu:
> We also persist the unhashed sign in tokens in the database, negating our entire strategy of encrypting stuff in the database. Is that a problem?

The tokens are only valid for 1 hour, I don't think it negates the *entire* strategy.

The question is: if someone can has obtained read access to the database, would they even care about sign in tokens?

db/migrate/20200219160002_create_emails.rb
```diff
@@ -0,0 +1,12 @@
+class CreateEmails < ActiveRecord::Migration[6.0]
+  def change
+    create_table :emails do |t|
+      t.string :to, null: false
```
tvararu: I'm wondering if it's worth associating emails to the application form they pertain to... the only emails which I can think of that wouldn't have an application form to attach to are sign up emails. Then again, my hunch is this feature isn't that interesting for sign in/up emails, but for transactional emails between users/referees/providers which can all be tied to an application form.

lib/email_log_interceptor.rb
```diff
@@ -0,0 +1,14 @@
+class EmailLogInterceptor
+  def self.delivering_email(mail)
+    Email.create!(
+      to: mail.to.first,
+      subject: mail.subject,
+      body: mail.body.encoded,
+      notify_reference: mail.header['reference']&.value,
+    )
```
tvararu: Ah, now that I've read this, I'm not sure if it's actually trivial to infer the `application_form` from the `mail` object that we get here? Depending on the recipient it could be tricky

db/migrate/20200219160002_create_emails.rb
tijmenb: The problem is that we don't know at the moment, we'd have to update every email to try and add that information. Perhaps we should do that anyway, to add the `reference` to every email.

What I was thinking is that we could link email addresses to this page, which then shows all emails to a certain user.

lib/email_log_interceptor.rb
tijmenb: Yeah, it would be tricky. Adding a reference everywhere at some point might be worth it though.

duncanjbrown:
Does Notify give us the id of emails when it sends them? If we don't want to store bodies, could we store subject lines and link out to the Notify log? e.g. `https://www.notifications.service.gov.uk/services/022acc23-c40a-4077-bbd6-fc98b2155534/notification/9b1de29e-a4db-4c0f-8fe2-9f6d0d1b35c3`. We then get status reports and sent_at for free (for what they're worth)

tijmenb:
@duncanjbrown I think this intercept happens before we send them out to Notify. The only way to link our email to Notifies logs/callbacks is to use `reference` I think.

duncanjbrown:
Right now we email all `ProviderUsers` of a Provider when something happens. If we had a timeline (which would be 👌 ) we'd probably want to pack that down into a single event. Having `reference` on everything might help that?

tijmenb:
@tvararu @duncanjbrown soooo, went down the rabbit hole.

- We now add the application form as a header in all emails
- Which is picked up by the inceptor
- The inceptor generates a reference is there isn't one, so that we can do the Notify callback stuff

tijmenb:
*Another* iteration: we now keep track of which email we're sending.

app/models/email.rb
```diff
@@ -0,0 +1,9 @@
+class Email < ApplicationRecord
+  # Note that most, but not all, emails that we send are in regards to an
+  # application form. For sign-ups and sign-ins application_form will be nil.
+  belongs_to :application_form, optional: true
```
tvararu: 🙌

PR #1438, opened 2020-02-20
Adding a course from find for a signed in candidate

davidgisbey:
## Context
**https://trello.com/c/c4ZYOuaI/1010-add-you-selected-a-course-page-into-find-to-apply-journey will be affected by this work**

This PR covers stretch goal 1 of the Trello card. This is to fix adding a course from find for candidates who are already signed in.
 
- Rework the start page controller to call the ExistingCandidateAuthentication service if the user is signed in and course params are present

## Guidance to review

I think the naming of the ExistingCandidateAuthentication does not really describe what it actually does,  but I can't come up with anything that seems to describe it well.

# Link to Trello card

https://trello.com/c/71lG72Y6/

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/start_page_controller.rb
+++ b/app/controllers/candidate_interface/start_page_controller.rb
@@ -2,7 +2,7 @@ module CandidateInterface
   class StartPageController < CandidateInterfaceController
     before_action :show_pilot_holding_page_if_not_open
     skip_before_action :authenticate_candidate!
-    before_action :redirect_signed_in_candidate
+    before_action :redirect_to_interstitial_path_if_signed_in
 
     def show; end
 
@@ -29,28 +29,11 @@ module CandidateInterface
 
   private
 
-    def redirect_signed_in_candidate
+    def redirect_to_interstitial_path_if_signed_in
       if current_candidate.present?
         add_course_from_find_id_to_candidate if params[:providerCode].present?
 
-        service = ExistingCandidateAuthentication.new(candidate: current_candidate)
-        service.execute
-
-        if service.candidate_does_not_have_a_course_from_find || service.candidate_has_submitted_application
-          redirect_to candidate_interface_interstitial_path
-        elsif service.candidate_has_already_selected_the_course
-          flash[:warning] = "You have already selected #{@course.name_and_code}."
-          redirect_to candidate_interface_course_choices_review_path
-        elsif service.candidate_has_new_course_added
-          redirect_to candidate_interface_course_choices_review_path
-        elsif service.candidate_should_choose_site
-          redirect_to candidate_interface_course_choices_site_path(@course.provider.code, @course.code)
-        elsif service.candidate_already_has_3_courses
-          flash[:warning] = "You cannot have more than 3 course choices. You must delete a choice if you want to apply to #{@course.name_and_code}."
-          redirect_to candidate_interface_course_choices_review_path
-        else
-          redirect_to candidate_interface_application_form_path
-        end
+        redirect_to candidate_interface_interstitial_path
       end
     end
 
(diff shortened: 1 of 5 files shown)


app/controllers/candidate_interface/start_page_controller.rb
```diff
@@ -26,5 +26,38 @@ def eligibility_params
...
         .transform_values(&:strip)
     end
+
+  private
+
+    def redirect_signed_in_candidate
+      if current_candidate.present?
+        add_course_from_find_id_to_candidate if params[:providerCode].present?
+
+        service = ExistingCandidateAuthentication.new(candidate: current_candidate)
```
davidgisbey: Would love any thoughts on what this service should be renamed to. The current naming doesn't really describe what is it doing

spec/system/candidate_interface/candidate_signed_in_user_with_course_params_spec.rb
```diff
@@ -0,0 +1,173 @@
+require 'rails_helper'
+
+RSpec.describe 'An existing candidate arriving from Find with a course and provider code' do
+  include CourseOptionHelpers
+  scenario 'candidate is not signed in and retains their course selection through the sign up process' do
```
fofr:
```suggestion
  scenario 'candidate is signed in' do
```

davidgisbey: thanks

spec/system/candidate_interface/candidate_signed_in_user_with_course_params_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+    and_i_am_signed_in
+    and_i_have_less_than_3_application_options
+
+    when_i_arrive_at_the_apply_from_find_page_with_course_params_with_multiple_sites
+    and_i_click_apply_on_apply
+    then_i_should_see_the_course_choices_site_page
+    and_i_see_the_form_to_pick_a_location
+
+    given_the_course_i_selected_has_multiple_sites
+    and_i_am_an_existing_candidate_on_apply
```
stevehook: Creating a new candidate in the middle of the test confused me a bit. But I think we should probably leave as is - we don't want to break this up and multiply the number of tests we have. What do you think?

spec/system/candidate_interface/candidate_signed_in_user_with_course_params_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+    create(:course_option, site: @site, course: @course, vacancy_status: 'B')
+  end
+
+  def and_i_am_an_existing_candidate_on_apply
+    @email = "#{SecureRandom.hex}@example.com"
+    @candidate = create(:candidate, email_address: @email)
+  end
+
+  def and_i_have_less_than_3_application_options
+    application_form = create(:application_form, candidate: @candidate)
```
stevehook: Does creating more than one application form for the same candidate cause problems? (We assume elsewhere that the `current_application` is `candidate.application_forms.first`

tvararu:
Works great in my testing on the review app 💯

spec/system/candidate_interface/candidate_signed_in_user_with_course_params_spec.rb
davidgisbey: Yea I get that and it definitely could be split over more test files. I agree though that the overhead of creating the extra files and duplicating the majority of the methods probably isn't worth it.

spec/system/candidate_interface/candidate_signed_in_user_with_course_params_spec.rb
davidgisbey: Good spot. I think following the pattern of creating a new candidate is probably preferable considering you already have to parse it once anyway.

PR #1439, opened 2020-02-20
Course choice refactorings

malcolmbaig:
## Context

A series of small refactors identified while working on Trello card 883. 

Code changes with the goal of improving readability.

## Guidance to review

See individual commits.

## Link to Trello card

https://trello.com/c/TU8oMtpf

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/course_choices_controller.rb
+++ b/app/controllers/candidate_interface/course_choices_controller.rb
@@ -18,10 +18,9 @@ module CandidateInterface
 
     def make_choice
       @choice_form = CandidateInterface::CourseChosenForm.new(application_choice_params)
+      render :have_you_chosen and return unless @choice_form.valid?
 
-      if !@choice_form.valid?
-        render :have_you_chosen
-      elsif @choice_form.chosen_a_course?
+      if @choice_form.chosen_a_course?
         redirect_to candidate_interface_course_choices_provider_path
       else
         redirect_to 'https://find-postgraduate-teacher-training.education.gov.uk'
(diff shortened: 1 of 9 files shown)


app/helpers/select_options_helper.rb
```diff
@@ -0,0 +1,19 @@
+module SelectOptionsHelper
```
duncanjbrown: I can't see this included anywhere - is there some magic?

app/models/candidate_interface/pick_provider_form.rb
```diff
@@ -5,8 +5,8 @@ class PickProviderForm
     attr_accessor :provider_id
     validates :provider_id, presence: true
 
-    def other?
-      Course.exposed_in_find.where(provider_id: provider_id).blank?
+    def courses_available?
```
duncanjbrown: This is so much better

app/helpers/select_options_helper.rb
tvararu: I think Rails autoloads everything in `app/helpers`, but I may be mistaken!

malcolmbaig: > I can't see this included anywhere - is there some magic?

Yes - Rails makes all helpers available in all views.

app/controllers/candidate_interface/course_choices_controller.rb
```diff
@@ -18,10 +18,9 @@ def have_you_chosen
 
     def make_choice
       @choice_form = CandidateInterface::CourseChosenForm.new(application_choice_params)
+      render(:have_you_chosen) && return if !@choice_form.valid?
```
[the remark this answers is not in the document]
malcolmbaig: Strong preference for `and return` but the linter complains with its current config. Making some changes to this based on an offline chat with @tijmenb - will revisit.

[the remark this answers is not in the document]
malcolmbaig: Updated to use `and`, with an accompanying commit that changes the rubocop config.

PR #1441, opened 2020-02-20
Don't try to look up dbd_date on a choice that might not have an offer

duncanjbrown:
## Context

It's possible for an application's first choice not to have a dbd_date on it, but the mailer assumes it does.

Should fix https://sentry.io/organizations/dfe-bat/issues/1523597528/?project=1765973&referrer=slack

### DIFF
--- a/app/mailers/candidate_mailer.rb
+++ b/app/mailers/candidate_mailer.rb
@@ -141,8 +141,8 @@ class CandidateMailer < ApplicationMailer
   end
 
   def chase_candidate_decision(application_form)
-    @dbd_date = application_form.application_choices.first.decline_by_default_at.to_s(:govuk_date).strip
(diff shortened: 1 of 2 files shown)


app/mailers/candidate_mailer.rb
```diff
@@ -141,8 +141,8 @@ def reference_received(reference)
   end
 
   def chase_candidate_decision(application_form)
-    @dbd_date = application_form.application_choices.first.decline_by_default_at.to_s(:govuk_date).strip
     @application_choices = application_form.application_choices.select(&:offer?)
+    @dbd_date = @application_choices.first.decline_by_default_at.to_s(:govuk_date).strip
```
C-gyorfi: Good catch!  💯

PR #1445, opened 2020-02-20
Fix for applications export

tijmenb:
## Context

The application timings CSV is supposed to show the latest time a column was updated on an application form. However, the method looks at when the column was last updated including the time the record was created. Since all columns are present in the changeset of a create, we basically fall back to the created_at of a application form, making it look like all columns were updated.

Fix the logic, add better tests.

## Guidance to review

Read the new tests.

## Link to Trello card

https://trello.com/c/dKBwLMPv

## Things to check

### DIFF
--- a/spec/services/support_interface/applications_export_spec.rb
+++ b/spec/services/support_interface/applications_export_spec.rb
@@ -1,6 +1,6 @@
 require 'rails_helper'
 
-RSpec.describe SupportInterface::ApplicationsExport do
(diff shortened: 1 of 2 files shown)


spec/services/support_interface/applications_export_spec.rb
```diff
@@ -15,25 +15,26 @@
       create(:application_choice, status: 'awaiting_provider_decision', application_form: application_form, updated_at: '2019-01-10')
```
duncanjbrown: This 2019 date is a bit confusing

PR #1446, opened 2020-02-20
Add migration for diversity information to ApplicationForm

C-gyorfi:
## Context

We need to ask users for Equality and Diversity info and we need to store this information somehow. 

This PR adds fields to record equality and diversity on the application form

## Guidance to review

Does this make sense?

## Link to Trello card

https://trello.com/c/TTUZPTgz/206-candidates-can-provide-equality-and-diversity-information-build

## Things to check


db/migrate/20200220161956_add_equality_and_diversity_to_application_form.rb
```diff
@@ -0,0 +1,10 @@
+class AddEqualityAndDiversityToApplicationForm < ActiveRecord::Migration[6.0]
+  def change
+    change_table :application_forms do |t|
+      t.string :sex
+      t.string :disability
```
tvararu: This is a multiple choice with checkboxes, string would be a bit clunky to use I think. Should also probably be `disabilities`

tvararu: I'm wondering if it would be better to structure this data in the database using a JSON field, e.g.:

```json
{
  "sex": "male",
  "disabilities": ["blind", "deaf"],
  "ethnic_group": "asian",
  "ethnic_background": "chinese"
}
```

cc @tijmenb @duncanjbrown WDYT?

C-gyorfi: I like the idea of using a JSON field

PR #1449, opened 2020-02-20
Add ok-computer for monitoring

duncanjbrown:
## Context

We could write our own monitoring, but this gem looks ok (computer). It's an engine we mount (here under integrations/monitoring) which provides a nice JSON format for healthchecks. Successful checks return 200. Failed checks return 500.

Supported out of the box are Redis, AR, and sidekiq latency, [amongst other things](https://github.com/sportngin/okcomputer/tree/master/lib/ok_computer/built_in_checks).

If we want to do it ~to~ for ourselves, there is a sensible API for adding [custom checks](https://github.com/sportngin/okcomputer#registering-custom-checks).

No surprises, hopefully.

the output looks like this. 

Capture

## Guidance to review

Test by curling `integrations/monitoring/all`. There is also an endpoint without `all` which serves only the "app is serving requests" status and doesn't exercise any of the ~pigs in cages on antibiotics~ databases.

## Link to Trello card

https://trello.com/c/AdpL18Up/1655-add-one-stop-healthcheck-endpoint-to-the-app

## Things to check

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -42,6 +42,9 @@ gem 'uk_postcode'
 gem 'business_time'
 gem 'holidays'
 
+# Monitoring
+gem 'okcomputer'
+
 # feature flags
 gem 'rollout'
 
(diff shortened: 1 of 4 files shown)


duncanjbrown:
Thanks @tvararu! Adding @mnacos as he may be working on this next week...

PR #1452, opened 2020-02-21
Call rejection email outside of transaction

tvararu:
## Context

We think there is a race condition that occurs when we schedule an email for delivery inside this transaction, where the delivery occurs before the transaction has time to finalise and commit the changes to the database. This means that `decline_by_default_at` can be set to nil, which lead to a sentry error in production.

Remove the audit comment in favour of the new email log.

Move email delivery call outside of transaction.

## Guidance to review

There isn't a good way to test this, because it's a race condition; our tests do not have a realistic queue setup.

## Link to Trello card

No card, tactical fix.

## Things to check


tvararu:
This breaks multiple tests lol

PR #1456, opened 2020-02-24
Add filtering to the Provider UI

willmcb:
## Context

Adds ability to filter on status to the provider UI applications page. Ticket not complete done yet but I think this would be a good point to get some feedback on my work.

I also think this could probably be released as is as it would be sort of an MVP of filter functionality.

Functionality so far is:

- Ability to show and hide filter dialogue (defaults to show)
- Ability to filter on status.
- Sorting and hiding and showing respect state of filter on page reloads.
- Adds new informational message to users if they filter and it returns no results.

TODO:
- sorting on provider names
- clear button / link
- extra UI options to do with deselecting filters.

**Before:**
Screenshot 2020-02-24 at 09 46 02

**After:**
Screenshot 2020-02-24 at 09 48 36

## Guidance to review

Please take note of:
- the implementation of the state object I have used to encapsulate the state of the page. What do you think of this implementation?
- the for filtering is a plain form - Is it worth refactoring this to use a form for and a model that encapsulates it? I thought not given that checkboxes are binary and don't require validation as far as I can see.

## Link to Trello card

## Things to check


app/frontend/styles/_filter.scss
```diff
@@ -0,0 +1,31 @@
+.moj-action-bar__filter:after {
+  background: none;
+}
+
+@media (min-width: 48.0625em) {
```
mnacos: @adamsilver is there a way to rename these to `.dfe-...`? Not a huge deal but jars a little.

app/components/provider_interface/Toggle_filter_component.rb
```diff
@@ -0,0 +1,22 @@
...
+      @filter_visible = page_state.filter_visible
+      @filter_options = page_state.filter_options
+    end
+
+    def toggle_filter
+      @filter_visible.eql?('true') ? 'false' : 'true'
+    end
+
+    def toggle_button_text
+      @filter_visible.eql?('true') ? 'Hide filter' : 'Show filter'
```
mnacos: I find 'Show/Hide filter' misleading, it should probably be 'Show/Hide filtering options' or equivalent. Something for @laurahermionetennant92, perhaps?

spec/models/provider_interface/provider_applications_page_state_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+      params = { 'filter_visible' => 'false' }
+
+      state = described_class.new(params: params)
+
+      expect(state.filter_visible).to eq('false')
+    end
+  end
+
+  describe '#ordering_arguments' do
+    it 'returns nil if "course", "updated", or "name" not presend as sort_by values' do
```
tvararu:
```suggestion
    it 'returns nil if "course", "updated", or "name" not present as sort_by values' do
```

app/components/provider_interface/Toggle_filter_component.rb
```diff
@@ -0,0 +1,22 @@
+module ProviderInterface
```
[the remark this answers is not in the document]
willmcb: Thanks good spot will camel_case that file name. 

I was also sceptical of the `PageState` name happy to change so that it is more descriptive.

spec/support/have_submit_button_matcher.rb
```diff
@@ -0,0 +1,5 @@
+RSpec::Matchers.define :have_submit_button do |value|
+  match do
+    page.has_selector?("input[type=submit][value='#{value}']")
+  end
+end
```
tvararu: Could you define this as a method inside the spec where you want to use it?

```ruby
def have_submit_button(value)
  page.has_selector?("input[type=submit][value='#{value}']")
end
```

I don't think we should add new matchers if we can avoid it.

app/components/provider_interface/Toggle_filter_component.rb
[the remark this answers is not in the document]
tvararu: `snake_case` 😉

app/components/provider_interface/filter_component.html.erb
```diff
@@ -0,0 +1,40 @@
...
+                                    <%= render ProviderInterface::FilterCheckboxComponent, name: "recruited", text: "Conditions met", filter_options: filter_options %>
+                                    <%= render ProviderInterface::FilterCheckboxComponent, name: "declined", text: "Declined", filter_options: filter_options %>
+                                    <%= render ProviderInterface::FilterCheckboxComponent, name: "awaiting_provider_decision", text: "New", filter_options: filter_options %>
+                                    <%= render ProviderInterface::FilterCheckboxComponent, name: "offer", text: "Offered", filter_options: filter_options %>
+                                    <%= render ProviderInterface::FilterCheckboxComponent, name: "rejected", text: "Rejected", filter_options: filter_options %>
+                                    <%= render ProviderInterface::FilterCheckboxComponent, name: "withdrawn", text: "Application withdrawn", filter_options: filter_options %>
+                                    <%= render ProviderInterface::FilterCheckboxComponent, name: "offer_withdrawn", text: "Offer withdrawn", filter_options: filter_options %>
```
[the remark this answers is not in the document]
willmcb: Absolutely good idea.

app/frontend/styles/_provider.scss
```diff
@@ -0,0 +1,3 @@
+.govuk-width-container {
+    max-width: 1220px;
+}
```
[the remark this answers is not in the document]
willmcb: I spoke to @adamsilver about this as I questioned in the design why the container was larger than in the app. He said that they changed it to accommodate the filter being added to the side of the page.

app/models/provider_interface/provider_applications_page_state.rb
```diff
@@ -0,0 +1,33 @@
...
+  private
+
+    def extract_filter_options
+      # i.e. url params
+      if @params[:filter].is_a?(Array)
+        @params[:filter]
+      else
+        # i.e. form params
+        @params.fetch('filter', false) ? @params['filter']['status'].keys : []
+      end
```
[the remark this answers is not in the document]
willmcb: Nice its a lot clearer to read this. 👍

spec/support/have_submit_button_matcher.rb
willmcb: Happy to do this 👍 - out of curiosity why should we avoid defining new matchers?

tvararu: It can lead to harder to understand tests, because if there are many, you end up reading test files while constantly looking things up to see what they're actually doing under the hood.

app/components/provider_interface/filter_component.rb
```diff
@@ -0,0 +1,22 @@
...
+      @current_sort_by = page_state.sort_by
+      @filter_options = page_state.filter_options
+    end
+
+    def status_filter_checkbox_names_text
+      [
+        ["pending_conditions", "Accepted"], ["recruited", "Conditions met"],
+        ["declined", "Declined"], ["awaiting_provider_decision", "New"],
+        ["offer", "Offered"], ["rejected", "Rejected"],
+        ["withdrawn", "Application withdrawn"], ["offer_withdrawn", "Offer withdrawn"]
```
stevehook: Are the human-readable labels already in a YAML file somewhere? If not would that be a better place for them?

app/models/provider_interface/provider_applications_page_state.rb
```diff
@@ -0,0 +1,33 @@
+module ProviderInterface
+  class ProviderApplicationsPageState
+    attr_reader :sort_order, :sort_by, :filter_visible, :filter_options
+
+    def initialize(params:)
+      @params = params
+      @sort_order = params[:sort_order].eql?('asc') ? 'asc' : 'desc'
+      @sort_by = params[:sort_by].presence || 'last-updated'
+      @filter_visible = params['filter_visible'] ||= 'true'
+      @filter_options = extract_filter_options
```
stevehook: `filter_options` only stores `status` values so the name may be a bit confusing - in future when we add a second filter it would need to be restructured or renamed anyway. Maybe just call it `status_filter_values`?

app/models/provider_interface/provider_applications_page_state.rb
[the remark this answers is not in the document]
stevehook: I'm curious about why there are two ways to structure the params. Don't we always send params via the query string so isn't `has_url_params` always true?

Also wondering what will happen when filtering is extended to cover other attributes (e.g. `provider`) - I guess you would need the structure that you have in the `has_form_params` case?

duncanjbrown:
🎉 Successfully split up and merged as separate PRs. Closing

PR #1457, opened 2020-02-24
Enable support users to export referee survey answers

davidgisbey:
## Context

We need to be able to export our referee results to a CSV so that we can track referee satisfaction with the service and identify improvements.

- Add an export referee survey feature to the support interface

Before 

image

After

image

## Guidance to review

I'm going to leave a couple of comments on bits of the implementation that are not ideal. I'd appreciate any thoughts people have.

## Link to Trello card

https://trello.com/c/AsfA3tr3/

## Things to check


app/services/support_interface/referee_survey_export.rb
```diff
@@ -0,0 +1,30 @@
+module SupportInterface
+  class RefereeSurveyExport
+    def self.call
+      references = ApplicationReference.where.not(questionnaire: nil)
+      references_with_feedback = references.reject do |reference|
+        reference.questionnaire.values.all? { |response| response == ' | ' }
```
davidgisbey: This is ugly. At the moment if a referee does not give any feedback. The questionnaire attribute is updated to ' | ' for each question due to this 

image


I'm not sure whether changing this code so if they do not respond the questionnaire attributes value remains nil, then running a data migration to fix the current data is preferable.

davidgisbey: It also will be a lot less taxing on the db as we scale. Not sure if I'm overthinking something that will be sparsely used though.

tvararu: Will look at this with you tomorrow 👍

spec/system/support_interface/referee_survey_csv_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'rails_helper'
+
+RSpec.feature 'Referee survery CSV' do
+  include DfESignInHelpers
+
+  scenario 'support user can downloada CSV with the survey results' do
```
tvararu:
```suggestion
  scenario 'support user can download a CSV with the survey results' do
```

PR #1451, opened 2020-02-21
Candidates can opt out of filling up equality and diversity info

C-gyorfi:
## Context

We need to ask users for Equality and Diversity info. As a first slice - Users can opt to provide Equality and Diversity information.

## Changes proposed in this pull request

This PR adds
- A start page for Equality and Diversity questionnaire and the option to opt to provide Equality and Diversity information
- Feature flag

image

## Link to Trello card

https://trello.com/c/TTUZPTgz/206-candidates-can-provide-equality-and-diversity-information-build

## Things to check
- [x] This code doesn't rely on migrations in the same Pull Request
- [x] If this code includes a migration adding or changing columns, it also backfills existing records for consistency
- [x] API release notes have been updated if necessary
- [x] New environment variables have been [added to the Azure config](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training#azure-hosting-devops-pipeline)

PR #1458, opened 2020-02-24
Add ability to choose and update sex for equality and diversity

chubberlisk:
## Context

We want to ask candidates for Equality and Diversity information. They can currently opt out of doing this and this slice was done in #1451.

This PR:

- adds a form object for updating the sex attribute of equality and diversity field of an application
- adds a review component for reviewing and changing the answer

and wraps everything together so that a candidate can choose and update the field.

### Screenshots

image

image

image

## Guidance to review

Would like feedback/help on:

- adding the caption for the `What is your sex?` page. As we are using the formbuilder to render the legend, there doesn't seem to be a way to pass in `Equality and diversity` as a caption. Any way around this? (see below)

Screenshot 2020-02-24 at 17 05 44

- refactoring this:

```ruby
def self.build_from_application(application_form)
  sex = application_form.equality_and_diversity ? application_form.equality_and_diversity['sex'] : nil

  new(sex: sex)
end

```

- review component looks a bit weird because we only have one field

## Link to Trello card

https://trello.com/c/TTUZPTgz/206-candidates-can-provide-equality-and-diversity-information-build

## Things to check

### DIFF
--- a/app/components/candidate_interface/equality_and_diversity_review_component.html.erb
+++ b/app/components/candidate_interface/equality_and_diversity_review_component.html.erb
@@ -1,0 +1,1 @@
+<%= render(SummaryCardComponent.new(rows: equality_and_diversity_rows, border: false, editable: @editable)) %>
--- a/app/controllers/candidate_interface/equality_and_diversity_controller.rb
+++ b/app/controllers/candidate_interface/equality_and_diversity_controller.rb
@@ -1,5 +1,27 @@
 module CandidateInterface
   class EqualityAndDiversityController < CandidateInterfaceController
     def start; end
+
+    def edit_sex
+      @sex = EqualityAndDiversity::SexForm.build_from_application(current_application)
+    end
+
+    def update_sex
+      @sex = EqualityAndDiversity::SexForm.new(sex: sex_param)
+
+      if @sex.save(current_application)
+        redirect_to candidate_interface_review_equality_and_diversity_path
+      else
+        render :edit_sex
+      end
+    end
+
+    def review; end
+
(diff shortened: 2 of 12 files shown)


tvararu:
Why is the form missing borders? :spinthinking:

Screenshot 2020-02-24 at 17 28 39

app/components/candidate_interface/equality_and_diversity_review_component.html.erb
```diff
@@ -0,0 +1 @@
+<%= render(SummaryCardComponent.new(rows: equality_and_diversity_rows, border: false, editable: @editable)) %>
```
tvararu: `border: false` 🤔 is that the design? If yes ignore me 👍

app/controllers/candidate_interface/equality_and_diversity_controller.rb
```diff
@@ -1,5 +1,27 @@
...
+      @sex = EqualityAndDiversity::SexForm.new(sex: sex_param)
+
+      if @sex.save(current_application)
+        redirect_to candidate_interface_review_equality_and_diversity_path
+      else
+        render :edit_sex
+      end
+    end
+
+    def review; end
```
tvararu: I think putting everything into this controller should work nicely 👍

app/components/candidate_interface/equality_and_diversity_review_component.html.erb
chubberlisk: Yush, also looks weird atm because we only have one field.

image

PR #1462, opened 2020-02-25
Add the 'You selected a course' page

stevehook:
## Context

When we built the Find to Apply feature we didn't include the "You selected a course" step, which should show after signing in or signing up. This page

- [x] Add new 'You selected a course' page via new `FindCourseSelectionsController`
- [x] Update the logic in `SignInController#interstitial` to route to the new page when we are about to add a new course from find
- [x] Factored the 'route selection' logic from `AddCourseFromFind` into a new service `InterstitialRouteSelector` (needs a better name).
- [x] Updated system specs to include new step
- [x] Introduced a new system spec to handle the 'No' option from the new course selection page
- [x] Feature flag?
- [x] Tidy up the styling in the new page (esp the link)

## Guidance to review

- I struggled a bit with the logic that starts in `SignInController#interstitial` and gets implemented in the `AddCourseFromFind` and `PickSiteForm` services. One problem is that `AddCourseFromFind` mixes up the where to redirect to decision with actually adding application choices and clearing the course selected in Find. I've tried to make this a bit easier to reason about by adding `InterstitialRouteSelector` but this is definitely WIP.
- The styling of the new page.
- 

## Link to Trello card

https://trello.com/c/c4ZYOuaI/1010-add-you-selected-a-course-page-into-find-to-apply-journey

## Things to check


app/controllers/candidate_interface/find_course_selections_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+      if !course_selection.confirm
+        redirect_to candidate_interface_course_choices_index_path
+        return
+      end
+
+      # TODO: refactor this into a service etc.?
+      if CourseOption.where(course_id: course.id).one?
+        course_option = CourseOption.where(course_id: course.id).first
+
+        pick_site_for_course(course, course_option.id)
```
davidgisbey: I think we probably still need to give them the option to apply via UCAS?

davidgisbey: so the flow would be confirm_selection ----> apply on ucas or apply ------> interstitial page/action

davidgisbey:
Looking good! I've just left a comment on the desired user flow. I may be wrong, but I think we still need to give the candidate the option to apply on UCAS.

app/controllers/candidate_interface/find_course_selections_controller.rb
stevehook: thanks @davidgisbey, that's an unintended side-effect, I'll reinstate that...

stevehook: Just had a look at this by testing manually and I still see the apply on new service or UCAS page if I hit the site with something like `/candidate/apply?providerCode=1N1&courseCode=238T` whether I am logged in or not - it's still implemented in `CandidateInterface::ApplyFromFindController`. So we are still giving the option but it's at the start of the process not after the new confirm selection page like it was before.

app/controllers/candidate_interface/find_course_selections_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      course = Course.find(params[:course_id])
+
+      course_selection = CourseSelectionForm.new(course, course_selection_params[:confirm])
+      if !course_selection.confirm
+        redirect_to candidate_interface_course_choices_index_path
+        return
+      end
+
+      # TODO: refactor this into a service etc.?
```
tvararu:
```suggestion
```

Don't think we need to yet

PR #1464, opened 2020-02-25
Set course choices section to incomplete when a course choice is added or deleted

davidgisbey:
## Context

Currently, when the course choices section is marked as complete, the candidate can add or delete courses without it setting the section to incomplete.

This means that a candidate can delete all their course choices and submit their application as our validations only check if all the sections have been completed.

-  When a candidate adds a new course the application_choices_compelted attribute on the application form is set to false
- When a candidate deletes a course the application_choices_compelted attribute on the application form is set to false

## Guidance to review

This is the first of probably quite a few PRs. After discussion with Theo, anytime a candidate updates something in a section the completed status should default to incomplete.

## Link to Trello card

https://trello.com/c/wrAbxDA8/

## Things to check

### DIFF
--- a/spec/system/candidate_interface/candidate_deletes_course_after_completing_section_spec.rb
+++ b/spec/system/candidate_interface/candidate_deletes_course_after_completing_section_spec.rb
@@ -1,0 +1,86 @@
+require 'rails_helper'
+
+RSpec.feature 'Candidate edits their choice section' do
+  include CandidateHelper
+
+  scenario 'Candidate deletes and adds additonal courses' do
+    given_i_am_signed_in
+    and_i_have_completed_the_course_choice_section
+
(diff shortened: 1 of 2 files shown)


spec/system/candidate_interface/candidate_deletes_course_after_completing_section_spec.rb
```diff
@@ -0,0 +1,86 @@
+require 'rails_helper'
+
+RSpec.feature 'Candidate edits their choice section' do
+  include CandidateHelper
+
+  scenario 'Candidate deletes and adds additonal courses' do
```
tvararu:
```suggestion
  scenario 'Candidate deletes and adds additional courses' do
```

PR #1469, opened 2020-02-26
1665 remove support users

steventux:
## Context

We can onboard support users via the UI but cannot remove them.

Adds delete buttons to the support users index per user. This takes the administrator to a confirmation screen before actual record deletion.

### Before

image

### After

image

(confirmation screen...)

image

## Guidance to review

@tvararu and I discussed putting this behind a feature flag purely to get used to the practice, arguably this feature isn't large enough to warrant it.

## Link to Trello card

## Things to check

### DIFF
--- a/app/controllers/support_interface/support_users_controller.rb
+++ b/app/controllers/support_interface/support_users_controller.rb
@@ -19,6 +19,16 @@ module SupportInterface
       end
     end
 
+    def confirm_destroy
+      @support_user = SupportUser.find(params[:id])
+    end
+
+    def destroy
+      SupportUser.find(params[:id]).destroy!
+      flash[:success] = 'Support user deleted'
+      redirect_to support_interface_support_users_path
+    end
+
   private
 
     def support_user_params
--- a/app/views/support_interface/support_users/index.html.erb
+++ b/app/views/support_interface/support_users/index.html.erb
@@ -25,6 +25,7 @@
     <tr class='govuk-table__row'>
       <th scope='col' class='govuk-table__header govuk-!-width-one-half'>Email</th>
       <th scope='col' class='govuk-table__header govuk-!-width-one-quarter'>DfE Sign-in UID</th>
+      <th scope='col' class='govuk-table__header govuk-!-width-one-quarter'>Actions</th>
     </tr>
   </thead>
 
@@ -36,6 +37,14 @@
         </td>
         <td class='govuk-table__cell'>
           <%= support_user.dfe_sign_in_uid %>
+        </td>
+        <td class='govuk-table__cell'>
+          <%= link_to(
+                support_interface_confirm_destroy_support_user_path(support_user),
+                class: 'govuk-link',
+          ) do %>
+            Delete user<span class='govuk-visually-hidden'> <%= support_user.email_address %></span>
+          <% end %>
         </td>
       </tr>
     <% end %>
--- a/spec/system/support_interface/remove_support_users_spec.rb
+++ b/spec/system/support_interface/remove_support_users_spec.rb
@@ -1,0 +1,51 @@
+require 'rails_helper'
+
+RSpec.feature 'Remove a support user' do
+  include DfESignInHelpers
+
+  scenario 'Confirming removal of a support user' do
+    given_i_am_a_support_user
+    and_there_are_some_support_users
+    when_i_visit_the_support_users_page
+    and_i_delete_a_support_user
+    then_i_should_see_a_confirmation_page
+    and_i_confirm_deletion
+    then_the_support_user_is_deleted_from_the_database
+    and_the_support_user_is_not_listed
+  end
+
+  def given_i_am_a_support_user
+    sign_in_as_support_user
+  end
+
+  def and_there_are_some_support_users
+    @support_users = create_list(:support_user, 2)
+    @deleted_user_email = @support_users.last.email_address
+  end
+
+  def when_i_visit_the_support_users_page
+    visit support_interface_support_users_path
+  end
+
+  def and_i_delete_a_support_user
+    within('.govuk-table__body') do
+      click_on("Delete user #{@deleted_user_email}")
+    end
+  end
+
+  def then_i_should_see_a_confirmation_page
+    expect(page.text).to include("Are you sure you want to delete support user #{@deleted_user_email}?")
+  end
+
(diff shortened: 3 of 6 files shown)


app/views/support_interface/support_users/index.html.erb
```diff
@@ -37,6 +40,14 @@
         <td class='govuk-table__cell'>
           <%= support_user.dfe_sign_in_uid %>
         </td>
+        <% if FeatureFlag.active?('delete_support_users') %>
+          <td class='govuk-table__cell'>
+            <%= link_to 'Delete',
```
fofr: I think this should just be a link, "Delete user", rather than a button.

app/controllers/support_interface/support_users_controller.rb
```diff
@@ -19,6 +19,18 @@ def create
       end
     end
 
+    def confirm_destroy
```
stevehook: Return 404 if feature flag is not active?

app/controllers/support_interface/support_users_controller.rb
```diff
@@ -19,6 +19,18 @@ def create
...
+      @support_user = SupportUser.find(params[:id])
+    end
+
+    def destroy
+      if FeatureFlag.active?('delete_support_users')
```
stevehook: Return 404 if feature flag is not active?

app/views/support_interface/support_users/index.html.erb
tvararu: This would benefit from some hidden text such as "Delete user <span class="govuk-visually-hidden">foo@example.com</span>" for accessibility 👍

spec/requests/support_interface/support_users_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'rails_helper'
```
tvararu: What do you think about writing a system spec instead of a request spec?

app/controllers/support_interface/support_users_controller.rb
stevehook: ...or as @tvararu suggested we could just drop the feature flag

steventux:
@tvararu Just noticed that it's possible to delete all support users including the one in session. There's no specific A/C around this but would it be sensible to guard against removing the user in session?

app/views/support_interface/support_users/index.html.erb
```diff
@@ -37,6 +38,11 @@
         <td class='govuk-table__cell'>
           <%= support_user.dfe_sign_in_uid %>
         </td>
+        <td class='govuk-table__cell'>
+          <%= link_to(support_interface_confirm_destroy_support_user_path(support_user)) do %>
```
chubberlisk: We should use a `govuk-link`, there's a helper called `ViewHelper#govuk_link_to` for this. 😉 

```suggestion
          <%= govuk_link_to(support_interface_confirm_destroy_support_user_path(support_user)) do %>
```

spec/system/support_interface/remove_support_users_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  scenario 'Confirming removal of a support user' do
+    given_i_am_a_support_user
+    and_there_are_some_support_users
+    and_i_visit_the_support_users_index_page
+    and_i_delete_a_support_user
+    then_i_should_see_a_confirmation_page
+    and_i_confirm_deletion
+    then_the_support_user_is_deleted_from_the_database
+    and_the_support_user_is_not_listed
```
chubberlisk: Suggestion to improve readability of the system spec:

```
given_i_am_a_support_user
and_there_are_some_support_users

when_i_visit_the_support_users_page
and_i_delete_a_support_user
then_i_should_see_a_confirmation_page

when_i_confirm_deletion
then_the_support_user_is_deleted_from_the_database
and_the_support_user_is_not_listed
```

What do you think?

app/controllers/support_interface/support_users_controller.rb
steventux: I've removed the feature flag commit, was useful to learn how to use the `rollout` gem in this app.

app/views/support_interface/support_users/index.html.erb
steventux: @chubberlisk I think for this to work with a block we would need to change `govuk_link_to` to have a similar method signature to `link_to`

```ruby
 def link_to(name = nil, options = nil, html_options = nil, &block)
```
For the sake of expediency I'll add the `govuk-link` css class to the html_options.

chubberlisk: Ah yush! Sorry, I didn't realise that it was a block. 🤦‍♀ 

Makes sense, do that.

tvararu:
> @tvararu Just noticed that it's possible to delete all support users including the one in session. There's no specific A/C around this but would it be sensible to guard against removing the user in session?

Hmm, good spot, I think it's fine to allow self-deletion. 👍

app/views/support_interface/support_users/index.html.erb
steventux: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1469/files#diff-e231e5182cc5a11d49224019f20a33a8R46

app/views/support_interface/support_users/index.html.erb
```diff
@@ -37,6 +38,14 @@
         <td class='govuk-table__cell'>
           <%= support_user.dfe_sign_in_uid %>
         </td>
+        <td class='govuk-table__cell'>
+          <%= link_to(
```
fofr:
```suggestion
          <%= govuk_link_to(
```

app/views/support_interface/support_users/index.html.erb
```diff
@@ -37,6 +38,14 @@
...
+                support_interface_confirm_destroy_support_user_path(support_user),
+                class: 'govuk-link',
```
fofr: (We've got a helper that includes this)

app/views/support_interface/support_users/index.html.erb
steventux: @fofr sadly `govuk_link_to` doesn't accept a block, we'd need to update this helper to have the same functionality / signature as the ActionView `link_to` helper.

PR #1472, opened 2020-02-26
Add pagination to the Provider interface

stevehook:
## Context

As we scale up to Providers that manage larger numbers of applications we will need to paginate the list of applications making it easier to manage.

This work is closely related to https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1456 but not really dependent on it.

- [x] Add kaminari gem to make pagination a bit easier
- [x] Add styles for pagination from MoJ design system
- [x] Generate custom views for the kaminari paginator so that we can implement the design
- [x] Add a `PaginatorComponent` that wraps the standard `kaminari` `paginate` helper and our own custom output mainly to show the  _Showing 26 to 29 of 29 results_ message within the markup specified by the design system. This component should be easy to drop in anywhere else we need pagination.
- [x] Put this behind a `provider_interface_pagination` feature flag. 

## Guidance to review

- Is Kaminari the right gem to use? We could do with a little more work.
- What should the page size be? It's just using the default 25 at the moment.
- Paging links will sit outside the filter form but we can achieve this mostly by adding the filter params to those links when they are generated server-side. The only snag here is that a user could select from filters, not press the Apply filters button and then click a paging link losing their most recent filter selections.
- Does adding a component make sense? 
- Note that the tests for the component deliberately stub out the Kaminari `paginate` method so that we can focus on testing the custom bits rather than retest the gem.

## Link to Trello card

https://trello.com/c/939aTO5R/1637-add-pagination-to-the-provider-ui-applications-index-%F0%9F%8F%88

## Things to check


app/components/paginator_component.rb
```diff
@@ -0,0 +1,26 @@
+class PaginatorComponent < ActionView::Component::Base
+  attr_reader :scope
+
+  def initialize(scope:)
+    @scope = scope
+  end
+
+  def render?
```
mnacos: Nice

PR #1474, opened 2020-02-26
Amend govuk_link_to so that it accepts a block

steventux:
## Context

We discovered in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1469 that `govuk_link_to` does not accept a block argument in the same way that ActionView `link_to` does.

This PR allows `govuk_link_to` to accept and yield a block. This is useful for more complex link text containing child elements.

Usage:

```ruby
<%= govuk_link_to(nil, root_path) do %>
  This link<span class="govuk-visually-hidden"> contains accessible text</span>
<% end %>
```

renders as:

```html
<a href="/" class="govuk-link">This link<span class="govuk-visually-hidden"> contains accessible text</span></a>
```

## Things to check

### DIFF
--- a/app/helpers/view_helper.rb
+++ b/app/helpers/view_helper.rb
@@ -1,6 +1,8 @@
 module ViewHelper
-  def govuk_link_to(body, url, html_options = {})
+  def govuk_link_to(body, url, html_options = {}, &_block)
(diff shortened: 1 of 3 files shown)


chubberlisk:
> @chubberlisk will be happy 👍

Chubberlisk _**is**_ happy. 😸

PR #1476, opened 2020-02-26
Performance dashboard improvements

benilovj:
## Context

There is lots of duplicate view code in the performance dashboard.

- fix some broken HTML
- extract a tile component that can handle all the flavours of tile that's currently being used in the dashboard
- add component previews

## Guidance to review

image
image
image
image
image

## Things to check

### DIFF
--- a/app/components/support_interface/tile_component.html.erb
+++ b/app/components/support_interface/tile_component.html.erb
@@ -1,4 +1,4 @@
-<div class="app-card">
+<div class="app-card<%= colour == :default ? '' : " app-card--#{colour}" %>">
   <span class="app-card__count"><%= count %></span>
   <%= label %>
 </div>
--- a/app/components/support_interface/tile_component.html.erb
+++ b/app/components/support_interface/tile_component.html.erb
@@ -1,0 +1,4 @@
+<div class="app-card">
+  <span class="app-card__count"><%= count %></span>
+  <%= label %>
+</div>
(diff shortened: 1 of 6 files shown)


app/components/support_interface/tile_component.html.erb
```diff
@@ -0,0 +1,4 @@
+<div class="app-card<%= !regular_type? ? " app-card--#{type_colour}" : '' %>">
+  <span class="<%= count_class %>"><%= count %></span>
+  <span><%= label %></span>
```
tvararu:
```suggestion
  <%= label %>
```

Unnecessary span? Can't find CSS that targets it.

app/components/support_interface/tile_component.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+
+    def type_colour
+      case @tile_type
+      when :headline then 'blue'
+      when :success then 'green'
+      when :error then 'red'
+      else ''
+      end
+    end
```
tvararu: `TagComponent` was changed to not use a semantic mapping and instead just ask the user to pass in the desired colour: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1018

I'm impartial to either approach and think either can work 👍

app/components/support_interface/tile_component.html.erb
benilovj: Thanks, I've removed this now.

PR #1480, opened 2020-02-27
Pre-check standard conditions for new offers

steventux:
## Context

Standard conditions are compulsory for a trainee to start a course and are usually included
in an offer. Providers can choose to omit these conditions but they should be included by default.

Standard conditions are pre-checked when a provider makes a new offer. The provide can omit them by unchecking the condition(s).

### Before

image

### After

image

## Guidance to review

I considered moving the repetition of checkbox values (they are pretty wordy) to a translation file. The use and guidance of translation files is under discussion though, would be happy to hear reviewer's thoughts on this. It would be DRYer to use translations but given these are form checkbox values I wonder if it's the correct use of i18n?

## Link to Trello card

https://trello.com/c/MpKzZ1HO/1698-make-default-conditions-not-optional-when-making-an-offer-in-the-provider-ui

## Things to check

### DIFF
--- a/app/services/make_an_offer.rb
+++ b/app/services/make_an_offer.rb
@@ -7,6 +7,7 @@ class MakeAnOffer
 
   MAX_CONDITIONS_COUNT = 20
   MAX_CONDITION_LENGTH = 255
+  STANDARD_CONDITIONS = ['Fitness to Teach check', 'Disclosure and Barring Service (DBS) check'].freeze
 
   validate :validate_course_data
   validate :validate_conditions_max_length
@@ -16,7 +17,7 @@ class MakeAnOffer
     actor:,
     application_choice:,
     offer_conditions: nil,
-    standard_conditions: nil,
+    standard_conditions: STANDARD_CONDITIONS,
     further_conditions: {},
     course_data: nil
   )
--- a/app/views/provider_interface/decisions/new_offer.html.erb
+++ b/app/views/provider_interface/decisions/new_offer.html.erb
@@ -24,7 +24,7 @@
             standard_conditions_checkboxes,
             :id,
             :name,
-            legend: { text: 'Standard conditions (optional)', size: 'm', tag: 'h2' },
+            legend: { text: 'Standard conditions', size: 'm', tag: 'h2' },
       ) %>
 
       <fieldset class="govuk-fieldset">
--- a/app/views/provider_interface/decisions/new_offer.html.erb
+++ b/app/views/provider_interface/decisions/new_offer.html.erb
@@ -19,10 +19,13 @@
 
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-      <%= f.govuk_check_boxes_fieldset :standard_conditions, legend: { text: 'Standard conditions (optional)', size: 'm', tag: 'h2' } do %>
-          <%= f.govuk_check_box :standard_conditions, 'Fitness to Teach check', label: { text: 'Fitness to Teach check' } %>
-          <%= f.govuk_check_box :standard_conditions, 'Disclosure and Barring Service (DBS) check', label: { text: 'Disclosure and Barring Service (DBS) check' } %>
-      <% end %>
+      <%= f.govuk_collection_check_boxes(
+            :standard_conditions,
+            standard_conditions_checkboxes,
+            :id,
+            :name,
+            legend: { text: 'Standard conditions (optional)', size: 'm', tag: 'h2' },
+      ) %>
 
       <fieldset class="govuk-fieldset">
         <legend class="govuk-fieldset__legend govuk-fieldset__legend--m">Further conditions (optional)</legend>
--- a/app/helpers/checkbox_options_helper.rb
+++ b/app/helpers/checkbox_options_helper.rb
@@ -8,4 +8,13 @@ module CheckboxOptionsHelper
       )
     end
   end
+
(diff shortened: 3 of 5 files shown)


app/controllers/provider_interface/decisions_controller.rb
```diff
@@ -22,6 +22,7 @@ def new_offer
       @application_offer = MakeAnOffer.new(
         actor: current_provider_user,
         application_choice: @application_choice,
+        standard_conditions: ['Fitness to Teach check', 'Disclosure and Barring Service (DBS) check'],
```
mnacos: Seems reasonable but may I suggest a small refactoring. How about moving these pre-checked standard conditions to a constant within `MakeAnOffer`, which then becomes the source of truth for what conditions are considered standard, instead of defining this in the controller.

You could then change `MakeAnOffer#initialize` to use that constant as a default value instead of `nil` and drop this line from the controller. There may be some tests that fail as a result (_if_ we have system spec exercising making an offer with no conditions, for example), but it'd be interesting to see what breaks.

steventux: @mnacos good point https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1480/commits/e854c7fe6d259dd7d461f2fa40225aa0fbb7ced6 makes those values defaults in the `MakeAnOffer` constructor as advised.

app/services/make_an_offer.rb
```diff
@@ -16,7 +16,7 @@ def initialize(
     actor:,
     application_choice:,
     offer_conditions: nil,
-    standard_conditions: nil,
+    standard_conditions: ['Fitness to Teach check', 'Disclosure and Barring Service (DBS) check'],
```
mnacos: Was hoping for something stubbable, such as
```ruby
  DEFAULT_STANDARD_CONDITIONS = ['Fitness to Teach check', 'Disclosure and Barring Service (DBS) check'].freeze
```
What do you think?

steventux:
@mnacos @stevehook thanks for the reviews, I think I've addressed all points now.
I wonder if reusing `MakeAnOffer::STANDARD_CONDITIONS` to populate the values of the checkboxes might make the value definitions less brittle.

mnacos:
> @mnacos @stevehook thanks for the reviews, I think I've addressed all points now.
> See [b329c48#diff-541dbd1e1d2e8eee85a7dd9c76c6b235R20](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/commit/b329c4858de95625177c12a767c516c7433fada6#diff-541dbd1e1d2e8eee85a7dd9c76c6b235R20)
> and
> [b329c48#diff-002f3970d414b723cc19c694b0d9c902R62](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/commit/b329c4858de95625177c12a767c516c7433fada6#diff-002f3970d414b723cc19c694b0d9c902R62)
> 
> I wonder if reusing `MakeAnOffer::STANDARD_CONDITIONS` to populate the values of the checkboxes might make the value definitions less brittle.
> 
> eg
> 
> ```
> <% MakeAnOffer::STANDARD_CONDITIONS.each do |condition| %>                                                                                                  
>   <%= f.govuk_check_box :standard_conditions, condition, label: { text: condition } %>                                                               
> <% end %>
> ```

That's a good idea and consistent with having the service itself be the source of truth for the default standard conditions. I'd probably use `@application_offer.standard_conditions` instead of the constant, but not too bothered.

Please add :)

app/views/provider_interface/decisions/new_offer.html.erb
```diff
@@ -19,10 +19,13 @@
...
-      <%= f.govuk_check_boxes_fieldset :standard_conditions, legend: { text: 'Standard conditions (optional)', size: 'm', tag: 'h2' } do %>
-          <%= f.govuk_check_box :standard_conditions, 'Fitness to Teach check', label: { text: 'Fitness to Teach check' } %>
-          <%= f.govuk_check_box :standard_conditions, 'Disclosure and Barring Service (DBS) check', label: { text: 'Disclosure and Barring Service (DBS) check' } %>
-      <% end %>
+      <%= f.govuk_collection_check_boxes(
+            :standard_conditions,
+            standard_conditions_checkboxes,
+            :id,
+            :name,
+            legend: { text: 'Standard conditions (optional)', size: 'm', tag: 'h2' },
```
mnacos: should this say 'optional'?

steventux: Hmm I haven't changed this text from the original copy. They are optional in that there is no validation to ensure these items are selected but I wonder about the wording...

steventux:
@mnacos I forced pushed since your approval, this was to make the test for the check box data less brittle https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1480/files#diff-935dc2f0c5c19b13b4a65dc6608f24b0R19-R25

app/views/provider_interface/decisions/new_offer.html.erb
steventux: @mnacos https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1480/commits/d1d0ce58bcb7eb33c58ff82c9735268f3b68c24d

mnacos: Thanks, I think it's good to merge.

PR #1484, opened 2020-02-27
1057 Pick Full Time or Part Time Course Options

malcolmbaig:
## Context

Some courses are available in a part-time study mode. This change adds the ability to select study mode for those courses that present an option.

- Add a study mode selection page after the candidate has chosen a course.
- Only show this screen if there is actually an option - where there isn't, resolve the study mode for that course automatically and simply display the next step (site selection, or the review page in cases where there's only one site).

Study mode selection page is as follows:

Screenshot 2020-02-27 at 15 25 36

Followed by site selection, filtered to only show sites that support the selected study mode:

Screenshot 2020-02-27 at 15 32 21

## Guidance to review

- I'd suggest pulling the branch down and clicking through locally. You may need to change a specific course and its course options to see the new page. Specifically, `Course#study_mode` needs to be `'full_time_or_part_time'`, and it should have at least two course options with different `study_mode` values set (`part_time` and `full_time`).
- Per-commit review may be useful as each commit describes a distinct part of the feature.

**Note** - the site selection functionality shown above should probably be modified so that if only one choice is available, we select this automatically and move the candidate on. This PR is probably large enough as it stands - once reviewed I'll write a followup PR to add this functionality.

## Link to Trello card

https://trello.com/c/JnW7TR8k

## Things to check

### DIFF
--- a/spec/system/candidate_interface/candidate_selecting_a_course_study_mode_spec.rb
+++ b/spec/system/candidate_interface/candidate_selecting_a_course_study_mode_spec.rb
@@ -4,6 +4,7 @@ RSpec.feature 'Selecting a study mode' do
   include CandidateHelper
 
   scenario 'Candidate selects a part time course' do
+    given_choosing_a_study_mode_is_active
     given_i_am_signed_in
     and_there_are_course_options
 
@@ -12,6 +13,10 @@ RSpec.feature 'Selecting a study mode' do
 
     when_i_select_a_site
     then_i_see_my_course_choice
+  end
+
+  def given_choosing_a_study_mode_is_active
+    FeatureFlag.activate('choose_study_mode')
   end
 
   def given_i_am_signed_in
--- a/spec/system/candidate_interface/candidate_selecting_a_course_study_mode_spec.rb
+++ b/spec/system/candidate_interface/candidate_selecting_a_course_study_mode_spec.rb
@@ -1,0 +1,78 @@
+require 'rails_helper'
+
+RSpec.feature 'Selecting a study mode' do
+  include CandidateHelper
+
+  scenario 'Candidate selects a part time course' do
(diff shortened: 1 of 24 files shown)


spec/system/candidate_interface/candidate_selecting_a_course_study_mode_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+RSpec.feature 'Selecting a study mode' do
+  include CandidateHelper
+
+  scenario 'Candidate selects a part time course' do
+    given_i_am_signed_in
+    and_there_are_course_options
+    when_i_select_a_part_time_course
+    then_i_can_only_select_sites_with_a_part_time_course
+    when_i_select_a_site
+    then_i_see_my_course_choice
```
tvararu:
```suggestion
    given_i_am_signed_in
    and_there_are_course_options

    when_i_select_a_part_time_course
    then_i_can_only_select_sites_with_a_part_time_course

    when_i_select_a_site
    then_i_see_my_course_choice
```

malcolmbaig:
> Everything looks 👍 to me, but this should be behind a feature flag so we can ship it and test it more easily.
> 
> Have also tested it on the review app.

Good point, will add one in.

malcolmbaig:
@tvararu I've opted to not hide the route changes in order to reduce the amount of feature flag noise in code. To quote the commit message:

> [this commit] does not hide the route changes or the updated method signature of
the site selection page path helper. These are largely cosmetic changes
from the user's perspective and do not alter pre-existing functionality.

This means the site selection page will have a `study_mode` in its URL, but that won't alter the behavior of the page. WDYT?

tvararu:
@malcolmbaig 👍

PR #1495, opened 2020-03-02
Add guidance to the provider sign in page when DSI has an incident

duncanjbrown:
Because the DSI error page is very unhelpful, warn our users before they
get there. This notice can be toggled via feature flag.

Screenshot 2020-03-02 at 10 44 13

### Context

https://ukgovernmentdfe.slack.com/archives/CAHBLU6ET/p1583141874153800

### Guidance to review

I chose not to include a test...

### Link to Trello card

https://trello.com/c/R1oZNMdc/1717-tell-providers-dfe-sign-in-is-having-an-incident

### DIFF
--- a/app/services/feature_flag.rb
+++ b/app/services/feature_flag.rb
@@ -26,6 +26,7 @@ class FeatureFlag
     equality_and_diversity
     provider_application_filters
     provider_interface_pagination
+    dfe_sign_in_incident
   ].freeze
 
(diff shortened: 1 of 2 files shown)


duncanjbrown:
Good point @tvararu. I'll see if there's a place to add this in the docs

duncanjbrown:
@emmajhf approved via Slack. Merging

PR #1506, opened 2020-03-02
[DNM] 1665 soft delete support users

steventux:
## Context

Depends on https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1504

We've recently added the ability to delete support users, but we need to retain the records for the audit history to work correctly so soft deletion is the better option.

- Adds the `paranoia` gem
- Applies soft deletion to `SupportUser` model

## Link to Trello card

## Things to check


duncanjbrown:
This looks great, but I notice the author of `paranoia` has deprecated his own gem! Should we consider `discard` instead? https://github.com/rubysherpas/paranoia/blob/core/README.md  https://github.com/jhawthorn/discard#why-not-paranoia-or-acts_as_paranoid

duncanjbrown:
Looking good! Comment above and inline

steventux:
Closing for now to find an alternative to using `paranoia` gem

tvararu:
_adds `paranoia` to his list of weirdly named ruby libraries_

PR #1507, opened 2020-03-02
Set the course choice drop downs default value to nil

davidgisbey:
## Context

At the moment if a user accidentally readds the same course the select retains the course's name and code.

This means the user has to manually delete the course's name before more options
become available to them.

- Set the selected value of drop down to nil

Before

image

After 

image

## Guidance to review

Add a course locally. Attempt to add the same course again.

## Link to Trello card

https://trello.com/c/XaJkYmar/

## Things to check


fofr:
The intended fix was that the original course name would be shown in the error message.

PR #1508, opened 2020-03-02
Resolve error when candidate edits their course choices after submission

C-gyorfi:
## Context
Deleting a course and adding a new course after submission causes an argument error:

image
This is because newly created application choices are in an `unsubmitted` state with `edit_by` set to `nil`. They should be in an `awaiting_references` state with `edit_by` set to the correct value.

This PR resolves the error by setting the `edit_by` date on course choices before completing the course section.

## Link to Trello card
https://trello.com/c/u9fFktHo/1048-application-form-errors-when-edit-application-feature-flag-is-on-and-user-edits-their-course-choices

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/course_choices_controller.rb
+++ b/app/controllers/candidate_interface/course_choices_controller.rb
@@ -146,9 +146,13 @@ module CandidateInterface
     def complete
       @application_form = current_application
 
-      if @application_form.application_choices.count.zero?
-        render :index
-      elsif @application_form.update(application_form_params)
+      render :index if @application_form.application_choices.count.zero?
+
+      if @application_form.submitted?
+        @application_form.application_choices.filter(&:unsubmitted?).each { |choice| SubmitApplicationChoice.new(choice).call }
+      end
+
+      if @application_form.update(application_form_params)
         redirect_to candidate_interface_application_form_path
       else
         @course_choices = current_candidate.current_application.application_choices
--- a/app/services/submit_application_choice.rb
+++ b/app/services/submit_application_choice.rb
@@ -1,0 +1,21 @@
+class SubmitApplicationChoice
+  def initialize(application_choice)
+    @application_choice = application_choice
+  end
+
+  def call
+    edit_by_time = time_limit_calculator.call[:time_in_future]
+    @application_choice.edit_by = edit_by_time
+    ApplicationStateChange.new(@application_choice).submit!
+  end
+
+private
+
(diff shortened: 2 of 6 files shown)


app/controllers/candidate_interface/course_choices_controller.rb
```diff
@@ -146,9 +146,14 @@ def destroy
...
       @application_form = current_application
 
-      if @application_form.application_choices.count.zero?
-        render :index
-      elsif @application_form.update(application_form_params)
+      render :index if @application_form.application_choices.count.zero?
+
+      if @application_form.submitted?
+        edit_by_time = TimeLimitCalculator.new(rule: :edit_by, effective_date: @application_form.submitted_at).call[:time_in_future]
+        @application_form.application_choices.each { |choice| choice.update!(edit_by: edit_by_time) }
```
C-gyorfi: Maybe we should call the `SubmitApplication` service here? I just noticed the state is still unsubmitted, not sure if it would cause other issues. Any suggestions?

duncanjbrown: This is an interesting one. I wonder if we should refactor `SubmitApplication` to call a `SubmitApplicationChoice` service, and call that service from here too?

C-gyorfi: Yes, I had a similar idea in my head this morning! I'll have a look

C-gyorfi: I made some changes, IMO it helped, what do you think? @duncanjbrown

app/services/submit_application_choice.rb
```diff
@@ -0,0 +1,23 @@
+class SubmitApplicationChoice
+  def initialize(application_choice)
+    @application_choice = application_choice
+  end
+
+  def call
+    return unless @application_choice.unsubmitted?
```
C-gyorfi: @duncanjbrown  I've written a test as discussed and added this line to prevent re-submitting existing application choices, good catch 🎣

duncanjbrown: This class already handles this condition by raising an error on `submit!`. imo it's the caller's responsibility not to give bad input. Maybe something like `@application_form.application_choices.filter(&:unsubmitted?).each { |choice| SubmitApplicationChoice.new(choice).call }`?

C-gyorfi: Make sense 😄

PR #1517, opened 2020-03-03
1665 Soft delete support users

steventux:
## Context

We currently delete `SupportUser` records which doesn't play nicely with our audit trail. 

See https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1516 for context on why we intend to use the `discard` gem for soft deletion.

This PR: 

- Introduces the `discard` gem as a lightweight approach to discarding records.
- The `SupportUser` model is configured for soft deletion.
- The support users index action omits discarded users.
- The `SupportUser.load_from_session` authentication method omits discarded users
- Rephrases 'Delete' calls to action to 'Remove' to reflect soft deletion

image

## Link to Trello card

## Things to check

### DIFF
--- a/app/models/support_user.rb
+++ b/app/models/support_user.rb
@@ -12,7 +12,7 @@ class SupportUser < ActiveRecord::Base
     dfe_sign_in_user = DfESignInUser.load_from_session(session)
     return unless dfe_sign_in_user
 
-    SupportUser.find_by(dfe_sign_in_uid: dfe_sign_in_user.dfe_sign_in_uid)
+    SupportUser.kept.find_by(dfe_sign_in_uid: dfe_sign_in_user.dfe_sign_in_uid)
   end
 
 private
--- a/app/models/support_user.rb
+++ b/app/models/support_user.rb
@@ -1,4 +1,6 @@
 class SupportUser < ActiveRecord::Base
+  include Discard::Model
+
   validates :dfe_sign_in_uid, presence: true
   validates :email_address, presence: true
 
--- a/spec/system/support_interface/removed_support_user_spec.rb
+++ b/spec/system/support_interface/removed_support_user_spec.rb
@@ -1,0 +1,45 @@
+require 'rails_helper'
+
+RSpec.describe 'A removed support user attempts to authenticate via DfE Sign-in' do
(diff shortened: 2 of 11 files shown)


app/models/support_user.rb
```diff
@@ -10,7 +12,7 @@ def self.load_from_session(session)
     dfe_sign_in_user = DfESignInUser.load_from_session(session)
     return unless dfe_sign_in_user
 
-    SupportUser.find_by(dfe_sign_in_uid: dfe_sign_in_user.dfe_sign_in_uid)
+    SupportUser.find_by(dfe_sign_in_uid: dfe_sign_in_user.dfe_sign_in_uid, discarded_at: nil)
```
duncanjbrown: Should we use the `kept` scope here instead of `discarded_at`?

spec/system/support_interface/removed_support_user_spec.rb
```diff
@@ -0,0 +1,45 @@
+require 'rails_helper'
+
+RSpec.describe 'A removed support user attempts to authenticate via DfE Sign-in' do
+  include DfESignInHelpers
+
+  scenario 'signing in as removed user' do
```
duncanjbrown: :+1: Great spec

app/models/support_user.rb
steventux: @duncanjbrown good point - amended here https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1517/files#diff-e24635ab750d5f5d41f7b82858b9ec65R15

steventux:
@duncanjbrown I've amended all visible instances of _Delete_ to read _Remove_ https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1517/files#diff-e0747ae2722aea43510b75a45cc9ac04

app/models/support_user.rb
duncanjbrown: > @duncanjbrown good point - amended here https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1517/files#diff-e24635ab750d5f5d41f7b82858b9ec65R15

looks good 👌

PR #1520, opened 2020-03-03
Render other_uk_qualification_type in qualifications title

steventux:
## Context

Candidates who choose the 'Other' GCSE qualification type are able to enter an additional value to denote the type, 'A Level' being a common case. 
We should expose this value in the qualification title rather than using the generic formatting of 'Other UK _subject_'.

This PR adds an additional way of rendering the qualification title text.
If `ApplicationQualification#other_uk_qualification_type` is present we prefer this to the generic `Other UK ...` title.

### Before
image

### After
image

## Guidance to review

The candidate qualification type input fields don't appear to have the conditional validation to enforce the user fills out the 'Other' qualification type, perhaps this is something to iterate. 

The approach here is to only render the specified type if a value for `ApplicationQualification#other_uk_qualification_type` is present. 

## Link to Trello card

https://trello.com/c/emBxNMKU/1712-display-otherukqualificationtype-in-provider-and-support-interface

## Things to check

### DIFF
--- a/app/components/qualification_title_component.rb
+++ b/app/components/qualification_title_component.rb
@@ -4,16 +4,31 @@ class QualificationTitleComponent < ActionView::Component::Base
   end
 
   def qualification_type
-    if @qualification.level == 'gcse' && @qualification.missing_qualification?
-      I18n.t('application_form.gcse.qualification_types_abbreviated.gcse')
-      ApplicationQualification.human_attribute_name('qualification_type.gcse')
-    elsif @qualification.level == 'gcse'
-      ApplicationQualification.human_attribute_name(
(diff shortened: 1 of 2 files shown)


fofr:
> The candidate qualification type input fields don't appear to have the conditional validation to enforce the user fills out the 'Other' qualification type, perhaps this is something to iterate.
We need to add this check.

fofr:
I think it should be:
`Other UK qualification: A level Maths` rather than `A level Maths`, which would read oddly in the GCSEs section 🤔

steventux:
@fofr I've updated the title as per your comment - see latest screenshot above.

PR #1521, opened 2020-03-03
Add radio button for providers with twenty or less course choices options

davidgisbey:
## Context

We currently show an autocomplete for picking a course, whether there is 1 course or 100.

Autocompletes are harder to use than Radios. When there's 20 or less courses (20 is the average number of courses a provider has) we should show them as a list of radio buttons instead.

For lists longer than 20, an autocomplete remains the best option.

The number of courses counted should be the number of available courses (eg those with vacancies)

- Use radio buttons on the options_for_course page if there are 20 or less available course options

## Guidance to review

- You can use Gorse to test the dropdown
- You can use the University of Chichester to test the radio buttons.

## Link to Trello card

https://trello.com/c/xGUWGWBe/

## Things to check

### DIFF
--- a/app/views/candidate_interface/course_choices/options_for_course.html.erb
+++ b/app/views/candidate_interface/course_choices/options_for_course.html.erb
@@ -11,7 +11,7 @@
         ) do |f| %>
       <%= f.govuk_error_summary %>
 
-      <% if @provider.courses.exposed_in_find.count > 20 %>
+      <% if @pick_course.available_courses.count > 20 %>
 
         <%= f.govuk_collection_select(
               :course_id,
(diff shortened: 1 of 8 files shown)


app/views/candidate_interface/course_choices/options_for_course.html.erb
```diff
@@ -11,14 +11,32 @@
...
+
+            <div class="govuk-!-margin-top-6">
+
+            <% @pick_course.available_courses.each_with_index do |course, i| %>
+              <%= f.govuk_radio_button :course_id, course.id, label: { text: "#{course.name} (#{course.code})" }, link_errors: i.zero? %>
+            <% end %>
+
+            </div>
+
+          <% end %>
```
tvararu:
```suggestion
        <%= f.govuk_radio_buttons_fieldset :course_id, legend: { size: 'xl', text: t('page_titles.which_course') } do %>

          <div class="govuk-!-margin-top-6">

          <% @pick_course.available_courses.each_with_index do |course, i| %>
            <%= f.govuk_radio_button :course_id, course.id, label: { text: "#{course.name} (#{course.code})" }, link_errors: i.zero? %>
          <% end %>

          </div>

        <% end %>
```

PR #1522, opened 2020-03-03
Add stubbing of course names in selecting a course study mode spec

chubberlisk:
## Context

We have a flakey test and it caused Master to fail, see https://dfe-ssp.visualstudio.com/Become-A-Teacher/_build/results?buildId=49348&view=ms.vss-test-web.build-test-results-tab&runId=1179756&resultId=100932&paneView=debug.

image

This PR stubs the name of courses when they are created in the spec to prevent getting an ambiguous match when selecting a course.

## Guidance to review

Does this make sense?

## Link to Trello card

N/A

## Things to check

### DIFF
--- a/spec/system/candidate_interface/candidate_selecting_a_course_study_mode_spec.rb
+++ b/spec/system/candidate_interface/candidate_selecting_a_course_study_mode_spec.rb
@@ -23,7 +23,7 @@ RSpec.feature 'Selecting a study mode' do
     @third_site = create(:site, provider: @provider)
 
     @single_site_course = create(
-      :course, :with_both_study_modes, :open_on_apply, provider: @provider
+      :course, :with_both_study_modes, :open_on_apply, provider: @provider, name: 'MS Painting'
     )
 
     create(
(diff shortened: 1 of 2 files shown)


spec/system/candidate_interface/candidate_selecting_a_course_study_mode_spec.rb
```diff
@@ -23,7 +23,7 @@ def given_there_is_a_single_site_full_time_course
     @third_site = create(:site, provider: @provider)
 
     @single_site_course = create(
-      :course, :with_both_study_modes, :open_on_apply, provider: @provider
+      :course, :with_both_study_modes, :open_on_apply, provider: @provider, name: 'MS Painting'
```
steventux: Doesn't the underlying problem still exist in the `:course` factory?
https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/spec/factories.rb#L155
Would it be better to fix this with some sort of FactoryBot sequence on the name?

tvararu: ^ I like this suggestion

steventux: @chubberlisk I misunderstood the issue, course names aren't unique so it's fair to fix the test in question.

PR #1527, opened 2020-03-04
Rake task to clean data on QA

duncanjbrown:
## Context

QA is full of old and incomplete data which makes testing a bit uncertain and causes weird errors.

Provide a way to delete it all and start again.

Considerations:

- we want to preserve support users as it's a pain to create them via the shell. proposed solution: cache existing ones in Redis and restore after deleting the data
- qa runs in `Rails.env.production` so `rake db:reset` is not available for trashing the db. also, we can't run that if there is another connection to the db, which there will be because the app is running. this code directly calls `ActiveRecord`'s `truncate_all` method 😬 

Refactor the local_dev_data task so we can reuse some of the bits. Add the redis serializing and deserializing. Add the rake task with suitable guards.

## Guidance to review

Rake dependencies are a funny way to order things. Maybe this should be a `bin/` script?
This doesn't trash redis, which might result in some sidekiq errors.

## Link to Trello card

https://trello.com/c/yRSmo96F/1658-investigate-providers-with-no-names-on-qa

## Things to check

### DIFF
--- a/lib/tasks/local_dev.rake
+++ b/lib/tasks/local_dev.rake
@@ -12,10 +12,11 @@ task setup_local_dev_data: %i[environment copy_feature_flags_from_production syn
   SupportUser.find_or_create_by!(dfe_sign_in_uid: 'dev-support', email_address: 'support@example.com')
 end
 
-desc 'Sync providers whitelisted in DEV_PROVIDERS_TO_SYNC and open all their courses'
+desc 'Sync some pilot-enabled providers and open all their courses'
 task sync_dev_providers_and_open_courses: :environment do
   puts 'Syncing data from Find...'
-  ENV['DEV_PROVIDERS_TO_SYNC'].split(',').each do |code|
+
+  %w[1N1 2LR C58].each do |code|
     SyncProviderFromFind.call(provider_code: code, sync_courses: true)
   end
 
--- a/lib/tasks/local_dev.rake
+++ b/lib/tasks/local_dev.rake
@@ -1,18 +1,5 @@
 desc 'Set up your local development environment with data from Find'
-task setup_local_dev_data: %i[environment copy_feature_flags_from_production] do
-  puts 'Syncing data from Find...'
-  ENV['DEV_PROVIDERS_TO_SYNC'].split(',').each do |code|
-    SyncProviderFromFind.call(provider_code: code, sync_courses: true)
-  end
-
-  puts 'Making all the courses open on Apply...'
-  Provider.all.each do |provider|
-    OpenProviderCourses.new(provider: provider).call
-  end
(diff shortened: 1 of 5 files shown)


lib/tasks/local_dev.rake
```diff
@@ -12,10 +12,11 @@ task setup_local_dev_data: %i[environment copy_feature_flags_from_production syn
   SupportUser.find_or_create_by!(dfe_sign_in_uid: 'dev-support', email_address: 'support@example.com')
 end
 
-desc 'Sync providers whitelisted in DEV_PROVIDERS_TO_SYNC and open all their courses'
+desc 'Sync some pilot-enabled providers and open all their courses'
 task sync_dev_providers_and_open_courses: :environment do
   puts 'Syncing data from Find...'
-  ENV['DEV_PROVIDERS_TO_SYNC'].split(',').each do |code|
+
+  %w[1N1 2LR C58].each do |code|
```
malcolmbaig: 👍

lib/tasks/reset_qa.rake
```diff
@@ -0,0 +1,37 @@
...
+  truncate_qa_database
+  restore_support_users
+  sync_dev_providers_and_open_courses
+  generate_test_applications
+]
+
+task truncate_qa_database: :environment do
+  raise 'You can’t do this anywhere but QA or development!' unless HostingEnvironment.qa? || Rails.env.development?
+
+  ActiveRecord::Tasks::DatabaseTasks.truncate_all
```
stevehook: Does this truncate the 'system' tables like `schema_migrations`?

lib/tasks/reset_qa.rake
```diff
@@ -0,0 +1,37 @@
...
+end
+
+task backup_support_users: :environment do
+  existing_users = SupportUser.pluck(:email_address, :dfe_sign_in_uid)
+  Redis.new.set('provider_users', JSON.generate(existing_users))
```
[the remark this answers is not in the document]
duncanjbrown: > 👍 Using Redis to stash these is a great idea. Should we use `support_users` as the key rather than `provider_users`?

oops — yes!

> 
> A (fairly unlikely) potential issue with this is that the process could fail after clearing the DB and before `restore_support_users` runs. One way to mitigate that would be to skip the backup if the `support_users` table is empty.

Good idea

lib/tasks/reset_qa.rake
duncanjbrown: Good question. Seems not to. Ran it locally and

Screenshot 2020-03-04 at 11 36 24

duncanjbrown:
@stevehook updated (and added tests)

PR #1528, opened 2020-03-04
Revert "Bump actionview-component from 1.10.0 to 1.13.0"

tvararu:
Reverts DFE-Digital/apply-for-postgraduate-teacher-training#1512

Unfortunately it looks like something in the [update from v1.11.1..v1.12.0](https://github.com/github/actionview-component/compare/v1.11.1...v1.12.0) breaks the app when running in production mode. We didn't catch this until it got to QA.


tvararu:
Tested this locally and it fixes the app when running in production mode 👍

PR #1530, opened 2020-03-04
Add before_you_start page to encourage candidate to select a course on first sign in

davidgisbey:
## Context

If we add a page that recommends making course choices before completing other sections of the application, then candidates won’t feel they’ve wasted their time entering information into a service they can’t use.

- Add a page that a user sees the first time they login which prompts them to add a course
- This only applies to candidates who have not arrived with a course from find

image

## Guidance to review

Create a new account. When you sign in you should see the above page with a flash message saying you have created your account. Click on choose a course and add a course as normal.

Do the same, but click on 'Go to your application form'. You should be taken to your application form.

## Link to Trello card

https://trello.com/c/nrFum7NU/

## Things to check


app/controllers/candidate_interface/sign_in_controller.rb
```diff
@@ -20,7 +20,11 @@ def interstitial
       if service.candidate_does_not_have_a_course_from_find || service.candidate_has_submitted_application
         if more_reference_needed? && FeatureFlag.active?('show_new_referee_needed')
           redirect_to candidate_interface_additional_referee_path
+        elsif flash[:success] == t('apply_from_find.account_created_message') && FeatureFlag.active?('before_you_start')
+          flash[:success] = t('apply_from_find.account_created_message')
```
davidgisbey: It seems that checking for the flash messages presence stops it beings passed on to the next controller action. Is there any way to retain the value and check for it?

duncanjbrown: There is also `#keep` https://apidock.com/rails/ActionController/Flash/FlashHash/keep

davidgisbey: Thanks for that. That's really helpful! I think Theo's right that I shouldn't be checking the flash, but that's really useful to know going forward.

app/models/application_form.rb
```diff
@@ -108,5 +108,15 @@ def within_amend_period?
...
 
+  def blank_application?
+    updated_at == created_at &&
+      application_choices.none &&
+      application_work_experiences.none &&
+      application_volunteering_experiences.none &&
+      application_qualifications.none &&
+      application_references.none &&
+      application_work_history_breaks.none
+  end
```
davidgisbey: This isn't the most elegant implementation of all time, but it was the simplest way I could think to do it.  Another option would be to add an additional column to the ApplicationForm which is updated to true if any of the above are updated.

tvararu: Hmm! What about using the [`unsubmitted_not_started_form` state](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/d9dafbafb220159ece0bf319b4f3cf7c66c95c08/app/services/process_state.rb#L19)?

davidgisbey: Hey, can do. I saw at. the top it says:

` the state of application choices. Do not rely on this for business logic.`

so I wasn't sure if it was okay to use it or not.

davidgisbey: I spoke to Tjmen and he said not to use  unsubmitted_not_started_form as various other things use it such as
 
image

The above solves the issue anyway. I hadn't realised that objects that belong to an application form update the applications forms updated_at.

spec/models/application_form_spec.rb
```diff
@@ -70,4 +70,61 @@ def application_form_with_course_option_for_provider_with(level:)
...
+
+    context 'when a candidate has added a volunteering experience' do
+      it 'returns false' do
+        application_form = create(:application_form)
+        create(:application_volunteering_experience, application_form: application_form)
+        expect(application_form.blank_application?).to be_falsey
+      end
+    end
+
+    context 'when a candidate has added a qualificaion' do
```
tvararu:
```suggestion
    context 'when a candidate has added a qualification' do
```

spec/models/application_form_spec.rb
```diff
@@ -70,4 +70,61 @@ def application_form_with_course_option_for_provider_with(level:)
...
+    end
+
+    context 'when a candidate has added a work history break' do
+      it 'returns false' do
+        application_form = create(:application_form)
+        create(:application_work_history_break, application_form: application_form)
+        expect(application_form.blank_application?).to be_falsey
+      end
+    end
+  end
```
tvararu: Do we still need these exhaustive tests? They're probably testing stuff that's already tested elsewhere.

davidgisbey: 💯 pulled most of them out

PR #1531, opened 2020-03-04
1641 add filtering to the provider ui add components behind feature flag

willmcb:
## Context

Third PR as per Theos comment here: #1456 (review)

Splitting up this large PR into smaller chunks

## Guidance to review

Note that the `app/controllers/provider_interface/application_choices_controller.rb` is messy at the moment - everything works but I am going to do another PR and introduce an `ApplicationsPageConfig` model to dry up some of the functionality and make it neater, and deal with things like pagination in a more streamlined way.

## Link to Trello card

## Things to check

### DIFF
--- a/app/components/provider_interface/filter_component.html.erb
+++ b/app/components/provider_interface/filter_component.html.erb
@@ -10,38 +10,10 @@
 
       <div class="moj-filter__content">
       <% if applied_filters.any? %>
-        <div class="moj-filter__selected">
-
-          <div class="moj-filter__selected-heading">
-
-            <div class="moj-filter__heading-title">
-              <h2 class="govuk-heading-m">Selected filters</h2>
-            </div>
-
-            <div class="moj-filter__heading-action">
-            <p class="govuk-body">
-               <%= link_to "Clear", filtering_page_path(additional_params), class: "govuk-link govuk-link--no-visited-state" %>
-            </p>
-            </div>
-
-          </div>
-
-          <% applied_filters.each do |applied_filters_key, checkbox_config_values| %>
-            <h3 class="govuk-heading-s govuk-!-margin-bottom-0"><%= applied_filters_key.capitalize %></h3>
-            <ul class="moj-filter-tags">
-            <% checkbox_config_values.each do |checkbox_config_key, checkbox_config_value| %>
-                <li>
-                <%= link_to retrieve_tag_text(applied_filters_key,checkbox_config_value), filtering_page_path(
-                  additional_params.merge({ filter_selections: build_tag_url_query_params(
-                                                 heading: applied_filters_key,
-                                                 tag_value: checkbox_config_key,
-                                                 applied_filters: applied_filters)})), class: "moj-filter__tag", id: "tag-#{checkbox_config_value}" %>
-                </li>
-              <% end %>
-            </ul>
-          <% end %>
-
-        </div>
+             <%= render ProviderInterface::CurrentFiltersComponent.new(path: path,
+                                                                       available_filters: available_filters,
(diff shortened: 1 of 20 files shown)


app/views/provider_interface/application_choices/index.html.erb
```diff
@@ -2,39 +2,84 @@
...
-        <td class='govuk-table__cell govuk-table__header--numeric'><%= application_choice.updated_at.to_s(:govuk_date_short_month) %></td>
-      </tr>
+<% if FeatureFlag.active?('provider_application_filters') %>
+
+<div class='moj-filter-layout'>
+  <% if @filter_visible.eql?('true') %>
+    <%= render ProviderInterface::FilterComponent.new(path: :provider_interface_applications_path,
+                                                      available_filters: @available_filters,
+                                                      preselected_filters: @filter_selections,
+                                                      additional_params: { sort_by: @sort_by,
```
duncanjbrown: This pattern is quite nice because it means the filter doesn't need to know about the sorting. As things are a change to the sorting params requires a change to this call to the filter, but that needn't be the case: we could just pass `params` here (or maybe a hash of eg `@sorting_and_filtering_options` built in the controller) and the `merge` would do the right thing. It would be possible to reuse this pattern in `ToggleFilterComponent` and `SortableColumnHeadingComponent`.

duncanjbrown: Also this would improve the component interfaces, eg change the constructor of `ToggleFilterComponent` to be `(filter_visible:, additional_params:)` which is a bit more intention-revealing

willmcb: This is a neater way to look at this. I'll look to do this when I checkout my final branch for this feature.

app/components/provider_interface/filter_component.html.erb
```diff
@@ -1,81 +1,80 @@
...
 
-      <div class="moj-filter__options">
-        <form method="get" action=<%= filtering_page_path %>>
-          <%= submit_tag "Apply filters", class: "govuk-button" %>
-          <% additional_params.each do |key, value| %>
-            <input class="govuk-checkboxes__input" id=<%= key %> name=<%= key %> type="hidden" value=<%= value %> %>
+          </div>
+
+          <% preselected_filters.each do |preselected_filters_key, checkbox_config_values| %>
+            <h3 class="govuk-heading-s govuk-!-margin-bottom-0"><%= preselected_filters_key.capitalize %></h3>
```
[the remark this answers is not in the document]
willmcb: Absolutely I was thinking similar to be honest - especially as it uses  another component further down, it makes sense to break this out as well.

app/views/provider_interface/application_choices/index.html.erb
```diff
@@ -2,39 +2,84 @@
...
-        <td class='govuk-table__cell'><%= render ProviderInterface::ApplicationStatusTagComponent.new(application_choice: application_choice) %></td>
-        <td class='govuk-table__cell govuk-table__header--numeric'><%= application_choice.updated_at.to_s(:govuk_date_short_month) %></td>
-      </tr>
+<% if FeatureFlag.active?('provider_application_filters') %>
+
+<div class='moj-filter-layout'>
+  <% if @filter_visible.eql?('true') %>
+    <%= render ProviderInterface::FilterComponent.new(path: :provider_interface_applications_path,
+                                                      available_filters: @available_filters,
+                                                      preselected_filters: @filter_selections,
```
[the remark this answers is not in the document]
willmcb: I like `applied_filters` I think I'll go with that 👍

PR #1541, opened 2020-03-05
Remove incorrect email from when_emails_are_sent page

duncanjbrown:
## Context

We say providers get an `application_withrawn` email on marking conditions not met. This isn't true.

Also, `application_withrawn` contains a typo. Will fix that in a separate PR (sidekiq might have references to it).

Before

Screenshot 2020-03-05 at 11 27 32

After

Screenshot 2020-03-05 at 11 28 14

### DIFF
--- a/config/locales/application_states.yml
+++ b/config/locales/application_states.yml
@@ -297,11 +297,13 @@ en:
       description: The provider says the candidate hasn’t met the conditions set out in the offer.
       emails:
         - candidate_mailer-conditions_not_met
-        - provider_mailer-application_withrawn
+
     pending_conditions-withdraw:
       name: Candidate withdraws
       by: candidate
       description: Candidates can withdraw at any time.
+      emails:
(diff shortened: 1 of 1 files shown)


config/locales/application_states.yml
```diff
@@ -297,7 +297,7 @@ en:
       description: The provider says the candidate hasn’t met the conditions set out in the offer.
       emails:
         - candidate_mailer-conditions_not_met
-        - provider_mailer-application_withrawn
```
tijmenb: I think was misplaced and it belonged in the transition below?

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/995af4fd81ae7411d85640d63933445f9b5577d2/config/locales/application_states.yml#L301-L304

duncanjbrown: good spot, will put it there

PR #1547, opened 2020-03-05
SPIKE: Provider UI application timeline 

duncanjbrown:
This is a spike into implementing the provider UI application timeline.

**Findings**

This is v doable, but we should consider a couple of upfront refactors:

- test data generation doesn't do right by the audit log. Everything is attributed to the Candidate. Also, everything happens at the same date/time which isn't realistic.
- we still audit provider updates as a string `email@example.com (Provider)`, not a link to a `ProviderUser`, which we ought to turn into a proper association

Incidentally, we bump into the question we faced around logging *transitions* as well as *states*, last seen when we were looking into how we handled an application transitioning between offer and reject (ie withdrawn by provider). There's some manual mapping in the component to accommodate this.

This code also needs to handle VendorApiUsers.

## Link to Trello card

https://trello.com/c/ZvgOpW5Z/1697-implement-the-new-timeline-design-for-applications-on-the-provider-ui-application-page

### DIFF
--- a/app/components/provider_interface/application_timeline_component.html.erb
+++ b/app/components/provider_interface/application_timeline_component.html.erb
@@ -1,19 +1,17 @@
 <h2 class="govuk-heading-l">Timeline</h2>
 <div class="app-timeline">
-  <div class="app-timeline__item">
+  <% entries.each do |entry| %>
+    <div class="app-timeline__item">
+      <div class="app-timeline__header">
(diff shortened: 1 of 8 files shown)


duncanjbrown:
This spike has served its purpose — have updated the Trello card https://trello.com/c/ZvgOpW5Z/1697-implement-the-new-timeline-design-for-applications-on-the-provider-ui-application-page.

Leaving the branch around in case it's useful when we come to implement.

PR #1551, opened 2020-03-06
Allow provider users to modify existing offers (offered course) via the UI

mnacos:
## Context

While it is possible for vendors to change the offered course through the API, provider users currently cannot make such changes throught the web UI. This PR aims to address this.

`app/components/provider_interface/status_box_components/offer_component.rb` has been updated to show 'Change' links for provider, course and location, if the relevant feature flag (`provider_change_response`) is active.

This work replaces the previous 'edit response' flow. Changes to offers are done via the 'Change' links, offer withdrawals via a link at the bottom of the status box component.

The new designs allow the provider user to change the provider, as part of modifying the offer, as long as an association between the current user and the new provider exists. Courses are then scoped to the new provider, and locations are scoped to the course. The current logic tries to preserve the `study_mode` (e.g. full-time, part-time) of the original application. Changing `study_mode` is outside the scope of this PR.

Changing the provider is not required. In most cases, provider users will probably just want to change either Course + Location, or just Location, and these flows are also supported.

Previously:
image

Now, with `provider_change_response` active:
image

Happy path:

ezgif com-optimize

## Guidance to review

The main new files are: `app/controllers/provider_interface/offer_changes_controller.rb` and `app/models/provider_interface/change_offer_form.rb`.

I've had to update many test helpers and factories because this is new territory (e.g. if courses are not `open_on_apply` they will not show up in the UI, other parts of the provider interface do not check things like that). It seems like our provider UI tests were unaffected by the lack of completed application forms past the :offer stage.

Some more thinking is required around visibility of applications when a provider user is associated with multiple providers, esp. if they can modify offers. Currently, the application is mainly associated with the provider offering the course the candidate has originally applied to, but there may be flows that require changing application_choice ownership when an offer is modified.

This PR also includes code to display the current offered course option instead of the course option the candidate requested, if a provider user has already modified it. Some additional design/UI work may be needed to make it clear the course option has been modified.

## Link to Trello card

[Providers can change the offered course via the UI](https://trello.com/c/Rudw3nXZ)

## Things to check

### DIFF
--- a/app/components/provider_interface/status_box_components/offer_component.rb
+++ b/app/components/provider_interface/status_box_components/offer_component.rb
@@ -26,16 +26,33 @@ module ProviderInterface
           {
             key: 'Provider',
             value: application_choice.offered_course.provider.name,
+            change_path: change_path(:provider),
           },
           {
             key: 'Course',
             value: application_choice.offered_course.name_and_code,
+            change_path: change_path(:course),
           },
           {
             key: 'Location',
             value: application_choice.offered_site.name_and_address,
+            change_path: change_path(:course_option),
           },
         ]
+      end
+
+      def paths
+        Rails.application.routes.url_helpers
+      end
+
+      def change_path(target)
+        return nil unless FeatureFlag.active?('provider_change_response')
+
+        case target
+        when :provider then paths.provider_interface_application_choice_change_offer_edit_provider_path(application_choice.id)
+        when :course then paths.provider_interface_application_choice_change_offer_edit_course_path(application_choice.id)
+        when :course_option then paths.provider_interface_application_choice_change_offer_edit_course_option_path(application_choice.id)
+        end
       end
     end
   end
--- a/app/components/provider_interface/status_box_components/offer_component.rb
+++ b/app/components/provider_interface/status_box_components/offer_component.rb
@@ -26,17 +26,17 @@ module ProviderInterface
           {
             key: 'Provider',
             value: application_choice.offered_course.provider.name,
-            change_path: change_path(:provider),
+            change_path: change_path(:provider), change_class: 'provider'
           },
           {
             key: 'Course',
             value: application_choice.offered_course.name_and_code,
-            change_path: change_path(:course),
+            change_path: change_path(:course), change_class: 'course'
           },
           {
             key: 'Location',
             value: application_choice.offered_site.name_and_address,
-            change_path: change_path(:course_option),
+            change_path: change_path(:course_option), change_class: 'course_option'
           },
         ]
       end
--- a/app/components/provider_interface/status_box_components/offer_component.rb
+++ b/app/components/provider_interface/status_box_components/offer_component.rb
@@ -26,17 +26,17 @@ module ProviderInterface
           {
             key: 'Provider',
             value: application_choice.offered_course.provider.name,
-            change_path: change_path(:provider), change_class: 'provider'
+            change_path: change_path(:provider), action: 'training provider'
           },
           {
             key: 'Course',
             value: application_choice.offered_course.name_and_code,
-            change_path: change_path(:course), change_class: 'course'
+            change_path: change_path(:course), action: 'course'
           },
           {
             key: 'Location',
             value: application_choice.offered_site.name_and_address,
-            change_path: change_path(:course_option), change_class: 'course_option'
+            change_path: change_path(:course_option), action: 'location'
           },
         ]
       end
--- a/app/controllers/provider_interface/offer_changes_controller.rb
+++ b/app/controllers/provider_interface/offer_changes_controller.rb
@@ -1,0 +1,154 @@
+module ProviderInterface
+  class OfferChangesController < ProviderInterfaceController
+    before_action :set_application_choice
+    before_action :requires_provider_change_response_feature_flag
+    before_action :set_change_offer_form
+
+    def edit_provider
+      @change_offer_form.step = :provider
+      set_alternative_providers
+    end
+
+    def edit_course
+      @change_offer_form.step = :course
+      if @change_offer_form.valid?
+        set_alternative_courses
+      else
+        render_providers
+      end
+    end
+
+    def edit_course_option
+      @change_offer_form.step = :course_option
+      if @change_offer_form.valid?
+        set_alternative_course_options
+      else
+        render_courses
+      end
+    end
+
+    def confirm_update
+      @change_offer_form.step = :confirm
+      if @change_offer_form.valid?
+        @future_application_choice = @application_choice.dup
+        @future_application_choice.offered_course_option_id = @change_offer_form.course_option_id
+      elsif @change_offer_form.errors[:provider_id].present?
+        render_providers
+      elsif @change_offer_form.errors[:course_id].present?
+        render_courses
+      else
+        render_course_options
+      end
+    end
+
+    def update
+      @change_offer_form.step = :update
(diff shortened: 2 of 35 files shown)


config/locales/en.yml
```diff
@@ -133,6 +133,15 @@ en:
...
               blank: Select if you want to make an offer or reject the application
+        provider_interface/change_offer_form:
+          attributes:
+            provider_id:
+              blank: You have not specified a provider
+            course_id:
+              blank: You have not specified a course
+              not_open_on_apply: Course not open on Apply
+            course_option_id:
+              blank: You have not specified a location
```
laurahermionetennant92:
```suggestion
              blank: You have not chosen a new location
```

config/locales/en.yml
```diff
@@ -133,6 +133,15 @@ en:
           attributes:
             decision:
               blank: Select if you want to make an offer or reject the application
+        provider_interface/change_offer_form:
+          attributes:
+            provider_id:
+              blank: You have not specified a provider
+            course_id:
+              blank: You have not specified a course
```
laurahermionetennant92:
```suggestion
              blank: You have not chosen a new course
```

config/locales/en.yml
```diff
@@ -133,6 +133,15 @@ en:
           attributes:
             decision:
               blank: Select if you want to make an offer or reject the application
+        provider_interface/change_offer_form:
+          attributes:
+            provider_id:
+              blank: You have not specified a provider
```
laurahermionetennant92:
```suggestion
              blank: You have not chosen an alternative provider
```

app/controllers/provider_interface/offer_changes_controller.rb
```diff
@@ -0,0 +1,159 @@
...
+      course_option_id = change_offer_params[:course_option_id]
+      CourseOption.find(course_option_id) if course_option_id.present?
+    end
+
+    def set_alternative_providers
+      @alternative_providers = current_provider_user.providers.order(:name)
+    end
+
+    def set_alternative_courses
+      @alternative_courses = Course.where(open_on_apply: true, provider: provider).order(:name)
```
stevehook: Should we exclude the currently selected course? What happens if I just select the same course again?

duncanjbrown: It could have a different study_mode, maybe?

duncanjbrown: If we were worried about providers ending up with an unsatisfying no-op, we could check at the end whether they were attempting to change the offered course to the existing offered course

app/controllers/provider_interface/offer_changes_controller.rb
```diff
@@ -0,0 +1,159 @@
...
+      @alternative_course_options = CourseOption.where(
+        course: course,
+        study_mode: current_option.study_mode, # preserving study_mode, for now
+        # TODO: check vacancy_status, e.g. 'B'
+      ).includes(:site).order('sites.name')
+    end
+
+    def change_offer_params
+      begin
+        params.require(:provider_interface_change_offer_form).permit(:provider_id, :course_id, :course_option_id)
```
duncanjbrown: Do we need to rescue? would `params.permit(provider_interface_change_offer_form: [:provider_id, :course_id, :course_option_id])` do the business?

app/models/application_choice.rb
```diff
@@ -36,13 +36,15 @@ def offer_withdrawn?
     rejected? && !offer_withdrawn_at.nil?
   end
 
+  def offered_option
```
duncanjbrown: 👌

app/models/provider_interface/change_offer_form.rb
```diff
@@ -0,0 +1,49 @@
+module ProviderInterface
+  class ChangeOfferForm
```
duncanjbrown: This class is lovely, nice work!

app/models/provider_interface/change_offer_form.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    validates :application_choice, presence: true
+
+    # This is a form for a multi-step process, in which each step depends on the previous ones.
+    # Validation needs to take this into account, e.g. it is fine for course_id to be blank if step == :provider
+    #
+    STEPS = %i(provider course course_option confirm update).freeze
+    validates :step, presence: true, inclusion: { in: STEPS, message: '%{value} is not a valid option' }
+
+    def step_after?(symbol)
```
duncanjbrown: Maybe give it a role-ier name?

```suggestion
    def step_after?(step_symbol)
```

app/controllers/provider_interface/offer_changes_controller.rb
mnacos: It's a good point, have just pushed a commit introducing this check in the service.

app/components/provider_interface/status_box_components/offer_component.rb
```diff
@@ -26,17 +26,34 @@ def rows
           {
             key: 'Provider',
             value: application_choice.offered_course.provider.name,
+            change_path: change_path(:provider), change_class: 'provider'
```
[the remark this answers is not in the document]
mnacos: I agree, it felt a little awkward adding this, but I also feel awkward about having multiple 'Change' links in the summary list component with no particular annotation to them. Does this impact accessibility? I propose we merge as is and tackle multiple 'Change' links as a separate PR. @adamsilver thoughts?

app/controllers/provider_interface/offer_changes_controller.rb
mnacos: No, sorry, doesn't do the trick. I like just doing `change_offer_params[:course_id]` and need to differentiate between a direct navigation to this step (e.g. by clicking to change the course from the application choice page) and a form submission from the previous page (:provider step). This is done partly on the basis of whether the top-level item `provider_interface_changer_offer_form` is present or not.

app/models/provider_interface/change_offer_form.rb
mnacos: Done

app/controllers/provider_interface/offer_changes_controller.rb
mnacos: Good catch, I've also added a `study_mode` check within `#set_alternative_courses`. I think the code did the right thing without this already, as during the last step it was filtering out study modes different to the current one. But it could have led to presenting courses that, if selected, would have produced empty option/locations lists to choose from in the next step.

mnacos: If we need a separate feature flag for withdrawing applications, I'll refactor then.

app/components/provider_interface/status_box_components/offer_component.rb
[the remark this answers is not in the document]
duncanjbrown: Agreed!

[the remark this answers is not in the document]
adamsilver: I think this should be addressed as part of this PR. We have done this in other places to make sure multiple Change links have context for screen readers.

There should be a visually-hidden property we can use and the text would be

‘training provider’, ‘course’, ‘location’ for each of the change links.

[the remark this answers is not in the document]
mnacos: Great, thanks Adam. Setting an `action:` against each change link solved both the accessibility question and allowed me to remove the custom :change_class and remain able to click on specific links within the system specs.

PR #1553, opened 2020-03-06
Add 'first' instead of '1st' for referees + change heading size

willmcb:
## Context

Fixes some discrepancy between app and design.

Referee headings were coming up as '1st reference', '2nd reference' this PR changes this so they read 'First referee', 'Second referee' etc. 

More details in the trello card.

Also changes the `<h2>` tags to govuk mediums rather than large.

**Before:**
image

**After**
Screenshot 2020-03-06 at 12 12 55

## Link to Trello card

https://trello.com/c/tU8HFZco/1732-fix-referee-sizing-and-content

## Things to check

### DIFF
--- a/app/services/text_ordinalizer.rb
+++ b/app/services/text_ordinalizer.rb
@@ -1,0 +1,18 @@
+class TextOrdinalizer
+  def self.call(value)
+    text_ordinalize(value)
+  end
+
+  class << self
+  private
+
+    def text_ordinalize(value)
+      ordinalize_mapping[value] || value.ordinalize
+    end
+
+    def ordinalize_mapping
+      %w[zeroth first second third fourth fifth sixth seventh
+         eighth ninth tenth]
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


app/services/text_ordinalizer.rb
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  class << self
+  private
+
+    def text_ordinalize(value)
+      ordinalize_mapping[value] || value.ordinalize
+    end
+
+    def ordinalize_mapping
```
stevehook: Maybe this should be a constant rather than a method?

stevehook: You could argue that it should be in the locale yml files but based on the conversation we had the other week I think we are best off avoiding that.

fofr:
Thank you 🙏

app/services/text_ordinalizer.rb
willmcb: Have constantised it 👍

PR #1555, opened 2020-03-06
Prepend Heroku app name to config.hosts

tvararu:
## Context

We call `authorised_hosts.first` in `HostingEnvironment#hostname` when `ENV["CUSTOM_HOSTNAME"]` is not defined. On Heroku, `authorised_hosts` always contains `apply-for-teacher-training.herokuapp.com` as the first item in a review app. Because of this, our review apps send magic link emails that tell users to go to `apply-for-teacher-training.herokuapp.com`.

This is a bit annoying, but we can easily work around it by prepending the desired hostname.

Prepend the Heroku app hostname instead of appending it. Do this in `HostingEnvironment` so we're modifying only one source of truth.

## Guidance to review

Test this by trying to log into the review app. The email you receive should contain a link that works.

## Link to Trello card

Nope.

## Things to check

### DIFF
--- a/app/lib/hosting_environment.rb
+++ b/app/lib/hosting_environment.rb
@@ -8,7 +8,9 @@ module HostingEnvironment
   end
 
   def self.authorised_hosts
-    ENV.fetch('AUTHORISED_HOSTS').split(',').map(&:strip)
+    hosts = ENV.fetch('AUTHORISED_HOSTS').split(',').map(&:strip)
+    hosts.prepend "#{ENV['HEROKU_APP_NAME']}.herokuapp.com" if ENV['HEROKU_APP_NAME']
+    hosts
   end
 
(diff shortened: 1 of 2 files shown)


tvararu:
Tested this on the review app, and it doesn't work yet.

tvararu:
Nice:

Screenshot 2020-03-06 at 17 59 08

tvararu:
@tijmenb have made an attempt to make this more resilient locally but I don't think it's great (because: it calls `ENV['HEROKU_APP_NAME']` in two places), pushed it up as a branch: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/compare/heroku-app-name?expand=1

If you have better ideas for how to improve, reply here or push to that branch 👍

PR #1557, opened 2020-03-06
1729-fix-summary-list-so-that-theres-no-gap-unless-at-least-one-row-has-an-action

willmcb:
Changes SummaryListComponent so that if all of the rows that are fed into it have no actions no <span> is output at the end of each row.

If any of the rows have actions the the span is output.

## Changes to UI:

**Before:**
Screenshot 2020-03-09 at 10 57 18

**After:**
Screenshot 2020-03-09 at 10 55 52

## Link to Trello card

https://trello.com/c/2kjPU7Aq/1729-fix-summary-list-so-that-theres-no-gap-unless-at-least-one-row-has-an-action

## Things to check

### DIFF
--- a/app/components/summary_card_component.rb
+++ b/app/components/summary_card_component.rb
@@ -2,7 +2,7 @@ class SummaryCardComponent < ActionView::Component::Base
   validates :rows, presence: true
 
   def initialize(rows:, border: true, editable: true)
-    @rows = determine_rows(rows, editable)
+    @rows = rows_including_actions_if_editable(rows, editable)
     @border = border
   end
 
@@ -14,7 +14,7 @@ private
 
   attr_reader :rows
 
-  def determine_rows(rows, editable)
+  def rows_including_actions_if_editable(rows, editable)
     rows.map do |row|
       row.tap do |r|
         unless editable
--- a/app/components/summary_card_component.rb
+++ b/app/components/summary_card_component.rb
@@ -20,6 +20,7 @@ private
         unless editable
           r.delete(:change_path)
           r.delete(:action_path)
+          r.delete(:action)
         end
       end
     end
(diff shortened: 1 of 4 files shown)


spec/components/summary_list_component_spec.rb
```diff
@@ -62,4 +62,35 @@
...
+            { key: 'Working pattern',
+             value: "Full-time\n Omnis itaque rerum. Velit in ." },
+            { key: 'Description',
+             value: 'Cumque autem veritatis..'  },
+            { key: 'Dates',
+             value: 'May 2003 - November 2019'  }]
+
+    result = render_inline(SummaryListComponent.new(rows: rows))
+
+    expect(result.to_html.include?('<span class="govuk-summary-list__actions"></span>')).to eq(false)
```
duncanjbrown: `expect(result.to_html).not_to include('<span...')`? https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/include-matcher

willmcb: Changed thanks for this!

app/components/summary_card_component.rb
```diff
@@ -20,6 +20,7 @@ def determine_rows(rows, editable)
         unless editable
           r.delete(:change_path)
           r.delete(:action_path)
+          r.delete(:action)
```
duncanjbrown: Nice. While we're here, this method name might be improved as it implies it's determining which rows to show rather than returning a set of rows with the relevant fields knocked out or included. How about `rows_including_actions_if_editable` or sth? cc @chubberlisk

willmcb: 🤔 I think that might make it a bit clearer.

chubberlisk: A much better name! 🚢

PR #1561, opened 2020-03-09
Fetch site statuses from Find API

malcolmbaig:
## Context

We want to stop candidates from applying to courses that are full, so that they don't waste time on their application unnecessarily. The first step in doing so is to retrieve data from the Find API which tells us the vacancy status of a given course/site choice.

Update the SyncProviderFromFind rake task to pull in and persist the required data. See commits for more detail.

## Guidance to review

- A per-commit review will step through the precursory refactors followed by the described changes.
- Testing locally, run `bundle exec rake sync_dev_providers_and_open_courses`. Inspecting your database, records in `course_options` should be populated with a `vacancy_status` value that matches the enum declaration in the CourseOption model.

## Link to Trello card

https://trello.com/c/LAh0iGIc

## Things to check

### DIFF
--- a/app/models/course_option.rb
+++ b/app/models/course_option.rb
@@ -11,6 +11,13 @@ class CourseOption < ApplicationRecord
     part_time: 'part_time',
   }
 
+  enum vacancy_status: {
+    both_full_time_and_part_time_vacancies: 'B',
+    part_time_vacancies: 'P',
+    full_time_vacancies: 'F',
+    no_vacancies: '',
+  }
+
   def validate_providers
     return unless site.present? && course.present?
 
--- a/app/models/course_option.rb
+++ b/app/models/course_option.rb
@@ -12,10 +12,10 @@ class CourseOption < ApplicationRecord
   }
 
   enum vacancy_status: {
-    both_full_time_and_part_time_vacancies: 'B',
-    part_time_vacancies: 'P',
-    full_time_vacancies: 'F',
-    no_vacancies: '',
+    both_full_time_and_part_time_vacancies: 'both_full_time_and_part_time_vacancies',
+    part_time_vacancies: 'part_time_vacancies',
+    full_time_vacancies: 'full_time_vacancies',
+    no_vacancies: 'no_vacancies',
   }
 
   def validate_providers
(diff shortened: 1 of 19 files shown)


malcolmbaig:
Missed a failure in `spec/system/support_interface/providers_spec.rb`, fixing now.

app/models/course_option.rb
```diff
@@ -11,6 +11,13 @@ class CourseOption < ApplicationRecord
     part_time: 'part_time',
   }
 
+  enum vacancy_status: {
+    both_full_time_and_part_time_vacancies: 'B',
+    part_time_vacancies: 'P',
+    full_time_vacancies: 'F',
+    no_vacancies: '',
```
tijmenb: I wonder if we should persist the string human readable strings like `full_time_vacancies`, instead of "B/P/F", to avoid coupling to Find?

app/services/sync_provider_from_find.rb
```diff
@@ -82,7 +86,7 @@ def self.create_or_update_course(find_course, provider)
           site: site,
```
tijmenb: I think we need to update the logic here to update rather than `find_or_create_by`, otherwise existing courses won't be updated.

malcolmbaig: Good point, updating...

app/models/course_option.rb
malcolmbaig: Agreed, clearer when inspecting records in the DB.

app/services/sync_provider_from_find.rb
```diff
@@ -78,12 +82,13 @@ def self.create_or_update_course(find_course, provider)
...
       study_modes.each do |mode|
-        CourseOption.find_or_create_by(
+        course_option = CourseOption.find_or_create_by(
           site: site,
           course_id: course.id,
           study_mode: mode,
-          vacancy_status: 'B', #TODO: Should this be reflected by `find_course.has_vacancies?`
         )
+
+        course_option.update(vacancy_status: site_status.vac_status)
```
tijmenb:
```suggestion
        course_option.update!(vacancy_status: site_status.vac_status)
```

To avoid silent failures (if we introduce validations or something).

PR #1562, opened 2020-03-09
Show work breaks in provider interface

stevehook:
## Context

A candidate needs to provide a full work-history, which includes explanations for any gaps e.g. maternity/paternity leave. 

At the moment a user can provide details in a capture all free text box, but we need to build in the gap calculations and the ability to provide specific explanations.

- Add new components for work history and volunteering history that implement the new design and include both explained and unexplained breaks.
- Add the these components to the Provider interface application show action behind a new `provider_interface_work_breaks` feature flag. 

image

## Guidance to review

- I was going to use the existing `work_breaks` feature flags but this is already active in all environments so there wouldn't be much point. I've add a new a `provider_interface_work_breaks` flag.
- I've added a new system spec with a view to deleting the old one when we remove the feature flag.
- I've added previews for the two new components - I'm not sure they are going to be fully featured without a few additions to the test data generators - to make sure we have a full work history with both kinds of break.
- We should probably use the new components to change the support interface too. Separate card/PR?

## Link to Trello card

https://trello.com/c/ZOkPYTd7/1479-show-work-breaks-calculated-breaks-and-reasons

## Things to check


app/components/previews/provider_interface/volunteering_history_component_preview.rb
```diff
@@ -0,0 +1,17 @@
+module ProviderInterface
+  class VolunteeringHistoryComponentPreview < ActionView::Component::Preview
+    def last_application
```
mnacos: In other components we've written the preview so that it picks a random model from a collection of models that match the use case. See e.g. https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/app/components/previews/provider_interface/status_box_component_preview.rb#L53

This helps preview the component with a range of data, e.g. by refreshing the preview page. Is it worth doing this here, too?

app/components/previews/provider_interface/volunteering_history_component_preview.rb
```diff
@@ -0,0 +1,17 @@
...
+      render_component_for application_form: ApplicationForm.last
+    end
+
+  private
+
+    def render_component_for(application_form:)
+      if !application_form.application_work_experiences.empty?
+        render ProviderInterface::VolunteeringHistoryComponent.new(application_form: application_form)
+      else
+        render template: 'support_interface/docs/missing_test_data'
```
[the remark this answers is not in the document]
stevehook: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1578 attempts to fix this

app/components/previews/provider_interface/volunteering_history_component_preview.rb
stevehook: I've changed the logic in the previews to select a 'random' application from the set of 'qualifying' applications (those that have the right combination of child objects - work experience, volunteering experience and work history breaks). Works with https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1578

PR #1563, opened 2020-03-09
Prevent a candidate from entering a future date for a start date

chubberlisk:
## Context

A candidate was able to submit their application despite including a role where the start date was in the future (April 2020).

This PR adds validation to prevent a start date being in the future for: adding a job, volunteering and work history break.

### After

image

## Guidance to review

FYI: The card specified that adding a job and volunteering but validation was also added for work history breaks.

## Link to Trello card

https://trello.com/c/Nqdbr71A/1116-dev-bug-user-added-future-date-as-start-date-%F0%9F%90%9B

## Things to check

### DIFF
--- a/app/helpers/date_validation_helper.rb
+++ b/app/helpers/date_validation_helper.rb
@@ -35,6 +35,13 @@ module DateValidationHelper
     end
   end
 
+  def start_date_before_current_year_and_month
+    if start_date.year > Time.zone.today.year || \
+        start_date.year == Time.zone.today.year && start_date.month > Time.zone.today.month
+      errors.add(:start_date, :in_the_future)
+    end
+  end
+
   def end_date_before_current_year_and_month
     if end_date.year > Time.zone.today.year || \
         end_date.year == Time.zone.today.year && end_date.month > Time.zone.today.month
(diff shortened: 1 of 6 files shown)


app/helpers/date_validation_helper.rb
```diff
@@ -35,6 +35,13 @@ def start_date_before_end_date
     end
   end
 
+  def start_date_before_current_year_and_month
+    if start_date.year > Time.zone.today.year || \
```
C-gyorfi: I think you could use date comparison like `start_date  < Date.today` or is there a reason you decided to go with this solution?

chubberlisk: Only reason was that `end_date_before_current_year_and_month` was implemented this way so I followed suit. Should have questioned it, thank you for doing so!

Will refactor to your solution. 🚢

chubberlisk: Resolved in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1563/commits/9c97aff4905fa20157502cff588d0c9bed3822a7.

C-gyorfi: Oh I think it was because of the start date does not have a day (or at least it is not displayed) and I assume it defaults to the first day of the month.  You could also use `Time.zone.today.at_beginning_of_month`, but I think it wouldn't make a difference.

PR #1572, opened 2020-03-09
Add button to render an application as PDF

steventux:
## Context

Providers often need to retain application details as files or in print.

Adds 'Download application (PDF)' link to provider application page which responds with a file download of the application in PDF format. 
`pdf-kit` gem is configured as middleware, generates the candidate application as an embedded PDF.

NOTE: This PR has been reworked after some team discussion, we decided to remove the javascript print link and replace with a link to PDF.

### After

image

## Guidance to review

- Does the PDF link work for you? Can you print, download and save the PDF?
- Are all application action buttons (Change, Respond etc.) and links hidden in print/PDF format?

## Link to Trello card
https://trello.com/c/5QZJlM2b/1778-pdf-generation-button

## Things to check

### DIFF
--- a/app/views/provider_interface/application_choices/show.html.erb
+++ b/app/views/provider_interface/application_choices/show.html.erb
@@ -19,7 +19,7 @@
       <span class="govuk-caption-xl"><%= @application_choice.application_form.support_reference %></span>
       <%= @application_choice.application_form.full_name %>
     </h1>
-    <p class="app-print-link app-!-print-display-none">
+    <p class="govuk-body govuk-!-margin-bottom-9 app-!-print-display-none">
       <%= govuk_link_to(
         'Download application (PDF)',
         provider_interface_application_choice_path(@application_choice.id, format: :pdf),
--- a/app/views/provider_interface/application_choices/show.html.erb
+++ b/app/views/provider_interface/application_choices/show.html.erb
@@ -13,23 +13,30 @@
   </div>
 <% end %>
 
-<h1 class="govuk-heading-xl">
-  <span class="govuk-caption-xl"><%= @application_choice.application_form.support_reference %></span>
-  <%= @application_choice.application_form.full_name %>
-</h1>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl govuk-!-margin-bottom-1">
+      <span class="govuk-caption-xl"><%= @application_choice.application_form.support_reference %></span>
+      <%= @application_choice.application_form.full_name %>
+    </h1>
+    <p class="app-print-link app-!-print-display-none">
+      <%= govuk_link_to(
+        'Download application (PDF)',
+        provider_interface_application_choice_path(@application_choice.id, format: :pdf),
+        download: @application_choice.application_form.support_reference
+      ) %>
+    </p>
+  </div>
+</div>
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <%= render ProviderInterface::StatusBoxComponent.new(application_choice: @application_choice) %>
+  <%= render ProviderInterface::StatusBoxComponent.new(application_choice: @application_choice) %>
   </div>
 </div>
 
-<div class="govuk-grid-row app-print-link app-!-print-display-none">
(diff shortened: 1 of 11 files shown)


spec/system/provider_interface/application_as_pdf_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+    expect(page).to have_content(@course_option.course.name_and_code)
+    expect(page).not_to have_link('View as PDF')
+  end
+
+  def convert_pdf_to_page
+    temp_pdf = Tempfile.new('pdf')
+    temp_pdf << page.source.force_encoding('UTF-8')
+    reader = PDF::Reader.new(temp_pdf)
+    pdf_text = reader.pages.map(&:text)
+    page.driver.response.instance_variable_set('@body', pdf_text)
```
duncanjbrown: This is witchcraft 😂

app/views/provider_interface/application_choices/show.html.erb
```diff
@@ -22,6 +22,16 @@
   <div class="govuk-grid-column-two-thirds">
   <%= render ProviderInterface::StatusBoxComponent.new(application_choice: @application_choice) %>
   </div>
+  <% unless rendering_pdf? %>
```
duncanjbrown: I think including the `print.css` stylesheet could make this work without the need for logic in the template. I'm not sure the `pdfkit` middleware supports setting the `stylesheet` property on the instance of `PDFKit`.

What would you think about adding a `respond_to { |format| format.pdf etc }` in the show action in the `ApplicationChoicesController`, and manually instantiating a `PDFKit` there?

steventux: We can set `print_media_type` in the config which will use the print version of styles - will update :+1:

steventux: We could make a similar method to the middleware and call it in a `respond_to` block.
It seems like the middleware intercepts and rewrites `response.body` and a few headers so this could presumably be wrapped around `render`.

steventux: @duncanjbrown updated to use print styles here https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1572/commits/f657bb8614c9f9dbc198315c0ff1ab083889e61f
and template logic is removed in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1572/commits/0a0c4c70a9b7538c6f7d181a050f9ab7e814c4a3

spec/system/provider_interface/application_as_pdf_spec.rb
steventux: It's also failing on CI, which makes me think it's probably not a reliable way of testing the response body. I'll do something simpler.

app/views/provider_interface/application_choices/show.html.erb
```diff
@@ -26,7 +26,7 @@
 
 <div class="govuk-grid-row app-print-link app-!-print-display-none">
   <div class="govuk-grid-column-two-thirds">
-    <%= govuk_link_to('Print this page', 'javascript:window.print();', class: 'app-print-link__link') %>
+    <%= govuk_link_to('View as PDF', provider_interface_application_choice_path(@application_choice.id, format: :pdf)) %>
```
adamsilver: 1. Let's put it in the top right of the page in line with the h1
2. Can you add a `download` attribute on the link so it doesn't open the PDF in the browser
3. I'll work on the text for the link and put it in the prototype which may help you with (1)

steventux: @adamsilver What's the rationale of downloading rather than opening in the browser? Wouldn't we want to offer the user the option to print directly from an embedded PDF viewer rather than download, then locate and open a file, then print?

adamsilver: The rationale is that clicking ‘Download X’ should ‘Download’ rather than show X. It would be slightly deceptive to open it.

It also means they're free to open the PDF in an application of their choosing, which I think will be something alternative to the browser.

steventux:
@adamsilver Preview app should be up to date with style changes now.

adamsilver:
@steventux awesome - ship it :)

PR #1575, opened 2020-03-09
Prevent application choice from being withdrawable during the edit period

C-gyorfi:
The previous bug fix introduced a weird behaviour in certain cases #1508. When a candidate edits an application choice while the application is still in the edit period, but already received references, the application choice has the option of `withdraw` rather than `delete` even though that application has not sent to providers. 

## Context

Update the review component to show withdraw only when an application exceeded the `edit_by` date(therefore sent to provider)

### Before
image

### After
image

## Link to Trello card

https://trello.com/c/u9fFktHo/1048-application-form-errors-when-edit-application-feature-flag-is-on-and-user-edits-their-course-choices

## Things to check

### DIFF
--- a/app/components/candidate_interface/course_choices_review_component.rb
+++ b/app/components/candidate_interface/course_choices_review_component.rb
@@ -29,7 +29,7 @@ module CandidateInterface
     end
 
     def withdrawable?(course_choice)
-      ApplicationStateChange.new(course_choice).can_withdraw?
+      ApplicationStateChange.new(course_choice).can_withdraw? && course_choice.edit_by < Time.zone.now
     end
 
     def any_withdrawable?
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -215,6 +215,7 @@ FactoryBot.define do
 
     status { ApplicationStateChange.valid_states.sample }
     personal_statement { 'hello' }
+    edit_by { Time.zone.now }
 
     factory :submitted_application_choice do
       status { 'awaiting_provider_decision' }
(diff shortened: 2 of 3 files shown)


duncanjbrown:
Per the state diagram, an application choice can be withdrawn before it's sent to providers. It's quite a fiddly distinction between that and deleting the choice! Is it correct to allow withdrawal at that stage? (I guess if they actually want to cancel their application — what happens if they delete all their choices?)

C-gyorfi:
> I guess if they actually want to cancel their application. What happens if they delete all their choices?

The problem is this creates a weird user journey: if you `withdraw` and application while it is within the editing period, then you will see the option to `delete` it. So to replace a course choice (that already received references) you would need to `withdraw` and then again `delete` and then add a new course choice

If they delete all course choices they are prompted to add a new one, of course, this can cause a weird journey if the edit by period ends before the candidate adds a new choice.

duncanjbrown:
Hm. What's the difference between withdrawing and deleting at this stage?

C-gyorfi:
If you withdraw, the application state changes to `withdrawn` but the application choice are still on the form

If you delete, the application choice is deleted from the form

duncanjbrown:
Is the correct fix to remove https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/c2296aeb359bdc2642a5772d2eca3d9d760c7479/app/services/application_state_change.rb#L29 in that case? i.e. to prevent withdrawal at this stage, as it doesn't make much sense and is likely to be confusing — deleting is effectively withdrawing. cc @tijmenb

C-gyorfi:
@duncanjbrown Yes, that seems like a clean solution! I'll wait to hear Tijmen's opinion

tvararu:
Removing the `event :withdraw` from the `application_complete` state does seem correct 🤔

tijmenb:
I think we should pause this until we know what the future is of the amending feature. We might need to replace all of the current code anyway.

PR #1577, opened 2020-03-09
Add safeguarding_concerns and relationship_corrections to references

C-gyorfi:
## Context 

We want to allow referees to say whether the applicant is a suitable candidate to work with children so that a provider can properly assess their application. We also want them to confirm and correct their relationship with the candidate. 

This PR adds a migration with two columns to the `referees table`
- `safeguarding_concerns` to record (if any) concerns as a string
- `relationship_correction` to record (if any) incorrect details are in the `relationship` field

## Guidance to review

Does this make sense? Shall we just amend the existing `relationship` field instead of using a separate `relationship_correction` field is cleaner? I'm open for suggestions

## Link to Trello card

https://trello.com/c/QlKuThen/1130-build-referee-view-of-candidate-working-with-children

## Things to check


C-gyorfi:
> I think this makes sense, having two separate fields as part of a `Reference` as it seems from the provider view (https://bat-design-history.netlify.com/apply-for-teacher-training/give-a-reference-second-iteration#provider-view-with-amended-relationship)
Thanks, Ting, this is super helpful. What do you think @tvararu, are you happy with these changes?

PR #1580, opened 2020-03-09
Make providers application_choices page wider without affecting rest of app

tvararu:
## Context

The `application_choices` index page now can optionally display a set of filters on the left which require a wider layout to look right at the moment.

This was previously done by overriding the main `govuk-width-container` class which has the undesirable effect of affecting the width of every other page in the application, as well as not looking great on breakpoints between ~800px and ~1240px.

- Add new `app-width-container--wide` override class, remove previous `govuk-width-container` declaration
- Use new class in header, body, and footer when a `page_width` variable is set to `"wide"`

## Guidance to review

I think the filters component deserves some more attention separately, as it doesn't behave nicely when scaling down to mobile and back. This can be carded up on its own.

Before:

Screenshot 2020-03-09 at 17 17 27

After:

Screenshot 2020-03-09 at 17 17 17

## Link to Trello card

None.

## Things to check

### DIFF
--- a/app/components/header_component.html.erb
+++ b/app/components/header_component.html.erb
@@ -1,5 +1,5 @@
 <header class="govuk-header <%= classes %>" role="banner" data-module="govuk-header">
-  <div class="govuk-header__container govuk-width-container">
+  <div class="govuk-header__container govuk-width-container<%= wide ? " app-width-container--wide" : "" %>">
     <div class="govuk-header__logo">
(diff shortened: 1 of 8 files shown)


duncanjbrown:
Added a checkbox to the main story card for the filters to check the mobile display https://trello.com/c/wMGLpXGS/1641-add-filtering-to-the-provider-ui

PR #1583, opened 2020-03-09
Allow support users to cancel references

tijmenb:
## Context

We've had 2 support requests from people who've filled in an incorrect email address for references. Not all incorrect emails bounce, so they would have to wait 10 days before adding a replacement referee.

This adds a "cancel reference" button, so we can manually remove the reference request. This allows candidates to provide a replacement. It doesn't email the candidate because this often happens via a Zendesk conversation anyway.

Given this application with references that haven't come in:

image

We can click on 'Cancel reference' and get the confirm screen:

image

This will change the status of the reference to 'cancelled':

image

Which will require the candidate to add a replacement:

image

## Guidance to review

Does it make sense? I don't think we need any checks, eg to prevent cancellation of a provided reference, because it might come in handy (eg someone filling in a bogus reference).

## Link to Trello card

https://trello.com/c/5rQqoe5P/1148-support-change-2x-referees-email-address

## Things to check

### DIFF
--- a/app/components/candidate_interface/additional_referees_start_component.rb
+++ b/app/components/candidate_interface/additional_referees_start_component.rb
@@ -28,6 +28,8 @@ module CandidateInterface
         "Our email requesting a reference didn’t reach #{referee.name}"
       when 'feedback_refused'
         "#{referee.name} said they won’t give a reference"
+      when 'cancelled'
+        "We've cancelled the reference request for #{referee.name}"
       else
         "#{referee.name} did not respond to our request"
       end
--- a/app/views/support_interface/references/cancel.html.erb
+++ b/app/views/support_interface/references/cancel.html.erb
@@ -1,0 +1,38 @@
+<% content_for :browser_title, 'Cancel reference' %>
+<% content_for :before_content, govuk_back_link_to(support_interface_application_form_path(@reference.application_form)) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">
+      Are you sure you want to cancel this reference?
+    </h1>
+
+    <p class="govuk-body">
+      Cancelling a reference will:
+    </p>
+
+    <ul class="govuk-list govuk-list--bullet">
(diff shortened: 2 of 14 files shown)


app/components/candidate_interface/additional_referees_start_component.rb
```diff
@@ -28,6 +28,8 @@ def reason_for_replacement(referee)
         "Our email requesting a reference didn’t reach #{referee.name}"
       when 'feedback_refused'
         "#{referee.name} said they won’t give a reference"
+      when 'cancelled'
+        "We've cancelled the reference request for #{referee.name}"
```
tvararu:
```suggestion
        "We’ve cancelled the reference request for #{referee.name}"
```

app/views/support_interface/references/cancel.html.erb
```diff
@@ -0,0 +1,38 @@
...
+      Cancelling a reference will:
+    </p>
+
+    <ul class="govuk-list govuk-list--bullet">
+      <li>prevent this referee from providing a reference</li>
+      <li>allow the candidate to provide a replacement referee</li>
+    </ul>
+
+    <p class="govuk-body">
+      Cancelling a reference does <strong>not send email notifications</strong>.
```
tvararu: Non-blocking design crit: [design system says to use bold sparingly](https://govuk-design-system-preview.netlify.com/styles/typography/#bold-text), not sure if it is worth it here. cc @fofr

spec/system/support_interface/cancel_reference_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  include DfESignInHelpers
+
+  scenario 'Support agent cancels reference' do
+    given_i_am_a_support_user
+    and_there_is_an_application
+    and_i_visit_the_application
+
+    when_i_click_to_cancel_the_reference
+    and_i_confirm_i_really_want_to_cancel_the_reference
+    then_the_reference_is_cancelled
```
tvararu: Thought it would have been cool if this test went on to check that the candidate is then able to add a replacement referee, but understand that would be out of the scope of this particular system test which is support-only

PR #1584, opened 2020-03-09
Clearer feature flag name

benilovj:
## Context

PR #1495 introduced a feature-toggled banner to show to provider users when DfE Sign-in was down. However, the feature flag name isn't self-explanatory when skimming down the flags on the Features page.

Before (slightly unfortunate wording on the button):

image

After:

image

## Guidance to review

It should be OK to rename the flag because it's turned off in production?

## Things to check


benilovj:
The Azure DevOps pipeline seems to be failing on an unrelated flaky test.

PR #1589, opened 2020-03-10
Link audit entries to provider users rather than just email addresses

stevehook:
## Context

At the moment we only record the email of the provider user. If we want to get eg the person's name we need to look the provider user up by email is unreliable and awkward.

- [x] Attribute provider audit actions to the current provider user instead of the email address
- [x] Update existing audit trail entries

## Guidance to review

- Is a rake task the right way to implement the upgrade? (I don't think it belongs in a migration because the data volume is pretty much unbounded so I'd be concerned about how long it takes).
- Not sure how to flag up 'manual post-deployment' tasks. Elsewhere I've seen release notes used to list things that the deployers need to do. Should we have a standard section on the Trello card?
- Another thing worth flagging is that the rake task and associated service can and should be deleted after they've served their purpose to avoid unnecessary code bloat. Should we just add a new Trello card?

## Link to Trello card

https://trello.com/c/urHsv4dj/1742-make-provider-auditing-associate-actions-with-provideruser-records

## Things to check

### DIFF
--- a/app/services/fix_provider_audits.rb
+++ b/app/services/fix_provider_audits.rb
@@ -1,0 +1,25 @@
+class FixProviderAudits
+  def initialize
+    @provider_users = {}
+  end
+
+  def call
+    Audited::Audit.where("username like '%(Provider)'").find_each do |audit|
+      email_address = email_address_from(audit)
+      provider_user = find_provider_user_by(email_address)
+      audit.update(user: provider_user) if provider_user
+    end
+  end
+
+private
+
+  def find_provider_user_by(email_address)
+    @provider_users[email_address] ||= ProviderUser.find_by(email_address: email_address)
+    @provider_users[email_address]
+  end
+
+  def email_address_from(audit)
+    match = audit.username.match(/(^.*) \(Provider\)/)
+    match[1] if match
+  end
+end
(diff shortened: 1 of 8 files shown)


app/services/fix_provider_audits.rb
```diff
@@ -0,0 +1,25 @@
+class FixProviderAudits
+  def initialize
+    @provider_users = {}
+  end
+
+  def call
+    Audited::Audit.where("username like '%(Provider)'").find_each do |audit|
+      email_address = email_address_from(audit)
+      provider_user = find_provider_user_by(email_address)
+      audit.update(user: provider_user) if provider_user
```
duncanjbrown: What should we do if there's no user found? It would be good if we could depend on there being something there

app/services/fix_provider_audits.rb
```diff
@@ -0,0 +1,25 @@
...
+    end
+  end
+
+private
+
+  def find_provider_user_by(email_address)
+    @provider_users[email_address] ||= ProviderUser.find_by(email_address: email_address)
```
duncanjbrown: :+1:

app/services/fix_provider_audits.rb
duncanjbrown: Going to check prod to see whether there are any in this state

duncanjbrown: Have checked prod in the following way

```ruby
found_users = Audited::Audit.where('username LIKE ?', '%Provider%')
  .map { |a| a.username.split(' (Provider)').first }
  .uniq
  .map { |e| ProviderUser.find_by(email_address: e) }
```

There are no `nil`s in this array so all the emails have corresponding `ProviderUsers`.

PR #1590, opened 2020-03-10
Show a course full page if selecting a course with no vacancies

malcolmbaig:
## Context

When a course has no vacancies in any of its course options, we want to show a page which informs the user of this fact. This is part of a wider piece of work to prevent candidates wasting time on applications to unavailable courses.

- Add a Course#full? method which checks if any course options have
  vacancies.
- Add the course full page, including links to the Provider contact
  details.
- Update course selection to show this page if the selected course is
  full.

### The course full page

Screenshot 2020-03-10 at 12 00 49

## Guidance to review

- If testing locally or on the review app, navigate to /candidate/application/courses/provider/1/courses/1/full (replacing the IDs as appropriate). This will show the newly introduced page.
- Testing the logical checks will require DB changes. Specifically, a course must have all its course options updated with `vacancy_status: 'no_vacancies'`.

## Link to Trello card

https://trello.com/c/LAh0iGIc/348-dev-stop-candidates-adding-full-courses-to-their-application

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/course_choices_controller.rb
+++ b/app/controllers/candidate_interface/course_choices_controller.rb
@@ -66,7 +66,7 @@ module CandidateInterface
 
       if !@pick_course.open_on_apply?
         redirect_to_ucas
-      elsif @pick_course.full?
+      elsif @pick_course.full? && FeatureFlag.active?('check_full_courses')
         redirect_to candidate_interface_course_choices_full_path(
           @pick_course.provider_id,
           @pick_course.course_id,
--- a/app/controllers/candidate_interface/course_choices_controller.rb
+++ b/app/controllers/candidate_interface/course_choices_controller.rb
@@ -66,6 +66,11 @@ module CandidateInterface
 
       if !@pick_course.open_on_apply?
         redirect_to_ucas
+      elsif @pick_course.full?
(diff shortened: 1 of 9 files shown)


app/controllers/candidate_interface/course_choices_controller.rb
```diff
@@ -83,6 +88,10 @@ def pick_course
       end
     end
 
+    def full
+      @course = Course.where(id: params[:course_id], provider_id: params[:provider_id]).first
```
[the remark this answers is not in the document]
malcolmbaig: It certainly could. There's a subtle difference in behavior when it comes to ordering, but as that's not relevant here and find_by is a bit tidier - updated.

app/controllers/candidate_interface/course_choices_controller.rb
```diff
@@ -83,6 +88,10 @@ def pick_course
       end
     end
 
+    def full
+      @course = Course.find_by(id: params[:course_id], provider_id: params[:provider_id])
```
tijmenb: I think we don't need the provider param here? Then this can be reduced to `Course.find(params[:course_id])`.

malcolmbaig: 🤦‍♂ updated!

PR #1596, opened 2020-03-10
1641 add filtering to the provider ui refactor

willmcb:
## Context

Forth PR as per Theos comment here: #1456 (review)

Refactor of solution to clean up controller and way that the filter_config is encapsulated.

Also reduced amount of information in the filter_config - now just takes `name:` and `text:` values in hash to compute what it should look like. See  `ProvidersApplicationPageState` for implementation of this (note the config hash has been moved here from the `ApplicationChoicesController` and somewhat adapted as per the above.

## Guidance for review

- haven't added unit tests for the `ProvidersApplicationPageState` class as I had in original PR as I think the system specs are pretty comprehensive - happy to write some though if you think this would be valuable.

## Link to Trello card

## Things to check

### DIFF
--- a/app/models/provider_interface/provider_applications_page_state.rb
+++ b/app/models/provider_interface/provider_applications_page_state.rb
@@ -20,6 +20,15 @@ module ProviderInterface
       }[@sort_by]
     end
 
+    def to_h
+      {
+        sort_order: @sort_order,
+        sort_by: @sort_by,
+        filter_visible: @filter_visible,
+        filter_selections: @filter_selections,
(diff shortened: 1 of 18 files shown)


app/models/provider_interface/provider_applications_page_state.rb
```diff
@@ -0,0 +1,109 @@
+module ProviderInterface
+  class ProviderApplicationsPageState
```
duncanjbrown: 🎉

PR #1602, opened 2020-03-11
Simplify CourseOption vacancy_status enum

malcolmbaig:
## Context

This enum currently uses the site vacancy status values provided by the
Find API. Those values are:

  - both_full_time_and_part_time_vacancies
  - full_time_vacancies
  - part_time_vacancies
  - no_vacancies

While Find stores these values at the Site level, in Apply we have the
concept of a CourseOption. Because this is a specific permutation of
course, study_mode, and site, we can get away with a simpler set of
descriptors:

  - vacancies
  - no_vacancies

  - Update the enum as described above
  - Implement a CourseOption method that translates the detailed
    descriptions from Find and persists one of the two simpler statuses
  - Invoke this method from the SyncProviderFromFind service.

## Guidance to review

- The use of a two-value enum rather than a boolean flag is kind of lame, but I've gone with this as it's the simpler change - it doesn't involve any database modifications and can be updated easily if we decide to adopt a different set of status values.

## Link to Trello card

https://trello.com/c/LAh0iGIc

## Things to check

### DIFF
--- a/app/models/course_option.rb
+++ b/app/models/course_option.rb
@@ -12,9 +12,7 @@ class CourseOption < ApplicationRecord
   }
 
   enum vacancy_status: {
-    both_full_time_and_part_time_vacancies: 'both_full_time_and_part_time_vacancies',
-    part_time_vacancies: 'part_time_vacancies',
-    full_time_vacancies: 'full_time_vacancies',
+    vacancies: 'vacancies',
     no_vacancies: 'no_vacancies',
   }
 
--- a/app/services/sync_provider_from_find.rb
+++ b/app/services/sync_provider_from_find.rb
@@ -72,7 +72,7 @@ class SyncProviderFromFind
       site.postcode = find_site.postcode&.strip
       site.save!
 
-      site_status = site_statuses.find { |ss| ss.site.id == find_site.id }
+      find_site_status = site_statuses.find { |ss| ss.site.id == find_site.id }
 
       study_modes = \
         if course.both_study_modes_available?
@@ -88,12 +88,33 @@ class SyncProviderFromFind
           study_mode: mode,
         )
 
-        course_option.update!(vacancy_status: site_status.vac_status)
+        vacancy_status = derive_vacancy_status(
+          status_description: find_site_status.vac_status,
+          study_mode: course_option.study_mode,
+        )
+        course_option.update(vacancy_status: vacancy_status)
       end
     end
 
     course
   end
+
+  def self.derive_vacancy_status(status_description:, study_mode:)
+    case status_description
+    when 'no_vacancies'
+      :no_vacancies
+    when 'both_full_time_and_part_time_vacancies'
+      :vacancies
+    when 'full_time_vacancies'
+      study_mode == 'full_time' ? :vacancies : :no_vacancies
+    when 'part_time_vacancies'
+      study_mode == 'part_time' ? :vacancies : :no_vacancies
+    else
+      raise InvalidFindVacancyStatusError, status_description
+    end
+  end
+
+  class InvalidFindVacancyStatusError < StandardError; end
 
   private_class_method :create_or_update_provider, :create_or_update_course
 end
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -192,7 +192,7 @@ FactoryBot.define do
     course
     site { association(:site, provider: course.provider) }
 
-    vacancy_status { 'both_full_time_and_part_time_vacancies' }
+    vacancy_status { 'vacancies' }
 
     trait :full_time do
       study_mode { :full_time }
--- a/spec/models/course_spec.rb
+++ b/spec/models/course_spec.rb
@@ -33,7 +33,7 @@ RSpec.describe Course, type: :model do
 
     context 'when a subset of course options have vacancies' do
       before do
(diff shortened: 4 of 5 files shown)


app/models/course_option.rb
```diff
@@ -12,12 +12,27 @@ class CourseOption < ApplicationRecord
...
 
   enum vacancy_status: {
-    both_full_time_and_part_time_vacancies: 'both_full_time_and_part_time_vacancies',
-    part_time_vacancies: 'part_time_vacancies',
-    full_time_vacancies: 'full_time_vacancies',
+    vacancies: 'vacancies',
     no_vacancies: 'no_vacancies',
   }
 
+  def update_vacancy_status_from_detailed_description!(description)
```
tijmenb: Could we move this into `SyncProviderFromFind` or a separate service, to avoid models from becoming too large?

app/models/course_option.rb
```diff
@@ -12,12 +12,27 @@ class CourseOption < ApplicationRecord
...
+    no_vacancies! and return if description == 'no_vacancies'
+    vacancies! and return if description == 'both_full_time_and_part_time_vacancies'
```
tijmenb: I had a little trouble following the logic here initially, I think because of the mixing of guard clauses and full `if` statements. Is there a way to write it differently, perhaps a case statement? That would allow you to remove the `and return`.

Also I've never seen the `vacancies!` API used in an enum, not sure if it's worth the suprise vs `update!(vacancy_status: 'vacancies')`?

malcolmbaig: I like the case statement idea. WDYT about this:

```
    case description
    when 'no_vacancies'
      no_vacancies!
    when 'both_full_time_and_part_time_vacancies'
      vacancies!
    when 'full_time_vacancies'
      full_time? ? vacancies! : no_vacancies!
    when 'part_time_vacancies'
      part_time? ? vacancies! : no_vacancies!
    else
      raise InvalidVacancyStatusDescriptionError, description
    end
```

And a variant without the enum methods:

```
    case description
    when 'no_vacancies'
      no_vacancies!
    when 'both_full_time_and_part_time_vacancies'
      vacancies!
    when 'full_time_vacancies'
      full_time? ? update(vacancy_status: :vacancies) : update(vacancy_status: :no_vacancies)
    when 'part_time_vacancies'
      part_time? ? update(vacancy_status: :vacancies) : update(vacancy_status: :no_vacancies)
    else
      raise InvalidVacancyStatusDescriptionError, description
    end

```

On the subject of those methods:

- Both `!` and `?` method variants are a [pretty well-documented](https://api.rubyonrails.org/classes/ActiveRecord/Enum.html) part of Rails' enum feature.
- Their conciseness can be useful in complex statements like the above.

However, happy to get a second opinion on which form is clearer/easier to understand. Let me know!

tvararu: I'm also less familiar with the bang enums but now that I know they exist I like them, the code is shouty but in a friendly kind of way ("yay vacancies!")

tijmenb: @malcolmbaig first one looks good! 👍 

(A version of option 2 could use a method that returns `:vacancies` / `:no_vacancies` and update it in a single place, but this is fine as well).

@tvararu lol, yeah it's like taking a ActiveRecord object by the collar and shouting VACANCIES! at it until it complies.

app/models/course_option.rb
malcolmbaig: Counterpoints:

- Moving behavior like this into the service layer pushes us towards an [Anemic Domain Model](https://martinfowler.com/bliki/AnemicDomainModel.html).
- The method acts solely on the `vacancy_status` field on records in the `course_options` table, deciding how to translate and persist valid input values. This seems like an appropriate responsibility of the `CourseOption` active record, IMHO.
- The method contains a piece of domain logic, rather than doing the coordination and delegating work that service objects typically do according to their definition in DDD.

Alternatives:

- Identify a distinct domain concept and put this logic into a separate model. Something like a `FindApiVacancyDescription`, which instantiates with the string from Find and implements a `vacancy_status(study_mode:)` method.
- Leave the method as-is, refactoring if the model becomes unwieldy at a later time.

tijmenb: Hah, you've thought about this :)

So it's worked out quite well to only add to the model if absolutely necessary. The models are easy to read, classes have few responsibilities and they're easily refactored.

In particular though, I think this method contains detailed knowledge about Find's domain model, which is outside of the responsibilities of the model. It will never be used outside of `SyncProviderFromFind`, so it seems quite clean to keep it in there. I see the translation of attributes from Find -> Apply no different than what we already do, like https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/184c780884bbf8a0599118d9172cd297d11f6464/app/services/sync_provider_from_find.rb#L67-L74

I think our definitions of services is different from others, we've tended to use "service" as a name for all POROs that we have, not just things that do things.

malcolmbaig: > It will never be used outside of SyncProviderFromFind, so it seems quite clean to keep it in there.

Fair point, I can see the value of co-locating this logic.

> I think our definitions of services is different from others, we've tended to use "service" as a name for all POROs that we have, not just things that do things.

Seems like another hint that the organization of that folder could do with a rethink at some point 😉 .

Will get an update pushed shortly.

PR #1609, opened 2020-03-11
Make test application generation spread events out over time

stevehook:
## Context

At the moment an application is created, submitted, receives references, is offered/accepted/enrolled etc all at the same moment. The data is useless for testing the timeline, sorting by date, etc.

- [x] Go back to a semi-random start point then fast forward time by a few days at a time between lifecycle steps.
- [x] Inline all Sidekiq jobs triggered by the test data generator so that they get done at the 'right' time.
- [x] Using built-in Rails test helpers for time travel rather than Timecop so as not to introduce another production dependency.

## Guidance to review

- Is `ActiveSupport::Testing::TimeHelpers` OK to use outside tests? Is it ok in production?

## Link to Trello card

https://trello.com/c/Vn06dsBl/1744-make-test-application-generation-spread-events-out-over-time
## Things to check

### DIFF
--- a/app/lib/test_applications.rb
+++ b/app/lib/test_applications.rb
@@ -12,6 +12,8 @@ module TestApplications
   end
 
   def self.create_application(states:, courses_to_apply_to: nil)
+    Timecop.freeze(rand(30..60).days.ago)
+
     raise ZeroCoursesPerApplicationError.new('You can\'t have zero courses per application') unless states.any?
 
     first_name = Faker::Name.first_name
@@ -63,9 +65,12 @@ module TestApplications
         SubmitApplication.new(application_form).call
         return if states.include? :awaiting_references
 
+        Timecop.travel(Time.zone.now + rand(1..2).days)
         application_form.application_references.each do |reference|
           reference.relationship_correction = [nil, Faker::Lorem.sentence].sample
           reference.safeguarding_concerns = [nil, Faker::Lorem.sentence].sample
+
+          Timecop.travel(Time.zone.now + rand(1..2).days)
 
           ReceiveReference.new(
             reference: reference,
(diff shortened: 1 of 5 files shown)


lib/tasks/generate_test_data.rake
```diff
@@ -5,5 +5,7 @@ end
 
 desc 'Generate test applications for existing providers'
 task generate_test_applications: :environment do
-  GenerateTestApplications.new.perform
+  Sidekiq::Testing.inline! do
```
duncanjbrown: Should this be in the rake task?

spec/lib/test_applications_spec.rb
```diff
@@ -10,6 +10,20 @@
     expect(choices.count).to eq(2)
   end
 
+  it 'creates a realistic timeline' do
```
duncanjbrown: good enough! 😄

lib/tasks/generate_test_data.rake
stevehook: I've moved it inside the `GenerateTestApplications` worker. I think this means that it will apply to provider initiated test data generation as well as the rake task.

app/lib/test_applications.rb
```diff
@@ -12,113 +12,139 @@ def self.generate_for_provider(provider:, courses_per_application:, count:)
...
+      end
     end
   end
 
   def self.put_application_choice_in_state(choice, state)
     # This is only supposed to happen after 7 days, but SendApplicationToProvider
-    # doesn't check the `edit_by` date of the ApplicationChoice
+    fast_forward(7..10)
     SendApplicationToProvider.new(application_choice: choice).call
     choice.update(edit_by: Time.zone.now)
```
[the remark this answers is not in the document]
stevehook: I've changed it to travel to the `edit_by` date unless `Time.zone.now` is already passed that date (due to earlier semi-random fast forwards).

app/lib/test_applications.rb
```diff
@@ -80,45 +89,66 @@ def self.create_application(states:, courses_to_apply_to: nil)
...
 
   def self.put_application_choice_in_state(choice, state)
-    # This is only supposed to happen after 7 days, but SendApplicationToProvider
-    # doesn't check the `edit_by` date of the ApplicationChoice
+    travel_to(choice.edit_by) if choice.edit_by > Time.zone.now
     SendApplicationToProvider.new(application_choice: choice).call
-    choice.update(edit_by: Time.zone.now)
     return if state == :awaiting_provider_decision
 
     if state == :offer
```
willmcb: What is your view on case statements? I think this `if elseif else` might be clearer as one, and would stop the repetition of 'state'.

stevehook: Yes it's as good a candidate for a case statement as you are likely to see. Thanks!

PR #1610, opened 2020-03-11
Fix missing section errors not showing when equality and diversity feature flag is on

chubberlisk:
## Context

The error summary on the "Review your application" page doesn't show when the `equality_and_diversity` feature flag is on which shows the sections that aren't complete.

This PR ensures that the missing section errors shows up by moving the redirect to the `ApplicationFormController` instead of linking to the equality and diversity start page in the view.

## Guidance to review

Try it using the Heroku preview app to submit an incomplete with the `equality_and_diversity` feature flag on and you should see the error summary, something like:

image

## Link to Trello card

https://trello.com/c/EfbebMO0/1151-fix-error-summary-not-showing-on-the-review-your-application-when-equality-and-diversity-feature-is-on

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/application_form_controller.rb
+++ b/app/controllers/candidate_interface/application_form_controller.rb
@@ -32,7 +32,11 @@ module CandidateInterface
       @application_form_presenter = CandidateInterface::ApplicationFormPresenter.new(current_application)
 
       if @application_form_presenter.ready_to_submit?
-        @further_information_form = FurtherInformationForm.new
+        if FeatureFlag.active?('equality_and_diversity')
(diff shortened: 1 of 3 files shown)


chubberlisk:
Ah... Not a super quick thing to fix as I thought.

Test failing because we end up getting looped back to the `ApplicationForm#submit_show` when a candidate says they don't want to fill in the equality and diversity questionnaire. I think we need a way to differentiate between when the candidate has seen the start questionnaire page and when they have chosen to not complete it. Looking at the `equality_and_diversity` field in both scenarios its value is `nil`.

PR #1611, opened 2020-03-11
Redis minimum TLS version change

tpinney:
## Context

On 31st March MS will be removing support for TLS versions 1.0 and 1.1 and will force all users to use version 1.2 as the minimum. This change is to pre-emptively make this change ourselves in a more controlled manner.

This is a service impacting change. The change will actually be made manually through the portal during a quiet period of operation and then this change will be merged in to ensure the templates persist the minimum TLS version advertised by the Redis cache.

Override default value of `minimumTlsVersion` in building block template from version 1.0 to version 1.2

## Guidance to review

N/A

## Link to Trello card

https://trello.com/c/9amwxUUS/1607-set-default-minimum-tls-version-to-12-on-redis-cache

## Things to check

### DIFF
--- a/azure/template.json
+++ b/azure/template.json
@@ -465,6 +465,9 @@
                     "redisCacheCapacity": {
                         "value": "[parameters('redisCacheCapacity')]"
                     },
+                    "minimumTlsVersion": {
+                        "value": "1.2"
+                    },
                     "dataPersistence": {
                         "value": "aof"
                     },


tpinney:
> Looks fine to me. Is there anything we need to check at the next deploy? (Other than making sure Redis is healthy after deploying `staging`).

You can check the advanced settings under the Redis blade on each environment in the Azure portal after deployment to make sure that the minimum TLS version is still set to 1.2 and hasn't been reverted back to a lower version.

PR #1613, opened 2020-03-11
Display additional course info on course review page

davidgisbey:
## Context

We need to enable candidates to differentiate between courses more easily on course review. page e.g. when a candidate applies to the same provider multiple times for the same subject.

Add rows to the courses review page for:
- financial support 
- qualification
- apply from date
- start date
- course length

Before

image

After

image

## Guidance to review

I've added apply from date as the card says to, but I'm pretty sure it's not needed. Unless I'm mistaken you cannot select courses that are currently not available to apply too right?

## Link to Trello card

https://trello.com/c/D4sPmcZI/1106-dev-display-additional-course-info-on-summary-page

## Things to check


app/components/candidate_interface/course_choices_review_component.rb
```diff
@@ -69,6 +74,41 @@ def study_mode_row(course_choice)
...
+        key: 'Financial support',
+        value: course.financial_support,
+      }
+    end
+
+    def qualification_row(course)
+      {
+        key: 'Qualification',
+        value: DisplayQualification.call(qualification: course.qualification),
+      }
```
davidgisbey: Should this service just be called in the sync provider from find service?

I'm not 100% sure how we want this so be persisted in the db.

app/components/candidate_interface/course_choices_review_component.rb
```diff
@@ -69,6 +74,41 @@ def study_mode_row(course_choice)
...
+    end
+
+    def course_length_row(course)
+      {
+        key: 'Course length',
+        value: DisplayCourseLength.call(course_length: course.course_length),
```
davidgisbey: Same with this

davidgisbey:
I seem to remember that apprenticeships are Salaried courses. When we pull from find they're either apprenticeship, salary or fee. Should we be showing all three options? @fofr

It's currently showing qualification and study. mode type.

PR #1617, opened 2020-03-11
Add migration to create ReferenceToken table

chubberlisk:
## Context

Currently, we have a single field `hashed_sign_in_token` on a `Reference`. This means we can only associate a single sign in token with a reference at a time. This causes problems when we send a chaser reference request email as it overwrites the saved token of the first reference request email, rendering the initial email useless.

We therefore want to be able to associate multiple tokens with a reference.

This PR adds a migration to create a `ReferenceToken` table.

## Guidance to review

- Wasn't sure about `t.index :hashed_token, unique: true`, twas added for `VendorApiToken`. Is this needed?

## Link to Trello card

https://trello.com/c/Io6avau8/1119-improve-user-journey-when-referees-click-on-their-first-email-after-a-chaser-has-been-sent

## Things to check


db/schema.rb
```diff
@@ -354,6 +363,7 @@
   add_foreign_key "emails", "application_forms", on_delete: :cascade
   add_foreign_key "provider_agreements", "provider_users"
   add_foreign_key "provider_agreements", "providers"
+  add_foreign_key "reference_tokens", "\"references\"", column: "reference_id", on_delete: :cascade
```
chubberlisk: Any ideas about the `\"`s? 🤔

app/models/application_reference.rb
```diff
@@ -13,6 +13,7 @@ class ApplicationReference < ApplicationRecord
   validates_presence_of :application_form_id
 
   belongs_to :application_form, touch: true
+  has_many :reference_tokens, dependent: :destroy
```
chubberlisk: Rubocop told me off for not having a `dependent`. Not 100% clear if this is correct.

chubberlisk: Hmm... This breaks `rspec ./spec/models/application_reference_spec.rb:7`.

app/models/reference_token.rb
```diff
@@ -0,0 +1,3 @@
+class ReferenceToken < ApplicationRecord
+  belongs_to :reference, class_name: 'ApplicationReference'
```
tvararu:
```suggestion
  belongs_to :application_reference
```

I think the reason we changed from using `reference` for the name is because it can conflict with build in methods in Rails

app/models/application_reference.rb
chubberlisk: Okay, the only way I've managed to satisfy Rubocop and tests is by doing: `has_many :reference_tokens, foreign_key: 'reference_id', inverse_of: :reference, dependent: :destroy`.

:U

tvararu: We found a cleaner way? ¯\_(ツ)_/¯

PR #1624, opened 2020-03-12
Display course options & vacancy status in support UI

tijmenb:
## Context

We're now syncing the vacancy status for courses, which will prevent candidates from applying to courses that are full.

This adds a list of course options & vacancy status to the support UI. It also serves as a better view into what different choices exist for a provider.

## Guidance to review

Long-ass screenshot alert:

image

## Link to Trello card

https://trello.com/c/yBDnIoUS

## Things to check

### DIFF
--- a/app/components/support_interface/course_choices_table_component.html.erb
+++ b/app/components/support_interface/course_choices_table_component.html.erb
@@ -1,0 +1,27 @@
+<table class='govuk-table'>
+  <thead class='govuk-table__head'>
+    <tr class='govuk-table__row'>
+      <th scope='col' class='govuk-table__header'>ID</th>
+      <th scope='col' class='govuk-table__header'>Course</th>
+      <th scope='col' class='govuk-table__header'>Vacancy status</th>
+    </tr>
+  </thead>
+
+  <tbody class='govuk-table__body'>
+    <% course_rows.each do |course_option| %>
+      <tr class='govuk-table__row'>
+        <td class='govuk-table__cell'><%= course_option.id %></td>
+        <td class='govuk-table__cell'><%= govuk_link_to(course_option.course.name_and_code, support_interface_course_path(course_option.course)) %> - <%= course_option.study_mode.humanize %> at <%= course_option.site.name %></td>
+        <td class='govuk-table__cell'>
+          <% if course_option.vacancies? %>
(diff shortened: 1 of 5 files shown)


malcolmbaig:
The screenshot doesn't appear to have the full list of sites - is that a problem with the screenshot, or did the page not actually render them all?

app/components/support_interface/course_choices_table_component.html.erb
```diff
@@ -0,0 +1,27 @@
...
+    </tr>
+  </thead>
+
+  <tbody class='govuk-table__body'>
+    <% course_rows.each do |course_option| %>
+      <tr class='govuk-table__row'>
+        <td class='govuk-table__cell'><%= course_option.id %></td>
+        <td class='govuk-table__cell'><%= govuk_link_to(course_option.course.name_and_code, support_interface_course_path(course_option.course)) %> - <%= course_option.study_mode.humanize %> at <%= course_option.site.name %></td>
+        <td class='govuk-table__cell'>
+          <% if course_option.vacancy_status == 'vacancies' %>
```
malcolmbaig: Alternatively, `course_option.vacancies?`  ⛳️

app/views/support_interface/providers/show.html.erb
```diff
@@ -38,6 +38,9 @@
   <%= render(SupportInterface::ProviderCoursesTableComponent.new(provider: @provider, courses: @provider.accredited_courses)) %>
 <% end %>
 
+<h2 class='govuk-heading-l'>Course choices & vacancies</h2>
+<%= render(SupportInterface::CourseChoicesTableComponent.new(course_options: @provider.course_options.includes(:course, :site))) %>
```
malcolmbaig: Suggest we do the AR query methods in the controller, assigning a separate `@course_options`.

tijmenb: 👍 Updated!

app/components/support_interface/course_choices_table_component.html.erb
tijmenb: 👍 Updated!

tijmenb:
@malcolmbaig updated, cheers. The screenshot seems to have been cut off, the page works.

PR #1627, opened 2020-03-12
Display additional information on course review page

davidgisbey:
## Context

We need to enable candidates to differentiate between courses more easily on course review. page e.g. when a candidate applies to the same provider multiple times for the same subject.

Add rows to the courses review page for:
- available study modes
- description/type
- start date
- course length

Before

image

After

image

## Guidance to review

Financial support is not in scope for this PR and should be a separate card as it requires a substantial amount of work to get it up and running.

image

image

image

image

image

## Link to Trello card

https://trello.com/c/D4sPmcZI/1106-dev-display-additional-course-info-on-summary-page

## Things to check


app/components/candidate_interface/course_choices_review_component.rb
```diff
@@ -62,13 +74,41 @@ def location_row(course_choice)
       }
     end
 
+    def study_mode_options_row(course)
+      {
+        key: 'Available options',
+        value: course.study_mode.humanize,
```
tvararu: Why is this slightly different from what `study_mode_row` does? That one does `.offered_option` as well 🤔

davidgisbey: You're correct it shouldn't be there. I had misunderstood what it was doing. and thought it selected the option the candidate chose rather than the courses study mode.

PR #1628, opened 2020-03-12
Actually fail when brakeman detects issues

tijmenb:
## Context

We are currently running brakeman as part of the build, but when it finds issues we do nothing! This is because `Brakeman.run` in rake will only return something, and not alter the exit code of the rake task. This means that CI won't know something is up.

- Fail on CI when brakeman detects issues 
- Fix the outstanding issue by ignoring it (it's a false positive)

## Things to check

### DIFF
--- a/brakeman.yml
+++ b/brakeman.yml
@@ -1,0 +1,3 @@
+---
+pager: false
+quiet: true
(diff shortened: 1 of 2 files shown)


tijmenb:
Confirmed this now fails:

Screenshot 2020-03-12 at 14 52 50

tijmenb:
@malcolmbaig I think Azure is having some problems? Should be okay now, I can see the output in GitHub Actions at least.

PR #1630, opened 2020-03-12
Add a page that lists all the courses that are full

tijmenb:
## Context

We now sync the vacancy status of courses (https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1561) to prevent applications to full courses (https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1590). We surface the courses that are full by provider (https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1624), but not across the system.

Add a page with all courses that are full. It's a quick thing to properly test the courses-full features on production.

image

It's in the performance section because I couldn't think of anything better.

## Link to Trello card

https://trello.com/c/yBDnIoUS

## Things to check

### DIFF
--- a/app/controllers/support_interface/course_options_controller.rb
+++ b/app/controllers/support_interface/course_options_controller.rb
@@ -1,0 +1,7 @@
+module SupportInterface
+  class CourseOptionsController < SupportInterfaceController
+    def index
+      @course_options = CourseOption.where('vacancy_status != ?', 'vacancies').includes(:course, :site)
+    end
+  end
+end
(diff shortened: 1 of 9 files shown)


app/controllers/support_interface/course_options_controller.rb
```diff
@@ -0,0 +1,7 @@
+module SupportInterface
+  class CourseOptionsController < SupportInterfaceController
+    def index
+      @course_options = CourseOption.where('vacancy_status != ?', 'vacancies').includes(:course, :site)
```
tijmenb: This includes everything non-vacancies to also surface the course options that have a different status accidentally.

PR #1634, opened 2020-03-12
Backfill reference tokens and create a new one when refreshing sign in token

chubberlisk:
## Context

Currently, we have a single field `hashed_sign_in_token` on a Reference. This means we can only associate a single sign in token with a reference at a time. This causes problems when we send a chaser reference request email as it overwrites the saved token of the first reference request email, rendering the initial email useless.

We therefore want to be able to associate multiple tokens with a reference and created a migration to add a new table called `ReferenceToken` in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1617.

This PR:

- adds a migration to create new `ReferenceToken`s for current references that have a `hashed_sign_in_token`
- updates `ApplicationReference#refresh_feedback_token!` to create a `ReferenceToken` with the hashed token for the reference

This will enable us to keep `ReferenceToken` up to date and have backward capability for when we want to solely use `ReferenceToken`.

## Guidance to review

- Should the backfilling migration be in a separate PR?
- Does the tests make sense?
- Should the `update!` and `create!` be in a `Transaction`?

## Link to Trello card

https://trello.com/c/Io6avau8/1119-improve-user-journey-when-referees-click-on-their-first-email-after-a-chaser-has-been-sent

## Things to check

- [x] This code doesn't rely on migrations in the same Pull Request (~~🤷‍♀~~)


tvararu:
> Should the update! and create! be in a Transaction?

That's not a bad idea, yes!

tvararu:
@chubberlisk this code does not rely on migrations in the same PR! If you remove the migration, nothing breaks, because nothing relies on the new behaviour. 😉

PR #1639, opened 2020-03-13
Add type column to references table

C-gyorfi:
## Context

We want to allow candidates to select the type of referee, so that the correct guidance can be shown to the referee on the style of reference that is required, improving the quality of reference.

We want to record weather a referee is an `Academic`  / `Professional` / `School-based` / `Character` reference

This PR adds the `type` column to the `references` table to store referee type information

## Guidance to review

Does the column name `type` make sense or is it too generic? Should we use more specific such as `type_of_referee`,  I decided to use `type` because I think `reference.type` reads better than  `reference.type_of_referee`.

## Link to Trello card

## Things to check


db/migrate/20200313124507_add_type_to_reference.rb
```diff
@@ -0,0 +1,5 @@
+class AddTypeToReference < ActiveRecord::Migration[6.0]
+  def change
+    add_column :references, :type, :string
```
tijmenb: FYI `type` is likely to conflict with Rails' Single Table Inheritance method. 

https://stackoverflow.com/questions/7134559/rails-use-type-column-without-sti

PR #1641, opened 2020-03-13
Disambiguate courses with the same name in the courses dropdown

davidgisbey:
## Context

Courses have a description on Find (such as "PCGE full time") and an accredited body (such as "Big SCITT"). When a course has the same name as another course, it usually has a different course description or accredited body. We should surface these.

On the course dropdown:
- Adds the description if a provider has two courses with the same name
- Adds the accrediting provider if two courses have the same name and description

## Guidance to review

After:

Screenshot 2020-03-25 at 17 03 52

## Link to Trello card

https://trello.com/c/OnwPZPWa/854-dev-disambiguate-courses-in-autocomplete-and-course-listing-page

## Things to check

### DIFF
--- a/app/helpers/select_options_helper.rb
+++ b/app/helpers/select_options_helper.rb
@@ -8,7 +8,7 @@ module SelectOptionsHelper
   def select_course_options(courses)
     [
       OpenStruct.new(id: '', name: t('activemodel.errors.models.candidate_interface/pick_course_form.attributes.code.blank')),
-    ] + courses.map { |course| OpenStruct.new(id: course.id, name: "#{course.name} (#{course.code})") }
+    ] + courses.map { |course| OpenStruct.new(id: course.id, name: course.name) }
   end
 
   def select_provider_options(providers)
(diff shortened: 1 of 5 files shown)


tijmenb:
I wonder if we need a different approach where we save the disambiguated name in the database. We're likely to need the name outside of the dropdown, which means that we may not want to recalculate all names every time we need a single name.

I suspect that we need the disambiguated name everywhere where we use `course.name_and_code`:

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/search?q=course.name_and_code

tvararu:
@tijmenb I think we should wait to have to use the disambiguated name outside of the dropdown before we do that, as it could be subtly different.

Have reviewed this with @fofr who gave it the 👍 from a design perspective, after we added an en-dash between the Course (code) – description. We also discussed customising the autocomplete to display the description and accrediting body on a separate line in the autocomplete, but that was more involved. (see "Custom results" from https://alphagov.github.io/accessible-autocomplete/examples/)

@tijmenb could you give this another look?

tvararu:
@tijmenb nice work, yeah a struct is much better!

PR #1643, opened 2020-03-15
Scrub equality and diversity audit data

tvararu:
## Context

We discovered that we were exposing this data to support users via the audit log.

Scrub equality and diversity audit data with a data migration.

Prevent it from being saved in the future by marking it as an excepted field.

## Guidance to review

Check the migration, no tests for the `except` because we'd be testing the library itself.

Tested this locally, here's how if you want to as well:

- Enable equality and diversity feature flag
- Log in with an unsubmitted but complete application
- Go to submit, fill in one equality and diversity field
- Check that it's in the audit log for that form
- Checkout to this branch
- Run pending migrations
- Check that the audit log is scrubbed
- Fill in another equality and diversity field
- Check that it's not in the audit log

## Link to Trello card

Ship it.

## Things to check

### DIFF
--- a/app/models/application_form.rb
+++ b/app/models/application_form.rb
@@ -112,5 +112,5 @@ class ApplicationForm < ApplicationRecord
     updated_at == created_at
   end
 
-  audited
+  audited except: :equality_and_diversity
 end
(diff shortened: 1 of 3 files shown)


tvararu:
Re-opening with Tijmen's approach.

PR #1646, opened 2020-03-16
Fix email logging

tijmenb:
## Context

In https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1433 we started logging emails in the database. In https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1614 we started tracking the delivery status. It turns out the tracking doesn't work yet. 

This fixes an issue where we aren't sending a `reference` to Notify. This means that we can't match Notify's callbacks to the emails in our database, and most emails are marked with delivery status "Unknown". Commit https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1646/commits/e820da02145702c8622877c490dcaeb593474d55.

In order to debug this, I've added:

- Start logging the status/reference in Logit so we can see which callbacks arrive
- Update the UI for the email log to better accomodate all the information - mostly things that I needed for debugging
- Add a undocumented MVP search, which allows us to search for specific emails by fiddling with the query string. Once @willmcb's work on filtering in the provider UI is complete, we'll replace the MVP with nice filters. 

## Link to Trello card

https://trello.com/c/AFbxsqs4

## Things to check

### DIFF
--- a/app/controllers/support_interface/email_log_controller.rb
+++ b/app/controllers/support_interface/email_log_controller.rb
@@ -2,6 +2,10 @@ module SupportInterface
   class EmailLogController < SupportInterfaceController
     def index
       @emails = Email.order(id: :desc).includes(:application_form).limit(1000)
+
+      %w[to subject mailer mail_template notify_reference application_form_id delivery_status].each do |column|
+        @emails = @emails.where(column => params[column]) if params[column]
+      end
     end
   end
 end
--- a/app/controllers/support_interface/email_log_controller.rb
+++ b/app/controllers/support_interface/email_log_controller.rb
@@ -4,7 +4,9 @@ module SupportInterface
       @emails = Email.order(id: :desc).includes(:application_form).limit(1000)
 
       %w[to subject mailer mail_template notify_reference application_form_id delivery_status].each do |column|
-        @emails = @emails.where(column => params[column]) if params[column]
+        next unless params[column]
+
+        @emails = @emails.where(column => params[column])
       end
     end
   end
--- a/spec/system/support_interface/email_log_spec.rb
+++ b/spec/system/support_interface/email_log_spec.rb
@@ -22,16 +22,24 @@ RSpec.feature 'Email log' do
   end
 
   def when_an_email_with_custom_reference_is_sent
+    @candidate = create(:candidate, email_address: 'harry@example.com')
+
     AuthenticationMailer.sign_up_email(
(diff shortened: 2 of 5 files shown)


app/controllers/support_interface/email_log_controller.rb
```diff
@@ -2,6 +2,10 @@ module SupportInterface
   class EmailLogController < SupportInterfaceController
     def index
       @emails = Email.order(id: :desc).includes(:application_form).limit(1000)
+
+      %w[to subject mailer mail_template notify_reference application_form_id delivery_status].each do |column|
+        @emails = @emails.where(column => params[column]) if params[column]
+      end
```
tvararu: I was confused here before I saw `if params[column]` as I thought this is doing ~7 `where` calls, what do you think of bringing it in the front like:

```suggestion
      %w[to subject mailer mail_template notify_reference application_form_id delivery_status]
        .select { |column| params[column] }
        .each { |column| @emails = @emails.where(column => params[column]) }
```

tijmenb: Fixed in a different way.

spec/system/support_interface/email_log_spec.rb
```diff
@@ -22,16 +22,24 @@ def given_i_am_a_support_user
...
+    @candidate = create(:candidate, email_address: 'harry@example.com')
+
     AuthenticationMailer.sign_up_email(
-      candidate: create(:candidate, email_address: 'harry@example.com'),
+      candidate: @candidate,
       token: '123',
     ).deliver_now
+
+    open_email('harry@example.com')
+    expect(current_email.header('reference')).to eql("test-sign_up_email-#{@candidate.id}")
```
C-gyorfi: This one is flaky when running locally, maybe:
```
expect(current_email.header('reference')).to include("sign_up_email-#{@candidate.id}")
```

or 

```
around do |example|
    ClimateControl.modify HOSTING_ENVIRONMENT_NAME: 'example_env' do
      example.run
    end
  end
```

?

C-gyorfi:
@tijmenb Today we tried to deploy and we were unable to receive emails on staging, once I reverted these changes and deployed to Heroku it worked again. I'm not sure what the issue is, but I'll revert this for now so we can continue the deployment. https://app.slack.com/client/T50RK42V7/CAHBLU6ET/thread/CAHBLU6ET-1584442452.055600

PR #1655, opened 2020-03-16
New application timeline in provider interface

stevehook:
## Context

Providers should be able to see a visual timeline of events for a given application. Initially this will just include status transitions. See https://manage-applications-beta.herokuapp.com/

- [x] `ApplicationTimelineComponent` that presents a timeline based on the design above.
- [x] Component previews
- [x] `FindStatusChangeAudits` service class that pulls status changes from the audit trail for a given application.
- [x] Plug the new component into the provider interface's application detail page.
- [x] Update system specs
- [x] Add `timeline` feature flag

image

## Guidance to review

- We need to map each type of state transition to a title, e.g. `unsubmitted -> submitted` is 'Application submitted'.  For now the approach is just a simple mapping (implemented Hash constant that maps the target status to a label. This might need to be refined if (a) we need to hide certain transitions in the timeline (b) the source status is relevant to the title.
- Do the previews work?
- There is a minor issue with the test data generator that means that all status changes are attributed to a candidate rather than the relevant provider. I think this should be addressed in a separate PR.
- For provider attribution we are just displaying the full name of the provider user that made the change. The design includes the provider name as well but this is omitted as the user may be a member of multiple providers.
- For updates that are triggered by background processing (e.g. decline by default) the change is attributed to 'system'.

## Link to Trello card

https://trello.com/c/ZvgOpW5Z/1697-%F0%9F%8F%88-implement-the-new-timeline-design-for-applications-on-the-provider-ui-application-page

## Things to check


app/services/find_status_change_audits.rb
```diff
@@ -0,0 +1,33 @@
...
+    application_choice.audits.select do |audit|
+      status_change?(audit)
+    end
+  end
+
+  def status_change_for(audit)
+    StatusChange.new(
+      audit.audited_changes['status'][1],
+      audit.created_at,
+      audit.user || audit.username,
```
duncanjbrown: I think falling back to `.username` will always cause this to result in "System" downstream? Maybe the calling code in the component should proactively check for a string and render that? Or perhaps we should supply a `User`-like thing in this position?

app/frontend/styles/_timeline.scss
```diff
@@ -0,0 +1,117 @@
+/* ==========================================================================
```
duncanjbrown: @adamsilver made some significant tweaks to the MoJ timeline and recommended we use the prototype version

duncanjbrown:
Taken it for a spin, and it works really well. I think we should fix the test data generation so it doesn't attribute everything to the candidate, as it will make this difficult to product review on QA

app/services/find_status_change_audits.rb
stevehook: I was seeing both Candidate and Provider attributed history items. However I got some odd results and it fell back to 'System' in dev whenever rails reloading the model classes because `audit.user.is_a? Candidate` returned false further downstream.

app/frontend/styles/_timeline.scss
stevehook: So this is the stylesheet copied from the prototype. I think that means that we've got your tweaks @adamsilver - if there are any further tweaks I will have to update and after merging we can just keep the copy up to date I guess.

duncanjbrown: (Sorry, badly phrased comment from me — no pointer intended, I saw your note in the commit message)

app/services/find_status_change_audits.rb
duncanjbrown: I found that too when I spiked this. `user.is_a? Candidate` just didn't work when it came out of an audit. I ended up checking `if audit.user_type == "Candidate"`

duncanjbrown: The mystery was pretty good — iirc `user.class.id` (ie the internal Ruby object id) and `Candidate.id` were identical and everything

stevehook:
> Taken it for a spin, and it works really well. I think we should fix the test data generation so it doesn't attribute everything to the candidate, as it will make this difficult to product review on QA

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1689

PR #1656, opened 2020-03-17
Clean up usage of single hashed token for a reference

chubberlisk:
## Context

In https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1638, we added in the ability to allow a reference to use the token in the initial and chaser reference request email. Before, it was only possible to use the most recent one.

This PR removes usage of `ApplicationReference#hashed_sign_in_token` the single field that stored the most recent token. This means that it is no longer updated when `#refresh_feedback_token!` is called and searched for in `.find_by_unhashed_token`.

It also adds a migration to remove the `hashed_sign_in_token` field from the `reference` table.

## Guidance to review

- Would I be correct to say that we should only merge this in when we have successfully deployed https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1638? 🤔 

## Link to Trello card

https://trello.com/c/Io6avau8/1119-dev-improve-user-journey-when-referees-click-on-their-first-email-after-a-chaser-has-been-sent

## Things to check

### DIFF
--- a/app/models/application_reference.rb
+++ b/app/models/application_reference.rb
@@ -39,20 +39,19 @@ class ApplicationReference < ApplicationRecord
   end
 
   def refresh_feedback_token!
-    unhashed_token, hashed_token = Devise.token_generator.generate(ApplicationReference, :hashed_sign_in_token)
+    unhashed_token, hashed_token = Devise.token_generator.generate(ReferenceToken, :hashed_token)
 
-    ActiveRecord::Base.transaction do
-      update!(hashed_sign_in_token: hashed_token)
-      ReferenceToken.create!(application_reference: self, hashed_token: hashed_token)
-    end
+    ReferenceToken.create!(application_reference: self, hashed_token: hashed_token)
 
     unhashed_token
   end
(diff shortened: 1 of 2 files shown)


chubberlisk:
This is interesting... So the tests are breaking because of how we generate and digest a token using the `hashed_sign_in_token` column:

- `Devise.token_generator.generate(ApplicationReference, :hashed_sign_in_token)`
- `Devise.token_generator.digest(ApplicationReference, :hashed_sign_in_token, unhashed_token)`

Which sort of means we need the column to continue to exist as Devise wants a real column. However, if we changed how we do the above e.g. a different column or something, that will mess up current tokens that exist. :sweat_smile:

Simplest thing would be to remove the migration from the PR, but then this smells - continuing to have a column that isn’t really used… 🤔

tvararu:
@chubberlisk have pushed some commits; yes we'll need the field for a while longer, but we can move the hashing to the new class and search for both tokens to maintain backwards compatibility.

PR #1657, opened 2020-03-17
Improve handling of offering a different course for new and existing offers

mnacos:
### Context

A [recent PR](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1551) introduced a `ChangeOffer` service and a UI for modifying courses for existing offers within the provider console. The ability to modify courses through the vendor API pre-dates this work. Although not well advertised/documented, it has been possible to perform multiple POST calls to `/applications/{application_id}/offer`, modifying the offer attributes such as conditions and course options after the initial offer was first made.

The aim of this work is to align the functionality of the provider interface and the vendor API, so that course modifications within an offer are handled by the same service, i.e. `ChangeOffer`, which will also perform email/Slack notifications.

Introducing `ChangeOffer` to `app/controllers/vendor_api/decisions_controller.rb` should have been a relatively easy task, however the vendor API operates on different assumptions than the provider interface. While the provider interface expects a `CourseOption` id, which is a very specific identifier for a particular {provider, year, course, site/location, study_mode} combination, the vendor API expects a `Course` json structure, consisting of the following keys:

-  `provider_code` (e.g. '2FR')
-  `course_code (e.g. '3CVK')`
-  `study_mode` (e.g. 'full_time')
-  `site_code` (e.g. 'K')
-  `recruitment_cycle_year` (e.g. 2020)

This PR moves this code matching/translation logic from `MakeAnOffer`to another service `GetCourseOptionFromCodes`, which is responsible for identifying the appropriate `CourseOption` id from such parameters. In the process, it fixes a `study_mode`-related future bug which would impact the service once the first part-time courses were opened on Apply.

This allows refactoring of `MakeAnOffer` to accept a `CourseOption` id, while preserving the current API contract around identifying courses via codes. As a result, the public interfaces of`MakeAnOffer`and `ChangeOffer`converge, both accepting optional `course_option_id` and `offer_conditions` keyword params in their constructors and modifying state through `#save`.

This, in turn, enables `app/controllers/vendor_api/decisions_controller.rb` to use these services interchangeably and paves the way for [1770 - Allow providers to offer a different course when they first make their offer](https://trello.com/c/J3gqkQK6).

### Guidance to review

Find an application in `offer` state within the provider interface (e.g. by visiting http://localhost:3000/provider/applications?commit=Apply+filters&sort_by=last-updated&sort_order=desc&filter_selections%5Bstatus%5D%5Bawaiting-provider-decision%5D=awaiting_provider_decision&filter_selections%5Bstatus%5D%5Boffer%5D=offer on your dev enviroment) and note its `id` (e.g. by clicking on the application and noting the `id` present in the url). Use the same process to identify the `id` of a 'New'/`awaiting_provider_decision` application.
image

Furthermore, you'll need codes for other courses run by this provider (e.g. 'Gorse SCITT'), in order to make/modify offers via the API. An easy way of obtaining these is through the 'change offer' screens within the provider interface. Codes are presented alongside the names of courses and sites in the relevant views. For example:

image

Finally, make sure you know/have a Vendor API key associated with the same provider as the one offering the courses associated with the applications you will use for testing.

image

You can then make API calls creating new offers or modifying existing offers using the application choice ids you identified before. The following example uses `curl`:

First, construct an appropriate `json` document to submit as part of the `POST`. For example:

```json
// offer.json
{
  "data": {
    "conditions": [ "First Condition", "Second Condition" ],
    "course": {
      "recruitment_cycle_year": 2020,
      "provider_code": "1N1",
      "course_code": "238T",
      "site_code": "5",
      "study_mode": "full_time"
    }
  },
  "meta": {
    "attribution": {
      "full_name": "Jane Jones",
      "email": "jane.jones@example.com",
      "user_id": "12345"
    },
    "timestamp": "2020-03-01T15:33:49.216Z"
  }
}
```

Then `POST` this document to the appropriate url:

```bash
curl -X POST -H 'Content-Type: application/json' -H 'Authorization: Bearer ko3_sXXi9VqNhgLxXzX1Xq' --data @offer.json http://localhost:3000/api/v1/applications/67/offer
```

You can then inspect the API responses and what happens to these applications.

Otherwise, inspect this PR and run specs: `spec/services/change_offer_spec.rb`, `spec/services/make_an_offer_spec.rb`, `spec/services/get_course_option_from_codes_spec.rb` and `spec/requests/vendor_api/post_make_an_offer_spec.rb`.

### Link to Trello card

[1754 - Update Vendor API to use the new ChangeOffer service if it's modifying an offer](https://trello.com/c/u86uvnXW)

## Things to check

### DIFF
--- a/app/controllers/vendor_api/decisions_controller.rb
+++ b/app/controllers/vendor_api/decisions_controller.rb
@@ -23,21 +23,13 @@ module VendorApi
         course_option = nil
       end
 
-      decision = if application_choice.offer?
-                   ChangeOffer.new(
-                     actor: api_user,
-                     application_choice: application_choice,
-                     course_option_id: course_option&.id,
-                     offer_conditions: params.dig(:data, :conditions),
-                   )
-                 else
-                   MakeAnOffer.new(
-                     actor: api_user,
-                     application_choice: application_choice,
-                     course_option_id: course_option&.id,
-                     offer_conditions: params.dig(:data, :conditions),
-                   )
-                 end
+      service = application_choice.offer? ? ChangeOffer : MakeAnOffer
+      decision = service.new(
+        actor: api_user,
+        application_choice: application_choice,
+        course_option_id: course_option&.id,
+        offer_conditions: params.dig(:data, :conditions),
+      )
 
       respond_to_decision(decision)
     end
--- a/app/controllers/vendor_api/decisions_controller.rb
+++ b/app/controllers/vendor_api/decisions_controller.rb
@@ -18,13 +18,7 @@ module VendorApi
           site_code: course_data[:site_code],
         ).call
 
-        change_offer_form = ProviderInterface::ChangeOfferForm.new step: :update,
-                                                                   application_choice: application_choice,
-                                                                   provider_id: (course_option.course.provider.id if course_option),
-                                                                   course_id: (course_option.course.id if course_option),
-                                                                   course_option_id: (course_option.id if course_option)
-
-        (render_cannot_find_course_option and return) unless change_offer_form.valid?
+        (render_cannot_find_course_option and return) unless course_option && course_option.course.open_on_apply
       else
         course_option = nil
       end
@@ -96,22 +90,20 @@ module VendorApi
     end
 
     def respond_to_decision(decision)
-      begin
-        if decision.save
-          render json: { data: SingleApplicationPresenter.new(application_choice).as_json }
-        else
-          render_validation_errors(decision.errors)
-        end
-      rescue ProviderAuthorisation::NotAuthorisedError => e
-        render status: :unprocessable_entity, json: {
-          errors: [
-            {
-              error: 'NotAuthorisedError',
-              message: "#{e.message} failed",
-            },
-          ],
-        }
+      if decision.save
+        render json: { data: SingleApplicationPresenter.new(application_choice).as_json }
+      else
+        render_validation_errors(decision.errors)
       end
+    rescue ProviderAuthorisation::NotAuthorisedError => e
+      render status: :unprocessable_entity, json: {
+        errors: [
+          {
+            error: 'NotAuthorisedError',
+            message: "#{e.message} failed",
+          },
+        ],
+      }
     end
 
     # Takes errors from ActiveModel::Validations and render them in the API response
--- a/app/controllers/vendor_api/decisions_controller.rb
+++ b/app/controllers/vendor_api/decisions_controller.rb
@@ -7,12 +7,43 @@ module VendorApi
         vendor_api_token_id: @current_vendor_api_token.id,
       )
 
-      decision = MakeAnOffer.new(
-        actor: api_user,
-        application_choice: application_choice,
-        offer_conditions: params.dig(:data, :conditions),
-        course_data: params.dig(:data, :course),
-      )
+      course_data = params.dig(:data, :course)
+
+      if course_data.present?
+        course_option = GetCourseOptionFromCodes.new(
+          provider_code: course_data[:provider_code],
+          course_code: course_data[:course_code],
+          recruitment_cycle_year: course_data[:recruitment_cycle_year],
+          study_mode: course_data[:study_mode],
+          site_code: course_data[:site_code],
+        ).call
+
+        change_offer_form = ProviderInterface::ChangeOfferForm.new step: :update,
+                                                                   application_choice: application_choice,
+                                                                   provider_id: (course_option.course.provider.id if course_option),
+                                                                   course_id: (course_option.course.id if course_option),
+                                                                   course_option_id: (course_option.id if course_option)
+
+        (render_cannot_find_course_option and return) unless change_offer_form.valid?
+      else
+        course_option = nil
+      end
+
+      decision = if application_choice.offer?
+                   ChangeOffer.new(
+                     actor: api_user,
+                     application_choice: application_choice,
+                     course_option_id: course_option&.id,
+                     offer_conditions: params.dig(:data, :conditions),
+                   )
+                 else
+                   MakeAnOffer.new(
+                     actor: api_user,
+                     application_choice: application_choice,
+                     course_option_id: course_option&.id,
+                     offer_conditions: params.dig(:data, :conditions),
+                   )
+                 end
 
       respond_to_decision(decision)
     end
@@ -61,15 +92,25 @@ module VendorApi
   private
 
     def application_choice
-      @application_choice ||= GetApplicationChoicesForProviders.call(providers: current_provider)
-        .find(params[:application_id])
+      @application_choice ||= GetApplicationChoicesForProviders.call(providers: current_provider).find(params[:application_id])
     end
 
     def respond_to_decision(decision)
-      if decision.save
-        render json: { data: SingleApplicationPresenter.new(application_choice).as_json }
-      else
-        render_validation_errors(decision.errors)
+      begin
+        if decision.save
+          render json: { data: SingleApplicationPresenter.new(application_choice).as_json }
+        else
+          render_validation_errors(decision.errors)
+        end
+      rescue ProviderAuthorisation::NotAuthorisedError => e
+        render status: :unprocessable_entity, json: {
+          errors: [
+            {
+              error: 'NotAuthorisedError',
+              message: "#{e.message} failed",
+            },
+          ],
+        }
       end
     end
 
@@ -77,6 +118,17 @@ module VendorApi
     def render_validation_errors(errors)
       error_responses = errors.full_messages.map { |message| { error: 'ValidationError', message: message } }
       render status: :unprocessable_entity, json: { errors: error_responses }
+    end
+
+    def render_cannot_find_course_option
+      render status: :unprocessable_entity, json: {
+        errors: [
+          {
+            error: 'CourseOptionError',
+            message: 'Cannot find an appropriate course option for these codes',
+          },
+        ],
+      }
     end
   end
 end
--- a/app/services/change_offer.rb
+++ b/app/services/change_offer.rb
@@ -1,16 +1,28 @@
 class ChangeOffer
   include ActiveModel::Validations
 
-  def initialize(actor:, application_choice:, course_option_id:)
+  attr_reader :application_choice, :course_option_id
+
+  validates_each :course_option_id do |record, attr, value|
+    record.errors.add(attr, :no_change) if value == record.application_choice.offered_option.id
+  end
+
+  def initialize(actor:, application_choice:, course_option_id:, offer_conditions: nil)
     @application_choice = application_choice
     @course_option_id = course_option_id
+    @offer_conditions = offer_conditions
     @auth = ProviderAuthorisation.new(actor: actor)
   end
 
-  def save!
+  def save
     @auth.assert_can_change_offer! application_choice: @application_choice, course_option_id: @course_option_id
-    if @course_option_id != @application_choice.offered_option.id
-      @application_choice.update!(offered_course_option_id: @course_option_id, offered_at: Time.zone.now)
+    if valid?
+      attributes = {
+        offered_course_option_id: @course_option_id,
+        offered_at: Time.zone.now,
+      }
+      attributes[:offer] = { 'conditions' => @offer_conditions } if @offer_conditions
+      @application_choice.update! attributes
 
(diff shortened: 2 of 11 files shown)


spec/requests/vendor_api/post_make_an_offer_spec.rb
```diff
@@ -121,12 +121,12 @@
...
       expect(response).to have_http_status(422)
       expect(parsed_response).to be_valid_against_openapi_schema('UnprocessableEntityResponse')
-      expect(error_response['message']).to match 'Offered course provider ABC does not exist'
+      expect(error_response['message']).to match 'Cannot find an appropriate course option for these codes'
     end
   end
 
   describe 'offering for application with a decision' do
-    it 'allows amending of existing offers' do
+    it 'allows amending of conditions for existing offers' do
```
mnacos: This has started failing now (with 'cannot find application with ID...'), even though the new scenario which uses the exact same methodology at L175 passes. Any ideas?

app/services/get_course_option_from_codes.rb
```diff
@@ -0,0 +1,65 @@
+class GetCourseOptionFromCodes
```
duncanjbrown: Very nice to have this all tied up in here

app/services/get_course_option_from_codes.rb
```diff
@@ -0,0 +1,65 @@
...
+  include ActiveModel::Validations
+  attr_accessor :provider_code, :course_code, :recruitment_cycle_year, :study_mode, :site_code
+  attr_accessor :provider, :course, :site, :course_option
+
+  validates_each :provider_code do |record, attr, value|
```
duncanjbrown: validates_each is a bit confusing at first here! is there a way around using it? if not, maybe add something to the commit message?

app/controllers/vendor_api/decisions_controller.rb
```diff
@@ -7,12 +7,43 @@ def make_offer
...
+      if course_data.present?
+        course_option = GetCourseOptionFromCodes.new(
+          provider_code: course_data[:provider_code],
+          course_code: course_data[:course_code],
+          recruitment_cycle_year: course_data[:recruitment_cycle_year],
+          study_mode: course_data[:study_mode],
+          site_code: course_data[:site_code],
+        ).call
+
+        change_offer_form = ProviderInterface::ChangeOfferForm.new step: :update,
```
duncanjbrown: It feels a bit suspect that this caller needs to know about the steps in the `ChangeOfferForm` in a different namespace.  Could we somehow pull the part which is actually shared into an independent service?

app/controllers/vendor_api/decisions_controller.rb
```diff
@@ -7,12 +7,43 @@ def make_offer
...
+                                                                   provider_id: (course_option.course.provider.id if course_option),
+                                                                   course_id: (course_option.course.id if course_option),
+                                                                   course_option_id: (course_option.id if course_option)
+
+        (render_cannot_find_course_option and return) unless change_offer_form.valid?
+      else
+        course_option = nil
+      end
+
+      decision = if application_choice.offer?
```
duncanjbrown: Really nice :+1:

app/controllers/vendor_api/decisions_controller.rb
```diff
@@ -61,15 +92,25 @@ def confirm_enrolment
...
-        render json: { data: SingleApplicationPresenter.new(application_choice).as_json }
-      else
-        render_validation_errors(decision.errors)
+      begin
+        if decision.save
+          render json: { data: SingleApplicationPresenter.new(application_choice).as_json }
+        else
+          render_validation_errors(decision.errors)
+        end
+      rescue ProviderAuthorisation::NotAuthorisedError => e
```
duncanjbrown: As the whole method is rescuable, no need for a `begin`? https://stackoverflow.com/a/1542682

app/controllers/vendor_api/decisions_controller.rb
```diff
@@ -78,5 +119,16 @@ def render_validation_errors(errors)
       error_responses = errors.full_messages.map { |message| { error: 'ValidationError', message: message } }
       render status: :unprocessable_entity, json: { errors: error_responses }
     end
+
+    def render_cannot_find_course_option
```
duncanjbrown: It's a shame not to use the existing `render_validation_errors` method. Could we pass `error`s down from the form?

app/services/change_offer.rb
```diff
@@ -1,16 +1,28 @@
 class ChangeOffer
   include ActiveModel::Validations
 
-  def initialize(actor:, application_choice:, course_option_id:)
+  attr_reader :application_choice, :course_option_id
+
+  validates_each :course_option_id do |record, attr, value|
+    record.errors.add(attr, :no_change) if value == record.application_choice.offered_option.id
```
duncanjbrown: What if they're only changing the conditions? It doesn't feel fair to reject that change because they also included the course option which hadn't changed.

app/services/change_offer.rb
```diff
@@ -1,16 +1,28 @@
...
+  def initialize(actor:, application_choice:, course_option_id:, offer_conditions: nil)
     @application_choice = application_choice
     @course_option_id = course_option_id
+    @offer_conditions = offer_conditions
     @auth = ProviderAuthorisation.new(actor: actor)
   end
 
-  def save!
+  def save
     @auth.assert_can_change_offer! application_choice: @application_choice, course_option_id: @course_option_id
```
duncanjbrown: I've just noticed we roll together the right to change conditions with the right to change course. Maybe one to be aware of in future.

app/services/make_an_offer.rb
```diff
@@ -64,9 +63,9 @@ def offer_conditions
   attr_reader :application_choice
 
   def offered_course_option
-    return nil if @course_data.nil?
+    return nil unless @course_option_id && @course_option_id != application_choice.course_option.id
```
duncanjbrown: This seems prudent — it would be a bit untidy to have identical `offered_course_options_id`s/`course_option_id`s hanging around just because vendors had happened to send back the existing course_data in the request body

app/services/make_an_offer.rb
```diff
@@ -78,48 +77,6 @@ def further_conditions
     ]
   end
 
-  def validate_course_data
```
duncanjbrown: :+1:

app/services/provider_authorisation.rb
```diff
@@ -5,21 +5,25 @@ def initialize(actor:)
...
-    @actor.is_a?(SupportUser) || @actor.providers.include?(application_choice.course.provider)
-  end
-
-  def can_change_offer?(application_choice:, course_option_id:)
-    course_option = CourseOption.find course_option_id
+  def can_make_offer?(application_choice:, course_option_id: nil)
+    supplied_course_option = CourseOption.find(course_option_id) if course_option_id
     if @actor.is_a?(SupportUser)
       true
+    elsif supplied_course_option && course_option_id != application_choice.course_option.id
```
duncanjbrown: If we check for a dupe in here, do we need to also check it in `MakeAnOffer#offered_course_option`?

app/services/provider_authorisation.rb
```diff
@@ -5,21 +5,25 @@ def initialize(actor:)
...
+      application_choice_belongs_to_user_providers?(application_choice: application_choice) && \
```
duncanjbrown: Right, so you can't transfer _off_ a course you don't control. That seems correct.

spec/requests/vendor_api/post_make_an_offer_spec.rb
mnacos: Turns out the problem isn't the test failing, it's the expectation. For existing offers we now call `ChangeOffer` (which was the main goal of this work!), and this service checks if `course_option_id` is being changed before making any changes, to avoid the API resetting DBD at will.

I think it's fine as it is. Currently we do not allow changes to conditions for an existing offer in the UI. If we start providing this functionality, we can revisit such modifications through the API and discuss what should happen with DBD etc.

app/services/change_offer.rb
mnacos: Yes, see also my comment on the failing test. Changing an offer right now means changing the course_option_id and (optionally) the associated conditions and we don't support changing the conditions alone, hence the permission is accurate.

app/services/provider_authorisation.rb
mnacos: We do :)

app/services/get_course_option_from_codes.rb
mnacos: Yes, I'd imagine it's a little confusing at first, but weirdly this is most direct way I could find validating all attributes against each other and populating the service with specific validation errors. I've done something similar in `ChangeOfferForm`. I think the underlying issue is validation for each attribute depends on the values of the other attributes of the instance.

See also: https://books.google.co.uk/books?id=GL2kAwAAQBAJ&pg=PA244&lpg=PA244&dq=alternatives+to+validates_each+rails&source=bl&ots=fAcRulJioj&sig=ACfU3U1UyXLx_oPCLcQlh43P7KggLWNvsA&hl=en&sa=X&ved=2ahUKEwixztXo4KjoAhWCYcAKHa4MDIEQ6AEwBXoECAkQAQ#v=onepage&q=alternatives%20to%20validates_each%20rails&f=false

app/controllers/vendor_api/decisions_controller.rb
mnacos: Yes, I think you're right. I compared the logic of `GetCourseOptionForCodes` and `ChangeOfferForm` and they both check that all attributes have been set and correspond to each other. The only additional value we get from using `ChangeOfferForm` here is a check on whether the new course is `open_on_apply`. So, I'll add an explicit check for that in the controller and remove this `ChangeOfferForm` dependency.

app/controllers/vendor_api/decisions_controller.rb
```diff
@@ -7,12 +7,43 @@ def make_offer
...
+                                                                   course_id: (course_option.course.id if course_option),
+                                                                   course_option_id: (course_option.id if course_option)
+
+        (render_cannot_find_course_option and return) unless change_offer_form.valid?
+      else
+        course_option = nil
+      end
+
+      decision = if application_choice.offer?
+                   ChangeOffer.new(
```
steventux: #### Coding style suggestion

`ChangeOffer` and `MakeAnOffer` constructors accept the same params so this could be reduced to:

```ruby
operation = application_choice.offer? ? ChangeOffer : MakeAnOffer
operation.new(
  actor: api_user,
  course_option_id: course_option&.id,
  offer_conditions: params.dig(:data, :conditions),
)
```

app/services/get_course_option_from_codes.rb
steventux: The majority of this class is validation code which makes me think it might warrant a separate validator class.

I'm possibly missing the point but I don't quite understand the use of `validates_each` here, it's documented as a way to apply the same block of validation logic on multiple attributes: https://guides.rubyonrails.org/active_record_validations.html#validates-each whereas we are applying it to a single attribute in several distinct blocks.
I think if we need validation logic which is conditional based on several record attributes a custom validator would serve this purpose https://guides.rubyonrails.org/active_record_validations.html#custom-validators as `ActiveModel::Validator#validate` receives the record as an argument.

app/services/provider_authorisation.rb
```diff
@@ -5,21 +5,25 @@ def initialize(actor:)
...
   end
 
-  def can_make_offer?(application_choice:)
-    @actor.is_a?(SupportUser) || @actor.providers.include?(application_choice.course.provider)
-  end
-
-  def can_change_offer?(application_choice:, course_option_id:)
-    course_option = CourseOption.find course_option_id
+  def can_make_offer?(application_choice:, course_option_id: nil)
+    supplied_course_option = CourseOption.find(course_option_id) if course_option_id
```
steventux: Suggest we make the first if clause a return statement as this saves performing an unnecessary `find`:

```ruby
return true if @actor is_a?(SupportUser)

if supplied_course_option && course_option_id != application_choice.course_option.id
...
```

app/services/get_course_option_from_codes.rb
steventux: Having discussed this, I think this isn't a significant issue, it's very particular validation for a very specific case and fine as is. :+1:

PR #1665, opened 2020-03-17
Fix bug when users double submit the find course selection form

tvararu:
## Context

If a user comes in from find, adds a course to their application that brings them up to 3 courses, and then goes back with their browser and resubmits the form, they'll get a 500 because `:options_for_site` is not defined in this controller.

Instead, we should display an error message and direct them to the index page for course choices.

Remove the `else` clause and always redirect to the index page, with an error message if appropriate.

## Guidance to review

I couldn't find a way to test this using system specs, because our `rack_test` driver does not allow `page.go_back`. `visit` will reset the session and log you out.

I have tested this manually as such:

- Log in and apply to 2 courses
- Find a course with only 1 site on Find
- Apply from Find for that course, you'll have 3 courses now
- Press back in your browser
- Resubmit the same form
- You should get a 500 (without the changes in this branch)

If you want to give it a go, the system spec to update is `spec/system/candidate_interface/course_selection/candidate_existing_user_with_course_params_spec.rb`.

## Link to Trello card

https://trello.com/c/Mpybdjth/1196-fix-live-bug-when-user-double-submits-the-form-when-they-arrive-from-find-on-a-course-with-only-one-site

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/find_course_selections_controller.rb
+++ b/app/controllers/candidate_interface/find_course_selections_controller.rb
@@ -44,11 +44,11 @@ module CandidateInterface
         course_option_id: course_option_id,
       )
 
-      if pick_site.save
-        redirect_to candidate_interface_course_choices_index_path
-      else
-        render :options_for_site
+      unless pick_site.save
+        flash[:warning] = pick_site.errors.full_messages.first
       end
+
+      redirect_to candidate_interface_course_choices_index_path
     end
 
     def course_selection_params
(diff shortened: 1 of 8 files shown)


app/controllers/candidate_interface/find_course_selections_controller.rb
```diff
@@ -44,11 +44,11 @@ def pick_site_for_course(course, course_option_id)
         course_option_id: course_option_id,
       )
 
-      if pick_site.save
-        redirect_to candidate_interface_course_choices_index_path
-      else
-        render :options_for_site
+      unless pick_site.save
+        flash[:warning] = "You cannot have more than 3 course choices. You must delete a choice if you want to apply to #{course.name_and_code}."
```
duncanjbrown: Should we use an error from the `PickSiteForm` itself? Less responsibility for the controller.

The string's also in `CandidateInterface::SignInController` — i18n?

tvararu: Yes, I was being lazy 👍

PR #1666, opened 2020-03-17
Change underscores in route paths to dashes

tvararu:
## Context

Underscores can be confused with spaces if they are printed out with an underline beneath the URL.

Update the route paths.

## Guidance to review

Should we set up redirects for the old URLs?

## Link to Trello card

None.

## Things to check

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -178,7 +178,8 @@ Rails.application.routes.draw do
         get '/choose' => 'course_choices#have_you_chosen', as: :course_choices_choose
         post '/choose' => 'course_choices#make_choice'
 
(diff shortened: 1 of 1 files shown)


tvararu:
@steventux yes, that's what I was wondering about as well when I suggested redirects. All the involved URLs are fairly "transactional" however, one being a `POST` for example. I don't think strongly one way or the other, if someone else reading this could also chip in one way or the other?

tijmenb:
Yeah, I'd just set up some redirects. It's a good habit to get into anyway. Also solves situation for when people view a page, then we deploy, and then they click a link.

tvararu:
Updated with 301s!

Screenshot 2020-03-18 at 15 30 32

PR #1668, opened 2020-03-17
Send all emails in the background

tijmenb:
## Context

GOV.UK Notify had a bit of temperature this morning and returned a 500. This caused a user to see an error.

https://sentry.io/organizations/dfe-bat/issues/1568618872/?referrer=slack

(Slack convo https://ukgovernmentdfe.slack.com/archives/CN1MCQCHZ/p1584451545393100)

This isn't necessary because we can use Sidekiq to send the email using Rails' magic `deliver_later` method. We did this for one type of email in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1663

- Send all emails in the background
- Make sure the tests work when doing this

## Link to Trello card

https://sentry.io/organizations/dfe-bat/issues/1568618872

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/additional_referees_controller.rb
+++ b/app/controllers/candidate_interface/additional_referees_controller.rb
@@ -32,7 +32,7 @@ module CandidateInterface
       references_to_confirm = not_requested_references.includes(:application_form).to_a
 
       references_to_confirm.each do |reference|
-        RefereeMailer.reference_request_email(current_candidate.current_application, reference).deliver_now
+        RefereeMailer.reference_request_email(current_candidate.current_application, reference).deliver_later
         reference.update!(feedback_status: 'feedback_requested')
       end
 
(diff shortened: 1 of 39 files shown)


duncanjbrown:
@stevehook a short doc sounds like a good idea to me :+1:

duncanjbrown:
Merging as this is good and @tijmenb is off today

tijmenb:
@stevehook the `deliver` method is a bit of a mystery, I can't find it defined anywhere. Even this returns nil:

```rb
> AuthenticationMailer.sign_in_without_account_email(to: 'foo').method(:deliver).source_location
=> nil
```

While:

```rb
> AuthenticationMailer.sign_in_without_account_email(to: 'foo').method(:deliver_later).source_location
=> ["/Users/tijmen/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/actionmailer-6.0.2.1/lib/action_mailer/message_delivery.rb", 93]
```

PR #1669, opened 2020-03-17
Allow candidates to choose referee type

C-gyorfi:
## Context
We want to allow candidates to select the type of referee, so that the correct guidance can be shown to the referee on the style of reference that is required, improving the quality of reference. We want to record weather a referee is an Academic / Professional / School-based / Character reference

- The previous PR added the `type` column to the `reference` table #1639
- This PR allows candidates to choose the reference type
- Next step is to allow candidates to review and change the reference type

Changes in this PR:
- Add RefereeTypeForm so that we can set and update referee type
- Add the referee type page and allow the candidate to choose the type of reference 

## After
image

image

image

image

## Guidance to review

Hopefully, the system specs are good guidance. 

## Link to Trello card

## Things to check

### DIFF
--- a/app/models/candidate_interface/reference/referee_type_form.rb
+++ b/app/models/candidate_interface/reference/referee_type_form.rb
@@ -1,0 +1,19 @@
+module CandidateInterface
+  class Reference::RefereeTypeForm
+    include ActiveModel::Model
+
+    attr_accessor :referee_type
+
+    validates :referee_type, presence: true
+
+    def self.build_from_reference(reference)
+      new(referee_type: reference.referee_type)
+    end
+
+    def save(reference)
+      return false unless valid?
+
+      reference.update!(referee_type: referee_type)
+    end
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -239,8 +239,12 @@ Rails.application.routes.draw do
 
       scope '/referees' do
         get '/' => 'referees#index', as: :referees
-        get '/new' => 'referees#new', as: :new_referee
-        post '/' => 'referees#create'
+
+        get '/type/(:id)' => 'referees#type', as: :referees_type
+        post '/update-type/(:id)' => 'referees#update_type', as: :update_referees_type
+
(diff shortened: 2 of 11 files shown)


config/routes.rb
```diff
@@ -239,8 +239,12 @@
 
       scope '/referees' do
         get '/' => 'referees#index', as: :referees
-        get '/new' => 'referees#new', as: :new_referee
-        post '/' => 'referees#create'
+
+        get '/type/(:id)' => 'referees#type', as: :referees_type
+        post '/update_type/(:id)' => 'referees#update_type', as: :update_referees_type
```
tvararu:
```suggestion
        post '/update-type/(:id)' => 'referees#update_type', as: :update_referees_type
```

app/models/candidate_interface/reference/referee_type_form.rb
```diff
@@ -0,0 +1,19 @@
+module CandidateInterface
+  class Reference::RefereeTypeForm
+    include ActiveModel::Model
+
+    attr_accessor :referee_type
+
+    validates :referee_type, presence: true
```
tvararu: How about a `validates :referee_type, inclusion: { in: REFEREE_TYPES }`?

C-gyorfi: Do you mean using a constant `REFEREE_TYPES` to store the possible types? I've added an enum to the references model in the next PR I'm working on:
```
enum referee_type: {
    academic: 'Academic',
    professional: 'Professional',
    school_based: 'School-based',
    character: 'Character',
  }
```
So it doesn't allow any other values as a referee_type, just like we did with status. What do you think?

C-gyorfi: I'll wait for your comment before merging this @tvararu  :)

tvararu: Whichever way works! You can also add it in the next PR 👍 You can re-use an enum or just an array of strings

PR #1675, opened 2020-03-18
Allow candidates to review and edit the type of their reference

C-gyorfi:
## Context
We want to allow candidates to select the type of referee, so that the correct guidance can be shown to the referee on the style of reference that is required, improving the quality of reference.

- The previous PR allows candidates to choose the reference type  #1669
- This PR is to allow candidates to review and change the reference type

Changes in this PR:
- Define possible referee types enum in ApplicationReference model
- Update RefereesReviewComponent to show Reference type row
- Complete the candidate journey to review and edit the type of their reference

### After
image

## Link to Trello card

## Things to check

### DIFF
--- a/app/components/candidate_interface/referees_review_component.rb
+++ b/app/components/candidate_interface/referees_review_component.rb
@@ -14,10 +14,10 @@ module CandidateInterface
       [
         name_row(referee),
         email_row(referee),
+        (reference_type_row(referee) if FeatureFlag.active?('referee_type')),
         relationship_row(referee),
         feedback_status_row(referee),
-      ]
-        .compact
+      ].compact
     end
 
     def minimum_references
@@ -56,6 +56,15 @@ module CandidateInterface
         value: referee.relationship,
         action: "relationship for #{referee.name}",
         change_path: candidate_interface_edit_referee_path(referee.id),
+      }
+    end
+
+    def reference_type_row(referee)
+      {
+        key: 'Reference type',
+        value: referee.referee_type.capitalize.dasherize || '',
+        action: "reference type for #{referee.name}",
+        change_path: candidate_interface_referees_type_path(referee.id),
       }
     end
 
--- a/app/controllers/candidate_interface/referees_controller.rb
+++ b/app/controllers/candidate_interface/referees_controller.rb
@@ -12,15 +12,31 @@ module CandidateInterface
     end
 
     def type
-      @reference_type_form = Reference::RefereeTypeForm.new
+      if params[:id]
+        set_referee_id
+
+        @reference_type_form = Reference::RefereeTypeForm.build_from_reference(@referee)
+      else
+        @reference_type_form = Reference::RefereeTypeForm.new
+      end
     end
 
     def update_type
       @reference_type_form = Reference::RefereeTypeForm.new(referee_type: referee_type_param)
 
-      return render :type unless @reference_type_form.valid?
+      if params[:id]
+        set_referee_id
 
-      redirect_to candidate_interface_new_referee_path(type: referee_type_param)
+        return redirect_to action: 'type', id: @id unless @reference_type_form.valid?
+
+        @reference_type_form.save(@referee)
+
+        redirect_to candidate_interface_review_referees_path
+      else
+        return render :type unless @reference_type_form.valid?
+
+        redirect_to candidate_interface_new_referee_path(type: referee_type_param)
+      end
     end
 
     def new
@@ -72,6 +88,12 @@ module CandidateInterface
                                     .application_references
                                     .includes(:application_form)
                                     .find(params[:id])
+    end
+
+    def set_referee_id
+      set_referee
+
+      @id = @referee.id
     end
 
     def set_referees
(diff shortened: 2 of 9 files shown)


app/views/candidate_interface/referees/type.html.erb
```diff
@@ -3,7 +3,7 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <%= form_with model: @reference_type_form, url: candidate_interface_update_referees_type_path, method: :post do |f| %>
+    <%= form_with model: @reference_type_form, url: candidate_interface_update_referees_type_path(id: @id), method: :post do |f| %>
       <%= f.govuk_error_summary %>
       <h1 class="govuk-heading-xl">
         <% title = @referees.any? ? 'Second referee' : 'First referee' %>
```
chubberlisk: Smol bug, when I:

1. Added my first referee
2. Click `Change` for `Reference type`
3. I see `Second referee` as the heading, I should see `First referee`

Maybe the below will solve this.

```suggestion
        <% title = @referees.one? ? 'First referee' : 'Second referee' %>
```

app/components/candidate_interface/referees_review_component.rb
```diff
@@ -59,6 +62,15 @@ def relationship_row(referee)
       }
     end
 
+    def reference_type_row(referee)
+      {
+        key: 'Reference type',
+        value: referee.referee_type ? t("referee_type.#{referee.referee_type}") : '',
```
chubberlisk: Wondering if it's possible to do:

```suggestion
        value: referee.referee_type.capitalize.dasherize || '',
```
E.g.
```ruby
'school_based'.capitalize.dasherize
=> "School-based"
```

Might remove the need for the translation additions and the enum?

app/components/candidate_interface/referees_review_component.rb
```diff
@@ -11,13 +11,16 @@ def initialize(application_form:, editable: true, heading_level: 2, show_incompl
...
         email_row(referee),
         relationship_row(referee),
         feedback_status_row(referee),
-      ]
-        .compact
+      ].compact
+
+      rows.insert(2, reference_type_row(referee)) if FeatureFlag.active?('referee_type')
+
+      rows
```
chubberlisk: Alternatively, to remove the need for the insert and assignment:

```ruby
[
  name_row(referee),
  email_row(referee),
  (reference_type_row(referee) if FeatureFlag.active?('referee_type')),
  relationship_row(referee),
  feedback_status_row(referee),
].compact
```

app/controllers/candidate_interface/referees_controller.rb
```diff
@@ -12,15 +12,31 @@ def index
...
 
-      return render :type unless @reference_type_form.valid?
+      if params[:id]
+        set_referee_id
+
+        return redirect_to action: 'type', id: @id unless @reference_type_form.valid?
+
+        @reference_type_form.save(@referee)
+
+        redirect_to candidate_interface_review_referees_path
```
chubberlisk: Personally I find it easier to follow a simple `if` statement than a `return ... unless ...`. What do you think of the below instead?

```suggestion
        set_referee_id

        if @reference_type_form.valid?
          @reference_type_form.save(@referee)

          redirect_to candidate_interface_review_referees_path
        else
          render :type
        end
```

spec/components/candidate_interface/referees_review_component_spec.rb
```diff
@@ -94,6 +94,28 @@
       expect(change_email).to eq("Change email address for #{first_referee.name}")
       expect(change_relationship).to eq("Change relationship for #{first_referee.name}")
     end
+
+    context 'When referee type feature flag is on' do
```
chubberlisk: Do we want to test for when the feature flag is off?

app/controllers/candidate_interface/referees_controller.rb
C-gyorfi: I tried to avoid nesting more if statements within `if params[:id]`, we need to render `:type` with param if there is an `id`.

app/views/candidate_interface/referees/type.html.erb
C-gyorfi: Aww, this will cause an issue when you add the first referee, because `@referees.one?` will be false

C-gyorfi: resolved this in 1066d44 along with the issue on the reference page(it was an existing bug it always showed First Referee)

app/components/candidate_interface/referees_review_component.rb
C-gyorfi: I'll keep the enum just to ensure we cannot save an incorrect referee type, but thanks for suggesting `dasherize` I tried to use `humanize` but that just added space, it good to know this! 😄

spec/components/candidate_interface/referees_review_component_spec.rb
C-gyorfi: Added in 0d851b4

PR #1677, opened 2020-03-18
Add UCAS downtime banner wherever we link off to the site

chubberlisk:
## Context

UCAS is planning downtime on 20 - 22nd March so we should inform our users that this is happening as there are a number of places we link off to the site.

This PR adds a banner to inform candidates of the downtime which is hidden under a feature flag called `banner_for_ucas_downtime` wherever we link off to UCAS i.e. wherever we call `UCAS.apply_url`.

### Screenshot

image

## Guidance to review

- Copy and pasta okay for this? 🍝 ~~Could be a component but seems overkill.~~ Much noicer.
- Didn't add a system spec for this. Should I?

## Link to Trello card

https://trello.com/c/vyOK6UCd/1154-update-content-ucas-down-time

## Things to check


chubberlisk:
> No need for a system spec but yes do make this a component 👍 will approve after!

Definitely happier with this now! 🙇‍♀

PR #1679, opened 2020-03-18
Make test application generation spread events out over time - second attempt

stevehook:
## Context

At the moment an application is created, submitted, receives references, is offered/accepted/enrolled etc all at the same moment. The data is useless for testing the timeline, sorting by date, etc.

The first attempt to do this used Rails testing time travel helpers https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1609. This ran into trouble when run on `qa` and we suspect may have interfered with the clock in other jobs (running concurrently in different threads). This PR uses a somewhat less 'clever' way to achieve the same effect.

- [x] Go back to a semi-random start point then fast forward time by a few days at a time between lifecycle steps.
- [x] Keep track of 'application time' in the `TestApplications` as we fast forward and use this time to update key attributes of the `ApplicationForm` and `ApplicationChoice` as well as audit trail entries.

## Guidance to review

- Is there are more elegant way of doing this?
- Note that I've kept the structure of the `travel_to` and `fast_forward` methods from the first attempt, just changed the implementation.

## Link to Trello card

https://trello.com/c/Vn06dsBl/1744-make-test-application-generation-spread-events-out-over-time
## Things to check

### DIFF
--- a/app/lib/test_applications.rb
+++ b/app/lib/test_applications.rb
@@ -60,7 +60,7 @@ module TestApplications
       return if states.include? :unsubmitted
 
       without_slack_message_sending do
-        SubmitApplication.new(application_form).call
+        SubmitApplication.new(application_form, skip_emails: true).call
         return if states.include? :awaiting_references
 
         application_form.application_references.each do |reference|
--- a/app/lib/test_applications.rb
+++ b/app/lib/test_applications.rb
@@ -65,6 +65,7 @@ class TestApplications
       return if states.include? :unsubmitted
 
       without_slack_message_sending do
+        fast_forward(1..2)
         SubmitApplication.new(@application_form, skip_emails: true).call
         @application_form.update_columns(submitted_at: @time)
         @application_form.application_choices.each do |application_choice|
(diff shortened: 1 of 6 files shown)


app/lib/test_applications.rb
```diff
@@ -85,54 +95,112 @@ def self.create_application(states:, courses_to_apply_to: nil)
...
 
-  def self.without_slack_message_sending
+  def without_slack_message_sending
     RequestStore.store[:disable_slack_messages] = true
     yield
     RequestStore.store[:disable_slack_messages] = false
   end
+
+  def travel_to(time)
+    @time = time
```
steventux: Is it worth providing an `attr_reader` for this instance var as it's referenced so often?

app/workers/generate_test_applications.rb
```diff
@@ -4,20 +4,20 @@ class GenerateTestApplications
...
-    TestApplications.create_application states: [:rejected] * 2
-    TestApplications.create_application states: [:offer_withdrawn]
-    TestApplications.create_application states: [:declined]
-    TestApplications.create_application states: [:accepted]
-    TestApplications.create_application states: [:accepted_no_conditions]
-    TestApplications.create_application states: [:recruited]
-    TestApplications.create_application states: [:conditions_not_met]
-    TestApplications.create_application states: [:enrolled]
-    TestApplications.create_application states: [:withdrawn]
+    TestApplications.new.create_application states: [:unsubmitted]
```
duncanjbrown: Would it work to reuse the same instance of `TestApplications` instead of `new`ing up a fresh one each line?

PR #1683, opened 2020-03-19
Audit redaction fix

duncanjbrown:
## Context

We got a production error associated with unexpected data passing through the component which renders audits. The error happened when we tried to redact data from an integer instead of a hash.

https://sentry.io/organizations/dfe-bat/issues/1571505476/?project=1765973&referrer=slack

The fix:
- Guard against trying to treat non-hash things as hashes

Also:
- Redact "create" actions as well as "update" actions
- Be cautious and redact top-level keys as well as nested keys

## Guidance to review

See tests.

## Link to Trello card

https://trello.com/c/BjmrVkde/1800-address-production-error-related-to-scrubbing-audit-data

## Things to check

### DIFF
--- a/app/components/support_interface/audit_trail_change_component.rb
+++ b/app/components/support_interface/audit_trail_change_component.rb
@@ -26,6 +26,8 @@ module SupportInterface
     end
 
     def redact_equality_and_diversity_data(value)
+      return value unless value.is_a? Hash
+
       %w[sex disabilities ethnic_group ethnic_background].each do |field|
         next unless value[field]
 
--- a/app/components/support_interface/audit_trail_change_component.rb
+++ b/app/components/support_interface/audit_trail_change_component.rb
@@ -13,11 +13,10 @@ module SupportInterface
 
     def format_audit_values
       if values.is_a? Array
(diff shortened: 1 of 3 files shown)


app/components/support_interface/audit_trail_change_component.rb
```diff
@@ -12,12 +14,13 @@ def initialize(attribute:, values:, last_change:)
     end
 
     def format_audit_values
-      if values.is_a? Array
-        before = values[0] ? redact_equality_and_diversity_data(values[0]) : 'nil'
-        after = values[1] ? redact_equality_and_diversity_data(values[1]) : 'nil'
+      if REDACTED_ATTRIBUTES.include? @attribute
```
steventux: More a question of code-style preference but could be adapted into return clauses:

```ruby
return '[REDACTED]' if REDACTED_ATTRIBUTES.include?(@attribute)
return values.map { |v| redact_equality_and_diversity_data(v) || 'nil' }.join(' → ') if value.is_a?(Array)

redact_equality_and_diversity_data(values) || 'nil'
```

duncanjbrown: Done. The `join` is better too, thank you!

PR #1686, opened 2020-03-19
Handle invalid sites during provider sync

malcolmbaig:
## Context

Followup to #1676 

When syncing Find data in SyncProviderFromFind, we create or update the
course options for each course. Pre-existing course options can become
invalid if the course is no longer available at a given site.

- Once we've finished syncing data for a given course, identify all invalid
  course options.
- Delete the ones that aren't associated with an ApplicationChoice, as
  these can be removed from the system safely.
- If any are associated with an ApplicationChoice, mark them as
  invalidated_by_find.
- Let Sentry know about the ones we've marked, so these can be investigated.

## Guidance to review

Comments left inline for areas of particular focus.

## Link to Trello card

https://trello.com/c/zUcpQrGi

## Things to check

### DIFF
--- a/app/services/sync_provider_from_find.rb
+++ b/app/services/sync_provider_from_find.rb
@@ -15,17 +15,14 @@ class SyncProviderFromFind
   def call
     @provider = create_or_update_provider
 
-    return unless provider.sync_courses?
+    return unless @provider.sync_courses?
 
     find_provider = fetch_provider_from_find_api
-
     update_provider_details_with_api_response(find_provider)
-
     find_provider.courses.each do |find_course|
       create_or_update_course(find_course)
     end
   end
-
 
 private
 
@@ -57,26 +54,12 @@ private
 
   def create_or_update_course(find_course)
     course = provider.courses.find_or_create_by(code: find_course.course_code)
-    course.name = find_course.name
-    course.level = find_course.level
-    course.study_mode = find_course.study_mode
-    course.description = find_course.description
-    course.start_date = find_course.start_date
-    course.course_length = find_course.course_length
-    course.recruitment_cycle_year = find_course.recruitment_cycle_year
-    course.exposed_in_find = find_course.findable?
+    assign_course_attributes_from_find(course, find_course)
+    add_accrediting_provider(course, find_course[:accrediting_provider])
 
-    if find_course[:accrediting_provider].present?
-      accrediting_provider = Provider.find_or_create_by(code: find_course[:accrediting_provider][:provider_code]) do |accredit_provider|
-        accredit_provider.name = find_course[:accrediting_provider][:provider_name]
-        accredit_provider.save
-      end
-      course.accrediting_provider = accrediting_provider
-    end
     course.save!
 
     site_statuses = find_course.site_statuses
-
     find_course.sites.each do |find_site|
       site = provider.sites.find_or_create_by(code: find_site.code)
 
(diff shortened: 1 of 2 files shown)


app/services/sync_provider_from_find.rb
```diff
@@ -112,7 +95,48 @@ def create_or_update_course(find_course)
...
+    course_options = course.course_options.joins(:site)
+    canonical_site_codes = find_course.sites.map(&:code)
+    invalid_course_options = course_options.where.not(sites: { code: canonical_site_codes })
+
+    return if invalid_course_options.blank?
+
+    chosen_course_option_ids = ApplicationChoice.where(course_option: invalid_course_options).pluck(:course_option_id)
+
+    not_part_of_an_application = invalid_course_options.where.not(id: chosen_course_option_ids)
+    not_part_of_an_application.delete_all
```
malcolmbaig: Not 100% about this - think it's safe given only it's only the `application_choices` table that has a course options foreign key. Should also get a foreign key error if an associated ApplicationChoice gets created by the time we get here. Would appreciate a second opinion.

app/services/sync_provider_from_find.rb
```diff
@@ -112,7 +95,48 @@ def create_or_update_course(find_course)
...
+
+    part_of_an_application = invalid_course_options.where(id: chosen_course_option_ids)
+    part_of_an_application.update_all(invalidated_by_find: true)
+
+    Raven.capture_message(
+      "#{part_of_an_application.count} invalid course options chosen by candidates.",
```
malcolmbaig: This will appear in Slack every sync for as long as invalid course options exist.

1. Too noisy?
2. Is there anything else that warrants including in this message?

app/services/sync_provider_from_find.rb
```diff
@@ -112,7 +95,48 @@ def create_or_update_course(find_course)
       end
     end
 
-    course
+    handle_course_options_with_invalid_sites(course, find_course)
```
malcolmbaig: When discussing this with @tijmenb originally, he noted that study modes were potentially another source of 'invalid' course options. That is, when a course study mode changes from `both_full_time_and_part_time` to `part_time`, let's say, the course options for `full_time` will no longer get picked up by the `CourseOption.find_or_create_by` above. From what I gather, the course study mode can change depending on the availability of vacancies for a given study mode at any given time.

I'm not sure these course options should be handled with the same 'delete or warn' approach taken in `handle_course_options_with_invalid_sites` below. Reasons:

- It's likely that a higher number of these course options will be part of an ApplicationChoice, and there's little value in warning about them because these are completely legitimate picks (whereas if the course site list changes, we probably want to be aware of any in-progress applications that have selected a site that no longer supports the chosen course).
- Deleting course options that aren't part of an ApplicationChoice is feasible, but it doesn't seem necessary - if a course's study mode changes to either FT or PT, we don't give the candidate a choice during course selection anyway.

Possible alternative:

When deciding what the study modes are further up in this method, rather than use `course.both_study_modes_available?`, we do something slightly more involved:

- Does this course have existing course options?
- If so, determine the `study_modes` array from the unique set of study modes among those course options.
- If not, just do the existing check.

This lets us operate against all CourseOptions created for a course historically, irrespective of the current course study mode returned from Find. We then set the vacancy status for each option, which will correspond with the study mode of the course (ie - if the course study mode is `full_time`, all of the `part_time` course options will have `no_vacancies`). WDYT?

duncanjbrown: This sounds sensible to me. My understanding of the issue:

1. We pull a Course (C) with `full_time_or_part_time` `study_mode`.
2. We create options (O1 and O2) with `full_time` and `part_time` `study_modes` respectively.
3. The `study_mode` of C changes to `full_time` because there are no longer any part time places.
4. On next sync, we iterate over the `study_mode`s of C to find the associated `CourseOptions`. We find O2 because it's part time, but don't find O1. Therefore we don't update O1's vacancy status, which is bad because it remains available for applications when it's actually full.

If this is right, the proposed solution sounds sensible.

malcolmbaig: @duncanjbrown that's correct, with particular emphasis on "we iterate over the study_modes of C". Specifically, we determine what the study modes are based on the value of a single enum field on C. We should instead be deriving the study modes from existing options, if they exist.

I'll put the study mode changes in a separate PR, leaving this one focused on the site-related changes.

app/services/sync_provider_from_find.rb
duncanjbrown: I think we can let this stand for now, as if it does blow up then a subsequent sync should correct it? We can keep an eye on Sentry in the meantime

app/services/sync_provider_from_find.rb
duncanjbrown: I think this is ok, and noisy is a good place to start

PR #1687, opened 2020-03-19
Update content on referee form feedback page

C-gyorfi:
## Context
We want to show more specific content/guidance for academic/non-academic references to improve the quality of the references. See Design: 
- https://bat-design-history.netlify.com/apply-for-teacher-training/give-a-reference-second-iteration#reference
- https://bat-design-history.netlify.com/apply-for-teacher-training/give-a-reference-second-iteration#reference-for-an-academic-referee

This PR updates the content on the reference feedback page

### Before
before

### After 
non-academic
after-academic

## Guidance to review

There is a tiny bit of logic to show/hide a paragraph based on academic or non-academic reference:
```ruby
  <% if @reference_form.reference.referee_type.nil? || @reference_form.reference.referee_type == 'academic' %>
```
I tested locally, but there is no system test for this, do you think we need to add to the system tests?
Its currently showing the academic reference content for referees that has no `referee type` because I don't think we can backfill referee types in the DB. Does this make sense?

I did not add feature flag around this because of its a minor content change, but I am open for suggestions! 

## Link to Trello card

https://trello.com/c/l267SMp9/1160-dev-content-change-for-academic-referees

## Things to check

### DIFF
--- a/app/views/referee_interface/reference/feedback.html.erb
+++ b/app/views/referee_interface/reference/feedback.html.erb
@@ -1,4 +1,4 @@
-<% content_for :title, title_with_error_prefix("Give a teacher training reference for #{@application.full_name}", @reference_form.errors.any?) %>
+<% content_for :title, title_with_error_prefix("Your reference for #{@application.full_name}", @reference_form.errors.any?) %>
 <% if FeatureFlag.active?('referee_confirm_relationship_and_safeguarding') %>
   <% content_for :before_content, govuk_back_link_to(referee_interface_safeguarding_path(token: @token_param)) %>
 <% end %>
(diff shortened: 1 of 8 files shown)


app/views/referee_interface/reference/feedback.html.erb
```diff
@@ -7,37 +7,35 @@
...
 
+      <p class="govuk-body">When giving your reference try to cover the following, but don’t worry if there are areas you can’t comment on:</p>
       <ul class="govuk-list govuk-list--bullet govuk-!-margin-bottom-6">
-        <% @application.application_choices.each do |application_choice| %>
-          <li><%= application_choice.course.name %> - <%= application_choice.site.name %></li>
+        <% if @reference_form.reference.referee_type.nil? || @reference_form.reference.referee_type == 'academic' %>
+          <li>
+            <h2 class="govuk-heading-s govuk-!-margin-bottom-1">Academic abilities</h2>
+            <p class="govuk-body">Does the candidate think critically and have good numeracy and literacy skills?</p>
+          </li>
```
chubberlisk: Regarding your question: Yeah, it does make sense to show this if there a reference doesn't have a referee type. I think system spec would be overkill for this. The alternative would be to create a component for this and then we can have coverage for this.

C-gyorfi: I wrapped it into a component as you suggested 🎉👌

spec/components/referee_interface/feedback_hints_component_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'rails_helper'
+
+RSpec.describe RefereeInterface::FeedbackHintsComponent do
+  context 'when its an academic refereence' do
```
chubberlisk:
```suggestion
  context 'when it is an academic reference' do
```

spec/components/referee_interface/feedback_hints_component_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    let(:reference) { build_stubbed(:reference, referee_type: :academic) }
+
+    it 'displays that the relationship is confirmed' do
```
chubberlisk:
```suggestion
    it 'displays the academic abilities guidance' do
```

spec/components/referee_interface/feedback_hints_component_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+      result = render_inline(described_class.new(reference: reference))
+
+      expect(result.text).to include('Academic abilities')
+    end
+  end
+
+  context 'when its non-academic refereence' do
```
chubberlisk:
```suggestion
  context 'when it is non-academic reference' do
```

spec/components/referee_interface/feedback_hints_component_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    let(:reference) { build_stubbed(:reference, referee_type: :school_based) }
+
+    it 'displays that the relationship is confirmed' do
```
chubberlisk:
```suggestion
    it 'does not display that the academic abilities guidance' do
```

spec/components/referee_interface/feedback_hints_component_spec.rb
C-gyorfi: 😅

PR #1693, opened 2020-03-20
Allow time limits to use calendar days instead of business days

tvararu:
## Context

We need `edit_by` to be set using calendar days instead of business days in order to allow RBD and DBD to be flexible if holidays change.

Add `use_business_days` to TimeLimitConfig rule struct, use it in TimeLimitCalculator.

## Guidance to review

Does this make sense?

## Link to Trello card

https://trello.com/c/iV0jnliU/1207-ucas-rbd-dbd-date-suspension-until-20th-april

## Things to check

### DIFF
--- a/app/lib/time_limit_config.rb
+++ b/app/lib/time_limit_config.rb
@@ -1,27 +1,27 @@
 class TimeLimitConfig
-  Rule = Struct.new(:from_date, :to_date, :limit)
+  Rule = Struct.new(:from_date, :to_date, :limit, :use_business_days)
 
   RULES = {
     reject_by_default: [
-      Rule.new(nil, nil, 40),
+      Rule.new(nil, nil, 40, true),
     ],
     decline_by_default: [
-      Rule.new(nil, nil, 10),
+      Rule.new(nil, nil, 10, true),
     ],
     edit_by: [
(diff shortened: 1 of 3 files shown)


tvararu:
This breaks a load of tests at the moment.

willmcb:
Makes sense and looks good 👍

app/lib/time_limit_config.rb
```diff
@@ -1,27 +1,27 @@
...
-      Rule.new(nil, nil, 5),
+      Rule.new(nil, nil, 7, false),
     ],
     chase_provider_before_rbd: [
-      Rule.new(nil, nil, 20),
+      Rule.new(nil, nil, 20, true),
     ],
     chase_referee_by: [
-      Rule.new(nil, nil, 5),
+      Rule.new(nil, nil, 5, true),
```
benilovj: should this also be a calendar day?

app/lib/time_limit_config.rb
```diff
@@ -1,27 +1,27 @@
...
     ],
     replace_referee_by: [
-      Rule.new(nil, nil, 10),
+      Rule.new(nil, nil, 10, true),
```
benilovj: should this also be a calendar day?

PR #1695, opened 2020-03-22
Fix for bugs in qa exposed by RecalculateDates

stevehook:
## Context

We need a way to pause/defer RBD and DBDs until providers can start processing applications again. Some of the work done to achieve this, in particular https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1692, has exposed a couple of problems that led to some errors on `qa`.

See https://sentry.io/organizations/dfe-bat/issues/1573923287/?project=1765973&referrer=slack

I _think_ that the reason for this error is a race condition in `DeclineOfferByDefault` and the `CandidateMailer`. The mailer assumes that there is a declined by default application choice attached to the application form that it is given. However it could (at least in theory) run before the transaction in `DeclineOfferByDefault` gets committed. The mailer call was changed to be async just a few days ago so this didn't matter until very recently.

It also highlighted an issue in `TestApplications` the `offered_at` status is being set to the `submitted_at` date due to some time errors in the logic around `edit_by` in this class.

- [x] Change references to `Time.zone.local` to `time` in `TestApplications` - we should be using the fake time that `TestApplications` manages to spread events across a realistic date range.
- [x] Move the invocation of `CandidateMailer` in `DeclineByDefault` outside the transaction so that it can't run the Sidekiq job until after the transaction has been committed. 

## Guidance to review

- Does the theory and solution to the race condition sound plausible?
- It looks like there are other dates set in `TestApplications` that don't set to reflect the fake timeline. e.g. `DeclineByDefault` sets `decline_by_default_at` based on the current time - it knows nothing of the fake time

## Link to Trello card

https://trello.com/c/iV0jnliU/1207-ucas-rbd-dbd-date-suspension-until-20th-april

## Things to check

### DIFF
--- a/app/services/decline_offer_by_default.rb
+++ b/app/services/decline_offer_by_default.rb
@@ -15,8 +15,8 @@ class DeclineOfferByDefault
           ProviderMailer.declined_by_default(provider_user, application_choice).deliver_later
         end
       end
-
-      CandidateMailer.declined_by_default(application_form).deliver_later
     end
+
+    CandidateMailer.declined_by_default(application_form).deliver_later
   end
 end
--- a/app/services/decline_offer_by_default.rb
+++ b/app/services/decline_offer_by_default.rb
@@ -6,14 +6,19 @@ class DeclineOfferByDefault
   end
 
   def call
+    application_choices = []
+
     ActiveRecord::Base.transaction do
       application_form.application_choices.offer.each do |application_choice|
         application_choice.update!(declined_by_default: true, declined_at: Time.zone.now)
         ApplicationStateChange.new(application_choice).decline_by_default!
+        application_choices << application_choice
+      end
+    end
 
-        application_choice.provider.provider_users.each do |provider_user|
-          ProviderMailer.declined_by_default(provider_user, application_choice).deliver_later
-        end
+    application_choices.each do |application_choice|
+      application_choice.provider.provider_users.each do |provider_user|
+        ProviderMailer.declined_by_default(provider_user, application_choice).deliver_later
       end
     end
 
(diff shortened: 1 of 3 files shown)


app/services/decline_offer_by_default.rb
```diff
@@ -15,8 +15,8 @@ def call
           ProviderMailer.declined_by_default(provider_user, application_choice).deliver_later
         end
       end
-
-      CandidateMailer.declined_by_default(application_form).deliver_later
     end
+
+    CandidateMailer.declined_by_default(application_form).deliver_later
```
duncanjbrown: :+1: This seems correct! This might be a good time to check we don't have irreversible side-effects inside any other transactions

duncanjbrown: Should we move the `ProviderMailer` invocation out here too?

stevehook: Maybe. I didn't change it earlier because the `ProviderMailer` doesn't do any database queries that depend on the state of the application choices - it's given the application choice id explicitly, whereas the `CandidateMailer` has to find the relevant application choices given only the application form.

duncanjbrown: Sure. Does it still seem a bit odd to make this side-effecting call inside the transaction? ie if we added a line below it which went on to throw an exception, the whole txn would be rolled back but the email would still be sent

stevehook: Thats a good point. Anyway I've had a go at pulling it out of the transaction.

stevehook:
@duncanjbrown is this one good to merge now?

PR #1708, opened 2020-03-23
Update provider emails content (COVID-19)

davidgisbey:
## Context

We need to update service notifications (emails) in light of the changes we're making as a result of changing RBD/DBD dates (in response to COVID-19). 

**Application rejected by default**

Before

image

After

image

**Application submitted**

Before

image

After

image

**Chase provider decision**

Before

image

After

image

## Guidance to review

Here's the content https://docs.google.com/document/d/1yjUxi5r9QIxSUamx5jekLRFEE08CPuRLcG017mdw7Og/edit#heading=h.5lh9uriclgum

It says DBD, but I'm pretty sure it should. be using RDB.

## Link to Trello card

https://trello.com/c/lUggdcuq/1215-dev-covid-19-changes-to-service-email-notifications

## Things to check

### DIFF
--- a/app/views/provider_mailer/application_rejected_by_default.text.erb
+++ b/app/views/provider_mailer/application_rejected_by_default.text.erb
@@ -4,7 +4,15 @@ Dear <%= @provider_user_name || 'colleague' %>
 
 <%= @application.candidate_name %> submitted an application for <%= @application.course_name_and_code %> on <%= @application.submitted_at %>.
 
+<% if FeatureFlag.active?('covid_19') %>
+
+This application has been rejected automatically because you did not respond in time.
+
+<% else %>
+
 This application has been rejected automatically because you did not respond within <%= @application.rbd_days %> working days.
+
+<% end %>
 
 # Give feedback or report a problem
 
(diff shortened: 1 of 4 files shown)


app/views/provider_mailer/application_rejected_by_default.text.erb
```diff
@@ -4,7 +4,7 @@ Dear <%= @provider_user_name || 'colleague' %>
 
 <%= @application.candidate_name %> submitted an application for <%= @application.course_name_and_code %> on <%= @application.submitted_at %>.
 
-This application has been rejected automatically because you did not respond within <%= @application.rbd_days %> working days.
```
C-gyorfi: Feature flags? Or we not meant to use feature flags on these?

davidgisbey: I wasn't sure, but i think you're right actually as things will just return to normal (hopefully) afterwards. Adding them noiw

C-gyorfi: Cool, I wasn't sure If I'm over complicating things 😅

spec/mailers/provider_mailer_spec.rb
```diff
@@ -46,39 +46,90 @@
                     'sign in path' => '/provider/sign-in')
   end
 
+    #
+    # describe 'Send provider decision chaser email' do
```
willmcb: Is there a reason to leave this comment in? looks like it was just commented out so it didn't run

davidgisbey: There is not. Thanks for that

PR #1710, opened 2020-03-23
Add a banner on application dashboard for COVID-19 guidance

chubberlisk:
## Context

We want to notify candidates about the potential effects of COVID-19 on their application post-submission.

This PR adds a `Covid19DashboardBannerComponent` component. This will only show when the `covid_19` feature flag is on and the application form has not been enrolled onto a course.

Further, we'll only display the COVID-19 banner when the new referees needed banner is not shown so that two banners aren't shown at the same time.

### Screenshot

image

## Guidance to review

- Does when we show the component make sense? Not enrolled and when new referees aren't needed.
- Testing only showing when new referees aren't need: I initially wanted to create a component something like `DashboardBannerComponent` but struggled with stubbing the components so changed to using system specs to test this. Is this okay?

## Link to Trello card

https://trello.com/c/VEtyaWji

## Things to check

### DIFF
--- a/app/components/candidate_interface/covid19_dashboard_banner_component.html.erb
+++ b/app/components/candidate_interface/covid19_dashboard_banner_component.html.erb
@@ -1,0 +1,9 @@
+<div class="app-banner govuk-!-margin-bottom-7">
+  <div class="app-banner__message">
+    <h2 class="govuk-heading-m">Coronavirus (COVID-19)</h2>
+    <p class="govuk-body">
+      Training providers may take longer to respond to your application.
+      You may also have longer to respond to any offers. Deadlines will show on this page.
+    </p>
+  </div>
+</div>
--- a/app/components/candidate_interface/covid19_dashboard_banner_component.rb
+++ b/app/components/candidate_interface/covid19_dashboard_banner_component.rb
@@ -1,0 +1,11 @@
+module CandidateInterface
+  class Covid19DashboardBannerComponent < ActionView::Component::Base
+    def initialize(application_form:)
+      @application_form = application_form
+    end
+
+    def render?
+      FeatureFlag.active?('covid_19') && @application_form.any_enrolled? == false
+    end
+  end
+end
--- a/app/views/candidate_interface/application_form/complete.html.erb
+++ b/app/views/candidate_interface/application_form/complete.html.erb
@@ -1,6 +1,11 @@
 <% content_for :title, t('page_titles.application_dashboard') %>
 
-<%= render CandidateInterface::NewReferencesNeededComponent.new(application_form: @application_form) %>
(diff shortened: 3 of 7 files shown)


chubberlisk:
Made a very smol content tweak after 👀 from @EmmaFrith. To `You will be able to see your deadlines **below**.` instead of `here`.

app/components/candidate_interface/covid19_dashboard_banner_component.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class="app-banner govuk-!-margin-bottom-7">
+  <div class="app-banner__message">
+    <h2 class="govuk-heading-m">Coronavirus (COVID-19)</h2>
+    <p class="govuk-body">
+      Training providers may take longer to respond to your application.
+      You may also have longer to respond to any offers. You will be able to see your deadlines below.
+    </p>
+  </div>
```
fofr:
```suggestion
    <p class="govuk-body">
      Training providers will take longer to respond to your application.
      You will also have longer to respond to any offers. You will be able to see your deadlines below.
    </p>
  </div>
```

Not sure who authored this content. cc @EmmaFrith

fofr: Suggestion to make the content more assertive.

chubberlisk: I'll check in with Emma F.

chubberlisk:
Need to update this PR regarding https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1713#issuecomment-602702531 about ensuring that this doesn't show when there's a flash message.

app/components/candidate_interface/covid19_dashboard_banner_component.html.erb
EmmaFrith: can't really say will because not everyone will have extended deadlines

EmmaFrith: can't really say will because not everyone will have extended deadlines

paulrobertlloyd: I thought it was best practice not to indicate layout in messages, because for screen reader users and others, words like ‘below’ don’t mean a great deal?

chubberlisk: Ah, should we change this back to "here" then?

EmmaFrith: i would just say 'on this page'

chubberlisk: https://ukgovernmentdfe.slack.com/archives/CN9Q1VB2M/p1585042323027100

PR #1718, opened 2020-03-23
1750 safeguarding data declaration

willmcb:
## Context

Adds message to application page if candidate has opted to declare any safeguarding information.

Adds the following to the provider single application view as per the design in the trello card:

**If there is no declaration:**

Screenshot 2020-03-23 at 15 44 30

**If there is a declaration:**
Screenshot 2020-03-23 at 15 46 24

## Guidance to review

System spec just tests that the component renders if the feature flag is on. There are a unit tests for the component's rendering logic - given this I didn't test permutations for the component in the system test. Do you agree with this approach?

## Link to Trello card

https://trello.com/c/vL0U4x2d/1750-build-safeguarding-data-declaration-yes-no

## Things to check


app/components/provider_interface/safeguarding_declaration_component.rb
```diff
@@ -8,12 +8,8 @@ def initialize(application_form:)
...
     end
 
-    def render?
-      application_form.disclose_disability? && disability_disclosure.present?
-    end
-
-    def disability_disclosure
-      @application_form.disability_disclosure
+    def has_disclosed_safeguarding_issues?
+      @application_form.safeguarding_issues.eql?('Yes') ? true : false
```
chubberlisk: This logic is 75% there, below is the form on the candidate side:

image

A candidate can:

- choose 'Yes' which means `application_form.safeguarding_issues` => `Yes` 
- choose 'No' which means `application_form.safeguarding_issues` => `No` 
- choose 'Yes' and then add details e.g. 'foo' which means `application_form.safeguarding_issues` => `foo` 

So I believe the logic should be:

```suggestion
    def has_disclosed_safeguarding_issues?
      @application_form.safeguarding_issues != 'No'
```

app/components/provider_interface/safeguarding_declaration_component.html.erb
```diff
@@ -0,0 +1,7 @@
+<h2 class="govuk-heading-m govuk-!-margin-top-8" id="qualifications">Criminal convictions and professional misconduct</h2>
```
chubberlisk: Copy and 🍝 ! 😉 

```suggestion
<h2 class="govuk-heading-m govuk-!-margin-top-8">Criminal convictions and professional misconduct</h2>
```

spec/components/provider_interface/safeguarding_declaration_component_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'rails_helper'
+
+RSpec.describe ProviderInterface::SafeguardingDeclarationComponent do
+  context 'when safeguarding issues has a value of "Yes"' do
+    it 'displays "Yes" for sharing safeguarding issues and "Not entered" for relevant information' do
```
chubberlisk:
```suggestion
    it 'displays that the candidate has shared information related to safeguarding' do
```

spec/components/provider_interface/safeguarding_declaration_component_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+      result = render_inline(described_class.new(application_form: application_form))
+
+      expect(result.text).to include('Criminal convictions and professional misconduct')
+      expect(result.text).to include('The candidate has shared information related to safeguarding. Please contact them directly for more details.')
+    end
+  end
+
+  context 'when safeguarding issues has a value of "No"' do
+    it 'displays "No" for sharing safeguarding issues and "Not entered" for relevant information' do
```
chubberlisk:
```suggestion
    it 'displays that the candidate has not shared information related to safeguarding' do
```

spec/components/provider_interface/safeguarding_declaration_component_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'rails_helper'
+
+RSpec.describe ProviderInterface::SafeguardingDeclarationComponent do
```
chubberlisk: We should add another spec to cover when a candidate has entered details for safeguarding e.g. `application_form = build_stubbed(:application_form, safeguarding_issues: 'I have a criminal conviction.')`. This should also display that the candidate has shared information related to safeguarding.

app/services/feature_flag.rb
```diff
@@ -15,6 +15,7 @@ class FeatureFlag
     pilot_open
     provider_application_filters
     provider_change_response
+    safeguarding_declaration
```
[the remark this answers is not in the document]
willmcb: Thanks I've gone with `provider_view_safeguarding` as I guess there may be a reason to turn these on and off independently in provider / candidate - given that one is input whereas one is a view.

app/components/provider_interface/safeguarding_declaration_component.rb
willmcb: 👍

app/components/provider_interface/safeguarding_declaration_component.html.erb
willmcb: 🍝 🤦‍♂

spec/components/provider_interface/safeguarding_declaration_component_spec.rb
willmcb: 🍝 🤦‍♂

spec/components/provider_interface/safeguarding_declaration_component_spec.rb
willmcb: 👍

spec/components/provider_interface/safeguarding_declaration_component_spec.rb
willmcb: 👍

spec/components/provider_interface/safeguarding_declaration_component_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  let(:heading) { 'Criminal convictions and professional misconduct' }
+  let(:has_shared_text) { 'The candidate has shared information related to safeguarding. Please contact them directly for more details.' }
+  let(:has_not_shared_text) { 'No information shared' }
```
chubberlisk: These can potentially be translations. Up to you, I'm happy either way.

willmcb: Think I will leave as is as other headings etc. on the page are inline currently.

PR #1720, opened 2020-03-23
Keep track of the time we sent an application to providers (2/2)

tijmenb:
## Context

Follow up to https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1719 to keep track of when we send an application to a provider.

- Set the attribute when sending to the provider
- Add the attribute to the support UI
- Backfill existing records

## Guidance to review

Is the backfill correct?

## Link to Trello card

https://trello.com/c/iV0jnliU

## Things to check

### DIFF
--- a/db/migrate/20200323161500_backfill_sent_to_provider_at.rb
+++ b/db/migrate/20200323161500_backfill_sent_to_provider_at.rb
@@ -1,0 +1,21 @@
+class BackfillSentToProviderAt < ActiveRecord::Migration[6.0]
+  def up
+    ApplicationChoice.all.each do |application_choice|
+      state_change = application_choice.audits.find do |audit|
+        audit.audited_changes['status'] == %w[application_complete awaiting_provider_decision]
+      end
+
+      next unless state_change
+
+      application_choice.update!(
+        sent_to_provider_at: state_change.created_at,
+        audit_comment: 'Backfill of `sent_to_provider_at`',
+      )
+      puts "Application form #{application_choice.application_form.id} was sent to the provider at #{state_change.created_at}"
+    end
+  end
+
+  def down
+    # Nothing to do
+  end
+end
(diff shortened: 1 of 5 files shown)


db/migrate/20200323161500_backfill_sent_to_provider_at.rb
```diff
@@ -0,0 +1,21 @@
+class BackfillSentToProviderAt < ActiveRecord::Migration[6.0]
+  def up
+    ApplicationChoice.all.each do |application_choice|
+      state_change = application_choice.audits.find do |audit|
+        audit.audited_changes["status"] == ["application_complete", "awaiting_provider_decision"]
+      end
```
tvararu: This is the first time that we're using the audit log for something like this!

db/migrate/20200323161500_backfill_sent_to_provider_at.rb
```diff
@@ -0,0 +1,21 @@
...
+  def up
+    ApplicationChoice.all.each do |application_choice|
+      state_change = application_choice.audits.find do |audit|
+        audit.audited_changes['status'] == %w[application_complete awaiting_provider_decision]
+      end
+
+      next unless state_change
+
+      application_choice.update!(
+        sent_to_provider_at: state_change.created_at,
```
tvararu: Is `state_change.created_at` not the same as `40.business_days.before(application_choice.reject_by_default_at)`? We wouldn't need to use the audit log if this was true.

tvararu: No, I think where this breaks is for applications in `awaiting_provider_decision`... right?

tijmenb: I think not because the `reject_by_default_at` is always at the end of the day, so it's not entirely accurate.

db/migrate/20200323161500_backfill_sent_to_provider_at.rb
tijmenb: But it's definitely why we've added it!

db/migrate/20200323161500_backfill_sent_to_provider_at.rb
tvararu: True

PR #1727, opened 2020-03-24
Update apply through UCAS page

davidgisbey:
## Context

Users currently see the "we are not ready for you" page when they try to add a provider with no courses open on Apply, or when a provider has no courses exposed on Find.

We should be giving them all the information they need and directing them to UCAs

Paul H was worked with the designs to they have been designer approved 👍 

Before

image

When a provider has no courses exposed in find

image

When a course is selected that is exposed in find

image

## Guidance to review

None really.

## Link to Trello card

https://trello.com/c/RQSDgAZD

## Things to check


paulrobertlloyd:
Would there ever be case that a provider is on Apply but doesn’t have any courses listed? (Understand that this can be the case from a technical perspective, but the reality should be different)

davidgisbey:
There are actually quite a few providers who don't have courses exposed on Find in prod @paulrobertlloyd

davidgisbey:
Which is why i'm not sure if we need to tell them to note the provider code down etc. as they won't be on UCAS Apply unless I'm missing something.

PR #1732, opened 2020-03-24
Make the choice between Apply and UCAS clearer

chubberlisk:
## Context

We want to make it easier for the candidate to choose between Apply and UCAS.

This PR updates the content on the Apply or UCAS page and adds radio buttons to choose between Apply and UCAS.

### Before

image

### After

image

## Guidance to review

- ~~Needs refactoring, currently there's duplication in the `ApplyFromFindController`.~~
- There's some spooky injection of HTML for the hint text for the radio buttons. Not sure how else to do this with `GOVUKDesignSystemFormBuilder::FormBuilder`.
- The redirect to UCAS isn't tested. I was hoping to do just check the content of the page for UCAS stuff but that doesn't work because it's not part of the site.
- Review commit by commit.

Any ideas about the above?

## Link to Trello card

https://trello.com/c/Ig9Gco50

## Things to check

### DIFF
--- a/app/views/candidate_interface/apply_from_find/apply_on_ucas_or_apply.html.erb
+++ b/app/views/candidate_interface/apply_from_find/apply_on_ucas_or_apply.html.erb
@@ -1,51 +1,53 @@
-<% content_for :title, 'You can apply for this course using a new GOV.UK service' %>
+<% content_for :title, 'Apply for this course' %>
 <% content_for :service_link, candidate_interface_apply_from_find_path(providerCode: @course.provider.code, courseCode: @course.code) %>
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <%= render CandidateInterface::UcasDowntimeComponent.new  %>
+    <%= render CandidateInterface::UcasDowntimeComponent.new %>
 
-    <h1 class="govuk-heading-xl">
-      <span class="govuk-caption-xl"><%= @course.name_and_code %></span>
-       You can apply for this course using a new GOV.UK service
-    </h1>
+    <%= form_with model: @apply_on_ucas_or_apply, url: candidate_interface_apply_from_find_path, method: :post do |f| %>
+      <%= f.govuk_error_summary %>
 
-    <p class="govuk-body">
-      <%= service_name %> is a streamlined new GOV.UK service which is
-      designed to make application easier. It comes with personalised support
-      for candidates.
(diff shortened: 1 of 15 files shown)


davidgisbey:
Tested it locally it's working as expected 👍

app/views/candidate_interface/apply_from_find/apply_on_ucas_or_apply.html.erb
```diff
@@ -1,51 +1,53 @@
...
+        <%= f.hidden_field :provider_code, value: @course.provider.code %>
+        <%= f.hidden_field :course_code, value: @course.code %>
+
+        <%= f.govuk_radio_buttons_fieldset :service, legend: { size: 'm', text: 'Do you want to apply using a new GOV.UK service?' } do %>
+          <%= f.govuk_radio_button :service, :apply, label: { text: 'Yes, I want to apply using the new service', size: 's' }, link_errors: true, hint_text:
+            "<p class=\"govuk-body govuk-!-margin-top-2\">Choose this option if:</p>
+            <ul class=\"govuk-list govuk-list--bullet\">
+              <li>all your chosen providers are available on the new service</li>
+              <li>you want to try a streamlined service with personalised support</li>
+            </ul>".html_safe %>
```
davidgisbey: image

Just had a little play and this is an option if you would prefer it.

davidgisbey: `      <%= f.govuk_radio_button :service, :apply, label: { text: 'Yes, I want to apply using the new service', size: 's' }, link_errors: true %>
            <p class="govuk-body govuk-!-margin-top- govuk-!-margin-left-8">Choose this option if:</p>

            <ul class= "govuk-list govuk-list--bullet govuk-body govuk-!-margin-top-2 govuk-!-margin-left-8">
              <li>all your chosen providers are available on the new service</li>
              <li>you want to try a streamlined service with personalised support</li>
            </ul>`

chubberlisk: Ah thanks for this! However, although this looks the same, there would be an accessibility issue for screen readers. Because that text is not part of the `govuk-radios__item` and a `govuk-hint govuk-radios__hint`, it won't get read out when a user is on that radio button. So unfortunately going to have to keep the weird HTML safe implementation. :/

PR #1733, opened 2020-03-24
Calculate work gaps based on submission date

stevehook:
## Context

We calculate whether a candidate has a gap at the end of their work history based on when their last job/volunteering ended (as we do now) and the date they submitted their application (rather than today's date).

When we work out whether a candidate has a work gap at the end of the work history we use something like this: `(Time.zone.now - last_job_end_date) > 1.month`. Because today's date is a moving target the outcome of this calculation is fairly likely to change from false to true while the provider is going through their process. Instead we should use `(date_application_submitted - last_job_end_date) > 1.month`
 
- [x] Use `ApplicationForm.submitted_at` as the 'baseline' for work break calculations.
- [x] Update specs including one new initially failing spec

## Guidance to review

- Is there anywhere else in the code for work gaps that uses current date in the same way that I've missed?

## Link to Trello card

https://trello.com/c/9KbrH4yC/1721-calculate-work-gaps-based-on-submission-date

## Things to check

### DIFF
--- a/app/services/work_history_with_breaks.rb
+++ b/app/services/work_history_with_breaks.rb
@@ -18,6 +18,7 @@ class WorkHistoryWithBreaks
   end
 
   def initialize(application_form)
+    @application_form = application_form
     @work_history = application_form.application_work_experiences.sort_by(&:start_date)
     @existing_breaks = application_form.application_work_history_breaks.sort_by(&:start_date)
     @current_job = nil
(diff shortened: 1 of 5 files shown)


app/services/work_history_with_breaks.rb
```diff
@@ -45,6 +46,11 @@ def timeline
 
 private
 
+  def submitted_at
+    # Time.zone.now
```
chubberlisk:
```suggestion
```

stevehook:
@chubberlisk I've updated `app/services/check_breaks_in_work_history.rb` - I think it makes sense to have that working the same way. What do you think?

PR #1739, opened 2020-03-25
Revert "Bump pry from 0.12.2 to 0.13.0"

chubberlisk:
## Context

This bump breaks `guard` which is super useful, see https://github.com/guard/guard/issues/957.

Reverts DFE-Digital/apply-for-postgraduate-teacher-training#1697 so `guard` can be used.


chubberlisk:
Reverting this does bring `guard` back! :)

image

PR #1740, opened 2020-03-25
Ensure all provider users have a first name and last name

duncanjbrown:
## Context

We have relied on DfE Sign-in to validate ProviderUsers having first and last names, but the validation doesn't work (it accepts " ", for example).

This matters because we address provider users by their first and last names in emails, and we email all users affiliated with a given provider when the application changes.

I'll also check the production database to make sure we have nobody in this state. (Done and updated on the Trello card)

Add validation to the form, plus translation strings for the error messages.

## Link to Trello card

https://trello.com/c/vf5xjN1L/1662-ensure-all-provider-users-have-a-first-name-and-last-name

## Things to check


duncanjbrown:
Looking at test failures

PR #1742, opened 2020-03-25
Use the referee as the user in audit log

tijmenb:
## Context

When a referee adds a reference we currently say that it was done by "Unknown user". 

We can fix that by using the reference as the audit user and displaying that in the log.

image

## Things to check


tijmenb:
@stevehook added a test. Probably too much hassle to backfill. Luckily it's clear from the context that the actor is the referee, so we don't really have to.

PR #1745, opened 2020-03-25
Change how study modes are determined in SyncProviderFromFind

malcolmbaig:
## Context

When we upsert course options during the sync, part of the process
involves iterating over available study modes.

Determining available study modes is currently subject to the following
issue:

- We pull a Course (C) with full_time_or_part_time study_mode.
- We create CourseOptions (O1 and O2) with full_time and part_time
  study_modes respectively.
- The study_mode of C changes to full_time because there are no longer
  any part time places.
- On next sync, we use the study_mode of C to find the associated
  CourseOptions. We find O2 because it's part time, but don't find O1.
  Therefore we don't update O1's vacancy status, which is bad because it
  remains available for applications when it's actually full.

(Shout out to @duncanjbrown for an issue description that articulates the problem more clearly than my earlier attempts.)

Fix this by introducing an object that helps determine the appropriate
study modes list. It returns either one or both study modes depending on
the current and historical study mode of the **course.**

## Guidance to review

Changes are probably best understood by reading the specs.

Screenshot 2020-03-25 at 13 51 01

## Link to Trello card

Related to https://trello.com/c/zUcpQrGi 

## Things to check

### DIFF
--- a/app/services/sync_provider_from_find.rb
+++ b/app/services/sync_provider_from_find.rb
@@ -73,13 +73,7 @@ private
 
       find_site_status = site_statuses.find { |ss| ss.site.id == find_site.id }
 
-
-      study_modes = \
-        if course.both_study_modes_available?
-          %i[full_time part_time]
-        else
-          [course.study_mode]
-        end
+      study_modes = CourseStudyModes.new(course).derive
 
       study_modes.each do |mode|
         course_option = CourseOption.find_or_initialize_by(
@@ -162,5 +156,19 @@ private
     end
 
     class InvalidFindStatusDescriptionError < StandardError; end
+  end
+
+  class CourseStudyModes
+    def initialize(course)
+      @course = course
+    end
+
+    def derive
+      both_modes = %w[full_time part_time]
+      return both_modes if @course.both_study_modes_available?
+
+      from_existing_course_options = @course.course_options.pluck(:study_mode).uniq
+      (from_existing_course_options + [@course.study_mode]).uniq
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


app/services/sync_provider_from_find.rb
```diff
@@ -162,4 +157,26 @@ def derive
...
+  class CourseStudyModes
+    def initialize(course)
+      @course = course
+    end
+
+    def derive
+      both_modes = %w[full_time part_time]
+      return both_modes if @course.both_study_modes_available?
+
+      from_existing_course_options = @course.course_options.pluck(:study_mode).uniq
```
stevehook: could you just do something like `return (from_existing_course_options + [@course.study_mode]).uniq` at this point?

duncanjbrown:
> This PR certainly addresses this case but what will happen if a course definition genuinely changes from both modes to full-time only (i.e. no part-time options should exist on our system). Can this happen?

Shall we ask Find?

malcolmbaig:
Had a few other things on the go and just returned to this PR.

> Have I understood correctly that the study_mode reported by Find changes when there are no vacancies for one of the modes?

@mnacos This is what I gather from reading some of the validation code in Find. Even if the value doesn't change as frequently as this assumption might imply, we should probably still have something like this PR to handle cases where the study_mode has been corrected or changed over time for other reasons. 

> My only practical suggestion is could you add a scenario to spec/services/sync_provider_from_find_spec.rb documenting this weird behaviour? i.e. Find changing the study mode according to the vacancy status.

@mnacos Definitely. I think this warrants a system spec as it'll help make a non-intuitive aspect of the sync behaviour a little more obvious. I wrote up a card for that here: https://trello.com/c/adHlQT3h

> Shall we ask Find?

We should! Also made that a part of the card linked above.

app/services/sync_provider_from_find.rb
malcolmbaig: Much nicer, updated!

PR #1747, opened 2020-03-25
Improve guidance around GCSE qualifications

chubberlisk:
## Context

We have had multiple instances of candidates applying to courses they don't meet the entry requirements for. 

This PR adds better guidance around each GCSE subject. The content for all three subjects are slightly different and sometimes depend on the qualification type. A `GcseGradeGuidanceComponent` component was created to deal with this.

Previously, we used for this `GcseQualificationHelper#guidance_for_gcse_edit_grade` but because of the tad complexity around this, a component was opted for instead and the helper removed as a result.

### Some screenshots

image

image

image

## Guidance to review

Would recommend reviewing commit by commit, looking at the tests in particular and then checking alongside the prototype that I've considered all the permutations.

## Link to Trello card

https://trello.com/c/GIpFCnEw

## Things to check

### DIFF
--- a/app/components/candidate_interface/gcse_grade_guidance_component.html.erb
+++ b/app/components/candidate_interface/gcse_grade_guidance_component.html.erb
@@ -1,0 +1,38 @@
+<p class="govuk-body">
+  <%= t('gcse_edit_grade.guidance.main') %>
+</p>
+
+<% if science? %>
+  <p class="govuk-body">
+    <% if scottish_national_5? %>
+      <%= t('gcse_edit_grade.guidance.triple_scottish_national_science') %>
+    <% elsif gcse? || gce_o_level? %>
+      <%= t('gcse_edit_grade.guidance.triple_gcse_science') %>
+    <% end %>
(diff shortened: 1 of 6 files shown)


paulrobertlloyd:
@chubberlisk Spotted two bugs, which may warrant separate PRs as slightly separate from this specific change.

1. The heading of the summary card for these qualifications should reflect the qualification type, _not_ say ‘GCSE’. So if a user selects ‘GCE O Level’, that’s what should appear as the card title. Example of what’s currently shown:

Screenshot 2020-03-26 at 11 38 53

2. On the qualification type page, can we insert an ‘or’ divider before the last option?

Production:
Screenshot 2020-03-26 at 11 39 08

Prototype:
Screenshot 2020-03-26 at 11 39 16

Otherwise, LGTM!

chubberlisk:
Thanks @paulrobertlloyd! In regards to the bugs you spotted, I'm going to create a separate card for those so I can push forward the guidance.

PR #1761, opened 2020-03-26
Fix GCSE qualifications design

C-gyorfi:
#### The bugs
- The heading of the summary card for the qualifications should reflect the qualification type, not say ‘GCSE’. So if a user selects ‘GCE O Level’, that’s what should appear as the card title.

- On the qualification type page, there should be an ‘or’ divider before the last option.
### Before
image

image

### After

image

image

## Guidance to review
👀 

## Link to Trello card
https://trello.com/c/gYaHY1qn/1251-fix-gcse-qualifications-design
## Things to check


paulrobertlloyd:
Love it, such a quick response time!

PR #1762, opened 2020-03-26
Implement the new design for applications in the provider applications index page

willmcb:
## Context

New "application card" design for Provider application choices index page. Will remove the need for page resizing - will remove this in another PR.

**Before:**
Screenshot 2020-03-26 at 14 39 57

**After:**
Screenshot 2020-03-26 at 14 40 50

## Guidance to review

- I have also amended the Covid-19 banner to be full width after speaking to @adamsilver about it (is two thirds on master currently)
- will do another PR to remove page resizing currently the page is 1220px which is larger than the rest of the app.
- Please note that notes are in the design but not implement in the app yet. When notes are added to the app this card will have to be further amended. 

## Link to Trello card

https://trello.com/c/csW7QE92/1810-implement-the-new-design-for-applications-in-the-provider-applications-index-page

## Things to check

### DIFF
--- a/app/components/provider_interface/application_card_component.html.erb
+++ b/app/components/provider_interface/application_card_component.html.erb
@@ -1,20 +1,20 @@
 <div class="app-application-card">
 
   <div class="app-application-card__primary">
-    <h3 class="govuk-heading-m govuk-!-margin-bottom-3">
+    <h3 id="applicant-name-<%= index %>" class="govuk-heading-m govuk-!-margin-bottom-3">
       <%= link_to candidate_name, provider_interface_application_choice_path(application_choice), class: 'govuk-link' %>
     </h3>
-    <p class="govuk-body govuk-!-font-size-16"><%= course_provider_name %></p>
-    <p class="govuk-body govuk-!-margin-bottom-0"><%= course_name_and_code %></p>
-    <p class="govuk-caption-m govuk-!-font-size-16">Accredited body: <%= accrediting_provider %></p>
+    <p id="course-provider-name-<%= index %>" class="govuk-body govuk-!-font-size-16"><%= course_provider_name %></p>
+    <p id="course-name-and-code-<%= index %>" class="govuk-body govuk-!-margin-bottom-0"><%= course_name_and_code %></p>
+    <p id="accredited-body-<%= index %>" class="govuk-body govuk-caption-m govuk-!-font-size-16">Accredited body: <%= accrediting_provider.name if accrediting_provider %></p>
   </div>
 
   <div class="app-application-card__secondary">
-    <div class="app-application-card__status">
+    <div id="status-<%= index %>" class="app-application-card__status">
       <%= render ProviderInterface::ApplicationStatusTagComponent.new(application_choice: application_choice) %>
     </div>
     <div>
-      <p class="govuk-!-margin-bottom-0 govuk-caption-m govuk-!-font-size-16">Updated on <%= updated_at %></p>
+      <p id="updated-at-<%= index %>" class="govuk-!-margin-bottom-0 govuk-caption-m govuk-!-font-size-16">Updated on <%= updated_at %></p>
     </div>
   </div>
 </div>
--- a/app/components/provider_interface/application_card_component.html.erb
+++ b/app/components/provider_interface/application_card_component.html.erb
@@ -13,8 +13,6 @@
     <div id="status-<%= index %>" class="app-application-card__status">
       <%= render ProviderInterface::ApplicationStatusTagComponent.new(application_choice: application_choice) %>
     </div>
-    <div>
-      <p id="updated-at-<%= index %>" class="govuk-!-margin-bottom-0 govuk-caption-m govuk-!-font-size-16">Updated on <%= updated_at %></p>
-    </div>
+    <p id="updated-at-<%= index %>" class="govuk-!-margin-bottom-1 govuk-caption-m govuk-!-font-size-16">Updated on <%= updated_at %></p>
   </div>
 </div>
--- a/app/components/provider_interface/application_card_component.html.erb
+++ b/app/components/provider_interface/application_card_component.html.erb
@@ -1,0 +1,20 @@
+<div class="app-application-card">
+
+  <div class="app-application-card__primary">
+    <h3 class="govuk-heading-m govuk-!-margin-bottom-3">
+      <%= link_to candidate_name, provider_interface_application_choice_path(application_choice), class: 'govuk-link' %>
+    </h3>
+    <p class="govuk-body govuk-!-font-size-16"><%= course_provider_name %></p>
+    <p class="govuk-body govuk-!-margin-bottom-0"><%= course_name_and_code %></p>
+    <p class="govuk-caption-m govuk-!-font-size-16">Accredited body: <%= accrediting_provider %></p>
+  </div>
+
+  <div class="app-application-card__secondary">
+    <div class="app-application-card__status">
+      <%= render ProviderInterface::ApplicationStatusTagComponent.new(application_choice: application_choice) %>
+    </div>
+    <div>
+      <p class="govuk-!-margin-bottom-0 govuk-caption-m govuk-!-font-size-16">Updated on <%= updated_at %></p>
+    </div>
+  </div>
+</div>
--- a/app/components/provider_interface/application_card_component.html.erb
+++ b/app/components/provider_interface/application_card_component.html.erb
@@ -6,7 +6,9 @@
     </h3>
     <p id="course-provider-name-<%= index %>" class="govuk-body govuk-!-font-size-16"><%= course_provider_name %></p>
     <p id="course-name-and-code-<%= index %>" class="govuk-body govuk-!-margin-bottom-0"><%= course_name_and_code %></p>
-    <p id="accredited-body-<%= index %>" class="govuk-body govuk-caption-m govuk-!-font-size-16">Accredited body: <%= accrediting_provider.name if accrediting_provider %></p>
+    <% if accrediting_provider %>
+      <p id="accredited-body-<%= index %>" class="govuk-body govuk-caption-m govuk-!-font-size-16">Accredited body: <%= accrediting_provider.name %> </p>
+    <% end %>
   </div>
 
   <div class="app-application-card__secondary">
--- a/app/components/provider_interface/application_card_component.html.erb
+++ b/app/components/provider_interface/application_card_component.html.erb
@@ -7,7 +7,9 @@
     <p class="govuk-body govuk-!-font-size-16"><%= course_provider_name %></p>
     <p class="govuk-body govuk-!-margin-bottom-0"><%= course_name_and_code %></p>
     <% if accrediting_provider %>
-      <p class="govuk-body govuk-caption-m govuk-!-font-size-16">Accredited body: <%= accrediting_provider.name %> </p>
+      <p class="govuk-body govuk-caption-m govuk-!-font-size-16 app-application-card__accredited-body">
+      Accredited body: <%= accrediting_provider.name %>
+      </p>
     <% end %>
   </div>
 
--- a/app/components/provider_interface/application_card_component.html.erb
+++ b/app/components/provider_interface/application_card_component.html.erb
@@ -1,22 +1,35 @@
-<div data-qa="application-card-<%= index %>"class="app-application-card">
+<div class="app-application-card">
 
-  <div class="app-application-card__primary">
-    <h3 class="govuk-heading-m govuk-!-margin-bottom-3">
-      <%= link_to candidate_name, provider_interface_application_choice_path(application_choice), class: 'govuk-link' %>
-    </h3>
-    <p class="govuk-body govuk-!-font-size-16"><%= course_provider_name %></p>
-    <p class="govuk-body govuk-!-margin-bottom-0"><%= course_name_and_code %></p>
-    <% if accrediting_provider %>
-      <p class="govuk-body govuk-caption-m govuk-!-font-size-16 app-application-card__accredited-body">
-      Accredited body: <%= accrediting_provider.name %>
-      </p>
-    <% end %>
-  </div>
-
-  <div class="app-application-card__secondary">
+  <h3 class="govuk-heading-m">
+    <%= link_to candidate_name, provider_interface_application_choice_path(application_choice), class: 'govuk-link' %>
     <div class="app-application-card__status">
       <%= render ProviderInterface::ApplicationStatusTagComponent.new(application_choice: application_choice) %>
+    <div>
+  </h3>
+
+  <dl>
+    <div class="app-application-card__provider" >
+      <dt class="govuk-visually-hidden">Provider:</dt>
+      <dd class="govuk-body"><%= course_provider_name %></dd>
     </div>
-    <p class="govuk-!-margin-bottom-1 govuk-caption-m govuk-!-font-size-16">Updated on <%= updated_at %></p>
-  </div>
+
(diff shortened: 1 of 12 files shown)


app/components/provider_interface/application_card_component.html.erb
```diff
@@ -0,0 +1,20 @@
...
+    <p id="course-provider-name-<%= index %>" class="govuk-body govuk-!-font-size-16"><%= course_provider_name %></p>
+    <p id="course-name-and-code-<%= index %>" class="govuk-body govuk-!-margin-bottom-0"><%= course_name_and_code %></p>
+    <p id="accredited-body-<%= index %>" class="govuk-body govuk-caption-m govuk-!-font-size-16">Accredited body: <%= accrediting_provider.name if accrediting_provider %></p>
+  </div>
+
+  <div class="app-application-card__secondary">
+    <div id="status-<%= index %>" class="app-application-card__status">
+      <%= render ProviderInterface::ApplicationStatusTagComponent.new(application_choice: application_choice) %>
+    </div>
+    <div>
```
tvararu: Do we need this div?

app/components/provider_interface/application_card_component.html.erb
```diff
@@ -0,0 +1,20 @@
...
+      <p id="updated-at-<%= index %>" class="govuk-!-margin-bottom-0 govuk-caption-m govuk-!-font-size-16">Updated on <%= updated_at %></p>
```
tvararu: There are too many IDs in this component; for testing purposes, we could just have one `id` at the top level and then refer to the nested elements using other CSS selectors.

If it's solely used for testing, I also prefer using `data-qa="application-card-13"` as it communicates the intention better.

app/components/provider_interface/application_card_component.html.erb
```diff
@@ -0,0 +1,20 @@
+<div class="app-application-card">
+
+  <div class="app-application-card__primary">
+    <h3 id="applicant-name-<%= index %>" class="govuk-heading-m govuk-!-margin-bottom-3">
+      <%= link_to candidate_name, provider_interface_application_choice_path(application_choice), class: 'govuk-link' %>
+    </h3>
+    <p id="course-provider-name-<%= index %>" class="govuk-body govuk-!-font-size-16"><%= course_provider_name %></p>
+    <p id="course-name-and-code-<%= index %>" class="govuk-body govuk-!-margin-bottom-0"><%= course_name_and_code %></p>
+    <p id="accredited-body-<%= index %>" class="govuk-body govuk-caption-m govuk-!-font-size-16">Accredited body: <%= accrediting_provider.name if accrediting_provider %></p>
```
tvararu: The markup for this component could be improved by not using `<p>` for the nested pieces of data. There are a few downsides currently:

- Without CSS, the hierarchy of the data could be unclear
- Screen reader users could have trouble figuring out what the values in the different paragraphs are meant to map to; apart from the `Accredited body:` which would be clear

Would you consider trying out 	[\<dl\> for this](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl)? Could look like:

```html
<h3>Kelly Jane Spears</h3>

<dl>
  <dt class="govuk-visually-hidden">Provider:</dt>
  <dd>Sommerset SCITT Consortium</dd>

  <dt class="govuk-visually-hidden">Course:</dt>
  <dd>Primary (X100)</dd>

  <dt>Accredited body:</dt>
  <dd>Sommerset SCITT Consortium</dd>

  <dt>Updated on</dt>
  <dd>10 Oct 2019</dd>
</dl>
```

Might require to re-jigger the CSS to lay it out as per the design.

@adamsilver WDYT of this suggestion?

app/frontend/styles/_application-card.scss
```diff
@@ -0,0 +1,70 @@
+.app-application-card {
+  border-bottom: 1px solid govuk-colour('mid-grey');
+  margin-bottom: 15px;
+  padding-bottom: 15px;
```
tvararu: Could these be `govuk-spacing` instead? (+ the rest of the file)

app/frontend/styles/_application-card.scss
```diff
@@ -0,0 +1,70 @@
...
+  // position: relative;
+  @include govuk-media-query($from: tablet) {
+    // overflow: hidden;
```
tvararu:
```suggestion
  @include govuk-media-query($from: tablet) {
```

app/components/provider_interface/application_card_component.html.erb
tvararu: The `Accredited body:` row is grey in the design.

spec/system/provider_interface/provider_applications_search_spec.rb
```diff
@@ -143,29 +141,29 @@ def when_i_search_for_candidate_name_with_odd_casing
...
-    expect(page).not_to have_css('.govuk-table__body', text: 'Duncan Idaho')
-    expect(page).not_to have_css('.govuk-table__body', text: 'Luke Smith')
+    expect(page).to have_css('.app-application-cards', text: 'Jim James')
+    expect(page).not_to have_css('.app-application-cards', text: 'Adam Jones')
+    expect(page).not_to have_css('.app-application-cards', text: 'Tom Jones')
+    expect(page).not_to have_css('.app-application-cards', text: 'Bill Bones')
+    expect(page).not_to have_css('.app-application-cards', text: 'Greg Taft')
+    expect(page).not_to have_css('.app-application-cards', text: 'Paul Atreides')
+    expect(page).not_to have_css('.app-application-cards', text: 'Duncan Idaho')
+    expect(page).not_to have_css('.app-application-cards', text: 'Luke Smith')
```
tvararu: Is there a way to scope `page` to a certain section so we don't need to repeat this CSS class?

Pseudocode of what I'm thinking would be better, I have no idea if this is possible:

```ruby
card = page.css('.app-application-cards')
expect(card).to have_text('Jim James')
```

tvararu: On reflection, do we need to test for `not_to have_css` of all these things? The affirmative `to have_css` makes sense.

app/views/provider_interface/application_choices/index.html.erb
```diff
@@ -30,12 +30,16 @@
     <% end %>
 
     <div class='moj-filter-layout__content'>
-      <%= render ProviderInterface::ToggleFilterComponent.new(filter_visible: @page_state.filter_visible,
-                                                              params_for_current_state: @page_state.to_h) %>
       <div class='moj-scrollable-pane'>
         <div class='moj-scrollable-pane__wrapper'>
         <% if @application_choices.any? %>
-          <%= render 'application_choices_table' %>
+          <div>
```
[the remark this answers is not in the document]
willmcb: It was in the prototype so I copied it over - I don't think it is needed though so will cut it.

app/components/provider_interface/application_card_component.html.erb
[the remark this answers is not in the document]
willmcb: True - think that all courses will have accredited bodies right? This isn't the case in the Dev test data generation though so I added this in so that the page would load (otherwise it calls a method on `nil`) Can amend to not output unless it is present though as this is cleaner.

app/components/provider_interface/application_card_component.html.erb
willmcb: Will check and delete if not.

app/components/provider_interface/application_card_component.html.erb
willmcb: Okay cool, will refactor.

app/components/provider_interface/application_card_component.html.erb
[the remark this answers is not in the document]
duncanjbrown: Not all courses have accredited bodies — it's only present where it's different from  `.provider`

willmcb: 👍 - copied from the design so unless @adamsilver has a reason for using `<p>` tags I'll refactor.

app/frontend/styles/_application-card.scss
willmcb: 👍

app/components/provider_interface/application_card_component.html.erb
willmcb: Good spot - will amend

spec/system/provider_interface/provider_applications_search_spec.rb
willmcb: Will look into this.

willmcb: Well I guess its actually testing that the names aren't present rather than the classes - will try to think of a neater way to do this.

app/components/provider_interface/application_card_component.html.erb
willmcb: 👍

app/components/provider_interface/application_card_component.html.erb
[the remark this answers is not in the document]
willmcb: 👍

app/views/provider_interface/application_choices/index.html.erb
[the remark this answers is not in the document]
willmcb: 👍

spec/components/provider_interface/application_card_component_spec.rb
```diff
@@ -0,0 +1,37 @@
+require 'rails_helper'
+
+RSpec.describe ProviderInterface::ApplicationCardComponent do
+  include CourseOptionHelpers
+
+  describe 'rendering' do
+    it 'renders what it should' do
```
chubberlisk: As a first time reader, I don't have the context of "what it should" render. We could be a bit clearer on what this component does in spec description. Below is not the best but a bit better. Also no need for the `describe` block IMO.

```suggestion
    it 'renders the details of the application' do
```

My preference would be to have spec for each of the attributes it displays e.g. `renders the name of the candidate`, `renders the name of the provider`, etc. But I think many would say that's overkill for something like this, which I can agree with. 😅

spec/system/provider_interface/provider_applications_search_spec.rb
tvararu: Yeah, I think if we test for the presence of the name we expect, that's good enough for me... I don't think we need the other assertions.

app/components/provider_interface/application_card_component.html.erb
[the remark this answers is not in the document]
willmcb: 👍

app/components/provider_interface/application_card_component.html.erb
willmcb: 👍

spec/components/provider_interface/application_card_component_spec.rb
willmcb: I think clarity is king - will incorporate these suggestions! Thanks :)

spec/system/provider_interface/provider_applications_search_spec.rb
willmcb: 👍

spec/components/provider_interface/application_card_component_spec.rb
willmcb: 👍

app/components/provider_interface/application_card_component.html.erb
willmcb: 👍

spec/system/provider_interface/provider_applications_search_spec.rb
willmcb: 👍

app/frontend/styles/_application-card.scss
willmcb: 👍

app/components/provider_interface/application_card_component.html.erb
```diff
@@ -0,0 +1,35 @@
+<div class="app-application-card">
+
+  <h3 class="govuk-heading-m">
+    <%= link_to candidate_name, provider_interface_application_choice_path(application_choice), class: 'govuk-link' %>
```
chubberlisk: We have a `ViewHelper` for adding the `govuk-link` class.

```suggestion
    <%= govuk_link_to candidate_name, provider_interface_application_choice_path(application_choice) %>
```

spec/components/provider_interface/application_card_component_spec.rb
```diff
@@ -0,0 +1,92 @@
+require 'rails_helper'
+
+RSpec.describe ProviderInterface::ApplicationCardComponent do
```
chubberlisk: Much clearer to me what this component displays! 💯

app/components/provider_interface/application_card_component.html.erb
willmcb: Thanks :) will change this.

willmcb: 👍

duncanjbrown:
This looks very good to me! Leaving approval to @tvararu and @chubberlisk :+1:.

One nit: should we tighten up the breakpoint which hides the filter? The colon-separated "Accredited body" and the Updated-at date broken over two lines don't feel quite right.

Screenshot 2020-03-31 at 09 29 47

willmcb:
@duncanjbrown have added a new breakpoint that should make the hiding cleaner - let me know what you think 👍

duncanjbrown:
> @duncanjbrown have added a new breakpoint that should make the hiding cleaner - let me know what you think 👍

👌

app/frontend/styles/application.scss
```diff
@@ -13,6 +13,7 @@ $govuk-image-url-function: frontend-image-url;
 $govuk-breakpoints: (
   mobile:  320px,
   tablet:  641px,
+  filter_hide: 945px,
```
tvararu: We should change this to a name that is not as tied to the one usecase here... I'm trying to think of something.

tvararu:
Everything else LGTM, just thinking about that breakpoint :+1:

app/frontend/styles/application.scss
tvararu:
```suggestion
  large: 945px,
```

app/frontend/styles/_filter.scss
```diff
@@ -29,3 +29,9 @@
 .moj-filter__options {
   box-shadow: none;
 }
+
+.app-filter {
+  @include govuk-media-query($until: filter_hide) {
+    display: none;
+  }
+}
```
tvararu: Here is how to rewrite this "mobile-first", which is functionally identical, but easier to add to in the future when designing for phones:

```suggestion
.app-filter {
  display: none;

  @include govuk-media-query($from: large) {
    display: block;
  }
}
```

tvararu:
I think we need a darker grey:

Screenshot 2020-03-31 at 11 29 10

app/frontend/styles/_application-card.scss
```diff
@@ -0,0 +1,54 @@
...
+
+    display: inline;
+  }
+
+  &__secondary {
+    width: 100%;
+
+    dd, dt {
+      font-size: govuk-font(16);
+      color: govuk-colour('mid-grey');
```
tvararu:
```suggestion
      color: govuk-colour('dark-grey');
```

app/frontend/styles/application.scss
willmcb: Yeah - I had the same thought but couldn't think of one 🤔 Will change to this.

app/frontend/styles/_filter.scss
willmcb: 👍

app/frontend/styles/_application-card.scss
willmcb: 👍

PR #1764, opened 2020-03-26
Add register or sign in page

malcolmbaig:
## Context

In order to simplify the sign up and sign in process, we want to add a
page that gives the candidate the option to choose one of these when
clicking through from the start page.

- Create an account prompt page and add its corresponding actions to the
  StartPageController.
- Link to this page from the root start page.
- For now, redirect to the sign in page when the user wants to sign in.
  The designs call for authentication to happen from this page, but that
  can be introduced in a later iteration.
- Modify the Find -> Apply route to go through this page.

**Note:** signing in from Find doesn't currently preserve the course/provider params. That's currently WIP.

## Guidance to review

- Per commit is advised.
- Easy to check out via the review app.

## Link to Trello card

https://trello.com/c/HlCpxZ5D

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/start_page_controller.rb
+++ b/app/controllers/candidate_interface/start_page_controller.rb
@@ -6,6 +6,21 @@ module CandidateInterface
 
     def show; end
 
+    def account_prompt
+      @account_prompt_form = AccountPromptForm.new
+    end
+
+    def account_prompt_handler
+      @account_prompt_form = AccountPromptForm.new(account_prompt_params)
+      render :account_prompt and return unless @account_prompt_form.valid?
+
+      if @account_prompt_form.existing_account?
+        redirect_to candidate_interface_sign_in_path
+      else
+        redirect_to candidate_interface_eligibility_path
+      end
+    end
+
     def eligibility
       @eligibility_form = EligibilityForm.new
     end
@@ -28,6 +43,10 @@ module CandidateInterface
     end
 
   private
+
+    def account_prompt_params
+      params.require(:candidate_interface_account_prompt_form).permit(:existing_account, :email)
+    end
 
     def redirect_to_interstitial_path_if_signed_in
       if current_candidate.present?
--- a/app/controllers/candidate_interface/start_page_controller.rb
+++ b/app/controllers/candidate_interface/start_page_controller.rb
@@ -17,7 +17,10 @@ module CandidateInterface
       if @account_prompt_form.existing_account?
         redirect_to candidate_interface_sign_in_path
       else
-        redirect_to candidate_interface_eligibility_path
+        redirect_to candidate_interface_eligibility_path(
+          providerCode: params[:providerCode],
+          courseCode: params[:courseCode],
+        )
       end
     end
 
--- a/app/controllers/candidate_interface/start_page_controller.rb
+++ b/app/controllers/candidate_interface/start_page_controller.rb
@@ -6,15 +6,15 @@ module CandidateInterface
 
     def show; end
 
-    def account_prompt
-      @account_prompt_form = AccountPromptForm.new
+    def create_account_or_sign_in
+      @create_account_or_sign_in_form = CreateAccountOrSignInForm.new
     end
 
-    def account_prompt_handler
-      @account_prompt_form = AccountPromptForm.new(account_prompt_params)
-      render :account_prompt and return unless @account_prompt_form.valid?
+    def create_account_or_sign_in_handler
+      @create_account_or_sign_in_form = CreateAccountOrSignInForm.new(create_account_or_sign_in_params)
+      render :create_account_or_sign_in and return unless @create_account_or_sign_in_form.valid?
 
-      if @account_prompt_form.existing_account?
+      if @create_account_or_sign_in_form.existing_account?
         redirect_to candidate_interface_sign_in_path
       else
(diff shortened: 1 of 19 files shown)


app/views/candidate_interface/start_page/account_prompt.html.erb
```diff
@@ -0,0 +1,21 @@
+<% content_for :title, t('page_titles.application') %>
```
chubberlisk: Looking at the prototype, I think we're missing a back link.

```suggestion
<% content_for :title, t('page_titles.application') %>
<% content_for :before_content, govuk_back_link_to(candidate_interface_start_path) %>
```

app/controllers/candidate_interface/start_page_controller.rb
```diff
@@ -6,6 +6,24 @@ class StartPageController < CandidateInterfaceController
 
     def show; end
 
+    def account_prompt
```
chubberlisk: I'm not sure how I feel about `account_prompt`. I have to think for a second about what this means. If this was named `create_account_or_sign_in`, `sign_up_or_in` or something along those lines, I would know exactly what page this action is for. That being said, I feel weird about having an `or` in a controller action... 🤦‍♀ 

Another thought: could this be in its own controller? We have `SignInController` and `SignUpController`, what do you think about that idea?

Tbh, I think this is a personal preference thing, feel free to ignore this comment and leave as is.

app/views/candidate_interface/start_page/show.html.erb
```diff
@@ -32,13 +32,21 @@
...
+    <% else %>
+      <%= link_to candidate_interface_eligibility_path, role: 'button', class: 'govuk-button govuk-!-margin-top-4 govuk-!-margin-bottom-4 govuk-button--start', 'data-module': 'govuk-button' do %>
+        <%= t('application_form.begin_button') %>
+        <svg class="govuk-button__start-icon" xmlns="http://www.w3.org/2000/svg" width="17.5" height="19" viewBox="0 0 33 40" role="presentation" focusable="false">
+          <path fill="currentColor" d="M0 0h13l20 20-20 20H0l20-20z" />
+        </svg>
+      <% end %>
     <% end %>
-    <p class="govuk-body">or</p>
-    <p class="govuk-body"><%= govuk_link_to 'Sign in to an existing application', candidate_interface_sign_in_path %></p>
```
chubberlisk: I'm not sure we should be removing this completely. I think this needs to be under the `create_account_or_sign_in_page` feature flag.

app/views/candidate_interface/start_page/show.html.erb
```diff
@@ -32,13 +32,21 @@
...
+          <path fill="currentColor" d="M0 0h13l20 20-20 20H0l20-20z" />
+        </svg>
+      <% end %>
+    <% else %>
+      <%= link_to candidate_interface_eligibility_path, role: 'button', class: 'govuk-button govuk-!-margin-top-4 govuk-!-margin-bottom-4 govuk-button--start', 'data-module': 'govuk-button' do %>
+        <%= t('application_form.begin_button') %>
+        <svg class="govuk-button__start-icon" xmlns="http://www.w3.org/2000/svg" width="17.5" height="19" viewBox="0 0 33 40" role="presentation" focusable="false">
+          <path fill="currentColor" d="M0 0h13l20 20-20 20H0l20-20z" />
+        </svg>
+      <% end %>
```
chubberlisk: Usually I'm okay with duplication like this but the `svg` makes it harder to spot the difference between when the feature flag is on and off. So as a first time viewer, I think it would be clearer to call out that the only difference is the link.

```erb
<% some_meaningful_var_name = FeatureFlag.active?('create_account_or_sign_in_page') ? candidate_interface_account_prompt_path : candidate_interface_eligibility_path %>

<%= link_to some_meaningful_var_name, role: 'button', class: 'govuk-button govuk-!-margin-top-4 govuk-!-margin-bottom-4 govuk-button--start', 'data-module': 'govuk-button' do %>
  <%= t('application_form.begin_button') %>
  <svg class="govuk-button__start-icon" xmlns="http://www.w3.org/2000/svg" width="17.5" height="19" viewBox="0 0 33 40" role="presentation" focusable="false">
    <path fill="currentColor" d="M0 0h13l20 20-20 20H0l20-20z" />
  </svg>
<% end %>
```

tvararu:
Also tested this on the Review app and it look 👌

app/controllers/candidate_interface/start_page_controller.rb
tvararu: I can +1 either of those names, don't think the `or` is jarring. But I may be unfamiliar with Rails conventions that might say it's wrong? 🤔

malcolmbaig: Agreed, naming can be better here. 

I think `create_account_or_sign_in` is the way to go. It's consistent with how we're referring to the feature in the Trello card and in the page title. 👍 

I like the separate controller idea - ironically, that's where I started when spiking out how this would work. I ended up moving it back into here because it seemed odd to set it apart from the start page and eligibility, given how tightly related they are in the UI. If anything though, eligibility should be in its own controller as well. I think I'll keep the actions here for now and extract into their own controller in a later refactor.

chubberlisk: 🚢

PR #1765, opened 2020-03-26
Refactor and use COVID-19 feature flag when calculating edit by date

chubberlisk:
## Context

A conversation came up about the time a candidate has to edit their application after submission during COVID-19 and not. This was because there was a question about whether the guidance around this on the "Application successfully submitted" page was correct. See Slack thread: https://ukgovernmentdfe.slack.com/archives/CN9Q1VB2M/p1585225722060500.

From this, we want:

- Normal: You have 5 working days to edit your application. We'll only send your application to your provider(s) when this editing period is over.
- COVID: You have 7 calendar days to edit your application. We'll only send your application to your provider(s) when this editing period is over.

Changing to calendar days for the `edit_by` date in `TimeLimitConfig` was done recently in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1716.

This PR does some refactoring in `TimeLimitConfig`:

- When the `covid_19` feature flag is on, we will return 7 calendar days, else return 5 working days.
- Returns a `Days` object with a helpful `to_s` method. This means we won't have to have `if... else..` statements for displaying `7 working days` or `7 calendar days` or update all the places when we do switch over.

And updates `SubmitApplicationChoice` service to take into consideration whether `edit_by` is in calendar or working days.

## Guidance to review

- Does any of this make sense?
- What do you think of the refactoring?
- Is a can of worm going to be opened?

## Link to Trello card

N/A

## Things to check

### DIFF
--- a/app/services/submit_application_choice.rb
+++ b/app/services/submit_application_choice.rb
@@ -14,7 +14,10 @@ private
     if HostingEnvironment.sandbox_mode?
       Time.zone.now
     else
-      TimeLimitConfig.edit_by.days.after(@application_choice.application_form.submitted_at).end_of_day
+      days = TimeLimitConfig.edit_by
+      date = days.type == :calendar ? days.count.days : days.count.business_days
+
+      date.after(@application_choice.application_form.submitted_at).end_of_day
     end
   end
 end
--- a/app/services/submit_application_choice.rb
+++ b/app/services/submit_application_choice.rb
@@ -14,10 +14,7 @@ private
     if HostingEnvironment.sandbox_mode?
       Time.zone.now
     else
-      days = TimeLimitConfig.edit_by
-      date = days.type == :calendar ? days.count.days : days.count.business_days
-
-      date.after(@application_choice.application_form.submitted_at).end_of_day
+      TimeLimitConfig.edit_by.to_days.after(@application_choice.application_form.submitted_at).end_of_day
     end
   end
 end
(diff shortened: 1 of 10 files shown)


app/services/submit_application_choice.rb
```diff
@@ -14,7 +14,10 @@ def edit_by_time
     if HostingEnvironment.sandbox_mode?
       Time.zone.now
     else
-      TimeLimitConfig.edit_by.days.after(@application_choice.application_form.submitted_at).end_of_day
+      days = TimeLimitConfig.edit_by
+      date = days.type == :calendar ? days.count.days : days.count.business_days
+
+      date.after(@application_choice.application_form.submitted_at).end_of_day
```
tvararu: Feels like this could be encapsulated inside the `Days` class? Finding a suitable method name though... `to_days`?

```ruby
# In app/lib/time_limit_config.rb
def to_days
  type == :calendar ? count.days : count.business_days
end
```

Then this could be:

```ruby
TimeLimitConfig.edit_by.to_days.after(@application_choice.application_form.submitted_at).end_of_day
```

chubberlisk: Agreed! 🚀

chubberlisk: Resolved in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1765/commits/bf49496e8900d59b89d9d96cf71b77b6304bdc3f and https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1765/commits/b2e80a73bdfa6f004c57d4c5500fcaaa36a58ee4.

PR #1767, opened 2020-03-26
Allow candidates to change the location of their course choice 

C-gyorfi:
## Context
We want to allow candidates to be able to change the course, full time or part-time, or location for a course. If they want to change their location, they currently need to delete the choice and add a new one, instead of editing the location directly. 

This PR is the first(and potentially simplest slice) to add change location link to the course choices review component.

This PR adds:
- Feature flag for `edit_course_choices` feature
- Add `#update` method to the `PickSiteForm` to allow updating an existing choice
- Updates the `CourseChoicesReviewComponent` to show the change link when there are multiple site options

### Before
image

### After
image

## Guidance to review

Does this approach make sense? 

## Link to Trello card
https://trello.com/c/awBTB88j/1171-change-links-are-missing-from-the-course-choices-summary-cards

## Things to check

### DIFF
--- a/app/components/candidate_interface/course_choices_review_component.rb
+++ b/app/components/candidate_interface/course_choices_review_component.rb
@@ -68,9 +68,20 @@ module CandidateInterface
     end
 
     def location_row(course_choice)
+      change_path = if FeatureFlag.active?('edit_course_choices') && has_multiple_sites?(course_choice)
+                      candidate_interface_course_choices_site_path(
+                        course_choice.provider.id,
+                        course_choice.course.id,
+                        course_choice.course.study_mode,
+                        course_choice_id: course_choice.id,
+                      )
+                    end
+
       {
         key: 'Location',
         value: "#{course_choice.offered_site.name}\n#{course_choice.offered_site.full_address}",
+        action: "location for #{course_choice.course.name_and_code}",
+        change_path: change_path,
       }
     end
 
@@ -141,6 +152,10 @@ module CandidateInterface
         key: 'Reason for offer withdrawal',
         value: course_choice.offer_withdrawal_reason,
       }
+    end
+
+    def has_multiple_sites?(course_choice)
+      CourseOption.where(course_id: course_choice.course.id).many?
     end
   end
 end
--- a/spec/components/candidate_interface/course_choices_review_component_spec.rb
+++ b/spec/components/candidate_interface/course_choices_review_component_spec.rb
@@ -58,6 +58,28 @@ RSpec.describe CandidateInterface::CourseChoicesReviewComponent do
         expect(result.css('.app-summary-card__title').text).to include(course_choice.provider.name)
       end
     end
+
+    context 'when course choice is single site and edit course choices feature is active' do
+      before { FeatureFlag.activate('edit_course_choices') }
+
+      it 'renders without the "Change" location links' do
+        result = render_inline(described_class.new(application_form: application_form))
+
+        expect(result.css('.govuk-summary-list__actions').text).not_to include('Change')
+      end
+    end
+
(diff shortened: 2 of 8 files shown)


chubberlisk:
Twas testing using the Heroku app and found a bug, steps to reproduce:

1. Add Coventry University as provider
2. Choose Computing
3. Choose full-time
4. Choose The National School of Education and Teaching Nottingham Hub, College Way, Nottingham, Nottinghamshire, NG8 4DQ
5. Get to the review courses page, then click Change for location

image

app/components/candidate_interface/course_choices_review_component.rb
```diff
@@ -142,5 +151,13 @@ def offer_withdrawal_reason_row(course_choice)
...
       }
     end
+
+    def has_multiple_sites?(course_choice)
+      !PickCourseForm.new(
+        provider_id: course_choice.provider.id,
+        course_id: course_choice.course.id,
+        application_form: course_choice.application_form,
+      ).single_site?
+    end
```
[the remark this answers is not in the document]
C-gyorfi: I agree will have a look and refactor!

C-gyorfi:
> Twas testing using the Heroku app and found a bug, steps to reproduce:

Resolved this bug with https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/e311a7c1516aeb5c3d6ca9a3ecb152191808368d/app/controllers/candidate_interface/course_choices_controller.rb#L143

The bug was caused by not passing in the study mode to `PickSiteForm` so it worked with courses that only had `full-time` study mode

app/components/candidate_interface/course_choices_review_component.rb
```diff
@@ -71,6 +71,15 @@ def location_row(course_choice)
       {
         key: 'Location',
         value: "#{course_choice.offered_site.name}\n#{course_choice.offered_site.full_address}",
+        action: "location for #{course_choice.course.name_and_code}",
+        change_path: (if has_multiple_sites?(course_choice) && FeatureFlag.active?('edit_course_choices')
```
tvararu: This is a tiny bit more efficient:

```suggestion
        change_path: (if FeatureFlag.active?('edit_course_choices') && has_multiple_sites?(course_choice)
```

tvararu: Suggest making this into a `change_path =` variable declaration at the start of the `location_row` method as it could be a bit easier on the indentation 👍

spec/components/candidate_interface/course_choices_review_component_spec.rb
```diff
@@ -58,6 +58,29 @@
...
       end
     end
+
+    context 'when course choice is single site' do
+      it 'renders without the "Change" location links' do
+        result = render_inline(described_class.new(application_form: application_form))
+
+        expect(result.css('.govuk-summary-list__actions').text).not_to include('Change')
+      end
+    end
```
tvararu: Think you might need `FeatureFlag.activate('edit_course_choices')` for this, right?

spec/components/candidate_interface/course_choices_review_component_spec.rb
```diff
@@ -58,6 +58,29 @@
...
+
+    context 'when there are multiple site options for course and edit course choices feature is active' do
+      before do
+        create(:course_option, course: application_form.application_choices.first.course)
+        FeatureFlag.activate('edit_course_choices')
```
tvararu: Do we need to clean this side-effect after this test?

C-gyorfi: No, double-checked and the teardown between tests takes care of them

PR #1773, opened 2020-03-29
Filtering applications by accredited body in Provider Interface

stevehook:
## Context

A user need was identified during research for providers to be able to filter applications by the accredited provider.

- [x] Extended `ProviderApplicationsPageState` to work with accredited providers as well as providers.
- [x] Changed `ProviderApplicationsPageState` to use the current provider user to work out the possible values for providers and accredited providers.
- [x] Added `ProviderOptionsService` to encapsulate a couple of queries that fetch the possible values for providers and accredited providers used to populate the options in the filter.
- [x] Update system spec

## Guidance to review

image

- The filtering rules applicable to provider users who are members of an accredited provider made me scratch my head a bit. As an accredited provider I'm assuming you have access to all the course that you 'accredit' (not really the right term). As a result I'm assuming you can select yourself in the new filter. It gets a more confusing for providers that can are both providers for some courses and accredited providers for others, in which case a given provider might appear in both filters as in the example above.
- I'm assuming that we are using the term 'Accredited Provider'
- To get the list of available accredited providers I used a query rather than iterating over the unfiltered list of application choices to get the possible values (as we did to get the list of possible provider values). See `ProviderOptionsService`
- For consistency I changed the way we get the list of possible provider values to a query also using the same service.
- One consequence of this is that you will now see providers and accredited providers that don't have any applications (yet). Not sure whether this is desirable or not)

## Link to Trello card

https://trello.com/c/xhVK2tff/1782-support-filtering-by-accredited-body-on-the-provider-ui-application-list

## Things to check


app/services/provider_options_service.rb
```diff
@@ -0,0 +1,31 @@
+class ProviderOptionsService
```
duncanjbrown: Should we namespace this inside `ProviderInterface`?

app/models/provider_interface/provider_applications_page_state.rb
```diff
@@ -125,22 +125,33 @@ def status_filters
...
 
-      provider_filters = {
+      {
         heading: 'provider',
-        input_config: input_config.uniq,
+        input_config: input_config,
       }
+    end
+
+    def accrediting_provider_filters_builder
```
duncanjbrown: This should be `accredited provider`: https://ukgovernmentdfe.slack.com/archives/CAHBLU6ET/p1585561134106700

app/services/provider_options_service.rb
```diff
@@ -0,0 +1,31 @@
...
+  attr_reader :provider_user
+
+  def initialize(provider_user)
+    @provider_user = provider_user
+  end
+
+  def accrediting_providers
```
duncanjbrown: Also `accredited`

app/models/provider_interface/provider_applications_page_state.rb
stevehook: just saw that Slack thread will update terminology in code. As a separate exercise it might be worth correcting `providers.accrediting_provider_id` in the DB schema but I'll leave that for now.

duncanjbrown: yes, I agree :+1:

stevehook: https://trello.com/c/3gHEY19s/1828-rename-accrediting-provider-to-accredited-provider-in-db-schema

duncanjbrown:
> The filtering rules applicable to provider users who are members of an accredited provider made me scratch my head a bit. As an accredited provider I'm assuming you have access to all the course that you 'accredit' (not really the right term). As a result I'm assuming you can select yourself in the new filter. It gets a more confusing for providers that can are both providers for some courses and accredited providers for others, in which case a given provider might appear in both filters as in the example above.

I think this is ok for now. It's good to draw out the (implicit, so far) assumption that if you are an accredited body then you ratify* your own courses.

In an extreme case I suppose you could be a SCITT which ratifies some SD courses and have some (but not all) of your own courses ratified by an HEI.

*for want of a better word (https://ukgovernmentdfe.slack.com/archives/CAHBLU6ET/p1585564547127700?thread_ts=1585561134.106700&cid=CAHBLU6ET)

duncanjbrown:
> To get the list of available accredited providers I used a query rather than iterating over the unfiltered list of application choices to get the possible values (as we did to get the list of possible provider values). See ProviderOptionsService

I think this is good. I don't think there's any expectation (yet) that we diminish the options based on what's already been filtered

duncanjbrown:
One more point in favour of "does it make sense to select yourself in the accredited bodies" filter: the new card design coming in #1762 also includes the offering provider in the "accredited body" space if there's no accredited body. eg a Gorse SCITT course has the "accredited body" field present, populated with "Gorse SCITT"

PR #1787, opened 2020-03-30
Make Slack poster channel aware

tijmenb:
## Context

We currently use a channel-specific webhook, but we now have a webhook configured that is allowed to post to all channels. This will make the app more flexible, for example we'll be able to post important messages to `#twd_apply` too.

This specifies the channel, username and emoji for the Slack posts, so that it goes to the correct one.

## Guidance to review

It should all work on QA - we'll know if it doesn't.

I've tried the code in https://ukgovernmentdfe.slack.com/archives/CN1MCQCHZ/p1585585959026100.

## Things to check

### DIFF
--- a/app/workers/slack_notification_worker.rb
+++ b/app/workers/slack_notification_worker.rb
@@ -19,15 +19,23 @@ private
   end
 
   def post_to_slack(text)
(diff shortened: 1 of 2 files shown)


tvararu:
Merging to reduce number of open PRs 👷

PR #1791, opened 2020-03-31
Rename accrediting provider to accredited provider in db schema

stevehook:
## Context

The database schema and model code (associations) use the wrong terminology for 'accredited providers'. 

- [x] Migration to copy values from `courses.accrediting_provider_id` to the new `accredited_provider_id` column added by https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/compare/1828-migration-to-add-courses-accredited-provider?expand=1
- [x] Update model associations and other references to be consistent.

## Guidance to review

- Is this migration safe? By leaving the old column in place (to be removed in a later deploy) we should avoid crashes that could result from old code running on the new database. There could be issues with updates to the old column happening after data is copied from old to new. The most likely cause for this would be the sync from Find background service but that _should_ correct itself next time it runs.

## Link to Trello card

https://trello.com/c/3gHEY19s/1828-rename-accrediting-provider-to-accredited-provider-in-db-schema

## Things to check

- [x] This code doesn't rely on migrations in the same Pull Request - See review notes above


duncanjbrown:
> Is this migration safe? By leaving the old column in place (to be removed in a later deploy) we should avoid crashes that could result from old code running on the new database. However there could be errors due to new code running before the migration completes.

Should we add the column in PR 1 and then the code change plus a backfill migration in PR 2?

stevehook:
> > Is this migration safe? By leaving the old column in place (to be removed in a later deploy) we should avoid crashes that could result from old code running on the new database. However there could be errors due to new code running before the migration completes.
> 
> Should we add the column in PR 1 and then the code change plus a backfill migration in PR 2?

Yes had the same thought as I was raising this PR so I've just raised https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1794

This will be PR2 as you say. PR3 can just drop the old column.

stevehook:
> There are a few dangling references to `accrediting`. Some belong to the Find API and we can't fix them, some will come out in the wash when we remove the column, but should we fix the var name and test descriptions?
> 
> ```
> app/components/provider_interface/application_card_component.html.erb:23:          <%= accrediting_provider.nil? ? course_provider_name : accrediting_provider.name %>
> app/components/provider_interface/application_card_component.rb:5:    attr_accessor :accrediting_provider, :application_choice, :application_choice_path,
> app/components/provider_interface/application_card_component.rb:9:      @accrediting_provider = application_choice.accredited_provider
> app/services/sync_provider_from_find.rb:58:    add_accredited_provider(course, find_course[:accrediting_provider])
> spec/components/support_interface/provider_courses_table_component_spec.rb:47:    it 'may include accrediting providers' do
> spec/support/test_helpers/find_api_helper.rb:68:                'accrediting_provider': nil,
> spec/support/test_helpers/find_api_helper.rb:174:                'accrediting_provider': {
> spec/support/test_helpers/find_api_helper.rb:293:              'accrediting_provider': nil,
> spec/services/get_application_choices_for_providers_spec.rb:105:  it 'returns application_choice that the provider is the accrediting body for' do
> spec/services/sync_provider_from_find_spec.rb:101:      it 'correctly handles accrediting providers' do
> ```

Yes I should have picked some of those up too. Have just pushed a commit to get rid of all references to `accrediting` except those for the Find API and legacy migrations.

PR #1792, opened 2020-03-31
Use .new rather than method call for Component rendering

willmcb:
Use `.new()` rather than method call for Component rendering 

It was raising warning when running tests.

## Link to Trello card
N/A

## Things to check


tvararu:
Thank you for doing something from my TODO list! 👍

PR #1795, opened 2020-03-31
Authenticate sign ins from 'create account or sign in' page

malcolmbaig:
## Context

On the 'create account or sign in' page, if a candidate selects the latter they currently get sent to the dedicated sign in page.

In order to make this slightly simpler for the candidate, change it so that we ask for an email and begin the magic link authentication process immediately without the additional step.

- Add an email field to the 'create account or sign in' page and its
  corresponding form.
- Extract existing sign in logic into a service object.
- Call this service from both the SignInController and
  StartPageController.

**Note:** I've opted to write the SignInCandidate service object such that one of its arguments is the controller context from which it's called. It's a little unconventional, but it seemed like the simplest way to ensure we have the same logic in two places. I considered simple duplication, but I think the sign in logic is complex enough that we'd want to define it in just one place.

## Updated Form
Screenshot 2020-03-31 at 12 39 12

## Updated Form With validation error
Screenshot 2020-03-31 at 12 39 33
## Existing sign in screen with validation behavior preserved
Screenshot 2020-03-31 at 12 40 00

## Guidance to review

* Easiest to test locally with a pre-existing user and both sidekiq and `tail -f log/mail.log` running for magic link emails.
* Worth checking sign in from the 'Create account or sign in' page and the dedicated sign in page (linked in header).
* Course selections from Find should also work across the new sign in mechanism.

## Link to Trello card

https://trello.com/c/HlCpxZ5D

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/start_page_controller.rb
+++ b/app/controllers/candidate_interface/start_page_controller.rb
@@ -15,10 +15,7 @@ module CandidateInterface
       render :create_account_or_sign_in and return unless @create_account_or_sign_in_form.valid?
 
       if @create_account_or_sign_in_form.existing_account?
-        redirect_to candidate_interface_sign_in_path(
-          providerCode: params[:providerCode],
-          courseCode: params[:courseCode],
-        )
+        SignInCandidate.new(@create_account_or_sign_in_form.email, self).call
       else
         redirect_to candidate_interface_eligibility_path(
           providerCode: params[:providerCode],
--- a/app/models/candidate_interface/create_account_or_sign_in_form.rb
+++ b/app/models/candidate_interface/create_account_or_sign_in_form.rb
@@ -2,9 +2,10 @@ module CandidateInterface
   class CreateAccountOrSignInForm
     include ActiveModel::Model
 
-    attr_accessor :existing_account
+    attr_accessor :existing_account, :email
 
     validates :existing_account, presence: true
+    validates :email, presence: true, email_address: true, if: -> { existing_account? }
 
     def existing_account?
       ActiveModel::Type::Boolean.new.cast(existing_account)
--- a/app/services/candidate_interface/sign_in_candidate.rb
+++ b/app/services/candidate_interface/sign_in_candidate.rb
@@ -1,0 +1,47 @@
+module CandidateInterface
+  class SignInCandidate
+    attr_reader :controller, :email
+
+    def initialize(email, controller)
+      @controller = controller
+      @email = email
+    end
+
+    delegate(
+      :params,
+      :redirect_to,
+      :render,
+      :candidate_interface_check_email_sign_in_path,
+      to: :controller,
+    )
+
+    def call
+      candidate = Candidate.for_email email
+
+      if candidate.persisted?
+        update_course_from_find(candidate)
+        MagicLinkSignIn.call(candidate: candidate)
+        controller.add_identity_to_log(candidate.id)
+        redirect_to candidate_interface_check_email_sign_in_path
+      elsif candidate.valid?
+        AuthenticationMailer.sign_in_without_account_email(to: candidate.email_address).deliver_now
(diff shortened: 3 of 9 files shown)


app/services/candidate_interface/sign_in_candidate.rb
```diff
@@ -20,26 +20,33 @@ def initialize(email, controller_context)
...
+  private
+
+    def update_course_from_find(candidate)
+      course_from_find = Provider
+        .find_by(code: params[:providerCode])
+        &.courses
+        &.find_by(code: params[:courseCode])
+
+      if course_from_find
+        candidate.update(course_from_find_id: course_from_find.id)
```
tijmenb:
```suggestion
        candidate.update!(course_from_find_id: course_from_find.id)
```

app/services/candidate_interface/sign_in_candidate.rb
```diff
@@ -0,0 +1,53 @@
...
+    )
+
+    def call
+      candidate = Candidate.for_email email
+
+      if candidate.persisted?
+        update_course_from_find(candidate)
+        MagicLinkSignIn.call(candidate: candidate)
+
+        controller_context.send :add_identity_to_log, candidate.id
```
tijmenb:
```suggestion
        controller_context.add_identity_to_log(candidate.id)
```

app/controllers/candidate_interface/start_page_controller.rb
```diff
@@ -15,9 +15,9 @@ def create_account_or_sign_in_handler
       render :create_account_or_sign_in and return unless @create_account_or_sign_in_form.valid?
 
       if @create_account_or_sign_in_form.existing_account?
-        redirect_to candidate_interface_sign_in_path(
-          providerCode: params[:providerCode],
-          courseCode: params[:courseCode],
+        SignInCandidate.call(
```
tijmenb: I would skip the `call` and call `new` here, I think it saves some code in the service object and makes it easier to refactor (you only have to rename a param in 2 places).

app/models/candidate_interface/create_account_or_sign_in_form.rb
```diff
@@ -2,9 +2,12 @@ module CandidateInterface
   class CreateAccountOrSignInForm
     include ActiveModel::Model
 
-    attr_accessor :existing_account
+    attr_accessor :existing_account, :email
 
     validates :existing_account, presence: true
+    with_options if: -> { existing_account? } do
+      validates :email, presence: true, email_address: true
+    end
```
tijmenb: Is this the same as this?

```suggestion
  validates :email, presence: true, email_address: true, if: -> { existing_account? }
```

app/services/candidate_interface/sign_in_candidate.rb
```diff
@@ -0,0 +1,53 @@
+module CandidateInterface
+  class SignInCandidate
+    def self.call(email:, controller_context:)
+      new(email, controller_context).call
+    end
+
+    attr_reader :controller_context, :email
```
tijmenb: Perhaps just `controller` is descriptive enough here.

app/services/candidate_interface/sign_in_candidate.rb
malcolmbaig: Sadly `add_identity_to_log` is a [private](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/app/controllers/candidate_interface/candidate_interface_controller.rb#L34) method of the `CandidateInterfaceController`. What do you think about moving it above the private keyword or removing the keyword altogether, as I don't think it serves much of a purpose?

malcolmbaig: Have moved it out of private, can revisit if opinions on this differ.

tijmenb: 👍

PR #1802, opened 2020-03-31
App service container start timeout pipeline variable

tpinney:
## Context

Due to exceptional demand on Azure resources we have seen app service staging slot startups timeout and therefore wish to increase this timeout.

The slot timeout is currently 600s and is set to a default value in the `azure-pipelines-deploy-template.yml` file. This change introduces a new pipeline variable named `appServiceContainerTimeoutSeconds` that is set in the shared variable group to apply to all environments. This parameter can be overridden on a per environment basis if required.

## Guidance to review

Verify in DevOps environment that the app service container environment variable `WEBSITES_CONTAINER_START_TIME_LIMIT` is now set to 1200 instead of the default of 600 as seen in all the other environment presently.
https://portal.azure.com/#@platform.education.gov.uk/resource/subscriptions/289869cc-7183-46bd-8131-f673c5eb94ba/resourceGroups/s106d02-apply/providers/Microsoft.Web/sites/s106d02-apply-as/configuration

## Link to Trello card

N/A

## Things to check

### DIFF
--- a/azure-pipelines-release.yml
+++ b/azure-pipelines-release.yml
@@ -68,6 +68,7 @@ stages:
       databaseStorageAutoGrow: 'disabled'
       databaseBackupRetentionDays: 7
       dockerhubUsername: '$(dockerHubUsername)'
+      containerStartTimeLimit: '$(appServiceContainerTimeoutSeconds)'
       railsSecretKeyBase: '$(railsSecretKeyBase)'
       railsEnv: 'production'
       basicAuthEnabled: '$(basicAuthEnabled)'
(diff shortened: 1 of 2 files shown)


duncanjbrown:
Do we need to add the variable in the other envs too? What happens if it's not set?

tpinney:
> Do we need to add the variable in the other envs too? What happens if it's not set?

Its set in the `APPLY - Shared variables` group, so will apply to all environments unless specifically overridden in their specific environment variable groups. Failing all else it will revert to the template default value of 600 seconds that it was using previously.

PR #1806, opened 2020-04-01
Add first page of the candidate satisfaction survey

davidgisbey:
## Context

Candidates need a way to give us feedback on our service. To do this we're going to add a satisfaction survey. This PR adds the initial page of this flow.

- Adds a feature flag 
- Adds a Give feedback button to the submit success page
- Adds the first page which asks if they would recommend the service

Before

image

After

image

image

## Guidance to review

https://bat-design-history.netlify.com/apply-for-teacher-training/satisfaction-survey

I've just taken it far enough to get the tests to pass on the complexity view so I can keep this PR fairly self-contained.

Happy with the approach?

## Link to Trello card

https://trello.com/c/tO5B8iK1/1257-dev-survey-to-candidates-when-they-have-submitted-an-application

## Things to check

### DIFF
--- a/app/models/candidate_interface/satisfaction_survey_form.rb
+++ b/app/models/candidate_interface/satisfaction_survey_form.rb
@@ -2,24 +2,24 @@ module CandidateInterface
   class SatisfactionSurveyForm
     include ActiveModel::Model
 
-    attr_accessor :question, :response
+    attr_accessor :question, :answer
 
     validates :question, presence: true
 
     def save(application_form)
       return false unless valid?
 
-      remove_strongly_agree_and_strongly_disagree_text(@response)
-      application_form.update(satisfaction_survey: { @question => @response })
+      remove_strongly_agree_and_strongly_disagree_text(@answer)
+      application_form.update(satisfaction_survey: { @question => @answer })
     end
 
   private
 
-    def remove_strongly_agree_and_strongly_disagree_text(response)
-      if response == '1 - strongly agree'
-        @response = '1'
-      elsif response == '5 - strongly disagree'
-        @response = '5'
+    def remove_strongly_agree_and_strongly_disagree_text(answer)
+      if answer == '1 - strongly agree'
+        @answer = '1'
+      elsif answer == '5 - strongly disagree'
+        @answer = '5'
       end
     end
   end
--- a/app/models/candidate_interface/satisfaction_survey_form.rb
+++ b/app/models/candidate_interface/satisfaction_survey_form.rb
@@ -1,0 +1,22 @@
+module CandidateInterface
+  class SatisfactionSurveyForm
+    include ActiveModel::Model
+
+    attr_accessor :question, :response
+
+    def save(application_form)
+      remove_strongly_agree_and_strongly_disagree_text(@response)
+      application_form.update(satisfaction_survey: { @question => @response })
+    end
+
+  private
+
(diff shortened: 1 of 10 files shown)


app/models/candidate_interface/satisfaction_survey_form.rb
```diff
@@ -0,0 +1,26 @@
...
+    include ActiveModel::Model
+
+    attr_accessor :question, :response
+
+    validates :question, presence: true
+
+    def save(application_form)
+      return false unless valid?
+
+      remove_strongly_agree_and_strongly_disagree_text(@response)
```
tvararu, line 12: Hmm - should we actually remove this? 🤔

We could also save it with, and remove it later upon presenting if we need to.

Does this complicate potentially amending your answers?

app/models/candidate_interface/satisfaction_survey_form.rb
```diff
@@ -0,0 +1,26 @@
+module CandidateInterface
+  class SatisfactionSurveyForm
+    include ActiveModel::Model
+
+    attr_accessor :question, :response
+
+    validates :question, presence: true
```
tvararu, line 7: Should we validate for the `inclusion` of the question in the set of answers we expect?

app/models/candidate_interface/satisfaction_survey_form.rb
```diff
@@ -0,0 +1,26 @@
+module CandidateInterface
+  class SatisfactionSurveyForm
+    include ActiveModel::Model
+
+    attr_accessor :question, :response
```
tvararu, line 5: Minor suggestion: maybe `answer` is a better name than `response`, which has a lot of overloaded meaning in full-stack applications.

app/models/candidate_interface/satisfaction_survey_form.rb
tvararu: Actually - why do we need to do this? How come we don't get back `'1'` since that's what we pass in as the `value=` of the radio button?

config/routes.rb
```diff
@@ -301,6 +301,13 @@
         post '/' => 'safeguarding#update'
         get '/review' => 'safeguarding#show', as: :review_safeguarding
       end
+
+      scope '/satisfaction-survey' do
+        get '/recommendation' => 'satisfaction_survey#recommendation', as: :satisfaction_survey_recommendation
+        post 'recommendation' => 'satisfaction_survey#submit_recommendation', as: :satisfaction_survey_submit_recommendation
```
tvararu, line 307:
```suggestion
        post '/recommendation' => 'satisfaction_survey#submit_recommendation', as: :satisfaction_survey_submit_recommendation
```

app/models/candidate_interface/satisfaction_survey_form.rb
davidgisbey: That's in a follow up PR 👍

davidgisbey: image

app/models/candidate_interface/satisfaction_survey_form.rb
davidgisbey: Good shout!

app/models/candidate_interface/satisfaction_survey_form.rb
davidgisbey: yea that's a 🤦‍♂ moment haha

PR #1807, opened 2020-04-01
Rename ProviderUsersProvider to ProviderPermissions

duncanjbrown:
## Context

`new`ing up a `ProviderUsersProvider` to work with provider permissions is a bit clunky, so rename the model to `ProviderPermissions`

Rename the model. As the model is not yet explicitly referenced anywhere, renaming is a question of setting `self.table_name` and updating associations.

## Guidance to review

- Does this make sense?
- Is the `ProviderPermissions` the right name?
- Do we need to rename the table too?

## Link to Trello card

Follow-up on https://trello.com/c/mPGFKdgO/1840-create-permissions-model-for-providerusers

## Things to check

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -5,8 +5,8 @@ class Provider < ApplicationRecord
   has_many :application_choices, through: :course_options
   has_many :accredited_courses, class_name: 'Course', inverse_of: :accrediting_provider, foreign_key: :accrediting_provider_id
 
-  has_many :provider_users_providers, dependent: :destroy
-  has_many :provider_users, through: :provider_users_providers
+  has_many :provider_permissions, dependent: :destroy
+  has_many :provider_users, through: :provider_permissions
   has_many :provider_agreements
 
   enum region_code: {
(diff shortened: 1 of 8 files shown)


steventux:
I like this change and think the naming makes sense. I was reading [this](http://www.jefferydurand.com/ruby/rails/domain/modeling/user/permissions/2016/06/15/user-permissions-through-join-table.html) yesterday which uses a similar naming convention.
@willmcb and I worked on a method to check the `manage_users` permission [here](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/compare/1840-create-permissions-method-for-providerusers) with the idea that we obfuscate the need to call the association directly outside of _whereever these permissions methods should live_.

duncanjbrown:
> @willmcb and I worked on a method to check the `manage_users` permission [here](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/compare/1840-create-permissions-method-for-providerusers) with the idea that we obfuscate the need to call the association directly outside of _whereever these permissions methods should live_.

that sounds 👌

PR #1808, opened 2020-04-01
Update change offer flow user experience

mnacos:
### Context

The current 'change offer' flow works well but additional UI/UX tweaks have been planned to improve the user experience. This PR focuses on modifying existing offers: in a nutshell, it hides 'Change' links when no options exist for the current provider user and makes some text conditional based on the initial trigger of the 'change offer' flow.

The guiding principle behind these changes is "hide links not relevant to this user or available choices". For example, if a provider user is only associated with one provider, they shouldn't see the 'Change provider' link when editing an offer. Likewise, if a provider only offers one course or one location, there is no point prompting the user to change course or location for an existing offer. Finally, if a user is trying to change the course of an offer, there is no point showing them the change provider option at the end of the flow.

1. This work introduces the notion of an entry point to the 'change offer' flow, capturing it and preserving it via a query param across the different parts of the multi-step process. This param, `entry`, is the basis for conditional text within the different steps.
2. The relevant component `provider_interface/status_box_components/offer_component.rb` is also modified to hide 'Change' links if no other valid options exist for the current provider user (e.g. if the user is only associated with one provider, no 'change provider' link is shown. Similar checks are performed for the the availability of courses and locations.
3. Headings at the various stages are now conditional on the `entry` point of the flow. 
4. The confirmation page at the end of the flow now contains conditional 'Change' links, based on what the entry point to the flow was. Support for these links has been added to `app/components/provider_interface/offer_summary_list_component.rb`

### Guidance to review

Refactorings:

- `components/provider_interface/offer_summary_list_component.rb` now accepts `options` so that it can be used to display Change links on the change offer confirmation page.
- `components/provider_interface/status_box_component.rb` also accepts `options` now, for the sub-components that need them. For offers, specifically, these are used to hide any Change links that are meaningless because no alternative options exist. The logic in this component is different to the logic in `OfferSummaryListComponent` because the status box component is the beginning of the process, while `OfferSummaryListComponent` with change links is used at the end.
- A new service `GetAllChangeOptionsFromOfferedOption` is introduced to obtain and feed the relevant data to the components above.

The views within `app/views/provider_interface/offer_changes/` are also used by another flow, that of changing the course details while making the initial offer. This should be unaffected by the current PR.

### Link to Trello card

[1789 - Update "offer" flows with new designs for consistency](https://trello.com/c/E144qHwK)

### Things to check

### DIFF
--- a/app/components/provider_interface/offer_summary_list_component.rb
+++ b/app/components/provider_interface/offer_summary_list_component.rb
@@ -18,17 +18,17 @@ module ProviderInterface
         },
         {
           key: 'Provider',
-          value: application_choice.offered_course.provider.name,
+          value: course_option.course.provider.name,
           change_path: change_path(:provider), action: 'training provider'
         },
         {
           key: 'Course',
-          value: application_choice.offered_course.name_and_code,
+          value: course_option.course.name_and_code,
           change_path: change_path(:course), action: 'course'
         },
         {
           key: 'Location',
-          value: application_choice.offered_site.name_and_address,
+          value: course_option.site.name_and_address,
           change_path: change_path(:course_option), action: 'location'
         },
       ]
@@ -38,14 +38,17 @@ module ProviderInterface
       Rails.application.routes.url_helpers
     end
 
+    # TODO: paths produced preserve the current (unsaved) provider/course/option selection
     def change_path(target)
-      case target
-      when :provider
-        paths.provider_interface_application_choice_change_offer_edit_provider_path(application_choice.id) if show_provider_link?
-      when :course
-        paths.provider_interface_application_choice_change_offer_edit_course_path(application_choice.id) if show_course_link?
-      when :course_option
-        paths.provider_interface_application_choice_change_offer_edit_course_option_path(application_choice.id) if show_course_option_link?
+      if new_course_option
+        case target
+        when :provider
+          paths.provider_interface_application_choice_change_offer_edit_provider_path(application_choice.id) if show_provider_link?
+        when :course
+          paths.provider_interface_application_choice_change_offer_edit_course_path(application_choice.id) if show_course_link?
+        when :course_option
+          paths.provider_interface_application_choice_change_offer_edit_course_option_path(application_choice.id) if show_course_option_link?
+        end
       end
     end
 
@@ -62,8 +65,14 @@ module ProviderInterface
     end
 
     def new_course_option
-      @course_option ||= CourseOption.find(@course_option_id) if @course_option_id
-      @course_option unless @course_option && @course_option.id != application_choice.offered_option.id
+      if @course_option_id
+        @new_course_option ||= CourseOption.find(@course_option_id)
+        @new_course_option if @course_option_id != application_choice.offered_option.id
+      end
+    end
+
+    def course_option
+      new_course_option || application_choice.offered_option
     end
   end
 end
--- a/app/components/provider_interface/offer_summary_list_component.rb
+++ b/app/components/provider_interface/offer_summary_list_component.rb
@@ -8,6 +8,9 @@ module ProviderInterface
       @header = header
       @course_option_id = extra_arguments[:course_option_id]
       @entry = extra_arguments[:entry]
+      @available_providers = extra_arguments[:available_providers]
+      @available_courses = extra_arguments[:available_courses]
+      @available_course_options = extra_arguments[:available_course_options]
     end
 
     def rows
@@ -53,15 +56,15 @@ module ProviderInterface
     end
 
     def show_provider_link?
-      @entry == 'provider'
+      @entry == 'provider' && @available_providers && @available_providers.count > 1
     end
 
     def show_course_link?
-      @entry != 'course_option'
+      @entry != 'course_option' && @available_courses && @available_courses.count > 1
     end
 
     def show_course_option_link?
-      true
+      @available_course_options && @available_course_options.count > 1
     end
 
     def new_course_option
--- a/app/components/provider_interface/offer_summary_list_component.rb
+++ b/app/components/provider_interface/offer_summary_list_component.rb
@@ -5,9 +5,11 @@ module ProviderInterface
 
     def initialize(application_choice:, header: 'Your offer', options: {})
       @application_choice = application_choice
+      @course_option = application_choice.offered_option
       @header = header
-      @course_option_id = options[:new_course_option_id]
-      @entry = options[:entry]
+      @change_provider_path = options[:change_provider_path]
+      @change_course_path = options[:change_course_path]
+      @change_course_option_path = options[:change_course_option_path]
     end
 
     def rows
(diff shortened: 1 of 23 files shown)


adamsilver:
working well, @mnacos.

The width of the check answers page should be 2/3 (one for another PR perhaps)?

mnacos:
> working well, @mnacos.
> 
> The width of the check answers page should be 2/3 (one for another PR perhaps)?

Thanks. Easy to fix, have just updated it.

duncanjbrown:
Please don't merge yet! In the middle of review 🙂

app/controllers/provider_interface/offer_changes_controller.rb
```diff
@@ -58,16 +70,37 @@ def update
 
     def render_providers
       set_alternative_providers
+      @page_title = \
```
steventux, line 73: The repeat conditionals in these `render_*` methods make me think `@page_title` generation could be wrapped up in a convenient method. I also wonder if the render methods are the right place to set this var.

app/components/provider_interface/offer_summary_list_component.rb
```diff
@@ -3,9 +3,14 @@ class OfferSummaryListComponent < ActionView::Component::Base
     include ViewHelper
     attr_reader :application_choice, :header
 
-    def initialize(application_choice:, header: 'Your offer')
+    def initialize(application_choice:, header: 'Your offer', extra_arguments: {})
       @application_choice = application_choice
       @header = header
+      @course_option_id = extra_arguments[:course_option_id]
+      @entry = extra_arguments[:entry]
+      @available_providers = extra_arguments[:available_providers]
```
steventux, line 11: I might be misunderstanding the change here but wouldn't passing the change path remove a bit of path determining logic in this component? We don't seem to use these collections other than to work out which path to render.

app/components/provider_interface/status_box_component.rb
```diff
@@ -4,8 +4,10 @@ class StatusBoxComponent < ActionView::Component::Base
 
     attr_reader :application_choice
 
-    def initialize(application_choice:)
+    def initialize(application_choice:, extra_arguments: nil)
```
steventux, line 7: I think the `StatusBoxComponent` pattern is a bit compromised now, not necessarily as part of this PR but because status boxes are conditionally rendered. Probably not a change needed in this PR but we have 2 cases of not rendering the box depending on status and this PR introduces `extra_arguments` to satisfy another piece of conditional logic.
Something to look at in the future, I think we've moved away from always rendering something based on `ApplicationChoice#status` and into some bespoke renderings.
Maybe that points towards different views or partials to satisfy this portion of the page as a one size fits all component seems to be a stretch now.

app/components/provider_interface/status_box_components/offer_component.rb
```diff
@@ -3,9 +3,13 @@ module StatusBoxComponents
     class OfferComponent < ActionView::Component::Base
       include ViewHelper
       attr_reader :application_choice
+      attr_reader :available_providers, :available_courses, :available_course_options
 
-      def initialize(application_choice:)
-        @application_choice = application_choice
+      def initialize(args)
+        @application_choice = args[:application_choice]
```
steventux, line 9: This seems to duplicate the pattern in `OfferSummaryListComponent` above of passing the collections when only the path is needed. I wonder if we can deduce paths in a common piece of logic before the component is initialized? A helper maybe?

app/components/provider_interface/offer_summary_list_component.rb
mnacos: From 83673a382bd06c13c2bc22e6c5b6494d3ad91c83 onwards we also use the relevant counts to determine if the links should be shown. But if Adam agrees we may simplify this to just check `@entry`. Application choice + new course_option_id would then be sufficient to construct the change links. But it's an interesting point, whether the component should decide visibility and url of the links or just be given them.

app/components/provider_interface/status_box_component.rb
mnacos: I agree the more conditional we make these renderings the more a different approach may be called for. Allowing the passing of optional bits of information is necessarily a bad thing, though. In fact, if we stick with the current pattern for now, Duncan suggested we go all the way and just pass a hash to the top-level initializer, then have the subcomponents complain if they don't get what they need. I think I'll just do that for now. Maybe something to discuss at tech retro?

app/components/provider_interface/status_box_components/awaiting_provider_decision_component.rb
```diff
@@ -4,8 +4,8 @@ class AwaitingProviderDecisionComponent < ActionView::Component::Base
       include ViewHelper
       attr_reader :application_choice
 
-      def initialize(application_choice:)
-        @application_choice = application_choice
+      def initialize(args)
```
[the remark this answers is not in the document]
mnacos: After my chat with Duncan I think we may go the other way and just use a hash everywhere, which will include an `:application_choice` key.

app/components/provider_interface/status_box_components/offer_component.rb
mnacos: Whether to show each link and what the link is, both of which can be controlled by the presence or absence of a link that's passed to the component (and these links would need to be passed through the top-level `StatusBoxComponent` the way things currently are). But then the controller needs to make that decision, which defeats the purpose of the current code structure around status boxes (them being self-contained, the controller doesn't need to worry about them too much).

With the 'change offer' stuff, I think we're pushing the limits of what can be done with standard MVC/Rails, in terms of multi-step wizard flows and steps referring to previous steps etc. The current approach (and action view components, for that matter) is probably influenced by Javascript/MVVM frameworks where a big state entity is made available to components and these just pick what they want. Happy to discuss further, but I think this may be a more general architecture/direction discussion.

app/services/get_all_change_options_from_offered_option.rb
```diff
@@ -0,0 +1,39 @@
...
+      available_courses: available_courses,
+      available_course_options: available_course_options,
+    }
+  end
+
+private
+
+  def study_mode_for_other_courses
+    current_study_mode = @application_choice.offered_option.study_mode
+    [current_study_mode, :full_time_or_part_time]
```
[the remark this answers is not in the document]
mnacos: No. Courses can have `{:full_time, :part_time, :full_time_or_part_time}`. Course options only have `{:full_time, :part_time}`.

app/components/provider_interface/offer_summary_list_component.rb
```diff
@@ -16,17 +18,68 @@ def rows
...
         },
       ]
     end
+
+    def paths
+      Rails.application.routes.url_helpers
+    end
+
+    # paths produced preserve the current (unsaved) provider/course/option selection
+    def change_path(target)
```
duncanjbrown, line 42: Were we planning to push these up into the controller?

app/components/provider_interface/status_box_components/conditions_not_met_component.rb
```diff
@@ -4,8 +4,8 @@ class ConditionsNotMetComponent < ActionView::Component::Base
       include ViewHelper
       attr_reader :application_choice
 
-      def initialize(application_choice:)
-        @application_choice = application_choice
+      def initialize(args)
```
duncanjbrown, line 7: Were we going leave this (and its like) as simply `application_choice:`?

mnacos: Yep, I've restored application_choice as an argument.

app/components/provider_interface/offer_summary_list_component.rb
mnacos: Pushed

PR #1811, opened 2020-04-01
Build the complexity page of the satisfaction survey

davidgisbey:
## Context

**Do not merge until 1806 is merged**

This is the next page in the satisfaction survey flow.

- Add the complexity page
- Update the SatisfactionSurvey to add/amend additional rows to the hash for the user's response 
- Adds a validation to check that the value of the question parameter is a question that we ask them.

image

## Guidance to review

In the SatisfactionSurvey model, I've added in logic for if a user clicks back and edits their previous response.

## Link to Trello card

https://trello.com/c/tO5B8iK1/1257-dev-survey-to-candidates-when-they-have-submitted-an-application

## Things to check

### DIFF
--- a/app/models/candidate_interface/satisfaction_survey_form.rb
+++ b/app/models/candidate_interface/satisfaction_survey_form.rb
@@ -6,11 +6,11 @@ module CandidateInterface
 
     QUESTIONS_WE_ASK = [
       I18n.t('page_titles.recommendation'),
-      I18n.t('page_titles.complexity')
+      I18n.t('page_titles.complexity'),
     ].freeze
 
     validates :question, presence: true
-    validates :question, inclusion: { in: QUESTIONS_WE_ASK, allow_blank: false }
+    validates :question, inclusion: { in: QUESTIONS_WE_ASK, allow_blank: false, message: 'Choose one of the options' }
 
     def save(application_form)
       return false unless valid?
--- a/app/models/candidate_interface/satisfaction_survey_form.rb
+++ b/app/models/candidate_interface/satisfaction_survey_form.rb
@@ -9,7 +9,24 @@ module CandidateInterface
     def save(application_form)
       return false unless valid?
 
-      application_form.update(satisfaction_survey: { @question => @answer })
+      if question_already_answered?(application_form)
+        application_form.satisfaction_survey[@question] = @answer
+        application_form.save
+      elsif application_form.satisfaction_survey.present?
+        application_form.update!(satisfaction_survey: merge_satisfaction_survey_and_answer(application_form))
+      else
+        application_form.update!(satisfaction_survey: { @question => @answer })
+      end
+    end
+
+  private
+
+    def question_already_answered?(application_form)
+      application_form.satisfaction_survey&.keys&.include?(@question)
+    end
+
+    def merge_satisfaction_survey_and_answer(application_form)
+      application_form.satisfaction_survey.merge!({ @question => @answer })
     end
   end
 end
--- a/app/models/candidate_interface/satisfaction_survey_form.rb
+++ b/app/models/candidate_interface/satisfaction_survey_form.rb
@@ -4,7 +4,13 @@ module CandidateInterface
 
     attr_accessor :question, :answer
 
+    QUESTIONS_WE_ASK = [
+      'I would recommend this service to a friend or colleague',
+      'I found this service unnecessarily complex',
+    ].freeze
+
     validates :question, presence: true
+    validates :question, inclusion: { in: QUESTIONS_WE_ASK, allow_blank: false }
 
     def save(application_form)
       return false unless valid?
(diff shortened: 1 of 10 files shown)


app/models/candidate_interface/satisfaction_survey_form.rb
```diff
@@ -4,12 +4,35 @@ class SatisfactionSurveyForm
 
     attr_accessor :question, :answer
 
+    QUESTIONS_WE_ASK = [
+      'I would recommend this service to a friend or colleague',
+      'I found this service unnecessarily complex',
+    ].freeze
+
     validates :question, presence: true
+    validates :question, inclusion: { in: QUESTIONS_WE_ASK, allow_blank: false }
```
tvararu, line 13: I think we need to validate that the `answer` is included in what we expect as well.

app/models/candidate_interface/satisfaction_survey_form.rb
```diff
@@ -4,12 +4,35 @@ class SatisfactionSurveyForm
...
 
-      application_form.update(satisfaction_survey: { @question => @answer })
+      if question_already_answered?(application_form)
+        application_form.satisfaction_survey[@question] = @answer
+        application_form.save
+      elsif application_form.satisfaction_survey.present?
+        application_form.update!(satisfaction_survey: merge_satisfaction_survey_and_answer(application_form))
+      else
+        application_form.update!(satisfaction_survey: { @question => @answer })
+      end
```
tvararu, line 25: Would this work for all the cases?

```suggestion
      application_form.satisfaction_survey ||= {}
      application_form.satisfaction_survey[@question] = @answer
      application_form.save
```

app/models/candidate_interface/satisfaction_survey_form.rb
```diff
@@ -4,12 +4,35 @@ class SatisfactionSurveyForm
 
     attr_accessor :question, :answer
 
+    QUESTIONS_WE_ASK = [
+      'I would recommend this service to a friend or colleague',
+      'I found this service unnecessarily complex',
```
tvararu, line 9: Should we pick these up from the titles?

app/models/candidate_interface/satisfaction_survey_form.rb
tvararu: Also, what do these errors show up as to the user? Do we need some localisation strings?

app/models/candidate_interface/satisfaction_survey_form.rb
davidgisbey: 💯

app/models/candidate_interface/satisfaction_survey_form.rb
davidgisbey: Oh, nice! Very elegant.

app/models/candidate_interface/satisfaction_survey_form.rb
davidgisbey: There are free text answers too, but we could definitely validate certain questions have certain answers @tvararu

davidgisbey: Yea good point regarding the errors. They currently don't show up and it just renders the view. The only way that page can be triggered is if they manually change the hidden field for the question. I'm not sure what the error message should be? 

Perhaps 'Stop it please' haha. But seriously, maybe 'Question was not recognised. Please input your answer again.'?

app/models/candidate_interface/satisfaction_survey_form.rb
```diff
@@ -4,12 +4,35 @@ class SatisfactionSurveyForm
...
+    ].freeze
+
     validates :question, presence: true
+    validates :question, inclusion: { in: QUESTIONS_WE_ASK, allow_blank: false }
```
tvararu, line 14: I think we also need `validates :answer, presence: true` and an appropriate message, like “Choose one of the options”

app/views/candidate_interface/satisfaction_survey/complexity.html.erb
```diff
@@ -1 +1,18 @@
...
+<%= form_with model: @survey, url: candidate_interface_satisfaction_survey_submit_complexity_path, method: :post do |f| %>
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+
+      <h2 class="govuk-heading-xl">
+        <%= t('page_titles.complexity') %>
+      </h2>
+
+      <%= f.hidden_field :question, value: t('page_titles.complexity') %>
```
tvararu, line 13: Instead of passing this using a hidden field, we could infer it from the route; when we're on the `post /complexity` route, we know which question they're trying to answer.

davidgisbey: That's a really good shout 👍

app/models/candidate_interface/satisfaction_survey_form.rb
davidgisbey: As discussed. Candidates should not be forced to give answers to all the questions.

app/models/candidate_interface/satisfaction_survey_form.rb
davidgisbey: This is longer necessary as we don;t allow users to mess with the question anymore since we use the params to get the question rather than the hidden field.

PR #1816, opened 2020-04-02
Simplify input arguments to the filter components

tijmenb:
## Context

I'm looking to reuse the provider applications filter component in the support interface. Doing some tidy up to get familiar with the code.

We don't need the `path` argument because we can construct all the URLs we want without it. This cleans up the code and makes it easier to reuse the code in other contexts like the support interface.

## Guidance to review

Nothing in particular.

## Things to check

### DIFF
--- a/app/components/provider_interface/current_filters_component.html.erb
+++ b/app/components/provider_interface/current_filters_component.html.erb
@@ -1,17 +1,15 @@
 <div class="moj-filter__selected">
 
   <div class="moj-filter__selected-heading">
-
     <div class="moj-filter__heading-title">
       <h2 class="govuk-heading-m">Selected filters</h2>
     </div>
 
     <div class="moj-filter__heading-action">
     <p class="govuk-body">
-       <%= link_to "Clear", filtering_page_path(params_for_current_state.except(:filter_selections)), class: "govuk-link govuk-link--no-visited-state" %>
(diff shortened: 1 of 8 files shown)


app/components/provider_interface/current_filters_component.html.erb
```diff
@@ -1,17 +1,15 @@
...
   <div class="moj-filter__selected-heading">
-
     <div class="moj-filter__heading-title">
       <h2 class="govuk-heading-m">Selected filters</h2>
     </div>
 
     <div class="moj-filter__heading-action">
     <p class="govuk-body">
-       <%= link_to "Clear", filtering_page_path(params_for_current_state.except(:filter_selections)), class: "govuk-link govuk-link--no-visited-state" %>
+       <%= link_to "Clear", '?' + params_for_current_state.except(:filter_selections).to_query, class: "govuk-link govuk-link--no-visited-state" %>
```
tijmenb, line 10: We could use `url_for` here. This works, but not in the unit tests for components because the component doesn't know the current URL. In reality the view just knows where they are and `url_for(foo: 'bar')` automatically translates into `/current/url?foo=bar`.

duncanjbrown: 🎉 Very nice. Broken component unit tests caused us to introduce `path:` in the first place.

willmcb: Yep that was the reason to begin with - this is nice @tijmenb!

PR #1823, opened 2020-04-02
Provider users with the appropriate permissions can invite other provider users

steventux:
## Context

We want to enable provider users to onboard/invite other provider users.
This needs to be restricted by a specific permission and also within the context of the providers the current provider user can see.

See  https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1800 and https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1807 for some low level implementation details.
 

Using the `manage_users` permission in `ProviderPermissions` we allow certain users to invite other provider users.
We use the existing `InviteProviderUser` service and email template to create the user and send an invite.

image

## Guidance to review

You'll need the `provider_add_provider_users` feature flag on and to set `manage_users: true` for at least one `ProviderPermissions` record for the current provider user.

## Link to Trello card

https://trello.com/c/Cms7JJCh/1841-providerusers-with-the-appropriate-permissions-can-invite-other-providerusers

Paired with @willmcb on this feature.

## Things to check

### DIFF
--- a/app/controllers/provider_interface/provider_users_controller.rb
+++ b/app/controllers/provider_interface/provider_users_controller.rb
@@ -20,17 +20,15 @@ module ProviderInterface
         provider_user_params.merge(current_provider_user: current_provider_user),
       )
       provider_user = @form.build
-      render :new && return unless provider_user
+      render :new and return unless provider_user
 
-      service = InviteProviderUser.new(provider_user: provider_user)
-      begin
-        service.save_and_invite!
-        flash[:success] = 'Provider user invited'
-        redirect_to provider_interface_provider_users_path
-      rescue DfeSignInApiError => e # show errors from api
-        e.errors.each { |error| @form.errors.add(:base, error) }
-        render :new
-      end
+      InviteProviderUser.new(provider_user: provider_user).save_and_invite!
+
+      flash[:success] = 'Provider user invited'
+      redirect_to provider_interface_provider_users_path
+    rescue DfeSignInApiError => e
+      handle_dsi_error(e)
+      render :new
     end
 
   private
@@ -42,6 +40,14 @@ module ProviderInterface
 
     def requires_provider_add_provider_users_feature_flag
       raise unless FeatureFlag.active?('provider_add_provider_users')
+    end
+
+    def handle_dsi_error(form, exception)
+      Raven.capture_exception(exception)
+      form.errors.add(
+        :base,
+        'A problem occurred inviting this user. Please try again. If problems persist, please contact support.',
+      )
     end
   end
 end
--- a/app/controllers/provider_interface/provider_users_controller.rb
+++ b/app/controllers/provider_interface/provider_users_controller.rb
@@ -1,0 +1,38 @@
+module ProviderInterface
+  class ProviderUsersController < ProviderInterfaceController
+    def index
+      # TODO: A scope for this?
+      @provider_users = ProviderUser
+        .joins(:provider_users_providers)
+        .where('provider_users_providers.provider': current_provider_user.providers)
+    end
+
+    def new
+      @form = ProviderUserForm.new
+      @form.current_provider_user = current_provider_user
+    end
+
+    def create
+      @form = ProviderUserForm.new(provider_user_params)
+      provider_user = @form.build
+      render :new && return unless provider_user
+
+      service = InviteProviderUser.new(provider_user: provider_user)
+      begin
+        service.save_and_invite!
+        flash[:success] = 'Provider user created'
+        redirect_to provider_interface_provider_users_path
+      rescue DfeSignInApiError => e # show errors from api
+        e.errors.each { |error| @form.errors.add(:base, error) }
+        render :new
+      end
+    end
+
+  private
+
+    def provider_user_params
+      params.require(:provider_interface_provider_user_form)
+            .permit(:email_address, :first_name, :last_name, provider_ids: [])
+    end
+  end
+end
--- a/app/controllers/provider_interface/provider_users_controller.rb
+++ b/app/controllers/provider_interface/provider_users_controller.rb
@@ -1,5 +1,7 @@
 module ProviderInterface
   class ProviderUsersController < ProviderInterfaceController
+    before_action :requires_provider_add_provider_users_feature_flag, only: %i[new create]
+
     def index
       # TODO: A scope for this?
       @provider_users = ProviderUser
@@ -33,6 +35,10 @@ module ProviderInterface
     def provider_user_params
       params.require(:provider_interface_provider_user_form)
             .permit(:email_address, :first_name, :last_name, provider_ids: [])
+    end
+
+    def requires_provider_add_provider_users_feature_flag
+      raise unless FeatureFlag.active?('provider_add_provider_users')
     end
   end
 end
--- a/app/controllers/provider_interface/provider_users_controller.rb
+++ b/app/controllers/provider_interface/provider_users_controller.rb
@@ -3,10 +3,7 @@ module ProviderInterface
     before_action :requires_provider_add_provider_users_feature_flag, only: %i[new create]
 
     def index
-      # TODO: A scope for this?
-      @provider_users = ProviderUser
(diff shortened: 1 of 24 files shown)


app/controllers/provider_interface/provider_users_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+
+      @form = ProviderUserForm.new(current_provider_user: current_provider_user)
+    end
+
+    def create
+      @form = ProviderUserForm.new(
+        provider_user_params.merge(current_provider_user: current_provider_user),
+      )
+      provider_user = @form.build
+      render :new && return unless provider_user
```
duncanjbrown, line 23: This renders `204 No content` which effectively hides all the errors. Propose
```suggestion
      render :new and return unless provider_user
```

app/models/provider_interface/provider_user_form.rb
```diff
@@ -0,0 +1,27 @@
+module ProviderInterface
+  class ProviderUserForm < SupportInterface::ProviderUserForm
+    attr_accessor :current_provider_user
+
+    validate :permitted_providers
+
+    def available_providers
+      @available_providers ||= Provider
```
duncanjbrown, line 8: Could this be a use case for `ProviderOptionsService`?

lib/tasks/local_dev.rake
```diff
@@ -10,6 +10,8 @@ task setup_local_dev_data: %i[environment copy_feature_flags_from_production syn
     u.providers = [ApplicationChoice.first.provider]
   end
   SupportUser.find_or_create_by!(dfe_sign_in_uid: 'dev-support', email_address: 'support@example.com')
+
+  ProviderPermissions.update_all(manage_users: true) if FeatureFlag.active?('provider_add_provider_users')
```
duncanjbrown, line 14: :+1: Nice

spec/services/text_cardinalizer_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'rails_helper'
+
+RSpec.describe TextCardinalizer do
```
duncanjbrown, line 3: I like this thing 👍

app/controllers/provider_interface/provider_users_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+
+      service = InviteProviderUser.new(provider_user: provider_user)
+      begin
+        service.save_and_invite!
+        flash[:success] = 'Provider user invited'
+        redirect_to provider_interface_provider_users_path
+      rescue DfeSignInApiError => e # show errors from api
+        e.errors.each { |error| @form.errors.add(:base, error) }
```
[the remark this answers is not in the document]
steventux: :+1: Yes I agree, this was copied from the support interface provider users controller but the errors aren't necessarily _customer friendly_.

app/models/provider_interface/provider_user_form.rb
```diff
@@ -0,0 +1,27 @@
+module ProviderInterface
+  class ProviderUserForm < SupportInterface::ProviderUserForm
```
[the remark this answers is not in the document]
steventux: Yes good point, will separate the two.

app/models/provider_interface/provider_user_form.rb
steventux: :+1: Would it be better to move the scope `Provider.with_users_manageable_by(provider_user)` into this service to keep the model lean? This is the only place the scope is used.

app/controllers/provider_interface/provider_users_controller.rb
steventux: Fixed in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1823/commits/8d3f0bbb122f763fcc290ce4e83f78b630e3bae0 I did wonder where the correct place to test for these error cases is. It seems like we may bloat system specs testing for validations but testing errors in controller flow doesn't seem to have a common place in our tests...

app/models/provider_interface/provider_user_form.rb
steventux: Amended in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1823/commits/bb5abe8e96c3434943666f0f195f06dbb2cf4211

app/models/provider_interface/provider_user_form.rb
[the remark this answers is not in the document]
steventux: Amended here https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1823/commits/34852f343c6c40e899ea5ba123cac73cf60dc502

app/controllers/provider_interface/provider_users_controller.rb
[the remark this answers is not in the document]
steventux: I've simplified here https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1823/commits/8d3f0bbb122f763fcc290ce4e83f78b630e3bae0

steventux:
Thanks @duncanjbrown I think I've addressed review comments now.

app/models/provider_interface/provider_user_form.rb
duncanjbrown: :+1: Nice

duncanjbrown:
Changes look good, thanks @steventux — giving it a final spin...

duncanjbrown:
2 nits:

- The error messages need translations as they're generic "can't be blank"s

Screenshot 2020-04-07 at 17 02 52


- The list of providers has an unnecessary scroll bar beside it. I think this makes sense in the Support UI where there are hundreds of providers, but there will only ever be a few here

Screenshot 2020-04-07 at 17 00 13

steventux:
@duncanjbrown thanks for spotting that, amended in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1823/commits/8d777d7b1307a39d1d8e2e1471153b4e04f49428 and https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1823/commits/0ab8ba2dff5136d20f108f4e139dde3a89f2c279

Error messages are consistent with support interface form now:

image

tijmenb:
Can we do a PR on the content after this is merged? 

- I think the designs always say "User" rather than "Provider user"
- Title of the New page doesn't match h1
- I think we don't use the "must not be blank" validation message

steventux:
@tijmenb had to rebase cos of stuck CI run so https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1823/commits/d735360c5fde56259499492c7d02d47eb8da70d8 and https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1823/commits/a661583a7462863f636e71b8a2d4efa804fca87b should address the content issues you spotted.

steventux:
@mnacos that was the intention here https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1823/commits/14db761e9c9ced7f1a567613055880654c85c318 but I notice it doesn't work set the permission on the review app.

duncanjbrown:
q: What if a provider wants to add an existing user to their org? Currently they get this:

Screenshot 2020-04-09 at 08 57 00

The short-term fix is to tell them to contact support. Or we could handle this gracefully, and associate the existing user with our provider.

steventux:
@duncanjbrown it's a good question, I'd regard this as an edit to an existing user which isn't in scope for the current story: _you cannot edit or delete users either (this is for a future card)_ though I can't find card for provider users editing other provider users.

mnacos:
> @mnacos that was the intention here [14db761](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/commit/14db761e9c9ced7f1a567613055880654c85c318) but I notice it doesn't work set the permission on the review app.

Ah yes, that's probably because the feature flag is off when this is run on the review app. I've manually switched on the feature flag and given `dev-support` the `manage_users` permissions. I've also added some missing env vars from the Heroku config.

PR #1824, opened 2020-04-02
Add JS to prevent users from accidentally leaving a page with unsaved changes

tvararu:
## Context

This shows users a prompt when they attempt to navigate away from a page with a text area after they have made changes but don't submit them.

Add a JS module and unit test it using Jest.

## Guidance to review

Review commit by commit.

Screen Recording 2020-04-02 at 17 38 44

## Link to Trello card

https://trello.com/c/SgplKNq9/1093-dev-add-js-to-prevent-users-from-accidentally-leaving-a-page-with-unsaved-changes

## Things to check

### DIFF
--- a/app/frontend/packs/application.js
+++ b/app/frontend/packs/application.js
@@ -12,3 +12,4 @@ govUKFrontendInitAll();
 initNationalityAutocomplete();
 initProvidersAutocomplete();
 initCoursesAutocomplete();
+initWarnOnUnsavedChanges();
--- a/app/frontend/packs/application.js
+++ b/app/frontend/packs/application.js
@@ -3,7 +3,7 @@ import { initAll as govUKFrontendInitAll } from "govuk-frontend";
 import initNationalityAutocomplete from "./nationality-autocomplete";
 import initCoursesAutocomplete from "./courses-autocomplete";
 import initProvidersAutocomplete from "./providers-autocomplete";
-
+import initWarnOnUnsavedChanges from "./warn-on-unsaved-changes";
 
 import "accessible-autocomplete/dist/accessible-autocomplete.min.css";
 import "../styles/application.scss";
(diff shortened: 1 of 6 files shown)


app/views/candidate_interface/volunteering/review/show.html.erb
```diff
@@ -11,7 +11,7 @@
 
     <%= govuk_button_link_to t('application_form.volunteering.another.button'), candidate_interface_new_volunteering_role_path, class: 'govuk-button--secondary' %>
 
-    <%= form_with model: @application_form, url: candidate_interface_complete_volunteering_path do |f| %>
+    <%= form_with model: @application_form, url: candidate_interface_complete_volunteering_path, data: { module: 'form-check-leave' } do |f| %>
```
tijmenb, line 14: I didn't immediately grasp `form-check-leave`. Perhaps:

```suggestion
    <%= form_with model: @application_form, url: candidate_interface_complete_volunteering_path, data: { module: 'warn-on-unsaved-changes' } do |f| %>
```

tvararu: That's a much better name 👍

tvararu:
@tijmenb the card calls for it to only be on forms that have a `text_area` or a `"I have completed this section"` checkbox. We wouldn't want it on sign in forms, add a new support user forms, provider filter forms...

I do share your concern that this is very easy to forget to add.

Other ideas:

1. Adding it to all forms as a default, and making it possible to opt-out (opens up the opposite problems, what if we forget to opt out and it makes our new forms annoying to use)
2. Creating a smarter Ruby form component that automatically adds this to itself if it has a nested `text_area` or `"I have completed this section"` checkbox (this sounds tricky to implement)
3. Making the JavaScript target all forms that have `text_area`s or `"I have completed this section"` checkboxes when adding this, instead of relying on `data-module` (this is less explicit and could also cause issues)

I think 3 sounds best, should I give it a try?

PR #1825, opened 2020-04-02
Currently available courses page should group providers by region

stevehook:
## Context

As the list of providers and courses on Apply grows we need to make the list of available providers and courses easier to navigate. It makes sense to group the providers into regions.

- [x] Group courses by provider region and change views to present an accordian per region under region headings.
- [x] Feature flag
- [x] New system spec
- [x] Refactor to components?
  
## Guidance to review

- Does the grouping logic make sense? I've moved the sorting into the DB and but the grouping happens in Ruby code still.
- I think we will run into issues with performance in the longer-term with this design (and the previous one) as more providers become available on Apply. Not sure what we can do to mitigate that without changing the design to present a subset of the results or load subsets on demand. Any thoughts on this welcome.

In mine tests I found the performance to be OK with 1000 and 5000 courses. Both page loads took about a second and it didn't degrade significantly with increased numbers of courses. So I think this is OK for now - we are probably going to have to address usability before performance.

image 

## Link to Trello card

https://trello.com/c/luWz8uHE/975-dev-currently-available-providers-page-grouped-by-region

## Things to check

### DIFF
--- a/app/components/grouped_provider_courses_component.html.erb
+++ b/app/components/grouped_provider_courses_component.html.erb
@@ -1,0 +1,21 @@
+<% courses_by_provider_and_region.each do |region_code, courses_by_provider| %>
+  <h2 class="govuk-heading-m govuk-!-margin-bottom-2 govuk-!-margin-top-8"><%= t("provider_regions.#{region_code}") %></h2>
+  <div class="govuk-accordion" data-module="govuk-accordion" id="accordion-default">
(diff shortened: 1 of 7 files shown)


tvararu:
> In mine tests I found the performance to be OK with 1000 and 5000 courses. Both page loads took about a second and it didn't degrade significantly with increased numbers of courses. So I think this is OK for now - we are probably going to have to address usability before performance.

💯 to you for testing!

PR #1826, opened 2020-04-02
Flag accredited bodies not on Apply when opening courses

duncanjbrown:
## Context

Some accredited bodies are responsible for sifting, making offers etc for applications to courses they ratify. During a meeting about access controls, it was pointed out that it would be bad if we opened courses for a provider but the accredited body responsible for some of those courses was locked out. We should help ourselves not to make this mistake.

Screenshot 2020-04-02 at 22 48 37

## Guidance to review

- I used the presence of a `ProviderAgreement` as a proxy for the provider being onboarded. I think this is probably good enough? It ought to also cover providers who are set up as clients of the API (it doesn't yet, but https://trello.com/c/qerAetcs/1855-api-access-should-depend-on-the-presence-of-a-dsa)
- No tests 🤠. We don't really have a place for this simple logic to live. Maybe it's simple enough?

## Link to Trello card

https://trello.com/c/L7Y8vxNf/1856-add-a-note-to-the-support-console-warning-that-we-should-be-careful-when-opening-all-courses-for-providers-with-accredited-bodie

## Things to check

### DIFF
--- a/app/components/support_interface/provider_courses_table_component.html.erb
+++ b/app/components/support_interface/provider_courses_table_component.html.erb
@@ -39,7 +39,12 @@
           <% end %>
         </td>
         <% if accredited_bodies_vary? %>
-          <td class='govuk-table__cell'><%= course[:accredited_body] %></td>
+          <td class='govuk-table__cell'><%= course[:accredited_body] %>
+            <% unless course[:accredited_body_onboarded] %>
+              <br />
+              (No users on Apply)
+            <% end %>
+          </td>
         <% end %>
       </tr>
     <% end %>
--- a/app/controllers/support_interface/providers_controller.rb
+++ b/app/controllers/support_interface/providers_controller.rb
@@ -14,7 +14,7 @@ module SupportInterface
     end
 
     def courses
-      @provider = Provider.includes(:courses).find(params[:provider_id])
+      @provider = Provider.includes(courses: [:accredited_provider]).find(params[:provider_id])
     end
 
     def vacancies
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -37,4 +37,13 @@ class Provider < ApplicationRecord
   def application_forms
     ApplicationForm.where(id: application_choices.select(:application_form_id))
   end
+
+  def onboarded?
+    provider_agreements.any?
+  end
+
+  def all_associated_accredited_providers_onboarded?
+    accredited_providers = courses.map(&:accredited_provider).uniq.compact
+    accredited_providers.all?(&:onboarded?)
+  end
 end
--- a/app/components/support_interface/provider_courses_table_component.rb
+++ b/app/components/support_interface/provider_courses_table_component.rb
@@ -17,6 +17,7 @@ module SupportInterface
           apply_from_find_link: link_to_apply_from_find_page(course),
           link_to_find_course_page: link_to_find_course_page(course),
           accredited_body: link_to_provider_page(course.accredited_provider),
+          accredited_body_onboarded: course.accredited_provider&.onboarded?,
         }
       end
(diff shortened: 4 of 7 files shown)


app/models/provider.rb
```diff
@@ -37,4 +37,8 @@ def accredited_courses
   def application_forms
     ApplicationForm.where(id: application_choices.select(:application_form_id))
   end
+
+  def onboarded?
+    ProviderAgreement.find_by(provider: self).present?
```
tijmenb, line 42:
```suggestion
    provider_agreements.any?
```

results in a more efficient query:

```
[1] pry(main)> ProviderAgreement.find_by(provider: Provider.last).present?
  Provider Load (1.8ms)  SELECT "providers".* FROM "providers" ORDER BY "providers"."id" DESC LIMIT $1  [["LIMIT", 1]]
  ProviderAgreement Load (0.4ms)  SELECT "provider_agreements".* FROM "provider_agreements" WHERE "provider_agreements"."provider_id" = $1 LIMIT $2  [["provider_id", 5], ["LIMIT", 1]]
=> false
[2] pry(main)> Provider.last.provider_agreements.any?
  Provider Load (0.4ms)  SELECT "providers".* FROM "providers" ORDER BY "providers"."id" DESC LIMIT $1  [["LIMIT", 1]]
  ProviderAgreement Exists? (0.4ms)  SELECT 1 AS one FROM "provider_agreements" WHERE "provider_agreements"."provider_id" = $1 LIMIT $2  [["provider_id", 5], ["LIMIT", 1]]
=> false
```

app/controllers/support_interface/providers_controller.rb
```diff
@@ -14,7 +14,11 @@ def show
     end
 
     def courses
-      @provider = Provider.includes(:courses).find(params[:provider_id])
+      @provider = Provider.includes(courses: [:accrediting_provider]).find(params[:provider_id])
+
+      accredited_providers = @provider.courses.map(&:accrediting_provider).uniq.compact
+
+      @all_accredited_providers_onboarded = accredited_providers.all?(&:onboarded?)
```
tijmenb, line 21: Perhaps these two lines can be extracted into a service/Provider model so you can add a unit test?

app/components/support_interface/provider_courses_table_component.html.erb
```diff
@@ -39,7 +39,12 @@
           <% end %>
         </td>
         <% if accredited_bodies_vary? %>
-          <td class='govuk-table__cell'><%= course[:accredited_body] %></td>
+          <td class='govuk-table__cell'><%= course[:accredited_body] %>
```
tijmenb, line 42: Can we link this to the provider page?

app/models/provider.rb
duncanjbrown: oops - yes!

app/components/support_interface/provider_courses_table_component.html.erb
duncanjbrown: It already links to the accredited body's provider page? https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/d2bf03c502ee7d74e917ed56dc5aa00ac918eb61/app/components/support_interface/provider_courses_table_component.rb#L19

app/controllers/support_interface/providers_controller.rb
duncanjbrown: Sure — maybe `Provider` isn't a bad place for this to live

duncanjbrown:
Thanks @tijmenb!

duncanjbrown:
@tijmenb should be fixed up now. Let me know if there's any more to do on the link

PR #1834, opened 2020-04-06
Expose applications that need support intervention

tijmenb:
## Context

There are ~3~ 4 situations in which applications need manual intervention:

- The course is removed from Apply
- The course is removed from Find
- The course's site is removed from Find
- The course is full

Add a page with applications that need intervention. It's fairly MVP! Also a few UI tweaks to make it easier to navigate the support section.

Screenshot 2020-04-06 at 14 44 15

## Guidance to review

Does it make sense?

## Link to Trello card

https://trello.com/c/fRAKJt36/1264-expose-applications-that-need-support-intervention

## Things to check

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -55,4 +55,8 @@ class Course < ApplicationRecord
   def full?
     course_options.all?(&:no_vacancies?)
   end
+
+  def find_url
+    @find_url ||= "https://www.find-postgraduate-teacher-training.service.gov.uk/course/#{provider.code}/#{code}"
+  end
 end
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -57,6 +57,6 @@ class Course < ApplicationRecord
   end
 
   def find_url
-    @find_url ||= "https://www.find-postgraduate-teacher-training.service.gov.uk/course/#{provider.code}/#{code}"
+    "https://www.find-postgraduate-teacher-training.service.gov.uk/course/#{provider.code}/#{code}"
   end
 end
--- a/app/services/support_interface/application_monitor.rb
+++ b/app/services/support_interface/application_monitor.rb
@@ -12,9 +12,15 @@ module SupportInterface
       ).map(&:application_form)
     end
 
-    def applications_to_full_courses
+    def applications_to_removed_sites
       active_applications.where(
         course_option: CourseOption.where(invalidated_by_find: true),
+      ).map(&:application_form)
+    end
+
+    def applications_to_courses_with_sites_without_vacancies
+      active_applications.where(
+        course_option: CourseOption.where(vacancy_status: 'no_vacancies'),
       ).map(&:application_form)
     end
 
--- a/app/services/support_interface/application_monitor.rb
+++ b/app/services/support_interface/application_monitor.rb
@@ -1,21 +1,27 @@
 module SupportInterface
   class ApplicationMonitor
     def applications_to_disabled_courses
-      ApplicationChoice.where(
+      active_applications.where(
         course_option: CourseOption.joins(:course).where('courses.open_on_apply' => false),
       ).map(&:application_form)
     end
 
     def applications_to_hidden_courses
-      ApplicationChoice.where(
+      active_applications.where(
         course_option: CourseOption.joins(:course).where('courses.open_on_apply' => true, 'courses.exposed_in_find' => false),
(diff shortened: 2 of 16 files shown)


app/services/support_interface/application_monitor.rb
```diff
@@ -0,0 +1,21 @@
+module SupportInterface
+  class ApplicationMonitor
```
duncanjbrown, line 2: Once an application has finished its life — ie enrolled, declined, withdrawn, etc, should we filter it out of these queries?

app/models/course.rb
```diff
@@ -55,4 +55,8 @@ def both_study_modes_available?
   def full?
     course_options.all?(&:no_vacancies?)
   end
+
+  def find_url
```
duncanjbrown, line 59: 🎉

tvararu: Long overdue 👍

app/services/support_interface/application_monitor.rb
tijmenb: Yeah, that's a good point. Pushed a commit.

app/models/course.rb
```diff
@@ -55,4 +55,8 @@ def both_study_modes_available?
...
+    @find_url ||= "https://www.find-postgraduate-teacher-training.service.gov.uk/course/#{provider.code}/#{code}"
```
tvararu, line 60: Is the memoization necessary?

```suggestion
    "https://www.find-postgraduate-teacher-training.service.gov.uk/course/#{provider.code}/#{code}"
```

app/services/support_interface/application_monitor.rb
```diff
@@ -1,21 +1,27 @@
...
-      ApplicationChoice.where(
+      active_applications.where(
         course_option: CourseOption.where(invalidated_by_find: true),
       ).map(&:application_form)
     end
+
+  private
+
+    def active_applications
+      ApplicationChoice.where('status NOT IN (?)', %w[rejected declined withdrawn enrolled])
```
duncanjbrown, line 24:
```suggestion
      ApplicationChoice.where('status NOT IN (?)', %w[rejected declined withdrawn enrolled conditions_not_met])
```

app/services/support_interface/application_monitor.rb
```diff
@@ -0,0 +1,27 @@
...
+      ).map(&:application_form)
+    end
+
+    def applications_to_hidden_courses
+      active_applications.where(
+        course_option: CourseOption.joins(:course).where('courses.open_on_apply' => true, 'courses.exposed_in_find' => false),
+      ).map(&:application_form)
+    end
+
+    def applications_to_full_courses
```
malcolmbaig, line 15: I think we have a couple of conflicting concepts here. In hindsight I should've given `invalidated_by_find` a more specific name, but what that field indicates currently is whether a site is still a valid selection for the chosen course, or if it's been removed. That's distinct from whether or not the site is full, which is indicated by `vacancy_status: :no_vacancies`.

"full_courses" is probably best described as any course where all course options are `vacancy_status: :no_vacancies`. At a simpler level, maybe we could make this `applications_to_full_sites` which would query:

```
      active_applications.where(
        course_option: CourseOption.where(vacancy_status: 'no_vacancies'),
      ).map(&:application_form)
```

tijmenb: O wow, great catch. I've changed it up here: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1834/commits/4746c8823e2f09ef1a2a338f69a0897400b32648

spec/services/support_interface/application_monitor_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      applications = described_class.new.applications_to_full_courses
+
+      expect(applications.map(&:id)).not_to include(with_vacancies.application_form_id)
+      expect(applications.map(&:id)).to include(without_vacancies.application_form_id)
+    end
+  end
+
+  describe '#applications_to_removed_sites' do
+    it 'returns applications to sites that have been removed from Apply' do
```
malcolmbaig, line 48: nitpick: `removed from Find`, as we don't actively delete sites on our end.

malcolmbaig:
Couple of additional comments, but otherwise LGTM 👍

PR #1837, opened 2020-04-06
Fix for safeguarding false positives 

stevehook:
## Context

There have been at least two support calls about safeguarding being flagged to providers when no such declaration was made by the candidate. This we believe is due to `null` values for applications that pre-date the addition of the `application_forms#safeguarding_issues` column.

- [x] Add a `ApplicationForm#safeguarding_issues_status` attribute and use `ApplicationForm#safeguarding_issues` just for the text of any safeguarding issue. In other words stop using special string values (`Yes` and `No`) in `ApplicationForm#safeguarding_issues`.
- [x] Give support users visibility of safeguarding issues by adding it to the application detail view.

## Guidance to review

- Does the upgrade cover all the cases? (unanswered questions, never asked as well as yes and no answers).
- Does the copy make sense? (Tried to be consistent with https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1828)
- I think we need to revisit `ApplicationReference#safeguarding_concerns` and break that into two attributes `safeguarding_concerns` and `safeguarding_concerns_status` too but I think it's best to do in a separate PR.
- There is a migration in this PR but it doesn't alter the schema so it's not going to cause any 'missing column' issues. Is this OK?
- This feature isn't switched on in production. Does this PR leak any new functionality?

Provider interface:

image

Support interface:

image

## Link to Trello card

https://trello.com/c/cl0rRz5F/1263-dev-suitability-to-work-with-children-incorrectly-flagged-to-providers-twice

## Things to check

### DIFF
--- a/app/components/provider_interface/safeguarding_declaration_component.html.erb
+++ b/app/components/provider_interface/safeguarding_declaration_component.html.erb
@@ -1,7 +1,3 @@
 <h2 class="govuk-heading-m govuk-!-margin-top-8">Criminal convictions and professional misconduct</h2>
 
-<% if has_disclosed_safeguarding_issues? %>
-  <p class="govuk-body">The candidate has shared information related to safeguarding. Please contact them directly for more details.</p>
-<% else %>
-  <p class="govuk-body">No information shared</p>
-<% end %>
+<p class="govuk-body"><%= message %></p>
--- a/app/components/provider_interface/safeguarding_declaration_component.rb
+++ b/app/components/provider_interface/safeguarding_declaration_component.rb
@@ -2,30 +2,17 @@ module ProviderInterface
   class SafeguardingDeclarationComponent < ActionView::Component::Base
     include ViewHelper
 
-    attr_reader :application_form
+    attr_reader :safeguarding_status
 
     def initialize(application_form:)
-      @application_form = application_form
+      @safeguarding_status = SafeguardingStatus.new(
+        application_form: application_form,
+        i18n_key: 'provider_interface.safeguarding_declaration_component',
+      )
     end
 
     def message
-      if has_no_answer?
-        I18n.t('provider_interface.safeguarding_declaration_component.no_answer_message')
-      elsif has_disclosed_safeguarding_issues?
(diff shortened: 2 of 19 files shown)


tijmenb:
@tvararu I'd prefer to keep them separate, there might be changes in the future that apply only to support (like exposing the data), and we don't want to accidentally leak between interfaces. This is what happened with displaying referee info.

I would try to DRY up the two components by extracting the content & conditionals to a separate class.

tvararu:
@tijmenb very reasonable 👍

tijmenb:
I've pondered this some more, and I think we should go even further and try to solve the underlying issue - which I think is that we rely on non-explicit values in the database.

As I understand the behaviour of the form:

* Answering No sets the column to "No".
* Answering Yes, but no explanation will set the column to "Yes"
* Answering Yes, and add an explanation sets the column to the explanation
* And if they've not answered, the column is nil

My proposal would be to add a new non-nil enum column to the application form that represents the (lack of) answer of the user:

- yes_has_safeguarding_issues_to_declare
- no_has_no_safeguarding_issues_to_declare
- never_asked (for backfill)
- not_answered_yet (default)

This will remove complexity from `SafeguardingIssuesDeclarationForm`, stop us from relying on magic strings, requires upfront thinking about defaults and backfilling, and be very explicit in the database.

Thoughts?

stevehook:
@tijmenb @tvararu I agree it's the right thing to do because I'm not all that happy about squeezing a boolean (or enum) as well as a text value into a single column as we are now. So happy to make the change you suggest.

Only slightly tricky bit is determining which records get the `never_asked` value rather than the `not_answered` default. I guess those the `nil` values need to be mapped to `never_asked` if the application has been submitted? Not sure about this though.

Also I was wondering whether the same applies to safeguarding attribute on references?

tijmenb:
> I guess those the nil values need to be mapped to never_asked if the application has been submitted?

That seems correct. Perhaps the reverse works too - we can set all `nils` to `never_asked` and then only update the unsubmitted forms to `not_answered`.

tijmenb:
> Also I was wondering whether the same applies to safeguarding attribute on references?

Yeah, that would make sense there as well I think.

PR #1842, opened 2020-04-07
Implement the new tabbed design for the provider ui single application page

willmcb:
## Context

Adds tabs component to application. This was deemed as a precursor / blocker for the notes work as discussed in the notes kick off - as such it was bumped up in priority.

**Before**
Screenshot 2020-04-07 at 16 38 23

**After**
Screenshot 2020-04-07 at 16 37 42

## Guidance to review

I have used the pattern defined elsewhere in the app to control if a tab is selected, or not (see `SubNavigationComponent` for an example). In short the component accepts a hash of 'items' that are interrogated to render the tabs. Which tab is selected depends on the presence of a `current: true` hash value that should differ depending on the page. Does this make sense? Do you like this implementation?

I have included the component in `app/views/provider_interface/application_choices/show.html.erb` as I needed to get the screenshots - however as the Notes and Timeline `show` views / routes don't exist yet these tabs have `#` as url values and lead nowhere thus far. 

Happy to remove the `TagsComponent` from rendering here until the other views are available if you think this is better (for instance there would be the chance that this goes live and the application show tabs have dead links in them - not a good look). 

This is also the reason why I haven't amended any system tests to check for tabs functionality.

## Link to Trello card

https://trello.com/c/U6usyWvE/1865-implement-the-new-tabbed-design-for-the-provider-ui-single-application-page

## Things to check

### DIFF
--- a/app/frontend/styles/_application-tabs.scss
+++ b/app/frontend/styles/_application-tabs.scss
@@ -1,0 +1,16 @@
+.app-tabs {
+  margin-top: govuk-spacing(8);
+
+  &__tab {
+    font-weight: bold !important;
+    text-align: left !important;
+    text-decoration: none !important;
+  }
+
+  &__list-item--selected {
+    background-color: govuk-colour("white") !important;
+    border-top: govuk-spacing(1) solid $govuk-link-colour !important;
+    color: $govuk-text-colour !important;
+    padding-top: govuk-spacing(2) !important;
+  }
+}
--- a/app/views/provider_interface/application_choices/show.html.erb
+++ b/app/views/provider_interface/application_choices/show.html.erb
@@ -27,6 +27,7 @@
   <%= @application_choice.application_form.full_name %>
   <%= render(ProviderInterface::ApplicationStatusTagComponent.new(application_choice: @application_choice)) %>
 </h1>
+
 <p class="govuk-body govuk-!-margin-bottom-9 app-!-print-display-none">
   <%= govuk_link_to(
     'Download application (PDF)',
--- a/spec/components/sub_navigation_component_spec.rb
+++ b/spec/components/sub_navigation_component_spec.rb
@@ -1,0 +1,27 @@
+require 'rails_helper'
+
+RSpec.describe SubNavigationComponent do
+  let(:items) {
+    [{ name: 'Application', url: '#', current: true },
+     { name: 'Notes', url: '#' },
+     { name: 'Timeline', url: '#' }]
+  }
+
+  context 'nav tabs appearing as selected' do
+    it 'when the item is "current" then that tab is selected' do
+      result = render_inline(described_class.new(items: items))
+
+      expect(result.css('.govuk-tabs__list-item--selected').text).to include('Application')
+    end
+  end
+
+  context 'rendering tabs' do
+    it 'renders all of the nav tabs specified in the items hash passed to it' do
+      result = render_inline(described_class.new(items: items))
+
+      expect(result.text).to include('Application')
+      expect(result.text).to include('Notes')
+      expect(result.text).to include('Timeline')
+    end
+  end
+end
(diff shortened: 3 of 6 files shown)


tijmenb:
Looks good! Instead of a new component can we update the `SubNavigationComponent` to the new style and use that? Then we'll have the new nice tabs everywhere 💯 

If you use `SubNavigationComponent` you can omit `current: true` because it will highlight the current page by default: 

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/4de59158ae725b4007a8c471151cf2866f850da1/app/components/sub_navigation_component.html.erb#L5

willmcb:
@tijmenb nice, thanks for that, it makes sense to have one thing rather than two! Looking at their implementations they are near  identical. 

Spoke to @adamsilver re this change he said that we may see some undesirable wrapping of the tabs on the support interface but this can be fixed by better organisation in any case.

willmcb:
See http://g.recordit.co/F2HjmRj8sH.gif for support tabs nav effects.

app/frontend/styles/_application-tabs.scss
```diff
@@ -0,0 +1,16 @@
...
+    text-align: left !important;
+    text-decoration: none !important;
+  }
+
+  &__list-item--selected {
+    background-color: govuk-colour("white") !important;
+    border-top: govuk-spacing(1) solid $govuk-link-colour !important;
+    color: $govuk-text-colour !important;
+    padding-top: govuk-spacing(2) !important;
+  }
```
tijmenb, line 15: @tvararu what's our opinion on SCSS shortcuts like this? I'd prefer explicit & non-nested CSS.

app/views/provider_interface/application_choices/show.html.erb
```diff
@@ -27,6 +27,11 @@
   <%= @application_choice.application_form.full_name %>
   <%= render(ProviderInterface::ApplicationStatusTagComponent.new(application_choice: @application_choice)) %>
 </h1>
+
+<%= render SubNavigationComponent.new(items: [{name: 'Application', url: provider_interface_application_choice_path(@application_choice), current: true},
+                                                        {name: 'Notes', url: '#'},
+                                                        {name: 'Timeline', url: '#'}]) %>
```
tijmenb, line 33: If we remove this for now this PR can be merged before we have notes & timeline.

spec/components/sub_navigation_component_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'rails_helper'
+
+RSpec.describe SubNavigationComponent do
```
tijmenb, line 3: 👍 for adding a test.

app/frontend/styles/_application-tabs.scss
willmcb: I have no strong opinions either way - but this is the style me and @tvararu used for in `_application-card.scss` for the `ApplicationCardComponent`.

app/views/provider_interface/application_choices/show.html.erb
willmcb: Done

app/frontend/styles/_application-tabs.scss
tijmenb: Yeah, that's actually where my preference comes from! I was fiddling with the card CSS yesterday and I searched for one of the classes in my editor. I was really confused why it didn't show up.

willmcb: haha, right yeah I can see that being an issue! I quite like the succinct / DRYness of being able to nest things though.

Is this something that you think I should address in this PR, or something we can take note of and discuss as a group? Similar to how we were speaking about JS styles / patterns - it might be good to have a conversation around SCSS as well. What do you think?

duncanjbrown: I quite like this per-file as it follows the BEM hierarchy & usually grepping for `__thing` gets what you need. Maybe one for dev retro?

PR #1846, opened 2020-04-08
Experimental: Add JS tests to GitHub actions

duncanjbrown:
Experiment following #1844

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -230,28 +230,8 @@ jobs:
     name: 'JavaScript tests'
     runs-on: ubuntu-latest
 
-    services:
-      postgres:
-        image: 'postgres:11.5'
-        ports:
-          - '5432:5432'
-        options: '--health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5'
-      redis:
-        image: redis
-        ports:
-          - '6379:6379'
-        options: '--entrypoint redis-server'
-
     steps:
       - uses: actions/checkout@v1
-
-      - name: Set up gem cache
-        uses: actions/cache@v1
-        with:
-          path: vendor/bundle
-          key: ${{ runner.os }}-gem-${{ hashFiles('**/Gemfile.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-gem-
 
       - uses: actions/setup-node@v1
         with:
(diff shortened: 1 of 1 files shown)


duncanjbrown:
Seems to work. Pushing another commit to make a test fail...

duncanjbrown:
It failed 🎉

duncanjbrown:
There's an unrelated transient failure in the acceptance specs. Card: https://trello.com/c/jU6gucuI/1884-fix-transient-spec-failure

.github/workflows/tests.yml
```diff
@@ -225,3 +225,72 @@ jobs:
...
+      postgres:
+        image: 'postgres:11.5'
+        ports:
+          - '5432:5432'
+        options: '--health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5'
+      redis:
+        image: redis
+        ports:
+          - '6379:6379'
+        options: '--entrypoint redis-server'
```
tvararu, line 243: Don't think these are needed to run JS tests 🤔

.github/workflows/tests.yml
```diff
@@ -225,3 +225,72 @@ jobs:
...
+          PGHOST: localhost
+          PGUSER: postgres
+          RAILS_ENV: test
+        run: |
+          yarn install
+          gem install bundler
+          bundle config path vendor/bundle
+          bundle install --jobs 4 --retry 3
+          bin/rails db:setup
+          RAILS_ENV=test bin/rails webpacker:compile
```
tvararu, line 292: None of these commands and their Ruby version prerequisites should be necessary to run JS tests 🤔

.github/workflows/tests.yml
```diff
@@ -225,3 +225,72 @@ jobs:
...
+
+      - name: 'Run tests'
+        run: |
+          npx jest
```
tvararu, line 296: The `npx` equivalent for `yarn` is simply `yarn`:

```suggestion
          yarn jest
```

.github/workflows/tests.yml
duncanjbrown: We need to run Webpacker, which depends on Rails, which depends on... everything

.github/workflows/tests.yml
duncanjbrown: Thanks — forgot which JS ecosystem I was in 😂

.github/workflows/tests.yml
tvararu: Turns out we don't; we need webpacker for the JavaScript that we want to run in the browser. Jest however runs in Node, so we don’t need to compile or do anything to the code, it just runs natively.

PR #1847, opened 2020-04-08
Handle case of nil region_code in grouped providers

stevehook:
## Context

There are a small number of providers that do not have `region_code` values when we sync them from the Find API. We don't handle these properly in the grouped providers interface. This PR fixes that bug.

- [x] Treat `region_code` `nil` values as 'No Region'.

image

## Guidance to review

- Do we need a special category for `nil` region codes?

## Link to Trello card

https://trello.com/c/9qiEqzo0/1299-ensure-providers-without-a-regioncode-are-shown-on-providers-page

## Things to check


paulrobertlloyd:
@laurahermionetennant92 @EmmaFrith There may be cases where we have to show providers on this page who don’t have a region associated with them.

Speaking to @stevehook, the plan is to flag up when a provider is synced and has no region associated with it, so hopefully we could prevent this from happening, or fix it when it does. But, should that not happen, what should we use as a title here?

EmmaFrith:
@paulrobertlloyd @stevehook @tvararu maybe if we do need this category it should be something like 'No region specified' because it's not like it isn't in a region!

tvararu:
@EmmaFrith I also thought about "No region specified," but thought maybe it doesn't fit nationwide providers so well, since they couldn't specify a region even if they wanted to. But I think it works 👍

PR #1848, opened 2020-04-08
Fix for out-of-order migrations

stevehook:
## Context

`20200330153007_rename_courses_accrediting_provider_id_part1` depends on `20200331091319_add_courses_accredited_provider_id` but with the original timestamps they ran in the wrong order.

These migrations already ran on production in the correct order but are causing problems in some local branches and review apps.

- [x] Modify the logic in each migration to allow for them being run in the wrong order.

## Guidance to review

Modified the original migrations rather than flipping the timestamps because that would have caused a different problem in any branch that had already run one of the migrations sucessfully.

I tested it by reset the DB back to 29th March. Then running the migrations (in the wrong order).

## Link to Trello card

n/a

## Things to check

### DIFF
--- a/db/migrate/20200330153007_rename_courses_accrediting_provider_id_part1.rb
+++ b/db/migrate/20200330153007_rename_courses_accrediting_provider_id_part1.rb
@@ -1,5 +1,9 @@
 class RenameCoursesAccreditingProviderIdPart1 < ActiveRecord::Migration[6.0]
   def up
+    unless column_exists?(:courses, :accredited_provider_id)
+      add_column :courses, :accredited_provider_id, :integer, null: true
+    end
+
     execute 'UPDATE courses SET accredited_provider_id = accrediting_provider_id'
   end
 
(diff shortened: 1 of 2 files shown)


db/migrate/20200330153007_rename_courses_accrediting_provider_id_part1.rb
```diff
@@ -1,5 +1,9 @@
 class RenameCoursesAccreditingProviderIdPart1 < ActiveRecord::Migration[6.0]
   def up
+    unless column_exists?(:courses, :accredited_provider_id)
+      add_column :courses, :accredited_provider_id, :integer, null: true
+    end
```
tvararu, line 5: Should we add the reverse of this in the `down`?

stevehook: I don't think so because this migration wasn't supposed to add it and doing so could mess up the other migration getting rolled back (maybe). If it needs to be dropped it would be dropped by the `add_courses_accredited_provider_id` migration whichever order they are run in.

PR #1849, opened 2020-04-08
Add notes model

willmcb:
## Context

Adds notes model - first in series of commits to build notes functionality:

Proposed specs:
1. models
2. feature flag, controller, routes, views, system spec, other specs
3. join up with tabs, amend system specs where needed.

Might make sense as a polymorphic model but was deemed to be out of scope for now.

- new model to represent notes and associations herein.

## Guidance to review

- Does this look correct to you? Is application choices the right model to `has_many: :notes`

- Not sure about - backfilling for consistency as mentioned in the checklist below - I guess not as this hasn't existed until now - please advise. 

- Is it right that we have notes as `dependant: :destroy` in `application_choices`? I was thinking yes as they relate to individual `application_choices`. I wonder though might we need them for auditing or something even if they are orphan records? Perhaps a better option would be to `NULL` out the content and keep the timestamps or something. 

## Link to Trello card
https://trello.com/c/mscEDZr5/1780-%F0%9F%8F%88-notes

## Things to check


app/models/application_choice.rb
```diff
@@ -11,6 +11,8 @@ class ApplicationChoice < ApplicationRecord
   has_one :provider, through: :course
   has_one :accredited_provider, through: :course, class_name: 'Provider'
 
+  has_many :notes, depenant: :destroy
```
tijmenb, line 14: Typo here, but we don't need `dependent: :destroy` because by default we do a "cascading delete" in the database. You can [see it in action in this spec](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/spec/models/candidate_spec.rb#L15-L35). It's probably worth adding an extra test for these models to that spec to make sure it does what we want.

spec/factories.rb
```diff
@@ -1,4 +1,10 @@
 FactoryBot.define do
+  factory :note do
```
tijmenb, line 2: I'd skip this until we need it, and we'd want random values for this.

spec/models/note_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
```
tijmenb, line 1: I'd also skip adding this file for now (not all models have a test).

db/migrate/20200408150738_create_notes.rb
```diff
@@ -0,0 +1,11 @@
+class CreateNotes < ActiveRecord::Migration[6.0]
+  def change
+    create_table :notes do |t|
+      t.string :title
+      t.text :message
+      t.belongs_to :application_choice, null: false, foreign_key: true
```
tijmenb, line 6: Should we add a relationship to the author (provider user) here as well?

willmcb: Good point - knowing who wrote a not is probably something we would want to know haha.

app/models/application_choice.rb
willmcb: This was tripping a rubocop so I have added a global disable for Disable `Rails/HasManyOrHasOneDependent`.

db/migrate/20200409090712_add_provider_user_to_notes.rb
```diff
@@ -0,0 +1,5 @@
+class AddProviderUserToNotes < ActiveRecord::Migration[6.0]
+  def change
+    add_reference :notes, :provider_user, null: false, foreign_key: true
```
tijmenb, line 3: Could this go into the original migration so the PR is smaller?

db/schema.rb
```diff
@@ -364,6 +375,8 @@
   add_foreign_key "course_options", "sites", on_delete: :cascade
   add_foreign_key "courses", "providers"
   add_foreign_key "emails", "application_forms", on_delete: :cascade
+  add_foreign_key "notes", "application_choices"
```
tijmenb, line 378: I was wrong about the default, we need to add something to make this a `on_delete: :cascade`.

willmcb: Do yo mean adding this as in the `ProviderUser` / `ApplicationChoice` model or to the DB as a whole?

app/models/application_choice.rb
stevehook: Having a cascading delete constraint in the DB is a good thing. Not sure whether we should have `dependent: destroy` is a bad thing though. We have lots of `presence: true` validations in Rubyland backed by `NOT NULL` constraints in the DB. It made me wonder what happens to the models when a cascading delete is triggered in the DB - does the collection of dependent models in memory get out of sync with the DB?

willmcb: 🤔 good point - have had a bit of a google around and haven't seen anything that would explain what is happening here.

db/schema.rb
```diff
@@ -255,8 +266,8 @@
   create_table "provider_users_providers", force: :cascade do |t|
     t.bigint "provider_id", null: false
     t.bigint "provider_user_id", null: false
-    t.datetime "created_at", default: -> { "now()" }, null: false
-    t.datetime "updated_at", default: -> { "now()" }, null: false
+    t.datetime "created_at", default: -> { "CURRENT_TIMESTAMP" }, null: false
```
mnacos, line 269: Can we preserve `now()`? Not sure `CURRENT_TIMESTAMP` will do what's expected.

duncanjbrown: I think this is an acknowledged difference in behaviour between postgres versions, and Rails has decided not to care? https://github.com/yahonda/rails/commit/4181b677e5164a18fe25767c5bf9eeb3df800231

PR #1851, opened 2020-04-08
Providers can make offers on courses they ratify

duncanjbrown:
## Context

This recently bit a provider who was associated with the accredited body for a course. They can see the choice and all the UI associated with offers, but they got a 500 when they tried to make the offer.

https://sentry.io/organizations/dfe-bat/issues/1603184950/?project=1765973&referrer=slack

We also need to fix the fact the user saw a 500 (though it did surface this error to us this time): https://trello.com/c/DRw9n8KJ/1892-show-a-sensible-error-when-a-provider-user-tries-to-make-an-offer-on-a-course-they-dont-have-permission-to.

When we decide whether to make an offer, use the existing GetApplicationChoicesForProviders service to see if they can see this application. If they can, OK.

## Guidance to review

Is it correct?

## Link to Trello card

https://trello.com/c/DRw9n8KJ/1892-show-a-sensible-error-when-a-provider-user-tries-to-make-an-offer-on-a-course-they-dont-have-permission-to

## Things to check

### DIFF
--- a/app/services/provider_authorisation.rb
+++ b/app/services/provider_authorisation.rb
@@ -10,17 +10,17 @@ class ProviderAuthorisation
 
     supplied_course_option = CourseOption.find(course_option_id) if course_option_id
     if supplied_course_option && course_option_id != application_choice.course_option.id
-      application_choice_belongs_to_user_providers?(application_choice: application_choice) && \
+      application_choice_visible_to_user?(application_choice: application_choice) && \
         course_option_belongs_to_user_providers?(course_option: supplied_course_option)
     else
-      application_choice_belongs_to_user_providers?(application_choice: application_choice)
+      application_choice_visible_to_user?(application_choice: application_choice)
     end
   end
 
   def can_change_offer?(application_choice:, course_option_id:)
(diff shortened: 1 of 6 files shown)


app/services/provider_authorisation.rb
```diff
@@ -10,17 +10,17 @@ def can_make_offer?(application_choice:, course_option_id: nil)
 
     supplied_course_option = CourseOption.find(course_option_id) if course_option_id
     if supplied_course_option && course_option_id != application_choice.course_option.id
-      application_choice_belongs_to_user_providers?(application_choice: application_choice) && \
+      application_choice_visible_to_user?(application_choice: application_choice) && \
         course_option_belongs_to_user_providers?(course_option: supplied_course_option)
```
tijmenb, line 14: Should this get a similar treatment? I'm assuming that `supplied_course_option` is there if the offered course is different?

duncanjbrown: We don’t know yet. The offer-a-different course UI is only populated with courses from directly associated providers. In theory an API user could attempt this. Feels like a separate piece of work?

PR #1858, opened 2020-04-09
Check if a RG exists before finding last deployment

vigneshmsft:
## Guidance to review

Script changes to check if a RG exists in the specified Azure subscription before proceeding to do other checks like finding last deployment on that RG.
The pipeline proceeds to do a full deployment if a RG doesn't already exist.

## Link to Trello card

https://trello.com/c/A3Am3PcM/1897-bug-creating-new-azure-environment-fails-because-resource-group-doesnt-exist

## Things to check

### DIFF
--- a/azure-pipelines-deploy-template.yml
+++ b/azure-pipelines-deploy-template.yml
@@ -102,75 +102,77 @@ jobs:
 
                 #Check if RG exists
                 Get-AzResourceGroup -Name $resourceGroupName -ErrorVariable rgDoesntExist -ErrorAction SilentlyContinue
-                if($rgDoesntExist){
-                  $fullDeployment = $true
-                  Write-Host "##vso[task.setvariable variable=fullDeployment]$fullDeployment"
-                  return
-                }
-
-                #Get array of deployment files, which if changed, should trigger a full ARM deployment.
-                #Add pipeline files to array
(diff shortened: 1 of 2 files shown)


azure-pipelines-deploy-template.yml
```diff
@@ -100,6 +100,14 @@ jobs:
                 $encodedAccessToken = [System.Convert]::ToBase64String([System.Text.Encoding]::UTF8.GetBytes($env:SYSTEM_ACCESSTOKEN))
                 $variableGroupLastEdit = @()
 
+                #Check if RG exists
+                Get-AzResourceGroup -Name $resourceGroupName -ErrorVariable rgDoesntExist -ErrorAction SilentlyContinue
+                if($rgDoesntExist){
+                  $fullDeployment = $true
+                  Write-Host "##vso[task.setvariable variable=fullDeployment]$fullDeployment"
```
tpinney, line 107: I'd delete this line and the following one and then enclose all of the following code in an `else`, except for the last line.
I'm not a huge fan of exiting from a script halfway through if I can avoid it, but sometimes it is unavoidable. In this case it avoids unnecessary code duplication.

vigneshmsft: changed as suggested 👍

PR #1862, opened 2020-04-09
Attempt to fix a flakey system spec for courses changes

stevehook:
## Context

Looks like this is failing when there are duplicate course names generated by the factories. e.g. https://dfe-ssp.visualstudio.com/Become-A-Teacher/_build/results?buildId=57595&view=ms.vss-test-web.build-test-results-tab&runId=1215328&resultId=101138&paneView=debug

- [x] Do content matches on the course name and code so that we get an exact match.

## Guidance to review

- Is there a better way?

## Link to Trello card

n/a

## Things to check


duncanjbrown:
I think we might have collided on #1859!

stevehook:
> I think we might have collided on #1859!

ha, yes I should have rebased again! Closing this one then...

PR #1864, opened 2020-04-14
Candidate can cancel reference request

davidgisbey:
## Context

During the 7 day editing period a candidate should be able to withdraw a reference request.

When a candidate removes a reference request thy are prompted to add another referee using the additional references flow. The withdrawn reference moves to a cancelled state but still shows in the UI.

- Adds a delete reference link to the completed application form page
- When a user cancels a reference request they are redirected to the add additional referee type page

Before 

image

image

After

image

image

image

## Guidance to review

Lots of this needs content review. Should the complete page display an add referee button? Or should it just link to the references review page?

## Link to Trello card

https://trello.com/c/hOgpOk8A/1280-dev-user-can-cancel-reference-request

## Things to check

### DIFF
--- a/app/components/candidate_interface/referees_review_component.html.erb
+++ b/app/components/candidate_interface/referees_review_component.html.erb
@@ -13,7 +13,7 @@
     <%= render(SummaryCardHeaderComponent.new(title: t('application_form.referees.nth_referee')[referee.ordinal], heading_level: @heading_level)) do %>
       <% if FeatureFlag.active?('candidate_cancels_reference') %>
         <div class="app-summary-card__actions">
-          <%= link_to candidate_interface_destroy_referee_path(referee.id), class: 'govuk-link' do %>
+          <%= link_to candidate_interface_confirm_destroy_referee_path(referee.id), class: 'govuk-link' do %>
             <%= t('application_form.referees.delete') %>
             <span class="govuk-visually-hidden"> <%= referee.name %></span>
           <% end %>
--- a/app/components/candidate_interface/referees_review_component.html.erb
+++ b/app/components/candidate_interface/referees_review_component.html.erb
@@ -11,7 +11,7 @@
 <% @application_form.application_references.includes(:application_form).each do |referee| %>
   <%= render(SummaryCardComponent.new(rows: referee_rows(referee), editable: @editable)) do %>
     <%= render(SummaryCardHeaderComponent.new(title: t('application_form.referees.nth_referee')[referee.ordinal], heading_level: @heading_level)) do %>
-      <% if @editable %>
+      <% if FeatureFlag.active?('candidate_cancels_reference') %>
         <div class="app-summary-card__actions">
           <%= link_to candidate_interface_destroy_referee_path(referee.id), class: 'govuk-link' do %>
             <%= t('application_form.referees.delete') %>
--- a/app/components/candidate_interface/referees_review_component.html.erb
+++ b/app/components/candidate_interface/referees_review_component.html.erb
@@ -1,5 +1,5 @@
 <p class="govuk-body">
-  <% if @application_form.submitted? && @application_form.application_references.count < 2 && FeatureFlag.active?('candidate_cancels_reference') %>
+  <% if @application_form.submitted? && @application_form.references_where_feedback_requested_or_feedback_provided.count < 2 && FeatureFlag.active?('candidate_cancels_reference') %>
     <%= t('application_form.referees.info.after_submission_with_replacement_needed') %>
   <% elsif @application_form.submitted?  %>
     <%= t('application_form.referees.info.after_submission') %>
@@ -11,7 +11,7 @@
 <% @application_form.application_references.includes(:application_form).each do |referee| %>
   <%= render(SummaryCardComponent.new(rows: referee_rows(referee), editable: @editable)) do %>
     <%= render(SummaryCardHeaderComponent.new(title: t('application_form.referees.nth_referee')[referee.ordinal], heading_level: @heading_level)) do %>
-      <% if FeatureFlag.active?('candidate_cancels_reference') %>
+      <% if FeatureFlag.active?('candidate_cancels_reference') || @editable %>
         <div class="app-summary-card__actions">
           <%= link_to candidate_interface_confirm_destroy_referee_path(referee.id), class: 'govuk-link' do %>
             <%= t('application_form.referees.delete') %>
--- a/app/components/candidate_interface/referees_review_component.html.erb
+++ b/app/components/candidate_interface/referees_review_component.html.erb
@@ -11,7 +11,7 @@
 <% @application_form.application_references.includes(:application_form).each do |referee| %>
   <%= render(SummaryCardComponent.new(rows: referee_rows(referee), editable: @editable)) do %>
     <%= render(SummaryCardHeaderComponent.new(title: t('application_form.referees.nth_referee')[referee.ordinal], heading_level: @heading_level)) do %>
-      <% if FeatureFlag.active?('candidate_cancels_reference') %>
+      <% if @editable %>
         <div class="app-summary-card__actions">
           <%= link_to candidate_interface_destroy_referee_path(referee.id), class: 'govuk-link' do %>
             <%= t('application_form.referees.delete') %>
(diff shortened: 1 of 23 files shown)


app/services/feature_flag.rb
```diff
@@ -3,6 +3,7 @@ class FeatureFlag
     add_additional_courses_page
     banner_about_problems_with_dfe_sign_in
     banner_for_ucas_downtime
+    candidate_cancels_reference
```
tijmenb, line 6: `candidate_can_cancel_reference`?

config/routes.rb
```diff
@@ -267,6 +267,7 @@
         patch '/update/:id' => 'referees#update', as: :update_referee
 
         get '/delete/:id' => 'referees#confirm_destroy', as: :confirm_destroy_referee
+        patch '/delete/:id' => 'referees#cancel_referee_request', as: :cancel_referee_request
```
tijmenb, line 270: Can we make this use a different URL just to be safe?

app/controllers/candidate_interface/referees_controller.rb
```diff
@@ -115,8 +125,8 @@ def referee_params
         .transform_values(&:strip)
     end
 
-    def redirect_to_review_referees_if_amendable
-      redirect_to candidate_interface_review_referees_path if current_application.amendable?
+    def redirect_to_review_referees_if_amendable_and_two_referees_are_present
+      redirect_to candidate_interface_review_referees_path if current_application.amendable? && current_application.application_references.count == ApplicationForm::MINIMUM_COMPLETE_REFERENCES
     end
```
tijmenb, line 130: I don't think this change is necessary because it won't do anything, `amendable?` is always false because the `edit_application` feature flag is permanently disabled.

app/views/candidate_interface/referees/review.html.erb
```diff
@@ -6,18 +6,24 @@
 </h1>
 
 <% editable = @application_form.amendable? ? false : true %>
```
tijmenb, line 8: From IRL convo: we don't need edits to this page because it's only visible when the form is not submitted yet.

spec/system/candidate_interface/candidate_cancels_a_reference_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+
+RSpec.describe 'Cancelling a reference' do
+  include CandidateHelper
+
+  scenario 'candidate cancels a reference request after completing their application' do
+    given_the_training_with_a_disability_flag_is_active
+    and_i_have_completed_and_submitted_my_application
+    and_the_candidate_cancels_reference_flag_is_active
+    and_the_referee_type_flag_is_active
+    and_edit_application_is_active
```
tijmenb, line 11: This is always false.

app/views/referee_mailer/reference_cancelled_email.text.erb
```diff
@@ -0,0 +1,5 @@
+Dear <%= @name %>
+
+#  <%= @candidate_name %> has cancelled their request for a reference from you.
```
davidgisbey, line 3: needs a content review

app/views/candidate_interface/referees/confirm_cancel.html.erb
```diff
@@ -0,0 +1,25 @@
...
+      <h1 class="govuk-heading-xl">
+        <span class="govuk-caption-xl"><%= @referee.name %></span>
+        <%= t('page_titles.referee_cancel') %>
+      </h1>
+
+      <p class='govuk-body'>You have <%= formatted_days_remaining %> (until <%= edit_by_date %>) to edit your application or change your referee.</p>
+
+      <p class='govuk-body'> You will need to provide another referee as we can't send your application to your teacher training providers without 2 complete references.</p>
+
+      <p class='govuk-body'> <% @referee.name %> will be informed that they do not need to provide a reference.</p>
```
davidgisbey, line 16: needs to be content reviewed

davidgisbey:
Have actioned all the feedback and stopped out the code that would only affect a user if the edit application flag was on. A few things need a content review.

app/components/candidate_interface/referees_review_component.html.erb
```diff
@@ -9,9 +9,17 @@
 <% @application_form.application_references.includes(:application_form).each do |referee| %>
   <%= render(SummaryCardComponent.new(rows: referee_rows(referee), editable: @editable)) do %>
     <%= render(SummaryCardHeaderComponent.new(title: t('application_form.referees.nth_referee')[referee.ordinal], heading_level: @heading_level)) do %>
-      <% if @editable %>
+
+      <% if FeatureFlag.active?('candidate_can_cancel_reference') && referee.feedback_status != 'cancelled' %>
```
tijmenb, line 13: This needs some thought - I think we only want to allow candidates to cancel references in very specific circumstances. This will show the cancel link when the application isn't submitted yet, and also when the feedback already has been provided and the application has been sent to the provider.

I *think* the logic should be that candidates can only cancel a reference that's in the `feedback_requested` state?

We also need to make sure we check the state in the controller before showing the confirmation / cancelling the reference for real.

app/controllers/candidate_interface/referees_controller.rb
```diff
@@ -70,10 +70,22 @@ def update
       end
     end
 
+    def confirm_cancel
+      @application_form = current_application
+    end
+
+    def cancel
+      @referee.update!(feedback_status: 'cancelled')
+      RefereeMailer.reference_cancelled_email(current_application, @referee).deliver_later
```
tijmenb, line 79: Can we send a Slack message to support? It'll be interesting to know how much this feature is used.

app/components/candidate_interface/referees_review_component.html.erb
davidgisbey: Good point. I've redirected them to the referee review page if it's not in the correct state. If they've submitted the before action will redirect them to the complete page.

app/controllers/candidate_interface/referees_controller.rb
```diff
@@ -118,5 +140,12 @@ def referee_params
     def redirect_to_review_referees_if_amendable
       redirect_to candidate_interface_review_referees_path if current_application.amendable?
     end
+
+    def send_slack_message
+      message = "A referee has cancelled their request for a reference from #{@referee.name}."
```
tijmenb, line 145: We shouldn't use full names/other PII in Slack messages, perhaps this?

```suggestion
      message = "Candidate #{candidate_first_name} has cancelled one of their references"
```

app/controllers/candidate_interface/referees_controller.rb
```diff
@@ -70,10 +70,32 @@ def update
...
+        redirect_to candidate_interface_review_referees_path
+      end
+    end
+
+    def cancel
+      if @referee.feedback_requested?
+        @referee.update!(feedback_status: 'cancelled')
+        RefereeMailer.reference_cancelled_email(current_application, @referee).deliver_later
+
+        send_slack_message
```
tijmenb, line 86: How about pulling these three lines into a service?

app/mailers/referee_mailer.rb
```diff
@@ -38,6 +38,17 @@ def reference_confirmation_email(application_form, reference)
     )
   end
 
+  def reference_cancelled_email(application_form, reference)
```
tijmenb, line 41: Could we use `reference.application_form` instead here instead of passing in `application_form`?

spec/system/candidate_interface/candidate_cancels_a_reference_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    given_the_training_with_a_disability_flag_is_active
+    and_i_have_completed_and_submitted_my_application
+    and_the_candidate_can_cancel_reference_flag_is_active
+    and_the_referee_type_flag_is_active
+
+    when_i_visit_the_application_complete_page
+    and_i_click_delete_on_my_first_reference
+    then_i_should_see_the_confirm_cancel_page
+
+    when_i_click_confirm_delete
```
tijmenb, line 16:
```suggestion
    when_i_click_to_confirm_the_cancellation
```

EmmaFrith:
On the first email @davidgisbey, we could go for: 

**Are you sure you want to cancel this request for a reference?**

You will need to give details of a new referee. 

We cannot send your application to your provider(s) without 2 references. 

We’ll tell ((referee name)) that they do not need to give reference. 


_I think we can remove the line 'you have 2 days to edit your application or change your referee' - we don't know how quickly the referee is going to respond - so can't say they've got x number or days right?_

PR #1866, opened 2020-04-14
Support users can grant permissions to provider users

steventux:
## Context

We have added a MVP of provider user permissions. We should allow support users to manage these.

- Splits `InviteProviderUser` into dual concerns of saving + updating permissions and inviting in separate services
- Introduces a service to call both of the above and handle errors
- Nests permissions per provider available to the provider user

image

## Link to Trello card

https://trello.com/c/KgRyOExk/1842-supportusers-can-grant-permissions-to-providerusers

## Things to check

### DIFF
--- a/app/models/support_interface/provider_user_form.rb
+++ b/app/models/support_interface/provider_user_form.rb
@@ -50,6 +50,7 @@ module SupportInterface
         last_name: provider_user.last_name,
         email_address: provider_user.email_address,
         provider_ids: provider_user.provider_ids,
+        manage_users: provider_user.can_manage_users?,
       )
     end
 
--- a/app/models/support_interface/provider_user_form.rb
+++ b/app/models/support_interface/provider_user_form.rb
@@ -24,7 +24,11 @@ module SupportInterface
     end
 
     def save
-      @provider_user.save! if build
+      if build
+        @provider_user.save!
+        assign_manage_users_permissions if manage_users
+        @provider_user
+      end
     end
 
     def email_address=(raw_email_address)
@@ -63,6 +67,13 @@ module SupportInterface
       return unless ProviderUser.exists?(email_address: email_address)
 
       errors.add(:email_address, 'This email address is already in use')
+    end
+
+    def assign_manage_users_permissions
+      ProviderPermissions.where(
+        provider_user_id: provider_user.id,
+        provider_id: provider_ids,
+      ).update_all(manage_users: true)
     end
   end
 end
--- a/app/models/support_interface/provider_user_form.rb
+++ b/app/models/support_interface/provider_user_form.rb
@@ -47,7 +47,7 @@ module SupportInterface
         last_name: provider_user.last_name,
         email_address: provider_user.email_address,
         provider_ids: provider_user.provider_ids,
-        permissions: permissions(provider_user),
+        permissions: permissions_for(provider_user),
       )
     end
 
@@ -57,10 +57,10 @@ module SupportInterface
       @provider_ids.reject(&:blank?)
     end
 
-    def self.permissions(provider_user)
-      {
-        manage_users: ProviderPermissions.manage_users.where(provider_user: provider_user).pluck(:provider_id),
-      }
+    def self.permissions_for(provider_user)
+      provider_permissions = ProviderPermissions.manage_users.where(provider_user: provider_user)
+
+      OpenStruct.new({ manage_users: provider_permissions.pluck(:provider_id) })
     end
 
   private
--- a/app/models/support_interface/provider_user_form.rb
+++ b/app/models/support_interface/provider_user_form.rb
@@ -60,7 +60,7 @@ module SupportInterface
     def self.permissions_for(provider_user)
       provider_permissions = ProviderPermissions.manage_users.where(provider_user: provider_user)
 
-      OpenStruct.new({ manage_users: provider_permissions.pluck(:provider_id) })
+      ProviderPermissionsOptions.new(manage_users: provider_permissions.pluck(:provider_id))
     end
 
   private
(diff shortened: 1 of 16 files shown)


spec/services/save_and_invite_provider_user_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+      email_address: 'test+invite_provider_user@example.com',
+      first_name: 'Firstname',
+      last_name: 'Lastname',
+      provider_ids: [provider.id],
+    )
+  }
+  let(:permissions) { { manage_users: [provider.id] } }
+  let(:provider_user) { form.build }
+  let(:save_service) { SaveProviderUser.new(provider_user: provider_user, permissions: permissions) }
+  let(:invite_service) { InviteProviderUser.new(provider_user: provider_user) }
```
duncanjbrown, line 16: If we're making these injectable, might we as well inject stubs implementing `save!`? Slightly less coupling

spec/services/save_and_invite_provider_user_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    context 'an error occurs in invite service' do
+      before { allow(invite_service).to receive(:call!).and_raise }
+
+      it 'rolls back the transaction and raises the error' do
+        expect { service.call }.to raise_error.and change(ProviderUser, :count).by(0)
+      end
+    end
+
+    context 'a DfeSignInApiError occurs' do
```
duncanjbrown, line 53: Should this object know about HTTP? The stub could raise on `save!`

spec/models/support_interface/provider_user_form_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      permissions = described_class.permissions_for(provider_user)
+
+      expect(permissions).to be_a(OpenStruct)
+      expect(permissions.manage_users).to eq(provider_user.providers.map(&:id))
+    end
+  end
+
+  describe '#save' do
+    context 'with invalid params' do
+      it 'returns nil' do
```
duncanjbrown, line 50: nit: should it return false, for ActiveRecord consistency? https://apidock.com/rails/ActiveRecord/Base/save

app/models/support_interface/provider_user_form.rb
```diff
@@ -55,6 +57,12 @@ def provider_ids
       @provider_ids.reject(&:blank?)
     end
 
+    def self.permissions_for(provider_user)
+      provider_permissions = ProviderPermissions.manage_users.where(provider_user: provider_user)
+
+      OpenStruct.new({ manage_users: provider_permissions.pluck(:provider_id) })
```
duncanjbrown, line 63: As a matter of taste, how would you feel about making this a class-level constant so it has a name outside the method, and a plain `Struct` so it's clearly the whole definition? eg `ProviderPermissionsOptions = Struct.new(:manage_users)`. I guess the downside of plain `Struct` is that struct arguments are positional so this might get less readable when further from the definition. Maybe an `OpenStruct` as a constant?

steventux: I like the `Struct` route as `Struct#members` gives us the valid permissions params keys for free.

duncanjbrown: Nice, I didn't know about that!

spec/services/save_and_invite_provider_user_spec.rb
steventux: :+1: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1866/commits/2819a377045a5bb37eb6d96ca0270ede0e0690dc

spec/services/save_and_invite_provider_user_spec.rb
steventux: :+1: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1866/commits/2819a377045a5bb37eb6d96ca0270ede0e0690dc

spec/models/support_interface/provider_user_form_spec.rb
steventux: I think this method is redundant, it was only ever called in the `SupportInterface::ProviderUserController#update` action and we now use the `SaveProviderUser` service to persist the user on update so https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1866/commits/a32e6da6908932a3dc9235c34597ea8ef90d7041

app/services/save_provider_user.rb
```diff
@@ -0,0 +1,36 @@
...
+    @provider_user = provider_user
+    @permissions = sanitized_permissions(permissions)
+  end
+
+  def call!
+    @provider_user.save!
+    update_permissions
+    @provider_user.reload
+  end
+
```
duncanjbrown, line 12:
```suggestion

private

```

spec/services/save_provider_user_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  describe '#call!' do
+    subject(:service) do
+      described_class.new(provider_user: new_provider_user_from_form, permissions: permissions)
+    end
+
+    it 'saves the provider user' do
+      expect { service.call! }.to change(ProviderUser, :count).by(1)
+    end
+
+    it 'updates permissions for the saved user' do
```
duncanjbrown, line 31: Is it worth adding a spec for the `sanitize_permissions` behaviour here? ie if I try to assign a permission for a provider I don't belong to, it shouldn't work

steventux: :+1: Thanks @duncanjbrown adding this to the spec helped me spot some unintentional behaviour where an invalid provider id would stop any permissions being set for the given key.
I've changed the method to filter out unassigned provider ids and retain valid ones.

PR #1872, opened 2020-04-15
Remove SUPPORT_USERNAME and SUPPORT_PASSWORD ENV vars

duncanjbrown:
These ENV vars are unused because Support access is authenticated via DfE Sign-in now.

We also still have ENV vars for basic auth on the candidate interface, which I think are superseded by the `pilot_open` flag. We should remove those too.

Once this is deployed we can remove these vars from each env in Azure too.

https://dfe-ssp.visualstudio.com/Become-A-Teacher/_library?itemType=VariableGroups

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -1,4 +1,2 @@
 FIND_BASE_URL=https://api2.publish-teacher-training-courses.service.gov.uk/api/v3/
-SUPPORT_USERNAME=test
-SUPPORT_PASSWORD=test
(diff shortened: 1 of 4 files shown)


duncanjbrown:
Ah, thank you! I didn't understand `ag`. Now I understand it a bit better. https://github.com/duncanjbrown/dotfiles/commit/d056166873a83cd3377b80d1fdc4f7ee7bfaf61c

PR #1873, opened 2020-04-15
Tweak logstash buffer settings to prevent full buffers

mnacos:
## Context

During my investigation into Logit log loss, I observed a few cases of `Buffer Full` in the logs. While this is not the main cause of log loss (this seems to be TCP-related), we should probably increase the size of the buffer. 

This PR raises `buffer_max_items` to 200 (we've been using the default of 50) and drops `buffer_max_interval` to 2 (we've been using the default of 5 seconds), to reduce the possibility of a full buffer.

## Guidance to review

It's only a configuration change.

## Link to Trello card

https://trello.com/c/e1YIWE8s

## Things to check


duncanjbrown:
Seems worth a try

PR #1876, opened 2020-04-15
Add 'Dataset 1' export for UCAS to support UI

aldavidson:
## Context

To temporarily support dual-running of Apply & UCAS, we have agreed to send UCAS a dataset ('Dataset 1' in the Data Sharing Agreement) which they will use to match candidates and applications against their systems. They will then send us back a file ('Dataset 2' in the DSA) which indicates any candidates who have applied on both systems. 

(NOTE: please merge #1882 first)

This PR adds a service object to generate CSV in the format specified in the DSA, and links to call it from the downloads section of the support interface. This allows us to easily generate example files on-demand so we can send them to UCAS to ask 'hey, how does _this_ one look?' etc.

Screen Shot 2020-04-15 at 14 14 56

Eventually this will be run automatically each night, at which point the service object which does the work can simply be called from a worker on a schedule.

## Guidance to review

1. Go to /support & log in
2. Click 'performance' in the top nav, scroll down to the bottom
3. Click 'Download Dataset 1 (CSV)'

This should download a .csv file, openable in the spreadsheet of your choice, containing all submitted (or later) applications, in the format defined in the [Data Sharing Agreement](https://docs.google.com/document/d/1TCESfSIBGBa3JqhNqRa1df-ioO6Hzk44IywgvAHmvTg/edit?usp=sharing)

Also note, I have added the import of course subjects from the Find API, when we run `SyncProviderFromFind`. To prevent us having to hit the Find API many times every time this runs (every 15 mins), it now caches this info in a JSONB field on the course record. This can be generalised into a full model later if needed. 

## Link to Trello card

[DEV: (Tactical) data matching to stop candidates making more than 6 applications](https://trello.com/c/ybZbTUR6/388-dev-tactical-data-matching-to-stop-candidates-making-more-than-6-applications)

## Things to check

- [x] This code doesn't rely on migrations in the same Pull Request - the migrations are in PR #1882 


tvararu:
Could you update the PR title for this? 👍

aldavidson:
> Could you update the PR title for this? 👍

How's this look now?

app/services/sync_provider_from_find.rb
```diff
@@ -128,6 +128,8 @@ def assign_course_attributes_from_find(course, find_course)
     course.course_length = find_course.course_length
     course.recruitment_cycle_year = find_course.recruitment_cycle_year
     course.exposed_in_find = find_course.findable?
+    course.subjects = find_course.subjects
```
tijmenb, line 131: Could we extract the subject codes here, so we don't have to get them out in `Course#subject_codes`? I'd like the database to be structured as possible.

app/views/support_interface/performance/index.html.erb
```diff
@@ -62,3 +62,16 @@
 <p class='govuk-body'>
   <%= govuk_link_to 'Download candidate survey results (CSV)', support_interface_candidate_survey_path %>
 </p>
+
+<h3 class='govuk-heading-m'>Applications export for UCAS</h3>
+
+<p class='govuk-body'>
```
tijmenb, line 68: 💯 for explaining this

app/views/support_interface/performance/index.html.erb
```diff
@@ -62,3 +62,16 @@
...
+  This generates a list of all submitted applications and choices in the form_updated_at
```
tijmenb, line 69:
```suggestion
  This generates a list of all submitted applications and choices in the form
```

spec/services/personal_details_controller.rb
```diff
@@ -0,0 +1,52 @@
+module CandidateInterface
```
tijmenb, line 1: This shouldn't be part of the diff as well, odd.

spec/services/application_choices_export_spec.rb
```diff
@@ -0,0 +1,123 @@
+require 'rails_helper'
```
tijmenb, line 1: How is this file part of the diff? It's already on master 🤔

app/services/sync_provider_from_find.rb
aldavidson: Sure - do you mean _instead of_ storing the whole structure, or _as well as_ ? 

In a previous chat with @tvararu we thought we would probably need to refer to the subjects more in future, so storing the whole structure would give us more options with not much overhead. But if you'd rather only store what we actually need right now, I'm fine with that too.

tijmenb: Yes, instead of. I'd like to store what we need, and avoid coupling our code to Find. The sync should ideally translate Find's data into something that's useful for us, so you don't have to understand the Find data model to understand how Apply works.

spec/services/application_choices_export_spec.rb
aldavidson: I had to rename it from `spec/services/applications_export_spec.rb` to `spec/services/application_choices_export_spec.rb` in order to get a clean Rubocop run

app/services/sync_provider_from_find.rb
aldavidson: Sure - i've updated the migration to just store `subject_codes` now, and pulled it out to a separate PR( #1882 )

db/schema.rb
```diff
@@ -209,6 +209,8 @@
     t.string "course_length"
     t.string "description"
     t.integer "accredited_provider_id"
+    t.jsonb "subjects"
+    t.string "funding_type"
```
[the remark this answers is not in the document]
aldavidson: done - #1882

aldavidson:
@tijmenb @tvararu ready for re-review, I reckon

app/models/find_api/provider.rb
```diff
@@ -4,6 +4,14 @@ class Provider < FindAPI::Resource
     has_many :courses
     has_many :sites
 
+    # There's a quirk in the JsonApiClient that means we have to do some
+    # counter-intuitive things with our resource models to get it to work.
+    # In this case, to get included course subjects to work, we have to define
+    # this as an inner class of Provider, despite it actually being a has-has_many
+    # of Course. Don't know why - if anyone can figure out a more
+    # elegant way, feel free.
+    class Subject < FindAPI::Resource; end
```
tvararu, line 13: 👍 for this comment.

app/services/applications_export_for_ucas.rb
```diff
@@ -0,0 +1,86 @@
...
+      nctl_subject: subject_codes(application_choice.course),
+      course_name: application_choice.course.name,
+      course_code: application_choice.course.code,
+    }
+  end
+
+  def subject_codes(course)
+    course.subject_codes.to_a.join('|')
+  end
+end
```
tvararu, line 86: This whole file is very clean 👍

app/models/find_api/course.rb
```diff
@@ -15,6 +15,12 @@ def name_and_code
       "#{name} (#{code})"
     end
 
+    def subject_codes
+      subjects.to_a.map do |subject|
+        subject.try(:attributes).try(:[], 'subject_code')
```
tijmenb, line 20: Would this work?

```suggestion
        subject.dig('attributes', 'subject_code')
```

app/services/applications_export_for_ucas.rb
```diff
@@ -0,0 +1,86 @@
...
+      .includes(
+        :candidate,
+        :application_choices,
+        :application_qualifications,
+        :application_work_experiences,
+        :application_references,
+        application_choices: %i[course provider],
+      )
+      .where('candidates.hide_in_reporting' => false)
+      .where.not(submitted_at: nil)
```
tijmenb, line 54: I suspect that we only should export applications that have been sent to the provider? You would want to scope to not being in `STATES_NOT_VISIBLE_TO_PROVIDER` instead.

db/migrate/20200416094721_add_course_subject_codes_and_funding_type.rb
```diff
@@ -0,0 +1,8 @@
+class AddCourseSubjectCodesAndFundingType < ActiveRecord::Migration[6.0]
```
tijmenb, line 1: This goes away after a rebase I think?

spec/factories.rb
```diff
@@ -227,6 +227,9 @@
     end
   end
 
+  factory :subject do
+  end
```
tijmenb, line 231: Do we need this?

app/models/find_api/course.rb
aldavidson: sadly not, subject is a `FindAPI::Provider::Subject` that doesn't respond to `dig`

app/services/applications_export_for_ucas.rb
aldavidson: Hmm, not sure - all the discussions I've been privy to have been around submitted(-or-further-on) applications. 
I think as the purpose of doing this is to detect duplicates and take preventive action, isn't it the case that if we candetect and correct a duplicate _before_ it goes to the provider, that's actually _better_ than after it's sent?

spec/factories.rb
aldavidson: as it turns out - no 👍

app/services/applications_export_for_ucas.rb
tijmenb: Yeah, that makes sense! Disregard comment.

db/migrate/20200416094721_add_course_subject_codes_and_funding_type.rb
aldavidson: it does indeed

PR #1877, opened 2020-04-15
Improve Slack message when candidates sign up

tijmenb:
## Context

The signup messages currently look like this:

Screenshot 2020-04-15 at 17 29 04

Make them look like:

> ✨ The 982th candidate just signed up

This is more typographically pleasing, provides a visual queue that about the type of message and uses a meaningful number in the message (so we can celebrate the 1000th real sign up).

## Guidance to review

However you want.

## Link to Trello card

https://trello.com/c/mGLdC4yg/185-post-new-sign-ups-to-slack-every-10th-candidate

## Things to check

### DIFF
--- a/app/lib/state_change_notifier.rb
+++ b/app/lib/state_change_notifier.rb
@@ -1,5 +1,13 @@
 class StateChangeNotifier
-  def self.call(event, candidate: nil, application_choice: nil, application_form: nil)
+  def self.sign_up(candidate)
+    helpers = Rails.application.routes.url_helpers
+    text = "New sign-up [candidate_id: #{candidate.id}]"
+    url = helpers.support_interface_candidate_url(candidate)
+
+    send(text, url)
+  end
+
+  def self.call(event, application_choice: nil, application_form: nil)
     helpers = Rails.application.routes.url_helpers
 
     if application_choice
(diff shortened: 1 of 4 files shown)


app/lib/state_change_notifier.rb
```diff
@@ -1,5 +1,14 @@
 class StateChangeNotifier
-  def self.call(event, candidate: nil, application_choice: nil, application_form: nil)
+  def self.sign_up(candidate)
+    helpers = Rails.application.routes.url_helpers
+    candidate_number = Candidate.where(hide_in_reporting: false).count
+    text = ":sparkles: Candidate ##{candidate_number} just signed up"
```
mnacos, line 5:
```suggestion
    text = ":sparkles: #{candidate_number.ordinalize} sign up"
```
What you've suggested is definitely an improvement, but developers may visually interpret hash sign + number as the active record id. Using `ordinalize` could address this, as it makes it look less like an id.

tvararu: Maybe? e.g. `The 987th candidate has signed up`

```ruby
text = ":sparkles: The #{candidate_number.ordinalize} candidate has signed up"
```

tijmenb:
@mnacos @tvararu updated!

PR #1878, opened 2020-04-15
Add 'notes' feature flag; move Timeline to its own tab

mnacos:
## Context

In preparation for adding the 'Provider user adds notes against an application choice' flow, we need to move the Timeline to its own tab/view and add a `notes` feature flag. 

This PR refactors `provider_interface/application_choices_controller.rb` and associated views to use:
- a separate view for the timeline component,
- a common `_application_choice_header.html` partial, for banners, status tags and newly introduced `SubNavigationComponent`, and
- a controller `before_action` setting `@application_choice` and `@sub_navigation_items` 

Page title and breadcrumbs are still set within the corresponding views. The `before_action` checks `notes` and `timeline` feature flags, so it's possible for the single application view to look like this:

image

## Guidance to review

- Should I have created a new `notes_controller.rb`? I thought about it, but notes will only be attached to `application_choices` and it's easier to reuse controller code and partials across the three views/tabs ([Application, Notes, Timeline]) if they are all part of `application_choices_controller`. Furthermore, if we decide to attach notes to other things, chances are the relevant views will be custom to these contexts, so a scoped pattern like `.../application_choices/1970/notes` makes sense.
- Do the adjusted page titles and breadcrumbs make sense? I've kept the Application title and breadcrumbs the same, but 'Notes' or 'Timeline' are appended to both title and breadcrumbs when a provider user navigates to these views (see image).

image

## Link to Trello card

https://trello.com/c/mscEDZr5

## Things to check

### DIFF
--- a/app/controllers/provider_interface/application_choices_controller.rb
+++ b/app/controllers/provider_interface/application_choices_controller.rb
@@ -1,5 +1,7 @@
 module ProviderInterface
   class ApplicationChoicesController < ProviderInterfaceController
+    before_action :set_application_choice_and_sub_navigation_items, except: %i[index]
+
     def index
       @page_state = ProviderApplicationsPageState.new(
         params: params,
@@ -21,12 +23,6 @@ module ProviderInterface
     end
 
     def show
-      available_providers = current_provider_user.providers
-
-      @application_choice = GetApplicationChoicesForProviders.call(
-        providers: available_providers,
-      ).find(params[:application_choice_id])
-
       @status_box_options = if @application_choice.offer?
                               GetAllChangeOptionsFromOfferedOption.new(
                                 application_choice: @application_choice,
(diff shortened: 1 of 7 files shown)


adamsilver:
@mnacos looks like the alignment is off between the contents of timeline tab and the timeline content itself.

app/views/provider_interface/application_choices/_application_choice_header.html.erb
```diff
@@ -0,0 +1,16 @@
...
+    </div>
+  </div>
+<% end -%>
+
+<h1 class="govuk-heading-xl govuk-!-margin-bottom-1">
+  <%= @application_choice.application_form.full_name %>
+  <%= render(ProviderInterface::ApplicationStatusTagComponent.new(application_choice: @application_choice)) %>
+</h1>
+
+<%= render SubNavigationComponent.new(items: @sub_navigation_items) %>
```
duncanjbrown, line 16: Should we render this if `@sub_navigation_items.count == 1`?

app/views/provider_interface/application_choices/notes.html.erb
```diff
@@ -0,0 +1,21 @@
...
+      <li class="govuk-breadcrumbs__list-item" aria-current="page">
+        Notes
+      </li>
+    </ol>
+  </div>
+<% end %>
+
+<%= render 'application_choice_header' %>
+
+Notes
```
duncanjbrown, line 21: Shall we have nothing rather than unstyled text? Less bad if it's accidentally turned on

mnacos:
> @mnacos looks like the alignment is off between the contents of timeline tab and the timeline content itself. The timeline doesn't need to be wrapped in a govuk-grid-row :)

Fixed, thanks for spotting.

app/controllers/provider_interface/application_choices_controller.rb
```diff
@@ -1,5 +1,7 @@
 module ProviderInterface
   class ApplicationChoicesController < ProviderInterfaceController
+    before_action :set_application_choice_and_sub_navigation_items, except: %i[index]
```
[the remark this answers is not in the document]
mnacos: I agree but all actions scoped to an application choice will require `@application_choice` except `index` so I think it's fine this once.

app/views/provider_interface/application_choices/_application_choice_header.html.erb
mnacos: I'll add a `>1` conditional.

app/views/provider_interface/application_choices/notes.html.erb
mnacos: I think I'll keep the file there and just `raise` from the view.

PR #1880, opened 2020-04-16
Introduce the basis for changing courses in support

tijmenb:
## Context

Currently we have a feature in support to add a course to an application that has been submitted. We want to build a feature to allow support to also remove courses from someone's application.

This is trickier than it sounds at first. Removing a course means different things depending on how many courses the candidate has applied to:

1. If the candidate has applied to a single course, removing a course will end the application. They will have to apply again via Apply 2 (currently only available on UCAS).
2. If the candidate has applied to a single course but wants to change it to another, we would want add a new course first, and then remove the original one to avoid scenario #1 from kicking in.
3. If the candidate has applied to multiple courses, we can just remove the course from the application without any side effects.

To implement the feature we'll implement a flow that allows support users to make specific changes:

1. Add a course (only available when 1 or 2 choices have been made)
2. Replace a course (always available)
3. Withdraw from a course (only available when 2 or 3 choices have been made)
4. Withdraw from all courses and cancel application (always available)

This will allow us to easily implement the changes and handle the side effects like email notifications, Apply 2, and cancelling referees.

This PR takes the existing feature and moves it into the flow, so we can build the individual features on top of it.

Screens that have changed:

image

image

## Guidance to review

Particularly interested in feedback on the plan described in this PR.

## Link to Trello card

https://trello.com/c/fmTuhZ0y/1276-dev-support-user-can-withdraw-a-candidate-course-whole-application

## Things to check


app/controllers/support_interface/change_course_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+      case change_course_form.change_type
+      when 'add_course'
+        redirect_to support_interface_add_course_to_application_path(application_form)
+      else
+        flash[:info] = "Sorry - we haven't built this feature yet"
+        redirect_to support_interface_change_course_path(application_form)
+      end
+    end
+
+    def select_course_to_add
```
stevehook, line 19: I think this controller will get a bit big when all 4 options are implemented so perhaps executing the individual options would be easier to manage as separate controllers? Can be looked at later I guess.

spec/system/support_interface/adding_a_course_spec.rb
```diff
@@ -28,8 +29,13 @@ def when_i_visit_the_application_form
...
 
-  def and_click_on_the_button_to_add_a_course
-    click_on 'Add course'
+  def and_click_on_the_button_to_change_courses
+    click_on 'Make changes to courses'
+  end
+
+  def and_i_select_the_option_to_add_a_course
+    choose 'Add a course'
+    click_on 'Continue'
```
stevehook, line 38: :+1: having only minor changes to system spec is a good sign!

config/routes.rb
```diff
@@ -458,12 +458,15 @@
     get '/applications' => 'application_forms#index'
     get '/applications/action-required' => 'application_forms#action_required', as: :action_required
     get '/applications/:application_form_id' => 'application_forms#show', as: :application_form
-    get '/applications/:application_form_id/add-course' => 'application_forms#select_course_to_add', as: :add_course_to_application
-    post '/applications/:application_form_id/add-course' => 'application_forms#add_course'
     get '/applications/:application_form_id/audit' => 'application_forms#audit', as: :application_form_audit
     get '/applications/:application_form_id/comments/new' => 'application_forms/comments#new', as: :application_form_new_comment
     post '/applications/:application_form_id/comments' => 'application_forms/comments#create', as: :application_form_comments
 
+    get '/applications/:application_form_id/change-course' => 'change_course#options', as: :change_course
```
stevehook, line 465: Just wondering whether it makes sense to have a `ChangeCourseOptionsController` with `index` and `create` actions/routes rather these custom routes. Not sure whether that leads to a clearer solution?

app/controllers/support_interface/change_course_controller.rb
tijmenb: Good point. I think the individual options might become quite big indeed, so I'll keep an eye on that.

PR #1881, opened 2020-04-16
Stop candidates from applying to courses taken off apply

stevehook:
## Context

Candidates need to be prevented from applying to a course that is no longer available on Apply (but is on Find).

- [x] Add warning to review page so that candidate can see that a course has been closed on Apply before submitting.
- [x] Add validations to prevent applications to closed courses being submitted.

## Guidance to review

image

image 

## Link to Trello card

https://trello.com/c/diQ6M64v/1307-dev-stop-candidates-from-applying-to-courses-taken-off-apply

## Things to check


app/presenters/candidate_interface/application_form_presenter.rb
```diff
@@ -73,6 +73,13 @@ def application_choice_errors
             )
             next
           end
+
+          if choice.course_closed_on_apply?
+            error_list << ApplicationChoiceError.new(
+              choice.course_closed_on_apply_error, choice.id
+            )
+            next
```
malcolmbaig, line 81: I wonder if it's worth making this the 1st or 2nd check - I'm thinking that if a course is closed on Apply, its course-level or site-level vacancy status may remain unchanged, in which case moving this block further up saves us a couple of unnecessary SQL queries.

spec/system/candidate_interface/candidate_reviewing_application_for_course_unavailable_on_apply_spec.rb
```diff
@@ -0,0 +1,89 @@
+require 'rails_helper'
```
malcolmbaig, line 1: It might make more organisational sense to consolidate this spec with `spec/system/candidate_interface/candidate_reviewing_application_with_unavailable_course_options_spec.rb`, perhaps by adding a couple of steps which swap out one invalid course option for another and asserting the presence of the new warning.

PR #1882, opened 2020-04-16
add migrations to support #1876 (Add UCAS Dataset1 export to support UI)

aldavidson:
## Context

These are the migrations for PR #1876 - please merge this PR _first_, before that one 
(doing migrations in a separate PR like this prevents issues with the workers starting before the migrations have run through)

add two fields to courses - funding_type and subject_codes. Both to be populated from the Find API in PR #1876 

## Link to Trello card

[DEV: (Tactical) data matching to stop candidates making more than 6 applications](https://trello.com/c/ybZbTUR6/388-dev-tactical-data-matching-to-stop-candidates-making-more-than-6-applications)

## Things to check


db/schema.rb
```diff
@@ -210,6 +210,8 @@
     t.string "course_length"
     t.string "description"
     t.integer "accredited_provider_id"
+    t.jsonb "subject_codes"
+    t.string "funding_type"
```
tvararu, line 214: Is this different from `financial_support`?

davidgisbey: funding type is apprenticeship, salaried or fee. Financial support details bursaries etc.

tvararu: Thanks @davidgisbey!

PR #1883, opened 2020-04-16
Prevent support users from adding more than 3 courses

tijmenb:
## Context

Follow up to https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1880.

We can currently add more than 3 choices to an application, while this shouldn't be possible. This removes the option from the new page to make a change to the courses.

When a course already has 3 choices, the support user will see this:

image

## Guidance to review

Does it make sense? Any content suggestions?

## Link to Trello card

https://trello.com/c/fmTuhZ0y

## Things to check

### DIFF
--- a/app/controllers/support_interface/change_course_controller.rb
+++ b/app/controllers/support_interface/change_course_controller.rb
@@ -1,5 +1,7 @@
 module SupportInterface
   class ChangeCourseController < SupportInterfaceController
+    before_action :make_sure_course_can_be_added, only: %w[select_course_to_add add_course]
+
     def options
       @pick_option_form = SupportInterface::ChangeCourseForm.new(application_form: application_form)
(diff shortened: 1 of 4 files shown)


app/controllers/support_interface/change_course_controller.rb
```diff
@@ -1,5 +1,7 @@
 module SupportInterface
   class ChangeCourseController < SupportInterfaceController
+    before_action :make_sure_course_can_be_added, only: %w[select_course_to_add add_course]
```
tijmenb, line 3: Another point of evidence for @stevehook suggestion to split up the controllers https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1880#discussion_r409413573.

PR #1884, opened 2020-04-16
Allow support users to withdraw choices and cancel applications

tijmenb:
## Context

This is part of the feature to allow support users to withdraw an application on the candidate's behalf. It's part of a series started in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1880.

The big question was whether to introduce a new state for "cancelled" applications. The reason I think it might be good is that we can keep track of applications that have been modified since submission. The alternative is to just `destroy!` the ApplicationChoice. This will cause weirdness once we implement the feature to withdraw from *all* of the courses, we would have forms without any choices in the system, which still have been submitted. The downside to adding a new state is that we add a bunch of overhead.

Important changes:

- Add cancelled state, including documentation
- Add UI to withdraw a course from an application
- Add UI to cancel the entire application
- Make sure the references code can handle cancelled applications

Screenies:

# 📹 Watch the video: https://www.loom.com/share/e98760bd996c4c17a45555aab82e3e7e

## Guidance to review

- Does the code work?
- Is it consistent?

## Link to Trello card

https://trello.com/c/fmTuhZ0y

## Things to check


tvararu:
> Should we introduce a new state for "cancelled" applications

I say we should!

app/models/support_interface/cancel_course_form.rb
```diff
@@ -0,0 +1,16 @@
+module SupportInterface
+  class CancelCourseForm
```
duncanjbrown, line 2: If it's the last course, should this also notify referees their responses are no longer needed?

tijmenb: No, that's why this PR doesn't allow cancellation of the last course (https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1884/files#diff-84183967e9402039054b942451a960d4R8-R10). Perhaps I'll combine that feature in this PR as well, so we can look at it as a single flow.

duncanjbrown: Ah, great. I wonder if it's a separate thing, then, as it might need to be invoked from the candidate side as well if they withdraw from the last course?

app/controllers/support_interface/change_course_controller.rb
```diff
@@ -12,6 +12,10 @@ def pick_option
       case change_course_form.change_type
       when 'add_course'
         redirect_to support_interface_add_course_to_application_path(application_form)
+      when 'cancel_course'
+        redirect_to support_interface_cancel_course_path(application_form)
+      when 'cancel_application'
+        redirect_to support_interface_cancel_application_path(application_form)
```
tijmenb, line 18: "Cancel course" and "Cancel application" makes me think the terminology isn't correct. Anyone better ideas?

app/models/support_interface/cancel_course_form.rb
tijmenb: I've added the feature to cancel the entire application in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1884/commits/e085fc88b96d68bc1c127f1012aec3539f762a50. Do you mean to say you think it's a separate thing or not a separate thing?

app/controllers/support_interface/change_course_controller.rb
duncanjbrown: I guess it's `Cancel {Apply 1|Apply 2} application`? `cancel_application_for_current_phase` for the whole thing? `remove_course`?

app/models/support_interface/cancel_application_form.rb
```diff
@@ -0,0 +1,17 @@
+module SupportInterface
+  class CancelApplicationForm
+    include ActiveModel::Model
+
+    attr_accessor :application_form
+
+    def save!
+      application_form.application_choices.includes(:course_option).where(status: %w[awaiting_references application_complete]).find_each do |application_choice|
```
davidgisbey, line 8: Are we using the course option for anything?

tijmenb: I don't know! Bullet is complaining about this. I'll investigate.

app/controllers/support_interface/change_course_controller.rb
davidgisbey: +1 on remove course

stevehook: remove course sounds better to me too. Cancel seems appropriate for the whole application

tijmenb: Ok let's go with "remove" in UI. Would the state still "cancelled" or also "removed"?

app/views/support_interface/application_forms/show.html.erb
```diff
@@ -54,7 +54,9 @@
 <% end %>
 
 <% if @application_form.application_choices.any?(&:awaiting_references?) %>
-  <%= govuk_button_link_to 'Make changes to courses', support_interface_change_course_path(@application_form) %>
+<div class='govuk-body'>
+  <%= govuk_link_to 'Add or withdraw course choices', support_interface_change_course_path(@application_form) %>
```
tvararu, line 58: I liked [this tweet](https://mobile.twitter.com/thetuttingtutor/status/1252710968817319936) the other day:

> General UX writing rule I follow:
> • Navigation = nouns
> • Buttons = verbs

This could be a button as it was before, but yeah, minor.

app/views/support_interface/change_course/options.html.erb
```diff
@@ -1,24 +1,28 @@
 <% content_for :browser_title, t('support_interface.change_course.page_title') %>
```
tvararu, line 1: Title on this page should contain `Error:` if the form below fails validation; there should be a helper we can use for this.

app/views/support_interface/change_course/select_choice_to_withdraw.html.erb
```diff
@@ -0,0 +1,19 @@
+<% content_for :before_content, govuk_back_link_to(support_interface_change_course_path(@form.application_form)) %>
```
tvararu, line 1: There isn't a title on this page; which should contain `Error:` like the above suggestion.

config/locales/support_interface.yml
```diff
@@ -7,4 +7,8 @@ en:
       restore: Restore support user
 
     change_course:
-      page_title: What kind of change do you want to make?
+      page_title: What do you want to do?
+
+      add_course: 'Add another course choice'
+      remove_course: 'Withdraw a course choices'
```
tvararu, line 13:
```suggestion
      remove_course: 'Withdraw a course choice'
```

PR #1889, opened 2020-04-16
Allow provider users to amend existing provider users.

steventux:
## Context

When a managing provider user attempts to add a user who already exists they currently get a validation error on email uniqueness. There's a need for managing users for a provider to onboard users from a different provider (and are therefore not visible to the managing user).

This PR allows the `create` action to detect an existing user and amend their details. This is different from editing a user as mentioned in the user need above. We have future work to allow editing a provider user.

## Link to Trello card

https://trello.com/c/eAbezhne/1894-dont-refuse-when-a-provideruser-tries-to-add-another-whos-already-in-the-system

## Things to check

### DIFF
--- a/app/controllers/provider_interface/provider_users_controller.rb
+++ b/app/controllers/provider_interface/provider_users_controller.rb
@@ -24,7 +24,7 @@ module ProviderInterface
         form: @form,
         save_service: SaveProviderUser.new(provider_user: provider_user),
         invite_service: InviteProviderUser.new(provider_user: provider_user),
-        new_user: !@form.persisted?,
+        new_user: @form.existing_provider_user.blank?,
       )
 
       render :new and return unless service.call
--- a/app/controllers/provider_interface/provider_users_controller.rb
+++ b/app/controllers/provider_interface/provider_users_controller.rb
@@ -24,6 +24,7 @@ module ProviderInterface
         form: @form,
         save_service: SaveProviderUser.new(provider_user: provider_user),
         invite_service: InviteProviderUser.new(provider_user: provider_user),
+        new_user: !@form.persisted?,
       )
 
       render :new and return unless service.call
--- a/app/models/provider_interface/provider_user_form.rb
+++ b/app/models/provider_interface/provider_user_form.rb
@@ -10,13 +10,17 @@ module ProviderInterface
     validates :email_address, :first_name, :last_name, presence: true
     validates :email_address, email: true
     validates :provider_ids, presence: true
-    validate :email_is_unique
     validate :permitted_providers
 
     def build
       return unless valid?
 
-      @provider_user ||= ProviderUser.new
+      if existing_provider_user
+        @provider_user = existing_provider_user
+      else
+        @provider_user ||= ProviderUser.new
+      end
+
       @provider_user.first_name = first_name
       @provider_user.last_name = last_name
       @provider_user.email_address = email_address
@@ -58,12 +62,8 @@ module ProviderInterface
 
   private
 
-    def email_is_unique
-      return if persisted? && provider_user.email_address == email_address
-
-      return unless ProviderUser.exists?(email_address: email_address)
-
-      errors.add(:email_address, 'This email address is already in use')
+    def existing_provider_user
+      @existing_provider_user ||= ProviderUser.find_by(email_address: email_address)
     end
 
     def permitted_providers
--- a/app/models/provider_interface/provider_user_form.rb
+++ b/app/models/provider_interface/provider_user_form.rb
@@ -7,7 +7,8 @@ module ProviderInterface
     attr_writer :provider_ids
     attr_reader :email_address
 
-    validates :email_address, :first_name, :last_name, presence: true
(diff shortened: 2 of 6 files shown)


spec/system/provider_interface/provider_invites_user_spec.rb
```diff
@@ -72,4 +76,21 @@ def and_the_user_should_be_sent_a_welcome_email
...
+    fill_in 'First name', with: 'Jane'
+    fill_in 'Last name', with: 'Smith'
+    fill_in 'Email address', with: @email_address
+    check @another_provider.name_and_code
+    click_on 'Invite user'
+  end
+
+  def then_the_provider_is_assigned_to_the_user
+    expect(page).to have_content('Provider user invited')
+    expect(@new_provider_user.reload.providers).to include(@another_provider)
```
tijmenb, line 94: If you add the following line the test fails:

```diff
+ expect(@new_provider_user.reload.providers).to include(@provider)
```

I think what you see here is that the invite overrides the original `providers` of the user because they're not in the form, so don't get passed to the form object.

Looking at this now, I suspect this bit will become quite tricky! With this feature and the edit feature you'll have to keep in mind that someone can have access to a completely unrelated provider, so the checked boxes in the form could be a subset of the providers a user should have access to.

app/models/provider_interface/provider_user_form.rb
```diff
@@ -10,13 +10,17 @@ class ProviderUserForm
...
 
-      @provider_user ||= ProviderUser.new
+      if existing_provider_user
+        @provider_user = existing_provider_user
+      else
+        @provider_user ||= ProviderUser.new
+      end
+
       @provider_user.first_name = first_name
       @provider_user.last_name = last_name
```
tijmenb, line 25: Similar to my other comment, this overrides the first & last name for existing users. Technically this will allow a trolly school administrator to change the names for anyone in the system, even if they're not affiliated with the school.

steventux: :+1: I think this should be left out of the _edit_ scenario so we only update providers and nothing else.

spec/system/provider_interface/provider_invites_user_spec.rb
steventux: Yes I see what you mean, we need to populate the form with the providers from the existing user to reliably preserve them. This can't be done until we've submitted the form and discovered we're dealing with an existing user. 
If we just add the submitted providers to the existing user then there's no way to remove a provider (though that could arguably be handled in the _real_ edit form).

One fairly clunky way to get round the problem would be to return to the form in 'edit' mode populated with the existing user details but I'm not sure this would be clear to the managing user. I was thinking of an explicit message like:

_A user with this email already exists, if you would like to add them to specific providers please select them and resubmit this form._

app/models/provider_interface/provider_user_form.rb
steventux: @tijmenb I've distinguished between creating and updating here: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/commit/92d097ad2b238071061b6ef4c708ee9381c84be2

app/controllers/provider_interface/provider_users_controller.rb
```diff
@@ -24,6 +24,7 @@ def create
         form: @form,
         save_service: SaveProviderUser.new(provider_user: provider_user),
         invite_service: InviteProviderUser.new(provider_user: provider_user),
+        new_user: !@form.persisted?,
```
duncanjbrown, line 27: Would `form.existing_provider_user.blank?`(or words to that effect) work?

app/models/provider_interface/provider_user_form.rb
```diff
@@ -10,17 +10,11 @@ class ProviderUserForm
     validates :email_address, :first_name, :last_name, presence: true
     validates :email_address, email: true
     validates :provider_ids, presence: true
-    validate :email_is_unique
     validate :permitted_providers
 
     def build
-      return unless valid?
+      @provider_user = existing_provider_user ? build_from_existing_user : build_new_user
```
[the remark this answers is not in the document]
steventux: Yes I think it's pretty incongruous where it currently happens. Will update.

app/controllers/provider_interface/provider_users_controller.rb
steventux: :+1: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1889/commits/f19c8a5022b2418410a5996a50b257ce80f2ddf5

app/models/provider_interface/provider_user_form.rb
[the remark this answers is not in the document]
steventux: @duncanjbrown https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1889/commits/ab4dba6625a3b5bb18406818e3cc2e75ee2a0184 we now only validate these name fields based on the presence of an existing user.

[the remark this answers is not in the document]
duncanjbrown: Much better than my suggestion!

PR #1891, opened 2020-04-16
Create CSV export for all active provider users

davidgisbey:
## Context

We need to have a list of all the active provider users so that we can email them with updates

- Create a service that returns all the provider users who have signed in and the providers they are associated with
- Adds an action in the support provider user controller that outputs them to a csv
- Add a link the provider user page on the support app that you can download it from

Before 

image

After

image

## Guidance to review

In the card, it says to return a list of providers with all the active providers so we can email them.

It seems to make more sense to return the active provider users to me.

Happy to change it to providers with an active user if I'm missing something.  

## Link to Trello card

https://trello.com/c/sgWrSrDL/1267-create-download-link-for-csv-of-all-active-providers?menu=filter&filter=label:Dev

## Things to check

### DIFF
--- a/app/controllers/support_interface/provider_users_controller.rb
+++ b/app/controllers/support_interface/provider_users_controller.rb
@@ -1,5 +1,3 @@
-require 'csv'
-
 module SupportInterface
   class ProviderUsersController < SupportInterfaceController
     def index
@@ -43,20 +41,6 @@ module SupportInterface
       else
         render :edit
       end
-    end
-
-    def active_users_export
-      provider_users = SupportInterface::ActiveProviderUsersExport.call
-
-      csv = CSV.generate do |rows|
-        rows << provider_users.first.keys
-
-        provider_users.each do |a|
-          rows << a.values
-        end
-      end
-
-      send_data csv, filename: "active-provider-users-#{Time.zone.today}.csv", disposition: :attachment
     end
 
   private
--- a/app/controllers/support_interface/provider_users_controller.rb
+++ b/app/controllers/support_interface/provider_users_controller.rb
@@ -1,3 +1,5 @@
+require 'csv'
+
 module SupportInterface
   class ProviderUsersController < SupportInterfaceController
     def index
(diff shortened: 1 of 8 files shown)


app/controllers/support_interface/provider_users_controller.rb
```diff
@@ -43,6 +45,20 @@ def update
...
+      csv = CSV.generate do |rows|
+        rows << provider_users.first.keys
+
+        provider_users.each do |a|
+          rows << a.values
+        end
+      end
+
+      send_data csv, filename: "active-provider-users-#{Time.zone.today}.csv", disposition: :attachment
+    end
```
tijmenb, line 60: Perhaps we can move this into `PerformanceController` so that all exports are in the same place, it might make refactoring easier in the future and you can reuse the new `to_csv` helper.

davidgisbey: Yea I think it makes sense that all of the CSV exports are bunched too. I'll move it across 👍

app/views/support_interface/performance/index.html.erb
```diff
@@ -38,6 +38,7 @@
   The list of providers that are being synced from the Find service, along
   with when they signed the data sharing agreement.
 </p>
+</p>
```
tijmenb, line 41:
```suggestion
```

app/controllers/support_interface/provider_users_controller.rb
```diff
@@ -1,3 +1,5 @@
+require 'csv'
+
```
tijmenb, line 2:
```suggestion
```

spec/services/support_interface/active_provider_users_export_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'rails_helper'
+
+RSpec.describe SupportInterface::ActiveProviderUsersExport do
+  describe '#call' do
+    it 'returns provider_users who have have signd in at least once' do
```
tijmenb, line 5:
```suggestion
    it 'returns provider_users who have have signed in at least once' do
```

PR #1894, opened 2020-04-17
Only sync a single provider on Heroku review apps

tijmenb:
This is to stop the deluge of daily emails that warn about us reaching the database row limit.


tijmenb:
This brings a new database to 6.100 records:

Screenshot 2020-04-17 at 08 24 18

tvararu:
👍

PR #1899, opened 2020-04-17
Add migration to create new table to store validation errors

stevehook:
## Context

At the moment validation errors are not visible to us. So that we can learn more about candidates behaviour and better understand areas where our service appears to repeatedly 'fail' we want a log of validation errors.

This PR is the migration only first part of this feature.

This work is initially base on the spike: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/compare/track-validation-errors?expand=1

- [x] Add a new `validation_errors` database table
- [x] Add columns for polymorphic association to user (e.g. candidate) who triggered the validation error.
- [x] Column to store URL / Path of request that triggered the validation error.

## Guidance to review

- Is the proposed schema sensible?

```
\d validation_errors
Table "public.validation_errors"
┌──────────────┬────────────────────────────────┬───────────┬──────────┬───────────────────────────────────────────────┐
│    Column    │              Type              │ Collation │ Nullable │                    Default                    │
├──────────────┼────────────────────────────────┼───────────┼──────────┼───────────────────────────────────────────────┤
│ id           │ bigint                         │           │ not null │ nextval('validation_errors_id_seq'::regclass) │
│ form_object  │ character varying              │           │ not null │                                               │
│ user_id      │ integer                        │           │ not null │                                               │
│ user_type    │ character varying              │           │ not null │                                               │
│ request_path │ character varying              │           │ not null │                                               │
│ details      │ jsonb                          │           │          │                                               │
│ created_at   │ timestamp(6) without time zone │           │ not null │                                               │
│ updated_at   │ timestamp(6) without time zone │           │ not null │                                               │
└──────────────┴────────────────────────────────┴───────────┴──────────┴───────────────────────────────────────────────┘
Indexes:
    "validation_errors_pkey" PRIMARY KEY, btree (id)
    "index_validation_errors_on_form_object" btree (form_object)
```

- We using the name `form_object` to describe the thing that is validated. This makes sense in most cases because we widely use form-level validations, however there are cases where we validate an underlying model directly in candidate interface controllers. Is ok to have this kind of ambiguity in the naming or should we come up with a more generic term?
- Validation errors are linked back to a candidate (etc.) so they are traceable to individuals. This is the same as the way that we link the audit trail to candidates (and other kinds of users).

## Link to Trello card

## Things to check


db/migrate/20200417145121_create_validation_errors.rb
```diff
@@ -0,0 +1,10 @@
+class CreateValidationErrors < ActiveRecord::Migration[6.0]
+  def change
+    create_table :validation_errors do |t|
+      t.string :form_object
+      t.json :details
```
tijmenb, line 5: I think this needed to be `jsonb` - https://stackoverflow.com/questions/39637370/difference-between-json-and-jsonb-in-postgres

PR #1900, opened 2020-04-19
Limit amount of referees to 10

davidgisbey:
## Context

At the moment there is no limit on how many times a candidate can cancel their referee request and adds new ones.

We also don't have anything in place for the seventh referee onwards in the nth referee locales so the table name is blank (see before image)

- Limit the number of referees a candidate can have associated with their application to 10
- Redirects them to a page asking them to contact the support team page if they try to add an 11th referee or cancel a reference request and already have 10 referees

Before 

image

After

image

image

## Guidance to review

The designs will need a review (ux and content)

## Link to Trello card

https://trello.com/c/CXTNkMUo/1323-dev-limit-amount-of-referees-a-candidate-can-have-to-10

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/additional_referees_controller.rb
+++ b/app/controllers/candidate_interface/additional_referees_controller.rb
@@ -91,6 +91,8 @@ module CandidateInterface
 
     def contact_support
       @maximum_referees = ApplicationForm::MAXIMUM_REFERENCES
+
+      redirect_to candidate_interface_review_referees_path unless current_application.application_references.count >= @maximum_referees
     end
 
   private
--- a/app/controllers/candidate_interface/additional_referees_controller.rb
+++ b/app/controllers/candidate_interface/additional_referees_controller.rb
@@ -1,7 +1,7 @@
 module CandidateInterface
   class AdditionalRefereesController < CandidateInterfaceController
     rescue_from ActiveRecord::RecordNotFound, with: :render_404
-    before_action :redirect_to_contact_support_if_at_maximum_reference_count, only: %i[type update_type new create confirm]
+    before_action :redirect_to_contact_support_if_at_maximum_reference_count, only: %i[type update_type new create]
 
     def type
       @page_title = page_title_for_new_page
@@ -116,13 +116,13 @@ module CandidateInterface
     end
 
     def redirect_to_dashboard_if_no_references_needed
-      return if !reference_status.still_more_references_needed?
+      return if reference_status.still_more_references_needed?
 
       redirect_to candidate_interface_application_form_path
     end
 
     def redirect_to_contact_support_if_at_maximum_reference_count
-      redirect_to candidate_interface_additional_referee_contact_support_path if at_maximum_referee_count && reference_status.still_more_references_needed?
+      redirect_to candidate_interface_additional_referee_contact_support_path if at_maximum_referee_count && reference_status.needs_to_draft_another_reference?
     end
 
     def at_maximum_referee_count
--- a/app/controllers/candidate_interface/additional_referees_controller.rb
+++ b/app/controllers/candidate_interface/additional_referees_controller.rb
@@ -1,6 +1,7 @@
 module CandidateInterface
   class AdditionalRefereesController < CandidateInterfaceController
     rescue_from ActiveRecord::RecordNotFound, with: :render_404
+    before_action :redirect_to_contact_support_if_at_maximum_reference_count, only: %i[type update_type new create confirm]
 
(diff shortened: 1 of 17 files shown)


davidgisbey:
I've found this a bit more confusing than I would like to admit haha

config/locales/application_form.yml
```diff
@@ -317,6 +317,10 @@ en:
         - Fourth referee
         - Fifth referee
         - Sixth referee
+        - Seventh referee
+        - Eighth referee
+        - Ninth referee
+        - Tenth referee
```
tijmenb, line 323: Can we replace this by the TextOrdinalizer?

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/app/services/text_ordinalizer.rb#L1

tvararu:
@tijmenb 

> Changed my mind, it's useful to limit these things because they could be used to spam people.

Yeah, we discussed this a bit offline; without having a limit in place, someone could e.g. submit an application and then use a browser automation script to cancel a reference -> add a new reference -> repeat ad infinitum. On every loop, we'd send the affected referee an email via Notify.

This can be abused by a bad actor, e.g. by combining it with `+123` addresses, they could flood someone's inbox. (which I think might lead to our BAT email becoming blacklisted, which wouldn't be good)

app/controllers/candidate_interface/additional_referees_controller.rb
```diff
@@ -125,6 +130,14 @@ def redirect_to_dashboard_if_no_references_needed
       redirect_to candidate_interface_application_form_path
     end
 
+    def redirect_to_contact_support_if_at_maximum_reference_count
+      redirect_to candidate_interface_additional_referee_contact_support_path if at_maximum_referee_count && reference_status.needs_to_draft_another_reference?
+    end
```
tvararu, line 135: Why is `reference_status.needs_to_draft_another_reference?` necessary?

app/controllers/candidate_interface/additional_referees_controller.rb
```diff
@@ -125,6 +130,14 @@ def redirect_to_dashboard_if_no_references_needed
...
+
+    def at_maximum_referee_count
+      current_application.application_references.count >= ApplicationForm::MAXIMUM_REFERENCES
+    end
```
tvararu, line 139: We could move this to be a local variable inside of `redirect_to_contact_support_if_at_maximum_reference_count` if it isn't used anywhere else.

app/services/reference_status.rb
```diff
@@ -40,6 +36,10 @@ def feedback_completed?
     application_references.select(&:feedback_provided?).size == 2
   end
 
+  def non_overdue_references
+    application_references.reject(&:feedback_overdue?)
+  end
```
tvararu, line 41: Do we need to `reject(&:feedback_overdue?)`?

Following from that, if the only place we use this is with the `select` from above, we could also hoist the `select` here and then rename this:

```suggestion
  def references_that_do_not_need_replacing
    application_references.select { |reference| reference.feedback_provided? || reference.feedback_requested? }
  end
```

tvararu: Could this also just be `application_references.reject(&:replacement?)`?

fofr:
Has https://user-images.githubusercontent.com/42515961/79689805-0835eb00-824f-11ea-923a-1ce31e133204.png been through content design?

davidgisbey:
@fofr no, it needs a design and content review. I've been. busy with procurement work so this has been on the back burner.

app/controllers/candidate_interface/additional_referees_controller.rb
davidgisbey: it isn't you're right 👍

app/services/reference_status.rb
davidgisbey: I might be missing something but as feedback_overdue? is a method, not an enum value/attribute/boolean,  it means a reference can be both overdue and requested. Do we not need to reject them before checking how many are in a provided/requested state. (i.e. the ones that don't need replacing)?

app/controllers/candidate_interface/referees_controller.rb
```diff
@@ -11,6 +11,8 @@ def index
     end
 
     def type
+      @nth_referee = "#{TextOrdinalizer::ORDINALIZE_MAPPING[(@referees.count + 1)].capitalize} referee"
```
tvararu, line 14: Could be a private method in this controller to DRY up 👀

PR #1905, opened 2020-04-20
Update ProviderUser permissions atomically

steventux:
## Context

In an attempt to make permissions updates performant we called `update_all` on relevant records.
See https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1866/files#diff-7ea8d6c1126f23dfafc8a299cbf13497R19
This has the unintentional side-effect of not creating audit entries.

Call active record `update` on each `ProviderPermission` record as this creates audit log entries and ensures validations are called.

## Things to check

### DIFF
--- a/app/services/save_provider_user.rb
+++ b/app/services/save_provider_user.rb
@@ -6,28 +6,28 @@ class SaveProviderUser
 
   def call!
     @provider_user.save!
-    update_permissions
+    update_permissions!
     @provider_user.reload
   end
 
 private
 
-  def update_permissions
+  def update_permissions!
     unselected_permissions = ProviderPermissions
       .where(provider_user: @provider_user)
       .where.not(provider_id: @permissions.values.flatten)
 
     ActiveRecord::Base.transaction do
-      ProviderPermissionsOptions::VALID_PERMISSIONS.each do |permission|
-        unselected_permissions.where(permission => true).each { |perm| perm.update(permission => false) }
+      ProviderPermissionsOptions::VALID_PERMISSIONS.each do |permission_name|
+        unselected_permissions.where(permission_name => true).each do |permission|
+          permission.update!(permission_name => false)
+        end
       end
 
-      @permissions.each do |permission, provider_ids|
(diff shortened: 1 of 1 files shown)


app/services/save_provider_user.rb
```diff
@@ -13,15 +13,21 @@ def call!
...
+    unselected_permissions = ProviderPermissions
+      .where(provider_user: @provider_user)
+      .where.not(provider_id: @permissions.values.flatten)
+
     ActiveRecord::Base.transaction do
-      ProviderPermissions
-        .where(provider_user: @provider_user)
-        .update_all(ProviderPermissionsOptions.reset_attributes)
+      ProviderPermissionsOptions::VALID_PERMISSIONS.each do |permission|
+        unselected_permissions.where(permission => true).each { |perm| perm.update(permission => false) }
```
tijmenb, line 22: To avoid silent failures if validation fails:

```suggestion
        unselected_permissions.where(permission => true).each { |perm| perm.update!(permission => false) }
```

app/services/save_provider_user.rb
```diff
@@ -13,15 +13,21 @@ def call!
...
+      end
 
       @permissions.each do |permission, provider_ids|
-        ProviderPermissions
-          .where(provider_user: @provider_user, provider_id: provider_ids)
-          .update_all(permission => true)
+        provider_ids.each do |provider_id|
+          ProviderPermissions
+            .where(provider_user: @provider_user, provider_id: provider_id, permission => false)
+            .update(permission => true)
```
tijmenb, line 29:
```suggestion
            .update!(permission => true)
```

steventux: Ah yes, I was wondering why that wouldn't work for `ActiveRecord::Relation` wasn't paying attention to the error - updated here https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1905/commits/20570652057d73a36bb057f241de4ba031e9e40a

PR #1906, opened 2020-04-20
Send application_submitted provider email in after_commit callback

malcolmbaig:
## Context

In several of our environments, including production, we see the
following exception:

    ActionView::Template::Error
    wrong number of arguments (given 1, expected 0)

It appears intermittently and occurs when trying to send the
ProviderMailer#application_submitted email. Its source is this line
in the email template:

    @application.rbd_date.to_s(:govuk_date)

The error occurs because the `rbd_date` is nil (this is just a struct
attribute that holds the `reject_by_default_at` value of the
application_choice record). When inspecting a sample of the application
choice records that are subject to this error, the
`reject_by_default_at` value appears to be correctly populated.

The email is sent as part of the SendApplicationToProvider service
object, which updates the `reject_by_default_at` just before doing an async
send of the email. The error above is possibly due to a worker
attempting to send the email before the database commit that stores the
RBD value has finished.

Attempt to fix this by sending the email in an after_commit callback on
the ApplicationChoice model. Pass in a temporary flag on the model so
that the email is only sent in this particular scenario.

## Guidance to review

Also considered:
- Raising a custom exception in the mailer when rbd date is blank and configuring Sentry to ignore it. Decided against this as it potentially masks actual problems with rbd date population, and there doesn't seem to be a way to access retry counts (eg - to only raise if this is the first or second attempt) in these mailers without defining a custom worker to do the sending.
- Defining an after_commit callback on the instance at runtime (in the body of the SendApplicationToProvider service object). Doesn't appear to be possible in Rails.

Other ideas welcome.

## Link to Trello card

https://trello.com/c/8g0yXIfH
## Things to check

### DIFF
--- a/app/models/application_choice.rb
+++ b/app/models/application_choice.rb
@@ -2,6 +2,8 @@ class ApplicationChoice < ApplicationRecord
   include Chased
 
   before_create :set_initial_status
+  after_commit :send_new_application_email_to_provider, on: :update
+  attr_accessor :allow_new_application_email
 
   belongs_to :application_form, touch: true
   belongs_to :course_option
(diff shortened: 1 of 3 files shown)


malcolmbaig:
Discussing with @tijmenb , this is more likely due to the nested transaction created here https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/app/services/submit_reference.rb#L36

Reworking PR.

PR #1907, opened 2020-04-20
Remove edit application feature

tijmenb:
## Context

This feature was never turned on in production, for reasons outlined in the design history:

https://bat-design-history.netlify.app/apply-for-teacher-training/amending-mvp

This removes the feature altogether. It renames the main before_action to `redirect_to_dashboard_if_submitted`, which is what it was before the feature was introduced. It also adds it to some missing controllers (it was easy to see here which controllers didn't have changes made to them when doing a project search-replace).

## Guidance to review

Does it remove everything? Does it accidentally remove other functionality? 

## Link to Trello card

https://trello.com/c/r6ieKOuA

## Things to check


app/models/application_form.rb
```diff
@@ -108,10 +108,6 @@ def full_name
     "#{first_name} #{last_name}"
   end
 
-  def amendable?
-    FeatureFlag.active?('edit_application') && submitted? && within_amend_period?
-  end
-
   def within_amend_period?
```
duncanjbrown, line 111: I think this can go too?

tijmenb:
Updated, I also found some i18n keys and a template that wasn't used anymore.

PR #1912, opened 2020-04-21
Fix intermittent exception when sending application to provider

malcolmbaig:
## Context

In several of our environments, including production, we see the
following exception:

    ActionView::Template::Error
    wrong number of arguments (given 1, expected 0)

It appears intermittently and occurs when trying to send the
ProviderMailer#application_submitted email. Its source is this line
in the email template:

    @application.rbd_date.to_s(:govuk_date)

The error occurs because the rbd_date is nil (this is just a struct
attribute that holds the reject_by_default_at value of the
application_choice record).

The likely cause of this seems to be the transaction nesting that occurs
when SubmitReference calls the SendApplicationsToProvider service. The
call to this service is part of the top-level transaction. It iterates
over eligible application forms and calls SendApplicationToProvider for
each of their application choices. This service then contains database
changes wrapped in another transaction, followed by async delivery of
the application_submitted email.

Because these database changes are only committed when the top-level
transaction block closes, it seems likely that on some occasions a
Sidekiq worker is attempting to send the email before the
reject_by_default_at is present in the database record.

Shout out to @tijmenb for identifying the root cause.

This PR removes the transaction nesting, and instead leaves a linear
series of database updates and transactions for each discrete step of
the SubmitReference process.

## Guidance to review

 
na

## Link to Trello card
https://trello.com/c/8g0yXIfH

## Things to check

### DIFF
--- a/app/services/submit_reference.rb
+++ b/app/services/submit_reference.rb
@@ -15,9 +15,11 @@ class SubmitReference
       raise 'Can\'t submit a reference without answers to all questions'
     end
 
-    @reference.update!(feedback_status: 'feedback_provided')
-
-    progress_application_if_enough_references_have_been_submitted
+    if enough_references_have_been_submitted?
+      progress_applications
+    else
+      reference_feedback_provided!
+    end
 
     CandidateMailer.reference_received(@reference).deliver_later
     RefereeMailer.reference_confirmation_email(application_form, reference).deliver_later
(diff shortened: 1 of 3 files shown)


duncanjbrown:
To @stevehook's point, what if we branched earlier? a flow something like

```
if(this_is_the_last_reference_we_need)
  # one transaction which updates the references and commits the application state changes
  SendApplicationsToProvider.new.call
else
  # just update the reference
end
```

malcolmbaig:
@stevehook You're right - a partial failure would be possible here. I figured initially each update is a discrete step in the process and that an exception at any stage would be easy to diagnose and fix, but ideally we'd want to rollback everything if SubmitReference fails (and, for that matter, simply call the service again to get to a correct state once any issues are fixed).

I think something like @duncanjbrown 's suggestion is needed here. Will rework.

app/services/submit_reference.rb
```diff
@@ -26,21 +27,28 @@ def save!
...
+
+  def progress_applications
+    ActiveRecord::Base.transaction do
+      reference_feedback_provided!
+      application_form.application_choices.each do |application_choice|
+        ApplicationStateChange.new(application_choice).references_complete!
+      end
+    end
+
+    SendApplicationsToProvider.new.call
```
stevehook, line 52: This service also writes to the database (it may be badly named) so there is still the potential for a partial failure here. Application choices could be moved to the `application_complete` state rather than all the way to `awaiting_provider_feedback`. I don't think this is necessarily a massive problem in this case because background processing would pick them up and progress them later. So I think this is OK, what do you think @duncanjbrown ?

If we are really concerned about having an atomic transaction then putting 10 second delay on the mail is the least messy alternative I can think of. Having separate services for DB and mail would also work but seems wrong to me.

malcolmbaig: > I don't think this is necessarily a massive problem in this case because background processing would pick them up and progress them later.

My commit message could have been clearer in this regard, but yes - basically this. My thinking was also that the two services are loosely coupled, and that SubmitReference shouldn't concern itself with the state changes introduced by SendApplicationsToProvider. Any errors in the latter can be dealt with in isolation, with the knowledge that SubmitReference leaves application choices in the `application_complete` state.

duncanjbrown: I agree, it's eventually consistent and that's a good thing

PR #1913, opened 2020-04-21
Deduplicate the list of ProviderUsers we show to providers who can manage users

duncanjbrown:
## Context

The list of provider users was confusing as users appeared in it multiple times

Screenshot 2020-04-21 at 10 12 42

- `DISTINCT` the ProviderUser.visible_to scope so it only returns each user once
- show only users for providers the current user has permission to manage

## Guidance to review

Is it right to limit the scope to `manage_users: true` relationships?

## Link to Trello card

https://trello.com/c/H5qgGRbo/2011-provider-users-appear-multiple-times-in-the-list-managing-users-can-see

## Things to check

### DIFF
--- a/app/models/provider_user.rb
+++ b/app/models/provider_user.rb
@@ -13,6 +13,7 @@ class ProviderUser < ActiveRecord::Base
   scope :visible_to, ->(provider_user) {
     joins(:provider_permissions)
       .where(ProviderPermissions.table_name => { provider_id: provider_user.providers.pluck(:id) })
+      .distinct
   }
 
   def self.load_from_session(session)
--- a/app/models/provider_user.rb
+++ b/app/models/provider_user.rb
@@ -11,9 +11,13 @@ class ProviderUser < ActiveRecord::Base
   has_associated_audits
 
   scope :visible_to, ->(provider_user) {
-    joins(:provider_permissions)
-      .where(ProviderPermissions.table_name => { provider_id: provider_user.providers.pluck(:id) })
-      .distinct
+    providers_that_user_can_manage = provider_user.provider_permissions.manage_users.select(:provider_id)
+
+    users_that_user_can_see = ProviderPermissions.where(
+      provider_id: providers_that_user_can_manage,
+    ).select(:provider_user_id)
+
+    where(id: users_that_user_can_see)
   }
 
   def self.load_from_session(session)
--- a/spec/models/provider_user_spec.rb
+++ b/spec/models/provider_user_spec.rb
@@ -82,8 +82,9 @@ RSpec.describe ProviderUser, type: :model do
     it 'returns provider users with access to the same providers as the passed user' do
       provider = create(:provider)
 
-      user_a = create(:provider_user, providers: [provider])
+      user_a = create(:provider_user)
       user_b = create(:provider_user, providers: [provider])
+      create(:provider_permissions, provider_user: user_a, provider: provider, manage_users: true)
 
       expect(ProviderUser.visible_to(user_a)).to include(user_b)
       expect(ProviderUser.visible_to(user_a).count).to eq(2) # user_a can see themselves plus user_b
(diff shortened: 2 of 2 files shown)


app/models/provider_user.rb
```diff
@@ -12,7 +12,8 @@ class ProviderUser < ActiveRecord::Base
 
   scope :visible_to, ->(provider_user) {
     joins(:provider_permissions)
-      .where(ProviderPermissions.table_name => { provider_id: provider_user.providers.pluck(:id) })
+      .where(ProviderPermissions.table_name => { provider_id: provider_user.provider_permissions.where(manage_users: true).pluck(:provider_id) })
```
tijmenb, line 15: I don't really grasp how this will return duplicate ProviderUsers in the first place? A normal `Model.where()` could never do that, or am I confused? 

Does it have something to do with the `joins`? Does it need to be a left/right/inner join instead of whatever it is now? 

Or does it have something to do with the fact that we do 2 queries when calling this scope? Might be worth doing this so it becomes a subquery:

```suggestion
      .where(ProviderPermissions.table_name => { provider_id: provider_user.provider_permissions.where(manage_users: true).select(:provider_id) })
```

steventux, line 15: [There's a scope defined here](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/app/models/provider_permissions.rb#L9) which means you might be able to reduce this to:

```ruby
.where(ProviderPermissions.table_name => { provider_id: provider_user.provider_permissions.manage_users.pluck(:provider_id) })
```

I'm not sure about scoping to the permission though. 
[Currently the only use of this particular scope is on the users index](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/ff72db8338eff848b72122db426c8e49803abb01/app/controllers/provider_interface/provider_users_controller.rb#L6). 
It wasn't obvious in the original story whether the current provider user could see all users from visible providers or all users from visible providers where they could manage users.
I don't know what the significance of visibility versus manageability is in this case.

duncanjbrown: Good shout on the subquery — will add. The generated SQL looks like this:

```
SELECT "provider_users".* FROM "provider_users" INNER JOIN "provider_users_providers" ON "provider_users_providers"."provider_user_id
" = "provider_users"."id" WHERE "provider_users_providers"."provider_id" IN (SELECT "providers"."id" FROM "providers" INNER JOIN "provider_users_providers" ON
 "providers"."id" = "provider_users_providers"."provider_id" WHERE "provider_users_providers"."provider_user_id" = 5);
```

Or with the subquery evaluated:

```
 SELECT "provider_users".* FROM "provider_users" INNER JOIN "provider_users_providers" ON "provider_users_providers"."provider_user_id" = "provider_users"."id" WHERE "provider_users_providers"."provider_id" IN (5, 1, 6)
```

In either case, you can see the dupes racking up:

Screenshot 2020-04-21 at 16 11 04

I guess the INNER JOIN might be distinctly satisfied by each provider_user_id/provider_id combination, as they're part of the where? 

I initially solved this with `GROUP BY` provider_users.id but preferred `DISTINCT`.

duncanjbrown: Ah, I knew there was something like that somewhere! Will revisit. Maybe we should leave it as it is for now anyway.

tijmenb: Ok, I tried to solve this with `left_joins` but I got really confused!

Here's a single-query scope that does the same thing, in case you find it simpler to read (I'm not sure it's more performant):

```rb
  scope :visible_to, ->(provider_user) {
    providers_that_user_can_manage = provider_user.provider_permissions.manage_users.select(:provider_id)

    users_that_user_can_see = ProviderPermissions.where(
      provider_id: providers_that_user_can_manage,
    ).select(:provider_user_id)

    where(id: users_that_user_can_see)
  }

```

duncanjbrown: More readable, it's going in!

duncanjbrown:
:+1: I've adopted @tijmenb's suggestion which in turn adopts @steventux's.

Having slept on it I think it is the right thing to do to restrict viewing to users with `manage_users` permission, as it's consistent with the way we control access to the `Users` tab on Manage.

PR #1917, opened 2020-04-21
Prevent candidates from being able to add a course twice

tvararu:
## Context

One of our candidates was able to add the same course 3 times (within 3 minutes) and submitted their application successfully.

We validate for this as part of `PickCourseForm`, which works fine for courses without multiple sites. For courses with multiple sites, candidates can get around this by navigating to the sites page directly and submitting it.

This adds checks to the site options page as well as the POST action, and redirects the user to the courses review page with an appropriate error flash.

## Guidance to review

I will investigate if we need something like this for choosing a study mode as well.

### Before

The gif is too large to upload:

- log in as a candidate without any course choices
- choose a course with multiple sites (Gorse SCITT / Arts and Design)
- copy the URL for the site selection page
- choose the first site
- manually go to the site selection page
- choose the first site
- there are now duplicate course choices

### After

Screen Recording 2020-04-21 at 16 53 03

## Link to Trello card

https://trello.com/c/JYjDiQUm/1387-dev-investigate-why-a-user-was-able-to-add-3-of-the-same-course-to-their-application-mp1198

## Things to check

### DIFF
--- a/spec/system/candidate_interface/course_selection/candidate_selecting_a_course_spec.rb
+++ b/spec/system/candidate_interface/course_selection/candidate_selecting_a_course_spec.rb
@@ -30,6 +30,9 @@ RSpec.feature 'Selecting a course' do
     and_i_choose_a_course
     then_i_see_a_message_that_ive_already_chosen_the_course
 
+    when_i_visit_the_site_page_for_a_course_i_have_already_selected
+    then_i_see_a_message_that_ive_already_chosen_the_course
+
     given_that_i_am_on_the_course_choices_review_page
     when_i_click_on_add_another_course
     and_i_choose_that_i_know_where_i_want_to_apply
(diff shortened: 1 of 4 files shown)


spec/system/candidate_interface/course_selection/candidate_selecting_a_course_spec.rb
```diff
@@ -204,13 +207,6 @@ def when_i_click_continue
     click_button 'Continue'
   end
 
-  def when_i_click_back
-    # TODO: the back link goes back to the '/candidate/application/courses/provider' instead of going back to the specific provider.
-    # Fix this and use click_link 'Back'
-
-    visit '/candidate/application/courses/provider/R55/courses'
-  end
```
tvararu, line 212: This is unused.

tvararu:
> I will investigate if we need something like this for choosing a study mode as well.

No, users are directed to the site page regardless even if it's single-site, which then does the checks in this PR, so they can't add duplicate sites that way. 👍

tijmenb:
I wonder if we should split up or simplify this controller, it has become quite hard to follow. The number of distinct flows and edge cases seem to be growing.

tvararu:
@tijmenb I agree, was thinking the same when building this, have added a card https://trello.com/c/5rSJREyS/240-refactor-coursescontroller

PR #1918, opened 2020-04-21
Alert the support channel when the 1000th candidate signs up

tvararu:
Small/fun thing suggested by Dan Williams.


tvararu:
We're missing an Azure build on this PR but I assume it's unlikely to break anything.

PR #1919, opened 2020-04-21
Add accredited body, study mode to the offer section of single application pages

duncanjbrown:
## Context

This information isn't present on the single application page in Manage.

- refactor the offer details so we can change them in one place
- change them in that place

This doesn't cover adding the funding type to the offer because there's a bit more to do on that first.

## Guidance to review

See commits.

## Link to Trello card

https://trello.com/c/HS9uplti/1994-build-provider-ui-add-view-of-accredited-body-study-mode-funding-route-to-the-offer-section-of-single-application-page

## Things to check

### DIFF
--- a/app/views/provider_interface/application_choices/show.html.erb
+++ b/app/views/provider_interface/application_choices/show.html.erb
@@ -49,9 +49,9 @@
 
     <%= render SummaryListComponent.new(rows: [
       { key: 'Provider', value: @application_choice.provider.name_and_code },
-      { key: 'Course', value: @application_choice.offered_course.name_and_code },
+      { key: 'Course', value: @application_choice.course.name_and_code },
       { key: 'Cycle', value: @application_choice.course.recruitment_cycle_year },
-      { key: 'Preferred location', value: @application_choice.offered_site.name_and_code },
+      { key: 'Preferred location', value: @application_choice.site.name_and_code },
       { key: 'Study mode', value: @application_choice.course_option.study_mode.humanize },
     ]) %>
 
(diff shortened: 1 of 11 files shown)


app/views/provider_interface/application_choices/show.html.erb
```diff
@@ -49,9 +49,9 @@
 
     <%= render SummaryListComponent.new(rows: [
       { key: 'Provider', value: @application_choice.provider.name_and_code },
-      { key: 'Course', value: @application_choice.offered_course.name_and_code },
+      { key: 'Course', value: @application_choice.course.name_and_code },
```
mnacos, line 52: Guessing the intention here is to display the original choice? It's not super clear, but guessing this will be redesigned soon enough.

app/views/provider_interface/application_choices/show.html.erb
```diff
@@ -49,9 +49,9 @@
...
       { key: 'Cycle', value: @application_choice.course.recruitment_cycle_year },
-      { key: 'Preferred location', value: @application_choice.offered_site.name_and_code },
+      { key: 'Preferred location', value: @application_choice.site.name_and_code },
```
mnacos, line 54: The use of 'Preferred' makes this clearer for site.

app/views/provider_interface/application_choices/show.html.erb
duncanjbrown: Yeah, I thought it was confusing to mutate this as we have a separate `offer` section

duncanjbrown:
Thanks @mnacos!

PR #1922, opened 2020-04-22
Improve visibility of validation errors

stevehook:
## Context

We need to see more about the validation errors our users are encountering so that we can learn more about their behaviour and better understand areas where our service appears to repeatedly 'fail'.

This PR will focus solely on recording validation errors in the `validation_errors` database table. A follow-up PR will add a Support user interface for viewing the errors captured.

Migration already deployed from https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1899

- [X] Add a `track_validation_errors` feature flag - nothing is recorded until this is enabled.
- [X] Add a `track_validation_error` method to the `CandidateInterfaceController` base class and call this from all the other controller actions in the candidate interface that can potentially trigger validation errors.
- [X] Modify a system spec to verify that errors are captured.

Example for `ValidationError#detail`:
```
{
  "phone_number" => {
    "value" => "NOT A NUMBER",
    "messages" => ["Enter a phone number, like 01632 960 001, 07700 900 982 or +44 0808 157 0192"]
  }
}
```

## Guidance to review

- We've decided to go with a solution that gets triggered at the controller level (rather than in the models or form objects that actually implement the validations) because we think we need to manage the error tracking per controller action. Doing it at the form object or model level could mean that we end up triggering the logic to store validation errors more often than we need to. The main downside to this approach is that it's potentially error prone as it requires developers to add the tracking call to every 'validating' action that we add in future.
- I'm not sure about the best approach to testing. System specs are part of the solution but it seems we need something a bit more low-level to test the nuts and bolts of the validation error tracking so I added some request specs (even though we don't often use them).

## Link to Trello card

## Things to check

### DIFF
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -465,4 +465,9 @@ FactoryBot.define do
     provider
     provider_user
   end
+
+  factory :validation_error do
+    form_object { 'RefereeInterface::ReferenceFeedbackForm' }
+    details { { feedback: 'must be present' } }
+  end
 end
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -468,7 +468,7 @@ FactoryBot.define do
 
   factory :validation_error do
     form_object { 'RefereeInterface::ReferenceFeedbackForm' }
-    details { { feedback: 'must be present' } }
+    details { { feedback: { messages: ['Enter feedback'], value: '' } } }
     association :user, factory: :candidate
     request_path { '/candidate' }
   end
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -469,5 +469,7 @@ FactoryBot.define do
   factory :validation_error do
     form_object { 'RefereeInterface::ReferenceFeedbackForm' }
     details { { feedback: 'must be present' } }
+    association :user, factory: :candidate
+    request_path { '/candidate' }
   end
 end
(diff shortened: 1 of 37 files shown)


spec/factories.rb
```diff
@@ -465,4 +465,11 @@
     provider
     provider_user
   end
+
+  factory :validation_error do
+    form_object { 'RefereeInterface::ReferenceFeedbackForm' }
+    details { { feedback: 'must be present' } }
```
tvararu, line 471: My understanding is the records should look like this:

```suggestion
    details { { feedback: { messages: ['must be present'], value: '' } } }
```

I could be wrong though, can you double check?

tvararu:
Tested on Heroku with success:

Screenshot 2020-04-25 at 14 27 22

```ruby
irb(main):002:0> ValidationError.all
=> #<ActiveRecord::Relation [#<ValidationError id: 1, form_object: "CandidateInterface::PersonalDetailsForm", user_id: 1, user_type: "Candidate", request_path: "/candidate/application/personal-details/review", details: {"first_nationality"=>{"value"=>"", "messages"=>["Select your nationality from the list"]}}, created_at: "2020-04-25 13:28:08", updated_at: "2020-04-25 13:28:08">]>
```

spec/factories.rb
stevehook: You're right - will fix that.

PR #1925, opened 2020-04-22
Providers have UI to change offers to rejections

duncanjbrown:
## Context

Providers need to do this and it's allowed. They can't do it, though, because there's no link to the flow they need: https://ukgovernmentdfe.slack.com/archives/CPH8J9G65/p1587483378323800

Add a `Make an offer` link to the single application page when the application is in a rejected or withdrawn-by-us state

Screenshot 2020-04-22 at 16 51 45

## Guidance to review

This is not quite as good as it could be because the user doesn't get the full "respond to application" flow which would allow them to change the provider, course etc — the link takes them straight to the "set conditions" page, ie the last step. To make a change like that, the provider would need to first make the offer, then use the "Change" links to correct it.

I did this because the "respond" flow also includes a "reject application" option in it, which leads to a "the application is not ready for that action" dead end. In due course it would probably be good to change "Make an offer" to "Change decision" and populate those radios accordingly, but hopefully this will do in the short term.

## Link to Trello card

https://trello.com/c/zf8GFsC4/2017-provide-a-link-for-providers-to-re-make-an-offer-after-theyve-rejected-a-candidate

## Things to check

### DIFF
--- a/app/components/provider_interface/status_box_components/rejected_component.html.erb
+++ b/app/components/provider_interface/status_box_components/rejected_component.html.erb
@@ -2,7 +2,18 @@
   <h2 class="govuk-heading-l">Offer</h2>
   <%= render SummaryListComponent.new(rows: offer_withdrawn_rows) %>
   <%= render ProviderInterface::ConditionsComponent.new(application_choice: application_choice) %>
+  <% if FeatureFlag.active?('provider_change_response') %>
+    <p class="govuk-body govuk-!-margin-bottom-0 app-!-print-display-none">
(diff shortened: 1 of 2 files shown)


adamsilver:
I am thinking for MVP...

‘Undo withdrawal and make offer’
‘Undo rejection and make offer’

duncanjbrown:
Updated per Slack with

- Change rejection to offer
- Undo withdrawal and make offer

Links still go to the conditions page

duncanjbrown:
Closing because this isn't the right answer https://ukgovernmentdfe.slack.com/archives/CPH8J9G65/p1587993510399500

PR #1927, opened 2020-04-23
Mark emails that we can't send to GOV.UK Notify

tijmenb:
## Context

This marks emails in the email log with the delivery status `notify_error` if we encounter an error sending the email to GOV.UK Notify. It comes on the back of the recent PaaS incident that took down the service for a couple of hours. We currently just retry errors from Notify, but we never mark the failed email in the email log as such, which means it will forever have the status "pending". 

Note that we `rescue_from` all Notify exceptions. I originally tried to distinguish between client errors like invalid API key or email address (anything descendent from `Notifications::Client::ClientError`) and server errors (`Notifications::Client::ServerError`). 

The idea was that we could have separate statuses for each, and perhaps skip the retries for client errors. However, this is not going to work because the way GOV.UK Notify system behaved during the incident.

We did see some Bad Gateway errors, which raised a ServerError (https://sentry.io/organizations/dfe-bat/issues/1568629644). But other errors were 404
(https://sentry.io/organizations/dfe-bat/issues/1623747428) and a 400 (https://sentry.io/organizations/dfe-bat/issues/1569990518). This means we can't reliably distinguish between client and server errors from Notify, and we'll have to treat them the same.

## Guidance to review

Does it make sense?

## Link to Trello card

https://trello.com/c/TcOj7XwZ

## Things to check

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -1,6 +1,16 @@
 class ApplicationMailer < Mail::Notify::Mailer
   GENERIC_NOTIFY_TEMPLATE = '2744ea53-34f1-431f-8173-8388fadd826a'.freeze
 
+  rescue_from Notifications::Client::RequestError do
+    # WARNING: this needs to be a block, otherwise the exception won't be
+    # re-raised and we won't be notified via Sentry, and the job won't retry.
+    #
+    # @see https://github.com/rails/rails/issues/39018
+    email = Email.find(self.headers['email-log-id'].to_s)
+    email.update!(delivery_status: 'notify_error')
+    raise
+  end
+
   def notify_email(headers)
     headers = headers.merge(rails_mailer: mailer_name, rails_mail_template: action_name)
     view_mail(GENERIC_NOTIFY_TEMPLATE, headers)
--- a/app/components/support_interface/email_log_row_component.rb
+++ b/app/components/support_interface/email_log_row_component.rb
@@ -31,6 +31,7 @@ module SupportInterface
         pending: 'govuk-tag--yellow',
         unknown: 'govuk-tag--grey',
         delivered: 'govuk-tag--green',
+        notify_error: 'govuk-tag--red',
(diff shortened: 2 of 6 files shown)


app/mailers/application_mailer.rb
```diff
@@ -1,6 +1,16 @@
 class ApplicationMailer < Mail::Notify::Mailer
   GENERIC_NOTIFY_TEMPLATE = '2744ea53-34f1-431f-8173-8388fadd826a'.freeze
 
+  rescue_from Notifications::Client::RequestError do
+    # WARNING: this needs to be a block, otherwise the exception won't be
+    # re-raised and we won't be notified via Sentry, and the job won't retry.
+    #
+    # @see https://github.com/rails/rails/issues/39018
+    email = Email.find(self.headers['email-log-id'].to_s)
```
duncanjbrown, line 9: Say there's an email with an unstable `notify_ref` and notify is down. Each time we try to deliver it we generate a new ref. Each time it fails, the log-id is plucked out of the header, and we mark that attempt as `notify_error`. Once notify gets the email then it calls us back with the last `notify_ref`, we do `Email.find_by(notify_reference: blah)` and we mark that `delivered`. All the failed ones are marked `failed` and the successful one is marked `successful`. This is good.

But if we have a stable `notify_ref` eg `production_reference-request_104`, then the `find_by(notify_reference: production_reference_request_104)` _could_ find one of the failures instead of the pending one. The email log would then contain a pending email as well as some `notify_error`s and one `successful`.

Should we change `find_by(notify_reference: blah)` to `find_by(notify_reference: blah, status: 'pending')`?

duncanjbrown: Alternatively, we could always append random hashes to `notify_ref`s

tijmenb: Good catch! PR here: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1928

PR #1928, opened 2020-04-23
Make Email#notify_reference unique

tijmenb:
## Context

This appends a random string to the notify reference for emails with a stable reference. We need this because when an email fails to send (GOV.UK Notify returns an error), we'll start retrying. This causes multiple emails in the log with the same `notify_reference`.

When getting the callback from Notify it becomes tricky to determine which of the emails the delivery receipt is for. This was raised by @duncanjbrown in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1927#discussion_r413645825.

Update the reference.

## Link to Trello card

https://trello.com/c/TcOj7XwZ/1405-wrap-up-tuesdays-incident-where-govuk-notify-was-down

## Things to check

### DIFF
--- a/app/mailers/authentication_mailer.rb
+++ b/app/mailers/authentication_mailer.rb
@@ -5,7 +5,7 @@ class AuthenticationMailer < ApplicationMailer
     notify_email(
       to: candidate.email_address,
       subject: t('authentication.sign_up.email.subject'),
-      reference: "#{HostingEnvironment.environment_name}-sign_up_email-#{candidate.id}",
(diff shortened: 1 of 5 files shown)


duncanjbrown:
Should we append it in `EmailLogInterceptor`? Then we wouldn't be able to forget

tijmenb:
@duncanjbrown not sure, I'd rather not mutate the value, it seems more explicit this way? I don't think we'll add any more custom references by the way.

PR #1931, opened 2020-04-23
Providers can assign permissions to their own users

steventux:
## Context

Provider users can invite other provider users if they have the required permissions. We need a way for managing provider users to grant permissions to others.

- Adds a details page per provider user listing their providers and permissions
- Adds a way to update these providers and permissions.

Paired with @willmcb on this feature.

image

and to update providers and permissions...

image

## Guidance to review

I think we need to make labels consistent when editing providers and permissions. The design focuses on changing permissions but adding and removing Providers is an inherent part of the nested checkbox layout. So it might be necessary to review labeling and phrasing around Providers and Permissions.

## Link to Trello card

https://trello.com/c/BTWwpO0Y/2004-providers-with-the-manageusers-permission-can-edit-users-associated-with-their-organisation

## Things to check

### DIFF
--- a/app/components/provider_interface/provider_permissions_list_component.html.erb
+++ b/app/components/provider_interface/provider_permissions_list_component.html.erb
@@ -1,0 +1,14 @@
+<% visible_provider_permissions.each do |provider_permission| %>
+  <h4><%= provider_permission.provider.name %></h4>
+  <ul id="provider-<%= provider_permission.provider.id %>-enabled-permissions">
+    <% ProviderPermissionsOptions::VALID_PERMISSIONS.each do |permission_name| %>
+      <li>
+        <% if provider_permission.send(permission_name) %>
+          <%= permission_name.to_s.humanize %>
+        <% else %>
+            No permissions
+        <% end %>
+      </li>
+    <% end %>
+  </ul>
+<% end %>
--- a/app/components/provider_interface/provider_permissions_list_component.rb
+++ b/app/components/provider_interface/provider_permissions_list_component.rb
@@ -1,0 +1,15 @@
+module ProviderInterface
+  class ProviderPermissionsListComponent < ViewComponent::Base
+    include ViewHelper
+    attr_reader :provider_permissions, :possible_permissions
+
+    def initialize(provider_permissions:, possible_permissions:)
+      @provider_permissions = provider_permissions
+      @possible_permissions = possible_permissions
+    end
+
+    def visible_provider_permissions
+      possible_permissions & provider_permissions
+    end
+  end
+end
--- a/app/components/provider_interface/provider_user_details_component.rb
+++ b/app/components/provider_interface/provider_user_details_component.rb
@@ -1,0 +1,35 @@
+module ProviderInterface
+  class ProviderUserDetailsComponent < ViewComponent::Base
+    include ViewHelper
+    attr_reader :header
+
+    def initialize(provider_user:, provider_permissions:, possible_permissions:)
+      @provider_user = provider_user
+      @provider_permissions = provider_permissions
+      @header = provider_user.full_name
+      @possible_permissions = possible_permissions
+    end
+
+    def rows
+      [
+        { key: 'Name', value: @provider_user.full_name },
+        { key: 'Email', value: @provider_user.email_address },
+        permissions_row,
+      ]
+    end
+
+    def permissions_row
+      {
+        key: 'Permissions',
+        value: render(
+          ProviderInterface::ProviderPermissionsListComponent.new(
+            provider_permissions: @provider_permissions,
+            possible_permissions: @possible_permissions,
+          ),
+        ),
+        change_path: provider_interface_provider_user_edit_providers_path(@provider_user),
+        action: 'Change',
+      }
+    end
+  end
+end
--- a/app/controllers/provider_interface/provider_users_controller.rb
+++ b/app/controllers/provider_interface/provider_users_controller.rb
@@ -1,9 +1,23 @@
 module ProviderInterface
   class ProviderUsersController < ProviderInterfaceController
-    before_action :requires_provider_add_provider_users_feature_flag, only: %i[new create]
+    before_action :requires_provider_add_provider_users_feature_flag
+    before_action :redirect_unless_permitted_to_manage_users
 
     def index
       @provider_users = ProviderUser.visible_to(current_provider_user)
+    end
+
+    def show
+      @provider_user = ProviderUser.visible_to(current_provider_user).find_by(id: params[:id])
+
+      redirect_to(action: :index) and return unless @provider_user
+
+      form = ProviderUserForm.new(
+        provider_user: @provider_user,
+        current_provider_user: current_provider_user,
+      )
+      @possible_permissions = form.possible_permissions
+      @provider_permissions = form.provider_permissions
     end
 
     def new
@@ -17,12 +31,18 @@ module ProviderInterface
 
     def create
       @form = ProviderUserForm.new(
-        provider_user_params.merge(current_provider_user: current_provider_user),
+        provider_user_params.merge(
+          current_provider_user: current_provider_user,
+          provider_permissions: provider_permissions_params,
+        ),
       )
       provider_user = @form.build
       service = SaveAndInviteProviderUser.new(
         form: @form,
-        save_service: SaveProviderUser.new(provider_user: provider_user),
+        save_service: SaveProviderUser.new(
+          provider_user: provider_user,
+          provider_permissions: @form.provider_permissions,
+        ),
         invite_service: InviteProviderUser.new(provider_user: provider_user),
         new_user: @form.existing_provider_user.blank?,
       )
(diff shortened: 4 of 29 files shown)


willmcb:
Found bug where if you remove a permission it currently renders a blank list item:

Screenshot 2020-04-24 at 09 48 17

Will put in fix for this and test for the component.

willmcb:
Also found issue where when you remove a provider it fails - will look into fix for this as well.

spec/system/provider_interface/manage_provider_user_permissions_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+  def and_the_provider_add_provider_users_feature_is_enabled
+    FeatureFlag.activate('provider_add_provider_users')
+  end
+
+  def when_i_click_on_the_users_link
+    click_on('Users')
+  end
+
+  def and_i_click_on_a_user
+    click_on(@provider.provider_users.first.full_name)
```
tijmenb, line 50: Is this the current user themself that we're editing? Or are we creating a separate user somewhere?

app/controllers/provider_interface/provider_users_controller.rb
```diff
@@ -33,15 +39,42 @@ def create
...
+      provider_user = ProviderUser.find(params[:provider_user_id])
+      @form = ProviderUserForm.from_provider_user(provider_user)
+      @form.current_provider_user = current_provider_user
+    end
+
+    def update_providers
+      provider_user = ProviderUser.find(params[:provider_user_id])
+      provider_user.assign_attributes(provider_user_params.except(:permissions))
+      @form = ProviderUserForm.from_provider_user(provider_user)
+      service = SaveProviderUser.new(provider_user: provider_user, permissions: permissions_params)
```
tijmenb, line 52: If the user we're editing has access to provider A and B, and the `current_provider_user` only has access to A, when the form is submitted, the user's access to B is accidentally revoked.

This is what I tried to warn about in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1889#discussion_r409808564. Do we need some kind of extra layer to avoid us accidentally doing this? It's a pretty tricky thing to keep in mind.

app/controllers/provider_interface/provider_users_controller.rb
```diff
@@ -33,15 +39,42 @@ def create
...
 
+    def edit_providers
+      provider_user = ProviderUser.find(params[:provider_user_id])
+      @form = ProviderUserForm.from_provider_user(provider_user)
+      @form.current_provider_user = current_provider_user
+    end
+
+    def update_providers
+      provider_user = ProviderUser.find(params[:provider_user_id])
+      provider_user.assign_attributes(provider_user_params.except(:permissions))
```
tijmenb, line 50: This allows users to fiddle with the HTML and edit the user's name and email address.

app/controllers/provider_interface/provider_users_controller.rb
```diff
@@ -1,11 +1,17 @@
...
-    before_action :requires_provider_add_provider_users_feature_flag, only: %i[new create]
+    before_action :requires_provider_add_provider_users_feature_flag
+    before_action :redirect_unless_permitted_to_manage_users
 
     def index
       @provider_users = ProviderUser.visible_to(current_provider_user)
     end
 
+    def show
+      @provider_user = ProviderUser.find_by(id: params[:id])
```
tijmenb, line 11: This allows someone with manage_users access to a provider to see all provider users in the system by changing the ID in the URL. 

It also allows them to change their provider organisations (but only to the ones the current user has access to).

app/controllers/provider_interface/provider_users_controller.rb
```diff
@@ -33,15 +39,42 @@ def create
...
+      @form.current_provider_user = current_provider_user
+    end
+
+    def update_providers
+      provider_user = ProviderUser.find(params[:provider_user_id])
+      provider_user.assign_attributes(provider_user_params.except(:permissions))
+      @form = ProviderUserForm.from_provider_user(provider_user)
+      service = SaveProviderUser.new(provider_user: provider_user, permissions: permissions_params)
+
+      render :edit_providers and return unless service.call!
```
tijmenb, line 54: This will still raise a 500 `ActionController::ParameterMissing` if no checkbox is checked on the edit providers page.

app/components/provider_interface/provider_permissions_list_component.html.erb
```diff
@@ -0,0 +1,11 @@
+<% provider_user.providers.each do |provider| %>
```
tijmenb, line 1: This will expose the fact that the user has access to a certain provider, even if the `current_provider_user` does not have access to that provider. This problem also exists on the index page by the way.

app/controllers/provider_interface/provider_users_controller.rb
steventux: Ah yes, we should only unassign what is present in the form and unchecked then. :+1:

app/controllers/provider_interface/provider_users_controller.rb
willmcb: Thanks - have scoped this to the `current_provider_user` now

app/controllers/provider_interface/provider_users_controller.rb
steventux: @tijmenb https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1931/commits/689a2b97504f82f7225cd55cd4d093836f12991b attempts to address this. I think we'll need to perform a similar operation for permissions as the same problem exists, will see if I can produce something common to both.

app/controllers/provider_interface/provider_users_controller.rb
```diff
@@ -33,15 +45,42 @@ def create
...
     end
 
+    def edit_providers
+      provider_user = ProviderUser.find(params[:provider_user_id])
+      @form = ProviderUserForm.from_provider_user(provider_user)
+      @form.current_provider_user = current_provider_user
+    end
+
+    def update_providers
+      provider_user = ProviderUser.find(params[:provider_user_id])
```
duncanjbrown, line 55: Should we scope this here and in the `edit` action, as in `show`? 

```ruby
provider_user = ProviderUser.visible_to(current_provider_user).find_by(id: params[:id])
```

app/controllers/provider_interface/provider_users_controller.rb
duncanjbrown: Managing a provider association is the same thing as managing a permission. If the form only had the power to act on a certain set of permissions then could we make the updates much dumber?

steventux: @duncanjbrown I see what you mean. Handling all this in the service layer would mean doing the updates in a single transaction. @willmcb and I have been pairing on a way to do this.

duncanjbrown: @steventux I did some sketching. Got a bit caught up with Rails form magic, but I think this approach might be worth a go.

commit 1, which pilots it on the `show` action: cb9e1029d225518bb95ed787b7cb904a4ebd7047

commit 2, which is an (incomplete) attempt to approach editing in a similar way
361e2743427f6ddba9bb1f4b30338ffd37b88bca

If this worked I think we could get rid of the intermediate representation of permissions in the OptionsService and various other things here and in support

duncanjbrown: slept on it and came up with https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/commit/fc8e646adf15d0b11f6b8f48f6987dd3d25be3ac which I think could work

app/components/provider_interface/provider_permissions_list_component.rb
```diff
@@ -0,0 +1,15 @@
+module ProviderInterface
+  class ProviderPermissionsListComponent < ActionView::Component::Base
```
tvararu, line 2:
```suggestion
  class ProviderPermissionsListComponent < ViewComponent::Base
```

app/components/provider_interface/provider_user_details_component.rb
```diff
@@ -0,0 +1,35 @@
+module ProviderInterface
+  class ProviderUserDetailsComponent < ActionView::Component::Base
```
tvararu, line 2:
```suggestion
  class ProviderUserDetailsComponent < ViewComponent::Base
```

steventux:
@tijmenb @duncanjbrown This PR has been widely refactored since https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1931/commits/c2c7d115c6552fef39eafc1fe1304d0663d981c2

Things we've tried to address:

- Confining visibility of provider users as per comments here: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1931#discussion_r415978976
- Ensuring we don't remove non-visible providers by params-omision mentioned here: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1931#discussion_r415962162
- Confining visibility of providers mentioned here: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1931#discussion_r415978976
- Ensuring the additions, removals and updates to permissions are atomic so the audit trail isn't noisy (removing and re-adding permissions would cause more noise than necessary).

We opted for @duncanjbrown 's idea of serializing `ProviderPermission` instances from the view via the `ProviderUserForm` but due to the shared service for saving provider users and updating their permissions some work was necessary on the support interface side of things as well.

app/controllers/provider_interface/provider_users_controller.rb
willmcb: Thanks @tijmenb this is cause by the checkboxes no being included in the form post requests if they are blank. Played around with this a bit this morning and came up with the following fix:

```ruby
    def update_providers
      provider_user = ProviderUser.find(params[:provider_user_id])

      redirect_to_edit_permissions_with_warning and return if provider_user_params.nil?

      provider_user.assign_attributes(provider_user_params.except(:permissions))
      @form = ProviderUserForm.from_provider_user(provider_user)
      service = SaveProviderUser.new(provider_user: provider_user, permissions: permissions_params)

      render :edit_providers and return unless service.call!

      flash[:success] = 'Providers updated'
      redirect_to provider_interface_provider_user_path(provider_user)
    end

  private

    def provider_user_params
      if params.has_key? (:provider_interface_provider_user_form)
        params.require(:provider_interface_provider_user_form)
              .permit(:email_address, :first_name, :last_name, provider_ids: [], permissions: {})
      end
    end

    def redirect_to_edit_permissions_with_warning
      flash[:warning] = 'A user needs to be associated with at least one provider'
      redirect_to provider_interface_provider_user_edit_providers_path(params[:provider_user_id])
    end
```
My understanding is that we would want users to be associated with at least one provider hence the use of the warning.

I think this is a bit messy though - the other option I was thinking about is to have hidden fields that submit with the checkboxes. This means that the form params can still be retrieved from the checkbox form and then I could add validations to the model rather than special control flow in the controller.

What are your thoughts?

Working on a spec for this in any case as the result should be the same I think.

willmcb: Ignore this I was acting on old information 😢

PR #1935, opened 2020-04-23
Enable the Zeitwerk autoloader

duncanjbrown:
This was introduced in Rails 6. Making this upgrade doesn't really enable anything except allowing us to upgrade ActionView::Component (now renamed to ViewComponent). I went down this detour because I noticed our component previews were broken, and wanted to bring the library up to date.

My process was:

- Enable zeitwerk in `application.rb`
- this fails out of the box because it can't find FindAPI.
- this is because FindAPI is defined in a file called find_api.rb.
- Fix this by setting acronym('API') in the inflector
- This breaks a lot of other autoloading
- Fix all those errors by renaming 'Api' to 'API' in class and module names
- be led by other capitalized classnames to add acronym('UCAS')
- fix all THOSE files too by renaming 'Ucas' to 'UCAS'

DO NOT rename VendorApiUser to VendorAPIUser - put an exemption in the
`zeitwerk.rb` initializer - because this is used as a polymorphic
type: column in the audits table. We can and (I think) should push a breaking change through to production to rename all instances of `VendorApiUser` in `auditable_audits.user_type` to `VendorAPIUser` as there are no instances of it in prod (because no vendor api users), and it errors only when we render out the audit trail — currently (but only just) possible only in Support.

## Guidance to review

Give this a thorough kicking on Heroku.

## Link to Trello card

https://trello.com/c/hJWf3p1y/188-migrate-to-zeitwerk-so-we-can-keep-up-with-viewcomponent

## Things to check


tvararu:
Sounds like quite the yak shave!

tvararu:
The Heroku app seems to fail:

```
Zeitwerk::NameError: expected file /app/app/components/previews/support_interface/tile_component_preview.rb to define constant Previews::SupportInterface::TileComponentPreview, but didn't
```

duncanjbrown:
So it seems Zeitwerk wants to autoload everything under `/app` and unlike the `:classic` autoloader it brooks no deviation from `path = module` and `file = class`. I set `config.eager_load = true` in dev to reproduce the bug and fixed it by following the advice in the ActionView::Component docs to move the previews under `/spec`.

I found the issue with previews, which is that the `previews_path` had a leading slash appended when we fixed up Rubocop recently: no previews found. Fixed that here and previews now work 🎉 

One more thing I just remembered to check is that no Sidekiq workers were renamed as part of this — more potential for mismatch-on-deserialization — but they aren't so I think we're ok.

duncanjbrown:
This change went in with #1937

PR #1940, opened 2020-04-24
Send an email when you log into support with a new device

tijmenb:
## Context

We'd like to make the Support UI more secure, eventually using 2FA. This is a step towards that:  on first login in we send users a notification and we save a long-lived cookie with the User ID. On subsequent logins, we check if the user ID matches. If it does, we don't send a new confirmation email. The cookie is valid forever.

This hooks into the DfE Signin flow to send the email. It also adds a preview for the email itself.

Screenshot 2020-04-24 at 11 50 22

## Guidance to review

Will this work?

## Link to Trello card

https://trello.com/c/bUJp6Bf5

## Things to check

### DIFF
--- a/app/controllers/dfe_sign_in_controller.rb
+++ b/app/controllers/dfe_sign_in_controller.rb
@@ -10,6 +10,11 @@ class DfESignInController < ActionController::Base
     if @local_user
       DsiProfile.update_profile_from_dfe_sign_in(dfe_user: @dfe_sign_in_user, local_user: @local_user)
       @local_user.update!(last_signed_in_at: Time.zone.now)
+
+      if @local_user.is_a?(SupportUser)
+        send_support_sign_in_confirmation_email
+      end
+
       redirect_to @target_path ? session.delete('post_dfe_sign_in_path') : default_authenticated_path
     else
       DfESignInUser.end_session!(session)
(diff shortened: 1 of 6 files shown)


app/views/support_mailer/confirm_sign_in.text.erb
```diff
@@ -0,0 +1,12 @@
...
+# New sign in to Support for Apply for teacher training
+
+We detected you signed into Support for Apply on a new device.
+
+IP address: <%= @device[:ip_address] %>
+Browser: <%= @device[:user_agent] %>
+
+If this was you, you can ignore this email.
+
+If this wasn't you, contact the development team immediately.
```
benilovj, line 12:
```suggestion
If this wasn’t you, contact the development team immediately.
```

app/controllers/dfe_sign_in_controller.rb
```diff
@@ -25,6 +30,27 @@ def callback
 
 private
 
+  def send_support_sign_in_confirmation_email
+    return unless FeatureFlag.active?('support_sign_in_confirmation_email')
+
+    return if cookies.signed[:sign_in_confirmation] == @local_user.id
```
duncanjbrown, line 36: Should we namespace this cookie as `support_sign_in_confirmation` for explicitness?

PR #1941, opened 2020-04-24
Add system spec for sync provider from find

davidgisbey:
## Context

The sync has some complex behaviour that's worth surfacing and making more obvious to any developers who have to work on it. One way of doing that is with a system spec.

Write a spec that covers how the sync handles:

- A site no longer being available for a given course between syncs.
- A course study_mode changing between syncs, and how this affects course options created in the first sync.

- Add tests demonstrating the above behviour

## Guidance to review

Is this sitting in the right place? I'm not 100% sure where it should go?

## Link to Trello card

https://trello.com/c/adHlQT3h/1249-dev-add-system-spec-for-find-sync

## Things to check

### DIFF
--- a/spec/system/sync_provider_from_find/site_is_deleted_spec.rb
+++ b/spec/system/sync_provider_from_find/site_is_deleted_spec.rb
@@ -1,0 +1,74 @@
+require 'rails_helper'
+
+RSpec.describe 'Sync from find' do
+  include FindAPIHelper
+
+  scenario 'a site is removed between syncs' do
+    given_there_is_a_course_on_find_with_multiple_sites
+
+    when_sync_provider_from_find_is_called
+    then_the_correct_course_options_are_created_on_apply
+
+    when_find_says_that_a_site_is_no_longer_listed_for_that_course
+    and_sync_provider_from_find_is_called
+    then_the_course_option_for_that_site_is_deleted
+
+    given_there_is_a_course_on_find_with_multiple_sites
+    and_sync_provider_from_find_has_been_called
+    and_raven_can_capture_exceptions
+
+    when_find_says_that_a_site_is_no_longer_listed_for_that_course
+    and_that_site_is_part_of_an_application
+    and_sync_provider_from_find_is_called
+    then_the_affected_course_options_invalidated_by_find_attribute_is_set_to_true
+    and_raven_captures_an_exception
+  end
+
+  def given_there_is_a_course_on_find_with_multiple_sites
+    stub_find_api_provider_200_with_multiple_sites(provider_name: 'ABC College', provider_code: 'ABC', study_mode: 'full_time')
+  end
+
+  def when_sync_provider_from_find_is_called
+    SyncProviderFromFind.call(provider_name: 'ABC College', provider_code: 'ABC', sync_courses: true)
+  end
+
+  def then_the_correct_course_options_are_created_on_apply
+    @provider = Provider.find_by!(code: 'ABC')
+
+    expect(@provider.courses.first.course_options.count).to eq 2
+  end
+
+  def when_find_says_that_a_site_is_no_longer_listed_for_that_course
+    stub_find_api_provider_200(provider_name: 'ABC College', provider_code: 'ABC', study_mode: 'full_time')
+  end
+
+  def then_the_course_option_for_that_site_is_deleted
+    expect(@provider.courses.first.course_options.count).to eq 1
+  end
+
+  def and_sync_provider_from_find_has_been_called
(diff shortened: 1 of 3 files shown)


spec/system/sync_provider_from_find/site_is_deleted_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    given_a_site_is_deleted_which_is_not_associated_with_an_application
+
+    when_sync_provider_from_find_is_called
+    then_the_course_option_for_that_site_is_deleted
+
+    given_there_is_a_course_on_find_with_multiple_sites
+    and_sync_provider_from_find_has_been_called
+    and_raven_can_capture_exceptions
+
+    when_a_site_is_deleted_which_is_associted_to_an_application
```
tvararu, line 21:
```suggestion
    when_a_site_is_deleted_which_is_associated_to_an_application
```

duncanjbrown:
Love this. Should we document what happens when a part-time course option which has application choices associated with it sees its parent course go to full-time only? Does that option become "invalidated by Find" too?

spec/system/sync_provider_from_find/site_is_deleted_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    SyncProviderFromFind.call(provider_name: 'ABC College', provider_code: 'ABC', sync_courses: true)
+  end
+
+  def then_the_correct_course_options_are_created_on_apply
+    @provider = Provider.find_by!(code: 'ABC')
+
+    expect(@provider.courses.first.course_options.count).to eq 2
+  end
+
+  def given_a_site_is_deleted_which_is_not_associated_with_an_application
```
malcolmbaig, line 41: I think we should describe this slightly differently - current wording is a little ambiguous about whether we're referring to an apply Site record or a site returned in the Find API call.

Maybe something like `when_find_says_that_a_site_is_no_longer_listed_for_that_course` ?

spec/system/sync_provider_from_find/site_is_deleted_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    when_sync_provider_from_find_is_called
+    then_the_course_option_for_that_site_is_deleted
+
+    given_there_is_a_course_on_find_with_multiple_sites
+    and_sync_provider_from_find_has_been_called
+    and_raven_can_capture_exceptions
+
+    when_a_site_is_deleted_which_is_associated_to_an_application
+    and_sync_provider_from_find_is_called
+    then_the_affected_course_options_invalidated_by_find_attriibute_is_set_to_true
```
malcolmbaig, line 23: typo on attribute

davidgisbey:
Currently, a course option's `invalidated_by_find` is not set to true if the parent course's study mode changes. The `invalidated_by_find` attribute is only used when a site has been deleted. When a study mode changes, invalidating a  course the vacancy status is set to 'no vacancies'.

spec/system/sync_provider_from_find/site_is_deleted_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  def then_the_course_option_for_that_site_is_deleted
+    expect(@provider.courses.first.course_options.count).to eq 1
+  end
+
+  def and_sync_provider_from_find_has_been_called
+    when_sync_provider_from_find_is_called
+  end
+
+  def when_a_site_is_deleted_which_is_associated_to_an_application
```
malcolmbaig, line 53: Same wording observation as above, and perhaps move the application choice into a separate step.

So maybe:
```
when_find_says_that_a_site_is_no_longer_listed_for_that_course
and_that_site_is_part_of_an_application
```

spec/system/sync_provider_from_find/study_mode_changes_spec.rb
```diff
@@ -0,0 +1,58 @@
+require 'rails_helper'
+
+RSpec.describe 'Sync from find' do
+  include FindAPIHelper
+
+  scenario 'a courses study mode changes between syncs' do
+    given_there_is_a_full_time_course_on_find
+
+    when_sync_provider_from_find_is_called
+    then_the_correct_course_option_are_created_on_apply
```
malcolmbaig, line 10: `_is_created`?

malcolmbaig:
Nicely done 👌 . Some small observations but overall LGTM.

spec/system/sync_provider_from_find/site_is_deleted_spec.rb
davidgisbey: Thanks for all the feedback

PR #1950, opened 2020-04-27
Provider user adds note flow

mnacos:
### Context

User research has identified a need for notes within the provider UI. This feature will give provider users the ability to write and attach notes to a specific application (application choice) and the ability to review notes within the single application view or its timeline.

Each note will have timestamps, title, body text and an author. They will be added via a form accessible from the single application view.

A couple of PRs have been merged into master already to provide a database-backed `Note` model and create the structure, routes and views for tabbed application view sub-components.

New `new_note` and `create_note` controller actions and routes have been added, scoped similarly to the `notes` action (index) within the context of a single application view.

`:title` in the Note model is renamed to `:subject`, in line with feedback from Content Design.

The 'Notes' tab contains an 'Add Note' button and a list of all notes for this application choice in reverse chronological order.

image

image
`ApplicationTimelineComponent` has been extended to show notes, as well as status change events, in reverse chronological order.

image

### Guidance to review

Ensure the 'notes' feature is active. Navigate to an application within the provider UI (any application state will do). Click on the 'Notes' tab. Add a note. Inspect both the 'Notes' and 'Timeline' tabs.

A system spec,`spec/system/provider_interface/provider_adds_note_to_application_spec.rb`, exercises this 'add note' flow.

### Link to Trello card

[Providers add Notes in UI](https://trello.com/c/mscEDZr5)

### Things to check

- [ ] This code doesn't rely on migrations in the same Pull Request - IT DOES! __but this is a brand new feature, so no data exists and even if it did, we are only changing the name of the column, which preserves existing data__


laurahermionetennant92:
Typeface not quite right: 
Screenshot 2020-04-28 at 11 04 59

mnacos:
> Typeface not quite right:
> Screenshot 2020-04-28 at 11 04 59

Yep, Adam and I have just worked out what it is. Will push an update.

laurahermionetennant92:
Content tweaks:
1. Here, under the h1 'Add note', add 'Notes will appear on the application timeline, labelled by subject line' as hint text.
2. 
Screenshot 2020-04-28 at 11 14 14
Change 'Title' to 'Subject line'

laurahermionetennant92:
Here on the timeline, add 'Note' – so 'Note: Blah by Support User A

Screenshot 2020-04-28 at 11 16 29

duncanjbrown:
Just had a look from the front. 2 minor things:

- If I don't have a name set on my user, there's a free-floating comma in the attribution

Screenshot 2020-04-28 at 11 10 24

- If I enter too many characters into the title field then I can't submit the form. But if I enter too many characters into the Message field I still can

app/components/provider_interface/application_timeline_component.rb
```diff
@@ -27,18 +27,39 @@ def render?
...
+            note.id,
+          )
+        end
+      else
+        []
+      end
+    end
+
+    def events
+      (status_change_events + note_events).sort_by(&:date).reverse
```
duncanjbrown, line 60: 👌

app/components/provider_interface/application_timeline_component.rb
```diff
@@ -27,18 +27,39 @@ def render?
 
   private
 
-    def events
-      changes = FindStatusChangeAudits.new(application_choice: application_choice).call.reverse
+    def status_change_events
+      changes = FindStatusChangeAudits.new(application_choice: application_choice).call
       changes = changes.select { |change| TITLES.has_key?(change.status) }
       changes.map do |change|
         Event.new(
```
duncanjbrown, line 34: What would you think about making this polymorphic now that we have two two kinds of Timeline Entry? We could make a component for each type of entry, and have `events` return an array of `new`'d up components we call `render` on, eg

```ruby
> timeline.events
=> [StatusChangeEventComponent, StatusChangeEventComponent, NoteEventComponent]

timeline.events.each do |e|
  render(e)
end
```

where eg StatusChangeEventComponent looks like this:

```ruby
class StatusChangeEventComponent
  TITLES = {massive hash only relevant to this kind of entry}

  def title
    # deal with the audit log
  end

  def actor
    # deal with the audit log
  end

  def date
    # deal with the audit log
  end
end
```

then audit log concerns remain focussed — we could find a suitable way to reuse them if we ever needed to

app/views/provider_interface/application_choices/notes.html.erb
```diff
@@ -18,4 +18,32 @@
 
 <%= render 'application_choice_header' %>
 
-<% raise 'Notes are wip' %>
+<%= form_with model: @new_note_form,
```
duncanjbrown, line 21: Could the button be a straight link and we don't need a `@new_note_form`?

app/controllers/provider_interface/application_choices_controller.rb
```diff
@@ -35,6 +35,31 @@ def show
 
     def notes
       redirect_to(action: :show) unless FeatureFlag.active?('notes')
+
+      @new_note_form = ProviderInterface::NewNoteForm.new
+      @notes = @application_choice.notes.order('created_at DESC')
+    end
+
+    def new_note
+      redirect_to(action: :show) unless FeatureFlag.active?('notes')
```
duncanjbrown, line 44: `before_action` maybe?

app/controllers/provider_interface/application_choices_controller.rb
```diff
@@ -35,6 +35,31 @@ def show
 
     def notes
       redirect_to(action: :show) unless FeatureFlag.active?('notes')
+
+      @new_note_form = ProviderInterface::NewNoteForm.new
+      @notes = @application_choice.notes.order('created_at DESC')
+    end
+
+    def new_note
```
duncanjbrown, line 43: `application_choices_controller` is losing focus a bit — what do you think about pulling this out into a separate controller?

app/controllers/provider_interface/application_choices_controller.rb
```diff
@@ -35,6 +35,31 @@ def show
...
+
+      @new_note_form = ProviderInterface::NewNoteForm.new
+    end
+
+    def create_note
+      if FeatureFlag.active?('notes')
+        @new_note_form = ProviderInterface::NewNoteForm.new new_note_params
+
+        if @new_note_form.valid?
+          @new_note_form.save
```
duncanjbrown, line 54: Might as well `!` if we're not looking at the return value

app/controllers/provider_interface/application_choices_controller.rb
```diff
@@ -35,6 +35,31 @@ def show
...
+
+          redirect_to provider_interface_application_choice_notes_path(@application_choice)
```
duncanjbrown, line 56: Add a flash?

app/lib/test_applications.rb
```diff
@@ -204,6 +208,26 @@ def confirm_enrollment(choice)
     end
   end
 
+  def add_note(choice)
+    provider_user = choice.provider.provider_users.first
+    provider_user ||= add_provider_user_to_provider(choice.provider)
+    as_provider_user(choice) do
+      @time = @time + rand(-5..5).days
```
duncanjbrown, line 215: Could we reuse (or make some extensions to) `fast_forward` here?

laurahermionetennant92:
@mnacos see below – 
'Title' > 'Subject'
'Message' > 'Note'




**Add note**

Subject
[hint]
When you add a note, its subject appears on your application list by the relevant candidate. 

Note




Screenshot 2020-04-28 at 15 03 52

adamsilver:
From Slack:

1. The h1 should not have the margin bottom override class
2. The hint should go under the subject field and read ‘This appears on your application list by the relevant candidate’ (without full stop as it is a hint)

app/controllers/provider_interface/application_choices_controller.rb
mnacos: Changed this to
```ruby
if @new_note_form.save
```
as #save includes a valid?

app/components/provider_interface/application_timeline_component.rb
mnacos: I think this is a great idea but will group these changes together with the attribution tweaks you and Adam has suggested in another PR.

app/lib/test_applications.rb
mnacos: `fast_forward` mainly deals with the timestamps of audited events. I've changed the code to use `travel_to`, which is used elsewhere, too.

app/controllers/provider_interface/application_choices_controller.rb
mnacos: The main reason not to is the routes are scoped to the application choice and the index view for notes requires the subnavigation components for the application's tabs. It feels a little weird moving e.g. `new_note` and `create_note` to a separate controller while the index view remains within `application_choices_controller`. Ideas?

duncanjbrown:
Teensy UI consideration, probably not for this PR: when I click "Note added" in the Timeline, I return to the notes index with an anchor, which is great. But the page is short enough (and probably will be, up to a dozen or so notes) that I can't tell which note I've clicked on. Should we highlight the anchored note or sth? I am not a designer.

app/views/provider_interface/application_choices/new_note.html.erb
```diff
@@ -0,0 +1,23 @@
...
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with model: @new_note_form,
+          url: provider_interface_application_choice_create_note_path(@application_choice.id),
+          method: :post do |f| %>
+
+      <%= f.govuk_error_summary %>
+
+      <h1 class="govuk-heading-xl">Add note</h1>
+
+      <% subject_hint = 'This appears on your application list by the relevant candidate' %>
```
duncanjbrown, line 15: Is this actually true as of this PR?

db/migrate/20200428163027_rename_title_to_subject_in_notes.rb
```diff
@@ -0,0 +1,7 @@
+class RenameTitleToSubjectInNotes < ActiveRecord::Migration[6.0]
+  def change
+    change_table :notes do |t|
+      t.rename :title, :subject
```
duncanjbrown, line 4: So glad we had this conversation now rather than later! 🎉

app/views/provider_interface/application_choices/new_note.html.erb
mnacos: No, but @adamsilver  requested it.

db/migrate/20200428163027_rename_title_to_subject_in_notes.rb
mnacos: > Teensy UI consideration, probably not for this PR: when I click "Note added" in the Timeline, I return to the notes index with an anchor, which is great. But the page is short enough (and probably will be, up to a dozen or so notes) that I can't tell which note I've clicked on. Should we highlight the anchored note or sth? I am not a designer.

I think @adamsilver is planning something for this. I've looked into highlighting the relevant note in the view, but rails cannot retrieve the anchor part of the url, so it'd need to be something javascript-driven. Probably best tackled when the designs are revised.

adamsilver:
> Teensy UI consideration, probably not for this PR: when I click "Note added" in the Timeline, I return to the notes index with an anchor, which is great. But the page is short enough (and probably will be, up to a dozen or so notes) that I can't tell which note I've clicked on. Should we highlight the anchored note or sth? I am not a designer.

The timeline should not be linking off to anywhere else. It should be read only at the moment for all events.

app/components/provider_interface/application_timeline_component.rb
```diff
@@ -7,7 +7,7 @@ def initialize(application_choice:)
       @application_choice = application_choice
     end
 
-    Event = Struct.new(:title, :actor, :date)
+    Event = Struct.new(:title, :actor, :date, :type, :id)
```
duncanjbrown, line 10: Do we need `:id` any more?

laurahermionetennant92:
This looks good @mnacos !
Thank you

app/components/provider_interface/application_timeline_component.rb
mnacos: Adam is planning a route along the lines of `/notes/:id` for linking and possiblly editing or deleting a specific note. So, these fields will be needed for the links then. I can remove these for now.

PR #1951, opened 2020-04-27
Support interface for validation error tracking

stevehook:
## Context

We need to see more about the validation errors our users are encountering so that we can learn more about their behaviour and better understand areas where our service appears to repeatedly 'fail'.

This PR follows on from https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1922 and adds a front-end to the support interface for viewing validation errors.

- Link from existing Performance page to a new validation index page. This summarise the errors logged grouped by form object or model (the thing that's validated).
image

- Detail page for individual form objects/models listing all the validation errors for that object.
image

- Pagination on the detail page because the number of validation errors is unlimited. 

## Guidance to review

- Does it make sense to group validation errors by form object or model? We could also group by request path or not at all.
- There is no pagination on the index page because the number of items it lists is limited to the number of validatable form objects and models in the application.
- The entries on the detail page are laid out in a table. It's getting pretty congested - some of the columns are too narrow. So would like to get a better solution for this.
- For the candidate that triggered the change I'm just displaying the candidate email address. It may be a lot more useful to have the candidate name and reference but those are attributes of the application form (of which there could be more than one per candidate). A link to the application form would also be nice. So should we also be capturing the application form as well as the candidate?

## Link to Trello card

## Things to check

### DIFF
--- a/app/views/support_interface/validation_errors/index.html.erb
+++ b/app/views/support_interface/validation_errors/index.html.erb
@@ -1,7 +1,7 @@
 <ul>
   <% @grouped_counts.each do |form_object, count| %>
     <li>
-      <a href='<%= candidate_interface_validation_error_path(form_object) %>'><%= form_object %></a><%= "(#{count})" %>
+      <a href='<%= support_interface_validation_error_path(form_object) %>'><%= form_object %></a><%= "(#{count})" %>
     </li>
   <% end %>
 </ul>
--- a/app/views/support_interface/validation_errors/index.html.erb
+++ b/app/views/support_interface/validation_errors/index.html.erb
@@ -1,0 +1,7 @@
+<ul>
+  <% @grouped_counts.each do |form_object, count| %>
+    <li>
+      <a href='<%= candidate_interface_validation_error_path(form_object) %>'><%= form_object %></a><%= "(#{count})" %>
+    </li>
+  <% end %>
+</ul>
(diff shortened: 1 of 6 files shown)


app/views/support_interface/validation_errors/index.html.erb
```diff
@@ -0,0 +1,22 @@
...
+    <tr class="govuk-table__row">
+      <th class="govuk-table__header govuk-table__header govuk-!-width-three-quarters">Form or model</th>
+      <th class="govuk-table__header govuk-table__header govuk-!-width-one-quarter">Error count</th>
+    </tr>
+  </thead>
+  <tbody class="govuk-table__body">
+  <% @grouped_counts.each do |form_object, count| %>
+    <tr class="govuk-table__row">
+      <td class="govuk-table__cell">
+        <a href='<%= support_interface_validation_error_path(form_object) %>'><%= form_object %></a>
```
tijmenb, line 14:
```suggestion
        <%= govuk_link_to form_object, support_interface_validation_error_path(form_object) %>
```

spec/system/support_interface/validation_errors_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    @validation_error = create(
+      :validation_error,
+      form_object: 'CandidateInterface::DegreeForm',
+      details: {
+        award_year: { value: '2222', messages: ['Enter a year before 2022'] },
+      },
+      request_path: '/candidate/application/degrees/51/year',
+      created_at: Time.zone.local(2020, 4, 24, 12, 35, 42),
+      user: @candidate,
+    )
```
tijmenb, line 35: We could consider going into the signup form and trigger an error for the full end-to-end experience.

PR #1953, opened 2020-04-27
Update candidate_sign_in_url helper to return magic link 

malcolmbaig:
## Context

This helper is used by several CandidateMailer emails. Currently it
returns the URL for the sign in page, with an encrypted candidate id in
the params. This simply redirects candidates to an 'expired token' page
from where they have to click a button to get a magic link.

Improve this experience by changing the helper to return a magic
link instead, ie - an authentication endpoint with the encrypted candidate id
and a sign in token in the params. This allows the candidate to
authenticate immediately if they click the link before token expiry.
It's then only after token expiry that they see the 'expired token'
page.
## Guidance to review

- Submitting an application is probably the easiest way to see one of the CandidateMailer emails. 

## Link to Trello card

https://trello.com/c/ap98xn4c
## Things to check

### DIFF
--- a/app/controllers/candidate_interface/sign_in_controller.rb
+++ b/app/controllers/candidate_interface/sign_in_controller.rb
@@ -4,12 +4,8 @@ module CandidateInterface
     before_action :redirect_to_application_if_signed_in, except: %i[authenticate]
 
     def new
-      if params[:u]
-        redirect_to candidate_interface_expired_sign_in_path(u: params[:u])
-      else
-        candidate = Candidate.new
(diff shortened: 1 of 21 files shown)


malcolmbaig:
Some of these mailer unit tests need reworking, closing temporarily.

tvararu:
Screenshot 2020-04-28 at 14 21 50

Sign in still works fine on the review app. However, I think we can remove the `u=` query string parameter, because we now have better token storage; we could have done this longer ago.

PR #1956, opened 2020-04-27
Better expose unavailable courses in Support

tijmenb:
## Context

When a course becomes unavailble, we show this to candidates (https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1841).

This exposes the same information in the support UI - but for applications in all states, not just when they're unsubmitted.

It also extracts a component to make the change easier, and adds the missing "study mode" attribute to the table.

### Before

image

### After

It looks like this when *everything* is wrong (this will never happen in practice):

image

## Link to Trello card

https://trello.com/c/fRAKJt36

## Things to check

### DIFF
--- a/app/components/support_interface/application_choice_component.html.erb
+++ b/app/components/support_interface/application_choice_component.html.erb
@@ -1,0 +1,43 @@
+<% course_name_and_status = capture do %>
+  <%= govuk_link_to(
+    application_choice.offered_course.name_and_code,
+    support_interface_course_path(application_choice.offered_course)
+  ) %>
+
+  <% if application_choice.course_not_available? %>
+    <%= render TagComponent.new(text: 'Course removed from Find', type: 'yellow') %>
+  <% end %>
+
+  <% if application_choice.course_closed_on_apply? %>
+    <%= render TagComponent.new(text: 'Course closed on Apply', type: 'yellow') %>
+  <% end %>
+
+  <% if application_choice.course_full? %>
+    <%= render TagComponent.new(text: 'Course is full', type: 'yellow') %>
+  <% end %>
+<% end %>
+
+<% location_and_status = capture do %>
+  <%= application_choice.offered_option.site.name_and_code %>
+
+  <% if application_choice.chosen_site_full? %>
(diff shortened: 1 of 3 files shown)


app/components/support_interface/course_choices_table_component.html.erb
```diff
@@ -11,7 +11,13 @@
     <% course_rows.each do |course_option| %>
       <tr class='govuk-table__row'>
         <td class='govuk-table__cell'><%= course_option.id %></td>
-        <td class='govuk-table__cell'><%= govuk_link_to(course_option.course.name_and_code, support_interface_course_path(course_option.course)) %> - <%= course_option.study_mode.humanize %> at <%= course_option.site.name %></td>
+        <td class='govuk-table__cell'>
+          <% if course_option.invalidated_by_find? %>
+            <%= render TagComponent.new(text: 'Site deleted', type: 'red') %>
```
tijmenb, line 16: @malcolmbaig should this read "Course no longer offered at this site"?

app/views/support_interface/providers/sites.html.erb
```diff
@@ -9,9 +9,15 @@
...
-    <% @provider.sites.each do |site| %>
+    <% @provider.sites.includes(:course_options).each do |site| %>
       <tr class='govuk-table__row'>
-        <td class='govuk-table__cell'><%= site.name_and_code %></td>
+        <td class='govuk-table__cell'>
+          <%= site.name_and_code %>
+
+          <% if site.deleted_from_find? %>
+            <%= render TagComponent.new(text: 'Deleted in Find', type: 'red') %>
+          <% end %>
```
tijmenb, line 19: This is just wrong, the site isn't actually deleted AFAWK, just that there are courses that used to be offered at the site, that no longer are.

app/components/support_interface/course_choices_table_component.html.erb
malcolmbaig: Yes, with the concept of 'sites being deleted on Find' handled separately pending confirmation that this is actually a thing.

PR #1959, opened 2020-04-27
Move mailer previews under /spec

duncanjbrown:
Zeitwerk will try to eager load anything under `app/` when we're in production, including these, resulting in:

```
expected file /Users/duncan/projects/apply-for-postgraduate-teacher-training/app/mailers/previews/referee_mailer_preview.rb to define constant Previews::RefereeMailerPreview
```

https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#autoload-paths

Official Rails guidance since 6 (when Zeitwerk was introduced) is to put them under `test/` or in our case `spec/` — not eager loaded in prod.

For some reason having `show_previews` enabled hides this error. That has the effect of adding `app/mailers/previews` to the autoload_path. Shouldn’t it fail faster then, though? Possibly adding `app/mailers`.

This change will fix the build — mailer previews are still broken, though.

## Context

Staging failed to deploy.

This is a good place for it to fail because it's a production-like environment where nothing like `Rails.env.development? || HostingEnvironment.qa? || HostingEnvironment.review?` will return true.


duncanjbrown:
Conjecture about autoload paths being root namespaces for Zeitwerk — ie if mailers/previews is on autoload path then its contents are not expected to be namespaced under Preview — seems to be correct: https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#concerns

PR #1961, opened 2020-04-28
Rename CourseOption#invalidated_by_find

malcolmbaig:
## Context

This field name is used to indicate whether the site/course relationship
described by a course option still exists, as determined by the Find API
sync. The name is a little ambiguous, so for clarity rename it to
site_still_valid.
- Database migrations to rename the column and invert its boolean values
- Code changes to use the new field

## Guidance to review
I considered a zero-downtime approach which would have involved three stages (in three separate deploys):

- Add a new column and ensure we're writing to both on any updates.
- Migrate the data and stop writing to the old column.
- Remove the old column.

This is obviously quite time-consuming so given our low user numbers, I opted for a single change which prioritises speed over stability. Areas that would potentially error during deploy are the Sync and course selection.

Happy to hear other opinions on this and open to reworking into three steps if considered worthwhile.

## Link to Trello card
https://trello.com/c/IjTL2YNR

## Things to check


malcolmbaig:
Chatted with @tvararu about this and on second thought let's phase this in. BRB.

PR #1962, opened 2020-04-28
Docker build performance improvments

tpinney:
## Context

The build pipeline duration has steadily been increasing in duration and this PR aims to stem some of that creep in duration with optimisations to the build pipeline stage.

Two key introductions:
1. Added BuildKit - This feature allows parallel building of stages in a multi stage Docker image. This option can be turned on/off by setting the `dockerBuildkitState` variable to 1/0 respectively in the pipeline variables. This is configured in the `APPLY - Shared Variables` variable group in AZDO. Enabling this feature knocks approximately two minutes off the build.
1. ~~Added Docker image caching feature - This feature is configurable by setting the `dockerCacheEnable` variable to true/false in the `APPLY - Shared Variables` variable group in AZDO. This feature when enabled will save the docker image and its intermediate layers to a file stored in an Azure pipeline cache which can then be loaded at the start of subsequent builds to save having to install the gems and run Bundler unnecessarily every time on a given branch. The cache will only update if any of the `Dockerfile`, `Gemfile` or `package.json` files change. Local testing showed this to save as much as another three minutes from the build pipeline. At present this feature is currently disabled because AZDO agents are configured to immediately remove all intermediate layers when a build completes meaning caching can't be leveraged on subsequent runs at present, however should this change the feature can easily be turned on.~~
1. Upversion Bundler to 2.1.4 to remove a warning that was present in the builds
1. Optimised the Dockerfile to remove unnecessary levels of layers from the build cache.

## Guidance to review

Builds from this branch can be seen here: https://dfe-ssp.visualstudio.com/Become-A-Teacher/_build?definitionId=49&_a=summary&repositoryFilter=37&branchFilter=8449

## Link to Trello card

https://trello.com/c/UWxiLzcl/1190-reduce-duration-of-devops-pipeline-build-stage

## Things to check

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,10 +8,10 @@ ARG DEV_PACKAGES="postgresql-dev git nodejs yarn graphviz ttf-ubuntu-font-family
 ARG RUBY_PACKAGES="tzdata"
 ARG bundleWithout=""
 
-ENV BUNDLER_VERSION="2.0.2"
-ENV BUNDLE_PATH="/gems"
-ENV BUNDLE_WITHOUT=${bundleWithout}
-ENV WKHTMLTOPDF_GEM=wkhtmltopdf-binary-edge-alpine
+ENV BUNDLER_VERSION="2.1.4" \
+    BUNDLE_PATH="/gems" \
+    BUNDLE_WITHOUT=${bundleWithout} \
+    WKHTMLTOPDF_GEM=wkhtmltopdf-binary-edge-alpine
 
 RUN apk update && \
     apk upgrade && \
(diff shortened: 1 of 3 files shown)


tpinney:
For reference, here is the detail about the layer caching limitation:
https://docs.microsoft.com/en-us/azure/devops/pipelines/ecosystems/containers/build-image?view=azure-devops#is-reutilizing-layer-caching-during-builds-possible-on-azure-pipelines

saliceti:
I'm not comfortable with this PR as it adds complexity to the code but caching can't be used. Let's discuss it.

tpinney:
> I'm not comfortable with this PR as it adds complexity to the code but caching can't be used. Let's discuss it.

If you want me to remove the caching steps I can.

tpinney:
Removed the disabled pipeline caching feature for Docker at @saliceti's request. This has been preserved for posterity in the `pipeline-docker-cache` branch: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/tree/pipeline-docker-cache.

PR #1970, opened 2020-04-29
Candidates can apply again - initial workflow

stevehook:
## Context

We need to allow candidates who have had an unsuccessful (for whatever reason) initial application to create a copy of their application form and then be able to submit it with a new course choice.

- [x] Feature flag `apply_again`
- [x] Link on dashboard to _Apply again_ if application ended without success
- [x] New interstitial page that kicks off the apply again process via _Start now_ button
- [x] `DuplicateApplication` service that creates a copy of the original application minus application choices.
- [x] New application should have a new support reference.

## Guidance to review

- Does the UI look OK?
image

image

image

- Are there any bits of this that could leak out from behind the feature flag? The flag will need to be there for a while after this gets merged because there are a few other PRs that will make up the feature.
- Are the tests sufficient?
- Is the `DuplicateApplication` logic right?

## Link to Trello card

https://trello.com/c/uQeHCQJ2/1390-candidates-can-apply-again-which-copies-the-form-with-temporary-interstitial

## Things to check

### DIFF
--- a/app/services/duplicate_application.rb
+++ b/app/services/duplicate_application.rb
@@ -5,7 +5,7 @@ class DuplicateApplication
     @original_application_form = original_application_form
   end
 
-  IGNORED_ATTRIBUTES = %w[id created_at updated_at submitted_at course_choices_completed phase].freeze
+  IGNORED_ATTRIBUTES = %w[id created_at updated_at submitted_at course_choices_completed phase support_reference].freeze
   IGNORED_CHILD_ATTRIBUTES = %w[id created_at updated_at application_form_id].freeze
 
   def duplicate
--- a/app/services/duplicate_application.rb
+++ b/app/services/duplicate_application.rb
@@ -1,0 +1,43 @@
+class DuplicateApplication
+  attr_reader :original_application_form
+
+  def initialize(original_application_form)
+    @original_application_form = original_application_form
+  end
+
+  def duplicate
+    attrs = original_application_form.attributes.except(*
+      %w[id created_at updated_at submitted_at course_choices_completed phase]
+    ).merge(
+      phase: 'apply_2'
+    )
+
+    new_application_form = ApplicationForm.create!(attrs)
+
+    original_application_form.application_work_experiences.each do |w|
+      new_application_form.application_work_experiences.create!(w.attributes.except(*%w[
+        id created_at updated_at application_form_id
+      ]))
+    end
+
+    original_application_form.application_volunteering_experiences.each do |w|
+      new_application_form.application_volunteering_experiences.create!(w.attributes.except(*%w[
+        id created_at updated_at application_form_id
+      ]))
+    end
+
+    original_application_form.application_qualifications.each do |w|
+      new_application_form.application_qualifications.create!(w.attributes.except(*%w[
+        id created_at updated_at application_form_id
+      ]))
+    end
+
+    original_application_form.references.each do |w|
+      new_application_form.references.create!(w.attributes.except(*%w[
+        id created_at updated_at application_form_id
+      ]))
+    end
+
+    true
+  end
+end
--- a/app/services/duplicate_application.rb
+++ b/app/services/duplicate_application.rb
@@ -38,6 +38,12 @@ class DuplicateApplication
       ]))
     end
 
-    true
+    original_application_form.application_work_history_breaks.each do |w|
+      new_application_form.application_work_history_breaks.create!(w.attributes.except(*%w[
+        id created_at updated_at application_form_id
+      ]))
+    end
+
+    new_application_form
   end
 end
--- a/app/services/duplicate_application.rb
+++ b/app/services/duplicate_application.rb
@@ -5,9 +5,12 @@ class DuplicateApplication
     @original_application_form = original_application_form
   end
 
+  IGNORED_ATTRIBUTES = %w[id created_at updated_at submitted_at course_choices_completed phase].freeze
+  IGNORED_CHILD_ATTRIBUTES = %w[id created_at updated_at application_form_id].freeze
+
   def duplicate
     attrs = original_application_form.attributes.except(
-      *%w[id created_at updated_at submitted_at course_choices_completed phase],
+      *IGNORED_ATTRIBUTES,
     ).merge(
       phase: 'apply_2',
     )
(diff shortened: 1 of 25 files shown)


app/services/duplicate_application.rb
```diff
@@ -0,0 +1,49 @@
...
+      ]))
+    end
+
+    original_application_form.application_qualifications.each do |w|
+      new_application_form.application_qualifications.create!(w.attributes.except(*%w[
+        id created_at updated_at application_form_id
+      ]))
+    end
+
+    original_application_form.application_references.each do |w|
```
tijmenb, line 35: We could consider copying only the references that were actualy provided (with status `feedback_provided`), so we skip the cancelled/rejected ones.

app/services/submit_application.rb
```diff
@@ -24,7 +24,7 @@ def call
 private
 
   def send_reference_request_email_to_referees(application_form)
-    application_form.application_references.includes(:application_form).each do |reference|
+    application_form.application_references.where(feedback: nil).each do |reference|
```
tijmenb, line 27: The presence of feedback doesn't mean the reference was submitted (the submission is a separate step now). This might be more robust:

```suggestion
    application_form.application_references.not_requested_yet.each do |reference|
```

app/views/candidate_interface/application_form/complete.html.erb
```diff
@@ -14,6 +14,14 @@
   Application submitted on <%= submitted_at_date %>. <%= govuk_link_to 'View application', candidate_interface_application_review_submitted_path %>
 </p>
 
+<% if ProcessState.new(@application_form).state == :ended_without_success %>
```
tijmenb, line 17: We'll need to come up with something that doesn't use `ProcessState`, because I don't really trust that code yet.

stevehook: It does the right thing for now but it's easy to see it getting out of sync with `ApplicationStateChange`. 

What would a better thing look like? Would it help if we had a list of `UNSUCCESSFUL_END_STATES` on the `ApplicationStateChange`? That way we could start to remove the references to specific statuses from `ProcessState`. Maybe also `SUCCESSFUL_END_STATES` and `UNFINISHED_STATES`.

tijmenb: Yeah, I think that would be good to introduce. I'd prefer to duplicate some logic from `ProcessState` for now.

app/services/submit_application.rb
```diff
@@ -53,6 +53,10 @@ def email_address_is_a_bot?(reference)
   def submit_application
     application_choices.each do |application_choice|
       SubmitApplicationChoice.new(application_choice).call
+
+      if application_form.apply_again? && application_form.references_complete?
```
tijmenb, line 57: Somehow I can't find where `references_complete?` is defined, is this correct? I think we don't need `application_form.apply_again?` here.

app/views/candidate_interface/course_choices/_guidance.html.erb
```diff
@@ -1,4 +1,3 @@
-<p class="govuk-body">You can apply for up to 3 courses at this stage of your application.</p>
```
tijmenb, line 1: This needs to stay in I think.

spec/system/candidate_interface/candidate_with_unsuccessful_application_applies_again_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+  scenario 'Can apply again' do
+    given_the_pilot_is_open
+    and_apply_again_feature_flag_is_active
+    and_i_am_signed_in_as_a_candidate
+
+    when_i_have_an_unsuccessful_application
+    and_i_visit_the_application_dashboard
+    and_i_click_on_apply_again
+    and_i_click_on_start_now
+
```
tijmenb, line 22: I wonder if this should include the submission, perhaps it would've caught the `references_complete?`  issue?

app/services/submit_application.rb
stevehook: I think I got this muddled up with the spike version. Things have changed a lot since then and I am pretty sure this piece of logic was moved out of this service. I don't think there is any reason for this PR to add this `if` statement so removing it.

spec/system/candidate_interface/candidate_with_unsuccessful_application_applies_again_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+RSpec.feature 'Candidate with unsuccessful application' do
+  include CandidateHelper
+
+  around do |example|
+    date_that_avoids_clocks_changing_by_ten_days = Time.zone.local(2020, 1, 13)
+    Timecop.freeze(date_that_avoids_clocks_changing_by_ten_days) do
+      example.run
+    end
+  end
```
tvararu, line 11: Do you need this? There are no time-based assertions in this test. 🤔

spec/system/candidate_interface/candidate_with_unsuccessful_application_applies_again_spec.rb
tvararu: Could be done as a follow-up on the same card, but yeah, we should test submission 👍

app/services/duplicate_application.rb
```diff
@@ -0,0 +1,52 @@
...
+      new_application_form.application_references.create!(
+        w.attributes.except(*IGNORED_CHILD_ATTRIBUTES),
+      )
+    end
+
+    original_application_form.application_work_history_breaks.each do |w|
+      new_application_form.application_work_history_breaks.create!(
+        w.attributes.except(*IGNORED_CHILD_ATTRIBUTES),
+      )
+    end
```
tvararu, line 48: These could be DRYer but in this situation I think explicit is preferable. 👍

app/services/submit_application.rb
tijmenb: I think the reason for the if-statement was that we need to progress the application to `application_complete` if there are already 2 references. Otherwise the application is stuck in `awaiting_references`. One for the follow up that introduces the submission test.

app/views/candidate_interface/application_form/start_apply_again.html.erb
```diff
@@ -0,0 +1,23 @@
+<h1 class="govuk-heading-xl">Do you want to apply again?</h1>
```
fofr, line 1: Looks like this page might be missing a page title (`<title>`).

stevehook: thanks @fofr - I've added a `title` in a follow up PR https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1995

PR #1975, opened 2020-04-29
Send Slack message to support when course becomes unavailable

tijmenb:
## Context

We currently get a Sentry error in `#twd_apply_tech` when a location is no longer available for a course. We then find out which people are affected and tell support to contact them. This tells the support team directly. Together with https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1974 will make it no longer necessary to involve a developer. 

Send a Slack message instead of Sentry error.

## Guidance to review

Does the message make sense?

## Link to Trello card

https://trello.com/c/fRAKJt36

## Things to check


malcolmbaig:
Nice 👌

PR #1976, opened 2020-04-29
Return the correct value from request.remote_ip on Azure

duncanjbrown:
`request.remote_ip` returns private Azure IPs in production, which means it's useless.

Following investigation https://ukgovernmentdfe.slack.com/archives/CN1MCQCHZ/p1588093070402800 we found out that the format Azure uses for IPs in the `X-Forwarded-For` header isn't compatible with the Rails code which tries to parse those IPs and they end up being discarded.

`Rack::Request` does the right thing. Use it to parse and rewrite the `HTTP_X_FORWARDED_FOR` value ahead of time so `ActionDispatch::Request` gets IPs it can deal with.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -112,4 +112,32 @@ Rails.application.configure do
   HostingEnvironment.authorised_hosts.each do |host|
     config.hosts << host
   end
+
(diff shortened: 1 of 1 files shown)


tvararu:
Love the comment and investigation that went into this

PR #1982, opened 2020-04-30
Add index to provider_users_provider table

willmcb:
Adds uniqueness index on provider_id and provider_user_id:

`index_provider_users_providers_by_provider_and_provider_user`

## Context

We perform a lot of operations on provider permissions, this index will help with speed and data integrity.

## Guidance for review

- we don't think there are any existing records that would be impacted by this, but we will test in QA to ensure that users aren't adversely affected. 

## Link to Trello card

https://trello.com/c/BTWwpO0Y/2004-providers-with-the-manageusers-permission-can-edit-users-associated-with-their-organisation

## Things to check


duncanjbrown:
Code looks good, failing spec

db/schema.rb
```diff
@@ -273,9 +273,10 @@
   create_table "provider_users_providers", force: :cascade do |t|
```
tijmenb, line 273: I would expect the schema number to have been updated in this PR. Did you run the migration locally?

willmcb: I have run the migrations locally and nothing changed on my branch.

tijmenb: Ah right, the number on this migration is lower than the last merged migration. 🚀

steventux: @tijmenb This PR sat around for a while because it couldn't be merged until after https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1931 and was then rebased, so I think the latest schema timestamp/number has been preferred over the orignal. The schema file does contain the added index though.

PR #1991, opened 2020-05-01
Show recent provider note subjects on applications list

mnacos:
## Context

We need to add this:
image

Adding `most_recent_note` to `ProviderInterface::ApplicationCardComponent` and relevant styles.

## Guidance to review

If an application choice has any notes, the most recent one should appear on applications list, just below the status tag.

## Link to Trello card

[Note should appear on the application list page](https://trello.com/c/ai0z23Ma)

## Things to check

### DIFF
--- a/app/components/provider_interface/application_card_component.html.erb
+++ b/app/components/provider_interface/application_card_component.html.erb
@@ -4,6 +4,10 @@
     <%= govuk_link_to candidate_name, provider_interface_application_choice_path(application_choice) %>
     <div class="app-application-card__status">
       <%= render ProviderInterface::ApplicationStatusTagComponent.new(application_choice: application_choice) %>
+      <% if most_recent_note %>
+        <div class="govuk-!-margin-top-0 govuk-!-margin-bottom-0"></div>
+        <p class="app-application-card__note"><%= most_recent_note.subject %></p>
+      <% end %>
     <div>
   </h3>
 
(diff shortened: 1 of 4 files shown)


adamsilver:
@mnacos when I signed in it says the account is not ready yet.

app/components/provider_interface/application_card_component.html.erb
```diff
@@ -4,6 +4,10 @@
     <%= govuk_link_to candidate_name, provider_interface_application_choice_path(application_choice) %>
     <div class="app-application-card__status">
       <%= render ProviderInterface::ApplicationStatusTagComponent.new(application_choice: application_choice) %>
+      <% if most_recent_note %>
+        <div class="govuk-!-margin-top-0 govuk-!-margin-bottom-0"></div>
```
adamsilver, line 8: You probably took this from the prototype but the prototype isn’t perfectly clean. it would be better if this div wasn’t here.

mnacos: Fixed

app/frontend/styles/_notes.scss
```diff
@@ -1,3 +1,34 @@
...
+.app-application-card__note {
+  position: relative;
+  padding: 1px 0px 1px 15px;
+  margin-top: 10px;
+  margin-bottom: 5px;
+  color: $govuk-text-colour;
+  line-height: 1.25;
+  font-weight: 400;
+  font-size: 1rem;
+  -webkit-font-smoothing: antialiased;
```
tvararu, line 16: Is this necessary? 🤔 We don't use this rule anywhere else in the app.

mnacos: It doesn't seem to make much difference, dropping.

PR #1993, opened 2020-05-01
Fix URLs for DSI fallback Slack notifications

tvararu:
:shipit:

### DIFF
--- a/app/controllers/provider_interface/sessions_controller.rb
+++ b/app/controllers/provider_interface/sessions_controller.rb
@@ -27,7 +27,7 @@ module ProviderInterface
 
         SlackNotificationWorker.perform_async(
           "Provider user #{provider_user.first_name} has requested a fallback signin link",
-          edit_support_interface_provider_user_path(provider_user),
+          edit_support_interface_provider_user_url(provider_user),
         )
       end
     end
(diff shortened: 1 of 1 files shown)


app/controllers/provider_interface/sessions_controller.rb
```diff
@@ -27,7 +27,7 @@ def sign_in_by_email
 
         SlackNotificationWorker.perform_async(
           "Provider user #{provider_user.first_name} has requested a fallback signin link",
-          edit_support_interface_provider_user_path(provider_user),
+          Rails.application.routes.url_helpers.edit_support_interface_provider_user_path(provider_user),
```
tijmenb, line 30: I think the link helper works allright, but it should just be url instead of path.

```suggestion
          edit_support_interface_provider_user_url(provider_user),
```

app/controllers/provider_interface/sessions_controller.rb
```diff
@@ -38,7 +38,7 @@ def authenticate_with_token
 
       SlackNotificationWorker.perform_async(
         "Provider user #{provider_user.first_name} has signed in via the fallback mechanism",
-        edit_support_interface_provider_user_path(provider_user),
+        Rails.application.routes.url_helpers.edit_support_interface_provider_user_path(provider_user),
```
tijmenb, line 41:
```suggestion
        edit_support_interface_provider_user_url(provider_user),
```

app/controllers/provider_interface/sessions_controller.rb
tvararu: I found the verbose version in a different controller, so I copied it over, but you're right. The only place these don't work OOTB I think is inside component `.rb` files.

PR #1995, opened 2020-05-01
Extend system spec for apply again to cover application submission

stevehook:
## Context

This PR is a follow up to https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1970 to improve tests and fix an issue with the Apply Again feature.

- [x] Extend system spec for starting an Apply Again application to include submission
- [x] Add missing title to template

## Guidance to review

- Does the spec make sense?
- Is the title correct?

## Link to Trello card

https://trello.com/c/uQeHCQJ2/1390-candidates-can-apply-again-which-copies-the-form-with-temporary-interstitial

## Things to check

### DIFF
--- a/app/services/submit_application.rb
+++ b/app/services/submit_application.rb
@@ -53,6 +53,18 @@ private
   def submit_application
     application_choices.each do |application_choice|
       SubmitApplicationChoice.new(application_choice).call
+
+      if application_form.apply_again? && enough_references_have_been_provided?
+        ApplicationStateChange.new(application_choice).references_complete!
+      end
     end
+  end
+
+  def enough_references_have_been_provided?
+    application_form
+      .application_references
+      .feedback_provided
+      .uniq
+      .count >= ApplicationForm::MINIMUM_COMPLETE_REFERENCES
   end
 end
--- a/app/services/submit_application.rb
+++ b/app/services/submit_application.rb
@@ -54,7 +54,7 @@ private
     application_choices.each do |application_choice|
       SubmitApplicationChoice.new(application_choice).call
 
-      if application_form.apply_again? && enough_references_have_been_provided?
+      if FeatureFlag.active?('apply_again') && enough_references_have_been_provided?
         ApplicationStateChange.new(application_choice).references_complete!
       end
     end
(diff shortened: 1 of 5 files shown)


spec/system/candidate_interface/candidate_with_unsuccessful_application_applies_again_spec.rb
```diff
@@ -56,4 +68,26 @@ def and_i_click_on_start_now
...
+  end
+
+  def and_submit_the_application
+    click_link 'Check and submit your application'
+    click_link 'Continue'
+    choose 'No' # "Is there anything else you would like to tell us?"
+
+    click_button 'Submit application'
+
+    @application = ApplicationForm.last
```
tijmenb, line 87: Do we need this?

spec/system/candidate_interface/candidate_with_unsuccessful_application_applies_again_spec.rb
```diff
@@ -56,4 +68,26 @@ def and_i_click_on_start_now
...
+  end
+
+  def then_application_is_submitted
+    expect(page).to have_content('Application successfully submitted')
```
tijmenb, line 91: The key thing to test here is whether or not the application is now in the "application_complete" state, since the references are already completed.

app/views/candidate_interface/application_form/start_apply_again.html.erb
```diff
@@ -1,4 +1,6 @@
-<h1 class="govuk-heading-xl">Do you want to apply again?</h1>
+<% content_for :title, t('page_titles.start_apply_again') %>
```
fofr, line 1: :+1:

app/services/submit_application.rb
```diff
@@ -53,6 +53,18 @@ def email_address_is_a_bot?(reference)
   def submit_application
     application_choices.each do |application_choice|
       SubmitApplicationChoice.new(application_choice).call
+
+      if application_form.apply_again? && enough_references_have_been_provided?
```
tijmenb, line 57: I think we could skip checking `apply_again?` here, as it makes sense without it too.

stevehook: You're right I think. I guess I was thinking of limiting the extent of the change but one less special case is better in the long run. But I do think this should strictly speaking be guarded by the feature flag?

PR #1997, opened 2020-05-04
Accept different types of email address input for sign in fallback

tijmenb:
## Context

This allows users to input their email address with extra spaces and mixed-case.

See commits.

## Link to Trello card

https://trello.com/c/t421tPrM/2073-deal-with-may-1st-dfe-signin-outage

## Things to check

### DIFF
--- a/app/controllers/provider_interface/sessions_controller.rb
+++ b/app/controllers/provider_interface/sessions_controller.rb
@@ -18,7 +18,7 @@ module ProviderInterface
     def sign_in_by_email
       raise unless FeatureFlag.active?('dfe_sign_in_fallback')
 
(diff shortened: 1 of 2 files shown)


app/controllers/provider_interface/sessions_controller.rb
```diff
@@ -18,7 +18,7 @@ def destroy
     def sign_in_by_email
       raise unless FeatureFlag.active?('dfe_sign_in_fallback')
 
-      provider_user = ProviderUser.find_by(email_address: params.dig(:provider_user, :email_address).downcase)
+      provider_user = ProviderUser.find_by(email_address: params.dig(:provider_user, :email_address).downcase.strip)
```
steventux, line 21: `params.dig` can return `nil` here so perhaps:

```ruby
params.dig(:provider_user, :email_address)&.downcase.strip
```

PR #1988, opened 2020-05-01
Apply again email content

davidgisbey:
## Context

Now we are building apply again, candidates need to be told about it when all of their options are rejected/withdrawn/declined.

This PR creates new emails for the above situations informing them about apply again.

## Changes proposed in this pull request

- Update the content of the all rejected email
- Create 2 new DBD emails, 1 when they have a rejected course choice and 1 without, informing the candidate of apply2
- Adds an email for when a candidate withdraws an application choice and has none which are successful
- Adds an email for when a candidate declines an application choice and has none which are successful

All rejected 

Before

image

After

image

DBD email - all withdrawn or declined

image

DBD email - all rejected/withdrawn/declined with at least 1 rejection

image

Withdraw email

image

Decline email 

image

## Guidance to review

Use the mailer previews to check the emails have the correct content

content document - https://docs.google.com/document/d/14S_40eWV8MYZnjGoW1Pop4LfdXuMXkrHRlilLpoBDdw/edit

## Link to Trello card

https://trello.com/c/uSMiuUHp/1388-update-email-content-to-allow-candidates-to-apply-again

## Things to check

- [x] This code doesn't rely on migrations in the same Pull Request
- [x] If this code includes a migration adding or changing columns, it also backfills existing records for consistency
- [x] API release notes have been updated if necessary
- [x] New environment variables have been [added to the Azure config](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training#azure-hosting-devops-pipeline)

PR #2000, opened 2020-05-04
Add apply2 emails for withdraw and decline

davidgisbey:
## Context

These are the last two emails that are sent to candidates informing them about Apply2.

They are sent when:

1. A candidate withdraws their final course choice and has none which are successful
2. A candidate declines their final course choice and has none which are successful

- Create a withdraw_last_course_choice email
- Create a decline_last_course_choice email
- Call them in the WIthdraw/Decline application services

Withdraw email

image

Decline email 

image

## Guidance to review

The previous PR for this is  #1988. and the content can be found here
https://docs.google.com/document/d/14S_40eWV8MYZnjGoW1Pop4LfdXuMXkrHRlilLpoBDdw/edit

Use the mailer previews to check the emails have the correct content
 
I'm unsure if how I've disabled bullet is the way to go. It breaks build stubbed if you use the includes method.

## Link to Trello card

https://trello.com/c/uSMiuUHp/1388-update-email-content-to-allow-candidates-to-apply-again

## Things to check


app/services/decline_offer.rb
```diff
@@ -1,15 +1,30 @@
...
   end
+
+private
+
+  def all_course_choices_unsuccessful
+    @application_choices.size == @application_choices.select { |application_choice|
+                                   application_choice.rejected? ||
+                                     application_choice.withdrawn? ||
+                                     application_choice.declined?
+                                 } .size
```
tijmenb, line 28: I believe you can use `ApplicationForm#ended_without_success?`:

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/3fccf306dbc4dd57f0b85f688aa35f7ba3d57aa2/app/models/application_form.rb#L122-L124

That also includes the `cancelled` state that we're missing here.

spec/system/candidate_interface/candidate_declines_offer_spec.rb
```diff
@@ -5,40 +5,60 @@
...
     login_as(@candidate)
   end
 
-  def and_i_have_an_offer
-    application_form = create(
+  def and_the_apply_again_flag_is_on
+    FeatureFlag.activate('apply_again')
+  end
+
+  def and_i_have_multiiple_offers
```
tijmenb, line 36:
```suggestion
  def and_i_have_multiple_offers
```

spec/system/support_interface/docs_spec.rb
```diff
@@ -29,6 +29,8 @@ def then_i_see_the_provider_flow_documentation
 
   def and_it_contains_documentation_for_all_emails
     emails_outside_of_states = %w[
+      candidate_mailer-decline_last_application_choice
+      candidate_mailer-withdraw_last_application_choice
```
tijmenb, line 33: These can go into the `decline` and `withdraw` events in the YAML.

app/services/decline_offer.rb
davidgisbey: Nice. I've reworked the first PR and will sort his now 👍

spec/system/support_interface/docs_spec.rb
davidgisbey: 👍

app/views/candidate_mailer/decline_last_application_choice.text.erb
```diff
@@ -0,0 +1,44 @@
...
+
+Sign in to review feedback and apply again:
+https://www.apply-for-teacher-training.education.gov.uk/candidate/account
+
+# Get help
+
+For advice on becoming a teacher, call Get Into Teaching for free on 0800 389 2500 or chat to an adviser online:
+https://getintoteaching.education.gov.uk/contact
+
+For any technical issues using Apply for teacher training, contact us at [becomingateacher@digital.education.gov.uk](mailto:becomingateacher@digital.education.gov.uk)
```
tvararu, line 42: Can we reuse content from this as a partial? Would shorten this email + others a fair bit.

app/services/decline_offer.rb
```diff
@@ -1,13 +1,18 @@
 class DeclineOffer
   def initialize(application_choice:)
     @application_choice = application_choice
+    @application_choices = application_choice.application_form.application_choices
```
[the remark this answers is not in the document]
davidgisbey: Before Tjimen told me about ended_without_success? i was using it to determine if all the candidates course choices were unsuccessful. It can come out now.

PR #2001, opened 2020-05-04
Modify max-width of applications list to 960px

mnacos:
## Context

Aim of this PR is to fix the width of applications index in the provider_interface. This should be 960px rather than 1220px, which is what it's set to currently.

Adam has pointed out that the root cause for the current width is the inclusion of custom styles, in particular `.app-width-container--wide` and an associated query param `wide` which has crept into various places. Removing all instances of this custom `wide`/`:page_width` logic seems to fix the width of all pages.

## Guidance to review

Should be straightforward. Please inspect all interfaces (candidate, provider and support) when reviewing this code.

## Link to Trello card

[Change application list page width](https://trello.com/c/67mVoJI7)

## Things to check

### DIFF
--- a/app/frontend/styles/_provider.scss
+++ b/app/frontend/styles/_provider.scss
@@ -1,9 +1,3 @@
-@include govuk-media-query($from: wide) {
-  .app-width-container--wide {
-    max-width: 1220px;
-  }
-}
-
 .moj-action-bar {
   overflow: hidden;
 }
--- a/app/components/header_component.html.erb
+++ b/app/components/header_component.html.erb
@@ -1,5 +1,5 @@
 <header class="govuk-header <%= classes %>" role="banner" data-module="govuk-header">
-  <div class="govuk-header__container govuk-width-container<%= wide ? " app-width-container--wide" : "" %>">
+  <div class="govuk-header__container govuk-width-container">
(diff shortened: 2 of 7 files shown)


adamsilver:
According to the screenshot, there the dot for the note is in the wrong position

app/frontend/styles/_provider.scss
```diff
@@ -1,9 +1,13 @@
 @include govuk-media-query($from: wide) {
   .app-width-container--wide {
-    max-width: 1220px;
+    max-width: 960px;
   }
 }
 
+.moj-scrollable-pane:after {
```
adamsilver, line 7: I don't believe this is needed both in css and in html now.

app/frontend/styles/_provider.scss
```diff
@@ -1,9 +1,13 @@
 @include govuk-media-query($from: wide) {
```
adamsilver, line 1: Can kill this now as no need for the override at all now.

app/frontend/styles/_provider.scss
```diff
@@ -1,9 +1,13 @@
...
   .app-width-container--wide {
```
adamsilver, line 2: And should be able to remove this from the html where it's been applied.

PR #2002, opened 2020-05-04
Allow candidates to replace referees when they use Apply Again

stevehook:
## Context

We need to allow candidates to replace referees when they use Apply Again.

The service already meets this requirement (as far as I can tell) so the only change here is a new system spec.

## Guidance to review

Is there something I'm missing? Are there other circumstances in which we should allow referee replacement but do not?

## Link to Trello card

https://trello.com/c/fmJ6jqMB/1396-candidates-can-delete-old-references-when-they-apply-again

## Things to check

### DIFF
--- a/spec/system/candidate_interface/candidate_replaces_reference_after_applying_again_spec.rb
+++ b/spec/system/candidate_interface/candidate_replaces_reference_after_applying_again_spec.rb
@@ -3,13 +3,6 @@ require 'rails_helper'
 RSpec.feature 'Candidate applying again' do
   include CandidateHelper
 
-  around do |example|
-    date_that_avoids_clocks_changing_by_ten_days = Time.zone.local(2020, 1, 13)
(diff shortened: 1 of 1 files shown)


spec/system/candidate_interface/candidate_replaces_reference_after_applying_again_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+
+RSpec.feature 'Candidate applying again' do
+  include CandidateHelper
+
+  around do |example|
+    date_that_avoids_clocks_changing_by_ten_days = Time.zone.local(2020, 1, 13)
+    Timecop.freeze(date_that_avoids_clocks_changing_by_ten_days) do
+      example.run
+    end
+  end
```
tijmenb, line 11: Do we need this bit?

stevehook: Good point, that's not needed in this spec because there are not date based assertions.

PR #2009, opened 2020-05-05
Add missing course information to application show view

willmcb:
## Context

Adds more course information (accredited body and funding type) to the course table in the applications show view.

**Before:**
Screenshot 2020-05-05 at 14 33 29

**After:**
Screenshot 2020-05-05 at 14 32 29

## Guidance to review

- If a course has no `accredited_provider` it is assumed that they are self accrediting and the provider name is rendered as the Accrediting bod (see spec); is this a correct assumption to make?
- Are the details in the right order?
- Is funding type the right thing to show to the user?
- are the names for the funding types correct? Underlying data for funding types is `apprenticeship`, `fee`, and `salary` which are rendered to the user as 'Apprenticeship', 'Fee paying' and 'Salaried' respectively.

## Link to Trello card

https://trello.com/c/Nj7rrX7e/2059-single-application-view-include-missing-information-in-course-section

## Things to check

### DIFF
--- a/spec/components/provider_interface/course_details_component_spec.rb
+++ b/spec/components/provider_interface/course_details_component_spec.rb
@@ -15,48 +15,77 @@ RSpec.describe ProviderInterface::CourseDetailsComponent do
 
   let(:render) { render_inline(described_class.new(application_choice: application_choice)) }
 
+  def row_text_selector(row_name, render)
+    rows = { provider: 0,
+             accredited_body: 1,
+             course: 2,
+             cycle: 3,
+             location: 4,
+             full_or_part_time: 5,
+             funding_type: 6 }
+
+    render.css('.govuk-summary-list__row')[rows[row_name]].text
+  end
+
   it 'renders the provider name and code' do
-    expect(render.css('.govuk-summary-list__row')[0].text).to include('Provider')
-    expect(render.css('.govuk-summary-list__row')[0].text).to include(application_choice.provider.name_and_code)
+    render_text = row_text_selector(:provider, render)
+
+    expect(render_text).to include('Provider')
+    expect(render_text).to include(application_choice.provider.name_and_code)
   end
 
   context 'when an accredited body is present' do
     it 'renders the accredited body name and code when it is present' do
       render = render_inline(described_class.new(application_choice: application_choice_with_accredited_body))
-      expect(render.css('.govuk-summary-list__row')[1].text).to include('Accredited body')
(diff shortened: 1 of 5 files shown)


spec/components/provider_interface/course_details_component_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  it 'renders the study mode' do
+    expect(render.css('.govuk-summary-list__row')[5].text).to include('Study mode')
+    expect(render.css('.govuk-summary-list__row')[5].text).to include(application_choice.course_option.study_mode.humanize)
+  end
+
+  it 'renders financing funding type of a course' do
+    expect(render.css('.govuk-summary-list__row')[6].text).to include('Funding type')
+    expect(render.css('.govuk-summary-list__row')[6].text).to include('Fee paying')
+  end
```
tijmenb, line 59: I appreciate the thoroughness of this testing!

Two questions:

- When we make changes in the ordering, we would need to change a lot of the assertions. Can we avoid that somehow?
- Some of the assertions duplicate the code that's being tested. For example, the study mode test in essence tests that `application_choice.course_option.study_mode.humanize` is the same as `application_choice.course_option.study_mode.humanize`. Perhaps you could make it specific, so we test that the resulting study mode string actually makes sense (and isn't a sneaky nil for example).

spec/factories.rb
```diff
@@ -275,6 +275,10 @@
     trait :part_time do
       study_mode { :part_time }
     end
+
+    trait :accredited_provider do
+      accredited_provider { provider }
+    end
```
tijmenb, line 281: Optional: this could go into the `let` where we're using it, to avoid the factories from becoming too large.

adamsilver:
Please change ‘Study mode’ to ‘Full or part time’

spec/factories.rb
willmcb: done 👍

willmcb:
> Please change ‘Study mode’ to ‘Full or part time’

spec/components/provider_interface/course_details_component_spec.rb
willmcb: have added a method that looks rows up by name and returns an index to make it easier to change row orders in future (they are now all in one place).

willmcb: I have added literals to the test now 👍

willmcb: Also moved to instance_doubles as there is no need for DB creation with this kind of component test in my mind.

PR #2011, opened 2020-05-05
Candidates can't edit old references when they apply again

stevehook:
## Context

In an Apply Again application we will normally have two completed references whereas in normal first-time applications the references are all unsubmitted. We need to stop candidates from editing their old references to prevent them from getting into weird states.

- [x] Drop the 'Change' links in the `ReviewReferencesComponent` unless the reference is unsubmitted.
- [x] Update system spec.
- [x] Return an invalid operation code if anybody hits the actions that update a references unless it's unsubmitted.
- [x] Look at what needs to be covered by the feature flag.

## Guidance to review

Does the logic around hiding the change links make sense?

## Link to Trello card

https://trello.com/c/pmArg5iR/1397-candidates-cant-edit-old-references-when-they-apply-again

## Things to check

### DIFF
--- a/spec/components/candidate_interface/referees_review_component_spec.rb
+++ b/spec/components/candidate_interface/referees_review_component_spec.rb
@@ -128,6 +128,15 @@ RSpec.describe CandidateInterface::RefereesReviewComponent do
       result = render_inline(described_class.new(application_form: application_form, editable: false))
 
       expect(result.css('.app-summary-list__actions').text).not_to include('Change')
+    end
+  end
+
+  context 'when referees are not deletable' do
+    let(:application_form) { create(:completed_application_form, references_count: 1, with_gces: true) }
+
+    it 'renders component without a delete link' do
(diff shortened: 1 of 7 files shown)


spec/components/candidate_interface/referees_review_component_spec.rb
```diff
@@ -50,8 +57,8 @@
       first_referee = application_form.application_references.first
       first_referee.update_columns(
         feedback_status: 'feedback_requested',
-        requested_at: 9.business_days.ago,
-        created_at: 9.business_days.ago,
+        requested_at: 5.days.ago,
+        created_at: 5.days.ago,
```
stevehook, line 61: This test started failing when I changed the `feedback_status`. I think it was wrong because the applicable time limit is 7 elapsed days so `9.business.days` is always going to be in the overdue state. With the Covid-19 break even `2.business.days` is more than 7 elapsed.

PR #2012, opened 2020-05-05
Providers with manage users permission can delete users associated with their org

steventux:
## Context

Provider users with specific permissions can manage other users, we should enable them to remove users from their Provider org(s).

Adds a way to remove a provider user from provider orgs common to the managing user (the current provider user) and the target user. This effectively deletes them from view. A confirmation step is presented in this journey.

Any associations to provider orgs not common to the managing and editable user are preserved as these are still valid associations.
 

image

and confirmation screen

image

## Link to Trello card

https://trello.com/c/pPKOrnfF/2019-build-providers-with-manageusers-permission-can-delete-users-associated-with-their-org

## Things to check

### DIFF
--- a/app/controllers/provider_interface/provider_users_controller.rb
+++ b/app/controllers/provider_interface/provider_users_controller.rb
@@ -74,6 +74,21 @@ module ProviderInterface
       redirect_to provider_interface_provider_user_path(provider_user)
     end
 
+    def confirm_remove
+      @provider_user = find_provider_user
+    end
+
+    def remove
+      @provider_user = find_provider_user
+      service = RemoveProviderUser.new(
+        current_provider_user: current_provider_user,
+        user_to_remove: @provider_user,
+      )
+
+      flash[:success] = 'User’s account successfully deleted' if service.call!
+      redirect_to provider_interface_provider_users_path
+    end
+
   private
 
     def provider_user_params
--- a/app/services/remove_provider_user.rb
+++ b/app/services/remove_provider_user.rb
@@ -1,0 +1,17 @@
+class RemoveProviderUser
+  attr_reader :current_provider_user, :user_to_remove
+
+  def initialize(current_provider_user:, user_to_remove:)
+    @current_provider_user = current_provider_user
+    @user_to_remove = user_to_remove
+  end
+
+  # Removes associations to providers common to the managing user
+  # and the user we're editing/removing.
+  def call!
+    managed_providers = current_provider_user.provider_permissions.manage_users.includes(:provider).map(&:provider)
+    shared_providers = managed_providers & user_to_remove.providers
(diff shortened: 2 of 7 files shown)


app/services/remove_provider_user.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  def initialize(current_provider_user:, user_to_remove:)
+    @current_provider_user = current_provider_user
+    @user_to_remove = user_to_remove
+  end
+
+  # Removes associations to providers common to the managing user
+  # and the user we're editing/removing.
+  def call!
+    shared_providers = current_provider_user.providers & user_to_remove.providers
```
tijmenb, line 12: Should this be `current_provider_user.providers.manage_users`?

steventux: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2012/commits/e659aabb6938ba7684f75fd56bd065e7ab80b07e#diff-1a42958a6311b2501e78d6b964484de7R12 scopes this to managed providers.
I wonder about the use of `#map` here though, I think we should be able to achieve this with a join. I know we saw some non-uniq results in the past using a join but I wonder if this was the result of a bug which caused duplicate `ProviderPermission` records which has been fixed by https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1931.

app/controllers/provider_interface/provider_users_controller.rb
```diff
@@ -82,6 +82,21 @@ def update_providers
       redirect_to provider_interface_provider_user_path(provider_user)
     end
 
+    def confirm_remove
+      @provider_user = ProviderUser.visible_to(current_provider_user).find(params[:provider_user_id])
```
tijmenb, line 86: This can use `find_provider_user` from https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2013 now.

app/controllers/provider_interface/provider_users_controller.rb
```diff
@@ -82,6 +82,21 @@ def update_providers
...
+    end
+
+    def remove
+      @provider_user = ProviderUser.visible_to(current_provider_user).find(params[:provider_user_id])
```
tijmenb, line 90: This can use `find_provider_user` from https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2013 now.

steventux: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2012/commits/481fdb517c6bb01a4f1c34e85eee421dc316760b#diff-6c36ce09043ee29972a80d5a8862669dR78-R82 :+1:

PR #2013, opened 2020-05-05
Show 404s in provider interface when user not found, feature flag is off

tijmenb:
## Context

Follow up to https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1931. 

Shows a 404 when:

- you don't have access to a user
- you don't have access to manage any users
- the feature flag is off 

## Link to Trello card

https://trello.com/c/BTWwpO0Y

## Things to check

### DIFF
--- a/app/controllers/provider_interface/sessions_controller.rb
+++ b/app/controllers/provider_interface/sessions_controller.rb
@@ -16,7 +16,7 @@ module ProviderInterface
     end
 
     def sign_in_by_email
-      raise unless FeatureFlag.active?('dfe_sign_in_fallback')
+      render_404 unless FeatureFlag.active?('dfe_sign_in_fallback')
 
       provider_user = ProviderUser.find_by(email_address: params.dig(:provider_user, :email_address).downcase.strip)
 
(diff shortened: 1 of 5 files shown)


app/controllers/provider_interface/sessions_controller.rb
```diff
@@ -16,7 +16,7 @@ def destroy
     end
 
     def sign_in_by_email
-      raise unless FeatureFlag.active?('dfe_sign_in_fallback')
+      render_404 unless FeatureFlag.active?('dfe_sign_in_fallback')
```
tvararu, line 19: I clicked on a magic link URL to QA, and it signed me in, even though the feature flag is off.

@tijmenb this renders a 404 and then proceeds to sign the user in anyway. This means that even if the feature flag is off, we accept the sign in attempt using the magic link. This seems incorrect; should it be:

```ruby
unless FeatureFlag.active?('dfe_sign_in_fallback')
  render_404
  return
end
```

tvararu: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2072

PR #2016, opened 2020-05-06
Candidate can see previous application when applying again

davidgisbey:
## Context

Candidates should be able to reference their previous application when applying again.

This PR adds a link on their application form page which allows them to review their previous application on the application_review_previous path.

- Adds a new page which takes an id and renders the review submitted page for that id.

## Guidance to review

I'll leave a comment on the bit of code I am not a fan of. I think there will be a far more elegant solution.

## Link to Trello card

https://trello.com/c/swT0QSj9/1394-candidates-can-view-their-previous-applications-when-they-apply-again

## Things to check


app/views/candidate_interface/application_form/review_submitted.html.erb
```diff
@@ -1,5 +1,10 @@
 <% content_for :title, t('page_titles.submitted_application') %>
-<% content_for :before_content, govuk_back_link_to(candidate_interface_application_form_path, 'Back to application dashboard') %>
+
+<% if @review_previous_application %>
```
davidgisbey, line 3: The pages are identical with the exception of the content of the back link. Perhaps the rest of the page could go into a partial to avoid this?

app/views/candidate_interface/application_form/show.html.erb
```diff
@@ -110,5 +110,13 @@
 
   <div class="govuk-grid-column-one-third">
     <%= render partial: '/candidate_interface/shared/need_help', locals: { support_reference: @application_form.support_reference } %>
+
+    <% if FeatureFlag.active?('apply_again') %>
```
stevehook, line 114: Should there be a check to see if `@previous_application.present?` here?

It crashes when I run it on a non-Apply Again application and in any case the heading is redundant if we don't have anything to link to.

app/views/candidate_interface/application_form/review_submitted.html.erb
stevehook: Not sure I understand what you mean by pages being identical. I can't see anything too bad about this apart from the `content_for` call being (almost) duplicated. I can't think of a neat way to avoid that. You could specify the default parameter value as a constant in the `ViewHelper` so that you could do:

```
<% content_for :before_content, govuk_back_link_to(candidate_interface_application_form_path, @review_previous_application ? 'Back to application dashboard') : ViewHelper::DEFAULT_BACK_LABEL %>
```

 But I think that is less clear.

app/views/candidate_interface/application_form/show.html.erb
davidgisbey: good shout!

app/views/candidate_interface/application_form/review_submitted.html.erb
davidgisbey: Sorry, that is misleading. Two actions render the same view but need different text in the back link. I think the current solution is probably fine then 👍

app/views/candidate_interface/application_form/show.html.erb
davidgisbey: Good catch. Thanks!

app/views/candidate_interface/application_form/show.html.erb
```diff
@@ -110,5 +110,13 @@
 
   <div class="govuk-grid-column-one-third">
     <%= render partial: '/candidate_interface/shared/need_help', locals: { support_reference: @application_form.support_reference } %>
+
+    <% if FeatureFlag.active?('apply_again') && @previous_application_form.present? %>
+      <h2 class="govuk-heading-s">Previous applications</h2>
+      <%= govuk_link_to 'First application', candidate_interface_application_review_previous_path(@previous_application_form.id) %>
+      <p class="govuk-body-s">
+        <%= "Submitted #{@previous_application_form.submitted_at.to_s(:govuk_date).strip}" %>
+      </p>
```
tvararu, line 119: Can we do the work now to make this work for N applications, or is it tricky currently? I think we could enumerate `current_candidate.application_forms.sort_by(&:submitted_at)` to come up with a list, `each` them, and then use the inflector to numeralize the link?

tvararu: Plus to ease testing burden, it could be extracted as a component and unit tested

app/controllers/candidate_interface/application_form_controller.rb
```diff
@@ -73,6 +73,13 @@ def review_submitted
       @application_form = current_application
     end
 
+    def review_previous_application
+      @application_form = ApplicationForm.find(current_application.previous_application_form_id)
```
tijmenb, line 77: This needs to be scoped to the current user.

davidgisbey: isn't current application scoped to the user?

config/routes.rb
```diff
@@ -56,6 +56,7 @@
       get '/edit' => 'application_form#edit', as: :application_edit
       get '/review' => 'application_form#review', as: :application_review
       get '/review/submitted' => 'application_form#review_submitted', as: :application_review_submitted
+      get '/review/submitted/:id' => 'application_form#review_previous_application', as: :application_review_previous
```
tijmenb, line 59: What does the `:id` do here?

app/views/candidate_interface/application_form/show.html.erb
davidgisbey: I don't see any reason this can't be done now

config/routes.rb
davidgisbey: I think we will need this for showing previous applications when there are more than 1. My current implementation doesn't reflect this. So if they have 3 previous applications it would pass the id specific to the first/second/third application.

app/controllers/candidate_interface/application_form_controller.rb
```diff
@@ -73,6 +73,13 @@ def review_submitted
       @application_form = current_application
     end
 
+    def review_previous_application
+      @application_form = ApplicationForm.find(params['id']) if params['id'].present?
```
tijmenb, line 77: This needs to be scoped to the current user, otherwise everyone can see everyone elses application.

Why does it check for the presence of the ID? It will always be here, otherwise the you can't visit the route.

app/controllers/candidate_interface/application_form_controller.rb
```diff
@@ -82,5 +93,11 @@ def further_information_params
       )
         .transform_values(&:strip)
     end
+
+    def build_previous_application_form
+      return if params['candidate_id']&.to_i != current_candidate.id
```
tijmenb, line 98: If I understand it correctly, is this meant to avoid people seeing other people's applications? It's doesn't work for that, since they can change the ID in the URL to their own, and this will not `return`.

Luckily, the `current_candidate.application_forms.find()` is the correct approach so we don't need this extra check here.

app/controllers/candidate_interface/application_form_controller.rb
```diff
@@ -82,5 +93,11 @@ def further_information_params
...
+
+      @application_form = current_candidate.application_forms.find(params['id']) if params['id'].present?
```
tijmenb, line 100: I tried to say before in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2016#discussion_r420847452 that it's impossible for `params['id']` to be nil. That's because the user can only reach this code by visiting the `/application/:id/submitted` route, so the `params[:id]` must always exist.

app/controllers/candidate_interface/application_form_controller.rb
```diff
@@ -73,6 +74,16 @@ def review_submitted
       @application_form = current_application
     end
 
+    def review_previous_application
+      if @application_form.nil?
```
tijmenb, line 78: This can never be true, because if the application form doesn't exist it'll raise a `ActiveRecord::RecordNotFound` exception.

spec/system/candidate_interface/candidate_can_see_their_previous_application_on_apply_again_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+    and_i_click_on_apply_again
+    and_i_click_on_start_now
+    and_i_click_on_the_link_to_my_previous_application
+    then_i_see_the_review_previous_application_page
+
+    when_i_click_back
+    then_i_see_my_current_application_page
+
+    when_i_try_and_visit_the_page_with_another_candidate_id
+    then_i_am_redirected_to_my_current_application_page
```
tijmenb, line 19: Can we add these tests to the main Apply 2 feature test, to avoid having to do the setup twice?

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/spec/system/candidate_interface/candidate_with_unsuccessful_application_applies_again_spec.rb

app/controllers/candidate_interface/application_form_controller.rb
```diff
@@ -73,6 +74,16 @@ def review_submitted
...
+        redirect_to candidate_interface_application_form_path
+      else
+        @review_previous_application = true
+
+        render :review_submitted
+      end
+    end
```
tijmenb, line 85: As a fix for my other comments, what would you make of something like this?

```suggestion
    def review_previous_application
      @application_form = current_candidate.application_forms.find(params[:id])
      @review_previous_application = true
      render :review_submitted
    rescue ActiveRecord::RecordNotFound
      render_404
    end
```

app/components/candidate_interface/links_to_previous_applications_component.html.erb
```diff
@@ -0,0 +1,6 @@
+<% submitted_applications.each do |application_form| %>
+  <%= govuk_link_to ordinalize(application_form), candidate_interface_application_review_previous_path(application_form.candidate.id, application_form.id) %>
```
tijmenb, line 2: Optional: this reads a bit awkwardly, perhaps:

```suggestion
  <%= govuk_link_to ordinalize(application_form), candidate_interface_review_previous_application_path(application_form.candidate.id, application_form.id) %>
```

app/views/candidate_interface/application_form/show.html.erb
```diff
@@ -16,8 +16,8 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <p class="govuk-hint govuk-!-margin-bottom-8 govuk-!-margin-top-2"><%= @application_form_presenter.updated_at %></p>
-    
-    <% if FeatureFlag.active?('apply_again') && @application_form.apply_2? && @previous_application_form.application_choices.rejected.present?  %>
+
+    <% if FeatureFlag.active?('apply_again') && @application_form.apply_2? && @previous_application_form&.application_choices&.rejected.present?  %>
```
tijmenb, line 20: Using `&` is a bit of a smell, this might be more robust:

```suggestion
    <% if FeatureFlag.active?('apply_again') && @previous_application_form && @previous_application_form.application_choices.rejected.any?  %>
```

spec/components/candidate_interface/links_to_previous_applications_component_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  let(:candidate) { build(:candidate) }
+  let(:application_form) { create(:application_form, candidate: candidate, phase: :apply_2) }
+  let(:previous_application_form) { create(:completed_application_form, candidate: candidate, phase: :apply_2, submitted_at: 3.days.ago) }
+  let(:first_application_form) { create(:completed_application_form, candidate: candidate, phase: :apply_1, submitted_at: 5.days.ago) }
+
+  before do
+    application_form
+    previous_application_form
+    first_application_form
+  end
```
tijmenb, line 13: FYI `let!` does what you want here:

```suggestion
  let!(:application_form) { create(:application_form, candidate: candidate, phase: :apply_2) }
  let!(:previous_application_form) { create(:completed_application_form, candidate: candidate, phase: :apply_2, submitted_at: 3.days.ago) }
  let!(:first_application_form) { create(:completed_application_form, candidate: candidate, phase: :apply_1, submitted_at: 5.days.ago) }
```

app/controllers/candidate_interface/application_form_controller.rb
davidgisbey: Yea i was having a struggle day yesterday haha. Thanks, i get it now.

app/controllers/candidate_interface/application_form_controller.rb
davidgisbey: Looks good to me 👍 The rescue should never be hit right other than manually messing with the URL, but it's still good to have in place!

spec/components/candidate_interface/links_to_previous_applications_component_spec.rb
davidgisbey: Okay cool. On Find banging lets was a capital crime haha. Good to know i can use them here.

app/components/candidate_interface/links_to_previous_applications_component.html.erb
davidgisbey: That makes sense. I was trying to stick with the convention, but this reads so much more clearly 👌

app/views/candidate_interface/application_form/show.html.erb
davidgisbey: Good suggestion. Thanks!

spec/system/candidate_interface/candidate_can_see_their_previous_application_on_apply_again_spec.rb
davidgisbey: Have done 👍

davidgisbey:
Sorry, Tijmen I feel like I'm really making you earn your money this week haha. Thanks for all the feedback!

config/routes.rb
```diff
@@ -50,6 +50,8 @@
 
     get '/interstitial', to: 'after_sign_in#interstitial', as: :interstitial
 
+    get '/:candidate_id/application/:id/submitted' => 'application_form#review_previous_application', as: :review_previous_application
```
tijmenb, line 53: Why do we need this?

davidgisbey: Theo asked for the work to be completed that would allow a candidate to see all of their previous applications. So moving it over from using the previous_application to taking an application_form id seemed like the most reasonable thing to do

tijmenb: Sorry, I meant `:candidate_id`. It's not used?

davidgisbey: Oh right. Yea good point 👍

PR #2019, opened 2020-05-06
Cast check icon values to booleans before rendering

willmcb:
We were getting false positives when rendering check icons for whether someone had experience of working with children.

PR adds casting to header component so that `check_icon` is rendered when needed.

## Context

Bug identified by user

N/A in terms of UI design.

## Guidance to review

- is the component level approach correct?

## Link to Trello card

https://trello.com/c/lXwtGC10/2114-bug-this-role-involved-working-with-children-incorrectly-rendering-on-heading-components

## Things to check


tvararu:
Discussed offline: I feel like adding casting to the component argument is a smell. It feels like a band-aid. We should have a way of preventing the data from reaching the component in a malformed way. 🤔

PR #2020, opened 2020-05-06
Restrict course selection when candidates Apply Again

malcolmbaig:
## Context

When a candidate's application is unsuccessful, they can opt to Apply
Again. One of the business rules for the Apply Again process is that
they can only apply to one course at a time.

- Add UI changes to prevent the adding of additional courses under Apply Again.
- Add a form validation to prevent the same when navigating to the course selection screens by URL.

## Application Form

Screenshot 2020-05-06 at 13 28 43

## Course Choice Review

Screenshot 2020-05-06 at 13 28 51

## Validation error

Screenshot 2020-05-06 at 13 30 38

## Guidance to review

- Can be tested manually with an application in the Apply Again phase. Probably easiest to use the DuplicateApplicationService in the console.
- Check the component refactor makes sense. 

## Link to Trello card

https://trello.com/c/0Tj02wBa
## Things to check


tvararu:
Screenshot 2020-05-06 at 17 13 37

Found this bug, steps to reproduce:

- Log in as `burt.beahan@example.com` on this Review app
- Go to https://apply-for-te-1391-apply-mzq34j.herokuapp.com/candidate/application/courses/provider
- Gorse SCITT
- Biology 7 to 14
- Boston Spa Academy
- 💥

stevehook:
Have done a bit of manual testing and looks like it's watertight (I wasn't able to work around the validation). Using the back button can be confusing but I think this is a longstanding issue. e.g. if I add my one permitted course and then hit _Back_ a couple of times I end up with a validation error:
image

malcolmbaig:
@tvararu That's the existing experience when you get a [validation error](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/app/controllers/candidate_interface/course_choices_controller.rb#L290) on [PickSiteForm](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/app/models/candidate_interface/pick_site_form.rb#L43). Agreed this could be better, though. My latest commit changes that render call, and instead redirects to the application form with a flash warning set.

Screenshot 2020-05-07 at 09 50 54

WDYT?

app/views/candidate_interface/application_form/show.html.erb
```diff
@@ -8,30 +8,32 @@
...
+    <% if FeatureFlag.active?('apply_again') && @application_form.apply_again? && @previous_application_form.application_choices.rejected.present?  %>
       <%= render(CandidateInterface::RejectionReasonsComponent.new(application_form: @previous_application_form)) %>
     <% end %>
 
-    <h2 class="govuk-heading-m govuk-!-font-size-27 govuk-!-margin-bottom-2">
-      <%= t('page_titles.course_choices') %>
-    </h2>
-    <% unless @application_form_presenter.application_choices_added? %>
-      <p class="govuk-body"><%= t('application_form.courses.intro') %></p>
+    <% if FeatureFlag.active?('apply_again') && @application_form_presenter.apply_again? %>
```
malcolmbaig, line 23: Thinking about this feature flag. Is there actually any point to checking it's active outside of the bit where a candidate can hit the button to start the Apply Again process? 

Once they've duped their application and entered the apply_again phase, if we were to disable the feature flag for whatever reason in the future, then I think it would be preferable for any existing applications in that state to continue seeing the apply_again UI. Not doing so leaves the system in an incorrect state where they can, eg - add three courses to an apply_again application.

tvararu: @malcolmbaig 👍

app/views/candidate_interface/course_choices/index.html.erb
```diff
@@ -6,7 +6,11 @@
     <h1 class="govuk-heading-xl">
       <%= t('page_titles.choosing_courses') %>
     </h1>
-    <%= render 'guidance' %>
+    <% if FeatureFlag.active?('apply_again') && apply_again %>
```
tvararu, line 9:
```suggestion
    <% if apply_again %>
```

Similar logic to above. 👍

app/views/candidate_interface/application_form/show.html.erb
tvararu, line 23:
```suggestion
    <% if @application_form_presenter.apply_again? %>
```

app/views/candidate_interface/course_choices/review.html.erb
```diff
@@ -2,7 +2,11 @@
 <% content_for :before_content, govuk_back_link_to(candidate_interface_application_form_path, 'Back to application') %>
 
 <h1 class="govuk-heading-xl">
+  <% if FeatureFlag.active?('apply_again') && @application_form.apply_again? %>
```
tvararu, line 5:
```suggestion
  <% if @application_form.apply_again? %>
```

app/views/candidate_interface/course_choices/review.html.erb
```diff
@@ -12,21 +16,23 @@
 
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-      <% if @course_choices.count.present? %>
-        <% if @course_choices.count.between?(1, 2) %>
+      <% if @course_choices.present? %>
+        <% if FeatureFlag.active?('apply_again') && @application_form.apply_again? %>
```
tvararu, line 20:
```suggestion
        <% if @application_form.apply_again? %>
```

spec/components/candidate_interface/application_form_course_choices_component_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+    result.css('a').first.attributes['href'].value
+  end
+
+  def status_text(result)
+    result.css('.govuk-tag').first.text.strip
+  end
+
+  def first_paragraph(result)
+    result.css('p').first
+  end
```
tvararu, line 93: Nice

app/views/candidate_interface/application_form/show.html.erb
malcolmbaig: cc: @stevehook @davidgisbey for visibility of the above

PR #2022, opened 2020-05-06
Show the correct number of providers next to each user when a provider is managing users

steventux:
## Context

We restrict the operations the current provider user can perform based on their provider permissions.
When they view the users index, the provider information for each user should reflect the same constraints on what provider information is visible.

Scope providers to those manageable by current provider user when displaying provider information on the provider users index.

## Link to Trello card

https://trello.com/c/dQnskexv/2069-show-the-correct-number-of-providers-next-to-each-user-when-a-provider-is-managing-users

## Things to check

### DIFF
--- a/app/components/provider_interface/user_list_card_component.rb
+++ b/app/components/provider_interface/user_list_card_component.rb
@@ -12,7 +12,7 @@ module ProviderInterface
     end
 
     def providers_text
-      providers = provider_user.providers & manageable_providers
+      providers = provider_user.providers.where(id: manageable_providers.pluck(:id)).order(:name)
 
       return providers.first.name if providers.size == 1
 
--- a/app/components/provider_interface/user_list_card_component.rb
+++ b/app/components/provider_interface/user_list_card_component.rb
@@ -2,16 +2,18 @@ module ProviderInterface
   class UserListCardComponent < ViewComponent::Base
     include ViewHelper
 
-    attr_reader :provider_user, :full_name, :email_address
+    attr_reader :provider_user, :manageable_providers, :full_name, :email_address
 
-    def initialize(provider_user:)
+    def initialize(provider_user:, manageable_providers:)
       @provider_user = provider_user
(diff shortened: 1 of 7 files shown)


app/models/provider.rb
```diff
@@ -26,6 +26,11 @@ class Provider < ApplicationRecord
 
   audited
 
+  scope :manageable_by, ->(provider_user) do
```
duncanjbrown, line 29: Is this technically `with_users_manageable_by`?

app/components/provider_interface/user_list_card_component.rb
```diff
@@ -2,16 +2,18 @@ module ProviderInterface
...
+    def initialize(provider_user:, manageable_providers:)
       @provider_user = provider_user
+      @manageable_providers = manageable_providers
       @full_name = provider_user.full_name
       @email_address = provider_user.email_address
     end
 
     def providers_text
-      providers = provider_user.providers
+      providers = provider_user.providers.where(id: manageable_providers.pluck(:id)).order(:name)
```
duncanjbrown, line 15: I think the `pluck` will run on each row render — could we push it up into the controller so it only runs once? Maybe the controller could generate structs or hashes like `{ user: user, providers: user.providers.where.etc }` and the component could be dumber

app/models/provider.rb
steventux: Amended here: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2022/commits/c76fe3e881ac35c3538a122dfc753425f8ae8eb3

app/components/provider_interface/user_list_card_component.rb
steventux: I've done something like that here https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2022/commits/d2e2a0ffaedccab44e33b08b12de803d3b8bd09f

PR #2023, opened 2020-05-06
Fix issue with booleans in volunteering_role_form

willmcb:
We were getting false positives when rendering check icons fo volunteering for whether someone had experience of working with children for a given volunteering experience.

## Context

Bug identified by user

N/A in terms of UI design.

## Guidance to review

- do these changes make sense to you?
- may be worth creating a new application - adding some volunteering experience at with with `working_with_children` set to 'Yes', and one set to 'No'. To test this is working correctly. It does for me but I may have done something silly, so a second pair of eyes would be good.

## Link to Trello card

https://trello.com/c/lXwtGC10/2114-bug-this-role-involved-working-with-children-incorrectly-rendering-on-heading-components

## Things to check


spec/models/candidate_interface/volunteering_role_form_spec.rb
```diff
@@ -6,7 +6,7 @@
       role: 'School Experience Intern',
       organisation: Faker::Educator.secondary_school,
       details: Faker::Lorem.paragraph_by_chars(number: 300),
-      working_with_children: [true, true, true, false].sample,
+      working_with_children: true,
```
tvararu, line 9: The intention of this is to test that it works with both false and true, but predominantly true; do we need to remove it?

willmcb: Good point will have a think.

willmcb:
@tvararu happy to apply it to the work experience as well - interestingly though when I tested the work experience it appears to work at the moment though 🤔

PR #2027, opened 2020-05-07
Show pre submission reference guidance under not_requsted_yet status

davidgisbey:
## Context

At the moment, we have static guidance about contacting referees on this page: 

https://bat-design-history.netlify.app/images/apply-for-teacher-training/improving-the-references-user-journey/static-guidance-about-contacting-your-referees-when-you-submit.png

We've changed the design so that guidance appears under the referee status tag instead, appearing only in the 'not requested' state - i.e., when someone has added or replaced a reference:

https://bat-design-history.netlify.app/images/apply-for-teacher-training/improving-the-references-user-journey/showing-guidance-at-the-relevant-point-when-replacing-referees.png

- Only show the pre-submission guidance when a reference is a  not_requsted_yet state

Before

image

After

image

Doesn't render post submission

image

## Guidance to review

I'll leave a comment about one thing 👍 

## Link to Trello card

https://trello.com/c/el
2X7sKpK/1442-dev-changes-to-references-section-for-apply-1-and-apply-2-turning-static-guidance-into-conditional-guidance-with-slight-change-t

## Things to check

### DIFF
--- a/app/components/candidate_interface/referees_review_component.rb
+++ b/app/components/candidate_interface/referees_review_component.rb
@@ -69,16 +69,18 @@ module CandidateInterface
     end
 
     def feedback_status_row(referee)
-      return nil unless show_status?(referee)
-
-      {
-        key: 'Status',
-        value: feedback_status_label(referee),
-      }
-    end
-
-    def show_status?(referee)
-      referee.application_form.submitted?
+      if referee.not_requested_yet? && !referee.application_form.submitted?
+        {
+          key: 'Status',
+          value: feedback_status_label(referee) +
+            content_tag(:p, t('application_form.referees.info.not_requested_yet'), class: 'govuk-body govuk-!-margin-top-2'),
+        }
+      else
+        {
+          key: 'Status',
+          value: feedback_status_label(referee),
+        }
+      end
     end
 
     def feedback_status_label(reference)
--- a/spec/components/candidate_interface/referees_review_component_spec.rb
+++ b/spec/components/candidate_interface/referees_review_component_spec.rb
@@ -30,9 +30,13 @@ RSpec.describe CandidateInterface::RefereesReviewComponent do
 
     it 'renders component with correct value for status for unrequested reference' do
       application_form.update_column(:submitted_at, nil)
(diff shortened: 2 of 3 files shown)


app/components/candidate_interface/referees_review_component.rb
```diff
@@ -69,16 +69,18 @@ def reference_type_row(referee)
...
-    def show_status?(referee)
-      referee.application_form.submitted?
+      if referee.not_requested_yet? && !referee.application_form.submitted?
+        {
+          key: 'Status',
+          value: feedback_status_label(referee) +
+            content_tag(:p, t('application_form.referees.info.not_requested_yet'), class: 'govuk-body govuk-!-margin-top-2'),
+        }
+      else
+        {
```
davidgisbey, line 79: Is there any other situation where we wouldn't want to render the status? I can't think of one.

stevehook: Can't think of one

spec/components/candidate_interface/referees_review_component_spec.rb
```diff
@@ -30,9 +30,13 @@
...
     it 'renders component with correct value for status for unrequested reference' do
       application_form.update_column(:submitted_at, nil)
+      first_referee = application_form.application_references.first
+      first_referee.update(feedback_status: 'not_requested_yet')
       result = render_inline(described_class.new(application_form: application_form))
 
-      expect(result.css('.govuk-summary-list__key').text).not_to include('Status')
+      expect(result.css('.govuk-summary-list__key').text).to include('Status')
+      expect(result.css('.govuk-summary-list__value').to_html).to include('Not requested')
+      expect(result.css('.govuk-summary-list__value').to_html).to include(t('application_form.referees.info.not_requested_yet'))
```
stevehook, line 39: Would be good to also see a couple of specs asserting this copy is not present when the status is different or the application is submitted.

davidgisbey: Good shout. I've added a couple in.

PR #2029, opened 2020-05-07
Fix access to mailer and component previews

mnacos:
## Context

Recent library upgrades have made mailer and view component previews unavailable.

For component previews, the mounted path changed from `/rails/components` to `/rails/view_components`, so it's just a case of updating our support footer link.

For mailer previews, it's the generation of the preview index that's problematic, as it relies on a `url_for` which errors. The actual email previews seem to work fine, if you know the path.

This is the commit responsible for the use of `url_for`, for (the slightly random reason of) supporting previews in a subdirectory: https://github.com/rails/rails/commit/f5132c19d52c4457aa2e09c647d2f52a73966c68

This PR updates the support footer links to point to the right paths for mailer and view component previews.

It also brings responsibility for generating the mailer previews index within the app. The code is pretty minimal, should be easy to maintain.

## Guidance to review

Inspect the code changes. Try to access previews through the support interface on various environments and locally.

## Link to Trello card

[Fix mailer previews on QA](https://trello.com/c/vkdrrBy7)

## Things to check

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -44,6 +44,8 @@ Rails.application.configure do
     end
   })
 
+  # for default_url_options, see config/initializers/default_url_options.rb
+
   # Do not buffer STDOUT in Ruby. This behaviour interacts weirdly with the docker-compose
   # log output and causes logs only to be printed when an exception occurs or the process
   # exits.
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -44,8 +44,6 @@ Rails.application.configure do
     end
   })
 
-  config.action_mailer.default_url_options = { host: 'localhost', port: 3000 }
-
   # Do not buffer STDOUT in Ruby. This behaviour interacts weirdly with the docker-compose
   # log output and causes logs only to be printed when an exception occurs or the process
   # exits.
--- a/config/initializers/default_url_options.rb
+++ b/config/initializers/default_url_options.rb
@@ -1,4 +1,9 @@
 uri = URI(HostingEnvironment.application_url)
-Rails.application.routes.default_url_options[:host] = uri.host
-Rails.application.routes.default_url_options[:port] = uri.port
-Rails.application.routes.default_url_options[:protocol] = uri.scheme
+application_config = Rails.application.routes
+mailer_config = ActionMailer::Base
+
+[application_config, mailer_config].each do |config|
+  config.default_url_options[:host] = uri.host
+  config.default_url_options[:port] = uri.port
+  config.default_url_options[:protocol] = uri.scheme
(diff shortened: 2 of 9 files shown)


spec/rails_helper.rb
```diff
@@ -5,7 +5,10 @@
 # Prevent database truncation if the environment is production
 abort('The Rails environment is running in production mode!') if Rails.env.production?
 require 'rspec/rails'
+
 # Add additional requires below this line. Rails is not loaded until this point!
+require 'view_component/test_helpers'
```
duncanjbrown, line 10: The changes in this file are already present in `spec/support/components.rb`

mnacos: Indeed :) missed that. Updated

config/initializers/default_url_options.rb
```diff
@@ -1,4 +1,9 @@
 uri = URI(HostingEnvironment.application_url)
-Rails.application.routes.default_url_options[:host] = uri.host
-Rails.application.routes.default_url_options[:port] = uri.port
-Rails.application.routes.default_url_options[:protocol] = uri.scheme
+application_config = Rails.application.routes
```
duncanjbrown, line 2: I think this file's duties could be inlined in `config.application.rb`, which is a less surprising place to find this config. (though exactly how, via `config.*` or direct update, isn't clear — see https://github.com/rails/rails/issues/29992)

config/environments/development.rb
```diff
@@ -44,8 +44,6 @@
     end
   })
 
-  config.action_mailer.default_url_options = { host: 'localhost', port: 3000 }
```
duncanjbrown, line 47: Replacing this with a parsed `HostingEnvironment.application_url` is good. We may as well get Foreman to send emails with the right port in links while we're at it, by changing `hosting_environment.rb` line 6:

```ruby
"http://localhost:3000"
```

to 

```ruby
"http://localhost:#{ENV.fetch('PORT', 3000)}"
```

tijmenb:
Do we have any idea what broke the previews? It'd be good to fix upstream so we can eventually remove this code.

duncanjbrown:
@tijmenb https://discuss.rubyonrails.org/t/understanding-a-conflict-between-two-activesupport-on-load-calls/74836

config/initializers/default_url_options.rb
mnacos: I agree but setting these values in `Rails.application.routes` has so far been the most reliable way to configure url options in all contexts outside mailers and somehow `routes` does not seem to be available at the time `config/application.rb` is loaded. I'd rather keep all config in one place.

PR #2035, opened 2020-05-11
Fix title notification for notes dot on application list in wrong place sometimes

willmcb:
## Context

Dot was in wrong place when note title was approx under 20 chars in length.

**Before:**
Screenshot 2020-05-11 at 16 27 25

**After:**
Screenshot 2020-05-11 at 16 28 01

## Guidance to review

- load provider applications index - make sure it looks right for you.

## Link to Trello card

https://trello.com/c/2gdRmYDC/2115-title-notification-dot-on-application-list-in-wrong-place-sometimes

## Things to check

### DIFF
--- a/app/components/provider_interface/application_card_component.html.erb
+++ b/app/components/provider_interface/application_card_component.html.erb
@@ -5,7 +5,7 @@
     <div class="app-application-card__status">
       <%= render ProviderInterface::ApplicationStatusTagComponent.new(application_choice: application_choice) %>
       <% if most_recent_note %>
-        <p class="app-application-card__note"><%= most_recent_note.subject %></p>
+        <span class="app-application-card__note"><%= most_recent_note.subject %></span>
       <% end %>
     <div>
   </h3>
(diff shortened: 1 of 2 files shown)


app/frontend/styles/_notes.scss
```diff
@@ -9,6 +9,8 @@
   padding: 1px 0px 1px 15px;
   margin-top: 10px;
   margin-bottom: 5px;
+  width: fit-content;
```
tvararu, line 12: Seems incompatible with IE11: https://caniuse.com/#feat=mdn-css_properties_width_fit-content

tvararu: Service manual still recommends "compliant" in IE11: https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices

tvararu: I think this is quite a small thing, but I'm going to take a look locally to see if there's an easy fix which doesn't use this property

willmcb:
@tvararu have changed to your suggestion and re-pushed.

PR #2036, opened 2020-05-11
Fix bug on the support interfaces candidate show page

davidgisbey:
## Context

image

We're getting the above error. The current way the component is being instantiated is no longer supported.

Update to use

` SupportInterface::ApplicationsTableComponent.new(application_forms: @application_forms)  `

instead of 

`SupportInterface::ApplicationsTableComponent, application_forms: @application_forms`

## Guidance to review

None

## Link to Trello card

I'll make one quickly

## Things to check


davidgisbey:
It is indeed. Everything was changed over a while ago, but it looks like this just slipped through.

PR #2054, opened 2020-05-13
Increase postgres connection and statement timeout

tvararu:
## Context

Azure seems to be sporadically timing out our postgres queries. The default is 5 seconds. Let's try bumping it to 10 to see if the errors subside.

Bump `timeout` and `statement_timeout` in database.yml.

## Guidance to review

Does this make sense?

## Link to Trello card

None.

## Things to check

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -1,7 +1,9 @@
 default: &default
   adapter: postgresql
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
-  timeout: 5000
+  variables:
+    statement_timeout: 10000
+    idle_in_transaction_session_timeout: 10000
   username: <%= ENV['DB_USERNAME'] %>
   password: <%= ENV['DB_PASSWORD'] %>
   host: <%= ENV['DB_HOSTNAME'] %>


duncanjbrown:
Apparently there is no `timeout:` key for the PG adapter: https://github.com/rails/rails/blob/ed29afcc03cd4c192944758cc97b91e189cb6c66/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L52-L69

Keys not appearing in that list are proxied to `libpq-connect` according to the above comment https://www.postgresql.org/docs/current/libpq-connect.html, and `timeout` isn't there either. We can pass `statement_timeout` as we do. If we want to have connection timeouts too I think we can add `connect_timeout` as a sibling of `adapter`. Otherwise I think we should remove the `timeout:` key

vigneshmsft:
yup doesn't look like `timeout` is supported https://guides.rubyonrails.org/configuring.html#configuring-a-postgresql-database
I think setting `idle_in_transaction_session_timeout` to a desired value is good housekeeping.

config/database.yml
```diff
@@ -1,7 +1,8 @@
 default: &default
   adapter: postgresql
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
-  timeout: 5000
+  variables:
+    statement_timeout: 10000
```
tvararu, line 5:
```suggestion
    statement_timeout: 10000
    idle_in_transaction_session_timeout: 10000
```

@vigneshmsft like this?

vigneshmsft: yup, looks good to me

PR #2057, opened 2020-05-13
Add view safeguarding information permission for provider users

steventux:
## Context

As a precursor to delivering permissions at a provider organisation level we need a user level permission which can be used in conjunction with the org level permission.

- Adds `view_safeguarding_information` permission to `ProviderPermissions`
- Adds permission checkbox on existing provider user provider permissions form.
- Adds permission checkbox on existing support interface form for managing provider users.
- Restricts visibility of safeguarding information based on the user-level permission for the relevant provider.

image

## Link to Trello card

https://trello.com/c/bwEGeJa3/2093-build-access-mgmt-safeguarding

## Things to check

### DIFF
--- a/app/components/provider_interface/provider_permissions_list_component.html.erb
+++ b/app/components/provider_interface/provider_permissions_list_component.html.erb
@@ -1,17 +1,11 @@
 <% visible_provider_permissions.each do |provider_permission| %>
   <h4><%= provider_permission.provider.name %></h4>
   <ul id="provider-<%= provider_permission.provider.id %>-enabled-permissions">
-    <%
-      humanized_permissions = [].tap do |ary|
-        ProviderPermissions::VALID_PERMISSIONS.each do |permission_name|
-          ary << permission_name.to_s.humanize if provider_permission.send(permission_name)
-        end
-      end
-    -%>
-    <% humanized_permissions.each do |humanized_permission| %>
-      <li><%= humanized_permission %></li>
+    <% formatted_permissions = humanized_permissions(provider_permission) -%>
+    <% formatted_permissions.each do |formatted_permission| %>
+      <li><%= formatted_permission %></li>
     <% end %>
-    <% if humanized_permissions.empty? %>
+    <% if formatted_permissions.empty? %>
       <li>No permissions</li>
     <% end %>
   </ul>
--- a/app/components/provider_interface/provider_permissions_list_component.html.erb
+++ b/app/components/provider_interface/provider_permissions_list_component.html.erb
@@ -1,14 +1,18 @@
 <% visible_provider_permissions.each do |provider_permission| %>
   <h4><%= provider_permission.provider.name %></h4>
   <ul id="provider-<%= provider_permission.provider.id %>-enabled-permissions">
-    <% ProviderPermissions::VALID_PERMISSIONS.each do |permission_name| %>
-      <li>
-        <% if provider_permission.send(permission_name) %>
-          <%= permission_name.to_s.humanize %>
-        <% else %>
-            No permissions
-        <% end %>
-      </li>
+    <%
+      humanized_permissions = [].tap do |ary|
+        ProviderPermissions::VALID_PERMISSIONS.each do |permission_name|
+          ary << permission_name.to_s.humanize if provider_permission.send(permission_name)
+        end
+      end
+    -%>
+    <% humanized_permissions.each do |humanized_permission| %>
+      <li><%= humanized_permission %></li>
+    <% end %>
+    <% if humanized_permissions.empty? %>
+      <li>No permissions</li>
     <% end %>
   </ul>
 <% end %>
--- a/app/models/support_interface/provider_user_form.rb
+++ b/app/models/support_interface/provider_user_form.rb
@@ -66,7 +66,11 @@ module SupportInterface
           provider_id: form.provider_permission[:provider_id],
           provider_user_id: provider_user.try(:id),
         )
-        permission.manage_users = form.provider_permission.fetch(:manage_users, false)
+
+        ProviderPermissions::VALID_PERMISSIONS.each do |permission_name|
+          permission.send("#{permission_name}=", form.provider_permission.fetch(permission_name, false))
+        end
+
         permission
       end
     end
--- a/app/views/provider_interface/application_choices/show.html.erb
+++ b/app/views/provider_interface/application_choices/show.html.erb
@@ -52,7 +52,9 @@
     <%= render ProviderInterface::TrainingWithDisabilityComponent.new(application_form: @application_choice.application_form) %>
 
     <% if FeatureFlag.active?('provider_view_safeguarding') %>
-      <%= render ProviderInterface::SafeguardingDeclarationComponent.new(application_form: @application_choice.application_form)%>
+      <% if current_provider_user.can_view_safeguarding_information_for?(@application_choice.provider) %>
+        <%= render ProviderInterface::SafeguardingDeclarationComponent.new(application_form: @application_choice.application_form)%>
+      <% end %>
     <% end %>
 
     <h2 class="govuk-heading-m govuk-!-margin-top-8" id="qualifications">Qualifications</h2>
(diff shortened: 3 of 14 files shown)


app/views/provider_interface/provider_users/edit_providers.html.erb
```diff
@@ -20,6 +20,8 @@
                   <%= ppf.hidden_field :provider_id %>
                   <%= ppf.govuk_check_boxes_fieldset :permissions, legend: { text: 'Choose permissions', size: 's' } do %>
                     <%= ppf.govuk_check_box :manage_users, true, multiple: false, label: { text: 'Manage users' } %>
+                    <%= ppf.govuk_check_box :view_safeguarding_information, true, multiple: false,
```
mnacos, line 23: What does `multiple:` mean here?

spec/models/provider_user_spec.rb
```diff
@@ -78,6 +78,25 @@
     end
   end
 
+  describe '#can_view_safeguarding_information_for?' do
+    let(:provider_user) { create :provider_user, :with_provider }
+
+    it 'is false without the correct permission' do
+      expect(provider_user.can_view_safeguarding_information_for?(build_stubbed(:provider, id: 1))).to be false
```
mnacos, line 85: Would this be slightly better with:

```
      provider = provider_user.providers.first
      expect(provider_user.can_view_safeguarding_information_for?(provider)).to be false
```

Just to prove provider affiliation alone does not grant view rights.

app/views/provider_interface/application_choices/show.html.erb
```diff
@@ -52,7 +52,9 @@
     <%= render ProviderInterface::TrainingWithDisabilityComponent.new(application_form: @application_choice.application_form) %>
 
     <% if FeatureFlag.active?('provider_view_safeguarding') %>
-      <%= render ProviderInterface::SafeguardingDeclarationComponent.new(application_form: @application_choice.application_form)%>
+      <% if current_provider_user.can_view_safeguarding_information_for?(@application_choice.provider) %>
```
mnacos, line 55: I wonder if this should be:

```suggestion
      <% if current_provider_user.can_view_safeguarding_information_for?(@application_choice.offered_option.provider) %>
```

offered_option will return the original course option unless an alternative one has been offered by the provider. The alternative one may be belong to a different provider that the user is associated with.

app/models/support_interface/provider_user_form.rb
```diff
@@ -66,7 +66,11 @@ def provider_permissions=(attributes)
           provider_id: form.provider_permission[:provider_id],
           provider_user_id: provider_user.try(:id),
         )
-        permission.manage_users = form.provider_permission.fetch(:manage_users, false)
+
+        ProviderPermissions::VALID_PERMISSIONS.each do |permission_name|
+          permission.send("#{permission_name}=", form.provider_permission.fetch(permission_name, false))
```
mnacos, line 71: I couldn't decide at first, but I think I like this. :)

spec/system/provider_interface/manage_provider_user_permissions_spec.rb
```diff
@@ -90,13 +95,31 @@ def when_i_remove_a_permissions_from_a_provider_user
...
+
+    click_on 'Update providers'
+  end
+
+  def then_i_can_see_the_view_safeguarding_permission_for_the_provider_user
+    within("#provider-#{@provider.id}-enabled-permissions") do
+      expect(page).to have_content 'View safeguarding information'
+    end
+
+    expect(@managed_user.provider_permissions.first.view_safeguarding_information).to be true
```
duncanjbrown, line 120: It would be good not to need to inspect the DB as part of a system spec. Could we assert that the checkbox is checked?

app/components/provider_interface/provider_permissions_list_component.html.erb
```diff
@@ -1,14 +1,18 @@
...
-      <li>
-        <% if provider_permission.send(permission_name) %>
-          <%= permission_name.to_s.humanize %>
-        <% else %>
-            No permissions
-        <% end %>
-      </li>
+    <%
+      humanized_permissions = [].tap do |ary|
+        ProviderPermissions::VALID_PERMISSIONS.each do |permission_name|
```
duncanjbrown, line 6: This is quite gnarly for the view — maybe a helper method on the component?

app/components/provider_interface/provider_permissions_list_component.html.erb
```diff
@@ -1,14 +1,18 @@
 <% visible_provider_permissions.each do |provider_permission| %>
```
duncanjbrown, line 1: I just noticed this hides the provider altogether when no permissions are set:

When manage users permission is set for Oriel:
Screenshot 2020-05-18 at 09 24 44

When no permissions are set:
Screenshot 2020-05-18 at 09 25 53

I'll add a separate card

spec/system/support_interface/managing_provider_users_spec.rb
```diff
@@ -179,6 +187,14 @@ def and_they_should_be_able_to_manage_users
     end
   end
 
+  def and_they_should_be_able_to_view_safeguarding_information
+    expect(@user.reload.provider_permissions.view_safeguarding_information.first.provider).to eq(@provider)
```
duncanjbrown, line 191: Could we get away without this assertion?

spec/models/provider_user_spec.rb
```diff
@@ -78,6 +78,25 @@
...
+    end
+
+    it 'is false without the correct permission for the given provider' do
+      provider_user.provider_permissions.update_all(view_safeguarding_information: true)
+      expect(provider_user.can_view_safeguarding_information_for?(build_stubbed(:provider, id: 1))).to be false
```
duncanjbrown, line 90: Do we need an `id` on this one?

app/views/provider_interface/application_choices/show.html.erb
duncanjbrown: This is a good question! I think we should tackle it separately to this PR, though?

app/views/provider_interface/provider_users/edit_providers.html.erb
steventux: https://govuk-form-builder.netlify.app/form-elements/checkboxes/#generating-a-single-checkbox
I think because this isn't a multiple selection on the same attribute it's necessary.

spec/system/provider_interface/manage_provider_user_permissions_spec.rb
steventux: I agree, I think this was a pattern I adopted to debug the system specs quickly but we should use expect the UI to reflect db state. Line 117 does this as it's the output from the permissions list component.

app/components/provider_interface/provider_permissions_list_component.html.erb
steventux: That's odd I definitely see the No permissions message:
image

steventux:
@mnacos @duncanjbrown I think I've addressed all comments apart from https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2057#discussion_r425233371 which may need some discussion.

app/models/support_interface/provider_user_form.rb
steventux: Metaprogramming considered harmful! 
I wasn't sure about it either, but this seems like a valid use as it simplifies adding further permissions using the `ProviderPermissions::VALID_PERMISSIONS` constant.

app/components/provider_interface/provider_permissions_list_component.html.erb
duncanjbrown: Probably me mixing up my branches locally, I'll check

duncanjbrown: My bad: I removed `manage users` permission for my current user on one of the provider I had access to. Not a bug

spec/system/provider_interface/see_individual_application_spec.rb
```diff
@@ -44,6 +45,11 @@ def and_the_safeguarding_declaration_feature_flag_is_active
     FeatureFlag.activate('provider_view_safeguarding')
   end
 
+  def and_i_am_permitted_to_see_safeguarding_information
+    ProviderUser.find_by(dfe_sign_in_uid: 'DFE_SIGN_IN_UID')
+      .provider_permissions.update_all(view_safeguarding_information: true)
+  end
+
```
tijmenb, line 52: @steventux I just accidentally broke this feature while rebasing, but the tests still pass. Can you add a test that the provider user _can't_ see the safeguarding data without the correct permission?

steventux: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2097

PR #2060, opened 2020-05-13
Migration to add `features` table (for auditing)

stevehook:
## Context

Support staff need to be able see when a particular feature was activated or deactivated and by whom. This is a migration only PR to create the `features` table that we will use together with the `audited` gem to audit feature flag updates.

- [x] Create a `features` table and associated unique index for the `name` column.

## Guidance to review

- Using the name `Feature` for the new model to avoid clashing with `FeatureFlag` service class so that we don't need to break the existing API and/or overload the model with lots of business logic.
- Anything missing from the schema?
- Update issues? We don't need to backfill this table because it will need to be populated 'on-demand' whenever we reference a feature flag. Absence of a record for a particular flag always means 'inactive'.

## Link to Trello card

https://trello.com/c/hkzUvtMz/1440-%F0%9F%8F%88-audit-log-of-when-features-were-enabled

## Things to check


stevehook:
Considering renaming the table/persistent model to `features`/`Feature`. This makes sense as a name and avoids the clash with the existing `FeatureFlag` service class.

tvararu:
@davidgisbey the description and owners can stay in the code, as I don't think we gain anything by putting them in the database

PR #2063, opened 2020-05-13
Add confirm completion checkbox for personal details section

davidgisbey:
## Context

First PR is #2051 - Adds a bunch of columns to the application_forms table for the completion of each section.

Currently, the majority of the sections of the application form, do not need to be explicitly marked as complete. This means that many sections are marked as complete by default as they still meet the logic required to do so. 

- Adds a confirm completion checkbox to the personal details section.

before

image

after

image

## Link to Trello card

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/personal_details_controller.rb
+++ b/app/controllers/candidate_interface/personal_details_controller.rb
@@ -28,22 +28,9 @@ module CandidateInterface
     end
 
     def complete
-      @application_form = current_application
+      current_application.update!(application_form_params)
 
-      if PersonalDetailsForm.build_from_application(current_application).valid?
-        current_application.update!(application_form_params)
-
-        redirect_to candidate_interface_application_form_path
-      else
-        flash[:warning] = 'You can’t mark this section as complete without adding all your personal details.'
-
-        current_application.personal_details_completed = false
-
-        personal_details_form = PersonalDetailsForm.build_from_application(current_application)
-        @personal_details_review = PersonalDetailsReviewPresenter.new(form: personal_details_form)
-
-        render :show
-      end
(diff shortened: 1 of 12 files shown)


app/controllers/candidate_interface/personal_details_controller.rb
```diff
@@ -21,14 +22,33 @@ def update
...
+
+        render :show
+      end
+    end
+
   private
 
+    def set_application_form_to_current_application
+      @application_form = current_application
+    end
```
tijmenb, line 50: Can we use `current_application` in the code directly?

davidgisbey: We absolutely can. It's just used in pretty much all the views so I thought it would make sense to create it before. I'll continue to use current_applicatiion in the controller actions.

app/controllers/candidate_interface/personal_details_controller.rb
```diff
@@ -21,12 +22,30 @@ def update
...
     end
 
+    def complete
+      @application_form = current_application
+
+      if PersonalDetailsForm.build_from_application(current_application).valid?
+        current_application.update!(application_form_params)
+
+        redirect_to candidate_interface_application_form_path
+      else
```
tvararu, line 37: I don't think this `else` clause can trigger under normal use. The user can't reach this form if they haven't filled in their personal details successfully. Do we need it?

davidgisbey: Yea I meant to left a comment saying the else should never trigger. It can come out 💯

tvararu: New version is shorter and nicer! 👍

PR #2067, opened 2020-05-13
Fix position of tag on application page

willmcb:
## Context

Fix to position of 'status' tag on application card had side effect on application show view. This PR fixes that.

**Before**
Screenshot 2020-05-14 at 08 56 22

**After**
Screenshot 2020-05-14 at 08 56 49

## Guidance to review

- Load the app - look at the any application summary.

## Link to Trello card

Was side effect of this: https://trello.com/c/2gdRmYDC/2115-title-notification-dot-on-application-list-in-wrong-place-sometimes

## Things to check

### DIFF
--- a/app/components/provider_interface/application_card_component.html.erb
+++ b/app/components/provider_interface/application_card_component.html.erb
@@ -3,7 +3,9 @@
   <h3 class="govuk-heading-m">
     <%= govuk_link_to candidate_name, provider_interface_application_choice_path(application_choice) %>
     <div class="app-application-card__status">
-      <%= render ProviderInterface::ApplicationStatusTagComponent.new(application_choice: application_choice) %>
+      <div>
+        <%= render ProviderInterface::ApplicationStatusTagComponent.new(application_choice: application_choice) %>
+      </div>
       <% if most_recent_note %>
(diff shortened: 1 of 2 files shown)


app/components/provider_interface/application_card_component.html.erb
```diff
@@ -3,7 +3,9 @@
   <h3 class="govuk-heading-m">
     <%= govuk_link_to candidate_name, provider_interface_application_choice_path(application_choice) %>
     <div class="app-application-card__status">
-      <%= render ProviderInterface::ApplicationStatusTagComponent.new(application_choice: application_choice) %>
+      <div>
```
duncanjbrown, line 6: Rather than a spare `<div>`, would it be better to  somehow pass a class down to TagComponent which gives it `display: block`?

tvararu: We have tried that. However, that also stretches it to fill the container, see here: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2035#pullrequestreview-409319761. You do need a separate block wrapping element with the way the HTML is currently structured.

PR #2070, opened 2020-05-14
Prevent code reuse between namespaces

tijmenb:
## Context

Splitting the app into namespaces is one of the approaches we use to avoid this application from becoming a big ball of spagetti ([see the ADR](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/adr/0005-application-structure.md)).

I'm looking into moving [a lot of the components into their appropriate namespaces](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/tree/933d7bbd5490f480bd32b874384b25ad55935b08/app/components), because there are way too many at the root-level. To do this properly I needed a way of quickly checking if a component was used in a single namespace or multiple. An automated test comes in handy. 

Add a test that fails if code from one namespace is used in another. Fixes a few minor violations.

## Link to Trello card

Related to https://trello.com/c/qfGonFDe.

## Things to check


spec/lib/namespaces_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+RSpec.describe 'Namespace' do
+  namespaces = %w[CandidateInterface ProviderInterface SupportInterface APIDocs RefereeInterface VendorAPI]
+
+  namespaces.each do |namespace|
+    describe namespace.constantize do
+      (namespaces - [namespace]).each do |other_namespace|
+        it "does not use code from the #{other_namespace} namespace" do
+          command = "grep -rnw app/**/#{namespace.underscore} -e '#{other_namespace}'"
+          usage_of_namespaced_code_in_other_namespace = `#{command}`
+          expect(usage_of_namespaced_code_in_other_namespace).to be_empty, usage_of_namespaced_code_in_other_namespace
```
malcolmbaig, line 12: Repeat of `usage_of_namespaced_code_in_other_namespace`. Not sure that's doing anything?

tijmenb: The second argument makes sure that the full `grep` output is displayed if there are multiple matches.

Without the second arg:

```
  1) Namespace CandidateInterface does not use code from the ProviderInterface namespace
     Failure/Error: expect(usage_of_namespaced_code_in_other_namespace).to be_empty
       expected `"app/controllers/candidate_interface/application_form_controller.rb:7:      @application_form_present..._again')\napp/controllers/candidate_interface/degrees/base_controller.rb:11:    ProviderInterface\n".empty?` to return true, got false
     # ./spec/lib/namespaces_spec.rb:12:in `block (5 levels) in <top (required)>'
``` 

With:

```
  1) Namespace CandidateInterface does not use code from the ProviderInterface namespace
     Failure/Error: expect(usage_of_namespaced_code_in_other_namespace).to be_empty, usage_of_namespaced_code_in_other_namespace
     
       app/controllers/candidate_interface/application_form_controller.rb:7:      @application_form_presenter = ProviderInterface::ApplicationFormPresenter.new(current_application)
       app/controllers/candidate_interface/application_form_controller.rb:21:      render_404 and return unless ProviderInterface::FeatureFlag.active?('apply_again')
       app/controllers/candidate_interface/degrees/base_controller.rb:11:    ProviderInterface
     # ./spec/lib/namespaces_spec.rb:12:in `block (5 levels) in <top (required)>'
```

PR #2071, opened 2020-05-14
Change approach for sharing UI with support interface

tijmenb:
## Context

This PR intends to solve two sets of related problems all having to do with the sharing of code between candidate, provider and support interfaces.

### Components mixup

* There's a large list of components in app/components that are sometimes used by one namespace, sometimes by two or three. There are also namespace-specific components in app/components/provider_interface, app/components/candidate_interface, etc. This makes it confusing where to put a new component.
* This situation sometimes causes problems with data separation. In https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1121 we introduced a feature the support section that we accidentally displayed to providers (fixed in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1267). We need a more consistent separation to make sure this doesn't happen again.
* Having a lot of components in the root namespace causes difficult situations, where we use namespaced components inside a non-namespaced component.

### Support UI

* The support interface sometimes lags behind provider and candidate interfaces, because they're usually prioritised. It would be better if we don't have to worry about the interfaces getting out of sync. 
* It can be sometimes hard to spot differences in how we display information to candidates and providers.

Instead of _sharing_ components, explicitly _import_ partials from provider and candidate interface into the support interface, and use them as such. Arguably this is also sharing code between the interfaces, but it's much more preferable to sharing because it's much more explicit. The result is that when we update the provider UI, the support UI is also updated in exactly the same way, and there's no risk of it going wrong.

It has the added benefit of being a lot more explicit and informative to the support users - they can now diagnose issues with the provider interface from support.

Once this is merged, we can move most components that are currently shared between Provider interface and Candidate interface (for example, the `LanguageSkillsComponent`) into the ProviderInterface namespace.

image

## Guidance to review

Does this make sense?

## Things to check

### DIFF
--- a/app/views/support_interface/application_forms/show.html.erb
+++ b/app/views/support_interface/application_forms/show.html.erb
@@ -62,25 +62,42 @@
   <% end %>
 <% end %>
 
-<h2 class="govuk-heading-l govuk-!-margin-top-8" id="qualifications">Qualifications</h2>
 
-<%= render QualificationsComponent.new(application_form: @application_form) %>
+<h2 class="govuk-heading-l govuk-!-margin-top-8">Application data</h2>
 
-<h2 class="govuk-heading-l govuk-!-margin-top-8">Work history</h2>
+<div class="govuk-tabs" data-module="govuk-tabs">
+  <h2 class="govuk-tabs__title">
+    Contents
+  </h2>
+  <ul class="govuk-tabs__list">
+    <li class="govuk-tabs__list-item govuk-tabs__list-item--selected">
+      <a class="govuk-tabs__tab" href="#what-the-candidate-sees">
+        What the candidate sees
+      </a>
+    </li>
+    <li class="govuk-tabs__list-item">
+      <a class="govuk-tabs__tab" href="#what-the-provider-sees">
+        What the provider sees
+      </a>
+    </li>
+  </ul>
 
-<%= render WorkHistoryReviewComponent.new(application_form: @application_form, editable: false, heading_level: 3) %>
+  <div class="govuk-tabs__panel govuk-tabs__panel--hidden" id="what-the-candidate-sees">
+    <h2 class="govuk-heading-l">What the candidate sees</h2>
+    <%= render 'candidate_interface/application_form/application_form_specific_data', application_form: @application_form, editable: false, missing_error: false %>
+  </div>
 
-<h2 class="govuk-heading-l govuk-!-margin-top-8">Unpaid experience and volunteering</h2>
-
-<%= render VolunteeringReviewComponent.new(application_form: @application_form, editable: false, heading_level: 3) %>
-
-<h2 class="govuk-heading-l govuk-!-margin-top-8">Language skills</h2>
-
-<%= render LanguageSkillsComponent.new(application_form: @application_form) %>
-
-<h2 class="govuk-heading-l govuk-!-margin-top-8">Personal statement</h2>
-
-<%= render PersonalStatementComponent.new(application_form: @application_form) %>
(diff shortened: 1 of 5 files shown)


app/views/support_interface/application_forms/show.html.erb
```diff
@@ -62,25 +62,33 @@
...
-
-<%= render PersonalStatementComponent.new(application_form: @application_form) %>
+<h2 class="govuk-heading-l govuk-!-margin-top-8">Application data</h2>
+
+<div class="govuk-tabs" data-module="govuk-tabs">
+  <h2 class="govuk-tabs__title">
+    Contents
+  </h2>
+  <ul class="govuk-tabs__list">
+    <li class="govuk-tabs__list-item">
```
tvararu, line 73:
```suggestion
    <li class="govuk-tabs__list-item govuk-tabs__list-item--selected">
```

app/views/support_interface/application_forms/show.html.erb
```diff
@@ -62,25 +62,33 @@
...
+      </a>
+    </li>
+    <li class="govuk-tabs__list-item">
+      <a class="govuk-tabs__tab" href="#what-the-candidate-sees">
+        What the candidate sees
+      </a>
+    </li>
+  </ul>
+
+  <div class="govuk-tabs__panel govuk-tabs__panel--hidden" id="what-the-provider-sees">
```
tvararu, line 85:
```suggestion
  <div class="govuk-tabs__panel" id="what-the-provider-sees">
```

tvararu: I think the selected panel is not meant to be hidden in the initial server-side render?

app/views/support_interface/application_forms/show.html.erb
tvararu: Seems like one of the panels should be selected by default, according to http://govuk-frontend-review.herokuapp.com/components/tabs

app/views/support_interface/application_forms/show.html.erb
tijmenb: I got this from the Design System: https://design-system.service.gov.uk/components/tabs/

tvararu: Seems legit then, `--hidden` is probably unused without `.js-enabled` on body

fofr:
Nice. I love the tabs and the different views.

- Does the provider tab always show, even when not submitted?
- Should the tab indicate that state better?
- Should candidate tab be first?

With components, what happens for components that are explicitly for support (I'm thinking of things like references where we show chase dates, etc)

tijmenb:
> Does the provider tab always show, even when not submitted?
> Should the tab indicate that state better?
> Should candidate tab be first?

Great point, I've swapped the tabs and hidden the content for unsubmitted applications. This might also prevent crashes where we try to show missing data. 

> With components, what happens for components that are explicitly for support (I'm thinking of things like references where we show chase dates, etc)

They'll stay like they're now. These tabs are for the "data" that doesn't need to be made specific for support UIs.

Things that will have support-specific layouts (and aren't in the tabs) are: 

- application / personal details (have things like states and more dates)
- course choices (which includes course availability warnings and edit buttons)
- references (which include the actual feedback)
- Criminal convictions and professional misconduct (for the reasons)

Does that make sense?

tijmenb:
I'm going to close this for a bit because I'm not sure what to do about the safeguarding information. It's now scoped to a provider user, which means we shouldn't show it in the support UI. I'll revisit when I have the time.

PR #2076, opened 2020-05-14
DSI fallback: expire magic link tokens after an hour

duncanjbrown:
## Context

We don't expire these tokens so they last forever (or until the user regenerates them).

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1989#discussion_r418512600

Pull the token handling code out of the controller and test it, then add the constraint that tokens must be < 1 hour old.

## Link to Trello card

https://trello.com/c/t421tPrM/2073-deal-with-may-1st-dfe-signin-outage


tvararu:
https://trello.com/c/JXRnucRj/265-add-expiry-to-provider-magic-link-sign-in

PR #2077, opened 2020-05-14
Test content of Download Application as PDF response

steventux:
## Context

Introducing middleware to add compression recently caused PDF downloads to fail.
PDFKit doesn't modify the status of the response if it encounters errors, it probably should but we don't control that.

Adds a way to read the text from the PDF response and verify the candidate name and support reference are present.

## Things to check

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -98,6 +98,7 @@ group :test do
   gem 'clockwork-test'
   gem 'deepsort'
   gem 'rspec-benchmark'
+  gem 'pdf-reader'
 end
 
 group :development, :test do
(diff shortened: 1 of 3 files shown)


tvararu:
The test breaks for me locally. We can pair on it when you have some time. 👍

duncanjbrown:
Shall we close this for the time being?

steventux:
Closing as we think we're experiencing this bug in tests https://github.com/wkhtmltopdf/wkhtmltopdf/issues/2148 also this https://github.com/pdfkit/pdfkit/pull/469 might solve the testing by status code issue.

PR #2083, opened 2020-05-15
Feature flag audit log and migration to PostgreSQL

stevehook:
## Context

Support staff need to be able see when a particular feature was activated or deactivated and by whom. This is  PR #3 for this card that switches reading of the feature flags from Redis to the `features` table in PostgreSQL. It will need to be deployed separately to and after https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2081

- [x] Switch `FeatureFlag.active?` etc. over to using PostgreSQL. API stays the same.
- [x] Auditing is already configured https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2081 and should start happening when we set up
- [x] Display the audit log for feature flags in the support interface.
- [x] Change base to `master` and rebased when upstream branch merged.
- [x] Update system specs for support interface

## Guidance to review

- We need to deploy this after the code change that migrates data from Redis to PostgreSQL. Ideally it would be two deploys in quick succession. Is there a better way to roll this out?
- There are not too many changes in areas of the code base that are dependent on the `FeatureFlag` API. Have I missed anything?
- We can drop the `rollout` gem if we remove the migrator class in future to keep our dependencies lean.

image

## Link to Trello card
 
https://trello.com/c/hkzUvtMz/1440-%F0%9F%8F%88-audit-log-of-when-features-were-enabled

## Things to check

### DIFF
--- a/app/components/support_interface/feature_audit_trail_component.html.erb
+++ b/app/components/support_interface/feature_audit_trail_component.html.erb
@@ -1,4 +1,4 @@
-<ul>
+<ul class="govuk-list govuk-list--bullet">
   <% audits.select { |audit| visible?(audit) }.each do |audit| %>
     <li>
       <%= l(audit.created_at.to_date, format: :long) %> <%= l(audit.created_at, format: :time) %> - <%= action_label_for(audit) %> <%= active_value_for(audit) %> by <%= user_label_for(audit) %>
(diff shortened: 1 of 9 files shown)


tvararu:
Screenshot 2020-05-18 at 17 06 19

Looks nice! I think if we don't have a known user, we could just say `Created Inactive`

app/components/support_interface/feature_audit_trail_component.html.erb
```diff
@@ -0,0 +1,7 @@
+<ul>
```
tvararu, line 1:
```suggestion
<ul class="govuk-list govuk-list--bullet">
```

stevehook:
> Looks nice! I think if we don't have a known user, we could just say `Created Inactive`. What do you think?

Thanks that's a good idea, done.

PR #2087, opened 2020-05-18
Add last completed checkboxes for each application form section

davidgisbey:
## Context

First PR is #2051 - Adds a bunch of columns to the application_forms table for the completion of each section.

Second PR is #2063 - Adds the implementation for personal details

Third PR is #2086 - Adds a completed checkbox for the first 5 sections

Currently, the majority of the sections of the application form, do not need to be explicitly marked as complete. This means that many sections are marked as complete by default as they still meet the logic required to do so. 

The following sections still need to be added.

- [x] personal statement 
- [x] subject knowledge
- [x] interview needs
- [x] references

For each of the above:

- Adds a route
- Adds a checkbox
- The course presenter uses the attribute_completed value for section completion
- If a section is complete and they edit the section it sets the value to false (incomplete)

## Guidance to review

This should be a lot easier to review than it looks if you review by commit. Most are very similar.

A further PR to backfill data will need to be raised.

## Link to Trello card

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/referees_controller.rb
+++ b/app/controllers/candidate_interface/referees_controller.rb
@@ -120,7 +120,7 @@ module CandidateInterface
 
         redirect_to candidate_interface_application_form_path
       else
-        flash[:warning] = 'You can’t mark this section complete without adding two referees.'
+        flash[:warning] = "You can’t mark this section complete without adding #{ApplicationForm::MINIMUM_COMPLETE_REFERENCES} referees."
         current_application.references_completed = false
         @application_form = current_candidate.current_application
 
--- a/app/controllers/candidate_interface/referees_controller.rb
+++ b/app/controllers/candidate_interface/referees_controller.rb
@@ -36,6 +36,8 @@ module CandidateInterface
 
         @reference_type_form.save(@referee)
 
+        current_application.update!(references_completed: false)
+
         redirect_to candidate_interface_review_referees_path
       else
         return render :type unless @reference_type_form.valid?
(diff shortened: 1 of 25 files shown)


app/controllers/candidate_interface/referees_controller.rb
```diff
@@ -109,6 +114,20 @@ def review
...
+        current_application.update!(application_form_params)
+
+        redirect_to candidate_interface_application_form_path
+      else
+        flash[:warning] = 'You can’t mark this section complete without adding two referees.'
+        current_application.references_completed = false
+        @application_form = current_candidate.current_application
+
+        render :review
+      end
```
tvararu, line 128: Interesting. I wondered here for a bit if we should instead hide the checkbox until two references are added, but that could be confusing to users.

app/controllers/candidate_interface/referees_controller.rb
```diff
@@ -109,6 +114,20 @@ def review
       @application_form = current_candidate.current_application
     end
 
+    def complete
+      if current_application.application_references.count >= ApplicationForm::MINIMUM_COMPLETE_REFERENCES
+        current_application.update!(application_form_params)
+
+        redirect_to candidate_interface_application_form_path
+      else
+        flash[:warning] = 'You can’t mark this section complete without adding two referees.'
```
tvararu, line 123: Could we add a tiny bit of gold plating to this and make the `two referees` based on `ApplicationForm::MINIMUM_COMPLETE_REFERENCES`?

PR #2091, opened 2020-05-19
Migration to backfill completed section booleans

davidgisbey:
## Context

~Do not merge until the whole feature has shipped.~ Edit: done

Candidates now need to mark sections of completed. A bunch of new attributes have been added for these completed sections.

Completed applications and applications in progress should be backfilled to reflect that candidates have the sections that auto-completed. 

- Completed applications show all sections completed (science for primary, but not secondary)
- In progress, applications use the completed logic in the ApplicationPresenter to populate the completed section booleans.

## Guidance to review

~This should not be merged until all other PRs have already been deployed.~ Edit: done

~When this is deployed we need to immediately switch the feature flag on.~ Edit: we have worked around this

~Is it possible to get a dump of prod data to test how long the migration would take to run?~ Edit: shouldn't matter

## Link to Trello card

https://trello.com/c/2YtoBPNB/1425-explicitly-mark-all-sections-as-complete-applyagain

## Things to check


tvararu:
> When this is deployed we need to immediately switch the feature flag on.

This isn't great. It means that if we discover a bug with the feature, that we can't easily switch it off. If we do, it means users can edit their applications into having `_completed` booleans that are desynced to what they should be. 🤔 

We've managed to work around these kind of situations, most recently when @stevehook migrated the feature flags to Postgres. The idea we used is that we can start using the boolean fields in a write-only way before we backfill and before we switch on the feature flag.

In this case, we could set `update! personal_details_completed: true` when a user currently submits their personal details form when the feature flag is deactivated. Combined with the backfill in this PR, this will sync everyone's `_completed` fields correctly, and syncing will be maintained if we toggle the feature flag on and off.

This sounds like it could be quite a bit of work, as you'd need to go back and retrofit some more code to all the affected controllers. If you like we can have a bit of a kick-off chat if I didn't explain it well enough, or maybe we can think of an easier solution. Give me a shout either way 👍

db/migrate/20200526122730_populate_completed_section_booleans.rb
```diff
@@ -0,0 +1,50 @@
...
+                               safeguarding_issues_completed: presenter.safeguarding_completed?,
+                               becoming_a_teacher_completed: presenter.becoming_a_teacher_completed?,
+                               subject_knowledge_completed: presenter.subject_knowledge_completed?,
+                               interview_preferences_completed: presenter.interview_preferences_completed?,
+                               references_completed: presenter.all_referees_provided_by_candidate?)
+    end
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
```
stevehook, line 48: I think it's ok to have this as a no-op (I must admit I do that for backfill `down`s) but I guess this is correct strictly speaking.

db/migrate/20200526122730_populate_completed_section_booleans.rb
```diff
@@ -0,0 +1,50 @@
...
+        })
+      end
+
+      application_form.update!(fields_to_update)
+    end
+
+    in_progress_application_forms = ApplicationChoice.includes(:application_form).unsubmitted.map(&:application_form).uniq
+    started_application_forms = in_progress_application_forms.reject(&:blank_application?)
+
+    started_application_forms.each do |application_form|
```
stevehook, line 31: ~~Doesn't really matter with data volumes as they are but this should be a `find_each` to batch the results.~~

Actually scratch that it's a bit more complicated than that because `find_each` would be needed before the `map` and `reject`.

PR #2093, opened 2020-05-19
Add 3rd Nudge Email for incomplete references 

malcolmbaig:
## Context

We already send nudge emails for incomplete references in the following
scenarios:

- Nudge both parties if no response after 5 working days.
- Ask candidate if they want to change referees after 10 working days.

We want to send a 3rd nudge email after 20 working days.

See commits.
## Guidance to review
- Commit-based review recommended as there's some refactoring/renaming work in this PR.

## Link to Trello card
https://trello.com/c/qsAuZsU7

## Things to check


app/services/send_reference_chase_email_to_referee_and_candidate.rb
```diff
@@ -1,4 +1,4 @@
-class SendChaseEmailToRefereeAndCandidate
+class SendReferenceChaseEmailToRefereeAndCandidate
```
stevehook, line 1: 👍

app/lib/time_limit_config.rb
```diff
@@ -37,6 +37,10 @@ def self.replace_referee_by
     14
   end
 
+  def self.additional_reference_chase_calendar_days
+    28
```
stevehook, line 41: The requirement says 20 business days. Are we doing it this way to avoid all the Covid-19 business time 'holidays'?

spec/support_specs/clockwork_spec.rb
```diff
@@ -4,7 +4,7 @@
 require './app/workers/clockwork_check'
 require './app/workers/send_applications_to_provider_worker'
 require './app/workers/decline_offers_by_default_worker'
-require './app/workers/send_chase_email_to_referees_worker'
+require './app/workers/send_chase_email_to_candidates_worker'
 require './app/workers/send_chase_email_to_providers_worker'
```
stevehook, line 8: Do you need to `require` the new workers here too? I guess not, in which case I wonder why we have these here at all.

spec/system/referee_interface/referee_does_not_respond_spec.rb
```diff
@@ -52,4 +55,16 @@ def then_an_email_is_sent_to_the_candidate_asking_for_a_new_referee
 
     expect(current_email.subject).to have_content('Give details of a new referee')
   end
+
+  def when_the_candidate_does_not_respond_within_20_working_days
+    Timecop.travel(29.business_days.from_now) do
```
stevehook, line 60: This should be `28.days.from_now` if we are using calendar days for the limit I think. And the method name is misleading.

app/lib/time_limit_config.rb
tvararu: Yes, seems good to me

app/workers/send_additional_reference_chase_email_to_candidates_worker.rb
```diff
@@ -0,0 +1,23 @@
+class SendAdditionalReferenceChaseEmailToCandidatesWorker
+  include Sidekiq::Worker
+
+  def perform
+    references_to_chase.each do |reference|
+      CandidateMailer.chase_reference_again(reference).deliver_later
+      ChaserSent.create!(chased: reference, chaser_type: :additional_reference_request)
```
tijmenb, line 7: Two questions:

- I wonder if we need to check here if the application has been submitted. With the replacement feature it's possible for the candidate to have 2 complete references and 1 still waiting. I think that case we wouldn't want to send a nudge.
- This sends a nudge email per outstanding reference. I think we should perhaps send it per candidate? Otherwise the candidate would receive 2 identical emails if they have both references outstanding.

To avoid similar issues, can we print out a list of emails that would be sent on prod, so we can sense check it?

spec/system/referee_interface/referee_does_not_respond_spec.rb
malcolmbaig: You're right, references to time aren't clear at all in this spec. Updated, including renaming the two existing methods that invoke Timecop.

app/workers/send_additional_reference_chase_email_to_candidates_worker.rb
malcolmbaig: Agreed on both counts. Will work something up. The existing nudge emails appear subject to the same issues, so will review those as well. 

Regarding this:

> To avoid similar issues, can we print out a list of emails that would be sent on prod, so we can sense check it?

Do you mean log these on each execution of the worker, or something else?

tijmenb: Sorry, that sentence wasn't very clear :) 

I meant doing some kind of dry-run on production so that we can produce a list of emails that we will send once this is deployed, and have that reviewed by a PM.

As we don't backfill the existing applications, I'm expecting 50-60 emails to be sent after deploy (https://www.apply-for-teacher-training.education.gov.uk/integrations/performance-dashboard).

malcolmbaig: @tijmenb New commits added with the suggested changes.

PR #2094, opened 2020-05-19
Add provider_relationship_permissions table

steventux:
## Context

We're working out how to implement provider permissions at an organisational level. We'll need somewhere to store these permissions.

Adds underlying `provider_relationship_permissions` table for provider-to-provider permissions.
Adds `view_safeguarding_information` permission as part of the table definition.

## Guidance to review

Naming things is hard, please suggest a good name for this permissions model. (`ProviderPermissions` is currently in use for user level permissions, we could change this though).

Some other names:

- `ProviderToProviderPermissions`
- `RelatedProviderPermissions`
- `ProviderOrganisationPermissions`

## Link to Trello card

https://trello.com/c/F6QCZXqx/2169-provider-to-provider-permissions-affect-what-users-can-do-in-the-system

## Things to check


duncanjbrown:
I like the way you've prevented STI by naming the columns. 👨‍⚕️ 

I think the name ought to have "permissions" in it. `ProviderRelationshipPermissions`?

steventux:
@duncanjbrown Ah! that wasn't intentional when naming the columns. This is very much a discussion PR. Happy to go with STI as we all agreed this was a good case for it.

duncanjbrown:
Oh yeah, we still need STI don't we. Sorry — my bad

steventux:
@duncanjbrown I've amended the table name to `provider_relationship_permissions` and the relationship columns to `provider_id` and `related_provider_id` so this should work with STI. There's also an added index for `type` + `provider_id`

db/migrate/20200519102317_add_provider_relationship_permissions.rb
```diff
@@ -0,0 +1,28 @@
...
+    add_foreign_key :provider_relationship_permissions, :providers, column: :related_provider_id
+
+    add_index(
+      :provider_relationship_permissions,
+      %i[type provider_id],
+      unique: true,
+    )
+    add_index(
+      :provider_relationship_permissions,
+      %i[provider_id related_provider_id],
```
duncanjbrown, line 23: Wouldn't this cause a bit of trouble? If we had provider 1 ratifying courses at provider 2, then provider 2 would define two relationships: the "as ratifying provider" relationship and the "as training provider" relationship. We'd have to make sure we didn't put the same IDs in `provider_id`/`related_provider_id` in each case, which in practice would mean checking for an existing relationship then reversing the positions. 

Would it work to rename these columns as they were (`training_provider_id` and `ratifying_provider_id`) and making a single, three-column unique index of `ratifying_provider_id`, `training_provider_id` and `type`?

steventux: I would expect that to produce 2 records:

```
<Provider id: 1> # Ratifying provider
<Provider id: 2> # Training provider

# The STI relationships:
<AccreditedBodyPermissions provider_id: 1, related_provider_id: 2, a_permission: false>
<TrainingProviderPermissions provider_id: 2, related_provider_id: 1, a_permission: true>
```

If we condensed these into a single record with specific column names for ratifying and training providers would we then lose the granularity of the permissions on each relationship?

I like the compound index idea and I'm happy to make the column names more specific to the relationship, maybe I'm misunderstanding the use of the `type` here.

duncanjbrown: Ah, I think we agree. I'm proposing we make the role of each provider explicit in the column name. I might well be missing something, though

```
AccreditedBodyPermissions.create!(training_provider: 1, accredited_body: 2) => OK
TrainingProviderPermissions.create!(training_provider: 1, accredited_body: 2) => OK
AccreditedBodyPermissions.create!(training_provider: 1, accredited_body: 2) => 💥
```

PR #2095, opened 2020-05-19
Deal with errors from find sync

tijmenb:
## Context

Calls to Find sometimes fail, which is expected. The problem is that it generates Sentry errors that aren't actionable.

- Ignore errors when making requests to Find in the sync
- Keep track of the last time the sync succeeded
- Sound the alarm when it hasn't happened successfully in an hour 

## Link to Trello card

https://trello.com/c/VG5mhgFw

## Things to check

### DIFF
--- a/spec/services/sync_all_providers_from_find_spec.rb
+++ b/spec/services/sync_all_providers_from_find_spec.rb
@@ -19,6 +19,21 @@ RSpec.describe SyncAllProvidersFromFind do
       expect { SyncAllProvidersFromFind.call }.to change { Provider.count }.by(2)
     end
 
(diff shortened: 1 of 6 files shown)


spec/services/sync_all_providers_from_find_spec.rb
```diff
@@ -19,6 +19,21 @@
       expect { SyncAllProvidersFromFind.call }.to change { Provider.count }.by(2)
     end
 
+    it 'sets the last updated timestamp' do
```
vigneshmsft, line 22: another test for "it doesn't set the last updated timestamp" in case of exception? useful?

PR #2098, opened 2020-05-19
Grant the view safeguarding permission to all test users

steventux:
## Context

Missed in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2057

Adds the `view_safeguarding_information` permission for all test users.

## Link to Trello card

https://trello.com/c/bwEGeJa3/2093-build-access-mgmt-safeguarding

## Things to check

### DIFF
--- a/lib/tasks/local_dev.rake
+++ b/lib/tasks/local_dev.rake
@@ -11,7 +11,8 @@ task setup_local_dev_data: %i[environment copy_feature_flags_from_production syn
   end
   SupportUser.find_or_create_by!(dfe_sign_in_uid: 'dev-support', email_address: 'support@example.com')
 
-  ProviderPermissions.update_all(manage_users: true) if FeatureFlag.active?('provider_add_provider_users')
+  ProviderPermissions.update_all(manage_users: true)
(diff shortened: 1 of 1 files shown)


lib/tasks/local_dev.rake
```diff
@@ -12,6 +12,8 @@ task setup_local_dev_data: %i[environment copy_feature_flags_from_production syn
   SupportUser.find_or_create_by!(dfe_sign_in_uid: 'dev-support', email_address: 'support@example.com')
 
   ProviderPermissions.update_all(manage_users: true) if FeatureFlag.active?('provider_add_provider_users')
+
+  ProviderPermissions.update_all(view_safeguarding_information: true) if FeatureFlag.active?('provider_view_safeguarding')
```
duncanjbrown, line 16: When this runs on Heroku the flag will be inherited from production so it won't set the permission until we turn on the feature in prod — should we let it happen irrespective of flag? And above too?

steventux: Yes good idea, it won't do any harm granting these regardless

PR #2099, opened 2020-05-19
Move #edit_by to ApplicationForm

davidgisbey:
## Context

The `edit_by` field lives on each `application_choice` it is duplicated for each choice on an application form. It should live on the `application_form`.

This PR needs to add in new functionality behind a feature flag and update the application forms edit by when a course choice's edit by is updated with the feature flag on.

- Change the various services that deal with edit_by  to use the new `edit_by` attribute on ApplicationForm

## Guidance to review

An additional migration will need to happen which populates the edit_by attribute for each application form and then drops the edit_by column on the application_choices table.

## Link to Trello card

https://trello.com/c/IbvLIdCn/1573-move-editby-to-application-form

## Things to check


app/models/application_dates.rb
```diff
@@ -16,7 +16,7 @@ def decline_by_default_at
   end
 
   def edit_by
-    @application_form.application_choices.first&.edit_by
```
tvararu, line 19: This line is exactly why we're doing this work!

app/services/submit_application_choice.rb
```diff
@@ -7,11 +7,11 @@ def initialize(application_choice, apply_again: false, enough_references: false)
 
   def call
     if @apply_again && @enough_references
-      @application_choice.edit_by = Time.zone.now
+      @application_choice.application_form.update(edit_by: Time.zone.now)
```
tvararu, line 10: Here we are calling `application_form.update` up to 3 times with the ~same `edit_by` when we submit an application. This seems wrong now. Should we bubble this up to `SubmitApplication`? [There's already a convenient `update!(submitted_at` we can attach to.](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/a62645398c7abc4fd2282f3948d7f6e6014ec213/app/services/submit_application.rb#L13)

tvararu:
> An additional migration will need to happen which populates the edit_by attribute for each application form

I think that should be part of this PR, right? Otherwise merging this will lead to every application having a `edit_by: nil`.

app/services/submit_application_choice.rb
davidgisbey: Good spot. I've moved all the edit_by logic up to the submit application service 👍

davidgisbey:
Yea sorry I should have made it explicit in the comments the backfill needed to go in first.

davidgisbey:
I think this should be good for a re-review now 👍 The application forms edit_by is now updated when the applcation_choice is with the flag off.

app/services/submit_application.rb
```diff
@@ -10,7 +10,12 @@ def initialize(application_form)
 
   def call
     ActiveRecord::Base.transaction do
-      application_form.update!(submitted_at: Time.zone.now)
+      if application_form.apply_again? && enough_references_have_been_provided? && FeatureFlag.active?('move_edit_by_to_application_form')
```
tvararu, line 13: This should happen irrespective of the feature flag because we need it to work whether it's on or not:

```suggestion
      if application_form.apply_again? && enough_references_have_been_provided?
```

app/services/submit_reference.rb
```diff
@@ -38,10 +38,14 @@ def progress_applications
       reference_feedback_provided!
       application_form.application_choices.awaiting_references.each do |application_choice|
         ApplicationStateChange.new(application_choice).references_complete!
-        if application_form.apply_again?
-          application_choice.update(edit_by: Time.zone.now)
+        if application_form.apply_again? && !FeatureFlag.active?('move_edit_by_to_application_form')
+          application_choice.update!(edit_by: Time.zone.now)
```
tvararu, line 42:
```suggestion
        if application_form.apply_again?
          application_choice.update!(edit_by: Time.zone.now)
          application_form.update!(edit_by: Time.zone.now)
```

app/services/submit_reference.rb
```diff
@@ -38,10 +38,14 @@ def progress_applications
...
         end
       end
+
+      if application_form.apply_again?
+        application_form.update!(edit_by: Time.zone.now)
+      end
```
tvararu, line 48:
```suggestion
```

tvararu: This suggestion is combined with the one above in the same file

app/services/submit_application.rb
```diff
@@ -10,7 +10,12 @@ def initialize(application_form)
...
+        application_form.update!(submitted_at: Time.zone.now, edit_by: Time.zone.now)
+      else
+        application_form.update!(submitted_at: Time.zone.now)
+        application_form.update!(edit_by: edit_by_time)
```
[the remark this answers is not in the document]
davidgisbey: this actually needs to be in two separate actions. Edit by time is reliant on submitted_at being set. I guess ii can change the edit_by_time method to always use Time.zone.now instead. I'll have a look into it.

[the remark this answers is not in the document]
tvararu: Ah fair dos, in that case you can ignore but it isn't obvious why there's two `update!` calls one after each other. Someone else might want to do the same fix and trip over this.

app/services/submit_reference.rb
davidgisbey: Yea i was trying to make it clear what should be stripped out later on, but it's just made it way more confusing to parse. Thanks for all the feedback!

tvararu:
On the review app, this submitted application form has an application choice with a nil `edit_by`; I think this is just bad seed data. When I ran the migrations, it set the `edit_by: nil` on the main application form https://apply-for-te-1573-chang-l8jeb1.herokuapp.com/support/applications/2/audit

PR #2107, opened 2020-05-20
Backfill ApplicationForm#edit_by

davidgisbey:
## Context

**Needs to be merged after #2099** 

We need to backfill edit_by dates on the application forms before deploying the rest of the changes to the code out.

- for submitted applications use the first application choices edit_by to populate its edit_by column

## Guidance to review

A further PR to drop the edit_by column from the application choices column will follow.

## Link to Trello card

## Things to check


tvararu:
I think this PR can go into #2099; then that PR will not rely on it, because the code that looks this up is behind a feature flag, but they'll still go out at the same time (because that makes sense, we can't test #2099 without this PR)

davidgisbey:
Okay cool will close this one 👍

PR #2113, opened 2020-05-20
Fix bug changing the type parameter on the add new referee

stevehook:
## Context

It's possible to manipulate the URL when adding a new reference to throw a server error.

- Use the existing `Reference::RefereeTypeForm` with an extra validation to enforce inclusion rules for `ApplicationReference#reference_type` and redirect if needed rather than throwing the exception. Both for initial two references `ReferencesController` and post-submission `AdditionalReferencesController`.
- Update system specs.

## Guidance to review

- Does this cover all the possible ways to set a reference type?
- Are there any issues with reusing the `Reference::RefereeTypeForm` for validation?

## Link to Trello card

https://trello.com/c/5wJxPkcy/1561-bug-changing-the-type-parameter-on-the-add-new-referee-flow-leads-to-a-server-error

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/additional_referees_controller.rb
+++ b/app/controllers/candidate_interface/additional_referees_controller.rb
@@ -41,6 +41,12 @@ module CandidateInterface
     def new
       redirect_to_confirm_if_no_more_reference_needed
 
+      referee_type_form = Reference::RefereeTypeForm.new(referee_type: params[:type])
+      unless referee_type_form.valid?
+        track_validation_error(referee_type_form)
+        return redirect_to action: 'type'
+      end
+
       @reference = current_candidate.current_application.application_references.build(referee_type: params[:type])
       @page_title = "Details of your new #{@reference.referee_type.downcase.dasherize} referee"
     end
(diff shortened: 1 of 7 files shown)


tvararu:
What @tijmenb said, with the add that I think instead of a 404, it should be a 301 to the page that asks you to input the type 👍

stevehook:
> Cool, I think this works. I wonder though if we should approach it differently, as the type isn't really something the user inputs, so validation and errors wouldn't necessarily have to be used.
> 
> It would also be appropriate to add a constraint in the routes to make sure that `params[:type]` is one of the known types, so that any other type would return a 404.
> 
> This is the approach taken for the GCSE subjects:
> 
> https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/8391c817f28d3e0383e9d10887a6da6decffc51a/config/routes.rb#L108

Have switched over to route constraints. The two routes pass the type in different ways, the simple constraint works for the path-based param but I had to use a lambda for the one that uses query parameters. Maybe we could align these so they both use path-based params?

PR #2114, opened 2020-05-20
Refactor filter_component to make it more extensible

willmcb:
Removes some sub components that didn't provide much in the
trade off between flexibility and maintainability.

Refactors ProviderApplicationsPageState so that it takes
fewer inputs and more cleanly builds the config options
for the filter

## Context

When trying to add locations filters to the provider applications page, it became clear that the component needs refactoring so as to allow further changes to be made cleanly.

No UI changes - only backend code.

## Guidance to review

- are the tests detailed enough?

## Link to Trello card

As a precursor to implementing: https://trello.com/c/kISeiFeX/2099-build-add-training-location-attribute-to-application-view-list-filter

## Things to check


app/components/provider_interface/filter_component.rb
```diff
@@ -2,24 +2,64 @@ module ProviderInterface
...
+          if option[:checked]
+            arr << { title: option[:label], remove_link: remove_checkbox_tag_link(filter[:name], option[:value]) }
+          end
+        end
+      end
     end
 
-    def is_candidates_name_search_field?(filter_group)
-      filter_group[:heading].parameterize.eql?('candidate-s-name')
+    def filter_should_render?(filter_options:)
```
duncanjbrown, line 25: Lovely method name! But... it looks generic when it's not. If there are no options for the filter in question, should it even be added to the filters array in the first place?

duncanjbrown:
A good way to review this might be to check out the branch and see if you can understand how the filters work. The specs all pass, and the diff nets out at -395 lines, so if it's comprehensible (we hope it is, and better than what came before) then it's an improvement.

app/components/provider_interface/filter_component.rb
willmcb: Good point, but thinking about it now, this method is incorrect. It is not that it is less than 1 that we should care about, but that it is be that it is less than 2. If it is less than two it means that there is only one possible filter so it is redundant as it would include all of the applications on that page. This was the original intention for this check.

For instance if there is only 1 provider - then all applications would belong to it, there fore filtering for it would just bring back everything in any case.

Does that make sense?

duncanjbrown: sorry — I see this is for the case where there's only one option

willmcb: Have moved logic to page_state and updated specs as discussed 👍 see: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2114/commits/af1312ba73bcb5b081e07a402cc7f4bce4caff30

app/models/provider_interface/provider_applications_page_state.rb
```diff
@@ -64,45 +44,55 @@ def status_filters
...
-        heading: 'provider',
-        input_config: input_config,
+      provider_filter = {
+        type: :checkboxes,
+        heading: 'Provider',
+        name: 'provider',
+        options: provider_options,
       }
+
+      return provider_filter if provider_filter[:options].size > 1
```
duncanjbrown, line 77: It might be friendlier to return as early as possible, ie when we first do the query for providers, rather than first building the hash and then throwing it away?

willmcb: Good point there could be a few of these so would make sense to make it a bit faster.

willmcb: Have refactored so that less work is done before the check 👍 - see: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2114/commits/9b9f81c50e5f69c30f6612a9f29f095ff19a5e2d

mnacos:
Haven't had the chance to review properly yet, but `FilterComponentPreview` needs updating (see `spec/components/previews/provider_interface/filter_component_preview.rb`)

PR #2115, opened 2020-05-20
Allow ratifying provider to be null

steventux:
## Context

This migration PR https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2094/files#diff-211254c6b60fe3ebd7ebb745c0f3486fR6 introduced the `provider_relationship_permissions` table, but added not-null constraints on the `ratifying_provider_id` column.

Self-ratifying training providers won't need to populate this field so allow null here.

## Link to Trello card

https://trello.com/c/F6QCZXqx/2169-provider-to-provider-permissions-affect-what-users-can-do-in-the-system

## Things to check


duncanjbrown:
I'm still not convinced we need records for self-ratifying providers. If there is no accredited body, then there's no relationship to check? Or to put it another way, if the training provider can't do anything with this course, and there's no accredited body either, then nobody can act on this application.

steventux:
@duncanjbrown I understand what you mean and I think you have a good point, perhaps my opinionated tests for the authorising service led me here, I'll close this PR.

PR #2117, opened 2020-05-20
Don’t show "Awaiting decision" if flash is present

duncanjbrown:
## Context

Spotted in the bug bash

Hide the "Awaiting decision" banner unless the flash is empty

## Guidance to review

No test! 🤠 

## Link to Trello card

https://trello.com/c/C5XZPFvU/2134-when-i-add-a-note-to-a-new-application-there-are-2-alert-boxes

## Things to check

### DIFF
--- a/app/views/provider_interface/application_choices/_application_choice_header.html.erb
+++ b/app/views/provider_interface/application_choices/_application_choice_header.html.erb
@@ -1,4 +1,4 @@
-<% if @application_choice.status == 'awaiting_provider_decision' -%>
+<% if @application_choice.status == 'awaiting_provider_decision' && flash.empty? -%>
   <div class="app-banner">
     <div class="app-banner__message">
       <h2 class="govuk-heading-m govuk-!-margin-bottom-0">


duncanjbrown:
> Okido. I do wonder if we need a design iteration on this. I think we're using the same pattern for immediate notifications (flash), BAU call to actions (make a decision) and semi-temporary notices (the COVID banner)?

Good point — I'll mention it to @adamsilver

adamsilver:
@tijmenb @duncanjbrown 

Yep so the two types of banner are meant to look the same. Here's the differences:

1. When someone does something and needs to know whether it worked or not it should be focused automatically (like the Error Summary)

2. It should be green when successful

3. It should be red when not

4. When there's a message like COVID or ‘Respond to application’ i.e. messages generated by the system, then it should be blue and not automatically focused.

5. Because they take up the same position on screen, it's best to only have one notification shown at a time. There are various different ways to go about it. But for this particular ticket I’m pretty happy with it.

Also the guidance above is part of the proposed notification banner component that's going into the GOV.UK Design System soon.

PR #2122, opened 2020-05-21
Refactor change offer with view_components

mnacos:
## Context

We need to add another step to the 'change offer' flow (for `study_mode`). This is relevant to both the initial offer and existing offer flows.

Move multi-step interface logic to view_components.

A single parent component (`ProviderInterface::ChangeOfferComponent`) accepts a `ChangeOfferForm` and decides which sub-component to render based on the supplied `:step` and the validity of the form.

The entire flow relies on a single route with a `*step` wildcard element. This makes it easier to troubleshoot user actions in the logs. 
```
GET    /provider/applications/:application_choice_id/offer/change/*step(.:format)
```

After this PR, adding a `study_mode` step to the process will be a matter of adding a step to `ChangeOfferForm`, creating the relevant `study_mode` sub-component and modifying the logic of listing courses and locations based on `study_mode`. 

## Guidance to review

Best place to start is the 'change offer' component previews. [See review app previews](https://apply-for-te-1995-refac-01ltkk.herokuapp.com/rails/view_components/provider_interface/change_offer)

Use 'split' view when inspecting file changes.

## Link to Trello card

https://trello.com/c/0u3Z8WLc

## Things to check


app/components/provider_interface/change_offer_component.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+
+    def set_entry_from_step
+      change_offer_form.entry ||= change_offer_form.step
+    end
+
+  private
+
+    def setup_component(component)
+      if component == ProviderInterface::ChangeOffer::ConfirmComponent
```
steventux, line 29: This seems like an unnecessary conditional given the callers all originate in a case statement.

app/components/provider_interface/change_offer_component.rb
```diff
@@ -0,0 +1,57 @@
...
+          change_offer_form: change_offer_form,
+          providers: providers,
+        )
+      end
+    end
+
+    def get_component_for_step(step)
+      case step
+      when :confirm
+        setup_component ProviderInterface::ChangeOffer::ConfirmComponent
```
steventux, line 47: Given we have the appropriate constructor args `change_offer_form, providers` and `completion_url, completion_method` as instance vars in this super component, why not just initialize these step components with args here? It seems indirect to pass the responsibility of initializing to another method with another conditional.

spec/components/previews/provider_interface/change_offer_preview.rb
```diff
@@ -0,0 +1,151 @@
...
+      if new_course_option
+        provider_id = new_course_option.provider.id
+        course_id = new_course_option.course.id
+        course_option_id = new_course_option.id
+      else
+        provider_id = application_choice.offered_course.provider.id
+        course_id = application_choice.offered_course.id
+        course_option_id = application_choice.offered_option.id
+      end
+      ProviderInterface::ChangeOfferForm.new \
```
steventux, line 121: Stylewise shouldn't we just use parenthesis here instead of a backslash? Applies to other uses of backslash for multiline constructor or method calls.

app/models/provider_interface/pick_response_form.rb
```diff
@@ -8,6 +8,25 @@ class PickResponseForm
     attr_accessor :decision
     validates :decision, inclusion: { in: VALID_DECISIONS }
 
+    def redirect_attrs
+      attrs = {}
+
+      if decision_is_change?
+        attrs[:controller] = 'offer_changes'
+        attrs[:action] = 'edit_offer'
+        attrs[:step] = case decision
```
steventux, line 17: Case statement could be simplified to
```ruby
attrs[:step] = decision.delete_prefix('edit')
```

spec/components/previews/provider_interface/change_offer_preview.rb
```diff
@@ -0,0 +1,151 @@
+module ProviderInterface
+  class ChangeOfferPreview < ViewComponent::Preview
+    def existing_offer_start_at_1_provider(provider_interface_change_offer_form: nil)
```
steventux, line 3: The first 3 methods here repeat the same code varying only on the `initial_step` method argument. That suggests something common could be called.

spec/components/previews/provider_interface/change_offer_preview.rb
```diff
@@ -0,0 +1,151 @@
...
+    def existing_offer_start_at_confirm_change(provider_interface_change_offer_form: nil)
+      initial_step :confirm
+      set_application_choice application_choice_with_offer
+
+      form = submitted_form(provider_interface_change_offer_form)
+      form ||= new_form(new_course_option: pick_different_option)
+      render_with form
+    end
+
+    def new_offer_start_at_1_provider(provider_interface_change_offer_form: nil)
```
steventux, line 36: Similar repetition here suggests these 3 methods could be simplified and DRY'd up.

app/components/provider_interface/status_box_components/offer_component.rb
```diff
@@ -54,11 +54,11 @@ def change_path(target)
 
         case target
         when :provider
-          paths.provider_interface_application_choice_change_offer_edit_provider_path(application_choice.id, entry: 'provider') if show_provider_link?
+          paths.provider_interface_application_choice_edit_offer_path(application_choice.id, step: 'provider') if show_provider_link?
```
steventux, line 57: There's no need for a case statement here, you can simplify as:

```ruby
paths.provider_interface_application_choice_edit_offer_path(application_choice.id, step: target.to_s)
```

You could check that `target` is valid beforehand if need be.

mnacos: Normally I would agree but there is some logic in `show_provider_link?`, `show_course_link?` and `show_course_option_link?` that needs to be checked according to `target`. This is probably the main reason for the current construct.

spec/components/previews/provider_interface/change_offer_preview.rb
mnacos: Most code in this file has been added to provide a scaffold so that clicking through the various sub-components works (more or less). However, this is not a controller, so I cannot rely on any `:before_action` magic or equivalent. This means I would need to extract the common code into a method and call this from each scenario, maybe saving 1-2 lines in each case but sacrificing readability.

The reason we don't have only two cases (existing offer, new offer) is that subcomponent titles and change links at the end of the process are conditional to `entry`, therefore starting the flow at different points is significant (mostly for the existing offer flow).

app/components/provider_interface/status_box_components/offer_component.rb
steventux: Ah yes didn't spot that link check. 
You could use the target to work out the collection needed to calculate whether the link was displayed. Then you'd just have a method called `show_link?(target)`. This would lead to less less `show_*_link?` methods and you wouldn't need this case statement:

```ruby
def show_link?(target)
  collection = case target
  when :course then available_course
  when :course_options then available_course_options
  else available_providers
  end
  collection.count > 1 if collection
end
```

app/components/provider_interface/change_offer_component.rb
mnacos: Yep, I agree

app/components/provider_interface/status_box_components/offer_component.rb
mnacos: I like

spec/components/previews/provider_interface/change_offer_preview.rb
steventux: For me it's not the number of lines of code we'd save so much as repeating the exact same lines several times (with the exception of the value passed to `initial_step`), maintaining methods like this will become a headache. 
I'm perhaps not fully understanding how these methods are called as I can't seem to find where they are called.

mnacos: These methods are only called in the context of component previews, see https://apply-for-te-1995-refac-01ltkk.herokuapp.com/rails/view_components/provider_interface/change_offer 

The previews themselves are used for easy visual inspection and testing components in isolation. They are similar to tests, in that they also document things.

The fact these methods are only ever called for as part of development and QA is why I think some repetition may be ok here.

app/components/provider_interface/change_offer/change_provider_component.rb
```diff
@@ -0,0 +1,36 @@
...
+  module ChangeOffer
+    class ChangeProviderComponent < ViewComponent::Base
+      include ViewHelper
+
+      attr_reader :change_offer_form, :application_choice, :providers
+
+      def initialize(change_offer_form:, providers:)
+        @change_offer_form = change_offer_form
+        @application_choice = change_offer_form.application_choice
+        @providers = providers.respond_to?(:order) ? providers.order(:name) : providers
```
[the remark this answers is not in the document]
mnacos: I've made ordering of radio buttons a component responsibility (see e.g. courses and locations), so ordering the providers should happen here, too. Contrary to courses and locations, which are queried directly from the components, `providers` is passed as an argument, so I guess I'm being a bit defensive/paranoid.

app/components/provider_interface/change_offer_component.rb
```diff
@@ -0,0 +1,49 @@
+module ProviderInterface
+  class ChangeOfferComponent < ViewComponent::Base
+    include ViewHelper
+
+    attr_reader :change_offer_form, :application_choice, :providers, :completion_url, :completion_method
+
+    def initialize(change_offer_form:, providers:, completion_url:, completion_method:)
```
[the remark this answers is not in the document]
mnacos: Yes, it's only relevant when completing the flow in view_components. The original intention was to override `completion_url` and `completion_method` according to the context and be able to support existing offers, new offers and previews with minimal changes. Too bad the new offer flow skips the confirmation component!

mnacos:
> This is definitely an improvement! Minor comments/questions inline.
> 
> I wonder if the relationship between `ChangeOfferForm` and `ChangeOfferComponent` has become a bit too cosy — there's some slightly funky stuff going on where the components are validating and mutating the form. I'm not sure it's worth refactoring at this stage, though.

In order to support this multi-step process and remove as much logic from the controller as possible, I've gone for a new pattern, one which increases the coupling between component and form. The component is useless without the form, but the form contains all validation logic and can be used or tested individually. The component mutates the form during the multi-step process but the form doesn't save or produce side-effects.

app/components/provider_interface/change_offer/change_provider_component.rb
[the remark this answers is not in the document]
duncanjbrown: If we're worried about this, then we might as well use `sort_by`, I think. That'll work for AR relations (by realising them) and arrays equally well. But I'm not convinced we do need to worry about it.

app/components/provider_interface/change_offer_component.rb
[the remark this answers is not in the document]
duncanjbrown: It feels quite a lot to add just for that. How would you feel about cutting it?

app/components/provider_interface/change_offer/change_provider_component.rb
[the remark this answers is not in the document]
mnacos: Ok, will remove the check and trust we can call `.order`

app/components/provider_interface/change_offer_component.rb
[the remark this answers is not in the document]
mnacos: I've removed `completion_method`.

PR #2125, opened 2020-05-21
Completed sections - Fix error links

davidgisbey:
## Context

Currently, when you go on the submit show page and try and submit an incomplete application you receive errors asking you to complete the section.

When you click on the link it should direct you to that section of the page so you can click on complete the section. 

This is not working

- If the feature flag is active and the relevant component is called from the submit_show page. Use the application_form_presenter complete logic for the link.
- If the component is called in any other context. Use the previous section completed validations. e.g. for contact details check the number and address are valid/completed.

## Guidance to review

- Create a new application, try and submit it with the flag on and off. Do the links work?

## Link to Trello card

## Things to check

### DIFF
--- a/app/components/candidate_interface/contact_details_review_component.rb
+++ b/app/components/candidate_interface/contact_details_review_component.rb
@@ -2,14 +2,14 @@ module CandidateInterface
   class ContactDetailsReviewComponent < ViewComponent::Base
     validates :application_form, presence: true
 
-    def initialize(application_form:, editable: true, missing_error: false, submit_show: false)
+    def initialize(application_form:, editable: true, missing_error: false, submit_application_attempt: false)
       @application_form = application_form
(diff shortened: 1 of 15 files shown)


app/components/candidate_interface/contact_details_review_component.rb
```diff
@@ -2,21 +2,26 @@ module CandidateInterface
   class ContactDetailsReviewComponent < ViewComponent::Base
     validates :application_form, presence: true
 
-    def initialize(application_form:, editable: true, missing_error: false)
+    def initialize(application_form:, editable: true, missing_error: false, submit_application_attempt: false)
```
stevehook, line 5: Would `submitting_application` be a better name for the new option? I'm not sure.

davidgisbey: yea i actually prefer submitting_application 

Thanks for the suggestion.

PR #2129, opened 2020-05-22
Ensure completed section booleans are updated with the flag off

davidgisbey:
## Context

When the `mark_section_as_complete` feature flag is off, the section completed booleans should still be updated.

- Change the `completed_section_component` to always update the section to true when a user clicks continue without being presented with a checkbox.

## Guidance to review

None really 

## Link to Trello card

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/contact_details/address_controller.rb
+++ b/app/controllers/candidate_interface/contact_details/address_controller.rb
@@ -1,8 +1,5 @@
 module CandidateInterface
   class ContactDetails::AddressController < ContactDetails::BaseController
-    before_action :redirect_to_dashboard_if_submitted
-    after_action :complete_section, only: %i[update]
-
     def edit
       @contact_details_form = ContactDetailsForm.build_from_application(
         current_application,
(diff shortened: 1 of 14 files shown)


tvararu:
There is a way to soft break this in the review app with the feature flag off:

- Complete the contact details section
- Click on Continue
- Note that `contact_details_completed: true`
- Go to contact details section
- Change something
- Note that `contact_details_completed: false`
- Click "Back to application" instead of Continue
- Note still that `contact_details_completed: false`

If the feature flag then is turned on, the user will encounter the error that they need to mark their contact details as completed to submit.

Is there an easy way to address this? Does it involve adding something to all the controllers?..

app/controllers/candidate_interface/contact_details/address_controller.rb
```diff
@@ -1,6 +1,7 @@
 module CandidateInterface
-  class ContactDetails::AddressController < CandidateInterfaceController
+  class ContactDetails::AddressController < ContactDetails::BaseController
     before_action :redirect_to_dashboard_if_submitted
+    after_action :complete_section, only: %i[update]
```
tvararu, line 4:
```suggestion
```

Do we need these if they're being done in `ContactDetails::BaseController`?

PR #2130, opened 2020-05-22
Iterate support view on validation errors for analysis

tijmenb:
## Context

When a user fills in a form we often perform some validations, like checking that a field is filled in or doesn’t have too many words. We’ve been tracking validation errors for about 3 weeks now. This PR updates the UI that we use for analysis.

This iterates the validation error pages to allow analysis of the errors.

Add list of individual errors so you can see which happened most often:

image

Allow search for errors by a specific user, a form, an attribute:

image

Allow deeplinking to an individual error (handy for illustrating individual examples in the analysis):

image

As seen above, it updates the layout to make it clearer what happened. This mostly comes from @paulrobertlloyd.

## Guidance to review

Could you analyse the errors using this?

## Link to Trello card

https://trello.com/c/4bMEwKdf/1587-analysis-of-validation-errors

## Things to check


paulrobertlloyd:
Where do you search from, or is it just by using URL params?

tijmenb:
@paulrobertlloyd yeah, you can click all the things you "search" for, which puts it in the URL. Perhaps "search" isn't the right word here.

tijmenb:
Merging so it can go out and I can try it with prod data soon. We'll iterate further!

PR #2134, opened 2020-05-22
Send application to provider immediately when applying again

tvararu:
## Context

We need to send applications off to providers as soon as they are submitted when users apply again.

Add an option to the SubmitApplicationChoice service that allows us to immediately send off an application choice.

## Guidance to review

Does this make sense? I will reuse this in another PR about allowing support users to add applications to forms that are in the "awaiting provider decision" state.

## Link to Trello card

https://trello.com/c/azfEJqBy/1593-dev-send-apply-again-application-straight-away-mvp

## Things to check

### DIFF
--- a/app/services/submit_application.rb
+++ b/app/services/submit_application.rb
@@ -18,7 +18,12 @@ class SubmitApplication
       submit_application
     end
 
-    CandidateMailer.application_submitted(application_form).deliver_later
+    if send_to_provider_immediately?
+      CandidateMailer.application_sent_to_provider(@application_form).deliver_later
+    else
+      CandidateMailer.application_submitted(application_form).deliver_later
+    end
+
     send_reference_request_email_to_referees(application_form)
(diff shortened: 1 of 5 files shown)


app/services/submit_application.rb
```diff
@@ -18,7 +18,12 @@ def call
       submit_application
     end
 
-    CandidateMailer.application_submitted(application_form).deliver_later
+    if send_to_provider_immediately?
+      CandidateMailer.application_sent_to_provider(@application_form).deliver_later
+    else
+      CandidateMailer.application_submitted(application_form).deliver_later
+    end
```
stevehook, line 25: I guess these could cause a kind of race condition if the mailer depends on something happening in the current transaction. But that's not a new problem. I think I'd be in favour of adding a `deliver_later(wait: 5.seconds)` or something if this is an issue.

tvararu: Hmm, not aware of that being an issue currently. Fine to merge without?

PR #2135, opened 2020-05-22
Allow support users to add courses when awaiting_provider_decision

tvararu:
## Context

Merge https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2134 first.

We have to add a course to a candidate's application because the provider is not replying. We have a form to do this in support, but it only works if the choices are in the `awaiting_references`.

Make the support form work when the application is in the `awaiting_provider_decision` state.

## Guidance to review

Does this make sense?

## Link to Trello card

https://trello.com/c/fiJcy82l/1586-manually-add-course-to-candidates-application

## Things to check

### DIFF
--- a/app/models/support_interface/add_course_to_application_form.rb
+++ b/app/models/support_interface/add_course_to_application_form.rb
@@ -49,7 +49,7 @@ module SupportInterface
     def application_form_in_correct_state
       return if awaiting_references? || awaiting_provider_decision?
 
-      errors[:base] << 'You can only add a course to an application that has at least 1 other application choice in the "awaiting references" state.'
+      errors[:base] << 'You can only add a course to an application that has at least 1 other application choice in the "awaiting references" or "awaiting_provider_decision" state.'
     end
   end
 end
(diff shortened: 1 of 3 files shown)


app/models/support_interface/add_course_to_application_form.rb
```diff
@@ -32,11 +35,19 @@ def course_option_exists
...
     # For simplicity we only support adding applications to forms that are in
-    # "Awaiting references" state, because we need to add the choices in the
-    # same state as other applications.
+    # "Awaiting references" or "Awaiting provider decision" state, because we
+    # need to add the choices in the same state as other applications.
     def application_form_in_correct_state
-      return if application_form.application_choices.any?(&:awaiting_references?)
+      return if awaiting_references? || awaiting_provider_decision?
 
       errors[:base] << 'You can only add a course to an application that has at least 1 other application choice in the "awaiting references" state.'
```
stevehook, line 52:
```suggestion
      errors[:base] << 'You can only add a course to an application that has at least 1 other application choice in the "awaiting references" or "awaiting_provider_decision" state.'
```

PR #2136, opened 2020-05-22
Fix some mailer previews

tvararu:
## Context

Mailer previews are broken due to a variety of reasons.

Un-break them and add tests so hopefully they stay un-broken.

- Stub the `.update!` method on the `build_stubbed` candidate that is passed into the emails that create magic links; prevents errors which warn that "you can't call ActiveRecord methods on a build_stubbed object"
- Reuse the same `site` so that we don't hit the `.uniq` limit for the codes and school addresses which is a bit of a rabbit hole
- Add request specs to test if mailers return 200; I set the tests that are currently failing to be `pending`

The same tweaks can be used to fix almost all the other emails, but I will do that in a follow-up PR.

## Guidance to review

Commit by commit if you like.

I made some small tweaks to the preview page so it looks prettier too.

### Before

Screenshot 2020-05-21 at 14 58 03

### After

Screenshot 2020-05-21 at 14 58 10

### Tests

Screenshot 2020-05-27 at 13 56 11

## Link to Trello card

https://trello.com/c/Rulh9wiA/1582-fix-broken-mailer-previews

## Things to check

### DIFF
--- a/app/views/support_interface/mailer_previews/index.html.erb
+++ b/app/views/support_interface/mailer_previews/index.html.erb
@@ -1,8 +1,8 @@
 <% @previews.each do |preview| %>
-<h3><%= preview.preview_name.titleize %></h3>
-<ul>
+<h3 class="govuk-heading-m"><%= preview.preview_name.titleize %></h3>
+<ul class="govuk-list govuk-list--bullet">
 <% preview.emails.each do |email| %>
-<li><%= link_to email, url_for(controller: 'rails/mailers', action: 'preview', path: "#{preview.preview_name}/#{email}") %></li>
+<li><%= govuk_link_to email, url_for(controller: 'rails/mailers', action: 'preview', path: "#{preview.preview_name}/#{email}") %></li>
 <% end %>
 </ul>
 <% end %>
--- a/spec/mailers/previews/candidate_mailer_preview.rb
+++ b/spec/mailers/previews/candidate_mailer_preview.rb
@@ -2,6 +2,7 @@ class CandidateMailerPreview < ActionMailer::Preview
   def application_submitted
     application_form = FactoryBot.build_stubbed(
       :completed_application_form,
+      candidate: candidate,
       support_reference: 'ABC-DEF',
     )
 
@@ -11,7 +12,8 @@ class CandidateMailerPreview < ActionMailer::Preview
   def application_sent_to_provider
     application_form = FactoryBot.build_stubbed(
       :completed_application_form,
-      application_choices: [FactoryBot.build_stubbed(:application_choice, :awaiting_provider_decision)],
+      candidate: candidate,
+      application_choices: [FactoryBot.build_stubbed(:application_choice, :awaiting_provider_decision, course_option: course_option)],
     )
 
     CandidateMailer.application_sent_to_provider(application_form)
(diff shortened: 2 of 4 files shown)


spec/requests/support_interface/mailer_previews_spec.rb
```diff
@@ -0,0 +1,216 @@
+require 'rails_helper'
+
+RSpec.describe 'Support interface - Mailer previews', type: :request do
```
tijmenb, line 3: Can we loop through the mailers for these tests, so they automatically are run when added?

tvararu: Hmm, yes, however in that case you'd only be able to `xit` an entire mailer instead of an individual broken preview. I think that's preferable once they're all green, so will update to that.

tvararu: @tijmenb see updated code. I've used `.emails.each` but also had to define a list of `broken_previews` which are currently failing. These are marked as pending. The idea is to fix in follow-up PRs.

tvararu:
Build is currently failing because Azure DevOps fails when we have warnings in the output.

app/views/support_interface/mailer_previews/index.html.erb
```diff
@@ -1,8 +1,8 @@
 <% @previews.each do |preview| %>
-<h3><%= preview.preview_name.titleize %></h3>
-<ul>
+<h3 class="govuk-heading-m"><%= preview.preview_name.titleize %></h3>
+<ul class="govuk-list govuk-list--bullet">
 <% preview.emails.each do |email| %>
-<li><%= link_to email, url_for(controller: 'rails/mailers', action: 'preview', path: "#{preview.preview_name}/#{email}") %></li>
+<li><%= govuk_link_to email, url_for(controller: 'rails/mailers', action: 'preview', path: "#{preview.preview_name}/#{email}") %></li>
```
willmcb, line 5: Would be nice to apply these styles to the component previews as well - obvs no need to be part of this PR, but would be cool to have consistency of styles

spec/mailers/previews/candidate_mailer_preview.rb
```diff
@@ -266,6 +258,18 @@ def conditions_not_met
 
 private
 
+  def candidate
+    candidate = FactoryBot.build_stubbed(:candidate)
+
+    # This is not great. It's necessary because some of our mail templates
+    # generates and send a new magic link token to candidates.
+    def candidate.update!(*)
```
willmcb, line 266: How does this work? Are you monkey patching the instance of candidate to override `.update()`? I assumed thats what this is doing and I understand 'why' but have never seen this syntax before so I'm not sure how 🤔?

app/views/support_interface/mailer_previews/index.html.erb
tvararu: I agree, can be done separately!

spec/mailers/previews/candidate_mailer_preview.rb
tvararu: Yup, but it's not a real candidate object, it's a `build_stubbed`. `update!` is already monkey patched to something that raises. For the purpose of the mailer previews, I want it to carry on. 🤔

PR #2142, opened 2020-05-27
Refactor course choices controller

malcolmbaig:
# Context

The controller has grown over time as we've added more features and complexity, and is now long and hard to navigate.

- Extract all actions in `CourseChoicesController` to more specific controllers. Those that relate to the process of choosing a course live under a new `CourseChoices` namespace. 
- The `pick_site_for_course` and `pick_new_site_for_course` methods are used in multiple places. Extract them into service objects called `PickCourseOption` and `PickReplacementCourseOption`, respectively.

## Guidance to review

- Per-commit review highly recommended.
- Is it clearer? Does the new structure make sense? Is any of the naming confusing?
- There's still plenty of scope to refactor some of the longer actions, eg - `pick_course`, but limiting to these changes for now to discuss the approach and keep the PR to a manageable size.

## Link to Trello card

https://trello.com/c/5rSJREyS

## Things to check


app/controllers/candidate_interface/course_choices/add_another_course_controller.rb
```diff
@@ -0,0 +1,28 @@
+module CandidateInterface
+  module CourseChoices
+    class AddAnotherCourseController < BaseController
+      def ask
```
stevehook, line 4: It looks like this `ask`/`decide` pattern is common to a few of the new simplified controllers. Do you think a naming convention like you've introduced here could be applied to some of the others?

At first I wondered whether standard 'RESTful' action names like `index` and `create` might make sense now that each controller is concerned with just one resource/thing.

app/controllers/candidate_interface/application_choices_controller.rb
```diff
@@ -0,0 +1,65 @@
+module CandidateInterface
+  class ApplicationChoicesController < CandidateInterfaceController
+    before_action :redirect_to_dashboard_if_submitted
+
+    def index
+      @course_choices = current_candidate.current_application.application_choices
```
stevehook, line 6: We use `course_choice` or similar in a few places to refer to an application choice. Would it be less confusing to just call the `application_choice` or `application_choices` etc.?

app/services/candidate_interface/pick_course_option.rb
```diff
@@ -0,0 +1,45 @@
+module CandidateInterface
+  class PickCourseOption
```
stevehook, line 2: This is great. One advantage is that we can now add more isolated specs for the service business logic. I think we are covered by the system specs for now, service specs can be added when we need them (when we want to change the logic etc.)

app/controllers/candidate_interface/course_choices/course_selection_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+          )
+        else
+          @pick_course = PickCourseForm.new(
+            provider_id: params.fetch(:provider_id),
+            application_form: current_application,
+          )
+        end
+      end
+
+      def pick_course
```
tvararu, line 22: This method still looks quite long, but it's now in a much shorter file. This is a lot better and likely enough for now.

tijmenb:
@tvararu I'd say that if we don't need a spec for the service, then we don't need to add it. Perhaps we sometimes over test some bits because we feel we need to add to the existing unit tests?

app/controllers/candidate_interface/course_choices/add_another_course_controller.rb
malcolmbaig: > Do you think a naming convention like you've introduced here could be applied to some of the others?

Not sure! I think it makes sense for those screens where the candidate makes a binary choice about where they go next. Not sure how well it applies to the others in this module

>At first I wondered whether standard 'RESTful' action names like index and create might make sense now that each controller is concerned with just one resource/thing.

Was thinking the same. I'll add a commit that covers this in my followup PR so that we can see what it looks like.

PR #2144, opened 2020-05-27
Refactor and simplify conditionals for "Apply Again"

tijmenb:
## Context

We've introduced the feature to allow candidates to apply again recently. This means we've introduced a bunch of conditionals in the code where we check whether or not the user is in the "apply 2" phase. This could make it harder to understand the codebase. 

It's also not always clear _why_ a certain thing is different in the different phases. In https://ukgovernmentdfe.slack.com/archives/CP18YJXPY/p1588250617304800 we decided to decouple the act of "applying again" from the 1-course choice restriction, but we haven't implemented that. 

- Replace `apply_again?` by different conditionals for `candidate_has_previously_applied?` and `candidate_can_choose_single_course?`. This will make it possible to allow candidates who have applied in a previous cycle to apply to 3 courses, by changing the `phase` for these applications.
- Use descriptive conditionals that explain what we're after, like `candidate_has_previously_applied?` or `can_edit_after_submission?`.
- Remove conditionals if they're not necessary
- Fixup some other duplicated code like the check how many choices the candidate can add

## Guidance to review

Review this *per commit* - it wouldn't make sense otherwise.

## Link to Trello card

https://trello.com/c/6aEZkNZi

## Things to check

### DIFF
--- a/app/services/submit_application.rb
+++ b/app/services/submit_application.rb
@@ -65,7 +65,7 @@ private
   end
 
   def send_to_provider_immediately?
-    application_form.apply_again? && enough_references_have_been_provided?
+    application_form.candidate_has_previously_applied? && enough_references_have_been_provided?
   end
 
   def enough_references_have_been_provided?
--- a/app/services/submit_application.rb
+++ b/app/services/submit_application.rb
@@ -65,7 +65,7 @@ private
   end
 
   def send_to_provider_immediately?
-    application_form.candidate_has_previously_applied? && enough_references_have_been_provided?
+    !application_form.can_edit_after_submission? && enough_references_have_been_provided?
   end
 
   def enough_references_have_been_provided?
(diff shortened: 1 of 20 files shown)


app/services/submit_application.rb
```diff
@@ -10,11 +10,7 @@ def initialize(application_form)
 
   def call
     ActiveRecord::Base.transaction do
-      if application_form.apply_again? && enough_references_have_been_provided?
-        application_form.update!(submitted_at: Time.zone.now, edit_by: Time.zone.now)
-      else
-        application_form.update!(submitted_at: Time.zone.now, edit_by: edit_by_time)
-      end
+      application_form.update!(submitted_at: Time.zone.now, edit_by: edit_by_time)
```
stevehook, line 13: My only concern here is that `edit_by` gets out of sync with the 'state' of the application. Like you say it doesn't fail any specs but could we make assumptions in future based on `edit_by` that are broken for apply again applications?

tijmenb: Yeah, I'll look at this separately when I'll look at https://ukgovernmentdfe.slack.com/archives/CN1MCQCHZ/p1590568483055500. It might have gthe same solution.

PR #2147, opened 2020-05-27
New header style for application

willmcb:
## Context

Changes the header design to match that of the prototype.

UPDATE: following speaking to @adamsilver new header is now only used in the ProviderInterface.

**Before**
Screenshot 2020-05-27 at 11 39 45

**After**
Screenshot 2020-05-27 at 16 44 48

## Guidance to review

- is it too reliant in MoJ imports?
- should the line under the heading match the colour of the environment tag?
- does it look okay throughout the app?
- does it collapse as you'd expect on smaller screens (I have removed the 'menu' button inline with the behaviour of the prototype.
- does this look okay if someone is signed out:
Screenshot 2020-05-27 at 16 44 14

## Link to Trello card

https://trello.com/c/st1Mq1h6/2005-build-implement-the-new-header-for-manage-applications

## Things to check


tijmenb:
> should the line under the heading match the colour of the environment tag?

Yes, definitely!

app/frontend/styles/_header.scss
```diff
@@ -0,0 +1,170 @@
+/* ==========================================================================
```
tijmenb, line 1: We already include the package in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/package.json#L5. Do we need to import this specifically?

willmcb: Good point - will try to load it from source. In other places they are copied over, but I think it makes sense to load them from node_modules.

willmcb: Done.

adamsilver:
> does it collapse as you'd expect on smaller screens (I have removed the 'menu' button inline with the behaviour of the prototype.

Yep

> does this look okay if someone is signed out

Yep

willmcb:
I've learnt from @adamsilver that we don't need to rely on the MoJ header styles and can instead configure the govuk styles to achieve the design we want. As this PR is becoming slightly convoluted I am going to close it and push up a separate cleaner PR.

PR #2153, opened 2020-05-27
Use DFE-Digital's fork of audited

duncanjbrown:
For various reasons that I don't understand fully, applying some Rails best practices(TM) to Audited's autoloading code makes mailer previews work with `mail-notify` again. I raised a PR on audited but they haven't looked at it in 2 weeks, so here's a fork with the fix in the meantime.


duncanjbrown:
Going to give this one last test locally before hitting merge

PR #2154, opened 2020-05-27
Handle part-time/full-time courses filling up

stevehook:
## Context

When a course fills up and has no further vacancies at the point a candidate submits an application for it we need to flag this on the review page and catch it if the candidate attempts to submit anyway. We need to distinguish between:

1.  Course completely full (at all locations)
2. Course full at chosen location but with vacancies elsewhere
3. Course full at chosen location for the chosen study mode but with vacancies for a different study mode.

- [x] New system specs to cover cases 2 and 3 above (1 is covered elsewhere).
- [x] Extend validation logic in `ApplicationChoicePresenter` to cover case 3 above.
- [x] Some additional specs for `ApplicationChoice` to cover some of the helper methods.

## Guidance to review

- I've added new system specs rather than extend `candidate_reviewing_application_with_unavailable_course_options_spec.rb` because there are too many permutations for one application. Not sure this is the right thing to do.
- The methods on `ApplicationChoice#chosen_site_full?` and `#chosen_study_mode_full?` really mean 'chosen site full but other locations are available' and 'chosen study mode full at given location but other study modes available there'. Maybe the naming is wrong here. Can't think of a way to put it more precisely in a concise way.
- For case 3 the potential actions are the same as case 2 for now. There is no 'change study mode' only action that I know of. Do we need to think about implementing this?

## Link to Trello card

https://trello.com/c/r7oohSNF/1344-dev-handle-location-pt-ft-courses-filling

## Things to check


app/components/candidate_interface/course_choices_review_component.html.erb
```diff
@@ -45,6 +45,22 @@
           to see if the course will re-open or discuss alternatives
         </li>
       </ul>
+    <% elsif course_choice.chosen_study_mode_full? %>
```
mnacos, line 48: How about:
```suggestion
    <% elsif course_choice.course_option_full? && course_choice.other_study_modes_available? %>
```
with `other_study_modes_available?` focusing on the the second query only.

Maybe something similar for sites, too?

app/models/application_choice.rb
```diff
@@ -69,20 +69,45 @@ def course_full_error
     I18n.t('errors.application_choices.course_full', descriptor: course.provider_and_name_code)
   end
 
-  def chosen_site_full?
+  def course_option_full?
```
malcolmbaig, line 72: Thinking about the API for application_choices, and wondering if there's potentially a way to simplify.

We have two clients for this collection of availability checking code - the `ApplicationFormPresenter` and the `CourseChoicesReviewComponent`. In both cases, they need to be aware of an implicit order in which to query these methods, based on increasing specificity of what's being checked.

What if we had:

```
def site_full?
  course.course_options.where(site: course_option.site).vacancies.blank?
end
```

```
def study_mode_full?
  course_option.no_vacancies?
end
```
We then evaluate them in that order. So if we reach `study_mode_full?`, then we know that there are vacancies at the same site in the other study mode. The specs for these methods will probably be simpler/easier to follow as well. 

The fact that there's a hierarchy to these method calls at all means we should consider extracting an object out that deals exclusively with the availability errors on a given application choice, but that's something for another time.

app/components/candidate_interface/course_choices_review_component.html.erb
```diff
@@ -45,6 +45,22 @@
...
+      <p class='app-review-warning__primary-message'><%= course_choice.chosen_study_mode_full_error %>.</p>
```
malcolmbaig, line 49: This might be a good opportunity to rename `course_choices` -> `application_choices`, likely we've started doing elsewhere.

PR #2155, opened 2020-05-27
Allow providers to change study_mode when modifying offers

mnacos:
## Context

As a provider user, when making an initial offer, I want to be able to change the offer `study_mode` to full time or part time, subject to the available course options.

As a provider user, when changing an existing offer, I want to be able to change the offer `study_mode` to full time or part time, subject to the available course options.

This PR adds `study_mode` as a step within `ChangeOfferForm` and related validations (study_mode must not be blank, be either 'full_time' or 'part_time', the course must support the selected `study_mode` and options of that `study_mode` must be available).

It also introduces the relevant `ChangeStudyModeComponent` to the family of change offer components and updates the controllers to pass `study_mode` information to the form supplied to `ChangeOfferComponent`.

## Guidance to review

When using the review app, interact with Coventry University related applications, as their courses support both full time and part time options. (e.g. https://apply-for-te-1995-chang-y7o4yn.herokuapp.com/provider/applications/24) Gorse SCITT courses tend to be full time only.

Here are some screenshots. Display is often conditional, see notes.

### ### Responding to an application:
image
Note: The option to make an offer but change the `study_mode` is new. This option will appear only if options of the other type are available for the chosen course.

---------------------------------------

### ### Confirming a new offer:
image
Note: Provider name and code, study_mode information are new additions.

---------------------------------------

### ### Looking at an existing offer:
image
Note: 'Change' link for study_mode is new.

---------------------------------------

### ### Changing the study_mode:
image
Note: This is a brand new step. This step is skipped if no other study modes are available for the chosen course.

---------------------------------------

### ### Confirming changes to an existing offer:
image
Note: Change links at the bottom are conditional to whether other study modes are available, as well as to `entry`-related concerns (as before).

---------------------------------------

### Other considerations

- The logic that determines if other study modes are available for a course does not take into account `recruitment_cycle_year`.
- The form will check whether a course is 'open on Apply' but not the `vacancies` status of the course.

## Link to Trello card

https://trello.com/c/0u3Z8WLc

## Things to check

### DIFF
--- a/app/components/provider_interface/change_offer/change_course_component.rb
+++ b/app/components/provider_interface/change_offer/change_course_component.rb
@@ -18,14 +18,8 @@ module ProviderInterface
         Course.where(
           open_on_apply: true,
           provider_id: change_offer_form.provider_id,
-          study_mode: relevant_study_modes,
           recruitment_cycle_year: recruitment_cycle_year,
         ).order(:name)
-      end
-
-      def relevant_study_modes
-        current = change_offer_form.study_mode || application_choice.offered_option.study_mode
-        [current, 'full_time_or_part_time']
       end
 
       def recruitment_cycle_year
--- a/app/components/provider_interface/change_offer/change_course_component.rb
+++ b/app/components/provider_interface/change_offer/change_course_component.rb
@@ -18,13 +18,18 @@ module ProviderInterface
         Course.where(
           open_on_apply: true,
           provider_id: change_offer_form.provider_id,
-          study_mode: study_mode_for_courses,
+          study_mode: relevant_study_modes,
+          recruitment_cycle_year: recruitment_cycle_year,
         ).order(:name)
       end
 
-      def study_mode_for_courses
-        current_study_mode = application_choice.offered_option.study_mode
-        [current_study_mode, :full_time_or_part_time]
+      def relevant_study_modes
+        current = change_offer_form.study_mode || application_choice.offered_option.study_mode
+        [current, 'full_time_or_part_time']
+      end
+
+      def recruitment_cycle_year
+        application_choice.offered_course.recruitment_cycle_year # same year as application
       end
 
       def page_title
--- a/app/components/provider_interface/change_offer_component.rb
+++ b/app/components/provider_interface/change_offer_component.rb
@@ -29,6 +29,10 @@ module ProviderInterface
         ProviderInterface::ChangeOffer::ChangeLocationComponent.new(
           change_offer_form: change_offer_form,
         )
+      when :study_mode
+        ProviderInterface::ChangeOffer::ChangeStudyModeComponent.new(
+          change_offer_form: change_offer_form,
+        )
       when :course
         ProviderInterface::ChangeOffer::ChangeCourseComponent.new(
           change_offer_form: change_offer_form,
--- a/app/components/provider_interface/change_offer_component.rb
+++ b/app/components/provider_interface/change_offer_component.rb
@@ -30,9 +30,18 @@ module ProviderInterface
           change_offer_form: change_offer_form,
         )
       when :study_mode
-        ProviderInterface::ChangeOffer::ChangeStudyModeComponent.new(
-          change_offer_form: change_offer_form,
-        )
+        study_modes = Course.find(change_offer_form.course_id).study_modes_present
+        if study_modes.count > 1
+          ProviderInterface::ChangeOffer::ChangeStudyModeComponent.new(
+            change_offer_form: change_offer_form,
+          )
+        else # skip to location
+          change_offer_form.study_mode = study_modes.first
+          change_offer_form.step = :course_option
+          ProviderInterface::ChangeOffer::ChangeLocationComponent.new(
+            change_offer_form: change_offer_form,
+          )
+        end
       when :course
         ProviderInterface::ChangeOffer::ChangeCourseComponent.new(
           change_offer_form: change_offer_form,
--- a/app/components/provider_interface/change_offer_component.rb
+++ b/app/components/provider_interface/change_offer_component.rb
@@ -30,17 +30,15 @@ module ProviderInterface
           change_offer_form: change_offer_form,
         )
       when :study_mode
-        study_modes = Course.find(change_offer_form.course_id).study_modes_present
+        study_modes = Course.find(change_offer_form.course_id).available_study_modes_from_options
         if study_modes.count > 1
           ProviderInterface::ChangeOffer::ChangeStudyModeComponent.new(
             change_offer_form: change_offer_form,
           )
-        else # skip to location
+        else # skip to next step
           change_offer_form.study_mode = study_modes.first
-          change_offer_form.step = :course_option
-          ProviderInterface::ChangeOffer::ChangeLocationComponent.new(
-            change_offer_form: change_offer_form,
-          )
+          change_offer_form.step = change_offer_form.next_step
+          sub_component
         end
       when :course
         ProviderInterface::ChangeOffer::ChangeCourseComponent.new(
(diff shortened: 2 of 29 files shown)


duncanjbrown:
Hello! Just beginning to review this. I noticed one UI quirk, which is that if I select "change to part time" or click the "change" link alongside study mode, I'm re-offered the "Location" screen afterwards with no prompting as to what the originally chosen location was. Unless I've carefully remembered which location the candidate originally applied to, I risk accidentally changing their location. If I had to do this IRL I would cautiously click back to the application, make a note of the location, then click "Change" again.

I accept that not all study modes are available at all locations, so we can't guarantee that the chosen mode is available at the original location, but would it be friendly to 

- select the same location, if it's still available
and/or
- include a note of the previously-selected selection at the top of the location page?

WDYT @adamsilver @mnacos?

mnacos:
> Hello! Just beginning to review this. I noticed one UI quirk, which is that if I select "change to part time" or click the "change" link alongside study mode, I'm re-offered the "Location" screen afterwards with no prompting as to what the originally chosen location was. Unless I've carefully remembered which location the candidate originally applied to, I risk accidentally changing their location. If I had to do this IRL I would cautiously click back to the application, make a note of the location, then click "Change" again.
> 
> I accept that not all study modes are available at all locations, so we can't guarantee that the chosen mode is available at the original location, but would it be friendly to
> 
> * select the same location, if it's still available
>   and/or
> * include a note of the previously-selected selection at the top of the location page?
> 
> WDYT @adamsilver @mnacos?

I guess I could try matching on `site_id`. Strictly speaking, we don't show a list of locations on that step, but the locations of all valid course options for `{course_id, study_mode}`, the `id`s of which are guaranteed to be different. So, the code would have to pre-select, in effect, the nearest course_option (nearest being the same `site_id`, although I sense a geolocation feature coming!).

mnacos:
I've added a bit of code preselecting the course option that matches the same `site_id`. This fixes the issue @duncanjbrown describes.

app/models/provider_interface/change_offer_form.rb
```diff
@@ -64,6 +75,15 @@ def self.course_option_matches_course?(record)
       record.course_option_id && CourseOption.find(record.course_option_id).course.id == record.course_id
     end
 
+    def self.study_mode_valid_for_course?(record)
+      course = Course.find(record.course_id) if record.course_id
+      course && (course.both_study_modes_available? || course.study_mode == record.study_mode)
```
duncanjbrown, line 80: Does this feel like it wants to be a method on course: `course.supports_study_mode?(mode)` or equivalent?

spec/system/provider_interface/provider_changes_an_offer_spec.rb
```diff
@@ -41,13 +44,30 @@ def and_i_am_permitted_to_see_applications_for_two_providers
...
+    @other_course = create(:course, :open_on_apply, :with_both_study_modes, provider: @provider)
+    @course_option_two = create(:course_option, :full_time, course: @other_course)
+    @course_option_three = create(:course_option, :part_time, course: @other_course)
+  end
+
+  def and_a_part_time_only_course_exists_for_this_provider
+    @part_time_course = create(:course, :open_on_apply, :part_time, provider: @provider)
+  end
+
+  def and_an_older_course_exists_for_this_provider
```
duncanjbrown, line 61: Nice to see this here! Hoping we can have a proper recruitment cycle kickoff very soon

app/components/provider_interface/change_offer_component.rb
```diff
@@ -29,6 +29,19 @@ def sub_component
         ProviderInterface::ChangeOffer::ChangeLocationComponent.new(
           change_offer_form: change_offer_form,
         )
+      when :study_mode
+        study_modes = Course.find(change_offer_form.course_id).study_modes_present
+        if study_modes.count > 1
+          ProviderInterface::ChangeOffer::ChangeStudyModeComponent.new(
+            change_offer_form: change_offer_form,
+          )
+        else # skip to location
```
duncanjbrown, line 38: This works!

A small bikeshed: it's a bit of a shame to fix the order of steps out here. How do you feel about recursion? Not ideal with a method that relies on mutable object state like this one does, but it feels better not to duplicate the stepping logic

```ruby
change_offer_form.study_mode = study_modes.first
change_offer_form.step = change_offer_form.next_step
sub_component
```

app/components/provider_interface/change_offer/change_course_component.rb
```diff
@@ -18,13 +18,18 @@ def courses
...
+          recruitment_cycle_year: recruitment_cycle_year,
         ).order(:name)
       end
 
-      def study_mode_for_courses
-        current_study_mode = application_choice.offered_option.study_mode
-        [current_study_mode, :full_time_or_part_time]
+      def relevant_study_modes
+        current = change_offer_form.study_mode || application_choice.offered_option.study_mode
+        [current, 'full_time_or_part_time']
```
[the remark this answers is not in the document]
mnacos: Yes, you are right. This logic is consistent with preserving the current `study_mode`, but the whole point of this PR is to let providers change it. I'll remove all of `relevant_study_modes` and list all (open/current) courses. The hint text for each course mentions whether a course has full_time, part_time or both modes available, so it should work great.

app/models/course.rb
```diff
@@ -83,6 +83,10 @@ def both_study_modes_available?
     study_mode == 'full_time_or_part_time'
   end
 
+  def study_modes_present
```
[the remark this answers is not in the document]
mnacos: I've changed this to `available_study_modes_from_options`, just because it's derived from course options rather than the course's `study_mode`.

app/components/provider_interface/change_offer_component.rb
mnacos: Hmm, recursion feels a bit riskier, but as long as `step` doesn't get stuck on `study_mode` we should be ok.

duncanjbrown:
Thank you! Last nit: the "study mode" previews are going straight to location: https://apply-for-te-1995-chang-y7o4yn.herokuapp.com/rails/view_components/provider_interface/change_offer/existing_offer_start_at_3_study_mode

mnacos:
> Thank you! Last nit: the "study mode" previews are going straight to location: https://apply-for-te-1995-chang-y7o4yn.herokuapp.com/rails/view_components/provider_interface/change_offer/existing_offer_start_at_3_study_mode

Not much to fix there, it's behaving as expected. Gorse SCITT courses seem to be exclusively full-time, so the code skips to location, as intended. I'll change the default provider for review app test applications to `C85` which will give us both full-time and part-time options.

laurahermionetennant92:
This looks good @mnacos @adamsilver  but where we say 'Study mode' it should be 'Full time/part time' (like it is on the first screenshot)
:)

mnacos:
> This looks good @mnacos @adamsilver but where we say 'Study mode' it should be 'Full time/part time' (like it is on the first screenshot)
> :)

updated in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2155/commits/b920f5167e6923a16dea0a4eaf5dc587d4713c0c

adamsilver:
This should say ‘Full time or part time’

image

adamsilver:
I just tried to change study mode but kept it as the previous selection and the location screen still comes up.

mnacos:
> I just tried to change study mode but kept it as the previous selection and the location screen still comes up.

Hi Adam, for this iteration there is no skipping of location. Also, the current behaviour with all radio button screens is the option that's pre-selected is the current option. Are you saying we should change this if we know the user is changing the study mode?

mnacos:
> This should say ‘Full time or part time’
> 
> image

Fixed

PR #2159, opened 2020-05-28
Improve insta-sending applications to providers

tijmenb:
## Context

This is a iteration of https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2134. It aims to solve 2 problems: a recurrence of the nested transaction bug described in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/1912 (see [slack thread](https://ukgovernmentdfe.slack.com/archives/CN1MCQCHZ/p1590568483055500)). Secondly it aims to improve the logical flow - it feels weird that we have to move an application through several states in order to eventually end up in the desired state. 

This adds a direct path between unsubmitted and sent to provider states. This allows for a bunch of cleanup and optimisation. The key conditional when submitting an application reads fairly well now:

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/30d4d6140c471e612da8e5d4d768af08a3031d28/app/services/submit_application.rb#L17-L32 

This also removes the transaction, since it's tricky to implement for the `SendApplicationToProvider` flow.

## Guidance to review

Look at the individual commits, but also look at the end state.

## Link to Trello card

https://trello.com/c/6aEZkNZi

## Things to check


app/services/submit_application.rb
```diff
@@ -15,7 +15,14 @@ def call
...
 
-      submit_application
+      application_choices.each do |application_choice|
+        ApplicationStateChange.new(application_choice).submit!
+
+        if send_to_provider_immediately?
+          ApplicationStateChange.new(application_choice).references_complete!
+          SendApplicationToProvider.new(application_choice: application_choice).call
+        end
+      end
```
tvararu: 👍

PR #2161, opened 2020-05-28
Do not fail a build if there are warnings

vigneshmsft:
do not fail a build run if there are warnings in test runs.
Also publish the test results as build artifacts.

## Context

Current build fails, if there are warnings in the test runs.
Warnings might be intentional and we do not want to fail the build unless there are errors. 

Do not fail rspec test task in case of warnings.
Example build with warnings: https://dev.azure.com/dfe-ssp/Become-A-Teacher/_build/results?buildId=65745&view=results

## Things to check

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -255,6 +255,12 @@ stages:
         testRunner: JUnit
         testResultsFiles: 'results/*.xml'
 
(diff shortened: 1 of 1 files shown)


duncanjbrown:
:+1: Skipping an RSpec test shouldn't fail the build

PR #2162, opened 2020-05-28
Remove setting edit_by when submitting references

tijmenb:
## Context

This was introduced in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/commit/e129c5be147f4bfe158546cf5c166034ced8edd6, but I think it's not needed anymore. The `edit_by` date is already set when the application is submitted:

https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/b607578b5536756e5a2e3b7da2a931a64dbacdd3/app/services/submit_application.rb#L13-L14

Remove the code.

## Guidance to review

Is this right? @malcolmbaig to confirm.

## Link to Trello card

https://trello.com/c/6aEZkNZi

## Things to check

### DIFF
--- a/app/services/submit_reference.rb
+++ b/app/services/submit_reference.rb
@@ -38,13 +38,6 @@ private
       reference_feedback_provided!
       application_form.application_choices.awaiting_references.each do |application_choice|
         ApplicationStateChange.new(application_choice).references_complete!
-
-        next unless application_form.candidate_has_previously_applied?
-
-        # If the candidate has previously applied, they have less of a need to
-        # edit the application. Hence, we clear out the usual 7-day edit window
-        # by resetting the `edit_by` time.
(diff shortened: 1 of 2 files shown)


tijmenb:
I think I just understood the original intention. The `edit_by` isn't actually set to `now` if the references haven't come in for this application (this can happen when the candidate swaps a reference). So this code intends to prevent editing in this case. I think that still holds, even without this code, if we check for the form state when determing whether the user can still edit.

malcolmbaig:
This looks sound to me 👍 . Now that we're:

- enforcing the fact that apply_again applications can't be edited after submission
- setting the edit_by for apply_again applications to `now` at point of submission

the code in SubmitReference doesn't need to know anything about the apply_again or edit_by concepts.

PR #2165, opened 2020-05-28
Make permissions changes human-readable in the audit log

duncanjbrown:
## Context

Before this change, assigning a provider user to a provider resulted in an audit log entry like "Create ProviderPermissions #24, provider_id: 9".

This is hopeless for seeing what actually happened, so render a better string based on the data in the audit log.

("Comment" fields are not present usually)

**Before**

Screenshot 2020-05-28 at 14 49 30

**After**

Screenshot 2020-05-28 at 14 49 38

## Guidance to review

It's a bit hacky to look up the original provider from another audit entry in the event the original Permissions record was deleted, but I don't think it's too bad under the circumstances?

## Link to Trello card

https://trello.com/c/y07UkxMp/1877-build-show-access-control-audit-to-support-users

## Things to check

### DIFF
--- a/app/components/support_interface/audit_trail_item_component.rb
+++ b/app/components/support_interface/audit_trail_item_component.rb
@@ -11,6 +11,8 @@ module SupportInterface
     def audit_entry_event_label
       if audit.comment.present? && audit.audited_changes.empty?
         "Comment on #{audit.auditable_type.titlecase}"
+      elsif audit.auditable_type == 'ProviderPermissions'
+        label_for_provider_permission_change(audit)
       else
         "#{audit.action.capitalize} #{audit.auditable_type.titlecase} ##{audit.auditable_id}"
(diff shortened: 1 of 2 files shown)


willmcb:
Hey Duncan, just a thought I think you can move `with_audited: true` to the context in the spec and avoid a little repetition: 
```  
context 'the audited item is a ProviderPermissions record', with_audited: true do
  # your specs
end
```

duncanjbrown:
> Hey Duncan, just a thought, I think you can move `with_audited: true` to the context in the spec and avoid a little repetition:
> 
> ```ruby
> context 'the audited item is a ProviderPermissions record', with_audited: true do
>   # your specs
> end
> ```

I didn't want to do this because that would mean auditing the provider and provider_user each time, which seemed wasteful.

PR #2173, opened 2020-05-29
Call to action email for previously unsuccessful applications to apply again

stevehook:
## Context

When we ship 'apply again' we want to notify applicants who were unsuccessful but would not at the time have been able to apply again that they can now do so.

- [x] New methods for `CandidateMailer` for each of the main four cases
  - [x] Rejected applications
  - [x] Declined applications
  - [x] Withdrawn applications
  - [x] DBD applications
- [ ] Handle case where there is a rejection as well as a one of the other reasons for unsuccessful application
- [ ] Mailer previews
- [ ] Worker that iterates all the unsuccessful applications before a given date and invokes the right call-to-action mailer above.
- [ ] rake task to kick off the above (possibly with a cutoff date as input - can default to now).

## Link to Trello card

https://trello.com/c/5J7sX8iQ/1591-dev-build-email-for-submitted-your-application-for-apply-again-mvp

## Things to check


stevehook:
Closing this because the requirement is much simpler.

PR #2176, opened 2020-05-29
Add ProviderInterface::HeaderComponent & spec

willmcb:
## Context

New header component for provider interface

**Before**
Screenshot 2020-05-27 at 11 39 45

**After**
Screenshot 2020-05-27 at 16 44 48

## Guidance to review

- does it look okay throughout the app?

## Link to Trello card

https://trello.com/c/st1Mq1h6/2005-build-implement-the-new-header-for-manage-applications

## Things to check

### DIFF
--- a/app/components/provider_interface/header_component.html.erb
+++ b/app/components/provider_interface/header_component.html.erb
@@ -1,0 +1,34 @@
+<header class="govuk-header <%= classes %>" role="banner" data-module="govuk-header">
+  <div class="govuk-header__container govuk-width-container">
+    <div class="govuk-header__logo">
+      <%= link_to 'https://gov.uk', class: 'govuk-header__link govuk-header__link--homepage' do %>
+        <span class="govuk-header__logotype">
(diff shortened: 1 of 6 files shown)


adamsilver:
I feel like there should be on component called Header which can be called with different params provider compared to candidate or support.

This looks like the bulk of the header HTML has been duplicated/hard coded. 

What's the thinking on this?

willmcb:
Thinking was that really it is a simple way to do it, given that this might change in the near future it is easy to just replace one with the other rather than baking conditional logic into the component. 

Currently there is already conditional logic in the `_header.erb` view that can be used to render different components so I thought this was the straightest way of going about this change. 

The new header is completely encapsulated in one place under a given namespace - if we need to use it more widely we can just remove it from that name space add it to the global name space and there is no real need to change the code much (other than some module names). 

If we have to add conditional logic to the component and then we decide to go with the newer version we would have to delete the redundant conditional logic in any case.

adamsilver:
No worries @willmcb @duncanjbrown – all good.

I know it's slightly different but that's fine.

The prototype doesn‘t have the email so let's leave it as it doesn't navigate anywhere.

duncanjbrown:
@adamsilver just confirming you're happy with the way this looks?

adamsilver:
Approved

PR #2182, opened 2020-05-29
Email candidates with unsuccessful applications before apply again

stevehook:
## Context

We want to send a call to action email to all candidates whose applications ended unsuccessfully before we released apply again.

- [x] `CandidateMailer#apply_again_call_to_action` email
- [x] Mailer preview for the above
- [x] `SendApplyAgainCallToAction` service class to email all unsuccessful candidates

## Guidance to review

image

- I didn't want to over-engineer this as it's a one-off mailshot but do we need a way to ensure that we don't send these emails twice to the same person? e.g. a field in the database.
- Do we need a rake task to trigger the service class?
- Do we need to parameterise the service class (and rake task) with a datetime for the time that we released the apply again feature or is it ok to assume this will be run soon enough after that happens?

## Link to Trello card

https://trello.com/c/yyvDNpCO/1607-dev-build-email-for-candidates-who-were-unsuccessful-when-we-didnt-have-apply-again

## Things to check

### DIFF
--- a/app/services/send_apply_again_call_to_action.rb
+++ b/app/services/send_apply_again_call_to_action.rb
@@ -13,10 +13,12 @@ private
   def unsuccessful_application_forms
     ApplicationForm
       .joins(:application_choices)
+      .joins("LEFT OUTER JOIN emails ON emails.application_form_id = application_forms.id AND emails.mailer = 'candidate_mailer' AND emails.mail_template = 'apply_again_call_to_action'")
       .where
       .not(application_choices: {
         status: ApplicationStateChange.valid_states - ApplicationStateChange::UNSUCCESSFUL_END_STATES.map(&:to_sym),
       })
+      .where(emails: { id: nil })
       .distinct
   end
 end
--- a/app/services/send_apply_again_call_to_action.rb
+++ b/app/services/send_apply_again_call_to_action.rb
@@ -14,11 +14,12 @@ private
     ApplicationForm
       .joins(:application_choices)
       .joins("LEFT OUTER JOIN emails ON emails.application_form_id = application_forms.id AND emails.mailer = 'candidate_mailer' AND emails.mail_template = 'apply_again_call_to_action'")
-      .where
-      .not(application_choices: {
+      .joins('LEFT OUTER JOIN application_forms AS subsequent_application_form ON application_forms.id = subsequent_application_form.previous_application_form_id')
+      .where.not(application_choices: {
         status: ApplicationStateChange.valid_states - ApplicationStateChange::UNSUCCESSFUL_END_STATES.map(&:to_sym),
       })
       .where(emails: { id: nil })
+      .where(subsequent_application_form: { id: nil })
       .distinct
   end
 end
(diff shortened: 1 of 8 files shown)


app/services/send_apply_again_call_to_action.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+private
+
+  def unsuccessful_application_forms
+    ApplicationForm
+      .joins(:application_choices)
+      .where
+      .not(application_choices: {
+        status: ApplicationStateChange.valid_states - ApplicationStateChange::UNSUCCESSFUL_END_STATES.map(&:to_sym),
```
malcolmbaig, line 18: Can we just query for `where(application_choices: { status: ApplicationStateChange::UNSUCCESSFUL_END_STATES })` ?

app/services/send_apply_again_call_to_action.rb
```diff
@@ -0,0 +1,22 @@
...
+  def perform
+    unsuccessful_application_forms.find_each do |application_form|
+      CandidateMailer.apply_again_call_to_action(application_form).deliver
+    end
+  end
+
+private
+
+  def unsuccessful_application_forms
+    ApplicationForm
```
malcolmbaig, line 14: I wonder if it's worth explicitly excluding applications that have already started apply_again, in case there's a significant delay between launching the feature and sending the email?

malcolmbaig:
> I didn't want to over-engineer this as it's a one-off mailshot but do we need a way to ensure that we don't send these emails twice to the same person? e.g. a field in the database.

Good question - not sure, but we're not doing any handling for this case in other emails, one-shot or otherwise.

>Do we need a rake task to trigger the service class?

I think we're fine to just `rails runner` the service class on the console.

> Do we need to parameterise the service class (and rake task) with a datetime for the time that we released the apply again feature or is it ok to assume this will be run soon enough after that happens?

The latter, IMHO.

app/services/send_apply_again_call_to_action.rb
stevehook: The intention here is to get applications (forms) for which all the application choices have ended unsuccessfully. So if a candidate got rejected for one choice and accepted an offer for another (or just hasn't heard yet) then we shouldn't be sending this email.

stevehook:
> > I didn't want to over-engineer this as it's a one-off mailshot but do we need a way to ensure that we don't send these emails twice to the same person? e.g. a field in the database.
> 
> Good question - not sure, but we're not doing any handling for this case in other emails, one-shot or otherwise.
> 
> > Do we need a rake task to trigger the service class?
> 
> I think we're fine to just `rails runner` the service class on the console.
> 
> > Do we need to parameterise the service class (and rake task) with a datetime for the time that we released the apply again feature or is it ok to assume this will be run soon enough after that happens?
> 
> The latter, IMHO.

I've added a condition to the query to exclude applications that have already started an apply again. I also added another condition to avoid the duplicate emails - I had forgotten that we log emails so it's fairly easy to do that.

PR #2189, opened 2020-06-01
Fix font size and spacing on applications appearing in provider application index

willmcb:
## Context

Fixes font sizes and spacing of note messages on application cards on the provider interface application index view.

**Before:**
Screenshot 2020-06-01 at 16 15 12

**After:**
Screenshot 2020-06-01 at 16 14 23

## Guidance to review

- Does it look like it matches the design to you?

## Link to Trello card

https://trello.com/c/d1WgjClI/2218-wrong-font-size-and-spacing-on-applications-appearing-in-the-list

## Things to check

### DIFF
--- a/app/components/provider_interface/application_card_component.html.erb
+++ b/app/components/provider_interface/application_card_component.html.erb
@@ -7,7 +7,7 @@
         <%= render ProviderInterface::ApplicationStatusTagComponent.new(application_choice: application_choice) %>
       </div>
       <% if most_recent_note %>
-        <span class="app-application-card__note"><span class='note-bullet-point'> </span><%= most_recent_note.subject %></span>
+        <span class="app-application-card__note"><span aria-hidden='true' class='note-bullet-point'> </span><%= most_recent_note.subject %></span>
       <% end %>
     <div>
   </h3>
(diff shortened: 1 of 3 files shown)


adamsilver:
Font size looks good

Spacing isn’t quite right. The long note text should wrap onto multiple lines instead of pushing the left hand content down.

willmcb:
Screenshot 2020-06-02 at 16 06 47

adamsilver:
Hey @willmcb – I made a video of the remaining issues.

https://www.loom.com/share/00b960c278174d5083160503000aa891

willmcb:
@adamsilver the bug also appears on the prototype - it is due to having three particularly long words and having `word-break: normal` set there is nowhere for things to break other than in whitespaces so it causes this issue. 

You can set `word-break: break-all` and this will break mid word but I'm not sure that is preferable:

Screenshot 2020-06-05 at 14 56 53

willmcb:
Have rolled back to add a `<span>` for the bullet point and added `aria-hidden="true"` and so that it is ignored by screen readers - result being:

Screenshot 2020-06-08 at 14 30 14

PR #2200, opened 2020-06-03
Remove provider user email from banner

willmcb:
This is not a nav item in the design, and causes a nav item overflow
issue with long email addresses due to the new slimmer banner.

Provider user email is replaced with a link to the provider applications
view as per the prototype design.

## Context

With the new banner long emails cause a mis-alignment of nav items, which was identified on QA in product review. 

The default Dev email is not long enough for this to have been an issue.

Emails are not part of the header designs on the prototype so given this they are removed in this PR.

Before:
Screenshot 2020-06-03 at 11 55 11

After:
Screenshot 2020-06-03 at 11 55 50

## Guidance to review

- is there a reason we shouldn't do this? 

## Link to Trello card

- follow up to: https://trello.com/c/st1Mq1h6/2005-build-implement-the-new-header-for-manage-applications

## Things to check

### DIFF
--- a/app/models/navigation_items.rb
+++ b/app/models/navigation_items.rb
@@ -39,7 +39,7 @@ class NavigationItems
     def for_provider_interface(current_provider_user)
       return [NavigationItem.new('Sign in', provider_interface_sign_in_path, false)] unless current_provider_user
 
-      items = [NavigationItem.new(current_provider_user.email_address, nil, false)]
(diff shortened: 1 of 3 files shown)


adamsilver:
@duncanjbrown this is the plan (if you fancy whipping up an MVP version of this quick):

image

PR #2202, opened 2020-06-03
Add organisations index and details pages

steventux:
## Context

Part of the provider relationship permissions journey relies on an organisations page for a provider user. This page links pages per organisation which display the permissions for that org.

- Adds `/provider/organisations` index page
- Adds a page displaying permissions for a single organisation relative to their relationship to various other ratifying and training providers.
- Adds a navigation link for organisations index page

#### Organisations index

image

#### Organisations permissions details
image

 

## Guidance to review

I haven't placed any of these pages behind feature flags as there are no links to actions to manage permissions etc. This might not be the right approach.

## Link to Trello card

https://trello.com/c/RDYz6s69/2251-build-provider-organisations-pages

## Things to check

### DIFF
--- a/app/views/provider_interface/organisations/_permissions_list.html.erb
+++ b/app/views/provider_interface/organisations/_permissions_list.html.erb
@@ -1,0 +1,12 @@
+<ul class="govuk-list">
+  <% if permission.view_safeguarding_information %>
+    <li>
+      <%= render 'check_icon' %>
+      access safeguarding information
+    </li>
+  <% end %>
+  <li>
+    <%= render 'check_icon' %>
+    view applications
+  </li>
+</ul>
--- a/app/views/provider_interface/organisations/show.html.erb
+++ b/app/views/provider_interface/organisations/show.html.erb
@@ -32,6 +32,13 @@
 
       <p class="govuk-!-margin-bottom-1 govuk-body"><%= ratifying_permission.ratifying_provider.name %> can:</p>
       <%= render 'permissions_list', permission: ratifying_permission %>
+
+      <p class="govuk-body">
+        <%= govuk_link_to 'Change', provider_interface_edit_provider_relationship_permissions_path(
+          training_provider_id: training_permission.training_provider.id,
+          ratifying_provider_id: training_permission.ratifying_provider.id,
+        ) if current_provider_user.can_manage_organisations? %>
+      </p>
     <% end %>
   </div>
 </div>
--- a/app/views/provider_interface/organisations/show.html.erb
+++ b/app/views/provider_interface/organisations/show.html.erb
@@ -1,0 +1,37 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl"><%= @provider.name %></h1>
+    <% @ratifying_permissions.each do |_id, permissions| %>
+      <% ratifying_permission, training_permission = permissions %>
+      <h2 class="govuk-heading-m">
+        For courses ratified by <%= ratifying_permission.ratifying_provider.name %> and run by <%= ratifying_permission.training_provider.name %>
+      </h2>
+      <div class="app-banner govuk-!-margin-bottom-4">
+        <div class="app-banner__message">
+          Contact <%= ratifying_permission.training_provider.name %> to change permissions.
+        </div>
+      </div>
+
+      <h2 class="govuk-heading-m">
+        For courses ratified by <%= ratifying_permission.ratifying_provider.name %> and run by <%= ratifying_permission.training_provider.name %>
+      </h2>
+      <p class="govuk-!-margin-bottom-1 govuk-body"><%= ratifying_permission.ratifying_provider.name %> can:</p>
+      <%= render 'permissions_list', permission: ratifying_permission %>
+
+      <p class="govuk-!-margin-bottom-1 govuk-body"><%= ratifying_permission.training_provider.name %> can:</p>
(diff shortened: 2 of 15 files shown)


duncanjbrown:
Code looks very good to me. 

One UI thing: if I don't have any relationships to manage the page looks oddly empty. Do we need an empty state listing all the permissions and/or explaining there's no relationship to manage? I'm not sure whether we should have the header link at all if there are no relationships to learn about. It does feel right that if there are relationships which are affecting what the user can and can't do then we should make them visible.

Screenshot 2020-06-04 at 13 43 18

Minor UI nit: spacing underneath the provider name looks a bit tight here compared with the prototype

Screenshot 2020-06-04 at 13 44 04

Screenshot 2020-06-04 at 13 58 47

steventux:
@duncanjbrown I see what you mean, it looks bad. I'd gone on the assumption that all provider users would belong to at least one provider, but I hadn't considered that if that provider was self-ratifying then no relationships to other providers would exist and we'd get this blank page.
I'll remove the nav link unless manageable relationships exist.

willmcb:
LGTM 👍

app/views/provider_interface/organisations/_permissions_list.html.erb
```diff
@@ -0,0 +1,12 @@
+<ul class="govuk-list">
+  <% if permission.view_safeguarding_information %>
+    <li>
+      <%= render 'check_icon' %>
+      access safeguarding information
+    </li>
+  <% end %>
+  <li>
```
mnacos, line 8: Do we need an org-level 'Make decisions` entry?

steventux: @mnacos not sure if this is an org level permission, I think if it is, we should do it in a separate PR to keep this moving as we'll need to amend `ProviderRelationshipPermissions` with a suitable migration etc.

mnacos: As an organisation, I'd want to be able to grant/withdraw an org-level 'Make decisions' but I guess we should double-check. Like you say, this can be done as a separate PR.

app/views/provider_interface/organisations/show.html.erb
```diff
@@ -0,0 +1,37 @@
...
+        For courses ratified by <%= ratifying_permission.ratifying_provider.name %> and run by <%= ratifying_permission.training_provider.name %>
+      </h2>
+      <div class="app-banner govuk-!-margin-bottom-4">
+        <div class="app-banner__message">
+          Contact <%= ratifying_permission.training_provider.name %> to change permissions.
+        </div>
+      </div>
+
+      <h2 class="govuk-heading-m">
+        For courses ratifyied by <%= ratifying_permission.ratifying_provider.name %> and run by <%= ratifying_permission.training_provider.name %>
```
duncanjbrown, line 16: Typo: `ratifyied`

steventux: @duncanjbrown good spot - fixed here https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2202/commits/435b376dc918e1839fa1f4b601feab409289e1e2#diff-979630c1c1029745e95aa0c3369a8b4fR16

duncanjbrown:
Almost there — the spacing is still off on the list of providers, and in the event I have a mixture of manageable and unmanageable providers, I can still click through to the unmanageable ones, resulting in the blank screen?

steventux:
Sorry @duncanjbrown, less haste, more speed. I think both outstanding issues are fixed by https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2202/commits/4abe78b5cc7389addedcbcfd727e3ffe04dd9751 now.

duncanjbrown:
More! Sorry.

- Selecting no checkboxes on the organisations page gives me an error:

Screenshot 2020-06-10 at 14 23 43

- I still get `Access Denied` when clicking on a provider that is ratifying-only from the list of providers

- I think we should have a gap under the "Change" link?

Screenshot 2020-06-11 at 15 03 40

- Maybe not for this PR. The end of the "change" flow is a bit odd when seen for the second time as it assumes I'm new to the process:

Screenshot 2020-06-10 at 14 22 45

(btw, I wonder if this page should always have the `Users` link — providers could have orgs but not users permission)

steventux:
@duncanjbrown thanks again for the review, I've attempted to address each point:

- The controller error should be fixed by https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2202/commits/be15794b122f9ab57dbdc652d608d7197d69b178 though we still need to address whether a user should be able to unset all permissions, perhaps in another story.
- The `Access Denied` issue should be rectified by widening the scope of `Provider.with_permissions_visible_to` as this now retrieves both training and ratifying providers as both should be visible to the user (though not necessarily _manageable_).
- I've placed the 'Change' link in a paragraph tag for additional bottom spacing https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2202/commits/7b3100e4358be2c196fd676cb4e93fe31d14a204
- The end of the change flow is odd here I agree. The confirmation and success steps are part of the onboarding journey so at the very least the wording could be _Permissions successfully updated_.
 @adamsilver should we do something different for post-setup updates to organisation permissions? (The design app doesn't seem to allow me to change permissions from the org page)

adamsilver:
@steventux I’ve updated the prototype to show how editing permissions work:
https://manage-applications-beta.herokuapp.com/organisations/show3

PR #2203, opened 2020-06-03
Lower RBD time limit to 20 days after 1st July 2020

stevehook:
## Context

UCAS have confirmed that the RBD date should be lowered to 20 (business) days from 1st July.

- [x] Add new configuration rule with a start date of 1/7/2020 that applies the new limit after that date.
- [x] Add a spec that tests the `TimeLimitCalculator` together with the new config (not stubbed).

## Guidance to review

- Nothing in particular

## Link to Trello card

https://trello.com/c/iawQmftw/1566-dev-lower-rbd-to-20-days-from-1st-july

## Things to check

### DIFF
--- a/app/components/candidate_interface/referee_guidance_component.rb
+++ b/app/components/candidate_interface/referee_guidance_component.rb
@@ -10,6 +10,13 @@ module CandidateInterface
       'provider'.pluralize(provider_count)
     end
 
+    def reject_by_default_days
+      @reject_by_default_days ||= TimeLimitCalculator.new(
+        rule: :reject_by_default,
+        effective_date: @application_form&.application_choices&.first&.sent_to_provider_at || Time.zone.now,
+      ).call[:days]
+    end
+
   private
 
     def provider_count
--- a/app/components/candidate_interface/referee_guidance_component.rb
+++ b/app/components/candidate_interface/referee_guidance_component.rb
@@ -13,7 +13,7 @@ module CandidateInterface
     def reject_by_default_days
       @reject_by_default_days ||= TimeLimitCalculator.new(
         rule: :reject_by_default,
-        effective_date: @application_form&.application_choices&.first&.sent_to_provider_at || Time.zone.now,
+        effective_date: @application_form.application_choices.first&.sent_to_provider_at || Time.zone.now,
       ).call[:days]
     end
 
--- a/app/components/candidate_interface/referee_guidance_component.html.erb
+++ b/app/components/candidate_interface/referee_guidance_component.html.erb
@@ -9,5 +9,5 @@
 <% end %>
 
 <% unless FeatureFlag.active?('covid_19') %>
-  <p class="govuk-body">Training providers then have <%= TimeLimitConfig.limits_for(:reject_by_default).first.limit %> working days to respond to your application.</p>
+  <p class="govuk-body">Training providers then have <%= reject_by_default_days %> working days to respond to your application.</p>
 <% end %>
--- a/app/lib/time_limit_config.rb
+++ b/app/lib/time_limit_config.rb
@@ -51,6 +51,7 @@ class TimeLimitConfig
     ],
     chase_provider_before_rbd: [
       Rule.new(nil, nil, 20),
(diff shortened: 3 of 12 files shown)


duncanjbrown:
Do we need to cover `chase_provider_before_rbd` too?

stevehook:
> chase_provider_before_rbd

Yes that's a good point @duncanjbrown, I guess we need to send the chaser half-way through that period? (The default `chase_provider_before_rbd` limit is 20 days when the RBD limit is 40 days)

duncanjbrown:
> > chase_provider_before_rbd
> 
> Yes that's a good point @duncanjbrown, I guess we need to send the chaser half-way through that period? (The default `chase_provider_before_rbd` limit is 20 days when the RBD limit is 40 days)

Sounds like a good place to start!

I've also just noticed we access `TimeLimitConfig.limits_for(:rule).first` directly quite a lot in emails, which will result in incorrect information going out during periods when different rules apply.

stevehook:
> > chase_provider_before_rbd
> 
> Yes that's a good point @duncanjbrown, I guess we need to send the chaser half-way through that period? (The default `chase_provider_before_rbd` limit is 20 days when the RBD limit is 40 days)

There is an assumption about `chase_provider_before_rbd` in ProviderMailer having only a default rule that I've tried to fix. Checking for any others.

stevehook:
@duncanjbrown I think this is ready for another look if you have a few moments. I've fixed one issue with the `ProviderMailer` due to a `TimeLimitConfig.limits_for(:rule).first`. There are a few others in specs, I had to freeze time to a known date to get one to pass and others already do this so they should be fine until we change the default rule. Can you think of anything else I need to address?

There is a subtle issue with `reject_by_default`/`chase_provider_before_rbd`. It's possible to set RBD using the default rule (40 days) and then have the chaser sent using the shortened time limit (10 days). This is because we use the current time to get the applicable limit and use it to count backwards from the RBD, so if RBD is more than 20 business days after 1st July we'll be applying the new rule for the chaser even though we may have used the default rule for the RBD. I don't think this is a major issue.

duncanjbrown:
Good point about changing RBDs! Agree that doesn't sound too bad. I suppose the most correct way to do it would be to establish when the rdb was calculated and call up the appropriate chase_provider_before_rbd definition. I think there's a case for writing the chase date onto the model at the time of calculation, as we do with the rbd date? Maybe next year!

I think the `:reject_by_default` date is also going to go wrong in `candidate_mailer/application_submitted.text.erb` and `candidate_interface/referee_guidance_component.html.erb` (neither shown at the moment as the COVID flag is on, but they are wrong).

stevehook:
> Good point about changing RBDs! Agree that doesn't sound too bad. I suppose the most correct way to do it would be to establish when the rdb was calculated and call up the appropriate chase_provider_before_rbd definition. I think there's a case for writing the chase date onto the model at the time of calculation, as we do with the rbd date? Maybe next year!?
> 
> I think the `:reject_by_default` date is also going to go wrong in `candidate_mailer/application_submitted.text.erb` and `candidate_interface/referee_guidance_component.html.erb` (neither shown at the moment as the COVID flag is on, but they are wrong).

I think asked about about storing the chase by date in the DB but there was some reason why it didn't happen. There are pros and cons - storing the dates makes most things easier except when you get a global pandemic and have to change the limits and recalculate them all! On balance though I think storing in DB is best.

I've pushed up fixes for `candidate_mailer/application_submitted.text.erb` and `candidate_interface/referee_guidance_component.html.erb`.

app/components/candidate_interface/referee_guidance_component.rb
```diff
@@ -10,6 +10,13 @@ def pluralize_provider
       'provider'.pluralize(provider_count)
     end
 
+    def reject_by_default_days
+      @reject_by_default_days ||= TimeLimitCalculator.new(
+        rule: :reject_by_default,
+        effective_date: @application_form&.application_choices&.first&.sent_to_provider_at || Time.zone.now,
```
duncanjbrown, line 16: We trust that `application_form.application_choices` are present (even if empty) in `provider_count` below — could we do away with the first two `&.`?

PR #2222, opened 2020-06-05
Consistent capitalisation for Get Into Teaching

paulrobertlloyd:
## Context

Get Into Teaching capitalise each part of the their name (whether they should or not, is another matter). We currently refer to this service as:

* Get Into Teaching
* Get into Teaching
* Get into teaching

We should be consistent.


fofr:
I wonder if a helper would stop us repeating this mistake? 🤔

paulrobertlloyd:
Perhaps this string should be localised, like we do for our own service names?

PR #2227, opened 2020-06-08
Add initial HESA dataset

malcolmbaig:
## Context
The Higher Education Statistics Agency provides structured data to
describe the following characteristics of degree courses:

- classes (grades)
- establishments
- subjects
- types

We want to use this data when candidates specify their degree as part of
their applications.

Using raw CSV files as a source, add an initial cut of this data in the
form of several arrays so that we can start using it in code.

## Guidance to review
A follow-on ticket will consider how we apply transformations to the data and will revisit how it's structured. The purpose of this PR is mainly visibility of what we're adding and broad agreement that we're happy with:

- the names of things
- storing the data in these constants, for now

## Link to Trello card

https://trello.com/c/K6UV4aVs
## Things to check


duncanjbrown:
Looks good! What do you think about namespacing these under `HESA_`, or even `HESA_2020_` so we know who they belong to?

PR #2229, opened 2020-06-08
Fix Withdrawn by us and Rejected filters

AgaDufrat:
## Context

'Withdrawn by us' filter was showing no results and 'Rejected' filter included
both rejected and withdrawn by us applications. This was caused by the fact
that the status filter was running WHERE status query. However
`offer_withdrawn` does not exist in the database; those applications have
`rejected` status in the database.

- separately add offer withdrawn application choices when 'Withdrawn by us' filter is selected
- when the 'Rejected' filter is selected, exclude 'Withdrawn by us' applications from the results

| Before  | After |
| ------------- | ------------- |
| Screenshot 2020-06-08 at 14 46 13 |Screenshot 2020-06-08 at 14 40 05|
|Screenshot 2020-06-08 at 14 39 48|Screenshot 2020-06-08 at 14 38 45|

## Guidance to review
- go to http://localhost:3000/provider/applications
- try different filters such as:
  - 'Withdrawn by us' only
  - 'Rejected' only
  - 'Withdrawn by us' & 'Rejected' 
  - 'Withdrawn by us' & some other filters
  - 'Rejected' & some other filters

_Note:_ In the service spec I explicitly specified the status to highlight it even though it's sometimes done by the factory - happy to change it.

## Link to Trello card

https://trello.com/c/U0OeofIp/2116-title-provider-interface-application-filter-by-status-wrong

## Things to check

### DIFF
--- a/spec/services/provider_interface/filter_application_choices_for_providers_spec.rb
+++ b/spec/services/provider_interface/filter_application_choices_for_providers_spec.rb
@@ -1,0 +1,70 @@
+require 'rails_helper'
+
+RSpec.describe FilterApplicationChoicesForProviders do
+  include CourseOptionHelpers
+
+  describe '#call' do
+    let(:submitted_application_choice) { create(:application_choice, :awaiting_provider_decision, status: 'awaiting_provider_decision') }
+    let(:offered_application_choice) { create(:application_choice, :with_offer, status: 'offer') }
+    let(:accepted_application_choice) { create(:application_choice, :with_accepted_offer, status: 'pending_conditions') }
+    let(:conditions_met_application_choice) { create(:application_choice, :awaiting_provider_decision, status: 'recruited') }
+    let(:enrolled_application_choice) { create(:application_choice, status: 'enrolled') }
+    let(:rejected_application_choice) { create(:application_choice, :with_rejection, status: 'rejected') }
+    let(:declined_application_choice) { create(:application_choice, :with_declined_offer, status: 'declined') }
+    let(:application_withdrawn_application_choice) { create(:application_choice, status: 'withdrawn') }
+    let(:conditions_not_met_application_choice) { create(:application_choice, status: 'conditions_not_met') }
+    let(:withdrawn_by_us_application_choice) { create(:application_choice, status: 'rejected', offer_withdrawn_at: 2.days.ago) }
+
+    context 'when filtering by status' do
+      let(:application_choices) { ApplicationChoice.all }
+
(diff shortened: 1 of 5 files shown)


duncanjbrown:
This is interesting. Without comparing dates, there is no way to determine that an application is definitely in one or the other state because it's possible for an application to be withdrawn, then rejected!

duncanjbrown:
I think if we additionally scope `offer_withdrawn` to rejected applications (the `push` below), the logic boils down to the below.

```ruby
def status(application_choices, statuses)
  return application_choices if statuses.blank?

  if statuses.include?('offer_withdrawn')
    statuses.push('rejected')
    application_choices = application_choices.where.not(offer_withdrawn_at: nil)
  elsif statuses.include?('rejected')
    application_choices = application_choices.where(offer_withdrawn_at: nil)
  end

  application_choices.where(status: statuses)
end
```

It's painful that we have this pseudo-state!

AgaDufrat:
Thanks 👍 , as discussed I'm adding a scope on the model and unit tests for this service.

spec/services/provider_interface/filter_application_choices_for_providers_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+          { filter: 'Conditions met', filter_status: 'recruited' },
+          { filter: 'Enrolled', filter_status: 'enrolled' },
+          { filter: 'Rejected', filter_status: 'rejected' },
+          { filter: 'Declined', filter_status: 'declined' },
+          { filter: 'Application withdrawn', filter_status: 'withdrawn' },
+          { filter: 'Conditions not met', filter_status: 'conditions_not_met' },
+          { filter: 'Withdrawn by us', filter_status: 'offer_withdrawn' },
+        ]
+
+      test_data.each_with_index do |example, index|
```
duncanjbrown, line 35: Nice approach! Good to cover everything. The order dependency is a shame, but preferable to more metaprogramming, I think?

spec/services/provider_interface/filter_application_choices_for_providers_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+        it "single #{example[:filter]} filter returns only #{example[:filter].downcase} application choices" do
+          subject = described_class.call(
```
duncanjbrown, line 37: `subject` has special meaning in RSpec and there's usually a better name for whatever it is — `returned_application_choices` or sth?

PR #2230, opened 2020-06-08
Migration adds `application_choices.course_unavailable_notification_sent_at`

stevehook:
## Context

We need to send email notifications to candidates who have applied to a course that gets full while they are waiting for references. This is a migration-only PR to support that feature.

To implement this feature we will poll the database for applications that are in `awaiting_references` status for which the selected course is no longer available.

Finding applications are in the right status with full courses is easy enough but we also need to exclude those for which we have already sent a notification. Initially I thought that we could use the `emails` table but there are a couple of problems with this:

1. It is associated with `ApplicationForm` not `ApplicationChoice`.
2. We would not be able to handle updates to the course choice and the edge case where you have a second choice filling up as well as the first.

So I think we need a specific attribute for this on the `ApplicationChoice` model hence this migration.

## Guidance to review

Is there another way to do this that doesn't involve adding to the database schema?

## Link to Trello card

https://trello.com/c/qX7OyNDH/1647-dev-notification-emails-when-a-course-becomes-full-while-waiting-for-references

## Things to check


stevehook:
Going to close this in favour of using `chasers_sent`.

PR #2231, opened 2020-06-08
Provider user needing to set organisation permissions sets them on first login

steventux:
## Context

Users with the `manage_organisations` permissions can assign provider relationship permissions for their training provider and the ratifying provider for relevant courses.
These relationships between providers are defined and populated from data in Find.
The relationships are treated as 'set up' once the designated user has set permissions and confirmed them.

- Adds a filter in `ProviderInterfaceController` to redirect the relevant user to set up organisation permissions.
- This takes the user into a set of steps which detail what the permissions mean and allows them to select the appropriate permissions for training and ratifying providers.
- Saves a timestamp on each `ProviderRelationshipPermissions` record confirmed by the user.
- The user exits these steps once all associated provider relationship permissions have been set up and confirmed.

#### 1. Setup step

image

#### 2. Assign permissions step

image

#### 3. Confirmation

image

#### 4. Success

image

## Link to Trello card

https://trello.com/c/hX3bkcvr/2236-%F0%9F%8F%88-a-provider-user-needing-to-set-organisation-permissions-sets-them-on-first-login

## Things to check

### DIFF
--- a/app/controllers/provider_interface/provider_interface_controller.rb
+++ b/app/controllers/provider_interface/provider_interface_controller.rb
@@ -15,6 +15,7 @@ module ProviderInterface
     before_action :authenticate_provider_user!
     before_action :add_identity_to_log
     before_action :check_data_sharing_agreements
+    before_action :check_provider_relationship_permissions
 
     layout 'application'
 
@@ -64,6 +65,29 @@ module ProviderInterface
       if GetPendingDataSharingAgreementsForProviderUser.call(provider_user: current_provider_user).any?
         redirect_to provider_interface_new_data_sharing_agreement_path
       end
+    end
+
+    def check_provider_relationship_permissions
+      return unless FeatureFlag.active?('enforce_provider_to_provider_permissions')
+      return if request.path == provider_interface_provider_relationship_permissions_setup_path
+      return unless current_provider_user
+
+      if provider_permissions_need_setup?
+        redirect_to provider_interface_provider_relationship_permissions_setup_path
+      end
+    end
+
+    def provider_permissions_need_setup?
+      permissions = TrainingProviderPermissions.find_by(
+        setup_at: nil,
+        training_provider: current_provider_user.providers,
+      )
+
+      return false if permissions.blank?
+
+      ProviderAuthorisation.new(actor: current_provider_user).can_manage_organisation?(
+        provider: permissions.training_provider,
+      )
     end
 
     def render_404
--- a/app/controllers/provider_interface/provider_interface_controller.rb
+++ b/app/controllers/provider_interface/provider_interface_controller.rb
@@ -69,8 +69,8 @@ module ProviderInterface
 
     def check_provider_relationship_permissions
       return unless FeatureFlag.active?('enforce_provider_to_provider_permissions')
-      return if request.path == provider_interface_provider_relationship_permissions_setup_path
       return unless current_provider_user
+      return if performing_provider_organisation_setup?
 
       if provider_permissions_need_setup?
         redirect_to provider_interface_provider_relationship_permissions_setup_path
@@ -88,6 +88,13 @@ module ProviderInterface
       ProviderAuthorisation.new(actor: current_provider_user).can_manage_organisation?(
         provider: permissions.training_provider,
       )
+    end
+
+    def performing_provider_organisation_setup?
+      [
+        ProviderInterface::ProviderAgreementsController,
+        ProviderInterface::ProviderRelationshipPermissionsController,
+      ].include?(request.controller_class)
     end
 
     def render_404
--- a/app/controllers/provider_interface/provider_relationship_permissions_controller.rb
+++ b/app/controllers/provider_interface/provider_relationship_permissions_controller.rb
@@ -7,7 +7,7 @@ module ProviderInterface
       @training_provider_permissions = TrainingProviderPermissions.where(
         setup_at: nil,
         training_provider: current_provider_user.providers,
-      )
+      ).includes(%i[training_provider ratifying_provider]).order(:created_at)
     end
 
     def success; end
--- a/app/controllers/provider_interface/provider_relationship_permissions_controller.rb
+++ b/app/controllers/provider_interface/provider_relationship_permissions_controller.rb
@@ -3,6 +3,13 @@ module ProviderInterface
     before_action :render_404_unless_permissions_found
     before_action :render_403_unless_access_permitted
 
+    def setup
+      @training_provider_permissions = TrainingProviderPermissions.where(
+        setup_at: nil,
+        training_provider: current_provider_user.providers,
+      )
+    end
+
     def success; end
 
     def edit
@@ -62,9 +69,10 @@ module ProviderInterface
     end
 
     def render_403_unless_access_permitted
-      render_403 unless current_provider_user.providers.include?(
-        training_provider_permissions.training_provider,
-      )
+      render_403 unless current_provider_user.can_manage_organisations? &&
+        current_provider_user.providers.include?(
+          training_provider_permissions.training_provider,
+        )
     end
   end
 end
--- a/app/controllers/provider_interface/provider_relationship_permissions_controller.rb
+++ b/app/controllers/provider_interface/provider_relationship_permissions_controller.rb
@@ -68,10 +68,10 @@ module ProviderInterface
     end
 
     def render_403_unless_access_permitted
-      render_403 unless current_provider_user.can_manage_organisations? &&
-        current_provider_user.providers.include?(
-          training_provider_permissions.training_provider,
-        )
+      training_provider = training_provider_permissions.training_provider
+
+      render_403 unless ProviderAuthorisation.new(actor: current_provider_user)
+        .can_manage_organisation?(provider: training_provider)
     end
   end
 end
--- a/app/controllers/provider_interface/provider_relationship_permissions_controller.rb
+++ b/app/controllers/provider_interface/provider_relationship_permissions_controller.rb
@@ -23,9 +23,8 @@ module ProviderInterface
 
     def update
       initialize_form
-      @form.assign_permissions_attributes(provider_relationship_permissions_form_params)
 
-      if @form.save!
+      if @form.update!(provider_relationship_permissions_form_params)
         redirect_to provider_interface_provider_relationship_permissions_success_path
       else
         flash[:warning] = 'Unable to save permissions, please try again. If problems persist please contact support.'
--- a/app/models/provider_user.rb
+++ b/app/models/provider_user.rb
@@ -36,10 +36,6 @@ class ProviderUser < ActiveRecord::Base
     end
   end
 
-  def can_view_safeguarding_information_for?(provider)
-    provider_permissions.view_safeguarding_information.exists?(provider: provider)
-  end
-
   def full_name
     "#{first_name} #{last_name}" if first_name.present? && last_name.present?
   end
--- a/app/models/provider_user.rb
+++ b/app/models/provider_user.rb
@@ -49,7 +49,7 @@ class ProviderUser < ActiveRecord::Base
   end
 
   def can_manage_organisations?
-    ProviderInterface::ProviderRelationshipPermissions.exists?(training_provider: providers)
+    provider_permissions.exists?(manage_organisations: true)
   end
 
 private
--- a/app/services/provider_authorisation.rb
+++ b/app/services/provider_authorisation.rb
@@ -48,6 +48,12 @@ class ProviderAuthorisation
     end
   end
 
+  def can_manage_organisation?(provider:)
+    return true if @actor.is_a?(SupportUser)
+
+    @actor.provider_permissions.exists?(provider: provider, manage_organisations: true)
+  end
+
   # automatically generates assert_can...! methods e.g. #assert_can_make_offer! for #can_make_offer?
   instance_methods.select { |m| m.match PERMISSION_METHOD_REGEXP }.each do |method|
     permission_name = method.to_s.scan(PERMISSION_METHOD_REGEXP).last.first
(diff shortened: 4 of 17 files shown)


app/controllers/provider_interface/provider_interface_controller.rb
```diff
@@ -48,6 +49,27 @@ def check_data_sharing_agreements
       end
     end
 
+    def check_provider_relationship_permissions
+      return unless FeatureFlag.active?('enforce_provider_to_provider_permissions')
+      return unless current_provider_user
+      return if performing_provider_organisation_setup?
+
+      if current_provider_user.can_manage_organisations? && provider_relationship_permissions_need_setup?
```
mnacos, line 57: `can_manage_organisations?` in its current form is not specific enough. If I am a provider user with 'manage organisations' for provider A but without 'manage organisations' for provider B, I'll still be able to set up org permissions for provider B. Maybe something like `can_manage_organisation?(provider)`?

app/models/provider_user.rb
```diff
@@ -48,6 +48,10 @@ def can_manage_users?
     provider_permissions.exists?(manage_users: true)
   end
 
+  def can_manage_organisations?
+    ProviderInterface::ProviderRelationshipPermissions.exists?(training_provider: providers)
```
mnacos, line 52: See previous comment, this is a little dangerous.

mnacos: How about adding a `can_manage_organisations?(provider:)` instead within `ProviderAuthorisation`, along the lines of `can_view_safeguarding_information?(course:)`?

Some of these methods within `ProviderUser` make sense, especially for quick global checks, such as `can_manage_users?`. If we need we a quick UI-related decisions (e.g. whether to show the Users link in the provider interface header) they are fine, but it's too easy to start relying on these in other parts of the code when we should be more specific (is this user allowed to manage users for this specific provider...).

Others should probably be removed, such as `can_view_safeguarding_information_for?(provider)`, which is no longer used, from what I can see.

spec/system/provider_interface/setup_provider_relationship_permissions_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+    when_i_click_continue
+    and_i_choose_permissions_for_the_provider_relationship
+    and_i_confirm_my_choices
+    then_i_see_permissions_were_successfully_saved
+
+    when_i_click_continue
+    then_i_can_see_the_permissions_setup_page
+
+    when_i_click_continue
+    and_i_choose_permissions_for_another_provider_relationship
```
mnacos, line 24: Maybe set up the test data so that the user only has 'Manage organisations' for one of the providers? This way we'll test the specificity of 'Manage organisations' instead of the ability to handle multiple sets of org permissions. As I understand this, this is an asymmetric feature, in that the training provider is the one deciding to give access to the ratifying provider and not the other way around.

You could always add a third provider, of course, and test both the specificity of the permission and the ability to handle multiple sets of permissions (for multiple providers).

app/models/provider_interface/provider_relationship_permissions_form.rb
```diff
@@ -9,6 +9,8 @@ def assign_permissions_attributes(params)
     end
 
     def save!
+      @accredited_body_permissions.setup_at = Time.current
+      @training_provider_permissions.setup_at = Time.current
```
mnacos, line 13: Smol feature request: could we also have a `setup_by` recording the id of the provider user who has last changed org permissions? It would be very useful for support/auditing.

Alternatively, we should be adding to the audit trail. I realise this may be beyond the scope of this PR. But we should either add the `setup_by` or create a ticket to add these actions to the audit trail.

app/models/provider_user.rb
steventux: :+1: I think this PR is a little behind how we intend to manage this permission, I've since added an explicit user-level permission for this https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/db/schema.rb#L317 so perhaps we should use this on the ProviderUser.
I agree we should be checking the orgs carefully, I've done something similar here https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2202/commits/2b08541e4931d55854a73d6aba363dd48d1f000f#diff-b4b81af4516e63df181af81c4c2c7d37R23 in a before filter for organisations pages. Will update.

app/models/provider_interface/provider_relationship_permissions_form.rb
steventux: I agree this should be using the audit trail, I was mistaken in thinking that using `save!` would do this but I need to use `update!(...)` - will amend.

steventux: @mnacos https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2231/commits/13d1cf59d5ace13af84deeb81661cffd5c81a891 calls `update!` to include changes in audit trail.

spec/system/provider_interface/setup_provider_relationship_permissions_spec.rb
steventux: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2231/commits/64852d66c3378a87206797339490eb6617697632#diff-29880b76c764d0e6550048eef5418477R49 :+1: I've removed the ratifying provider from the providers associated with the user.

app/controllers/provider_interface/provider_interface_controller.rb
steventux: This is now https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2231/commits/136e838c36d3805972582f3e73c6b4f6d08f4815#diff-4402a7a0a04b6d6e60e6fb0da05a3a8fR52 a user-level permissions check.

app/models/provider_user.rb
steventux: @mnacos https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2231/commits/26ad3e533832eb611ef31ecf73a22de4e9e6cd50 removes the `can_view_safeguarding_information_for?(provider)` method. Thanks for spotting this.
I've now started using the user-level permission `can_manage_organisations` in conjunction with a check on the training provider: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/64852d66c3378a87206797339490eb6617697632/app/controllers/provider_interface/provider_relationship_permissions_controller.rb#L71-L74

app/controllers/provider_interface/provider_interface_controller.rb
duncanjbrown: I think it's still the case that if I'm associated with A and B, A is set up but B isn't, and I can only manage organisations for A, this returns true even though I can't manage B's relationships?

spec/models/provider_user_spec.rb
```diff
@@ -78,36 +78,17 @@
...
-    it 'is false without the correct permission for the given provider' do
-      provider_user.provider_permissions.update_all(view_safeguarding_information: true)
-      expect(provider_user.can_view_safeguarding_information_for?(build_stubbed(:provider))).to be false
-    end
-
-    it 'is true with the correct permission for the given provider' do
-      provider_user.provider_permissions.update_all(view_safeguarding_information: true)
-      provider = provider_user.providers.first
-      expect(provider_user.can_view_safeguarding_information_for?(provider)).to be true
+    it 'is false for users without the manage users permission' do
```
duncanjbrown, line 84: Typo: `users` for `organisations`

app/controllers/provider_interface/provider_relationship_permissions_controller.rb
```diff
@@ -61,9 +68,10 @@ def render_404_unless_permissions_found
     end
 
     def render_403_unless_access_permitted
-      render_403 unless current_provider_user.providers.include?(
-        training_provider_permissions.training_provider,
-      )
+      render_403 unless current_provider_user.can_manage_organisations? &&
```
duncanjbrown, line 70: Will this also return true in the event that the user has manage_orgs permission for a provider that isn't the current training provider?

app/models/provider_interface/provider_relationship_permissions_form.rb
duncanjbrown: +1 to the audit trail. `save!` feels like it ought to work.

app/views/provider_interface/provider_relationship_permissions/success.html.erb
```diff
@@ -11,7 +11,7 @@
     </h2>
 
     <p class="govuk-body">
-      Select <a href="/organisations">Organisations</a> (at the top of every page) to update your permissions.
+      Select <%= govuk_link_to 'Organisations', '/provider/organisations' %> (at the top of every page) to update your permissions.
```
duncanjbrown, line 14: Should we use the path helper here?

app/models/provider_user.rb
```diff
@@ -36,10 +36,6 @@ def self.onboard!(dsi_user)
     end
   end
 
-  def can_view_safeguarding_information_for?(provider)
```
duncanjbrown, line 39: :+1: As @mnacos has suggested, let's have a review of what permissions methods we've got and where they all live once the current PRs are in

app/models/provider_interface/provider_relationship_permissions_form.rb
steventux: @duncanjbrown I had a quick look at the audited gem and it seemed to work on `create` `update!` and `destroy`, I think it stores the parameters passed to the `create` and `update!` methods and uses them as details for the trail. https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/commit/13d1cf59d5ace13af84deeb81661cffd5c81a891 modifies the form to use `update!` so we should now record these permissions being changed.

app/controllers/provider_interface/provider_relationship_permissions_controller.rb
steventux: Yes sorry as @mnacos pointed out this should be scoped to a particular provider. I'll add a method to `ProviderAuthorization` for this.

steventux: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2231/commits/f062678cfbf7db0d391234cf4d9b0367db0e0d95 and the commit before it https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2231/commits/2f8bc6a1bb687e12f8c31d364597d570e05e625a use a more specific scope to check the permissions now.

app/views/provider_interface/provider_relationship_permissions/success.html.erb
steventux: :+1: This was in the other PR - will update now it's merged.

steventux: Added and rebased here https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2231/commits/c6cdf577614afac2cc132af319f11c3392ef63fd#diff-797a45328232ed22024a3002790f4897R14

spec/models/provider_user_spec.rb
steventux: Updated here https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2231/files#diff-ee4451eba1a2bcdccf60d8f9bfea364cR84-R88

app/controllers/provider_interface/provider_interface_controller.rb
steventux: @duncanjbrown https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2231/files#diff-be572e8dffe260e5ccdf6095c4e62b25R43 adds an organisation specific check.
So we have 
`ProviderUser#can_manage_organisaions?` - A general check used for displaying navigation items etc.
`ProviderAuthorization#can_manage_organisation?(provider:)` A check that the acting user can manage a specific provider organisation.

duncanjbrown: I think this would still redirect the user if they had manage_orgs but one of the orgs they couldn't manage was a ratifying provider?

steventux: Ah yes this needs to use `ProviderAuthorisation#can_manage_organisation?(provider:)` thanks!

steventux: Ah yes I think the best way to mitigate this is to find the first `TrainingProviderPermission` with a training provider who is one of the current user's providers and `setup_at` is nil. Then pass this provider to `ProviderAuthorisation#can_manage_organisation?` which will check the user has permission specific to that org.

app/services/provider_authorisation.rb
```diff
@@ -48,6 +48,12 @@ def can_view_safeguarding_information?(course:)
     end
   end
 
+  def can_manage_organisation?(provider:)
+    return true if @actor.is_a?(SupportUser)
+
+    @actor.provider_permissions.exists?(provider: provider, manage_organisations: true)
```
mnacos, line 54: Actor may also be a VendorApiUser. This model responds to .providers but not `.provider_permissions'. Not particularly important as api keys cannot currently manage org relationships, but worth a mention.

app/controllers/provider_interface/provider_interface_controller.rb
steventux: @duncanjbrown @mnacos I've amended the `ProviderInterfaceController` filter here https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2231/files#diff-9817fde727b731ffab81417f046ad3a9R70-R91 to check for training provider permissions needing setup and then to check that the training provider on this org-level permission is a provider where the user has user-level permissions to manage organisations. Hopefully this tightens up the check properly.

duncanjbrown: Thanks - looks good to me!

PR #2232, opened 2020-06-08
Upgrade to Ruby 2.7

tijmenb:
## Context

2.7 came out some time ago, time to upgrade!

See commits.

## Guidance to review

Per commit.

## Link to Trello card

https://trello.com/c/srHI6130

## Things to check


Dockerfile
```diff
@@ -59,10 +57,12 @@ ENV RAILS_ENV=production \
...
-RUN rm -rf $BUNDLE_PATH/$GEM_PATH/cache/*.gem && \
-    find $BUNDLE_PATH/$GEM_PATH/gems -name "*.c" -delete && \
-    find $BUNDLE_PATH/$GEM_PATH/gems -name "*.h" -delete && \
-    find $BUNDLE_PATH/$GEM_PATH/gems -name "*.o" -delete
+RUN rm -rf /usr/local/bundle/cache/*.gem && \
+    find /usr/local/bundle/gems -name "*.c" -delete && \
+    find /usr/local/bundle/gems -name "*.h" -delete && \
+    find /usr/local/bundle/gems -name "*.o" -delete && \
+    find /usr/local/bundle/gems -name "*.md" -delete && \
+    find /usr/local/bundle/gems -name "*.html" -delete
```
duncanjbrown, line 65: This makes me a bit anxious as gems could conceivably use `html` files for providing functionality.

vigneshmsft: ah okie, size gain its worth the risk then, I'll not delete the html and md files then.

vigneshmsft: I could see only about 4MB gain by removing html files in the gems folder, have removed this line.

PR #2233, opened 2020-06-08
Add application reference to Vendor API responses

AgaDufrat:
## Context

One of the vendors pointed out that it would be useful to have
application_reference since it’s displayed to the candidates and providers in
the UI. Application reference is often used when candidates contact providers.
Adding it to the API will allow provider's system to display this information
and therefore to identify candidates/application by the reference.

 Return application_reference in the response. 

## Guidance to review

- GET /applications/{application_id} returns application_reference
- API docs are updated:
  - http://localhost:3000/api-docs/reference#get-applications-application_id HTTP 200 example
  - http://localhost:3000/api-docs/reference#applicationattributes-object

## Link to Trello card

https://trello.com/c/ObzwukFU/2246-add-application-reference-to-vendor-api-responses

## Things to check

### DIFF
--- a/app/presenters/vendor_api/single_application_presenter.rb
+++ b/app/presenters/vendor_api/single_application_presenter.rb
@@ -10,6 +10,7 @@ module VendorAPI
         id: application_choice.id.to_s,
         type: 'application',
         attributes: {
+          application_reference: application_form.support_reference,
           status: application_choice.status,
           phase: application_form.phase,
           updated_at: application_choice.updated_at.iso8601,
--- a/app/views/api_docs/pages/release_notes.md
+++ b/app/views/api_docs/pages/release_notes.md
@@ -1,3 +1,9 @@
+### 9th June 2020
+
+New attributes:
+
+- `ApplicationAttributes` now has an `application_reference` attribute of type string.
+
 ### 20th May 2020
 
 Corrections to documentation
--- a/config/vendor-api-v1.yml
+++ b/config/vendor-api-v1.yml
@@ -337,6 +337,7 @@ components:
       type: object
       additionalProperties: false
       required:
+      - application_reference
       - candidate
       - phase
       - contact_details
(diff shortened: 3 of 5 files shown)


app/presenters/vendor_api/single_application_presenter.rb
```diff
@@ -19,6 +19,7 @@ def as_json
           reject_by_default_at: application_choice.reject_by_default_at&.iso8601,
           candidate: {
             id: "C#{application_form.candidate.id}",
+            application_reference: application_form.support_reference,
```
duncanjbrown, line 22: I think this is the wrong place for the attribute as it's possible for the same candidate to have multiple applications, in the event that a) their first application fails and they go into "Apply again" or b) they Apply in the following cycle. One level up, in `attributes`, might be better?

config/vendor-api-v1.yml
```diff
@@ -464,6 +465,11 @@ components:
           description: The candidate’s ID in the Apply system
           maxLength: 10
           example: C5432
+        application_reference:
+          type: string
+          description: The candidate’s reference in the Apply system
```
duncanjbrown, line 470: This suggests the reference is a pointer to the candidate rather than a reference for them to use. Consider something more explicit, like "The candidate's reference number for their application in the Apply system"?

app/views/api_docs/pages/release_notes.md
```diff
@@ -1,3 +1,9 @@
+### 9th June 2020
+
+New attributes:
+
+- `ApplicationAttributes` now has a unique `application_reference` attribute of type string.
```
duncanjbrown, line 5: Sorry, one more thing! This attr isn't unique as the candidate may select two courses at the same provider, which would result in two applications coming over the API with the same application reference.

app/presenters/vendor_api/single_application_presenter.rb
```diff
@@ -10,6 +10,7 @@ def as_json
         id: application_choice.id.to_s,
         type: 'application',
         attributes: {
+          application_reference: application_form.support_reference,
```
tijmenb, line 13: Sorry to be late to this party, but can we rename this to "support reference"? It would be consistent with the field name, and avoid confusion with the `ApplicationReference` model.

AgaDufrat: sure, will do!

PR #2234, opened 2020-06-08
Show a sensible error when making an offer twice over the API

duncanjbrown:
## Context

UCB were confused when they made the same `make_offer` request twice: the first returned 200 and the second returned 404 with the message `Could not find an Application with ID...`. This isn't normal client behaviour but it's also very confusing API behaviour.

- Add stricter validation to `ChangeOffer` so we always have to specify what we want to change the offer _to_.
- Bubble its messages back to the front of the API.

Interesting things:

- the ChangeOffer flow on the front end doesn't actually check whether the `ChangeOffer` is valid — it merrily shows a green flash and goes on its way. This is weird behaviour and we shouldn't do it.
- sharing error messages between the API and front end needs a look over — I cheated a bit in the YAML here, building a sentence from the attribute name and the error message, as that's how we build errors in the API. We'll hit this when we try to fix the above.

## Link to Trello card

https://trello.com/c/pvFqo8AX/2219-api-bugs-reported-by-ucb

## Things to check

### DIFF
--- a/app/controllers/provider_interface/decisions_controller.rb
+++ b/app/controllers/provider_interface/decisions_controller.rb
@@ -18,20 +18,26 @@ module ProviderInterface
     end
 
     def new_offer
-      @course_option_id = params[:course_option_id] # from offer_changes/edit_offer
+      course_option = if params[:course_option_id]
+                        CourseOption.find(params[:course_option_id])
+                      else
+                        @application_choice.course_option
+                      end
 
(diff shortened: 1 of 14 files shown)


duncanjbrown:
It’s weird that we create errors in the controller. Going to try to push these down into validations in the service

duncanjbrown:
Closing to re-raise

PR #2240, opened 2020-06-09
Enhance HESA dataset

malcolmbaig:
## Context

The raw HESA data's formatting is all over the place. We want to tidy it up so we can use it in our degree selection screens.

A series of manual changes, see individual commits for detail.

## Guidance to review
- Changes were all applied by some combination of manual edits and recorded macros. No code has been added to deal with automating this cleanup. The data is supposedly quite static and changes infrequently. It's also quite inconsistent, meaning any automation would have to be somewhat complex. This probably isn't a worthwhile tradeoff given how infrequently we expect to make updates.
- hesa_degree_establishments have not been modified - the formatting of that data appears workable as is.
- hesa_degree_classes have also not been modified - I suspect we might remove this as the [prototype design](https://apply-beta-prototype.herokuapp.com/application/233GP/degree/237G0/grade) doesn't suggest we need the standardised HESA descriptors for grades. 

## Link to Trello card

https://trello.com/c/5Nfb93le
## Things to check


config/initializers/hesa_degree_types.rb
```diff
@@ -1,92 +1,86 @@
...
+  [98, 'B PHIL', 'Bachelor of Philosophy', :bachelor],
+
+  [200, 'MA', 'Master of Arts', :master],
+  [201, 'M LIB', 'Master of Librarianship', :master],
+  [202, 'M LITT', 'Master of Literature', :master],
+  [203, 'M MUS', 'Master of Music', :master],
+  [204, 'M PHIL', 'Master of Philosophy', :master],
+  [205, 'M SC', 'Master of Science', :master],
+  [206, 'M THEOL', 'Master of Theology', :master],
+  [207, 'Certificate of Membership of Cranfield Institute of Technology', :master],
```
tvararu, line 64: Missing second field?

```suggestion
  [207, nil, 'Certificate of Membership of Cranfield Institute of Technology', :master],
```

paulrobertlloyd:
> **Remove honours degrees from HESA degree types**
> The current designs for the degree selection workflow we intend to build with this data don't distinguish between honours and non-honours degrees. Remove entries marked 'Hons' from the dataset.

This is correct. We determine that a qualification is an honours degree if the candidate selects a class with honours (which you can only get if you complete the third year of your degree). So, if we wanted to submit the structured data to HESA for these 7 particular degrees (codes `2`, `4`, `6`, `8`, `10`, `13` and `14`) we’d need to combine the answer for degree type and the grade in these instances.

> **Restructure HESA degree types data**
> For each entry:
> - extract the short description (eg - 'BSc') where available

Some of the `:short_description:`s use uncommon capitalisation and spacing, i.e. `M CHEM` instead of `MChem`, `B ENG` instead of `BEng`. We’ll probably want to review these before we launch this feature.

But otherwise, looks 👌

config/initializers/hesa_degree_types.rb
```diff
@@ -1,92 +1,86 @@
...
+  [76, 'B MED SC', 'Bachelor of Medical Science', :bachelor],
+  [77, 'BM', 'Bachelor of Medicine', :bachelor],
+  [78, 'B MET', 'Bachelor of Metallurgy', :bachelor],
+  [79, 'B MET/EN', 'Bachelor of Metallurgy and Engineering', :bachelor],
+  [80, 'B MU', 'Bachelor of Music', :bachelor],
+  [81, 'BN', 'Bachelor of Nursing', :bachelor],
+  [82, 'B PHARM', 'Bachelor of Pharmacy', :bachelor],
+  [83, 'B SC', 'Bachelor of Science', :bachelor],
+  [84, 'BSC/ECO', 'Bachelor of Science Economics', :bachelor],
+  [85, 'BSC/ENG', 'Bachelor of Science &amp; Engineering', :bachelor],
```
duncanjbrown, line 42: Do we need to escape ampersands here? (we're quite far from the view)

config/initializers/hesa_degree_types.rb
```diff
@@ -1,92 +1,86 @@
...
-  [402, "Degree equivalent (inc. foreign qualifications)"],
-  [999, "Unknown"],
+  [1, nil, 'BEd', :bachelor],
+  [3, nil, 'BSc/Education', :bachelor],
+  [5, nil, 'BTech/Education', :bachelor],
+  [7, nil, 'BA/Education', :bachelor],
+  [9, nil, 'BA Combined Studies/Education of the Deaf', :bachelor],
+  [12, nil, 'BA with intercalated PGCE', :bachelor],
+  [51, 'BA', 'Bachelor of Arts', :bachelor],
+  [52, 'BA ECON', 'Bachelor of Arts Economics', :bachelor],
```
duncanjbrown, line 9: The structs in the prototype have the traditional casing (`BAEcon` rather than `BA ECON`). Should we use prototype-style `BAEcon`?

config/initializers/hesa_degree_types.rb
malcolmbaig: Good spot, thanks :+1:

config/initializers/hesa_degree_types.rb
malcolmbaig: We don't, fixed!

config/initializers/hesa_degree_types.rb
duncanjbrown: Oops, I missed @paulrobertlloyd's comment when I reviewed this by moonlight 🙄

PR #2242, opened 2020-06-09
Send email to candidate when course fills while awaiting references

stevehook:
## Context

Candidates should be notified when courses become full while they are waiting for references.

- [x] Feature flag `unavailable_course_notifications` - the worker still runs without this flag being set but posts a message to slack rather than sending the email.
- [X] New `CandidateMailer#course_unavailable_notification` method
- [X] New `SendCourseFullNotificationsWorker` worker to poll database for application choices that are awaiting references and have become full, and invoke mailer
- [x] Configure clock to call new worker
- [x] Mail previews
- [x] Use `ChaserSent` to record whether an email has already been sent for a particular application choice.
- [x] Refactor to extract the logic for determining the 'reason' for a course being unavailable into something new.

There are several distinct reasons for a course being unavailable:
- [X] Course full
- [X] Location full (but other locations for the same course available)
- [X] Study mode full (but other study modes available for the same course/location)
- [x] Course withdrawn

## Guidance to review

- Do the reasons make sense?
- Are the tests adequate?

## Link to Trello card

https://trello.com/c/qX7OyNDH/1647-dev-notification-emails-when-a-course-becomes-full-while-waiting-for-references

## Things to check

### DIFF
--- a/app/workers/send_course_full_notifications_worker.rb
+++ b/app/workers/send_course_full_notifications_worker.rb
@@ -6,6 +6,10 @@ class SendCourseFullNotificationsWorker
 
     GetApplicationChoicesWithNewlyUnavailableCourses.call.each do |application_choice|
       reason = reason_course_is_unavailable(application_choice)
+      ChaserSent.create!(
+        chased: application_choice,
+        chaser_type: :course_unavailable_notification,
+      )
       CandidateMailer.course_unavailable_notification(application_choice, reason).deliver_later
     end
   end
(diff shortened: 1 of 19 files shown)


app/workers/send_course_full_notifications_worker.rb
```diff
@@ -0,0 +1,34 @@
+class SendCourseFullNotificationsWorker
+  include Sidekiq::Worker
+
+  def perform
+    GetApplicationChoicesWithNewlyUnavailableCourses.call.each do |application_choice|
+      reason = ReasonCourseNotAvailable.new(application_choice).call
+      if FeatureFlag.active?(:unavailable_course_notifications)
+        ChaserSent.create!(
```
tijmenb, line 8: Should this be outside of the conditional? Or we'll end up with constantly repeating slack messages, since they're never marked as sent?

PR #2244, opened 2020-06-10
Allow candidate to select a course choice to replace

davidgisbey:
## Context

If a user has more than 1 course choice that has become full or withdrawn, when they click on the update course choice banner, they should be able to select a course choice to replace/make a decision on.

- Adds the ReplaceChoiceController
- Adds the pick_course_to_replace page

## Guidance to review

designs - https://docs.google.com/document/d/1FwPaGSYzSViXtbB2INHwO7bdxr0fFFUpVs4z978tW9Q/edit.

If anyone has any thoughts on the naming that would be great!

## Link to Trello card

https://trello.com/c/kUjDNNbH/1646-dev-update-course-choices-if-course-becomes-full-while-waiting-for-references

## Things to check

### DIFF
--- a/app/models/candidate_interface/pick_choice_to_replace_form.rb
+++ b/app/models/candidate_interface/pick_choice_to_replace_form.rb
@@ -1,0 +1,9 @@
+module CandidateInterface
+  class PickChoiceToReplaceForm
+    include ActiveModel::Model
+
+    attr_accessor :id
+
+    validates :id, presence: true
+  end
+end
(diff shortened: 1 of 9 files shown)


app/models/candidate_interface/pick_choice_to_replace_form.rb
```diff
@@ -0,0 +1,9 @@
+module CandidateInterface
+  class PickChoiceToReplaceForm
+    include ActiveModel::Model
+
+    attr_accessor :id
+
+    validates :id, presence: true
```
tvararu, line 7: We should keep an eye on this as the feature progresses. Currently we only check that an ID is present, but we don't check if that ID is of a course choice that actually belongs to the candidate's application form. Could open us up to a security hole, but it's too early to say if it will be exploitable once everything else is fleshed out.

davidgisbey: Yea that's a really good point 👍

PR #2248, opened 2020-06-10
Enforce make_decisions user permission if feature flag is on

mnacos:
### Context

Only provider users who have the make_decisions permission should be able to make decisions on applications. A "decision" for this purpose means an offer, a rejection or a change to an existing offer. Provider users who can't make decisions should not see UI relating to any of these operations. They should also be forbidden from carrying out the operations via a crafted request (ie the service layer should enforce the permissions too).

The `:provider_make_decisions_restriction` flag is meant to stop provider users from making decisions, unless they have the `make_decisions` user-level permission. This is a big departure from the current logic, which assumes you can respond to any application that is visible to you (visibility is a function of user/provider/course associations).

The prompt to respond to a 'submitted' application is hidden if the provider user lacks `make_decisions`:

__Before__:

image

__After__:

image

'Change' links are hidden for users without `make_decisions` viewing applications in `offer` state.

__Before__:
image

__After__:
image

Besides, if a user without the relevant permission tries to access a url that requires `make_decisions` (e.g. if a colleague emails them a link or they manually enter the url), they will see a page explaining why the action is unavailable:

image
I think a 403 is better than a 404 in this case. Also, being descriptive about which permission is missing should help with troubleshooting and support.

The main `make_decisions` checks have been added to `ProviderAuthorisation`, which means users cannot alter the frontend or craft requests to bypass this permission requirement.

This PR does not affect the ability of vendors to make decisions on applications, as API keys remain omnipotent (within the realm of a single provider).

Users of ratifying providers retain their ability to make or change offers, as long they have the `make_decisions` permission for the provider ratifying the course in question. I have removed `can_change_offer?` in favour of `can_make_offer?`, as they share the same logic and restrictions. 

For releasing this to production, the current plan is:
- Merge and release this code, keeping the feature flag off.
- Prepare a database migration to grant `make_decisions` to all existing provider users, optionally turning the feature flag on at the same time (feature flags are postgres-backed now).
- Monitor the feature and remove the feature flag from the codebase in due course.

### Guidance to review

Run the code with the feature flag on and off, using provider users with and without the `make_decisions` permission.

If running the tests locally, run the tests with and without this line: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2248/files#diff-be572e8dffe260e5ccdf6095c4e62b25R24

The tests pass without this line, which suggests we could remove the feature flag (which equates to having the feature flag on permanently) without breaking anything.

If you are using the review environment, I've removed `make_decisions` from `dev-provider` but preserved it for `dev-support`. Please log in as `dev-provider` or `dev-support`, depending on which scenario you are testing.

### Link to Trello card

[Provider users have to have make_decisions permission before they can make decision](https://trello.com/c/CwAkdmkP)

### Things to check


app/services/provider_authorisation.rb
```diff
@@ -71,4 +87,9 @@ def training_provider_can_view_safeguarding_information?(course:)
         .view_safeguarding_information
         .exists?(ratifying_provider: course.accredited_provider, training_provider: course.provider)
   end
+
+  def actor_has_permission_to_make_decisions?(providers:)
+    @actor.is_a?(VendorApiUser) ||
+      @actor.provider_permissions.make_decisions.exists?(provider: providers)
```
duncanjbrown, line 93: I think we should assume all potential non-API `@actors` will implement `can_make_decisions_for?` and call that

app/services/provider_authorisation.rb
```diff
@@ -9,19 +9,35 @@ def can_make_offer?(application_choice:, course_option_id: nil)
...
 
   def can_change_offer?(application_choice:, course_option_id:)
-    new_course_option = CourseOption.find course_option_id
-    @actor.is_a?(SupportUser) || \
-      application_choice_visible_to_user?(application_choice: application_choice) && \
-        course_option_belongs_to_user_providers?(course_option: new_course_option)
+    can_make_offer?(
+      application_choice: application_choice,
+      course_option_id: course_option_id,
+    )
```
duncanjbrown, line 40: Should we lose this method altogether and use `can_make_offer?` everywhere? Not 100% sure but it feels like it might be less to think about.

app/controllers/provider_interface/decisions_controller.rb
```diff
@@ -127,6 +128,20 @@ def requires_provider_change_response_feature_flag
       render_404 unless FeatureFlag.active?('provider_change_response')
     end
 
+    def requires_provider_user_make_decisions_permission
+      provider = @application_choice.offered_course.provider
+
+      if FeatureFlag.active?('provider_make_decisions_restriction') &&
+          !current_provider_user.can_make_decisions_for?(provider)
+
+        redirect_to provider_interface_missing_permission_path(
```
duncanjbrown, line 137: Should we redirect or should we raise, rescue and render? I think I prefer raise and render as we don't have an addressable error page, and the error is served with the correct status code.

app/views/provider_interface/provider_user_permissions/missing_permission.html.erb
```diff
@@ -0,0 +1,13 @@
+<%= content_for :title, "Requires ‘#{@permission}’" %>
```
duncanjbrown, line 1: Page looks good!

spec/support/test_helpers/provider_user_permissions_helper.rb
```diff
@@ -0,0 +1,17 @@
+module ProviderUserPermissionsHelper
+  def permit_make_decisions(dfe_sign_in_uid: 'DFE_SIGN_IN_UID', provider: nil)
```
duncanjbrown, line 2: Nice!
- should it have a bang, as we're side-effecting?
- would be nice to include see_safeguarding_information here? Maybe a follow up refactor.

spec/support/test_helpers/provider_user_permissions_helper.rb
```diff
@@ -0,0 +1,17 @@
...
+    permissions = if provider
+                    provider_user.provider_permissions.where(provider: provider)
+                  else
+                    provider_user.provider_permissions
+                  end
+
+    permissions.update_all(make_decisions: true)
+  end
+
+  alias_method :and_i_am_permitted_to_make_decisions_for_my_provider, :permit_make_decisions
```
duncanjbrown, line 15: The rule of thumb for helpers is to share implementations, not steps, so I think the project-standard way to do this would be to implement `and_i_am_permitted_to...` in each spec that uses it, and call out to `permit_make_decisions` each time ([Rule 3](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/master/docs/testing-styleguide.md#rules))

app/services/provider_authorisation.rb
```diff
@@ -9,19 +9,35 @@ def can_make_offer?(application_choice:, course_option_id: nil)
     return true if @actor.is_a?(SupportUser)
 
     supplied_course_option = CourseOption.find(course_option_id) if course_option_id
```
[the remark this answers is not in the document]
mnacos: Yes, this branch needs a rebase and some tweaks now.

app/services/provider_authorisation.rb
mnacos: Is this a suggestion to attach `can_make_decisions_for?` methods to `ProviderUser`, 'VendorApiUser` and `SupportUser`? I think I prefer keeping all authorisation logic in this service.

See also https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2231#discussion_r439458862

mnacos: Although we could add a nicer interface for the query to the actor classes, instead of accessing the one-to-many directly.

app/services/provider_authorisation.rb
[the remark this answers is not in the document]
mnacos: Rebased

app/services/provider_authorisation.rb
mnacos: I've added a `provider.users_with_make_decisions` and removed `can_make_decisions_for?` in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2248/commits/288d8c2b50a0256627481db1e3020735910074e5

app/controllers/provider_interface/decisions_controller.rb
```diff
@@ -127,6 +128,20 @@ def requires_provider_change_response_feature_flag
       render_404 unless FeatureFlag.active?('provider_change_response')
     end
 
+    def requires_provider_user_make_decisions_permission
+      provider = @application_choice.offered_course.provider
```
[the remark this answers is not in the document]
mnacos: Considering application handover to another provider is a valid use case, I would suggest preserving `offered_option`, which has been the behaviour so far. But I agree we need to be clear on this and we'll need some corresponding tests, perhaps added as another PR.

app/controllers/provider_interface/provider_interface_controller.rb
```diff
@@ -14,6 +14,13 @@ class ProviderInterfaceController < ActionController::Base
       render template: 'provider_interface/account_creation_in_progress', status: :forbidden
     }
 
+    rescue_from MissingPermission, with: lambda { |e|
+      Raven.capture_exception(e)
```
duncanjbrown, line 18: 👌

app/controllers/provider_interface/provider_interface_controller.rb
```diff
@@ -14,6 +14,13 @@ class ProviderInterfaceController < ActionController::Base
       render template: 'provider_interface/account_creation_in_progress', status: :forbidden
     }
 
+    rescue_from MissingPermission, with: lambda { |e|
```
duncanjbrown, line 17: Should we have a test for this behaviour? 

I think you could rescue with a method called `missing_permission` here and it would know which view to use? Would be good not to need a separate controller.

app/lib/missing_permission.rb
```diff
@@ -0,0 +1,9 @@
+class MissingPermission < RuntimeError
+  attr_reader :permission, :provider, :provider_user
+
+  def initialize(permission:, provider:, provider_user:)
```
duncanjbrown, line 4: It's against Liskov (and feels a bit weird) to change the constructor of `RuntimeError` like this. I think it would be better to take a message in the first position as normal and not use keyword args (hash args we could splat, maybe?)

app/controllers/provider_interface/provider_user_permissions_controller.rb
```diff
@@ -0,0 +1,9 @@
+module ProviderInterface
```
duncanjbrown, line 1: Is this used anywhere?

app/lib/missing_permission.rb
```diff
@@ -0,0 +1,9 @@
+class MissingPermission < RuntimeError
```
duncanjbrown, line 1: Could this be namespaced in the `ProviderInterface`?

app/controllers/provider_interface/provider_interface_controller.rb
mnacos: I've added a 'respond to an application' scenario covering the case of hand-crafting urls exercising this path. I've also removed the separate directory and started rescuing with a named method, but specifying the template is still needed.

app/controllers/provider_interface/provider_user_permissions_controller.rb
mnacos: Removed

app/lib/missing_permission.rb
mnacos: Yep but not in `app/lib`.

app/lib/missing_permission.rb
mnacos: The intention there is to use a custom error class and require this information, otherwise we don't have enough information for an informative 'access denied' page. But I see your point.

I've changed the error to subclass `StandardError` and accept a hash with the required bits. I've also switched callers to `raise <error instance>, message` syntax.

PR #2252, opened 2020-06-11
Migration to add application_qualifications.start_year

stevehook:
## Context

As part of the structured data for degree qualifications we need to capture the start year for degree qualifications.

Just a migration to add the `start_year` column to the `application_qualifications` table.

## Guidance to review

- Is the name `start_year` OK?
- There is no backfill for this column. It's not really possible to work out the start date for degree qualifications accurately and the attribute is optional.

## Link to Trello card

https://trello.com/c/eqwtUVGx/1662-dev-add-the-year-candidates-started-their-degree-for-when-did-you-study-for-your-degree

## Things to check


davidgisbey:
start_year or year_started both are good depending on what tense we would prefer 👌

PR #2253, opened 2020-06-11
Remove the safeguarding question from the referee questionnaire

davidgisbey:
## Context

Gemma was getting an error when trying to export the referee survey results.

A record was manually updated due a 500. One question's response was true when it should have been 'true'. Due to this, when calling .split triggered an error. 

While solving the above, i realised that at some point, the referee reference flow has changed. 

A question on safeguarding has been asked on a separate page.

This means that it does the additional safeguarding question on the questionnaire should be removed.

- Remove the safeguarding question from the referee questionnaire

## Things to check

### DIFF
--- a/app/controllers/referee_interface/questionnaire_form.rb
+++ b/app/controllers/referee_interface/questionnaire_form.rb
@@ -6,15 +6,13 @@ module RefereeInterface
                   :experience_explanation_ok, :experience_explanation_good, :experience_explanation_very_good,
                   :guidance_rating, :guidance_explanation_very_poor,
                   :guidance_explanation_poor, :guidance_explanation_ok, :guidance_explanation_good,
-                  :guidance_explanation_very_good, :safe_to_work_with_children,
-                  :safe_to_work_with_children_explanation, :consent_to_be_contacted,
+                  :guidance_explanation_very_good, :consent_to_be_contacted,
                   :consent_to_be_contacted_details
 
     def save(reference)
       questionnaire = {
         'Please rate your experience of giving a reference' => "#{experience_rating} | #{experience_explanation}",
         'Please rate how useful our guidance was' => "#{guidance_rating} | #{guidance_explanation}",
-        'If we asked whether a candidate was safe to work with children, would you feel able to answer?' => safe_to_work_with_children_response,
         'Can we contact you about your experience of giving a reference?' => consent_to_be_contacted_response,
       }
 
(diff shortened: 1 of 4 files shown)


app/services/support_interface/referee_survey_export.rb
```diff
@@ -17,8 +17,8 @@ def call
           'Experience explanation' => extract_explanation(reference, 'Please rate your experience of giving a reference'),
           'Consent to be contacted' => extract_rating(reference, 'Can we contact you about your experience of giving a reference?'),
           'Contact details' => extract_explanation(reference, 'Can we contact you about your experience of giving a reference?'),
-          'Safe to work with children?' => extract_rating(reference, 'If we asked whether a candidate was safe to work with children, would you feel able to answer?'),
-          'Safe to work with children explanation' => extract_explanation(reference, 'If we asked whether a candidate was safe to work with children, would you feel able to answer?'),
+          'Safe to work with children?' => reference.safeguarding_concerns_status,
+          'Safe to work with children explanation' => reference.safeguarding_concerns,
         }
```
tijmenb, line 22: I don't have much context, but I'm a little confused by this. Why are we adding this information to the export?

davidgisbey: It doesn't have to change, but we ask a referee whether there are safeguarding issues twice. It is now on a mandatory page with it's own radio button, whereas previously it was simply a question on the questionnaire.

When the new page was added, the question was not removed from the questionnaire.

tijmenb: Aha - I think we're conflating two things here. The questionnaire asks "If we asked whether a candidate was safe to work with children, would you feel able to answer?". We now have a question in the form, which is "Do you know of any reason why #{@application.full_name} should not work with children?". Those are not the same. Does that make sense?

Now that's cleared up, I don't entirely understand why https://sentry.io/organizations/dfe-bat/issues/1721431101 happened. It seems to occur when the `If we asked whether a candidate was safe to work with children, would you feel able to answer?` question is `nil` in the JSON object. Any idea why that could have happened?

tvararu: My understanding is that it was refactored here but the survey export was not updated: https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2197

stevehook: This makes sense to me. The only bit I'm unclear about is how the boolean values got into the questionnaire data?

tijmenb:
Before merging, can you update the PR title & description to explain what we found yesterday?

PR #2260, opened 2020-06-12
Make pagination align with application cards

willmcb:
## Context

Align pagination with application cards so that app matches the design.

**Before:**
Screenshot 2020-06-12 at 09 29 38

**After:**
Screenshot 2020-06-12 at 09 27 45

## Guidance to review

- is this correct from a design perspective. Note that if there are a lot of applications then the pagination will fold up into two lines:

Screenshot 2020-06-12 at 09 28 43

## Link to Trello card

https://trello.com/c/FKkvBXXs/2117-title-pagination-should-be-aligned-with-the-list-within-the-right-hand-column

## Things to check

### DIFF
--- a/app/frontend/styles/_filter.scss
+++ b/app/frontend/styles/_filter.scss
@@ -54,6 +54,10 @@
   box-shadow: none;
 }
 
+.moj-pagination__list {
+  padding-left: 3px;
+}
+
 .app-filter {
   @include govuk-media-query($from: tablet) {
     margin-bottom: govuk-spacing(3);
--- a/app/components/header_component.html.erb
+++ b/app/components/header_component.html.erb
@@ -17,7 +17,7 @@
       <%= link_to service_name, service_url, class: 'govuk-header__link govuk-header__link--service-name' %>
       <% if navigation_items.any? %>
         <button type="button" class="govuk-header__menu-button govuk-js-header-toggle app-!-print-display-none" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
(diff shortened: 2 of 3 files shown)


app/frontend/styles/_filter.scss
```diff
@@ -30,6 +30,10 @@
   box-shadow: none;
 }
 
+.moj-pagination__list {
+  padding-left: 3px;
```
adamsilver, line 34: Why is this needed? (It's not on the prototype.)

app/frontend/styles/_filter.scss
```diff
@@ -30,6 +30,10 @@
...
+}
+
 .app-filter {
   display: none;
```
adamsilver, line 38: I know that this isn't part of the review but this means without JS the user won't be able to see filters on small screens which is inaccessible. See prototype code.

app/frontend/styles/_filter.scss
willmcb: For some reason without this it is the chevrons don't fit correctly.

willmcb:
> Works nicely for me locally. I wondered why the padding was needed also. Removing didn't seem to break the layout for me.

Odd it looks like this for me: 
Screenshot 2020-06-12 at 11 45 01

adamsilver:
The padding is unfortunately needed for now because when you're on page 2 the arrow is partially cut off by the filter layout divs which have overflow hidden.

Ideally the layout needs to change to avoid the overflow hidden technique.

For now I’d say merge with the padding

PR #2262, opened 2020-06-12
Fix menu collapse on small screens

willmcb:
## Context

The menu button wasn't working on small screens. It seems to be that the `govuk-header__navigation-end` class does not play well with the collapsable menu. I have replaced this with the standard `govuk-header__navigation` class and added some CSS to get the correct alignment.

**Before:**
Screenshot 2020-06-12 at 10 54 55

The menu button had no effect.

**After:**
Before menu is clicked:
Screenshot 2020-06-12 at 10 56 08

After menu is clicked:
Screenshot 2020-06-12 at 11 00 32

## Guidance to review
- Does this work? Is there a more elegant way to do the CSS for this?

## Link to Trello card

https://trello.com/c/Dr0rpyTf/2262-redundant-menu-item-in-header-on-small-screens

## Things to check

### DIFF
--- a/app/components/header_component.html.erb
+++ b/app/components/header_component.html.erb
@@ -17,7 +17,7 @@
       <%= link_to service_name, service_url, class: 'govuk-header__link govuk-header__link--service-name' %>
       <% if navigation_items.any? %>
         <button type="button" class="govuk-header__menu-button govuk-js-header-toggle app-!-print-display-none" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
-        <ul id="navigation" class="govuk-header__navigation govuk-header__navigation--end app-!-print-display-none" aria-label="Top Level Navigation">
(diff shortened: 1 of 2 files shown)


app/frontend/styles/_provider_header.scss
```diff
@@ -5,7 +5,13 @@
 }
 
 .provider-header__content {
+  .govuk-header__navigation { float: right }
```
adamsilver, line 8: Not sure u need this

app/frontend/styles/_provider_header.scss
```diff
@@ -5,7 +5,13 @@
...
+
   @include govuk-media-query($from: tablet) {
     width: 40%;
   }
+
+  @include govuk-media-query($until: desktop) {
+    .govuk-header__navigation { float: left }
```
adamsilver, line 15: Or this

willmcb: Without this it floats left on small screens - if this is okay we don't need this CSS:

Screenshot 2020-06-12 at 11 37 35

PR #2265, opened 2020-06-12
Fix mailer previews

tvararu:
## Context

Some mailer previews are broken.

Stub out some more calls so that we don't hit the database and so on.

## Guidance to review

Makes sense?

## Link to Trello card

https://trello.com/c/gsCe5tH7/1686-fix-mailer-previews-⛏

## Things to check

### DIFF
--- a/spec/mailers/previews/candidate_mailer_preview.rb
+++ b/spec/mailers/previews/candidate_mailer_preview.rb
@@ -82,7 +82,6 @@ class CandidateMailerPreview < ActionMailer::Preview
   end
 
   def new_offer_single_offer
-    course_option = FactoryBot.build_stubbed(:course_option)
     application_choice = application_form.application_choices.build(
       course_option: course_option,
       status: :offer,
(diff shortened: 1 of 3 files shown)


tvararu:
@tijmenb The way I made tests fail is like this:

```diff
diff --git a/spec/mailers/mailer_previews_spec.rb b/spec/mailers/mailer_previews_spec.rb
index 5e360d266..1bd731923 100644
--- a/spec/mailers/mailer_previews_spec.rb
+++ b/spec/mailers/mailer_previews_spec.rb
@@ -5,7 +5,9 @@ RSpec.describe 'Mailer previews' do
     describe preview do
       preview.emails.each do |email|
         it email do
-          expect { preview.call(email) }.not_to raise_error
+          100.times do
+            expect { preview.call(email) }.not_to raise_error
+          end
         end
       end
     end
```

But that's slow. I don't think we should have that in the codebase. Maybe worth documenting or adding it as a conditional way to run the tests?

tijmenb:
Right, makes sense. No need to add that test!

PR #2267, opened 2020-06-12
optimise Dockerfile

vigneshmsft:
remove build and dev tools/packages after gem installation.
remove intermediate dev layer.
image size reduced to 325MB from 420MB.

## Context
Optimize Dockerfile, remove intermediate layer which adds little value.

Uninstall dev tools and packages after `bundle install`
Remove intermediate `dev-build` layer.
New dockre image size is 325MB

## Link to Trello card

https://trello.com/c/ZDYPJ62v/2274-reduce-docker-layers-and-image-size

## Things to check


Dockerfile
```diff
@@ -1,101 +1,66 @@
...
+    gem update --system && \
+    find / -wholename '*default/bundler-*.gemspec' -delete && \
+    rm -rf /usr/local/bin/bundle && \
+    gem install bundler -v 2.1.4 && \
+    bundle install --no-binstubs --jobs=3 --retry=5 && \
+    yarn install --check-files && \
+    rm -rf yarn.lock && \
+    bundle exec rake assets:precompile && \
+    apk del .gem-installdeps && \
+    rm -rf tmp log app/assets lib/assets vendor/assets node_modules && \
```
duncanjbrown, line 34: None of these files except `node_modules` exist in our app. We could probably remove `app/frontend` but it's only 160K.

Dockerfile
```diff
@@ -1,101 +1,66 @@
...
-
-#### Prod minify stage builds the app on the common build env image ####
-
-FROM common-build-env AS prod-minify
-
-ARG APP_HOME=/app
-
 # These variables are required for running Rails processes like assets:precompile
-ENV RAILS_ENV=production \
+ENV BUNDLE_WITHOUT=${bundleWithout} \
```
duncanjbrown, line 9: Do we use `BUNDLE_WITHOUT` anywhere?

spec/azure/configuration_spec.rb
```diff
@@ -6,16 +6,4 @@
       JSON.parse(File.read('azure/template.json'))
     end
   end
-
-  describe 'azure-pipelines.yml' do
```
duncanjbrown, line 10: I'm not sure the space saving from deleting our pipeline files from the image is worth the cost of risking broken pipelines in the future?

Dockerfile
vigneshmsft: the tmp and lib folders where created in my local dev env during the build

Dockerfile
vigneshmsft: bundler uses `BUNDLE_WITHOUT` as an alternative to `--without` switch
https://bundler.io/v1.16/bundle_config.html#CONFIGURATION-KEYS

Dockerfile
duncanjbrown: Sorry, yes — I missed those!

spec/azure/configuration_spec.rb
vigneshmsft: this was not for space saving, feel this test is redundant and the tests run inside the container and the container wouldn't have been built if the yaml was not proper, further the file could still be a valid yaml but divert away from the pipeline schema.

Dockerfile
vigneshmsft: removed `app/assets`, `lib/assets`, `vendor/assets` from the remove list and added `app/frontend` to be deleted

Dockerfile
duncanjbrown: Nice! TIL

spec/azure/configuration_spec.rb
duncanjbrown: We do sometimes change the files and run specs outside the container, eg when we're adding ENV vars, so I think this still has some value. Separately, it might be interesting to run what we've got through https://github.com/microsoft/azure-pipelines-vscode/blob/master/service-schema.json.

vigneshmsft: not ignoring yml files and adding valid YAML test for all pipeline files.

PR #2274, opened 2020-06-15
Add conditional logic for radio buttons and content n the replace course choice flow

davidgisbey:
## Context

When a candidate wants to replace a course choice. They are presented with 5 options.

1. Replace the location
2. replace the study mode
3. replace the course
4. remove the course 
5. keep the course

Based on whether other locations and study modes are available, the replace location and study mode radio buttons should not be available.

There is also different content based on the above, in addition to if a course has been withdrawn.

- Add in conditional logic to render the change location/study mode radio buttons.
- Add in conditional logic for content.

When the course has been withdrawn

image

When a course is full (no other locations with vacancies and no other study mode available)

image

When a course has other locations available, but not another study mode 

image

When a course has another study mode available, but no other location 

image

When a course has other locations and another study mode available

image

## Guidance to review

Designs - https://docs.google.com/document/d/1FwPaGSYzSViXtbB2INHwO7bdxr0fFFUpVs4z978tW9Q/edit

probably the best way to review this is to change state via the console and check its behaving as expected.

## Link to Trello card

https://trello.com/b/aRIgjf0y/candidate-team-board?menu=filter&filter=label:Dev

## Things to check


app/controllers/candidate_interface/course_choices/replace_choices/decision_controller.rb
```diff
@@ -9,6 +9,9 @@ def choose_action
           @course_name_and_code = @course_choice.course.name_and_code
           @provider_name = @course_choice.provider.name
           @site_name = @course_choice.site.name
+          @study_mode = @course_choice.course_option.study_mode.humanize.downcase
+          @other_locations_available = create_pick_site_form(@course_choice, @course_choice.course_option).available_sites.present?
+          @other_study_mode = @course_choice.course_option.get_alternative_study_mode
```
stevehook, line 14: This looks good and works for me locally (when I tried manually changing the availability in my database) but there is one (edge) case that isn't covered. When the course I applied to no longer has vacancies but another location has vacancies with a different study mode I am just told that the course is full. I'm not sure this covered in the design but it doesn't seem right to say that the whole course is just full.

stevehook, line 14: This looks good and works for me locally (when I tried manually changing the availability in my database) but there is one (edge) case that isn't covered. When the course I applied to no longer has vacancies but another location has vacancies with a different study mode I am just told that the course is full. I'm not sure this covered in the design but it doesn't seem right to say that the whole course is just full.

davidgisbey: Yea that is a really good point. Maybe @paulrobertlloyd has some thoughts on it. I'm going to see if can do a quick run-through of what we've got prior to release with him and Emma if they've got time. I'll bring it up then 👍

spec/system/candidate_interface/candidate_is_only_presented_with_valid_radio_buttons_when_replacing_a_course_choice_spec.rb
```diff
@@ -0,0 +1,161 @@
...
+    when_i_arrive_at_my_application_dashboard
+    and_i_click_update_my_course_choice
+    then_i_arrive_on_the_replace_course_choice_page
+    and_i_should_be_told_that_the_course_has_been_withdrawn
+    and_i_should_have_the_option_to_replace_my_course_choice
+    and_i_should_have_the_option_to_delete_my_course_choice
+    and_i_should_have_the_option_to_keep_my_course_choice
+    and_i_should_not_have_the_option_to_change_the_location_of_my_course_choice
+    and_i_should_not_have_the_option_to_replace_the_study_mode_course_choice
+  end
```
tvararu, line 67: Would it have been easier to test this using a component and multiple unit tests? This seems like quite a lot of effort!

app/controllers/candidate_interface/course_choices/replace_choices/decision_controller.rb
tvararu: Let me know what you get out of this 👍

spec/system/candidate_interface/candidate_is_only_presented_with_valid_radio_buttons_when_replacing_a_course_choice_spec.rb
davidgisbey: It's largely just chaining the same methods over and over uselessly tbh. But yea i think I'm going to put it into a component anyway actually. Just so i can pull all that logic out of the controller if nothing else.

app/controllers/candidate_interface/course_choices/replace_choices/decision_controller.rb
davidgisbey: @tvararu we discussed it. There are going to be a few content changes. 

1. The submit application anyway content will change
2. The contact support page will change
3. A bit more will be added to the withdraw application page

PR #2276, opened 2020-06-15
Add replace study mode and keep course flow 

davidgisbey:
## Context

Adds the flow for replacing study mode. 
Adds the flow for keeping a course.

No UI changes. They just click change to x and the controller uses the existing confirmation page.

## Link to Trello card

https://trello.com/c/kUjDNNbH/1646-dev-update-course-choices-if-course-becomes-full-while-waiting-for-references

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/course_choices/replace_choices/decision_controller.rb
+++ b/app/controllers/candidate_interface/course_choices/replace_choices/decision_controller.rb
@@ -16,6 +16,12 @@ module CandidateInterface
             replacement_course_option_id = @course_choice.course_option.get_alternative_study_mode.id
 
             redirect_to candidate_interface_confirm_replacement_course_choice_path(@course_choice.id, replacement_course_option_id)
+          elsif @pick_replacement_action.replacement_action == 'keep_choice'
+            if only_one_course_choice_needs_replacing?
+              redirect_to candidate_interface_application_complete_path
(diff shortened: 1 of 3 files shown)


app/controllers/candidate_interface/course_choices/replace_choices/decision_controller.rb
```diff
@@ -20,6 +20,16 @@ def route_action
...
+          elsif @pick_replacement_action.valid? && @pick_replacement_action.replacement_action == 'replace_study_mode'
+            replacement_course_option_id = @course_choice.course_option.get_alternative_study_mode.id
+
+            redirect_to candidate_interface_confirm_replacement_course_choice_path(@course_choice.id, replacement_course_option_id)
+          elsif @pick_replacement_action.replacement_action == 'keep_choice'
+            if only_one_course_choice_needs_replacing?
+              redirect_to candidate_interface_application_complete_path
+            else
+              redirect_to candidate_interface_replace_course_choices_path
+            end
```
davidgisbey, line 32: Should there be a flash here? Nothing is actually being updated, but it might make sense to show one anyway.

tvararu: I think it's OK not to 👌

PR #2279, opened 2020-06-15
Add a DraftDegree model

malcolmbaig:
## Context
The degree flow is being reworked to split out the type, subject, and
institution sections into separate screens. In order to support this, we
need a mechanism to store form values as we move from one screen to the
other. The existing flow captures these three fields in one form, and
then has separate steps for the degree year and grade. It persists an
ApplicationQualification in order to hold state through the entire
process.

Rather than build on the existing use of ApplicationQualification
records, this commit adds a DraftDegree model and corresponding table.
Each application_form has (at most) one draft_degree, which will be
converted into an application_qualification when the candidate completes
the last step.

This approach has the following benefits:

- It creates some separation from the ApplicationQualification records
  by providing an ephemeral space to store values for any 'work in
  progress' degrees being added by a candidate.
- This consequently allows us to fix a slightly broken user journey
  whereby partially-completed ApplicationQualification degrees appear on
  the candidate's degree review page if they abandon the form partway
  through.

## Link to Trello card

https://trello.com/c/x9RTYwgQ
## Things to check


malcolmbaig:
@tvararu @stevehook In implementing the 'edit' functionality for these degree fields, I'm starting to see some confusing code and complexity emerge from this approach. Specifically, each controller ends up dealing with two entities (the DraftDegree or the ApplicationQualification) depending on which action you're in. We can mitigate this by having separate controllers but then you end up with an inordinate number of controllers just to create a single degree record. 

I've spiked an alternative where we basically just stop the candidate marking this section as complete if they have any incomplete data. It's a simpler alternative with an acceptable UX, particularly as it's a minority of users who'll end up with incomplete degrees on their review page. Will get that work in a PR so we can discuss - closing this one for now.

PR #2287, opened 2020-06-16
Stop emails from sandbox

duncanjbrown:
## Context

The sandbox sends a lot of emails, many of them related to mass-generated applications. This is noisy for providers and for us.

- Stop delivering emails from sandbox
- Add an "Emails" tab to applications in the sandbox where providers can view the email history for an application.

Screenshot 2020-06-16 at 15 18 19

## Guidance to review

- Are emails correctly stopped
- Does the log work?

## Link to Trello card

https://trello.com/c/9MUQ9jrv/2292-stop-emails-from-sandbox

## Things to check

### DIFF
--- a/app/views/api_docs/pages/release_notes.md
+++ b/app/views/api_docs/pages/release_notes.md
@@ -1,3 +1,9 @@
+### 16th June 2020
+
+Sandbox changes:
+
+- Sandbox no longer sends emails to providers about application state changes.
+
 ### 15th June 2020
 
 New attributes:
--- a/app/views/api_docs/pages/release_notes.md
+++ b/app/views/api_docs/pages/release_notes.md
@@ -2,7 +2,7 @@
 
 Sandbox changes:
 
-- Sandbox no longer sends emails to providers about application state changes.
+- Sandbox no longer sends emails to providers about application state changes
 
 ### 15th June 2020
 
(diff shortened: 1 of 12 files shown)


config/application.rb
```diff
@@ -52,5 +52,7 @@ class Application < Rails::Application
     config.active_job.queue_adapter = :sidekiq
 
     config.middleware.use PDFKit::Middleware, { print_media_type: true }, disposition: 'attachment', only: [%r[^/provider/applications/\d+]]
+
+    config.action_mailer.perform_deliveries = false if HostingEnvironment.sandbox_mode?
```
tijmenb, line 56: This is a bit too broad I think, we're going to want emails for authentication, otherwise the candidate flow can't be tested. Similarly candidate notifications, referee emails, initial email to users to say that they've been invited are very useful (and I think wouldn't be too spammy, as they're triggered by actual user behaviour, not RDB/DBD robots).

Perhaps we can tag the providers notifications somehow, and add an interceptor to ignore those?

duncanjbrown: That’s a very good point! Interceptor sounds like a good idea. I’ll have a look tomorrow.

duncanjbrown:
broken with master — will fix with rebase

duncanjbrown:
Transient failure on Azure. This should be ready for another look

app/views/api_docs/pages/release_notes.md
```diff
@@ -1,3 +1,9 @@
+### 16th June 2020
+
+Sandbox changes:
+
+- Sandbox no longer sends emails to providers about application state changes. The email history of each application is available under an "Emails" tab on the web interface.
```
tijmenb, line 5: Perhaps leave this out, if this tab isn't going to be permanent.

```suggestion
- Sandbox no longer sends emails to providers about application state changes.
```

duncanjbrown: Sure

PR #2288, opened 2020-06-16
Make course queries cycle aware

duncanjbrown:
## Context

As we prepare for Rollover, we're going to start syncing courses for the next recruitment cycle.

This means that courses for 2020 and 2021 will live side by side in our database.

As a first step, we need to make sure that we don't serve up courses for the wrong recruitment cycle. We also want a way to preview the app when the next cycle has started.

- Move the definition of the current recruitment cycle out of `FindAPI` and into a module of its own called `RecruitmentCycle`.
- Visit each place in the app where we look up courses and scope the query to the current cycle using a scope on `Course`.
- When loading applications in the provider interface, make sure we only get applications for courses in the current cycle. This might need to change in future (eg when a provider needs to retain access to courses from the previous cycle) but I think this is a reasonable place to start.
- Add a feature flag to pretend we're in the 2021 cycle so we can see what it's like and what breaks

This PR makes no changes to support, as we already display recruitment cycle information .

## Guidance to review

I'm not 100% sure I've got all the places where we look up courses — are there others?

## Link to Trello card

https://trello.com/c/yijdh9OU/2252-rollover-make-course-queries-cycle-aware

## Things to check

### DIFF
--- a/app/models/recruitment_cycle.rb
+++ b/app/models/recruitment_cycle.rb
@@ -1,0 +1,9 @@
+module RecruitmentCycle
+  def self.current_year
+    if FeatureFlag.active?('switch_to_2021_recruitment_cycle')
+      2021
+    else
+      2020
+    end
+  end
+end
--- a/app/services/feature_flag.rb
+++ b/app/services/feature_flag.rb
@@ -18,6 +18,7 @@ class FeatureFlag
     [:dfe_sign_in_fallback, 'Use this when DfE Sign-in is down', 'Tijmen Brommet'],
     [:force_ok_computer_to_fail, 'OK Computer implements a health check endpoint, this flag forces it to fail for testing purposes', 'Michael Nacos'],
     [:pilot_open, 'Enables the Apply for Teacher Training service', 'Tijmen Brommet'],
+    [:switch_to_2021_recruitment_cycle, 'Sync and serve courses for the 2021 recruitment cycle. DO NOT ENABLE IN PRODUCTION.', 'Duncan Brown'],
   ].freeze
 
   TEMPORARY_FEATURE_FLAGS = [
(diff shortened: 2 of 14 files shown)


app/services/feature_flag.rb
```diff
@@ -18,6 +18,7 @@ def feature
     [:dfe_sign_in_fallback, 'Use this when DfE Sign-in is down', 'Tijmen Brommet'],
     [:force_ok_computer_to_fail, 'OK Computer implements a health check endpoint, this flag forces it to fail for testing purposes', 'Michael Nacos'],
     [:pilot_open, 'Enables the Apply for Teacher Training service', 'Tijmen Brommet'],
+    [:preview_2021_recruitment_cycle, 'Sync and serve courses for the 2021 recruitment cycle. DO NOT ENABLE IN PRODUCTION.', 'Duncan Brown'],
```
mnacos, line 21: How about `:use_2021_recruitment_cycle` instead? We could actually use this feature flag for real to migrate production environments to the new cycle.

duncanjbrown: Yes, preview's not much good! `use_` is not bad... but really you don't "use" a recruitment cycle, you're sort of in it. Maybe `switch_to_`? `enter_`?

app/models/recruitment_cycle.rb
```diff
@@ -0,0 +1,11 @@
+module RecruitmentCycle
+  def self.current_year
+    ENV.fetch('RECRUITMENT_CYCLE_YEAR') do
+      if FeatureFlag.active?('preview_2021_recruitment_cycle')
```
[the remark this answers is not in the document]
duncanjbrown: Yes, exactly —  having courses from both years will be the eventual state of prod but we don't want it until the real courses are out. In the meantime, we can test on other environments (or maybe a dedicated rollover environment).

You're right about the ENV var being redundant — I cargo-culted it from the previous implementation. And I've just checked and it's not configured in Azure anyway! Will take it out. :+1:

app/services/feature_flag.rb
mnacos: `:switch_to_2021_recruitment_cycle` is great

PR #2290, opened 2020-06-17
Add org-level `make_decisions` checks to ProviderAuthorisation

mnacos:
## Context

We might as well add the `provider_relationship_permissions` checks to `can_make_offer?` if the relevant feature flags are on. The UI doesn't expose the org-level `make_decisions` checkboxes yet, but the database already supports `make_decisions` at an organisation level.

A provider user may have org-level `make_decisions` for a course because they are associated with the training provider or the ratifying_provider for the course or both, as long as at least one of these organisations has org-level `make_decisions`.

## Guidance to review

Please review `provider_authorisation_spec.rb` and `ProviderAuthorisation`.

## Link to Trello card

https://trello.com/c/CwAkdmkP

## Things to check


app/services/provider_authorisation.rb
```diff
@@ -64,6 +69,26 @@ def course_option_belongs_to_user_providers?(course_option:)
     @actor.providers.include?(course_option.course.provider)
   end
 
+  def provider_level_permissions(training_provider:, ratifying_provider:)
+    permissions = []
```
steventux, line 73: We return an array of `ProviderRelationshipPermissions` here, possibly containing a permissions record on each side of the relationship ie. one `AccreditedBodyPermissions` record and one `TrainingProviderPermissions` record. The end result is to then call the `make_decisions` method on both. 
Couldn't we add `make_decisions` to the `find_by` scope and just test the result for presence/emptiness?
Alternatively if the `provider_level_permissions` method is meant to be an ease-of-use utility method we could pass filtering arguments to it eg.
```ruby
def provider_level_permisions(training_provider:, ratifying_provider:, filter_attrs: {})
   attrs = { training_provider: training_provider, ratifying_provider: ratifying_provider }.merge(filter_attrs)
   [
     ProviderInterface::TrainingProviderPermissions.find_by(attrs),
     ProviderInterface::AccreditedBodyPermissions.find_by(attrs),
   ].compact
end
```
Either way I think making the finding method specific to the permission we're checking, will be clearer.

mnacos: Hi Steve, your suggestions make sense but they interrogate `provider_relationship_permissions` without taking `@actor` into account. I started refactoring `provider_level_permissions` to add `filter_attrs` but once I started introducing @actor-based conditionals the elegance/brevity of AR-based filters was lost.

I'll keep the implementation as is, for now, but rename the method to `provider_relationship_permissions_for_actor` to emphasise the role of actor in all this.

PR #2299, opened 2020-06-18
Split up Degrees::BaseController

malcolmbaig:
## Context

The add degree flow starts with a page that asks for type/subject/institution on one page. The new HESA/structured data flow has them all separated. We can separate them out first and enhance with structured data after.

Major changes:

- Split the Base controller and the DegreeForm into three separate equivalents.
- Rework the test suite to reflect these changes.
- Add a validation to the degrees review page to prevent marking the section complete if any degree information is missing. This fixes a pre-existing bug whereby the candidate can submit incomplete degrees if they abandon the flow partway through and then return to the review page.

## Guidance to review

- Per-commit review recommended.
- Should be fairly straightforward to manually test the add degree flow in the review app.

There are a few (less crucial) bits that need doing which will follow in another PR, as this one is quite large already. These are:

- Consistently resetting `ApplicationForm#degrees_completed?` to `false` whenever there's an add/edit to the degree information.
- Consistent use of `track_validation_error` across all degree controllers.
- Clean up of obsolete code and translation strings.

## Link to Trello card
https://trello.com/c/x9RTYwgQ/1684-split-up-degree-controller

## Things to check


app/models/application_qualification.rb
```diff
@@ -16,4 +25,16 @@ class ApplicationQualification < ApplicationRecord
   def missing_qualification?
     qualification_type == 'missing'
   end
+
+  def incomplete_degree_information?
+    return false unless degree?
+
+    EXPECTED_DEGREE_FIELDS.each do |field|
+      if send(field).nil?
```
stevehook, line 33: Could any of these contain empty strings? If so would that mean it's incomplete?

stevehook: Can this be shortened to something like: `EXPECTED_DEGREE_FIELDS.any? { |field| send(field).nil? }`

malcolmbaig: @stevehook 
Empty strings are unlikely given the presence validations across all the forms, but should be considered incomplete.

Have addressed both questions in https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2299/commits/dd499b41d4a7da192fdabe39cfb1e12c778a4b99

WDYT?

stevehook: Looks good :+1:

PR #2300, opened 2020-06-18
Rename AccreditedBodyPermissions to RatifyingProviderPermissions

steventux:
## Context

We have several terms for providers who ratify courses for other providers.

- Accredited body
- Accredited provider
- Ratifying provider

Attempt to make the permissions relationship between the two types of providers clearer. 
The ratifying provider ratifies courses for the training provider, this seems to be more common terminology that using the phrase _accredited body_. It's a bit more consistent with the naming of the counterpart STI class `TrainingProviderPermissions`

Also demodulize the permissions classes `ProviderRelationshipPermissions`, `TrainingProviderPermissions` and `RatifyingProviderPermissions`

## Things to check


steventux:
@tijmenb @duncanjbrown what's the best / preferred strategy for migrating any existing `ProviderRelationshipPermissions` records which currently have a type of `ProviderInterface::AccreditedBodyPermissions` ? I have created a migration but I assume this can't be part of this PR because we can't guarantee whether the migration runs before the code is deployed? Is it better to fix these in the console(s) with
```ruby
ProviderRelationshipPermissions.where(type: 'ProviderInterface::AccreditedBodyPermissions').update_all(type: 'RatifyingProviderPermissions')
```

PR #2314, opened 2020-06-19
Fix broken test on master

davidgisbey:
## Context

The test is deleting clicking delete on the undergraduate degree, not the masters. It then checks to see that the master's degree has been deleted and blows up

## Things to check

### DIFF
--- a/spec/system/candidate_interface/candidate_entering_degrees_spec.rb
+++ b/spec/system/candidate_interface/candidate_entering_degrees_spec.rb
@@ -64,7 +64,7 @@ RSpec.feature 'Entering their degrees' do
     when_i_fill_in_my_additional_degree
     then_i_can_check_my_additional_degree
 
-    when_i_click_on_delete_degree
+    when_i_click_delete_on_my_additional_degree
     and_i_confirm_that_i_want_to_delete_my_additional_degree
     then_i_can_only_see_my_undergraduate_degree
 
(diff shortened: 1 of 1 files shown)


spec/system/candidate_interface/candidate_entering_degrees_spec.rb
```diff
@@ -257,6 +257,12 @@ def then_i_can_check_my_additional_degree
     expect(page).to have_content 'Masters Maths'
   end
 
+  def when_i_click_delete_on_my_additional_degree
+    within(find_all('.app-summary-card__header')[1]) do
```
malcolmbaig, line 261: The additional degree appears first in list, so need [0] here i think.

davidgisbey: Hm for whatever reason locally it's showing up after the masters.

davidgisbey: image

image

davidgisbey: I'm assuming this is due to the undergrad degree being deleted and then added again causing an ordering issue.

stevehook: I'm seeing the same thing - Masters degree appearing consistently second in the list at the point the test fails.

malcolmbaig: For some reason the test run on Azure seems to be failing because this order isn't deterministic. I would've thought implicit ordering by primary key would apply but apparently not. 

Getting a pass on Azure using the fix [here](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/2322). Only remaining failure there is the Slack-related test that @duncanjbrown and @vigneshmsft are looking at.

davidgisbey: Great. Will close this then 👍

PR #2315, opened 2020-06-19
Refactor provider UI to rely on #can_make_offer?

mnacos:
## Context

The controller before filter raising/rendering 'Access Denied' when a user would 'hack' an application decision url was not taking into account org-level permissions, even though other parts of the code did.

All UI-level decisions now rely on `#can_make_offer?`, which is the most reliable authority for such checks. This PR also fixes a bug in `#can_make_offer?` to do with org-level checks when the accedited body is `nil`.

When the course is self-ratified, i.e. ratified by the training provider, no relevant `ProviderRelationshipPermission` records exist in the database. If `accredited_body` is `nil`, absence of records should not be interpreted as lack of permissions. Training providers should have full permission to `make_decisions` on such courses, from an organisation point of view. 

## Guidance to review

Is my description correct about what happens with `ProviderRelationshipPermission` records when courses are self-ratified?

## Link to Trello card

https://trello.com/c/CwAkdmkP

## Things to check


app/controllers/provider_interface/decisions_controller.rb
```diff
@@ -1,9 +1,10 @@
 module ProviderInterface
   class DecisionsController < ProviderInterfaceController
     before_action :set_application_choice
-    before_action :requires_provider_user_make_decisions_permission
     before_action :requires_provider_change_response_feature_flag, only: %i[new_withdraw_offer confirm_withdraw_offer withdraw_offer]
 
+    include RequiresMakeDecisionsPermission
```
duncanjbrown, line 6: :+1: Pulling this out makes a lot of sense to me. Would it be more Rails-y to mix in the module in `ProviderInterfaceController` without the `included...` block then continue to use a traditional `before_action : requires_provider_user_make_decisions_permission`?

app/services/provider_authorisation.rb
```diff
@@ -10,7 +10,7 @@ def can_make_offer?(application_choice:, course_option_id:)
 
     course_option = CourseOption.find(course_option_id)
     training_provider = course_option.provider
-    ratifying_provider = course_option.course.accredited_provider
+    ratifying_provider = course_option.course.accredited_provider || training_provider
```
duncanjbrown, line 13: Changing the roles around here seems surprising. Maybe we need to have explicit language about "self-ratifying" providers?

spec/system/provider_interface/provider_responds_to_application_spec.rb
```diff
@@ -15,22 +16,64 @@
...
     and_i_am_not_permitted_to_make_decisions_for_my_provider
+    and_my_organisation_is_permitted_to_make_decisions
     and_i_sign_in_to_the_provider_interface
 
     when_i_try_to_respond_to_an_application
     then_i_get_access_denied
   end
 
-  scenario 'Provider can respond to an application currently awaiting_provider_decision' do
+  scenario 'Provider cannot respond to an application if organisation lacks make_decisions' do
```
duncanjbrown, line 30: This system spec now contains several scenarios, which isn't how we do system specs. Could we break this up and/or find a way to test the edges at a lower level?

app/controllers/provider_interface/decisions_controller.rb
mnacos: Ok, I'll remove the `included...` block and use explicit before actions.

app/services/provider_authorisation.rb
mnacos: Turns out, the logic is wrong. The data sync task will not create `ProviderRelationshipPermissions` records for self-ratified courses and we currently interpret the absence of records as lack of permissions, denying self-ratifying providers the ability to make decisions.

Discussing this with Duncan, we've come to the conclusion we should be skipping org-level permissions checks for self-ratified courses.

spec/system/provider_interface/provider_responds_to_application_spec.rb
mnacos: The way `#can_make_offer?` behaves has been thoroughly tested in the relevant service spec. What's missing is a way of asserting controller flows in the absence or presence of these feature flags. I'll remove these scenaria from here and create a new 'provider responds to application' file for post-launch (all feature flags on).

app/lib/requires_make_decisions_permission.rb
```diff
@@ -0,0 +1,17 @@
+module RequiresMakeDecisionsPermission
```
tijmenb, line 1: I'd like to avoid `includes` / concerns where possible, since they make it harder to find the place where a method is defined. Can we move this into `ProviderInterfaceController` instead?

tijmenb: For what it's worth, this is the nightmare scenario I'm thinking about when dealing with includes:

https://github.com/alphagov/whitehall/blob/98287c46f7ee7802759e8df123ca1e06519a92dc/app/models/edition.rb#L4-L35

https://github.com/alphagov/whitehall/blob/98287c46f7ee7802759e8df123ca1e06519a92dc/app/models/case_study.rb#L2-L10

mnacos: Nice

PR #2320, opened 2020-06-22
International addresses

stevehook:
## Context

Candidates who live overseas need to provide postal address in a localised format so that providers can contact them by post.

- [x] We added the `international_address` and `address_type` columns to `application_forms` in an earlier merged PR.
- [x] Add an `international_addresses` feature flag.
- [x] Add new form to the contact details flow to prompt for the type of address (UK or international) that sits between the phone number and address forms.
- [x] Change the existing address form to use a single text area for address input if the candidate specified international.
- [x] Update the `ContactDetailsReviewComponent` to show international addresses when appropriate and have the change button link to the address type page (so that the candidate can switch from one type of address to the other as well as change the actual address).
- [x] Add a country drop down to the _Where do you live?_ page. (Following discussions about how we can fit international addresses into the existing Vendor API schema).
- [x] `ApplicationForm#country` stores an ISO 8166 two-letter country code. Added `config/initializers/countries.rb` to store a country list obtained from the GOV.UK countries register.

image

image

image

## Guidance to review

- Is the feature flag watertight?
- Would it make sense to rename the feature flag `international` to cover other cards in the epic or keep it small?
- We would run into problems if we enable the feature, candidates start adding international addresses and we switch it off again - the review page doesn't show international addresses unless the feature is on.
- Not covered in this PR - changes to the Vendor API - separate PR to follow if needed
- Should we clear the `address_line1`, `address_line2` etc. if a candidate switches from a UK to international address and clear `international_address` if a candidate switches from international to UK?

## Link to Trello card

https://trello.com/c/Wif90Hdc/1683-dev-🌐-international-addresses

## Things to check


tvararu:
> Should we clear the address_line1, address_line2 etc. if a candidate switches from a UK to international address and clear international_address if a candidate switches from international to UK?

Yes, I think we should!

stevehook:
> > Should we clear the address_line1, address_line2 etc. if a candidate switches from a UK to international address and clear international_address if a candidate switches from international to UK?
> 
> Yes, I think we should!

👍 done

app/frontend/packs/nationality-autocomplete.js
```diff
@@ -6,7 +6,8 @@ const initNationalityAutocomplete = () => {
       "#candidate-interface-personal-details-form-first-nationality-field",
       "#candidate-interface-personal-details-form-first-nationality-field-error",
       "#candidate-interface-personal-details-form-second-nationality-field",
-      "#candidate-interface-personal-details-form-second-nationality-field-error"
+      "#candidate-interface-personal-details-form-second-nationality-field-error",
+      "#candidate-interface-contact-details-form-country-field"
```
tvararu, line 10: Is there an -error version of this new field?

app/controllers/candidate_interface/contact_details/address_type_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+    def update
+      render_404 and return unless FeatureFlag.active?(:international_addresses)
+
+      @contact_details_form = ContactDetailsForm.new(address_type_params)
+
+      if @contact_details_form.save_address_type(current_application)
+        current_application.update!(contact_details_completed: false)
+
+        # TODO: redirect to the correct form
+        redirect_to candidate_interface_contact_details_edit_address_path
```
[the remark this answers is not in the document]
stevehook: I was originally thinking the flow would branch here but it's simpler than I thought. Removed the comment.

app/frontend/packs/nationality-autocomplete.js
stevehook: Yes there is, thanks for spotting that

PR #2326, opened 2020-06-22
Remove confusing line of content

davidgisbey:
## Context

We were telling users they could go back to change their study mode or location. However, the radio buttons only show if they are available options. 

This may cause confusion.

## Link to Trello card

## Things to check


fofr:
Shouldn't this line remain but be conditional on the same thing?

PR #2327, opened 2020-06-23
Add withdrawal feedback column to application choice

davidgisbey:
## Context

We're adding a page for candidates to ask them to provide a withdrawal reason and asking if we can contact them for research.

- Add a withdrawal_feedback column which will store their responses as a hash

## Link to Trello card

https://trello.com/c/fHkzCjyC/1702-dev-reasons-for-candidate-withdrawing-application

## Things to check


davidgisbey:
Yea i did think about other approaches, but i as you say it's in line with other questionnaires and f we want to add additional questions at any point this makes it more flexible.

davidgisbey:
although a consent_to_be_contacted and withdrawl_feedback column would probably be more straightforward, but less complicated.

PR #2329, opened 2020-06-23
SyncProviderFromFind can work on providers from different recruitment cycles

duncanjbrown:
## Context

Syncing providers has two steps:

1. getting the list of providers for the current recruitment cycle
2. loading the courses, sites etc for each provider we got in step 1. This operation needs to know which recruitment cycle the provider is in

We want to be able to handle syncing courses for two cycles at the same time as this will mean we can properly test the cycles living side by side in our database. And hopefully, in the future, Find will be able to publish courses as they roll over, rather than in a block, as they plan to this year.

So far we've always been in the 2020 cycle, so we've been able to use `RecruitmentCycle.current_year` inside `SyncAllProvidersFromFind` (which does step 1)  and `SyncProviderFromFind` (which does step 2).

This PR is about pulling that hardcoded value out of those places and turning it into an argument. 
- A lot of specs need to be changed to pass the new arg, but they're all changing in the same way.
- Stubs are going to need to be aware of which recruitment cycle to put their fake responses in. In a spec with `FindAPIHelper` included you can call `recruitment_cycle_year` to get the currently-stubbed recruitment cycle, and `set_stubbed_recruitment_cycle_year!` to set it. (Which doesn't happen in this PR, but will happen in the next one). I preferred `set_stubbed_recruitment_cycle_year!` as just `stubbed_recruitment_cycle_year = whatever` feels weird & magical.

The next PR will start syncing 2021 courses and handling the carrying over of "open_on_apply" status from cycle to cycle.

## Guidance to review

Does it make sense? Are there any references left to `current_year`?

## Link to Trello card

https://trello.com/c/Tgq41OB9/2253-rollover-start-syncing-dummy-2021-data

## Things to check

### DIFF
--- a/app/models/find_api/provider.rb
+++ b/app/models/find_api/provider.rb
@@ -12,8 +12,8 @@ module FindAPI
     # elegant way, feel free.
     class Subject < FindAPI::Resource; end
 
-    def self.current_cycle
-      where(recruitment_cycle_year: RecruitmentCycle.current_year)
+    def self.recruitment_cycle(year)
+      where(recruitment_cycle_year: year)
     end
   end
 end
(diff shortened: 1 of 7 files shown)


duncanjbrown:
Thanks @malcolmbaig

PR #2332, opened 2020-06-23
pipeline to build a base docker image with gem node installed for reducing main CI build times

vigneshmsft:
## Context

Current docker build takes ~5 mins, it can be reduced by using a cached image with gems and node modules pre-installed.

refactor Dockerfile into stages to use cached layers.
new pipeline to build gem-node-dependencies images on change to any of Gemfile, Gemfile.lock, yarn.lock and package.json
new Github action to tigger a new build of gem-node-dependencies when a new branch is created.
use this base image in the man CI pipeline to build the final application image.

## Link to Trello card

https://trello.com/c/6oJaGrIC/2320-create-a-base-image-with-gem-npm-dependencies-to-reduce-build-times

## Things to check

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,61 +1,86 @@
-#### Common Build Image Stage ####
+# To use or update to a ruby version, change {BASE_RUBY_IMAGE}
+ARG BASE_RUBY_IMAGE=ruby:2.7.1-alpine
+# BASE_RUBY_IMAGE_WITH_GEMS_AND_NODE_MODULES will default to apply-for-teacher-training-gems-node-modules
+# building all layers above it if a value is not specidied during the build
+ARG BASE_RUBY_IMAGE_WITH_GEMS_AND_NODE_MODULES=apply-for-teacher-training-gems-node-modules
 
-FROM ruby:2.7.1-alpine AS common-build
+# Stage 1: install-gems-node-modules, download gems and node modules.
+FROM ${BASE_RUBY_IMAGE} AS install-gems-node-modules
 
-ARG BUILD_DEPS="git gcc libc-dev make nodejs yarn postgresql-dev graphviz-dev graphviz-doc build-base libxml2-dev libxslt-dev ttf-ubuntu-font-family"
-ARG bundleWithout=""
-
-# These variables are required for running Rails processes like assets:precompile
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -1,61 +1,85 @@
-#### Common Build Image Stage ####
+# To use or update to a ruby version, change {BASE_RUBY_IMAGE}
+ARG BASE_RUBY_IMAGE=ruby:2.7.1-alpine
+# BASE_PACKAGES_IMAGE will default to apply-for-teacher-training-gem-node-dependencies
+# building all layers above it if a value is not specidied during the build
+ARG BASE_PACKAGES_IMAGE=apply-for-teacher-training-gem-node-dependencies
 
-FROM ruby:2.7.1-alpine AS common-build
+# Stage 1: gem-node-packages, download gem and node packages.
+FROM ${BASE_RUBY_IMAGE} AS gem-node-packages
```
duncanjbrown, line 8: I think the language around dependencies and packages here could do with one more revision.

We're building a thing in that's going to be the `BASE_PACKAGES_IMAGE` which is really `gem-node-dependencies` which is a slimmed-down version of `gem-node-packages`. 

- `gem` and `node` aren't equivalent — `gem` is a package but `node` is a runtime.
- we're mixing up `dependencies` and `packages`

I think we should call them `dependencies` throughout as `packages` implies OS packages.

Propose something like:

```
ARG RUBY_BASE_IMAGE_WITH_BUNDLED_GEMS_AND_NODE_MODULES=gems-and-node-modules-without-build-dependencies
...
from ${BASE_RUBY_IMAGE} gems-and-node-modules
...
from ${BASE_RUBY_IMAGE} gems-and-node-modules-without-build-dependencies
...
from ${BASE_RUBY_IMAGE_WITH_BUNDLED_GEMS_AND_NODE_MODULES} assets-precompile
...
from ${BASE_RUBY_IMAGE} production
```

vigneshmsft: thanks @duncanjbrown have renamed as suggested, but have kept the second layer name as `apply-for-teacher-training-gems-node-modules` instead of `gems-and-node-modules-without-build-dependencies` to keep is same as the name of the image being published.

PR #2337, opened 2020-06-23
Fix first login 'Access Denied' bug when other org setups are pending

mnacos:
## Context

There is a weird bug affecting provider users with `manage_organisations` permission when they try to log in when the `:enforce_provider_to_provider_permissions` flag is active, serving them Access Denied. Full steps to reproduce the this error can be found below.

Specify the `training_provider_id` when redirecting provider users to the org-level permissions setup page. The previous code was picking up organisations unrelated to the user.

## Guidance to review

Steps to reproduce bug and verify the fix:
- Clear your local data.
- Recreate and seed the database (e.g. `setup_local_dev_data`)
- Use the support console to add Oriel High School to the providers
- Add an Oriel High School user (e.g. `oriel@example.com`)
- Switch the `:enforce_provider_to_provider_permissions` feature flag on
- Log into the provider interfaces as the Oriel High School user
- Accept the Data Sharing Agreement
- What do you see next? Access Denied or the organisation permissions setup page? 

I've added `training_provider_id` as a query param. Should it be part of the url?

## Link to Trello card

https://trello.com/c/TfyXM7kv

## Things to check

### DIFF
--- a/app/controllers/provider_interface/provider_interface_controller.rb
+++ b/app/controllers/provider_interface/provider_interface_controller.rb
@@ -73,22 +73,23 @@ module ProviderInterface
       return unless current_provider_user
       return if performing_provider_organisation_setup?
 
-      if provider_permissions_need_setup?
-        redirect_to provider_interface_provider_relationship_permissions_setup_path
+      if (provider_id = next_training_provider_to_setup&.id)
+        redirect_to provider_interface_provider_relationship_permissions_setup_path(training_provider_id: provider_id)
       end
     end
 
-    def provider_permissions_need_setup?
+    def next_training_provider_to_setup
       permissions = TrainingProviderPermissions.find_by(
         setup_at: nil,
         training_provider: current_provider_user.providers,
       )
 
-      return false if permissions.blank?
+      if permissions.present?
+        auth = ProviderAuthorisation.new(actor: current_provider_user)
 
-      ProviderAuthorisation.new(actor: current_provider_user).can_manage_organisation?(
-        provider: permissions.training_provider,
-      )
+        training_provider = permissions.training_provider
+        training_provider if auth.can_manage_organisation?(provider: training_provider)
+      end
     end
 
     def performing_provider_organisation_setup?
--- a/app/controllers/provider_interface/provider_relationship_permissions_controller.rb
+++ b/app/controllers/provider_interface/provider_relationship_permissions_controller.rb
@@ -4,7 +4,7 @@ module ProviderInterface
     before_action :render_403_unless_access_permitted
 
     def setup
-      @training_provider_permissions = TrainingProviderPermissions.where(
+      @relationships_pending_setup = TrainingProviderPermissions.where(
         setup_at: nil,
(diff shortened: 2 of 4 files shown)


duncanjbrown:
> It seems like this bug has arisen from two slightly differing checks in the controllers and I wonder if that's a good argument to consolidate them in one place like `ProviderAuthorisation`?. Then both controllers share the same logic and there's no need to pass the training provider id around.

> My concern is that `ProviderRelationshipPermissionsController` should probably be the start of the setup process regardless of any other filters or hooks elsewhere in the application. This creates a reliance on something else feeding it the training provider, which isn't a big deal but I wonder if the onboarding steps should be more self-sufficient.

I think relying on contextual state is always going to be less robust than going back to the source of truth for a fresh picture of the state each time, so I think I'd also be in favour of pushing this up into `ProviderAuthorisation` and taking the business logic out of the controller.

mnacos:
> > It seems like this bug has arisen from two slightly differing checks in the controllers and I wonder if that's a good argument to consolidate them in one place like `ProviderAuthorisation`?. Then both controllers share the same logic and there's no need to pass the training provider id around.
> 
> > My concern is that `ProviderRelationshipPermissionsController` should probably be the start of the setup process regardless of any other filters or hooks elsewhere in the application. This creates a reliance on something else feeding it the training provider, which isn't a big deal but I wonder if the onboarding steps should be more self-sufficient.
> 
> I think relying on contextual state is always going to be less robust than going back to the source of truth for a fresh picture of the state each time, so I think I'd also be in favour of pushing this up into `ProviderAuthorisation` and taking the business logic out of the controller.

Plenty of good suggestions here but I've approached this as a bugfix, not a refactor :)

There are multiple concerns in the comments:
- using `ProviderAuthorisation` as the source of truth for restricting access
- self-sufficiency of the onboarding/setup process (which could be refactored to combine DSA/org setup)
- making setup provider-specific (e.g. by passing a provider id as part of the route)

I think the key question is what do we intend to do with the onboarding process. Keeping with the current approach (`ProviderRelationshipPermissionsController`) my suggestion would be to make `training_provider_id` a part of the url and a dependency of the setup flow. The setup process can only deal with one training provider at a time, so we might as well make this explicit. Deciding which provider to trigger the setup flow for would be a job for `ProviderInterfaceController`. What's missing perhaps is an onboarding layer (e.g. service) performing the queries that decide whether to divert a user to the onboarding flow. But the logic is still pretty simple so it may be tolerable to keep it as a method in the controller.

Now, if we were to rework the onboarding process to incorporate both DSA checks and the org relationship checks, this would be a different story. But it would also be a sizeable refactor and its own story. WDYT?

steventux:
_The setup process can only deal with one training provider at a time, so we might as well make this explicit._  I had this thought too and I can see the logic in structuring the setup this way but I don't think we should do it. I'd prefer to see parity in the permissions checks and a _cold-start_ behaviour in the permissions controller. Even if this means duplicating the permissions check to get a bug fix out and then DRYing it up in another PR.

PR #2339, opened 2020-06-23
Vendor API updates for international addresses

stevehook:
## Context

Candidates who live overseas need to provide postal address in a localised format so that providers can contact them by post.

The upstream PR https://github.com/DFE-Digital/apply-for-teacher-training/pull/2320 implements the main part of this feature. This PR just updates the Vendor API so that vendors have access to international addresses.

- [x] Update the `SingleApplicationPresenter` to serialize international addresses - the address body is output to the `address_line1` field to avoid making any significant changes to the schema.
- [x] Update schema to allow nil values for UK-specific address fields.
- [x] Update release notes.
- [x] Rebase onto `master`

## Guidance to review

- This is designed to be a non-breaking change to the API and vendor applications. Is this true?

## Link to Trello card

https://trello.com/c/BUYJXkju/1735-extend-vendor-api-to-support-international-addresses

## Things to check

### DIFF
--- a/app/presenters/vendor_api/single_application_presenter.rb
+++ b/app/presenters/vendor_api/single_application_presenter.rb
@@ -188,7 +188,14 @@ module VendorAPI
     end
 
     def contact_details
-      if application_form.uk?
+      if application_form.international? && FeatureFlag.active?(:international_addresses)
+        {
+          phone_number: application_form.phone_number,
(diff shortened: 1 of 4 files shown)


duncanjbrown:
Approach looks good. Should we add a `description` to `ContactDetails` or the `address_line_1` field (or both) explaining what the policy is?

PR #2342, opened 2020-06-24
Update the degree type page to use HESA data

malcolmbaig:
## Context

We need to allow users to choose from HESA options when they choose degree type during the 'add degree' flow.

- Make degree type HESA data available to the frontend by rendering it into a data attribute on the page.
- Add JS to enhance the existing degree type input field with autocomplete based on this data.
- Allow the user to submit their own custom type description in addition to selecting a HESA type.
- Ensure we store the HESA code if the candidate selects one of the HESA options.
- Add a model (`Hesa::DegreeType`) to ease the process of retrieving and querying the degree type data.

### Before

---
Screenshot 2020-06-24 at 09 21 20

---

### After

---
Screenshot 2020-06-24 at 09 21 01

---

## Guidance to review

- Per-commit recommended, followed by a quick check of the review app.
- Does the model around the HESA data make sense? Is there anything you'd change about its API?
- Missing from this PR is the scoping by undergrad/postgrad described on the card. This will follow in its own PR shortly.

## Link to Trello card

https://trello.com/c/rXUPhuBC
## Things to check

### DIFF
--- a/app/frontend/packs/degree-type-autocomplete.js
+++ b/app/frontend/packs/degree-type-autocomplete.js
@@ -19,28 +19,35 @@ const initDegreeTypeAutocomplete = () => {
   }
 
   try {
-    const inputId = "degree-type";
-    const input = document.getElementById(inputId);
-    const containerId = "degree-type-autocomplete";
-    const container = document.getElementById(containerId);
+    const inputIds = [
+      "candidate-interface-degree-type-form-type-description-field",
+      "candidate-interface-degree-type-form-type-description-field-error"
+    ];
 
-    if (!container) return;
+    inputIds.forEach(inputId => {
+      const input = document.getElementById(inputId);
+      if (!input) return;
 
-    const sourceData = JSON.parse(container.dataset.source);
+      const containerId = "degree-type-autocomplete";
+      const container = document.getElementById(containerId);
+      if (!container) return;
 
-    accessibleAutocomplete({
-      element: container,
-      id: inputId,
-      name: input.name,
-      source: sourceData,
-      showNoOptionsFound: true,
-      templates: {
-        inputValue: inputTemplate,
-        suggestion: suggestionTemplate
-      }
+      const sourceData = JSON.parse(container.dataset.source);
+
+      accessibleAutocomplete({
+        element: container,
+        id: inputId,
+        name: input.name,
+        source: sourceData,
+        showNoOptionsFound: true,
+        templates: {
+          inputValue: inputTemplate,
+          suggestion: suggestionTemplate
+        }
+      });
+
+      input.remove();
     });
-
-    input.remove();
 
   } catch (err) {
     console.error("Could not enhance degree type input:", err);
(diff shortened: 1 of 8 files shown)


tvararu:
Screenshot 2020-06-24 at 11 12 14

On the review app, if the feature flag is on, the JS fails to initialise with an error. This isn't a big deal however, as we have progressive enhancement.

app/views/candidate_interface/degrees/type/_form_fields.html.erb
```diff
@@ -1,7 +1,10 @@
 <%= f.govuk_text_field(
   :type_description,
+  id: 'degree-type',
```
tvararu, line 3: This breaks the label `for=` attribute

app/frontend/packs/degree-type-autocomplete.js
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    function suggestionTemplate(result) {
+      const descriptor = result.split('|');
+
+      if (descriptor[0]) {
+        return `<strong>${descriptor[0]}</strong> <span class="autocomplete__option--hint">${descriptor[1]}</span>`
+      }
+      return `<strong>${descriptor[1]}</strong>`
+    }
```
tvararu, line 26: I would hoist these outside of this function to make them less indented. 📐

app/frontend/packs/degree-type-autocomplete.js
```diff
@@ -0,0 +1,48 @@
...
+
+    if (!container) return;
+
+    input.remove();
```
tvararu, line 30: You should do this after the `accessibleAutocomplete` call; if for whatever reason the JS fails below, the user will be left without an input!

spec/forms/candidate_interface/degree_type_form_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+          type_description: 'Doctor of Rap Battles',
+          application_form: create(:application_form),
+        )
+      end
+
+      it 'persists type description but no HESA code' do
+        form.save
+
+        degree = form.application_form.application_qualifications.degree.first
+        expect(degree.qualification_type).to eq 'Doctor of Rap Battles'
```
tvararu, line 35: Nice

app/views/candidate_interface/degrees/type/_form_fields.html.erb
malcolmbaig: Good shout, fixing..

malcolmbaig: Addressed in latest commit.

app/frontend/packs/degree-type-autocomplete.js
```diff
@@ -0,0 +1,57 @@
...
+      if (!container) return;
+
+      const sourceData = JSON.parse(container.dataset.source);
+
+      accessibleAutocomplete({
+        element: container,
+        id: inputId,
+        name: input.name,
+        source: sourceData,
+        showNoOptionsFound: true,
```
tvararu, line 42: We should make sure the input is pre-filled with what the user typed in:

```suggestion
        showNoOptionsFound: true,
        defaultValue: input.value,
```

PR #2343, opened 2020-06-24
Refactor start of org permissions setup flow

mnacos:
## Context

Just a small refactor based on conversations around https://github.com/DFE-Digital/apply-for-teacher-training/pull/2337

Extract query for pending org permissions into a service.

## Guidance to review

Should be straightforward.

## Link to Trello card

https://trello.com/c/TfyXM7kv

## Things to check

### DIFF
--- a/app/controllers/provider_interface/provider_interface_controller.rb
+++ b/app/controllers/provider_interface/provider_interface_controller.rb
@@ -73,22 +73,8 @@ module ProviderInterface
       return unless current_provider_user
       return if performing_provider_organisation_setup?
 
-      if (provider_id = next_training_provider_to_setup&.id)
-        redirect_to provider_interface_provider_relationship_permissions_setup_path(training_provider_id: provider_id)
-      end
-    end
-
-    def next_training_provider_to_setup
-      permissions = TrainingProviderPermissions.find_by(
-        setup_at: nil,
-        training_provider: current_provider_user.providers,
-      )
-
-      if permissions.present?
-        auth = ProviderAuthorisation.new(actor: current_provider_user)
-
-        training_provider = permissions.training_provider
-        training_provider if auth.can_manage_organisation?(provider: training_provider)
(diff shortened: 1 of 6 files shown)


app/controllers/provider_interface/provider_interface_controller.rb
```diff
@@ -73,22 +73,8 @@ def check_provider_relationship_permissions
...
-        setup_at: nil,
-        training_provider: current_provider_user.providers,
-      )
-
-      if permissions.present?
-        auth = ProviderAuthorisation.new(actor: current_provider_user)
-
-        training_provider = permissions.training_provider
-        training_provider if auth.can_manage_organisation?(provider: training_provider)
+      if ProviderSetup.new(provider_user: current_provider_user).next_relationship_pending
```
duncanjbrown, line 76: Yes!

spec/services/provider_setup_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      3.times.each do |idx|
+        expected_relationship = relationships[idx]
+        expect(next_relationship_pending).to eq(expected_relationship)
+        expected_relationship.update(setup_at: Time.current)
+      end
+
+      expect(next_relationship_pending).to be_nil
+    end
+
+    it 'returns nil if all manageable relationships have been set up' do
```
duncanjbrown, line 43: I think this spec effectively duplicates the above — do we need it?

app/controllers/provider_interface/provider_interface_controller.rb
duncanjbrown: We should have a follow-up to move the DSA code into here too

spec/services/provider_setup_spec.rb
mnacos: I guess it's covered by the previous one. Removing

PR #2348, opened 2020-06-24
Support interface magic link sign in

tvararu:
## Context

We need to allow support users to sign in via the a fallback mechanism when DSI is down.

- Add tokens table, in the future we'll transition Candidate + ProviderUser + ApplicationReference to this
- Create AuthenticationToken class and wire up SupportUser to it
- Copy+paste controllers, views, service, test from how ProviderUser does it, with the idea to refactor and consolidate this later

## Guidance to review

Review commit by commit.

There is a migration in this PR but because it's behind a feature flag, we do not rely on it / it won't break background workers.

## Link to Trello card

https://trello.com/c/1KWg2n91/1741-dev-provide-access-to-support-via-magic-link

## Things to check


db/schema.rb
```diff
@@ -43,7 +43,7 @@
     t.string "offer_withdrawal_reason"
     t.datetime "offer_withdrawn_at"
     t.datetime "sent_to_provider_at"
-    t.jsonb "structured_rejection_reasons"
+    t.text "structured_rejection_reasons"
```
tvararu, line 46: Why is this a thing?

duncanjbrown:
Tokens table is a good idea. Should we do a preliminary refactoring PR to set it up and move `ProviderUser` and `Candidate` tokens into it?

tvararu:
@duncanjbrown my and @tijmenb's preference would be to ship the end to end implementation for SupportUser end to end, and then transition everything else to the tokens table:

- Candidates
- ProviderUser
- ApplicationReference

tvararu:
Re-doing commits in a sec but marking as ready for review.

db/schema.rb
duncanjbrown: We were thinking about serialising to YAML instead of JSON

tvararu: How come it's coming is as a migration on my PR? 😳

app/models/support_user_token.rb
```diff
@@ -0,0 +1,3 @@
+class SupportUserToken < Token
```
tijmenb, line 1: I think we don't need STI, but use polymorphic associations:

```rb
class AuthenticationToken
 belongs_to :authenticable, polymorpic: true
end

# and then

support_user = AuthenticationToken.find_by(token: token).authenticable
```

db/schema.rb
tvararu: These get updated on `master` when you run `db:migrate`, I won't make this changest part of this PR!

tvararu:
Commits redone!

app/models/support_user.rb
```diff
@@ -4,6 +4,8 @@ class SupportUser < ActiveRecord::Base
   validates :dfe_sign_in_uid, presence: true
   validates :email_address, presence: true, uniqueness: true
 
+  has_many :tokens, class_name: 'SupportUserToken', foreign_key: :model_id, dependent: :destroy
```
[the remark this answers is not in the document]
tvararu: Good call! I didn't want to call it `support_user_token` because RedundancyRedundancy

app/models/support_user_token.rb
tvararu: 👍

db/schema.rb
```diff
@@ -403,6 +403,16 @@
     t.index ["email_address"], name: "index_support_users_on_email_address", unique: true
   end
 
+  create_table "tokens", force: :cascade do |t|
```
tvararu, line 406: @tijmenb keep the `tokens` table name or rename it to `authentication_tokens`?

app/models/support_user_token.rb
tvararu: I'm struggling a bit to make this work, could you either refine your example or give me a shout when you have a minute to pair?

db/schema.rb
duncanjbrown: Must have been a rebase snafu after https://github.com/DFE-Digital/apply-for-teacher-training/pull/2336/files. I'll raise a PR

app/mailers/support_mailer.rb
```diff
@@ -8,4 +8,13 @@ def confirm_sign_in(support_user, device:)
       subject: 'New sign in to Support for Apply for teacher training',
     )
   end
+
+  def fallback_sign_in_email(support_user, token)
+    @magic_link = support_interface_authenticate_with_token_url(token: token)
```
tijmenb, line 13: Can we pass in the token and use the path helper in the view?

db/schema.rb
tijmenb: +1 for `authentication_tokens`

app/mailers/support_mailer.rb
tvararu: Fixed!

app/models/support_user_token.rb
tvararu: Done!

app/controllers/support_interface/sessions_controller.rb
```diff
@@ -4,12 +4,46 @@ class SessionsController < SupportInterfaceController
...
 
       redirect_to action: :new
     end
+
+    def sign_in_by_email
+      render_404 and return unless FeatureFlag.active?('dfe_sign_in_fallback')
+
+      support_user = SupportUser.find_by(email_address: params.dig(:support_user, :email_address).downcase.strip)
+
+      if support_user
```
tijmenb, line 23: Does this need the check on UID being present?

app/controllers/support_interface/sessions_controller.rb
```diff
@@ -4,12 +4,46 @@ class SessionsController < SupportInterfaceController
...
+        SupportInterface::MagicLinkAuthentication.send_token!(support_user: support_user)
+      end
+
+      redirect_to support_interface_check_your_email_path
+    end
+
+    def authenticate_with_token
+      redirect_to action: :new and return unless FeatureFlag.active?('dfe_sign_in_fallback')
+
+      support_user = SupportInterface::MagicLinkAuthentication.get_user_from_token!(token: params.fetch(:token))
```
tijmenb, line 33: This could use https://github.com/DFE-Digital/apply-for-teacher-training/pull/2362 as well.

app/controllers/support_interface/sessions_controller.rb
tvararu: No! I had it in initially, with a test to check that we correctly dodge if it isn't present, but the test failed because of this https://github.com/DFE-Digital/apply-for-teacher-training/blob/support-magic-link/app/models/support_user.rb#L4

So I removed both the spec + check

PR #2358, opened 2020-06-25
Provider emails offer on the course applied to not the course actually offered

steventux:
## Context

Changes to offered courses are not being accounted for in emails to providers.
This was spotted here https://becomingateacher.zendesk.com/agent/tickets/7704 where an alternative course option was offered.

Use `offered_option` instead of `course_option` and `offered_course` instead of `course` as these correctly check for the presence of the alternative offered course option and default to the original.

## Link to Trello card

https://trello.com/c/48hvol3c/2261-providers-are-emailed-saying-candidates-accepted-an-offer-on-the-course-applied-to-not-the-course-actually-offered

## Things to check


app/views/provider_mailer/application_withdrawn.text.erb
```diff
@@ -2,7 +2,7 @@ Dear <%= @provider_user.full_name || 'colleague' %>
 
 # Application withdrawn
 
-<%= @application_choice.application_form.full_name %> withdrew their application for <%= @application_choice.course_option.course.name_and_code %>.
+<%= @application_choice.application_form.full_name %> withdrew their application for <%= @application_choice.offered_option.course.name_and_code %>.
```
duncanjbrown, line 5: Could we use `offered_course` here?

app/views/provider_mailer/offer_accepted.text.erb
```diff
@@ -2,7 +2,7 @@ Dear <%= @provider_user.full_name || 'colleague' %>
 
 # Offer accepted
 
-<%= @application_choice.application_form.full_name %> has accepted your offer to study <%= @application_choice.course_option.course.name_and_code %>.
+<%= @application_choice.application_form.full_name %> has accepted your offer to study <%= @application_choice.offered_option.course.name_and_code %>.
```
duncanjbrown, line 5: And `offered_course` here too?

PR #2371, opened 2020-06-26
Show DSA signed date in the providers list in support

AgaDufrat:
## Context

This will help the transition team see at a glance which providers are yet to sign the Data Sharing Agreement.

Add a `DSA signed date` to the table on the providers page in support interface which shows.

<kbd>Screenshot 2020-06-25 at 16 48 41</kbd>

## Guidance to review

- visit http://localhost:3000/support/providers

Notes: not sure how to test this since the action of accepting DSA is done in provider interface and the results of it are are visible in support interface. We only seem to be testing if syncing from Find worked in support interface. 

## Link to Trello card

https://trello.com/c/pdAy717B/2299-show-dsa-signed-date-in-the-provider-list-in-support

## Things to check

### DIFF
--- a/app/controllers/support_interface/providers_controller.rb
+++ b/app/controllers/support_interface/providers_controller.rb
@@ -1,7 +1,9 @@
 module SupportInterface
   class ProvidersController < SupportInterfaceController
     def index
-      @providers = Provider.where(sync_courses: true).includes(:sites, :courses).order(:name)
+      @providers = Provider.where(sync_courses: true)
+        .includes(:sites, :courses, :provider_agreements)
+        .order(:name)
     end
 
     def other_providers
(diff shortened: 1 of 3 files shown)


mnacos:
Looking good. For testing, you could add a `see_providers_spec.rb` system spec, along the lines of `see_applications_spec.rb` or `see_candidates_spec.rb`.

Like you said, DSA signing is a provider interface action and this is a support interface thing, so you wouldn't test end-to-end, but I think it's fine to set up state programmatically and just verify the DSA dates are shown/correct. There are some convenient traits in `spec/factories.rb`, such as `create(:provider, :with_signed_agreement)` which should help. Doing this is not super-critical but would improve coverage of the support console.

PR #2373, opened 2020-06-26
Update degree subject form to use HESA data

malcolmbaig:
## Context
We need to allow users to choose from HESA options when they choose degree type/subject/institution/grade.

- Candidates can choose from HESA degree subjects when they choose their degree subject
- Candidates can still submit free text

See commits for implementation details.

Screenshot 2020-06-26 at 12 51 18
## Guidance to review

- Easily tested locally by adding a degree to an application form.

## Link to Trello card

https://trello.com/c/PYf7wdSf
## Things to check


config/initializers/hesa_degree_subjects.rb
```diff
@@ -1,7 +1,5 @@
 # rubocop:disable Style/StringLiterals
 HESA_DEGREE_SUBJECTS = [
-  [999999, "Not known"],
-  [999998, "Not applicable"],
```
tvararu, line 4: 👍

PR #2375, opened 2020-06-26
Update content suggesting choosing a course first

paulrobertlloyd:
## Context

From @EmmaFrith:

> The changes to the content reflect the following issues:
>
> - 'you'll be guided back' to UCAS - who is to say they were ever at UCAS? we should also use more active language where possible
> - 'by choosing a course first' - we should be explicit in what we want the user to do, rather than saying it in a round-about way 
> - UCAS should be UCAS Teacher Training on first mention because Apply isn't replacing the whole of UCAS
> - 'you can only apply for some courses using this service' - it is not clear to which service this refers because 'UCAS' immediately precedes this 
> - we use plain English where possible, so rather than 'make an application', we'd say 'do an application' 

Before:
Before

After:
After

## Link to Trello card

https://trello.com/c/JkxCQhyX

## Things to check

### DIFF
--- a/app/views/candidate_interface/application_form/before_you_start.html.erb
+++ b/app/views/candidate_interface/application_form/before_you_start.html.erb
@@ -1,19 +1,15 @@
 <h1 class="govuk-heading-xl">We suggest you choose a course first</h1>
 
 <div class="govuk-grid-row">
-
   <div class="govuk-grid-column-two-thirds">
-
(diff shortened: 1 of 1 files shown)


EmmaFrith:
there's a sentence missing from the content outlined here:https://docs.google.com/document/d/1JKGc3xbUnLWkrrUaQBQrA5XK4BsTf5FkSGyrTVXHx3Y/edit?usp=sharing @paulrobertlloyd

PR #2377, opened 2020-06-26
Show degree subject HESA code in support section

malcolmbaig:
## Context

We want to start displaying HESA codes in the support section to easily check if the feature is working as we roll it out.
- Add a subject column to the qualifications table and move the subject over from the degree column.
- Add the degree type and subject HESA codes (if available) to the table.

Screenshot_20200629_144656

## Link to Trello card
https://trello.com/c/PYf7wdSf

## Things to check

### DIFF
--- a/app/components/qualification_subject_component.rb
+++ b/app/components/qualification_subject_component.rb
@@ -1,0 +1,13 @@
+class QualificationSubjectComponent < ViewComponent::Base
+  def initialize(qualification:)
+    @qualification = qualification
+  end
+
+  def subject
+    @qualification.subject&.titleize
+  end
+
+  def hesa_code
+    @qualification.subject_hesa_code
+  end
+end
--- a/app/lib/hesa/degree_type.rb
+++ b/app/lib/hesa/degree_type.rb
@@ -1,6 +1,10 @@
 module Hesa
   class DegreeType
-    DegreeTypeStruct = Struct.new(:hesa_code, :abbreviation, :name, :level)
+    DegreeTypeStruct = Struct.new(:hesa_code, :abbreviation, :name, :level) do
+      def shortest_display_name
+        abbreviation || name
+      end
+    end
     UNDERGRADUATE_LEVELS = %i[bachelor master].freeze
 
     class << self
@@ -22,6 +26,10 @@ module Hesa
 
       def find_by_name(name)
         all.find { |type| type.name == name }
+      end
+
+      def find_by_hesa_code(code)
+        all.find { |type| type.hesa_code == code }
       end
     end
   end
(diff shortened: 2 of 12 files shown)


malcolmbaig:
Raising as a draft PR to run the visual change past @paulrobertlloyd . Code feedback still welcome.

@paulrobertlloyd does that screenshot above look OK as a way for us to start displaying the subject HESA code?

paulrobertlloyd:
Looks good, though I’d be more explicit, changing the column title to `Degree (HESA code)`.

paulrobertlloyd:
Was `Masters (Design)` manually entered? If we were to show `Degree (HESA code)` in the column heading, we’d want to avoid using brackets elsewhere.

malcolmbaig:
@paulrobertlloyd latest set of commits look like the following. The `Masters/Arts` entry is freeform input, the `BA/Product Design`  is HESA selections. The grade HESA codes haven't been implemented yet but will display in a similar way once they're in.

Screenshot_20200629_134518

paulrobertlloyd:
Love it! I’d keep the HESA codes as small grey text as you had initially, but other than that, 👍

malcolmbaig:
> +1 Code looks good and it works for me locally.
> 
> One question about this feature (not this PR): Do we need to update the test application generator to include HESA codes or does this just work?

We'd need to add them in. Currently HESA codes are only populated if degrees are updated via one of the forms (`DegreeTypeForm`, `DegreeSubjectForm`, and, imminently, `DegreeInstitutionForm`/`DegreeGradeForm`).

app/components/qualification_subject_component.rb
```diff
@@ -0,0 +1,13 @@
+class QualificationSubjectComponent < ViewComponent::Base
+  def initialize(qualification:)
+    @qualification = qualification
+  end
+
+  def subject
+    @qualification.subject&.titleize
```
duncanjbrown, line 7: Do we allow this to be nil? We validate :subject on GcseQualificationTypeForm, OtherQualificationForm and DegreeSubjectForm

app/lib/hesa/degree_type.rb
```diff
@@ -1,6 +1,10 @@
 module Hesa
   class DegreeType
-    DegreeTypeStruct = Struct.new(:hesa_code, :abbreviation, :name, :level)
+    DegreeTypeStruct = Struct.new(:hesa_code, :abbreviation, :name, :level) do
+      def shortest_display_name
```
duncanjbrown, line 4: Nice!

app/components/qualification_subject_component.rb
malcolmbaig: We do, yes. A candidate can abandon the 'add degree' flow partway through, meaning this table may need to render a partially-completed degree.

PR #2379, opened 2020-06-26
Update status labels and colours

paulrobertlloyd:
## Context

We should be consistent in how we assign colours to states, and ensure labels are clear an unambiguous.

This PR makes the following changes:

* Provider: ‘Withdrawn by us‘ status label → ‘Offer withdrawn’
* Candidate: ‘Pending’ status label → ‘Awaiting decision’
* Candidate: ‘Offer’ status label → ‘Offer made’
* Candidate: ‘Accepted’ status label → ‘Offer accepted’
* Candidate: ‘Declined’ status label → ‘Offer declined’
* Candidate: ‘Cancelled’ status label → ‘Application cancelled’
* Candidate: ‘Withdrawn’ status label → ‘Application withdrawn’
* All: ‘Declined’ status label → ‘Reference declined’
* All: Update reference status colours, using colours consistent with usage elsewhere
* All: Update (and, in some cases, add missing) application status colours, using colours consistent with usage elsewhere
* ~Support: Use red rather than yellow for full/unavailable course status colour~

Probably easiest to review these changes by looking at each commit separately.

## Guidance to review

Audit of existing label and colour assignments, and proposed changes can be found here: https://docs.google.com/spreadsheets/d/12xE9NKfsQpfnaROICUXTSJZf68E4qA479Sldhx1LxCU/edit#gid=812555744 

## Link to Trello card

https://trello.com/c/XuUfSfPB/1744-review-tag-colours-and-usage

## Things to check


paulrobertlloyd:
Will try and get some screenshots next week.

tijmenb:
This looks good! I'm going to add some refactoring on top to make it easier to keep these up to date.

tijmenb:
Added some commits.

- There are now status components for candidate, provider and support
- The components all have automatically generated component previews
- We raise an error if a colour is missing. This will be picked up by the automated tests because the preview components won't render.

Preview URLs:

- https://apply-for-te-status-lab-mz656u.herokuapp.com/rails/view_components/candidate_interface/application_status_tag_component
- https://apply-for-te-status-lab-mz656u.herokuapp.com/rails/view_components/provider_interface/application_status_tag_component
- https://apply-for-te-status-lab-mz656u.herokuapp.com/rails/view_components/support_interface/application_status_tag_component

paulrobertlloyd:
I ❤️  THIS! Couple of things…

* For some states, we may want to default to no colour, i.e. white text on dark blue (although @adamsilver has opinions on that); this however would throw an error.
* The `offer_withdrawn` state seems to be missing from all of these components (if I understand the states correctly)
* We’ve included some states that providers will never see, and so these pages throw an error. (`application_complete`, `awaiting_references`, `cancelled`, `unsubmitted`)
* Checking the previews, checking the `offer` state for providers throw an error, and then all subsequent previews error too.

tijmenb:
Oh nice catch! "This will be picked up by the automated tests because the preview components won't render." was wrong because I was confused with the automated testing we have for the email previews.

> For some states, we may want to default to no colour, i.e. white text on dark blue (although @adamsilver has opinions on that); this however would throw an error.

I've added an explicit `:default` option for these cases (only `enrolled` afaic?).

> The offer_withdrawn state seems to be missing from all of these components (if I understand the states correctly)

This is very tricky! It's not really a status so it's too hard to auto-generate these tests for now. - https://trello.com/c/wNkQdYUH. I'll look at this separately.

> We’ve included some states that providers will never see, and so these pages throw an error. (application_complete, awaiting_references, cancelled, unsubmitted)

Added them explicitly in the component.

> Checking the previews, checking the offer state for providers throws an error, and then all subsequent previews error too.

I _may_ have fixed this.

paulrobertlloyd:
Thanks @tijmenb, great to have these statuses all in one place and reviewable. :shipit:

PR #2383, opened 2020-06-29
Choose a course first copy tweak

paulrobertlloyd:
## Context

Adds missed paragraph from the change made in #2375:

> If your course is not available, we’ll send you to UCAS to apply there.


EmmaFrith:
thanks @paulrobertlloyd

PR #2384, opened 2020-06-29
Update content on start page

paulrobertlloyd:
## Context

* Suggested content changes from @EmmaFrith 
* No longer open links on this page in new tabs

Before:

Before

After:

After
## Link to Trello card

https://trello.com/c/gp4IKjeb/1718-start-page-edit

### DIFF
--- a/app/views/candidate_interface/start_page/show.html.erb
+++ b/app/views/candidate_interface/start_page/show.html.erb
@@ -2,36 +2,26 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-xl">
-      <%= service_name %>
-    </h1>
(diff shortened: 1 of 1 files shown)


EmmaFrith:
@paulrobertlloyd - just a couple of small things which aren't the same as outlined here: https://docs.google.com/document/d/1PnKkbWbh4QgowOnS0whK8Wny5xxhGIeXJpuv3GNMryA/edit?usp=sharing

1. can you change the link text to 'Learn more about training to teach if **you’re disabled**' - rather than 'have a disability'
2. can you remove the full stops from the 2 sentences which are entirely link text

See here for examples on how to use punctuation with links on gov.uk: https://www.gov.uk/government/how-government-works. When the whole sentence is a link, and says something like 'Read more...' 'See...', we don't need the full stops.

Thanks!

PR #2388, opened 2020-06-29
Explicitly assign permissions from pairing

steventux:
## Context

We retrieve permissions in pairs for each provider relationship. The underlying query doesn't have a consistent order so
be explicit and use the type when setting this variables.

Assign the local view variables based on the permissions type as this is more explicit and obvious than relying on query ordering.

## Link to Trello card

https://trello.com/c/k5hlzAcA/2348-organisation-pages-permissions-are-displayed-in-reverse-for-training-and-ratifying-providers

## Things to check

### DIFF
--- a/app/controllers/provider_interface/organisations_controller.rb
+++ b/app/controllers/provider_interface/organisations_controller.rb
@@ -7,14 +7,14 @@ module ProviderInterface
     end
 
     def show
-      @ratifying_permissions = ProviderRelationshipPermissions
-        .includes(:training_provider, :ratifying_provider)
-        .where(ratifying_provider: provider)
-        .group_by(&:training_provider_id)
-      @training_permissions = ProviderRelationshipPermissions
-        .includes(:training_provider, :ratifying_provider)
-        .where(training_provider: provider)
-        .group_by(&:ratifying_provider_id)
+      scope = ProviderRelationshipPermissions.includes(:training_provider, :ratifying_provider)
+
+      @ratifying_permissions = ProviderRelationshipPermissionsPair.pairs_from_collection(
+        scope.where(ratifying_provider: provider),
+      )
+      @training_permissions = ProviderRelationshipPermissionsPair.pairs_from_collection(
+        scope.where(training_provider: provider),
+      )
     end
 
   private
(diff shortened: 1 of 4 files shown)


duncanjbrown:
Fun bug! Is the fact that we're depending on the order of an unsorted pair a signal that we ought to build the data structure differently? Maybe a struct or a hash would be more usable?

steventux:
@duncanjbrown Yes the underlying data structure from the `group_by` is sub-optimal:
```
{ 1010 => [<AccreditedBodyPermissions #32ew6fw49>, <TrainingProviderPermissions #327r6g23>], 1012 => [... }
```
I couldn't find a way to do this pairing in SQL at the time, that's not to say it can't be done.
The alternative 'fix' I could think of was to add an `order(:type)` clause to the queries but again it doesn't really deal with the problem of how we get the pairs of permissions.
I suppose another option is we could make more trips to the database for the pairs. Suggestions welcome.

duncanjbrown:
Could reduce to a hash immediately on getting it out of the db. Bit awkward to set the keys but nicer downstream

```ruby
@ratifying_permissions = ProviderInterface::ProviderRelationshipPermissions
  .includes(:training_provider, :ratifying_provider)
  .where(ratifying_provider: provider)
  .group(:id, :training_provider_id).each_with_object({}) do |permission, hash|
    key = permission.type.match(/Training/) ? :training_provider : :ratifying_provider
    hash[key] = permission
  end
```

steventux:
@duncanjbrown I've attempted to tidy up how we pair permissions here https://github.com/DFE-Digital/apply-for-teacher-training/pull/2388/commits/90315ce80f305818f135476f61ef3441a405337e

app/helpers/permissions_helper.rb
```diff
@@ -0,0 +1,9 @@
+module PermissionsHelper
```
duncanjbrown, line 1: Do we need this any more?

steventux: @duncanjbrown thanks I've removed the old helper implementation and squashed down to https://github.com/DFE-Digital/apply-for-teacher-training/pull/2388/commits/73f442863a0ab3648c1584bd1a5e003e31c0257b

PR #2390, opened 2020-06-29
Candidate journey tracking report (CSV download)

stevehook:
## Context

The data extract we currently have from the service is helpful but there are gaps in the overall journey that would be helpful to report on, this would then improve our Performance Analysis capability, particularly in relation to apply again.

- [x] Added link to new report (CSV download) on the Support Performance page
- [x] Service classes `CandidateJourneyTrackingExport` and `CandidateJourneyTracker`

image

Include the following timestamps:

- [x] `form_not_started`
- [x] `form_started_and_not_submitted`
- [x] `submitted_and_awaiting_references`
- [x] `reference_1_received`
- [x] `reference_2_received`
- [x] `reference_reminder_email_sent`
- [x] `new_reference_request_email_sent`
- [x] `new_reference_added`
- [x] `references_complete` - from audit trail when `ApplicationForm#references_completed` gets set to true? (can happen more than once)
- [x] `waiting_to_be_sent_to_provider` - when application progresses to `application_complete`.
- [x] `application_sent_to_provider` - `ApplicationChoice#sent_to_provider_at`
- ~`awaiting_decision` - duplicate of `sent_to_provider_at`?)~
- [x] `rbd_date` - `ApplicationChoice#sent_to_provider_at`
- [x] `rbd_reminder_sent` - Is this the `chase_provider_decision` email? `ChaserSent#provider_decision_request` ?
- [x] `application_rbd` - Combination of `ApplicationChoice#rejected_at` and `rejected_by_default`
- [x] `provider_decision` (Reject/Offer) - `ApplicationChoice#rejected_at` or `ApplicationChoice#offered_at`
- [x] `offer_made` awaiting decision from candidate - `ApplicationChoice#offered_at`
- [ ] `email_sent_to_candidate` - I'm not sure how to evaluate this one. I thought it could be extracted from the email log but the `Email` is only associated with `ApplicationForm` so it can't be tied to a specific `ApplicationChoice` as far as I can see (without reading the subject or body of the email). 
- [x] `dbd_date` - `ApplicationChoice#decline_by_default_at`
- [x] `dbd_reminder_email` - `ChaserSent#chaser_type`
- [x] `candidate_decision` (accept/decline) - `ApplicationChoice#accepted_at` or `ApplicationChoice#declined_at`
- [x] `offer_declined` - `ApplicationChoice#declined_at`
- [x] `offer_accepted` - `ApplicationChoice#accepted_at`
- ~`email_sent_to_candidate_2` - leaving this for now~
- [x] `pending_conditions` - from audit trail
- [x] `conditions_outcome` - from audit trail
- [x] `conditions_met` - from audit trail
- [x] `conditions_not_met` - from audit trail
- [x] `enrolled` - from audit trail
- [x] `ended_without_success` - from audit trail?
- [ ] `send_rejection_email` - I'm not sure how to evaluate this one. See `email_sent_to_candidate` above.

## Guidance to review

- There are a lot of data points in this report that are a little painful/inefficient to extract (and a couple that I don't think can be extracted) from the data we have available. We might want to review these and consider recording explicit timestamps in the relevant business logic as we do with a number of 'events' e.g. `ApplicationForm#submitted_at`, `ApplicationChoice#rejected_at`.
- It would be useful for a couple of the data points to have an `Email#application_choice` association (in addition to the existing `Email#application_form` association) so that we could use the email log to extract certain dates. Might be hard (or impossible in some cases) to backfill though.
- This report is potentially going to be very slow, and might cause knock-on effects to users other than the person invoking the report. In future we should consider either limiting the scope by returning say a month's worth of applications at a time or pre-calculating the results in a background process. For now all I'm doing to mitigate this is to use `find_each` to iterate the result set in batches and some `includes` to eager load associations.
- There are a few data points in the report that refer to events that could happen more than once, e.g. `references_complete` would happen more than once if a candidate needed to add an additional reference. In such cases I've assumed that we are interested in the first of any matching events.

## Link to Trello card

https://trello.com/c/Fde3OX6R/1704-dev-data-extract-from-service-for-pa

## Things to check

### DIFF
--- a/app/controllers/support_interface/performance_controller.rb
+++ b/app/controllers/support_interface/performance_controller.rb
@@ -19,8 +19,8 @@ module SupportInterface
     end
 
     def application_choice_timings
-      # TODO: Implementation to follow
-      csv = to_csv([])
+      application_choices = SupportInterface::ApplicationChoiceTimingsExport.new.application_choices
+      csv = to_csv(application_choices)
 
       send_data csv, filename: "application-choices-timings-#{Time.zone.today}.csv", disposition: :attachment
     end
--- a/app/controllers/support_interface/performance_controller.rb
+++ b/app/controllers/support_interface/performance_controller.rb
@@ -18,6 +18,13 @@ module SupportInterface
       send_data csv, filename: "submitted-application-choices-#{Time.zone.today}.csv", disposition: :attachment
     end
 
+    def application_choice_timings
+      # TODO: Implementation to follow
+      csv = to_csv([])
+
+      send_data csv, filename: "application-choices-timings-#{Time.zone.today}.csv", disposition: :attachment
+    end
+
     def providers_export
       providers = SupportInterface::ProvidersExport.new.providers
       csv = to_csv(providers)
--- a/app/controllers/support_interface/performance_controller.rb
+++ b/app/controllers/support_interface/performance_controller.rb
@@ -18,11 +18,11 @@ module SupportInterface
       send_data csv, filename: "submitted-application-choices-#{Time.zone.today}.csv", disposition: :attachment
     end
 
-    def application_choice_timings
-      application_choices = SupportInterface::ApplicationChoiceTimingsExport.new.application_choices
+    def candidate_journey_tracking
+      application_choices = SupportInterface::CandidateJourneyTrackingExport.new.application_choices
       csv = to_csv(application_choices)
 
-      send_data csv, filename: "application-choices-timings-#{Time.zone.today}.csv", disposition: :attachment
+      send_data csv, filename: "candidate-journey-tracking-#{Time.zone.today}.csv", disposition: :attachment
     end
 
     def providers_export
--- a/app/services/support_interface/candidate_journey_tracker.rb
+++ b/app/services/support_interface/candidate_journey_tracker.rb
@@ -112,6 +112,13 @@ module SupportInterface
       @application_choice.enrolled_at
     end
 
+    def ended_without_success
+      earliest_update_audit_for(
+        @application_choice,
+        status: ApplicationStateChange::UNSUCCESSFUL_END_STATES,
+      )
+    end
+
   private
 
     def all_references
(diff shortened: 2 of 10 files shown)


app/controllers/support_interface/performance_controller.rb
```diff
@@ -18,6 +18,13 @@ def submitted_application_choices
       send_data csv, filename: "submitted-application-choices-#{Time.zone.today}.csv", disposition: :attachment
     end
 
+    def candidate_journey_tracking
```
tijmenb, line 21: If we add this then we can delete the `application_timings` (ApplicationChoicesExport) export, it serves the same purpose.

stevehook: 👍 I'd like to do this as a follow-up PR. This one has already dragged on for too long.

stevehook: @tijmenb Mike has asked us to keep the old report - he needs it for the short-term at least.

spec/services/support_interface/candidate_journey_tracker_spec.rb
```diff
@@ -0,0 +1,562 @@
...
+    it 'returns nil if the application form has not been submitted' do
+      application_form = create(:application_form)
+      application_choice = create(:application_choice, status: :unsubmitted, application_form: application_form)
+
+      expect(described_class.new(application_choice).submitted_and_awaiting_references).to be_nil
+    end
+  end
+
+  describe '#reference_1_received' do
+    it 'returns nil if not references were received' do
```
davidgisbey, line 62:
```suggestion
    it 'returns nil if no references were received' do
```

app/services/support_interface/candidate_journey_tracker.rb
```diff
@@ -0,0 +1,195 @@
...
+            change = audit.audited_changes[attribute.to_s]
+            if value.is_a?(Array)
+              change && value.include?(change[1])
+            else
+              change && change[1] == value
+            end
+          end
+      end
+
+      audits.map(&:created_at).min
```
davidgisbey, line 182: This was really smart 👌

PR #2402, opened 2020-06-30
Disallow UCAS data download in production

tijmenb:
##  Context

We allow support users to check the UCAS export via a download. This is currently behind a feature flag, so it would still allow it to be used on production if ever activated. We don't want that because it contains sensitive info.

Change the feature flag into a check on environment.

## Link to Trello card

## Things to check

### DIFF
--- a/app/controllers/support_interface/performance_controller.rb
+++ b/app/controllers/support_interface/performance_controller.rb
@@ -47,10 +47,7 @@ module SupportInterface
 
     def applications_export_for_ucas
       if HostingEnvironment.production?
-        # The unauthorized page expects an instance var that's only set in
-        # the dfe_sign_in_controller
-        @dfe_sign_in_user = dfe_sign_in_user
-        render 'support_interface/unauthorized', status: :forbidden
+        render_404
       else
         applications = ApplicationsExportForUCAS.new.applications
         header_row = ApplicationsExportForUCAS.csv_header(applications)
--- a/app/controllers/support_interface/performance_controller.rb
+++ b/app/controllers/support_interface/performance_controller.rb
@@ -46,17 +46,17 @@ module SupportInterface
     end
 
     def applications_export_for_ucas
-      if FeatureFlag.active?('download_dataset1_from_support_page')
+      if HostingEnvironment.production?
+        # The unauthorized page expects an instance var that's only set in
+        # the dfe_sign_in_controller
+        @dfe_sign_in_user = dfe_sign_in_user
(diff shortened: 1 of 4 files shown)


app/controllers/support_interface/performance_controller.rb
```diff
@@ -46,17 +46,17 @@ def candidate_survey
     end
 
     def applications_export_for_ucas
-      if FeatureFlag.active?('download_dataset1_from_support_page')
+      if HostingEnvironment.production?
+        # The unauthorized page expects an instance var that's only set in
+        # the dfe_sign_in_controller
```
duncanjbrown, line 51: This is a thing because we're overloading the current "unauthorized" page. I think the right thing to do would be name that properly and create a generic "unauthorized" page for generic unauthorized actions?

spec/system/support_interface/exporting_applications_for_ucas_spec.rb
```diff
@@ -36,12 +36,12 @@ def and_i_visit_the_service_performance_page
     visit support_interface_performance_path
   end
 
-  def when_the_feature_is_not_enabled
-    FeatureFlag.deactivate('download_dataset1_from_support_page')
+  def when_i_am_on_production
+    ENV['HOSTING_ENVIRONMENT_NAME'] = 'production'
```
duncanjbrown, line 40: I think this and the other `ENV` assignment are unsafe because they might hang around once the test is done? `ClimateControl` could do the cleanup for us.

tijmenb: Can I use `ClimateControl` without a block? Not sure how I'd use it otherwise in these tests.

app/controllers/support_interface/performance_controller.rb
tijmenb: Raising a 404 now, which is better anyway.

spec/system/support_interface/exporting_applications_for_ucas_spec.rb
duncanjbrown: I don't think so — we'd have to indent in the body of the spec.

How about something like this?

```ruby
around do |e|
  existing_env = ENV
  e.run
  ENV = existing_env
end
```

tijmenb: I've checked, and tests that rely on a specific value already use climate control or set it explicitly, so I think this actually won't cause an issue?

PR #2404, opened 2020-06-30
Give dev-provider and dev-support names

duncanjbrown:
## Context

Noticed as part of reviewing access controls.

Give them names when we create them in `setup_local_dev_data`. This means they display properly in the manage-users UI

Screenshot 2020-06-30 at 12 46 15

## Link to Trello card

https://trello.com/c/sCTl8idK/2333-review-ac

### DIFF
--- a/lib/tasks/local_dev.rake
+++ b/lib/tasks/local_dev.rake
@@ -1,15 +1,30 @@
 desc 'Set up your local development environment with data from Find'
 task setup_local_dev_data: %i[environment copy_feature_flags_from_production sync_dev_providers_and_open_courses] do
   puts 'Creating a provider-only user with DfE Sign-in UID `dev-provider` and email `provider@example.com`...'
-  ProviderUser.find_or_create_by!(dfe_sign_in_uid: 'dev-provider', email_address: 'provider@example.com') do |u|
+  ProviderUser.create!(
+    dfe_sign_in_uid: 'dev-provider',
+    email_address: 'provider@example.com',
+    first_name: 'Peter',
+    last_name: 'Rovider',
+  ) do |u|
     u.providers = Provider.where(code: '1JA').all
   end
 
   puts 'Creating a support & provider user with DfE Sign-in UID `dev-support` and email `support@example.com`...'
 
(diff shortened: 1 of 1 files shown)


lib/tasks/local_dev.rake
```diff
@@ -1,15 +1,30 @@
...
   end
 
   puts 'Creating a support & provider user with DfE Sign-in UID `dev-support` and email `support@example.com`...'
 
-  SupportUser.find_or_create_by!(dfe_sign_in_uid: 'dev-support', email_address: 'support@example.com')
+  SupportUser.find_or_create_by!(
+    dfe_sign_in_uid: 'dev-support',
+    email_address: 'support@example.com',
+    first_name: 'Apply',
+    last_name: 'Support',
```
tijmenb, line 19: Can these users ever exist already? In that case this would crash because it'll try to create a new record with an existing email/dsi-uid.

tvararu: I don't 🤔 not sure why you'd run this without a fresh db

tijmenb: Fair. Can we change it to `create!` then?

lib/tasks/local_dev.rake
```diff
@@ -1,15 +1,30 @@
 desc 'Set up your local development environment with data from Find'
 task setup_local_dev_data: %i[environment copy_feature_flags_from_production sync_dev_providers_and_open_courses] do
   puts 'Creating a provider-only user with DfE Sign-in UID `dev-provider` and email `provider@example.com`...'
-  ProviderUser.find_or_create_by!(dfe_sign_in_uid: 'dev-provider', email_address: 'provider@example.com') do |u|
+  ProviderUser.find_or_create_by!(
+    dfe_sign_in_uid: 'dev-provider',
+    email_address: 'provider@example.com',
+    first_name: 'Apply',
+    last_name: 'Provider',
```
tijmenb, line 8: As discussed, a real-sounding name would avoid confusion in the UI. 

Peter Rovider and Susan Upport?

PR #2406, opened 2020-06-30
Introduce a real "offer withdrawn" status

tijmenb:
## Context

When a provider has made an offer, they can withdraw it. The application is then in a "rejected" state. This was for practical reasons (we didn't have to have a new event and state for this process) and for backwards compatibility in the API (a new status constitutes a breaking change). However, since the functionality was introduced it turns out we want to treat applications with a withdraw offer different from rejected applications - we've introduced the `offer_withdrawn` state as a kind of virtual state dependent on the `offer_withdrawn_at` attribute. This makes things inconsistent, and introduces a lot of `if` statements everywhere. Recently, it made https://github.com/DFE-Digital/apply-for-teacher-training/pull/2379 weird.

Introduce a real "offer withdrawn" state, and convert any existing applications to it. Make an exception for the API - we'll expose the real status in v2.

## Guidance to review

Makes sense?

## Link to Trello card

https://trello.com/b/aRydsKwk/apply-tech

## Things to check

### DIFF
--- a/app/services/filter_application_choices_for_providers.rb
+++ b/app/services/filter_application_choices_for_providers.rb
@@ -17,15 +17,7 @@ class FilterApplicationChoicesForProviders
     def status(application_choices, statuses)
       return application_choices if statuses.blank?
 
-      filtered_application_choices = application_choices.where(status: statuses)
-
-      if statuses.include?('offer_withdrawn')
-        filtered_application_choices = filtered_application_choices.or(application_choices.offer_withdrawn)
-      elsif statuses.include?('rejected')
-        filtered_application_choices = filtered_application_choices.where(offer_withdrawn_at: nil)
-      end
-
-      filtered_application_choices
+      application_choices.where(status: statuses)
     end
 
     def provider(application_choices, providers)
--- a/config/locales/application_states.yml
+++ b/config/locales/application_states.yml
@@ -158,6 +158,13 @@ en:
       description: |
         The provider has rejected the candidate’s application.
 
+    offer_withdrawn:
+      name: Offer withdrawn
+      description: |
+        The provider has made an offer but withdrew it later.
+
+        This is called `rejected` in the API.
+
     withdrawn:
       name: Withdrawn
       description: |
@@ -350,6 +357,15 @@ en:
         - provider_mailer-application_withdrawn
 
     rejected-make_offer:
+      name: Provider makes offer
+      by: provider
+      description: ""
+      emails:
+        - candidate_mailer-new_offer_single_offer
+        - candidate_mailer-new_offer_multiple_offers
+        - candidate_mailer-new_offer_decisions_pending
+
+    offer_withdrawn-make_offer:
       name: Provider makes offer
       by: provider
       description: ""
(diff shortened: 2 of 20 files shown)


app/services/filter_application_choices_for_providers.rb
```diff
@@ -17,15 +17,7 @@ def search(application_choices, candidates_name)
...
-      filtered_application_choices = application_choices.where(status: statuses)
-
-      if statuses.include?('offer_withdrawn')
-        filtered_application_choices = filtered_application_choices.or(application_choices.offer_withdrawn)
-      elsif statuses.include?('rejected')
-        filtered_application_choices = filtered_application_choices.where(offer_withdrawn_at: nil)
-      end
-
-      filtered_application_choices
+      application_choices.where(status: statuses)
```
duncanjbrown, line 20: 🎉

db/migrate/20200630141828_update_application_status_to_offer_withdrawn.rb
```diff
@@ -0,0 +1,13 @@
+class UpdateApplicationStatusToOfferWithdrawn < ActiveRecord::Migration[6.0]
+  def change
+    ApplicationChoice
+      .where(status: 'rejected')
```
duncanjbrown, line 4: This is not watertight as it's possible for an offer to have transitioned from offer to withdrawal to offer to rejection

app/services/filter_application_choices_for_providers.rb
duncanjbrown: We might be able to dial down the spec for this file now that the logic is simpler

config/locales/application_states.yml
```diff
@@ -158,6 +158,11 @@ en:
       description: |
         The provider has rejected the candidate’s application.
 
+    offer_withdrawn:
+      name: Rejected
```
duncanjbrown, line 162: Should this be "Offer withdrawn"?

app/services/filter_application_choices_for_providers.rb
tijmenb: Done!

db/migrate/20200630141828_update_application_status_to_offer_withdrawn.rb
tijmenb: Is that true? Offer to rejection isn't possible, just offer to offer withdrawn.

Otherwise it's an existing bug in the scope & attribute:

https://github.com/DFE-Digital/apply-for-teacher-training/blob/2ebbc1127650f277fb6795228ed70d95bd81206b/app/models/application_choice.rb#L34-L38

config/locales/application_states.yml
tijmenb: Done - also mentioned the difference in API.

db/migrate/20200630141828_update_application_status_to_offer_withdrawn.rb
duncanjbrown: Yes, you're right!

PR #2408, opened 2020-06-30
Fix a cycle-related bug in GetApplicationChoicesForProviders

duncanjbrown:
## Context

We made this query cycle-aware... but we missed a bit.

See commit.

## Link to Trello card

Spotted while working on https://trello.com/c/sCTl8idK/2333-review-access-controls-in-preparation-for-launch

## Things to check

### DIFF
--- a/app/queries/get_application_choices_for_providers.rb
+++ b/app/queries/get_application_choices_for_providers.rb
@@ -19,7 +19,7 @@ class GetApplicationChoicesForProviders
     ApplicationChoice.includes(*includes)
       .where('courses.provider_id' => providers, 'courses.recruitment_cycle_year' => RecruitmentCycle.current_year)
       .or(ApplicationChoice.includes(*includes)
-        .where('courses.accredited_provider_id' => providers))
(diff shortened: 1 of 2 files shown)


app/queries/get_application_choices_for_providers.rb
```diff
@@ -19,7 +19,7 @@ def self.call(providers:)
     ApplicationChoice.includes(*includes)
       .where('courses.provider_id' => providers, 'courses.recruitment_cycle_year' => RecruitmentCycle.current_year)
       .or(ApplicationChoice.includes(*includes)
-        .where('courses.accredited_provider_id' => providers))
+        .where('courses.accredited_provider_id' => providers, 'courses.recruitment_cycle_year' => RecruitmentCycle.current_year))
```
tijmenb, line 22: I'd imagine `GetApplicationChoicesForProviders` to be called for previous cycles as well. Should this be a parameter, set by the controller?

duncanjbrown: It will be soon

duncanjbrown: When we implement this: https://manage-applications-beta.herokuapp.com/switch-cycle

PR #2413, opened 2020-07-01
Account for duplicates which may exist for RatifyingProviderPermissions

steventux:
## Context

Records created by the sync from Find may have used the new STI type `RatifyingProviderPermissions` effectively creating new records which will have the same foreign keys as existing records using the STI type `ProviderInterface::AccreditedBodyPermissions`
Attempts to update them to the use the new STI type will fail on the `[type, ratifying_provider_id, training_provider_id]` uniqueness constraint. 
This currently causes the migration to fail.

This commit checks for the existence of a newer record (using the RatifyingProviderPermissions type) and either disgards the old
record or updates it depending on whether a 'duplicate' exists.

## Things to check


steventux:
@duncanjbrown running this locally I found I needed to specify the alias https://github.com/DFE-Digital/apply-for-teacher-training/pull/2413/commits/33d51c1edd8a404a1f255d7771ef449902c365fb#diff-c29047a57f77b0b8298664710ca8fc89R1 to avoid `NameError: uninitialized constant ProviderInterface::AccreditedBodyPermissions`.

PR #2416, opened 2020-07-01
API docs reflect our change-offer behaviour

duncanjbrown:
It's not currently possible to change the conditions of an offer without changing the course.

This is a bit weird and I think it's a defect in the system, but we'll tackle it when we cover the individual conditions workflow.

## Context

UCB's developer was confused by the docs, which implied he would be able to change the conditions on an existing offer.

Clarify this behaviour in the API docs.

## Link to Trello card

https://trello.com/c/LTwAEdTO/2384-update-api-documentation-on-changing-offers-to-clarify-how-changing-an-offer-works

## Things to check

### DIFF
--- a/app/views/api_docs/pages/release_notes.md
+++ b/app/views/api_docs/pages/release_notes.md
@@ -1,3 +1,10 @@
+### 2nd July 2020
+
+The documentation around the `/offer` endpoint has been clarified to show that:
+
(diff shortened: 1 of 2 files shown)


duncanjbrown:
Merging bc the flaky test is fixed in #2419

PR #2426, opened 2020-07-02
Sort by reject by default date

steventux:

Adds sorting options to allow sorting applications by RBD date.

#### Sort options select

image

#### Formatting of application dates...

image

## Link to Trello card

https://trello.com/c/kUuiiWro/2373-implement-rbd-sorting

## Things to check

### DIFF
--- a/app/components/provider_interface/application_card_component.rb
+++ b/app/components/provider_interface/application_card_component.rb
@@ -20,6 +20,7 @@ module ProviderInterface
 
     def contextual_date
       return changed_at_date unless sort_by && sort_by == 'Days left to respond'
+      return changed_at_date unless application_choice.status == 'awaiting_provider_decision'
       return changed_at_date unless reject_by_default_in_future?
 
       return '1 day to respond' if days_to_respond == 1
--- a/app/components/provider_interface/application_card_component.rb
+++ b/app/components/provider_interface/application_card_component.rb
@@ -4,9 +4,9 @@ module ProviderInterface
 
     attr_accessor :accredited_provider, :application_choice, :application_choice_path,
                   :candidate_name, :course_name_and_code, :course_provider_name, :changed_at,
-                  :most_recent_note, :site_name_and_code
+                  :most_recent_note, :site_name_and_code, :sort_by
 
-    def initialize(application_choice:)
+    def initialize(application_choice:, sort_by: nil)
       @accredited_provider = application_choice.accredited_provider
       @application_choice = application_choice
       @candidate_name = application_choice.application_form.full_name
@@ -15,6 +15,32 @@ module ProviderInterface
       @changed_at = application_choice.updated_at.to_s(:govuk_date_and_time)
       @site_name_and_code = application_choice.site.name_and_code
       @most_recent_note = application_choice.notes.order('created_at DESC').first
+      @sort_by = sort_by
+    end
+
+    def contextual_date
+      return changed_at_date unless sort_by && sort_by == 'Days left to respond'
+      return changed_at_date unless reject_by_default_in_future?
+
+      return '1 day to respond' if days_to_respond == 1
+      return 'Less than 1 day to respond' if days_to_respond < 1
+
+      "#{days_to_respond} days to respond"
+    end
+
+  private
+
+    def changed_at_date
+      "Changed #{changed_at}"
+    end
+
+    def reject_by_default_in_future?
+      application_choice.reject_by_default_at.present? &&
+        application_choice.reject_by_default_at > Time.current
+    end
+
+    def days_to_respond
+      @days_to_respond ||= (application_choice.reject_by_default_at.to_date - Date.current).to_i
     end
   end
 end
--- a/app/components/provider_interface/application_card_component.rb
+++ b/app/components/provider_interface/application_card_component.rb
@@ -4,9 +4,9 @@ module ProviderInterface
 
     attr_accessor :accredited_provider, :application_choice, :application_choice_path,
                   :candidate_name, :course_name_and_code, :course_provider_name, :changed_at,
-                  :most_recent_note, :site_name_and_code, :sort_by
+                  :most_recent_note, :site_name_and_code, :show_date
 
-    def initialize(application_choice:, sort_by: nil)
+    def initialize(application_choice:, show_date: :updated_at)
       @accredited_provider = application_choice.accredited_provider
       @application_choice = application_choice
       @candidate_name = application_choice.application_form.full_name
@@ -15,11 +15,11 @@ module ProviderInterface
       @changed_at = application_choice.updated_at.to_s(:govuk_date_and_time)
       @site_name_and_code = application_choice.site.name_and_code
       @most_recent_note = application_choice.notes.order('created_at DESC').first
-      @sort_by = sort_by
+      @show_date = show_date
     end
 
     def contextual_date
-      return changed_at_date unless sort_by && sort_by == 'Days left to respond'
+      return changed_at_date unless show_date == :reject_by_default_at
       return changed_at_date unless application_choice.status == 'awaiting_provider_decision'
       return changed_at_date unless reject_by_default_in_future?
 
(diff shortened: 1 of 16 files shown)


app/components/provider_interface/application_card_component.rb
```diff
@@ -4,9 +4,9 @@ class ApplicationCardComponent < ViewComponent::Base
 
     attr_accessor :accredited_provider, :application_choice, :application_choice_path,
                   :candidate_name, :course_name_and_code, :course_provider_name, :changed_at,
-                  :most_recent_note, :site_name_and_code
+                  :most_recent_note, :site_name_and_code, :sort_by
 
-    def initialize(application_choice:)
+    def initialize(application_choice:, sort_by: nil)
```
tijmenb, line 9: We could rename this to conceptually decouple the "sorting" from how we display the date:
 
```suggestion
    def initialize(application_choice:, show_date:)
```

Where `show_date` can be either `:last_updated` or `:reject_by_default`. This would also removes the magic string in the `contextual_date` method.

spec/components/provider_interface/application_card_component_spec.rb
```diff
@@ -109,4 +130,78 @@
...
+      build_stubbed(
+        :application_choice,
+        reject_by_default_at: reject_by_default_at,
+        updated_at: updated_at,
+        course_option: course_option,
+        status: status,
+      )
+    end
+
+    subject(:component) { described_class.new(application_choice: application_choice, show_date: show_date) }
```
duncanjbrown, line 149: could the subject be `contextual_date`?

spec/components/provider_interface/application_card_component_spec.rb
```diff
@@ -109,4 +130,78 @@
...
+      it 'presents the last changed date' do
+        expect(component.contextual_date).to eq('Changed  2 June 2020 at  9:05am')
+      end
+    end
+
+    context 'when reject_by_default_at is less than a day away' do
+      let(:reject_by_default_at) { 1.hour.from_now }
+
+      it 'presents "Less than 1 day to respond"' do
+        expect(described_class.new(application_choice: application_choice, show_date: show_date).contextual_date).to eq('Less than 1 day to respond')
```
duncanjbrown, line 187: I found these still passed relying on the subject. With `contextual_date` in the subject, this could be

```ruby
      it 'presents "Less than 1 day to respond"' do
        expect(contextual_date).to eq('Less than 1 day to respond')
      end
```

or even

```ruby
it { is_expected.to eq 'Less than 1 day to respond' }
```

app/components/provider_interface/application_card_component.rb
steventux: thanks @tijmenb amended here https://github.com/DFE-Digital/apply-for-teacher-training/pull/2426/commits/6ccfeb16667e05507b46545f2bce0c014f55ee78

app/views/provider_interface/application_choices/index.html.erb
```diff
@@ -45,7 +45,10 @@
     <% if @application_choices.any? %>
       <div class="app-application-cards">
         <% @application_choices.each_with_index do |choice, index| %>
-          <%= render ProviderInterface::ApplicationCardComponent.new(application_choice: choice, sort_by: @page_state.sort_by) %>
+          <%= render ProviderInterface::ApplicationCardComponent.new(
+            application_choice: choice,
+            show_date: @page_state.sort_by == 'Days left to respond' ? :reject_by_default_at : :updated_at,
```
duncanjbrown, line 50: Would it be tidy to map this in `PageState`? Then it's the only place for this mapping to live.

spec/components/provider_interface/application_card_component_spec.rb
```diff
@@ -109,4 +130,78 @@
...
+      it 'presents the last changed date' do
+        expect(component.contextual_date).to eq('Changed  2 June 2020 at  9:05am')
+      end
+    end
+
+    context 'when reject_by_default_at is nil' do
+      let(:reject_by_default_at) { nil }
+
+      it 'presents the last changed date' do
+        expect(component.contextual_date).to eq('Changed  2 June 2020 at  9:05am')
```
duncanjbrown, line 171: Is it right that `govuk_date_and_time` puts spaces around single-digit dates and hours?

spec/components/provider_interface/sort_order_component_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    let(:params) do
+      ActionController::Parameters.new({
+        'status' => %w[awaiting_provider_decision pending_conditions],
+        'candidate_name' => 'Susan',
+      })
+    end
+    let(:page_state) do
+      ProviderInterface::ProviderApplicationsPageState.new(params: params, provider_user: provider_user)
+    end
+    let(:rendered_component) { render_inline(component) }
```
duncanjbrown, line 15: This shadows `view_component`'s magic `rendered_component` method (like `response` in a request spec). https://github.com/github/view_component#testing-1. Could use a `before` maybe, and leave the tests as is?

app/models/provider_interface/provider_applications_page_state.rb
```diff
@@ -16,7 +16,32 @@ def filtered?
...
+
+    def sort_by
+      sort_options.include?(@params[:sort_by]) ? @params[:sort_by] : 'Last changed'
+    end
+
+    def sort_options
+      ['Last changed', 'Days left to respond']
+    end
+
+    def sort_order
```
duncanjbrown, line 30: Maybe for the future: at the cost of some indirection, we could move the actual sorting logic into a `SortApplicationChoicesForProviders` service to mirror `FilterApplicationChoicesForProviders` — `PageState` could go back to being ignorant of how the params are handled and give us just `last_changed` or `days_to_respond`

spec/components/provider_interface/application_card_component_spec.rb
steventux: The definition is here https://github.com/DFE-Digital/apply-for-teacher-training/blob/43fe487c9630ba1a3c52867c92b0ffbc49cea70b/config/initializers/date_formats.rb#L8 and I don't see the additional whitespace but the test fails without this spacing. The concatenation in `changed_at_date` doesn't appear to add it either.

app/models/provider_interface/provider_applications_page_state.rb
steventux: I'll do this in this PR, I agree that there's a lot of sorty sort of stuff going on in '`PageState`

steventux: @duncanjbrown https://github.com/DFE-Digital/apply-for-teacher-training/pull/2426/commits/fad8080ace092790a326319cbf93537de6b0c1cb I've delegated from sorting method calls to the sort service from page state. Not sure if this is a clean enough separation though, what do you think?

duncanjbrown: That's definitely a cleaner separation. I think I'd go further — have page state know nothing at all about sorting, something like:

```
state = PageState.new(params, user)

application_choices = # query, filter, etc

application_choices = SortApplicationChoicesForProviders.call(
  application_choices: application_choices,
  sort_by: state.sort_by # returns :days_to_respond or :last_changed
)
```

i.e. `SortApplicationChoicesForProviders#call` would take the current `ApplicationChoices` relation and a `sort_by` and use a switch on the `sort_by` to select the sorting method and do the sorting at the same time, a bit like `FilterApplicationChoicesForProviders`.

steventux: Ah! yes that's makes more sense I've amended the last commit https://github.com/DFE-Digital/apply-for-teacher-training/pull/2426/commits/ae9b93b0d4a6a4d0bca14aedf6f57b931b8fea66

duncanjbrown: Nice! one last suggestion: https://github.com/DFE-Digital/apply-for-teacher-training/commit/71145b7e3a564c4503d8152608f9650b32e7a422 to make everything except `PageState` deal in `days_left_to_respond` and `last_changed`

steventux: :+1: I've re-amended here https://github.com/DFE-Digital/apply-for-teacher-training/pull/2426/commits/dab9b2be664b2d0c7924ca5d4b0980636101e7b2 and included these in the option values which seemed to simplify things.

PR #2431, opened 2020-07-02
Remove SupportUser clauses in ProviderAuthorisation

tijmenb:
##  Context

The ProviderAuthorisation class is quite complicated already. As far as I can see, we aren't actually using the class in the Support UI, so removing these statements will be fine, and makes the class smaller.

## Link to Trello card

https://trello.com/c/sCTl8idK

## Things to check

### DIFF
--- a/app/services/provider_authorisation.rb
+++ b/app/services/provider_authorisation.rb
@@ -6,8 +6,6 @@ class ProviderAuthorisation
   end
 
   def can_make_offer?(application_choice:, course_option_id:)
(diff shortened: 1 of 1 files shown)


tijmenb:
@duncanjbrown before I fix the tests, does this make sense?

duncanjbrown:
Yes, I think so. It'll come back if we need it.

tijmenb:
Turns out the SupportUser clauses are used/misused in the test application generation, for some reason. Will look into this another time.

PR #2432, opened 2020-07-02
Extract class for determining authorisation of offer

tijmenb:
##  Context

This extracts the offer logic into a separate class, so that it can be more easily refactored.

Extract offer-related methods into an `OfferAuthorisation` class.

## Link to Trello card

https://trello.com/c/sCTl8idK/2333-review-access-controls-in-preparation-for-launch

## Things to check

### DIFF
--- a/app/services/offer_authorisation.rb
+++ b/app/services/offer_authorisation.rb
@@ -1,0 +1,74 @@
+class OfferAuthorisation
+  def initialize(actor:)
+    @actor = actor
+  end
+
+  def can_make_offer?(application_choice:, course_option_id:)
+    return true if @actor.is_a?(SupportUser)
+
+    course_option = CourseOption.find(course_option_id)
+    training_provider = course_option.provider
+    ratifying_provider = course_option.course.accredited_provider
+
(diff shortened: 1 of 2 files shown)


app/services/offer_authorisation.rb
```diff
@@ -0,0 +1,74 @@
+class OfferAuthorisation
+  def initialize(actor:)
+    @actor = actor
+  end
+
+  def can_make_offer?(application_choice:, course_option_id:)
```
duncanjbrown, line 6: What if we passed all the params in the constructor and then asked `granted?` or something? To remove the redundancy from the name

```ruby
authorisation = MakeOfferAuthorisation.new(actor: x, application_choice: y, course_option_id: z)
if authorisation.granted?
  do_stuff
end
```

tijmenb: I didn't want to refactor too much for this, it's likely to change anyway. Okay if we look at this later? There's a chance that this goes back into the class anyway, or that we extract a bunch of other `XAuthorisation` classes.

PR #2433, opened 2020-07-02
SPIKE signing out of DSI when we sign out of support or Manage

duncanjbrown:
Proof of concept — not for merging!

## Context

When a user signs out of Support or Manage, they stay signed in to DfE Sign-in. This means it's possible to log back in as them by clicking on the "Sign in with DfE Sign-in" button — DSI recognises them and forwards them back to us for a new session.

We would like to sign them out of DSI at the same time. Thankfully, TVS are doing this already so we can look at their work to see what we need to do.

The process is outlined in the (draft!) OpenID Connect spec here: https://openid.net/specs/openid-connect-session-1_0.html#RPLogout.

The basic idea is that when we sign a user out, we send them to DfE Sign-in first, then DSI sends the user back to us for a final logout.

I've attempted this in the following way:

**Add two methods to DfESignInUser: `logout_url_for_support_interface` and `logout_url_for_provider_interface`**. These return the special sign-me-out path on DfE Sign-in, which will sign the user and and then redirect them to us. We have separate methods because this can happen in two contexts, and we need to know which context we started in so we can send the user to the appropriate 'just signed out' page. We persist this information using the `state` param on the DSI URL.

**When we hit the `destroy` actions in `SupportInterface::SessionsController` or `ProviderInterface::SessionsController` we redirect immediately to DSI's sign-me-out path**. I preferred to leave these actions in the controllers rather than write the DSI urls directly into the template as we want to continue to support bypass mode and authentication by email. **These modes are both broken by this PR** — fixing the `destroy` actions should make them work again. Maybe we could still "optimistically" destroy the session in those actions.

**After DSI has signed us out, it sends the user to `DfESignInController#callback_after_signout`**. This destroys the session on our side (but see above) and redirects to the right page using the `state` param we passed earlier.

ALSO

We store `id_token` in the DSI user details in the session. This is "recommended" in the spec and TVS do it, but I'm not sure what happens if it's missing. We could well have user sessions in an `id_token`-less state when we first ship this.

### DIFF
--- a/app/controllers/dfe_sign_in_controller.rb
+++ b/app/controllers/dfe_sign_in_controller.rb
@@ -28,13 +28,12 @@ class DfESignInController < ActionController::Base
 
   alias_method :bypass_callback, :callback
 
-  # This is called by a redirect from DfE Sign-in after visiting
-  # DfESignInUser#logout_url_for(interface: (provider|support)).
+  # This is called by a redirect from DfE Sign-in after visiting the signout
+  # link on DSI. We tell DSI to redirect here using the
+  # post_logout_redirect_uri parameter - see DfESignInUser#dsi_logout_url
   #
-  # The given interface will appear here in the :state param.
-  def callback_after_signout
-    DfESignInUser.end_session!(session)
-
+  # The interface we signed out from will appear here in the :state param.
+  def redirect_after_dsi_signout
     if params[:state] == 'support'
       redirect_to support_interface_path
     else
--- a/app/controllers/dfe_sign_in_controller.rb
+++ b/app/controllers/dfe_sign_in_controller.rb
@@ -28,6 +28,20 @@ class DfESignInController < ActionController::Base
 
   alias_method :bypass_callback, :callback
 
+  # This is called by a redirect from DfE Sign-in after visiting
+  # DfESignInUser#logout_url_for(interface: (provider|support)).
+  #
+  # The given interface will appear here in the :state param.
+  def callback_after_signout
+    DfESignInUser.end_session!(session)
+
+    if params[:state] == 'support'
+      redirect_to support_interface_path
+    else
+      redirect_to provider_interface_path
+    end
+  end
+
 private
 
   def send_support_sign_in_confirmation_email
--- a/app/controllers/support_interface/sessions_controller.rb
+++ b/app/controllers/support_interface/sessions_controller.rb
@@ -10,9 +10,7 @@ module SupportInterface
     end
 
     def destroy
-      DfESignInUser.end_session!(session)
-
-      redirect_to action: :new
+      redirect_to dfe_sign_in_user.support_interface_logout_url
     end
 
     def sign_in_by_email
--- a/app/controllers/support_interface/sessions_controller.rb
+++ b/app/controllers/support_interface/sessions_controller.rb
@@ -10,7 +10,9 @@ module SupportInterface
     end
 
     def destroy
-      redirect_to dfe_sign_in_user.support_interface_logout_url
+      dsi_logout_url = dfe_sign_in_user.support_interface_dsi_logout_url
+      DfESignInUser.end_session!(session)
+      redirect_to dsi_logout_url
     end
 
     def sign_in_by_email
--- a/app/controllers/support_interface/sessions_controller.rb
+++ b/app/controllers/support_interface/sessions_controller.rb
@@ -10,9 +10,14 @@ module SupportInterface
     end
 
     def destroy
-      dsi_logout_url = dfe_sign_in_user.support_interface_dsi_logout_url
+      post_signout_redirect = if dfe_sign_in_user.needs_dsi_signout?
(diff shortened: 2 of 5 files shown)


duncanjbrown:
https://trello.com/c/HofRikzT/2381-spike-signing-out-of-manage-also-signs-you-out-of-dfe-sign-in

tijmenb:
This looks good, but I have questions about `DfESignInController#callback_after_signout`. Is that URL also called if the user signs out from DSI independently (say, from Find or from the DSI UI directly).

duncanjbrown:
> This looks good, but I have questions about `DfESignInController#callback_after_signout`. Is that URL also called if the user signs out from DSI independently (say, from Find or from the DSI UI directly)?

I think `callback` is probably not the right name for this action as it implies a webhook or sth, whereas in reality the user's own browser visits it. DSI knows to send them there bc we include our `post_logout_redirect_uri` when we build the link to sign out of DSI

https://github.com/DFE-Digital/apply-for-teacher-training/pull/2433/files#diff-e2f6e0bbd2ff99fd653c1c30472258aaR65-R72

duncanjbrown:
Small updates:

- on re-reading the spec, it actually prescribes the optimistic sign-out (ie kill the apply session before redirecting), so this branch now does that
- I've hopefully clarified some of the method names to indicate that DSI is what's being signed out of, and that the `callback` as was is really a redirect

I've discovered that missing out the `id_hint` parameter yields an (unhelpful) error — see below . We'll have to handle this case when we write this code proper.

Screenshot 2020-07-06 at 12 43 29

duncanjbrown:
https://trello.com/c/k9crMTdY/531-when-a-user-signs-out-of-manage-also-sign-them-out-of-dfe-sign-in

tijmenb:
Thanks for the clarification @duncanjbrown! In my head the DSI logout system did have a callback system, so that we could log out users even if they signed out in a different service. But this improves the thing quite a bit! I'll do a proper review tomorrow.

duncanjbrown:
Added a commit to handle the case where there's no `id_token`, which should unbreak email and dev-mode

app/models/dfe_sign_in_user.rb
```diff
@@ -16,6 +32,7 @@ def self.begin_session!(session, omniauth_payload)
       'first_name' => omniauth_payload['info']['first_name'],
       'last_name' => omniauth_payload['info']['last_name'],
       'last_active_at' => Time.zone.now,
+      'id_token' => omniauth_payload['credentials']['id_token'],
```
tijmenb, line 35: Can we come up with a more self-explanatory name? Would this be correct?

```suggestion
      'dfe_sign_in_sign_out_token' => omniauth_payload['credentials']['id_token'],
```

app/controllers/dfe_sign_in_controller.rb
```diff
@@ -28,6 +28,19 @@ def callback
 
   alias_method :bypass_callback, :callback
 
+  # This is called by a redirect from DfE Sign-in after visiting the signout
+  # link on DSI. We tell DSI to redirect here using the
+  # post_logout_redirect_uri parameter - see DfESignInUser#dsi_logout_url
+  #
+  # The interface we signed out from will appear here in the :state param.
+  def redirect_after_dsi_signout
+    if params[:state] == 'support'
```
tijmenb, line 37: Perhaps:

```suggestion
    if params[:state] == 'user_signed_out_from_support_interface'
```

app/controllers/support_interface/sessions_controller.rb
```diff
@@ -10,9 +10,14 @@ def new
     end
 
     def destroy
-      DfESignInUser.end_session!(session)
+      post_signout_redirect = if dfe_sign_in_user.needs_dsi_signout?
+                                dfe_sign_in_user.support_interface_dsi_logout_url
+                              else
+                                provider_interface_path
```
tijmenb, line 16: Should this be `support_interface_path`?

app/models/dfe_sign_in_user.rb
```diff
@@ -36,11 +53,28 @@ def self.load_from_session(session)
...
     session.delete('post_dfe_sign_in_path')
     session.delete('dfe_sign_in_user')
   end
+
+private
+
+  # a URL the user can visit to log them out of DSI and be redirected to our
+  # after-sign-out path where we'll delete their local session
+  def dsi_logout_url(interface:)
+    logout_url = URI.parse("#{ENV.fetch('DFE_SIGN_IN_ISSUER')}/session/end").tap do |url|
```
tijmenb, line 70: Not important but I think this can be visually simplified:

```rb
query = {
  post_logout_redirect_uri: auth_dfe_sign_out_url,
  id_token_hint: @id_token,
  state: interface,
}

"#{ENV.fetch('DFE_SIGN_IN_ISSUER')}/session/end?#{query.to_query}"
```

duncanjbrown:
Thanks for the notes @tijmenb — I'm going to close this branch now and link it from the Trello card. We'll make sure to fix these up when we do this for production.

PR #2434, opened 2020-07-02
Validation error summary page

stevehook:
## Context

We have built validation errors into the service and can now track these to a good level of detail. Tijmen has shown the value of this and where we can track user behavious. However to undertake analysis quickly effectively it would be helpful to have a summary table highlighting key areas for analysis focus.

- [x] New support interface page that summarise the error totals broken down by error type (form and attribute) over the past week, month and all time:

image

- [x] System spec
- [x] Extract query logic into service class with specs
- [x] Links to search page (drill-down)

## Guidance to review

- Is there a better way to present this info?
- Are the tests sufficient?
- I tried implement as much of this as possible in a single database query rather than doing in-memory data manipulation in Ruby. Does the query make sense?

## Link to Trello card

https://trello.com/c/iI7lB1Cc/1711-dev-validation-error-summary

## Things to check

### DIFF
--- a/app/controllers/support_interface/validation_errors_controller.rb
+++ b/app/controllers/support_interface/validation_errors_controller.rb
@@ -14,32 +14,7 @@ module SupportInterface
     end
 
     def summary
-      @validation_error_summary = ValidationErrorSummary.new.call
-    end
-
-    class ValidationErrorSummary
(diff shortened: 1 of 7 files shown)


stevehook:
Closing and re-opening to try to fix the review app

PR #2451, opened 2020-07-06
Add job to send matching data to UCAS

tijmenb:
##  Context

We're going to be sharing data with UCAS to find candidates who've applied to both services. 

This is the first step in that process, it uploads the data every night to UCAS' Movit instance. It uses the Movit API, documented at https://transfer.ucasenvironments.com/swagger/ui.

Authentication works by fetching an access token using a username and password, configured in environment variables. Currently we have one email address that has to the folder we're uploading to - if you need access than you can look at the environment variables.

This uploads to a *test environment*. I'm not a 100% sure what will change between production and testing environments - definitely the username & password, likely the folder ID, perhaps the URL as well. We'll see once we have the production credentials. 

## Guidance to review

There's a follow up this to _download_ matched data from the servers. At that point we'll probably refactor this code a bit, because things will be shared (auth, folder IDs, etc).

## Link to Trello card

## Things to check

### DIFF
--- a/app/services/ucas_matching/upload_matching_data.rb
+++ b/app/services/ucas_matching/upload_matching_data.rb
@@ -1,0 +1,51 @@
+require 'csv'
+
+module UCASMatching
+  class UploadMatchingData
+    include Sidekiq::Worker
+
+    # The `dfe_apply_itt_applications` folder in Movit. This will probably
+    # change for the production version.
+    UPLOAD_FOLDER = 685520099
+
+    def perform
+      unless ENV['UCAS_USERNAME']
+        Rails.logger.info 'UCAS credentials aren\'t configured, assuming that this this is a test environment. Not uploading data.'
+        return
+      end
+
+      filename = "/tmp/dfe_apply_itt_applications_#{DateTime.now}.csv"
+
+      File.open(filename, 'w') do |f|
+        f.write(csv_as_string)
+      end
+
+      # https://transfer.ucasenvironments.com/swagger/ui/index#/Folders/POSTapi%2Fv1%2Ffolders%2F%7BId%7D%2Fmove-1.0
+      response = HTTP
+        .auth(UCASAPI.auth_string)
+        .post(
+          "#{UCASAPI.base_url}/folders/#{UPLOAD_FOLDER}/files",
+          form: { file: HTTP::FormData::File.new(filename) },
+        )
+
+      unless response.status.success?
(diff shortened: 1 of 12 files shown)


app/services/ucas_matching/upload_matching_data.rb
```diff
@@ -0,0 +1,35 @@
+require 'csv'
+
+module UCASMatching
+  class UploadMatchingData
+    include Sidekiq::Worker
+
+    def perform
+      filename = "/tmp/dfe_apply_itt_applications_#{DateTime.now}.csv"
+
+      File.open(filename, 'a') do |f|
```
duncanjbrown, line 10: If this file is unique by timestamp, would `w` mode make more sense?

tijmenb: Done!

spec/services/ucas_matching/upload_matching_data_to_ucas_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+          body: { access_token: '123456789' }.to_json,
+        )
+
+      request = stub_request(:post, 'https://transfer.ucasenvironments.com/api/v1/folders/685520099/files')
+
+      create(:submitted_application_choice)
+
+      UCASMatching::UploadMatchingData.perform_async
+
+      expect(request).to have_been_made
```
duncanjbrown, line 17: Should we also check the authorization header here to make sure it's all hanging together?

tijmenb:
@duncanjbrown I've updated the PR. Replaced the unit test with a system test that also tests what happens either the API token or the upload fails. Added error handling.

app/services/ucas_matching/upload_matching_data.rb
```diff
@@ -0,0 +1,51 @@
...
+      # https://transfer.ucasenvironments.com/swagger/ui/index#/Folders/POSTapi%2Fv1%2Ffolders%2F%7BId%7D%2Fmove-1.0
+      response = HTTP
+        .auth(UCASAPI.auth_string)
+        .post(
+          "#{UCASAPI.base_url}/folders/#{UPLOAD_FOLDER}/files",
+          form: { file: HTTP::FormData::File.new(filename) },
+        )
+
+      unless response.status.success?
+        raise ApiError, "HTTP #{response.status} when fetching access token: '#{response}'"
```
duncanjbrown, line 32:
```suggestion
        raise ApiError, "HTTP #{response.status} when uploading to Movit: '#{response}'"
```

PR #2454, opened 2020-07-07
Remove old test data endpoint promote experimental endpoints

steventux:
## Context

Our API contract with vendors is necessarily very stable and inflexible. We have a policy of not removing things from the API.
We had treated the "new" test data endpoints `/experimental/test-data` as stable too, but users are getting confused about which endpoints to use.

- Clarify the difference in stability between `/applications` and `/test-data` endpoints.
- Removes `/test-data/regenerate` endpoint
- Promotes `/experimental/test-data/*` endpoints to top-level `/test-data/` path.

## Link to Trello card

https://trello.com/c/c9YEaQS4/2312-remove-old-test-data-endpoint-promote-experimental-endpoints

## Things to check


app/views/api_docs/pages/release_notes.md
```diff
@@ -1,3 +1,13 @@
+### 7th July 2020
+
+Documentation has been amended to emphasise the stability of `/applications` endpoints
+in contrast to the experimental and unsupported nature of `/test-data` endpoints.
```
duncanjbrown, line 4: "experimental and unsupported" sounds like we _really_ don't care! Like, we do support them. Maybe just "in contrast to" is enough?

steventux: @duncanjbrown I've revised the apocalyptic language and added a 308 redirect for the `/experimental` routes.

config/routes.rb
```diff
@@ -461,8 +461,11 @@
     post '/applications/:application_id/confirm-enrolment' => 'decisions#confirm_enrolment'
 
     post '/test-data/regenerate' => 'test_data#regenerate'
-    post '/experimental/test-data/generate' => 'test_data#generate'
-    post '/experimental/test-data/clear' => 'test_data#clear!'
+    post '/test-data/generate' => 'test_data#generate'
+    post '/test-data/clear' => 'test_data#clear!'
+
+    post '/experimental/test-data/generate' => redirect('/test-data/generate', status: 308)
```
duncanjbrown, line 467: This gets me a routing error:

Screenshot 2020-07-08 at 16 54 28

Adding `api/v1` to the front makes it work. Interestingly the params aren't passed so I generated the default 100 applications. The route redirect doesn't seem to work like a normal HTTP redirect — it seems to substitute one route for another. Have we got to redirect from the controller?

steventux: Ah OK this didn't work as expected, I tested this with `CURL -XPOST localhost:3000/api/v1/experimental/test-data/clear` and got a 'You are being redirected' HTML response. I wasn't sure this was a desireable outcome. I'll take another look.

mnacos: I don't think we should redirect POST requests to the new/non-experimental endpoints. First of all, a full redirect (making the call succeed) is not an option, secondly, the calling code may not be able to follow redirects (HTTP/REST libraries used for API calls do not necessarily follow redirects automatically).

I think we should just return an appropriate HTTP error code (so they know their call was unsuccessful) and a JSON response explaining that the endpoints have moved (so they know what is wrong). Bonus points for telling consumers where the new endpoints are in the JSON response (so they know how to fix).

steventux: @mnacos I've attempted this here https://github.com/DFE-Digital/apply-for-teacher-training/pull/2454/commits/469eef70c90639b0bfb227665e785538154e5984 I'm not sure about `410` as a response but all the redirect responses expect that we modify the `Location` header which isn't really what we want to do (Duncan also mentioned there is a side effect in redirecting). Opinions on which status to return welcome.

mnacos: I think `410 Gone` is perfect for this.

PR #2464, opened 2020-07-08
Generate fake providers with courses in the sandbox

AgaDufrat:
## Context

There are some providers, who need a "fake" provider to test features they can't test on the real one. In the case of UCB, they have many courses on Find, but only one is published. At some future date more of their courses might become available via Find and Apply. They therefore have an interest in testing cross-course features like making offers to different courses, but using the current data they only have a single course to work with so they can't.

It's also plausible that providers will need to test various relationships with accredited bodies.

Also some courses were getting full on Find so we need to add more courses for vendors to test with.

- Add a new task to Support interface which generates a fake provider with 10 courses and 3 ratified courses as well as an API token
- Refactor GenerateVendorProviders to use the new service and therefore
  - create 10 courses
  - create a Test providers as a part of Generate vendor providers
  - create 3 courses as a ratifying provider for 10 providers with the Test provider running the training

<kbd>Screenshot 2020-07-14 at 19 46 19</kbd>
<kbd>Screenshot 2020-07-14 at 19 45 48</kbd>

## Guidance to review

- click "Generate providers for vendors" http://localhost:3000/support/tasks 
- check in the console if the providers and courses were created

- click "Generate a fake provider for a vendor"

## Link to Trello card

https://trello.com/c/FID7u4VU/2385-generate-a-fake-provider-with-multiple-courses-in-sandbox

## Things to check

### DIFF
--- a/app/services/generate_vendor_providers.rb
+++ b/app/services/generate_vendor_providers.rb
@@ -16,7 +16,7 @@ class GenerateVendorProviders
     ]
 
     providers.each do |provider|
-      Provider.find_or_create_by(provider)
+      GenerateFakeProvider.generate_provider(provider)
     end
   end
 end
--- a/app/views/support_interface/tasks/create_fake_provider.html.erb
+++ b/app/views/support_interface/tasks/create_fake_provider.html.erb
@@ -1,0 +1,12 @@
+<% content_for :before_content, govuk_back_link_to(support_interface_tasks_path) %>
+<% content_for :title, 'New fake provider' %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render SummaryListComponent.new(rows: [
+      { key: 'Provider name', value: @new_provider.name },
+      { key: 'Provider code', value: @new_provider.code },
+      { key: 'Vendor API token', value: @vendor_api_token },
+    ]) %>
+  </div>
+</div>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -653,6 +653,7 @@ Rails.application.routes.draw do
     get '/performance/candidate-journey-tracking', to: 'performance#candidate_journey_tracking', as: :candidate_journey_tracking
 
     get '/tasks' => 'tasks#index', as: :tasks
+    post '/tasks/create-fake-provider' => 'tasks#create_fake_provider'
     post '/tasks/:task' => 'tasks#run', as: :run_task
 
     get '/validation-errors' => 'validation_errors#index', as: :validation_errors
--- a/app/controllers/support_interface/tasks_controller.rb
+++ b/app/controllers/support_interface/tasks_controller.rb
@@ -24,5 +24,15 @@ module SupportInterface
         render_404
       end
     end
+
+    def create_fake_provider
+      @new_provider = GenerateFakeProvider.generate_provider(
+        {
+          name: Faker::Educator.unique.university,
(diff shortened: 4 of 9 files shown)


app/services/generate_vendor_providers.rb
```diff
@@ -13,10 +13,28 @@ def self.call
...
       { name: 'University College Birmingham', code: 'B35' },
+      { name: 'Test Provider', code: 'TEST' },
     ]
 
     providers.each do |provider|
       Provider.find_or_create_by(provider)
     end
+
+    providers.reject { |p| p[:code] == 'TEST' }.each do |provider|
+      generate_courses_as_an_accredited_body(provider[:code])
```
steventux, line 24: It seems like you can just pass `provider` here to `generate_courses_as_an_accredited_body` and then you won't have to `Provider.find_by(code: accredited_body_code)` in this method. (See review comment above)

app/services/generate_vendor_providers.rb
```diff
@@ -13,10 +13,28 @@ def self.call
...
       { name: 'University of Newcastle Upon Tyne', code: 'N21' },
       { name: 'University College Birmingham', code: 'B35' },
+      { name: 'Test Provider', code: 'TEST' },
     ]
 
     providers.each do |provider|
       Provider.find_or_create_by(provider)
     end
+
+    providers.reject { |p| p[:code] == 'TEST' }.each do |provider|
```
steventux, line 23: Could we assign `Provider.find_by(code: 'TEST')` to a variable and pass it to `generate_courses_as_an_accredited_body` as `accredited_body_provider` (or whatever name makes sense).
This will mean we don't need to `find_by` within the method.

```ruby
accredited_body_provider = providers.find { |p| p[:code] == 'TEST' }

providers.except(accredited_body_provider).each do |provider|
  generate_courses_as_an_accredited_body(provider, accredited_body_provider)
end
```

app/views/support_interface/tasks/create_fake_provider.html.erb
```diff
@@ -0,0 +1,12 @@
+<% content_for :before_content, govuk_back_link_to(support_interface_tasks_path) %>
+<% content_for :title, 'New fake provider' %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render SummaryListComponent.new(rows: [
+      { key: 'Provider name', value: @new_provider.name },
+      { key: 'Provider code', value: @new_provider.code },
+      { key: 'Vendor API token', value: @vendor_api_token },
```
duncanjbrown, line 9: Nice!

config/routes.rb
```diff
@@ -636,6 +636,7 @@
 
     get '/tasks' => 'tasks#index', as: :tasks
     post '/tasks/:task' => 'tasks#run', as: :run_task
+    get '/tasks/create-fake-provider' => 'tasks#create_fake_provider'
```
duncanjbrown, line 639: This should be a POST as it creates data

AgaDufrat: Ah! I started with a POST and changed it to GET because it's not sending any params 🙈. Good to know, thanks!

duncanjbrown:
One more thing — maybe this should be marked sandbox only?

app/views/support_interface/tasks/create_fake_provider.html.erb
```diff
@@ -0,0 +1,14 @@
+# This is the view I want to render
```
duncanjbrown, line 1: Stray comment

PR #2467, opened 2020-07-08
Merge "provider_add_provider_users" and "provider_make_decisions_restriction" flags

tijmenb:
## Context

We currently have 3 feature flags that interact with each other. The `provider_add_provider_users` flag allows providers to add other providers. `provider_make_decisions_restriction` allows the providers to then restrict some users from making decisions (the third `enforce_provider_to_provider_permissions` then extends this to provider-to-provider permissions, which we're not touching at the moment).

This PR merges the 2 feature flags, because we're going to launch the features at the same time. It also makes it easier to understand the code. Additionally it makes some refactoring related to the flags.

## Guidance to review

Per commit.

## Link to Trello card

https://trello.com/c/sCTl8idK

## Things to check

### DIFF
--- a/app/views/provider_interface/provider_users/index.html.erb
+++ b/app/views/provider_interface/provider_users/index.html.erb
@@ -2,9 +2,7 @@
 
 <h1 class="govuk-heading-xl">Users</h1>
 
-<% if FeatureFlag.active?('provider_add_provider_users') -%>
-  <%= link_to 'Invite user', new_provider_interface_provider_user_path, class: 'govuk-button' %>
-<% end -%>
+<%= link_to 'Invite user', new_provider_interface_provider_user_path, class: 'govuk-button' %>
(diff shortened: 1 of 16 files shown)


app/views/provider_interface/provider_users/index.html.erb
```diff
@@ -2,9 +2,7 @@
 
 <h1 class="govuk-heading-xl">Users</h1>
 
-<% if FeatureFlag.active?('provider_add_provider_users') -%>
-  <%= link_to 'Invite user', new_provider_interface_provider_user_path, class: 'govuk-button' %>
-<% end -%>
+<%= link_to 'Invite user', new_provider_interface_provider_user_path, class: 'govuk-button' %>
```
duncanjbrown, line 5: Redundancy!

PR #2468, opened 2020-07-08
Update design for provider users in Support

tijmenb:
## Context

We're going to be introducing more permissions for providers, but we haven't updated the provider user screens in a long time. 

The provider list before:

image

After:

image

Things that have changed:

- Add the name of the provider user because that's useful to know
- Remove the DSI UID because it's only for 🤓 
- Add a list of permissions the user has. Controversy alert: I've added the implicit permission "View applications" as well, because otherwise the list would be empty, and I feel it's a permission. It's in a component that we might be able to reuse for the provider UI.

Provider user page before:

image

And after:

image

Things that have changed:

- A list of things we know about the user in a summary list
- Remove the form fields - these attributes can only be edited in DfE Signin 
- Add provider name to the title

## Link to Trello card

https://trello.com/c/aIg23O4H/438-turn-on-providers-managing-provider-users-feature-flag

## Things to check

### DIFF
--- a/app/components/check_icon.rb
+++ b/app/components/check_icon.rb
@@ -1,0 +1,2 @@
+class CheckIcon < ViewComponent::Base
+end
--- a/app/views/support_interface/provider_users/edit.html.erb
+++ b/app/views/support_interface/provider_users/edit.html.erb
@@ -39,14 +39,6 @@
         'Last sign in at' => @form.provider_user.last_signed_in_at&.to_s(:govuk_date_and_time) || 'Not signed in yet',
       }) %>
 
-      <% if @form.provider_user.dfe_sign_in_uid.present? %>
-        <%= f.govuk_text_field :email_address, label: { text: 'Email address', size: 'm' } %>
-      <% else %>
-        <%= f.govuk_text_field :email_address, label: { text: 'Email address', size: 'm' }, disabled: true, hint_text: 'The email address is not editable until the user has signed in for the first time. This is to ensure it continues to match the email on their DfE Sign-in invitation' %>
-      <% end %>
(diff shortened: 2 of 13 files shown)


app/components/check_icon.rb
```diff
@@ -0,0 +1,2 @@
+class CheckIcon < ViewComponent::Base
```
duncanjbrown, line 1: :+1: We could use this in the `ProviderInterface` too, where we currently have a partial — we can pick it up when we fix up all the account UI next week

app/views/support_interface/provider_users/edit.html.erb
```diff
@@ -24,17 +30,14 @@
...
-      <% else %>
-        <%= f.govuk_text_field :email_address, label: { text: 'Email address', size: 'm' }, disabled: true, hint_text: 'The email address is not editable until the user has signed in for the first time. This is to ensure it continues to match the email on their DfE Sign-in invitation' %>
-      <% end %>
-      <%= f.govuk_text_field :first_name, label: { text: 'First name' } %>
-      <%= f.govuk_text_field :last_name, label: { text: 'Last name' } %>
+      <%= render SummaryListComponent.new(rows: {
+        'First name' => @form.provider_user.first_name,
+        'Last name' => @form.provider_user.last_name,
+        'Email address' => @form.provider_user.email_address,
+        'DfE Signin UID' => @form.provider_user.dfe_sign_in_uid,
```
duncanjbrown, line 37:
```suggestion
        'DfE Sign-in UID' => @form.provider_user.dfe_sign_in_uid,
```

app/components/check_icon.rb
tijmenb: 👍

PR #2469, opened 2020-07-08
SPIKE: Set up ProviderUser permissions with a session-backed wizard

duncanjbrown:
## Context

We don't have any session-backed wizards in the app. This is what one might look like.

The need here is for a wizard which:

- can add and remove pages according to what is selected at different points in the process
- has "change" links that preserve the existing state of the form
- does not persist anything intermediate to the database. We don't want users hanging around in a half-invited state where we have to go and clean them up with cron

The prototype flow is here:

https://manage-applications-beta.herokuapp.com/users/new

You can try out the flow in this PR by navigating to `provider/users/new` locally or on the review app.

## Questions

- ⚠️ There may well be a security issue here where in the final POST, a user could assign permissions to a provider they don't have access to. `ProviderUserInvitationWizard#save!` could handle that.
- ~Not sure how best to test this as the controller is closely coupled with the Wizard. Probably a longish system spec?~ The wizard could be mostly tested in isolation following the below changes
- ~I'm not totally convinced the balance of responsibilites between controller and wizard, esp with respect to saving, is right. Only the controller can see the session, obviously, but maybe the wizard should be responsible for saving itself?~ the wizard can't know when to save itself (it's different for `update` actions) but moving the implementation of saving into the wizard feels better than the controller knowing about it. It also means we don't need to pass in the state-to-be-saved
- ~in the same way, the controller propbably knows too much about routing logic — it might be nicer for next_redirect to get a symbol from the wizard and dispatch accordingly. this would make the wizard more testable in isolation.~ did this — it's better!
- ~@adamsilver would like to lose the query param `?change=true` when we head back from the check answers page — we could perhaps set it pre-emptively in the session as soon as we get to that page, which might simplify the controller too. Maybe the wizard should know it's in the "checking" state?~ done this, and it's simpler: once we hit the check-answers page once we're in checking-answers mode from then on. No need for query param.

## Link to Trello card

https://trello.com/c/lTQDSi7i/2401-spike-a-wizard-for-the-provider-adding-a-provider-user-flow

### DIFF
--- a/app/controllers/provider_interface/provider_users_invitations_controller.rb
+++ b/app/controllers/provider_interface/provider_users_invitations_controller.rb
@@ -98,21 +98,13 @@ module ProviderInterface
     end
 
     def next_redirect(wizard)
-      if wizard.returning_to_answered_question
-        if wizard.has_permissions_to_set?
-          { action: :edit_permissions, provider_id: wizard.next_provider_requiring_permissions_setup }
-       else # carry on back to the check answers page
-          { action: :check }
-        end
-      elsif wizard.current_step == 'details'
-        { action: :edit_providers }
-      elsif wizard.current_step == 'providers' && wizard.has_permissions_to_set?
-        { action: :edit_permissions, provider_id: wizard.next_provider_requiring_permissions_setup }
-      elsif wizard.current_step == 'permissions' && wizard.has_permissions_to_set?
-        { action: :edit_permissions, provider_id: wizard.next_provider_requiring_permissions_setup }
-      else
-        { action: :check }
-      end
+      step, provider_id = wizard.next_step
+
+      {
+        check: { action: :check },
+        providers: { action: :edit_providers },
+        permissions: { action: :edit_permissions, provider_id: provider_id },
+      }.fetch(step)
     end
 
     def wizard_params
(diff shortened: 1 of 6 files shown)


duncanjbrown:
Here is a minimal example of this wizard pattern

```ruby
class ABasicWizard
  include ActiveModel::Model

  # Where we'll store the wizard's state in the state_store hash (probably the
  # session, IRL). This should be unique to the wizard and it should be a literal
  # string so we can rename/move the class in future.
  STATE_STORE_KEY = :a_basic_wizard

  # Required for the wizard to work
  attr_accessor :current_step, :checking_answers

  # Whatever you like - these are the params the wizard will collect
  attr_accessor :param_one, :param_two

  # on: causes the validation only to be applied on that step
  validates :param_one, presence: true, on: :start_page
  validates :param_two, presence: true, on: :second_page

  # Every time we create this object, load past state Clients are responsible
  # for deciding when to save_state!  (see below)
  def initialize(state_store, attrs = {})
    @state_store = state_store
    super(JSON.parse(last_saved_state).deep_merge(attrs))

    # Once we get to the check answers page, this will be true for the rest of
    # the session
    @checking_answers = true if current_step == 'check'
  end

  # Validates the wizard for the current step only
  def valid_for_current_step?
    valid?(current_step.to_sym)
  end

  # Returns [step, *params] for the next step. Controllers can interpret this
  # how they like.
  #
  # This way the wizard is responsible for its own routing logic but it doesn't
  # need to know about HTTP, so we can test it in isolation
  def next_step
    if checking_answers
      [:check]
    elsif current_step == 'start_page'
      [:second_page, :some_param_for_the_controller]
    else
      [:check]
    end
  end

  # This is the end of the wizard's life - it persists its contents and then it
  # clears its state
  def save!
    # Do some saving stuff
    clear_state!
  end

  # Clients should call this:
  # - immediately after instantiation, when showing a form
  # - only after calling valid_for_current_step? when receiving input
  def save_state!
    @state_store[STATE_STORE_KEY] = state
  end

  private

  # The current state of the object, minus some ActiveModel cruft and
  # state_store, which is received fresh on each .new
  def state
    as_json(except: %w[state_store errors validation_context]).to_json
  end

  def last_saved_state
    @state_store[STATE_STORE_KEY].presence || '{}'
  end

  def clear_state!
    @state_store.delete(STATE_STORE_KEY)
  end
end
```

duncanjbrown:
Closing ahead of https://trello.com/c/BcoX8Mt9/2431-implement-the-provider-user-invitation-wizard

PR #2425, opened 2020-07-02
Graph provider relationships

tijmenb:
## Context

We're working on implementing provider-to-provider permissions. We don't have a way to see what permissions exist between providers at the moment. I'm playing around with the code and data to get a better understanding of the domain.

## Changes proposed in this pull request

This generates a diagram that shows the relationships and permissions that exist. It may be useful to get an idea about the relationship between providers. It's very MVP and doesn't have any tests - I'd like to see this in qa/prod first.

image

## Guidance to review

https://apply-for-te-graph-prov-rycd5a.herokuapp.com/support/providers/relationships

## Link to Trello card

https://trello.com/c/sCTl8idK/2333-review-access-controls-in-preparation-for-launch

## Things to check

- [X] This code doesn't rely on migrations in the same Pull Request
- [X] If this code includes a migration adding or changing columns, it also backfills existing records for consistency
- [X] API release notes have been updated if necessary
- [X] New environment variables have been [added to the Azure config](https://github.com/DFE-Digital/apply-for-teacher-training#azure-hosting-devops-pipeline)

PR #2475, opened 2020-07-09
Revert "Graph provider relationships"

tijmenb:
Reverts DFE-Digital/apply-for-teacher-training#2425. I've pondered this over the last week, and I think a graph isn't a good way of showing this information. A graph like this implies that the nodes are the same similar, while they aren't in this case. I'll come back with a different way of visualising the relationships between providers.


tijmenb:
Actually, I can reuse bits of this code. Closing again!

PR #2476, opened 2020-07-09
RFC: change modelling of provider-to-provider relationships

tijmenb:
## Context

Provider to provider relationships are modelled using 2 classes.

The following means that for courses accredited by `provider_b`, the staff of `provider_a` will have the power to make decisions:

```rb
TrainingProviderPermissions.create!(
  training_provider: provider_a, 
  ratifying_provider: provider_b, 
  make_decisions: true,
)
```

And this means that for courses accredited by `provider_b`, the staff of `provider_b` will have the power to make decisions:

```rb
RatifyingProviderPermissions.create!(
  training_provider: provider_a,
  ratifying_provider: provider_b,
  make_decisions: true,
)
```

This is a spike into changing the model so we have a single database record that explains what the relationship is between 2 providers. The previous examples change into the following:

```rb
ProviderRelationshipPermissions.create!(
  training_provider: provider_a,
  ratifying_provider: provider_b,
  training_provider_can_make_decisions: true,
  ratifying_provider_can_make_decisions: true,
)
```

The spike only goes as far as fixing the `ProviderAuthorisation` code and specs (which pass now). Hopefully that's enough to see whether this is a worthwile change to make.

## Guidance to review

Focus on:

- app/services/offer_authorisation.rb
- spec/services/provider_authorisation_spec.rb

## Link to Trello card

https://trello.com/c/sCTl8idK/2333-review-access-controls-in-preparation-for-launch

## Things to check


steventux:
We considered this kind of approach when we first set out the provider relationship permissions classes, there's a doc somewhere detailing the options we went through. We favoured the two-STI-class approach as it avoided pairing permissions columns in the same db record (I don't remember any other reason for splitting the permissions this way but there may have been others).
Having worked with the 2 STI class approach and done some of the implementation I think it's an awkward thing to manage, there are very few instances where we use only one of the two classes in a permissions pair and I don't see any limitations in consolidating into one permissions class.
There are more instances where we need to combine the training/ratifying provider ids from two columns from both permissions records which is painful and feels off somehow. 
In short I don't think we're gaining anything by maintaining a pairing of permissions like this. I think the cost in understanding and working with the current modelling is quite high.

I'd be in favour of bringing these two permissions classes together like this, the dual permissions column overhead seems less worse than dual permissions classes.

mnacos:
I think the suggested re-modelling improves things in two ways: a) there is one db record per (directed) relationship instead of two b) it's a bit clearer who's allowed to do what. I don't particularly like the long column names for the boolean fields, but if was new to the code I would immediately know what they mean, so this is another advantage.

So, all things considered, :+1:

duncanjbrown:
I agree with all the above - this is better! As a nice side effect it will make validating these relationships simpler, eg checking that at least one side has make_decisions.

duncanjbrown:
As @tijmenb is away this week and this doesn't block any of the work in the current sprint, I'm closing this and converting it to a Trello card to pick up with the other org-org work. https://trello.com/c/cW38R9Pd/545-finish-refactoring-provider-provider-permissions

PR #2477, opened 2020-07-09
Don't try to upload data to UCAS if we don't have a username

tijmenb:
## Context

I forgot that Azure sets these environment variables to an empty string if the variable isn't set in the library. This now does the correct thing.

Fixes https://sentry.io/organizations/dfe-bat/issues/1774232835/?referrer=slack

See commit.

## Link to Trello card

## Things to check

### DIFF
--- a/app/services/ucas_matching/upload_matching_data.rb
+++ b/app/services/ucas_matching/upload_matching_data.rb
@@ -9,7 +9,7 @@ module UCASMatching
     UPLOAD_FOLDER = 685520099
 
     def perform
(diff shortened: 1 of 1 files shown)


tijmenb:
This didn't work because if the variable isn't set:

```
irb(main):001:0> ENV['UCAS_USERNAME']
=> "$(ucasUsername)"
```

cc @stevehook @AgaDufrat @mnacos

I've added empty vars to all envs, hopefully that fixes the issue of the sync running on prod.

PR #2480, opened 2020-07-09
Rename first nationality to nationality

davidgisbey:
## Context

With the new designs, it no longer makes sense for there to be a first and second nationality as there is a free text for them to input details on multiple nationalities. 

This PR creates a migration to change first_nationality to nationality on the application form.

- Migration to rename first_nationality column to nationality
- Use the new value in the code base

## Link to Trello card

https://trello.com/c/AsOuRv9u/1760-dev-%F0%9F%8C%90-international-residency-visa-status

## Things to check


tvararu:
Turns out the design has changed, so we don't need this anymore.

PR #2483, opened 2020-07-10
Add non-uk GCSEs columns to the application qualifications table

davidgisbey:
## Context

International candidates are asked to provide additional details on their GCSE equivalency qualifications.

- Add additional columns to application qualifications 

## Guidance to review

Is the naming okay? I struggled to come up with a good name for the country in which the qualification was awarded.

## Link to Trello card

## Things to check


db/migrate/20200710101438_add_non_uk_gcse_columnns_to_application_qualification.rb
```diff
@@ -0,0 +1,8 @@
+class AddNonUkGcseColumnnsToApplicationQualification < ActiveRecord::Migration[6.0]
+  def change
+    add_column :application_qualifications, :non_uk_qualification_type, :string
+    add_column :application_qualifications, :accrediting_country, :string
```
davidgisbey, line 4: I'm not convinced about this, but can't think of anything better. Maybe awarding_country?

tvararu: I prefer `awarding_country` 👍

stevehook: For international degrees we've got `institution_country` but I think that's a different thing.

db/migrate/20200710101438_add_non_uk_gcse_columnns_to_application_qualification.rb
```diff
@@ -0,0 +1,8 @@
...
+    add_column :application_qualifications, :naric_reference_number, :string
+    add_column :application_qualifications, :comparable_uk_qualification, :string
```
stevehook, line 6: We've already got a `naric_reference` and `comparable_uk_degree`. I wonder whether we should be using the same columns at least for the reference. Maybe I should drop `comparable_uk_degree` and use `comparable_uk_qualification` for everything? It could get very confusing otherwise.

davidgisbey: Yea I think it probably makes sense to use `comparable_uk_qualification` for all of the qualifications.

PR #2484, opened 2020-07-10
Zip the file we're sending to UCAS

tijmenb:
## Context

UCAS would like to have the CSV we send them zipped and encrypted.

This does that using the `archive_zip` gem. I also tried to use Rubyzip but the documentation scared me off (https://github.com/rubyzip/rubyzip#password-protection-experimental says the password thing is "experimental").

## Guidance to review

Run the export, test that you can extract the ZIP into a valid CSV.

## Link to Trello card

## Things to check

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -158,3 +158,7 @@ Layout/MultilineMethodArgumentLineBreaks:
 
 Layout/FirstMethodArgumentLineBreak:
   Enabled: false
+
+# 🤷‍♂️
+Style/AsciiComments:
+  Enabled: false
--- a/app/services/ucas_matching/matching_data_file.rb
+++ b/app/services/ucas_matching/matching_data_file.rb
@@ -1,0 +1,42 @@
+require 'csv'
+# require 'archive/zip'
+
+module UCASMatching
+  class MatchingDataFile
+    # @returns filename
+    def create_file
+      timestamp = DateTime.now.strftime('%Y%m%d_%H%M%S_%z').gsub('+', '')
+      filename_prefix = "/tmp/dfe_apply_itt_applications_#{timestamp}-#{SecureRandom.hex}"
+      csv_filename = "#{filename_prefix}.csv"
+      zip_filename = "#{filename_prefix}.zip"
+
+      File.open(csv_filename, 'w') do |f|
+        f.write(csv_as_string)
+      end
+
+      Archive::Zip.archive(
+        zip_filename,
+        csv_filename,
+        encryption_codec: Archive::Zip::Codec::TraditionalEncryption,
+        password: ENV.fetch('UCAS_ZIP_PASSWORD'),
+      )
+
+      zip_filename
+    end
+
+  private
+
+    def csv_as_string
+      applications = MatchingDataExport.new.applications
+      header_row = MatchingDataExport.csv_header(applications)
+
+      CSV.generate do |rows|
+        rows << header_row
+
+        applications.each do |application|
+          rows << application.values
+        end
+      end
+    end
+  end
+end
--- a/azure/pipelines/build.yml
+++ b/azure/pipelines/build.yml
@@ -291,6 +291,7 @@ stages:
       dsiApiSecret: '$(dsiApiSecret)'
       ucasUsername: '$(ucasUsername)'
       ucasPassword: '$(ucasPassword)'
+      ucasZipPassword: '$(ucasZipPassword)'
       sandbox: '$(sandbox)'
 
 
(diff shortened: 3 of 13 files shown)


app/services/ucas_matching/matching_data_file.rb
```diff
@@ -0,0 +1,41 @@
...
+module UCASMatching
+  class MatchingDataFile
+    # @returns filename
+    def create_file
+      filename_prefix = "#{Rails.root}/tmp/dfe_apply_itt_applications_#{Time.zone.now.iso8601}-#{SecureRandom.hex}"
+      csv_filename = "#{filename_prefix}.csv"
+      zip_filename = "#{filename_prefix}.zip"
+
+      File.open(csv_filename, 'w') do |f|
+        f.write(csv_as_string)
```
duncanjbrown, line 13: Do we need the intermediate CSV file — could we pass it to `Archive::Zip` as a `StringIO`, maybe?

duncanjbrown: Maybe that wouldn't work — I guess the ZIP needs to take named files on disk or what would it unzip to?

tijmenb: Yeah, I think `archive-zip` can't do it without the intermediate file.

azure/pipelines/build.yml
```diff
@@ -291,6 +291,7 @@ stages:
       dsiApiSecret: '$(dsiApiSecret)'
       ucasUsername: '$(ucasUsername)'
       ucasPassword: '$(ucasPassword)'
+      ucasZipPassword: '$(ucasZipPassword)'
```
mnacos, line 294: I recall @vigneshmsft said entries in build.yml and release.yml are no longer needed when I added ucasUploadBaseUrl etc.

azure/pipelines/release.yml
```diff
@@ -97,6 +97,7 @@ stages:
       dsiApiSecret: '$(dsiApiSecret)'
       ucasUsername: '$(ucasUsername)'
       ucasPassword: '$(ucasPassword)'
+      ucasZipPassword: '$(ucasZipPassword)'
```
mnacos, line 100: Ditto, see comment on build.yml

azure/pipelines/templates/deploy.yml
```diff
@@ -243,6 +244,7 @@ jobs:
                 -dsiApiSecret "${{parameters.dsiApiSecret}}"
                 -ucasUsername "${{parameters.ucasUsername}}"
                 -ucasPassword "${{parameters.ucasPassword}}"
+                -ucasZipPassword "${{parameters.ucasZipPassword}}"
```
mnacos, line 247:
```suggestion
                -ucasZipPassword "$(ucasZipPassword)"
```

I think this is the new recommended way of doing this.

spec/system/ucas_matching/upload_matching_data_to_ucas_spec.rb
```diff
@@ -79,7 +79,6 @@ def then_the_file_has_been_uploaded
     expect(@latest_exception).to be_nil
 
     expect(WebMock).to have_requested(:post, 'https://transfer.ucasenvironments.com/api/v1/folders/685520099/files')
```
mnacos, line 81: Perhaps test the file can be unzipped here?

tijmenb: No that doesn't work. See the comment in the PR for the background.

azure/pipelines/build.yml
tijmenb: Let's fix that in a separate PR, I don't want to introduce inconsistency between different sets of variables, since this stuff is so hard to test.

azure/pipelines/release.yml
tijmenb: Same as https://github.com/DFE-Digital/apply-for-teacher-training/pull/2484#discussion_r457368555

azure/pipelines/templates/deploy.yml
tijmenb: I'd prefer to do this in a different PR for everything to avoid inconsistency here as well.

spec/system/ucas_matching/upload_matching_data_to_ucas_spec.rb
tijmenb: @duncanjbrown made it work! https://github.com/DFE-Digital/apply-for-teacher-training/pull/2484/commits/3228e74c3f9d0146a03ee2ffae5b871e7643dc3b

.rubocop.yml
```diff
@@ -158,3 +158,7 @@ Layout/MultilineMethodArgumentLineBreaks:
 
 Layout/FirstMethodArgumentLineBreak:
   Enabled: false
+
+# 🤷‍♂️
+Style/AsciiComments:
+  Enabled: false
```
duncanjbrown, line 164: 🎉

PR #2486, opened 2020-07-10
add support for multiple custom domain bindings in App Service Template

vigneshmsft:
## Context

Changing main service under service.gov.uk domain, 
while keeping support for education.gov.uk to enable redirection from education -> service domain.

Change ARM templates to support multiple custom domain bindings.

Certificate for service domain is created in keyvault and the thumbprint is referenced in
the customDomains parameter which is set per environment in Azure DevOps variable groups.

Removing few unused params in `template.json`

## Things to check


saliceti:
I think this has debug code as it doesn't rely on the bat-platform-building-blocks

vigneshmsft:
> k this has debug code as it do

Yes, I was planning to remove the local copy of `app-service-linux.json` once the PR in bat-building-block is merged.

PR #2488, opened 2020-07-10
Show per-provider permissions change links when editing provider users

mnacos:
### Context

We'll be adding per-provider edit permissions pages within the provider interface. These will be accessed via per-provider 'change' links within the 'show user' view. This PR changes ` ProviderUserDetailsComponent` to display a 'change' link per provider listed.

Editing a user within the provider interface

Before:
image

After:
image
The target of these 'change' links has not changed yet, but 'Manage users' is not live yet, so this page is not accessible. Merging this will provide entry points for other bits of work.

### Guidance to review

Do the code changes make sense? The main objective of this PR is to provide a scaffold for two other bits of work developed in parallel: a new provider-specific permissions list component (which will be extended to handle org relationships) and b. adding provider-specific edit permissions views.

### Link to Trello card

This work is part of [Implement edit permissions page for provider users](https://trello.com/c/GU90aVjA)

### Things to check


app/components/provider_interface/provider_user_details_component.rb
```diff
@@ -14,22 +14,44 @@ def rows
...
-      }
+    def visible_provider_permissions
+      @possible_permissions & @provider_permissions
+    end
+
+    def permission_rows
+      visible_provider_permissions.map do |permission|
+        {
+          key: "Permissions: #{permission.provider.name}",
+          value: temp_code_this_will_be_a_component(permission),
```
mnacos, line 28: This is where your component will be included/rendered @AgaDufrat :)

AgaDufrat: I made some changes to Tijmen's component in the end so we can go with:
```suggestion
          value: render(PermissionsList.new(permission)),
```

PR #2492, opened 2020-07-13
Add provider permissions to the account page

AgaDufrat:
## Context

Users need to be able to view what permissions they have for their account

Update DfE Sign-in link to mention password

-  Update DfE Sign-in link to mention password change
- Change order of permissions in PermissionsList component
- Show "View applications only" when user doesn't have other permissions
- Show user permissions on the provider account page

<kbd>Screenshot 2020-07-13 at 12 10 15</kbd>

## Guidance to review

- visit http://localhost:3000/provider/account

## Link to Trello card

https://trello.com/c/V6p6TK33/2412-add-provider-permissions-to-account-page

## Things to check

### DIFF
--- a/app/components/permissions_list.html.erb
+++ b/app/components/permissions_list.html.erb
@@ -1,19 +1,19 @@
 <ul class="govuk-list" id="provider-<%= @permission_model.provider.id %>-enabled-permissions">
-  <li><%= render CheckIcon.new %> View applications</li>
+  <% if @permission_model.manage_organisations? %>
+    <li><%= render CheckIcon.new %> Manage organisations</li>
+  <% end %>
 
   <% if @permission_model.manage_users? %>
     <li><%= render CheckIcon.new %> Manage users</li>
   <% end %>
 
-  <% if @permission_model.manage_organisations? %>
-    <li><%= render CheckIcon.new %> Manage organisations</li>
+  <% if @permission_model.make_decisions? %>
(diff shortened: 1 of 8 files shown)


app/components/permissions_list.rb
```diff
@@ -2,4 +2,13 @@ class PermissionsList < ViewComponent::Base
   def initialize(permission_model)
     @permission_model = permission_model
   end
+
+  def view_applications_only?
```
mnacos, line 6: I like this but the list of user-level permissions may change in the future and this method could fall behind. How about we move it to `app/models/provider_permissions.rb` and use `VALID_PERMISSIONS` like so:

```ruby
  def view_applications_only?
    VALID_PERMISSIONS.map { |permission| send(permission) }.all?(false)
  end
```

app/components/permissions_list.html.erb
```diff
@@ -1,19 +1,21 @@
...
+    <li><%= render CheckIcon.new %> Make decisions</li>
   <% end %>
 
   <% if @permission_model.view_safeguarding_information? %>
     <li><%= render CheckIcon.new %> View safeguarding information</li>
   <% end %>
 
-  <% if @permission_model.make_decisions? %>
-    <li><%= render CheckIcon.new %> Make decisions</li>
+  <% if view_applications_only? %>
```
mnacos, line 18:
```suggestion
  <% if @permission_model.view_applications_only? %>
```

Please see my other comment below.

PR #2494, opened 2020-07-13
Persist filter state across requests

duncanjbrown:
## Context

This is out of sync with the prototype

- when visiting `/applications`, you can select filters, searches, ordering etc
- if you navigate away, and then click "Applications" in the header, your filters are preserved
- submitting new filters (or blank filters) will clear the saved state

This is achieved by having `ProviderApplicationsPageState` write its state into the session (handed down from the controller) whenever it changes. The above logic is used to decide whether to read the existing state, or update and rewrite that state.

## Guidance to review

Does this make sense?

## Link to Trello card

https://trello.com/c/qPEBdMux/2411-ensure-that-users-filters-persist-across-sessions

## Things to check

### DIFF
--- a/app/models/provider_interface/provider_applications_page_state.rb
+++ b/app/models/provider_interface/provider_applications_page_state.rb
@@ -3,11 +3,18 @@ module ProviderInterface
     attr_accessor :available_filters, :filter_selections, :provider_user
     attr_reader :applied_filters
 
+    STATE_STORE_KEY = :provider_interface_applications_page_state
 
     def initialize(params:, provider_user:, state_store:)
       @provider_user = provider_user
       @applied_filters = parse_params(params)
       @state_store = state_store
+
+      if @applied_filters.empty?
+        @applied_filters = last_saved_filter_state
+      else
+        save_filter_state!
+      end
     end
 
     def filters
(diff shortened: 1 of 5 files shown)


app/models/provider_interface/provider_applications_page_state.rb
```diff
@@ -1,10 +1,22 @@
...
 
-    def initialize(params:, provider_user:)
-      @params = params
+    STATE_STORE_KEY = :provider_interface_applications_page_state
+
+    def initialize(params:, provider_user:, state_store:)
       @provider_user = provider_user
+      @applied_filters = parse_params(params)
+      @state_store = state_store
+      filtering_params = parse_params(params)
```
stevehook, line 12: is this redundant? could you just use `@applied_filters`?

app/models/provider_interface/provider_applications_page_state.rb
```diff
@@ -1,10 +1,22 @@
...
+
+      if filtering_params.empty?
+        @applied_filters = last_saved_filter_state
+      else
+        @applied_filters = parse_params(params)
```
stevehook, line 17: also redundant?

duncanjbrown:
Thanks for comments @stevehook — will address!

PR #2499, opened 2020-07-14
Add country to international GCSE flow 

davidgisbey:
## Context

When an international candidate is inputting their GCSE equivalency they need to be able to input which country they studied in.

- Add the institution country page 

image

## Guidance to review

I'm not the institution_country should be saved as a country code? If so this will need to be tweaked.

## Link to Trello card

## Things to check

### DIFF
--- a/app/forms/candidate_interface/gcse_institution_country_form.rb
+++ b/app/forms/candidate_interface/gcse_institution_country_form.rb
@@ -4,6 +4,8 @@ module CandidateInterface
 
     attr_accessor :institution_country
 
+    validates :institution_country, presence: true
+
     validates :institution_country,
               inclusion: { in: COUNTRIES_BY_NAME }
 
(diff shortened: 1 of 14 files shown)


app/forms/candidate_interface/gcse_institution_country_form.rb
```diff
@@ -0,0 +1,26 @@
+module CandidateInterface
+  class GcseInstitutionCountryForm
+    include ActiveModel::Model
+
+    attr_accessor :institution_country
+
+    validates :institution_country, presence: true
+
+    validates :institution_country,
+              inclusion: { in: COUNTRIES_BY_NAME }
```
davidgisbey, line 10: Should this be being saved as a country code i.e. 'US' rather than 'United States'.

tvararu: Hmm, I think that's a good idea, could you update to that?

davidgisbey: Have done 👍

tvararu:
Looks like a bug with prefilling the existing value:

Screen Recording 2020-07-17 at 10 13 49

davidgisbey:
Yea sorry forgot to update the build_from_qualificaton method to convert the ISO. Should be all working now 👍

PR #2501, opened 2020-07-14
Speed up SupportInterface::ApplicationsExport

duncanjbrown:
This did a lot of queries and loaded a lot of AR objects into memory, and it stopped running on prod because it caused database timeouts (the threshold is 10 seconds).

Using pluck and storing more objects in RAM rather than fetching from the DB sacrifices a bit of readability for performance. Even though we retain more bits of data this still seems to be cheaper in memory terms.

Not 100% sure this will fix the problem on prod, but it ought to help.

Performance stats, using 2020 forms with 2594 application choices:

**Before**

_Benchmark_
real time, seconds: 20.882151999976486

_Memory profile_
Total allocated: 1062854626 bytes (7731962 objects)
Total retained:  858798 bytes (9154 objects)

**After**

_Benchmark_
real time, seconds: 6.385142999934033

_Memory profile_
Total allocated: 141258273 bytes (1403405 objects)
Total retained:  168 bytes (1 objects)

## Context

https://sentry.io/organizations/dfe-bat/issues/1689909459/?referrer=slack

## Guidance to review

Does the code still do the same thing?

## Link to Trello card

https://trello.com/c/A9JuYHfs/2436-speed-up-supportinterfaceapplicationsexport

## Things to check

### DIFF
--- a/app/services/support_interface/applications_export.rb
+++ b/app/services/support_interface/applications_export.rb
@@ -1,7 +1,25 @@
 module SupportInterface
   class ApplicationsExport
     def applications
-      relevant_applications.map do |application_form|
+      # Load the ApplicationForm audits into memory first, which is faster.
+      # Avoid the overhead of object creation and storage using #pluck.
+      audits = Audited::Audit.where(action: 'update', auditable_type: 'ApplicationForm')
+        .order(created_at: :desc).pluck(:auditable_id, :audited_changes, :created_at).to_a
+
+      applications_of_interest = ApplicationForm.includes(
+        :candidate,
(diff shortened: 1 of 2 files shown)


duncanjbrown:
> This looks like a sensible optimisation.
> 
> I think the same sort of problems will crop up with the candidate journey report too. As we scale up we'll perhaps have to think about doing these reports a bit differently, either by splitting them somehow or pre-calculating them overnight (or more regularly) in the background.

Yes. In theory the audit log knows everything, it just holds the information in an slightly awkward way. Maybe we could do something with that.

Updated to use `.dig`, much better!

PR #2507, opened 2020-07-15
Don't assume TimeLimitConfig(:rule).first.limit is the correct limit

duncanjbrown:
Should fix the build (see https://dfe-ssp.visualstudio.com/Become-A-Teacher/_build/results?buildId=93068&view=ms.vss-test-web.build-test-results-tab&runId=1334394&resultId=100162&paneView=debug).

There can be multiple time limits per rule (eg summer deadlines changing) so taking the first limit is not always the right thing to do.

`TimeLimitCalculator` knows how to choose the right date, so always go through that.

There are still a couple of system specs using `TimeLimitConfig(:rule).first.limit` directly but they need more surgery.

## Link to Trello card

https://trello.com/c/UZg93KLf/1731-we-use-timelimitconfig-directly-in-mailers-which-can-cause-problems-over-the-summer-when-time-limits-change?menu=filter&filter=label:Dev

## Things to check

### DIFF
--- a/app/mailers/candidate_mailer.rb
+++ b/app/mailers/candidate_mailer.rb
@@ -142,6 +142,7 @@ class CandidateMailer < ApplicationMailer
   def chase_candidate_decision(application_form)
     @application_choices = application_form.application_choices.select(&:offer?)
     @dbd_date = @application_choices.first.decline_by_default_at.to_s(:govuk_date).strip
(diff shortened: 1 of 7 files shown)


duncanjbrown:
Broke some mailer previews... fixing 🔧

PR #2485, opened 2020-07-10
Add non-uk qualifications to the GCSE type page

davidgisbey:
## Context

International candidates need to be able to input their GCSE equivalencies.

In an attempt to keep my PRs small, this just covers adding an additional radio button to the type page and makes sure the review page shows the correct values. 

The behaviour of non_uk_qualifcation_type is almost identically to other_uk_qualification_type. 

## Changes proposed in this pull request

- Add a radio button and text area for non-uk qualifications
- Update the GCSE qualification type from to use the new fields
- Update the review component to display the input qualification name

Type page

| Before  | After |
| ------------- | ------------- |
|image| image|

Review page

image

## Guidance to review

I think it's fairly straight forward to review. Just literally clicking through and adding one and checking the behaviour is okay.

## Link to Trello card

https://trello.com/c/Sy4RX8lL/1763-dev-%F0%9F%8C%90-international-gcse-equivalents

## Things to check

- [x] This code doesn't rely on migrations in the same Pull Request
- [x] If this code includes a migration adding or changing columns, it also backfills existing records for consistency
- [x] API release notes have been updated if necessary
- [x] New environment variables have been [added to the Azure config](https://github.com/DFE-Digital/apply-for-teacher-training#azure-hosting-devops-pipeline)

PR #2509, opened 2020-07-15
Add NARIC reference number page to the international GCSE flow

davidgisbey:
## Context

Previous PRs are:
 
#2485 - Add non_uk to qualification the qualification type page
#2499 - Add the institution country page 

Candidates need to able to provide the NARIC reference number and the comparable UK qualification
for their GCSE equivalency.

- Add the naric reference page which allows a user to provide their NARIC reference number and the comparable uk equivalent

image

## Guidance to review

There will be one more PR which will allow you to change the grade page.

## Link to Trello card

## Things to check

### DIFF
--- a/spec/components/candidate_interface/gcse_qualification_review_component_spec.rb
+++ b/spec/components/candidate_interface/gcse_qualification_review_component_spec.rb
@@ -2,13 +2,14 @@ require 'rails_helper'
 
 RSpec.describe CandidateInterface::GcseQualificationReviewComponent do
   context 'with the international_gcses flag on' do
-    it 'displays award year, qualification type, grade and institution country' do
+    it 'displays award year, qualification type, grade, institution country, naric reference and comparable UK qualification' do
       FeatureFlag.activate('international_gcses')
       application_form = build :application_form
       @qualification = application_qualification = build(
         :application_qualification,
         application_form: application_form,
-        qualification_type: 'GCSE',
+        qualification_type: 'non_uk',
+        non_uk_qualification_type: 'High school diploma',
         level: 'gcse',
         grade: 'c',
         institution_country: 'US',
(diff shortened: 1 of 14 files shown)


config/initializers/countries.rb
```diff
@@ -200,3 +200,4 @@
 }.freeze
 
 COUNTRIES_BY_NAME = COUNTRIES.map(&:second)
+REVERSE_COUNTRIES_HASH = COUNTRIES.map(&:reverse).to_h
```
malcolmbaig, line 203: I think it might be worth revisiting some of this naming.

`COUNTRIES_BY_NAME` could simply be `COUNTRY_NAMES`, as the 'by' suggests are more complex data structure than a simple list.

`REVERSE_COUNTRIES_HASH` could be `COUNTRY_NAMES_TO_ISO_CODES`, so its clearer what the constant contains without looking up its definition.

spec/components/candidate_interface/gcse_qualification_review_component_spec.rb
```diff
@@ -2,25 +2,53 @@
 
 RSpec.describe CandidateInterface::GcseQualificationReviewComponent do
   context 'with the international_gcses flag on' do
-    it 'displays award year, qualification type, grade and institution country' do
+    before do
+      FeatureFlag.activate('international_gcses')
+    end
+
+    it 'displays award year, qualification type, grade, institution country, naric reference and comparable UK qualification' do
```
malcolmbaig, line 9: Suggest renaming to `it 'renders a non-uk GCSE equivalent'` (or words to that effect) - the expectations at the end of the `it` block are easily scannable for anyone interested in the detail.

spec/components/candidate_interface/gcse_qualification_review_component_spec.rb
```diff
@@ -2,25 +2,53 @@
...
       FeatureFlag.activate('international_gcses')
```
malcolmbaig, line 10: Already activated by the context block.

config/initializers/countries.rb
davidgisbey: All good points. I'll change them in the previous PR dealing with nationalities 👍

spec/components/candidate_interface/gcse_qualification_review_component_spec.rb
davidgisbey: Good point. I have changed it. Thanks for the feedback.

PR #2511, opened 2020-07-15
Missing link classes

paulrobertlloyd:
Fixes two places where `govuk-link` was not given to links, meaning they use browser default colour. 

Screenshot 2020-05-26 at 18 12 36

Screenshot 2020-07-15 at 16 30 04

### DIFF
--- a/app/components/provider_interface/reference_with_feedback_component.rb
+++ b/app/components/provider_interface/reference_with_feedback_component.rb
@@ -41,7 +41,7 @@ module ProviderInterface
     def email_address_row
       {
         key: 'Email address',
-        value: mail_to(email_address),
+        value: mail_to(email_address, email_address, class: 'govuk-link'),
       }
     end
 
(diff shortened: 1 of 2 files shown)


app/components/provider_interface/reference_with_feedback_component.rb
```diff
@@ -41,7 +41,7 @@ def name_row
     def email_address_row
       {
         key: 'Email address',
-        value: mail_to(email_address),
+        value: mail_to(email_address, email_address, class: 'govuk-link'),
```
paulrobertlloyd, line 44: @duncanjbrown Am assuming this is correct. I copied the code used for email in personal details summary.

duncanjbrown: Looks good. We should really have a `govuk_mail_to` I reckon

paulrobertlloyd: It’s actually possible to [style links like this by default](https://frontend.design-system.service.gov.uk/sass-api-reference/#govuk-global-styles), using `$govuk-global-styles: true;` in `application.scss`. That said, it’s probably best to use one approach or the other, not a mix of the two.

PR #2520, opened 2020-07-16
Use a single model per set of ProviderRelationshipPermissions

duncanjbrown:
## Context

We used to use two models to model this two-sided relationship, but @tijmenb's spike in #2476 convinced us that one would be better.

This PR (hopefully) complete's Tijmen's refactor. In order to keep this from being totally enormous, I've deleted a lot of code rather than refactor it. The deleted code all pertains to the editing of provider-provider permissions, which we're going to tackle here: https://trello.com/c/g4jGI6XK/487-implement-new-checkbox-structure-for-managing-provider-level-permissions. When we ship that story we'll reincorporate a lot.

## Guidance to review

Commit-by-commit.

## Things to check


spec/factories.rb
```diff
@@ -323,16 +323,7 @@
...
-    training_provider { create(:provider) }
-  end
-
-  factory :training_provider_permissions,
-          class: 'TrainingProviderPermissions' do
-    type { 'TrainingProviderPermissions' }
+  factory :provider_relationship_permissions do
     ratifying_provider { create(:provider) }
     training_provider { create(:provider) }
     setup_at { Time.zone.now }
```
stevehook, line 329: Curious about this, is it meant to distinguish permissions that are saved in the DB after import from Find but not fully populated by the providers?

db/migrate/20200709073219_add_permissions_to_table.rb
```diff
@@ -0,0 +1,21 @@
...
+  def down
+    add_column :provider_relationship_permissions, :make_decisions, :boolean, null: false, default: false
+    add_column :provider_relationship_permissions, :view_safeguarding_information, :boolean, null: false, default: false
+    add_column :provider_relationship_permissions, :type, :string, null: false
+    remove_column :provider_relationship_permissions, :training_provider_can_make_decisions
+    remove_column :provider_relationship_permissions, :training_provider_can_view_safeguarding_information
+    remove_column :provider_relationship_permissions, :ratifying_provider_can_make_decisions
+    remove_column :provider_relationship_permissions, :ratifying_provider_can_view_safeguarding_information
+  end
+end
```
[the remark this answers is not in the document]
mnacos: I guess renaming columns is not an option because of STI (interpretation is dependent on `type`) and adding logic to preserve existing org relationships is not worth it, considering all existing relationships are test data. So we'll need to set up org relationships again in our dev environments, which should be fine but is hampered by the disappearance of `ProviderInterface::ProviderRelationshipPermissionsController`. Any chance we could bring it back?

app/models/provider_relationship_permissions.rb
```diff
@@ -2,8 +2,24 @@ class ProviderRelationshipPermissions < ApplicationRecord
   belongs_to :ratifying_provider, class_name: 'Provider'
   belongs_to :training_provider, class_name: 'Provider'
 
-  VALID_PERMISSIONS = %i[view_safeguarding_information make_decisions].freeze
+  # For ease of use in views, this struct acts like a ProviderPermissions model
+  PermissionsSet = Struct.new(:make_decisions?, :view_safeguarding_information?) do
```
mnacos, line 6: This is fine for display purposes, but considering we'll be passing form objects to components to handle updates, perhaps we should make our default org permissions interface return form objects rather than structs. As in, `#ratifying_provider_permissions` would return a `RatifyingProviderPermissions` ActiveModel-backed object.

On a more general note, it seems to me we're trying to simplify those complex unwieldy new column names (`training_provider_can_make_decisions`) and start using simpler permissions names like `make_decisions` in the code again. Wasn't the original refactor intention to be as specific as possible with the column/method names? Another direction would be to pass the actual permissions model to components, but use different components for showing ratifying and training provider org permissions. I think I would prefer this.

db/migrate/20200709073219_add_permissions_to_table.rb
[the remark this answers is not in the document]
duncanjbrown: This should work, I think because it will effectively delete all the existing permissions but retain the training_provider/ratifying_provider relationships. If the sync happens to run during the migration it will fall over, but the subsequent sync will work.

@mnacos it will come back shortly — would be happy to discuss but my instinct is that the cost of refactoring that code is much greater than us having to type eg `ProviderRelationshipPermissions.find_by(training_provider: x, ratifying_provider: y).update(training_provider_can_manage_organisations: true)` for a few days

db/schema.rb
```diff
@@ -331,15 +331,15 @@
...
   create_table "provider_relationship_permissions", force: :cascade do |t|
-    t.string "type", null: false
     t.integer "training_provider_id", null: false
     t.integer "ratifying_provider_id", null: false
-    t.boolean "view_safeguarding_information", default: false, null: false
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
     t.datetime "setup_at"
-    t.boolean "make_decisions", default: false, null: false
-    t.index ["type", "training_provider_id", "ratifying_provider_id"], name: "index_provider_relationship_permissions_provider_ids_and_type", unique: true
```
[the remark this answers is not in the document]
duncanjbrown: Yes, great idea :+1:

spec/factories.rb
duncanjbrown: Yes - we use this to decide whether to redirect providers to set up orgs when they log in, if the orgs aren't set up yet

app/models/provider_relationship_permissions.rb
duncanjbrown: My understanding is that the form will deal with and validate a pair of permissions at a time, so there's no need to carve out separate form objects, eg https://manage-applications-beta.herokuapp.com/organisations/edit. That form will deal directly with `ratifying_provider_can` et al.

The struct is just so we can use the same template/logic to render permissions on either side of the relationship, as this is the same on both sides, eg: https://manage-applications-beta.herokuapp.com/organisations/show3.

db/migrate/20200709073219_add_permissions_to_table.rb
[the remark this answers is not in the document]
mnacos: Fair enough, we'll use the console for a few days.

I've verified this happens, existing relationships are maintained but permissions disappear (all org relationships fall back to View only).

app/models/provider_relationship_permissions.rb
mnacos: The form object suggestion was just an exploration to see if we can return something a bit more useful than the structs, if `#ratifying_provider_permissions` and `#training_provider_permissions` become the main way of accessing these subsets for a given relationship.

My main reservation is these structs could become an interface for checking permissions, along the lines of ```relationship.ratifying_provider_permissions.make_decisions?```, which doesn't actually look that bad written like this but could become bug-prone if the struct itself is passed to another part of the code, where `make_decisions?` will have lost some of its context. As long as anything decision-related goes via `ProviderAuthorisation`, this is not an issue, of course, but I'd prefer the structs to be generated in the component code that uses them, rather than the model, if their main purpose is to allow components that work seamlessly for both training and ratifying provider permissions.

As already discussed on a call, I'm happy for this to be merged as is and perhaps refactor the way components are made to work for both training and ratifying provider permissions.

duncanjbrown:
@stevehook @mnacos I've added the index _but_ to do it I've had to truncate the table, or there will be records which duplicate tp_id and rp_id — there used to be two per relationship after all. https://github.com/DFE-Digital/apply-for-teacher-training/pull/2520/commits/ec9a75c8f2f7e510347004235e1058ef0d277dda

I've done this in the migration. I think this is ok as the migration is guaranteed to run once only and the table does not contain any real data? I'm inclined to go back and remove it afterwards anyway?

app/models/provider_relationship_permissions.rb
duncanjbrown: @mnacos Thanks for writing this up! We will revisit after @AgaDufrat's next PR is in.

PR #2523, opened 2020-07-17
Provider user invitation wizard - part 1

stevehook:
## Context

This PR is the first part of the implementation of a new wizard based design for the provider interface invite new user flow. It follows on from the spike PR https://github.com/DFE-Digital/apply-for-teacher-training/pull/2469.

This PR covers the forms needed to set basic details, select accessible providers and set their permissions. It doesn't cover the review page and change feature. It's behind a `providers_can_manage_users_and_permissions` feature flag and not yet linked to the users index page.

- [x] Forms for the first 3 steps to match the design https://manage-applications-beta.herokuapp.com/users

image

image

image

- [x] System spec that goes as far as a dummy review page
- [x] Navigation rules for the wizard (back as well as forward).
- [x] Specs for the wizard class.

## Guidance to review

- I've simplified the wizard `next_step` method a bit to avoid complicating it with the logic for making changes via the review page. e.g having the `checking_answers` attribute serialised to wizard state causes problems when the user deviates from the 'normal' path (e.g. by clicking a back button from the review page). In a follow up PR I think this will need a mechanism that's a bit more decoupled from the data for the new user.
- I've left the 'Invite user' button on the users index page pointing to the new implementation for now to keep this small (though I think the existing implementation is behind the same feature flag). So you'll need to navigate directly to `/provider/users/new` to test this -  it should 404 unless the `providers_can_manage_users_and_permissions` feature flag.

## Link to Trello card

## Things to check

### DIFF
--- a/app/forms/provider_interface/provider_user_invitation_wizard.rb
+++ b/app/forms/provider_interface/provider_user_invitation_wizard.rb
@@ -16,6 +16,8 @@ module ProviderInterface
       @state_store = state_store
 
       super(JSON.parse(last_saved_state).deep_merge(attrs))
+
+      self.checking_answers = false if current_step == 'check'
     end
 
     def providers
@@ -96,7 +98,7 @@ module ProviderInterface
   private
 
     def state
-      as_json(except: %w[state_store errors validation_context checking_answers current_step current_provider_id]).to_json
+      as_json(except: %w[state_store errors validation_context current_step current_provider_id]).to_json
     end
 
     def last_saved_state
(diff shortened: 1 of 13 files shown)


app/forms/provider_interface/provider_user_invitation_wizard.rb
```diff
@@ -0,0 +1,134 @@
...
+      if checking_answers
+        [:check]
+      elsif current_step == 'details'
+        [:index]
+      elsif current_step == 'providers'
+        [:details]
+      elsif current_step == 'permissions'
+        previous_provider_id.present? ? [:permissions, previous_provider_id] : [:providers]
+      elsif current_step == 'check'
+        providers.present? ? [:permissions, providers.last] : [:providers]
```
duncanjbrown, line 80: It shouldn't be possible to get here without providers, so I think it's always `[permissions, providers.last]`?

PR #2526, opened 2020-07-17
Add a phase banner to the provider landing page

duncanjbrown:
## Context

This PR follows #2522 (closed). It does some preliminary work for that feature:

- corrects an error in the header html (unclosed `<li>` tags) which corrects the spacing of the menu items
- updates the attrs for the crown svg to match the example of the Design System website
- extracts the phase banner into a component and renders it on all layouts, which means it now also appears on the provider landing page

**Before**
Screenshot 2020-07-17 at 11 21 06

**After**
Screenshot 2020-07-17 at 11 20 27

## Link to Trello card

https://trello.com/c/OTojnjhu/2440-implement-the-new-two-level-header-for-manage

## Things to check


app/components/provider_interface/header_component.html.erb
```diff
@@ -3,7 +3,7 @@
     <div class="govuk-header__logo provider-header__logo">
```
paulrobertlloyd, line 3:
```suggestion
    <div class="govuk-header__logo app-provider-header__logo">
```

Wondering if we should be consistent in how we namespace custom classes… `app-` is typically the norm across GOV.UK services.

duncanjbrown: We should. I've added a card to our tech debt backlog about this: https://trello.com/c/RzV6xdX8/333-review-css-classnames-preferring-app-prefixes-for-custom-css

PR #2527, opened 2020-07-17
Add the two-level header to Manage (v2)

duncanjbrown:
## Context

This PR implements the lower-level grey nav bar in the prototype.

- Add `moj-frontend` bits
- Parameterize the `PhaseBanner` component so it can have its border removed when stacked on top of another element
- Refactor/add `NavigationItems#for_provider_account_nav` and `NavigationItems#for_primary_account_nav`
- touch it in a system spec

## Link to Trello card

https://trello.com/c/OTojnjhu/2440-implement-the-new-two-level-header-for-manage

## Things to check

### DIFF
--- a/app/components/phase_banner.html.erb
+++ b/app/components/phase_banner.html.erb
@@ -1,5 +1,5 @@
 <div class="govuk-width-container">
-  <div class="govuk-phase-banner app-!-print-display-none">
+  <div class="govuk-phase-banner <%= @stacked ? 'app-phase-banner__stacked' : '' %> app-!-print-display-none">
     <p class="govuk-phase-banner__content"><strong class="govuk-tag govuk-phase-banner__content__tag app-environment-tag app-environment-tag--<%= HostingEnvironment.environment_name %>">
       <%= HostingEnvironment.phase %>
     </strong>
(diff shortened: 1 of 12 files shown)


app/components/phase_banner.html.erb
```diff
@@ -1,5 +1,5 @@
 <div class="govuk-width-container">
-  <div class="govuk-phase-banner app-!-print-display-none">
+  <div class="govuk-phase-banner <%= @stacked ? 'app-phase-banner__stacked' : '' %> app-!-print-display-none">
```
adamsilver, line 2: What does ‘stacked’ mean? Is it just that there's no border?

duncanjbrown: That it's to be stacked on top of another element. Suggestion for a better name for this param v welcome!

duncanjbrown:
> Looks great. I would just question the ‘stacked’ class name and param name. If it's literally about removing the border then I think we can be on the nose about it. ‘app-phase-banner--noborder`.
> 
> If we want to be more semantic and generic then `provider-phase-banner`.

Great — thanks. I've made it `app-phase-banner--no-border`.

duncanjbrown:
Just noticed the underline is required when we're on a non-primary nav page. Going to fix

PR #2529, opened 2020-07-17
Fix course replacement bug

tvararu:
## Context

It looks like that for certain courses which have part_time/full_time study modes, the course replacement flow is broken. It will result in a 404, because ActiveRecord attempts to find a CourseOption with an ID of a Course.

Use a `.find_by` instead of a `.find` and exhaustively test all the paths.

## Guidance to review

Opening this as draft since I'm not working on Monday; can someone pick it up:

- [x] Add tests
- [x] Fix other places affected with the same issue (search for `CourseOption.find(params['course_id']).id`)

## Link to Trello card

https://trello.com/c/B7hp3orE/1852-users-cant-add-a-new-course-to-an-application-post-submit

## Things to check


davidgisbey:
:shipit:

PR #2530, opened 2020-07-20
Add provider-specific edit user permissions flow to provider_interface

mnacos:
## Context

A previous PR, https://github.com/DFE-Digital/apply-for-teacher-training/pull/2488, introduced provider-specific permissions 'Change' links within the provider_interface. This PR implements the provider-specific user permissions page these links are intended to point to.

Add two new routes for editing per-provider permissions for a user:

```
GET /provider/provider-users/2/providers/5/permissions
PATCH /provider/provider-users/2/providers/5/permissions
```

Add `edit_permissions` and `update_permissions` actions to `provider_interface/provider_users_controller.rb`.

Add an `edit_permissions.html.erb` view displaying checkboxes. The checkboxes are generated by a component.

Add a new form `ProviderInterface::ProviderPermissionsForm`, responsible for basic validations and persisting permission changes.

## Guidance to review

image

## Link to Trello card

https://trello.com/c/GU90aVjA

## Things to check

### DIFF
--- a/app/components/provider_interface/edit_provider_user_permissions_component.html.erb
+++ b/app/components/provider_interface/edit_provider_user_permissions_component.html.erb
@@ -1,0 +1,38 @@
+<legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
+  <h1 class="govuk-fieldset__heading">
+    <span class="govuk-caption-xl"><%= permissions_form.provider_user.full_name %></span>
+    Change permissions: <%= permissions_form.provider.name %>
+  </h1>
+</legend>
+
+<span class="govuk-hint">
+  The user will still be able to view all applications made to courses at this organisation.
+</span>
+
+<%= form_with model: permissions_form,
+              method: :patch,
+              url: provider_interface_provider_user_update_permissions_path(
+                provider_id: permissions_form.provider.id,
+                provider_user_id: permissions_form.provider_user.id,
+              ) do |f| %>
+  <%= f.govuk_error_summary %>
+  <%= f.hidden_field :provider_id %>
+  <%= f.govuk_check_boxes_fieldset :permissions do %>
+    <%= f.govuk_check_box :manage_organisations, true, multiple: false,
+                          label: { text: 'Manage organisation' },
+                          hint_text: 'Change permissions between organisations' %>
+
+    <%= f.govuk_check_box :manage_users, true, multiple: false,
+                          label: { text: 'Manage users' },
+                          hint_text: 'Invite or delete users and set their permissions' %>
+
+    <%= f.govuk_check_box :view_safeguarding_information, true, multiple: false,
+                          label: { text: 'View safeguarding information' },
+                          hint_text: 'View sensitive material about the candidate' %>
+
+    <%= f.govuk_check_box :make_decisions, true, multiple: false,
+                          label: { text: 'Make decisions' },
+                          hint_text: 'Make offers, amend offers and reject applications' %>
+  <% end %>
+  <%= f.govuk_submit 'Save' %>
+<% end %>
--- a/app/components/provider_interface/provider_user_details_component.rb
+++ b/app/components/provider_interface/provider_user_details_component.rb
@@ -26,7 +26,7 @@ module ProviderInterface
         {
           key: "Permissions: #{permission.provider.name}",
           value: render(PermissionsList.new(permission)),
-          change_path: provider_interface_provider_user_edit_providers_path(@provider_user),
+          change_path: provider_interface_provider_user_edit_permissions_path(@provider_user, provider_id: permission.provider.id),
           action: 'Change',
         }
       end
--- a/app/components/provider_interface/provider_user_details_component.rb
+++ b/app/components/provider_interface/provider_user_details_component.rb
@@ -27,7 +27,7 @@ module ProviderInterface
           key: "Permissions: #{permission.provider.name}",
           value: render(PermissionsList.new(permission)),
           change_path: provider_interface_provider_user_edit_permissions_path(@provider_user, provider_id: permission.provider.id),
-          action: 'Change',
+          action: "Change permissions for #{permission.provider.name}",
         }
       end
     end
--- a/app/controllers/provider_interface/provider_users_controller.rb
+++ b/app/controllers/provider_interface/provider_users_controller.rb
@@ -143,7 +143,7 @@ module ProviderInterface
     end
 
     def assert_current_user_can_manage_users_for(provider)
-      access_denied_for_provider(provider) unless current_user_can_manage_users_for?(provider)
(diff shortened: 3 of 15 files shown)


duncanjbrown:
I get a 400 if I try to remove all permissions:

Screenshot 2020-07-20 at 16 09 48

app/controllers/provider_interface/provider_users_controller.rb
```diff
@@ -120,5 +145,12 @@ def find_provider_user
     rescue ActiveRecord::RecordNotFound
       render_404
     end
+
+    def find_provider_permissions_model
+      ProviderPermissions.find_by(
```
tijmenb, line 150: This will raise `ActiveRecord::RecordNotFound`:

```suggestion
      ProviderPermissions.find_by!(
```

duncanjbrown:
Adam confirmed that "View applications" should only appear when there are no permissions assigned and that the wording should be `The user will only be able to view applications`

adamsilver:
...and on the edit screens in the same sitch will read ‘The user can only view application’

mnacos:
> I get a 400 if I try to remove all permissions:
> 
> Screenshot 2020-07-20 at 16 09 48

Of course, no permissions leads to no `:provider_interface_provider_user_permissions_form` params, which is required. I've fixed this by adding a hidden field :-\ Any better ideas?

app/controllers/provider_interface/provider_users_controller.rb
mnacos: Makes sense, thanks

app/controllers/provider_interface/provider_users_controller.rb
```diff
@@ -1,7 +1,7 @@
 module ProviderInterface
   class ProviderUsersController < ProviderInterfaceController
     before_action :require_feature_flag
-    before_action :redirect_unless_permitted_to_manage_users
+    before_action :require_manage_users_permission
```
tijmenb, line 4: These don't have to change once https://github.com/DFE-Digital/apply-for-teacher-training/pull/2543 is merged.

app/controllers/provider_interface/provider_users_controller.rb
```diff
@@ -120,5 +175,12 @@ def find_provider_user
     rescue ActiveRecord::RecordNotFound
       render_404
     end
+
+    def find_provider_permissions_model!
+      ProviderPermissions.find_by!(
+        provider_user: find_provider_user,
+        provider: Provider.find(params[:provider_id]),
+      )
```
tijmenb, line 183: If we change this into the query below, it won't be possible to find the model for permissions the user does not have access to. I think that means we can get rid of `assert_current_user_can_manage_users_for` and associated methods?

```suggestion
      find_provider_user.provider_permissions.find_by!(
        provider_id: params[:provider_id],
        manage_users: true,
      )
```

mnacos: On the surface it appears so, but the difference is subtle. `find_provider_permissions_model!` returns the relationship between the user being edited and a provider, but the `manage_users` check is for the `current_provider_user` and their relationship to the same provider.

app/components/provider_interface/edit_provider_user_permissions_component.html.erb
```diff
@@ -0,0 +1,38 @@
+<legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
```
tijmenb, line 1: Future PR: I think this should go into the `<form>` (specifically `govuk_check_boxes_fieldset`).

app/components/provider_interface/provider_user_details_component.rb
```diff
@@ -26,7 +26,7 @@ def permission_rows
         {
           key: "Permissions: #{permission.provider.name}",
           value: render(PermissionsList.new(permission)),
-          change_path: provider_interface_provider_user_edit_providers_path(@provider_user),
+          change_path: provider_interface_provider_user_edit_permissions_path(@provider_user, provider_id: permission.provider.id),
           action: 'Change',
```
tijmenb, line 30: Future PR: to disambiguate the links for specs & screenreaders:

```suggestion
          action: "Change permissions for #{permission.provider.name}"
```

app/controllers/provider_interface/provider_users_controller.rb
tijmenb: Ah right!

app/controllers/provider_interface/provider_users_invitations_controller.rb
```diff
@@ -39,10 +39,22 @@ def update_providers
...
+      permission_struct = Struct.new(:slug, :name)
+      @permissions = [
+        permission_struct.new('manage_users', 'Manage users'),
+        permission_struct.new('make_decisions', 'Make decisions'),
+      ]
+
+      permissions_form_struct = Struct.new(:id, :provider_id, :permissions)
+      @permissions_form = permissions_form_struct.new(
+        @provider.id,
+        @provider.id,
```
tijmenb, line 55: Is it correct that we have `@provider.id` twice here? Perhaps this change needs another look, the creation of `Structs` inside the method seems like something that could be cleaned up.

tijmenb: Talked about this on slack, this is a bit gnarly but generally harmless. Refactor later!

app/components/provider_interface/edit_provider_user_permissions_component.html.erb
mnacos: Ok, will keep in mind.

app/controllers/provider_interface/provider_users_invitations_controller.rb
mnacos: Yes, this is best handled as a wizard refactor.

app/components/provider_interface/provider_user_details_component.rb
mnacos: Decided to add this, didn't break anything.

app/components/provider_interface/edit_provider_user_permissions_component.html.erb
```diff
@@ -0,0 +1,38 @@
...
+                          label: { text: 'Manage users' },
+                          hint_text: 'Invite or delete users and set their permissions' %>
+
+    <%= f.govuk_check_box :view_safeguarding_information, true, multiple: false,
+                          label: { text: 'View safeguarding information' },
+                          hint_text: 'View sensitive material about the candidate' %>
+
+    <%= f.govuk_check_box :make_decisions, true, multiple: false,
+                          label: { text: 'Make decisions' },
+                          hint_text: 'Make offers, amend offers and reject applications' %>
```
AgaDufrat, line 35: I'm working on this page now. Is it ok if I swap those permissions around so the order is more consistent with the prototype and other pages where we list these?

PR #2541, opened 2020-07-21
Provider user invitation wizard - part 2 - Saving new users

stevehook:
## Context

This PR is the second part of the implementation of a new wizard based design for the provider interface invite new user flow. It follows on from PR https://github.com/DFE-Digital/apply-for-teacher-training/pull/2523

This PR covers persistence of new and existing users to the database following input via the wizard. It doesn't cover the review page and change feature. It's behind a providers_can_manage_users_and_permissions feature flag and not yet linked to the users index page.

- [x] Added a new `ProviderInterface#SaveProviderUserService` to implement the logic to either create or update a provider user based on state of the form model (`ProviderUserInvitationWizard `)
- [x] Removed the `ProviderUserInvitationWizard#save!` method - controller interacts directly with services to do persistence.
- [x] Integrate the new service with the existing `SaveAndInviteProviderUser` and `InviteProviderUser` services to implement the `ProviderUsersInvitationsController#commit` action.
- [x] Extend new system spec to assert that user is created in the database and that they get an invitation email.

## Guidance to review

- To some extent `ProviderInterface#SaveProviderUserService` is repeating work done by the existing `SaveProviderUser` service. I looked at reusing that but it seemed too tightly coupled to other parts to work with the new wizard.
- To get `SaveAndInviteProviderUser` and `InviteProviderUser` to work with the new service I had to extend `InviteProviderUser` to accept the `provider_user` as an email address as well as a model. This seemed like the easiest way to keep the various bits reasonably decoupled.

## Link to Trello card

## Things to check

### DIFF
--- a/spec/system/provider_interface/provider_invites_user_with_new_wizard_spec.rb
+++ b/spec/system/provider_interface/provider_invites_user_with_new_wizard_spec.rb
@@ -76,7 +76,7 @@ RSpec.feature 'Provider invites a new provider user using wizard interface' do
     @another_provider = Provider.find_by(code: 'DEF')
   end
 
-  def and_i_can_manage_users_for_two_provider
+  def and_i_can_manage_users_for_two_providers
     @provider_user.provider_permissions.find_by(provider: @provider).update(manage_users: true)
     @provider_user.provider_permissions.find_by(provider: @another_provider).update(manage_users: true)
   end
--- a/spec/system/provider_interface/provider_invites_user_with_new_wizard_spec.rb
+++ b/spec/system/provider_interface/provider_invites_user_with_new_wizard_spec.rb
@@ -46,6 +46,7 @@ RSpec.feature 'Provider invites a new provider user using wizard interface' do
 
     when_i_commit_changes
     then_i_see_the_new_user_on_the_index_page
+    and_new_user_gets_an_invitation_email
   end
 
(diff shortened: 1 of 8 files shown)


spec/system/provider_interface/provider_invites_user_with_new_wizard_spec.rb
```diff
@@ -19,7 +19,7 @@
     when_i_visit_the_invite_user_wizard
     then_i_see_a_404_page
 
-    and_i_can_manage_users_for_a_provider
+    and_i_can_manage_users_for_two_provider
```
duncanjbrown, line 22:
```suggestion
    and_i_can_manage_users_for_two_providers
```

duncanjbrown:
This looks very good — just need to resolve the above design question

duncanjbrown:
Just spoke to @adamsilver about this. His view is that the invite process should be additive — ie if we don't select things in this process they shouldn't be knocked out of an existing record. This seems sensible to me

stevehook:
> Just spoke to @adamsilver about this. His view is that the invite process should be additive — ie if we don't select things in this process they shouldn't be knocked out of an existing record. This seems sensible to me

ok thanks @duncanjbrown that makes sense given the way this is going to be used, will get that change made now.

PR #2543, opened 2020-07-21
Centralise `can_manage_users?` permission

tijmenb:
## Context

In our retro we discussed whether the `ProviderAuthorisation` class is now the source of truth for authorisation-related things. It isn't, but we thought it should be. This is a small refactor to start.

We check if the user has "manage_users" permissions in 3 places. This PR centralises the usage into `ProviderAuthorisation`.

To make this easier, I've introduced a `ProviderUser#authorisation` shortcut to make it easier to query whether or not they have the correct permissions.

## Guidance to review

Does it make sense? Is `provider_user.authorisation` a good idea?

## Link to Trello card

https://trello.com/c/DqwdYmu1/335-we-dont-have-canviewapplication-in-provider-authorisation

## Things to check

### DIFF
--- a/app/controllers/provider_interface/provider_users_invitations_controller.rb
+++ b/app/controllers/provider_interface/provider_users_invitations_controller.rb
@@ -111,8 +111,7 @@ module ProviderInterface
     end
 
     def redirect_unless_permitted_to_manage_users
-      can_manage_users = ProviderPermissions.exists?(provider_user: current_provider_user, manage_users: true)
-      render_404 unless can_manage_users
+      render_404 unless current_provider_user.authorisation.can_manage_users?
     end
   end
 end
--- a/app/controllers/provider_interface/provider_users_invitations_controller.rb
+++ b/app/controllers/provider_interface/provider_users_invitations_controller.rb
@@ -1,7 +1,7 @@
 module ProviderInterface
   class ProviderUsersInvitationsController < ProviderInterfaceController
-    before_action :require_feature_flag
-    before_action :redirect_unless_permitted_to_manage_users
+    before_action :require_feature_flag!
+    before_action :require_manage_user_permission!
 
     def edit_details
       @wizard = ProviderUserInvitationWizard.new(session, current_step: 'details')
@@ -106,11 +106,11 @@ module ProviderInterface
         .permit(provider_permissions: {})
     end
 
-    def require_feature_flag
+    def require_feature_flag!
       render_404 unless FeatureFlag.active?(:providers_can_manage_users_and_permissions)
     end
 
-    def redirect_unless_permitted_to_manage_users
+    def require_manage_user_permission!
       render_404 unless current_provider_user.authorisation.can_manage_users?
     end
   end
--- a/app/controllers/provider_interface/provider_users_invitations_controller.rb
+++ b/app/controllers/provider_interface/provider_users_invitations_controller.rb
@@ -111,7 +111,7 @@ module ProviderInterface
     end
 
     def require_manage_user_permission!
-      render_404 unless current_provider_user.authorisation.can_manage_users?
+      render_404 unless current_provider_user.authorisation.can_manage_users_for_at_least_one_provider?
     end
   end
 end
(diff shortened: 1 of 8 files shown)


app/controllers/provider_interface/provider_users_invitations_controller.rb
```diff
@@ -106,13 +106,12 @@ def permissions_params
...
-    def require_feature_flag
+    def require_feature_flag!
       render_404 unless FeatureFlag.active?(:providers_can_manage_users_and_permissions)
     end
 
-    def redirect_unless_permitted_to_manage_users
-      can_manage_users = ProviderPermissions.exists?(provider_user: current_provider_user, manage_users: true)
-      render_404 unless can_manage_users
+    def require_manage_user_permission!
+      render_404 unless current_provider_user.authorisation.can_manage_users?
```
mnacos, line 114: How about a 403 in this case (user lacks permission)? 404 is still a good fit for feature flags.

app/services/provider_authorisation.rb
```diff
@@ -3,6 +3,13 @@ def initialize(actor:)
     @actor = actor
   end
 
+  def can_manage_users?
```
duncanjbrown, line 6: I did a quick double take when I saw there was no provider here, but it makes sense under the circumstances. I wonder if it should be `can_manage_users_for_any_providers?`. Then we could have `can_manage_users_for(provider:)`

app/models/provider_user.rb
```diff
@@ -40,8 +40,8 @@ def full_name
     "#{first_name} #{last_name}" if first_name.present? && last_name.present?
   end
 
-  def can_manage_users?
-    provider_permissions.exists?(manage_users: true)
+  def authorisation
```
duncanjbrown, line 43: I think this is good

app/services/provider_authorisation.rb
mnacos, line 6: There is no universal `can_manage_users?` permission, the ability to manage users is granted on a per-provider basis.

Something like 
```suggestion
  def can_manage_users_for?(provider)
```
would be more accurate.

See also https://github.com/DFE-Digital/apply-for-teacher-training/pull/2530/files#diff-6c36ce09043ee29972a80d5a8862669dR154

and https://github.com/DFE-Digital/apply-for-teacher-training/pull/2530/files#diff-6c36ce09043ee29972a80d5a8862669dR141

app/controllers/provider_interface/provider_users_invitations_controller.rb
duncanjbrown, line 114: I like seeing `.authorisation.` in the chain here rather than not due to some magic delegation

app/services/provider_authorisation.rb
tijmenb: Renamed in https://github.com/DFE-Digital/apply-for-teacher-training/pull/2543/commits/553111bebab0b4bcebaf6074d2f77767676bbca0.

tijmenb: I think it's just named confusingly, not flawed. Its purpose is to check whether or not the user has permission to manage users on at least one org so we can show the page in the navigation. @duncanjbrown had the same comment in https://github.com/DFE-Digital/apply-for-teacher-training/pull/2543#discussion_r458228462. I've renamed the method in https://github.com/DFE-Digital/apply-for-teacher-training/commit/553111bebab0b4bcebaf6074d2f77767676bbca0.

app/controllers/provider_interface/provider_users_invitations_controller.rb
tijmenb: I think it's good practice to never return 403s, because it reveals the existence of things that the user shouldn't be able to see. In this case it doesn't matter much, but it's best to avoid it altoghether to not accidentally introduce it when it does.

mnacos: We've been using 404s for most things so far and this is fine, but for permissions I wanted to suggest 403s in conjunction with https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/controllers/provider_interface/provider_interface_controller.rb#L40, which explains to the user which permission is missing and the relevant provider, so that users can self-serve (i.e. ask the relevant provider to grant them access).

image

Not a big deal either way.

app/services/provider_authorisation.rb
mnacos: Great, thanks

PR #2546, opened 2020-07-21
Hide Languages section of Personal Details when EFL is active

malcolmbaig:
## Context

We're adding an 'English as a foreign language' section to the
application form, visible to any candidate with a nationality other than
British and Irish. This has a lot of overlap with the existing Languages
page of the Personal Details section, so we should hide it if the EFL
feature is active.

- Add an object which checks if we should hide the Languages page. This
  checks if the EFL feature is active, and if one of the following is
  true:
  * The application_form is unsubmitted
  * The application_form#english_main_language field is nil (ie - it's
    never been completed)
- Use this object to:
  * Skip the Languages page when filling in Personal Details
  * Hide the corresponding Review rows.
  * Allow the section to be marked complete even if the Languages data
    is missing.
## Guidance to review

- Enable efl_section feature flag
- Via support, go to an application that's unsubmitted
- Step through Personal Details, noting that it doesn't require you to complete the Languages page.

### Questions:

- Does this approach make sense? Is there a simpler alternative?
- Could we just rework the existing questions so that we don't necessarily need to hide that part of the form? Information about other languages the candidate speaks is one thing that the new EFL section doesn't cover, so we may want to preserve this somehow.
- What are the implications for the API? Is there anything that needs updating? 
- Do we need a migration strategy in place to eventually remove the DB fields related to this page? Is this something we'd even want to do?

## Link to Trello card

https://trello.com/c/9fk7yluU

## Things to check

### DIFF
--- a/app/components/candidate_interface/personal_details_review_component.rb
+++ b/app/components/candidate_interface/personal_details_review_component.rb
@@ -21,9 +21,13 @@ module CandidateInterface
     end
 
     def rows
-      CandidateInterface::PersonalDetailsReviewPresenter
-        .new(personal_details_form: @personal_details_form, nationalities_form: @nationalities_form, languages_form: @languages_form, right_to_work_form: @right_to_work_or_study_form)
-        .rows
+      CandidateInterface::PersonalDetailsReviewPresenter.new(
+        personal_details_form: @personal_details_form,
+        nationalities_form: @nationalities_form,
+        languages_form: @languages_form,
+        application_form: @application_form,
+        right_to_work_form: @right_to_work_or_study_form,
+      ).rows
     end
 
     def show_missing_banner?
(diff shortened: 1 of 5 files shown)


tvararu:
I'll give the code itself a proper review when the previous PRs are merged and this is rebased on master 👍

The review app behaves correctly!

> Does this approach make sense? Is there a simpler alternative?

Will re-review the code later to say for sure but overall sounds good!

> Could we just rework the existing questions so that we don't necessarily need to hide that part of the form? Information about other languages the candidate speaks is one thing that the new EFL section doesn't cover, so we may want to preserve this somehow.

Sounds like a design question, could you pass this by @paulrobertlloyd in Slack?

> What are the implications for the API? Is there anything that needs updating?

I'm not sure... maybe we could massage the IELTS/TOEFL info into `english_language_details` on the current V1 API? One to chat with @duncanjbrown, I can start a thread in Slack.

> Do we need a migration strategy in place to eventually remove the DB fields related to this page? Is this something we'd even want to do?

🤔 I think we should keep them for now

malcolmbaig:
Going to rework this in light of changes made in the International Personal Details feature. Will reopen soon.

malcolmbaig:
Reworked PR here https://github.com/DFE-Digital/apply-for-teacher-training/pull/2585

PR #2547, opened 2020-07-21
Implement new checkbox structure for managing provider level permissions

steventux:
## Context

We've iterated the design for managing provider relationship permissions so that we group inputs by permission rather than provider.

Revives some of the provider relationship permissions UI temporarily removed in https://github.com/DFE-Digital/apply-for-teacher-training/pull/2520 but using the simpler single permissions model.

Reorganises the edit permissions form so that fields are grouped by permission.

image

#### Validating that at least one organisation has permission

image

## Guidance to review

The bulk of the complexity in this PR is how we validate seemingly unrelated attributes on the `ProviderRelationshipPermissions` model as pairs of permissions. This makes for a slightly awkward params structure like:

```ruby
[
  { make_decisions: { training_provider_can_make_decisions: 'true', ratifying_provider_can_make_decisions: 'true' },
  { view_safeguarding_information: { ... },
}]
```
This allows us to highlight the correct fieldset in the form when no permissions have been selected.
The last few commits deal with validation so it might make sense to review as separate commits.

## Link to Trello card

https://trello.com/c/g4jGI6XK/2457-implement-new-checkbox-structure-for-managing-provider-level-permissions

## Things to check

### DIFF
--- a/app/controllers/provider_interface/provider_relationship_permissions_controller.rb
+++ b/app/controllers/provider_interface/provider_relationship_permissions_controller.rb
@@ -1,0 +1,61 @@
+module ProviderInterface
+  class ProviderRelationshipPermissionsController < ProviderInterfaceController
+    before_action :render_404_unless_permissions_found
+    before_action :render_403_unless_access_permitted
+
+    def edit
+      initialize_form
+    end
+
+    def update
+      initialize_form
+
+      if @form.update!(provider_relationship_permissions_form_params)
+        flash[:success] = 'Permissions successfully changed'
+        redirect_to provider_interface_organisation_path(provider_relationship_permissions.training_provider)
+      else
+        flash[:warning] = 'Unable to save permissions, please try again. If problems persist please contact support.'
+        render :edit
+      end
+    end
+
+  private
+
+    def initialize_form
+      @form = ProviderRelationshipPermissionsForm.new(permissions: provider_relationship_permissions)
+    end
+
+    def provider_relationship_permissions
+      @provider_relationship_permissions ||= ProviderRelationshipPermissions.find_by(provider_relationship_params)
+    end
+
+    def provider_relationship_params
+      params.permit(:ratifying_provider_id, :training_provider_id).to_h
+    end
+
+    def provider_relationship_permissions_form_params
+      return {} unless params.key?(:provider_interface_provider_relationship_permissions_form)
+
+      params
+        .require(:provider_interface_provider_relationship_permissions_form)
+        .require(:permissions)
+        .permit(*ProviderRelationshipPermissions.permissions_fields)
+        .to_h
+    end
+
+    def provider_relationship_permissions_params
+      provider_relationship_permissions_form_params.fetch(:permissions, {})
+    end
+
+    def render_404_unless_permissions_found
+      render_404 if provider_relationship_permissions.blank?
+    end
+
+    def render_403_unless_access_permitted
+      training_provider = provider_relationship_permissions.training_provider
+
+      render_403 unless ProviderAuthorisation.new(actor: current_provider_user)
+        .can_manage_organisation?(provider: training_provider)
+    end
+  end
+end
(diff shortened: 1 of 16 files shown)


app/controllers/provider_interface/provider_relationship_permissions_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+    def initialize_form
+      @form = ProviderRelationshipPermissionsForm.new(permissions: provider_relationship_permissions)
+    end
+
+    def provider_relationship_permissions
+      @provider_relationship_permissions ||= ProviderRelationshipPermissions.find_by(provider_relationship_params)
+    end
+
+    def provider_relationship_params
+      params.permit(:ratifying_provider_id, :training_provider_id).to_h
```
duncanjbrown, line 33: Given that all these relationships will exist in the database anyway, would it be simpler to use `:relationship_id`?

spec/system/provider_interface/manage_provider_relationship_permissions_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+    end
+  end
+
+  def and_i_remove_safeguarding_permissions_from_all_providers_and_attempt_to_save
+    within(find('.view-safeguarding-information', match: :first)) { uncheck @ratifying_provider.name }
+    click_on 'Save permissions'
+  end
+
+  def then_i_can_see_a_validation_error_telling_me_to_assign_an_org_to_a_permission
+    expect(page).to have_content('At least one organisation must have permission to view safeguarding information')
```
adamsilver, line 156:
```suggestion
    expect(page).to have_content('Select which organisations can view safeguarding information')
```

app/models/provider_relationship_permissions.rb
```diff
@@ -11,4 +22,14 @@ def training_provider_can_view_applications_only?
   def ratifying_provider_can_view_applications_only?
     PERMISSIONS.map { |permission| send("ratifying_provider_can_#{permission}") }.all?(false)
   end
+
+private
+
+  def at_least_one_active_permission_in_pair
+    PERMISSIONS.each do |permission|
+      if !send("training_provider_can_#{permission}") && !send("ratifying_provider_can_#{permission}")
+        errors.add(permission, "At least one organisation must have permission to #{permission.to_s.humanize.downcase}")
```
adamsilver, line 31:
```suggestion
        errors.add(permission, "Select which organisations can #{permission.to_s.humanize.downcase}")
```

app/controllers/provider_interface/provider_relationship_permissions_controller.rb
steventux: Yes I think so, the only actions that flow into this controller logic will come from the organisations pages which already lookup permissions records or from the setup redirect which have to retrieve these records anyway.

steventux:
@duncanjbrown I've incorporated your suggestion about reworking params and rebased into https://github.com/DFE-Digital/apply-for-teacher-training/pull/2547/commits/78d6170469ff1773e22fb308fd8b17a308fdda3d

app/controllers/provider_interface/provider_relationship_permissions_controller.rb
steventux: @duncanjbrown https://github.com/DFE-Digital/apply-for-teacher-training/pull/2547/commits/45c09755185c84227644894e4aa87ab68a624558 should cover routing to these actions via the permissions ID.

app/models/provider_relationship_permissions.rb
```diff
@@ -2,7 +2,18 @@ class ProviderRelationshipPermissions < ApplicationRecord
   belongs_to :ratifying_provider, class_name: 'Provider'
   belongs_to :training_provider, class_name: 'Provider'
 
-  PERMISSIONS = %w[make_decisions view_safeguarding_information].freeze
+  PERMISSIONS = %i[make_decisions view_safeguarding_information].freeze
+
+  validate :at_least_one_active_permission_in_pair, if: -> { setup_at.present? }
+
+  def self.permissions_fields
```
duncanjbrown, line 9: Do we use this anywhere?

steventux: @duncanjbrown Regarding validation - _I think_ the heroku provider relationships all have `setup_at: nil`, the initial state of a provider-to-provider relationship, this means the validation won't add any errors. Validation works for me locally.
We should never have this situation once the redirect to the permissions setup screens are back in place as this process sets that date.
I've removed the `permissions_fields` method.

PR #2555, opened 2020-07-22
Provider user invitation wizard - part 3 - check page

stevehook:
## Context

This PR is the third part of the implementation of a new wizard based design for the provider interface invite new user flow. It follows on from PR https://github.com/DFE-Digital/apply-for-teacher-training/pull/2541

This PR covers the new design for the 'check' page and change feature. It's behind a `providers_can_manage_users_and_permissions` feature flag.

- [x] Implement summary list with Change actions for each attribute of the new provider user
- [x] Implement mechanism to navigate back to the 'check' page after making an amendment (rather than going to the next page in the wizard)
- [x] Implement/Integrate component to display permissions for each provider to match design
- [x] Integrate into 'Invite user' button on provider users index page
- [x] Fix permissions page - doesn't render currently selected permissions
- [x] Make permissions page consistent with the design and the edit flow.
- [x] Validations on permission and providers pages?

## Guidance to review

- Does the mechanism used to control the flow from Change link to form and back to check page make sense? (using a URL param to set the `checking_answers` switch.)
- Is the authorisation check added to `SaveProviderUserService` enough to prevent bad actors?
- Does the system spec now cover everything it needs to?
- There model behind the permissions page of the wizard has been simplified to fit with changes in other branches. It could still be refactored to be a bit clearer and less verbose perhaps. I would prefer to do that as a separate PR but any ideas welcome.

## Link to Trello card

## Things to check


duncanjbrown:
Looking good — giving a final check!

The top alignment looks a bit off here, and the text should be bold

**Prototype**

Screenshot 2020-07-27 at 11 46 05

**App**

Screenshot 2020-07-27 at 11 46 10

app/forms/provider_interface/provider_user_invitation_wizard.rb
```diff
@@ -122,11 +127,23 @@ def previous_provider_id
...
+          provider_permissions[p.to_s]['permissions'].reject(&:blank?).blank?
+      end
     end
 
-    def any_provider_needs_permissions_setup
+    def any_provider_needs_permissions_setup?
       next_provider_needing_permissions_setup.present?
     end
+
+    def at_least_one_provider_permission
```
duncanjbrown, line 140: This validation works, but I don't think it's required. It should be possible for a user to view applications only for a provider.

app/services/provider_interface/save_provider_user_service.rb
```diff
@@ -16,6 +18,13 @@ def call!
...
 
+    def assert_permissions_for_providers!
+      authorisation = ProviderAuthorisation.new(actor: @actor)
+      return if @wizard.provider_permissions.keys.all? { |provider_id| authorisation.can_manage_users_for?(Provider.find(provider_id)) }
+
+      raise ProviderAuthorisation::NotAuthorisedError, 'You are not allowed to add users to these providers'
+    end
+
     def email_exists?
       ProviderUser.find_by(email_address: wizard.email_address).present?
```
mnacos, line 29: This will fail if a mixed-case email address has been entered.

stevehook: Good point. Will downcase like we do in other 'forms'.

stevehook:
> Looks good to me. Thank you @stevehook!

Thankyou for the pointers @duncanjbrown & @mnacos, will just try to correct that styling and then merge...

stevehook:
> > Looks good to me. Thank you @stevehook!
> 
> Thankyou for the pointers @duncanjbrown & @mnacos, will just try to correct that styling and then merge...

image

PR #2557, opened 2020-07-22
Implement separate page for editing providers

tijmenb:
## Context

This implements the page to add/remove a user to providers, as proposed in the prototype: https://manage-applications-beta.herokuapp.com/users/change-providers 

Change the existing "edit providers" screen into a screen that only allows the editing of providers. This is combined with https://github.com/DFE-Digital/apply-for-teacher-training/pull/2530, which adds a new page to allow permission editing per provider.

The show page looks like this:

image

The edit page now looks like this:

image

And when no orgs are selected:

image

## Guidance to review

Is this as designed?

## Link to Trello card

https://trello.com/c/T4OMwXqA/2466-provider-user-to-adds-removes-other-provider-users-to-from-organisations

## Things to check

### DIFF
--- a/app/forms/provider_interface/provider_user_providers_form.rb
+++ b/app/forms/provider_interface/provider_user_providers_form.rb
@@ -1,0 +1,51 @@
+module ProviderInterface
+  class ProviderUserProvidersForm
+    include ActiveModel::Model
+    attr_accessor :current_provider_user, :provider_user, :provider_ids
+
+    validate :at_least_one_provider_is_selected
+
+    def self.from_provider_user(attributes)
+      form = new(attributes)
+      form.provider_ids = form.provider_user.provider_permissions.pluck(:provider_id)
+      form
+    end
+
+    def providers_that_actor_can_manage_users_for
+      @providers_that_actor_can_manage_users_for ||= begin
+        current_provider_user.authorisation.providers_that_actor_can_manage_users_for
+      end
+    end
+
+    def persisted?
+      true
+    end
+
+    def save
+      return unless valid?
+
+      selected_providers.each do |provider|
+        ProviderPermissions.find_or_create_by!(provider: provider, provider_user: provider_user)
+      end
+
+      not_selected_providers = providers_that_actor_can_manage_users_for - selected_providers
+
+      not_selected_providers.each do |provider|
+        permission = ProviderPermissions.find_by(provider: provider, provider_user: provider_user)
+        permission.destroy! if permission
+      end
+
+      true
+    end
+
+    def at_least_one_provider_is_selected
+      if selected_providers.none?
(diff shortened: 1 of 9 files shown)


app/forms/provider_interface/provider_user_providers_form.rb
```diff
@@ -0,0 +1,52 @@
...
+
+      not_selected_providers.each do |provider|
+        permission = ProviderPermissions.find_by(provider: provider, provider_user: provider_user)
+        permission.destroy! if permission
+      end
+
+      true
+    end
+
+    def at_least_one_provider_is_selected
```
mnacos, line 42: I don't understand this validation. The user in question may belong to providers A,B,C and the current user (then one with `manage_users`) may belong to B, C, D. The current user will see providers B,C,D in the UI, why shouldn't they be allowed to remove the user in question from both B and C?

spec/forms/provider_interface/provider_user_providers_form_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'rails_helper'
+
+RSpec.describe ProviderInterface::ProviderUserProvidersForm do
+  describe '#save' do
+    it 'prevents adding people to organisations they do not have access to' do
+      provider_user = create(:provider_user)
+
+      provider = create(:provider)
+      provider_i_dont_have_access_to = create(:provider)
```
mnacos, line 9: What happens if the current user is associated with a provider but does not have `manage_users` permission for that provider?

app/forms/provider_interface/provider_user_providers_form.rb
duncanjbrown: From our meeting earlier, I think we agreed that this would be handled by an explicit "delete" action, as removing the user from all providers the current user could manage users for would cause them to disappear from the manage-users UI?

mnacos: Ok thanks, makes sense for this iteration.

spec/forms/provider_interface/provider_user_providers_form_spec.rb
mnacos: Ok, turns out we only show the intersection of providers common to both current_provider_user and the user in question, __if__ the current user has `manage_users`. This is because `possible_providers` has a built-in `.manage_users` scope. It's not very clear, perhaps could do with a refactor.

tijmenb: I'm not sure I understand, where would you do a refactor?

mnacos: Sorry, not related to this PR, I think we should rename/move `possible_providers`.

tijmenb: Done!

PR #2560, opened 2020-07-22
Small fixes for the permissions interface

tijmenb:
## Context

Some small UI fixes for the support interface.

See individual commits.

## Link to Trello card

https://trello.com/c/llkC80ei/2394-things-to-check-once-new-provider-user-designs-are-implemented

## Things to check


tijmenb:
Removed the URL change bit because I couldn't get it to work with the wizard, which uses the same path. Will revisit after the wizard is ready.

PR #2561, opened 2020-07-22
Don't show users providers whose relationships they can't manage

duncanjbrown:
## Context

Noticed while working on https://trello.com/c/N3Vdwg4X/2473-org-relationships-are-duplicated-on-the-organisationsshow-page, which will be fixed next.

This PR is about the "Organisations" section of Manage.

When you click on "Organisations", you see a list of providers.

This used to show all the providers you had access to who had relationships to manage AND all the providers on the other side of those relationships.

This PR reduces the contents of that list to providers

1. that have relationships to manage
2. that the current user has manage_organisations permission for

For example, in the old world, you might have manage_organisations permission for provider A, who ratifies courses at providers B, C and D. You would see providers A, B, C, and D in the list of organisations. Clicking on B, C or D would show you each one's relationship with A. Clicking on A would show you all A's relationships (i.e. the same relationships, grouped under A).

After this PR, only provider A will appear in the list. **No information is hidden, it's just less duplicated**.

While we're here, follow #2543 and #2553 in moving permissions-related stuff inside `ProviderAuthorisation`.

**Before**

Screenshot 2020-07-22 at 15 26 46

**After**

Screenshot 2020-07-22 at 15 26 18

## Guidance to review

Does this make sense?

There's a bit of fudge where `providers_that_actor_can_manage_organisations_for` doesn't technically return that, it returns the providers _with organisational relationships_ that the actor can manage organisations for. But I think this is probably right nonetheless.

## Link to Trello card

Towards https://trello.com/c/N3Vdwg4X/2473-org-relationships-are-duplicated-on-the-organisationsshow-page

## Things to check

- [Xx] New environment variables have been [added to the Azure config](https://github.com/DFE-Digital/apply-for-teacher-training#azure-hosting-devops-pipeline)

### DIFF
--- a/app/controllers/provider_interface/organisations_controller.rb
+++ b/app/controllers/provider_interface/organisations_controller.rb
@@ -19,11 +19,11 @@ module ProviderInterface
   private
 
     def manageable_providers
-      @_manageable_providers ||= Provider.with_permissions_visible_to(current_provider_user)
+      @_manageable_providers ||= current_provider_user.authorisation.providers_that_actor_can_manage_organisations_for
     end
 
     def render_403_unless_organisation_valid_for_user
-      render_403 unless Provider.with_permissions_visible_to(current_provider_user).include?(provider)
+      render_403 unless manageable_providers.include?(provider)
     end
 
     def provider
--- a/app/models/navigation_items.rb
+++ b/app/models/navigation_items.rb
@@ -45,7 +45,8 @@ class NavigationItems
 
       items = []
 
-      if current_provider_user.can_manage_organisations? && Provider.with_permissions_visible_to(current_provider_user).exists?
+      if current_provider_user.can_manage_organisations? && \
+          current_provider_user.authorisation.can_manage_organisations_for_at_least_one_provider?
         items << NavigationItem.new('Organisations', provider_interface_organisations_path, is_active(current_controller, %w[organisations provider_relationship_permissions]))
       end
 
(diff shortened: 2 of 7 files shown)


app/controllers/provider_interface/organisations_controller.rb
```diff
@@ -19,11 +19,11 @@ def show
   private
 
     def manageable_providers
-      @_manageable_providers ||= Provider.with_permissions_visible_to(current_provider_user)
+      @_manageable_providers = current_provider_user.authorisation.providers_that_actor_can_manage_organisations_for
```
duncanjbrown, line 22: oops `||=`

app/models/navigation_items.rb
```diff
@@ -45,7 +45,8 @@ def for_provider_account_nav(current_provider_user, current_controller)
 
       items = []
 
-      if current_provider_user.can_manage_organisations? && Provider.with_permissions_visible_to(current_provider_user).exists?
+      if current_provider_user.can_manage_organisations? && \
```
mnacos, line 48: How about a 
```suggestion
      if current_provider_user.autorisation.can_manage_at_least_one_organisation? && \
```
in the style of https://github.com/DFE-Digital/apply-for-teacher-training/pull/2543

duncanjbrown: Nice — done!

PR #2566, opened 2020-07-23
use sslmod=require for database connection

vigneshmsft:
https://www.postgresql.org/docs/9.1/libpq-ssl.html#LIBPQ-SSL-SSLMODE-STATEMENTS

We have ssl enforced on the server, use the require mode on client as well.
use `sslmode=require`

## Things to check

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -22,3 +22,4 @@ test:
 
 production:
   <<: *default
+  sslmode: require


duncanjbrown:
Looks like docker-compose's postgres as-is doesn't support SSL. Which seems fair enough — dev envs shouldn't need to either. Maybe we should set this in prod only?

vigneshmsft:
I'll try and enable ssl in docker-compose, unless we prefer to have it just in production?

duncanjbrown:
> I'll try and enable ssl in docker-compose, unless we prefer to have it just in production?

I think it might be a bit of a pain to set up — & I'm not sure there's any advantage to having it in test?

tvararu:
+1 for limiting this to prod only for now instead of adding SSL to docker-compose 👍

PR #2569, opened 2020-07-23
Update content when a provider can only view applications

steventux:
## Context

We will be validating that at least one of the two providers in a relationship will be authorised for each permission. see: https://github.com/DFE-Digital/apply-for-teacher-training/pull/2547 
So there will only ever be one provider having the default _view applications_ permission.
It doesn't make sense to present this in list form.
  

If a provider has no authorisation for any permission we display this as a sentence rather than a list.

image

## Link to Trello card

https://trello.com/c/YWI3gDG6/2461-handle-when-an-organisation-gets-no-additional-permissions-on-the-details-view

## Things to check


duncanjbrown:
In the event that permissions haven't been set up for those providers — say you're a ratifying provider and the training provider's staff haven't got round to it — I think there will be two providers in the `view_applications_only` state? Should we hide the other headings too, in that case? cc @adamsilver

duncanjbrown:
Just discussed with @adamsilver.

If we're in a pre-setup state, we should show

```
Blackfriars Teaching School Alliance can only view applications.
Gorse SCITT can only view applications.
```

and hide both the (now-empty) "The following organisations can..." sections.

adamsilver:
@steventux On @duncanjbrown’s comment please use one paragraph without check icons as follows:

```
{{training provider}} has not set up permissions for this partnership yet.
```

We have a ticket to refine the content further but this will do for the moment.

PR #2570, opened 2020-07-23
Deal with nil values when listing provider-provider permissions

mnacos:
## Context

Minor fix for https://github.com/DFE-Digital/apply-for-teacher-training/pull/2567 I didn't review in time.

The component breaks because `map` can return nils.

Use `.compact`.

## Guidance to review

Difficult to review, requires the setup of provider-provider relationships in a rails console.

## Link to Trello card

https://trello.com/c/ygSVGvI5/2481-add-provider-provider-permissions-to-edit-view

## Things to check

### DIFF
--- a/app/components/permissions_list.rb
+++ b/app/components/permissions_list.rb
@@ -4,15 +4,15 @@ class PermissionsList < ViewComponent::Base
   end
 
   def training_providers_that_can(permission)
-    permissions_as_ratifying_provider.map do |permission_relationship|
+    permissions_as_ratifying_provider.map { |permission_relationship|
       permission_relationship.training_provider if permission_relationship.send("training_provider_can_#{permission}?")
-    end
+    }.compact
   end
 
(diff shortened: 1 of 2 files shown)


AgaDufrat:
Good spot. We have the same problem here https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/components/provider_interface/edit_provider_user_permissions_component.rb

mnacos:
> Good spot. We have the same problem here https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/components/provider_interface/edit_provider_user_permissions_component.rb

You mean the PermissionsList component, right? Applied the same fix.

mnacos:
> Great. Extra  in the `reduce` tank if you like it
> 
> ```ruby
> things_we_might_not_want_all_of.reduce([]) do |list, thing|
>   list << thing if condition
> end
> ```

Thanks but I prefer `.compact` :)

PR #2572, opened 2020-07-23
Add code for launching `make_decisions` and `manage_users`

mnacos:
## Context

We are preparing to launch a major access controls update to the 'Manage' interface. In particular:
- Provider users without `make_decisions` will be unable to respond to applications
- Provider users can invite/remove other users and manage their permissions (if they have `manage_users`)

To assist the launch and preserve the ability of existing users to respond to applications, we'll ensure:
- All existing provider users automatically acquire the `make_decisions` permission
- The user that has signed the Data Sharing Agreement for each organisation automatically acquires `manage_users` permission for that organisation.

## Guidance to review

I've implemented this code as a worker, rather than a database migration so that we can:
- write tests
- and be able to schedule or trigger the launch manually, outside of a deployment

There is an automated check that `all_providers_have_at_least_one_user_with_manage_users?`. I've excluded providers with no users from this check. This check will also fail and the process will get stuck if a providers with users has not yet signed Data Sharing Agreement (probably because they have never signed in). I've added an error so that we are notified if this happens.

Is there anything else the worker should do? There have been mentions of an automatic email sent to users (all users? users that acquire `manage_users`?) but this could be done as a separate PR.

## Link to Trello card

https://trello.com/c/aIg23O4H

## Things to check


duncanjbrown:
This PR does all the right things... but I think it should do them in a slightly different way!

At the moment the manual intervention required to turn this feature on is to shell in via PIM and run the worker. It would be more convenient to do this by toggling a feature flag.

The user data isn't going to change for a few days, so I think we can safely migrate now and toggle the flag when we're ready. (confirmation from Tranisition team: https://ukgovernmentdfe.slack.com/archives/CNRBP3GMV/p1595516189038200).

Would you mind moving the update code to a migration and removing the flag-toggle?

mnacos:
> This PR does all the right things... but I think it should do them in a slightly different way!
> 
> At the moment the manual intervention required to turn this feature on is to shell in via PIM and run the worker. It would be more convenient to do this by toggling a feature flag.
> 
> The user data isn't going to change for a few days, so I think we can safely migrate now and toggle the flag when we're ready. (confirmation from Tranisition team: https://ukgovernmentdfe.slack.com/archives/CNRBP3GMV/p1595516189038200).
> 
> Would you mind moving the update code to a migration and removing the flag-toggle?

Ok, but I'll remove the check, as we don't really want the migration to fail.

spec/workers/launch_make_decisions_and_manage_users_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    it 'gives all existing provider users make_decisions for all their providers' do
+      expect(ProviderPermissions.count).to eq(4)
+      expect(ProviderPermissions.find_each.all?(&:make_decisions)).to be_truthy
+    end
+  end
+
+  describe '#give_manage_users_to_the_user_who_has_signed_the_dsa!' do
+    it 'does what it says on the tin' do
```
duncanjbrown, line 27: 😂

app/workers/launch_make_decisions_and_manage_users.rb
```diff
@@ -0,0 +1,28 @@
...
+  include Sidekiq::Worker
+  sidekiq_options retry: 0
+
+  # This class is temporary code to assist with this launch and will be removed
+
+  def update_provider_user_permissions!
+    Audited.audit_class.as_user('LaunchMakeDecisionsAndManageUsers task') do
+      give_manage_users_to_the_user_who_has_signed_the_dsa!
+
+      ProviderPermissions.update_all(make_decisions: true)
```
tijmenb, line 11: Note that this doesn't update the audit log, that would have been useful.

PR #2575, opened 2020-07-23
Make the provider signin fallback use authentication tokens

tijmenb:
## Context

We've introduced a better way of dealing with authentication tokens in #2348. The support signin fallback uses a table with tokens instead of a single token on the SupportUser model. This means that if the user requests multiple tokens, they'll all keep working.

Make the provider user fallback use this method as well.

## Guidance to review

Will the functionality be the same?

## Link to Trello card

## Things to check


tijmenb:
I just realised that this allows a provider to sign in to the support UI. Closing to have a good think.

PR #2585, opened 2020-07-27
Hide Languages section of Personal Details when EFL is active

malcolmbaig:
## Context

We're adding an 'English as a foreign language' section to the
application form, visible to any candidate with a nationality other than
British and Irish. This has a lot of overlap with the existing Languages
page of the Personal Details section, so we should hide it if the EFL
feature is active.

- Add a policy object to handle the checks for whether or not we should hide the Languages page, and use it where required.
- Rework all affected tests.

## Guidance to review
- Take particular note of the last two commits, which illustrate a refinement/simplification of the `LanguagesSectionPolicy` boolean check. 
- Does the boolean check make sense? Are there any edge cases this behaviour doesn't account for?

## Link to Trello card

https://trello.com/c/9fk7yluU
## Things to check

### DIFF
--- a/spec/presenters/candidate_interface/personal_details_review_presenter_spec.rb
+++ b/spec/presenters/candidate_interface/personal_details_review_presenter_spec.rb
@@ -225,6 +225,23 @@ RSpec.describe CandidateInterface::PersonalDetailsReviewPresenter do
     end
   end
 
+  context 'when the language rows should be hidden' do
+    before { allow(LanguagesSectionPolicy).to receive(:hide?).and_return true }
+
+    it 'does not show the language rows' do
+      languages_form = build(
+        :languages_form,
+        english_main_language: 'Yes',
+        english_language_details: '',
+        other_language_details: 'Glossolalia',
+      )
+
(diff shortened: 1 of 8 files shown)


spec/presenters/candidate_interface/personal_details_review_presenter_spec.rb
```diff
@@ -51,211 +70,222 @@
...
-          other_nationality1: 'French',
-          other_nationality2: 'German',
-          other_nationality3: 'Spanish',
-        )
-
-        expect(rows(personal_details_form, nationalities_form, languages_form, right_to_work_form)).to include(
-          row_for(:nationality, 'British, French, German, and Spanish', Rails.application.routes.url_helpers.candidate_interface_edit_nationalities_path),
-        )
-      end
+    it 'includes a hash with up to 5 nationaltties when international_personal_details is on' do
```
davidgisbey, line 105:
```suggestion
    it 'includes a hash with up to 5 nationalities when international_personal_details is on' do
```

That was me ^

PR #2590, opened 2020-07-27
Prevent audit trail errors if provider_id cannot be found for permission updates

mnacos:
## Context

It's not always possible to know the provider for which permissions were changed, if the provider user is no longer associated with this provider. This is because removing provider users from an organisation removes the `provider_users_providers` record on which the relevant `ProviderPermissions` model is based.

This is a band aid fix so we can ship the feature. We should consider soft-deleting records from `provider_users_providers` table, which will preserve the relevant audit information.

## Guidance to review

Should be an easy review.

## Link to Trello card

Part of general preparation for https://trello.com/c/k4LxRi4u

## Things to check

### DIFF
--- a/app/components/support_interface/audit_trail_item_component.rb
+++ b/app/components/support_interface/audit_trail_item_component.rb
@@ -64,8 +64,12 @@ module SupportInterface
                             auditable_id: audit.auditable_id,
                             action: 'create',
                           )
-                          provider_id = creation_record&.audited_changes&.dig('provider_id')
-                          Provider.find_by(id: provider_id)&.name || 'a provider'
+
+                          if creation_record
+                            Provider.find_by(id: creation_record.audited_changes['provider_id']).name
+                          else
+                            'a provider'
+                          end
                         end
         "Permissions changed for #{provider_name}"
       when 'destroy'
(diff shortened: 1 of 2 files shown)


duncanjbrown:
Logic looks good  and I think the paranoia is healthy — but maybe too much!

If there's a creation_record it has to have an `audited_changes` because its an `Audited::Audit`. It must also have a `provider_id` or postgres would have rejected it (it's `null: false`). And we never delete `Provider`s. So inside the `else` the logic could be:

```ruby
if creation_record
  Provider.find_by(id: creation_record.audited_changes['provider_id']).name
else
  'a provider'
end
```

Maybe a test too?

mnacos:
> Logic looks good and I think the paranoia is healthy — but maybe too much!
> 
> If there's a creation_record it has to have an `audited_changes` because its an `Audited::Audit`. It must also have a `provider_id` or postgres would have rejected it (it's `null: false`). And we never delete `Provider`s. So inside the `else` the logic could be:
> 
> ```ruby
> if creation_record
>   Provider.find_by(id: creation_record.audited_changes['provider_id']).name
> else
>   'a provider'
> end
> ```
> 
> Maybe a test too?

Makes sense. I've added a test case for when the creation record is missing.

PR #2593, opened 2020-07-28
Remove duplicated org relationships on the organisations show page

AgaDufrat:
## Context

If the user has access to both the ratifying and training providers we should
not be showing the relationship twice, in both directions. We keep the
relationship as a training provider since they are can make changes and the
removed duplicated entry.

Screenshot 2020-07-28 at 12 17 30

## Guidance to review

- visit http://localhost:5000/provider/organisations/:id

## Link to Trello card

https://trello.com/c/N3Vdwg4X/2473-org-relationships-are-duplicated-on-the-organisationsshow-page

## Things to check

### DIFF
--- a/app/controllers/provider_interface/organisations_controller.rb
+++ b/app/controllers/provider_interface/organisations_controller.rb
@@ -7,13 +7,12 @@ module ProviderInterface
     end
 
     def show
-      scope = ProviderRelationshipPermissions.includes(:training_provider, :ratifying_provider)
+      @training_permissions = ProviderRelationshipPermissions.where(training_provider_id: params[:id])
+        .or(ProviderRelationshipPermissions.where(training_provider_id: manageable_providers, ratifying_provider_id: params[:id]))
+        .includes(:training_provider)
 
-      permissions = scope.where(training_provider: manageable_providers)
-        .or(scope.where(ratifying_provider: manageable_providers))
-
-      @ratifying_permissions = permissions.select { |p| manageable_providers.include?(p.ratifying_provider) }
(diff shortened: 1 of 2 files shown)


app/controllers/provider_interface/organisations_controller.rb
```diff
@@ -12,8 +12,8 @@ def show
       permissions = scope.where(training_provider: manageable_providers)
         .or(scope.where(ratifying_provider: manageable_providers))
```
duncanjbrown, line 13: We're still getting every manageable provider! I think replacing `manageable_providers` with `params[:id]` (or equivalent) ought to fix it

AgaDufrat: What would be better 🤔 
```
permissions ||= ProviderRelationshipPermissions.where(training_provider: params[:id])
                .or(ProviderRelationshipPermissions.where(ratifying_provider: params[:id]))
                .includes(:training_provider, :ratifying_provider)

@training_permissions = permissions.where(training_provider_id: params[:id])
@ratifying_permissions = permissions.where(ratifying_provider_id: params[:id])
```
or 
```
@training_permissions = ProviderRelationshipPermissions.where(training_provider_id:                  
       params[:id]).includes(:training_provider)
@ratifying_permissions = ProviderRelationshipPermissions.where(ratifying_provider_id:
       params[:id]).includes(:ratifying_provider)
```

duncanjbrown: I reckon the second one is a bit more readable? As long as the deduplication still works

PR #2598, opened 2020-07-28
Relax check for next provider needing permissions setup

steventux:
## Context

We do not validate that permissions should be non-empty so it doesn't make sense
to exclude empty permissions when calculating the next provider.

This commit attempts to rectify a bug where a user amending a set of permissions
in the wizard will be returned to the same permissions set unless they turn a permission
'on', this isn't a validation requirement so remove the check for empty permissions.
This allows them to return to the review step of the wizard.

## Guidance to review

Steps to replicate original bug:

- Start the invite user wizard flow
- Add a provider
- Assign permissions
- Go to "change" those permissions on the review step
- Remove the permissions and submit
   
you're stuck on the same permissions page

## Link to Trello card

https://trello.com/c/yJUqNuQn/2531-its-possible-to-get-stuck-in-the-invite-provider-user-wizard

## Things to check

### DIFF
--- a/app/forms/provider_interface/provider_user_invitation_wizard.rb
+++ b/app/forms/provider_interface/provider_user_invitation_wizard.rb
@@ -163,10 +163,7 @@ module ProviderInterface
     end
 
     def next_provider_needing_permissions_setup
-      providers.find do |p|
-        provider_permissions.keys.exclude?(p.to_s) ||
-          provider_permissions[p.to_s]['permissions'].reject(&:blank?).blank?
-      end
+      providers.find { |p| provider_permissions.keys.exclude?(p.to_s) }
     end
 
     def any_provider_needs_permissions_setup?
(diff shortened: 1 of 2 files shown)


app/forms/provider_interface/provider_user_invitation_wizard.rb
```diff
@@ -163,10 +163,7 @@ def previous_provider_id
     end
 
     def next_provider_needing_permissions_setup
-      providers.find do |p|
-        provider_permissions.keys.exclude?(p.to_s) ||
-          provider_permissions[p.to_s]['permissions'].reject(&:blank?).blank?
-      end
+      providers.find { |p| provider_permissions.keys.exclude?(p.to_s) }
```
stevehook, line 166: can you just do something like `(providers - provider_permissions.keys).first` or `(providers - provider_permissions.keys.map(&:to_i)).first`?

duncanjbrown:
Fixed! I think we're missing "The user will only be able to view applications" in the `ProviderUserInvitationPermissionsComponent`:

Screenshot 2020-07-28 at 21 32 59

duncanjbrown:
NVM fixed it in #2600 🕯️

PR #2599, opened 2020-07-28
Present new EFL qualification data in Manage and on the API

malcolmbaig:
## Context

We've introduced an 'English as a foreign language' section to the
application. It appears if a candidate selects a nationality other than
British or Irish. Switching on the feature also hides the Languages
section of Personal Details if the candidate hasn't already filled this
in, as there is overlap between the two sets of questions and EFL is
intended as a replacement.

If the Languages section is hidden, the application can be submitted
with the following database fields containing null values:

- english_main_language
- english_language_details

Both are used in the Vendor API and in the Manage interface. To ensure
we continue to provide consistent data in these fields, we need to infer
values for them if they are blank.

Override methods as follows:

- english_main_language
  * Return true if the candidate is British/Irish OR if they declare
    that they don't need an English qualification.
  * Return false otherwise.
- english_language_details
  * Return a formatted string description of the candidate's EFL
    qualification, if one has been provided.

## Guidance to review
- Is the method overriding likely to confuse at all further down the road? Is there an approach that could be clearer?
- Suggested steps for testing in a review app:
  - Ensure `efl_section` feature flag is enabled.
  - Complete application as a new candidate, selecting a nationality other than British/Irish in the Personal Details section.
  - You shouldn't see the Languages page when completing Personal Details.
  - Complete the EFL section.
  - Submit the application.
  - View it via the Manage interface and the API.

## Link to Trello card

https://trello.com/c/nlQgcKnR
## Things to check

### DIFF
--- a/spec/system/vendor_api/vendor_receives_application_spec.rb
+++ b/spec/system/vendor_api/vendor_receives_application_spec.rb
@@ -101,7 +101,7 @@ RSpec.feature 'Vendor receives the application' do
           uk_residency_status: nil,
           english_main_language: true,
           other_languages: "I'm great at Galactic Basic so English is a piece of cake",
-          english_language_qualifications: '',
+          english_language_qualifications: nil,
           disability_disclosure: 'I have difficulty climbing stairs',
         },
         qualifications: {
--- a/app/forms/candidate_interface/languages_form.rb
+++ b/app/forms/candidate_interface/languages_form.rb
@@ -13,7 +13,9 @@ module CandidateInterface
 
     def self.build_from_application(application_form)
       new(
-        english_main_language: boolean_to_word(application_form[:english_main_language]),
+        english_main_language: boolean_to_word(
+          application_form.english_main_language(fetch_database_value: true),
+        ),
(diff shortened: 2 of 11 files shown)


duncanjbrown:
Hi @malcolmbaig — sorry I haven't got to this yet! will do before standup tomorrow

spec/system/vendor_api/vendor_receives_application_spec.rb
```diff
@@ -101,7 +101,7 @@ def then_it_should_include_the_data_from_the_application_form
           uk_residency_status: nil,
           english_main_language: true,
           other_languages: "I'm great at Galactic Basic so English is a piece of cake",
-          english_language_qualifications: '',
+          english_language_qualifications: nil,
```
duncanjbrown, line 104: Not for this PR, but maybe we should update `CandidateHelper.candidate_completes_application_form` for the new flow anyway?

malcolmbaig:
> Would you consider changing the "legacy" addressing to take a parameter on the method instead of the more subtle `[]` syntax?

That's such a good idea. Will get this updated.

PR #2600, opened 2020-07-28
Fix UX inconsistencies around ProviderUsers

duncanjbrown:
## Context

These cropped up in a final design review.

See commits. Note that this changeset introduces a possible _new_ bug — 🎉 — the "Account" page now shows "This user can only view applications" if you're in that state. A bit weird if you don't usually refer to yourself in the third person.

Fixing that is marginally more complicated than fixing all this, so I've preferred the latter & will add a backlog card.

## Link to Trello card

https://trello.com/c/6C0lcv6X/2533-manage-user-permissions-ux-discrepancies

## Things to check

### DIFF
--- a/spec/components/permissions_list_component_spec.rb
+++ b/spec/components/permissions_list_component_spec.rb
@@ -13,15 +13,34 @@ RSpec.describe PermissionsList do
 
   it 'renders permissions' do
     permission_model = create(:provider_permissions, manage_organisations: true)
-    result = render_inline(described_class.new(permission_model))
+    result = render_inline(described_class.new(permission_model, user_is_viewing_their_own_permissions: false))
 
     expect(result.css('li').text).to include('Manage organisations')
     expect(result.css('li').text).not_to include('The user can only view applications')
   end
 
+  describe 'rendering View applications only' do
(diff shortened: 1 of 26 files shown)


spec/components/permissions_list_component_spec.rb
```diff
@@ -13,17 +13,36 @@
...
+  describe 'rendering View applications only' do
+    it 'shows an appropriate message when the user is viewing another user’s permissions' do
+      permission_model = create(:provider_permissions)
+      result = render_inline(described_class.new(permission_model, user_is_viewing_their_own_permissions: false))
+
+      expect(result.css('li').text).to include('The user can only view applications')
+      expect(result.css('li').text).not_to include('Manage organisations')
+      expect(result.css('li').text).not_to include('Make manage users')
+      expect(result.css('li').text).not_to include('Make decistions')
+      expect(result.css('li').text).not_to include('View safeguarding information')
```
mnacos, line 31: Super minor, shouldn't this be 'Access safeguarding information'?

duncanjbrown: Yes! Will amend

PR #2601, opened 2020-07-29
Spike for stopping new applications on 24th August

stevehook:
# :rotating_light: DO NOT MERGE :rotating_light:

## Context

We need to prevent candidates from submitting applications after 24th August so that candidates have 2 weeks to get references before the 'end of cycle' date. This PR is a Spike to explore the requirement and a potential implementation.

- [x] Feature flag `stop_new_applications` - maybe it should be a more generic name like `end_of_cycle`.
- [ ] Info banner displayed n weeks prior to cutoff explaining that time is running out for candidates that have not yet submitted. (Content needed).
- [x] Info banner explaining to candidates that new applications are suspended. (Content needed).

image

- [x] Hide the _Continue_ button on the application review page.
- [ ] Redirect to application page with an error message (content needed) if user does somehow...
  - [x] Confirm their application
  - [ ] Submit their application
  - [ ] Visit the Equality / Diversity page
  - [ ] Visit the Submit page
- [x] `EndOfCycleTimetable` class to encapsulate the significant dates in the calendar. (`apply_1_cutoff`, `apply_2_cutoff`, `reopen`)
- [x] Hide the _Add another course_ button

Before:
image
After:
image

- [x] Redirect to application page with an error message (content needed) if user does visit the add course pages
- [ ] Do we need a banner on the signup pages to warn candidates that they will have to wait until 9th September to choose courses and submit their application?
- [ ] Throw an exception in the `SubmitApplication`, `PickCourseOption` and any other relevant services if they are invoked during the closure?
- [ ] Change the copy for _Check and submit_ section of the application page to just say _Review_ ?

Before:
image

## Guidance to review

Does the list of proposed changes above (including the bits not done yet) make sense? What's missing?

Is the technical approach partly implemented in this spike sensible for...
- Using a feature flag just to control the rollout of this feature (spanning several cards/PRs)?
- Having the cutoff dates in a service class (`EndOfCycleTimetable`)? These are hard-coded with a view to reusing the functionality in future years.
- Stopping candidates submit their forms? (combination of hiding buttons/links and redirecting with error message)
- Stopping candidates adding more courses to their applications? (combination of hiding buttons/links and redirecting with error message) - I think the requirement is also to remove all existing application choices from applications that are unsubmitted at the cutoff point too.
- Where in the application and when do we need to warn candidates about the closure? e.g. do we need to warn in advance of the actual cutoff?

## Link to Trello card

https://trello.com/c/PJXub5CK/1873-🏈-tech-spike-stop-apply-1-on-24th-aug

## Things to check


paulrobertlloyd:
This looks good, me and @EmmaFrith have been thinking along similar lines.

One thing however, is the distinction between candidates who have yet to submit their first application (Apply 1), and those applying again (Apply 2). The later group can still submit their application until 18 September, so we have a situation where between 24 August and 18 September, some people can submit, and some can’t. Does this take account of this?

paulrobertlloyd:
Thought was that once you can’t choose a course, we no longer link to that page from the application index, which may make things a bit simpler. (I guess we’d still have to redirect away from course choice pages if candidates somehow found themselves there).

stevehook:
> This looks good, me and @EmmaFrith have been thinking along similar lines.
> 
> One thing however, is the distinction between candidates who have yet to submit their first application (Apply 1), and those applying again (Apply 2). The later group can still submit their application until 18 September, so we have a situation where between 24 August and 18 September, some people can submit, and some can’t. Does this take account of this?

This spike was only meant to take account of Apply 1 but looking at it now I think we need to deal with both 1 & 2 together.

stevehook:
> Thought was that once you can’t choose a course, we no longer link to that page from the application index, which may make things a bit simpler. (I guess we’d still have to redirect away from course choice pages if candidates somehow found themselves there).

That would work, though it would mean that candidates can't drill down to just view the courses they already selected.

paulrobertlloyd:
> though it would mean that candidates can't drill down to just view the courses they already selected.

Additional thought is that once you can no longer choose courses for this cycle, we remove previous course choices from your application. That idea may need some discussion, but avoids the risk of having a course on an application that’s no longer available, and possibly easier to explain than if had a mix of choices that are/are not available in the next cycle.

stevehook:
I've modified the spike implementation to cater for Apply 1 and 2 by adding _another_ feature flag. But this is starting to feel messy - feature flags are not meant for fine-grained changes to functionality at pre-determined times. So I think a better approach is to put the overall 'feature' behind one flag and have a timetable of the various cutoff dates that can be hard-coded for now. These would include at least:

- 24th August: the cutoff for new Apply 1 applications
- 18th September: the cutoff for new Apply 2 applications
- 13th October: open 2021 cycle

This would be structured to allow for a different set of dates for each year (we would want the implementation to be generic wrt the yearly timetable).

stevehook:
Closing this for now as we've learnt what we needed to learn from it. After the design is specified we'll probably break the work down into smaller tasks but can potentially reuse some of the ideas and code here.

PR #2602, opened 2020-07-29
Properly remove provider users

tijmenb:
## Context

@mnacos noticed that the audit log wasn't being updated when a user is deleted from the provider UI. It turns out that `users.providers =` does not update the audit log (added in https://github.com/DFE-Digital/apply-for-teacher-training/pull/2012, cc @steventux).

This PR changes the way we "delete" a user so that we remove individual permissions, which updates the log. For extra info we also add an audited comment.

## Guidance to review

image

## Link to Trello card

https://trello.com/c/GvQ0UIli/2339-removing-a-user-from-a-provider-doesnt-always-update-the-audit-log

## Things to check

### DIFF
--- a/spec/system/provider_interface/removing_provider_user_spec.rb
+++ b/spec/system/provider_interface/removing_provider_user_spec.rb
@@ -3,7 +3,7 @@ require 'rails_helper'
 RSpec.describe 'Removing a provider user' do
   include DfESignInHelpers
 
-  scenario 'removing a user from all providers' do
+  scenario 'removing a user from all providers', with_audited: true do
     FeatureFlag.activate(:providers_can_manage_users_and_permissions)
 
     given_i_am_a_provider_user_with_dfe_sign_in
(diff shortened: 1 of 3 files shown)


spec/system/provider_interface/removing_provider_user_spec.rb
```diff
@@ -68,6 +69,10 @@ def then_the_deleted_user_has_no_visible_provider_permissions
     expect(@user_to_remove.reload.providers).to eq([@non_visible_provider])
   end
 
+  def and_the_audit_log_has_been_updated
+    expect(@user_to_remove.own_and_associated_audits.first(2).map(&:action)).to eql(%w[destroy destroy])
```
duncanjbrown, line 73: Could be more explicit about the type that's being destroyed?

```ruby
@user_to_remove.own_and_associated_audits.limit(2).pluck(:action, :auditable_type).to eql [["destroy", "ProviderPermissions"], ["destroy", "ProviderPermissions"]]
```

tijmenb: Done!

PR #2603, opened 2020-07-29
When inviting users, list only organisations the provider user can manage

mnacos:
## Context

If a user is allowed to `manage_users` for Blackfriars, but not Keele, the user invitation wizard nevertheless lists both when adding the new user to providers.

image

Fix this by restricting the list of providers to the list of organisations the current provider user can manage.

## Guidance to review

Is there a better way to write such queries?

## Link to Trello card

https://trello.com/c/6BAC6Bxt

## Things to check


duncanjbrown:
ah, of course – that's better!

PR #2606, opened 2020-07-29
Support the single provider journey through the invite provider user wizard

steventux:
## Context

The latest design has different conditional rendering logic for the scenario where an inviting user belongs to one provider organisation.

Adds conditional rendering logic for scenario where inviting user belongs to one provider organisation.

image

## Link to Trello card

https://trello.com/c/gH5jrsuu/2532-support-the-single-provider-journey-through-the-invite-provider-user-wizard

## Things to check

### DIFF
--- a/app/controllers/provider_interface/provider_users_invitations_controller.rb
+++ b/app/controllers/provider_interface/provider_users_invitations_controller.rb
@@ -4,7 +4,7 @@ module ProviderInterface
     before_action :require_manage_user_permission!
 
     def edit_details
-      @wizard = wizard_for(current_step: 'details', single_provider: current_provider_user.providers.size == 1)
+      @wizard = wizard_for(wizard_setup_options)
       @wizard.save_state!
     end
 
@@ -23,7 +23,7 @@ module ProviderInterface
       @wizard = wizard_for(current_step: 'providers')
       @wizard.save_state!
 
-      @available_providers = current_provider_user.authorisation.providers_that_actor_can_manage_users_for
+      @available_providers = available_providers
     end
 
     def update_providers
@@ -98,6 +98,19 @@ module ProviderInterface
 
   private
 
+    def available_providers
+      current_provider_user.authorisation.providers_that_actor_can_manage_users_for
+    end
+
+    def wizard_setup_options
+      setup_options = { current_step: 'details' }
+      if available_providers.count == 1
+        setup_options[:providers] = available_providers.map(&:id)
+        setup_options[:single_provider] = true
+      end
+      setup_options
+    end
+
     def wizard_for(options)
       options[:checking_answers] = true if params[:checking_answers] == 'true'
       ProviderUserInvitationWizard.new(session, options)
@@ -115,7 +128,7 @@ module ProviderInterface
 
     def details_params
       params.require(:provider_interface_provider_user_invitation_wizard)
-        .permit(:first_name, :last_name, :email_address, :single_provider, providers: [])
+        .permit(:first_name, :last_name, :email_address)
     end
 
     def providers_params
--- a/app/controllers/provider_interface/provider_users_invitations_controller.rb
+++ b/app/controllers/provider_interface/provider_users_invitations_controller.rb
@@ -4,7 +4,7 @@ module ProviderInterface
     before_action :require_manage_user_permission!
 
     def edit_details
-      @wizard = wizard_for(current_step: 'details')
+      @wizard = wizard_for(current_step: 'details', single_provider: current_provider_user.providers.size == 1)
       @wizard.save_state!
     end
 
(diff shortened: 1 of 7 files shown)


app/views/provider_interface/provider_users_invitations/edit_details.html.erb
```diff
@@ -13,6 +13,11 @@
       <%= f.govuk_text_field :first_name, label: { text: 'First name', size: 'm' } %>
       <%= f.govuk_text_field :last_name, label: { text: 'Last name', size: 'm' } %>
 
+      <% if current_provider_user.providers.size == 1 %>
+        <%= f.hidden_field :single_provider, value: true %>
```
duncanjbrown, line 17: Maybe we could set this on the wizard as soon as the user starts, rather than posting it back?

steventux: @duncanjbrown rebased into https://github.com/DFE-Digital/apply-for-teacher-training/pull/2606/commits/03120fcb607db8c66e193d7f4bf9a6ef087226c5

app/controllers/provider_interface/provider_users_invitations_controller.rb
```diff
@@ -4,7 +4,7 @@ class ProviderUsersInvitationsController < ProviderInterfaceController
     before_action :require_manage_user_permission!
 
     def edit_details
-      @wizard = wizard_for(current_step: 'details')
+      @wizard = wizard_for(current_step: 'details', single_provider: current_provider_user.providers.size == 1)
```
duncanjbrown, line 7: This needs to be `current_provider_user.authorisation.providers_that_actor_can_manage_users_for` or we get the checkbox when the user has 2 providers but can only manage for 1

app/views/provider_interface/provider_users_invitations/edit_details.html.erb
steventux: @duncanjbrown We skip the orgs page here https://github.com/DFE-Digital/apply-for-teacher-training/pull/2606/commits/239d6baba7115a59f53ac217463f1c4a90882646#diff-2397cfec288e45bd0d29bd8a1c428527R99 is this not working as expected for you? Seems to work on PR app...

duncanjbrown:
I've just realised we could simplify more by making `single_provider` dynamic - then we could lose the postback to `update_details`

```ruby
    def edit_details
      # set a provider if there is only one to set
      wizard_setup_options = if available_providers.count == 1
                               { providers: available_providers.map(&:id) }
                             else
                               {}
                             end
      @wizard = wizard_for(wizard_setup_options.merge(current_step: 'details'))
      @wizard.save_state!
    end
```

app/controllers/provider_interface/provider_users_invitations_controller.rb
steventux: @duncanjbrown thanks! 
Done here https://github.com/DFE-Digital/apply-for-teacher-training/pull/2606/commits/10aa54773817413adb9aecdba831e4e2c45b912d

app/controllers/provider_interface/provider_users_invitations_controller.rb
```diff
@@ -98,6 +98,19 @@ def previous_page
...
+      current_provider_user.authorisation.providers_that_actor_can_manage_users_for
+    end
+
+    def wizard_setup_options
+      setup_options = { current_step: 'details' }
+      if available_providers.count == 1
+        setup_options[:providers] = available_providers.map(&:id)
+        setup_options[:single_provider] = true
+      end
+      setup_options
```
duncanjbrown, line 111: Nice!

app/controllers/provider_interface/provider_users_invitations_controller.rb
```diff
@@ -115,7 +128,7 @@ def path_for(step, provider_id)
 
     def details_params
       params.require(:provider_interface_provider_user_invitation_wizard)
-        .permit(:first_name, :last_name, :email_address)
+        .permit(:first_name, :last_name, :email_address, providers: [])
```
duncanjbrown, line 131: Can this param go now?

app/controllers/provider_interface/provider_users_invitations_controller.rb
```diff
@@ -98,6 +98,19 @@ def previous_page
...
 
+    def available_providers
+      current_provider_user.authorisation.providers_that_actor_can_manage_users_for
+    end
+
+    def wizard_setup_options
+      setup_options = { current_step: 'details' }
+      if available_providers.count == 1
+        setup_options[:providers] = available_providers.map(&:id)
+        setup_options[:single_provider] = true
```
duncanjbrown, line 109: Of course, it makes sense to pass this in so the wizard doesn't know about current_provider_user

PR #2617, opened 2020-07-30
Hide header navigation links from provider interface during onboarding

mnacos:
## Context

There are links in the header during onboarding. These won't work because users will be redirected back to the onboarding screens.

image

Hide the specified links from the header if the provider user is `performing_setup`. I've added a `@provider_setup.pending?` method which is used to set an optional boolean `performing_setup` as an extra argument to `NavigationItem.for_provider_*_nav` methods. We currently only show 'Applications' in the sub-header, but it should be easy to add the other links when they are ready.

## Guidance to review

Is this approach reasonable? `NavigationItems` code does not have access to `@provider_setup`, set in the controller, hence the need for the new argument. 

I've also moved `NavigationItems` to `app/helpers`, as it's not a model.

## Link to Trello card

https://trello.com/c/ejHRNikf

## Things to check

### DIFF
--- a/app/helpers/navigation_items.rb
+++ b/app/helpers/navigation_items.rb
@@ -36,25 +36,32 @@ class NavigationItems
       end
     end
 
-    def for_provider_primary_nav(current_controller)
-      [NavigationItem.new('Applications', provider_interface_applications_path, is_active(current_controller, %w[application_choices decisions offer_changes]))]
+    def for_provider_primary_nav(current_controller, performing_setup = false)
+      if performing_setup
+        []
+      else
+        [NavigationItem.new('Applications', provider_interface_applications_path, is_active(current_controller, %w[application_choices decisions offer_changes]))]
+      end
     end
 
-    def for_provider_account_nav(current_provider_user, current_controller)
+    def for_provider_account_nav(current_provider_user, current_controller, performing_setup = false)
       return [NavigationItem.new('Sign in', provider_interface_sign_in_path, false)] unless current_provider_user
 
       items = []
 
-      if FeatureFlag.active?('enforce_provider_to_provider_permissions') &&
-          current_provider_user.authorisation.can_manage_organisations_for_at_least_one_provider?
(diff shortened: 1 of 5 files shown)


app/helpers/navigation_items.rb
```diff
@@ -36,25 +36,29 @@ def for_support_interface(current_support_user, current_controller)
...
+        items << NavigationItem.new('Organisations', provider_interface_organisations_path, is_active(current_controller, %w[organisations provider_relationship_permissions])) unless performing_setup
       end
 
       if FeatureFlag.active?(:providers_can_manage_users_and_permissions) && current_provider_user.authorisation.can_manage_users_for_at_least_one_provider?
-        items << NavigationItem.new('Users', provider_interface_provider_users_path, is_active(current_controller, 'provider_users'))
+        items << NavigationItem.new('Users', provider_interface_provider_users_path, is_active(current_controller, 'provider_users')) unless performing_setup
       end
 
-      items << NavigationItem.new('Account', provider_interface_account_path, is_active(current_controller, 'account'))
+      items << NavigationItem.new('Account', provider_interface_account_path, is_active(current_controller, 'account')) unless performing_setup
```
duncanjbrown, line 61: Maybe this could all be wrapped in an `unless performing_setup`

steventux: @mnacos, @duncanjbrown seems sensible to do this, we don't use these anywhere in setup, they are present in the prototype app though https://manage-applications-beta.herokuapp.com/onboard/step3

duncanjbrown: @adamsilver agreed on Slack we should do this (https://ukgovernmentdfe.slack.com/archives/CPH8J9G65/p1594635006466700) — I've put a card into the design backlog to clean up the prototype (https://trello.com/c/4gbM3p5t/630-remove-header-links-during-onboarding)

mnacos:
Thanks! I've added a system spec assertion that these links are hidden during DSA signing. When we have the new org setup pages in, we can also check header links are hidden for that flow, too @steventux

PR #2621, opened 2020-07-30
do not mount /public assets local folder

vigneshmsft:
## Context

when running docker-compose for local development, 
`/public` folder might not contain the compiled packs folder

also not run the clock and worker services for local development

## Things to check

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -4,6 +4,13 @@ services:
   web:
     volumes:
       - ${APP_ROOT_FOLDER:-.}:/app
+      # do not mount these below local files/folders into the container
+      - /app/public
     environment:
       - RAILS_ENV=development
       - AUTHORISED_HOSTS=localhost
+# disable clock and worker when running locally
+  clock:
+    entrypoint: ["/bin/true"]
+  worker:
+    entrypoint: ["/bin/true"]
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ shell: ## Open a shell on the app container
 
 .PHONY: serve
 serve: ## Run the service
-	docker-compose up --build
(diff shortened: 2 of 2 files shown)


docker-compose.override.yml
```diff
@@ -4,6 +4,13 @@ services:
   web:
     volumes:
       - ${APP_ROOT_FOLDER:-.}:/app
+      # do not mount these below local files/folders into the container
+      - /app/public
```
tijmenb, line 8: Can you add some more info on why this is necessary? I don't all the context.

docker-compose.override.yml
```diff
@@ -4,6 +4,13 @@ services:
...
     environment:
       - RAILS_ENV=development
       - AUTHORISED_HOSTS=localhost
+# disable clock and worker when running locally
+  clock:
+    entrypoint: ["/bin/true"]
+  worker:
+    entrypoint: ["/bin/true"]
```
tijmenb, line 16: Would be useful to be able to run the whole environment using Docker? Can it be run manually?

docker-compose.override.yml
vigneshmsft: earlier I just mounted the root folder in the local disk as the work directory of the container, so that you don't have to rebuild the container when doing active development.
doing so looks for assets inside the `public/packs` folder, but this might not be present in the local directory without running `assets:precompile`, so am excluding this folder from being mapped and use assets from inside the container.

vigneshmsft: coming to think of it, I could also set the a different target so that the assets are generated/compiled dynamically?
But I am not aware of anyone other than Laura and myself who uses docker to run in locally and decided it would be simpler to do this change

PR #2623, opened 2020-07-31
Delete test candidates and applications via support interface on QA

stevehook:
## Context

There is a lot of test data on qa much of which is invalid or just unnecessary. We need a way to purge the data in test environments (`qa`, `development`, `test` and `review`). This PR adds a background worker that gets invoked via a button in the support interface.

- [x] Add button to the support interface tasks page and an 'are you sure?' step. Hidden in a non-test environment.
- [x] New worker `DeleteTestApplications` called via button. Could be used independently. Raises error if called in a non-test environment.
- [x] System spec to exercise the above.

image

image

## Guidance to review

- Is it safe? Are the test-only checks watertight? Is it ok to use in Review apps? (this wasn't specified in the card).
- Is the worker re-usable?
- Are the tests sufficient?
- Copy could probably be improved.

## Link to Trello card

https://trello.com/c/4RZNyokS/1807-add-button-to-purge-qa-candidates-and-their-applications

## Things to check


config/routes.rb
```diff
@@ -709,6 +709,7 @@
     get '/tasks' => 'tasks#index', as: :tasks
     post '/tasks/create-fake-provider' => 'tasks#create_fake_provider'
     post '/tasks/:task' => 'tasks#run', as: :run_task
+    get '/tasks/:task/confirm' => 'tasks#confirm', as: :confirm_task
```
duncanjbrown, line 712: Would it be simpler to have  an explicit path for this? Then no need for a controller action at all, I think, and 404 would work out of the box?

app/workers/purge_test_applications.rb
```diff
@@ -0,0 +1,29 @@
...
+  def perform(*)
+    raise 'You can only purge test applications in a test environment' unless DeleteTestApplications.can_run_in_this_environment?
+
+    candidates_to_purge.find_each do |candidate|
+      candidate.application_forms.each do |application_form|
+        application_form.application_choices.each(&:destroy)
+        application_form.destroy
+      end
+      candidate.destroy
+    end
```
tijmenb, line 13: Because we've set up [cascading deletes correctly](https://github.com/DFE-Digital/apply-for-teacher-training/blob/65491ad206b25723058eb38a31da737bc1776e51/spec/models/candidate_spec.rb#L15-L35), we should be able to do this:

```suggestion
    candidates_to_purge.delete_all
```

app/workers/delete_test_applications.rb
```diff
@@ -0,0 +1,29 @@
+class DeleteTestApplications
```
tijmenb, line 1: This file is duplicated I think.

stevehook: oops, sorry, botched rename - fixed

app/views/support_interface/tasks/confirm_delete_test_applications.html.erb
```diff
@@ -0,0 +1,26 @@
...
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with model: nil,
+      url: support_interface_run_task_path('delete_test_applications') do |f|
+    %>
+      <h1 class="govuk-heading-xl">
+        Delete test applications
+      </h1>
+
+      <p class="govuk-body">
+        This will delete all candidates with emails that end in "@example.com", their applications and associated data. Are you sure?
```
duncanjbrown, line 13:
```suggestion
        This will delete all candidates with emails that end in “@example.com”, their applications and associated data. Are you sure?
```

app/views/support_interface/tasks/index.html.erb
```diff
@@ -16,6 +16,16 @@
   <%= button_to 'Recalculate dates', support_interface_run_task_path('recalculate_dates'), class: 'govuk-button' %>
 </div>
 
+<% if DeleteTestApplications.can_run_in_this_environment? %>
+  <div>
+    <h2 class='govuk-heading-m'>Delete test applications</h2>
+    <div class='govuk-body'>
+      This will delete all candidates with emails that end in "@example.com", their applications and associated data.
```
duncanjbrown, line 23:
```suggestion
      This will delete all candidates with emails that end in  “@example.com”, their applications and associated data.
```

PR #2624, opened 2020-07-31
Bug fix alteration of flow and addition of view so that candidate can give one referee at a time

george-bit:
## Context

Post submission, Candidates are currently unable to submit referees independently of one another. This causes a potential issue if the candidate only has one referee to hand and wishes to request the other at a later date as currently they can only be confirmed and submitted in tandem.

Split the confirmation flow to appear separately for each referee and introduce a new view to appear in between addition of referees to allow candidate to return to the application dashboard and add another referee later.

single referee confirmation screen:
image

New Add another referee view:
image

## Guidance to review

Use the review app to check that the UI flow makes sense. 

Design doc here: https://docs.google.com/drawings/d/1mQxu3S7mR8IYxuREzfwOguWRc98e6gzqZZ1D-GF4UWE/edit

Review the code against best practices.

## Link to Trello card

https://trello.com/c/SWsfc89S/1811-bug-i-only-want-to-give-one-referee-at-a-time

## Things to check

### DIFF
--- a/app/services/feature_flag.rb
+++ b/app/services/feature_flag.rb
@@ -38,8 +38,12 @@ class FeatureFlag
     [:efl_section, 'Allow candidates with nationalities other then British or Irish to specify their English as a Foreign Language experience', 'Malcolm Baig'],
     [:international_degrees, 'Changes to the model and forms for degree qualifications to cater for non-UK degrees.', 'Steve Hook'],
     [:international_gcses, 'Candidates can provide details of international GCSE equivalents.', 'George Holborn'],
+<<<<<<< HEAD
     [:international_other_qualifications, 'Candidates can provide details of Other international qualifications .', 'David Gisbey'],
     [:separate_additional_referees, 'Candidates can submit and confirm their referees independently of each other in the additional referees flow']
+=======
+    [:separate_additional_referees, 'Candidates can submit and confirm their referees independently of each other in the additional referees flow', 'George Holborn'],
+>>>>>>> Update to feature test to follow flow & tweak to redirect on controller
   ].freeze
 
   FEATURES = (PERMANENT_SETTINGS + TEMPORARY_FEATURE_FLAGS).map { |name, description, owner|
--- a/app/services/feature_flag.rb
+++ b/app/services/feature_flag.rb
@@ -38,16 +38,8 @@ class FeatureFlag
     [:efl_section, 'Allow candidates with nationalities other then British or Irish to specify their English as a Foreign Language experience', 'Malcolm Baig'],
     [:international_degrees, 'Changes to the model and forms for degree qualifications to cater for non-UK degrees.', 'Steve Hook'],
     [:international_gcses, 'Candidates can provide details of international GCSE equivalents.', 'George Holborn'],
-<<<<<<< HEAD
-<<<<<<< HEAD
     [:international_other_qualifications, 'Candidates can provide details of Other international qualifications .', 'David Gisbey'],
-    [:separate_additional_referees, 'Candidates can submit and confirm their referees independently of each other in the additional referees flow']
-=======
-    [:separate_additional_referees, 'Candidates can submit and confirm their referees independently of each other in the additional referees flow', 'George Holborn'],
->>>>>>> Update to feature test to follow flow & tweak to redirect on controller
-=======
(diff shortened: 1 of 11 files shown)


tvararu:
Have tested it in the review app and it seems to work 👌

app/services/feature_flag.rb
```diff
@@ -38,6 +38,7 @@ def feature
     [:efl_section, 'Allow candidates with nationalities other then British or Irish to specify their English as a Foreign Language experience', 'Malcolm Baig'],
     [:international_degrees, 'Changes to the model and forms for degree qualifications to cater for non-UK degrees.', 'Steve Hook'],
     [:international_gcses, 'Candidates can provide details of international GCSE equivalents.', 'George Holborn'],
+    [:separate_additional_referees, 'Candidates can submit and confirm their referees independently of each other in the additional referees flow', 'George Holborn'],
```
tvararu, line 41: For consistency:

```suggestion
    [:separate_additional_referees, 'Candidates can submit and confirm their referees independently of each other in the additional referees flow.', 'George Holborn'],
```

tvararu:
I did spot one bug! You'll need to add a translation for the form model validation error:

Screenshot 2020-07-31 at 15 47 06

That will go in the [en.yml file, in the errors section](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/config/locales/en.yml#L167). Take a look at the other ones to figure out how to add one for your form model. :)

spec/forms/candidate_interface/add_another_referee_form_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'rails_helper'
+
+RSpec.describe CandidateInterface::AddAnotherRefereeForm, type: :model do
+  let(:form) { described_class.new(add_another_referee) }
+  let(:add_another_referee) do
+    { 'add_another_referee' => 'yes' }
+  end
```
tvararu, line 7:
```suggestion
  let(:add_another_referee) { 'add_another_referee' => 'yes' }
```

spec/forms/candidate_interface/add_another_referee_form_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    context 'when the add_another_referee value is yes' do
+      it 'returns true' do
+        expect(form.add_another_referee?).to be_truthy
+      end
+    end
+
+    context 'when the add_another_referee value is no' do
+      let(:add_another_referee) do
+        { 'add_another_referee' => 'no' }
+      end
```
tvararu, line 21:
```suggestion
      let(:add_another_referee) { 'add_another_referee' => 'no' }
```

spec/system/candidate_interface/candidate_needs_to_provide_two_new_referees_with_separate_additional_referees_flag_on_spec.rb
```diff
@@ -0,0 +1,251 @@
...
+  scenario "Candidate provides a new referee because 2 didn't respond" do
+    given_i_am_signed_in_as_a_candidate
+    and_i_have_submitted_my_application
+    and_both_of_my_referees_havent_responded_within_a_reasonable_timeframe
+    and_the_separate_additional_referees_flag_is_on
+
+    when_i_visit_the_application_dashboard
+    then_i_see_that_i_need_new_references
+
+    when_i_visit_the_start_page
```
tvararu, line 15: Could you click on the banner here instead of visiting the page directly?

spec/forms/candidate_interface/add_another_referee_form_spec.rb
george-bit: Thanks Theo

spec/system/candidate_interface/candidate_needs_to_provide_two_new_referees_with_separate_additional_referees_flag_on_spec.rb
george-bit: So this was my idea originally but David explained why we should do it this way. Effectively we are trying to mimic the behaviour as if the candidate proceeded by choosing yes from the new view.

They definitely could add another referee by clicking the banner... however, they would then see the normal add new referee page. To mimic the flow more accurately we want them to proceed from the add_another_referee page which should redirect them instead to the slightly different add a second referee page :)

PR #2627, opened 2020-07-31
Fix for country autocomplete saving wrong value

stevehook:
## Context

On the contact details pages for international address we were saving the country name e.g. `France` rather than the ISO code `FR`.

Other components assume we store the ISO code and this is consistent with other places that we store nationality or country values.

- [x] add a dedicated `country-autocomplete.js` for declaring autocompletes for country values rather than dumping them in with the nationality ones.
- [x] just use the default settings for `accessible-autocomplete` specifically leaving the name of the original `select` element intact so that it's value gets submitted when the form is posted. This ensures that the value of the selected item (which is an ISO code) gets submitted rather than the value of the autocomplete input (which is then human-readable country name).

## Guidance to review

There is some inconsistency in the way we use `accessible-autocomplete` that confused me. The recommended way seems to be to use a `select` element for progressive enhancement and have the autocomplete sync the selected value back the the `select` element so that it gets submitted as normal. However we toggle the element names in a few places so that the autocomplete `input` value is submitted instead (e.g. https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/frontend/packs/nationality-autocomplete.js#L32-L36). Unfortunately this only works if the display value is the same as the value so it doesn't work with ISO country code/name selects. There are a few places where we workaround this issue by converting the display values submitted to actual values in server-side code which seems a bit messy.

## Link to Trello card

https://trello.com/c/i54Wkj7X/1854-international-address-country-autocomplete-not-saving-correct-value

## Things to check


davidgisbey:
I think this is the right approach. There's a few more fields that should be using this in the other qualifications.

PR #2630, opened 2020-08-03
Add JavaScript to the personal details nationalities page

davidgisbey:
## Context

Candidates need to be able to input multiple nationalities. Paul built the design using some JS.

This should do a few things. 

It should have an add nationalities button. When clicked this should display an additional select for candidates to pick an additional nationality. This button should not display once all 3 selects are visible.

It should have a remove link on the second and third select. When clicked it should hide the select and clear its value.

If multiple nationalities are present, when a user goes to edit their nationality they should pre-populate.

If an error occurs, use the non-js solution.

- Build the JS functionality. for the nationalities page

image

## Guidance to review

Screenshots of the designs are on the Trello ticket.

Probably easiest to just manually test it and see if you can break it.

## Link to Trello card

https://trello.com/c/AsOuRv9u/1760-dev-%F0%9F%8C%90-international-residency-visa-status

## Things to check

### DIFF
--- a/app/frontend/packs/nationalities-component.js
+++ b/app/frontend/packs/nationalities-component.js
@@ -1,6 +1,5 @@
 const nationalitiesComponent = () => {
-
-  const pageHasErrors = document.querySelector('.govuk-error-summary');
+  const pageHasErrors = document.querySelector(".govuk-error-summary");
   if (pageHasErrors) return;
 
   const secondSelect = document.getElementById(
@@ -18,27 +17,19 @@ const nationalitiesComponent = () => {
     "[for=candidate-interface-nationalities-form-other-nationality3-field]"
   );
 
-  let addNationalityButton = null
+  let addNationalityButton = null;
 
-  addRemoveLink(
-    secondSelect.id
-  );
+  addRemoveLink(secondSelect.id);
 
-  addRemoveLink(
-    thirdSelect.id
-  );
+  addRemoveLink(thirdSelect.id);
 
   addAddNationalityButton(
     "#candidate-interface-nationalities-form-other-other-conditional"
   );
 
-  hideSection(
-    secondSelect.id
-  );
+  hideSection(secondSelect.id);
 
-  hideSection(
-    thirdSelect.id
-  );
+  hideSection(thirdSelect.id);
 
   function addRemoveLink(selector) {
     const labelEl = document.querySelector(`[for=${selector}]`);
@@ -51,7 +42,7 @@ const nationalitiesComponent = () => {
     removeLink.href = "#";
     labelEl.appendChild(removeLink);
 
-    addNthNationalityHiddenSpan(removeLink, selector)
+    addNthNationalityHiddenSpan(removeLink, selector);
 
     removeLink.addEventListener("click", function () {
       handleRemoveLinkClick(labelEl, selector);
@@ -60,13 +51,12 @@ const nationalitiesComponent = () => {
 
   function addNthNationalityHiddenSpan(removeLink, selector) {
     const nthNationalitySpan = document.createElement("span");
-    nthNationalitySpan.classList.add("govuk-visually-hidden")
+    nthNationalitySpan.classList.add("govuk-visually-hidden");
 
     if (selector == secondSelect.id) {
-      nthNationalitySpan.innerHTML = 'Second nationality';
-    }
-    else {
-      nthNationalitySpan.innerHTML = 'Third nationality';
+      nthNationalitySpan.innerHTML = "Second nationality";
+    } else {
+      nthNationalitySpan.innerHTML = "Third nationality";
     }
 
     removeLink.appendChild(nthNationalitySpan);
@@ -77,7 +67,10 @@ const nationalitiesComponent = () => {
     addNationalityButton = document.createElement("button");
     addNationalityButton.innerHTML = "Add another nationality";
     addNationalityButton.id = "add-nationality-button";
-    addNationalityButton.classList.add("govuk-button", "govuk-button--secondary");
+    addNationalityButton.classList.add(
+      "govuk-button",
+      "govuk-button--secondary"
+    );
     parent.appendChild(addNationalityButton);
 
     if (secondSelect.value && thirdSelect.value) {
@@ -111,8 +104,7 @@ const nationalitiesComponent = () => {
       thirdFormLabel.parentElement.style.display === "none"
     ) {
       secondFormLabel.parentElement.style.display = "";
-    }
-    else if (secondFormLabel.parentElement.style.display === "none") {
+    } else if (secondFormLabel.parentElement.style.display === "none") {
       secondFormLabel.parentElement.style.display = "";
       addNationalityButton.style.display = "none";
     } else {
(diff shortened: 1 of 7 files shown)


davidgisbey:
Thanks for the feedback Steve. I'll sort the bug with the button and link with the flag off now.

_Why are English, Welsh and Northern Irish listed as nationalities in the Other autocomplete?_

Some people self-identify as Scottish, Welsh etc. I'm pretty sure Paul said that those should be valid options.

I think @tvararu had some thoughts on the other points.

app/frontend/packs/nationalities-component.js
```diff
@@ -0,0 +1,118 @@
...
+    removeLink.classList.add("govuk-link", "remove-link");
+
+    // This has to be a link and not a button as the govuk-link class requires an
+    // href to apply  it's styling
+    removeLink.href = "#";
+    labelEl.appendChild(removeLink);
+
+    removeLink.addEventListener("click", function () {
+      addRemoveLinkEvent(labelEl, selector);
+    });
```
tvararu, line 46:
```suggestion
    removeLink.addEventListener("click", (event) => {
      event.preventDefault();
      addRemoveLinkEvent(labelEl, selector);
    });
```

app/frontend/packs/nationalities-component.js
```diff
@@ -0,0 +1,118 @@
...
+  );
+
+  hideSection(
+    "candidate-interface-nationalities-form-other-nationality3-field"
+  );
+
+  function addRemoveLink(selector) {
+    const labelEl = document.querySelector(`[for=${selector}]`);
+    const removeLink = document.createElement("a");
+    removeLink.innerHTML = "Remove";
```
tvararu, line 36: This could a bit better, it could be `Remove <span class="govuk-visually-hidden">{second/third} nationality` depending on what's passed into the `addRemoveLink` function. Would help screen reader users. Up for giving that a spin? ♿

app/frontend/packs/nationalities-component.js
```diff
@@ -0,0 +1,118 @@
...
+    removeLink.classList.add("govuk-link", "remove-link");
```
tvararu, line 37: Custom CSS should be prefixed with `app-`:

```suggestion
    removeLink.classList.add("govuk-link", "app-remove-link");
```

Going to add a suggestion to the CSS file as well.

app/frontend/packs/nationalities-component.js
tvararu: Also, I prefer calling event handlers e.g. `handleRemoveLinkClick`, but that's more of a style suggestion.

app/frontend/packs/nationalities-component.js
```diff
@@ -0,0 +1,118 @@
...
+      "[id=candidate-interface-nationalities-form-other-nationality3-field]"
+    );
+
+    if (secondSelect.value && thirdSelect.value) {
+      nationalityButton.style.display = "none";
+    }
+
+    nationalityButton.addEventListener("click", function () {
+      event.preventDefault();
+      addNationalityEvent(nationalityButton);
```
tvararu, line 70: Similar to my earlier naming suggestion:

```suggestion
      handleAddNationalityClick(nationalityButton);
```

app/frontend/packs/nationalities-component.js
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  function addNationalityEvent(nationalityButton) {
+
+    const secondFormLabel = document.querySelector(
+      "[for=candidate-interface-nationalities-form-other-nationality2-field]"
+    );
+    const thirdFormLabel = document.querySelector(
+      "[for=candidate-interface-nationalities-form-other-nationality3-field]"
+    );
```
tvararu, line 100: You could hoist all the `document.querySelector` calls to the start of this file, assign to variables, and then reuse those wherever you need later on:

```js
const secondFormLabel = document.querySelector(...)
const thirdFormLabel = document.querySelector(...)
let addNationalityButton = null

function addAddNationalityButton(...) {
  addNationalityButton = document.createElement("button")
  // ...
}

function addNationalityEvent(...) {
  if (secondFormLabel.parentElement.style.display === "none") {
    secondFormLabel.parentElement.style.display = "";
  } else {
    thirdFormLabel.parentElement.style.display = "";
  }
  // ...
}
```

This could shorten your functions and might make things a bit less repetitive. If that makes sense, do you want to give it a go?

app/frontend/packs/nationalities-component.js
```diff
@@ -0,0 +1,118 @@
...
+
+  function addRemoveLink(selector) {
+    const labelEl = document.querySelector(`[for=${selector}]`);
+    const removeLink = document.createElement("a");
+    removeLink.innerHTML = "Remove";
+    removeLink.classList.add("govuk-link", "remove-link");
+
+    // This has to be a link and not a button as the govuk-link class requires an
+    // href to apply  it's styling
+    removeLink.href = "#";
```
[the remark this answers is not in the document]
davidgisbey: Yea i can't either :sad_panda:

app/frontend/packs/nationalities-component.js
davidgisbey: Have done 👍

app/frontend/packs/nationalities-component.js
```diff
@@ -0,0 +1,117 @@
...
+  addAddNationalityButton(
+    "#candidate-interface-nationalities-form-other-other-conditional"
+  );
+
+  hideSection(secondSelect.id);
+
+  hideSection(thirdSelect.id);
+
+  function addRemoveLink(selector) {
+    const labelEl = document.querySelector(`[for=${selector}]`);
```
tvararu, line 35: You don't need to run a `querySelector` here anymore now that `secondFormLabel` is available You could just pass it in as an argument where it's called, like `addRemoveLink(secondFormLabel)`, and then:

```suggestion
  function addRemoveLink(labelEl) {
```

app/frontend/packs/nationalities-component.js
```diff
@@ -0,0 +1,117 @@
...
+    const removeLink = document.createElement("a");
+    removeLink.innerHTML = "Remove";
+    removeLink.classList.add("govuk-link", "app-nationality-remove-link");
+
+    // This has to be a link and not a button as the govuk-link class requires an
+    // href to apply  it's styling
+    removeLink.href = "#";
+    labelEl.appendChild(removeLink);
+
+    addNthNationalityHiddenSpan(removeLink, selector);
```
tvararu, line 45: You shouldn't need to pass in the `selector` here either as it's only used to decide which content to use. You could pass a number instead (2 for the second label):

```suggestion
    addNthNationalityHiddenSpan(removeLink, 2);
```

app/frontend/packs/nationalities-component.js
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    addNationalityButton.addEventListener("click", function () {
+      event.preventDefault();
+      handleAddNationalityClick();
+    });
+  }
+
+  function hideSection(selector) {
+    const select = document.getElementById(selector);
```
tvararu, line 87: See above suggestion about reusing the label variables here too!

PR #2638, opened 2020-08-03
Update the display of safeguarding information

AgaDufrat:
## Context

Since we released Access Controls [the design of displaying safeguarding information changed](https://bat-design-history.netlify.app/manage-teacher-training-applications/view-safeguarding-information/). Users with view safeguarding information permission should be shown the details.

- always display 'Criminal convictions and professional misconduct' header
- update message copy
- show details to users with permissions (html stolen from candidate's interface)

| Without permissions | With permissions |
| ------------- | ------------- |
|Screenshot 2020-08-03 at 12 51 07| Screenshot 2020-08-03 at 12 47 32 |

## Guidance to review

- to declare safeguarding issues `ApplicationForm.find_by(support_reference: "XXXX").update!(safeguarding_issues_status: "has_safeguarding_issues_to_declare", safeguarding_issues: "I have a criminal conviction.")`
- test with different user and provider-provider permissions
- other states like "Not answered yet" also work

## Link to Trello card

https://trello.com/c/hfwlPpk7/2500-implement-new-design-for-safeguarding-information

## Things to check

### DIFF
--- a/app/components/provider_interface/safeguarding_declaration_component.rb
+++ b/app/components/provider_interface/safeguarding_declaration_component.rb
@@ -21,6 +21,10 @@ module ProviderInterface
         .can_view_safeguarding_information?(course: application_choice.course)
     end
 
+    def details
+      application_choice.application_form.safeguarding_issues
+    end
+
   private
 
     def status
--- a/app/components/provider_interface/safeguarding_declaration_component.rb
+++ b/app/components/provider_interface/safeguarding_declaration_component.rb
@@ -2,13 +2,37 @@ module ProviderInterface
   class SafeguardingDeclarationComponent < ViewComponent::Base
     include ViewHelper
 
-    attr_reader :message
+    attr_reader :application_choice, :current_provider_user
 
-    def initialize(application_form:)
-      @message = SafeguardingStatus.new(
-        status: application_form.safeguarding_issues_status,
+    def initialize(application_choice:, current_provider_user:)
+      @application_choice = application_choice
+      @current_provider_user = current_provider_user
+    end
+
+    def message
+      SafeguardingStatus.new(
+        status: status,
         i18n_key: 'provider_interface.safeguarding_declaration_component',
       ).message
+    end
+
+    def current_user_has_permission_to_view_safeguarding_information?
+      current_provider_user.authorisation
+        .can_view_safeguarding_information?(course: application_choice.course)
+    end
+
+  private
+
+    def status
+      if safeguarding_issues_declared? && !current_user_has_permission_to_view_safeguarding_information?
+        'has_safeguarding_issues_to_declare_no_permissions'
+      else
+        application_choice.application_form.safeguarding_issues_status
+      end
+    end
+
+    def safeguarding_issues_declared?
+      application_choice.application_form.has_safeguarding_issues_to_declare?
     end
   end
 end
--- a/app/components/provider_interface/safeguarding_declaration_component.html.erb
+++ b/app/components/provider_interface/safeguarding_declaration_component.html.erb
@@ -1,1 +1,13 @@
 <p class="govuk-body"><%= message %></p>
+<% if current_user_has_permission_to_view_safeguarding_information? %>
+  <details class="govuk-details govuk-!-margin-bottom-2" data-module="govuk-details">
+    <summary class="govuk-details__summary">
+      <span class="govuk-details__summary-text">
+        View information disclosed by the candidate
+      </span>
+    </summary>
+    <div class="govuk-details__text">
+      <%= details %>
+    </div>
+  </details>
+<% end %>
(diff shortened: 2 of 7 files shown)


app/components/provider_interface/safeguarding_declaration_component.rb
```diff
@@ -2,13 +2,43 @@ module ProviderInterface
...
+      set_details if safeguarding_issues_declared? && display_safeguarding_information?
+
+      SafeguardingStatus.new(
+        status: status,
         i18n_key: 'provider_interface.safeguarding_declaration_component',
       ).message
     end
+
+    def display_safeguarding_information?
+      ProviderAuthorisation.new(actor: current_provider_user)
```
tijmenb, line 22:
```suggestion
      current_provider_user.authorisation
```

app/components/provider_interface/safeguarding_declaration_component.rb
```diff
@@ -2,13 +2,43 @@ module ProviderInterface
...
+    def status
+      if safeguarding_issues_declared? && !display_safeguarding_information?
+        'has_safeguarding_issues_to_declare_no_permissions'
+      else
+        application_choice.application_form.safeguarding_issues_status
+      end
+    end
+
+    def safeguarding_issues_declared?
+      application_choice.application_form.safeguarding_issues_status == 'has_safeguarding_issues_to_declare'
```
tijmenb, line 37: `enums` automatically expose helper methods like this:

```suggestion
      application_choice.application_form.has_safeguarding_issues_to_declare?
```

They have the advantage of preventing typos, as it'll raise a `NoMethodError` if you make one.

app/components/provider_interface/safeguarding_declaration_component.rb
```diff
@@ -2,13 +2,43 @@ module ProviderInterface
...
+    end
+
+    def set_details
+      @details = application_choice.application_form.safeguarding_issues
+    end
```
tijmenb, line 42: Is `@details` used? I can't find it in the HTML.

AgaDufrat: It's used here: https://github.com/DFE-Digital/apply-for-teacher-training/blob/3f122fb59ad13134c2cb70b532d38268ab6b4ea4/app/components/provider_interface/safeguarding_declaration_component.html.erb#L10

tijmenb: Ah, it's exposed via `attr_reader`. Should we make this into a method so it's similar to how we use `message`?

AgaDufrat: I was a bit unsure about this bit and that I'm setting the details within message:
```
def message
      set_details if safeguarding_issues_declared? && display_safeguarding_information?
```
but then I only wanted to assign whatever candidate disclosed to `details` if the provider_user can view it. 
I guess we have `<% if display_safeguarding_information? %>` in the view so it should be fine.

Should I go with public
```
   def details
      application_choice.application_form.safeguarding_issues
    end
```
and remove `set_details if ...` from #message?

tijmenb: Yeah, that sounds correct. It removes some duplication as well!

app/components/provider_interface/safeguarding_declaration_component.rb
```diff
@@ -2,13 +2,41 @@ module ProviderInterface
...
+    def details
+      application_choice.application_form.safeguarding_issues
+    end
+
+  private
+
+    def status
+      if safeguarding_issues_declared? && !display_safeguarding_information?
+        'has_safeguarding_issues_to_declare_no_permissions'
+      else
```
duncanjbrown, line 33: Double checking: is it correct that if the user hasn't got permission but issues are not declared that we still show the status? i.e. if you don't have permission you are still told that there are no issues.

app/components/provider_interface/safeguarding_declaration_component.rb
```diff
@@ -2,13 +2,41 @@ module ProviderInterface
...
+    end
+
+    def message
+      SafeguardingStatus.new(
+        status: status,
         i18n_key: 'provider_interface.safeguarding_declaration_component',
       ).message
     end
+
+    def display_safeguarding_information?
```
duncanjbrown, line 19: I think this does the right thing but I could see this method name inviting people to put unrelated logic in the body — how about something like `current_user_has_permission_to_view_safeguarding_information?`

app/components/provider_interface/safeguarding_declaration_component.rb
AgaDufrat: That's my understanding based on [the design history entry](https://bat-design-history.netlify.app/manage-teacher-training-applications/view-safeguarding-information/).

duncanjbrown: ok great :+1:

PR #2640, opened 2020-08-03
Fix for focus appearing on mousedown of checkbox labels

edwardhorsford:
## Context
Quick for [this issue](https://trello.com/c/GSHg4UkG/2552-focus-issue-on-filters-box-when-selecting-items-in-the-filters-box). Because the filter has `tabindex="-1"`, it briefly gets focus each mousedown on a checkbox label.

My fix involves having a _second_ focusable element inside the one we already have. That element has focus styles suppressed. When a user clicks on a checkbox label, it receives focus, but nothing visually changes. When they release their cursor, the checkboxes are focused as normal.

As that second element has `tabindex="-1"` users can't normally tab in to it. As our main filter has normal focus styles, it looks correct when the popover is opened.

## Before
filter-a-list-focus-bug-before

## After
filter-a-list-focus-bug-after2

Note I've [raised an issue upstream](https://github.com/ministryofjustice/moj-design-system/issues/39) with the MOJ design system about the issue.

## Other changes

I fixed the incorrect focus styling on the filter component.

### DIFF
--- a/app/components/provider_interface/filter_component.html.erb
+++ b/app/components/provider_interface/filter_component.html.erb
@@ -1,5 +1,5 @@
 <div class="moj-filter-layout__filter app-filter">
-  <div class="moj-filter" tabindex="-1">
+  <div class="moj-filter">
     <div class="moj-filter__header">
       <div class="moj-filter__header-title">
         <h2 class="govuk-heading-m">Filter</h2>
@@ -36,7 +36,9 @@
         </div>
       <% end %>
 
-      <div class="moj-filter__options">
+      <!-- Div made focusable to 'catch' focus from checkboxes. see https://github.com/DFE-Digital/apply-for-teacher-training/pull/2640
+       -->
+      <div class="moj-filter__options" tabindex="-1">
         <form method="get">
           <%= hidden_field_tag :sort_by, page_state.sort_by %>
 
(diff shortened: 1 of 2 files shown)


edwardhorsford:
@adamsilver updated with my revised solution - the wider component can now show correct focus styles - but you won't get a weird flash when users click the checkbox.

app/components/provider_interface/filter_component.html.erb
```diff
@@ -36,7 +36,9 @@
         </div>
       <% end %>
 
-      <div class="moj-filter__options">
+      <!-- Div made focusable to 'catch' focus from checkboxes. see https://github.com/DFE-Digital/apply-for-teacher-training/pull/2640
```
duncanjbrown, line 39: No need for this comment here or in the CSS — GitHub's merge commit will include the PR reference & we'll be able to get back to it from the diff. Possibly it ought to go into the commit message.

edwardhorsford: @duncanjbrown normally I'd agree - but in this case I think what I'm doing is unexpected and it might be helpful to note this is intentionally weird.

duncanjbrown: As you like!

PR #2641, opened 2020-08-03
Add a feature flag check to ProviderSetup#next_relationship_pending

duncanjbrown:
It was possible to get into a state where we weren't redirected to the setup page because the feature flag was off. However the header menu items were still hidden because `ProviderSetup` reported we still had a relationship to configure.

## Context

Noticed when working locally

Check the flag in `ProviderSetup`, which controls both the controller and the menu. Previously only the controller checked the flag.

## Guidance to review

If the feature flag is turned off, the usual header menu items should appear on the /applications page.

## Link to Trello card

bugfix on https://trello.com/c/ejHRNikf/2509-remove-links-from-header-during-onboarding-process

## Things to check

### DIFF
--- a/app/services/provider_setup.rb
+++ b/app/services/provider_setup.rb
@@ -23,6 +23,8 @@ class ProviderSetup
   end
 
   def next_relationship_pending
+    return unless FeatureFlag.active?('enforce_provider_to_provider_permissions')
+
     permissions = ProviderRelationshipPermissions.find_by(
       setup_at: nil,
       training_provider: @provider_user.providers,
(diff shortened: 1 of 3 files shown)


app/services/provider_setup.rb
```diff
@@ -23,6 +23,8 @@ def next_agreement_pending
   end
 
   def next_relationship_pending
+    return unless FeatureFlag.active?('enforce_provider_to_provider_permissions')
```
mnacos, line 26: Yes, that's a much better place for this check.

PR #2642, opened 2020-08-03
Restrict 'confirm conditions' to provider users with `make_decisions`

mnacos:
## Context

The 'confirm conditions' (met/not met) flow has been unrestricted so far, subject to application visibility. Now that we have started restricting provider user actions based on their per-provider permissions, we have decided to change this and require `make_decisions` permission for any provider action which affects the application lifecycle.

This PR hides the 'Confirm conditions' button in the provider interface when viewing 'accepted' offers and restricts access to the relevant screens. It also goes beyond the UI to enforce this restriction within the backend services, to prevent issues arising from changes to the UI.

Considering we have started using the `can_make_offer?` logic to authorise any decision-related provider action, this PR also renames all instances of `can_make_offer` to `can_make_decisions` for simplicity and consistency.

It also adds an `actor:` context to the `ConfirmEnrolment` service for completeness, even though this is not yet accessible via the UI (but may be triggered via the vendor API).

## Guidance to review

Should be straightforward for visible elements, such as the 'Confirm conditions' button and pages. Ideally, it shouldn't be possible to see any `NotAuthorised` errors from these services via the UI, because our policy is to hide the relevant navigation elements from unauthorised users.

## Link to Trello card

https://trello.com/c/Zfj9ZyLP

## Things to check

### DIFF
--- a/app/controllers/provider_interface/conditions_controller.rb
+++ b/app/controllers/provider_interface/conditions_controller.rb
@@ -1,6 +1,7 @@
 module ProviderInterface
   class ConditionsController < ProviderInterfaceController
     before_action :set_application_choice
+    before_action :requires_make_decisions_permission
 
     def edit
       @conditions_form = ConfirmConditionsForm.new
--- a/app/controllers/provider_interface/conditions_controller.rb
+++ b/app/controllers/provider_interface/conditions_controller.rb
@@ -22,10 +22,18 @@ module ProviderInterface
       redirect_to action: :edit unless @conditions_form.valid?
 
       if @conditions_form.conditions_met?
-        ConfirmOfferConditions.new(application_choice: @application_choice).save || raise('ConfirmOfferConditions failure')
+        ConfirmOfferConditions.new(
+          actor: current_provider_user,
+          application_choice: @application_choice,
+        ).save || raise('ConfirmOfferConditions failure')
+
         flash[:success] = 'Conditions successfully marked as met'
       else
-        ConditionsNotMet.new(application_choice: @application_choice).save || raise('ConditionsNotMet failure')
+        ConditionsNotMet.new(
+          actor: current_provider_user,
+          application_choice: @application_choice,
+        ).save || raise('ConditionsNotMet failure')
+
         flash[:success] = 'Conditions successfully marked as not met'
       end
 
--- a/app/services/conditions_not_met.rb
+++ b/app/services/conditions_not_met.rb
@@ -1,11 +1,14 @@
 class ConditionsNotMet
   include ActiveModel::Validations
 
-  def initialize(application_choice:)
+  def initialize(actor:, application_choice:)
+    @auth = ProviderAuthorisation.new(actor: actor)
     @application_choice = application_choice
   end
 
   def save
+    @auth.assert_can_make_offer!(application_choice: @application_choice, course_option_id: @application_choice.offered_option.id)
+
     ApplicationStateChange.new(@application_choice).conditions_not_met!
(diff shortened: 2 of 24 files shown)


app/services/conditions_not_met.rb
```diff
@@ -1,11 +1,14 @@
...
   include ActiveModel::Validations
 
-  def initialize(application_choice:)
+  def initialize(actor:, application_choice:)
+    @auth = ProviderAuthorisation.new(actor: actor)
     @application_choice = application_choice
   end
 
   def save
+    @auth.assert_can_make_offer!(application_choice: @application_choice, course_option_id: @application_choice.offered_option.id)
```
duncanjbrown, line 10: What do you think about renaming this (and its sister `can_make_offer?`, and the class `OfferAuthorisation`) `assert_can_make_decisions` etc? cc @tijmenb

spec/services/withdraw_offer_spec.rb
```diff
@@ -6,15 +6,24 @@
       application_choice = create(:application_choice, status: :offer)
 
       withdrawal_reason = 'We are so sorry...'
-      WithdrawOffer.new(application_choice: application_choice, offer_withdrawal_reason: withdrawal_reason).save
+      WithdrawOffer.new(
+        actor: create(:support_user),
```
duncanjbrown, line 10: Should we have one test where we make `ProviderAuthorisation` blow up? (here and in the other places it's added)

app/services/conditions_not_met.rb
tijmenb: Makes sense, but also fine to do in a separate PR.

app/controllers/provider_interface/conditions_controller.rb
```diff
@@ -22,10 +23,18 @@ def update
       redirect_to action: :edit unless @conditions_form.valid?
 
       if @conditions_form.conditions_met?
-        ConfirmOfferConditions.new(application_choice: @application_choice).save || raise('ConfirmOfferConditions failure')
+        ConfirmOfferConditions.new(
+          actor: current_provider_user,
+          application_choice: @application_choice,
+        ).save || raise('ConfirmOfferConditions failure')
```
tijmenb, line 29: I know it's not part of this PR, but this should raise a more informative error - this won't tell us what has gone wrong (an invalid state transition I think).

app/services/conditions_not_met.rb
mnacos: I think it's the right thing to do, adding to this PR. `OfferAuthorisation` is only called from `ProviderAuthorisation` and the rest is just a global search/replace for `can_make_offer`.

app/controllers/provider_interface/conditions_controller.rb
mnacos: It seems to be a pattern in many services corresponding to provider actions to `.save` instead of `.save!` and rescue `Workflow::NoTransitionAllowed`, returning `false`. We could refactor these services as a separate PR. If these services raise a `ProviderInterface::NotAuthorisedError` now that we check `make_decisions`, I think we'll be able to see it.

spec/services/withdraw_offer_spec.rb
mnacos: I've added some `NotAuthorised` expectations to `WithdrawOffer`, `ConfirmOfferConditions`, `ConditionsNotMet` and `ConfirmEnrolment` service specs.

duncanjbrown: Nice, thank you!

PR #2643, opened 2020-08-03
Update existing code to be backward compatible for other and non_uk qualifications

davidgisbey:
## Context

If we switch the flag off, users need to be able to update their non_uk and other_uk qualifications.

This PR refactors the non-feature flagged path to update the correct values in the DB (non_uk_qualification_type, other_uk_qualfication_type and institution_country) and use them if populated.

While implementing this I found the following bug on production 

image

- Updating non_uk and other_uk qualifications without the feature flag updates the correct attributes in the DB
- It continues to work correctly by using qualification_type if the above values are not present

## Guidance to review

Once this is merged a data migration will need to take place which sets the other_uk_qualification_type value for non gcse/a-levels/as-levels, then sets the qualificaiton_type to 'Other' so it is in line with how the new implementation works. 

## Link to Trello card

https://trello.com/c/hcn07OYW/1764-dev-%F0%9F%8C%90-other-international-qualification

## Things to check


app/controllers/candidate_interface/other_qualifications/base_controller.rb
```diff
@@ -60,9 +60,17 @@ def other_qualification_params
...
       params.require(:candidate_interface_other_qualification_form).permit(
-        :qualification_type, :subject, :institution_name, :grade, :award_year
-      ).merge!(id: params[:id])
-        .transform_values(&:strip)
+        :subject, :institution_name, :grade, :award_year, :other_uk_qualification_type,
+        :non_uk_qualification_type, :institution_country
+      ).merge!(
+        id: params[:id],
+        qualification_type: params.dig('candidate_interface_other_qualification_form', 'qualification_type' || get_qualification.qualification_type),
+        institution_country: get_qualification.institution_country,
```
stevehook, line 68: Do we always need to ignore the value of `params[:institution_country]`?

app/controllers/candidate_interface/other_qualifications/base_controller.rb
```diff
@@ -60,9 +60,17 @@ def other_qualification_params
...
     def other_qualifcations_update_params
       params.require(:candidate_interface_other_qualification_form).permit(
-        :qualification_type, :subject, :institution_name, :grade, :award_year
-      ).merge!(id: params[:id])
-        .transform_values(&:strip)
+        :subject, :institution_name, :grade, :award_year, :other_uk_qualification_type,
+        :non_uk_qualification_type, :institution_country
+      ).merge!(
+        id: params[:id],
+        qualification_type: params.dig('candidate_interface_other_qualification_form', 'qualification_type' || get_qualification.qualification_type),
```
stevehook, line 67: This line doesn't look quite right but I'm not sure:
```suggestion
        qualification_type: params.dig('candidate_interface_other_qualification_form', 'qualification_type') || get_qualification.qualification_type,
```

app/controllers/candidate_interface/other_qualifications/base_controller.rb
tvararu: The base controller isn't being used in the new implementation, should be okay 👌

PR #2652, opened 2020-08-04
Email candidates when they are rejected by default

stevehook:
## Context

We don't send candidates any emails when their application is rejected by default. We should.

- [x] Modify templates and `CandidateMailer` so that they work for rejected by default as well as plain rejected applications.
- [x] Invoke `SendCandidateRejectionEmail` from `RejectApplicationByDefault` service
- [x] Mailer previews and manual testing

## Guidance to review

- There are 3 new mailer previews that can be checked on the review app:

https://apply-for-te-1891-email-1bj22u.herokuapp.com/rails/mailers/candidate_mailer/application_rejected_by_default_awaiting_decisions
https://apply-for-te-1891-email-1bj22u.herokuapp.com/rails/mailers/candidate_mailer/application_rejected_by_default_all_rejected
https://apply-for-te-1891-email-1bj22u.herokuapp.com/rails/mailers/candidate_mailer/application_rejected_by_default_offers_made

- I've not attempted to cover scenarios with `apply_again` feature flag off - it's probably gone past the point of no return and going to be removed soon.

## Link to Trello card

https://trello.com/c/B77dAIjA/1891-dev-📧-email-candidates-when-their-application-is-rejected-by-default

## Things to check


spec/system/provider_interface/candidate_receives_email_when_an_application_has_been_rejected_by_default_spec.rb
```diff
@@ -0,0 +1,48 @@
+require 'rails_helper'
+
+RSpec.feature 'An application has been rejected by default' do
```
tijmenb, line 3: Perhaps we can rewrite https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/spec/system/provider_interface/provider_receives_email_when_an_application_has_been_rejected_by_default_spec.rb to make it test for both emails? Perhaps a single test `app/system/reject_by_default_spec.rb` for the whole feature.

We can even move https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/spec/system/provider_interface/provider_receives_email_when_an_application_is_getting_close_to_the_reject_by_default_date_spec.rb into that.

app/services/reject_application_by_default.rb
```diff
@@ -14,5 +14,6 @@ def call
 
     StateChangeNotifier.call(:reject_application_by_default, application_choice: application_choice)
     SendRejectByDefaultEmailToProvider.new(application_choice: application_choice).call
+    SendCandidateRejectionEmail.new(application_choice: application_choice).call
```
tijmenb, line 17: I'm wondering how this will work in practice. If the candidate has 3 application choices, they will get 3 emails, but the last one will be different. The "application_rejected_awaiting_decisions" email is technically possible, but only by milliseconds.

app/mailers/candidate_mailer.rb
```diff
@@ -78,7 +78,10 @@ def application_rejected_all_rejected(application_choice)
 
     email_for_candidate(
       application_choice.application_form,
-      subject: I18n.t!('candidate_mailer.application_rejected.all_rejected.subject', provider_name: @course.provider.name),
+      subject: I18n.t!(
+        "candidate_mailer.application_rejected#{application_choice.rejected_by_default ? '_by_default' : ''}.all_rejected.subject",
```
tijmenb, line 82: Since they're all the same, perhaps this is clearer:

```suggestion
        application_choice.rejected_by_default ? "candidate_mailer.application_rejected_by_default.subject" : "candidate_mailer.application_rejected.all_rejected.subject",
```

app/services/reject_application_by_default.rb
stevehook: Yes the RBDs would probably all happen at once for candidates with more than one. I wondered whether we should refactor the `RejectApplicationsByDefault` and `RejectApplicationByDefault` to batch these together somehow but not sure how that would work - it would certainly need new content. Is that something for a follow-up or should we try to tackle it here?

We could defer the rejection emails by a few seconds to give time for `RejectApplicationsByDefault` to finish its cycle. That would at least mean that the multiple emails would be consistent because they would be rendered after all the application choices had been rejected.

spec/system/provider_interface/candidate_receives_email_when_an_application_has_been_rejected_by_default_spec.rb
stevehook: yes that's much better, 3-into-1.

app/services/reject_application_by_default.rb
stevehook: @tvararu & @tijmenb I've changed the way we call the candidate mailer from the `RejectApplicationByDefault` to put in a short delay (1 minute) so that we can avoid the problem that we discussed where all but the last of a set of multiple RBD emails are misleading because they imply that there are still pending decisions. I added an explicit order to the the query that fetches the application choices to process so that the application choices for a given candidate are not potentially spread out over a large result set.

PR #2661, opened 2020-08-05
Require tests to be explicit about which feature flags they rely on

tijmenb:
## Context

There are 2 somewhat related issues with feature flags:

- It is possible to write code that isn't being tested by our integration tests. For example, if you were to add `raise if FeatureFlag.active?(:introduce_bug)` to a controller, the tests will pass without a problem because all flags are off by default.

- When introducing a feature flag, we typically add tests where we turn _on_ the flag and observe the behaviour. 

  When removing feature flags, I've often noticed that there are specs that fail because they actually rely on a feature flag being turned off. But since we don't explicitly turn it off in the code, it isn't easy to find these cases. What complicates the situation is that you have to go back into this failing test and understand it before you can make the changes. This is tricky because you don't have context on the feature, or you're not the person who wrote the tests.

Run the RSpec tests once with all flags on by default, and once with all flags off by default. It all runs in parallel so it won't take more time.

This will force us to explicitly activate or deactivate flags when adding them, using:

```rb
before do
  FeatureFlag.deactivate(:whatever_feature)
end
```

```rb
before do
  FeatureFlag.activate(:whatever_feature)
end
```

It's a little more work upfront, but it'll make it easier and safer to remove feature flags.  

## Guidance to review

Is this sensible?

## Link to Trello card

https://trello.com/c/f6K25rS3/354-feature-flags

## Things to check

### DIFF
--- a/app/services/feature_flag.rb
+++ b/app/services/feature_flag.rb
@@ -57,12 +57,6 @@ class FeatureFlag
     FEATURES[feature_name].feature.active?
   end
 
-  def self.reset!
-    return if Rails.env.production?
-
-    Feature.update_all(active: false)
-  end
-
   def self.sync_with_database(feature_name, active)
     feature = Feature.find_or_initialize_by(name: feature_name)
     feature.active = active
--- a/azure/pipelines/build.yml
+++ b/azure/pipelines/build.yml
@@ -226,16 +226,34 @@ stages:
   - template: templates/rspec-job.yml
     parameters:
       testCommand: 'integration-tests'
-      displayName: 'Integration Tests'
-      jobId: 'integration_tests'
+      displayName: 'Integration Tests with all features off'
+      jobId: 'integration_tests_features_off'
       jobAttempt: $(System.JobAttempt)
+      featureFlagState: 'off'
 
   - template: templates/rspec-job.yml
     parameters:
       testCommand: 'unit-tests'
-      displayName: 'Unit Tests'
-      jobId: 'unit_tests'
+      displayName: 'Unit Tests with all features off'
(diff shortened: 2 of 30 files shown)


stevehook:
I really like the presumption that all feature flags are on by default (with the exception of maybe one or two ‘permanent’ ones like the DSI fallback). I'm not sure how the specs would work if they all had to be run with flags both on and off. I guess you would need to have explicit `FeatureFlag#deactivate`s anywhere that tested the pre-feature functionality and we would still need to add more tests to cover the post-feature functionality.

Maybe this would work with what we talked about yesterday. For a given new flag we might change the functionality covered by a few tests and we could clone each of those, rename one of the clones as `thing_without_feature_flag_x_spec.rb`  and add a `FeatureFlag.deactivate`. The modify `thing_spec.rb` to handle the new behaviour. So when we drop feature flag 'x' it's just a matter of deleting `thing_without_feature_flag_x_spec.rb`.

In practice the spec cloning approach might not work in all cases - probably good for system specs, less so for things like model specs, not sure.

vigneshmsft:
If the problem is that the specs are not removed when a feature flag is removed/turned off, wouldn't it be simpler to assume that the feature flag is `active ` instead of mandating the feature flag be configured for the test to run?
`return true` instead of raising error in https://github.com/DFE-Digital/apply-for-teacher-training/blob/1e910e0086e31cbb6416b1db07ca38f063073b39/app/services/feature_flag.rb#L56

so `raise if FeatureFlag.active?(:introduce_bug)` will fail if `introduce_bug` feature hasn't been configured or has been removed
so a dev will know to remove the no longer relevant test code as well

PR #2668, opened 2020-08-06
Help users move on from error pages

mnacos:
## Context

Our current error pages trap users, as the link at the top of the page just mirrors the requested url (which is how you got to this page in the first place). The text of the link is always 'Apply for teacher training', even if you were visiting a provider or support-interface page.

Make error pages reflect the interface the user intended to visit by varying the header title. Link to the start page of each interface so that the user can get unstuck. 

image
image
image

## Guidance to review

Easiest way to test is request pages that do not exist. Additional work may be needed for custom 'Manage' error pages.

## Link to Trello card

https://trello.com/c/xEwXApm3

## Things to check

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -5,7 +5,7 @@ module ApplicationHelper
   end
 
   def service_name
-    case current_namespace
+    case intended_namespace
     when 'candidate_interface'
       t('service_name.apply')
     when 'provider_interface'
@@ -21,7 +21,7 @@ module ApplicationHelper
     custom_link = content_for(:service_link)
     return custom_link if custom_link
 
-    case current_namespace
+    case intended_namespace
     when 'provider_interface'
       provider_interface_path
     when 'candidate_interface'
@@ -35,5 +35,9 @@ module ApplicationHelper
 
   def current_namespace
     params[:controller].split('/').first
+  end
+
+  def intended_namespace
+    "#{request.path.split('/').second}_interface"
   end
 end
--- a/app/views/errors/not_found.html.erb
+++ b/app/views/errors/not_found.html.erb
@@ -9,10 +9,21 @@
     <p class="govuk-body">
       If you pasted the web address, check you copied the entire address.
     </p>
-    <p class="govuk-body">
-      If the web address is correct or you selected a link or button and you
-      need to speak to someone about this problem, contact the Becoming a Teacher team:
-      <%= bat_contact_mail_to %>.
-    </p>
+    <% if intended_namespace == 'provider_interface' %>
+      <p class="govuk-body">
+        You can
+        <%= govuk_link_to "return to #{service_name}", service_link %>.
+      </p>
+      <p class="govuk-body">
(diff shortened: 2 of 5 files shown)


app/helpers/application_helper.rb
```diff
@@ -36,4 +36,8 @@ def service_link
   def current_namespace
     params[:controller].split('/').first
   end
+
+  def intended_namespace
```
duncanjbrown, line 40: Looking at this, shall we just change the implementation of `current_namespace`? Bit confusing to have two methods that return the same thing in all circs except 404 pages

mnacos: I've started using `intended_namespace` for the footer, too, which will display the relevant footers for candidate and provider error pages e.g.

image

The only thing stopping us from turning `intended_namespace` into the new `current_namespace` is, ironically, the lack of custom provider error pages. This is because the header component will then start displaying the sub-navigation items, but the error page lacks a user session context. For example (for a signed-in provider user):

image

How about we merge this as is (with both `current_namespace` and `intended_namespace`) and merge the two methods in another PR providing interface-specific error controllers and pages (so that they get access to `current_candidate`, `current_provider_user` and `current_support_user`)?

edwardhorsford:
Already looks much better, especially with the more specific footers. 👍

laurahermionetennant92:
Content here @mnacos (FYI @edwardhorsford)

Page not found

If you entered a web address, check it is correct.

If you pasted the web address, check you copied the entire address.

You can [return to your Manage teacher training applications dashboard](url).

Or, if you need to speak to someone about this problem, email us on becomingateacher@digital.education.gov.uk.

mnacos:
> Content here @mnacos (FYI @edwardhorsford)
> 
> Page not found
> 
> If you entered a web address, check it is correct.
> 
> If you pasted the web address, check you copied the entire address.
> 
> You can [return to your Manage teacher training applications dashboard](url).
> 
> Or, if you need to speak to someone about this problem, email us on [becomingateacher@digital.education.gov.uk](mailto:becomingateacher@digital.education.gov.uk).

Great, have updated the page:

image

edwardhorsford:
@laurahermionetennant92 Should the line about 'return to dashboard' only show if signed in?

laurahermionetennant92:
@edwardhorsford I was wondering about that – so you landed on that page without being signed in, is that what you were explaining?
If a user isn't signed in, the line should read, Sign in to your Manage teacher training applications account

??

mnacos:
> @laurahermionetennant92 Should the line about 'return to dashboard' only show if signed in?

Within the context of the current error pages we cannot detect if the user is signed in or not. But Manage _is_ about the dashboard, so maybe ok to say 'dashboard' regardless?

laurahermionetennant92:
New content @mnacos :)


Page not found

If you entered a web address, check it is correct.

If you pasted the web address, check you copied the entire address.

You can return to Manage teacher training applications.

Or, if you need to speak to someone about this problem, email us on becomingateacher@digital.education.gov.uk.

PR #2669, opened 2020-08-06
Fix AbstractController::DoubleRenderError in ProviderUsersController

AgaDufrat:
## Context

`find_provider_user` was called directly in an action. We think this may have resulted in [double rendering error](https://sentry.io/organizations/dfe-bat/issues/1826861551/?project=1765973) as the action would not have halted at the first render. So we moved this method to a before filter. 

_If a "before" filter renders or redirects, the action will not run. If there are additional filters scheduled to run after that filter, they are also cancelled._ [https://guides.rubyonrails.org/action_controller_overview.html#filters](https://guides.rubyonrails.org/action_controller_overview.html#filters)

Make `find_provider_user` a before filter.

## Guidance to review

The controller still works

## Link to Trello card

https://trello.com/c/ibPX7YvE/2574-fix-abstractcontrollerdoublerendererror

## Things to check

### DIFF
--- a/app/controllers/provider_interface/provider_users_controller.rb
+++ b/app/controllers/provider_interface/provider_users_controller.rb
@@ -2,6 +2,7 @@ module ProviderInterface
   class ProviderUsersController < ProviderInterfaceController
     before_action :require_feature_flag!
     before_action :require_manage_users_permission!
+    before_action :find_provider_user, except: %i[index]
 
     def index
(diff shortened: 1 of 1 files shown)


AgaDufrat:
I couldn't find if rescue_from behaves like the before filter and stop rendering/redirecting.
I'll merge it as it is because I read that if a render or redirect is already in progress when the exception is raised, it's possible to get double render error from the `rescue_from`.

PR #2671, opened 2020-08-07
[SPIKE] Stop candidates being able to add a course/be sent to find 

davidgisbey:
## Context

This is a spike related to adding/replacing courses around the EOC.

This PR tackles three issues.

1. Candidates should not be able to add new course once their applications phases deadline has passed (apply1/apply again)
2. Same with replacing an existing course
3. All links to Find need to be removed while Find is down
 
- Remove all links to FIND outside of the add/replace course flow when Find is down
- Redirect the add/replace course choice controllers to the review/dashboard page if a candidate cannot add a new course
- Remove the replace course banner (post-submission)
- Add a new banner on the review page informing candidates the current cycle has finished 

## Guidance to review

This is very light on tests. Happy to add more. I'm not 100% sure how many you should be writing while spiking.

I'm assuming we are still going to use a feature flag for testing purposes so all our other tests don't just start suddenly failing during these time periods.

Have I missed anything?

## Link to Trello card

https://trello.com/c/MaIKxcbX/1909-spike-freeze-stop-candidates-looking-at-new-courses-during-find-close-down-25-8-a1-19-septaa

## Things to check

### DIFF
--- a/app/components/candidate_interface/find_down_course_choices_banner.html.erb
+++ b/app/components/candidate_interface/find_down_course_choices_banner.html.erb
@@ -1,0 +1,15 @@
+<% if @find_down %>
+  <div class="app-banner">
+    <div class="app-banner__message">
+      <div class="govuk-heading-m">
+        <p>
+          The 2019/20 Application Cycle has ended.
+        </p>
+
+        <p class='govuk-body govuk-!-font-size-24'>
+          You will be able to add course choices for the 2020/21 Application Cycle from <%= EndOfCycleTimetable.find_reopens.to_s(:govuk_date).strip %>.
+        </p>
+      </div>
+    </div>
+  </div>
+<% end %>
--- a/app/components/candidate_interface/find_down_course_choices_banner.html.erb
+++ b/app/components/candidate_interface/find_down_course_choices_banner.html.erb
@@ -1,4 +1,4 @@
-<% if @find_down %>
+<% unless CandidateInterface::CanAddCourseChoice.can_add_course_choice?(application_form: @application_form) %>
   <div class="app-banner">
     <div class="app-banner__message">
       <div class="govuk-heading-m">
--- a/app/components/candidate_interface/rejection_reasons_component.rb
+++ b/app/components/candidate_interface/rejection_reasons_component.rb
@@ -19,9 +19,7 @@ module CandidateInterface
     def course_details_row(application_choice)
       {
         key: 'Course',
-        value: govuk_link_to(application_choice.offered_course.name_and_code,
-                             application_choice.offered_course.find_url, target: '_blank', rel: 'noopener') +
-          tag.p(application_choice.course.description, class: 'govuk-body'),
+        value: course_details_row_value(application_choice),
       }
     end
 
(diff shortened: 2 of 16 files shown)


app/services/candidate_interface/can_add_course_choice.rb
```diff
@@ -0,0 +1,16 @@
+module CandidateInterface
+  class CanAddCourseChoice
+    def self.can_add_course_choice?(application_form:)
```
davidgisbey, line 3: if we use a feature flag i'll add something like `return true unless FeatureFlag.active('end_of_cycle')`

app/components/candidate_interface/rejection_reasons_component.rb
```diff
@@ -31,5 +29,15 @@ def rejection_reasons_row(application_choice)
         value: application_choice.rejection_reason,
       }
     end
+
+    def course_details_row_value(application_choice)
+      if EndOfCycleTimetable.find_down?
+        "#{application_choice.offered_course.name_and_code} <br> #{application_choice.course.description}".html_safe
```
malcolmbaig, line 35: Do we know if the course data from Find is adequately sanitised? If not, we may need to rethink how to display this so that we aren't marking the entire string safe.

app/components/candidate_interface/find_down_course_choices_banner.html.erb
```diff
@@ -0,0 +1,15 @@
+<% unless CandidateInterface::CanAddCourseChoice.can_add_course_choice?(application_form: @application_form) %>
```
malcolmbaig, line 1: Should this check be `EndOfCycleTimetable.find_down?`, given the subtle logical difference between that and `can_add_course_choice?`

app/components/candidate_interface/rejection_reasons_component.rb
davidgisbey: Provider and Course codes are generated on the backend. I'm pretty certain it's not possible to change them via the UI.

app/components/candidate_interface/find_down_course_choices_banner.html.erb
davidgisbey: I think the banner should actually be renamed to candidate_cannot_add_courses_banner.rb or something. 

It's a bit of a moot point anyway though as I just looked at the designs and you there's no longer a link to get to the courses section. It now looks like this:

image

app/services/candidate_interface/can_add_course_choice.rb
```diff
@@ -0,0 +1,16 @@
+module CandidateInterface
+  class CanAddCourseChoice
```
malcolmbaig, line 2: Thinking about the interface here - the question of 'can I add courses to this application form?' is much broader than the end of cycle logic implemented in the class body. For example, we prevent adding courses when the max number of choices has been reached.

I wonder if something like `EndOfCyclePolicy.can_add_course_choice?(application_form)` would be clearer? That way, at each call site it's obvious this check relates to end of cycle concerns.

app/components/candidate_interface/rejection_reasons_component.rb
malcolmbaig: Same with course descriptions?

app/services/candidate_interface/can_add_course_choice.rb
davidgisbey: Thanks Malcolm, that is a really good suggestion. I was struggling to come with a good name for this 👍

tvararu:
:shipit:

PR #2672, opened 2020-08-07
Replace dumb quotes with smart ones

paulrobertlloyd:
…because we’re not animals.

* Replaces all user-facing ‘dumb’ quotes I could find with ‘smart’ ones
* Adds some consistency around quoting in YAML in the few places I came across while doing this (not exhaustive)
* Tidies up header syntax in a markdown file

### DIFF
--- a/config/vendor-api-v1.yml
+++ b/config/vendor-api-v1.yml
@@ -7,7 +7,7 @@ info:
     name: DfE
     email: becomingateacher@digital.education.gov.uk
   description: |
-    API for DfE's Apply for postgraduate teacher training service.
+    API for DfE’s Apply for teacher training service.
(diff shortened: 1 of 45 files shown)


config/vendor-api-v1.yml
```diff
@@ -7,7 +7,7 @@ info:
     name: DfE
     email: becomingateacher@digital.education.gov.uk
   description: |
-    API for DfE's Apply for postgraduate teacher training service.
+    API for DfE’s Apply for postgraduate teacher training service.
```
duncanjbrown, line 10: Might as well fix this while we’re here

```suggestion
    API for DfE’s Apply for teacher training service.
```

PR #2674, opened 2020-08-10
Generate safeguarding issues correctly

AgaDufrat:
## Context

We have `safeguarding_issues_status` column in a database which is set to "not_answered_yet" by default.
There is also `safeguarding_issues` column which contains whatever the candidate disclosed. 

The factory sets `safeguarding_issues` but the `safeguarding_issues_status` is "not_answered_yet".
If something is disclosed, the `safeguarding_issues_status` should be "has_safeguarding_issues_to_declare".

Update GenerateTestData and TestApplication class used by the worker to set the status and `safeguarding_issues` correctly by introducing new factory traits. 

## Guidance to review

- test GenerateTestData: `GenerateTestData.new(10).generate`
- test the worker
  - Comment out all the states in except `create states: [:awaiting_provider_decision]` in app/workers/generate_test_applications.rb
  - Comment out throwing exemption when applying to the same course multiple times in app/lib/test_applications.rb L55-57
  - Create some courses open on apply `FactoryBot.create(:course_option, course: FactoryBot.create(:course, :open_on_apply))`
  - Run the worker  `GenerateTestApplications.new.perform`

## Link to Trello card

https://trello.com/c/9NxZFKHc/2397-correctly-display-the-criminal-convictions-and-professional-misconduct-answer-to-providers-when-there-is-no-information-provided

## Things to check

### DIFF
--- a/app/lib/test_applications.rb
+++ b/app/lib/test_applications.rb
@@ -57,8 +57,15 @@ class TestApplications
     end
 
     Audited.audit_class.as_user(candidate) do
+      traits = %i[
+        with_safeguarding_issues_disclosed
+        with_no_safeguarding_issues_to_declare
+        with_safeguarding_issues_never_asked
+      ].sample
+
       @application_form = FactoryBot.create(
         :completed_application_form,
+        traits,
         application_choices_count: 0,
         full_work_history: true,
         volunteering_experiences_count: 1,
--- a/app/services/generate_test_data.rb
+++ b/app/services/generate_test_data.rb
@@ -22,6 +22,9 @@ class GenerateTestData
       Audited.audit_class.as_user(candidate) do
         traits = [:with_completed_references]
         traits << :with_equality_and_diversity_data if rand < 0.55
+        traits << %i[with_safeguarding_issues_disclosed
+                     with_no_safeguarding_issues_to_declare
+                     with_safeguarding_issues_never_asked].sample
         application_form = FactoryBot.create(
           :completed_application_form,
           *traits,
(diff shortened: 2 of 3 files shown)


app/lib/test_applications.rb
```diff
@@ -57,8 +57,11 @@ def create_application(states:, courses_to_apply_to: nil, apply_again: false, co
     end
 
     Audited.audit_class.as_user(candidate) do
+      traits = %i[with_safeguarding_issues_disclosed with_no_safeguarding_issues_to_declare].sample
+
       @application_form = FactoryBot.create(
         :completed_application_form,
+        *traits,
```
duncanjbrown, line 64: Unlike below, won't this just be a symbol anyway? could be passed in directly without a splat

app/lib/test_applications.rb
```diff
@@ -57,8 +57,11 @@ def create_application(states:, courses_to_apply_to: nil, apply_again: false, co
     end
 
     Audited.audit_class.as_user(candidate) do
+      traits = %i[with_safeguarding_issues_disclosed with_no_safeguarding_issues_to_declare].sample
```
duncanjbrown, line 60: Should we have `never_asked` here too?

app/services/generate_test_data.rb
```diff
@@ -22,6 +22,7 @@ def generate
       Audited.audit_class.as_user(candidate) do
         traits = [:with_completed_references]
         traits << :with_equality_and_diversity_data if rand < 0.55
+        traits << %i[with_safeguarding_issues_disclosed with_no_safeguarding_issues_to_declare].sample
```
duncanjbrown, line 25: Should we do it here too?

```suggestion
        traits << %i[with_safeguarding_issues_disclosed with_no_safeguarding_issues_to_declare with_safeguarding_issues_never_asked].sample
```

PR #2678, opened 2020-08-10
Deprecate Withdrawal.reason in the API

duncanjbrown:
## Context

We don't collect this information so this field has always been empty. We're now going to actually deprecate it.

Deprecate the field 

- set `deprecated: true` in the openAPI spec
- show bold "DEPRECATED" text in the API docs for such fields
- add a note to the field description explaining that it's deprecated

## Link to Trello card

https://trello.com/c/nSOpT0Cs/2499-%F0%9F%8F%88-deprecate-candidate-reason-for-withdrawal-in-api-responses

## Things to check

### DIFF
--- a/app/presenters/api_docs/api_schema.rb
+++ b/app/presenters/api_docs/api_schema.rb
@@ -48,6 +48,10 @@ module APIDocs
         attributes['nullable']
       end
 
+      def deprecated?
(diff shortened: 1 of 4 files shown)


duncanjbrown:
@tijmenb you might want to upstream this into the docs library?

PR #2679, opened 2020-08-10
Use print helper from govuk-frontend

paulrobertlloyd:
`govuk-frontend` v3.6.0 introduced a new utility helper `govuk-!-display-none-print`, which replicates the custom one we created (`app-!-print-display-none`). Given we are now on `govuk-frontend` v3.8.0, we can use the helper in frontend instead of our own.


edwardhorsford:
So close with your classname too!

PR #2688, opened 2020-08-11
Correct button markup for invite and delete provider user actions

steventux:
## Context

We've [defined a helper method](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/helpers/view_helper.rb#L20-L24) to consistently render links in the style of `govuk-button` so we should use it throughout the app where applicable.

- Changes candidate, provider and support interface instances of `link_to` and `govuk_link_to` with the `govuk-button` class
- Removes any duplicated attributes provided by helper.

## Link to Trello card

https://trello.com/c/wLY6kAjh/2542-correct-button-markup-for-invite-and-delete-provider-user-actions

## Things to check

### DIFF
--- a/app/views/provider_interface/provider_users/index.html.erb
+++ b/app/views/provider_interface/provider_users/index.html.erb
@@ -2,10 +2,9 @@
 
 <h1 class="govuk-heading-xl">Users</h1>
 
-<%= link_to(
+<%= govuk_button_link_to(
   'Invite user',
   provider_interface_edit_invitation_basic_details_path,
-  class: 'govuk-button',
 ) %>
 
 <div class="govuk-body">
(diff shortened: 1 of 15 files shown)


app/views/provider_interface/provider_users/index.html.erb
```diff
@@ -2,10 +2,9 @@
 
 <h1 class="govuk-heading-xl">Users</h1>
 
-<%= link_to(
+<%= govuk_button_link_to(
```
duncanjbrown, line 5: There's also `Delete user` in `provider_users/show.html.erb`

steventux: Fixed up here https://github.com/DFE-Digital/apply-for-teacher-training/pull/2688/commits/47f24401712baa5758f4f6c9924a5316f9b97939#diff-4fc2354ff570c561338be2d352678508R23

PR #2695, opened 2020-08-12
Make magic links single-use

malcolmbaig:
## Context

Magic links currently last for one hour, with no limit on their use
during that time. A recent pentest recommended that we make magic links
single-use.

- Expire the token after a successful sign in.
- Amend the copy on the expired page to reflect the new behaviour.
## Link to Trello card

https://trello.com/c/aNcIaroP

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/sign_in_controller.rb
+++ b/app/controllers/candidate_interface/sign_in_controller.rb
@@ -28,6 +28,7 @@ module CandidateInterface
         sign_in(candidate, scope: :candidate)
         add_identity_to_log candidate.id
         candidate.update!(last_signed_in_at: Time.zone.now)
+        candidate.expire_magic_link_token!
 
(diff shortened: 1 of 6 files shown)


malcolmbaig:
@tijmenb @tvararu was thinking something like this. WDYT?

benilovj:
On the [Get Help With Tech service](https://get-help-with-tech.education.gov.uk), we hit an issue with single use magic links that's worth being aware of, in case it starts happening when this is merged:

https://github.com/DFE-Digital/get-help-with-tech/pull/149

We saw these issues with users from 2 local authorities but we only had a small number of LAs on the service at the time, so it was hard to gauge the actual scale of the problem.

edwardhorsford:
I believe Notify also implemented an interstitial because of email clients pre-visiting links.

malcolmbaig:
@benilovj @edwardhorsford thanks for your comments, adding the additional step definitely makes sense. I'll update the PR :+1:

PR #2699, opened 2020-08-12
Sanitise CSV output by removing spreadsheet formulae

stevehook:
## Context

We are at risk of executing unwanted code on our devices when downloading CSV exports from Support.

- [x] Sanitise all text output in CSV values to filter out values that would be interpreted as formulae when imported into Excel etc.
- [x] `CsvHelper` service to encapsulate sanitisation rules. The rules are borrowed from https://github.com/DFE-Digital/get-help-with-tech/pull/38 as linked in the card.

## Guidance to review

- Is there are better name for `CsvHelper`?
- Have I missed anything?

## Link to Trello card

https://trello.com/c/N6xPaDWM/1935-pentest-44-sanitise-excel-csv-formulas-when-exporting-medium-⚠%EF%B8%8F

## Things to check


app/services/csv_helper.rb
```diff
@@ -0,0 +1,11 @@
+class CsvHelper
```
tvararu, line 1:
```suggestion
class CSVHelper
```

Nit! I think the lowercase version looks weird.

stevehook: Yes I agree, I was having issues with Zeitwerk expecting to find it in `c_s_v_helper.rb` but worked out the fix for that now.

PR #2701, opened 2020-08-12
Minor UI tweaks to OtherQualifications

davidgisbey:
## Context

I ran through how other qualifications work with Paul. He made a couple of comments on how it needed to be changed.

This PR implements those changes.

1. The key for each row should show optional if it is (for int quals this is subject and grade)
2. If a value has no been provided for a row, it should show 'Not entered'

| Before | After |
| ------------ | ------------ |
| image | image |

## Things to check


paulrobertlloyd:
LGTM. Screenshot shows `Grade(optional)` but looking at the code maybe this is an older screenshot, and that’s been fixed?

davidgisbey:
Yea sorry i spotted that while testing locally and fixed it.

PR #2706, opened 2020-08-13
Add a JSON 500 error response to API

steventux:
## Context

If API clients trigger 500s they get a splurge of HTML, which looks unprofessional and isn't useful

- Adds a JSON format handler in `ErrorsController` to catch unexpected errors
- Responds with a machine readable error

image

## Link to Trello card

https://trello.com/c/D4aVT7Ar/2599-we-dont-have-a-json-500-error-response

## Things to check

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -18,6 +18,21 @@ class ErrorsController < ApplicationController
   end
 
   def internal_server_error
-    render 'internal_server_error.html', status: :internal_server_error
+    respond_to do |format|
+      format.json do
+        render json: {
+          errors: [
+            {
+              error: 'InternalServerError',
+              message: 'The server encountered an unexpected condition that prevented it from fulfilling the request',
+            },
+          ],
+        }, status: :internal_server_error
+      end
+
+      format.any do
+        render 'internal_server_error.html', status: :internal_server_error
+      end
+    end
   end
 end


app/controllers/vendor_api/vendor_api_controller.rb
```diff
@@ -3,6 +3,7 @@ class VendorAPIController < ActionController::API
     include ActionController::HttpAuthentication::Token::ControllerMethods
     include LogQueryParams
 
+    rescue_from StandardError, with: :internal_server_error
```
tijmenb, line 6: I think this will prevent the errors from reaching Sentry, can you check?

app/controllers/vendor_api/vendor_api_controller.rb
```diff
@@ -44,6 +45,10 @@ def parameter_invalid(e)
       render json: { errors: [{ error: 'ParameterInvalid', message: e }] }, status: :unprocessable_entity
     end
 
+    def internal_server_error(e)
+      render json: { errors: [{ error: 'InternalServerError', message: e }] }, status: :internal_server_error
```
tijmenb, line 49: This exception message is probably going to be uninformative, perhaps use a hardcoded string:

```suggestion
      render json: { errors: [{ error: 'InternalServerError', message: 'The server encountered an unexpected condition that prevented it from fulfilling the request' }] }, status: :internal_server_error
```

app/controllers/vendor_api/vendor_api_controller.rb
steventux: :+1: I don't think they will reach Sentry from what I can see here https://gitlab.com/gitlab-org/gitlab-foss/-/issues/21043 and here https://github.com/getsentry/raven-ruby/search?q=rescue_from&unscoped_q=rescue_from. I'll ensure this happens

steventux: @tijmenb https://github.com/DFE-Digital/apply-for-teacher-training/pull/2706/commits/af66c5ae6d0ff0e43dcf3343a01a27b1fecac37d pushes exceptions to Sentry.
It will also raise and return if in development mode as mentioned.

steventux:
@tijmenb yes this works given the appropriate headers, I assume this is preferable in terms of simplicity?

steventux:
@tijmenb updated, a lot simpler - thanks :+1:

PR #2709, opened 2020-08-13
Communicate date 2021 cycles open

george-bit:
## Context

As a candidate
I need to know when applications for courses starting in 2021 open
So that I can submit my application to providers

Have created a new banner component to display reopening message and date. We have modified the apply again component to not show between cycles.

Before:
image

After:
image

## Guidance to review

Steve and I decided against a Feature flag on the basis that his is a relatively minor change that will have limited scope for error. Would be interested to know if there any differing views on this? As for the code please sanity check. In order to test the banner logic in practice you will need to load the review app and toggle your system clock to inside and then outside the between cycle window.

## Link to Trello card

https://trello.com/c/seZyyjfm/1933-dev-%F0%9F%9A%B2%F0%9F%94%9A-communicate-date-when-applications-for-2021-open

## Things to check

### DIFF
--- a/app/components/candidate_interface/reopen_banner_component.rb
+++ b/app/components/candidate_interface/reopen_banner_component.rb
@@ -1,0 +1,43 @@
+class CandidateInterface::ReopenBannerComponent < ViewComponent::Base
+  attr_accessor :phase, :flash_empty
+
+  def initialize(phase:, flash_empty:)
+    @phase = phase
+    @flash_empty = flash_empty
+  end
+
+  def show?
+    flash_empty &&
+      (show_apply_1_deadline_banner? || show_apply_2_deadline_banner?)
+  end
+
+  def
+    apply_1? ? apply_1_deadline : apply_2_deadline
+  end
+
+private
+
+  def show_apply_1_deadline_banner?
+    apply_1? &&
+      EndOfCycleTimetable.show_apply_1_deadline_banner? &&
+      FeatureFlag.active?(:deadline_notices)
+  end
+
+  def show_apply_2_deadline_banner?
+    apply_2? &&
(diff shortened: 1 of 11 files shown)


app/components/candidate_interface/reopen_banner_component.rb
```diff
@@ -0,0 +1,39 @@
...
+    flash_empty &&
+      (show_apply_1_reopen_banner? || show_apply_2_reopen_banner?)
+  end
+
+private
+
+  def show_apply_1_reopen_banner?
+    apply_1? &&
+      EndOfCycleTimetable.show_apply_1_reopen_banner?
+    # FeatureFlag.active?(:deadline_notices)
```
tvararu, line 19: What's this comment for?

spec/components/candidate_interface/reopen_banner_component_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+          phase: application_form.phase,
+          flash_empty: flash.empty?,
+        ),
+      )
+
+      expect(result.text).to include('Applications for courses starting this academic year have now closed')
+      expect(result.text).to include('Submit your application from 13 October 2020 for courses starting in the next academic year.')
+    end
+
+    it 'does NOT render when we not between cycles' do
```
tvararu, line 44:
```suggestion
    it 'does NOT render when we are not between cycles' do
```

app/components/candidate_interface/reopen_banner_component.rb
stevehook: Oops that's a mistake, we were debating whether we needed a feature flag or not but opted against

george-bit: Thanks for fixing everything Steve :)

PR #2710, opened 2020-08-13
Signing out of Manage should also sign you out of DfE Sign-In

mnacos:
## Context

This PR tweaks a previous spike https://github.com/DFE-Digital/apply-for-teacher-training/pull/2433 and prepares the code for release.

Currently, provider and support users signing out of Apply retain their DSI session and are able to sign-in to Apply again as the same user, unchallenged, simply by clicking on the Sign-in button. This is problematic, as:
- one provider/support user may easily impersonate another (even inadvertently), when sharing computers
- users with multiple identities find it hard to switch between different DSI profiles 

This PR addresses such cases by attempting to log out provider and support users from DSI immediately after destroying their Apply session. This is not guaranteed to succeed, of course, but should work fine in the majority of cases. 

The focus here is Apply-initiated logouts. If the user signs out of DSI either directly from their pages or via a similar logout process triggered by another DfE service, our service will not know until the Apply session expires. There is nothing we can do about this unless DSI starts supporting back-channel logout flows, at which point we should probably start persisting `id_token` values in the database.

Only minor changes to the original spike code were necessary (https://github.com/DFE-Digital/apply-for-teacher-training/pull/2710/commits/ced4912752851ca44d06934a62cbb300fe68ab7e).

I investigated whether we could use the built-in redirect provided by the `omniauth_openid_connect` gem. If you visit to `/auth/dfe/logout`, the gem is pre-programmed to redirect you to `end_session_endpoint`, which is configured automatically via the discovery process (see https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/config/initializers/omniauth.rb#L10). This auto-discovered endpoint resolves to the same IdP path `/session/end` that we have hardcoded. However, this built-in redirect seems to lack query params such as `id_token_hint` (which perhaps we could pass) and we have decided to first destroy the Apply session, then redirect, so it's not worth pursuing this, especially as `end_session_endpoint` is unlikely to ever change.

I have also verified this PR does not break the DSI bypass mechanism we use for development and review apps or sign-outs for real users who have signed-in before this code goes live (and therefore lack an `id_token` in their session. Signing-in and out when `:dfe_sign_in_fallback` feature flag is on works as expected, too.

## Guidance to review

I've fixed a few things mentioned in the original PR but did not change:
- The name of `id_token`. 'ID Token' is a very specific OIDC term, and such tokens are required for logout flows. OIDC can be a little obscure, but anyone familiar with the spec will instantly recognise it. The rest of us can probably read the code and see it's only ever used for logging out. :)
- The values passed to `state=`. `provider` and `support` are concise and clear to anyone familiar with this service.

Finally, the defensive `dfe_sign_in_user&.needs_dsi_signout?` code is there in case the redirect to `end_session_endpoint` fails for some reason (DSI misconfiguration or outage) and the user clicks 'Back' on their browser, which will take them back to the destroy action without a `dfe_sign_in_user`.

To test interactively, a working DfE Sign-In integration is needed. Then the DSI service configuration needs to be altered to include the relevant logout redirect for your environment. For example,

image

Do we need more tests?

## Link to Trello card

https://trello.com/c/k9crMTdY

## Things to check

### DIFF
--- a/app/controllers/support_interface/sessions_controller.rb
+++ b/app/controllers/support_interface/sessions_controller.rb
@@ -1,6 +1,6 @@
 module SupportInterface
   class SessionsController < SupportInterfaceController
-    skip_before_action :authenticate_support_user!
+    skip_before_action :authenticate_support_user!, except: :destroy
 
     def new
       session['post_dfe_sign_in_path'] ||= support_interface_path
@@ -13,7 +13,7 @@ module SupportInterface
       post_signout_redirect = if dfe_sign_in_user.needs_dsi_signout?
                                 dfe_sign_in_user.support_interface_dsi_logout_url
                               else
-                                provider_interface_path
+                                support_interface_path
                               end
 
       DfESignInUser.end_session!(session)
--- a/app/controllers/support_interface/sessions_controller.rb
+++ b/app/controllers/support_interface/sessions_controller.rb
@@ -10,9 +10,14 @@ module SupportInterface
     end
 
     def destroy
-      DfESignInUser.end_session!(session)
+      post_signout_redirect = if dfe_sign_in_user.needs_dsi_signout?
+                                dfe_sign_in_user.support_interface_dsi_logout_url
+                              else
+                                provider_interface_path
+                              end
 
-      redirect_to action: :new
+      DfESignInUser.end_session!(session)
+      redirect_to post_signout_redirect
     end
 
     def sign_in_by_email
--- a/app/controllers/dfe_sign_in_controller.rb
+++ b/app/controllers/dfe_sign_in_controller.rb
@@ -28,6 +28,19 @@ class DfESignInController < ActionController::Base
 
   alias_method :bypass_callback, :callback
 
+  # This is called by a redirect from DfE Sign-in after visiting the signout
+  # link on DSI. We tell DSI to redirect here using the
+  # post_logout_redirect_uri parameter - see DfESignInUser#dsi_logout_url
+  #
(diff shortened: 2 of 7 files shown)


app/controllers/support_interface/sessions_controller.rb
```diff
@@ -10,9 +10,14 @@ def new
     end
 
     def destroy
-      DfESignInUser.end_session!(session)
+      post_signout_redirect = if dfe_sign_in_user&.needs_dsi_signout?
```
duncanjbrown, line 13: How could we get here and not have a `dfe_sign_in_user`?

mnacos: Well, dfe_sign_in_user is really just `DfESignInUser.load_from_session(session)` so if you visit `/provider/sign-out` or `/support/sign-out` while unauthenticated (possible because we skip auth in this controller) this will be `nil` and without the paranoia you get a 500.

I encountered this while testing the feature when the test DSI env wasn't properly configured, which stopped the flow of redirects and presented me with a DSI error page. I then clicked the browser's back button, which showed me the applications list, with a Sign out link at the top. So I tried to sign out again.

It's not super important but I'm trying to prevent generating 500 errors from simply visiting sign-out urls.

mnacos: On second thoughts, I'll remove the `&.` and ensure `:authenticate_*_user!` methods are called for session `:destroy` actions.

PR #2712, opened 2020-08-13
Make /application/:id/offer idempotent

duncanjbrown:
## Context

A vendor is concerned that offers will get lost due to transient network issues and have requested that this endpoint in particular be made idempotent. Presumably the reliability of on-premise networks at providers varies quite a bit, so we might want to do this for other endpoints in time.

We can turn it into a proper PUT in a future version of the API.

- cause `ChangeOffer` to throw a validation error only when it receives a changeset which would not change the offer OR conditions. Old behaviour was to reject anything that would not change the offer, ignoring the conditions. As a (beneficial) side effect it is now possible to change conditions without changing the offered course.
- one option was to not raise a validation error under those circs and simply no-op. But having the `save` method return `true` after doing nothing seems confusing and likely to encourage bugs. Therefore to implement idempotence I decided to...
- expose `ChangeOffer#identical_to_existing_offer?` so clients can check if they're going to submit such an offer. Check this in the API `DecisionsController` and behave accordingly

## Guidance to review

- Do you agree with the above reasoning?
- Fire up an HTTP client and make some repeated offers. See that they do not 422.

## Link to Trello card

https://trello.com/c/0rHELylc/2518-allow-repeated-posts-to-the-offer-endpoint-with-the-same-details

## Things to check

### DIFF
--- a/app/controllers/vendor_api/decisions_controller.rb
+++ b/app/controllers/vendor_api/decisions_controller.rb
@@ -17,16 +17,31 @@ module VendorAPI
                         application_choice.course_option
                       end
 
-      service = application_choice.offer? ? ChangeOffer : MakeAnOffer
+      if application_choice.offer?
+        change_offer = ChangeOffer.new(
+          actor: @api_user,
+          application_choice: application_choice,
+          course_option: course_option,
+          offer_conditions: params.dig(:data, :conditions),
+        )
 
-      decision = service.new(
-        actor: @api_user,
-        application_choice: application_choice,
-        course_option: course_option,
-        offer_conditions: params.dig(:data, :conditions),
-      )
+        if change_offer.identical_to_existing_offer?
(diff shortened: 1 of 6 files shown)


tijmenb:
To achieve elite HTTP level, we could respond with 200 for non-changed offers, and 201 otherwise.

duncanjbrown:
> To achieve elite HTTP level, we could respond with 200 for non-changed offers, and 201 otherwise.

YES

duncanjbrown:
> To achieve elite HTTP level, we could respond with 200 for non-changed offers, and 201 otherwise.

Actually, this is a bit complicated because right now every response is generated by  `respond_to_decision` and I'm not convinced that 201 is equally applicable for, say, marking conditions met and creating an offer. Preferring to be slightly imprecise than more precise but sometimes wrong.

PR #2717, opened 2020-08-14
Add missing audible label for first line of address

paulrobertlloyd:
## Context

According to [the pattern described in the design system](https://design-system.service.gov.uk/patterns/addresses/), the label for first line of the address should announce that it is the first line of 2 (the second label is entirely hidden, and announces that it is line 2 of 2).

Add the missing (but visually hidden) `line 1 of 2` text to the label for `address_line1`. This follows the method recommended by @peteryates in https://github.com/DFE-Digital/govuk_design_system_formbuilder/issues/177#issuecomment-674234646

## Guidance to review

Are we happy using an inline `safe_join` method?

## Link to Trello card

https://trello.com/c/lQPig396

### DIFF
--- a/app/views/candidate_interface/contact_details/address/edit.html.erb
+++ b/app/views/candidate_interface/contact_details/address/edit.html.erb
@@ -17,7 +17,7 @@
               <%= t('page_titles.address') %>
             </h1>
           </legend>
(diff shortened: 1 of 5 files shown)


paulrobertlloyd:
Not quite sure what’s going on with the 2 failures (I updated the tests to fix the other 16 failures by using `match: :prefer_exact`). Any clues?

tvararu (approved):
LGTM as long as tests can be fixed to pass!

PR #2726, opened 2020-08-17
Prevent submitting applications between cycles

stevehook:
## Context

We need to prevent candidates from submitting their application during the 'closed period' between recruitment cycles.

- [x] The 'Review and submit your application' button is replaced with a 'Review' section and notice to explain that submissions are closed when between cycles:

image

- [x] The 'Submit' button is removed from the 'Review' page between cycles.
- [x] System spec to cover these changes.
- [x] Refactored and re-named a couple of methods on `EndOfCycleTimetable` to generalise them.

## Guidance to review

- Testing manually is tricky without resetting your system clock. We've discussed ways to make this easier for all the end-of-cycle features and this will be the subject of a separate PR.
- Are the tests sufficient?
- Do the renamings in `EndOfCycleTimetable` make sense?

## Link to Trello card

https://trello.com/c/XhLkMhQ9/1902-dev-🚲🔚-prevent-candidates-with-unsubmitted-applications-from-submitting-their-application

## Things to check

### DIFF
--- a/app/components/candidate_interface/apply_again_banner_component.rb
+++ b/app/components/candidate_interface/apply_again_banner_component.rb
@@ -12,7 +12,7 @@ module CandidateInterface
     end
 
     def render?
-      !EndOfCycleTimetable.show_apply_2_reopen_banner?
+      !EndOfCycleTimetable.between_cycles_apply_2?
     end
   end
 end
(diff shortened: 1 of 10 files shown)


stevehook:
> It's not within the scope of this card, but as far as I'm aware we still need to figure out how we're going to address all our tests failing once the apply1 deadline passes. Not being able to add course choices and submit the application is going to break the candidate helper. We probably need to discuss that sooner rather than later since the Apply 1 deadline is looming.

Yes agree we need to come up with a proper plan for that pretty quickly!

PR #2729, opened 2020-08-17
Use the degrees HESA abbreviation for the title on the review component

davidgisbey:
## Context

At the moment we the qualification_type is used in the title on the summary card components for degrees.

We should be using the abbreviation if one exists i.e. BA for Bachelor of Arts.

- Use the abbreviation for degree when one exists for the summary card title
- Update degree factory to use the HESA data

| Before | After | 
| ------------ | ------------ |
| image | image |

## Link to Trello card

https://trello.com/c/kpovALhr/1955-defect-use-degree-type-abbreviation-in-summary-card

## Things to check

### DIFF
--- a/app/components/candidate_interface/degrees_review_component.rb
+++ b/app/components/candidate_interface/degrees_review_component.rb
@@ -38,10 +38,14 @@ module CandidateInterface
 
     def degree_type_with_honours(degree)
       if degree.grade&.include? 'honours'
-        "#{degree.qualification_type} (Hons)"
+        "#{abbreviate_degree(degree.qualification_type)} (Hons)"
       else
-        degree.qualification_type
+        abbreviate_degree(degree.qualification_type)
       end
+    end
+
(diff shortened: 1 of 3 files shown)


app/components/candidate_interface/degrees_review_component.rb
```diff
@@ -38,12 +38,16 @@ def title(degree)
...
-        "#{degree.qualification_type} (Hons)"
+        "#{abbreviate_degree(degree.qualification_type)} (Hons)"
       else
-        degree.qualification_type
+        abbreviate_degree(degree.qualification_type)
       end
     end
 
+    def abbreviate_degree(name)
+      Hesa::DegreeType.find_by_name(name)&.abbreviation || name
```
stevehook, line 48: 👍 looks like there are a few HESA degree types without abbreviations but this handles that situation neatly

davidgisbey:
Good point about the test. I'll add one now.

PR #2730, opened 2020-08-17
Implement design for permissions not yet set by provider

steventux:
## Context

Ratifying providers may access permissions via the organisations pages. If permissions for these organisations have yet to be set up they should see an explanation as per the design.

See design story:
https://trello.com/c/pYlxvx3H/2488-design-for-when-accredited-body-user-logs-in-and-org-permissions-not-set-up-yet-by-training-provider

- Adds explanation about permissions setup as per design 
- Removes duplicate explanation from permissions list component

image

## Link to Trello card

https://trello.com/c/HE5czvGF/2625-implement-design-permissions-not-yet-set-by-provider

## Things to check

### DIFF
--- a/app/components/provider_relationship_permissions_list.html.erb
+++ b/app/components/provider_relationship_permissions_list.html.erb
@@ -28,13 +28,8 @@
   </ul>
 <% end %>
 
-<% if show_view_applications_only_section? %>
-  <% providers_that_can_only_view_applications = providers_that_can('view_applications_only') %>
+<% if show_view_applications_only_section? && permission_model.setup_at.present? %>
   <p class="govuk-!-margin-bottom-5 govuk-body">
-    <% if providers_that_can_only_view_applications.size > 1 %>
-      <%= training_provider.name %> has not set up permissions for this partnership yet.
-    <% else %>
-      <%= providers_that_can('view_applications_only').first.name %> can only view applications.
-    <% end %>
+    <%= providers_that_can('view_applications_only').first.name %> can only view applications.
   </p>
 <% end %>
--- a/app/views/provider_interface/organisations/show.html.erb
+++ b/app/views/provider_interface/organisations/show.html.erb
@@ -7,7 +7,13 @@
       </h2>
       <div class="app-banner govuk-!-margin-bottom-4">
         <div class="app-banner__message">
-          Contact <%= permissions.training_provider.name %> to change permissions.
+          <p>
+            <%= permissions.training_provider.name %> have not set up permissions yet - only they can set up permissions. Contact them to do this.
+          </p>
+
+          <p>
+            Staff at <%= permissions.ratifying_provider.name %> can still view applications in the meantime.
+          </p>
         </div>
       </div>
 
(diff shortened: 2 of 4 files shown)


app/components/provider_relationship_permissions_list.html.erb
```diff
@@ -17,14 +17,3 @@
...
   </ul>
 <% end %>
-
-<% if show_view_applications_only_section? %>
-  <% providers_that_can_only_view_applications = providers_that_can('view_applications_only') %>
-  <p class="govuk-!-margin-bottom-5 govuk-body">
-    <% if providers_that_can_only_view_applications.size > 1 %>
-      <%= training_provider.name %> has not set up permissions for this partnership yet.
-    <% else %>
-      <%= providers_that_can('view_applications_only').first.name %> can only view applications.
```
AgaDufrat, line 27: Do we want to remove this whole section? What if a training provider has set up the permission in such a way that only one provider in the relationship has all the power?

app/views/provider_interface/organisations/show.html.erb
```diff
@@ -7,7 +7,13 @@
       </h2>
       <div class="app-banner govuk-!-margin-bottom-4">
         <div class="app-banner__message">
-          Contact <%= permissions.training_provider.name %> to change permissions.
```
AgaDufrat, line 10: I probably misunderstood the card but I would think that we need a conditional here to check if permission has been set up.

```
if permissions.setup_at.nil?
  "<%= permissions.training_provider.name %> have not set up permissions yet..."
else
  "Contact <%= permissions.training_provider.name %> to change permissions."
 end
```

steventux: I think this is what I was unsure about and mentioned earlier, and I think you're right, there are 2 cases, one for when permissions have yet to be set and one when they've been set, thanks - will amend :+1:

app/components/provider_relationship_permissions_list.html.erb
steventux: I didn't quite understand the relevance of the `size` condition but I agree we want one of two messages depending on whether setup is required or not.
https://github.com/DFE-Digital/apply-for-teacher-training/pull/2730/commits/5ee272d15c3f8f7ef6f4bfb5c23e29afbb274291 should do this.

app/views/provider_interface/organisations/show.html.erb
steventux: https://github.com/DFE-Digital/apply-for-teacher-training/pull/2730/commits/5ee272d15c3f8f7ef6f4bfb5c23e29afbb274291 conditionally shows the different messages, it seems awkward that they are not in the same component though, will investigate bringing them into the same place.

app/components/provider_relationship_permissions_list.html.erb
AgaDufrat: I don't think we need to check the size.
The `at_least_one_active_permission_in_pair` validation on the model means that we can either have 1 or zero providers who can only view application and `show_view_applications_only_section?` already checks if there is a provider who can only see applications (`providers_that_can(:view_applications_only).any?`)

The whole thing LGTM.

AgaDufrat (approved):

PR #2736, opened 2020-08-18
Add support for testing EoC features

stevehook:
## Context

Because most of the features we are implementing to support the changeover from 2020 to 2021 recruitment cycles are date dependent there is no simple way to test them in `qa`, review apps or local development systems. At a minimum we need to setup up a system to test the features that are activated when the current cycle ends but before the new one begins.

- Add a `simulate_time_between_cycles` feature flag
- Use this feature flag to change the behaviour of `EndOfCycleTimetable` to return results consistent with the closed period between recruitment cycles regardless of the true date.
- Add a check to ensure that the new feature flag has no effect in production environments.

## Guidance to review

- This is meant to be the simplest thing that will work - to make the EoC features testable. There are better (more flexible) ways to do this such as storing the configuration for `EndOfCycleTimetable` in the database. In particular this implementation makes no distinction between Apply 1 & 2 applications (despite them having different deadlines in real life). Is this a reasonable first step?
- Do we need to cache the date values in `EndOfCycleTimetable` to avoid the repeated feature flag lookups (DB hits).

## Link to Trello card

https://trello.com/c/IWsNEUtT/1957-implement-a-way-to-test-eoc-features-in-qa-and-review-apps

## Things to check

### DIFF
--- a/spec/services/end_of_cycle_timetable_spec.rb
+++ b/spec/services/end_of_cycle_timetable_spec.rb
@@ -1,70 +1,144 @@
 require 'rails_helper'
 
 RSpec.describe EndOfCycleTimetable do
-  describe '.show_apply_1_deadline_banner?' do
-    it 'returns true before the configured date' do
-      Timecop.travel(Time.zone.local(2020, 8, 24, 23, 0, 0)) do
-        expect(EndOfCycleTimetable.show_apply_1_deadline_banner?).to be true
+  context 'when `simulate_time_between_cycles` feature flag is NOT active' do
+    describe '.show_apply_1_deadline_banner?' do
+      it 'returns true before the configured date' do
+        Timecop.travel(Time.zone.local(2020, 8, 24, 23, 0, 0)) do
+          expect(EndOfCycleTimetable.show_apply_1_deadline_banner?).to be true
+        end
+      end
+
+      it 'returns false after the configured date' do
+        Timecop.travel(Time.zone.local(2020, 8, 25, 1, 0, 0)) do
+          expect(EndOfCycleTimetable.show_apply_1_deadline_banner?).to be false
+        end
       end
     end
 
-    it 'returns false after the configured date' do
-      Timecop.travel(Time.zone.local(2020, 8, 25, 1, 0, 0)) do
-        expect(EndOfCycleTimetable.show_apply_1_deadline_banner?).to be false
+    describe '.show_apply_2_deadline_banner?' do
+      it 'returns true before the configured date' do
+        Timecop.travel(Time.zone.local(2020, 9, 18, 23, 0, 0)) do
+          expect(EndOfCycleTimetable.show_apply_2_deadline_banner?).to be true
+        end
+      end
+
+      it 'returns false after the configured date' do
+        Timecop.travel(Time.zone.local(2020, 9, 19, 1, 0, 0)) do
+          expect(EndOfCycleTimetable.show_apply_2_deadline_banner?).to be false
+        end
+      end
+    end
+
+    describe '.between_cycles_apply_1?' do
+      it 'returns false before the configured date' do
+        Timecop.travel(Time.zone.local(2020, 8, 24, 21, 0, 0)) do
+          expect(EndOfCycleTimetable.between_cycles_apply_1?).to be false
+        end
+      end
+
+      it 'returns true after the configured date' do
+        Timecop.travel(Time.zone.local(2020, 8, 25, 6, 0, 0)) do
+          expect(EndOfCycleTimetable.between_cycles_apply_1?).to be true
+        end
+      end
+
+      it 'returns false after the new cycle opens' do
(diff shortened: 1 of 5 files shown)


spec/services/end_of_cycle_timetable_spec.rb
```diff
@@ -1,70 +1,144 @@
...
 
-  describe '.show_apply_1_reopen_banner?' do
-    it 'returns false before the configured date' do
-      Timecop.travel(Time.zone.local(2020, 8, 24, 21, 0, 0)) do
-        expect(EndOfCycleTimetable.show_apply_1_reopen_banner?).to be false
+  context 'when `simulate_time_between_cycles` feature flag is active' do
+    before { FeatureFlag.activate(:simulate_time_between_cycles) }
+
+    describe '.show_apply_1_deadline_banner?' do
+      it 'returns true before the configured date' do
```
davidgisbey, line 78: should be false right?

spec/services/end_of_cycle_timetable_spec.rb
```diff
@@ -1,70 +1,144 @@
...
+          expect(EndOfCycleTimetable.show_apply_1_deadline_banner?).to be false
+        end
       end
     end
 
-    it 'returns false after the new cycle opens' do
-      Timecop.travel(Time.zone.local(2020, 10, 13, 12, 0, 0)) do
-        expect(EndOfCycleTimetable.show_apply_1_reopen_banner?).to be false
+    describe '.show_apply_2_deadline_banner?' do
+      it 'returns true before the configured date' do
```
davidgisbey, line 92: same here

spec/services/end_of_cycle_timetable_spec.rb
```diff
@@ -1,70 +1,144 @@
...
+          expect(EndOfCycleTimetable.show_apply_2_deadline_banner?).to be false
+        end
       end
     end
 
-    it 'returns true after the configured date' do
-      Timecop.travel(Time.zone.local(2020, 9, 19, 12, 0, 0)) do
-        expect(EndOfCycleTimetable.show_apply_2_reopen_banner?).to be true
+    describe '.show_apply_1_reopen_banner?' do
+      it 'returns false before the configured date' do
```
davidgisbey, line 106: and true here

spec/services/end_of_cycle_timetable_spec.rb
```diff
@@ -1,70 +1,144 @@
...
+        end
+      end
+
+      it 'returns true after the configured date' do
+        Timecop.travel(Time.zone.local(2020, 8, 25, 6, 0, 0)) do
+          expect(EndOfCycleTimetable.show_apply_1_reopen_banner?).to be true
+        end
+      end
+
+      it 'returns false after the new cycle opens' do
```
davidgisbey, line 118: and here

spec/services/end_of_cycle_timetable_spec.rb
```diff
@@ -1,70 +1,144 @@
...
+          expect(EndOfCycleTimetable.show_apply_1_reopen_banner?).to be true
+        end
       end
     end
 
-    it 'returns false after the new cycle opens' do
-      Timecop.travel(Time.zone.local(2020, 10, 13, 12, 0, 0)) do
-        expect(EndOfCycleTimetable.show_apply_2_reopen_banner?).to be false
+    describe '.show_apply_2_reopen_banner?' do
+      it 'returns false before the configured date' do
```
davidgisbey, line 126: and true here

spec/services/end_of_cycle_timetable_spec.rb
```diff
@@ -1,70 +1,144 @@
...
+        end
+      end
+
+      it 'returns true after the configured date' do
+        Timecop.travel(Time.zone.local(2020, 9, 19, 12, 0, 0)) do
+          expect(EndOfCycleTimetable.show_apply_2_reopen_banner?).to be true
+        end
+      end
+
+      it 'returns false after the new cycle opens' do
```
davidgisbey, line 138: and here. Sorry for the comment spam

davidgisbey:
The approach LGTM. Just a few minor comments on the test names :shipit:

malcolmbaig:
>     This is meant to be the simplest thing that will work - to make the EoC features testable. There are better (more flexible) ways to do this such as storing the configuration for EndOfCycleTimetable in the database. In particular this implementation makes no distinction between Apply 1 & 2 applications (despite them having different deadlines in real life). Is this a reasonable first step?

Yes, can refine and think about changing the storage mechanism for these dates if the need arises. :+1: 

> Do we need to cache the date values in EndOfCycleTimetable to avoid the repeated feature flag lookups (DB hits).

I think this is fine, as the way you've written the boolean check means that the DB lookup will only happen on the test environments.

malcolmbaig (approved):

tvararu (approved):
LGTM, tested on the review app 👍

Screenshot 2020-08-19 at 12 21 44

PR #2742, opened 2020-08-18
SPIKE change sorting to support the new applications Task View and its groupings

mnacos:
Not for merging!

## Context

In the proposed design (as per prototype), the Applications page contains a set of search results that are grouped based on discrete fixed criteria. The filters determine whether an application is included in the results set, and the group criteria determine which group an application belongs in.

Concerns were raised around the complexity of implementing the design so we need to validate prior to committing to implementation.
image

The more we can re-use existing code for filters and pagination, the better. The best way to achieve this is to perform all grouping and pagination in a single Postgresql query, which will also be good for performance and scalability.

To summarise, we need to:
- [x] show applications in groups
- [x] sort these groups using a custom order
- [x] have a defined header/title for each group
- [x] sort applications within groups by either updated_at or RBD (if awaiting provider decision)
- [x] restrict at least one of the groups to one cycle
- [x] paginate applications as before, across categories. This means a page may start and/or finish with an incomplete group.

Change the `SortApplicationChoices` service to split applications into groups. The criteria for these groups vary, but tend to rely on combinations of `status`, `reject_by_default_at` and `recruitment_cycle_year`.

In the controller, ActiveRecord combines queries for us in order to satisfy concerns such as authorisation, filtering, sorting and pagination. I've kept to this approach, but modified the sorting code to generate and use some virtual/computed columns in Postgresql: `task_view_group`, `pg_days_left_to_respond` and `recruitment_cycle_year`.

According to the current End-of-Cycle plans, provider users will be able to access applications from both the current and the previous cycle (e.g. they should be able to see applications with `pending_conditions` status from the previous year). Therefore, some groups will be restricted to `current_year`, others to `previous_year`, while some could show applications from both.

Unfortunately, while 
```ruby
ProviderInterface::SortApplicationChoices.call(
  application_choices: ApplicationChoice.all,
).first.task_view_group
```
returns a number,
```ruby
ProviderInterface::SortApplicationChoices.call(
  application_choices:  GetApplicationChoicesForProviders.call(providers: providers),
).first.task_view_group
```
throws an exception, because the resulting SQL is more complex and the virtual attributes are lost. This means we may have to add another presentation layer for the section/group headers, mirroring the criteria for groups such as `previous_cycle_pending_conditions` etc. Unless we find a way to preserve `task_view_group` in the combined queries, in which case we could do an enum-style mapping.

UPDATE: this is fixed now by making the sorting SELECT the root of the query tree.

This PR's approach seems to satisfy all requirements so far, ~~except specifying how headers/titles are maintained and displayed for each group, which requires further exploration~~.

## Guidance to review

Check the applications list loads and the number of applications shown is what you'd expect for your user. Then inspect the grouping, sorting and pagination. In this version of the code, the 'order by' dropdown is broken, as we'll probably end up removing it entirely or changing it completely. Also, the applications list page is not broken up into groups visibly, but records will appear sorted by a combination of status, RBD and updated_at within their groups. The order of these groups is also defined in the code.

I've also added a `TaskViewHelper` for defining and looking up the appropriate header titles for each `task_view_group`, in order to implement a proof-of-concept Task View applications list.

image

----
Filtering and searching are expected to work as before:
image

You can interact with the review app here: https://apply-for-te-2594-spike-6vrvxm.herokuapp.com/provider/applications?sort_by=days_left_to_respond

## Link to Trello card

https://trello.com/c/lf6NTGQW

## Things to check

### DIFF
--- a/app/services/provider_interface/sort_application_choices.rb
+++ b/app/services/provider_interface/sort_application_choices.rb
@@ -1,10 +1,10 @@
 module ProviderInterface
   class SortApplicationChoices
     def self.call(application_choices:)
-      with_task_view_group(application_choices).order(sort_order)
+      for_task_view(application_choices).order(sort_order)
     end
 
-    def self.with_task_view_group(application_choices)
+    def self.for_task_view(application_choices)
       application_choices.from <<~WITH_TASK_VIEW_GROUP.squish
         (
           SELECT *,
@@ -13,7 +13,7 @@ module ProviderInterface
               WHEN #{offered} THEN 2
               ELSE 999
             END AS task_view_group,
-            #{time_left_to_respond} AS time_left_to_respond
+            #{pg_days_left_to_respond} AS pg_days_left_to_respond
 
             FROM application_choices
         ) AS application_choices
@@ -37,21 +37,20 @@ module ProviderInterface
       OFFERED
     end
 
-    def self.time_left_to_respond
-      <<~TIME_LEFT_TO_RESPOND.squish
+    def self.pg_days_left_to_respond
+      <<~PG_DAYS_LEFT_TO_RESPOND.squish
         CASE
-          WHEN (DATE(reject_by_default_at) > '#{Time.zone.now.iso8601}')
-          THEN (DATE(reject_by_default_at) - '#{Time.zone.now.iso8601}')
-          ELSE NULL
-        END
-      TIME_LEFT_TO_RESPOND
+          WHEN (DATE(reject_by_default_at) > DATE('#{Time.zone.now.iso8601}'))
+          THEN (DATE(reject_by_default_at) - DATE('#{Time.zone.now.iso8601}'))
+          ELSE NULL END
+      PG_DAYS_LEFT_TO_RESPOND
     end
 
     def self.sort_order
       <<~ORDER_BY.squish
         task_view_group,
         status,
-        time_left_to_respond,
+        pg_days_left_to_respond,
         application_choices.updated_at
       ORDER_BY
     end
--- a/app/services/provider_interface/sort_application_choices.rb
+++ b/app/services/provider_interface/sort_application_choices.rb
@@ -1,24 +1,45 @@
 module ProviderInterface
   class SortApplicationChoices
-    def self.call(application_choices:, sort_by:)
-      application_choices.order(sort_order(sort_by))
+    def self.call(application_choices:)
+      with_task_view_group(application_choices).order(sort_order)
     end
 
-    def self.sort_order(sort_by)
-      return { updated_at: :desc } if sort_by != 'days_left_to_respond'
-
-      Arel.sql(
-        <<-ORDER_BY.strip_heredoc,
+    def self.with_task_view_group(application_choices)
+      application_choices.from <<~WITH_TASK_VIEW_GROUP.squish
         (
-          CASE
-            WHEN (status='awaiting_provider_decision' AND (DATE(reject_by_default_at) > '#{Time.zone.now.iso8601}')) THEN 1
-            ELSE 0
-          END
-        ) DESC,
-        reject_by_default_at ASC,
-        application_choices.updated_at DESC
-        ORDER_BY
-      )
+          SELECT *,
+            CASE
+              WHEN #{awaiting_provider_decision} THEN 1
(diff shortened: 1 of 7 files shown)


app/services/provider_interface/sort_application_choices.rb
```diff
@@ -1,24 +1,101 @@
...
+            AND c.recruitment_cycle_year = #{RecruitmentCycle.previous_year}
+        )
+      PREVIOUS_CYCLE_PENDING_CONDITIONS
+    end
+
+    def self.about_to_be_rejected_automatically
+      <<~AWAITING_PROVIDER_DECISION.squish
+        (
+          status='awaiting_provider_decision'
+            AND c.recruitment_cycle_year = #{RecruitmentCycle.current_year}
```
duncanjbrown, line 50: This makes sense bc everything is RBD'd before we start the new cycle, right?

Screenshot 2020-08-19 at 09 43 28

app/services/provider_interface/sort_application_choices.rb
```diff
@@ -1,24 +1,101 @@
...
+            )
+        )
+      AWAITING_PROVIDER_DECISION
+    end
+
+    def self.waiting_on_candidate
+      <<~WAITING_ON_CANDIDATE.squish
+        (
+          status='offer'
+            AND c.recruitment_cycle_year = #{RecruitmentCycle.current_year}
```
duncanjbrown, line 78: I _think_ the offer could have been made in the previous cycle and still require action here?

duncanjbrown (commented):
Great! Going to poke that exception you talked about in the description

duncanjbrown:
I think the exception happens bc ApplicationChoice.all hasn't actually run the query yet so we can still affect the SELECT. GetApplicationChoicesForProviders does run the query, I think. So at an efficiency cost, this seems to work:

```
choices = ProviderInterface::SortApplicationChoices.call(
  application_choices: ApplicationChoice.where(id: GetApplicationChoicesForProviders.call(providers: u.providers).map(&:id)),
)
```

steventux (commented):
It's a clean approach, I was expecting more SQL spaghetti than this. :+1:

I do wonder about maintainability and testing vs. performance with the SQL approach, I guess we have a fallback to paginate an array should be need to use Ruby to sort and group https://github.com/kaminari/kaminari#paginating-a-generic-array-object

mnacos:
> I think the exception happens bc ApplicationChoice.all hasn't actually run the query yet so we can still affect the SELECT. GetApplicationChoicesForProviders does run the query, I think. So at an efficiency cost, this seems to work:

Thanks for looking! I've followed your suggestion but removed the `pluck(:id)` and it seems like ActiveRecord is still only doing one query, while preserving `task_view_group` in the results. I'm guessing this is because `GetApplicationChoicesForProviders` is treated like a subquery and the sorting SELECT is now at the root of the query tree. :tada:  No efficiency cost

I've updated the code with some proof-of-concept headers in the applications list.

app/services/provider_interface/sort_application_choices.rb
mnacos: I guess! I think we can scrutinise this further when implementing TaskView for real.

app/services/provider_interface/sort_application_choices.rb
mnacos: I think you are right :) Some rigorous testing will be needed when we proceed to the implementation, especially around EoC edge cases.

mnacos:
> I do wonder about maintainability and testing vs. performance with the SQL approach, I guess we have a fallback to paginate an array should be need to use Ruby to sort and group https://github.com/kaminari/kaminari#paginating-a-generic-array-object

Thanks Steve. I've tried to keep this clean, esp. with maintainability in mind. As it stands, it's just one query that is using raw SQL (in order to use `CASE` within the sql, which we were already doing). I guess the new thing is defining the criteria for groups in SQL, but group definitions are not expected to change very much. ActiveRecord is doing the heavy lifting combining the queries, so with some good tests in place I'd be confident about mixing the SQL and pre-existing `.where` approaches. Unit testing should still be possible by exercising the `SortApplicationChoices` class.

mnacos:
Closing as the spike is complete. Will create another pull request for the production build.

PR #2744, opened 2020-08-18
Make magic links single-use and add confirmation page

malcolmbaig:
## Context
Magic links currently last for one hour, with no limit on their use
during that time. A recent pentest recommended that we make magic links
single-use.

This is a continuation of https://github.com/DFE-Digital/apply-for-teacher-training/pull/2695 .

- Change magic links so that they point to a sign in confirmation page.
- When the user clicks 'Continue' from this page, expire their token and
  sign them in.
- Amend the copy on the expired token page to reflect the new behaviour.

Screenshot 2020-08-18 at 20 32 01

## Guidance to review

- Any changes needed to content on the confirmation page?
- Per-commit review recommended
- May be easier to test multiple sign ins locally
- Do the spec changes make sense?

## Link to Trello card

https://trello.com/c/aNcIaroP
## Things to check

### DIFF
--- a/app/controllers/candidate_interface/sign_in_controller.rb
+++ b/app/controllers/candidate_interface/sign_in_controller.rb
@@ -14,7 +14,14 @@ module CandidateInterface
 
     def confirm_authentication
       candidate = retrieve_candidate_by_token || retrieve_candidate_by_user_id
-      redirect_to action: :new if candidate.nil?
+
+      redirect_to(action: :new) and return if candidate.nil?
+
+      if FindCandidateByToken.token_not_expired?(candidate)
+        render 'confirm_authentication'
+      else
+        redirect_to candidate_interface_expired_sign_in_path(u: params[:u])
+      end
     end
 
     def authenticate
--- a/app/controllers/candidate_interface/sign_in_controller.rb
+++ b/app/controllers/candidate_interface/sign_in_controller.rb
@@ -1,7 +1,7 @@
 module CandidateInterface
   class SignInController < CandidateInterfaceController
     skip_before_action :authenticate_candidate!
-    before_action :redirect_to_application_if_signed_in, except: %i[authenticate]
+    before_action :redirect_to_application_if_signed_in, except: %i[confirm_authentication authenticate]
 
     def new
       candidate = Candidate.new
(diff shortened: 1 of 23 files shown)


spec/mailers/candidate_mailer_spec.rb
```diff
@@ -73,7 +73,7 @@
         'Your application is being considered',
         'heading' => 'Dear Bob',
         'working days the provider has to respond' => '10 working days',
-        'magic link to authenticate' => 'http://localhost:3000/candidate/authenticate?token=raw_token&u=encrypted_id'
```
stevehook, line 76: One consequence of this is that candidates who receive a magic link just before we deploy this and try to use it after will get a 404. This is a bit confusing. Would it make sense to redirect the (legacy) `authenticate` route to `/sign-in/expired`?

app/controllers/candidate_interface/sign_in_controller.rb
```diff
@@ -12,24 +12,27 @@ def create
...
-      elsif token_not_expired
+      redirect_to(action: :new) and return if candidate.nil?
+
+      if token_not_expired
         flash[:success] = t('apply_from_find.account_created_message') if candidate.last_signed_in_at.nil?
         sign_in(candidate, scope: :candidate)
         add_identity_to_log candidate.id
-        candidate.update!(last_signed_in_at: Time.zone.now)
-
-        redirect_to candidate_interface_interstitial_path(providerCode: params[:providerCode], courseCode: params[:courseCode])
```
stevehook, line 32: Is there another route that accepts these params?

app/models/candidate.rb
```diff
@@ -36,6 +36,14 @@ def refresh_magic_link_token!
     magic_link_token.raw
   end
 
+  def update_sign_in_fields!
+    update!(
+      magic_link_token: nil,
+      magic_link_token_sent_at: nil,
+      last_signed_in_at: Time.zone.now,
+    )
+  end
```
stevehook, line 45: :+1:

stevehook (commented):
Tested locally and was able to log in once and only once using the magic link. The only odd thing I saw has that if I am logged in and then try to use an already used magic link I get the 'Confirm sign in' page and this takes me to the application dashboard. Would it be possible to redirect straight to the application dashboard skipping the confirm sign in page?

app/controllers/candidate_interface/sign_in_controller.rb
malcolmbaig: Appears to be handled by both SignUpController and ApplyFromFindController.

See:
https://github.com/DFE-Digital/apply-for-teacher-training/blob/1936-spike-auth-after-button-press/spec/system/candidate_interface/course_selection/candidate_new_user_with_course_params_spec.rb
https://github.com/DFE-Digital/apply-for-teacher-training/blob/1936-spike-auth-after-button-press/spec/system/candidate_interface/course_selection/candidate_existing_user_with_course_params_spec.rb
https://github.com/DFE-Digital/apply-for-teacher-training/blob/1936-spike-auth-after-button-press/spec/system/candidate_interface/course_selection/candidate_signed_in_user_with_course_params_spec.rb

spec/mailers/candidate_mailer_spec.rb
malcolmbaig: Good idea, will add

malcolmbaig:
> Would it be possible to redirect straight to the application dashboard skipping the confirm sign in page?

Yep, that seems sensible. Will update.

tvararu:
If the link is expired, can we present the user immediately with the "The link you clicked has expired" page?

Screen Recording 2020-08-19 at 16 48 31

tvararu (approved):
Code looks really clean! I just realised we're not using a `has_many` relationship for magic links for the candidate model. I'll try to prioritise that in the next sprint.

tvararu (approved):

PR #2747, opened 2020-08-19
Force all course codes to be unique when generating fake data

duncanjbrown:
Collisions are rare but possible, and they are annoying to deal with when they happen (failing specs, half-baked test envs)

## Context

Tests flaked https://dfe-ssp.visualstudio.com/Become-A-Teacher/_build/results?buildId=103402&view=ms.vss-test-web.build-test-results-tab&runId=1367106&resultId=100700&paneView=debug

## Link to Trello card

https://trello.com/c/sd3pw4Lc/2636-two-flaky-tests

## Things to check

### DIFF
--- a/app/services/generate_fake_provider.rb
+++ b/app/services/generate_fake_provider.rb
@@ -9,13 +9,13 @@ class GenerateFakeProvider
   end
 
   def self.generate_courses_for(training_provider)
-    10.times do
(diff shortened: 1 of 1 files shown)


mnacos (approved):
Makes sense :+1:

tvararu (approved):

PR #2749, opened 2020-08-19
Add DSA success page with contextual content

steventux:
## Context

When a user who needs to set up organisation permissions agrees to all DSAs they are currently taken straight to the permissions wizard first step. In order to make the journey less surprising we're adding a bridging page which explains next steps.

Adds a page which acts as a mid-step when a user has accepted all DSAs and needs to then set up permissions for their organisation.
If the user does not have permissions to set up org permissions they see a link which can take them to their applications.
In this scenario they may also see content about managing users if they have the relevant permissions.

### User has org permissions to set up

image

### User has no org permissions to set up

image

## Guidance to review

Reviewing is quite tricky given the preconditions needed on the provider and permissions data.

To see the permissions based DSA success page you'll need to: 

- Add a new provider for your provider user (or delete an existing DSA)
- Find or create a ProviderRelationshipPermissions record where the same provider is the training provider, set the `setup_at` timestamp to `nil`
- Log in and you should have to agree to the DSA then you should see the permissions-based success page.

To see the DSA success page without the permissions flow content:

- Add a new provider for your provider user (or delete an existing DSA)
- Log in and you should have to agree to the DSA then you should see the confirmation success page.

## Link to Trello card

https://trello.com/c/pMFlaOle/2644-build-interstitial-page-for-end-of-dsa-start-of-org-permissions-wizard

## Things to check

### DIFF
--- a/app/controllers/provider_interface/provider_agreements_controller.rb
+++ b/app/controllers/provider_interface/provider_agreements_controller.rb
@@ -12,11 +12,15 @@ module ProviderInterface
 
     def create_data_sharing_agreement
       @provider_agreement = ProviderAgreement.new(provider_agreement_params.merge(provider_user: current_provider_user))
-      if @provider_agreement.save
-        redirect_to provider_interface_path
-      else
-        render :data_sharing_agreement
-      end
+
+      render :data_sharing_agreement and return unless @provider_agreement.save
+
+      @provider_setup = ProviderSetup.new(provider_user: current_provider_user)
+      @provider_relationship_pending = @provider_setup.next_relationship_pending.present?
+
+      render :success and return if @provider_setup.next_agreement_pending.blank?
+
+      redirect_to provider_interface_new_data_sharing_agreement_path
     end
 
   private
--- a/app/views/provider_interface/provider_agreements/success.html.erb
+++ b/app/views/provider_interface/provider_agreements/success.html.erb
@@ -2,7 +2,7 @@
   <div class="govuk-grid-column-two-thirds">
     <div class="govuk-panel govuk-panel--confirmation">
       <h1 class="govuk-panel__title">
-        Thank you for signing up to our data sharing agreement
+        You've successfully signed the data sharing agreement
       </h1>
     </div>
     <% if @provider_relationship_pending %>
--- a/app/views/provider_interface/provider_agreements/success.html.erb
+++ b/app/views/provider_interface/provider_agreements/success.html.erb
@@ -1,0 +1,41 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <div class="govuk-panel govuk-panel--confirmation">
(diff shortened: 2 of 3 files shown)


app/controllers/provider_interface/provider_agreements_controller.rb
```diff
@@ -12,11 +12,16 @@ def new_data_sharing_agreement
 
     def create_data_sharing_agreement
       @provider_agreement = ProviderAgreement.new(provider_agreement_params.merge(provider_user: current_provider_user))
-      if @provider_agreement.save
-        redirect_to provider_interface_path
-      else
-        render :data_sharing_agreement
-      end
+
+      render :data_sharing_agreement and return unless @provider_agreement.save
```
mnacos, line 16: Somehow I would find a nested `if` easier to follow here. But don't feel strongly about it. :)

```ruby
  if @provider_agreement.save
    @provider_setup = ProviderSetup.new(provider_user: current_provider_user)

    if @provider_setup.next_agreement_pending
      redirect_to provider_interface_new_data_sharing_agreement_path
    else
      @provider_relationship = @provider_setup.next_relationship_pending
      render :success
    end
  else
    render :data_sharing_agreement
  end
```

app/controllers/provider_interface/provider_agreements_controller.rb
```diff
@@ -12,11 +12,16 @@ def new_data_sharing_agreement
...
+
+      @provider_setup = ProviderSetup.new(provider_user: current_provider_user)
+      @provider_relationship = @provider_setup.next_relationship_pending
+      @provider_agreement = @provider_setup.next_agreement_pending
```
mnacos, line 20: Is `@provider_agreement` needed? `:success` will render only `@provider_agreement` is blank. The redirect will initialise its own variables.

mnacos (commented):
Looks good. Added some minor comments related to the routing code.

app/controllers/provider_interface/provider_agreements_controller.rb
steventux: I think maybe there's an issue with having 3 differing rendering/redirecting outcomes from a single action and that makes for a lot of conditions. I went with a more guard-clause-like format for brevity.

app/controllers/provider_interface/provider_agreements_controller.rb
steventux: :+1: No you're right, this is redundant, I had just been rendering the `:new` template but changed it to a redirect, this is a relic of that earlier approach, will remove

steventux: @mnacos removed and rebased here https://github.com/DFE-Digital/apply-for-teacher-training/pull/2749/commits/7ffe04852222acbd661ba6aea50b3ac373dc9287

mnacos (approved):
:+1:

PR #2753, opened 2020-08-20
Generate test applications with correct equality and diversity data

AgaDufrat:
## Context

How we generate `equality_and_diversity` on application form is different to what it is in production.

1. Update the factory trait:
- Remove `other_disability` and `disability_status` keys as they are not saved
  in the database. When other_disability is declared it gets added to
`disabilities` array. When disability_status is 'no', `disabilities` is an
empty array, when it’s 'Prefer not to say’, this string gets added to
disabilities.
- Use long names of disabilities as they are saved in the database
- When sampling disability array allow it to be empty which represents that
  user answered "No"
- Add 'Prefer not to say' as a disability option

2. Update the worked to generate test applications with equality and diversity data.

## Guidance to review

Generate some applications

Note: I decided not to generate other_disability so it's easier to test. We are not doing it for safeguarding issues either.

## Link to Trello card

https://trello.com/c/vRibYMne/2643-fix-how-we-generate-equalityanddiversity-data

## Things to check

### DIFF
--- a/app/lib/test_applications.rb
+++ b/app/lib/test_applications.rb
@@ -57,15 +57,14 @@ class TestApplications
     end
 
     Audited.audit_class.as_user(candidate) do
-      traits = %i[
-        with_safeguarding_issues_disclosed
-        with_no_safeguarding_issues_to_declare
-        with_safeguarding_issues_never_asked
-      ].sample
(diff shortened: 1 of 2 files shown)


tvararu (approved):
👍

PR #2754, opened 2020-08-20
1897 back links on edit pages to redirect to review page

george-bit:
## Context

Alteration to the back links in edit views to route them back to the section review screen as opposed to the application review screen.

I have altered the path on each view listed in the trello card. These views and routes are not unilateral in their file structure, but I have predominantly replaced "Back to Application" links with straightforward back links that route back to the previous page. I feel this means that the views become relevant to both possible scenarios of initial application and editing an application.

image

## Guidance to review

Review whether this is a viable solution in keeping with convention. Attempt to access each view from the review screen and check that the links direct back to the correct place, also check that initial application process is not adversely affected by these changes.

## Link to Trello card

https://trello.com/c/n7A5psjB/1897-%F0%9F%94%99-when-editing-information-back-links-should-link-to-review-page

## Things to check


app/views/candidate_interface/contact_details/address/edit.html.erb
```diff
@@ -1,8 +1,8 @@
 <% content_for :title, title_with_error_prefix(t('page_titles.address'), @contact_details_form.errors.any?) %>
 <% if FeatureFlag.active?(:international_addresses) %>
-  <% content_for :before_content, govuk_back_link_to(candidate_interface_contact_details_edit_address_type_path) %>
+  <% content_for :before_content, govuk_back_link_to(:back) %>
```
tvararu, line 3: `:back` uses the referrer and where it's not available, it uses JavaScript to go back in the user's history: https://apidock.com/rails/ActionView/Helpers/UrlHelper/link_to

The design system has some guidance regarding this (emphasis added) https://design-system.service.gov.uk/components/back-link/:

> Make sure the link takes users to the previous page they were on, in the state they last saw it. Where possible, **ensure it works even when JavaScript is not available.**
> 
> If this is not possible, **you should hide the back link when JavaScript is not available.**

So... we can't use `:back` in this way, because it doesn't work without JavaScript and where the browser history queue is empty (e.g. you open the page in a new window/tab, the button doesn't do anything).

tvararu (changes requested):
This solution won't work unfortunately without JS, and even with JS there are some situations where it could break.

I've a few ideas. 💡 

Firstly, I think `govuk_back_link_to` should default to `:back` as the URL when nothing is passed in. It'll make it slightly easier to use.

Secondly, we should make the link meet the behaviour of the design system. Something like:

- If `@controller.request.env["HTTP_REFERER"]` is available, use that
- Else, use 'javascript:history.back()', but add a CSS class that makes it `display: none` by default
- Once JavaScript kicks in, check if the JavaScript `history.length` is `> 0`; if yes, unhide the link

app/views/candidate_interface/contact_details/address/edit.html.erb
```diff
@@ -1,8 +1,8 @@
...
 <% else %>
-  <% content_for :before_content, govuk_back_link_to(candidate_interface_contact_details_edit_base_path) %>
+  <% content_for :before_content, govuk_back_link_to(:back) %>
```
tvararu, line 5:
```suggestion
  <% content_for :before_content, govuk_back_link_to %>
```

PR #2755, opened 2020-08-20
Reject applications that are awaiting references at the end of the cycle

davidgisbey:
## Context

When the apply 2 deadline passes. All applications that are awaiting references should be rejected.

A new state 'rejected_at_end_of_cycle' is used (as discussed yesterday) and all the documentation has been updated.

- Add a worker that runs between apply2 closing and the next cycle launches, that collects awaiting_references applications and rejects them
- Updated all the relevant docs

## Guidance to review

I'm not 100% sure on whether this new state should be exposed in the vendor API. I don't see why not? Might be worth checking with Provendor though.

## Link to Trello card

https://trello.com/c/hX5tD7PX/1959-dev-%F0%9F%9A%B2%F0%9F%94%9A-reject-all-submitted-applications-including-apply-again-as-incomplete-which-havent-been-sent-to-providers-at-the-end-of-t

## Things to check


tvararu:
Re:

> I'm not 100% sure on whether this new state should be exposed in the vendor API. I don't see why not? Might be worth checking with Provendor though.

I think it should, @duncanjbrown would this be possible?

config/clock.rb
```diff
@@ -21,6 +21,8 @@ class Clock
   every(1.hour, 'SendChaseEmailToCandidates', at: '**:40') { SendChaseEmailToCandidatesWorker.perform_async }
   every(1.hour, 'SendCourseFullNotifications', at: '**:45') { SendCourseFullNotificationsWorker.perform_async }
 
+  every(1.hour, 'RejectAwaitingReferencesCourseChoices', at: '**:50') { RejectAwaitingReferencesCourseChoicesWorker.perform_async }
```
davidgisbey, line 24: This should really only need to run once. Maybe every(1.month, .....

app/services/candidate_interface/get_previous_cycles_awaiting_references_course_choices.rb
```diff
@@ -0,0 +1,9 @@
+module CandidateInterface
```
tvararu, line 1: Do the services in this PR need to live in the `CandidateInterface` namespace?

config/locales/application_states.yml
```diff
@@ -94,6 +96,10 @@ en:
         - referee_mailer-reference_request_chaser_email
         - referee_mailer-reference_request_chase_again_email
 
+    rejected_at_end_of_cycle:
+      name: Reject at the end of the cycle
+      description: Application choices that are awaiting references at are automatically rejectd at the end of the cycle
```
tvararu, line 101:
```suggestion
      description: Application choices that are awaiting references are automatically rejected at the end of the cycle
```

config/locales/application_states.yml
```diff
@@ -219,6 +225,11 @@ en:
       by: candidate
       description: If the application has not been sent to the provider, the candidate can still cancel the application.
 
+    awaiting_references-reject_at_end_of_cycle:
+      name: Awaiting references applications rejected
+      by: system
+      description: Application choices that are awaiting references at are automatically rejectd at the end of the cycle
```
tvararu, line 231:
```suggestion
      description: Application choices that are awaiting references are automatically rejected at the end of the cycle
```

spec/services/candidate_interface/reject_awaiting_references_application_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'rails_helper'
+
+RSpec.describe CandidateInterface::RejectAwaitingReferencesApplication do
+  describe '#call' do
+    let(:application_choice) { create(:awaiting_references_application_choice) }
+
+    it 'sets the course_is_on_apply and course_on_find attributes to true' do
```
tvararu, line 7:
```suggestion
    it 'rejects an application at the end of the cycle' do
```

app/workers/reject_awaiting_references_course_choices_worker.rb
```diff
@@ -0,0 +1,9 @@
+class RejectAwaitingReferencesCourseChoicesWorker
+  include Sidekiq::Worker
+
+  def self.perform
+    CandidateInterface::GetPreviousCyclesAwaitingReferencesCourseChoices.call&.each do |application_choice|
```
tvararu, line 5: Why do we need the `&`?

```suggestion
    CandidateInterface::GetPreviousCyclesAwaitingReferencesCourseChoices.call.each do |application_choice|
```

tvararu (approved):

config/clock.rb
tvararu: Could it miss the window when it's meant to run if we run it only once a month?

PR #2756, opened 2020-08-20
Show the selected course option's study mode in the API response

duncanjbrown:
We accidentally showed the available study modes for the course, which meant that if you e.g. offered a part time place on a course that also supported full time, you would see 'full_or_part_time' in this field instead of 'part_time'.

## Context

Noticed while investigating an API issue for UCB (https://trello.com/c/3iC0aiNg/2648-investigate-bug-report-from-ucb)

Show the correct data, which is `course_option.study_mode`, not `course.study_mode`.

## Things to check


duncanjbrown:
> 👍 We may need to update the date on release note before we merge.

Amended to August 24th

PR #2759, opened 2020-08-21
Carry over unsubmitted applications automatically

stevehook:
## Context

Applications that are unsubmitted when a recruitment cycle ends should be carried over to the new recruitment. Unlike unsuccessful applications the carry-over process for unsubmitted applications should be automatic.

- [x] Add `CarryOverUnsubmittedApplicationsWorker` that searches for applications from the last cycle and automatically carries them across using the existing `CarryOverApplication` service.
- [x] Allow unrequested references to be carried across (was just completed references).

## Guidance to review

- It's not clear how `CarryOverUnsubmittedApplicationsWorker` should be invoked (or exactly when). I am assuming we trigger it with `clock` and it runs once as a Sidekiq job.
- I think there was some discussion about whether duplicating unsubmitted applications was necessary to move them into the new recruitment cycle. I've gone with this for now because it's consistent with other carry-overs and it allows us to retain the original course choices but this is open to debate.

## Link to Trello card

## Things to check

### DIFF
--- a/app/workers/carry_over_unsubmitted_applications_worker.rb
+++ b/app/workers/carry_over_unsubmitted_applications_worker.rb
@@ -1,0 +1,17 @@
+class CarryOverUnsubmittedApplicationsWorker
+  include Sidekiq::Worker
+
+  def perform
+    unsubmitted_applications_from_earlier_cycle.each do |application_form|
+      CarryOverApplication.new(application_form).call
+    end
+  end
+
+private
+
+  def unsubmitted_applications_from_earlier_cycle
+    # TODO: Remove hard-coded year
+    # TODO: Omit applications that have already been carried over (have a subsequent_application_form)
+    ApplicationForm.joins(application_choices: :course).where(submitted_at: nil).where('courses.recruitment_cycle_year' => 2020).distinct
+  end
+end
--- a/app/workers/carry_over_unsubmitted_applications_worker.rb
+++ b/app/workers/carry_over_unsubmitted_applications_worker.rb
@@ -11,7 +11,14 @@ private
 
   def unsubmitted_applications_from_earlier_cycle
     # TODO: Remove hard-coded year
-    # TODO: Omit applications that have already been carried over (have a subsequent_application_form)
-    ApplicationForm.joins(application_choices: :course).where(submitted_at: nil).where('courses.recruitment_cycle_year' => 2020).distinct
+    ApplicationForm
+      .joins(application_choices: :course)
+      .where(submitted_at: nil)
+      .where('courses.recruitment_cycle_year' => 2020)
+      .where(
+        'application_forms.id NOT IN (:duplicated_applications)',
+        duplicated_applications: ApplicationForm.where.not(previous_application_form_id: nil).select(:previous_application_form_id),
+      )
+      .distinct
   end
 end
--- a/app/workers/carry_over_unsubmitted_applications_worker.rb
+++ b/app/workers/carry_over_unsubmitted_applications_worker.rb
@@ -10,11 +10,10 @@ class CarryOverUnsubmittedApplicationsWorker
 private
 
   def unsubmitted_applications_from_earlier_cycle
-    # TODO: Remove hard-coded year
     ApplicationForm
       .joins(application_choices: :course)
       .where(submitted_at: nil)
-      .where('courses.recruitment_cycle_year' => RecruitmentCycle.current_year)
+      .where('courses.recruitment_cycle_year' => RecruitmentCycle.current_year - 1)
       .where(
         'application_forms.id NOT IN (:duplicated_applications)',
         duplicated_applications: ApplicationForm.where.not(previous_application_form_id: nil).select(:previous_application_form_id),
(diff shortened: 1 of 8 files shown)


spec/system/candidate_interface/automatically_carry_over_unsubmitted_application_to_new_cycle_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+    create(
+      :application_choice,
+      status: :unsubmitted,
+      application_form: @application_form,
+    )
+    @unrequested_references = create_list(:reference, 2, feedback_status: :not_requested_yet, application_form: @application_form)
+  end
+
+  def and_the_recruitment_cycle_ends
+    Timecop.safe_mode = false
```
stevehook, line 56: Is this reasonable given that there is an `around` block wrapping the whole example that does have a safe mode-compliant block call?

app/workers/carry_over_unsubmitted_applications_worker.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+private
+
+  def unsubmitted_applications_from_earlier_cycle
+    # TODO: Remove hard-coded year
+    ApplicationForm
+      .joins(application_choices: :course)
+      .where(submitted_at: nil)
+      .where('courses.recruitment_cycle_year' => 2020)
```
mnacos, line 17:
```suggestion
      .where('courses.recruitment_cycle_year' => RecruitmentCycle.current_year)
```

spec/system/candidate_interface/automatically_carry_over_unsubmitted_application_to_new_cycle_spec.rb
```diff
@@ -0,0 +1,136 @@
+require 'rails_helper'
+
+RSpec.feature 'Automatically carry over unsubmitted applications' do
+  include CandidateHelper
+
+  around do |example|
+    Timecop.freeze(Date.new(2020, 8, 1)) do
```
mnacos, line 7:
```suggestion
    Timecop.freeze(Date.new(RecruitmentCycle.current_year, 8, 1)) do
```

spec/system/candidate_interface/automatically_carry_over_unsubmitted_application_to_new_cycle_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+      :application_choice,
+      status: :unsubmitted,
+      application_form: @application_form,
+    )
+    @unrequested_references = create_list(:reference, 2, feedback_status: :not_requested_yet, application_form: @application_form)
+  end
+
+  def and_the_recruitment_cycle_ends
+    Timecop.safe_mode = false
+    Timecop.travel(Time.zone.local(2020, 10, 15, 12, 0, 0))
```
mnacos, line 57:
```suggestion
    Timecop.travel(Time.zone.local(RecruitmentCycle.current_year, 10, 15, 12, 0, 0))
```

mnacos (commented):
Quick comment to request the use of `RecruitmentCycle.current_year`, since I'm working on similar things but from the perspective of what happens after EoC.

app/workers/carry_over_unsubmitted_applications_worker.rb
```diff
@@ -0,0 +1,24 @@
...
+  def perform
+    unsubmitted_applications_from_earlier_cycle.each do |application_form|
+      CarryOverApplication.new(application_form).call
+    end
+  end
+
+private
+
+  def unsubmitted_applications_from_earlier_cycle
+    # TODO: Remove hard-coded year
```
tvararu, line 13:
```suggestion
```

spec/system/candidate_interface/automatically_carry_over_unsubmitted_application_to_new_cycle_spec.rb
tvararu: 👍

tvararu (approved):
👍

spec/system/candidate_interface/automatically_carry_over_unsubmitted_application_to_new_cycle_spec.rb
stevehook: I'm going to keep this hard-coded date in the test because it needs to be relative to the dates defined in `EndOfCycleTimetable`.

app/workers/carry_over_unsubmitted_applications_worker.rb
stevehook: :+1: this needs to be `RecruitmentCycle.current_year - 1` so that it picks up applications from the previous cycle.

spec/system/candidate_interface/automatically_carry_over_unsubmitted_application_to_new_cycle_spec.rb
stevehook: I have removed the hard-coded date in the implementation and changed `EndOfCycleTimetable` to use `RecruitmentCycle`.

spec/system/candidate_interface/automatically_carry_over_unsubmitted_application_to_new_cycle_spec.rb
stevehook: as above

app/workers/carry_over_unsubmitted_applications_worker.rb
stevehook: Removed as part of change following Michael's comment

PR #2760, opened 2020-08-21
Rough sketch of 'Cancel applications to full courses after 7th Sep'

malcolmbaig:
## Context

From the card:

As a: DfE Apply
I need to: stop applications from being sent to providers of full courses after 7th September and reject as course full
So that: candidates are aware of what the status of their application is at the end of cycle and why the application was rejected

- Add 7th Sep to the EndOfCycleTimetable and expose a helper that checks if we're between that date and the opening of the new cycle.
- When a reference is provided, if the helper returns true and the course choice is full, cancel the choice and send some kind of email.

## Guidance to review

This PR illustrates a case-by-case approach that gets invoked on every reference submission. The alternative would be to batch cancel apps at the end of the 7th, but that wouldn't catch courses that become full somewhere between the 7th and the opening of the new cycle.

### Questions

- So we're clear, should we be checking that the chosen course option (study mode at a site) is full, or the whole course?
- What about withdrawn courses? Should we handle them the same way?
- What about courses that aren't full? Should they still go to providers after 7th Sep?

(these 3 addressed here: https://ukgovernmentdfe.slack.com/archives/CP18YJXPY/p1598263053032300)

- It looks like we're tracking content design of the relevant emails on this card https://trello.com/c/gaGLhi6K/1976-eoc-email-content . 
  - Am I correct in thinking that emails 2 and 3 are the ones that we'd send in the scenario outlined in this PR? cc: @EmmaFrith 
  - Email 1 looks like a separate thing that we'd need to implement.

## Link to Trello card
https://trello.com/c/iExqduky/1834-tech-spike-prevent-applications-to-full-courses-from-going-to-provider-after-7th-september

## Things to check

### DIFF
--- a/app/services/submit_reference.rb
+++ b/app/services/submit_reference.rb
@@ -37,14 +37,18 @@ private
     ActiveRecord::Base.transaction do
       reference_feedback_provided!
       application_form.application_choices.awaiting_references.each do |application_choice|
-        ApplicationStateChange.new(application_choice).references_complete!
+        if EndOfCycleTimetable.stop_sending_apply_1_to_providers? && application_choice.course_option_full?
+          EndOfCycle::CancelApplicationToFullCourse.new(application_choice).call
+        else
+          ApplicationStateChange.new(application_choice).references_complete!
 
-        next unless application_form.candidate_has_previously_applied?
+          next unless application_form.candidate_has_previously_applied?
 
-        # If the candidate has previously applied, they have less of a need to
-        # edit the application. Hence, we clear out the usual 7-day edit window
-        # by resetting the `edit_by` time.
-        application_form.update!(edit_by: Time.zone.now)
+          # If the candidate has previously applied, they have less of a need to
+          # edit the application. Hence, we clear out the usual 7-day edit window
+          # by resetting the `edit_by` time.
+          application_form.update!(edit_by: Time.zone.now)
+        end
       end
     end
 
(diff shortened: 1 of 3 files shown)


app/services/submit_reference.rb
```diff
@@ -37,14 +37,18 @@ def progress_applications
...
-        ApplicationStateChange.new(application_choice).references_complete!
-
-        next unless application_form.candidate_has_previously_applied?
-
-        # If the candidate has previously applied, they have less of a need to
-        # edit the application. Hence, we clear out the usual 7-day edit window
-        # by resetting the `edit_by` time.
-        application_form.update!(edit_by: Time.zone.now)
+        if EndOfCycleTimetable.stop_sending_apply_1_to_providers? && application_choice.course_option_full?
+          EndOfCycle::CancelApplicationToFullCourse.new(application_choice).call
```
malcolmbaig, line 41: @tvararu @stevehook on second thought, I think this logical check and the invocation of `CancelApplicationToFullCourse` should sit here instead https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/services/send_application_to_provider.rb#L10 , to account for the edit_by period.

WDYT?

tvararu: 👍

malcolmbaig:
Closing spike, tracking the actual implementation here: https://trello.com/c/tGI5SMcV

PR #2763, opened 2020-08-21
Reject applications that are awaiting references at the end of the cycle

tvararu:
Originally https://github.com/DFE-Digital/apply-for-teacher-training/pull/2755, but it broke QA.

Original PR context:

## Context

When the apply 2 deadline passes. All applications that are awaiting references should be rejected.

A new state 'rejected_at_end_of_cycle' is used (as discussed yesterday) and all the documentation has been updated.

- Add a worker that runs between apply2 closing and the next cycle launches, that collects awaiting_references applications and rejects them
- Updated all the relevant docs

## Guidance to review

I'm not 100% sure on whether this new state should be exposed in the vendor API. I don't see why not? Might be worth checking with Provendor though.

## Link to Trello card

https://trello.com/c/hX5tD7PX/1959-dev-%F0%9F%9A%B2%F0%9F%94%9A-reject-all-submitted-applications-including-apply-again-as-incomplete-which-havent-been-sent-to-providers-at-the-end-of-t

## Things to check

### DIFF
--- a/app/services/candidate_interface/reject_awaiting_references_application.rb
+++ b/app/services/candidate_interface/reject_awaiting_references_application.rb
@@ -1,0 +1,13 @@
+module CandidateInterface
+  class RejectAwaitingReferencesApplication
+    def self.call(application_choice)
+      ActiveRecord::Base.transaction do
+        ApplicationStateChange.new(application_choice).reject_at_end_of_cycle!
+        application_choice.update!(
+          rejection_reason: 'Awaiting references when the recruitment cycle closed.',
+          rejected_at: Time.zone.now,
+        )
+      end
+    end
+  end
+end
--- a/app/workers/reject_awaiting_references_course_choices_worker.rb
+++ b/app/workers/reject_awaiting_references_course_choices_worker.rb
@@ -1,7 +1,7 @@
 class RejectAwaitingReferencesCourseChoicesWorker
   include Sidekiq::Worker
 
-  def self.perform
+  def perform
     CandidateInterface::GetPreviousCyclesAwaitingReferencesCourseChoices.call.each do |application_choice|
       CandidateInterface::RejectAwaitingReferencesApplication.call(application_choice)
     end
--- a/app/workers/reject_awaiting_references_course_choices_worker.rb
+++ b/app/workers/reject_awaiting_references_course_choices_worker.rb
@@ -1,0 +1,9 @@
+class RejectAwaitingReferencesCourseChoicesWorker
+  include Sidekiq::Worker
+
+  def self.perform
+    CandidateInterface::GetPreviousCyclesAwaitingReferencesCourseChoices.call.each do |application_choice|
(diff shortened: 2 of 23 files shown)


app/workers/reject_awaiting_references_course_choices_worker.rb
```diff
@@ -0,0 +1,9 @@
+class RejectAwaitingReferencesCourseChoicesWorker
+  include Sidekiq::Worker
+
+  def self.perform
+    CandidateInterface::GetPreviousCyclesAwaitingReferencesCourseChoices.call&.each do |application_choice|
```
duncanjbrown, line 5: I think no need for `&` as it returns `[]` or an empty relation when there's nothing to do?

config/vendor-api-v1.yml
```diff
@@ -394,6 +394,7 @@ components:
           - pending_conditions
           - recruited
           - rejected
+          - rejected_at_end_of_cycle
```
duncanjbrown, line 397: I don't think we can add this here at this stage unfortunately, as it could be a breaking change. I think we ought to map to `rejected` for now and consider adding a boolean `rejected_at_end_of_cycle` field if this information turned out to be useful to vendors. Adding a card for this status in our draft API v2 feature list https://trello.com/c/4PcIvMCC/704-add-rejectedatendofcycle-status

app/services/candidate_interface/reject_awaiting_references_application.rb
```diff
@@ -0,0 +1,13 @@
+module CandidateInterface
+  class RejectAwaitingReferencesApplication
+    def self.call(application_choice)
+      ActiveRecord::Base.transaction do
+        ApplicationStateChange.new(application_choice).reject_at_end_of_cycle!
```
duncanjbrown, line 5: Should we email them too?

config/vendor-api-v1.yml
tvararu: Fixt!

app/workers/reject_awaiting_references_course_choices_worker.rb
tvararu: Fixt!

app/services/candidate_interface/reject_awaiting_references_application.rb
tvararu: Going to do as a follow-up card!

app/workers/reject_awaiting_references_course_choices_worker.rb
```diff
@@ -0,0 +1,9 @@
+class RejectAwaitingReferencesCourseChoicesWorker
+  include Sidekiq::Worker
+
+  def self.perform
```
stevehook, line 4:
```suggestion
  def perform
```
This fixes the error in the Sidekiq worker for me locally

stevehook (commented):
As discussed on Slack...

stevehook (approved):
Looks good!

PR #2765, opened 2020-08-24
Make Faker's fake university names more diverse

duncanjbrown:
## Context

We see flaky tests and our fake data generation is not very diverse because Faker::Educator.university only has a handful of possible combinations

- Use Faker::University.name
- Stop it from assigning the names of American states to the universities it creates because that's not good fake data

## Link to Trello card

Flaky tests: https://trello.com/c/sd3pw4Lc/2636-two-flaky-tests

## Things to check


mnacos (approved):
:+1:  `Faker::University` seems a bit more international, too.

PR #2768, opened 2020-08-24
Make it easier to find applications from provider notification emails

duncanjbrown:
## Context

Long ago during the bug bash we noticed it was sometimes hard to get to an application from an email about it, especially if the user had a common name.

We did supply a link to the application if further action was needed, but this didn't cover all cases and it costs nothing to include it anyway. Also, we never included the support reference in the past.

- Always include the support reference in the subject line
- Always include a link to the application

While we're here, tidy up some translations that were spread across two files.

## Link to Trello card

https://trello.com/c/ZXol9U6s/2132-title-hard-to-find-application-from-email

## Guidance to review

All the emails have previews at `rails/mailers/provider_mailer`

## Things to check


spec/mailers/provider_mailer_spec.rb
```diff
@@ -39,20 +39,20 @@
                                  ))
     @provider_user = @application_choice.provider.provider_users.first
     @provider_user.update(first_name: 'Johny', last_name: 'English')
-    @application_choice.application_form.update(submitted_at: Time.zone.now - 5.days)
+    @application_choice.application_form.update(submitted_at: Time.zone.now - 5.days, support_reference: '123A')
```
AgaDufrat, line 42: I was wondering if it would be better to define the support_reference when we are creating the application form on lines 35-39.
```
create(
  :completed_application_form,
   first_name: 'Harry',
   last_name: 'Potter',
   support_reference: '123A',
 ))
```

AgaDufrat (approved):
The previews look good 👍 📬

spec/mailers/provider_mailer_spec.rb
duncanjbrown: Yes, much better, thanks

PR #2772, opened 2020-08-25
Carry over unsuccessful applications manually

stevehook:
## Context

This PR implements the front-end that allows candidates to carry over applications that were unsuccessful in an earlier recruitment cycle. It is modelled on existing Apply Again functionality with some adjustments to the content and application flow.

- [x] Implement new `carry_over` and `start_carry_over` actions for `SubmittedApplicationFormController`.
- [x] Adapt the `ApplyAgainBannerComponent` so that it can be used for both same-cycle (Apply 2) and later-cycle (carry over) re-applications. When the earlier application is from a different cycle it is now linked to the `SubmittedApplicationFormController#start_carry_over` rather than `start_apply_again` action.
- [x] System spec that walks through the carry over process.
- [x] Refine the content for the 'carry over application' journey 

I've made only minor changes to the copy on the pages that appear as part of the carry over flow. Just to remove the line _You could still get a place on a course this year_ and on the choose courses page removed the _new_ from the flash message _Your new application is ready for editing_ as well as the reference to only being able to pick one course.

We want to remove the _we suggest you choose a course first_ page but that will be covered in a different card/PR.

image

image

image

## Guidance to review

- To test manually you'll need some kind of unsuccessful application from 2020 (any of those created by the test application generator will do) and then change to the 2021 recruitment cycle by flipping the `switch_to_2021_recruitment_cycle` and `simulate time mid cycle` feature flags.
- Does the system spec make sense?
- For now at least it should only be possible to carry over an application after the 2021 cycle reopens so there is no feature to allow candidates to carry over mid cycle and start preparing their application before being able to submit it. Not sure whether this is going to be a requirement but I think it's best dealt with as a separate card/PR.

## Link to Trello card

## Things to check

### DIFF
--- a/app/components/candidate_interface/apply_again_banner_component.html.erb
+++ b/app/components/candidate_interface/apply_again_banner_component.html.erb
@@ -3,9 +3,9 @@
     <div class="govuk-heading-m">
       <p>
       <% if @application_form.application_choices.all?(&:cancelled?) %>
-        Your application has been withdrawn. <%= govuk_link_to 'Do you want to apply again?', candidate_interface_start_apply_again_path %>
+        Your application has been withdrawn. <%= govuk_link_to 'Do you want to apply again?', start_path %>
       <% else %>
-        <%= govuk_link_to 'Do you want to apply again?', candidate_interface_start_apply_again_path %>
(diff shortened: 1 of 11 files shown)


tvararu (approved):
Tested manually as well, great work 👍

PR #2774, opened 2020-08-25
Allow providers to filter applications by recruitment cycle

mnacos:
## Context

When the current recruitment cycle ends, all current applications will suddenly become previous year applications. Some of these applications will be carried over to the new cycle (handled in another piece of work), but many will remain in the previous cycle. Some of these will be actionable and should surface in the provider interface still. The plan is to break down the applications list into prioritised sections (see https://github.com/DFE-Digital/apply-for-teacher-training/pull/2742).

This PR allows provider users to filter applications based on the recruitment cycle.

Change the default visibility of applications to span two years (`:previous_year` and `:current_year`). Present users with the relevant controls.

image

I've reused the existing filtering mechanism, so combinations of search and status, provider, location filters can be used as well as recruitment_cycle. Changing the default visibility to span two years means that search will surface appllications from the previous year, too -- which has been identified as a user expectation in research.

To be able to merge this, I have created a feature flag `:providers_can_filter_by_recruitment_year`. When this flag is off, any cycle-related filtering param is ignored and the relevant UI controls are hidden.

This PR also changes the `setup_local_dev_data` task to sync courses from the previous year and generate test applications for the previous cycle, too. This is meant to assist with reviewing/testing this PR but all other End-of-Cycle bits of work, too. Finally, I've started adding `:previous_year` traits to `spec/factories.rb`.

## Guidance to review

Best way to test is interact with the review app. To test locally, you may have to drop, recreate your database and run `rake setup_local_dev_data`. You'll then need to turn the feature flag on to see the cycle filtering controls.

Is the feature flag needed? Should it be merged with `:switch_to_2021_recruitment_cycle`?

Some generated previous year applications may not look right, e.g. they should have been declined by default but aren't. I've dealt with rejection by default for previous cycle applications because we currently support sorting based on 'days left to respond', which depends on applications being rejected when they should be. In future work we could also handle 'decline by default' and automated application EoC transitions.

Sorting by 'days left to respond' may appear broken within the review app for the applications you cannot respond to, as we currently fall back to:
```
  reject_by_default_at ASC,
  application_choices.updated_at DESC
```
and there are many applications with `reject_by_default_at` in the past now, which takes precedence over `updated_at`. Not worth fixing, as we'll soon change the default applications order.

## Link to Trello card

https://trello.com/c/kO6hOtJG

## Things to check

### DIFF
--- a/app/lib/test_applications.rb
+++ b/app/lib/test_applications.rb
@@ -16,8 +16,8 @@ class TestApplications
   def create_application(states:, courses_to_apply_to: nil, apply_again: false, course_full: false)
     candidate = nil
 
-    min_days_in_the_past = courses_to_apply_to&.first&.recruitment_cycle_year == RecruitmentCycle.previous_year ? 395 : 30
-    travel_to rand(min_days_in_the_past..(min_days_in_the_past + 60)).days.ago
+    min_days_in_the_past = courses_to_apply_to&.first&.recruitment_cycle_year == RecruitmentCycle.previous_year ? 385 : 20
+    travel_to rand(min_days_in_the_past..(min_days_in_the_past + 30)).days.ago
 
     if apply_again
       raise OnlyOneCourseWhenApplyingAgainError, 'You can only apply to one course when applying again' unless states.one?
@@ -116,7 +116,7 @@ class TestApplications
         # time to decide whether edit_by has passed, temporarily set edit_by to
         # a date in the future relative to now
         @application_form.application_choices.each { |choice| choice.update_columns(updated_at: time) }
-        @application_form.update_columns(submitted_at: time, edit_by: Time.zone.now + 1.day, updated_at: time)
+        @application_form.update_columns(submitted_at: time, edit_by: Time.zone.now + 7.days, updated_at: time)
 
         return if states.include? :awaiting_references
 
@@ -159,10 +159,10 @@ class TestApplications
     travel_to(choice.application_form.edit_by) if choice.application_form.edit_by > time
     SendApplicationToProvider.new(application_choice: choice).call
     choice.update_columns(sent_to_provider_at: time)
-    choice.audits.last.update_columns(created_at: time)
+    choice.audits.last&.update_columns(created_at: time)
     SetRejectByDefault.new(choice).call
     choice.update_columns(updated_at: time)
-    choice.audits.last.update_columns(created_at: time)
+    choice.audits.last&.update_columns(created_at: time)
 
     return if state == :awaiting_provider_decision
 
@@ -202,21 +202,21 @@ class TestApplications
     fast_forward(1..3)
     AcceptOffer.new(application_choice: choice).save!
     choice.update_columns(accepted_at: time, updated_at: time)
-    choice.audits.last.update_columns(created_at: time)
+    choice.audits.last&.update_columns(created_at: time)
   end
 
   def withdraw_application(choice)
     fast_forward(1..3)
     WithdrawApplication.new(application_choice: choice).save!
     choice.update_columns(withdrawn_at: time, updated_at: time)
-    choice.audits.last.update_columns(created_at: time)
+    choice.audits.last&.update_columns(created_at: time)
   end
 
   def decline_offer(choice)
     fast_forward(1..3)
     DeclineOffer.new(application_choice: choice).save!
     choice.update_columns(declined_at: time, updated_at: time)
-    choice.audits.last.update_columns(created_at: time)
+    choice.audits.last&.update_columns(created_at: time)
   end
 
   def make_offer(choice, conditions: ['Complete DBS'])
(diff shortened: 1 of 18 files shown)


app/queries/get_application_choices_for_providers.rb
```diff
@@ -17,9 +17,9 @@ def self.call(providers:)
     ]
 
     ApplicationChoice.includes(*includes)
-      .where('courses.provider_id' => providers, 'courses.recruitment_cycle_year' => RecruitmentCycle.current_year)
+      .where('courses.provider_id' => providers, 'courses.recruitment_cycle_year' => RecruitmentCycle.visible_years)
```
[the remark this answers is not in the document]
mnacos: :) Yes, well, first of all we need to decide if we'll be providing historical access to applications! But it sounds like we may have to separate the logic for granting access to a single application from that of populating the listview at some point in the future.

spec/models/provider_interface/provider_applications_page_state_spec.rb
```diff
@@ -28,11 +30,33 @@
...
+
+      page_state = described_class.new(
+        params: ActionController::Parameters.new,
+        provider_user: provider_user,
+        state_store: {},
+      )
+
+      expected_number_of_filters = 4
+      recruitment_cycle_index = 1
+      providers_array_index = 3
```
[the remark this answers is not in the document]
mnacos: I'll try and make the expectation clearer. The position of later filter sections in this array is dependent on the state of the feature flag, maybe a lookup would address this.

lib/tasks/local_dev.rake
```diff
@@ -44,6 +44,7 @@ task sync_dev_providers_and_open_courses: :environment do
 
   provider_codes = HostingEnvironment.review? ? %w[1JA 24J] : %w[1JA 24J 1N1]
   provider_codes.each do |code|
+    SyncProviderFromFind.call(provider_code: code, sync_courses: true, provider_recruitment_cycle_year: RecruitmentCycle.previous_year)
```
stevehook, line 47: :+1: This will be very useful

spec/models/provider_interface/provider_applications_page_state_spec.rb
```diff
@@ -99,12 +125,12 @@
 
     it 'returns true if filers have been applied' do
```
stevehook, line 126:
```suggestion
    it 'returns true if filters have been applied' do
```

stevehook (commented):
This all looks good to me. I've tried to integrate `RecruitmentCycle` into a couple of PRs I've been working on and the new method will help too.

More generally I wonder how we are going to handle the switch over from one year to the next, assuming we are not going to add a new feature flag for each year! Could we determine `RecruitmentCycle#current_year` from the dates we store in `EndOfCycleTimetable` so that the switchover is automatic? Might be a couple of issues with this I guess - we may want to retain the manual switch and it raises the question of when exactly does the new cycle begin? Is it on the day that the new cycle opens for applications or earlier?

app/lib/test_applications.rb
```diff
@@ -111,7 +115,8 @@ def create_application(states:, courses_to_apply_to: nil, apply_again: false, co
         # To prevent immediate sending to provider, which uses the current
         # time to decide whether edit_by has passed, temporarily set edit_by to
         # a date in the future relative to now
-        @application_form.update_columns(submitted_at: time, edit_by: Time.zone.now + 1.day)
+        @application_form.application_choices.each { |choice| choice.update_columns(updated_at: time) }
+        @application_form.update_columns(submitted_at: time, edit_by: Time.zone.now + 1.day, updated_at: time)
```
duncanjbrown, line 119: From call: we could push this "real" edit_by date further out so the application remained in `application_complete` for a few days, which is a bit more realistic

app/lib/test_applications.rb
```diff
@@ -127,22 +132,23 @@ def create_application(states:, courses_to_apply_to: nil, apply_again: false, co
...
+        application_choices.each { |choice| choice.update_columns(updated_at: time) }
 
         states.zip(application_choices).each do |state, application_choice|
           put_application_choice_in_state(application_choice, state)
         end
       end
 
       application_choices.each do |application_choice|
-        rand(0..3).times { add_note(application_choice) }
+        rand(0..3).times { add_note(application_choice.reload) }
```
duncanjbrown, line 151: From call: May not be needed

duncanjbrown (approved):
Talked through in a call and this is ready to go! 🚀 Minor comments inline as discussed

mnacos:
> More generally I wonder how we are going to handle the switch over from one year to the next, assuming we are not going to add a new feature flag for each year! Could we determine `RecruitmentCycle#current_year` from the dates we store in `EndOfCycleTimetable` so that the switchover is automatic? Might be a couple of issues with this I guess - we may want to retain the manual switch and it raises the question of when exactly does the new cycle begin? Is it on the day that the new cycle opens for applications or earlier?

That's an interesting possibility, definitely something to discuss.

PR #2775, opened 2020-08-25
Introduce an API field indicating whether safeguarding information has been supplied

AgaDufrat:
## Context

We do not want to send safeguarding information via the API as it means our clients become liable for looking after this sensitive data. They do need to know whether this information is present, though. Provider users will need to sign in to the web UI to see safeguarding information.

Add new fields to the Application object:
- `safeguarding_issues_status`
- `safeguarding_issues_details_url`

## Guidance to review

- do the descriptions make sense?
- is the example URL (with id) ok?
- is it ok to include url_helpers?

## Link to Trello card

https://trello.com/c/4CO0fhJ7/2584-introduce-an-api-field-indicating-whether-safeguarding-information-has-been-supplied

## Things to check

### DIFF
--- a/app/presenters/vendor_api/single_application_presenter.rb
+++ b/app/presenters/vendor_api/single_application_presenter.rb
@@ -43,6 +43,7 @@ module VendorAPI
           rejection: get_rejection,
           withdrawal: withdrawal,
           further_information: application_form.further_information,
+          safeguarding_issues_status: application_form.safeguarding_issues_status,
         },
       }
       if application_choice.status == 'enrolled'
--- a/app/presenters/vendor_api/single_application_presenter.rb
+++ b/app/presenters/vendor_api/single_application_presenter.rb
@@ -1,5 +1,7 @@
 module VendorAPI
   class SingleApplicationPresenter
+    include Rails.application.routes.url_helpers
+
     def initialize(application_choice)
       @application_choice = application_choice
       @application_form = application_choice.application_form
@@ -44,6 +46,7 @@ module VendorAPI
           withdrawal: withdrawal,
           further_information: application_form.further_information,
           safeguarding_issues_status: application_form.safeguarding_issues_status,
+          safeguarding_issues_details_url: safeguarding_issues_details_url,
         },
       }
       if application_choice.status == 'enrolled'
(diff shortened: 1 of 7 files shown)


duncanjbrown:
Is `safeguarding_issues_status` really a boolean? The enum in `ApplicationForm` has four options — I guess we'll never see `not_answered_yet`, but the other three are all still possible.

AgaDufrat:
The A/C on the card said _"New field in API response: safeguarding_issues_status (true/false)"_ so I thought that's what was decided. Happy to change it to appropriate strings.

duncanjbrown:
> > Is `safeguarding_issues_status` really a boolean? The enum in `ApplicationForm` has four options — I guess we'll never see `not_answered_yet`, but the other three are all still possible.
> 
> The A/C on the card said _"New field in API response: safeguarding_issues_status (true/false)"_ so I thought that's what was decided. Happy to change it to appropriate strings.

I think that was my bad, sorry. Let's do strings.

app/presenters/vendor_api/single_application_presenter.rb
```diff
@@ -260,5 +264,9 @@ def work_history_breaks
         ''
       end
     end
+
+    def safeguarding_issues_details_url
+      application_form.has_safeguarding_issues_to_declare? ? provider_interface_application_choice_url(application_choice).concat('#criminal-convictions-and-professional-misconduct') : nil
```
duncanjbrown, line 269: I think Rails path helpers support an `anchor:` property (https://stackoverflow.com/questions/51895357/rails-5-appending-an-html-anchor-to-rails-route-path-helper)

config/vendor-api-v1.yml
```diff
@@ -471,6 +473,20 @@ components:
           description: Other personal or professional issues relevant to the application
             which are not covered in the form
           example: I am expecting to be called for jury service in the next 12 months
+        safeguarding_issues_status:
+          type: string
+          description: Status of candidate's response to the safeguarding issues declaration
+          enum:
+          - not_answered_yet
```
duncanjbrown, line 480: I don't think they'll ever get this?

duncanjbrown (approved):
Looks good — can we lose `not_answered_yet` as the form isn't complete (and therefore isn't sendable) without it?

config/vendor-api-v1.yml
AgaDufrat: You are right, on providers side it can't be not_answered_yet.

PR #2776, opened 2020-08-25
Update back links to use gov_back_link_to so that they work contextually from both review screens

george-bit:

Change the context of the back links so that they direct to the correct review page dependent on where the last

## Guidance to review

Using the review app check that all links flow correctly to the review page. From the individual and application review pages.

## Link to Trello card

https://trello.com/c/x7Hw1Waq/1896-%F0%9F%94%99-when-editing-information-back-links-and-save-buttons-should-return-candidate-to-relevant-review-page

## Things to check


paulrobertlloyd:
@george-bit Looks good to me. Uncovered a few other bugs while reviewing this one, but they are unrelated, so :shipit:!

PR #2777, opened 2020-08-25
Fix spec failures following end-of-cycle deadline

stevehook:
## Context

Since the 2020 recruitment cycle ended last night some tests have begun to fail because they exercise functionality that is turned off by design between cycles.

- [x] Run all specs on a mid-cycle date by default.
- [x] Adjusted a few specs that have expectations based on dates calculated relative to `now` outside examples to use lambdas so that they get evaluated inside the example.

## Guidance to review

I think this could be improved in a few ways. So we should iterate. Is this ok to get us back to a green state?

## Link to Trello card

n/a

## Things to check

### DIFF
--- a/spec/mailers/candidate_mailer_spec.rb
+++ b/spec/mailers/candidate_mailer_spec.rb
@@ -96,7 +96,7 @@ RSpec.describe CandidateMailer, type: :mailer do
       it_behaves_like(
         'a mail with subject and content', :chase_candidate_decision,
         I18n.t!('chase_candidate_decision_email.subject_singular'),
-        'Date to resbond by' => -> { "Respond by #{10.business_days.from_now.to_s(:govuk_date).strip}" }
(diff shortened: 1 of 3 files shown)


spec/mailers/candidate_mailer_spec.rb
```diff
@@ -95,7 +96,7 @@
       it_behaves_like(
         'a mail with subject and content', :chase_candidate_decision,
         I18n.t!('chase_candidate_decision_email.subject_singular'),
-        'Date to resbond by' => "Respond by #{10.business_days.from_now.to_s(:govuk_date).strip}"
+        'Date to resbond by' => -> { "Respond by #{10.business_days.from_now.to_s(:govuk_date).strip}" }
```
tvararu, line 99:
```suggestion
        'Date to respond by' => -> { "Respond by #{10.business_days.from_now.to_s(:govuk_date).strip}" }
```

Spotted a typo

duncanjbrown (approved):
Green is good

tvararu (approved):

PR #2779, opened 2020-08-25
Add feature flag to enable normal features between recruitment cycles

stevehook:
## Context

Vendors/Providers cannot currently test 'normal' functionality (e.g. signing up candidates, adding courses and submitting applications normally) because we are between recruitment cycles. We need a simple way to configure a system to restore 'business as usual' behaviour at any time of the year rather than have them switched off between cycles.

- [x] Add `simulate_time_mid_cycle` feature flag to simulate mid recruitment cycle behaviour
- [x] Implement feature flag in `EndOfCycleTimetable` by moving the dates for EoC into the fairly distant future.

## Guidance to review

- This solution assumes that all users of a given system, i.e. all providers and vendors using Sandbox will either be in a recruitment cycle or not.
- `EndOfCycleTimetable` is getting a bit big now. I don't think we should be adding any more to it after this.

## Link to Trello card

https://trello.com/c/ReTgYcOl/2023-providers-vendors-cannot-test-normal-functionality-in-sandbox-environment-between-recruitment-cycles

## Things to check

### DIFF
--- a/app/services/end_of_cycle_timetable.rb
+++ b/app/services/end_of_cycle_timetable.rb
@@ -54,8 +54,12 @@ class EndOfCycleTimetable
   end
 
   def self.date(name)
-    if HostingEnvironment.test_environment? && FeatureFlag.active?(:simulate_time_between_cycles)
-      return simulate_time_between_cycles_dates[name]
+    if HostingEnvironment.test_environment? || HostingEnvironment.sandbox_mode?
+      if FeatureFlag.active?(:simulate_time_between_cycles)
+        return simulate_time_between_cycles_dates[name]
+      elsif FeatureFlag.active?(:simulate_time_mid_cycle)
+        return simulate_time_mid_cycle_dates[name]
+      end
     end
 
     DATES[name]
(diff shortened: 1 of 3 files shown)


duncanjbrown:
This looks good to me — just going to test locally before ✅

tvararu (approved):

app/services/end_of_cycle_timetable.rb
```diff
@@ -54,8 +54,12 @@ def self.between_cycles_apply_2?
   end
 
   def self.date(name)
-    if HostingEnvironment.test_environment? && FeatureFlag.active?(:simulate_time_between_cycles)
-      return simulate_time_between_cycles_dates[name]
+    if HostingEnvironment.test_environment? || HostingEnvironment.sandbox_mode?
+      if FeatureFlag.active?(:simulate_time_between_cycles)
+        return simulate_time_between_cycles_dates[name]
+      elsif FeatureFlag.active?(:simulate_time_mid_cycle)
```
duncanjbrown, line 60: These flags should really be mutually exclusive — I wonder if there's a way to enforce that. This solves the immediate problem tho

spec/services/end_of_cycle_timetable_spec.rb
```diff
@@ -71,6 +71,78 @@
...
+      it 'returns true after the configured date' do
+        Timecop.travel(Time.zone.local(2020, 9, 19, 1, 0, 0)) do
+          expect(EndOfCycleTimetable.show_apply_2_deadline_banner?).to be true
+        end
+      end
+    end
+
+    describe '.between_cycles_apply_1?' do
+      it 'returns false before the configured date' do
+        Timecop.travel(Time.zone.local(2020, 8, 24, 21, 0, 0)) do
```
duncanjbrown, line 107: no blocker — it would be good to somehow label these dates

duncanjbrown (approved):
It works! Thanks @stevehook

app/services/end_of_cycle_timetable.rb
stevehook: I did wonder about that but I don't think these switches should be feature flags in the longer term. We could add validation to the `FeatureFlag` model to enforce it? Going to merge for now but happy to add that as a follow-up if you think worthwhile.

PR #2788, opened 2020-08-27
Stop applications with unavailable course options after Sep 7

malcolmbaig:
## Context

Providers don't expect to receive applications to full or withdrawn course options after Sep 7th. We need to stop these at the point they're ready to send and notify the candidate.

- Add methods to `EndOfCyclePolicy` and `EndOfCycleTimetable` to handle checking if we should apply this behaviour.
- Update `SendApplicationsToProvider` to cancel the application if appropriate.
- Send one of two notification emails to the candidate depending on their application state.

## Guidance to review

- Per-commit recommended
- Mailer preview paths:
  - /rails/mailers/candidate_mailer/eoc_choice_unavailable_and_no_other_choices
  - /rails/mailers/candidate_mailer/eoc_choice_unavailable_and_still_waiting_on_other_choices
- The card states that these applications should be "rejected as full". This isn't a pre-existing state transition from `application_complete`, so for simplicity I've opted to use the existing transition to `cancelled`. Are there likely to be any issues with this?

---

### Email sent when there are no other active course choices on the application:

**Subject: Find a different course and re-submit your application**

Screenshot_20200828_113436

---

### Email sent when there are other active course choices on the application:

**Subject: Application could not be sent for (course name and code) at (provider name)**

#### One other course choice
Screenshot_20200828_114748

#### More than one other course choice
Screenshot_20200828_114404

## Link to Trello card

https://trello.com/c/iExqduky

## Things to check

### DIFF
--- a/app/mailers/candidate_mailer.rb
+++ b/app/mailers/candidate_mailer.rb
@@ -262,8 +262,8 @@ class CandidateMailer < ApplicationMailer
       @application_form,
       subject: I18n.t!(
         'candidate_mailer.eoc_choice_unavailable_has_other_choices.subject',
-        course_name: application_choice.course.name_and_code,
-        provider_name: application_choice.course.provider.name,
+        course_name: application_choice.course_option.course.name_and_code,
+        provider_name: application_choice.course_option.course.provider.name,
       ),
     )
   end
--- a/app/mailers/candidate_mailer.rb
+++ b/app/mailers/candidate_mailer.rb
@@ -253,6 +253,16 @@ class CandidateMailer < ApplicationMailer
     )
   end
 
+  def eoc_choice_unavailable_and_no_other_choices(application_choice)
+    @application_choice = application_choice
+    @application_form = application_choice.application_form
+
+    email_for_candidate(
+      @application_form,
+      subject: I18n.t!('candidate_mailer.eoc_choice_unavailable_no_other_choices.subject'),
+    )
+  end
+
 private
 
   def new_offer(application_choice, template_name)
--- a/app/mailers/candidate_mailer.rb
+++ b/app/mailers/candidate_mailer.rb
@@ -253,6 +253,21 @@ class CandidateMailer < ApplicationMailer
     )
   end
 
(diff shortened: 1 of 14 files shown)


tvararu (commented):
Would be nice to test this locally first before merging; I've greened the other PR with the feature flag to help with this.

Only bit I'm concerned about is the fact that this PR makes `app/services/send_applications_to_provider.rb` a lot more complex, whereas before it was just 4 lines. I can't think of a better implementation however.

tvararu (approved):
Screenshot 2020-08-27 at 14 52 41

Works 💯

stevehook (approved):
Just managed to get it working locally for the first case. As discussed code looks great.

> The card states that these applications should be "rejected as full". This isn't a pre-existing state transition from application_complete, so for simplicity I've opted to use the existing transition to cancelled. Are there likely to be any issues with this?

I think this is OK from a tech point of view. From an analysis point of view the extra status might be very useful (though not sure given that this is not going to be a very common thing). Anyway it can be dealt with as a separate card if needed.

EmmaFrith:
Re. the first screenshot - the content has since been edited following discussions - I'm sorry! Here's the updated content:

Subject: Find a different course and resubmit your application 

Dear [Candidate]

# You may still be able to get on a course

Due to [course name and code] filling up at [provider name], your recent application did not lead to a place on a course. 

# Resubmit no later than 18 September

You can resubmit your application for an available course until 18 September. 

Find out if there are any other suitable courses available:

[link to find]

If you find a course, continue your application:

[link to apply 2 interstitial ‘do you want to apply again’]

# Get help from a teacher training adviser

You can talk to a teacher training adviser on the phone or online: https://register.getintoteaching.education.gov.uk/register

EmmaFrith:
@malcolmbaig  - on the second screenshot - is it possible to use 'another course' or 'other courses' depending on if it's 1 or 2? If this is unnecessarily complex, I'll think of something which applies to both scenarios. 

(the existing content doesn't work when read in the singular)

EmmaFrith:
thanks @malcolmbaig - content comments resolved offline

PR #2789, opened 2020-08-27
Be strict about how clients use TextOrdinalizer

duncanjbrown:
## Context

Not a likely scenario irl, but a user had > 10 applications  on the sandbox: https://sentry.io/organizations/dfe-bat/issues/1861573888/?referrer=slack

`LinksToPreviousApplicationsComponent` blew up because it (legitimately) tried to ordinalize a number > 10 without going through `TextOrdinalizer`'s proper interface, and ended up with an unexpected nil.

Being strict about the public interface of `TextOrdinalizer` means this isn't possible any more. This change does not damage `LinksToPreviousApplicationsComponent` (or the other call sites) because `TextOrdinalizer`'s fallback to Rails's `#ordinalize` method will return eg `11th` which noops when you try to `#capitalize` it.

See commit

## Link to Trello card

https://sentry.io/organizations/dfe-bat/issues/1861573888/?referrer=slack

## Things to check

### DIFF
--- a/app/components/candidate_interface/links_to_previous_applications_component.rb
+++ b/app/components/candidate_interface/links_to_previous_applications_component.rb
@@ -14,7 +14,7 @@ module CandidateInterface
   private
 
     def ordinalize(application_form)
-      "#{TextOrdinalizer::ORDINALIZE_MAPPING[ordinal(application_form)].capitalize} application"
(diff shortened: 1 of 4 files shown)


tvararu (approved):

stevehook (approved):
:+1:

PR #2795, opened 2020-08-27
Store ProviderRelationshipPermissions wizard data in Redis

duncanjbrown:
## Context

We used the session to hold data from the wizard. Unfortunately we overflowed it and this caused 500s because cookies can only be 4k. It's insane that there can be 4k of data in these things but it's possible that other things are taking up space in the session too (flash messages, DSI data, etc).

We had a choice: shrink the data, or change the persistence method. In this PR I have chosen to change the persistence method.

I tried compressing the data with Ruby's zlib library but it gave us < 50% compression per permission set, which wasn't enough to make me feel safe. We could probably have hand-rolled a compression scheme and stored all the data we need in a couple of bytes but that would have been difficult and complicated.

Therefore I chose to store the data in Redis instead. I did not choose to move the whole session into Redis because that's not a decision to take lightly and I want to fix this bug without having to make it.

We have two wizards, one for invitations and one for setting permissions.

**Step 1** was to refactor the invitations wizard to depend on a `WizardStateStore::SessionStore` instead of a plain `session`.

**Step 2** was to do the same for the permissions wizard.

**Step 3** was to implement a `WizardStateStore::RedisStore` and have the permissions wizard use that.

There is a slight risk to changing the storage backend for one of these wizards because the user might have data in the session already. In the case of permissions this is not a risk because the feature is not live yet. Therefore this PR remains deployable. We could subsequently ship a PR which enabled Redis for the invitation wizard and migrated any extant sessions to Redis. Or, of course, we could simply YOLO it and switch it over in the middle of the night when no providers are at work. Note that in any case having `SessionStore` is handy because we can test the units without Redis.

Now that the backend is passed into the Wizard, the controller is in charge of setting the persistence key because it has the context of the current user etc. This doesn't seem too terrible.

## Guidance to review

Does all this make sense? Any suggestions for simplifying it? Should `SessionStore` be called `HashStore` or something? Could the persistence keys ever conflict with each other?

## Link to Trello card

2650-modifying-provider-permissions-loads-more-than-4kb-into-the-cookie-which-causes-a-cookieoverflow-error

## Things to check

### DIFF
--- a/app/controllers/provider_interface/provider_users_invitations_controller.rb
+++ b/app/controllers/provider_interface/provider_users_invitations_controller.rb
@@ -113,7 +113,11 @@ module ProviderInterface
 
     def wizard_for(options)
       options[:checking_answers] = true if params[:checking_answers] == 'true'
-      ProviderUserInvitationWizard.new(session, options)
+
+      ProviderUserInvitationWizard.new(
+        WizardStateStores::SessionStore.new(session: session, key: persistence_key_for_current_user),
+        options,
+      )
     end
 
     def path_for(step, provider_id)
@@ -156,6 +160,10 @@ module ProviderInterface
         provider_id: @provider.id,
         permissions: @wizard.permissions_for_provider(@provider.id),
       )
+    end
+
+    def persistence_key_for_current_user
+      "provider_user_invitation_wizard-#{current_provider_user.id}"
     end
   end
 end
--- a/app/lib/wizard_state_stores/redis_store.rb
+++ b/app/lib/wizard_state_stores/redis_store.rb
@@ -8,7 +8,7 @@ module WizardStateStores
     end
 
     def write(value)
-      @redis.set(@key, value)
+      @redis.set(@key, value, ex: ActiveSupport::Duration::SECONDS_PER_DAY)
     end
 
     def read
--- a/app/lib/wizard_state_stores/redis_store.rb
+++ b/app/lib/wizard_state_stores/redis_store.rb
@@ -1,0 +1,22 @@
+require 'redis'
+
+module WizardStateStores
+  class RedisStore
+    def initialize(key:)
+      @redis = Redis.new
+      @key = key
+    end
+
+    def write(value)
+      @redis.set(@key, value)
+    end
+
+    def read
+      @redis.get(@key)
+    end
+
+    def delete
+      @redis.del(@key)
+    end
+  end
+end
--- a/app/lib/wizard_state_stores/redis_store.rb
+++ b/app/lib/wizard_state_stores/redis_store.rb
@@ -3,7 +3,7 @@ require 'redis'
 module WizardStateStores
   class RedisStore
     def initialize(key:)
-      @redis = Redis.new
+      @redis = Redis.current
       @key = key
     end
 
--- a/app/controllers/provider_interface/provider_relationship_permissions_setup_controller.rb
+++ b/app/controllers/provider_interface/provider_relationship_permissions_setup_controller.rb
@@ -112,7 +112,10 @@ module ProviderInterface
 
     def wizard_for(options)
       options[:checking_answers] = true if params[:checking_answers] == 'true'
-      ProviderRelationshipPermissionsSetupWizard.new(session, options)
+      ProviderRelationshipPermissionsSetupWizard.new(
+        WizardStateStores::SessionStore.new(session: session, key: persistence_key_for_current_user),
+        options,
+      )
     end
 
     def require_feature_flag!
@@ -146,6 +149,10 @@ module ProviderInterface
       @permissions_form = ProviderInterface::ProviderRelationshipPermissionsSetupWizard::PermissionsForm.new(
         @wizard.permissions_for_relationship(@permissions_model.id).merge(id: @permissions_model.id),
       )
+    end
+
+    def persistence_key_for_current_user
+      "provider_user_permissions_wizard-#{current_provider_user.id}"
     end
   end
 end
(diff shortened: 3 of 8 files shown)


app/lib/wizard_state_stores/redis_store.rb
```diff
@@ -0,0 +1,22 @@
...
+
+module WizardStateStores
+  class RedisStore
+    def initialize(key:)
+      @redis = Redis.new
+      @key = key
+    end
+
+    def write(value)
+      @redis.set(@key, value)
```
duncanjbrown, line 11: We should probably expire this key after a few hours

duncanjbrown: Added 24-hr expiry

app/controllers/provider_interface/provider_users_invitations_controller.rb
```diff
@@ -157,5 +161,9 @@ def setup_permission_form
         permissions: @wizard.permissions_for_provider(@provider.id),
       )
     end
+
+    def persistence_key_for_current_user
+      "provider_user_invitation_wizard-#{current_provider_user.id}"
```
mnacos, line 166: It's not a big deal, but the controller shouldn't have to worry about generating keys. We have `STATE_STORE_KEY` in the wizard, and it'd be better if the wizard decides how to generate its own keys. Having said that, the wizard has no access to `current_user` currently.

We could start passing a user context to the wizard, but I think a better approach would be to pass `session:` to both wizards, making the two store implementations have a uniform API, making them truly interchangeable. The session object exposes an `.id` which is unique to the browser session. How would you feel about using a combination of `STATE_STORE_KEY` and `session.id` as the key instead? `session.id` can be `nil`, but this happens only when a browser first visits the site or if it rejects cookies and is guaranteed to be initialised after authentication.

mnacos (commented):
Looks great. I'd be tempted to switch the entire app's `session_store` to Redis, but realise it would either involve messing with `cache_store` or using `redis-session-store` but I haven't used the latter with Rails 6 yet. So for the purposes of unblocking org permissions and supporting the wizards I think this implementation is fine.

However, I'm a bit worried about opening a new Redis connection each time the wizard is used (I think the current code opens one connection per page load, if the page accesed involves a wizard). Page load turnaround is quick and our volumes are low, so this would be fine if connections were garbage-collected, but some quick tests on my laptop and some reports (see https://github.com/redis/redis-rb/issues/524) suggest they are not. We may have to add explicit `.disconnect` calls to be safe, or add some connection pooling.

I've also left a suggestion to change the keys to be session-based, rather than user_id-based, in order to make key generation the responsibility of each wizard without needing to pass a user context.

app/controllers/provider_interface/provider_users_invitations_controller.rb
duncanjbrown: This is a nice idea, thank you. Will have a go.

duncanjbrown:
> Looks great. I'd be tempted to switch the entire app's `session_store` to Redis, but realise it would either involve messing with `cache_store` or using `redis-session-store` but I haven't used the latter with Rails 6 yet. So for the purposes of unblocking org permissions and supporting the wizards I think this implementation is fine.

I think we're on the same page here! Maybe we can revisit. It occurred to me that it's actually quite nice having these things separately in Redis because we can easily nullify them (or even migrate them to a new format) without having to get superstitious about whether there are any clients left in the world with stale session data.

> 
> However, I'm a bit worried about opening a new Redis connection each time the wizard is used (I think the current code opens one connection per page load, if the page accesed involves a wizard). Page load turnaround is quick and our volumes are low, so this would be fine if connections were garbage-collected, but some quick tests on my laptop and some reports (see [redis/redis-rb#524](https://github.com/redis/redis-rb/issues/524)) suggest they are not. We may have to add explicit `.disconnect` calls to be safe, or add some connection pooling.

Good point — this hadn't occurred to me! Would like very much not to have to worry about it. Looks like Sidekiq depends on `connection_pool` anyway so we may able to use `Redis.current`. Do you think it's worth running your local tests again with `Redis.current` in place of `Redis.new`?

app/lib/wizard_state_stores/redis_store.rb
```diff
@@ -0,0 +1,22 @@
+require 'redis'
+
+module WizardStateStores
+  class RedisStore
+    def initialize(key:)
+      @redis = Redis.new
```
mnacos, line 6:
```suggestion
      @redis = Redis.current
```

works well for re-using connections.

mnacos:
> Good point — this hadn't occurred to me! Would like very much not to have to worry about it. Looks like Sidekiq depends on `connection_pool` anyway so we may able to use `Redis.current`. Do you think it's worth running your local tests again with `Redis.current` in place of `Redis.new`?

Excellent idea, seems to work well.

app/controllers/provider_interface/provider_users_invitations_controller.rb
duncanjbrown: I've had a go, and had a think, and I think I prefer the current implementation for the following reasons:

- It's simply more complicated to have the state store depend on a session object than a string, and it's even a bit weird when the state store has nothing to do with sessions in the first place
- It feels more logical to have the store itself,  rather than some ingredient of the store, as a dependency of the wizard
- I don't think having equivalent constructors is worth the trouble of taking a more complicated dependency

mnacos: The current implementation works fine so I'm happy to go with it. There is another side-effect to applying `user_id` as part of the wizard key, wizard state will be persisted across browsers, incognito sessions or even machines (up to the 24h expiry). Using a browser session id would probably be more in line with user expectations, but it's not that important.

mnacos (approved):
Works great. I'd still prefer wizard keys based on the browser session rather than the user id but persisting progress across browser sessions could even be a feature :shipit:

PR #2797, opened 2020-08-28
Add clock entry for automatically carrying over unsubmitted apps

stevehook:
## Context

This is a follow up to https://github.com/DFE-Digital/apply-for-teacher-training/pull/2759 that sets up a clock entry to run the automatic carry over of unsubmitted applications to the new recruitment cycle.

- Add clock entry to run existing `CarryOverUnsubmittedApplicationsWorker` (in Sidekiq) at 1 minute past midnight on the day that the new recruitment cycle opens.

## Guidance to review

- Is a one-off job the right approach here?

## Link to Trello card

## Things to check

### DIFF
--- a/config/clock.rb
+++ b/config/clock.rb
@@ -22,6 +22,7 @@ class Clock
   every(1.hour, 'SendCourseFullNotifications', at: '**:45') { SendCourseFullNotificationsWorker.perform_async }
 
   every(1.hour, 'RejectAwaitingReferencesCourseChoices', at: '**:50') { RejectAwaitingReferencesCourseChoicesWorker.perform_async }
+  every(1.day, 'CarryOverUnsubmittedApplications', at: '00:01', if: ->(t) { t.to_date == EndOfCycleTimetable.date(:next_cycle_opens) }) { CarryOverUnsubmittedApplicationsWorker.perform_async }
 
   every(1.day, 'UCASMatching::UploadMatchingData', at: '06:23') do
     if Time.zone.today.weekday?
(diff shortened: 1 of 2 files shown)


config/clock.rb
```diff
@@ -22,6 +22,7 @@ class Clock
   every(1.hour, 'SendCourseFullNotifications', at: '**:45') { SendCourseFullNotificationsWorker.perform_async }
 
   every(1.hour, 'RejectAwaitingReferencesCourseChoices', at: '**:50') { RejectAwaitingReferencesCourseChoicesWorker.perform_async }
+  every(1.day, 'CarryOverUnsubmittedApplications', at: '00:01', if: ->(t) { t.to_date == EndOfCycleTimetable.date(:next_cycle_opens) }) { CarryOverUnsubmittedApplicationsWorker.perform_async }
```
tvararu, line 25: Wondering if we should do the same for the `RejectAwaitingReferencesCourseChoicesWorker` above?

tvararu (approved):

config/clock.rb
stevehook: Yes it's running redundantly most of the time. Does it need to happen on the same day? When the cycle opens or earlier than that?

stevehook: I think it should the `RejectAwaitingReferencesCourseChoicesWorker` should run after apply 2 closes (19 Sept). Slightly more complicated than I thought so going to raise a separate PR and get this one merged.

https://github.com/DFE-Digital/apply-for-teacher-training/compare/1959-change-reject-awaiting-refs-to-oneoff-job?expand=1

PR #2800, opened 2020-08-28
Add null option to GCSE grade select

malcolmbaig:
Not doing so defaults to the first option in the list. This leads to a confusing UI when autocomplete JS is applied to the element, as you can only see a few matches that start with the first option (ie - grades beginning with '9').

## Context

See Trello card.
## Link to Trello card
https://trello.com/c/0GhLI2EE/2053-drop-down-arrow-not-working-and-grade-pre-filled-for-gcse

## Things to check

### DIFF
--- a/app/forms/candidate_interface/gcse_qualification_details_form.rb
+++ b/app/forms/candidate_interface/gcse_qualification_details_form.rb
@@ -14,7 +14,7 @@ module CandidateInterface
     Grade = Struct.new(:value, :option)
 
     def self.all_grade_drop_down_options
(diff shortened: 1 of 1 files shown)


paulrobertlloyd (commented):
I think we can probably improve the UX further (when editing a grade, candidate should be able to see all possible answers not just those filtered by currently input value), but in terms of this particular issue, this fixes it. :shipit:

stevehook (approved):
:+1: I think this solves the biggest problem with the autocomplete. Let's get this PR merged and keep the card open to address the other issues.

PR #2803, opened 2020-09-01
Add support page with all info about recruitment cycles

tijmenb:
## Context

We have a lot of settings related to recruitment cycles. They can be overridden by feature flags and per-environment. 

Add a page in support that exposes all the current variables and derived flags:

image

## Guidance to review

Are there other interesting bits of info that we can display?

## Things to check

### DIFF
--- a/app/helpers/view_helper.rb
+++ b/app/helpers/view_helper.rb
@@ -83,6 +83,12 @@ module ViewHelper
     (application_choice.reject_by_default_at.to_date - Date.current).to_i
   end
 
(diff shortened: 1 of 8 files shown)


mnacos (approved):
Thanks, very helpful. Why `/support/*docs*/cycles`? The other tabs have routes like `/support/when-emails-are-sent`.

tijmenb:
@mnacos Trying to organise it bettter - I'll move the other ones in a follow up PR!

PR #2804, opened 2020-09-01
Add offer_deferred status

AgaDufrat:
## Context

Providers need to be able to defer offers to the next recruitment cycle.

- Add a new `offer_deferred` status to the Application Choice
- Add rules to the state machine
  - Only applications that are in 'Offer accepted' (`pending_conditions`) and 'Conditions met' (`recruited`) can change to `offer_deferred`
  - From `offer_deferred` we can go back to `recruited` and `pending_conditions` states and `withdrawn` as candidates should be able to withdraw 
- Update status tag components
- Update support documentation
- Update performance statistics
- Hide deferred applications from the API

## Guidance to review

- Check preview for each interface eg: http://localhost:3000/rails/view_components/provider_interface/application_status_tag_component/offer_deferred
- Check provider and candidate application flows: http://localhost:3000/support/provider-flow

- Not sure about the way of hiding deferred offers from the API.  

## Link to Trello card

https://trello.com/c/UeS5yjDh/2689-add-offerdeferred-status

## Things to check


app/services/application_state_change.rb
```diff
@@ -5,8 +5,8 @@ class ApplicationStateChange
   # as they do not need references or the 7-day cooling off period
   STATES_THAT_MAY_BE_SENT_TO_PROVIDER = %i[application_complete unsubmitted].freeze
   STATES_NOT_VISIBLE_TO_PROVIDER = %i[unsubmitted awaiting_references application_complete cancelled].freeze
-  STATES_VISIBLE_TO_PROVIDER = %i[awaiting_provider_decision offer pending_conditions recruited rejected declined withdrawn conditions_not_met offer_withdrawn rejected_at_end_of_cycle].freeze
-  ACCEPTED_STATES = %i[pending_conditions conditions_not_met recruited].freeze
+  STATES_VISIBLE_TO_PROVIDER = %i[awaiting_provider_decision offer pending_conditions recruited rejected declined withdrawn conditions_not_met offer_withdrawn rejected_at_end_of_cycle offer_deferred].freeze
+  ACCEPTED_STATES = %i[pending_conditions conditions_not_met recruited offer_deferred].freeze
```
tijmenb, line 9: Should this go into `OFFERED_STATES` instead? cc @duncanjbrown as it impacts TAD reporting.

app/queries/get_application_choices_for_providers.rb
```diff
@@ -20,6 +20,6 @@ def self.call(providers:)
       .where('courses.provider_id' => providers, 'courses.recruitment_cycle_year' => RecruitmentCycle.visible_years)
       .or(ApplicationChoice.includes(*includes)
         .where('courses.accredited_provider_id' => providers, 'courses.recruitment_cycle_year' => RecruitmentCycle.visible_years))
-      .where('status IN (?)', ApplicationStateChange.states_visible_to_provider).includes([:accredited_provider])
+      .where('status IN (?)', ApplicationStateChange.states_visible_to_provider_in_vendor_api).includes([:accredited_provider])
```
tijmenb, line 23: Out of interest, what's the reason for not showing deferred offers via the API?

duncanjbrown: It's an unexpected state, so potentially a breaking change

app/services/application_state_change.rb
duncanjbrown: In fact this _is_ an accepted state bc it's not possible to enter it without first accepting the offer. However I take your point that TAD might not want to count it. Propose leaving it in here and either defining another constant below (eg `ACCEPTED_STATES_FOR_TAD`) or doing some work in the TAD export to filter it out. Think I prefer the latter?

app/queries/get_application_choices_for_providers.rb
duncanjbrown, line 23: Will this not also hide these applications in Manage? https://github.com/DFE-Digital/apply-for-teacher-training/blob/ceeb842702cd47f3ae94b49cb9194f422c8cf060/app/controllers/provider_interface/application_choices_controller.rb#L12-L14

app/services/application_state_change.rb
AgaDufrat: Thanks, I will have a go at filtering it out of the TAD export.

app/queries/get_application_choices_for_providers.rb
AgaDufrat: ah, yes! I will. Thanks!

duncanjbrown:
It's possible to crash the app by putting an application in `offer_deferred` state and trying to visit it. Fixing.

Screenshot 2020-09-07 at 17 00 09

duncanjbrown (approved):
props @AgaDufrat  — merging in your absence!

PR #2805, opened 2020-09-01
Grant default organisation permissions

steventux:
## Context

Do not merge until we're ready to enable the org to org permissions feature.

We'd like to enable the organisation level permissions feature with minimal disruption to existing provider organisations.

This migration:
- Grants `make_decisions` permission for all existing provider relationships and marks them as set up.
- Grants `manage_organisations` permission to all existing users with the manage_users permission.
- Grants `manage_organisations` permission to for orgs who don't have a managing user, this permission is granted to the user who agreed the DSA.
- Grants `manage_organisations` permission to the first user of any orgs who have an unsigned DSA.

## Guidance to review

- Should these updates be recorded as audit trail entries?
- The first statement to update `ProviderRelationshipPermissions` updates them into an invalid state because any 'set up' permissions should have enable a provider on each permission flag but we are only enabling `make_decisions`. Do we care about this? We're not forcing these providers to  fix this via the setup wizard. 

## Link to Trello card

https://trello.com/c/bQXtxL9Y/2682-data-migration-to-pre-set-all-org-org-permissions

## Things to check


db/migrate/20200922092516_add_default_permissions_for_provider_relationships.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    # Grant manage organisations permissions for all provider users who can currently manage other provider users.
+    ProviderPermissions.where(manage_users: true, manage_organisations: false).update_all(manage_organisations: true)
+
+    # Find any remaining provider organisations without a user with manage_organisations permissions.
+    provider_ids_with_no_manage_org_permissions = ProviderPermissions.where(manage_organisations: false).pluck(:provider_id).uniq
+
+    if provider_ids_with_no_manage_org_permissions.any?
+      # Fetch users who accepted DSA for these unmanaged organisations.
+      provider_user_ids_from_agreements = ProviderAgreement.where(provider_id: provider_ids_with_no_manage_org_permissions).pluck(:provider_user_id).uniq
```
duncanjbrown, line 18: We should double check on prod that all these providers have such a user

duncanjbrown (approved):

db/migrate/20200922092516_add_default_permissions_for_provider_relationships.rb
steventux: @duncanjbrown I can build this into the migration fairly easily.

steventux: On the other hand, so long as the provider permissions for the removed user were also removed, the update statement won't have any effect.

duncanjbrown (approved):
Tested locally — looks good to me!

PR #2807, opened 2020-09-02
Refactor EndOfCycleTimetable

tijmenb:
##  Context

We're trying to improve the testability of the end of cycle changes. This is a refactor to make that easier.

This -hopefully- simplifies the time table code by introducing the concept of a "schedule", which can be either real (the one we'll use on production), or one of the simulated schedules (where we're currently in mid-cycle, course-stop, or between cycles).

The idea is that we can make the schedule a single setting in the database, instead of different feature flags (which are currently weird because they're mutually exclusive - you can't simulate both mid-cycle and between cycles at the same time).

## Guidance to review

Does it make sense?

## Link to Trello card

## Things to check

### DIFF
--- a/app/services/end_of_cycle_timetable.rb
+++ b/app/services/end_of_cycle_timetable.rb
@@ -20,6 +20,10 @@ class EndOfCycleTimetable
     date(:apply_1_deadline)
   end
 
+  def self.stop_applications_to_unavailable_course_options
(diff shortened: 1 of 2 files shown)


stevehook (approved):
:+1: This looks like a step in the right direction. I wonder how it will look next year when we need to consider schedules that span multiple cycles.

PR #2811, opened 2020-09-02
Reuse the wizard's check-answers permissions list on the Organisation edit page

duncanjbrown:
## Context

The "show permissions" page accessible via Organisations > Organisation name is different from the prototype and from the Check answers page in the wizard. There's no need for two different implementations of this permissions list and the other implementation is better, so standardise on that.

**The old version**

Screenshot 2020-09-02 at 11 12 15

**The new version**

Screenshot 2020-09-02 at 11 14 20

The problem with using the wizard's component is that it depends on the wizard!

The first job is to break that dependency, which I've done by having it work on `ProviderRelationshipPermissions` models. I then changed the wizard to be able to emit those without the help of `SetupProviderRelationshipPermissions`. It used to be the case that that class took in the wizard's data and turned it into `ProviderRelationshipPermissions`. Now neither `SetupProviderRelationshipPermissions` nor the view component have any idea whether they're being fed by the wizard or by something else.

Then it was necessary to make the `Change` links work in both contexts. To do this I made a couple of change link builders (ick?) and passed them in.

Finally the component needed to decide whether or not to show the change links (those shown on the Organisations pages are not necessarily editable), so I added a final `editable` parameter.

## Guidance to review

The multiple change links should have correct visually-hidden text and should link to the checkboxes.

The wizard should continue to work.

## Link to Trello card

https://trello.com/c/6ZwbrG2o/2706-bring-the-organisations-edit-page-into-parity-with-the-wizards-check-answers-page

## Things to check


app/services/provider_interface/setup_provider_relationship_permissions.rb
```diff
@@ -2,27 +2,17 @@ module ProviderInterface
...
   class SetupProviderRelationshipPermissions
-    def self.call(permissions_data = {})
+    def self.call(models)
       ProviderRelationshipPermissions.transaction do
-        permissions_data.each do |id, permissions|
-          record = ProviderRelationshipPermissions.find(id)
+        models.map do |record|
           raise PermissionsSetupError, 'Permissions record has already been setup' if record.setup_at.present?
 
-          record.update!(permissions_attributes_for_persistence(permissions))
```
steventux, line 11: Does `save!`  provider the same audit trail info about attributes? I'm gonna assume it does (the docs don't seem to show an example of this).

app/components/provider_interface/provider_relationship_permissions_list_component.html.erb
```diff
@@ -5,17 +5,19 @@
       <%= row[:key] %>
     </dt>
     <dd class="govuk-summary-list__value">
-      <ul class="govuk-list make-decisions">
+    <ul class="govuk-list <%= row[:permission_slug] %>">
```
steventux, line 8: Needs indentation.

app/forms/provider_interface/provider_relationship_permissions_setup_wizard.rb
```diff
@@ -37,6 +37,28 @@ def permissions_for_relationship(relationship_id = current_provider_relationship
       provider_relationship_permissions.fetch(relationship_id.to_s, {})
     end
 
+    # We hold data in the form
+    # { permission_id => { permission1 => [ "training" ], permission2 => [ "training", "ratifying" ] } }
+    #
+    # This method massages it into attrs suitable for ProviderRelationshipPermissions models
+    # and returns those models ready for persistence, either for actually saving or for previewing
+    # on the check-answers page
+    def permissions_for_persistence
```
steventux, line 46: :+1: To this refactor, much better placed here.

steventux (approved):
This looks good to me, I've added a question about audit entries more out of ignorance than anything else.

app/services/provider_interface/setup_provider_relationship_permissions.rb
duncanjbrown: Good question, checking this...

duncanjbrown: Looks OK:

Screenshot 2020-09-07 at 16 23 38

PR #2812, opened 2020-09-02
Make support interface deal better with courses from multiple cycles

tijmenb:
## Context

We want to start syncing courses from the 2021 cycle. This starts to make the support UI deal with that better.

Before:

image

After:

image

## Guidance to review

See individual commits.

## Link to Trello card

https://trello.com/c/lMdqoepW/2069-make-sure-that-syncing-courses-from-the-new-cycle-works

## Things to check


duncanjbrown (approved):
Looks great. Should we make the open courses button only open courses in this cycle? I mean, it doesn't really matter because the queries are scoped, but it feels a bit weird that it's possible to open past and future courses

tijmenb:
@duncanjbrown great idea, I'll do that.

PR #2817, opened 2020-09-02
Adapt apply again for unsuccessful candidates between cycles

stevehook:
## Context

This PR allows candidates who were unsuccessful in a closed recruitment cycle to begin the processing of carrying over their application to the next recruitment cycle before it begins.

- [x] Change `CarryOverApplication` service to work between cycles (previously it only let you carry over after the new cycle started).
- [x] Update the `ApplyAgainBannerComponent` with the new content for this case.
- [x] Renamed `EndOfCycleTimetable.next_cycle_opens` to `apply_reopens`
- [x] System spec to cover the between cycles carry over journey

image

image

image

## Guidance to review

- To test manually you can either mess about with your system clock or activate the simulate between cycles feature flag and set the 2021 recruitment cycle feature flag to inactive.
- Are the tests sufficient?

## Link to Trello card

https://trello.com/c/x6hpRtsT/2042-dev-🚲🔚-adapt-apply-again-journey-for-returners-on-6th-13th

## Things to check

### DIFF
--- a/app/components/candidate_interface/apply_again_banner_component.html.erb
+++ b/app/components/candidate_interface/apply_again_banner_component.html.erb
@@ -1,18 +1,32 @@
 <div class="app-banner">
   <div class="app-banner__message">
     <div class="govuk-heading-m">
-      <p>
-      <% if @application_form.application_choices.all?(&:cancelled?) %>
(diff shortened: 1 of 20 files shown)


tvararu (approved):
Tested, seems to work!

PR #2821, opened 2020-09-03
Add recruitment time inversion to support

tijmenb:
## Context

To accurately test the end-of-cycle features, we need to be able to simulate when the system is in different states (mid cycle, between cycles, partially closed). 

By default, we're on the "Real" schedule:

image

You can then pick another schedule, which will change the times:

image

## Guidance to review

Does this make sense? I've introduced a "SiteSetting" model, which we could use for other things.

## Link to Trello card

## Things to check


app/models/site_setting.rb
```diff
@@ -0,0 +1,13 @@
+class SiteSetting < ApplicationRecord
+  audited
```
stevehook, line 2: :+1:

db/migrate/20200901201150_create_site_settings.rb
```diff
@@ -0,0 +1,11 @@
+class CreateSiteSettings < ActiveRecord::Migration[6.0]
+  def change
+    create_table :site_settings do |t|
+      t.string :name
+      t.text :value
+      t.timestamps
+    end
+
+    add_index :site_settings, :name, unique: true
```
stevehook, line 9: :+1:

app/services/end_of_cycle_timetable.rb
```diff
@@ -82,29 +80,73 @@ def self.schedules
...
+
+      today_is_after_apply_reopens: {
+        apply_1_deadline: 6.days.ago.to_date,
+        stop_applications_to_unavailable_course_options: 5.days.ago.to_date,
+        apply_2_deadline: 4.days.ago.to_date,
+        find_closes: 3.days.ago.to_date,
+        find_reopens: 2.days.ago.to_date,
+        apply_reopens: 1.day.ago.to_date,
       },
     }
```
stevehook, line 151: I guess the reason this hash isn't a constant is that you want the timetable to be relative to `Time.zone.now` rather than set in stone at application start up? Just wondering how expensive it is to calculate all these dates on each lookup.

app/services/end_of_cycle_timetable.rb
```diff
@@ -60,12 +60,10 @@ def self.date(name)
   end
 
   def self.current_cycle_schedule
-    return :real unless HostingEnvironment.test_environment? || HostingEnvironment.sandbox_mode?
-    return :today_is_between_cycles if FeatureFlag.active?(:simulate_time_between_cycles)
-    return :today_is_mid_cycle if FeatureFlag.active?(:simulate_time_mid_cycle)
-    return :today_applications_are_unavailable_to_stopped_courses if FeatureFlag.active?(:simulate_stop_applications_to_unavailable_course_options)
+    # Make sure this setting only has effect on non-production environments
+    return :real if HostingEnvironment.production?
```
stevehook, line 64: Does this force `sandbox` to always use `:real`? (`sandbox` doesn't seem to be a hosting environment so I'm not sure). If so that will disrupt vendor/provider testing.

stevehook (commented):
Much nicer. Tested locally and was able to exercise the carry over between cycles that I worked on recently so all looks good. I just had a question about `sandbox`.

app/services/end_of_cycle_timetable.rb
tijmenb: No, but we should make sure to turn on `today_is_mid_cycle` on sandbox as soon as this is deployed.

stevehook (approved):

PR #2822, opened 2020-09-03
Send a slack notification when a ProviderUser is added

steventux:
## Context

We'd like to have a sense of how users are using access controls and to know when a new user is added, in case we want to follow up with them.

Sends a slack notification when a provider user is invited.

## Link to Trello card

https://trello.com/c/HIpayjgd/2671-send-a-slack-notification-when-a-provideruser-adds-another-provideruser

## Things to check

### DIFF
--- a/app/services/invite_provider_user.rb
+++ b/app/services/invite_provider_user.rb
@@ -14,6 +14,7 @@ class InviteProviderUser
     invite_user_to_dfe_sign_in
 
     send_welcome_email
+    send_slack_notification
   end
 
   def dfe_invite_url
@@ -40,6 +41,14 @@ private
 
   def send_welcome_email
     ProviderMailer.account_created(@provider_user).deliver_later
+  end
+
+  def send_slack_notification
+    providers = @provider_user.providers.map(&:name).to_sentence
+    message = "#{@provider_user.first_name} has been invited to join #{providers}"
+    url = Rails.application.routes.url_helpers.support_interface_provider_user_url(@provider_user)
+
+    SlackNotificationWorker.perform_async(message, url)
   end
 
   def lookup_provider_user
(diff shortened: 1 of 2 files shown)


app/services/invite_provider_user.rb
```diff
@@ -42,6 +43,14 @@ def send_welcome_email
     ProviderMailer.account_created(@provider_user).deliver_later
   end
 
+  def send_slack_notification
+    providers = @provider_user.providers.map(&:name).to_sentence
+    message = "#{@provider_user.first_name} has been invited to join #{providers}"
+    url = Rails.application.routes.url_helpers.provider_interface_provider_user_url(@provider_user)
```
mnacos, line 49: Considering this notification is for internal use, how about linking to a support interface url instead?

app/services/invite_provider_user.rb
```diff
@@ -42,6 +43,14 @@ def send_welcome_email
     ProviderMailer.account_created(@provider_user).deliver_later
   end
 
+  def send_slack_notification
+    providers = @provider_user.providers.map(&:name).to_sentence
+    message = "#{@provider_user.first_name} has been invited to join #{providers}"
```
mnacos, line 48: Do we store who invited each user? The inviter could be as important as the invitee when tracking usage.

mnacos (commented):
Looking good. A couple of quick comments below.

app/services/invite_provider_user.rb
mnacos: The audit trail seems to capture enough information, including whether the inviter is a support or provider user.

app/services/invite_provider_user.rb
steventux: :+1: Thanks @mnacos, amended here https://github.com/DFE-Digital/apply-for-teacher-training/pull/2822/commits/4cb64dfcf1d8c842849cbae2ceaa73bbfdce2727

mnacos (approved):
:rocket:

PR #2826, opened 2020-09-03
Automatically infer the current cycle from the schedule

tijmenb:
## Context

We currently have the recruitment cycle hard coded.

This changes it to be inferred from the cycle schedule that we're on. This means that if you change the cycle in support, you'll be able to test the system using the new recruitment cycle (and see things like filtering in the provider UI change).

## Guidance to review

Makes sense?

## Link to Trello card

## Things to check

### DIFF
--- a/app/models/recruitment_cycle.rb
+++ b/app/models/recruitment_cycle.rb
@@ -1,9 +1,9 @@
 module RecruitmentCycle
   def self.current_year
-    if FeatureFlag.active?('switch_to_next_recruitment_cycle')
-      2021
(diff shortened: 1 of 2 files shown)


steventux (approved):
Makes sense :+1:

stevehook (approved):
:+1: for dropping another feature flag. The only thing I'm unsure about is exactly which date the new cycle starts. I vaguely remember somebody saying it's 6 October.

tvararu (approved):

PR #2832, opened 2020-09-04
Move degree autocompletes into correct form group

paulrobertlloyd:
## Context

Most degree autocompletes, rather than enhance a `select`, instead use a `div` with a data source. This `div` is placed below the form group that contains the original `input`, which is removed. However, in the error state, this means the autocomplete doesn’t inherit the error styling, or appear inside the form group which shows a border down it’s left side. 

Update the JavaScript for these autocompletes to find the form group that is a `previousElementSibling`, and if that contains the `input` we’re replacing, appends the autocomplete container to that form group.

Before:

Screenshot 2020-09-04 at 14 36 07

After:

Screenshot 2020-09-04 at 14 34 53

## Guidance to review

I think it’s probably wise to ensure the form group contains the input we’re replacing, but is using `previousElementSibling` too defensive?

## Link to Trello card

https://trello.com/c/GynRIQqK

## Things to check

### DIFF
--- a/app/frontend/packs/degree-grade-autocomplete.js
+++ b/app/frontend/packs/degree-grade-autocomplete.js
@@ -15,6 +15,12 @@ const initDegreeGradeAutocomplete = () => {
       const container = document.getElementById(containerId);
       if (!container) return;
 
+      // Move autocomplete to the form group containing the input to be replaced
+      const inputFormGroup = container.previousElementSibling
+      if (inputFormGroup.contains(input)) {
+        inputFormGroup.appendChild(container)
+      }
+
       const sourceData = JSON.parse(container.dataset.source);
 
       accessibleAutocomplete({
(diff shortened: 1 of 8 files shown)


app/frontend/packs/degree-grade-autocomplete.js
```diff
@@ -15,6 +15,12 @@ const initDegreeGradeAutocomplete = () => {
       const container = document.getElementById(containerId);
       if (!container) return;
 
+      // Move autocomplete to the form group containing the input to be replaced
+      const inputFormGroup = container.previousElementSibling
+      if (inputFormGroup.contains(input)) {
+        inputFormGroup.appendChild(container)
+      }
```
tvararu, line 22: Could you place this into a `helpers.js` file and import and use it in all the necessary components?

tvararu (approved):

paulrobertlloyd:
@tvararu Have refactored this change into a new helper, `accessibleAutocompleteFromSource`. Removes a lot of the repetition that existed previously, and include the previous fix made in this PR. Naming things is hard etc, but is this obvious? (Have added JSDoc for the helper function).

tvararu (approved):

PR #2833, opened 2020-09-04
Correct input width for IELTS score

paulrobertlloyd:
## Context

The autocomplete for IELTS band does not give an indication of the length of the expected value.

Before:

Screenshot 2020-09-04 at 15 52 45

After:

Screenshot 2020-09-04 at 15 52 20

## Guidance to review

This works, but will break should this page contain another autocomplete. Do we worry about that at this stage?

## Things to check

### DIFF
--- a/app/frontend/packs/ielts-band-score-autocomplete.js
+++ b/app/frontend/packs/ielts-band-score-autocomplete.js
@@ -19,6 +19,9 @@ const initIeltsBandScoreAutocomplete = () => {
         showNoOptionsFound: true,
         confirmOnBlur: false
       });
+
+      const accessibleAutocompleteWrapper = document.querySelector(".govuk-form-group--ielts-band-score .autocomplete__wrapper");
+      accessibleAutocompleteWrapper.classList.add("govuk-input--width-5");
     });
   } catch (err) {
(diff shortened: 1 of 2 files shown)


app/frontend/packs/ielts-band-score-autocomplete.js
```diff
@@ -19,6 +19,9 @@ const initIeltsBandScoreAutocomplete = () => {
         showNoOptionsFound: true,
         confirmOnBlur: false
       });
+
+      const accessibleAutocompleteWrapper = document.querySelector(".autocomplete__wrapper");
```
stevehook, line 23: I don't think it's an issue on the page where it's used but should/could this selector be more specific to avoid picking up other autocompletes (maybe in the future)?

stevehook (approved):
:+1: Worked for me on review app. One question, don't think it's a blocker.

app/frontend/packs/ielts-band-score-autocomplete.js
paulrobertlloyd: @stevehook Probably not a blocker, but have updated to add a class to the parent form group, and made this a more specific selector.

PR #2840, opened 2020-09-07
Change rejected_at_end_of_cycle to applications_not_sent

davidgisbey:
## Context

The state for application choices being rejected at the end of a cycle should be called application_not_sent rather than rejected at the end of the cycle. There are also a few minor UI tweaks on the application dashboard

- Refactor rejected_at_end_of_cycle to application_not_sent
- Add a description underneath the status informing candidates why their choice was rejected
- Make the status pink

| Before | After |
|-----------|-----------|
| image | image|

## Link to Trello card

https://trello.com/c/51nJicmW/2080-%F0%9F%9A%B2%F0%9F%94%9A-dev-add-new-application-not-sent-status

## Things to check

### DIFF
--- a/app/components/candidate_interface/application_complete_content_component.html.erb
+++ b/app/components/candidate_interface/application_complete_content_component.html.erb
@@ -19,7 +19,7 @@
   <h2 class="govuk-heading-m govuk-!-font-size-27">
     <%= t('application_complete.dashboard.all_withdrawn') %>
   </h2>
-<% elsif application_not_sent? %>
+<% elsif all_applications_not_sent? %>
   <h2 class="govuk-heading-m govuk-!-font-size-27">
     <%= t('application_complete.dashboard.application_not_sent', count: choice_count) %>
   </h2>
--- a/app/components/candidate_interface/application_complete_content_component.html.erb
+++ b/app/components/candidate_interface/application_complete_content_component.html.erb
@@ -19,6 +19,10 @@
   <h2 class="govuk-heading-m govuk-!-font-size-27">
     <%= t('application_complete.dashboard.all_withdrawn') %>
   </h2>
+<% elsif application_not_sent? %>
+  <h2 class="govuk-heading-m govuk-!-font-size-27">
+    <%= t('application_complete.dashboard.application_not_sent', count: choice_count) %>
+  </h2>
 <% elsif all_provider_decisions_made? %>
   <h2 class="govuk-heading-m govuk-!-font-size-27">
     <%= t('application_complete.dashboard.all_provider_decisions_made', count: choice_count) %>
--- a/app/components/candidate_interface/course_choices_review_component.rb
+++ b/app/components/candidate_interface/course_choices_review_component.rb
@@ -157,12 +157,7 @@ module CandidateInterface
     end
 
     def status_row_value(application_choice)
-      if ApplicationStatusTagComponent.new(application_choice: application_choice).status == 'application_not_sent'
-        render(ApplicationStatusTagComponent.new(application_choice: application_choice)) +
-          tag.p('Your application was not sent for this course because references were not given before the deadline.', class: 'govuk-body govuk-body govuk-!-margin-top-2')
-      else
-        render(ApplicationStatusTagComponent.new(application_choice: application_choice))
-      end
+      render(ApplicationStatusTagComponent.new(application_choice: application_choice))
     end
 
     def rejection_reason_row(application_choice)
--- a/app/components/candidate_interface/course_choices_review_component.rb
+++ b/app/components/candidate_interface/course_choices_review_component.rb
@@ -152,8 +152,17 @@ module CandidateInterface
     def status_row(application_choice)
       {
         key: 'Status',
-        value: render(ApplicationStatusTagComponent.new(application_choice: application_choice)),
+        value: status_row_value(application_choice),
       }
+    end
+
+    def status_row_value(application_choice)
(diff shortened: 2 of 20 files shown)


app/components/candidate_interface/course_choices_review_component.rb
```diff
@@ -152,10 +152,19 @@ def start_date_row(course)
     def status_row(application_choice)
       {
         key: 'Status',
-        value: render(ApplicationStatusTagComponent.new(application_choice: application_choice)),
+        value: status_row_value(application_choice),
       }
     end
 
+    def status_row_value(application_choice)
+      if ApplicationStatusTagComponent.new(application_choice: application_choice).status == 'application_not_sent'
```
tijmenb, line 160: Would this work?

```suggestion
      if application_choice.application_not_sent?
```

app/components/candidate_interface/course_choices_review_component.rb
```diff
@@ -152,10 +152,19 @@ def start_date_row(course)
...
+        render(ApplicationStatusTagComponent.new(application_choice: application_choice)) +
+          tag.p('Your application was not sent for this course because references were not given before the deadline.', class: 'govuk-body govuk-body govuk-!-margin-top-2')
```
tijmenb, line 162: Did you consider making the extra `p` part of the `ApplicationStatusTagComponent`?

app/models/application_form.rb
```diff
@@ -119,6 +119,10 @@ def any_offers?
     application_choices.map.any?(&:offer?)
   end
 
+  def application_not_sent?
```
tijmenb, line 122: Perhaps this, to differentiate it from `ApplicationChoice#application_not_sent?`

```suggestion
  def all_applications_not_sent?
```

app/models/application_form.rb
```diff
@@ -119,6 +119,10 @@ def any_offers?
...
+    application_choices.map.all?(&:application_not_sent?)
```
tijmenb, line 123: I don't think you need the map:

```suggestion
    application_choices.all?(&:application_not_sent?)
```

app/services/application_state_change.rb
```diff
@@ -5,11 +5,11 @@ class ApplicationStateChange
   # as they do not need references or the 7-day cooling off period
   STATES_THAT_MAY_BE_SENT_TO_PROVIDER = %i[application_complete unsubmitted].freeze
   STATES_NOT_VISIBLE_TO_PROVIDER = %i[unsubmitted awaiting_references application_complete cancelled].freeze
-  STATES_VISIBLE_TO_PROVIDER = %i[awaiting_provider_decision offer pending_conditions recruited rejected declined withdrawn conditions_not_met offer_withdrawn rejected_at_end_of_cycle].freeze
+  STATES_VISIBLE_TO_PROVIDER = %i[awaiting_provider_decision offer pending_conditions recruited rejected declined withdrawn conditions_not_met offer_withdrawn application_not_sent].freeze
```
tijmenb, line 8: I think this is wrong? The application is never sent to the provider, so it should not be visible to providers?

app/components/candidate_interface/application_complete_content_component.html.erb
```diff
@@ -19,6 +19,10 @@
   <h2 class="govuk-heading-m govuk-!-font-size-27">
     <%= t('application_complete.dashboard.all_withdrawn') %>
   </h2>
+<% elsif application_not_sent? %>
```
tijmenb, line 22: This won't show if one of the choices is cancelled or withdrawn. Is that the intended behaviour?

app/components/candidate_interface/course_choices_review_component.rb
davidgisbey: I did not, but it definitely should be. You're right 👍

app/services/application_state_change.rb
davidgisbey: There was some discussion with Duncan and Theo about whether this should be exposed. They decided it should be. Happy to discuss with them 👍

stevehook:
@davidgisbey just FYI I'm going to use `application_not_sent` for un-submitted applications that reach EoC (https://github.com/DFE-Digital/apply-for-teacher-training/pull/2853). Might also be a good idea if we do the same for applications 'cancelled' because courses are unavailable when references come in between cycles (https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/services/send_applications_to_provider.rb#L11).

app/components/candidate_interface/application_complete_content_component.html.erb
davidgisbey: Good point. Have fixed.

app/services/application_state_change.rb
tijmenb: Pinging @duncanjbrown @tvararu. Feels like we shouldn't expose these to providers?

davidgisbey: Looking back at this I agree it should be treated the same as unsubmitted and therefore should not be visible.

stevehook (approved):
I tried out the review app, and was able to move some applications to the new status by manually running `RejectAwaitingReferencesCourseChoicesWorker` via rails console. These seem to show up correctly for candidates and support users. I looked in the provider interface and didn't see them, then realised I shouldn't see them, so all good there too. So looks good to me if @tijmenb is happy?

davidgisbey:
I'm pretty sure he will be as I've actioned his change 👍

tijmenb:
I indeed was! 👍

PR #2841, opened 2020-09-07
Use greater or equal than in query to fix 'Less than 1 day to respond'

mnacos:
## Context

When there's less than one day to respond, the 'days to respond' label does not appear in the applications list.

image

Fix it.

image

## Guidance to review

Inspect changes.

## Link to Trello card

No card

## Things to check


tijmenb (commented):
Could we add a test for this?

duncanjbrown (approved):

PR #2844, opened 2020-09-07
Add provider satisfaction survey to footer and header

duncanjbrown:
## Context

Providers get a link to a survey instead of an email address for giving feedback.

- refactor the PhaseBanner so it, not HostingEnvironment (!), is responsible for rendering its contents
- let PhaseBanner take the feedback link as an optional param and test it
- pass in `ProviderInterface::FEEDBACK_LINK` when we're building provider headers
- add the days we offer support to both candidate and provider headers (it was missing before)

## Guidance to review

Hopefully straightforward 🤞

## Link to Trello card

https://trello.com/c/MfZeMRX8/2677-add-satisfaction-survey-into-footer-and-header-of-live-service

## Things to check

### DIFF
--- a/app/components/phase_banner.html.erb
+++ b/app/components/phase_banner.html.erb
@@ -4,7 +4,7 @@
       <%= HostingEnvironment.phase %>
     </strong>
     <span class="govuk-phase-banner__text">
-      <%= HostingEnvironment.phase_banner_text %>
+      <%= text %>
     </span>
     </p>
   </div>
--- a/app/components/phase_banner.rb
+++ b/app/components/phase_banner.rb
@@ -13,7 +13,7 @@ class PhaseBanner < ViewComponent::Base
 
     case HostingEnvironment.environment_name
     when 'production'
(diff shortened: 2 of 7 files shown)


config/locales/en.yml
```diff
@@ -146,6 +146,7 @@ en:
   layout:
     support:
       title: Get support
+      support_days: Monday to Friday (except public holidays)
```
paulrobertlloyd, line 149: @duncanjbrown I removed this because it was felt it was unnecessary given the SLA gives a response time that is working days. See [this thread on Slack](https://ukgovernmentdfe.slack.com/archives/CN9Q1VB2M/p1599115518014200?thread_ts=1598949188.002000&cid=CN9Q1VB2M).

duncanjbrown: Oops — thanks! Will remove

duncanjbrown: Gone

benweedon-DfE (approved):
Looks good to me!

tijmenb (approved):
🥇 Very nice, excellent refactoring as well.

PR #2845, opened 2020-09-07
Capitalise environment name tag

paulrobertlloyd:
## Context

Just a smol thing…

In the environment tag, capitalise the word. So `beta => Beta`, `sandbox => Sandbox` (but `qa => QA`)

### DIFF
--- a/app/lib/hosting_environment.rb
+++ b/app/lib/hosting_environment.rb
@@ -21,9 +21,9 @@ module HostingEnvironment
 
   def self.phase
     if production?
-      'beta'
+      'Beta'
     else
-      environment_name
+      environment_name.capitalize
     end
   end
 
--- a/config/initializers/inflector.rb
+++ b/config/initializers/inflector.rb
@@ -4,6 +4,7 @@ ActiveSupport::Inflector.inflections(:en) do |inflect|
   inflect.acronym('UCAS')
   inflect.acronym('TAD')
   inflect.acronym('CSV')
+  inflect.acronym('QA')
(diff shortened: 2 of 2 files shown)


app/lib/hosting_environment.rb
```diff
@@ -21,9 +21,13 @@ def self.hostname
 
   def self.phase
     if production?
-      'beta'
+      'Beta'
+    elsif sandbox_mode?
+      'Sandbox'
```
tijmenb, line 26: This should stay as is was. We sometimes put other environments in sandbox _mode_, and it shouldn't be confused with the sandbox environment.

app/lib/hosting_environment.rb
```diff
@@ -21,9 +21,13 @@ def self.hostname
...
+    elsif qa?
+      'QA'
```
tijmenb, line 28: You can avoid this conditional by adding `QA` to config/initializers/inflector.rb.

paulrobertlloyd: Oooh, nice!

app/lib/hosting_environment.rb
paulrobertlloyd: Reverted this change. I wonder… will capitalising `environment_name.capitalize` mean `sandbox` appears capitalised?

tijmenb (approved):

app/lib/hosting_environment.rb
tijmenb: Yes it will!

PR #2848, opened 2020-09-07
Simplify query necessary to check permissions  on application

steventux:
## Context

[We have Sentry errors for query timeouts coming from the same authorisation check](https://sentry.io/organizations/dfe-bat/issues/1882290403/?project=1765973&referrer=slack)

The underlying SQL performs a lot of LEFT OUTER JOIN statements, more than necessary to check whether an actor is eligible to make decisions for a given application.

Simplify the underlying authorisation method to use a simplified version of the application check. This uses minimal joins to courses.

### Before (dev environment)

image

### After (dev environment)

image

## Link to Trello card

https://trello.com/c/5ARv2hji/2727-investigate-potential-slow-query-performance-issue-with-getapplicationchoicesforproviders

## Things to check

### DIFF
--- a/app/queries/get_application_choices_for_providers.rb
+++ b/app/queries/get_application_choices_for_providers.rb
@@ -1,20 +1,18 @@
 class GetApplicationChoicesForProviders
-  def self.call(providers:, vendor_api: false)
+  DEFAULT_INCLUDES = [
+    :accredited_provider,
+    :offered_course_option,
+    :provider,
+    :site,
+    application_form: %i[candidate application_references application_work_experiences application_volunteering_experiences application_qualifications],
+    course: %i[provider],
+    course_option: [{ course: %i[provider] }, :site],
+  ].freeze
+
+  def self.call(providers:, vendor_api: false, includes: DEFAULT_INCLUDES)
     providers = Array.wrap(providers).select(&:present?)
 
     raise MissingProvider if providers.none?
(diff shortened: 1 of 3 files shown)


duncanjbrown (commented):
This looks a lot simpler! It looks like the `includes` part is the only thing that needs to vary from vanilaa `GetApplicationChoicesForProviders`. Would it make sense to keep what's there as a default, but in this case pass the `includes` as an argument?

steventux:
@duncanjbrown :+1: your suggestion is a nicer approach and keeps the complexities of the query in one place, I've amended here: https://github.com/DFE-Digital/apply-for-teacher-training/pull/2848/commits/f403953db9ea6e8d746a0b98d8f6798fd9c1f106
I've assumed that folding the `:accredited_provider` include into the main set is OK, also I've de-duped the repetition of `:course` and `:course_option` in the default includes.

duncanjbrown (approved):
LVGTM. I bet we can shrink this some more (or at least scope the mega-include more tightly) if we come back to it

steventux:
@duncanjbrown I've deduped `:application_form` in a similar manner and amended the query spec to assert we eager load based on the includes https://github.com/DFE-Digital/apply-for-teacher-training/pull/2848/commits/cff9341df4a2bd4a7885382643f02acdc5b56063

PR #2853, opened 2020-09-08
Carry over un-submitted applications manually after EoC

stevehook:
## Context

Candidates need to be able to manually continue an un-submitted application when the next recruitment cycle opens.

Ready for review but probably needs rebasing after https://github.com/DFE-Digital/apply-for-teacher-training/pull/2840 gets merged.

image
image
image

- [x] Application course choices for un-submitted applications are cancelled at the end of the recruitment cycle by a new `CancelUnsubmittedApplicationsWorker`.
- [x] Use the existing carry-over implementation for un-submitted applications with application choices.
- [x] Add `ApplicationForm#recruitment_cycle_year` (earlier PR). All applications are associated with a recruitment cycle year (even if they do not have any application choices). This enables us to trigger the carry-over for such applications so that they work the same way as all the others.
- [x] Drop the scheduled one-off job (in `clock`) to automatically carry over un-submitted applications (with application choices) automatically.
- [x] Refactor to extract `CarryOverBannerComponent` from `ApplyAgainBannerComponent`.
- [x] Alter logic for displaying the carry over banner to also work for un-submitted applications with no course choices.
- [x] Skip the application page and go straight to the carry over interstitial when candidates log back into the application after the new cycle opens.

Not covered in this PR:
- Add buttons to support interface (under the Tasks menu) for all one-off EoC jobs rather than running them as scheduled jobs (`clock`).
- Un-submitted applications being carried over between cycles (only addressing what happens after 13 October here).

## Guidance to review

- To test this you need a system in the ‘Today is after apply reopens’ schedule and with the 2021 recruitment cycle activated feature flag active (though this is to be removed soon). Try it with un-submitted applications from 2020 both with and without application choices.
- Are the specs adequate?
- Does the banner refactoring make sense?

## Link to Trello card

https://trello.com/c/DagOgy6H/2077-unsubmitted-applications-should-be-carried-over-to-next-recruitment-cycle-manually
## Things to check


app/services/application_state_change.rb
```diff
@@ -29,6 +29,7 @@ def initialize(application_choice)
     state :unsubmitted do
       event :submit, transitions_to: :awaiting_references
       event :send_to_provider, transitions_to: :awaiting_provider_decision
+      event :cancel, transitions_to: :cancelled
```
tijmenb, line 32: I wonder if this should be either it's own state (`never_submitted`) or `application_not_sent` (https://github.com/DFE-Digital/apply-for-teacher-training/pull/2840). What do you think?

stevehook: I think `application_not_sent` would make more sense. We don't want to get these mixed up with 'real' cancelled applications from a reporting point of view. I don't _think_ we need a brand new state because these applications would be easily distinguished from the other `application_not_sent` ones as they are un-submitted.

The only reason for going with `cancelled` was to be consistent with https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/services/send_applications_to_provider.rb#L11 . Maybe we should change that too?

tvararu (approved):

PR #2855, opened 2020-09-08
Show fewer page links when paginating above a threshold

mnacos:
## Context

As the number of applications on the system grows, we are generating too many page links at the bottom of the page, which do not add any value and force the paginator component to wrap lines.
image

Enforce a limit to the number of page links included at the bottom of the page (`KAMINARI_LINKS_LIMIT`). This is set to five page links, which means we will never display more than six (explanation below).

image
image
image
image
image
image
image
image
image
image
image

## Guidance to review

The `kaminari` gem, which handles our pagination, supports configuring `window`, `left` and `right` parameters to control how page links are displayed at the bottom of the page. It is also a bit opinionated and refuses to leave a gap / show an ellipsis where a single number fits. Thus, at the transition points (pages 4 and 8 in the example above) we show an additional number compared to our spec (`1 2 3 4 5 ...` instead of `1 ... 3 4 5 ...`).

Inspect the automated tests. Increase the total number of pages available on your local environment by running `GenerateTestApplications.new.perform` or using the support console.

## Link to Trello card

https://trello.com/c/SowrrBPb/2398-pagination-behaviour

## Things to check

### DIFF
--- a/app/components/paginator_component.html.erb
+++ b/app/components/paginator_component.html.erb
@@ -1,9 +1,13 @@
 <nav class="moj-pagination" role="navigation" aria-label="pager">
   <p class="govuk-visually-hidden" aria-labelledby="pagination-label">Pagination navigation</p>
 
-  <%= paginate @scope %>
+  <%= paginate @scope, paginate_configuration %>
 
   <p class="moj-pagination__results govuk-body">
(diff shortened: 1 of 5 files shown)


adamsilver (approved):

duncanjbrown (approved):
This works well! Bit sad we have to take on extra logic to support this. Approving — maybe there's a better gem out there

PR #2861, opened 2020-09-08
Support ability to filter 'offer_deferred' with test data and DeferOffer service

mnacos:
## Context

Because of we rely on `states_visible_to_provider` (see https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/models/provider_interface/provider_applications_page_state.rb#L79) we can already filter applications with 'offer_deferred' status. This PR just adds some deferred applications to local test data and review apps.

Add `:offer_deferred` applications to `GenerateTestApplications` and `spec/factories.rb`.

Add `offer_deferred_at` database field to `ApplicationChoice`.

Add minimal `DeferOffer` service.

After dropping the database and re-creating it (fresh test data):
image

## Guidance to review

Run the tests. Drop your database, re-create it, and re-populate it using `bundle exec rake setup_local_dev_data`.

## Link to Trello card

https://trello.com/c/Ixu6aApj/2702-providers-can-filter-by-deferred-status

## Things to check

- [ ] This code does not rely on migrations in the same Pull Request -- IT DOES but it's ok
- [ ] If this code includes a migration adding or changing columns, it also backfills existing records for consistency -- IT DOESN'T but we don't offer the ability to defer offers yet, so it's ok


app/lib/test_applications.rb
```diff
@@ -174,6 +174,9 @@ def put_application_choice_in_state(choice, state)
     when :offer_withdrawn
       make_offer(choice)
       withdraw_offer(choice)
+    when :offer_deferred
+      make_offer(choice)
```
duncanjbrown, line 178: I think this breaks the rules of the state machine — should we `accept_offer` first? And maybe we should sometimes have `recruited` applications.

app/lib/test_applications.rb
```diff
@@ -251,6 +254,15 @@ def withdraw_offer(choice)
     choice.audits.last&.update_columns(created_at: time)
   end
 
+  def defer_offer(choice)
+    as_provider_user(choice) do
+      fast_forward(1..3)
+      # TODO: should be calling a service for this - not yet implemented
+      choice.update_columns(status: 'offer_deferred', updated_at: time)
```
duncanjbrown, line 261: Maybe we should just implement the service now including `ApplicationStateChange`, however thin it is, then this code doesn't need to change again?

app/lib/test_applications.rb
mnacos: Well spotted. We already have `recruited` applications in the test data (showing up as 'Conditions met')

app/lib/test_applications.rb
mnacos: I've added a minimal `DeferOffer` service.

app/lib/test_applications.rb
mnacos: Ah, you mean deferred applications _via_ `recruited`. Will add.

duncanjbrown (approved):

PR #2870, opened 2020-09-09
Set sync_courses: true for fake providers

duncanjbrown:
## Context

When we create fake providers the courses are not automatically opened because the provider does not have `sync_courses` set to true. This means there's a manual step to open courses, and it's easy to forget.

- always mark fake providers with `sync_courses` true

## Link to Trello card

Came up as part of https://trello.com/c/3iC0aiNg/741-investigate-bug-report-from-ucb and also https://ukgovernmentdfe.slack.com/archives/CTHFLRFHB/p1599573003004300

## Things to check


mnacos (approved):
:rocket:

PR #2871, opened 2020-09-09
Fix training with a disability section to route to review page

george-bit:
## Context

When Training with a Disability have been filled out but section is not marked as complete the Application presenter should route the candidate to the review page rather than to the edit path.

Tweaks to the Application Form presenter and Training with a Disability show view.

## Guidance to review

Check that behaviour is now as required.

## Link to Trello card

https://trello.com/c/t98DVkIA/2059-returning-to-some-completed-sections-not-marked-as-complete-takes-you-to-the-wrong-page

## Things to check

### DIFF
--- a/spec/presenters/candidate_interface/application_form_presenter_spec.rb
+++ b/spec/presenters/candidate_interface/application_form_presenter_spec.rb
@@ -191,14 +191,14 @@ RSpec.describe CandidateInterface::ApplicationFormPresenter do
 
   describe '#training_with_a_disability_completed?' do
     it 'returns true if training with a disabilitty section is completed' do
-      application_form = FactoryBot.build(:application_form, training_with_a_disability_completed: true)
+      application_form = FactoryBot.build(:completed_application_form)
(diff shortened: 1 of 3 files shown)


spec/presenters/candidate_interface/application_form_presenter_spec.rb
```diff
@@ -189,6 +205,22 @@
...
+
+      expect(presenter).to be_training_with_a_disability_valid
+    end
+
+    it 'returns true if training with a disability section is incomplete' do
+      application_form = FactoryBot.build(:application_form, training_with_a_disability_completed: false, disclose_disability: true)
+      presenter = CandidateInterface::ApplicationFormPresenter.new(application_form)
+
+      expect(presenter).to be_training_with_a_disability_valid
+    end
```
tvararu, line 221: Could you add a `returns true...` test similar to the suggestion in the prior PR?

george-bit: Test updated

PR #2876, opened 2020-09-10
Trigger one-off EoC jobs manually via support interface

stevehook:
## Context

There are a couple of one-off background workers that need to be run at a certain time in the EoC timetable. Rather than have these run automatically in the middle of the night we want to trigger them manually so that they can be more easily monitored.

Leaving in draft for now because it depends on https://github.com/DFE-Digital/apply-for-teacher-training/pull/2853

- [x] Add button to the support interface tasks page for the _Cancel unsubmitted applications that reach end-of-cycle_ job
- [x] Add button to the support interface tasks page for the _Reject applications awaiting references at end-of-cycle_ job
- [x] System specs to cover the above
- [x] Some fixes to `CancelUnsubmittedApplicationsWorker`
- [x] Add a confirmation page so we don't accidentally trigger this job.
- [x] Changes `GetPreviousCyclesAwaitingReferencesApplication` to find applications based on `ApplicationForm#recruitment_cycle_year` rather than `Course#recruitment_cycle_year`

image

## Guidance to review

- To manually test your need to use the _Today is after apply 2 deadline passed_ schedule in `/support/cycles`. Unsuccessful applications and those awaiting references, from the 2020 cycle, should then be moved to the `application_not_sent` status.
- Should we have a confirmation page for each of these actions?

## Link to Trello card

https://trello.com/c/DagOgy6H/2077-unsubmitted-applications-should-be-carried-over-to-next-recruitment-cycle-manually

## Things to check


config/routes.rb
```diff
@@ -745,6 +745,8 @@
     post '/tasks/create-fake-provider' => 'tasks#create_fake_provider'
     post '/tasks/:task' => 'tasks#run', as: :run_task
     get '/tasks/confirm_delete_test_applications' => 'tasks#confirm_delete_test_applications', as: :confirm_delete_test_applications
+    get '/tasks/confirm_cancel_applications_at_end_of_cycle' => 'tasks#confirm_cancel_applications_at_end_of_cycle', as: :confirm_cancel_applications_at_end_of_cycle
+    get '/tasks/confirm_reject_applications_awaiting_references_at_end_of_cycle' => 'tasks#confirm_reject_applications_awaiting_references_at_end_of_cycle', as: :confirm_reject_applications_awaiting_references_at_end_of_cycle
```
tvararu, line 749: Could these be hyphenated instead?

tvararu (approved):

config/routes.rb
stevehook: oops, fixed

PR #2880, opened 2020-09-10
don't trigger build if another one is inprogress

vigneshmsft:
## Context
Concurrent builds could cause the swap operation to be reversed and the production slot to be stopped.
This is a less probable risk when deploying to staging/production env as the release pipeline is triggered manually 
https://ukgovernmentdfe.slack.com/archives/CN1MCQCHZ/p1599643603067100

When a build is triggered from the `gems-node-packages` pipeline, only trigger the down stream pipeline if there isn't an inprogress build for the same commit.

Start the production slot after a swap operation as a fail-safe to prevent production slot to be stopped in the event of concurrent builds.

## Link to Trello card

https://trello.com/c/CQ77rcgp/2743-pipeline-improvement-start-production-slot-before-deployment

## Things to check

### DIFF
--- a/azure/pipelines/build-gems-node-modules.yml
+++ b/azure/pipelines/build-gems-node-modules.yml
@@ -107,6 +107,13 @@ steps:
 
 - powershell: |
     $azureDevOpsAuthorizationHeader = @{ Authorization = "Bearer $env:SYSTEM_ACCESSTOKEN" }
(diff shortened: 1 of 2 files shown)


saliceti (approved):
ok

tvararu (approved):

PR #2884, opened 2020-09-10
Show EFL qualifications on Provider interface

malcolmbaig:
## Context

Providers need to be able to see an international candidate's EFL qualification.

- Add a qualification card component that summarises the detail captured
  by the application's english_proficiency record.
- Only render the component if an english_proficiency record is actually
  present, and the candidate nationality is not in our list of
  "English-speaking" nationalities.

Screenshot_20200914_154755

## Guidance to review
- Per-commit separates out the key parts of the implementation

## Link to Trello card
https://trello.com/c/mHiHJUrd

## Things to check


tvararu (approved):
Very clean 👍

app/components/efl_qualification_card_component.html.erb
```diff
@@ -0,0 +1,19 @@
+<% if show_component && english_proficiency.present? %>
+  <div class='app-qualification-card'>
+    <h3 class='govuk-heading-m'> English as a foreign language</h2>
+
+    <p class='govuk-body'><%= qualification_status %></p>
+
+    <% if english_proficiency.has_qualification?%>
```
duncanjbrown, line 7:
```suggestion
    <% if english_proficiency.has_qualification? %>
```

app/components/efl_qualification_card_component.rb
```diff
@@ -0,0 +1,32 @@
+class EflQualificationCardComponent < ViewComponent::Base
+  attr_reader :application_form
+
+  def initialize(application_form)
+    @application_form = application_form
+  end
+
+  def show_component
```
duncanjbrown, line 8: This is built-in! https://github.com/github/view_component#conditional-rendering

duncanjbrown (commented):
Looks very good to me! Should we also remove the "Is English your main language?" row from the "Language skills" section?

app/components/efl_qualification_card_component.rb
malcolmbaig: Nice, thanks for the link! Will rework this.

malcolmbaig:
> Should we also remove the "Is English your main language?" row from the "Language skills" section?

@duncanjbrown Good point - I wonder if we should go a step further and only show the "Language skills" section if the candidate actually filled it in (given we've been hiding that part of the form since the EFL feature was made active in prod).

If we were to hide the "Is English your main language?" row only, here's what a Provider would see:

### Scenario: Candidate \*did not\* fill in Personal Details>Languages

We infer the answer to "Is English your main language?" based on their nationality and EFL answers, so they see one of the following:

Screenshot_20200911_115348
Screenshot_20200911_115458

### Scenario: Candidate filled in Personal Details>Languages

As per current behavior, we display one of these depending on their answer to "Is English your main language?"
Screenshot_20200911_115743
Screenshot_20200911_115714


Seems like it might make sense to hide the section altogether in the first scenario. @paulrobertlloyd WDYT?

paulrobertlloyd:
> Seems like it might make sense to hide the section altogether in the first scenario. @paulrobertlloyd WDYT?

Yes, I think we should only show those questions if you have explicitly answered them.

Arguably, we might not need to support these questions at all, but there’s probably a broader discussion to be had about supporting previously collected answers that we no longer ask current candidates, especially as we enter a new cycle.

malcolmbaig:
@duncanjbrown @paulrobertlloyd this is now ready for a final review.

duncanjbrown (approved):
Looks good to me!

paulrobertlloyd:
So… after some UR on the qualifications, I updated the qualifications card design to ensure the different values had labels. But for some reason I didn’t update the prototype to apply these changes to the EFL qualification. 😬 I’ve committed a change that adds in these latest changes.

However, the side-effect of this is, in the EFL qualification card component can we:

* Change `Score or grade` to say `Overall band score` if IELTS, `Total score` if TOEFL (else default to `Score or grade`)
* Change `Reference number` to say `TRF number` if IELTS, `Registration number` if TOEFL, else not show this information

malcolmbaig:
>
    Change Score or grade to say Overall band score if IELTS, Total score if TOEFL (else default to Score or grade)Change Reference number to say TRF number if IELTS, Registration number if TOEFL, else not show this information

malcolmbaig:
> Change Score or grade to say Overall band score if IELTS, Total score if TOEFL (else default to Score or grade)
   Change Reference number to say TRF number if IELTS, Registration number if TOEFL, else not show this information

@paulrobertlloyd done!

app/components/efl_qualification_card_component.html.erb
```diff
@@ -0,0 +1,27 @@
...
+        <div class="govuk-grid-column-one-third">
+          <div class="app-card app-card--outline">
+            <h3 class="govuk-heading-s govuk-!-margin-bottom-1"><%= name %></h3>
+            <dl class="app-qualification">
+              <dt class="app-qualification__key govuk-visually-hidden">Awarded</dt>
+              <dd class="app-qualification__value app-qualification__value--caption"><%= award_year %></dd>
+              <dt class="app-qualification__key"><%= grade_title %></dt>
+              <dd class="app-qualification__value"><%= grade %></dd>
+              <dt class="app-qualification__key"><%= reference_number_title %></dt>
+              <dd class="app-qualification__value"><%= unique_reference_number %></dd>
```
paulrobertlloyd, line 17: @malcolmbaig All of this HTML should not appear if there is no reference number, not just the text within in it (affects spacing in card, otherwise). 

```suggestion
              <dt class="app-qualification__key"><%= reference_number_title %></dt>
              <dd class="app-qualification__value"><%= unique_reference_number %></dd>
```

paulrobertlloyd (changes requested):
Looking good! Just one comment around showing the reference number, then :shipit: 

There’s also a question of spacing between the different parts of the qualifications section, but we can review that once the other structured quals are in place.

paulrobertlloyd (approved):
:shipit:

PR #2886, opened 2020-09-11
Set references_completed correctly when carrying over application

tvararu:
## Context

If we change the number of references on the application when carrying it over, we should set it to false.

Set `references_completed` correctly when carrying over application.

## Guidance to review

Does this make sense?

## Link to Trello card

None, spotted while working on other things.

## Things to check

### DIFF
--- a/app/services/duplicate_application.rb
+++ b/app/services/duplicate_application.rb
@@ -45,6 +45,10 @@ class DuplicateApplication
       )
     end
 
+    if new_application_form.can_add_reference?
+      new_application_form.update! references_completed: false
+    end
+
     original_application_form.application_work_history_breaks.each do |w|
       new_application_form.application_work_history_breaks.create!(
         w.attributes.except(*IGNORED_CHILD_ATTRIBUTES),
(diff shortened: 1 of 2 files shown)


app/services/duplicate_application.rb
```diff
@@ -45,6 +45,11 @@ def duplicate
       )
     end
 
+    number_of_references_has_changed = original_application_form.application_references.size != new_application_form.application_references.size
+    if number_of_references_has_changed
+      new_application_form.update! references_completed: false
```
stevehook, line 50: If the original had say 1 declined reference and 2 approved references then we'd strip out the declined one leaving the 2 approved ones. So even if the size changes I think you could end up in a 'completed' state. Could we check for the presence of at least 2 references in the new application form?

tvararu: That makes sense!

stevehook: Hadn't thought of using `can_add_reference?` :+1:

stevehook (approved):

PR #2902, opened 2020-09-15
Check if find is down only once on providers page

tijmenb:
## Context

The list of available courses on Apply (https://www.apply-for-teacher-training.service.gov.uk/candidate/providers) is a bit slow according to Skylight:

https://www.skylight.io/app/applications/t8bEzG0cuIkd/recent/6h/endpoints/CandidateInterface::ContentController%23providers?responseType=html

Query is find is down only once. This saves a lot of `SiteSetting` queries and calculation, and might solve the performance issues with this page.

## Link to Trello card

https://trello.com/c/x0OxYLae/2143-speed-up-the-providers-page-in-the-candidate-ui

## Things to check

### DIFF
--- a/app/components/grouped_provider_courses_component.html.erb
+++ b/app/components/grouped_provider_courses_component.html.erb
@@ -1,3 +1,5 @@
+<% find_down = EndOfCycleTimetable.find_down? %>
+
 <% courses_by_provider_and_region.each do |region_code, courses_by_provider| %>
   <h2 class="govuk-heading-m govuk-!-margin-bottom-2 govuk-!-margin-top-8"><%= label_for(region_code) %></h2>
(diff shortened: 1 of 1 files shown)


davidgisbey (approved):
🤦 good spot

tijmenb:
@davidgisbey it was my bad - I introduced the `SiteSetting` lookup in the timetable, which made this slow.

tvararu (approved):
Alternative could be to memoize the `find_down?` call, but I prefer this tbh.

tijmenb:
@tvararu should have mentioned that, but it's a class method, so memoizing it would persist across requests and not change if today's date or the cycle schedule changes.

PR #2918, opened 2020-09-16
Change how we display GCSEs on the Manage application choice page 

malcolmbaig:
## Context

Providers need to be able to see an international candidate's GCSE-level qualifications. We want to display these (and all other GCSE variants) in a new qualifications card component.

Introduce a GcseQualificationCardsComponent that renders a set of
three cards, one each for Maths, English, and Science.

Each qualification card can render in one of four states:

- A 'missing' qualification, ie - the candidate has declared that they
  don't have the qualification.
- An international qualification
- A UK qualification, picked from one of the standard types we list
  (GCSE, O Level, and Scottish National 5).
- An 'other' UK qualification, specified by the candidate.

International cards display a NARIC statement of comparability if the
candidate has provided a NARIC reference.

### Before

Screenshot_20200916_094543

### After
Screenshot_20200916_093836

## Guidance to review

- Should be possible to submit different applications on the review app to then view in Manage. May need to switch the app to 'Mid-cycle' in support.
- An outline of the `GcseQualificationCardsComponent` behaviour from the spec output:

Screenshot_20200916_094436

## Link to Trello card

https://trello.com/c/Kc48lCfJ
## Things to check

### DIFF
--- a/app/views/provider_interface/application_choices/show.html.erb
+++ b/app/views/provider_interface/application_choices/show.html.erb
@@ -33,11 +33,9 @@
     <% end %>
 
     <h2 class="govuk-heading-l govuk-!-margin-top-8" id="course-info">Course details</h2>
-
     <%= render ProviderInterface::CourseDetailsComponent.new(application_choice: @application_choice)%>
 
     <h2 class="govuk-heading-l govuk-!-margin-top-8 govuk-!-margin-bottom-2" id="disabilities">Disability, access and other needs</h2>
-
     <%= render ProviderInterface::TrainingWithDisabilityComponent.new(application_form: @application_choice.application_form) %>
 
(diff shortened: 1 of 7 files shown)


paulrobertlloyd (approved):

tvararu (approved):

app/views/provider_interface/application_choices/show.html.erb
```diff
@@ -33,33 +33,29 @@
     <% end %>
 
     <h2 class="govuk-heading-l govuk-!-margin-top-8" id="course-info">Course details</h2>
-
```
duncanjbrown, line 36: :+1: thank you!

duncanjbrown (approved):

PR #2920, opened 2020-09-16
Allow provider users to reinstate deferred offers from previous cycle

mnacos:
## Context

The full version of this flow is complicated, as it will allow provider users to change the details of the courses they re-offer, if they wish to. For now, we are giving provider users the ability to reinstate deferred offers without making any changes to their details. This means last year's offered course and all of its details (e.g. location, study mode) must be available as an option in the current recruitment cycle.

The plan is to allow the marking of individual of offer conditions as met/not met, as well as the modification of the conditions list as part of this flow. However, our system currently supports only the marking of _all_ conditions as met/not met, so this is what provider users will be able to do while reinstating an offer. No changes can be made yet to the list of conditions.

#### Course is available

image
---
image
---
image
---
image
---
image

#### Course is not available

image
---
image

#### Implementation details

This PR uses the same 'wizard' pattern as provider user invitations and provider relationship setup, as the complexity of this feature will increase over time. A new ApplicationChoice database field `:status_before_deferral` has been added, because applications can be in either 'pending_conditions' or 'recruited' states when they are deferred and we need this information in order to present the status of conditions as they were before the offer was deferred.

## Guidance to review

- Would we like to store more metadata when reinstating an offer?
- Should the introduction copy in this MVP be updated to be more course_option-centric? -- _Yes, the text shown if any of the course details are not available in the current cycle has been updated._ 
- Should there be some text advising users who would like to change the list of conditions they need to contact support?

## Link to Trello card

https://trello.com/c/zrLd3K8W

## Things to check

- [ ] This code does not rely on migrations in the same Pull Request -- IT DOES
- [ ] If this code includes a migration adding or changing columns, it also backfills existing records for consistency -- we'll backfill this column manually (using the audit log) for the few applications affected.


adamsilver:
Do not use bold on the prompt for ‘not available’.

The first paragraph on the review deferred offer page should just be:

```
The details of the course you offered previously are not available in the current cycle.
```

The rest of the page isn't right and should look like this:

image

The prompt should read:

```
The details of the course you offered previously are not available in the current cycle.
```

duncanjbrown:
Should there be a "Change" link on the conditions on the Check page?

mnacos:
> Should there be a "Change" link on the conditions on the Check page?

Have just added an 'Update status of conditions' link, as per the design.

adamsilver:
PR

Proto

Issues:

1. the panel should not show a link to update the status of the conditions
2. the last condition should not have a bottom border

adamsilver:
There should be no bottom border on last condition:

image

adamsilver:
image

Again no bottom border on last condition.

adamsilver:
image

The note at the bottom should say:

`We’ll email the candidate to confirm their new offer and remind them of the conditions for it.`

mnacos:
> image
> 
> The note at the bottom should say:
> 
> `We’ll email the candidate to confirm their new offer and remind them of the conditions for it.`

Hiding "and remind them of the conditions for it" is on purpose, when the conditions are 'still met'.

mnacos:
> > Should there be a "Change" link on the conditions on the Check page?
> 
> Have just added an 'Update status of conditions' link, as per the design.

I've removed it, as per @adamsilver 's suggestion.

adamsilver:
image

Too much space at the bottom of the defer panel - please compare to proto (or we can pair)

adamsilver:
image

This error message should read:

`Select yes if the candidate has met all of the conditions`

and

`Select yes if the candidate has still met all of the conditions`

adamsilver (approved):

mnacos:
> This error message should read:
> 
> `Select yes if the candidate has met all of the conditions`
> 
> and
> 
> `Select yes if the candidate has still met all of the conditions`

Discussed on Slack, keeping this simple with one error message for both cases: `Select yes if the candidate has met all of the conditions`

app/controllers/provider_interface/application_choices_controller.rb
```diff
@@ -39,6 +39,14 @@ def show
         course_option_id: @application_choice.offered_option.id,
       )
 
+      if @application_choice.status == 'offer_deferred'
+        @deferred_offer_wizard_applicable =
+          @application_choice.recruitment_cycle == RecruitmentCycle.previous_year
```
duncanjbrown, line 44: Maybe a `can_be_reinstated?` method?

app/controllers/provider_interface/reconfirm_deferred_offers_controller.rb
```diff
@@ -0,0 +1,112 @@
...
+        @wizard.errors.add(
+          :base,
+          'Unable to confirm offer, please try again. If problems persist please contact support',
+        )
+        render action: step
+      end
+    end
+
+    def next_path
+      step, _id = @wizard.next_step
```
duncanjbrown, line 48: The wizard could return a symbol or `nil` which would make this method simpler

app/views/provider_interface/reconfirm_deferred_offers/check.html.erb
```diff
@@ -0,0 +1,31 @@
+<% content_for :browser_title, title_with_error_prefix('Review offer', @wizard.errors.any?) %>
+<% content_for :before_content, govuk_back_link_to(previous_path) %>
+
+<%= form_with model: @wizard, url: next_path, method: :post do |f| %>
```
duncanjbrown, line 4: `next_path` Could be hardcoded as this will always hit the same action

app/views/provider_interface/application_choices/_application_choice_header.html.erb
```diff
@@ -20,6 +20,30 @@
   </div>
 <% end -%>
 
+<% if @deferred_offer_wizard_applicable && flash.empty? && @provider_can_respond -%>
```
duncanjbrown, line 23: Maybe `elsif` just to be sure we never show two?

app/views/provider_interface/reconfirm_deferred_offers/start.html.erb
```diff
@@ -0,0 +1,61 @@
...
+        </p>
+
+        <p class="govuk-body-m">
+          We’ll contact you within one working day to confirm that the offer has been updated. We’ll also contact the candidate confirming the new details.
+        </p>
+      <% end %>
+    </div>
+
+    <% if @wizard.applicable? %>
+      <%= govuk_link_to 'Continue', next_path, class: 'govuk-button' %>
```
duncanjbrown, line 58: If they haven't given us any information to branch off here we could hardcode

duncanjbrown (commented):
Notes from call!

app/controllers/provider_interface/application_choices_controller.rb
mnacos: After some thinking, it feels a bit niche to attach this to `ApplicationChoice`.

app/forms/provider_interface/reconfirm_deferred_offer_wizard.rb
```diff
@@ -0,0 +1,137 @@
...
+
+    validates :application_choice_id, :current_step, presence: true
+
+    validate :validate_step,
+             :validate_application_choice,
+             :is_course_option_still_available
+
+    validates :conditions_status, presence: true, if: proc { |f|
+      %w[conditions check].include? f.current_step
+    }
```
duncanjbrown, line 19: It's slightly more idiomatic to use validation contexts for this: see https://github.com/DFE-Digital/apply-for-teacher-training/blob/15cc371bd7926a46bcc8ff7fd049bb0846591d81/app/forms/provider_interface/provider_user_invitation_wizard.rb#L10 and https://github.com/DFE-Digital/apply-for-teacher-training/blob/15cc371bd7926a46bcc8ff7fd049bb0846591d81/app/forms/provider_interface/provider_user_invitation_wizard.rb#L63-L65

`on:` declares a validation context for each validation in the list, so you can say (as `valid_for_current_step?` does, `valid(:conditions_status)`

duncanjbrown (approved):
Thank you — this is great! One minor comment about validations inline, but this should be good to go

PR #2922, opened 2020-09-16
Add application card for applications list

tijmenb:

This adds the application card to the support UI, and updates the design (by @adamsilver and @paulrobertlloyd).

## Guidance to review

Does it look nice?

## Link to Trello card

None, this is underground work.

## Things to check


paulrobertlloyd:
@tijmenb Updated application card design for support, and in doing so, refactored existing the application card component used in the provider UI. This means both components can share the same styles – and these styles are less constrained, more adaptable. Need to run this change past @adamsilver.

## Support UI

### Mobile

support-new-mobile

### Desktop

support-new-desktop

## Provider UI

Commentary here largely for the benefit of @adamsilver. 

### Mobile (before)

provider-old-mobile

### Mobile (after)

Main difference here is that the status tag is no longer floated to the right; my thinking was that it’s best to have consistent position for statuses given names can be quite long.

provider-new-mobile

### Desktop (before)

provider-old-desktop

### Desktop (before)

Largely unchanged, though adjusted some of the spacing to better group related information.

provider-new-desktop

adamsilver:
@paulrobertlloyd I purposely made the left and right sections not overlap. Partly because I think it's neater but more importantly is that there may be information there – see screenshot:

image

paulrobertlloyd:
@adamsilver Let me add a screenshot of this with the says to respond text – I tucked this under the status tag.

paulrobertlloyd:
Me and @adamsilver reviewed the application card pattern and paired on a redesign (which also addresses provider feedback on needing cards to be more compact. Here’s the revised design:

## Provider

Screenshot 2020-09-17 at 14 23 37

## Support

Screenshot 2020-09-17 at 14 22 32

tijmenb:
I've rebased the changes onto master. 

@paulrobertlloyd @adamsilver Is this how the user card is supposed to look?

Screenshot 2020-09-21 at 13 36 56

paulrobertlloyd:
Latest commit should fix issue with users list, but was unable to find this page to double-check.

tvararu (approved):
Code LGTM!

PR #2930, opened 2020-09-16
Mitigate empty permissions for invited user

steventux:
## Context

https://sentry.io/organizations/dfe-bat/issues/1897001575/?project=1765973

image

When a provider user is invited and no permissions are granted in the invitation wizard the underlying data will not contain a permissions list. The details component which renders the granted permissions relied on the presence of this collection.

Calls `fetch` on permissions data saved in the wizard so that the collection of assigned permissions defaults to an empty array. This allows the details component to correctly render the _The user will only be able to view applications_ message. 

## Link to Trello card

https://trello.com/c/YLJ6awIP/2755-when-inviting-a-new-user-but-not-adding-any-permissions-i-see-a-sorry-there-is-a-problem-with-the-service-error-message

## Things to check

### DIFF
--- a/app/components/provider_interface/provider_user_invitation_details_component.rb
+++ b/app/components/provider_interface/provider_user_invitation_details_component.rb
@@ -61,7 +61,7 @@ module ProviderInterface
           key: key,
           value: render(
             ProviderInterface::ProviderUserInvitationPermissionsComponent.new(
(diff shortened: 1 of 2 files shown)


AgaDufrat (approved):
👍

PR #2935, opened 2020-09-17
Use background task fan-out for provider sync

tijmenb:
## Context

We currently only sync the courses for providers that we've opted-in. This is because syncing all 2000 providers would be too slow (about 90 minutes, according to unscientific calculations). We'd like to start syncing all courses, because it will be what we'll do anyway at some point next year (as all providers will be on Apply).

Kickoff notes: https://ukgovernmentdfe.slack.com/archives/CN1MCQCHZ/p1600334976169700

This is the minimum we have to do to change the approach. Instead of syncing everything in one background task, we schedule a background task for each provider/recruitment year combination. The idea is that we can run these in parallel and perform the whole task faster. 

In follow up PRs we can start syncing all of the courses, once we've fixed a couple of issues that would cause.

## Guidance to review

Are there any unintended side effects?

## Link to Trello card

https://trello.com/c/HwAFHcKS/1965-tech-performance-sync-all-of-the-courses-from-find

## Things to check


app/workers/sync_courses_from_find.rb
```diff
@@ -0,0 +1,182 @@
...
+
+    find_provider.courses.each do |find_course|
+      create_or_update_course(find_course)
+    end
+  end
+
+private
+
+  def find_provider
+    # https://api2.publish-teacher-training-courses.service.gov.uk/api/v3/recruitment_cycles/2020/providers/1N1/?include=sites,courses.sites
```
tvararu, line 19: I think the comment is out of date, it's probably this now, but could you sense check it?

```suggestion
    # https://api2.publish-teacher-training-courses.service.gov.uk/api/v3/recruitment_cycles/2020/providers/1N1/?include=sites,courses.subjects,courses.sites,courses.site_statuses.site
```

tvararu (approved):

duncanjbrown (approved):

PR #2936, opened 2020-09-17
Change how we display degrees on the Manage application choice page 

malcolmbaig:
## Context

Providers need to be able to see an international candidate's degrees. We want to display these in a new qualifications card component.

Introduce a `DegreeQualificationCardsComponent` that renders a card for
each degree held by a candidate.

- Each card contains the details expected by the provider (subject,
  grade, degree type, etc)
- A NARIC statement is displayed if the corresponding database fields
  contain values
- An '(Hons)' is appended to the degree type if the awarded grade is an
  honours
- Predicted grades are prefixed with 'Predicted:'
- The component can be initialised with an optional
  `application_choice_state` argument. If provided, we only show the
  institution name if the application is in an 'offer accepted' state.
  This was implemented as an optional argument because this component is
  also used in the Support interface, where hiding the institution isn't
  a requirement.

### Before
Screenshot_20200917_115059

### After
Screenshot_20200917_115035

## Guidance to review

- Should be possible to submit different applications on the review app to then view in Manage. May need to switch the app to 'Mid-cycle' in support.
- An outline of the `DegreeQualificationCardsComponent` behaviour from the spec output:

Screenshot_20200917_113800

## Link to Trello card

https://trello.com/c/66hZWX9d
## Things to check

### DIFF
--- a/app/components/degree_qualification_cards_component.html.erb
+++ b/app/components/degree_qualification_cards_component.html.erb
@@ -1,0 +1,38 @@
+<div id="degrees" class="govuk-!-margin-bottom-8">
+  <h2 class="govuk-heading-m"><%= section_title %></h2>
+
+  <div class="govuk-grid">
+    <div class="govuk-grid-row app-grid-row--flex">
+      <% degrees.each do |degree| %>
+        <div class="govuk-grid-column-one-third">
+          <div class="app-card app-card--outline">
+            <h3 class="govuk-heading-s govuk-!-margin-bottom-1">
+              <%= degree_type_and_subject(degree) %>
+            </h3>
+            <dl class="app-qualification">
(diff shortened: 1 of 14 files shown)


malcolmbaig:
Review app link: https://apply-for-te-2014-degre-soj19k.herokuapp.com/ (not sure why Github didn't pick it up)

tvararu (approved):

paulrobertlloyd:
@malcolmbaig Just a few changes in commit 589052c590ef483d0ad22763c4a24d2e3caa3d69

* Ensure we are using the correct heading levels on this page for different sections and their individual items
* Fix a few cases where heading `ids did not match the respective title
* Move some of the grid layout `div`s into components to prevent layout breakages

Be good to see this page with all types of qualifications present (Degree, GCSE, Other and EFL), but couldn't see a dummy application. How can I simulate that?

PR #2938, opened 2020-09-17
Cancel outstanding references at the end of cycle

davidgisbey:
## Context

Any outstanding reference requests which have not been received by the 18th September should be cancelled.

- Add a 'cancelled_at_end_of_cycle' state to ApplicationReference
- Set all references which have not been received by EOC to the 'cancelled_at_end_of_cycle' state
- Carry over non-overdue referees details in the DuplicateApplication service

image

## Guidance to review

There might be a potential issue with this implementation. If a Candidate added a reference just before EoC say 16/09 then their reference will be cancelled on 19/09. If they then don't log in for a month then those references will not be carried over as the feedback will be overdue. 

Not a huge issue though as it's a pretty niche context.

This will need a design and content review.

## Link to Trello card

https://trello.com/c/L93EMrXk/2125-dev-%F0%9F%9A%B2%F0%9F%94%9A-cancel-outstanding-ref-requests-on-18th

## Things to check

### DIFF
--- a/config/locales/application_form.yml
+++ b/config/locales/application_form.yml
@@ -377,7 +377,7 @@ en:
         awaiting_reference_sent_less_than_5_days_ago: We’ve emailed your referee. Keep in touch with them to ensure they’re planning on giving a reference as soon as possible.
         awaiting_reference_sent_more_than_5_days_ago: Your referee has not responded yet. Ask them if they got the email - it may have gone to junk or spam.
         cancelled: You cancelled your reference request.
-        cancelled_at_end_of_cycle: Your reference was cancelled at the end of the cycle.
+        cancelled_at_end_of_cycle: The referee did not respond before courses closed.
         feedback_overdue: Your referee has not responded yet. Keep chasing the referee or cancel the request in favour of another referee.
   activemodel:
     errors:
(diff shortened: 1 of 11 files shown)


config/locales/application_form.yml
```diff
@@ -377,6 +377,7 @@ en:
         awaiting_reference_sent_less_than_5_days_ago: We’ve emailed your referee. Keep in touch with them to ensure they’re planning on giving a reference as soon as possible.
         awaiting_reference_sent_more_than_5_days_ago: Your referee has not responded yet. Ask them if they got the email - it may have gone to junk or spam.
         cancelled: You cancelled your reference request.
+        cancelled_at_end_of_cycle: Your reference was cancelled at the end of the cycle.
```
davidgisbey, line 380: This line needs a content review please @EmmaFrith

tvararu (approved):
Code LGTM 👍

EmmaFrith:
thanks @davidgisbey 

could we look at the content for this to make it more plain English? not sure a candidate would know what 'cycle' means? 

eg. 

tag: Cancelled  
description: The referee did not respond before courses closed.

cc. @paulrobertlloyd

paulrobertlloyd:
@davidgisbey Changes as @EmmaFrith said, but can we make the tag use same colour we do for the other cancelled state – I think this is orange.

(Doing so kinda goes against my own made up rules regarding colour usage, but seeing as they are only in my head and not candidates – ! – but probably makes sense to use the same colours for statuses when then say and amount to the same thing.)

EmmaFrith:
i agree @paulrobertlloyd - it's all 'cancelled' - just for a different reason - which we explained in the content underneath

tvararu:
@paulrobertlloyd @EmmaFrith like this? (PR updated)

Screenshot 2020-09-18 at 10 51 31

paulrobertlloyd:
@tvararu Perfect!

PR #2944, opened 2020-09-18
Don't show the referee banner at end of cycle

tvararu:
## Context

We shouldn't prompt candidates to add/replace referees at the end of the cycle.

Don't `render?` when we're at the end of the cycle.

## Guidance to review

Makes sense?

### Before

Screenshot 2020-09-18 at 10 56 24

### After

Screenshot 2020-09-18 at 10 56 35

## Link to Trello card

None.

## Things to check

### DIFF
--- a/spec/components/candidate_interface/new_references_needed_component_spec.rb
+++ b/spec/components/candidate_interface/new_references_needed_component_spec.rb
@@ -1,0 +1,25 @@
+require 'rails_helper'
+
+RSpec.describe CandidateInterface::NewReferencesNeededComponent do
+  describe '#render?' do
+    let(:application_form) { create(:completed_application_form) }
+
+    context 'when an application form needs more references' do
+      it 'renders the component' do
+        allow(CandidateInterface::EndOfCyclePolicy).to receive(:can_add_course_choice?).and_return(true)
+        create(:reference, :requested, application_form: application_form, requested_at: Time.zone.now - 30.days)
+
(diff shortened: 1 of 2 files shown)


spec/components/candidate_interface/new_references_needed_component_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'rails_helper'
+
+RSpec.describe CandidateInterface::NewReferencesNeededComponent do
+  describe '#render?' do
+    let(:application_form) { create(:completed_application_form) }
+
+    context 'when an application form needs more references' do
+      it 'renders the component' do
+        create(:reference, :requested, application_form: application_form, requested_at: Time.zone.now - 30.days)
```
stevehook, line 9: Do you also need to stub `can_add_course_choice?` in this example like you do in the other or freeze Time to a known date? Otherwise would it fail after 18 September?

stevehook (approved):
:+1: Makes sense! One question about the specs.

spec/components/candidate_interface/new_references_needed_component_spec.rb
tvararu: Hmm, let's stub it

tvararu: Done

stevehook: :+1:

PR #2951, opened 2020-09-21
Update PR template to include design checks

paulrobertlloyd:
## Context

We want to close the loop between design and implementation by catching any design interaction and content decisions that may have been misinterpreted during implementation to ensure the services works as intended.

Adds to additional checks to the ‘Things to check’ section on the PR template.

## Guidance to review

Is this useful? If only as a gentle reminder… 😉


tijmenb (changes requested):
To be discussed on Slack!

PR #2955, opened 2020-09-21
overrideFullDeployment variable to control full ARM deployment

vigneshmsft:
## Context

Currently the full ARM deployment from the pipeline happens only if there are changes to the ARM template or azure pipeline files.
it might be required to run a full ARM template deployment irrespective of code changes.

introduce a `overrideFullDeployment` variable to force a full ARM template deployment irrespective of file changes.

## Things to check


saliceti (approved):
OK

PR #2959, opened 2020-09-22
Fixes for A11y issues on candidate landing page

stevehook:
## Context

The _qa_ tag is not deemed to have sufficient contrast for accessibility by Axe. Rather than just pick a different colour we've switched to new `govuk-tag` styles for all non-production environments.

- [x] Workaround issue with aria-hidden/focus no navigation element - see https://github.com/alphagov/govuk-frontend/issues/1340
- [x] Use new environment tag styles to fix A11y issues

Before:

image

After:

image
image
image
image
image

There are a couple of other places where tags appear for Sandbox specific features, e.g.

image

The production interface is unchanged:

image

## Guidance to review

- Does it look OK?
- Is there any more CSS that we can remove?

## Link to Trello card

https://trello.com/c/ALvwnGc6/1844-include-serious-accessibility-issues-in-axe-tests

## Things to check


fofr:
Are there screenshots of the other environments too?

stevehook:
> Are there screenshots of the other environments too?

@fofr I've added 'after' screenshots for all the other environments.

tvararu (approved):
:+1:

paulrobertlloyd (approved):
Love to see some CSS removed! Wondering if the remaining CSS in `_environments.scss` file could be moved into `_header.scss`, but that can be a separate PR if we think that’s a good idea.

PR #2968, opened 2020-09-22
Remove to work details when a candidate selects they are British or Irish

davidgisbey:
## Context

There is a bug that can be reproduced with the following steps on QA:

1. Complete the personal details section as an Albanian candidate
2. Edit nationality to British and Albanian
3. See the right to work row is there when it should not be

- Reset the right_to_work_or_study and  right_to_work_or_study_details fields to nil when a candidate selects British or Irish as one of their nationalities
- Update the logic for rendering the right_to_work row to check for Britsh/Irish in all nationalities

## Guidance to review

Previously the logic for whether to render the right to work row only checked the first nationality. The nationalities are stored alphabetically so if someone has picked Austrian/Albanian etc. and British as nationalities it renders the row when it shouldn't.

## Link to Trello card

https://trello.com/c/V6N7Dv0K/2050-%F0%9F%8C%90-dont-show-or-save-residency-status-if-nationality-is-british-or-irish

## Things to check

### DIFF
--- a/app/forms/candidate_interface/nationalities_form.rb
+++ b/app/forms/candidate_interface/nationalities_form.rb
@@ -33,13 +33,25 @@ module CandidateInterface
       if FeatureFlag.active?('international_personal_details')
         nationalities = candidates_nationalities
 
-        application_form.update!(
-          first_nationality: nationalities[0],
-          second_nationality: nationalities[1],
-          third_nationality: nationalities[2],
-          fourth_nationality: nationalities[3],
-          fifth_nationality: nationalities[4],
-        )
+        if british.present? || irish.present?
+          application_form.update!(
+            first_nationality: nationalities[0],
+            second_nationality: nationalities[1],
+            third_nationality: nationalities[2],
+            fourth_nationality: nationalities[3],
+            fifth_nationality: nationalities[4],
+            right_to_work_or_study: nil,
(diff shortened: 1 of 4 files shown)


paulrobertlloyd:
It’s possibly a bit simpler than this fix implies. If a candidate has either British or Irish nationality, don’t show (or save) any previous answer for residency status.

So, for example, in both the following examples, my previous answer for ‘Residency status’ should not be shown.

Screenshot 2020-09-22 at 18 23 51


Screenshot 2020-09-22 at 18 23 39

davidgisbey:
Cool. I think we should probably also set the values for the right to work response to nil on the application form though. They are no longer relavant and it could be potentially misleading if for example someone queries for all application forms in which those attributes are not blank.

paulrobertlloyd:
> Cool. I think we should probably also set the values for the right to work response to nil on the application form though. They are no longer relavant and it could be potentially misleading if for example someone queries for all application forms in which those attributes are not blank.

Yup. If British or Irish, don’t show or save anything regarding residency status, and remove it if previously added.

davidgisbey:
Perfect. This is good to go once a dev approves then 👍

paulrobertlloyd:
Lovely stuff.

tvararu (approved):

PR #2970, opened 2020-09-23
Restructure the Find Sync code for easier changing

tijmenb:
## Context

We're intending to sync all of the courses on Find into our database, instead of just a subset. This needs some changes to make this performant and reliable. The intended design of the sync is:

- We sync all providers first
- Then sync the courses for each provider, in a separate job

The Find sync code has grown a lot in the last year, so it needs some untangling to make this possible.

These are just some foundational changes, mostly moving things into a namespace so it's well organised, and updating some tests to be less dependent on the current implementation.

## Guidance to review

Per commit.

## Link to Trello card

https://trello.com/c/HwAFHcKS

## Things to check


tvararu (approved):
Reviewed commit by commit, most of the code is maintained as-is and didn't seem to change. The new services make sense. LGTM!

duncanjbrown (approved):
Namespace is 👌. Should we deduplicate `Find` from the class names, eg  `FindSync::SyncAllProviders` rather than `FindSync::SyncAllProvidersFromFind`?

PR #2971, opened 2020-09-23
Add a rake task to create large numbers of test applications 

malcolmbaig:
## Context

We're testing some of the performance characteristics of the Apply
service in preparation for the larger number of candidates expected next
cycle. One of these tests involves seeing how the service handles
display and navigation across a much larger number of applications.

This is a first pass at writing a rake task that creates applications en masse, 
with a view to creating ~200K applications on the QA environment. It emphasizes 
speed over strict 'completeness' of the applications, the goal being a task that's 
relatively quick to run while providing a reasonable approximation of what a new 
application looks like in the system.

The rake task added in this commit provides a mechanism to create a
(candidate > application_form > application_choice) set multiple times.

- It accepts the id of an example ApplicationForm record to derive its
  test data from.
- It accepts a number specifying how many distinct applications should be
  created.
- Most of the work is delegated to the `BulkCreateTestApplications` class.

Inside `BulkCreateTestApplications`, we compose some SQL to create the
three records in a single query. The logic here endeavours to create a
minimal representation of an application in the system, and
intentionally leaves out some of the related data that would typically be 
found (eg - english_proficiency, references, qualifications, etc).

### Provider interface
provider-applications

### Support interface (candidates)

support-candidates

### Support interface (applications)

support-applications

## Guidance to review

To test manually, get a copy of the branch locally, find an application form to replicate and then run 
```
bundle exec rails 'bulk_create_test_applications[x,y]'
```
Where `x` is the `application_form id` and `y` is the number of applications to create.

### Questions

- Is the data generated by the task a reasonable representation of a new set of application forms in the system?
- Are there any further changes that should be made to the generated data?

## Link to Trello card

https://trello.com/c/tRweeqzJ
## Things to check

### DIFF
--- a/app/services/bulk_create_test_applications.rb
+++ b/app/services/bulk_create_test_applications.rb
@@ -8,24 +8,33 @@ class BulkCreateTestApplications
   end
 
   def call
-    random_id = SecureRandom.hex(10)
+    begin
+      tries ||= 0
+      sql = compose_insert_statements(unique_id: SecureRandom.hex(10))
+      ActiveRecord::Base.connection.execute(sql)
+    rescue ActiveRecord::RecordNotUnique
+      retry unless (tries += 1) > 3
+      raise UniqueCandidateError, 'Failed to create a uniquely identifiable candidate, tried 3 times'
+    end
+  end
 
+  def compose_insert_statements(unique_id:)
     candidate_field_names_sql, candidate_values_to_persist = compose_sql_fragments(
       candidate,
       ignore_fields: %w[id magic_link_token magic_link_token_sent_at],
-      custom_fields: { 'email_address' => "#{random_id}@example.com" },
+      custom_fields: { 'email_address' => "#{unique_id}@example.com" },
     )
     application_form_field_names_sql, application_form_values_to_persist = compose_sql_fragments(
       template_application_form,
       ignore_fields: %w[id candidate_id],
-      custom_fields: { 'last_name' => random_id },
+      custom_fields: { 'last_name' => unique_id },
     )
     application_choice_field_names_sql, application_choice_values_to_persist = compose_sql_fragments(
       application_choice,
       ignore_fields: %w[id application_form_id],
     )
 
-    query = <<~SQL
+    <<~SQL
       WITH created_candidate AS (
         INSERT INTO "candidates" ( #{candidate_field_names_sql})
         VALUES ( #{candidate_values_to_persist})
@@ -41,8 +50,6 @@ class BulkCreateTestApplications
       VALUES ( #{application_choice_values_to_persist}, (SELECT id FROM created_application_form) )
       RETURNING "id"
     SQL
-
-    ActiveRecord::Base.connection.execute(query)
   end
 
   def compose_sql_fragments(model, ignore_fields: [], custom_fields: {})
@@ -66,4 +73,6 @@ class BulkCreateTestApplications
 
     [field_names_sql, values_to_persist]
   end
+
+  class UniqueCandidateError < StandardError; end
 end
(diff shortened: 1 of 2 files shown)


tvararu:
Going to run this on the review app :button_mashing:

tvararu:
Worked locally, 100k generated in around ~3m:

image

app/services/bulk_create_test_applications.rb
```diff
@@ -0,0 +1,69 @@
...
+  attr_reader :template_application_form, :candidate, :application_choice
+
+  def initialize(template_application_form)
+    @template_application_form = template_application_form
+    @candidate = template_application_form.candidate
+    @application_choice = template_application_form.application_choices.first
+  end
+
+  def call
+    random_id = SecureRandom.hex(10)
```
tvararu, line 11: Should we look up if the ID is already in use and retry if so?

app/services/bulk_create_test_applications.rb
```diff
@@ -0,0 +1,69 @@
...
+        VALUES ( #{candidate_values_to_persist})
+        RETURNING "id"
+      ),
+       created_application_form AS (
+        INSERT INTO "application_forms" ( #{application_form_field_names_sql}, candidate_id )
+        VALUES ( #{application_form_values_to_persist}, (SELECT id FROM created_candidate) )
+        RETURNING "id"
+      )
+
+      INSERT INTO "application_choices" ( #{application_choice_field_names_sql}, application_form_id )
```
tvararu, line 40: If we clone an application that is awaiting a provider decision, will cloning the application choices cause the provider users to receive 100k+ emails when they are rejected by default or similar?

app/services/bulk_create_test_applications.rb
malcolmbaig: Rather than add the additional query, a couple of alternatives might be:

- Let it crash out on the email unique index on the rare occasion that there's a conflict, and just restart is manually with however many records you have left to create
- Switch to using UUIDs. Significantly uglier in the UI but extremely low chance of conflict.

app/services/bulk_create_test_applications.rb
malcolmbaig: Good point, yes I think we will. Could include a guard clause protecting against certain application states? Are there any other states we'd want to treat the same way?

app/services/bulk_create_test_applications.rb
tvararu: Hmm. Instead of an additional query to check if it's available, could we wrap the `ActiveRecord.execute` with `begin/rescue` and if we rescue from the `RecordAlreadyExistsWhatchacallit` then we re-assign `random_id = SecureRandom.hex(10)` again, upwards of 3 tries or so?

app/services/bulk_create_test_applications.rb
tvararu: Guard clause so it only applies to `form started but unsubmitted` sounds like the way to go :+1: and throw a helpful error message if we use it on another type of application

malcolmbaig: Done, ready for another review.

app/services/bulk_create_test_applications.rb
```diff
@@ -0,0 +1,89 @@
...
+
+  def call
+    error_if_application_choice_state_is_not_unsubmitted
+
+    sql = compose_insert_statements(unique_id: SecureRandom.hex(10))
+    begin
+      tries ||= 0
+      ActiveRecord::Base.connection.execute(sql)
+    rescue ActiveRecord::RecordNotUnique
+      retry unless (tries += 1) > 3
```
tvararu, line 18: Does the `sql =` declaration need to be in the begin block, otherwise `retry` will just try running the same unique_id?

tvararu (approved):

app/services/bulk_create_test_applications.rb
malcolmbaig: :facepalm: it does! Fixed.

PR #2972, opened 2020-09-23
Handle when permissions are empty in save provider user service

AgaDufrat:
## Context

When the user decides not to give any permissions to the new user (view applications only) permissions array doesn't exists. This causes the following error https://sentry.io/organizations/dfe-bat/issues/1906888247/?project=1765973

Screenshot 2020-09-23 at 16 41 45

Call `fetch` on permissions data so when it is nil, we can return an empty array.

## Things to check


steventux (approved):
:+1: LGTM

PR #2973, opened 2020-09-23
[Spike] Vendor api tracing

steventux:
## Context

We'd like to trace Vendor API usage.

- Adds the model VendorAPIRequest which captures some information on the request and response to and from the vendor API.
- Vendor API requests are intercepted via middleware and enqueued for async processing.

image

image

## Guidance to review

TBC: Do we need timings? Should be use existing metrics like Skylight?

## Link to Trello card

https://trello.com/c/vk7ZdPUP/2774-track-requests-to-the-api

## Things to check

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -17,6 +17,7 @@ Bundler.require(*Rails.groups)
 
 require './app/lib/hosting_environment'
 require './app/middlewares/redirect_to_service_gov_uk_middleware'
+require './app/middlewares/vendor_api_request_middleware'
 
 require 'pdfkit'
 
@@ -53,6 +54,7 @@ module ApplyForPostgraduateTeacherTraining
     config.active_job.queue_adapter = :sidekiq
 
     config.middleware.insert_after ActionDispatch::HostAuthorization, RedirectToServiceGovUkMiddleware
+    config.middleware.use VendorAPIRequestMiddleware
     config.middleware.use PDFKit::Middleware, { print_media_type: true }, disposition: 'attachment', only: [%r[^/provider/applications/\d+]]
     config.skylight.environments = ENV['SKYLIGHT_ENABLE'].to_s == 'true' ? [Rails.env] : []
   end
--- a/app/controllers/support_interface/vendor_api_requests_controller.rb
+++ b/app/controllers/support_interface/vendor_api_requests_controller.rb
@@ -1,0 +1,7 @@
+module SupportInterface
+  class VendorAPIRequestsController < SupportInterfaceController
+    def index
+      @vendor_api_requests = VendorAPIRequest.all.order(created_at: :desc)
+    end
+  end
+end
--- a/app/controllers/support_interface/vendor_api_requests_controller.rb
+++ b/app/controllers/support_interface/vendor_api_requests_controller.rb
@@ -1,7 +1,22 @@
 module SupportInterface
   class VendorAPIRequestsController < SupportInterfaceController
     def index
-      @vendor_api_requests = VendorAPIRequest.all.order(created_at: :desc)
+      @filter = SupportInterface::VendorAPIRequestsFilter.new(params: params)
+      @vendor_api_requests = VendorAPIRequest.order(created_at: :desc).page(params[:page] || 1).per(15)
+
+      if params[:q]
+        @vendor_api_requests = @vendor_api_requests.where("CONCAT(request_path, ' ', request_body, ' ', response_body, ' ', hashed_token) ILIKE ?", "%#{params[:q]}%")
+      end
+
+      if params[:status_code]
+        @vendor_api_requests = @vendor_api_requests.where('status_code IN (?)', params[:status_code])
+      end
+
+      %w[created_at status_code request_path].each do |column|
+        next unless params[column]
+
+        @vendor_api_requests = @vendor_api_requests.where(column => params[column])
+      end
     end
   end
 end
--- a/app/middlewares/vendor_api_request_middleware.rb
+++ b/app/middlewares/vendor_api_request_middleware.rb
@@ -1,0 +1,57 @@
+require 'redis'
+require './app/workers/vendor_api_request_worker'
+
+class VendorAPIRequestMiddleware
+  REQUEST_HEADER_KEYS = %w[
+    HTTP_VERSION
+    HTTP_HOST
+    HTTP_USER_AGENT
+    HTTP_ACCEPT
+    HTTP_ACCEPT_ENCODING
+    HTTP_AUTHORIZATION
+    HTTP_CONNECTION
+    HTTP_CACHE_CONTROL
+  ].freeze
+
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    started_at = Time.zone.now
+
+    @request = Rack::Request.new(env)
+    status, headers, response = @app.call(env)
+
+    begin
+      if vendor_api_path?
+        finished_at = Time.zone.now
+        response_time = (finished_at - started_at) * 1000
+
+        VendorAPIRequestWorker.perform_async(request_data, response.body, status, response_time, Time.zone.now)
+      end
+    rescue Redis::BaseError => e
+      Rails.logger.warn e.message
+    end
+
+    [status, headers, response]
+  end
+
+private
+
+  def request_data
+    {
+      path: @request.path,
+      params: @request.params,
+      headers: request_headers,
+    }
+  end
+
+  def request_headers
+    @request.env.slice(*REQUEST_HEADER_KEYS)
+  end
+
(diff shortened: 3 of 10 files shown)


app/workers/vendor_api_trace_worker.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def response_hash(response_body, status)
+    return {} unless status > 299
+
+    JSON.parse(response_body)
+  rescue JSON::ParserError
+    # FIXME: We don't have JSON responses for all response types eg. 404
+    # and we don't want to save HTML error page responses in the db.
+    # Should this be a no-op?
+    { body: "#{status} did not respond with JSON" }
```
duncanjbrown, line 25: I think this is a nice feature, we can see when we're wrongly serving HTML!

app/models/vendor_api_trace.rb
```diff
@@ -0,0 +1,2 @@
+class VendorAPITrace < ApplicationRecord
```
tijmenb, line 1: Perhaps:

```suggestion
class VendorAPIRequest < ApplicationRecord
```

app/middlewares/vendor_api_tracing_middleware.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+private
+
+  def request_hash
+    { base_url: @request.base_url, path: @request.path, params: @request.params }
+  end
+
+  def vendor_api_path?
+    @request.path =~ /^\/api\/v1|v2\/.*$/
```
duncanjbrown, line 36: Ha - v2! Maybe just `/^\/api\/`?

config/application.rb
```diff
@@ -53,6 +54,7 @@ class Application < Rails::Application
     config.active_job.queue_adapter = :sidekiq
 
     config.middleware.insert_after ActionDispatch::HostAuthorization, RedirectToServiceGovUkMiddleware
+    config.middleware.use VendorAPITracingMiddleware
```
duncanjbrown, line 57: I suppose it doesn't matter where it goes in the stack?

app/middlewares/vendor_api_tracing_middleware.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def call(env)
+    started_at = Time.zone.now
+
+    @request = Rack::Request.new(env)
+    status, headers, response = @app.call(env)
+
+    begin
+      if vendor_api_path?
+        finished_at = Time.zone.now
```
tijmenb, line 17: We might want to pass in the current time explicitly into the worker, otherwise the `created_at` will be misleading in case the worker retries.

db/migrate/20200923101308_create_vendor_api_traces.rb
```diff
@@ -0,0 +1,13 @@
+class CreateVendorAPITraces < ActiveRecord::Migration[6.0]
+  def change
+    create_table :vendor_api_traces do |t|
+      t.jsonb     :headers
```
tijmenb, line 4:
```suggestion
      t.jsonb     :request_headers
```

db/migrate/20200923101308_create_vendor_api_traces.rb
```diff
@@ -0,0 +1,13 @@
...
+      t.jsonb     :request
```
tijmenb, line 5:
```suggestion
      t.jsonb     :request_body
```

app/middlewares/vendor_api_tracing_middleware.rb
```diff
@@ -0,0 +1,38 @@
...
+class VendorAPITracingMiddleware
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    started_at = Time.zone.now
+
+    @request = Rack::Request.new(env)
+    status, headers, response = @app.call(env)
```
duncanjbrown, line 13: Maybe name these `response_status`, `response_headers` etc?

db/migrate/20200923101308_create_vendor_api_traces.rb
```diff
@@ -0,0 +1,13 @@
...
+      t.jsonb     :response
```
tijmenb, line 6:
```suggestion
      t.jsonb     :response_body
```

db/migrate/20200923101308_create_vendor_api_traces.rb
```diff
@@ -0,0 +1,13 @@
...
+      t.integer   :status
```
tijmenb, line 7:
```suggestion
      t.integer   :status_code
```

db/migrate/20200923101308_create_vendor_api_traces.rb
```diff
@@ -0,0 +1,13 @@
...
+      t.bigint    :request_time
```
tijmenb, line 8: Not sure:

```suggestion
      t.bigint    :response_time
```

app/middlewares/vendor_api_tracing_middleware.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    begin
+      if vendor_api_path?
+        finished_at = Time.zone.now
+        request_time = (finished_at - started_at) * 1000
+
+        VendorAPITraceWorker.perform_async(headers, request_hash, response.body, status, request_time)
```
duncanjbrown, line 20: I think this is response headers here? do we want `@request.headers`?

db/migrate/20200923101308_create_vendor_api_traces.rb
```diff
@@ -0,0 +1,13 @@
...
+
```
tijmenb, line 9: We also want a way to match the requests to providers, so we'll need a reference to the token that made the request I think.

duncanjbrown (commented):
Interested to hear what others think, but I quite like this approach. All in one place, knows nothing about business logic, slots in as ordinary middleware. Keeping request headers is nice too — will surely be useful for troubleshooting. We can add `Export as cURL` in future!

tijmenb (commented):
Looking good! I'm assuming that middlewares will work for this.

Can we whip up a quick UI in support? Probably something that looks like [the email log](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/views/support_interface/email_log/index.html.erb). 

That will make it easier to test (you could even just do a system test that makes an API request and then checks the UI in support), and will also make sure that we persist the correct request data (for example, I'm not sure if we should persist the request URL separately, or that this is fine).

config/application.rb
tijmenb: It probably does, if we want to keep track of the response time? Not sure though.

duncanjbrown:
Is it worth splitting request_path and request_body in the DB?

db/migrate/20200923101308_create_vendor_api_traces.rb
tijmenb: That leads to the question what to do with requests with a wrong token. Should we ignore those?

duncanjbrown: The app will serve an error response in the usual way, which would be good to know about (maybe someone's integration is broken)

tijmenb: But if the token is wrong, we won't know who made the request I think.

duncanjbrown: We might have the headers including forwarded-for and possibly user-agent?

duncanjbrown: But maybe this is just one for Kibana? Could get noisy too

duncanjbrown:
Do we need `request_time` (ie duration) if we have Skylight?

db/migrate/20200923101308_create_vendor_api_traces.rb
steventux: Should we persist a column with values something like...
```ruby
provider_uuid: Digest::MD5.hexdigest(request_headers.delete('Authorization'))`
```

steventux:
@tijmenb @duncanjbrown thanks for all the useful suggestions and comments. I've amended quite a bit of this spike PR as advised. Conscious of spending too much time on this now but most of what you've suggested is done.

#### TODO:

- Decide how we record timings of requests and whether we use existing metrics like Skylight?
- Which request headers do we need. Ive made a best guess.
- Added a Support UI but this is very limited, I guess we can pick this up in a proper implementation story.
- I've hashed the Provider API Token in the same way Devise does, I assume this to be one-way. Need to check this is secure. Also not sure how useful it is for searching.

steventux:
Closing in favour of https://github.com/DFE-Digital/apply-for-teacher-training/pull/3009

PR #2974, opened 2020-09-23
Support: Update design of pages in Tasks section

paulrobertlloyd:
## Context

There are a lot of important buttons on this page, so we should make sure its super clear what headings relate to which buttons, and that we use common conventions on confirmation pages that can have a destructive result.

* Updates layout and tweak content/headings on ‘Tasks’ page
* Updated content for each task conformation pages that follows conventions for such pages
* Updates `app-product-section` to `app-section` to make it a common shared style
* Add styling for `code` in `app-styled-content`.

## Guidance to review

| Before | After |
| - | - |
| tasks-index | tasks-index-redesigned |
| tasks-confirmation | tasks-confirmation-redesigned |

### DIFF
--- a/app/frontend/styles/_section.scss
+++ b/app/frontend/styles/_section.scss
@@ -1,4 +1,4 @@
-.app-product-section {
+.app-section {
   @include govuk-responsive-padding(6, "top");
   @include govuk-responsive-padding(4, "bottom");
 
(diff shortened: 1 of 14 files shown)


fofr:
Nice 💯

tvararu (approved):
:heart_eyes: designer love!

PR #2979, opened 2020-09-24
Remove open_on_apply constraint on broken view component

davidgisbey:
## Context

The ChangeOfferPreview view component is still borked 🤦 

Currently, the pick_different_option method is returning nil as it is scoped to only return courses that are open_on_apply which is returning an empty array. There is no need for the replacement course choice to be open_on_apply for a preview component.

image

- Remove the constraint that courses need to be open on apply

## Guidance to review

Is my assumption that this is not necessary correct?

## Link to Trello card

https://trello.com/c/Ejg12FYm/1954-fix-component-previews-and-enable-automated-accessibility-tests

## Things to check


stevehook (approved):
:+1: Makes sense to not be too fussy about courses in the previews.

PR #2983, opened 2020-09-24
Notify in Slack when the cycle is changed through support

tvararu:
## Context

It would be nice to see when the cycle was updated as it could help us track down issues.

Add a Slack ping to the controller.

## Guidance to review

Makes sense?

## Link to Trello card

None.

## Things to check


davidgisbey (approved):

stevehook (approved):
:+1: Nice

malcolmbaig (approved):

PR #2985, opened 2020-09-24
Add data migration to update legacy degree grade values 

malcolmbaig:
## Context

When entering a degree grade, in addition to allowing free text input we
also offer the candidate a selection of radio buttons mapped to one of
several common values. We've provided radio buttons historically,
however the actual stored database value for these standard options
changed when we added structured HESA data to degrees. Rather than
saving a short string such as `upper_second`, we now simply persist the
HESA description (`Upper second-class honours (2:1)`).

This prompted a change to how we display these grades on the frontend. Instead
of translating database values such as `upper_second` to a presentable
form via I18n, we now simply show the database value. This works fine
for anything with a standard HESA grade or a free-text input, but
doesn't look so great for older degree records that simply display
`upper_second` on the frontend.

We can easily map those older values to the current HESA descriptions.
This commit adds a migration which applies that mapping and also adds
the corresponding HESA code to the degree record.

### Before migration
Screenshot_20200924_203241

### After migration
Screenshot_20200924_203538

## Link to Trello card
https://trello.com/c/kRTjP8aV

## Things to check

### DIFF
--- a/db/migrate/20200924121000_migrate_old_degree_grade_values_to_hesa_descriptions.rb
+++ b/db/migrate/20200924121000_migrate_old_degree_grade_values_to_hesa_descriptions.rb
@@ -1,0 +1,15 @@
+class MigrateOldDegreeGradeValuesToHesaDescriptions < ActiveRecord::Migration[6.0]
+  def change
+    first = ApplicationQualification.degrees.where(grade: %w[first First])
+    first.update_all(grade: 'First class honours', grade_hesa_code: 1)
+
+    upper_second = ApplicationQualification.degrees.where(grade: %w[upper_second 2:1])
+    upper_second.update_all(grade: 'Upper second-class honours (2:1)', grade_hesa_code: 2)
+
+    lower_second = ApplicationQualification.degrees.where(grade: %w[lower_second 2:2])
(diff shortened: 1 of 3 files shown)


davidgisbey (approved):
lgtm :shipit:

db/migrate/20200924121000_migrate_old_degree_grade_values_to_hesa_descriptions.rb
```diff
@@ -0,0 +1,15 @@
+class MigrateOldDegreeGradeValuesToHesaDescriptions < ActiveRecord::Migration[6.0]
+  def change
```
stevehook, line 2: As this is non-reversible (and doesn't need to be) would it be best to make this an `up` and have an empty `down` method?

tvararu (approved):

stevehook (approved):
Looks good. Minor comment inline.

db/migrate/20200924121000_migrate_old_degree_grade_values_to_hesa_descriptions.rb
malcolmbaig: Doesn't trigger an exception on rollback which I think is probably fine?

PR #2987, opened 2020-09-25
Refactor QualificationsTableComponent

malcolmbaig:
## Context

This component was previously used to render a generic table populated
with qualifications. Most qualifications are now displayed in cards
(see, for example - `EflQualificationCardComponent`). This table
component is now only used for rendering the 'other' qualification
type. It's also in a slightly odd state - it still accepts a
`type_label` argument while having a hard-coded 'Other qualifications'
header in its markup.

There are a couple of options to deal with this inconsistency -

- make the component specific to 'other' qualifications, renaming it and
  removing the `type_label` argument
- leave it as a generic component and use the argument to populate the
  header

This commit opts for the latter. There isn't much of a gain in code clarity
by making the component more specific, and it can be used to render
similar tables if required later on.

## Things to check

### DIFF
--- a/app/components/qualifications_component.html.erb
+++ b/app/components/qualifications_component.html.erb
@@ -14,6 +14,6 @@
 <div class="govuk-grid-row">
   <%= render DegreeQualificationCardsComponent.new(application_form.application_qualifications.degrees, application_choice_state: application_choice_state) %>
   <%= render GcseQualificationCardsComponent.new(application_form) %>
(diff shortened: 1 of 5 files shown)


davidgisbey (approved):
makes sense to make it reusable 👍

stevehook (approved):
:+1:

tvararu (approved):

PR #2988, opened 2020-09-25
Show HESA codes for qualifications in support

stevehook:
## Context

We updated the display of Degree qualifications in the provider UI. As this used a shared component, this also meant updating the same information in the support UI. Yet, in the support UI we also need to show HESA codes, which are now missing.

- [x] Add markup for HESA codes to `DegreeQualificationCardsComponent` when new `show_hesa_codes` option is true.
- [x] Apply the new option only in the Support interface
- [x] Update `degree_qualification` factory so that HESA codes are populated by default so that sample data will be a fully populated.

image

## Guidance to review

- I think that previously generated sample data doesn't include the codes, but if you re-run `rails setup_local_dev_data` they should be populated going forward.

## Link to Trello card

https://trello.com/c/TtA6Bu47/2208-📐-show-hesa-codes-for-qualifications-in-support

## Things to check

### DIFF
--- a/app/components/degree_qualification_cards_component.html.erb
+++ b/app/components/degree_qualification_cards_component.html.erb
@@ -29,6 +29,15 @@
                 <dt class="app-qualification__key">Comparability</dt>
                 <dd class="app-qualification__value"><%= naric_statement(degree) %></dd>
               <% end %>
+
+              <% if show_hesa_codes? %>
+                <dt class="app-qualification__key">HESA codes</dt>
+                <dd class="app-qualification__value">
+                  <% hesa_code_values(degree).each do |key, value| %>
+                    <%= key %>: <%= value %><br/>
+                  <% end %>
+                </dd>
+              <% end %>
             </dl>
           </div>
         </div>
--- a/app/components/degree_qualification_cards_component.html.erb
+++ b/app/components/degree_qualification_cards_component.html.erb
@@ -32,11 +32,11 @@
 
               <% if show_hesa_codes? %>
                 <dt class="app-qualification__key">HESA codes</dt>
-                <dd class="app-qualification__value">
-                  <% hesa_code_values(degree).each do |key, value| %>
-                    <%= key %>: <%= value %><br/>
-                  <% end %>
-                </dd>
+                <% hesa_code_values(degree).each do |key, value| %>
+                  <dd class="app-qualification__value govuk-!-margin-bottom-0">
+                    <%= key %>: <%= value %>
+                  </dd>
+                <% end %>
               <% end %>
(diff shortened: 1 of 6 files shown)


app/components/degree_qualification_cards_component.html.erb
```diff
@@ -29,6 +29,15 @@
                 <dt class="app-qualification__key">Comparability</dt>
                 <dd class="app-qualification__value"><%= naric_statement(degree) %></dd>
               <% end %>
+
+              <% if show_hesa_codes? %>
+                <dt class="app-qualification__key">HESA codes</dt>
+                <dd class="app-qualification__value">
```
paulrobertlloyd, line 35: Can we have this output the following HTML:

```
<dt class="app-qualification__key">HESA codes</dt>
<dd class="app-qualification__value app-qualification__value--grey govuk-!-margin-bottom-0">Type: 72</dd>
<dd class="app-qualification__value app-qualification__value--grey govuk-!-margin-bottom-0">Subject: 100121</dd>
<dd class="app-qualification__value app-qualification__value--grey govuk-!-margin-bottom-0">Establishment: 0012</dd>
<dd class="app-qualification__value app-qualification__value--grey govuk-!-margin-bottom-0">Class: 01</dd>
```

paulrobertlloyd (changes requested):

spec/factories.rb
```diff
@@ -257,10 +257,29 @@
...
+          hesa_institution = Hesa::Institution.all.sample
+          qualification.institution_name = hesa_institution.name
+          qualification.institution_hesa_code = hesa_institution.hesa_code
+        end
+        if qualification.grade.blank?
+          hesa_grade = Hesa::Grade.all.sample
+          qualification.grade = hesa_grade.description
+          qualification.grade_hesa_code = hesa_grade.hesa_code
+        end
+      end
```
malcolmbaig, line 282: We're currently doing something like this in master, introduced in one of my PRs at some point. A rebase ought to reconcile the two!

malcolmbaig (approved):
Code lgtm :+1:

app/components/degree_qualification_cards_component.html.erb
stevehook: Thanks! Tried with multiple `dd`s first - should have realised I only needed to override the margin to get it looking right.

spec/factories.rb
stevehook: :+1: I think yours is better because it handles the case where the string values and not the codes are specified by the caller.

paulrobertlloyd (approved):

PR #2994, opened 2020-09-28
Sandbox feature: force courses to stay open

duncanjbrown:
### Context

HEIs testing with their SRSs in advance of the new cycle starting aren't able to apply to their own courses as they are marked `no_vacancies` in the data we get from Find. This is making them nervous because they have processes which depend on the legitimate course data so fake providers don't let them test as thoroughly as they'd like.

### Changes proposed in this PR

- small refactorings between course_option and course
- add a `hold_open` boolean to course options and respect it on the `available` and `vacancies` scope, plus the `no_vacancies` method

Proposed workflow is to manually set that boolean on SB using the shell, hence no UI. Should be one-off per provider who needs it.

### Guidance to review

This won't be with us long (just til Oct 13)!

It's a bit magic. Should we mark these courses in support somehow?

### Link to Trello card

https://trello.com/c/9ggZMAtB/2807-allow-providers-to-use-existing-courses-to-test-in-sandbox


duncanjbrown:
Discussed offline with @tijmenb — we're going to use a sandbox-wide feature flag for this and not a db field

app/models/course.rb
```diff
@@ -99,13 +99,21 @@ def available_study_modes_from_options
...
 
   def available?
     course_options.available.present?
   end
 
+  def closed_on_apply?
+    !open_on_apply
+  end
+
+  def not_available?
```
malcolmbaig, line 113: Perhaps one for another PR, but this isn't exactly the best method name I've ever picked. Thinking we should be more explicit here and just call it something like `not_available_on_find?`

malcolmbaig:
> Discussed offline with @tijmenb — we're going to use a sandbox-wide feature flag for this and not a db field

Refactors LGTM, if you wanted to rework this PR to focus solely on those.

malcolmbaig (approved):

duncanjbrown:
Closing in favour of #3003

PR #2995, opened 2020-09-28
Remove flash message when a candidate carries over their application

davidgisbey:
## Context

When a candidate carries over an application they see a flash message. They should be told that their application is ready for editing.

Before 

image

After

image

## Link to Trello card

https://trello.com/c/HqHoLTQ9/2184-bug-%F0%9F%94%9A%F0%9F%9A%B2-incorrect-red-banner-showing-on-carried-over-applications

## Things to check


stevehook (approved):
I think this makes sense. We don't want to show this banner for carried over applications and we already show a carry-over prompt banner if the application is from the earlier cycle.

PR #3001, opened 2020-09-28
Update to Back links on contact details section

george-bit:
## Context

Making this change to iron out bugs with previous back link solution for the contact details section.

New route, controller actions have been defined to split the new and edit functions to allow specific paths for each.

## Guidance to review

Please check that behaviour of back links in this section is consistent with: https://trello-attachments.s3.amazonaws.com/5cb88ec4be601441218204a0/5f47f079f0272516088b14c8/44e05334bb9f28deee5fc24425b9d819/Back_link_journeys.pdf

## Link to Trello card

https://trello.com/c/nrMBwBLm/2188-%F0%9F%94%99-back-links-on-contact-details-flow-should-not-use-history-but-context-prevent-back-loops

## Things to check

### DIFF
--- a/app/views/candidate_interface/contact_details/address/_form_fields.html.erb
+++ b/app/views/candidate_interface/contact_details/address/_form_fields.html.erb
@@ -1,0 +1,20 @@
+<%= f.govuk_error_summary %>
+
+<% if @contact_details_form.uk? %>
+  <fieldset class="govuk-fieldset">
+    <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl govuk-!-margin-bottom-8">
+      <h1 class="govuk-fieldset__heading">
+        <%= t('page_titles.address') %>
+      </h1>
+    </legend>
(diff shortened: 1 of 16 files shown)


app/views/candidate_interface/contact_details/address/_form_fields.html.erb
```diff
@@ -0,0 +1,20 @@
+<%= f.govuk_error_summary %>
+
+<% if @contact_details_form.uk? %>
+  <fieldset class="govuk-fieldset">
```
paulrobertlloyd, line 4: Can we use `govuk_fieldset` from the form builder for this?

george-bit: @paulrobertlloyd - Sorry not sure I follow, The fieldset's class is already set to "govuk-fieldset"? does this need to be an underscore?

paulrobertlloyd: Sorry, should have been clearer. I meant the `govuk_frontend` helper from the form builder gem. See: https://govuk-form-builder.netlify.app/building-blocks/fieldsets/

paulrobertlloyd:
Spotted one issue: when I click ‘Change’ next to an address on the section review page, then click ‘Back’ from the ‘What is your address page’, I get a ‘Page not found’ error (back links to `/candidate/application/contact-details/address_type/edit`

Trying to test the rest of the journey, but possibly found a bug with support, so will check the rest when that’s been fixed/addressed.

PR #3004, opened 2020-09-28
HESA codes (sex, disabilities & ethnicity) - write to and read from the database

TobyRet:
## Context
```
As a Provendor team
I need application HESA data codes for Sex, Ethnicity, Disability to be collected at source
So that I can supply Providers with a download they can share with HESA/DfE
```

- This PR takes the information inputted by the user in the equality and diversity survey, looks up the HESA codes for sex, disability and ethnicity and writes them to the database (stored in the `equalities_and_diversity` JSON blob in the application form.
- This information is now exposed in the Provender Application API

## Guidance to review

- Fill out the equality and diversity survey and you will see the values written to the db.
- One of the challenges has been how to deal with strings, strings, strings. Some complexity / indirection arises in cases where we need to convert a form value to a HESA value in order to get the HESA code (see ethnicity and disbabilities). Perhaps there is a better way to do this? 
- Note that backfilling values will be in a separate PR.

## Link to Trello card

https://trello.com/c/mxKdgR5L/2178-dev-collect-hesa-codes-at-source-in-ed-qs

## Things to check

### DIFF
--- a/app/forms/candidate_interface/equality_and_diversity/ethnic_background_form.rb
+++ b/app/forms/candidate_interface/equality_and_diversity/ethnic_background_form.rb
@@ -30,18 +30,18 @@ module CandidateInterface
       other_background_present = ethnic_background == EthnicBackgroundHelper::OTHER_ETHNIC_BACKGROUNDS[group].first && other_background.present?
 
       background = other_background_present ? other_background : ethnic_background
-      ethnicity_type = hesa_ethnicity_type(ethnic_background)
+      ethnicity_value = hesa_ethnicity_value(ethnic_background)
 
       if application_form.equality_and_diversity.nil?
         application_form.update(
           equality_and_diversity: {
             'ethnic_background' => background,
-            'hesa_ethnicity' => hesa_ethnicity_code(ethnicity_type),
+            'hesa_ethnicity' => hesa_ethnicity_code(ethnicity_value),
           },
         )
       else
         application_form.equality_and_diversity['ethnic_background'] = background
-        application_form.equality_and_diversity['hesa_ethnicity'] = hesa_ethnicity_code(ethnicity_type)
+        application_form.equality_and_diversity['hesa_ethnicity'] = hesa_ethnicity_code(ethnicity_value)
         application_form.save
       end
     end
@@ -52,13 +52,13 @@ module CandidateInterface
 
   private
 
-    def hesa_ethnicity_type(background)
-      Hesa::Ethnicity.convert_to_hesa_type(background)
+    def hesa_ethnicity_value(background)
+      Hesa::Ethnicity.convert_to_hesa_value(background)
     end
 
     def hesa_ethnicity_code(type)
       Hesa::Ethnicity
-        .find_by_type(type, RecruitmentCycle.current_year)
+        .find_by_value(type, RecruitmentCycle.current_year)
         &.hesa_code
     end
   end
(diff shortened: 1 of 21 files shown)


config/initializers/hesa_disabilities.rb
```diff
@@ -1,12 +1,26 @@
+module HesaDisabilityTypes
```
TobyRet, line 1: I'm in two minds about this. The use of constants has been used in order to have 'one source of truth'. This is useful in scenarios where a form value does not directly map onto a HESA value. On the other hand I'm worried this creates more indirection and makes the code less easier to understand. Any opinions on this welcome!

config/initializers/hesa_disabilities.rb
```diff
@@ -1,12 +1,26 @@
+module HesaDisabilityValues
```
TobyRet, line 1: I'm in two minds about this. The use of constants has been used in order to have 'one source of truth'. This is useful in scenarios where a form value does not directly map onto a HESA value. On the other hand I'm worried this creates more indirection and makes the code less easier to understand. Any opinions on this welcome!

duncanjbrown: If i understand correctly we now map twice — once from the form to the type, and then from the type to the code? How would it be if the form values were codes, then store just codes, no strings? Might need to override the labels in some cases.

TobyRet: Yeah that's an interesting idea @duncanjbrown 

We could change the values of the radio buttons like follows, which would remove the need to have globals (via the initialisers) or any conversion:

```
          <%= f.govuk_radio_button :sex, {value: 'female', hesa_code: '00'}, label: { text: 'Female' }, link_errors: true %>
```
This would have implications for other parts of the code which are already storing the Hesa code via the conversion method (e.g. Subjects, Grades) i.e. refactoring - what do you think @tvararu ?

TobyRet: Actually, that said, we would still need the conversion code when we come to backfilling ...

tvararu: The approach to save codes instead of strings makes sense, but let's do that only for these new HESA codes. The old ones can stay as they are and be revisited later as tech debt. :+1:

TobyRet: Further work to add hesa codes to the forms has revealed a number of additional complexities, and results in far more changes as well as spreading hesa code logic in many places. @tvararu and I believe that it makes sense to stick with the original solution proposed in this PR, which is also consistent with previous implementations

app/forms/candidate_interface/equality_and_diversity/ethnic_background_form.rb
```diff
@@ -30,17 +30,36 @@ def save(application_form)
       other_background_present = ethnic_background == EthnicBackgroundHelper::OTHER_ETHNIC_BACKGROUNDS[group].first && other_background.present?
 
       background = other_background_present ? other_background : ethnic_background
+      ethnicity_type = hesa_ethnicity_type(ethnic_background)
```
tvararu, line 33: Doesn't look like we use `ethnicity_type` directly in this method, maybe just fetch `ethnicity_code = ...` and use that? V minor

tvararu (approved):

app/forms/candidate_interface/equality_and_diversity/ethnic_background_form.rb
TobyRet: Woops - `ethnicity_type` and `hesa_ethnicity_type` should actually be renamed to `ethnicity_value` and `hesa_ethnicity_value` respectively. This variable is used in a couple of places (lines 39 & 44) - I'll  make the change

PR #3009, opened 2020-09-29
Trace vendor api requests

steventux:
## Context

We'd like to trace Vendor API usage.
[See spike here for preliminary work and discussion](https://github.com/DFE-Digital/apply-for-teacher-training/pull/2973)

image

- Adds the model `VendorAPIRequest` where request and response info lives. (See https://github.com/DFE-Digital/apply-for-teacher-training/pull/2999) for migration
- Adds middleware to intercept vendor API requests and enqueue them for processing by a Sidekiq worker.
- Provides support users with a basic UI to list and filter logged API requests

## Guidance to review

It would be good to componentise some of the view here. This might be simpler once we have a `DetailsComponent`.

## Link to Trello card

https://trello.com/c/LsxYKL3n/2815-build-vendor-api-request-tracing

## Things to check

### DIFF
--- a/app/controllers/support_interface/vendor_api_requests_controller.rb
+++ b/app/controllers/support_interface/vendor_api_requests_controller.rb
@@ -1,0 +1,7 @@
+module SupportInterface
+  class VendorAPIRequestsController < SupportInterfaceController
+    def index
+      @vendor_api_requests = VendorAPIRequest.all.order(created_at: :desc)
+    end
+  end
+end
--- a/app/controllers/support_interface/vendor_api_requests_controller.rb
+++ b/app/controllers/support_interface/vendor_api_requests_controller.rb
@@ -2,14 +2,18 @@ module SupportInterface
   class VendorAPIRequestsController < SupportInterfaceController
     def index
       @filter = SupportInterface::VendorAPIRequestsFilter.new(params: params)
-      @vendor_api_requests = VendorAPIRequest.order(created_at: :desc).page(params[:page] || 1).per(15)
+      @vendor_api_requests = VendorAPIRequest.includes(:provider).order(created_at: :desc).page(params[:page] || 1).per(15)
 
       if params[:q]
-        @vendor_api_requests = @vendor_api_requests.where("CONCAT(request_path, ' ', request_body, ' ', response_body, ' ', hashed_token) ILIKE ?", "%#{params[:q]}%")
+        @vendor_api_requests = @vendor_api_requests.where("CONCAT(request_path, ' ', request_body, ' ', response_body) ILIKE ?", "%#{params[:q]}%")
       end
 
       if params[:status_code]
         @vendor_api_requests = @vendor_api_requests.where('status_code IN (?)', params[:status_code])
+      end
+
+      if params[:provider_id]
+        @vendor_api_requests = @vendor_api_requests.where('provider_id IN (?)', params[:provider_id])
       end
 
       %w[created_at request_path].each do |column|
--- a/app/controllers/support_interface/vendor_api_requests_controller.rb
+++ b/app/controllers/support_interface/vendor_api_requests_controller.rb
@@ -1,7 +1,22 @@
 module SupportInterface
   class VendorAPIRequestsController < SupportInterfaceController
     def index
-      @vendor_api_requests = VendorAPIRequest.all.order(created_at: :desc)
+      @filter = SupportInterface::VendorAPIRequestsFilter.new(params: params)
+      @vendor_api_requests = VendorAPIRequest.order(created_at: :desc).page(params[:page] || 1).per(15)
+
+      if params[:q]
+        @vendor_api_requests = @vendor_api_requests.where("CONCAT(request_path, ' ', request_body, ' ', response_body, ' ', hashed_token) ILIKE ?", "%#{params[:q]}%")
+      end
+
+      if params[:status_code]
+        @vendor_api_requests = @vendor_api_requests.where('status_code IN (?)', params[:status_code])
+      end
+
+      %w[created_at request_path].each do |column|
+        next unless params[column]
+
+        @vendor_api_requests = @vendor_api_requests.where(column => params[column])
+      end
     end
   end
 end
--- a/app/middlewares/vendor_api_request_middleware.rb
+++ b/app/middlewares/vendor_api_request_middleware.rb
@@ -46,6 +46,10 @@ private
     @request.env.slice(*REQUEST_HEADER_KEYS)
   end
 
+  def trace_request?
+    FeatureFlag.active?('vendor_api_request_tracing') && vendor_api_path?
+  end
+
   def vendor_api_path?
     @request.path =~ /^\/api\/.*$/
   end
--- a/app/middlewares/vendor_api_request_middleware.rb
+++ b/app/middlewares/vendor_api_request_middleware.rb
@@ -1,0 +1,52 @@
+require 'redis'
+require './app/workers/vendor_api_request_worker'
+
+class VendorAPIRequestMiddleware
+  REQUEST_HEADER_KEYS = %w[
+    HTTP_VERSION
+    HTTP_HOST
+    HTTP_USER_AGENT
+    HTTP_ACCEPT
+    HTTP_ACCEPT_ENCODING
+    HTTP_AUTHORIZATION
+    HTTP_CONNECTION
+    HTTP_CACHE_CONTROL
+  ].freeze
+
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    @request = Rack::Request.new(env)
+    status, headers, response = @app.call(env)
+
+    begin
+      if trace_request?
+        VendorAPIRequestWorker.perform_async(request_data, response.body, status, Time.zone.now)
+      end
+    rescue Redis::BaseError => e
+      Rails.logger.warn e.message
+    end
+
+    [status, headers, response]
+  end
+
+private
+
+  def request_data
+    {
+      path: @request.path,
+      params: @request.params,
+      headers: request_headers,
+    }
+  end
+
+  def request_headers
+    @request.env.slice(*REQUEST_HEADER_KEYS)
+  end
+
+  def vendor_api_path?
+    @request.path =~ /^\/api\/.*$/
+  end
+end
--- a/app/middlewares/vendor_api_request_middleware.rb
+++ b/app/middlewares/vendor_api_request_middleware.rb
@@ -23,7 +23,7 @@ class VendorAPIRequestMiddleware
 
     begin
       if trace_request?
-        VendorAPIRequestWorker.perform_async(request_data, response.body, status, Time.zone.now)
+        VendorAPIRequestWorker.perform_async(request_data, body_from(response), status, Time.zone.now)
       end
     rescue Redis::BaseError => e
       Rails.logger.warn e.message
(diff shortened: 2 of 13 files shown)


app/controllers/support_interface/vendor_api_requests_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+
+      if params[:q]
+        @vendor_api_requests = @vendor_api_requests.where("CONCAT(request_path, ' ', request_body, ' ', response_body, ' ', hashed_token) ILIKE ?", "%#{params[:q]}%")
+      end
+
+      if params[:status_code]
+        @vendor_api_requests = @vendor_api_requests.where('status_code IN (?)', params[:status_code])
+      end
+
+      %w[created_at status_code request_path].each do |column|
```
tijmenb, line 15: Should `status_code` be in here? I think it might conflict with the `where('status_code IN (?)', params[:status_code])` line.

app/workers/vendor_api_request_worker.rb
```diff
@@ -0,0 +1,37 @@
+class VendorAPIRequestWorker
+  include Sidekiq::Worker
+  sidekiq_options retry: 3, queue: :low_priority
+
+  def perform(request_data, response_body, status_code, response_time, created_at)
+    request_headers = request_data['headers']
+    hashed_token = hashed_auth_token(request_headers.delete('HTTP_AUTHORIZATION'))
+
+    VendorAPIRequest.create(
```
tijmenb, line 9:
```suggestion
    VendorAPIRequest.create!(
```

app/workers/vendor_api_request_worker.rb
```diff
@@ -0,0 +1,37 @@
...
+      request_path: request_data['path'],
+      request_headers: request_headers,
+      request_body: request_data['params'],
+      response_body: response_hash(response_body, status_code),
+      status_code: status_code,
+      response_time: response_time,
+      hashed_token: hashed_token,
```
tijmenb, line 16: Still a bit unsure about this. Should we look up the provider and save that? Or alternatively a link to the VendorApiToken?

app/middlewares/vendor_api_request_middleware.rb
```diff
@@ -0,0 +1,57 @@
...
+  def call(env)
+    started_at = Time.zone.now
+
+    @request = Rack::Request.new(env)
+    status, headers, response = @app.call(env)
+
+    begin
+      if vendor_api_path?
+        finished_at = Time.zone.now
+        response_time = (finished_at - started_at) * 1000
```
tijmenb, line 29: Will this be accurate if there are middleware that wrap this middleware?

app/middlewares/vendor_api_request_middleware.rb
```diff
@@ -0,0 +1,57 @@
...
+
+        VendorAPIRequestWorker.perform_async(request_data, response.body, status, response_time, Time.zone.now)
```
tijmenb, line 31: Can we add a feature flag here so we can quickly disable if this causes trouble?

tijmenb (commented):
Looking good! Couple of questions.

app/workers/vendor_api_request_worker.rb
steventux: :+1: Yeah I think it's possible to look up the api user based on this value. Will add.

app/middlewares/vendor_api_request_middleware.rb
steventux: No it's not going to account for time taken by any subsequent middleware. @duncanjbrown asked the question (in the spike PR) of whether we needed to record this as Skylight should have a more accurate metric. Happy to remove this and the `response_time` db field if we think it's better to leave timings out.

duncanjbrown: I think it's better to leave out the timings if they could be misleading

steventux: See https://github.com/DFE-Digital/apply-for-teacher-training/pull/3029 for db amends

app/middlewares/vendor_api_request_middleware.rb
```diff
@@ -0,0 +1,56 @@
...
+    @app = app
+  end
+
+  def call(env)
+    @request = Rack::Request.new(env)
+    status, headers, response = @app.call(env)
+
+    begin
+      if trace_request?
+        VendorAPIRequestWorker.perform_async(request_data, response.body, status, Time.zone.now)
```
mnacos, line 26: `response.body` will fail here, as `response` in this context is just an array of output lines, e.g. ` ['<html><body>You are being <a href=\"http://localhost:3000/support/candidates\">redirected</a>.</body></html>']`

Passing `response` as an argument to `perform_async` may still work, though; it's just an array. 

Another concern is the size of enqueued messages, e.g. if response is a json list of applications. I thought we would only store responses if the API call is unsuccessful?

app/middlewares/vendor_api_request_middleware.rb
```diff
@@ -0,0 +1,56 @@
...
+      end
+    rescue Redis::BaseError => e
```
mnacos, line 28: nice

mnacos (commented):
Very promising. I've checked and our current sidekiq configuration (without weights) will process the `default` and `mailers` queues ahead of any `low_priority` buildup, so the en-queuing  approach carries far less risk than I originally thought.

Just a couple of comments on the middleware about what is and isn't available in that context and the need for a middleware spec.

app/middlewares/vendor_api_request_middleware.rb
```diff
@@ -0,0 +1,56 @@
...
+      Rails.logger.warn e.message
```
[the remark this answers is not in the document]
steventux: @mnacos this appears to be working as expected. I certainly see logged messages coming from the middleware in my development console.

app/middlewares/vendor_api_request_middleware.rb
steventux: Ah yes we need to mitigate this similarly to https://github.com/pdfkit/pdfkit/blob/master/lib/pdfkit/middleware.rb#L24-L25

steventux:
@tijmenb @mnacos thanks for reviews I think I've covered the points raised now.

app/models/support_interface/vendor_api_requests_filter.rb
```diff
@@ -0,0 +1,58 @@
...
+        name: 'status_code',
+        options: options,
+      }
+    end
+
+    def provider
+      options = VendorAPIRequest.includes(:provider).distinct(:provider_id).pluck(:provider_id).compact.map do |provider_id|
+        {
+          value: provider_id,
+          label: Provider.find(provider_id).name,
```
tijmenb, line 45: This looks like an N+1. We should be able to get all providers in one request I think.

app/workers/vendor_api_request_worker.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+private
+
+  def provider_id_from_auth_token(auth_header)
+    return if auth_header.blank?
+
+    token = auth_header[7..-1]
+    hashed_token = Devise.token_generator.digest(VendorAPIToken, :hashed_token, token)
+    VendorAPIToken.find_by(hashed_token: hashed_token)&.provider_id
```
tijmenb, line 27: You could use `VendorAPIToken.find_by_unhashed_token` here, which does the hashing of the token.

spec/system/support_interface/view_vendor_api_requests_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+  end
+
+  def and_some_applications_exist
+    applications = create_list(:application_choice, 2)
+    @first_application_choice = applications.first
+    @last_application_choice = applications.last
+  end
+
+  def and_vendor_api_requests_for_applications_have_been_made
+    visit vendor_api_path(@first_application_choice)
```
tijmenb, line 35: Can we add a test for the happy path here as well? An authenticated request that results in a 200, that we can filter via the provider filters.

app/workers/vendor_api_request_worker.rb
```diff
@@ -0,0 +1,37 @@
...
+      created_at: created_at,
+    )
+  end
+
+private
+
+  def provider_id_from_auth_token(auth_header)
+    return if auth_header.blank?
+
+    token = auth_header[7..-1]
```
tijmenb, line 25: Not sure, but perhaps we can use the same mechanism as the controller to extract the token?

See https://github.com/DFE-Digital/apply-for-teacher-training/commit/36c71e86d37e4e4fbffee969ee839edc4645edf3

Otherwise it's possible that the middlware doesn't recognise tokens that the controller does - I think Rails' `authenticate_with_http_token` is more liberal in what it accepts.

tijmenb (commented):
Looking good! Couple of last questions.

app/workers/vendor_api_request_worker.rb
tijmenb: Feel free to ignore if it doesn't work.

steventux: Thanks @tijmenb yeah it seems prudent to be consistent with the controller. 
`authenticate_with_http_token` calls an underlying method `ActionController::HttpAuthentication::Token#token_and_options` https://github.com/rails/rails/blob/070d4afacd3e9721b7e3a4634e4d026b5fa2c32c/actionpack/lib/action_controller/metal/http_authentication.rb#L457 which seems like a better fit for a Sidekiq worker. Added here https://github.com/DFE-Digital/apply-for-teacher-training/pull/3009/commits/da01857080650c9abf1027a6d78fb7f2e805c7c5#diff-ea82dc5cde409c49f45d4efc687841c9R29

spec/system/support_interface/view_vendor_api_requests_spec.rb
steventux: :+1: https://github.com/DFE-Digital/apply-for-teacher-training/pull/3009/commits/dd73730484c46cbcf741b859b7ee098de8789b9b#diff-780c12b896f5644b1bb63115000759feR20-R27

app/workers/vendor_api_request_worker.rb
steventux: :+1: https://github.com/DFE-Digital/apply-for-teacher-training/pull/3009/files#diff-ea82dc5cde409c49f45d4efc687841c9R29-R30

app/models/support_interface/vendor_api_requests_filter.rb
steventux: :+1: Strange this didn't get picked up by the lazy-loading linter, anyway I've updated to https://github.com/DFE-Digital/apply-for-teacher-training/pull/3009/files#diff-2c9f873e52850265633344b9c21301c8R42

app/models/support_interface/vendor_api_requests_filter.rb
```diff
@@ -0,0 +1,58 @@
...
+      {
+        type: :checkboxes,
+        heading: 'Status code',
+        name: 'status_code',
+        options: options,
+      }
+    end
+
+    def provider
+      options = Provider.where(id: VendorAPIRequest.distinct(:provider_id).pluck(:provider_id).compact).map do |provider|
```
tijmenb, line 42: I think this makes it one query instead of 2:

```suggestion
      options = Provider.where(id: VendorAPIRequest.distinct(:provider_id).select(:provider_id)).map do |provider|
```

app/models/support_interface/vendor_api_requests_filter.rb
tijmenb: Yeah, I don't think it works very well.

tijmenb (approved):
👍

PR #3011, opened 2020-09-29
Update to back links across work history section

george-bit:
## Context

Currently on Add job page when adding a new job the nav link is a "back to application" link when it should be a back link to the length section. Also, previous back link solution using JS needs to be refactored less ambiguously.

Links updated across the section as required

image

## Guidance to review

Test in review app to check that section routing now acceptable

## Link to Trello card

## Things to check


paulrobertlloyd:
**Adding a job**

* ‘Back’ link when adding a job should only go back to length question for the first job you add. Subsequent jobs it should take you to the section review page (as you have already answered length question, and section review page is the most sensible other destination).

PR #3013, opened 2020-09-29
Add new flow for decoupled references (part 1)

davidgisbey:
## Context

This is the initial flow for decoupling references. This PR covers changing the application form show page through to a candidate being asked to input the type of reference.

- Update the application show page to use the new design

|Before|After|
|------------|------------|
|image|image|

- Add the start page 

image

## Guidance to review

I'm going to keep these PRs small as I think it's going to end up being a fairly substantial piece of work.

Out of scope adding a type, name, email, description and review page

prototype - https://apply-beta-prototype.herokuapp.com/application/P3TB2

## Link to Trello card

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/referees_controller.rb
+++ b/app/controllers/candidate_interface/referees_controller.rb
@@ -14,6 +14,8 @@ module CandidateInterface
       end
     end
 
+    def start; end
+
     def type
       if params[:id]
         set_referee_id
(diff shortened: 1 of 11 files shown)


paulrobertlloyd (approved):
LGTM – love how this ties into the existing flow.

One question; are we making sure to delete old and deprecated translation strings etc. as we do this rebuild?

malcolmbaig:
Thinking about how to structure this work and potentially make it easier to work on parts in parallel. 

Is it worth implementing decoupled references in a separate set of route definitions and controllers? Something like https://github.com/DFE-Digital/apply-for-teacher-training/pull/3017

It would mean less branching logic in controllers and views, and will probably make deleting the legacy implementation a bit easier further down the road. WDYT? @davidgisbey @stevehook @tvararu

PR #3015, opened 2020-09-30
Fix rescuing errors when syncing with Find

tijmenb:
## Context

The error is currently named `FindSync::SyncAllProvidersFromFind::SyncFindApiError` because it was moved into a namespace. This means we no longer ignore the error in Sentry.

This updates the error to have a better name, rescue exceptions in the new worker as well, and make sure we ignore
this error in Sentry.

## Guidance to review

Ok?

## Link to Trello card

https://trello.com/c/HwAFHcKS/1965-tech-performance-sync-all-of-the-courses-from-find

## Things to check

### DIFF
--- a/app/services/find_sync/find_sync_error.rb
+++ b/app/services/find_sync/find_sync_error.rb
@@ -1,0 +1,3 @@
+module FindSync
+  class SyncError < StandardError; end
+end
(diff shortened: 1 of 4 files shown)


mnacos (approved):
:+1:

PR #3016, opened 2020-09-30
Remove not null constraint from the references email address table

davidgisbey:
## Context

For the decoupled references flow we need to be able to save a referees details incrementally as they add one attribute per page. This means that the validations on the model will need to be changed/removed. This validation directly on the DB column needs to be removed.

- remove the not null constraint on the references email address column

## Link to Trello card

## Things to check


stevehook (approved):
This kind of design does mean that you end can (and will) have half-completed (invalid) references in the DB that need to be handled. It might be worth considering using a session state backed wizard if that is likely to cause problems.

davidgisbey:
Good spot. I think just making it irreversible should be okay for now right?

PR #3019, opened 2020-09-30
Check provider association before allowing access to 'reconfirm offer' wizard

mnacos:
## Context

The controller does a straight `.find` on the application choice id, which means a provider could theoretically guess the url of a deferred offer that doesn't belong to them and attempt to reinstate it.

Back-end services already check the association via their authorisation calls, so the issue is limited to UI access.

Fix this

## Guidance to review

Straightfoward

## Link to Trello card

No Trello card.

## Things to check

### DIFF
--- a/app/controllers/provider_interface/reconfirm_deferred_offers_controller.rb
+++ b/app/controllers/provider_interface/reconfirm_deferred_offers_controller.rb
@@ -79,7 +79,9 @@ module ProviderInterface
   private
 
     def find_application_choice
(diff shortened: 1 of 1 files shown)


tijmenb:
I've solved this differently in https://github.com/DFE-Digital/apply-for-teacher-training/pull/3020/files#diff-0914a7f5731d890204d8f73f8a3bbf0d, is it okay if we use that one?

mnacos:
Sure, closing...

PR #3020, opened 2020-09-30
Increase debuggability of provider interface problems

tijmenb:
## Context

This PR addresses a couple of issues that would have made debugging https://github.com/DFE-Digital/apply-for-teacher-training/pull/3010 easier.

- In Sentry and Logit, add a support UI link to the provider user that made the request (currently we only log the DSI UUID, which you then have to look up)
- If the request relates to an application choice, add a link to the Support UI in Sentry and Logit. Currently we only have the `application_choice_id`, which is tricky to quickly match to a application form
- The issue solved in https://github.com/DFE-Digital/apply-for-teacher-training/pull/3010 was masked by the fact that we rescue `ActiveRecord::RecordNotFound` errors, regardless of where they occur. This PR makes the rescuing more specific so that we would have been notified via Sentry of the problem.

To do this, I've extracted a common `set_application_choice` method that also sets the debugging info. This solves a data scoping issue as well (https://github.com/DFE-Digital/apply-for-teacher-training/commit/12fd12aab5c7742ece8e2e561a65420ad902a276).

## Link to Trello card

https://trello.com/c/5YqXBdtK/2817-fix-possible-bug-that-prevents-providers-from-seeing-certain-courses

## Things to check

### DIFF
--- a/app/controllers/provider_interface/application_choices_controller.rb
+++ b/app/controllers/provider_interface/application_choices_controller.rb
@@ -1,6 +1,7 @@
 module ProviderInterface
   class ApplicationChoicesController < ProviderInterfaceController
-    before_action :set_application_choice_and_sub_navigation_items, except: %i[index]
+    before_action :set_application_choice, except: %i[index]
+    before_action :set_sub_navigation_items, except: %i[index]
 
     def index
       @filter = ProviderApplicationsFilter.new(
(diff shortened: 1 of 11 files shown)


mnacos (approved):
Very cool

tvararu (approved):

PR #3022, opened 2020-09-30
Create a separate DecoupledRefereesController for the new decoupled references implementation

davidgisbey:
## Context

It makes sense to implement this flow completely independently so they codebase doesn't get too entwined.

See https://ukgovernmentdfe.slack.com/archives/CN1MCQCHZ/p1601459837044900

- Create a new controller for the flow
- Move the relevant routes/views over

## Guidance to review

For now, I've just added decoupled_ before the controller/routes. These can all be removed once the flag is removed.

Any thoughts on naming would be great 👍 

## Link to Trello card

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/decoupled_referees_controller.rb
+++ b/app/controllers/candidate_interface/decoupled_referees_controller.rb
@@ -1,0 +1,11 @@
+module CandidateInterface
+  class DecoupledRefereesController < CandidateInterfaceController
+    def start; end
+
+    def type
+      @reference_type_form = Reference::RefereeTypeForm.new
+    end
+
+    def update_type; end
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -393,11 +393,11 @@ Rails.application.routes.draw do
         patch '/cancel/:id' => 'referees#cancel', as: :cancel_referee
       end
 
-      scope '/decoupled-referees' do
-        get '/start' => 'decoupled_referees#start', as: :decoupled_referees_start
+      scope '/references' do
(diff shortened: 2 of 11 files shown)


app/controllers/candidate_interface/decoupled_referees_controller.rb
```diff
@@ -0,0 +1,11 @@
+module CandidateInterface
+  class DecoupledRefereesController < CandidateInterfaceController
```
malcolmbaig, line 2: The resource we're creating is a Reference, so maybe `DecoupledReferences`? Consistent with the way we're describing the feature too.

config/routes.rb
```diff
@@ -393,6 +393,13 @@
         patch '/cancel/:id' => 'referees#cancel', as: :cancel_referee
       end
 
+      scope '/decoupled-referees' do
```
malcolmbaig, line 396: I like the prefixing of controllers and path helpers for clarity while this features is in-progress, but maybe the scope could just be `/references`? That way no visible change to any clients of this code when we eventually come back and remove the feature flag.

app/controllers/candidate_interface/decoupled_referees_controller.rb
davidgisbey: Yea i agree with this 👍

config/routes.rb
davidgisbey: And also this. I'll make the changes now

malcolmbaig (approved):

tvararu (approved):
Happy to merge like this now and remove the `Decoupled` later :+1:

PR #3024, opened 2020-09-30
Tidy up error handling in DeferOffer

tijmenb:
## Context & changes

The error handling here is not consistent.

- We return `false` and populate the `errors` if the state transition is illegal. This is something that shouldn't happen in the normal user flow, so it's legitimate to raise an error that will notify us. Currently the provider user would see a message "The application is not ready for that action", which isn't helpful
- We're using `update` to set `status_before_deferral` and `offer_deferred_at`, which will return false if the validation fails, which we ignore - because we just return true in the last line of the method. This makes sure that we fail if the update fails.
- Rename the `save` method to indicate that it'll raise an error if it fails.
- Remove the error checking in the controller because it will never be called anyway.

## Things to check

### DIFF
--- a/app/controllers/provider_interface/decisions_controller.rb
+++ b/app/controllers/provider_interface/decisions_controller.rb
@@ -143,18 +143,13 @@ module ProviderInterface
     end
 
     def defer_offer
-      @defer_offer = DeferOffer.new(
+      DeferOffer.new(
         actor: current_provider_user,
         application_choice: @application_choice,
-      )
-      if @defer_offer.save
-        flash[:success] = 'Offer successfully deferred'
-        redirect_to provider_interface_application_choice_path(
-          application_choice_id: @application_choice.id,
-        )
-      else
-        render action: :new_defer_offer
-      end
+      ).save!
+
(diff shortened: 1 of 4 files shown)


mnacos (approved):
I agree. We should probably update all of the other services, too.

```
app/services/conditions_not_met.rb:  rescue Workflow::NoTransitionAllowed
app/services/confirm_offer_conditions.rb:  rescue Workflow::NoTransitionAllowed
app/services/reinstate_pending_conditions.rb:  rescue Workflow::NoTransitionAllowed
app/services/make_an_offer.rb:  rescue Workflow::NoTransitionAllowed
app/services/defer_offer.rb:  rescue Workflow::NoTransitionAllowed
app/services/reinstate_conditions_met.rb:  rescue Workflow::NoTransitionAllowed
app/services/withdraw_offer.rb:  rescue Workflow::NoTransitionAllowed
app/services/reject_application.rb:  rescue Workflow::NoTransitionAllowed
```

tijmenb:
@mnacos good point - added https://trello.com/c/gxe5NsBz/375-inconsistent-error-handling-in-services

PR #3031, opened 2020-09-30
Send DSI API errors to Sentry

duncanjbrown:
## Context

From Slack thread where Bill found a user who was created in Manage but not in DSI. Need more information to understand what went on here, so this should help us in the future.

We currently swallow these exceptions and don't report them to sentry, preferring to show the user a flash in the form.

Send them to Sentry so we know about them.

## Guidance to review

Sending them to Sentry right?

## Link to Trello card

https://ukgovernmentdfe.slack.com/archives/CPH8J9G65/p1601469929065800

## Things to check


steventux (approved):
:+1:

PR #3033, opened 2020-09-30
Support: Provider/Provider User page layout improvements

paulrobertlloyd:
## Context

Tidy up of navigation and layout for provider and provider user views, including removing repetition of page titles three times 😮

(Was there any reason why we show provider user permissions within a scrollable `div`? Doing this creates two scroll separate vertical scrollbars on the page, making this more difficult to use.)

| Before | After |
| - | - |
| provider-details-before | provider-details-after |
| provider-user-before | provider-user-after |

## Guidance to review

Updated `ProviderUsersController` to provide same page variables for `edit` and `audits`; are my changes correct?

## Things to check

### DIFF
--- a/app/views/support_interface/provider_users/_provider_user_navigation.html.erb
+++ b/app/views/support_interface/provider_users/_provider_user_navigation.html.erb
@@ -1,0 +1,9 @@
+<% content_for :title do %>
+  <span class="govuk-caption-xl">Provider user</span>
+  <%= @form.provider_user.full_name %>
+<% end %>
+
+<%= render TabNavigationComponent.new(items: [
+  { name: 'Details', url: edit_support_interface_provider_user_path(@form.provider_user), current: local_assigns[:current] == 'Details' },
+  { name: 'History', url: support_interface_provider_user_audits_path(@form.provider_user), current: local_assigns[:current] == 'History' },
+]) %>
--- a/app/views/support_interface/provider_users/_provider_user_navigation.html.erb
+++ b/app/views/support_interface/provider_users/_provider_user_navigation.html.erb
@@ -4,6 +4,6 @@
 <% end %>
 
 <%= render TabNavigationComponent.new(items: [
-  { name: 'Details', url: edit_support_interface_provider_user_path(@form.provider_user), current: local_assigns[:current] == 'Details' },
-  { name: 'History', url: support_interface_provider_user_audits_path(@form.provider_user), current: local_assigns[:current] == 'History' },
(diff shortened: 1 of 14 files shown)


duncanjbrown:
The scrollable `div` is just because the list can be very long, I think. It's only providers with synced courses but that list is about to get quite a lot longer.

One thing I always miss on this page is, because the list is long, it's hard to find the providers the person is actually associated with. Would it be good to bubble those to the top of the list? Probably shortening the list would be better, though.

app/views/support_interface/provider_users/_provider_user_navigation.html.erb
```diff
@@ -0,0 +1,9 @@
+<% content_for :title do %>
+  <span class="govuk-caption-xl">Provider user</span>
+  <%= @form.provider_user.full_name %>
+<% end %>
+
+<%= render TabNavigationComponent.new(items: [
+  { name: 'Details', url: edit_support_interface_provider_user_path(@form.provider_user), current: local_assigns[:current] == 'Details' },
+  { name: 'History', url: support_interface_provider_user_audits_path(@form.provider_user), current: local_assigns[:current] == 'History' },
```
tijmenb, line 8: I don't think the `current:` stuff is necessary here, the `TabNavigationComponent` should automatically highlight the current page (you can use `current:` to highlight tabs that link to a different page, for example if you're on a subpage).

tijmenb (commented):
Looking good! Minor comment.

@duncanjbrown I'm looking at the provider list as part of syncing all providers so we can leave that for later.

app/views/support_interface/provider_users/_provider_user_navigation.html.erb
paulrobertlloyd: Ah, nice! Updated.

paulrobertlloyd:
@adamsilver reminded me why we have the scrollable `div` – it’s so the button doesn’t get lost at the end of a long, long page. So, reinstated the scrollable area, but made it more explicit with visible scrollbar and border, and made it so that it takes up 75% of the viewport height.

Screenshot 2020-10-01 at 11 11 58

adamsilver (approved):
Approved regardless, but could you use a breadcrumb instead of back link here?

paulrobertlloyd:
> Approved regardless, but could you use a breadcrumb instead of back link here?

Yeah, was thinking the same. We don’t have a breadcrumb component yet, but I may have a go at creating one 😀

PR #3035, opened 2020-10-01
Decoupled reference flow (part2)

davidgisbey:
## Context

Previous PRs are: 

This is the second part of the add referee flow and covers submitting a references type and name. 

- Updates the TypeForm's save method to be called update
- Adds a new save method to the TypeForm which creates a new reference with a specific type
- Removes validations from the references model which require name/email/description to be present on save
- Refactors the type action into their own controller 
 - Adds a new controller, routes, view and NameForm for the name page

image

## Guidance to review

Does the approach seem okay?

## Link to Trello card

## Things to check


app/models/application_reference.rb
```diff
@@ -3,13 +3,14 @@ class ApplicationReference < ApplicationRecord
 
   self.table_name = 'references'
 
-  validates :name, presence: true, length: { minimum: 2, maximum: 200 }
+  validates :name, presence: true, length: { minimum: 2, maximum: 200 }, unless: -> { FeatureFlag.active?('decoupled_references') }
```
tvararu, line 6: Instead of `unless: ->`, could we use a validation scope?

app/forms/candidate_interface/reference/referee_name_form.rb
```diff
@@ -0,0 +1,19 @@
+module CandidateInterface
+  class Reference::RefereeNameForm
+    include ActiveModel::Model
+
+    attr_accessor :name
+
+    validates :name, presence: true
```
tvararu, line 7: This is a replacement for the validation that is `unless`'s in the main `application_reference` model, but it's less restrictive. It should also check for length, like the other one.

What if we kept the validation in the model, but used validation scopes? (mentioned below as well)

app/views/candidate_interface/decoupled_references/name/new.html.erb
```diff
@@ -0,0 +1,18 @@
...
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with model: @reference_name_form, url: candidate_interface_decoupled_references_create_name_path(@reference.id) do |f| %>
+      <%= f.govuk_error_summary %>
+
+      <h1 class="govuk-heading-xl">
+        <%= t('page_titles.references_name') %>
+      </h1>
+
+      <%= f.govuk_text_field :name, label: { text: '' }  %>
```
tvararu, line 13: Empty label? Can the label be the H1?

app/models/application_reference.rb
tvararu: Nevermind

tvararu (approved):

PR #3037, opened 2020-10-01
Support navigation (performance, settings, docs)

paulrobertlloyd:
## Context

Plan is to use navigation below the header for primary navigation, but in order to do that, pages further down the navigation hierarchy need to be adapted first. This is the first step towards enabling that (candidate and provider sections to follow separately)

| Before | After |
| - | - |
| support-docs-before | support-docs-after |
| support-performance-before | support-performance-after |
| support-settings-before | support-settings-after |

## Things to check


adamsilver (approved):
Shipito

PR #3038, opened 2020-10-01
Ensure presence of equivalency_details in API responses

mnacos:
## Context

Trivial PR, as `equivalency_details` is already included in API responses.

Add spec verifying `equivalency_details` is part of the API responses.

## Guidance to review

Easiest review ever.

## Link to Trello card

https://trello.com/c/bOYfMiKj

## Things to check

### DIFF
--- a/app/presenters/vendor_api/single_application_presenter.rb
+++ b/app/presenters/vendor_api/single_application_presenter.rb
@@ -207,8 +207,18 @@ module VendorAPI
         award_year: qualification.award_year,
         institution_details: institution_details(qualification),
         awarding_body: qualification.awarding_body,
-        equivalency_details: qualification.equivalency_details,
+        equivalency_details: composite_equivalency_details(qualification),
       }.merge HesaQualificationFieldsPresenter.new(qualification).to_hash
+    end
+
+    def composite_equivalency_details(qualification)
+      details = [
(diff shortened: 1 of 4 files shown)


duncanjbrown:
Discussed offline, we're going to make a composite field from `naric_reference` and `comparable_uk_degree`

app/presenters/vendor_api/single_application_presenter.rb
```diff
@@ -206,10 +206,20 @@ def qualification_to_hash(qualification)
...
         awarding_body: qualification.awarding_body,
-        equivalency_details: qualification.equivalency_details,
+        equivalency_details: composite_equivalency_details(qualification),
       }
     end
 
+    def composite_equivalency_details(qualification)
+      details = [
+        ("Naric: #{qualification.naric_reference}" if qualification.naric_reference),
+        ("- #{qualification.comparable_uk_qualification} -" if qualification.comparable_uk_qualification),
```
duncanjbrown, line 216: Is it right to bake the dashes into this string? What if it's missing. Should we rather `join(' - ')`?

mnacos: If it's missing, the `if qualification.comparable_uk_qualification` makes it `nil`. But we might as well use `join(' - ')`. :+1:

duncanjbrown (approved):

PR #3039, opened 2020-10-01
Only send completed references over the API

duncanjbrown:
We were sending all references, including ones that had been cancelled
or had no feedback.

## Context

Pointed out by an HEI 😬

Put a scope on the application_references in the API


duncanjbrown:
> Could we have caught this in another way?

I think if we were running our own client we would have noticed pretty quickly

PR #3040, opened 2020-10-01
Decoupled reference flow (part3)

davidgisbey:
## Context

Previous PRs are:
#3035 - Up to adding a referees email 

This PR adds the email address view, controller actions, form and routes. The description page onwards is out of scope.

image

- With no value 

image

- With the incorrect format 

image

- With a duplicate email

image

## Guidance to review

I'm not overly happy with the validations. I'd love some input on them if anyone has any ideas.

## Link to Trello card

## Things to check

### DIFF
--- a/app/views/candidate_interface/decoupled_references/email_address/new.html.erb
+++ b/app/views/candidate_interface/decoupled_references/email_address/new.html.erb
@@ -4,7 +4,7 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <%= form_with model: @reference_email_address_form, url: candidate_interface_decoupled_references_create_email_path(@reference.id) do |f| %>
+    <%= form_with model: @reference_email_address_form, url: candidate_interface_decoupled_references_create_email_address_path(@reference.id) do |f| %>
       <%= f.govuk_error_summary %>
 
       <h1 class="govuk-heading-xl">
--- a/app/views/candidate_interface/decoupled_references/email_address/new.html.erb
+++ b/app/views/candidate_interface/decoupled_references/email_address/new.html.erb
@@ -7,13 +7,6 @@
     <%= form_with model: @reference_email_address_form, url: candidate_interface_decoupled_references_create_email_address_path(@reference.id) do |f| %>
       <%= f.govuk_error_summary %>
 
-      <h1 class="govuk-heading-xl">
-        <span class="govuk-caption-xl">
-          <%= @reference.name %>
-        </span>
-        <%= t('page_titles.references_email_address') %>
-      </h1>
-
       <% if HostingEnvironment.sandbox_mode? %>
         <div class="app-status-box app-status-box--sandbox govuk-!-margin-bottom-4">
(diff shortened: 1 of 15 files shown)


app/views/candidate_interface/decoupled_references/email_address/new.html.erb
```diff
@@ -0,0 +1,34 @@
...
+            <strong class="govuk-tag govuk-phase-banner__content__tag app-environment-tag govuk-tag--purple">
+              sandbox feature
+            </strong>
+          </p>
+          <p class="govuk-body govuk-!-margin-bottom-0">
+            Enter <code>refbot1@example.com</code> and <code>refbot2@example.com</code> as your referee email addresses to have the references automatically completed
+          </p>
+        </div>
+      <% end %>
+      <%= f.govuk_text_field :email_address, type: 'email', label: { text: '' }, hint_text: 'In most cases, this should be a work address.' %>
```
tvararu, line 29: Does the label need to be empty? Can it be the <h1> above somehow?

config/routes.rb
```diff
@@ -394,10 +392,18 @@
       end
 
       scope '/references' do
-        get '/start' => 'decoupled_references#start', as: :decoupled_references_start
+        get '/start' => 'decoupled_references/base#start', as: :decoupled_references_start
+
+        get '/type' => 'decoupled_references/type#new', as: :decoupled_references_new_type
+        post '/update-type' => 'decoupled_references/type#create', as: :decoupled_references_create_type
```
malcolmbaig, line 398: This is something we can circle back to if we're low on time, but some suggestions about the routing:

```
      get '/type' => 'decoupled_references/type#new', as: :decoupled_references_type
      post '/type' => 'decoupled_references/type#create'
```
Means you only need the one path helper and can use its path for both GETs and POSTs.

config/routes.rb
```diff
@@ -394,10 +392,18 @@
...
+
+        get '/name/:id' => 'decoupled_references/name#new', as: :decoupled_references_new_name
+        post '/name/:id' => 'decoupled_references/name#create', as: :decoupled_references_create_name
```
malcolmbaig, line 401: Making the `post` a `patch` means we use the correct HTTP verb to get here, and provides another hint that we're updating an existing resource at this endpoint. Also, an `edit/update` action pair is probably more accurate than `new/create`.

davidgisbey: Okay cool. We need additional actions to go back and edit/update their original choice, then redirect to the unsubmitted review page. I think for now it's probs best to leave it at this until the flow is working. Then I'll come back and make the requested changes. I've added them to the card.

config/routes.rb
davidgisbey: Same as below 👍

malcolmbaig (approved):

app/views/candidate_interface/decoupled_references/email_address/new.html.erb
davidgisbey: Done 👍

PR #3041, opened 2020-10-01
Stop candidates being able to create incomplete other qualifications

stevehook:
## Context

Because other qualifications are saved as soon as the candidate fills in just the qualification type (e.g. A-level) it's possible for drop out of the normal application flow and end up with incomplete (invalid) qualifications.

This PR was originally raised several weeks back and paused. It's been rebased and re-tested as of 2020-11-09 and had a number of missing/broken bits fixed.

The 'ideal' way to solve this is to defer persistence of other qualifications until the candidate has filled in the last (of two) forms. To do this we are using the Redis-backed state storage model that ProVendor have successfully adopted recently.

- [x] Add a `OtherQualificationWizard` to handle un-persisted state in Redis. This effectively replaces `OtherQualificationTypeForm` and `OtherQualificationForm` and adds some simple next/previous page logic (there are only 2 pages).
- [x] Update the controllers that implement create/update/delete of other qualifications to use the new wizard and defer persistence.
- [x] Move business logic from the forms to the wizards
- [x] Fix existing specs
- [x] Remove any remaining references to the form objects and delete
- [x] Additional specs for `OtherQualificationWizard`

Out of scope: Extract common code so that it can be reused and to tidy up the controllers and wizard itself. I think this would be an important next step if we want to re-use this approach elsewhere but should be handled in a separate PR. 
We should also remove the `international_other_qualification` feature flag.

## Guidance to review

- I've tried to test this manually as thoroughly as I can. There are quite a few permutations that hit this wizard, new/edit cases, there is behaviour specific to edits that update the `qualification_type` and there is logic to copy a previous qualification where the candidate selects one of the _create another_ radio buttons.
- I have not attempted to make this all work with the `international_other_qualification` feature flag OFF (we've agreed that we are not ever going to switch it off).

## Link to Trello card

https://trello.com/c/DPlksrfx/2049-blank-qualifications-generated-if-there-are-validation-errors

## Things to check


tvararu:
image

I logged in as a candidate who had already submitted an application so the section was completed on their end.

The link is:

`https://apply-for-te-2049-blank-xzxuuc.herokuapp.com/candidate/application/other-qualifications/review`

Clicking on it takes me here:

image

The link is `https://apply-for-te-2049-blank-xzxuuc.herokuapp.com/candidate/application/other-qualifications`.

Seems like a bug.

tvararu:
@stevehook shall we close this now that it's probably too far out of date to merge?

tvararu:
Works well in the review app! I did find two bugs.

First one:

- Sign in with a candidate that can apply again
- Apply again
- Section shows up like this:

image

- Clicking results in this page:

image

This seems wrong.

Second bug:

- Pick qualification type
- Fill in form with anything you like:

image

- Submit with "No, not at the moment"
- Arrive on Review, all is good
- Press back
- Submit again
- You get an error:

image

tvararu:
Once the bug is fixed I'll give this another look and we can :shipit:

stevehook:
Thanks @tvararu !

First issue seems to happen on `qa` too. I guess if you have not added any other qualifications and marked it as complete then you get the new qualification flow when you click there because there's nothing to review.

The second issue is because I have to clear state at the end of the create flow so going back to the details page means you've effectively started a clean slate but skipped the first page of the wizard. The first solution that comes to mind is to redirect to the first page if the qualification type values are missing from the current state but this doesn't work because the details controller will (typically) be served from the browser's cache. What the candidate would expect is that the back button gives them an opportunity to amend the qualification that they just created - I can't think of an easy way to do that with the wizard approach (it would need some kind of unique key allocated before save, e.g. a Guid primary key). So still looking for ideas about how to solve this one...

stevehook:
@tvararu I've updated the flow a bit to handle the back button issue. In the end I went with the flash message because this issue happens after updates as well as creates and I felt it was getting too messy trying update the right qualification in either case.

image

malcolmbaig:
Some good stuff in here :+1: 

One initial observation - would it make sense for the form objects to retain responsibility for validating and filling out forms? Thinking about how the solution in this PR might apply to other, longer forms, we could end up with quite large wizard classes if they take on validation for the entire flow.

Would it work if the form objects invoked the wizard in their respective `save` methods? That leaves the wizard with the following responsibilities:
- Persisting a bag of attributes in Redis, maybe namespaced according to the step in the flow.
- Clearing all state as required.
- Committing to the database once the final step is submitted.
- Checking if the first expected value of the create flow is present (allowing the controller to redirect to the review page with a flash message).

tvararu (approved):
```

                     ,
                    /|
                   /*|
                  /.+|
                 /* .|
                 |:+.:\
                 /+.:*.\
                |:.*.:+|
               /+.---.Z\
             ,(((/o^o\))\
            (())) ' > ()))
            )(())(/~\))))(
           /((()'`))((()(\)
          /::)).&& (()))():\
         /:*::)'||.   /\+:::\
        /:::::( || \ / |:/:::)
        \:::+/-'L|, &  |::*:/
         |::(|_  _'   _\+::|
         |*::\ `-'   //,):/\
         |:::+| ||   
```

Seems to work well! Didn't cover the code very well but it seems sensible. Ship.

malcolmbaig:
As discussed in Slack - this is going to be merged as is and followed up with a refactoring PR.

malcolmbaig (approved):

PR #3043, opened 2020-10-01
Remove GenerateVendorProviders

duncanjbrown:
We don't use this any more — vendors can use the "keep courses open" feature flag or they can create fake providers using the below button. Having two "Fake providers" buttons is confusing.

## Context

Various confusing conversations in Slack

Remove the button and the service that it triggers

Screenshot 2020-10-01 at 15 57 12


tvararu (approved):

tijmenb (approved):
Props for the screenshot.

PR #3046, opened 2020-10-02
Prevent submit without two references provided

stevehook:
## Context

Candidates should not be able to submit an application until they have 2 references.

- [x] Update `ApplicationFormPresenter` to include new ready for submission rule. (Behind `decoupled_references` feature flags).
- [x] Update `RefereesReviewComponent` to include new error message. (Behind `decoupled_references` feature flags).
- [x] Update `SectionMissingBannerComponent` to allow a non-standard link text to support above.

image

image

## Guidance to review

- Have I missed anything?
- I've added some unit tests but not tried to alter the system specs here mainly because the other parts of the decoupled reference journey are not yet in place. We could modify one of the existing specs if needed.

## Link to Trello card

https://trello.com/c/tlQBYyRa/2216-💔-dev-prevent-submitting-application-without-2-references

## Things to check

### DIFF
--- a/app/components/candidate_interface/referees_review_component.rb
+++ b/app/components/candidate_interface/referees_review_component.rb
@@ -26,7 +26,13 @@ module CandidateInterface
     end
 
     def show_missing_banner?
-      !@application_form.references_completed && @editable if @submitting_application
+      !@application_form.references_completed && @editable && @submitting_application
+    end
+
+    def show_references_not_provided?
+      return false unless FeatureFlag.active?(:decoupled_references)
+
+      !@application_form.enough_references_have_been_provided? && @editable && @submitting_application
     end
 
   private
--- a/app/components/candidate_interface/referees_review_component.rb
+++ b/app/components/candidate_interface/referees_review_component.rb
@@ -26,6 +26,8 @@ module CandidateInterface
     end
 
     def show_missing_banner?
+      return false if FeatureFlag.active?(:decoupled_references)
+
       !@application_form.references_completed && @editable && @submitting_application
     end
 
--- a/app/components/section_missing_banner_component.html.erb
+++ b/app/components/section_missing_banner_component.html.erb
@@ -1,6 +1,6 @@
 <div class="app-banner<% if @error %> app-banner--warning<% end %> app-banner--missing-section" id="<%= "missing-#{section}-error" %>">
   <div class="app-banner__message">
     <p class="govuk-body"><%= @text %></p>
-    <%= link_to t("review_application.#{section}.complete_section"), @section_path, class: 'govuk-link govuk-link--no-visited-state', 'data-qa': "missing-#{section}" %>
+    <%= link_to @link_text, @section_path, class: 'govuk-link govuk-link--no-visited-state', 'data-qa': "missing-#{section}" %>
   </div>
 </div>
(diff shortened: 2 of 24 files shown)


spec/presenters/candidate_interface/application_form_presenter_spec.rb
```diff
@@ -301,6 +301,45 @@
...
+
+    it 'returns true if the referees section has been created and two references have been provided' do
+      application_form = create(:application_form, references_completed: true)
+      create_list(:reference, 2, application_form: application_form, feedback_status: :feedback_provided)
+      presenter = CandidateInterface::ApplicationFormPresenter.new(application_form)
+
+      expect(presenter).to be_enough_references_provided
+    end
+
+    it 'returns false if the referees section has been created and only one reference has been provided' do
```
malcolmbaig, line 315: s/created/completed/ ?

app/components/section_missing_banner_component.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="app-banner<% if @error %> app-banner--warning<% end %> app-banner--missing-section" id="<%= "missing-#{section}-error" %>">
   <div class="app-banner__message">
     <p class="govuk-body"><%= @text %></p>
-    <%= link_to t("review_application.#{section}.complete_section"), @section_path, class: 'govuk-link govuk-link--no-visited-state', 'data-qa': "missing-#{section}" %>
+    <%= link_to @link_text, @section_path, class: 'govuk-link govuk-link--no-visited-state', 'data-qa': "missing-#{section}" %>
```
malcolmbaig, line 4: can use the `link_text` attr_reader here, rather than the instance var

app/components/candidate_interface/referees_review_component.rb
```diff
@@ -26,7 +26,13 @@ def minimum_references
...
 
     def show_missing_banner?
-      !@application_form.references_completed && @editable if @submitting_application
+      !@application_form.references_completed && @editable && @submitting_application
+    end
+
+    def show_references_not_provided?
+      return false unless FeatureFlag.active?(:decoupled_references)
+
+      !@application_form.enough_references_have_been_provided? && @editable && @submitting_application
```
malcolmbaig, line 35: FYI, I'm writing a new component to replace the existing referees review one. I've made a note on my card to move this over when it's ready.

stevehook:
@malcolmbaig this PR breaks about 14 system specs - ones that submit an application before receiving a reference. I've fixed (most of) them by adding an explicit `FeatureFlag.deactivate(:decoupled_references)`. That means we should probably have a corresponding spec for each one with the FF activated and whatever changes are needed to make it green though I'm not sure how stable they would be while we are changing other related bits. Just wondering what you think? Is it best to revisit the main parts of the new references flow is merged?

malcolmbaig:
@stevehook 

Initially I was thinking we could just have a single spec that tests application submit end-to-end with the feature switched on.

That would leave us with:

- a set of system specs testing various behaviors with decoupled references switched off.
- a single spec that covers submitting an app with decoupled references switched on.

Then shortly after we go live, the first set of specs get reworked when we eventually get round to removing the feature flag. The existence of the submit application spec gives us reasonable confidence that this specific piece of functionality behaves as expected. However, it means we'll have a period of time where several of our system specs test behavior that no longer exists in production (ie - references received after submit, 5-day edit window).

Alternatively, I wonder if we should write a card that needs to be done prior to the feature going live which entails reviewing every system spec where we've disabled the feature, creating a copy in `spec/system/candidate_interface/decoupled_references/specs_for_deletion` (or similar), and reworking the originals for the feature being enabled. That way we don't get bogged down with these specs while decoupled references is in progress, but by the time we go live we've got decent test coverage across both states.

Interested to hear other opinions on this. @tvararu @tijmenb ?

malcolmbaig (approved):

PR #3050, opened 2020-10-02
Fix incorrect role on navigation components

paulrobertlloyd:
## Context

We are using `role=tab` in our navigation components. We shouldn’t be doing this as links in these components do not behave like tabs (i.e. switching a view within a page). We can however [use `aria-current` to indicate which item in a set of items is currently selected](https://tink.uk/using-the-aria-current-attribute/).

## Guidance to review

@adamsilver Value for `aria-current`: `page` or `true`? Can we have two navigation lists with the same `aria-current` value (I believe you can as current is relative to sibling links).

## Things to check


tvararu:
`aria-current: page` is correct 👍

tvararu (approved):

steventux (approved):

PR #3055, opened 2020-10-05
Add caching for edit provider users page

tijmenb:
## Context

The new/edit provider user page is very slow, because we're generating a lot of checkboxes.

First, configure caching in all environments. We use in-memory cache, which is not fantastic (it's per node, so users won't get a consistent caching experience) but better than nothing. I'll investigate using Redis for this later.

Use key-based cache expiration to cache the set of checkboxes for a single provider. The checkbox cache is invalidated when:

- the HTML changes
- the provider changes (the updated_at attribute)
- the provider permission changes (when a checkbox is changed)

## Guidance to review

Does this make sense? Any problems that we can anticipate with this caching strategy?

## Link to Trello card

https://trello.com/c/jpKDACFO/2285-make-sure-add-provider-user-page-in-support-can-deal-with-a-lot-of-providers

## Things to check


duncanjbrown (approved):

tvararu (approved):

tvararu:
Problems are same as discussed yesterday; it's per-instance and it gets invalidated on deploy, but still in practice I think this will be useful!

PR #3062, opened 2020-10-05
Support: Use summary card component for feature flags

paulrobertlloyd:
## Context

Use summary card for feature flags, use same design for reference request history for feature flag audit trail

Before

Screenshot 2020-10-05 at 20 21 47

After

Screenshot 2020-10-05 at 20 20 42

## Guidance to review

@adamsilver What colour the button? Previously was grey, but can’t use grey button on grey background. Can’t really use green, as that might suggest an active feature flag. So added a blue variant… I’m not sure this is right.

Are the updated tests correctly written?

## Things to check

### DIFF
--- a/app/components/support_interface/feature_audit_trail_component.html.erb
+++ b/app/components/support_interface/feature_audit_trail_component.html.erb
@@ -1,7 +1,8 @@
 <ul class="govuk-list govuk-list--bullet">
   <% audits.select { |audit| visible?(audit) }.each do |audit| %>
     <li>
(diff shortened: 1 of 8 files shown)


adamsilver:
@paulrobertlloyd for public recording purposes, thinking we should:

1. Stick to blue button (given we don't have an ‘are you sure’ step)
2. Make it normal size
3. Vertically align the left hand text to the middle

paulrobertlloyd:
@adamsilver Done!

Screenshot 2020-10-06 at 16 24 03

tvararu (approved):

PR #3063, opened 2020-10-06
Prevent candidates from accidentally creating multiple carry-over applications

tijmenb:
## Context

https://www.apply-for-teacher-training.service.gov.uk/support/applications/3701 carried over their application twice, probably by doing some kind of back-button thing (see https://kibana.logit.io/goto/fb8e25702d6f321da484fe0698e6883d).

This disallows carry-over if the application form is not in a state that needs carrying over. Prevents candidates from accidentally making multiple applications by using the back button.

## Link to Trello card

https://trello.com/c/fbxCHROO/2293-prevent-candidates-from-accidentally-creating-multiple-carry-over-applications

## Things to check

### DIFF
--- a/app/models/application_form.rb
+++ b/app/models/application_form.rb
@@ -154,6 +154,14 @@ class ApplicationForm < ApplicationRecord
     apply_2?
   end
 
+  def must_be_carried_over?
+    if ended_without_success?
+      recruitment_cycle_year < RecruitmentCycle.current_year || EndOfCycleTimetable.between_cycles_apply_2?
+    elsif !submitted?
+      recruitment_cycle_year < RecruitmentCycle.current_year && !EndOfCycleTimetable.between_cycles_apply_1?
+    end
+  end
+
   def choices_left_to_make
     number_of_choices_candidate_can_make - application_choices.size
   end
--- a/app/components/candidate_interface/carry_over_banner_component.rb
+++ b/app/components/candidate_interface/carry_over_banner_component.rb
@@ -8,13 +8,7 @@ module CandidateInterface
     end
 
     def render?
-      if @application_form.ended_without_success?
-        @application_form.recruitment_cycle_year < RecruitmentCycle.current_year ||
-          EndOfCycleTimetable.between_cycles_apply_2?
-      elsif !@application_form.submitted?
-        @application_form.recruitment_cycle_year < RecruitmentCycle.current_year &&
-          !EndOfCycleTimetable.between_cycles_apply_1?
-      end
+      @application_form.must_be_carried_over?
     end
 
     def references_did_not_come_back_in_time?
--- a/app/controllers/candidate_interface/carry_over_controller.rb
+++ b/app/controllers/candidate_interface/carry_over_controller.rb
@@ -1,5 +1,7 @@
 module CandidateInterface
   class CarryOverController < CandidateInterfaceController
+    before_action :redirect_if_already_carried_over
+
     def start
       if EndOfCycleTimetable.between_cycles_apply_2?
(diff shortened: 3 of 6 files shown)


stevehook (commented):
Makes sense to me and testing the review app I wasn't able to create more than one new application with a previously unsuccessful application but with an un-submitted one I got into a redirect loop, need to have a proper look after stand-up to understand what's going on there...

Aside: this was meant to prevent double carry-overs but it looks like it doesn't work if the candidate attempts to carry over the original application again rather than the new current application. https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/services/carry_over_application.rb#L19-L21

tijmenb:
Ooh good shout - I think https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/services/carry_over_application.rb#L30 is possibly wrong. It may be generating applications for the next cycle currently.

stevehook:
The reason for the redirect loop is that we automatically redirect to the carry over page based on a rule in `UnsubmittedApplicationController#redirect_to_carry_over_if_unsubmitted_previous_cycle` and that rules is inconsistent with `ApplicationForm#must_by_carried_over?`(the former returns true when the later returns false). So the solution to that is I guess to use just call `ApplicationForm#must_by_carried_over?` from `UnsubmittedApplicationController#redirect_to_carry_over_if_unsubmitted_previous_cycle` so they are reusing the same logic and therefore one redirect cannot happen at the same time as the other.

tijmenb:
@stevehook fixed 🤞

app/models/application_form.rb
```diff
@@ -154,6 +154,14 @@ def candidate_can_choose_single_course?
     apply_2?
   end
 
+  def must_be_carried_over?
+    if ended_without_success?
+      recruitment_cycle_year < RecruitmentCycle.current_year || EndOfCycleTimetable.between_cycles_apply_2?
+    elsif !submitted?
+      recruitment_cycle_year < RecruitmentCycle.current_year && !EndOfCycleTimetable.between_cycles_apply_1?
```
stevehook, line 161: This means that you can't now carry over an un-submitted application until the new cycle opens. I don't think that's right (but I'm not sure), it's certainly a change of behaviour.

stevehook (commented):
Looks good now - no redirect loops! I just have a question inline about what is the correct behaviour for un-submitted applications. Should we permit un-submitted application carry over between cycles?

app/models/application_form.rb
tvararu: To the best of my knowledge that wasn't specified behaviour; it's better to be consistent and use the same checks in multiple places.

tvararu (approved):

app/models/application_form.rb
stevehook: Agreed consistency is most important. The one-off task that already 'failed' any un-submitted applications around 18 Sept should have moved all these into a state that can be carried over anyway (in the same way as other unsuccessful applications). So this piece of business logic will pick up un-submitted applications too.

stevehook (approved):

stevehook:
@tijmenb it looks like there is just one spec that's failing here. I think we just need to drop these two lines because with this PR it's not longer possible to manually carry over an application in the `unsubmitted` state between cycles:

https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/spec/system/candidate_interface/candidate_cannot_add_new_course_once_the_cycle_has_closed_spec.rb#L43-L45

I think we can change this because in production such applications would have been processed by `CancelUnsubmittedApplicationsWorker` and moved to the `application_not_sent` status so that they can be carried over manually.

PR #3064, opened 2020-10-06
Fix UCAS matches summary in support

AgaDufrat:
## Context

We are getting errors for individual UCAS match pages: 
https://sentry.io/organizations/dfe-bat/issues/1936503787/?project=1765973

Screenshot 2020-10-06 at 09 03 28

**- Consider recruitment cycle when looking up course for UCAS matches**
Now that we are approaching the next recruitment cycle we have two courses with the same code and provider. One for each recruitment cycle year. Decided to hardcore it instead of using `RecruitmentCycle.current_year`, `RecruitmentCycle.previous_year` as it’s going to change soon but we still want to compare courses in 2020 recruitment cycle. 

**- Fix UCAS matches summary for apply again candidates**
When a candidate applied again they would have two application forms. `find_by` was returning the latest application form, which in most cases was incomplete and not sent to providers and therefore had no application choices. Candidate’s ids for which this was the case: 96, 471, 3410, 1030, 2485, 2850, 3327, 3616.

## Guidance to review

Review questions:
- thought on hardcoding recruitment cycle year
- is there a fancy join I can do to avoid 2 queries in the 2nd commit?

Don't think we can reproduce the same issue locally. We can get the same error by running `GenerateTestData.new(10).generate` a few times but this is because some test application forms are generated without application choices, _I think._

## Things to check

### DIFF
--- a/app/models/ucas_matched_application.rb
+++ b/app/models/ucas_matched_application.rb
@@ -1,5 +1,4 @@
 class UCASMatchedApplication
-
   def initialize(matching_data, recruitment_cycle_year)
     @matching_data = matching_data
     @recruitment_cycle_year = recruitment_cycle_year
@@ -33,8 +32,9 @@ class UCASMatchedApplication
     if ucas_scheme?
       mapped_ucas_status
     else
-      ApplicationForm.find_by(candidate_id: @matching_data['Apply candidate ID'])
-        .application_choices.find_by(course_option: course.course_options)
+      ApplicationChoice.includes(:application_form)
+        .where('application_forms.candidate_id = ?', @matching_data['Apply candidate ID']).references(:application_forms)
+        .find_by(course_option: course.course_options)
         .status
     end
   end
--- a/app/models/ucas_matched_application.rb
+++ b/app/models/ucas_matched_application.rb
@@ -1,10 +1,16 @@
 class UCASMatchedApplication
-  def initialize(matching_data)
+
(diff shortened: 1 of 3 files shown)


app/models/ucas_matched_application.rb
```diff
@@ -1,10 +1,16 @@
 class UCASMatchedApplication
+  RECRUITMENT_CYCLE_YEAR_FOR_UCAS_MATCHES = 2020
```
duncanjbrown, line 2: What if we looked at `created_at` on the match, and checked which cycle bucket it fell into? I don't think any new matches should be created after the cycle closes.

app/models/ucas_matched_application.rb
```diff
@@ -33,9 +33,10 @@ def status
     if ucas_scheme?
       mapped_ucas_status
     else
-      ApplicationForm.find_by(candidate_id: @matching_data['Apply candidate ID'])
-        .application_choices.find_by(course_option: course.course_options)
-        .status
+      ApplicationChoice.find_by(
```
duncanjbrown, line 36: You could do something like `ApplicationChoice.includes(:application_form).where(course_option: co, application_form: { candidate_id: c })`

tvararu:
https://trello.com/c/RAdk0NeY/2301-fix-ucas-matching-sentry-error

duncanjbrown (approved):
LGTM!

PR #3068, opened 2020-10-06
Add HESA codes to qualifications in Apply API responses

mnacos:
## Context

We already store HESA qualification codes in the database, in fields such as `qualification_type_hesa_code`, `subject_hesa_code` and `grade_hesa_code`. This work adds HESA qualification code fields to Qualification entities in Apply API responses.

See also: https://www.hesa.ac.uk/collection/c19053

Add `hesa_degtype`, `hesa_degsbj`, `hesa_degclss`, `hesa_degest`, `hesa_degctry`, `hesa_degstdt`, `hesa_degenddt` attributes to each qualification. Leave their values `nil` if data is unavailable, or if the qualification is not at degree level. The names of these attributes mirror HESA attributes (e.g. `DEGTYPE`).

Add a `domicile` attribute to the candidate. This value is extracted from the candidate's address (as specified in the application).

Update the relevant API docs and release notes.

## Guidance to review

I've decided to zero-pad the codes (`01` instead of `1` etc.) to variable lengths depending on the attribute, because this is how the HESA documentation lists them. We seem to store them as bigints in the database, hence the 20 chars `maxLength` in the API docs.

Should the new qualifications attributes (e.g. `hesa_degtype`) be present (as keys) within non-degree qualifications or is it better to omit them altogether? I've gone for consistency and added them to all types of qualification, as our API docs contain only one type of Qualification object.

Are the new descriptions in the API docs and release notes clear?

## Link to Trello card

https://trello.com/c/S17ZnBOV

## Things to check


app/presenters/vendor_api/hesa_qualification_fields_presenter.rb
```diff
@@ -8,11 +8,11 @@ def to_hash
       return {} unless @qualification.level == 'degree'
 
       {
-        hesa_degtype: @qualification.qualification_type_hesa_code,
-        hesa_degsbj: @qualification.subject_hesa_code,
-        hesa_degclss: @qualification.grade_hesa_code,
-        hesa_degest: @qualification.institution_hesa_code,
-        hesa_degctry: @qualification.institution_country,
+        hesa_degtype: @qualification.qualification_type_hesa_code&.to_s.rjust(3, '0'),
+        hesa_degsbj: @qualification.subject_hesa_code&.to_s.rjust(6, '0'),
```
duncanjbrown: Do we need to pad this one? It always starts with 1 or 9

app/presenters/vendor_api/hesa_qualification_fields_presenter.rb
```diff
@@ -8,11 +8,11 @@ def to_hash
...
+        hesa_degclss: @qualification.grade_hesa_code&.to_s.rjust(2, '0'),
+        hesa_degest: @qualification.institution_hesa_code&.to_s.rjust(4, '0'),
+        hesa_degctry: @qualification.institution_country&.to_s.rjust(2, '0'),
```
duncanjbrown: These are 2-char ISO codes so shouldn't need padding

config/vendor-api-v1.yml
```diff
@@ -753,6 +766,48 @@ components:
...
+          type: string
+          description: HESA code for degree subject
+          example: '100048'
+          maxLength: 20
+          nullable: true
+        hesa_degclss:
+          type: string
+          description: HESA code for degree class
+          example: '01'
+          maxLength: 20
```
duncanjbrown, line 785: Shouldn't these fields all have maxlength of the legitimate data? ie in this case it would be 2

duncanjbrown (commented):
Couple of qs inline! Shouldn't the `candidate_completes_application_form` routine in `vendor_receives_application_spec` cause HESA codes to be stored?

config/vendor-api-v1.yml
mnacos: Depends... I've treated `maxLength` as an attribute of the Apply API and different to `Length` that is part of the HESA spec. `maxLength`, from this perspective, is how large a field vendors should be allocating in their database to store this information, not necessarily the length of values received. As we don't cap the length of values returned in the presenter, the current arrangement seems the safest option. Furthermore, if HESA codes were to change in terms of length, we could easily cope with these changes in the code while keeping Apply's `maxLength` spec constant.

What do you think?

app/presenters/vendor_api/hesa_qualification_fields_presenter.rb
mnacos: The refactored version of this code doesn't pad `institution_country`.

app/presenters/vendor_api/hesa_qualification_fields_presenter.rb
mnacos: Probably not, but legit values are all 6-chars long. Worst case scenario would be vendors receiving an unrecognisable code (e.g. `000000`) if we have `0` values. If our value is `nil` they'll get `nil`, too.

config/vendor-api-v1.yml
duncanjbrown: Sure — agree it makes sense to leave the fields a bit baggy in case the coding changes.

app/presenters/vendor_api/hesa_qualification_fields_presenter.rb
duncanjbrown: If we've got bad data no amount of padding is going to help! I think we should send what we've got. (Then fix the data in the DB).

mnacos: Ok, updated

duncanjbrown:
Thanks! Last q: what do you think about this?

> Shouldn't the candidate_completes_application_form routine in vendor_receives_application_spec cause HESA codes to be stored?

mnacos:
> Shouldn't the candidate_completes_application_form routine in vendor_receives_application_spec cause HESA codes to be stored?

Good news, it does: https://github.com/DFE-Digital/apply-for-teacher-training/blob/2015-hesa-qualification-fields-in-apply-api/spec/system/vendor_api/vendor_receives_application_spec.rb#L161

The other fields are `nil` because the degree supplied by https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/spec/support/test_helpers/candidate_helper.rb#L157 uses values such as 'University of Much Now' and 'Doge', which do not resolve.

An interesting point to be made is our HESA code matching seems to be exact and case sensitive. In theory, the supplied degree type ('BA') should resolve, but `type_decription` used in https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/forms/candidate_interface/degree_type_form.rb#L48 does not exist in the database. The value present is 'Bachelor of Arts'. 'BA' exists in the abbreviation column.

An improvement here would be to check the abbreviation column, too.

duncanjbrown (approved):
Looks good to me! Wondered about `degctry` but it can be omitted for uk degrees:

```
This field must be completed with the country where an entrant's previous degree was awarded. Where the qualification was awarded in the UK but the provider is not listed in PreviousQualification.DEGEST then England, Northern Ireland, Scotland or Wales should be coded in this field.
```

https://www.hesa.ac.uk/collection/c19053/e/degctry

Screenshot 2020-10-08 at 17 09 12

mnacos:
> Looks good to me! Wondered about `degctry` but it can be omitted for uk degrees:

Probably another case of our matching being too exact. There is no 'United Kingdom' in the HESA vocab, only 'United Kingdom, not otherwise specified'.

PR #3070, opened 2020-10-06
Manage: Use full-width header border

paulrobertlloyd:
## Context

Use full-width border under header to match width of primary navigation (matching design in prototype)

Before:
Screenshot 2020-10-06 at 16 11 39

After:

Screenshot 2020-10-06 at 16 12 45

## Things to check


adamsilver (approved):
You.
Are.
Awesome.
Thanks.
Raised.
Hands.

PR #3079, opened 2020-10-07
Update application submit page for decoupled references MVP

stevehook:
## Context

This PR updates the content of the submit page to reflect the decoupled references prototype, with content that talks about sending rather than submitting.

- [x] Copy the old template `submit_show.html.erb` to `submit_show_without_decoupled_references.html.erb` and update the content for `submit_show.html.erb`.
- [x] Copy the old localised strings referenced by the template above to a new `without_decoupled_references` key and use those for `submit_show_without_decoupled_references.html.erb`. Update the original keys with new content.
- [x] Put `submit_show.html.erb` behind the existing `decoupled_references` feature flag.
- [x] Fix specs

image

## Guidance to review

- I've tried to make it as simple as possible to remove the feature flag by duplicating the view template etc. - is there a simpler way?

## Link to Trello card

https://trello.com/c/waoPgAbN/2243-💔-dev-update-application-submit-page-mvp

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/unsubmitted_application_form_controller.rb
+++ b/app/controllers/candidate_interface/unsubmitted_application_form_controller.rb
@@ -26,7 +26,7 @@ module CandidateInterface
         @errors = @application_form_presenter.section_errors
         @application_choice_errors = @application_form_presenter.application_choice_errors
 
-        render 'candidate_interface/unsubmitted_application_form/review'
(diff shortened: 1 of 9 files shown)


paulrobertlloyd (approved):
I appreciate the effort gone in to having different content based on the decoupled feature flag (arguably, this content could change for all users, regardless of that flag, but good to be consistent with how we roll this out, I guess).

tvararu (approved):

PR #3080, opened 2020-10-07
Cascade candidate deletes down to UCAS matches

duncanjbrown:
Otherwise this causes foreign key errors when sandbox users try to clear test data: that routine works by deleting all the candidates and letting the deletes cascade.

## Context

https://ukgovernmentdfe.slack.com/archives/CN1MCQCHZ/p1602057528142400

Add a cascading delete in the DB as we do with other candidate associations

## Guidance to review

Not going to delete anything we don't want to delete? In practice this only ever happens in sandbox, or should.

## Link to Trello card

N/A, sentry error from yesterday's error fiesta

## Things to check


tvararu (approved):
:ship:

PR #3083, opened 2020-10-07
 Decoupled references review page (2/2)

malcolmbaig:
## Context
Builds on code introduced by https://github.com/DFE-Digital/apply-for-teacher-training/pull/3081

As a... candidate
I need to... see the status of my referees and reference requests
So that... I now what I need to do next to complete this section
## Changes proposed in this pull request
- Simplify the component implementation
- Move the app submit error messaging over from another location where it was previously implemented by @stevehook 
- Add component and system tests

## Guidance to review
- Path: `/candidate/application/references/review`
- Also worth checking the app submit error (if not enough references) and the post-submit application review.
- Delete, Cancel, and Change behaviour not implemented yet, will follow in another PR.

Probably best reviewed locally, might need to create some references via the existing references flow and modify state on the console until the decoupled references flow is ready.

## Link to Trello card

https://trello.com/c/7autFjC7
## Things to check

- [x] This code does not rely on migrations in the same Pull Request
- [x] If this code includes a migration adding or changing columns, it also backfills existing records for consistency
- [x] API release notes have been updated if necessary
- [x] New environment variables have been [added to the Azure config](https://github.com/DFE-Digital/apply-for-teacher-training#azure-hosting-devops-pipeline)

PR #3084, opened 2020-10-07
Change SubmitReference behaviour for decoupled references

malcolmbaig:
## Context

With `decoupled_references` active, when a referee submits a reference
we only want to update the reference state. None of the
`progress_applications` logic is relevant or desirable, as the
application will be sent to providers on submit.

Arose out of the work on https://github.com/DFE-Digital/apply-for-teacher-training/pull/3083 . It's not obviously related, but this change was prompted by the requirement that we no longer allow the candidate to mark the section complete manually. Instead, the section is considered complete as soon as two references are provided.

See diff.

## Link to Trello card

https://trello.com/c/7autFjC7
## Things to check

### DIFF
--- a/spec/services/submit_reference_spec.rb
+++ b/spec/services/submit_reference_spec.rb
@@ -2,79 +2,101 @@ require 'rails_helper'
 
 RSpec.describe SubmitReference do
   describe '#save!' do
-    let(:application_form) { create(:completed_application_form) }
+    describe 'decoupled_references feature ON' do
+      before { FeatureFlag.activate(:decoupled_references) }
 
-    context 'minimum required references received' do
-      it 'progresses the application choices to the "application complete" status' do
-        application_form = create(:completed_application_form, edit_by: 1.day.from_now)
-        active_application_choice = create(:application_choice, application_form: application_form, status: 'awaiting_references')
-        cancelled_application_choice = create(:application_choice, application_form: application_form, status: 'cancelled')
+      it 'updates the reference state to "feedback_provided"' do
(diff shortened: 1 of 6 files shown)


spec/services/submit_reference_spec.rb
```diff
@@ -2,79 +2,101 @@
 
 RSpec.describe SubmitReference do
   describe '#save!' do
-    let(:application_form) { create(:completed_application_form) }
+    describe 'decoupled_references feature ON' do
```
malcolmbaig, line 5: This diff looks noisy, but all that's happened here is I've introduced two separate `describe` blocks for `decoupled_references feature ON` and `decoupled_references feature OFF`. A new example sits in the former, all the old examples are nested under the latter.

tvararu: Looks fine with "ignore whitespace"!

tvararu (approved):

davidgisbey (approved):
lgtm :shipit:

spec/services/submit_reference_spec.rb
davidgisbey: It's not overly important, but out of interest I would have put the flag being on and off in context blocks. What's the thought process of using describe? I'm mainly interested as I think I've been doing it wrong.

malcolmbaig: @davidgisbey completely stylistic, these could readily have been context blocks. I guess here I was thinking that the method behaviour is almost completely different with the flag active, rather than the minor variation that a context probably suggests.

PR #3088, opened 2020-10-07
Manage org/org relationships in support

duncanjbrown:
An page which allows Support users to manage org/org relationshis

## Context

The support team are going to need this as masses of SDs come on along with their HEIs

Introduce a page in support which allows these relationships to be managed all at the same time. 

Known defects

- the error summary does not link to the checkboxes

This is fiddly because the permissions are models nested in a form object. There is surely a way to fix this but I don't think it's worth holding up this feature for.

Screenshot 2020-10-07 at 23 54 55

## Link to Trello card

https://trello.com/c/Ov5yNRrS/2870-allow-changing-org-permissions-via-the-support-console

## Things to check

### DIFF
--- a/app/controllers/support_interface/providers_controller.rb
+++ b/app/controllers/support_interface/providers_controller.rb
@@ -51,6 +51,39 @@ module SupportInterface
       @relationship_diagram = SupportInterface::ProviderRelationshipsDiagram.new(provider: @provider)
     end
 
+    def relationships
+      @provider = Provider.find(params[:provider_id])
+      relationships = ProviderRelationshipPermissions.where(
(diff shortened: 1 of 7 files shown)


steventux (commented):
The approach looks good to me, the only thing that was hard to understand/scan was the UI column headings above the checkboxes as it looks like a duplication (until you read the words carefully).
Perhaps this could be simplified or reworded in future.
A form that support users can assist in setting up permissions is undoubtedly useful, having a system spec for the happy path and circuit breaker for providers without relationship permissions seems like all this needs.

steventux (approved):
:+1:

PR #3091, opened 2020-10-08
Change 'Your account is not ready yet' page to an email address focused one

mnacos:
## Context

Sometimes newly invited provider users get stuck because they have previously signed up to DSI (e.g. for Publish) with a different email address and they end up using this other email address when signing into Apply. Our system does not recognise this other email address and they are shown the 'Your account is not ready' page. Unfortunately, this page is not very helpful. With more information, provider users may be able to resolve the issue themselves, instead of contacting support.

Change the message of the page to hint the problem is the email address. The 'Your account is not ready yet' message dates back to a time when inviting users was a multi-step process. It is now fully automated/single-step. 

Before:

image

After:

image

## Guidance to review

You can log in here: https://apply-for-te-improve-yo-p1s3dt.herokuapp.com/provider/sign-in

If you use random `uid` and `email` values you'll be able to see these changes.

Is the copy ok?

## Link to Trello card

No Trello card

## Things to check

### DIFF
--- a/app/views/provider_interface/account_creation_in_progress.html.erb
+++ b/app/views/provider_interface/account_creation_in_progress.html.erb
@@ -6,13 +6,24 @@
       Your account is not ready yet
     </h1>
     <p class="govuk-body">
-      You’re seeing this page because you’ve signed in
-      to the <%= t('service_name.apply') %> service but your account
-      is not ready to use yet.
+      You have signed in to DfE as
+      <span style="font-weight: bold; white-space: nowrap;">
+        <%= @dfe_sign_in_user.email_address %>
+      </span>
+      but we cannot find your
+      <span style="white-space: nowrap;">
+        <%= govuk_link_to t('service_name.apply'), provider_interface_applications_path %>
+      </span>
+      account.
     </p>
 
     <p class="govuk-body">
-      If you’ve registered to work with us and you think your account should be ready, contact
+      Is this the email address that has been invited to this service?
+    </p>
+
+    <p class="govuk-body">
+      If you’ve registered to work with us with this email address and you think your account should be ready, please contact:<br><br>
+
       <%= bat_contact_mail_to %>.
     </p>
   </div>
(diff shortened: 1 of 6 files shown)


app/views/provider_interface/account_creation_in_progress.html.erb
```diff
@@ -6,13 +6,24 @@
       Your account is not ready yet
```
TobyRet, line 6: Potentially this line could end up being confusing given the proposed changes

app/views/provider_interface/account_creation_in_progress.html.erb
```diff
@@ -6,13 +6,24 @@
...
     </h1>
     <p class="govuk-body">
-      You’re seeing this page because you’ve signed in
-      to the <%= t('service_name.apply') %> service but your account
-      is not ready to use yet.
+      You have signed in to DfE as
```
TobyRet, line 9: Would adding this additional info still follow security good practice?
https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Authentication_Cheat_Sheet.md#authentication-and-error-messages

app/views/provider_interface/account_creation_in_progress.html.erb
mnacos: I agree. Happy to change this but let's wait for the final version of the copy first.

app/views/provider_interface/account_creation_in_progress.html.erb
mnacos: It's good to check these things, but I think the inclusion of an email address on this page is ok, as they have already authenticated by the time they see this. It's more of an authorisation error (the authenticated user is not associated with any accounts). To perform an enumeration attack, they'd basically need log in as every user they are trying to assess.

mnacos:
Paul Genders has suggested the following copy:

**Your email address is not recognised**

You have signed into DfE as different@example. We can’t find an Apply for teacher training account associated with that address. 
Please sign in again, using the email address associated with your Apply account. 
If you are using the correct email address to sign in to Apply and you’re still receiving this message, please contact us at becoming@etc.

adamsilver:
I think avoid bold and put that sentence into its own paragraph.

adamsilver (approved):

PR #3095, opened 2020-10-08
Support: Use breadcrumbs to navigate deep hierarchies

paulrobertlloyd:
## Context

Adds breadcrumb navigation to several sections of the support UI.

## Guidance to review

I have a few questions about these changes, will add them inline.

## Things to check

### DIFF
--- a/app/views/support_interface/application_forms/_application_navigation.html.erb
+++ b/app/views/support_interface/application_forms/_application_navigation.html.erb
@@ -1,4 +1,27 @@
+<% content_for :before_content do %>
+  <%= render BreadcrumbComponent.new(items: [
+    {
+      text: 'Candidates',
+      path: support_interface_candidates_path
+    },
+    {
+      text: @application_form.candidate.email_address,
+      path: support_interface_candidate_path(@application_form.candidate)
+    },
+    {
+      text: @application_form.support_reference,
+      path: support_interface_application_form_path(@application_form)
+    }
+  ].concat(title != 'Details' ? [{ text: title }] : [])) %>
+<% end %>
+
+<p class="govuk-body govuk-!-margin-bottom-6">
+  <%= govuk_link_to 'Emails about this application', support_interface_email_log_path(application_form_id: @application_form.id), class: 'govuk-link--no-visited-state' %>
+</p>
+
 <%= render TabNavigationComponent.new(items: [
   { name: 'Details', url: support_interface_application_form_path(@application_form) },
   { name: 'History', url: support_interface_application_form_audit_path(@application_form) },
 ]) %>
+
+<h2 class="govuk-visually-hidden"><%= title %></h2>
--- a/app/views/support_interface/provider_users/_provider_user_navigation.html.erb
+++ b/app/views/support_interface/provider_users/_provider_user_navigation.html.erb
@@ -1,6 +1,20 @@
-<% content_for :title do %>
-  <span class="govuk-caption-xl">Provider user</span>
-  <%= @form.provider_user.full_name %>
+<% content_for :title, provider_user_name(@form.provider_user) %>
+
+<% content_for :before_content do %>
+  <%= render BreadcrumbComponent.new(items: [
+    {
+      text: 'Providers',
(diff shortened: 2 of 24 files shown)


app/models/provider_user.rb
```diff
@@ -37,7 +37,11 @@ def self.onboard!(dsi_user)
   end
 
   def full_name
-    "#{first_name} #{last_name}" if first_name.present? && last_name.present?
+    if first_name.present? && last_name.present?
+      "#{first_name} #{last_name}" if first_name.present? && last_name.present?
+    else
+      email_address
+    end
```
paulrobertlloyd, line 44: The edit provider page sometimes shows no title because a provider user doesn’t have a name. Thinking in this case `full_name` could fall back to their email address. Firstly, is this a good idea, and if so, is this the right place in the code base to achieve this?

app/views/support_interface/application_forms/_application_navigation.html.erb
```diff
@@ -1,4 +1,46 @@
...
+      {
+        text: @application_form.support_reference,
+        path: support_interface_candidate_path(@application_form.candidate)
+      },
+      {
+        text: title
+      }
+    ]) %>
+  <% end %>
+<% end %>
```
paulrobertlloyd, line 35: I’ve done this in a few places in this PR. When a page is the first ‘Details’ tab, we don’t to show that in the breadcrumb (only for subsequent tabs). Is there a smarter way of writing this - either in ERB template or perhaps somewhere in a controller/helper?

app/models/provider_user.rb
tvararu: This sounds more like a view concern, I wouldn't put it in the provider model, but it's not the worst place either! Perhaps you could create a `view_helper` for it, or make it part of a view component?

app/views/support_interface/application_forms/_application_navigation.html.erb
tvararu, line 35: I think the approach makes sense but you could DRY it:

```suggestion
  <%= render BreadcrumbComponent.new(items: [
    {
      text: 'Candidates',
      path: support_interface_candidates_path
    },
    {
      text: @application_form.candidate.email_address,
      path: support_interface_candidate_path(@application_form.candidate)
    },
    {
      text: @application_form.support_reference,
      path: support_interface_candidate_path(@application_form.candidate)
    }
  ].concat(title == 'Details' ? [{ text: title }] : [])) %>
<% end %>
```

app/views/support_interface/provider_users/_provider_user_navigation.html.erb
```diff
@@ -1,6 +1,39 @@
-<% content_for :title do %>
-  <span class="govuk-caption-xl">Provider user</span>
-  <%= @form.provider_user.full_name %>
+<% content_for :title, @form.provider_user.full_name %>
+
+<% content_for :before_content do %>
+  <% if title == 'Details' %>
```
tvararu, line 4: See previous suggestion!

app/views/support_interface/providers/_provider_navigation.html.erb
```diff
@@ -1,11 +1,37 @@
 <% content_for :title, "#{@provider.name_and_code}" %>
 
-<% content_for :before_content, govuk_back_link_to(support_interface_providers_path, 'Back to providers') %>
+<% content_for :before_content do %>
+  <% if title == 'Details' %>
```
tvararu, line 4: See above!

tvararu (approved):

app/models/provider_user.rb
paulrobertlloyd: Nice! Have made it a `view_helper`: `provider_user_name`.

app/views/support_interface/application_forms/_application_navigation.html.erb
paulrobertlloyd: A-ha! Awesome, have made these all dryer. Knew they’d be a way. Thanks @tvararu 😃

PR #3096, opened 2020-10-08
Add new submit application behaviour for decoupled references

malcolmbaig:
## Context
With decoupled references active, applications can't be submitted until references are received. As such, once the application is submitted it should go immediately to the provider with no edit window.

- Add a `SubmitApplicationWithDecoupledReferences` (temporary name) to handle the different submit behaviour.
- Invoke this service from `UnsubmittedApplicationFormController` when the `decoupled_references` flag is active.

## Guidance to review

- Manual testing - submitting an application and then checking the appropriate candidate/provider emails have been sent ought to do it.
- Does the new service object cover everything it needs to? Has anything been missed from the existing `SubmitApplication` that should also be copied over?
- This card https://trello.com/c/oUQxxBUp will remove the bit about references in `CandidateMailer.application_submitted`

## Link to Trello card
https://trello.com/c/0hOHcSaf

## Things to check

### DIFF
--- a/app/services/submit_application_with_decoupled_references.rb
+++ b/app/services/submit_application_with_decoupled_references.rb
@@ -1,0 +1,28 @@
+class SubmitApplicationWithDecoupledReferences
+  attr_reader :application_form, :application_choices
+
+  def initialize(application_form)
+    @application_form = application_form
+    @application_choices = application_form.application_choices
+  end
+
+  def call
+    # TODO: the decoupled_references feature makes the edit window redundant.
+    # Drop the ApplicationForm#edit_by column when removing the
+    # decoupled_references feature flag.
+    application_form.update!(
+      submitted_at: Time.zone.now,
+      edit_by: Time.zone.now,
+    )
+
+    application_choices.each do |application_choice|
+      SendApplicationToProvider.call(application_choice)
+    end
+
+    if application_form.apply_2?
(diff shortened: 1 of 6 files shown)


app/services/submit_application_with_decoupled_references.rb
```diff
@@ -0,0 +1,28 @@
...
+    @application_choices = application_form.application_choices
+  end
+
+  def call
+    # TODO: the decoupled_references feature makes the edit window redundant.
+    # Drop the ApplicationForm#edit_by column when removing the
+    # decoupled_references feature flag.
+    application_form.update!(
+      submitted_at: Time.zone.now,
+      edit_by: Time.zone.now,
```
davidgisbey, line 15: This makes sense for now, but I'm wondering if it should be set to nil when the flag is removed? Not that it's a big deal either way. We could potentially just drop the column though since it won't be used for anything going forward.

stevehook (approved):
Code looks good and tested on review app - emails to provider and candidate were sent and the application choice moved to the `awaiting_provider_decision` status so all good. Can't see anything missing in the new service.

app/services/submit_application_with_decoupled_references.rb
malcolmbaig: Worth discussing when removing the flag :+1:

PR #3098, opened 2020-10-09
Refactor - HESA Disability codes

TobyRet:
## Context

Some of the HESA disability code logic is currently replicated in more than one place - in the back fill and in the form objects. 
It's possible that HESA disability codes may change year on year and there is a danger that bugs could be silently introduced as a result of having the duplication.

Centralise more of the HESA disability code logic so that it is in one place and reduces the chance of future bugs occurring between the backfill and form completion. Specifically, duplicated logic (conversion and edge casing) is moved into each of the `Hesa` modules.

## Guidance to review

- Changes make sense and are worth it?
- Is there enough test coverage - should we be testing that all form inputs match the correct HESA code?

## Link to Trello card

https://trello.com/c/mxKdgR5L/2178-dev-collect-hesa-codes-at-source-in-ed-qs

## Things to check

### DIFF
--- a/app/forms/candidate_interface/equality_and_diversity/disabilities_form.rb
+++ b/app/forms/candidate_interface/equality_and_diversity/disabilities_form.rb
@@ -59,8 +59,7 @@ module CandidateInterface
 
     def hesa_disability_codes
       disabilities.map { |disability|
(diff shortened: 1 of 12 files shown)


tvararu (approved):
:100:

PR #3103, opened 2020-10-09
Support: Refactor course pages

paulrobertlloyd:
## Context

The current course page lists all applications and all vacancies, which makes it difficult to use (and I suspect can take a while to load).

This PR splits this page up into its component parts, and moves opening a course on Apply into it’s own page, accessible from a change link.

### Before

course-before

### After

course-details

course-applications

course-vacancies

course-open-on-apply

## Things to check


app/components/support_interface/course_apply_status_component.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= render TagComponent.new(
```
TobyRet, line 1: I'm not sure what the team policy is on component specs but I would imagine they would be good to have in instances where they are doing more than simply displaying html

app/views/support_interface/course/open_on_apply.html.erb
```diff
@@ -0,0 +1,17 @@
+<% content_for :browser_title, 'Can candidates use Apply?' %>
```
TobyRet, line 1: Breadcrumb required also on this page?

TobyRet (approved):
LGTM. Couple of minor things. I'll leave it up to you whether you decide to address in this PR or another one. 👍

app/components/support_interface/course_apply_status_component.html.erb
paulrobertlloyd: Would the alternative to be to have this in the view’s controller?

app/views/support_interface/course/open_on_apply.html.erb
paulrobertlloyd: Following a policy of using back links on pages like this (forms, flows, confirmation steps), and breadcrumbs on other pages where you’re browsing page within a deep hierarchy.

app/components/support_interface/course_apply_status_component.html.erb
paulrobertlloyd: @tvararu @duncanjbrown Thoughts? Is this component worthy, or should the code lie somewhere else?

duncanjbrown: Matter of taste! I think I'd turn it into a single `if` with duplication in preference to two ternaries which quietly depend on each other?

tijmenb (commented):
Could you check this with Josh? I strongly suspect that the Apply on/off setting is actually the only thing this page is used for, so moving this functionality would result in an extra click in 99% of journeys.

tijmenb:
I've checked the data, and this page was used 328 times in he last 48 hours. In 88% of the time (290) the form was used, and in 12% of the time the page was only viewed, without an update.

paulrobertlloyd:
@tijmenb Having spoken to Josh, have reverted details page to something closer to original layout, with the status setting at the bottom of the page. Not pushed below any vacancies or applications any more, should also be a little bit easier to update a course.

paulrobertlloyd:
Update design for details tab:

Screenshot 2020-10-13 at 14 42 49

PR #3105, opened 2020-10-11
Add query param to edit links on DecoupledReferencesReviewComponent

malcolmbaig:
## Context
When clicking the 'Change' link on any attribute displayed in the review
component, the candidate is sent to the corresponding edit page. As
currently implemented, saving any changes here redirects the candidate
to the summary screen at the end of the flow where they are prompted to
send the email request. This happens because this summary screen also
provides edit links for each attribute, and the default behaviour is to
return to this screen after any edit.

What should happen is that the candidate returns to the review page if
that's where they initiated the edit. Achieve this by introducing a
`return_to` query param on each edit link in the review component, and
make all controllers in the decoupled references flow aware of this
param. If present, the `update` action in each controller redirects to
the value of the param, otherwise we redirect to the summary screen by
default.

## Guidance to review
- Edit the attributes of a reference from the review page. You should be redirected back to the review page.
- Edit the attributes of a reference from the summary/send a request page you see when adding a reference. You should be redirected back to the summary after each edit.

## Link to Trello card
https://trello.com/c/UnHaN661

## Things to check

### DIFF
--- a/app/components/candidate_interface/decoupled_references_review_component.rb
+++ b/app/components/candidate_interface/decoupled_references_review_component.rb
@@ -32,7 +32,9 @@ module CandidateInterface
         key: 'Name',
         value: reference.name,
         action: "name for #{reference.name}",
-        change_path: candidate_interface_decoupled_references_edit_name_path(reference.id),
+        change_path: candidate_interface_decoupled_references_edit_name_path(
+          reference.id, return_to: :review
+        ),
       }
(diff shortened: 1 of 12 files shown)


malcolmbaig:
Reworking this for better security (Brakeman warning for the unprotected redirect).

tvararu (approved):

PR #3106, opened 2020-10-12
Fix incorrect recruitment cycle year for carried over applications

tijmenb:
## Context

The current code assumes that if we're between cycles, the carried over application will be for "next" year. However, the `current_cycle` changes halfway when we're between cycles when Find reopens. So this results in a recruitment year of 2022 currently.

Fix it by incrementing instead of looking up the year.

## Guidance to review

Does this make sense?

## Link to Trello card

https://trello.com/c/fbxCHROO

## Things to check

### DIFF
--- a/app/services/carry_over_application.rb
+++ b/app/services/carry_over_application.rb
@@ -26,10 +26,6 @@ private
   end
 
   def recruitment_cycle_year
-    if EndOfCycleTimetable.between_cycles_apply_2? && EndOfCycleTimetable.current_cycle?(@application_form)
-      RecruitmentCycle.next_year
-    else
-      RecruitmentCycle.current_year
-    end
+    @application_form.recruitment_cycle_year + 1
   end
 end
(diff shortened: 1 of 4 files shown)


app/services/carry_over_application.rb
```diff
@@ -26,10 +26,6 @@ def application_from_current_cycle?
   end
 
   def recruitment_cycle_year
-    if EndOfCycleTimetable.between_cycles_apply_2? && EndOfCycleTimetable.current_cycle?(@application_form)
-      RecruitmentCycle.next_year
-    else
-      RecruitmentCycle.current_year
-    end
+    application_form.recruitment_cycle_year + 1
```
stevehook, line 29: This makes the `raise_if_application_from_current_cycle` and `application_from_current_cycle?` methods above a bit redundant, they are basically testing `n + 1 == n` now. Not sure whether we need the check now that you've covered the issue of double carry overs in https://github.com/DFE-Digital/apply-for-teacher-training/pull/3063/files. So maybe just remove those methods?

stevehook (approved):
Yes I think this makes sense!

PR #3108, opened 2020-10-12
Display UCAS match course details when course is not on Apply

AgaDufrat:
## Context
Fixes https://sentry.io/organizations/dfe-bat/issues/1937345314/?project=1765973

Screenshot 2020-10-12 at 10 53 40

This was happening for courses, which are not in Apply's db

If a course is not on Apply, take details directly from UCAS matching data.

## Link to Trello card

https://trello.com/c/t6GQeL0D/2837-fix-ucas-matches-summary-in-support

## Things to check

### DIFF
--- a/spec/models/ucas_matched_application_spec.rb
+++ b/spec/models/ucas_matched_application_spec.rb
@@ -25,6 +25,20 @@ RSpec.describe UCASMatchedApplication do
 
       expect(ucas_matching_application.course).to eq(course)
     end
+
+    it 'returns the course details for a course which is not on Apply' do
+      ucas_matching_data =
+        { 'Scheme' => 'U',
+          'Course code' => '123',
+          'Course name' => 'Not on Apply',
+          'Provider code' => course.provider.code.to_s,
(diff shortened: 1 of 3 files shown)


spec/models/ucas_matched_application_spec.rb
```diff
@@ -25,6 +25,24 @@
...
+          'Apply candidate ID' => candidate.id.to_s }
+      ucas_matching_application = UCASMatchedApplication.new(ucas_matching_data, recruitment_cycle_year)
+
+      course_struct = OpenStruct.new(
+        code: '123',
+        name: 'Not on Apply',
+        provider: Provider.find_by(code: course.provider.code.to_s),
+      )
+
+      expect(ucas_matching_application.course).to eq(course_struct)
```
duncanjbrown, line 44: Rather than checking equality for the whole struct, perhaps assert that the `course` exposes the right interface, eg `course.name == 'Not on apply'`?

duncanjbrown (approved):
LGTM! Optional nit inline

PR #3110, opened 2020-10-12
Complete the request a referee flow

davidgisbey:
## Context

Previous PRs are:
#3035 - Up to adding a referees name
#3040 - Add referees email
#3044 - Adds the relationship page 
#3094 - Adds the unsubmitted review page

This is the final PR for the new decoupled references flow that links into the Candidate Name page steve added.

1. It links the flow in with Steve's work
2. It moves validations from the Review controller to the SubmitRefereeForm and adds tests for them
3. Renames the RefereeSubmitForm to SubmitRefereeForm

## Link to Trello card

https://trello.com/c/npGvdYqK/2213-%F0%9F%92%94-dev-request-a-referee-flow-mvp

## Things to check


app/controllers/candidate_interface/decoupled_references/review_controller.rb
```diff
@@ -11,29 +11,15 @@ def show
...
-            # TODO: Link up with Steve's work
-          else
-            # TODO: Once the forms are moved into objects this won't be necessary
-            # It can just call valid? and render the errors.
-            flash[:warning] = 'You must complete all of your referees details before your reference request can be submitted'
-
-            render :unsubmitted
-          end
+        if @submit_reference_form.send_request?
+          redirect_to candidate_interface_decoupled_references_create_candidate_name_path(@reference.id)
```
stevehook, line 22: Does this work as the `create` action is a post?

davidgisbey: Good point. As discussed i've added some logic to the new action. The duplication is taken care of by Malcolm's PR 
image

malcolmbaig: I reckon we should do those checks on CandidateNameForm as part of ReviewController#submit (can be refactored into a private method). 

Something like:
```
if @reference_candidate_name_form.valid?
  CandidateInterface::DecoupledReferences::RequestReference.call(@reference, flash)
else
  redirect_to candidate_interface_decoupled_references_new_candidate_name_path
end
```

That way we aren't triggering side effects via a GET request.

malcolmbaig (changes requested):
LGTM, just one comment about the interaction between the Review and CandidateName controller. Using the `DecoupledReferences::RequestReference` service object that's now in master will also ensure we don't send repeat emails via a stale form.

app/controllers/candidate_interface/decoupled_references/review_controller.rb
davidgisbey: Have done 👍

malcolmbaig (approved):
A system spec needs updating, and there are a couple of typos in commits 560bafd0a5d989ac18bf13a7170edec7d521e6f6 and b9f8e313fb6131a6cd5989a70dc67689b8d306b5, but otherwise this is good to go.

stevehook (commented):
Testing manually I still get prompted for my name if I select 'No, not at the moment'

tvararu (approved):
Haven't tested but code LGTM!

malcolmbaig (changes requested):
Good spot Steve! https://github.com/DFE-Digital/apply-for-teacher-training/pull/3110#pullrequestreview-506602141

davidgisbey:
Yea just fixed it.

app/controllers/candidate_interface/decoupled_references/review_controller.rb
```diff
@@ -11,31 +11,21 @@ def show
...
-        else
+        @candidate_name_form = Reference::CandidateNameForm.build_from_reference(@reference)
+
+        if @submit_reference_form.send_request? && @candidate_name_form.valid?
+          CandidateInterface::DecoupledReferences::RequestReference.call(@reference, flash)
+
           redirect_to candidate_interface_decoupled_references_review_path
+        else
+          redirect_to candidate_interface_decoupled_references_new_candidate_name_path(@reference.id)
         end
```
stevehook:
```
       if !@candidate_name_form.valid?
           redirect_to candidate_interface_decoupled_references_review_path
         else
           CandidateInterface::DecoupledReferences::RequestReference.call(@reference, flash) if @submit_reference_form.send_request?

           redirect_to candidate_interface_decoupled_references_new_candidate_name_path(@reference.id)
         end
```

app/controllers/candidate_interface/decoupled_references/review_controller.rb
```diff
@@ -11,31 +11,21 @@ def show
...
       end
```
stevehook: Might be a neater way to express it but I think this is the logic needed:
```
if @submit_reference_form.send_request? && !@candidate_name_form.valid?
  redirect_to candidate_interface_decoupled_references_new_candidate_name_path(@reference.id)
elsif @submit_reference_form.send_request?
  CandidateInterface::DecoupledReferences::RequestReference.call(@reference, flash)
  redirect_to candidate_interface_decoupled_references_review_path
else
  redirect_to candidate_interface_decoupled_references_review_path
end

```

davidgisbey: Yea that's neater than mine was 👍 thanks

malcolmbaig (approved):

stevehook (approved):

PR #3111, opened 2020-10-12
Add service object for sending decoupled references emails

malcolmbaig:
## Context
Currently, if a candidate navigates back to a stale page that asks them
to confirm if they want to send a reference request email, and they
click 'yes', we send an email each time. We want to put a check in place
to ensure only one email may be sent per referee.

Introduce a separate service object to accommodate the change in behaviour
from the existing SendReferenceRequest.

Differences in this implementation:

- Pass the flash object in to conditionally set success/failure messages
- Check the state of the reference and then either send the request or
  set a failure message.
## Guidance to review
- Brand new app, no first/last name filled in. Add a reference. Don't send it yet. Click 'Send request' on the review page. Fill in names and confirm send. Click back button and confirm again. Should see error and only one email should be sent.
- Add another reference and repeat the above (same flow, except name form won't appear). 

## Link to Trello card

https://trello.com/c/Bt876U5Q
## Things to check

### DIFF
--- a/app/controllers/candidate_interface/decoupled_references/candidate_name_controller.rb
+++ b/app/controllers/candidate_interface/decoupled_references/candidate_name_controller.rb
@@ -17,9 +17,7 @@ module CandidateInterface
 
         @reference_candidate_name_form.save(@reference)
 
-        # TODO: Refactor this to DRY wrt to RequestController
-        CandidateInterface::RequestReference.call(@reference)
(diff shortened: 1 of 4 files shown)


stevehook (approved):
Looks good, @davidgisbey is working in the same area with https://github.com/DFE-Digital/apply-for-teacher-training/pull/3110 so I guess it might move but the service object is applicable in any case.

PR #3113, opened 2020-10-12
Cache requests to Find for 5 minutes

tijmenb:
## Context

We're launching the new cycle tomorrow, and we're expecting higher traffic than normal. One of the first pages that users will hit is the "Apply from Find" page (eg. https://www.apply-for-teacher-training.service.gov.uk/candidate/apply?courseCode=2QW6&providerCode=G48). This page makes a request to the Find API if it can't find the course locally.

Cache the response for 5 minutes, so we avoid hammering the API if there's a lot of traffic.

Also implements better logging - we'll be able to see outgoing API requests in Logit.

## Guidance to review

Ok?

## Link to Trello card

Not sure, @tvararu ?

## Things to check

### DIFF
--- a/app/models/find_api/resource.rb
+++ b/app/models/find_api/resource.rb
@@ -4,3 +4,24 @@ module FindAPI
     self.connection_options = { headers: { user_agent: 'Apply for teacher training' } }
   end
 end
+
+require 'faraday/logging/formatter'
+
+class APIRequestLoggingFormatter < Faraday::Logging::Formatter
+  # Don't do logging when the request starts
+  def request(env); end
+
+  def response(env)
+    info('Response') do
+      {
+        api_request_url: "#{env.method.upcase} #{env.url}",
+        response_status: env.status,
+        runtime_ms: (env.response_headers['x-runtime'].to_f * 1000).to_i,
+      }
+    end
+  end
+end
+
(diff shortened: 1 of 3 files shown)


app/models/find_api/resource.rb
```diff
@@ -1,5 +1,6 @@
 module FindAPI
   class Resource < JsonApiClient::Resource
     self.site = ENV.fetch('FIND_BASE_URL')
+    self.connection_options = { headers: { user_agent: 'Apply for teacher training' } }
```
duncanjbrown, line 4: 💯

app/models/find_api/resource.rb
```diff
@@ -4,3 +4,24 @@ class Resource < JsonApiClient::Resource
     self.connection_options = { headers: { user_agent: 'Apply for teacher training' } }
   end
 end
+
+require 'faraday/logging/formatter'
+
+class APIRequestLoggingFormatter < Faraday::Logging::Formatter
```
duncanjbrown, line 10: Should we namespace this under `FindAPI`?

duncanjbrown (approved):
LGTM, nit _a volonté_

tvararu (approved):

tijmenb:
@duncanjbrown je decide de merge la request du pull en ce interést de timing

PR #3116, opened 2020-10-12
Downcase referee emails when saving

tvararu:
## Context

Emails should be downcased when submitted.

Add a `.downcase` call when building the form object.

## Guidance to review

Makes sense?

## Link to Trello card

https://trello.com/c/jorKBGZt/2324-save-case-insensitive-email-for-referees

## Things to check

### DIFF
--- a/app/forms/candidate_interface/reference/referee_email_address_form.rb
+++ b/app/forms/candidate_interface/reference/referee_email_address_form.rb
@@ -14,7 +14,7 @@ module CandidateInterface
 
     def self.build_from_reference(reference)
       new(
-        email_address: reference.email_address,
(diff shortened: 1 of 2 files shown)


malcolmbaig (approved):

davidgisbey:
Out of interest, why don't we just downcase the email when we save it to the db?

tvararu:
@davidgisbey there's not much difference. If we downcase when building the form however, a user that types in the email "BAD@" will then get back "bad@" in their email field (alongside the validation error), which kind of promps them that it should be lowercase. But this isn't significant enough... I did it there just because it was the first place I thought to!

PR #3117, opened 2020-10-12
Fix the decoupled references link from the application page

malcolmbaig:
Fix the decoupled references link from the application page

- Link to the start page if no references present. Link to the review page
  if at least one reference present.
- Make sure the Incomplete/Complete label next to the link is correct,
  based on how many complete references are present.

## Guidance to review
- Check the link to the references section on the application form page:

Screenshot_20201012_160926

- Should link to start page if no references, review page once a reference has been added.
- Should switch to Complete once two references given.

## Things to check

### DIFF
--- a/app/views/candidate_interface/unsubmitted_application_form/show.html.erb
+++ b/app/views/candidate_interface/unsubmitted_application_form/show.html.erb
@@ -53,7 +53,7 @@
         <li class="app-task-list__item">
           <%= render(
             TaskListItemComponent.new(
-              text: 'Add your references',
+              text: @application_form_presenter.references_link_text,
               completed: @application_form_presenter.enough_references_provided?,
               path: @application_form_presenter.references_path,
             )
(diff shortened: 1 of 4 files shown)


app/views/candidate_interface/unsubmitted_application_form/show.html.erb
```diff
@@ -54,8 +54,8 @@
           <%= render(
             TaskListItemComponent.new(
               text: 'Add your references',
```
davidgisbey, line 56: Not essential, but I think this should be conditional too right? So 'Review your references' or 'Add your references' based on the application_references.present? boolean.

davidgisbey, line 56:
```suggestion
              text: @application_form.application_references.present? ? 'Review your references' : 'Add your references',
```

davidgisbey (approved):
Minor comment, but it  looks good other than that 🚢

tvararu (approved):

PR #3131, opened 2020-10-13
Make "Applications" the first page in the support UI 

tijmenb:
## Context

This page is much more used, and is already the first element in the navigation.

- Redirect /support to /support/applications (instead of /support/candidates)
- Rename "Candidates" in the primary nav to "Applications"

## Guidance to review

OK?

### DIFF
--- a/app/helpers/navigation_items.rb
+++ b/app/helpers/navigation_items.rb
@@ -21,7 +21,7 @@ class NavigationItems
     def for_support_primary_nav(current_support_user, current_controller)
       if current_support_user
         [
-          NavigationItem.new('Candidates', support_interface_applications_path, is_active(current_controller, %w[candidates change_course import_references application_forms ucas_matches])),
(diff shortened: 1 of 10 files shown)


app/helpers/navigation_items.rb
```diff
@@ -21,7 +21,7 @@ def for_candidate_interface(current_candidate, _current_controller)
     def for_support_primary_nav(current_support_user, current_controller)
       if current_support_user
         [
-          NavigationItem.new('Candidates', support_interface_applications_path, is_active(current_controller, %w[candidates change_course import_references application_forms ucas_matches])),
```
paulrobertlloyd, line 24: Not sure we need to change the label in the primary nav: Candidates > Applications is fine.

tijmenb: Hmm, I found it a bit confusing when I tried it. Mind if we keep it "Applications"? Mostly because otherwise I have to untangle the changes in this P.

paulrobertlloyd: Sure. As I say, this section needs some work; the relationship between candidates and applications isn’t super clear (I’m trying to make it so, gradually)

PR #3133, opened 2020-10-13
Unique email address validation should only trigger on save

davidgisbey:
## Context

The review page calls valid on all the individual reference forms here.

image

If email address is nil, It fails on the unique email address validation as there is no email address input to downcase. This breaks the review page.

This validation should only be called on save.

- Only call the unique email address validation on save

## Things to check


tvararu (approved):

tijmenb (commented):
Is it possible to add a test for this?

tijmenb (approved):

PR #3137, opened 2020-10-13
Support: Refactor provider user

paulrobertlloyd:
## Context

* Shows summary of a user’s providers on table view (not all permissions), adds direct link to edit permissions
* Splits provider user into edit and show views
* Shows user’s providers and permissions on view page, with link to edit

provider-users

provider-user

edit-provider-user

## Guidance to review

* Does this seem right, easy to use
* Tests are failing… and I can’t for the life of me figure out what’s wrong!

## Things to check

### DIFF
--- a/app/controllers/support_interface/provider_users_controller.rb
+++ b/app/controllers/support_interface/provider_users_controller.rb
@@ -73,7 +73,7 @@ module SupportInterface
       provider_user = ProviderUser.find(params[:provider_user_id])
       provider_user.update!(send_notifications: !provider_user.send_notifications)
       flash[:success] = 'Provider user updated'
-      redirect_to edit_support_interface_provider_user_path(provider_user)
+      redirect_to support_interface_provider_user_path(provider_user)
     end
 
   private
--- a/app/controllers/support_interface/provider_users_controller.rb
+++ b/app/controllers/support_interface/provider_users_controller.rb
@@ -13,7 +13,7 @@ module SupportInterface
     end
 
     def show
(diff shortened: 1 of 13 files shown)


adamsilver (approved):

app/views/support_interface/provider_users/_provider_user_navigation.html.erb
```diff
@@ -18,6 +18,6 @@
 <% end %>
 
 <%= render TabNavigationComponent.new(items: [
-  { name: 'Details', url: edit_support_interface_provider_user_path(@form.provider_user) },
```
duncanjbrown, line 21: I think this partial could depend on a plain `provider_user` not a `form` with a `provider_user` property?

app/controllers/support_interface/provider_users_controller.rb
```diff
@@ -13,7 +13,8 @@ def index
     end
 
     def show
-      redirect_to edit_support_interface_provider_user_path(params[:id])
+      provider_user = ProviderUser.find(params[:id])
+      @form = ProviderUserForm.from_provider_user(provider_user)
```
duncanjbrown, line 17: I think we should pass `@provider_user` instead of form — confusing to have a `@form` on a show page

duncanjbrown (commented):
Massive improvement, thank you! Lmk if comment about `@form` unclear

paulrobertlloyd:
@duncanjbrown Ready for review. Fixed one last test that was failing after rebase, 2 other failing tests aren’t related to this PR, I don’t think.

PR #3146, opened 2020-10-14
Decoupled references - update application page #2

stevehook:
## Context

As a candidate I need to quickly understand the status of my references so that so I know if I need to manage my references.

This PR is the second of 2 or more and adds a list of references with their current status.

- [x] Add `TaskListItemReferencesComponent` to render the list of individual references and add this to the existing template.
- [x] Update system spec.

image

## Guidance to review

- Added a lookup table of sorts for the colours to the existing `references.yml`. Is there are better way or prior art for this? (Colours are taken from the related card https://trello.com/c/V9nKgrg3/2225-💔-dev-review-reference-states).
- Anything missing?

## Link to Trello card

https://trello.com/c/ZdPrMUfo/2219-💔-dev-update-your-application-page

## Things to check


app/frontend/styles/application.scss
```diff
@@ -37,5 +37,6 @@ $govuk-breakpoints: (
 @import "_qualification";
 @import "_section";
 @import "_status-box";
+@import "_status-indicator";
```
paulrobertlloyd, line 40: I think this stylesheet can be scoped to just candidate, for now rather than be included for all interfaces.

app/views/candidate_interface/unsubmitted_application_form/show.html.erb
```diff
@@ -53,14 +53,23 @@
...
+              )
+            ) %>
+          </div>
+          <div class="app-task-list__content">
+            <%= render(
+              TaskListItemReferencesComponent.new(
+                references: @application_form_presenter.references,
+              )
+            ) %>
+          </div>
```
paulrobertlloyd, line 72: Can we have this additional `div.app-task-list__content` appear only for the references section?

paulrobertlloyd (changes requested):
Will have a look when deploy ready, but a few comments in the meantime.

app/components/task_list_item_references_component.html.erb
```diff
@@ -0,0 +1,9 @@
+<ul class="govuk-list">
+  <% references.each do |reference| %>
+    <li class="govuk-body-s">
+      <span class="app-status-indicator app-status-indicator--<%= colour_for(reference) %>">
+      </span>&nbsp;
+      <b><%= reference.name %></b>: <%= status_label_for(reference) %>
```
paulrobertlloyd, line 6: Can this be on one line? Affects the space between the indicator and the name, otherwise.

```suggestion
      <span class="app-status-indicator app-status-indicator--<%= colour_for(reference) %>"></span>&nbsp;<b><%= reference.name %></b>: <%= status_label_for(reference) %>
```

app/views/candidate_interface/unsubmitted_application_form/show.html.erb
stevehook: This section of code is only for the references section but I've moved the `<div>` inside the component and will only render it when there is 1 or more references to show.

paulrobertlloyd: Perfect. Perhaps we can do the same for `TaskListItemComponent`, to be consistent?

paulrobertlloyd (approved):
Thanks @stevehook!

PR #3147, opened 2020-10-14
Generate exports via Sidekiq

tijmenb:
## Context

Our data exports are increasingly large, causing long-running requests - some take over 2 minutes to generate now.

This PR lets us generate the files in a Sidekiq worker instead. The worker saves the result in the database, and support users can download the exports via the UI.

This page now shows  "Generate .. export":

image

If you click that the export will start, and you will be redirected to the export's page that tells you to wait:

image

Once it's completed a download button appears:

image

## Things to check


spec/system/support_interface/application_choices_performance_spec.rb
```diff
@@ -23,7 +29,28 @@ def when_i_visit_the_service_performance_data_page
...
+    Sidekiq::Worker.clear_all
+    click_link 'Generate candidate journey tracking export'
+  end
+
+  def then_i_see_that_the_export_has_started
+    expect(page).to have_content 'This export is being generated'
+  end
+
+  def when_the_sidekiq_worker_has_finished
+    Sidekiq::Worker.drain_all
```
stevehook, line 42: :+1: not seen this way of testing with Sidekiq before.

stevehook (approved):
This all looks good to me. It worked perfectly running locally though for some reason the review app isn't working (the job gets queued in Sidekiq but not processed).

I wondered about the security implications of this feature. If I'm logged in I can navigate to `/support/data-exports/<guess-a-number>` but it doesn't work if I'm signed out so unless we have different privilege levels for different support users this is all good. I guess we could match the `initiator` to `current_support_user` if this ever mattered.

I was curious about why you picked `binary` for the `data` column rather than `text` but I guess this makes the feature a little more general - it could handle images etc. if ever needed?

spec/system/support_interface/application_choices_performance_spec.rb
tijmenb: Took me a while to figure out how to do it!

tijmenb:
Re: security, it's indeed just available to support users. That should be fine for now, people should be able to see eachothers reports.

I picked binary because that looked like the correct thing to store a file:

https://www.postgresql.org/docs/9.0/datatype-binary.html

> A binary string is a sequence of octets (or bytes). Binary strings are distinguished from character strings in two ways. First, binary strings specifically allow storing octets of value zero and other "non-printable" octets (usually, octets outside the range 32 to 126). Character strings disallow zero octets, and also disallow any other octet values and sequences of octet values that are invalid according to the database's selected character set encoding. Second, operations on binary strings process the actual bytes, whereas the processing of character strings depends on locale settings. In short, **binary strings are appropriate for storing data that the programmer thinks of as "raw bytes"**, whereas character strings are appropriate for storing text.

PR #3148, opened 2020-10-14
Rework specs for decoupled references

davidgisbey:
## Context

Several specs throughout the test suite (mostly system specs) have the `decoupled_references` feature flag deactivated in their test setup. This was to get them passing while the decoupled references MVP was being built. We should update these so that they're passing with the feature enabled.

This is the first of a few PRs

- copy existing specs into a new specs_to_be_deleted folder
- Fix the existing spec so it works
- If the decoupled flow doesn't have the required functionality add a TODO
- Updates the candidate_helper so it now receives reference responses before submission

## Guidance to review

This isn't as daunting as the line count suggests. The new spec files can be ignored as they have just been duped and moved to a new location.

I think reviewing by commit is the way to go with this one.

## Link to Trello card

https://trello.com/c/pizcHLeY/2307-%F0%9F%92%94-dev-rework-existing-specs-for-decoupled-references

## Things to check


tijmenb (approved):

spec/system/candidate_interface/candidate_replaces_reference_after_applying_again_spec.rb
```diff
@@ -4,6 +4,9 @@
   include CandidateHelper
 
   scenario 'Can replace a completed reference' do
+    # TODO: this feature cannot pass with the flag off at the moment as currently,
```
malcolmbaig, line 7: `flag on`

PR #3150, opened 2020-10-14
Re-request a failed reference

stevehook:
## Context

As a candidate I need to re-open a reference request so that I can request reference that may have been cancelled by me, cancelled during the end-of-cycle carry over or failed to be delivered.

- [x] Add _Send reference again_ link to review component
- [x] Extend system spec
- [x] Handle cancelled at end-of-cycle status
- [x] Hide link when 2 references have been provided

image

## Guidance to review

- To test manually I had to put a reference into a `cancelled`, `cancelled_at_end_of_cycle` or `email_bounced` state by changing the data manually.

## Link to Trello card

https://trello.com/c/Lwt9lQMb/2237-💔-dev-re-request-a-cancelled-failed-reference-request

## Things to check

### DIFF
--- a/app/components/candidate_interface/decoupled_references_review_component.html.erb
+++ b/app/components/candidate_interface/decoupled_references_review_component.html.erb
@@ -25,7 +25,7 @@
             <li class="app-summary-card__actions-list-item">
               <%= link_to candidate_interface_decoupled_references_new_request_path(reference.id), class: 'govuk-link' do %>
                 <%= t('application_form.referees.resend_request') %>
-                <span class="govuk-visually-hidden"> <%= reference.name %></span>
+                <span class="govuk-visually-hidden"> to <%= reference.name %></span>
               <% end %>
             </li>
           <% end %>
--- a/app/components/candidate_interface/decoupled_references_review_component.html.erb
+++ b/app/components/candidate_interface/decoupled_references_review_component.html.erb
@@ -21,6 +21,13 @@
                 <span class="govuk-visually-hidden"> <%= reference.name %></span>
               <% end %>
             </li>
+          <% elsif can_resend?(reference) %>
+            <li class="app-summary-card__actions-list-item">
+              <%= link_to candidate_interface_decoupled_references_new_request_path(reference.id), class: 'govuk-link' do %>
+                <%= t('application_form.referees.resend_request') %>
+                <span class="govuk-visually-hidden"> <%= reference.name %></span>
+              <% end %>
+            </li>
           <% end %>
         </ul>
       </div>
--- a/app/components/candidate_interface/decoupled_references_review_component.rb
+++ b/app/components/candidate_interface/decoupled_references_review_component.rb
@@ -30,7 +30,8 @@ module CandidateInterface
     end
 
     def can_resend?(reference)
-      (reference.cancelled? || reference.email_bounced?) &&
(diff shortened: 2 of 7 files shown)


app/components/candidate_interface/decoupled_references_review_component.html.erb
```diff
@@ -21,6 +21,13 @@
                 <span class="govuk-visually-hidden"> <%= reference.name %></span>
               <% end %>
             </li>
+          <% elsif can_resend?(reference) %>
+            <li class="app-summary-card__actions-list-item">
+              <%= link_to candidate_interface_decoupled_references_new_request_path(reference.id), class: 'govuk-link' do %>
+                <%= t('application_form.referees.resend_request') %>
+                <span class="govuk-visually-hidden"> <%= reference.name %></span>
```
paulrobertlloyd, line 28:
```suggestion
                <span class="govuk-visually-hidden"> to <%= reference.name %></span>
```

paulrobertlloyd (approved):
Looks like I can’t test this easily in the review app, but everything else looks good 😀

app/components/candidate_interface/decoupled_references_review_component.rb
```diff
@@ -29,6 +29,15 @@ def can_send?(reference)
...
     end
 
+    def can_resend?(reference)
+      (reference.cancelled? || reference.cancelled_at_end_of_cycle? || reference.email_bounced?) &&
+        !reference.application_form.enough_references_have_been_provided? &&
+        CandidateInterface::Reference::SubmitRefereeForm.new(
+          submit: 'yes',
+          reference_id: reference.id,
+        ).valid?
+    end
```
malcolmbaig, line 39: Is it possible to continuously cycle through cancel/resend on a reference, thus spamming the provider? There wouldn't be a sensible reason to do this, but is it worth guarding against it just in case?

stevehook: @malcolmbaig Do you think an `ApplicationReference#resent_at` attribute together with a sensible minimum time limit between resends would solve the problem? Can you think of a simpler solution?

malcolmbaig: Given the low likelihood, let's review if this is a reported problem.

spec/system/candidate_interface/decoupled_references/candidate_requests_a_reference_spec.rb
```diff
@@ -43,10 +43,17 @@
...
     then_i_should_not_see_a_send_reference_link
     when_i_navigate_to_the_send_request_page
     then_i_should_see_a_not_found_message
+
+    when_i_have_a_cancelled_reference
+    and_i_visit_the_all_references_review_page
+    and_i_click_the_resend_reference_link
+    and_i_confirm_that_i_am_ready_to_send_a_reference_request
+    then_i_see_a_confirmation_message
+    and_the_reference_is_moved_to_the_requested_state
```
malcolmbaig, line 56: is it worth including an assertion about the email that's sent as well?

malcolmbaig (approved):
LGTM, just one question about the system spec

spec/system/candidate_interface/decoupled_references/candidate_requests_a_reference_spec.rb
stevehook: Yes should have added that. Done now.

PR #3151, opened 2020-10-14
Support: Consolidate settings pages

paulrobertlloyd:
## Context

Tidy up routes and paths for settings related pages.

* Moves feature flags and cycle views and controllers to live under settings
* Moves `/support/feature-flags` → `/support/settings/feature-flags`
* Moves `/support/cycles` → `/support/settings/cycles `
* Moves `/support/tasks` → `/support/settings/tasks `
* Adds a redirect for `/support/settings ` → `/support/settings/feature-flags `

## Guidance to review

Should tasks live under settings as well (they share the same navigation, after all)? Or should nothing live under settings? Trying to be consistent and tidy, but don’t want to too tightly couple things to current IA either.

## Things to check


duncanjbrown (approved):
This is great, thank you!

On whether or not Tasks should live under Settings (and leaving aside the question of whether a Task is a Setting 😉 ) — I guess you mean the controller? — I think what's here is fine for now. `routes.rb` is doing the work to choose the controller based on URL, and that's what it's there for. If we did end up enriching the cycles or feature-flags parts of settings maybe we could consider a `Settings` module, eg `support_interface/settings/cycles_controller`, `support_interface/settings/tasks_controller` etc.

PR #3152, opened 2020-10-14
Flag which UCAS matches need an action in Support

AgaDufrat:
## Context

We need to flag which UCAS matches need our action in support so the support agents can send emails to the candidates.
In most cases, candidates have applications on both systems but there is no conflict.

Add a "Action needed" tag on the index page
Screenshot 2020-10-14 at 16 55 04

## Guidance to review

Note: since we still have some data for applications for courses run by Welsh providers with no indication of status on UCAS,  we don't want to consider those applications hence `return false if provider_not_on_apply?`

## Link to Trello card

https://trello.com/c/CGTMARVG/2810-show-ucas-matches

## Things to check


app/models/ucas_match.rb
```diff
@@ -8,4 +8,31 @@ class UCASMatch < ApplicationRecord
     new_match: 'new_match',
     processed: 'processed',
   }
+
+  def action_needed?
+    return false if processed?
+
+    application_for_the_same_course_in_progress_on_both_services? || application_accepted_on_ucas_and_in_progress_on_apply? || application_accepted_on_apply_and_in_progress_on_ucas?
```
duncanjbrown, line 15: 👌

duncanjbrown (approved):
Great stuff @AgaDufrat !

PR #3153, opened 2020-10-14
Add cancel a request functionality to decoupled references

george-bit:
## Context

As part of the decoupled references feature candidates will need the ability to cancel their reference requests.

Minor tweaks to content as per labels and wording but otherwise the same functionality that exists within the current referees section.

## Guidance to review

Check against the card criteria with Feature Flag on and off in review. Let me know if test coverage is suffiecient

## Link to Trello card

https://trello.com/c/mAoR8HLs/2222-%F0%9F%92%94-dev-cancel-a-reference-request

## Things to check


config/locales/en.yml
```diff
@@ -115,6 +115,11 @@ en:
     referee_type: What kind of referee do you want to add?
     referee_destroy: Are you sure you want to delete this referee?
     referee_cancel: Are you sure you want to cancel this request for a reference?
+<<<<<<< HEAD
```
TobyRet, line 118: @george-bit - woops!

george-bit: removed marker 👍

george-bit:
@davidgisbey after speaking with @TobyRet,  I've created a spec file to cover the new CancelReference service, noticed that this does not exist currently for the RequestReference service? Do we have a plan for handling deprecation of the old referees code once the feature flag is removed? Would be good to have a discussion.

spec/system/candidate_interface/decoupled_references/review_references_spec.rb
```diff
@@ -115,6 +116,17 @@ def and_i_can_delete_a_reference
...
 
+  def and_i_can_cancel_a_sent_reference
+    within '#references_sent' do
+      click_link 'Cancel request'
+    end
+    click_button 'Yes I’m sure'
+
+    expect(page).to have_current_path candidate_interface_decoupled_references_review_path
+    expect(page).to have_content("Reference request cancelled for #{@requested_reference.name}")
+    expect(page).to have_content('Cancelled')
```
malcolmbaig, line 127: Worth checking the referee received the `reference_cancelled_email`

spec/services/cancel_referee_spec.rb
```diff
@@ -1,4 +1,5 @@
 require 'rails_helper'
+# TODO: Remove file with Decoupled references FeatureFlag as part of deprication
```
malcolmbaig, line 2: `deprecation`

george-bit: D'oh

config/locales/en.yml
```diff
@@ -115,6 +115,8 @@ en:
     referee_type: What kind of referee do you want to add?
     referee_destroy: Are you sure you want to delete this referee?
     referee_cancel: Are you sure you want to cancel this request for a reference?
+    decoupled_reference_cancel: Are you sure you want to cancel this reference request?
```
malcolmbaig, line 118: Maybe call this `references_cancel_request` for consistency with the other recent additions. Means no need to revisit this when we do clean up further down the road.

config/locales/en.yml
```diff
@@ -115,6 +115,8 @@ en:
...
+    reference_destroy: Are you sure you want to delete this reference request?
```
malcolmbaig, line 119: Stray addition? Not seeing this used anywhere.

malcolmbaig (changes requested):
LGTM, just a few bits to revisit.

config/locales/en.yml
george-bit: Hmm yeah I think it must be. Thought it was in place already but must've added it experimentally. Got rid of it.

davidgisbey:
LGTM once Malcolms changes are made. In terms of getting rid of the old code, all the referees/additional referees code should just be able to be deleted. The card i'm working on at the moment has pulled up a bunch of work that needs to be completed beforehand, including getting all the specs passing with the decoupled flag on, and moving the old specs into a specs_to_be_deleted folder. 

The final step of that card will be deleting the feature flag and all the old referee/additional referee code.

spec/system/candidate_interface/decoupled_references/review_references_spec.rb
george-bit: Given it a go. Think it's okay, let me know if not!

malcolmbaig (approved):

PR #3156, opened 2020-10-15
Only show references that have feedback to providers

tijmenb:
## Context

We're currently showing the provider all references for a candidate, even unsubmitted/cancelled/refused ones.

When it was built we only ever sent applications to providers when there were 2 references, so we didn't think to filter.

We noticed this in the API 2 weeks ago https://github.com/DFE-Digital/apply-for-teacher-training/pull/3039.

Only show references that have feedback and add a test.

## Link to Trello card

https://trello.com/c/r2ae9tZs

### DIFF
--- a/app/services/support_interface/tad_application_export.rb
+++ b/app/services/support_interface/tad_application_export.rb
@@ -1,0 +1,71 @@
+module SupportInterface
+  class TADApplicationExport
+    attr_reader :application_choice
+
+    delegate :course_option, :course, :application_form, to: :application_choice
+    delegate :candidate, to: :application_form
+
+    def initialize(application_choice)
+      @application_choice = application_choice
+    end
+
+    def as_json
+      accrediting_provider = course.accredited_provider || application_choice.provider
+
+      # https://ukgovernmentdfe.slack.com/archives/CP18YJXPY/p1602583138300500
+      {
+        application_choice_id: application_choice.id,
+        application_id: application_form.id,
+        status: application_choice.status,
+        phase: application_form.phase,
+        first_name: application_form.first_name,
+        last_name: application_form.last_name,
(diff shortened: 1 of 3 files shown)


spec/system/provider_interface/see_individual_application_spec.rb
```diff
@@ -132,19 +132,26 @@ def and_my_organisation_has_received_an_application
...
            application_form: application_form,
            name: 'C3PO',
            email_address: 'c3p0@rebellion.org',
            relationship: 'Companion droid',
            feedback: 'The possibility of successfully navigating training is approximately three thousand seven hundred and twenty to one')
 
+    create(:reference,
+          :refused,
+          application_form: application_form,
+          name: 'BB-8')
```
duncanjbrown, line 153: 📐

duncanjbrown (approved):
We've now missed this twice! Clients have to remember to filter 😬

duncanjbrown:
Bit inclined to introduce a single function which returns appropriate references. Can't enforce usage, though. How would life be if we had a `ProviderApplication` decorator?

tijmenb:
@duncanjbrown we should explore `ProviderApplication`. We could raise if it's instantiated with a unsubmitted application for example. Are there other things like that?

duncanjbrown:
> @duncanjbrown we should explore `ProviderApplication`. We could raise if it's instantiated with a unsubmitted application for example. Are there other things like that?

That would be a good thing. I don't think there's another place where we scope queries right now. Think it's worth a little spike?

app/services/support_interface/tad_application_export.rb
```diff
@@ -0,0 +1,71 @@
+module SupportInterface
```
tijmenb, line 1: This file was accidentally committed.

PR #3157, opened 2020-10-15
Rework referee system specs with flag toggled off and fix chaser email workers

davidgisbey:
## Context

Previous PRs 
#3148 duplicated the first set of specs. Moved them into the specs_to_be_deleted folder and made the originals pass

This PR does the same as above with the referee system specs.

It also updates the workers which send the first and second reference-request chaser emails to use references in the awaiting feedback state where the application form has not been submitted.

- Move ref system specs with the decoupled references flag the specs_to_be_deleted_folder
- Get the originals passing with the flag active
- Update the workers to send chasers to the correct candidates 

## Guidance to review

Some eyes on the worker-related behaviour would be great. I'll leave a comment.

## Link to Trello card

https://trello.com/c/pizcHLeY/2307-%F0%9F%92%94-dev-rework-existing-specs-for-decoupled-references

## Things to check

### DIFF
--- a/app/queries/get_references_to_chase.rb
+++ b/app/queries/get_references_to_chase.rb
@@ -1,9 +1,17 @@
 class GetReferencesToChase
   def self.call
-    ApplicationReference
-      .feedback_requested
-      .where(['requested_at < ?', chase_referee_time_limit])
-      .where.not(id: ChaserSent.reference_request.select(:chased_id))
+    if FeatureFlag.active?(:decoupled_references)
+      ApplicationReference
+        .feedback_requested
+        .where(['requested_at < ?', chase_referee_time_limit])
+        .where(application_form_id: ApplicationForm.where(submitted_at: nil).pluck(:id))
+        .where.not(id: ChaserSent.reference_request.select(:chased_id))
+    else
+      ApplicationReference
+        .feedback_requested
+        .where(['requested_at < ?', chase_referee_time_limit])
+        .where.not(id: ChaserSent.reference_request.select(:chased_id))
+    end
   end
 
   def self.chase_referee_time_limit
--- a/app/workers/send_additional_reference_chase_email_to_both_parties_worker.rb
+++ b/app/workers/send_additional_reference_chase_email_to_both_parties_worker.rb
@@ -22,11 +22,19 @@ private
       .feedback_requested
       .where(['requested_at < ?', time_limit])
 
-    application_form_ids =
-      outstanding_reference_requests
-      .where(application_form_id: applications_waiting_on_reference_feedback.pluck(:id))
-      .distinct(:application_form_id)
-      .pluck(:application_form_id)
+    if FeatureFlag.active?(:decoupled_references)
+      application_form_ids =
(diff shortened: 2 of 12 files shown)


app/queries/get_references_to_chase.rb
```diff
@@ -1,9 +1,17 @@
...
-    ApplicationReference
-      .feedback_requested
-      .where(['requested_at < ?', chase_referee_time_limit])
-      .where.not(id: ChaserSent.reference_request.select(:chased_id))
+    if FeatureFlag.active?(:decoupled_references)
+      ApplicationReference
+        .feedback_requested
+        .where(['requested_at < ?', chase_referee_time_limit])
+        .where(application_form_id: ApplicationForm.where(submitted_at: nil).pluck(:id))
+        .where.not(id: ChaserSent.reference_request.select(:chased_id))
```
davidgisbey, line 8: Does this seem correct? Getting all references past the time limit, whose application forms have not been submitted and have not previously been chased.

app/workers/send_additional_reference_chase_email_to_both_parties_worker.rb
```diff
@@ -22,11 +22,19 @@ def applications_that_need_chasing
...
-      outstanding_reference_requests
-      .where(application_form_id: applications_waiting_on_reference_feedback.pluck(:id))
-      .distinct(:application_form_id)
-      .pluck(:application_form_id)
+    if FeatureFlag.active?(:decoupled_references)
+      application_form_ids =
+        outstanding_reference_requests
+        .where(application_form_id: unsubmitted_application_forms.pluck(:id))
+        .distinct(:application_form_id)
+        .pluck(:application_form_id)
```
davidgisbey, line 30: Same as above. Does it make sense?

app/queries/get_references_to_chase.rb
malcolmbaig: This looks correct to me, although the explicit check on `submitted_at` hints that we probably also need to cancel any outstanding reference requests when a candidate submits their application. With this logic we won't send chasers for those but it does seem like a loose end to tie up. Will check and write a card if needed.

stevehook (approved):

app/workers/send_additional_reference_chase_email_to_both_parties_worker.rb
malcolmbaig: Agreed, this seems like a good opportunity to move that logic into something like a `GetReferencesToChaseAgain` (or similar) query object. There's probably some further refactoring to be done there, but at least this way it's clear that two similar pieces of logic are involved and they're cleanly abstracted away.

Another issue (outside the scope of this PR) is that we model each chase differently. The first is recorded as a chase on the reference, the second is a chase on the entire application form. Can't recall why we did it this way but worth revisiting.

app/queries/get_references_to_chase.rb
davidgisbey: Yea this is now being picked up via a separate card.

PR #3161, opened 2020-10-15
Enact permissions onboarding flow for orgs with 'invalid' permissions

steventux:
## Context

We deliberately omitted enabling safeguarding and diversity permissions for existing organisations when setting up their default permissions.
This made some relationship permissions 'invalid' as at least one provider in the relationship should have each permission enabled. If the managing user in the organisation has not amended the permissions they are likely still in this state. We identified 181 records with this issue.

This has the side effect of restricting the visibility of safeguarding/diversity information completely if neither provider has the permission enabled. Users who are not able to manage permissions will be unaware of why they cannot see this information.

- Amend the conditions of `ProviderSetup#next_pending_relationship` to return 'invalid' relationships. This will direct the managing user through the permissions setup flow when they log in.
- Unskip a test for `next_pending_relationship` which was relying on record order, this adds an order clause to ensure we retrieve the oldest records first.

## Link to Trello card

This PR covers the task _When a user with manage org/user permissions logs in and permissions haven't been set up, prompt them to set them up_

https://trello.com/c/UAsvSnET/2868-dev-changes-tweaks-for-permissions

## Things to check

### DIFF
--- a/app/services/provider_setup.rb
+++ b/app/services/provider_setup.rb
@@ -23,14 +23,11 @@ class ProviderSetup
   end
 
   def next_relationship_pending
-    permissions = ProviderRelationshipPermissions
-      .order(:created_at)
-      .where(training_provider: @provider_user.providers)
-      .find { |prp| prp.setup_at.blank? || prp.invalid? }
+    relationships_pending&.first
+  end
 
-    if permissions.present?
-      auth = ProviderAuthorisation.new(actor: @provider_user)
-      permissions if auth.can_manage_organisation?(provider: permissions.training_provider)
-    end
+  def relationships_pending
+    auth = ProviderAuthorisation.new(actor: @provider_user)
+    auth.training_provider_relationships_that_actor_can_manage_organisations_for.select { |prp| prp.setup_at.blank? || prp.invalid? }
   end
 end
--- a/app/services/provider_setup.rb
+++ b/app/services/provider_setup.rb
@@ -23,10 +23,10 @@ class ProviderSetup
   end
 
   def next_relationship_pending
-    permissions = ProviderRelationshipPermissions.find_by(
-      setup_at: nil,
-      training_provider: @provider_user.providers,
-    )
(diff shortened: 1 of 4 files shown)


app/services/provider_setup.rb
```diff
@@ -23,11 +23,21 @@ def next_agreement_pending
...
 
+    if permissions.blank?
+      setup_permissions = ProviderRelationshipPermissions
+        .includes(:training_provider, :ratifying_provider)
+        .where(training_provider: @provider_user.providers)
+        .where.not(setup_at: nil)
+        .order(:created_at)
+
+      permissions = setup_permissions.reject(&:valid?)&.first
+    end
```
duncanjbrown, line 39: Isn't the logic here 

- get all the permissions on which I'm the training provider
- if any lack setup_at, we need to set them up
- if any are invalid, we need to set them up

so rather than two passes we could do something like

```
ProviderRelationshipPermissions
  .where(training_provider: @provider_user.providers)
  .select { |prp| prp.setup_at.blank? || prp.invalid? }
```

duncanjbrown:
I think we'll also need to fix `#provider_relationship_permissions_needing_setup` in `ProviderRelationshipPermissionSetupController`?

steventux:
@duncanjbrown I've had a go at using the same piece of code to identify relationships manageable by the current provider user and then filtering these as suggested https://github.com/DFE-Digital/apply-for-teacher-training/pull/3161/commits/d2a614101ade8ae868114e34e8cd2542cdc3ed94

duncanjbrown:
Looking good — left one comment! (on a commit 🤦) https://github.com/DFE-Digital/apply-for-teacher-training/commit/d2a614101ade8ae868114e34e8cd2542cdc3ed94#r43406191

steventux:
@duncanjbrown I've amended as per your suggestion, I omitted the sort because the underlying query orders by `created_at` so the first selected relationship will be the oldest.

duncanjbrown (approved):
Great

PR #3162, opened 2020-10-15
Remove start page and redirect to '/candidate/account'

TobyRet:
## Context

- GOV.UK will be responsible for the Apply's start page
- We no longer need the existing start page 

- Redirect traffic to '/candidate/account'
- Remove the start page

## Guidance to review
- Question for devs - the GOVUK Apply start page url is currently littered throughout the code. Was thinking it might make sense to make this a global via the initialisers folder. What do people think?

## Link to Trello card

https://trello.com/c/ndQPEHst/2284-dev-too-many-apply-start-pages

## Things to check


app/views/candidate_interface/content/providers.html.erb
```diff
@@ -6,7 +6,7 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <p class="govuk-body">Use <%= govuk_link_to service_name, candidate_interface_start_path %> if the courses you want to do are listed on this page.</p>
+    <p class="govuk-body">Use <%= govuk_link_to service_name, candidate_interface_create_account_or_sign_in_path %> if the courses you want to do are listed on this page.</p>
```
paulrobertlloyd, line 9: I think in this instance, we’ll want to link to our start page on GOV.UK, as that has eligibility criteria etc, and should remain the start of the journey.

paulrobertlloyd (changes requested):

app/views/candidate_interface/content/providers.html.erb
TobyRet: @paulrobertlloyd - yep makes sense. Do we have the GOV.UK link yet? Couldn't see it in the ticket

paulrobertlloyd: Yup, it’s: https://www.gov.uk/apply-for-teacher-training

app/views/candidate_interface/start_page/eligibility.html.erb
```diff
@@ -1,5 +1,5 @@
 <% content_for :title, title_with_error_prefix(t('page_titles.eligibility'), @eligibility_form.errors.any?) %>
-<% content_for :before_content, govuk_back_link_to(candidate_interface_start_path) %>
+<% content_for :before_content, govuk_back_link_to(candidate_interface_create_account_or_sign_in_path) %>
```
paulrobertlloyd, line 2: No longer relevant as this page is behind a feature flag, but looks like this back link was wrong!

app/views/candidate_interface/content/providers.html.erb
TobyRet: Ta!

TobyRet: 👍

app/views/candidate_interface/start_page/eligibility.html.erb
```diff
@@ -1,5 +1,5 @@
 <% content_for :title, title_with_error_prefix(t('page_titles.eligibility'), @eligibility_form.errors.any?) %>
-<% content_for :before_content, govuk_back_link_to(candidate_interface_start_path) %>
+<% content_for :before_content, govuk_back_link_to(govuk_apply_start_page) %>
```
paulrobertlloyd, line 2:
```suggestion
<% content_for :before_content, govuk_back_link_to(candidate_interface_create_account_or_sign_in_path) %>
```

I think this back link was actually wrong before, and should have taken users back to create account/sign in page, which is the previous page they would have seen. Somewhat academic as this page is now hidden behind a feature flag.

paulrobertlloyd: @TobyRet Just this outstanding issue; back link on the eligibility page should take you back to create an account/sign in page.

TobyRet: Ah sorry my bad. Fixed

paulrobertlloyd (approved):

config/routes.rb
```diff
@@ -11,12 +11,11 @@
...
     root to: 'content#sandbox'
   else
-    root to: redirect('/candidate')
+    root to: redirect('/candidate/account')
   end
 
   namespace :candidate_interface, path: '/candidate' do
-    get '/' => 'start_page#show', as: :start
-
+    get '/', to: redirect('https://www.gov.uk/apply-for-teacher-training')
```
TobyRet, line 18: The GovUK Apply start page url is littered throughout the code. 
Would it make sense to make this a global via an initialiser?

app/views/layouts/_header.html.erb
```diff
@@ -2,7 +2,7 @@
 <% when 'candidate_interface' %>
   <%= render(HeaderComponent.new(classes: "app-header--#{HostingEnvironment.environment_name}",
                                  service_name: service_name,
-                                 service_url: service_link,
+                                 service_url: govuk_apply_start_page,
```
paulrobertlloyd, line 5: Header should link to dashboard/application page when signed in, but not be linked at all when signed out. Not sure this achieves that?

TobyRet: 🤔 So a few things to consider here.

1. I think this needs to revert back to the `service_url` - this url dynamically changes based on the path ('/provider'. '/support' etc) and redirects back to their respective starting (home?) pages
2. Given the `service_url` is based on `path`, this then may present a bigger technical problem if we want to then also change the url based on whether the user is signed in or not. This requires more logic and code changes.
3. I was looking at Find and the header does not link back to the GOVUK start page. Is this definitely the behaviour we want?
4. Longing for that 'one line code change' mvp mentioned in the ticket! 😆

paulrobertlloyd: 1. Yeah, that sounds best. Just need to work out where `service_url` goes when signed out.
2. Desirable, but not necessary, so ignore for now.
3. Ideally, but not important.
4. 😭

stevehook (approved):
Code looks good. I'm not sure how we would test the round-trip from apply-startpage-apply outside of production now without setting up a dummy start page somewhere for each environment. Not sure whether we need that.

The redirects etc. are a bit complex. If @paulrobertlloyd is happy then it's ok with me.

tvararu (approved):
Approved pending constantisation of the GOV.UK link :+1:

PR #3163, opened 2020-10-15
Restructure how we do data exports

tijmenb:
## Context

Follow up to https://github.com/DFE-Digital/apply-for-teacher-training/pull/3147. This PR aims to make it really easy to add more reports. After this PR, you only have add a hash to the DataExports class to make your class exportable.

- Use a constant in the DataExport class to populate the UI with reports, so you only have to define it once
- Add an index page for data exports
- Move the report generation to a "/new" page
- Make all exports run in the background
- Delete individual system tests for report downloads, since the code is the same everywhere now. The exports themselves are tested in unit tests.

## Guidance to review

Test it out locally

## Link to Trello card

https://trello.com/c/OCkzzyo6/2770-generate-tad-export


app/views/support_interface/data_exports/index.html.erb
```diff
@@ -0,0 +1,25 @@
...
+<table class="govuk-table">
+  <thead class="govuk-table__head">
+    <tr class="govuk-table__row">
+      <th scope="col" class="govuk-table__header">Name</th>
+      <th scope="col" class="govuk-table__header">Initiated by</th>
+    </tr>
+  </thead>
+
+  <tbody class="govuk-table__body">
+    <% @data_exports.each do |data_export| %>
```
stevehook, line 14: Probably will need to paginate this sooner or later

stevehook (approved):
:+1: Tested review app and all seems to work fine.

PR #3164, opened 2020-10-15
Add reminder_sent_at column to references table

malcolmbaig:
## Context
We want to allow candidates to manually send a reminder to referees
after the initial reference request. 

Add a timestamp to mark when this
occurs so that we can:

- eventually include this event in the reference history
- limit manual reminders to one per reference by checking for the
  presence of the timestamp

## Guidance to review
- An alternative to this is using the `ChaserSent` model and making the manual reminder another type of chaser. An attribute on the reference seemed like a better fit, as the chaser abstraction appears to deal solely with automated messaging we take care of in batch (I think?). This way there's a separation between state change triggered by the candidate (`requested_at`, `replaced_at`, `reminder_sent_at`) vs state change triggered automatically by our system (records in the `chasers_sent` table). WDYT?

## Link to Trello card
https://trello.com/c/rjXbj7ky

## Things to check


tijmenb (approved):
This makes sense if we want to limit it to one (or one per day or something). It updates the audit history as well, which is useful. 👍

stevehook (approved):
Separate column makes sense to me. I wonder how this will interact with the automated chasers though (could get a bit spammy from the referees point of view if the two types of reminder happen to be sent over a short time).

malcolmbaig:
> Separate column makes sense to me. I wonder how this will interact with the automated chasers though (could get a bit spammy from the referees point of view if the two types of reminder happen to be sent over a short time).

Yeah, I think the only mitigation we have for this currently is letting the candidate know when the next automated reminder will go out (working on that currently). May be worth revisiting if provider feedback suggests it's a problem.

PR #3167, opened 2020-10-16
Review reference states

stevehook:
## Context

As a candidate I need to know the status of my references so that I can take the appropriate follow-on action.

This PR updates the reference review page with the latest design and content.

- [x] Updates to the content and colours for the `DecoupledReferenceReviewComponent`

image
image
image
image
image
image
image
image

## Guidance to review

- Colours are now looked up from `reference.yml`
- Some of the status title changes will have leaked into the old `ReferenceReviewComponent` (I've not attempted to keep all these changes behind the `decoupled_references` feature flag).
- To test every single combination will probably need a bit of direct database manipulation.

## Link to Trello card

https://trello.com/c/V9nKgrg3/2225-💔-dev-review-reference-states

## Things to check

### DIFF
--- a/app/components/candidate_interface/decoupled_references_review_component.rb
+++ b/app/components/candidate_interface/decoupled_references_review_component.rb
@@ -106,38 +106,36 @@ module CandidateInterface
     end
 
     def feedback_status_content(reference)
-      if reference.not_requested_yet?
-        tag.p(t('application_form.referees.info.not_requested_yet'), class: 'govuk-body govuk-!-margin-top-2')
-      elsif reference.feedback_refused?
-        tag.p(t('application_form.referees.info.declined'), class: 'govuk-body govuk-!-margin-top-2')
-      elsif reference.cancelled_at_end_of_cycle?
-        tag.p(t('application_form.referees.info.cancelled_at_end_of_cycle'), class: 'govuk-body govuk-!-margin-top-2')
-      elsif reference.cancelled?
-        tag.p(t('application_form.referees.info.cancelled'), class: 'govuk-body govuk-!-margin-top-2')
-      elsif reference.feedback_overdue?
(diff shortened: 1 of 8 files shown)


app/components/candidate_interface/decoupled_references_review_component.rb
```diff
@@ -106,38 +106,36 @@ def feedback_status_text(reference)
...
+        elsif reference.feedback_requested?
+          t('application_form.referees.info.feedback_requested')
+        elsif reference.email_bounced?
+          t('application_form.referees.info.email_bounced')
+        end
+
+      return '' if text.blank?
+
+      if text.is_a?(Array)
+        text.inject('') { |prev, line| prev + tag.p(line, class: 'govuk-body govuk-!-margin-top-2') }.html_safe
```
malcolmbaig, line 127: First thought - does returning an array of `tag.p` rather than a string have the same effect when rendering?

If not, this might be clearer:
 ` text.each_with_object('') { |line, content| content + tag.p(line, class: 'govuk-body govuk-!-margin-top-2') }.html_safe`

malcolmbaig (approved):
LGTM :rainbow:

malcolmbaig:
Worth squashing these commits behind a single, explanatory commit message.

app/components/candidate_interface/decoupled_references_review_component.rb
stevehook: I tried with an array of 'tags' but I doesn't render properly - I think because the return value from this method is already appended to another (html_safe) string. I couldn't help thinking there was some kind of `array_of_tags` psuedo-tag-helper type thing but didn't find it. So just changed to use `each_with_object`.

paulrobertlloyd (approved):

PR #3168, opened 2020-10-16
Add recruitment cycle year to UCAS matches

AgaDufrat:
## Context

When checking application status on Apply we need to know which recruitment cycle the application was in. 
There is nothing in the data we get from UCAS to explicitly determine the cycle so we need to store it in the database.
UCAS said they will only send us matches for the current cycles so this is the year we will store in the db.

See commit messages for details

## Guidance to review

- generate some new applications `/support/tasks` -> Generate test applications
- there is no nil if you do `UCASMatch.pluck(:recruitment_cycle_year).uniq` in the console
- individual match pages still work: /support/ucas-matches

## Link to Trello card

https://trello.com/b/5IiPW0Ok/provendor-team-board

## Things to check


app/services/ucas_matching/process_matching_data.rb
```diff
@@ -88,6 +88,7 @@ def process_csv_file(file)
       candidates.each do |candidate_id, matching_data|
         match = UCASMatch.find_or_initialize_by(candidate_id: candidate_id)
         match.matching_data = matching_data
+        match.recruitment_cycle_year = RecruitmentCycle.current_year
```
duncanjbrown, line 91: Should this be inside a block attached to `find_or_initialize_by`, just in case we accidentally overwrite the year on an existing match?

db/migrate/20201015135045_backfill_recruitment_cycle_year_in_ucas_matches.rb
```diff
@@ -0,0 +1,10 @@
+class BackfillRecruitmentCycleYearInUCASMatches < ActiveRecord::Migration[6.0]
+  def change
+    start_of_2021_cycle = Date.new(2020, 10, 13)
+    ucas_matches_in_2020_cycle = UCASMatch.where('created_at < ?', start_of_2021_cycle)
+
+    ucas_matches_in_2020_cycle.update_all(recruitment_cycle_year: 2020)
```
duncanjbrown, line 6: Are we sure we have no 2021 matches? (we could have sent some relevant applications, as we send everything right now)

AgaDufrat: The last update was on 21 September so I think it's safe.

AgaDufrat: I will amend to update all that were created after the new recruitment cycle date to `2021`

app/services/ucas_matching/process_matching_data.rb
AgaDufrat: I understand what the problem may be here but not sure what you mean by the block. 
Would something like this do it?
`match.recruitment_cycle_year = RecruitmentCycle.current_year if match.recruitment_cycle_year.nil?`

duncanjbrown: If you pass a block to `find_or_create_by` or `find_or_initialize_by` it only runs if the record doesn't exist yet: https://apidock.com/rails/v4.0.2/ActiveRecord/Relation/find_or_create_by

duncanjbrown (approved):
Block comment optional!

PR #3171, opened 2020-10-17
Allow candidates to delete feedback_provided references

davidgisbey:
## Context

Candidates should be able to delete referees that are in the feedback_provided state. 

|Before|After|
|------------|------------|
|image|image|

## Link to Trello card

https://trello.com/c/T5mZ6gMS/2345-decoupled-references-should-we-allow-candidates-to-delete-feedback-provided-references

## Guidance to review

I'm wondering if this should be a soft delete so it can be reinstated later if needed. Any thoughts? Maybe it's a nice to have further down the line.

## Things to check

### DIFF
--- a/app/views/candidate_interface/decoupled_references/review/show.html.erb
+++ b/app/views/candidate_interface/decoupled_references/review/show.html.erb
@@ -17,7 +17,7 @@
   <div id='references_given'>
     <h2 class="govuk-heading-m">References that have been given</h2>
     <%= render(
-      CandidateInterface::DecoupledReferencesReviewComponent.new(references: @references_given, editable: false, show_history: true)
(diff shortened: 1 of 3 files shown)


tijmenb (commented):
Can you check if the deletion shows up in the support interface audit log? If that's the case deleting might be fine.

davidgisbey:
Good shout. All good in terms of it being captured in the logs.

image

tijmenb (approved):
Cool, this should be good then!

tijmenb (changes requested):
Wait, we should have (had) validation on this. Can we add a check to only allow deletes for certain states?

davidgisbey:
Cool, I'll add something in now.

tijmenb:
👍

PR #3180, opened 2020-10-19
Update copy for changed offer email notifications

benswannack:
- Update copy for emails so that new offer details are clear
- Commonise wait or respond section of emails

## Context

Current change offer emails do not make it clear what was changed in the offer. This change adds details to the notification to explain this better.

### Emails in question:
Before:
Before
After:
After

## Guidance to review

- Is `offered_course_option` the right property to use for the newly updated course option?
- There is a component already in the codebase for displaying course details, we'd have to modify it to use it here, but it would commonise content a little more.

## Link to Trello card

https://trello.com/c/mgQsZZXg/2855-update-e-mail-to-highlight-that-the-offer-changed-in-the-emails-sent-to-candidates

## Things to check


app/views/candidate_mailer/_wait_or_respond.erb
```diff
@@ -0,0 +1,7 @@
+^You can wait to hear back about your other application(s) before making a decision. When the last decision is in, you’ll have 10 working days to respond.
```
duncanjbrown, line 1: Stray `^`s in this file? I see they were in the original template but I don't think they do anything?

duncanjbrown: Maybe retain the `.text` part of the file extension for consistency?

app/views/candidate_mailer/changed_offer.text.erb
```diff
@@ -1,13 +1,50 @@
...
+<% if @is_awaiting_decision %>
+
+# You can accept the offer or wait to hear back about other applications
+
+<%= render "wait_or_respond" %>
+
+<% else %>
 
-You can sign in to Apply for teacher training to check the status of your application(s):
+# Make a decision by <%= @application_choice.decline_by_default_at&.to_s(:govuk_date) %>
```
duncanjbrown, line 26: I think there should be no need for an `&` here as all offered choices should have a `decline_by_default_at` and we'd want to fail loudly if one didn't — not least because this sentence wouldn't be complete!

app/views/candidate_mailer/changed_offer.text.erb
```diff
@@ -1,13 +1,50 @@
...
+
+If you do not reply by <%= @application_choice.decline_by_default_at&.to_s(:govuk_date) %>, your application will be withdrawn.
```
duncanjbrown, line 28: As above

app/views/candidate_mailer/changed_offer.text.erb
```diff
@@ -1,13 +1,50 @@
...
+
+<% if @offers.count > 1 %>
+
+# Offers you've received
```
duncanjbrown, line 32: Does indenting here work? Might be easier to follow

duncanjbrown (commented):
Looking good! Couple of qs inline

app/views/candidate_mailer/changed_offer.text.erb
```diff
@@ -1,13 +1,50 @@
...
 
-Contact <%= @course_option.course.provider.name %> directly if you have any questions about this.
+Your offer from <%= @course_option.course.provider.name %> to study <%= @course_option.course.name_and_code %> at <%= @course_option.site.name %> has been changed.
+
+<%= @offered_course_option.course.provider.name %> has offered you the following course:
+
+Provider: <%= @offered_course_option.course.provider.name %>
+Course: <%= @offered_course_option.course.name_and_code %>
+Location: <%= @offered_course_option.site.name %>
+Full time or part time: <%= @offered_course_option.course.study_mode.humanize %>
```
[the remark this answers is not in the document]
tvararu: I don't think we need to just yet, my 2c :+1:

tvararu (approved):
LGTM, thanks for partialising/componentising the shared parts :+1:

duncanjbrown (approved):

PR #3181, opened 2020-10-19
Patch flaky UCAS matching spec

malcolmbaig:
## Context

A flaky spec, and an attempt to fix it. The reasoning below is an educated guess based on the error message. Also, this solution could probably be smarter if I understood the feature and the underlying spec better. Thought this was worth a punt though.

This spec fails intermittently because:

- it explicitly creates candidates with hard-coded ids (to match with
  CSV test data that the spec relies on)
- it creates additional candidates via calls to FactoryBot. These
  presumably create candidate records with a primary key of whatever
  PostgreSQL's `nextval()` function returns. This will cause a conflict
  if the candidate table id sequence is just prior to any of the
  hard-coded ids at time of test execution.

Set the hard-coded ids to very high numbers (high enough that a test
suite run on a fresh test database is unlikely to ever reach sequence
numbers that high in the candidate table).

## Things to check


tijmenb (approved):
Nice! Let's hope this works.

PR #3182, opened 2020-10-19
Add errors array to ProviderAuthorisation for can_view_safeguarding_information?

mnacos:
## Context

In order to help provider users understand why they are not given access to safeguarding information, we need to know why the authorisation call has failed, not just that it has.

Add an `errors` array to `ProviderAuthorisation`. Make `can_view_safeguarding_information?` populate this with error symbols.

## Guidance to review

Is the approach reasonable?

## Link to Trello card

https://trello.com/c/gIujbsgQ

## Things to check

### DIFF
--- a/app/services/provider_authorisation.rb
+++ b/app/services/provider_authorisation.rb
@@ -1,6 +1,9 @@
 class ProviderAuthorisation
+  attr_reader :errors
+
   def initialize(actor:)
     @actor = actor
+    @errors = []
   end
 
   def providers_that_actor_can_manage_users_for
@@ -59,11 +62,21 @@ class ProviderAuthorisation
   end
 
   def can_view_safeguarding_information?(course:)
-    @actor.provider_permissions.view_safeguarding_information
-      .exists?(provider: [course.provider, course.accredited_provider].compact) &&
-      (course.accredited_provider.blank? ||
-        ratifying_provider_can_view_safeguarding_information?(course: course) ||
-          training_provider_can_view_safeguarding_information?(course: course))
+    errors << :requires_provider_user_permission unless
+      @actor.provider_permissions.view_safeguarding_information
+        .exists?(provider: [course.provider, course.accredited_provider].compact)
+
+    if course.accredited_provider.present?
+      if @actor.providers.include?(course.provider)
+        errors << :requires_training_provider_permission unless
+          training_provider_can_view_safeguarding_information?(course: course)
+      else
+        errors << :requires_ratifying_provider_permission unless
+          ratifying_provider_can_view_safeguarding_information?(course: course)
+      end
+    end
+
+    errors.blank?
   end
 
   def can_view_diversity_information?(course:)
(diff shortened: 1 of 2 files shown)


app/services/provider_authorisation.rb
```diff
@@ -59,11 +62,21 @@ def can_make_decisions?(application_choice:, course_option_id:)
...
+      if @actor.providers.include?(course.provider)
+        errors << :requires_training_provider_can_view_safeguarding_information unless
+          training_provider_can_view_safeguarding_information?(course: course)
+      else
+        errors << :requires_ratifying_provider_can_view_safeguarding_information unless
+          ratifying_provider_can_view_safeguarding_information?(course: course)
+      end
+    end
+
+    errors.blank?
```
duncanjbrown, line 79: Just like Rails validation :+1:

app/services/provider_authorisation.rb
```diff
@@ -1,6 +1,9 @@
 class ProviderAuthorisation
+  attr_accessor :errors
```
duncanjbrown, line 2: Maybe just `attr_reader`?

app/services/provider_authorisation.rb
```diff
@@ -59,11 +62,21 @@ def can_make_decisions?(application_choice:, course_option_id:)
   end
 
   def can_view_safeguarding_information?(course:)
-    @actor.provider_permissions.view_safeguarding_information
-      .exists?(provider: [course.provider, course.accredited_provider].compact) &&
-      (course.accredited_provider.blank? ||
-        ratifying_provider_can_view_safeguarding_information?(course: course) ||
-          training_provider_can_view_safeguarding_information?(course: course))
+    errors << :requires_view_safeguarding_information unless
```
duncanjbrown, line 65: In two minds about this. We usually check one permission at a time, right? So maybe instead of 

```
requires_view_safeguarding_information
requires_training_provider_can_view_safeguarding_information
requires_ratifying_provider_can_view_safeguarding_information
```

we could get away with

```
requires_provider_user_permission
requires_training_provider_permission
requires_ratifying_provider_permission
```

which is a bit less to chew on for clients? Wdyt?

duncanjbrown (commented):
Code looks great, one q about what we call the errors. And should we do this for everything while we're here? (Maybe that's a follow-up PR)

app/services/provider_authorisation.rb
mnacos: I thought matching the error symbol with the name of the permission is a good idea, but what you've suggested may be better, as it makes error symbol names the same across all different authorisation calls (e.g. `can_view_diversity_information?`). We can easily infer which _specific_ permission is missing because we know we are calling `can_view_safeguarding_information?`, for example.

mnacos:
> Code looks great, one q about what we call the errors. And should we do this for everything while we're here? (Maybe that's a follow-up PR)

I suspect we'll be adding more permissions-related conditional content soon for the other permissions. We could either do a separate PR standardising the style across all authorisation calls or just wait until the UI needs this information for other types of permission.

duncanjbrown (approved):

duncanjbrown:
Let's add more as we need them :+1:

PR #3185, opened 2020-10-19
Support: Redirect to applications page

paulrobertlloyd:
## Context

Addresses part of the issue uncovered in #3131. When you click on the link in the header, or visit `/support/` when signed in, currently taken to the lesser used Candidates view, which is also the second nav item under the first primary nav item, which is a bit weird.

This change simply redirects support interface path to Candidates → Applications (first primary nav item, first secondary nav item).

Updates redirects and affected tests.

## Things to check

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -704,7 +704,7 @@ Rails.application.routes.draw do
   end
 
   namespace :support_interface, path: '/support' do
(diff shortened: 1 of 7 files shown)


tijmenb (approved):
🤟

PR #3189, opened 2020-10-20
Fix back-links in decoupled references

tijmenb:
## Context

These pages are in the decoupled references flow, but link out to the old-style referees flow.

Remove the back links for now, because I don't know what the correct links are. @davidgisbey ?

## Link to Trello card

https://trello.com/c/ykZiAQ1s/2344-%F0%9F%92%94-decoupled-references-delete-feature-flag-and-clean-up-code

## Things to check

### DIFF
--- a/app/views/candidate_interface/referees/confirm_cancel.html.erb
+++ b/app/views/candidate_interface/referees/confirm_cancel.html.erb
@@ -1,5 +1,5 @@
 <% content_for :title, t('page_titles.referee_cancel') %>
-<% content_for :before_content, govuk_back_link_to(candidate_interface_review_referees_path, 'Back') %>
+<% content_for :before_content, govuk_back_link_to %>
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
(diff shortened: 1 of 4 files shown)


app/views/candidate_interface/referees/confirm_cancel.html.erb
```diff
@@ -1,5 +1,4 @@
 <% content_for :title, t('page_titles.referee_cancel') %>
-<% content_for :before_content, govuk_back_link_to(candidate_interface_review_referees_path, 'Back') %>
```
davidgisbey, line 2: I think this should still be in there right?

davidgisbey (approved):
I think that the referees review link should be there until we pull out the original referees code. I've added the correct links.

app/views/candidate_interface/referees/confirm_cancel.html.erb
tijmenb: Oh right, perhaps deleted by accident. Can't do any harm though.

PR #3194, opened 2020-10-20
Reference history spike

malcolmbaig:
## Context

A draft PR outlining the approach I have in mind for reference histories, intended for visibility and feedback.

- Add a reference_events table to capture event histories for each reference.
- Add a ReferenceEvent model with an enum defining event names we want to track.
- Use the captured events to populate the history row when reviewing references.

image

## Guidance to review
- Main motivation behind this approach is that the design calls for a full history for a reference, so repeat occurrences of events would need to be visible (eg - request->cancel->send again sequence would result in a two request_sent events for that reference).
- Does this approach seem sound? Are there any shortcomings to be aware of?
- Should we clear events for an application's references upon submit, to prevent the table getting too large? Or leave that for now?
- This approach would need to migrate the existing requested_at/reminder_sent_at timestamps that drive the very basic histories we have in master currently.

## Link to Trello card

https://trello.com/c/JAWa9Xzt
## Things to check


tijmenb:
Should we investigate using the audit log for this? That would automatically be backfilled, because we've been tracking changes since the start.

malcolmbaig:
> Should we investigate using the audit log for this? That would automatically be backfilled, because we've been tracking changes since the start.

That's a really good idea, every event we want to track will leave some kind of audit record. Will put a spike together to figure out what that looks like.

malcolmbaig:
Closing for now.

PR #3195, opened 2020-10-20
Fix auto approving references on sandbox

davidgisbey:
## Context

References auto approving on sandbox was inadvertently broken by the new decouple references flow.

- Move auto approving references from the SubmitApplication service to the RequestReference service
- Adds some tests for the RequestReference service

## Guidance to review

Test it locally by setting SANDBOX=false in your .env file and adding a reference.

## Link to Trello card

https://trello.com/c/6xR0U5M3/2342-%F0%9F%92%94-decoupled-references-fix-auto-approving-references-in-sandbox

## Things to check


app/services/candidate_interface/decoupled_references/request_reference.rb
```diff
@@ -1,15 +1,43 @@
...
+      def auto_approve_reference(reference)
+        reference.update!(
+          feedback_status: :feedback_provided,
+          relationship_correction: '',
+          safeguarding_concerns: '',
+          safeguarding_concerns_status: :no_safeguarding_concerns_to_declare,
+          feedback: I18n.t('new_referee_request.auto_approve_feedback'),
+        )
+
+        SubmitReference.new(
```
tijmenb, line 33: Should this use SubmitApplicationWithDecoupledReferences?

davidgisbey: I don't think so. It's not actually submitting the application at this stage. It's just auto-returning the reference request. With the new flow, they need to have received 2 references so submit their application.

app/services/candidate_interface/decoupled_references/request_reference.rb
```diff
@@ -1,15 +1,43 @@
...
         end
       end
+
+    private
+
+      def auto_approve_reference_in_sandbox(reference)
+        auto_approve_reference(reference) if HostingEnvironment.sandbox_mode? && email_address_is_a_bot?(reference)
+      end
+
+      def auto_approve_reference(reference)
```
duncanjbrown, line 24: This routine is almost identical to the one in `TestApplications`, though that one sometimes adds corrections and concerns. Should we pull out an `AutoApproveReference` service?

spec/services/candidate_interface/decoupled_references/request_reference_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'rails_helper'
+
+RSpec.describe CandidateInterface::DecoupledReferences::RequestReference do
+  describe '#call' do
+    let(:reference) { create(:reference, feedback_status: 'not_requested_yet') }
+    let(:execute_service) { described_class.new.call(reference, flash) }
```
duncanjbrown, line 6: Is it worth pulling a `let` for something that only happens once?

spec/services/candidate_interface/decoupled_references/request_reference_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    before do
+      allow(RefereeMailer).to receive(:reference_request_email).and_return(mail)
+      execute_service
+    end
+
+    it 'updates the references feeedback status to requested' do
+      expect(reference.feedback_status).to eq('feedback_requested')
+    end
+
+    it 'sends a refeence request email to the referee mailer' do
```
duncanjbrown, line 19: Typo: `refeence` missing an `r`

spec/services/candidate_interface/decoupled_references/request_reference_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'rails_helper'
+
+RSpec.describe CandidateInterface::DecoupledReferences::RequestReference do
+  describe '#call' do
```
duncanjbrown, line 4: There's a lot of mocking here, but we really do care about the side effects. & we're tending towards double-entry testing where the spec basically knows how the class is implemented and a small change over there (or to, for example `ActionMailer`) can mean changes over here too. I'd not bother to mock the mailer, for example — would prefer to check an email's sent and that it was the right one.

spec/services/candidate_interface/decoupled_references/request_reference_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    it 'updates the references feeedback status to requested' do
+      expect(reference.feedback_status).to eq('feedback_requested')
+    end
+
+    it 'sends a refeence request email to the referee mailer' do
+      expect(RefereeMailer).to have_received(:reference_request_email).with(reference)
+    end
+
+    context 'when running in a provider sandbox', sandbox: true do
```
duncanjbrown, line 23: That said, this spec's very good!

spec/services/candidate_interface/decoupled_references/request_reference_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'rails_helper'
+
+RSpec.describe CandidateInterface::DecoupledReferences::RequestReference do
+  describe '#call' do
+    let(:reference) { create(:reference, feedback_status: 'not_requested_yet') }
+    let(:execute_service) { described_class.new.call(reference, flash) }
+    let(:mail) { instance_double(ActionMailer::MessageDelivery, deliver_later: true) }
+    let(:flash) { ActionDispatch::Flash::FlashHash.new }
```
duncanjbrown, line 8: Depending on flash feels a bit weird to me — does it feel weird to you?

davidgisbey: Yea it definitely does. I've not come across a flash being passed into the service. What approach would you have taken here? Should I be injecting a double in?

spec/services/candidate_interface/decoupled_references/request_reference_spec.rb
davidgisbey: 100% agree. I was following a pattern, but this pattern definitely tests implementation, not behaviour.

app/services/candidate_interface/decoupled_references/request_reference.rb
tijmenb: Oh yeah facepalm.

app/services/candidate_interface/decoupled_references/request_reference.rb
tijmenb: We'll refactor this in a later PR.

tijmenb (approved):
👍 good to merge for now, we'll take another pass at the refactors suggested by @duncanjbrown once we've untangled some of the other references bits.

spec/services/candidate_interface/decoupled_references/request_reference_spec.rb
duncanjbrown: I think this service shouldn't depend on flash at all — only the controller should know about the flash, and it should use the return value of this thing to make decisions on what to do with the flash. Really hardly a problem here because it's such a tiny object! But in general it's good to avoid this kind of coupling from controllers to the service layer

PR #3200, opened 2020-10-21
Update FIND_BASE_URL

TobyRet:
## Context

The Teacher Training API url needs updating as a result of a domain change.

- update `FIND_BASE_URL` in `.env.development`
- add `FIND_BASE_URL` to `.env.example`

## Guidance to review

Questions for reviewers:

- Possibly a naming issue to consider. Should the Teaching Training API be referred to as `FIND_BASE_URL`? May potentially confuse now that the Find UI is now managed by Apply? Would a better alternative be `TEACHER_TRAINING_API_BASE_URL`?
- Removing `FIND_BASE_URL` from `.env.development` with break the app locally for devs. Best to leave in there?

## Things to check

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -1,3 +1,3 @@
-FIND_BASE_URL=https://api2.publish-teacher-training-courses.service.gov.uk/api/v3/
+FIND_BASE_URL=https://api.publish-teacher-training-courses.service.gov.uk/api/v3/
 HOSTING_ENVIRONMENT_NAME=development
 STATE_CHANGE_SLACK_URL=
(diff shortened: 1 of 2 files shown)


.env.development
```diff
@@ -1,3 +1,3 @@
-FIND_BASE_URL=https://api2.publish-teacher-training-courses.service.gov.uk/api/v3/
+FIND_BASE_URL=https://api.publish-teacher-training-courses.service.gov.uk/api/v3/
```
duncanjbrown, line 1: Don't the `.env` files stack up and override each other, and we could lose this here? https://github.com/bkeepers/dotenv#can-i-use-dotenv-in-production

duncanjbrown: Just tried removing this locally and it worked ok

TobyRet:
I've added a tech debt ticket to rename `FIND_BASE_URL` 
https://trello.com/c/5El83bB4/383-rename-findbaseurl

tvararu (approved):

PR #3202, opened 2020-10-21
Fix the final chaser email for overdue references

tijmenb:
## Context

The current system of selecting references to chase after 28 days doesn't work with the new decoupled references feature, because it only looks at applications that are "awaiting references", which is no longer a valid state.

Paired on this with @davidgisbey.

This is a full refactor of the chaser system. It merges a bunch of classes, and removes duplicated specs. Everything should be covered by the system spec.

## Guidance to review

The best way to review is to look at the final ChaseReference class and see if it makes sense.

## Link to Trello card

https://trello.com/c/ykZiAQ1s

## Things to check


spec/system/referee_interface/referee_does_not_respond_spec.rb
```diff
@@ -16,63 +15,72 @@
...
+  def given_there_is_an_application_with_a_reference
+    @reference = create(:reference, :requested, email_address: 'anne@other.com', name: 'Anne Other')
+    @application = create(:application_form, first_name: 'F', last_name: 'B', application_references: [@reference])
   end
 
   def and_the_referee_does_not_respond_within_7_days
     Timecop.travel(7.days.from_now) do
-      SendReferenceChaseEmailToBothPartiesWorker.perform_async
+      ChaseReferences.perform_async
+      ChaseReferences.perform_async
```
malcolmbaig, line 30: Is this to test only one email is sent? An explanatory comment might be useful here, to make it more obvious.

malcolmbaig (approved):
Nice one :+1:

PR #3205, opened 2020-10-21
Decoupled references - edit email address for failed request

stevehook:
## Context

As a candidate I need to edit the address on a reference request and resend so that I can re-request a reference that may have been sent to an incorrect or unreachable email address.

- [x] For failed reference requests (`email_bounced` status) allow the user to _Retry request_ when there are less than two other references provided.
- [x] The user is prompted to optionally change the email address before resending. Otherwise the original email address is used again.
- [x] Extend `ReferenceHistoryComponent` to include the recipient's email address for retry and send reference request operations.

image
image
image

## Guidance to review

- Should be testable via review app
- Is the conditional logic for the change link correct?
- I have not added the edit email address feature for applications in the cancelled state. Is this correct?

## Link to Trello card

https://trello.com/c/FqNl6q0u/2238-💔-dev-edit-email-address-of-failed-reference-request

## Things to check

### DIFF
--- a/app/components/candidate_interface/decoupled_references_review_component.html.erb
+++ b/app/components/candidate_interface/decoupled_references_review_component.html.erb
@@ -1,5 +1,9 @@
 <% references.each do |reference| %>
-  <%= render(SummaryCardComponent.new(rows: reference_rows(reference), editable: editable && reference.editable?, ignore_editable: ['History'])) do %>
+  <%= render(SummaryCardComponent.new(
+    rows: reference_rows(reference),
+    editable: editable && reference.editable?,
+    ignore_editable: ignore_editable_for(reference),
+  )) do %>
     <%= render(SummaryCardHeaderComponent.new(title: card_title(reference))) do %>
       <div class="app-summary-card__actions">
         <ul class="app-summary-card__actions-list">
--- a/app/components/candidate_interface/decoupled_references_review_component.html.erb
+++ b/app/components/candidate_interface/decoupled_references_review_component.html.erb
@@ -46,6 +46,13 @@
                 <span class="govuk-visually-hidden"> to <%= reference.name %></span>
               <% end %>
             </li>
+          <% elsif can_retry?(reference) %>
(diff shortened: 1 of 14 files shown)


tijmenb (commented):
I'm a little unsure about the flow.

If the candidate changes the email of a bounced reference, they will have a odd UI: it will show the new email address, but it looks like it will show "Request failed" as the status and it isn't clear that they need to resend the request. Resending requests is limited to once, so if already retried (as is probable) it won't work.

I was expecting something like: if the email bounces, the user can update the email addres, which sends a new request email, updates the `requested_at` time (for chasers) and updates the status. 

Also, it looks like this only shows the edit email address link. If think it means that we actually allow editing of references that have been sent (if you know the URL).

stevehook:
> I'm a little unsure about the flow.
> 
> If the candidate changes the email of a bounced reference, they will have a odd UI: it will show the new email address, but it looks like it will show "Request failed" as the status and it isn't clear that they need to resend the request. Resending requests is limited to once, so if already retried (as is probable) it won't work.
> 
> I was expecting something like: if the email bounces, the user can update the email addres, which sends a new request email, updates the `requested_at` time (for chasers) and updates the status.
> 
> Also, it looks like this only shows the edit email address link. If think it means that we actually allow editing of references that have been sent (if you know the URL).

I think I agree about the flow. It would be better to have the email address update and re-request happen in the same transaction. I might have misunderstood the design (I don't think this part is in the prototype - understandably). If we drop the _Change_ link and prompt the candidate to change the email address what is currently the re-request confirmation form would that make sense? The only objection I can think of is that the candidate might not attempt to re-request because they may not realise that they can change the email address as part of that process. What do you think @paulrobertlloyd ?

I think we already allow editing email addresses that have been sent. I tried this locally and was able to do so. Have logged a separate card to address this - https://trello.com/c/TPXBtAdm/2385-dont-allow-edits-to-references-that-have-already-been-requested

stevehook:
Closing this to create a new PR partly to recreate the review app

PR #3207, opened 2020-10-21
Move 404 catch-all route to application.rb

tvararu:
## Context

When the route is defined in `routes.rb`, it overrides any custom routes that are defined in engines from 3rd party libraries.

This breaks in the newest version of ViewComponent, which has switched from `.prepend` to `.append` for routes: https://github.com/github/view_component/pull/493/files#diff-fe26dc052312860cc076b99a7f3c0913bf834fc73fe246def378024ffc038164R91

Move 404 catch-all route to application.rb, so that it can kick-in after engines have defined their custom routes. This is recommended in the README for `vidibus-routing_error`: https://github.com/vidibus/vidibus-routing_error#alternative-two

## Guidance to review

Does this make sense?

## Link to Trello card

https://trello.com/c/V7FqNNN3/2380-re-enable-accessibility-end-to-end-tests

## Things to check


mnacos (approved):
sweet, I need these previews

davidgisbey (approved):

PR #3208, opened 2020-10-21
Remove RejectAwaitingReferencesCourseChoicesWorker

tijmenb:
## Context

This worker and associated code was used to "cancel" applications and references at the end of last cycle. From this cycle on we have a new system for "decoupled" references, which means that we won't have to cancel these applications next year.

Remove all code.

## Guidance to review

Should I have removed more?

## Link to Trello card

https://trello.com/c/ykZiAQ1s

## Things to check


davidgisbey (approved):
Yea this makes sense 👍

PR #3214, opened 2020-10-21
Change references factory traits to be in line with the feedback status enum

davidgisbey:
## Context

At the moment, the traits in the references factory don't align the feedback_status enum in the application reference model. 

Ideally, these should align to allow for things like creating and iterating across each state in testing.

It's also just more intuitive.

Update the traits in the references factory to align with the feedback status enum.

## Guidance to review

It might be easier to check per commit as I've broken it down by trait.

## Things to check


tvararu (approved):
image

PR #3215, opened 2020-10-21
Redirect to /candidate/account from sandbox homepage

benswannack:
/candidate no longer exists within the apply service, so linking there causes us to forget we want to be in sandbox mode

## Context
Clicking the Continue as Candidate button currently results in being redirected away from sandbox to the live gov.uk site, and then clicking start redirects to the live apply service

Specify /candidate/account so that we stay on the sandbox environment

## Link to Trello card
https://trello.com/c/mDekNsqF/2948-sandbox-continue-as-candidate-button-redirects-through-to-live-govuk-site

## Things to check


paulrobertlloyd:
Should this use a rails path and not a hard coded value?

duncanjbrown:
> Should this use a rails path and not a hard coded value?

Yes, and so should the `/provider` link — let's fix this while we're here! I think there's a case for leaving the API docs link hardcoded to production as that might avoid vendor confusion ("I was looking at the sandbox docs but maybe production is different?").

TobyRet (dismissed):

tvararu (approved):
LGTM assuming the hard-coded /api-docs URL is good

PR #3217, opened 2020-10-21
Add super_diff gem for better hash comparisons in RSpec

duncanjbrown:
## Context

I discovered a gem which makes hash comparisons much more readable in RSpec output. Proposing we include it in the `Gemfile`.

### Before `super_diff`

Screenshot 2020-10-21 at 20 33 30

### After `super_diff`

Screenshot 2020-10-21 at 20 33 20

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -98,6 +98,7 @@ group :test do
   gem 'clockwork-test'
   gem 'deepsort'
   gem 'ruby-jmeter'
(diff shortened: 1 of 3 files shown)


TobyRet (approved):
Good idea :-)

malcolmbaig (approved):
:star:

tvararu (approved):

PR #3218, opened 2020-10-21
Add API fields to support Tribal

duncanjbrown:
## Context

Student Record System vendor Tribal need some extra data to support various provider workflows in their software so we've agreed to add some fields to the API.

- `ApplicationAttributes.recruited_at`
- `Offer.offer_made_at`
- `Offer.accepted_at`
- `Offer.declined_at`
- `Reference.referee_type`
- `Reference.safeguarding_concerns`

Each of these fields should be

- added to the OpenAPI YAML file
- added to `SingleApplicationPresenter` which renders the API responses, plus its spec
- mentioned in the release notes

## Guidance to review

Do the docs make sense? Are the dates correct? Are the release notes complete?

## Link to Trello card

https://trello.com/c/kRGZanSy/2867-add-api-fields-to-support-tribal

## Things to check


config/vendor-api-v1.yml
```diff
@@ -679,6 +689,23 @@ components:
           - Completion of professional skills test
         course:
           "$ref": "#/components/schemas/Course"
+        offer_made_at:
+          type: string
+          format: date-time
+          description: Date and time the offer was made
+          example: 2019-09-18T15:33:49.216Z
+        accepted_at:
```
benswannack, line 697: Is it worth prefixing this and `declined_at` with `offer_` for consistency?

benswannack (commented):
Looks good!

config/vendor-api-v1.yml
duncanjbrown: Yeah, I wondered about that! I think you're right — will update

duncanjbrown:
Updated per @benswannack's suggestion and added a field we missed first time, `Qualification.non_uk_qualification_type`

mnacos (approved):
LGTM - tests need to be updated with the current state of references.

spec/support/test_helpers/candidate_helper.rb
```diff
@@ -27,6 +27,7 @@ def candidate_completes_application_form(with_referees: true)
     FeatureFlag.deactivate(:international_personal_details)
     FeatureFlag.deactivate(:efl_section)
     FeatureFlag.deactivate(:international_degrees)
+    FeatureFlag.deactivate(:international_other_qualifications)
```
benswannack, line 30: Is this acceptable? If this feature flag is active, then in the tests, for the "other" qualification, we get a blank string as the `non_uk_qualification_type` rather than `nil` in the API response

PR #3224, opened 2020-10-22
Expand Reference History

malcolmbaig:
## Context
**As a...** candidate
**I need to...** know what has happened to my reference request
**So that...** so I can follow up with my referee or when contacting support.

- Add a `ReferenceHistory` model. Wraps a single reference instance and returns a history of relevant events derived from the audit log.
- Use this model's output in `ReferenceHistoryComponent`.
Screenshot_20201022_103330

## Guidance to review
- Per-commit recommended.
- Can be tested in the review app by adding a reference and clicking various actions (request, remind, cancel, etc).
- The current implementation of the reference history isn't particularly efficient (multiple iterations across the audit array with a single call to `all_events`). I think this is fine given the low number of audits we're typically dealing with per reference, and the fact that a non-optimised implementation is probably easier to follow. WDYT?

## Link to Trello card

https://trello.com/c/JAWa9Xzt

## Things to check

### DIFF
--- a/app/components/reference_history_component.html.erb
+++ b/app/components/reference_history_component.html.erb
@@ -1,8 +1,8 @@
-<ul class='govuk-list qa-reference-history'>
+<ul class="govuk-list" data-qa="reference-history">
   <% history.each do |event| %>
     <li>
       <%= formatted_title event %>
-      <span class='govuk-hint govuk-!-margin-bottom-0 govuk-!-font-size-16'><%= formatted_date event.time %></span>
+      <span class="govuk-hint govuk-!-margin-bottom-0 govuk-!-font-size-16"><%= formatted_date event.time %></span>
     </li>
   <% end %>
 </ul>
--- a/app/components/reference_history_component.html.erb
+++ b/app/components/reference_history_component.html.erb
@@ -1,15 +1,8 @@
 <ul class='govuk-list qa-reference-history'>
-  <% if requested_at %>
+  <% history.each do |event| %>
     <li>
-      Request sent
-      <span class='govuk-hint govuk-!-margin-bottom-0 govuk-!-font-size-16'><%= requested_at %></span>
-    </li>
-  <% end %>
-
-  <% if reminder_sent_at %>
-    <li>
-      Reminder sent
-      <span class='govuk-hint govuk-!-margin-bottom-0 govuk-!-font-size-16'><%= reminder_sent_at %></span>
+      <%= formatted_title event %>
+      <span class='govuk-hint govuk-!-margin-bottom-0 govuk-!-font-size-16'><%= formatted_date event.time %></span>
     </li>
   <% end %>
 </ul>
--- a/app/models/reference_history.rb
+++ b/app/models/reference_history.rb
@@ -59,7 +59,7 @@ class ReferenceHistory
 private
 
   def audits
-    reference.audits.updates
+    @audits ||= reference.audits.updates
   end
 
   def status_change(audit, to:)
--- a/app/models/reference_history.rb
+++ b/app/models/reference_history.rb
@@ -25,7 +25,7 @@ class ReferenceHistory
 
   def request_cancelled
     audits
-      .select { |a| status_change(a, to: 'cancelled') }
+      .select { |a| status_change(a, to: 'cancelled') || status_change(a, to: 'cancelled_at_end_of_cycle') }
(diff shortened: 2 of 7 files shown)


paulrobertlloyd:
Looking at Min Blanda’s application on the review app, and it’s not showing the initial reference request in the history:

Screenshot 2020-10-22 at 11 08 29

malcolmbaig:
> Looking at Min Blanda’s application on the review app, and it’s not showing the initial reference request in the history:
> 

As discussed on Slack, this is caused by how we're creating references in the test data generation. We create them in a single step in the `feedback_requested` state. This means we don't generate the audit history required to show the request sent event. Not something you'd see in normal use of the system.

I've raised a PR to fix: https://github.com/DFE-Digital/apply-for-teacher-training/pull/3226

app/components/reference_history_component.html.erb
```diff
@@ -1,15 +1,8 @@
...
-      <span class='govuk-hint govuk-!-margin-bottom-0 govuk-!-font-size-16'><%= requested_at %></span>
-    </li>
-  <% end %>
-
-  <% if reminder_sent_at %>
-    <li>
-      Reminder sent
-      <span class='govuk-hint govuk-!-margin-bottom-0 govuk-!-font-size-16'><%= reminder_sent_at %></span>
+      <%= formatted_title event %>
+      <span class='govuk-hint govuk-!-margin-bottom-0 govuk-!-font-size-16'><%= formatted_date event.time %></span>
```
paulrobertlloyd, line 5:
```suggestion
      <span class="govuk-hint govuk-!-margin-bottom-0 govuk-!-font-size-16"><%= formatted_date event.time %></span>
```

app/components/reference_history_component.html.erb
```diff
@@ -1,15 +1,8 @@
 <ul class='govuk-list qa-reference-history'>
```
paulrobertlloyd, line 1:
```suggestion
<ul class="govuk-list" data-qa="reference-history">
```

spec/system/candidate_interface/decoupled_references/candidate_views_reference_history_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+  end
+
+  def then_i_see_a_history_of_these_events_on_the_review_page
+    expect(page).to have_content 'History'
+    expected_history = [
+      { event_name: 'Request sent', timestamp: '1 January 2020 at 2:00pm' },
+      { event_name: 'Reminder sent', timestamp: '2 January 2020 at 2:00pm' },
+      { event_name: 'Request cancelled', timestamp: '3 January 2020 at 2:00pm' },
+    ]
+    within '.qa-reference-history' do
```
paulrobertlloyd, line 60: Not sure if this is the right selector, but should work…

```suggestion
    within '[data-qa="reference-history"]' do
```

paulrobertlloyd (changes requested):

app/models/reference_history.rb
```diff
@@ -0,0 +1,68 @@
...
+  def reference_given
+    audits
+      .select { |a| status_change(a, to: 'feedback_provided') }
+      .map { |a| Event.new('reference_given', a.created_at) }
+  end
+
+private
+
+  def audits
+    reference.audits.updates
```
stevehook, line 62: Should we memoize this? Testing in rails console `updates` seems to hit the database on each repeated call.

app/models/reference_history.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  def request_sent
+    audits
+      .select { |a| status_change(a, to: 'feedback_requested') }
+      .map { |a| Event.new('request_sent', a.created_at) }
+  end
+
+  def request_cancelled
+    audits
+      .select { |a| status_change(a, to: 'cancelled') }
```
stevehook, line 28: Should we include the `cancelled_at_end_of_cycle` status here?

malcolmbaig: Good question. @paulrobertlloyd - should we have a 'cancelled at end of cycle' event in the reference history?

paulrobertlloyd: I think so. The request status should also say cancelled (and the reason why, although that’s currently missing, pending design/content), but I don’t think it hurts to repeat this in the history.

stevehook (approved):

paulrobertlloyd (approved):

PR #3237, opened 2020-10-23
Terms of service checkbox bug

raamSoftwire:
## Context

The terms of service checkbox doesn't persist answer if there is a validation error with the email address on the sign up page.

Edit the sign up form template to read in the `@sign_up_form` object for the checkbox value.

## Guidance to review

To test this manually:

- Create new account on staging
- Navigate to terms of service agreement page
- Check checkbox, but omit email address
- Submit page
- Get returned to page without checkbox persisted.

## Link to Trello card

https://trello.com/c/zWvC0NDW/2035-bug-terms-of-service-checkbox-doesnt-persist-answer-if-there-is-a-validation-error-with-the-email-address

## Things to check

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,6 @@ spec/system/ucas_matching/matching_data_example.zip
 # jmeter stuff
 ruby-jmeter.jmx
 jmeter.log
+
+# RubyMine files
+.idea
--- a/app/views/candidate_interface/sign_up/new.html.erb
+++ b/app/views/candidate_interface/sign_up/new.html.erb
@@ -29,7 +29,7 @@
           <li>checking you’re safe to work with children</li>
         </ul>
 
-        <%= f.govuk_check_box :accept_ts_and_cs, true, multiple: false, link_errors: true, label: { text: t('authentication.sign_up.accept_terms_checkbox') } %>
+        <%= f.govuk_check_box :accept_ts_and_cs, 'true', multiple: false, link_errors: true, label: { text: t('authentication.sign_up.accept_terms_checkbox') } %>
       <% end %>
       <%= f.govuk_submit t('authentication.sign_up.button_continue') %>
     <% end %>
--- a/spec/system/candidate_interface/candidate_account_spec.rb
+++ b/spec/system/candidate_interface/candidate_account_spec.rb
@@ -14,6 +14,12 @@ RSpec.feature 'Candidate account' do
 
     when_i_visit_the_signup_page
     and_i_accept_the_ts_and_cs
+    and_i_submit_without_entering_an_email
+    then_i_see_form_errors_on_the_page
+    and_the_ts_and_cs_are_still_checked
+
+    when_i_visit_the_signup_page
+    and_i_accept_the_ts_and_cs
     and_i_submit_my_email_address
     then_i_receive_an_email_with_a_signup_link
     when_i_click_on_the_link_in_my_email
@@ -75,6 +81,14 @@ RSpec.feature 'Candidate account' do
   def and_i_submit_my_email_address(email = @email)
     fill_in t('authentication.sign_up.email_address.label'), with: email
     click_on t('authentication.sign_up.button_continue')
+  end
+
+  def and_i_submit_without_entering_an_email
+    click_on t('authentication.sign_up.button_continue')
+  end
+
+  def and_the_ts_and_cs_are_still_checked
+    expect(page).to have_checked_field t('authentication.sign_up.accept_terms_checkbox')
   end
 
   def and_i_submit_my_email_address_in_uppercase


.gitignore
```diff
@@ -59,3 +59,6 @@ spec/system/ucas_matching/matching_data_example.zip
 # jmeter stuff
 ruby-jmeter.jmx
 jmeter.log
+
+# RubyMine files
+.idea
```
raamSoftwire, line 64: I'm not sure how you feel about this, so I've included it here for discussion. 

I spent some time trying to configure a global excludes file, as discussed in the comment on line 5 of the gitignore but I was unsuccessful. This is a relatively low impact solution and will mean that any other team members in future also won't have to worry about it or spend time on it.

Equally, I'm happy to remove this change and pair with someone to figure out how to get it set up correctly. And then update the docs for future devs.

app/views/candidate_interface/sign_up/new.html.erb
```diff
@@ -29,7 +29,7 @@
           <li>checking you’re safe to work with children</li>
         </ul>
 
-        <%= f.govuk_check_box :accept_ts_and_cs, true, multiple: false, link_errors: true, label: { text: t('authentication.sign_up.accept_terms_checkbox') } %>
+        <%= f.govuk_check_box :accept_ts_and_cs, @sign_up_form.accept_ts_and_cs, multiple: false, link_errors: true, label: { text: t('authentication.sign_up.accept_terms_checkbox') } %>
```
raamSoftwire, line 32: This solves the bug, but I'm not convinced it is the correctly solution. I've had a look through the (docs)[https://www.rubydoc.info/gems/govuk_design_system_formbuilder/GOVUKDesignSystemFormBuilder/Builder#govuk_check_box-instance_method] which suggests that this is the attribute for controlling the checkbox value. 

But having a look at other checkboxes in the codebase, we haven't implemented similar changes elsewhere. The second attribute is left as `true` rather than taking in the property on the object.

tvararu: `'true'` seems to work just as well

tvararu (commented):
Discussed offline, LGTM once we add a system spec :+1:

app/views/candidate_interface/sign_up/new.html.erb
raamSoftwire: `'true'` does work on the page but I ran in to some difficulty when writing the tests.

Both `true` and `'true'` add a `value='true'` attribute to the input element in both the unchecked and checked states, which meant that you couldn't distinguish between the states for the test. I don't really understand how it is working on the page.

`@sign_up_form.accept_ts_and_cs` only adds `value='on'` in the checked state, which is means we can test it easily and it is easier to reason about how it is working.

spec/system/candidate_interface/candidate_account_spec.rb
```diff
@@ -77,6 +83,14 @@ def and_i_submit_my_email_address(email = @email)
     click_on t('authentication.sign_up.button_continue')
   end
 
+  def and_i_submit_without_entering_an_email
+    click_on t('authentication.sign_up.button_continue')
+  end
+
+  def and_the_ts_and_cs_are_still_checked
+    expect(find_field(t('authentication.sign_up.accept_terms_checkbox')).value).to eq 'on'
```
tijmenb, line 91: Capybara has a cool helper for this:

```suggestion
    expect(page).to have_checked_field t('authentication.sign_up.accept_terms_checkbox')
```

app/views/candidate_interface/sign_up/new.html.erb
tijmenb: Took me a bit to understand this! 

For my understanding: what the form builder does is compare `@sign_up_form.accept_ts_and_cs` with whatever is the second argument to `govuk_check_box`.

- When we use the boolean `true`, the boolean is cast to the string `"true"` as the `value=`. This means that the for compares `true == "true"`
- When we use `"true"` or _any_ other string, it will work because we compare the the correct things
- When we use  `@sign_up_form.accept_ts_and_cs`, this is initially nil, so no `value=` is provided. When submitting we then submit the string "`on`", which is compared correctly again

tijmenb (approved):

spec/system/candidate_interface/candidate_account_spec.rb
raamSoftwire: Thanks, I've added that in now

raamSoftwire:
Thanks @tijmenb for explaining what is going on and for raising the issue in the form builder repo. I've tidied up my commits and tested it all and so I think this is now ready to merge in.

tijmenb (approved):
👍 Nice, thanks!

PR #3240, opened 2020-10-23
Remove the decoupled references feature flag

davidgisbey:
## Context

This is the second to last PR for the decoupled references clean up.

It fixes/removes the last few specs that have the flag deactivated, then removes the flag.

- Fix/remove the last few specs where the flag is deactivated  
- Removes the flag from the feature flag file
 - Removes all the conditional logic and strips code that is unused when the flag is on
- Removes all the `FeatureFlag.activate`'s from the tests

## Guidance to review

Probably easiest to review by commit.

The last thing I can think of that needs to be done is delete unused i18n strings and move the ones being used that are used in the `referees:` section of the application_form.yml. file to the ` references:` section.

We can also rename everything from decoupled_references to references.

Have I missed anything?

## Link to Trello card

https://trello.com/c/ykZiAQ1s/2344-%F0%9F%92%94-decoupled-references-delete-feature-flag-and-clean-up-code

## Things to check


app/models/application_reference.rb
```diff
@@ -3,14 +3,6 @@ class ApplicationReference < ApplicationRecord
...
   self.table_name = 'references'
 
-  validates :name, presence: true, length: { minimum: 2, maximum: 200 }, unless: -> { FeatureFlag.active?('decoupled_references') }
-  validates :email_address, presence: true,
-                            email_address: true,
-                            length: { maximum: 100 },
-                            uniqueness: { case_sensitive: false, scope: :application_form_id },
-                            unless: -> { FeatureFlag.active?('decoupled_references') }
-  validate :email_address_not_own
-  validates :relationship, presence: true, word_count: { maximum: 50 }, unless: -> { FeatureFlag.active?('decoupled_references') }
```
tvararu, line 13: I think there are some orphaned validation messages you can also remove from the yml locale files.

tvararu (approved):

app/models/application_reference.rb
davidgisbey: Will do! Thanks for all the reviews today 👍

PR #3250, opened 2020-10-26
Remove code that sends applications to providers when references are complete

tijmenb:
## Context

Candidates now send the applications themselves once they have 2 references. This code is no longer being used.

🔥

## Guidance to review

Did I remove everything? Note that the `SendApplicationToProvider` class is used in the new code path, so should not go.

## Link to Trello card

https://trello.com/c/ykZiAQ1s


davidgisbey (approved):
lgtm

PR #3258, opened 2020-10-26
Fail Rspec test task if rspec process has errored.

vigneshmsft:
Rspec test results xml will be empty if tests aren't run due to code errors.
<testsuite name="rspec" tests="0" skipped="0" failures="0" errors="0">

## Context

Rspec test run always exists with error code 0, and the test results are arrived by parsing for number of test failures in the xml files. 
So even in the case where rspec could fail because of a missing file the process exits with error code 0 but this case is not handled in the script which parses the rspec results xml file.

https://ukgovernmentdfe.slack.com/archives/CN1MCQCHZ/p1603725811094000

Fail the rspec task if `tests="0"` in the rspec xml output.

## Things to check

### DIFF
--- a/azure/pipelines/templates/run-rspec-test.yml
+++ b/azure/pipelines/templates/run-rspec-test.yml
@@ -17,6 +17,12 @@ steps:
         then
           seedvalue=$(grep -Pio 'name="seed" value="\d+"' $testResultFile)
           echo "##[command]$seedvalue"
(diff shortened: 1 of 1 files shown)


tijmenb (approved):
This looks like it fixes the issue, so 👍 

However - I think the root cause of the issue if that we have a lot of complexity in the build pipeline, and adding another `if` statement might only make it more so. I wonder if we could remove some of this code - it feels like it should be something that could be handled natively by Azure?

PR #3265, opened 2020-10-27
Temporary fix for UCAS matching page crashing

tijmenb:
## Context

We can't figure out why the index page is crashing: https://sentry.io/organizations/dfe-bat/issues/1937345465

Remove the offending code so the list will be visible, and we can find out why this is happening.

## Link to Trello card


steventux (approved):
:+1:

PR #3267, opened 2020-10-27
Increase OkComputer sidekiq retries threshold to 50

mnacos:
## Context

We already have queue latency checks on `default` and `mailers` queues, so this sidekiq retries check will mostly detect high number of errors when pulling data from other services (Find, UCAS) (`low_priority` queue). We also used to have a single Sync with Find task, but now we trigger parallel per-provider syncs instead of one big task, so exceeding the limit has become very easy even with the tiniest network connectivity glitches. This causes too much noice. If there is something seriously wrong with Find Sync we have another check for that, too.

Increase the threshold from 10 retries to 50. Production reached ~70 retries during the recent Postgresql maintanance window/outage.

## Guidance to review

Reasonable?

## Link to Trello card

No Trello card

## Things to check

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -15,7 +15,7 @@ class SidekiqRetriesCheck < OkComputer::Check
   def check
     retries_queue_length = Sidekiq::RetrySet.new.size
     queue_length_text = " (#{retries_queue_length})"
-    if retries_queue_length > 10
+    if retries_queue_length > 50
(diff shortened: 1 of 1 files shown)


vigneshmsft (approved):
LGTM.
considering that we have many tasks running in the background, we should hit the 50 number soon enough (in case of issues) to indicate problems with the queue. 
But number is high enough to avoid false availability alerts.

PR #3269, opened 2020-10-28
Refactor course availability logic in the course model 

davidgisbey:
## Context

There are two methods in the course model which deal with course option availability.

The `both_study_modes_available?` method checks if a course's study mode is `:full_time_or_part_time` and the `available_study_modes_from_options` which plucks it's associated course options study modes and uniques them.

This causes some undesirable behaviour. When the SyncProviderFromFind job runs, it invalidates course_options whose site has been deleted on find if they are not associated with an application. 

If they are associated with an application it sets the site_still_valid boolean on the course option to false.

With the current implementation, we do not check if the course options have been invalidated by find in the  `available_study_modes_from_options` method.

We also don't check to see if there are valid course_options for both study modes before passing them onto the select study mode page in the add course flow.

Change the `available_study_modes_from_options` so it rejects course options which have `site_still_valid: false` before performing unique.

Change `both_study_modes_available?` so it only passes if `available_study_modes_from_options.count == 2`

## Guidance to review

Does this make sense? I don't think that for the most part using `study_mode == 'full_time_or_part_time'` makes sense for the add course flow, as it will continue to show invalidated course options. 

## Link to Trello card

https://trello.com/c/kxl26vEI/2416-course-availability-logic-is-confusing

## Things to check

### DIFF
--- a/spec/models/course_spec.rb
+++ b/spec/models/course_spec.rb
@@ -8,17 +8,23 @@ RSpec.describe Course, type: :model do
     it { is_expected.to validate_uniqueness_of(:code).scoped_to(%i[recruitment_cycle_year provider_id]) }
   end
 
-  describe '#both_study_modes_available?' do
-    it 'is true when the study_mode value indicates both modes are available' do
-      course = build_stubbed(:course, study_mode: 'full_time_or_part_time')
-      expect(course.both_study_modes_available?).to be true
+  describe '#currently_has_both_study_modes_available?' do
+    it 'is true when a course has full time and part time course options' do
+      course_option1 = build_stubbed(:course_option, :full_time)
+      course_option2 = build_stubbed(:course_option, :part_time)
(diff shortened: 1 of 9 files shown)


spec/models/course_spec.rb
```diff
@@ -8,17 +8,42 @@
     it { is_expected.to validate_uniqueness_of(:code).scoped_to(%i[recruitment_cycle_year provider_id]) }
   end
 
-  describe '#both_study_modes_available?' do
-    it 'is true when the study_mode value indicates both modes are available' do
-      course = build_stubbed(:course, study_mode: 'full_time_or_part_time')
-      expect(course.both_study_modes_available?).to be true
+  describe '#currently_has_both_study_modes_available?' do
+    it 'is true when the a course has full time and part time course options' do
```
malcolmbaig, line 12: the/a

malcolmbaig (approved):
This makes sense to me. :+1:

PR #3274, opened 2020-10-28
Track validation errors for decoupled references

davidgisbey:
## Context

The rest of the application form tracks validation errors. 

This should have been built into the flow (my fault).

- Adds validation error tracking to all controllers that use Forms
- Updates the system spec to show they're being tracked

## Link to Trello card

https://trello.com/c/mIUgBACD/2377-%F0%9F%92%94-decoupled-references-theres-no-validation-error-tracking-for-the-decoupled-references-flow

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/decoupled_references/candidate_name_controller.rb
+++ b/app/controllers/candidate_interface/decoupled_references/candidate_name_controller.rb
@@ -13,12 +13,15 @@ module CandidateInterface
           first_name: first_name_param,
           last_name: last_name_param,
         )
-        return render :new unless @reference_candidate_name_form.valid?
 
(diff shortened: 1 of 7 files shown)


tijmenb (approved):
This looks good, but I wonder if we need to simplify the tests for the validation errors. Each `and_a_validation_error_is_logged_for_type` is similar, and test the internals of the validation error system. I wonder if we can test less here - maybe just that a `ValidationError` has been created?

davidgisbey:
Cool. I did think about just testing the count was increasing so I'll do that 👍 We know the correct one is being called anyway as the message being displayed is already being tested.

PR #3277, opened 2020-10-28
Add invariant check for weird reference state

malcolmbaig:
## Context
An important property of the system is that submitted application forms
not have any outstanding reference requests. Such a state could lead to
unexpected reference chase emails being sent, or a bad user experience
if a referee tries to provide a reference when it's no longer needed.

This invariant check looks for this state in the database and sends a
message to Sentry with the ids of any applications it finds.

## Guidance to review
Does the SQL query look right?

## Link to Trello card
https://trello.com/c/qSOCkAhQ

## Things to check

### DIFF
--- a/app/workers/detect_invariants.rb
+++ b/app/workers/detect_invariants.rb
@@ -4,6 +4,7 @@ class DetectInvariants
 
   def perform
     detect_application_choices_in_old_states
+    detect_outstanding_references_on_submitted_applications
   end
 
   def detect_application_choices_in_old_states
@@ -17,6 +18,24 @@ class DetectInvariants
         `application_complete` state, but all these states have been removed:
 
         #{choices_in_wrong_state.map(&:id).join("\n")}
+      MSG
+
+      Raven.capture_exception(WeirdSituationDetected.new(message))
+    end
+  end
+
+  def detect_outstanding_references_on_submitted_applications
+    applications_with_reference_weirdness = ApplicationChoice
+      .joins(application_form: [:application_references])
+      .where.not(application_choices: { status: 'unsubmitted' })
+      .where(references: { feedback_status: :feedback_requested })
+      .pluck(:application_form_id).uniq
+
+    if applications_with_reference_weirdness.any?
+      message = <<~MSG
+        One or more references are still pending on these applications,
+        even though they've already been submitted:
+        #{applications_with_reference_weirdness.join("\n")}
       MSG
 
       Raven.capture_exception(WeirdSituationDetected.new(message))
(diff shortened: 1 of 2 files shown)


app/workers/detect_invariants.rb
```diff
@@ -23,5 +24,23 @@ def detect_application_choices_in_old_states
     end
   end
 
+  def detect_outstanding_references_on_submitted_applications
+    applications_with_reference_weirdness = ApplicationChoice
+      .joins(application_form: [:application_references])
+      .where.not(application_choices: { status: 'unsubmitted' })
```
stevehook, line 30: Are there any other pre-submission statuses we need to consider? I know we've got rid of most of these but does `application_not_sent` count as un-submitted?

In other places we test for the presence of `ApplicationForm#submitted_at`. Would that work?

app/workers/detect_invariants.rb
```diff
@@ -23,5 +24,23 @@ def detect_application_choices_in_old_states
...
+      .where(references: { feedback_status: :feedback_requested })
```
stevehook, line 31: Now that candidates can add more than two `feedback_requested` references before submission is it possible for a submitted application to have two `feedback_provided` and one or more `feedback_requested` references?

(I couldn't see any logic for cancelling 'spare' pending references in the submit application logic but I might be missing it).

stevehook (commented):
I had a couple of questions inline...

app/workers/detect_invariants.rb
malcolmbaig: This raises an interesting question actually - should we cancel outstanding references if an application gets [rejected at end of cycle](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/services/candidate_interface/cancel_unsubmitted_application_at_end_of_cycle.rb) (and moves to an `application_not_sent` state)? I think we probably should. @tijmenb - WDYT?

A consequence of that is that this query remains unchanged, ie - "only unsubmitted applications should have pending references".

app/workers/detect_invariants.rb
malcolmbaig: It shouldn't be possible, following this change https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/services/submit_reference.rb#L11

stevehook: I thought that logic must be somewhere. Makes sense.

app/workers/detect_invariants.rb
tijmenb: I *think* we've cancelled those references (but since deleted the code). We'll see once we run this check!

stevehook (approved):
:shipit:

PR #3278, opened 2020-10-28
Decoupled references - edit email address for failed request

stevehook:
## Context

As a candidate I need to edit the address on a reference request and resend so that I can re-request a reference that may have been sent to an incorrect or unreachable email address.

This PR replaces https://github.com/DFE-Digital/apply-for-teacher-training/pull/3205

- [x] For failed reference requests (`email_bounced` status) allow the user to _Retry request_ when there are less than two other references provided.
- [x] The user is prompted to optionally change the email address before resending. Otherwise the original email address is used again.
- [x] Extend `ReferenceHistoryComponent` to include the recipient's email address for retry and send reference request operations.

image
image
image

## Guidance to review

- Should be testable via review app
- Is the conditional logic for the change link correct?
- I have not added the edit email address feature for applications in the cancelled state. Is this correct?

## Link to Trello card

https://trello.com/c/FqNl6q0u/2238-💔-dev-edit-email-address-of-failed-reference-request

## Things to check


paulrobertlloyd (approved):
:shipit:

spec/components/candidate_interface/decoupled_references_review_component_spec.rb
```diff
@@ -97,17 +97,28 @@
     end
   end
 
-  context 'when reference state is "cancelled" and the reference is complete' do
-    let(:feedback_requested) { create(:reference, :feedback_requested) }
-    let(:cancelled) { create(:reference, :cancelled) }
+  context 'when reference state is "cancelled" or "email_bounced" and the reference is complete' do
```
malcolmbaig, line 100: The contents of this block are a little tricky to follow. Might it make sense to have separate contexts for `cancelled` and `email_bounced`? A side benefit is that then your `email_bounced` context doesn't have to state " and the reference is complete", which isn't relevant to the behaviour we want to test (ie - according to other specs in this file, we could never have sent it in the first place if it was incomplete).

spec/components/candidate_interface/decoupled_references_review_component_spec.rb
```diff
@@ -97,17 +97,28 @@
...
-      result = render_inline(described_class.new(references: [feedback_requested, cancelled]))
+    it 'a retry request link is available' do
+      result = render_inline(described_class.new(references: [feedback_requested, email_bounced]))
 
       feedback_requested_summary = result.css('.app-summary-card')[0]
       feedback_cancelled_summary = result.css('.app-summary-card')[1]
-      expect(feedback_requested_summary.text).not_to include 'Send request again'
-      expect(feedback_cancelled_summary.text).to include 'Send request again'
+      expect(feedback_requested_summary.text).not_to include 'Retry request'
+      expect(feedback_cancelled_summary.text).to include 'Retry request'
```
malcolmbaig, line 114: Should this be called something like `email_bounced_summary`?

spec/components/candidate_interface/decoupled_references_review_component_spec.rb
```diff
@@ -97,28 +97,40 @@
...
+      expect(feedback_provided_summary.text).not_to include 'Retry request'
+      expect(email_bounced_summary.text).to include 'Retry request'
     end
+  end
+
+  context 'when reference state is "cancelled" but there are already 2 references provided' do
+    let(:email_bounced) { create(:reference, :email_bounced) }
+    let(:provided_references) { create_list(:reference, 2, :feedback_provided, application_form: email_bounced.application_form) }
 
     it 'a retry request link is not available if there are sufficient references provided' do
```
malcolmbaig, line 118: Can just call this block `a retry request link is not available`

spec/components/candidate_interface/decoupled_references_review_component_spec.rb
```diff
@@ -97,28 +97,40 @@
...
 
-      feedback_cancelled_summary = result.css('.app-summary-card')[2]
-      expect(feedback_cancelled_summary.text).not_to include 'Retry request'
+  context 'when reference state is "cancelled" and the reference is complete' do
+    let(:cancelled) { create(:reference, :cancelled) }
+
+    it 'a send request link is available' do
+      result = render_inline(described_class.new(references: [cancelled]))
+
+      feedback_not_requested_summary = result.css('.app-summary-card')[0]
```
malcolmbaig, line 132: Should this be `cancelled_summary`, so as to mirror the reference status?

spec/system/candidate_interface/decoupled_references/candidate_requests_a_reference_spec.rb
```diff
@@ -176,7 +191,40 @@ def when_i_have_a_cancelled_reference
...
+  def when_i_change_the_email_address
+    fill_in 'Referee’s email address', with: 'john@example.com'
+  end
+
+  def and_i_confirm_that_i_am_ready_to_retry_a_reference_request
+    click_button 'Send reference request'
+  end
+
+  def and_the_reference_email_address_has_been_updated
+    expect(@reference.reload.email_address).to eq('john@example.com')
```
malcolmbaig, line 228: For clarity, maybe we should create the reference in `when_i_have_a_failed_reference` with an explicit email that differs from this one.

malcolmbaig (approved):
Few more spec-related comments, but otherwise this LGTM :+1: 

Worth squashing the commit history prior to merge.

PR #3283, opened 2020-10-29
Additions to the performance dashboard

stevehook:
## Context

We need to update the performance dashboard with the following additions:

- Apply Again Submissions
- Apply Again Offer Accepted
- % of providers onboarded
- No. applications Rejected by Default (RBD)

- [x] Change the query to group by `phase` as well as application form id and application choice id so that we can get the numbers for Apply again.
- [x] Add a separate query to get the percentage of providers onboarded
- [x] Add a separate 'state' to the query that picks out rejected by default applications
- [ ] Organise the tiles (design)

image

## Guidance to review

- How should we layout the new metrics? Is the separate table correct?
- Does grouping by `phase` make sense from an implementation point of view?
- Is it OK to assume that providers are onboarded when have one or more courses open on apply?
- Does it make sense to show the % or _Providers onboarded_ on any other tab than the _All years_ one? (It's find of meaningless to filter by recruitment cycle without some kind of historical data).

## Link to Trello card

https://trello.com/c/bSoFxjhp/2399-dev-pa-dashboard-for-ministerial-reports

## Things to check

### DIFF
--- a/app/models/performance_statistics.rb
+++ b/app/models/performance_statistics.rb
@@ -65,7 +65,10 @@ class PerformanceStatistics
   end
 
   def [](key)
-    candidate_status_counts.find { |x| x['status'] == key.to_s }&.[]('count')
+    candidate_status_counts
+      .select { |x| x['status'] == key.to_s }
+      .map { |row| row['count'] }
+      .sum
   end
 
   def total_candidate_count(only: nil, except: [], phase: nil)
@@ -82,6 +85,12 @@ class PerformanceStatistics
       .connection
       .execute(candidate_query)
       .to_a
+  end
+
+  def candidate_status_total_counts
+    candidate_status_counts.group_by { |row| row['status'] }.map do |status, rows|
+      { 'status' => status, 'count' => rows.map { |r| r['count'] }.sum }
+    end
   end
 
   def total_submitted_count
--- a/app/models/performance_statistics.rb
+++ b/app/models/performance_statistics.rb
@@ -73,7 +73,7 @@ class PerformanceStatistics
   def total_candidate_count(only: nil, except: [], phase: nil)
     candidate_status_counts
       .select { |row| only.nil? || row['status'].to_sym.in?(only) }
-      .select { |row| phase.nil? || row['phase'].to_sym == phase.to_sym }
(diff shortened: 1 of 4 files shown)


app/models/performance_statistics.rb
```diff
@@ -80,15 +87,40 @@ def candidate_status_counts
...
+    total_candidate_count(only: %i[pending_conditions recruited offer_deferred], phase: :apply_2)
+  end
+
+  def rejected_by_default_count
+    total_candidate_count(only: %i[rejected_by_default])
+  end
+
+  def percentage_of_providers_onboarded
+    @percentage_of_providers_onboarded ||= begin
+      counts = Provider.group(:sync_courses).count
```
tijmenb, line 122: Not sure if this is the correct definition of onboarded. I would consider it to be the "providers with open courses" stat from here: https://www.apply-for-teacher-training.service.gov.uk/support/performance/course-stats. Maybe we don't need it here?

config/locales/application_states.yml
```diff
@@ -65,6 +65,20 @@ en:
     conditions_not_met: Conditions not met
     offer_deferred: Deferred
 
+  performance_dashboard_other_metrics:
+    apply_again_submitted:
+      name: Apply again submitted
+      description: The candidate has applied again and submitted an application.
+    apply_again_accepted_offer:
+      name: Apply again offers accepted
+      description: The candidate has applied again an accepted an offer.
```
stevehook, line 74:
```suggestion
      description: The candidate has applied again and accepted an offer.
```

tvararu (approved):
LGTM

Not sure if the "Providers onboarded" metric makes sense or works when viewing data from previous cycles but we can iterate and fix it later if it doesn't work. On the review app it says it's 0% which is wrong but it probably works on the real app.

stevehook:
> LGTM
> 
> Not sure if the "Providers onboarded" metric makes sense or works when viewing data from previous cycles but we can iterate and fix it later if it doesn't work. On the review app it says it's 0% which is wrong but it probably works on the real app.

There are two providers on the review app that are onboarded (by our definition). Maybe rounding to the nearest whole number was a mistake. `0.1%` rounds to `0%`.

PR #3285, opened 2020-10-29
Only show referee first name in reference reminder Slack message

malcolmbaig:
## Context

We're showing the full name currently, better for privacy to show first name only.

Use a string split to extract the first part of the name.

## Guidance to review
See diff

## Link to Trello card
n/a

## Things to check

### DIFF
--- a/app/services/candidate_interface/decoupled_references/send_reference_reminder.rb
+++ b/app/services/candidate_interface/decoupled_references/send_reference_reminder.rb
@@ -21,18 +21,22 @@ module CandidateInterface
 
           send_slack_message(application_form, message)
 
-          flash[:success] = "Reminder sent to #{reference_name}"
+          flash[:success] = "Reminder sent to #{referee_name}"
         end
       end
 
     private
 
       def message
-        "Candidate #{application_form.first_name} has sent a reminder to #{reference_name}"
+        "Candidate #{application_form.first_name} has sent a reminder to #{referee_first_name}"
       end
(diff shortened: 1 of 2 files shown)


TobyRet (approved):

app/services/candidate_interface/decoupled_references/send_reference_reminder.rb
```diff
@@ -21,20 +21,24 @@ def call
...
-          flash[:success] = "Reminder sent to #{reference_name}"
+          flash[:success] = "Reminder sent to #{referee_name}"
         end
       end
 
     private
 
       def message
-        "Candidate #{application_form.first_name} has sent a reminder to #{reference_name}"
+        "Candidate #{application_form.first_name} has sent a reminder to #{referee_first_name}"
```
tijmenb, line 31: We can use `reference.first_name` instead, it's inputted separately.

tijmenb (approved):

app/services/candidate_interface/decoupled_references/send_reference_reminder.rb
malcolmbaig: We just capture the whole name in a single field, unfortunately! https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/views/candidate_interface/decoupled_references/name/_shared_form.html.erb

tijmenb (approved):

PR #3286, opened 2020-10-29
Regenerate application choice state diagram

malcolmbaig:
## Context
Reflect recent changes in available states.

## Guidance to review
View the diagram, check it looks right.

## Things to check


stevehook (approved):
Looks right to me. I guess `cancelled` is unreachable for any applications going forward but needs to be retained for legacy applications?

PR #3287, opened 2020-10-29
Add work history data export

raamSoftwire:
## Context

To validate/support DfE policy view that we should be enforcing collection of full rather than 5-years worth of Work history, we need to analyse how many candidates only provide 5-years of history (when they could have provided more).

I have chosen to create a data export over running queries directly on the DB because 
- we can reuse a lot of business logic for calculating unexplained breaks in work history
- it makes it easier to rerun the export again in future if needed
- we can write tests
- it doesn't require production access

Add a new data export for work history

## Guidance to review

These changes broadly follow the changes made for another [data export recently](https://github.com/DFE-Digital/apply-for-teacher-training/commit/f8b45fb6dc802ffa273649ab5646e92ca3c39e39).

It also reuses a lot of logic from [work_history_with_breaks.rb](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/services/work_history_with_breaks.rb) and [work_history_item_component](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/components/work_history_item_component.rb).

## Link to Trello card

https://trello.com/c/jPJLjr4Q/2392-dev-how-much-work-history

## Things to check

### DIFF
--- a/app/services/support_interface/unexplained_breaks_in_work_history_export.rb
+++ b/app/services/support_interface/unexplained_breaks_in_work_history_export.rb
@@ -27,12 +27,16 @@ module SupportInterface
         next if unexplained_breaks.nil?
 
         total_unexplained_time = unexplained_breaks.sum(&:length)
+        unexplained_breaks_in_last_five_years = unexplained_breaks.select {
+            |unexplained_break| unexplained_break.start_date > (submitted_at(application_form) - 5.years).to_date }
+
         output = {
             'Candidate id' => application_form.candidate_id,
             'Application id' => application_form.id,
             'Start of working life' => get_start_of_working_life(application_form),
+            'Total unexplained time (months)' => total_unexplained_time,
             'Number of unexplained breaks' => unexplained_breaks.length,
-            'Total unexplained time (months)' => total_unexplained_time
+            'Number of unexplained breaks in last 5 years' => unexplained_breaks_in_last_five_years.count,
         }
         output
       end
--- a/app/services/support_interface/unexplained_breaks_in_work_history_export.rb
+++ b/app/services/support_interface/unexplained_breaks_in_work_history_export.rb
@@ -27,16 +27,22 @@ module SupportInterface
         next if unexplained_breaks.nil?
 
         total_unexplained_time = unexplained_breaks.sum(&:length)
-        unexplained_breaks_in_last_five_years = unexplained_breaks.select {
+
+        unexplained_breaks_in_last_five_years = unexplained_breaks.count {
             |unexplained_break| unexplained_break.start_date > (submitted_at(application_form) - 5.years).to_date }
 
+        degrees = application_form.application_qualifications.degrees
+        unexplained_breaks_that_coincide_with_degrees =  unexplained_breaks.count {
+            |unexplained_break| unexplained_break_coincides_with_a_degree(unexplained_break, degrees)}
+        
         output = {
             'Candidate id' => application_form.candidate_id,
             'Application id' => application_form.id,
             'Start of working life' => get_start_of_working_life(application_form),
             'Total unexplained time (months)' => total_unexplained_time,
             'Number of unexplained breaks' => unexplained_breaks.length,
-            'Number of unexplained breaks in last 5 years' => unexplained_breaks_in_last_five_years.count,
+            'Number of unexplained breaks in last 5 years' => unexplained_breaks_in_last_five_years,
+            'Number of unexplained breaks that coincide with studying for a degree' => unexplained_breaks_that_coincide_with_degrees,
         }
         output
       end
@@ -45,6 +51,12 @@ module SupportInterface
     end
 
     private
+
+    def unexplained_break_coincides_with_a_degree(unexplained_break, degrees)
+      degrees.select { |degree| Date.new(degree.start_year.to_i,1,1) < unexplained_break.end_date &&
+          unexplained_break.start_date < Date.new(degree.award_year.to_i,12,31)}
+          .any?
+    end
 
     def get_unexplained_breaks(application_form)
       work_history = application_form.application_work_experiences.sort_by(&:start_date)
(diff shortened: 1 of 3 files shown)


app/services/support_interface/unexplained_breaks_in_work_history_export.rb
```diff
@@ -0,0 +1,108 @@
...
+      def start_date
+        @month_range.first.prev_month
+      end
+
+      def end_date
+        @month_range.last.next_month
+      end
+
+      def length
+        ((end_date.year * 12) + end_date.month) - ((start_date.year * 12) + start_date.month)
```
raamSoftwire, line 17: This was necessary to make sure the length calculated matched up with the one shown on the work history timeline on the `support/application/{id}` page.

This logic has been taken from [work_history_item_component.rb](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/components/work_history_item_component.rb).

app/services/support_interface/unexplained_breaks_in_work_history_export.rb
```diff
@@ -0,0 +1,108 @@
...
+      end
+    end
+
+    def data_for_export
+      data_for_export = Candidate.all.map do |candidate|
+        application_form = candidate.application_forms.order(:submitted_at).last
```
raamSoftwire, line 23: This line is to handle the Apply Again case where a candidate has 2 applications.

app/services/support_interface/unexplained_breaks_in_work_history_export.rb
```diff
@@ -0,0 +1,108 @@
...
+        next if application_form.nil?
```
raamSoftwire, line 24: This is for the case where a candidate hasn't created an application yet.

app/services/support_interface/unexplained_breaks_in_work_history_export.rb
```diff
@@ -0,0 +1,108 @@
...
+        @month_range.last.next_month
+      end
+
+      def length
+        ((end_date.year * 12) + end_date.month) - ((start_date.year * 12) + start_date.month)
+      end
+    end
+
+    def data_for_export
+      data_for_export = Candidate.all.map do |candidate|
```
raamSoftwire, line 22: I'm aware that this loads all of the candidates in to memory and will be quite slow in production. If anyone is able to suggest an improvement here I'd be interested to know

app/services/support_interface/unexplained_breaks_in_work_history_export.rb
```diff
@@ -0,0 +1,108 @@
...
+        )
+        break_months_in_timeline = remove_months(timeline: timeline_in_months, entries: work_history, application_form: application_form)
+        remaining_months = remove_months(timeline: break_months_in_timeline, entries: existing_breaks, application_form: application_form)
+
+        create_unexplained_breaks(remaining_months)
+      end
+    end
+
+    def get_start_of_working_life(application_form)
+      application_form.date_of_birth.beginning_of_month + 18.years
```
raamSoftwire, line 66: The `.beginning_of_month` was to sanitise the data so that birthdays don't leak in to the export, since we don't need it for the calculations.

I'd be interested to know if you think this step was necessary.

app/services/support_interface/unexplained_breaks_in_work_history_export.rb
TobyRet: These exports run as a Sidekiq job so hopefully that will alleviate any major memory issues. The only other suggestion I can think of is whether you could use `select` to bring back only the data you require in each application form (e.g. https://github.com/DFE-Digital/apply-for-teacher-training/commit/f8b45fb6dc802ffa273649ab5646e92ca3c39e39#diff-f3cf4595d7d06f89258aa90c9f8553e0a18288e0b7d303469a05efd582e1ff60R31)

TobyRet: Perhaps use of [includes](https://apidock.com/rails/v6.0.0/ActiveRecord/QueryMethods/includes) might be an idea also - Candidate.includes(:application_forms)

app/services/support_interface/unexplained_breaks_in_work_history_export.rb
TobyRet: Makes sense to me but I'd probably run that one passed Mike Emery as it sounds more like a policy / decision rule

raamSoftwire: I spoke to Mike and he thought it was a good idea 👍

TobyRet (commented):
Looking good. Added a few comments. You'll need to run your linter also

app/services/support_interface/unexplained_breaks_in_work_history_export.rb
raamSoftwire: I'm struggling to write a Rails query to get the most recent application for each candidate before I start. 

@TobyRet has helped me to get to  `ApplicationForm.order(:submitted_at).group_by{|a| a.candidate_id}.limit(1)`, but this still isn't quite right as it only returns a single ApplicationForm rather than one for each Candidate.

I've added selects and includes

```
data_for_export = Candidate.includes(:application_forms).all.map do |candidate|
        application_form = candidate.application_forms
                               .select(:submitted_at, :date_of_birth, :candidate_id, :id)
                               .includes(:application_qualifications, :application_work_experiences, :application_work_history_breaks)
                               .order(:submitted_at).last
```
It doesn't make a difference at this scale, but it might once we run it on the production data

app/services/support_interface/unexplained_breaks_in_work_history_export.rb
```diff
@@ -19,15 +19,12 @@ def length
...
-          .all.map do |candidate|
-        application_form = candidate.application_forms
-          .select(:submitted_at, :date_of_birth, :candidate_id, :id)
-          .includes(:application_qualifications, :application_work_experiences, :application_work_history_breaks)
-          .order(:submitted_at).last
-        next if application_form.nil?
+      applications = ApplicationForm
+                         .select(:id, :candidate_id, :submitted_at, :date_of_birth)
+                         .includes(:application_qualifications, :application_work_experiences, :application_work_history_breaks)
+                         .order(submitted_at: :desc).uniq(&:candidate_id)
```
raamSoftwire, line 25: This works because `uniq` traverses in order and keeps the first occurrence.

TobyRet (approved):
LGTM

PR #3288, opened 2020-10-29
Structured pre-degree quals | Decoupling Gcse::GradeController

TobyRet:
## Context

As a developer
I want to reduce coupling across the GSCE grade update flow
So that it will be easier to make the necessary changes to allow candidates to enter multiple english and science GCSEs (https://trello.com/c/jJPtfRY3/1779-%F0%9F%8F%88-allow-candidates-to-enter-multiple-english-and-science-gcses)

- Create separate controllers and views per GSCE subject for updating the grade
- Move common logic across the new controllers to a concern

## Guidance to review

Although there is now more duplication, I imagine much of this will disappear once https://trello.com/c/jJPtfRY3/1779-%F0%9F%8F%88-allow-candidates-to-enter-multiple-english-and-science-gcses is implemented. 

Hopefully this also gives us a better platform to begin work on the next spikes:
- https://trello.com/c/6hn5shti/2421-spike-structured-pre-degree-quals-science-gcse-data-model
- https://trello.com/c/XKkHzPM2/2422-spike-structured-pre-degree-quals-english-gcse-data-model

## Link to Trello card

https://trello.com/c/tg6YxlE2/2420-spike-structured-pre-degree-quals-decoupling-gcsegradecontroller

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/gcse/english/grade_controller.rb
+++ b/app/controllers/candidate_interface/gcse/english/grade_controller.rb
@@ -1,0 +1,38 @@
+module CandidateInterface
+  class Gcse::English::GradeController < CandidateInterfaceController
+    include Gcse::GradeControllerConcern
+
+    before_action :redirect_to_dashboard_if_submitted
+    before_action :set_subject
+
+    def edit
+      @application_qualification = details_form
+      @qualification_type = details_form.qualification.qualification_type
+    end
+
+    def update
+      @qualification_type = details_form.qualification.qualification_type
+
+      details_form.grade = details_params[:grade]
+      details_form.other_grade = details_params[:other_grade]
+
+      @application_qualification = details_form.save_grade
+
+      if @application_qualification
+        update_gcse_completed(false)
+        redirect_to next_gcse_path
+      else
+        @application_qualification = details_form
+        track_validation_error(@application_qualification)
+
+        render :edit
+      end
+    end
+
+  private
+
+    def set_subject
+      @subject = 'english'
+    end
+  end
+end
--- a/app/controllers/candidate_interface/gcse/grade_controller_concern.rb
+++ b/app/controllers/candidate_interface/gcse/grade_controller_concern.rb
@@ -1,0 +1,38 @@
+module CandidateInterface
+  module Gcse
+    module GradeControllerConcern
+      extend ActiveSupport::Concern
+
+      included do
+        helper_method :autocomplete_grades?
+      end
+
+      def autocomplete_grades?
+        @subject.in?(%w[maths english]) && @qualification_type == 'gcse'
+      end
+
+      def next_gcse_path
+        if details_form.award_year.nil?
+          candidate_interface_gcse_details_edit_year_path(subject: @subject)
+        else
+          candidate_interface_gcse_review_path(subject: @subject)
+        end
+      end
+
+      def details_params
+        params.require(:candidate_interface_gcse_qualification_details_form).permit(%i[grade award_year other_grade])
+      end
+
+      def details_form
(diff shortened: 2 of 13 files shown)


app/controllers/candidate_interface/gcse/naric_reference_controller.rb
```diff
@@ -47,5 +47,16 @@ def naric_reference_params
       params.require(:candidate_interface_naric_reference_form)
         .permit(:naric_reference_choice, :naric_reference, :comparable_uk_qualification)
     end
+
+    def grade_edit_path
```
TobyRet, line 51: This method appears in a number of places and could be extracted

davidgisbey:
This seems like a good approach to me. Like you say I think that a lot of the duplication in the views will go during the next spikes.

app/controllers/candidate_interface/gcse/naric_reference_controller.rb
malcolmbaig: Some kind of `ResolveGcseEditPath` service object would be useful.

config/routes.rb
```diff
@@ -127,6 +127,16 @@
         post '/complete' => 'contact_details/review#complete', as: :contact_details_complete
       end
 
+      scope '/gcse' do
+        get '/maths/grade' => 'gcse/maths/grade#edit', as: :gcse_maths_edit_grade
+        get '/science/grade' => 'gcse/science/grade#edit', as: :gcse_science_edit_grade
+        get '/english/grade' => 'gcse/english/grade#edit', as: :gcse_english_edit_grade
+
+        patch '/maths/grade' => 'gcse/maths/grade#update', as: :gcse_maths_update_grade
```
malcolmbaig, line 135: No need for the additional path helpers here, we can just PATCH the edit paths above and they'll resolve the same.

app/controllers/candidate_interface/gcse/english/grade_controller.rb
```diff
@@ -0,0 +1,38 @@
+module CandidateInterface
+  class Gcse::English::GradeController < CandidateInterfaceController
+    include Gcse::GradeControllerConcern
+
+    before_action :redirect_to_dashboard_if_submitted
+    before_action :set_subject
+
+    def edit
+      @application_qualification = details_form
```
malcolmbaig, line 9: Worth revisiting this naming in the next bit of this work, especially if we split out the details form. Will probably be a `@grade_form` variable that we end up dealing with.

app/controllers/candidate_interface/gcse/grade_controller_concern.rb
```diff
@@ -0,0 +1,38 @@
+module CandidateInterface
+  module Gcse
+    module GradeControllerConcern
```
malcolmbaig, line 3: I think the concern is a good interim step :+1: . I suspect we may end up refactoring a lot of it away as the feature progresses.

config/routes.rb
TobyRet: Sorry @malcolmbaig - not sure I understand. Are you saying the path helpers (e.g. `:gcse_maths_update_grade`) should be deleted?

malcolmbaig: Yes - the `gcse_maths_edit_grade_path` helper (for example) already gives you access to the '/maths/grade' endpoint, so the HTML form you build can just do a `method: :patch` and use the edit helper. This lines up with the path conventions Rails uses when it generates routes for you via the `resources` method, IIRC.

TobyRet: 👌

malcolmbaig (approved):

PR #3291, opened 2020-10-29
Set Commit SHA as ENV variable

vigneshmsft:
Register OKComputer:AppVersionCheck, which checks for presense of ENV['SHA']
[`/check/version`](https://dev.apply-for-teacher-training.service.gov.uk/check/version) endpoint to respond with current version.
Set version info in Raven configuration for identifying version in Sentry.

## Context

Make it easy to identify offending versions in [Sentry](https://docs.sentry.io/platforms/ruby/configuration/releases/).
And the current version in each environment.

`SHA` env variable set to the git commit (same as the docker image tag).

Adds optional [`OkComputer:AppVersionCheck`](https://github.com/sportngin/okcomputer/blob/master/lib/ok_computer/built_in_checks/app_version_check.rb#L4)

## Things to check

### DIFF
--- a/config/initializers/raven.rb
+++ b/config/initializers/raven.rb
@@ -1,6 +1,7 @@
 Raven.configure do |config|
   config.silence_ready = true
   config.current_environment = HostingEnvironment.environment_name
+  config.release = ENV['SHA']
   config.sanitize_fields = Rails.application.config.filter_parameters.map(&:to_s)
   config.inspect_exception_causes_for_exclusion = true
 
--- a/docker-compose.azure.yml
+++ b/docker-compose.azure.yml
@@ -6,6 +6,7 @@ services:
       args:
         bundleWithout: 'development'
         BASE_RUBY_IMAGE_WITH_GEMS_AND_NODE_MODULES: ${gemsNodeModulesImageName:-apply-for-teacher-training-gems-node-modules}
+        VERSION: ${dockerHubImageTag}
     image: ${dockerHubUsername}/${dockerHubImageName}:${dockerHubImageTag}
     environment:
       - RAILS_ENV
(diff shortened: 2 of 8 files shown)


config/initializers/raven.rb
```diff
@@ -1,6 +1,7 @@
 Raven.configure do |config|
   config.silence_ready = true
   config.current_environment = HostingEnvironment.environment_name
+  config.release = "#{HostingEnvironment.environment_name}@#{ENV['SHA']}"
```
tijmenb, line 4: I'm not sure if we should prefix this. In Sentry you can say "Resolve in next release", which means the error will only re-alert if it happens in a next version. If the version is hosting environment-specific, it won't work across environments.

davidgisbey (approved):

docker-compose.azure.yml
```diff
@@ -6,6 +6,7 @@ services:
       args:
         bundleWithout: 'development'
         BASE_RUBY_IMAGE_WITH_GEMS_AND_NODE_MODULES: ${gemsNodeModulesImageName:-apply-for-teacher-training-gems-node-modules}
+        SHA: ${dockerHubImageTag}
```
tijmenb, line 9: Maybe something like `DEPLOYED_VERSION` instead of `SHA` would be clearer?

tijmenb (commented):
Cool! Will help a lot with error tracking. One comment about the release we send to Sentry and one about the naming.

docker-compose.azure.yml
vigneshmsft: began with `VERSION`, I'll change this to `VERSION` but realised OKComputer used `SHA` and changed all occurances,
will change this entry to `VERSION`

PR #3292, opened 2020-10-29
Invariant check for unauthorised application edits

malcolmbaig:
## Context

To ensure we aren't allowing unscoped access to application form data.
An invariant check that looks for audit log entries describing application edits
by a candidate who isn't the owner of the application in question.
 

## Guidance to review
The query relies on any `audited` declarations in a model to be associated with the application form. Here's what we cover:
```
app/models/application_choice.rb
app/models/application_work_experience.rb
app/models/application_work_history_break.rb
app/models/application_qualification.rb
app/models/application_reference.rb
app/models/application_volunteering_experience.rb
```
Does anything else need covering by this check?

## Link to Trello card
https://trello.com/c/hChP4g58/2395-run-query-against-audit-log-to-find-any-historical-un-scoped-updates

## Things to check

### DIFF
--- a/app/workers/detect_invariants.rb
+++ b/app/workers/detect_invariants.rb
@@ -10,7 +10,7 @@ class DetectInvariants
 
   def detect_application_choices_in_old_states
     choices_in_wrong_state = begin
-      ApplicationChoice.where(status: %w[awaiting_references application_complete])
+      ApplicationChoice.where(status: %w[awaiting_references application_complete]).map(&:id).sort
     end
 
     if choices_in_wrong_state.any?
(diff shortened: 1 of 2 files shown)


app/workers/detect_invariants.rb
```diff
@@ -17,7 +18,7 @@ def detect_application_choices_in_old_states
         One or more application choices are still in `awaiting_references` or
         `application_complete` state, but all these states have been removed:
 
-        #{choices_in_wrong_state.map(&:id).join("\n")}
+        #{choices_in_wrong_state.map(&:id).sort.join("\n")}
```
raamSoftwire, line 21: Is there a reason why we sort this output but not the other one?

raamSoftwire (commented):
👍 What you have here looks good to me

> Does anything else need covering by this check?

I don't feel like I know enough about the codebase to answer this confidently yet

app/workers/detect_invariants.rb
malcolmbaig: Just that the associated spec expects ordered output. No reason not to sort the others tho, probably nicer that way. Will add. :+1:

davidgisbey (approved):
Looks good to me 👍 

I think this covers everything.

PR #3294, opened 2020-10-29
Add deferred offer reminder email

benswannack:
## Context
This email is being added so that candidates might chase providers about deferred offers that should be reinstated.
The job to send all these emails can be triggered in the support interface tasks section

New email looks like: 
image

## Guidance to review
I can't think of a good reason we would want to limit the number of times this button can be pressed per cycle - it's not a big deal if it happens twice and it's unlikely.

## Link to Trello card
https://trello.com/c/2FdZQ2SY/2854-dev-deferral-email-notifications

## Things to check

### DIFF
--- a/app/queries/get_deferred_application_choices_for_current_cycle.rb
+++ b/app/queries/get_deferred_application_choices_for_current_cycle.rb
@@ -1,0 +1,8 @@
+class GetDeferredApplicationChoicesForCurrentCycle
+  def self.call
+    ApplicationChoice
+      .joins(:course)
+      .where(status: :offer_deferred)
+      .where('courses.recruitment_cycle_year': RecruitmentCycle.previous_year)
+  end
+end
--- a/app/views/candidate_mailer/deferred_offer_reminder.text.erb
+++ b/app/views/candidate_mailer/deferred_offer_reminder.text.erb
@@ -1,0 +1,15 @@
+Dear <%= @application_form.first_name %>,
+
+# Reminder of your deferred offer
+
+On <%= @application_choice.offer_deferred_at.to_s(:govuk_date) %>, <%= @course_option.course.provider.name %> deferred your offer to study <%= @course_option.course.name_and_code %>.
+
+You'll shortly receive another email from us, confirming that you’ll start training in the new academic year (<%= RecruitmentCycle.current_year %>). If you don’t receive this message, you should contact us at [becomingateacher@digital.education.gov.uk](mailto:becomingateacher@digital.education.gov.uk).
+
+Sign in to your account to view the details of your deferred offer:
+
+<%= candidate_magic_link(@application_choice.application_form.candidate) %>
+
+# Give feedback or report a problem
+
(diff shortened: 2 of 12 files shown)


tvararu (approved):

app/views/candidate_mailer/deferred_offer_reminder.text.erb
```diff
@@ -0,0 +1,15 @@
+Dear <%= @application_form.first_name %>,
+
+# Reminder of your deferred offer
+
+On <%= @application_choice.offer_deferred_at.to_s(:govuk_date) %>, <%= @course_option.course.provider.name %> deferred your offer to study <%= @course_option.course.name_and_code %>.
+
+You'll shortly receive another email from us, confirming that you’ll start training in the new academic year (<%= RecruitmentCycle.current_year %>). If you don’t receive this message, you should contact us at [becomingateacher@digital.education.gov.uk](mailto:becomingateacher@digital.education.gov.uk).
```
mnacos, line 7: In order for training to commence, the provider must first reinstate the offer, confirm the conditions of the offer are met (which can happen at the same time as reinstating the offer, but this is less likely) and only then is the candidate 'recruited'. When their training actually starts depends on the course etc. and we won't be emailing them about that. This is the provider's responsibility.

Considering all this, this sentence appears a little inaccurate. Any suggestions @adamsilver ?

app/queries/get_deferred_application_choices_for_current_cycle.rb
```diff
@@ -0,0 +1,8 @@
+class GetDeferredApplicationChoicesForCurrentCycle
+  def self.call
+    ApplicationChoice
+      .joins(:course)
+      .where(status: :offer_deferred)
+      .where('courses.recruitment_cycle_year': RecruitmentCycle.previous_year)
```
mnacos, line 6: I'd recommend adding a service spec for this, considering the button triggers multiple emails. Also, we'll soon be changing the way we target these deferred offers, finding them based on `offer_deferred_at` rather than the course recruitment_cycle_year.

mnacos (commented):
Looks good so far. Just a couple of comments about:

- the reminder email content (may require some tweaking), and
- I'd like to see a `get_deferred_application_choices_for_current_cycle_spec.rb` service spec, to make sure we're getting this right

app/views/candidate_mailer/deferred_offer_reminder.text.erb
benswannack: Yes that's a good point. How about changing this line to something like:
`You'll shortly receive another email from us, confirming the offer. If you don’t receive this message, you should contact {{Provider}}`
This makes it clear that the candidate should expect an email confirming (reinstating) their offer, and it they don't get one they should chase the provider.
What do you think @adamsilver ?

adamsilver: I'm thinking:

`You’ll shortly receive an email to confirm the offer. If you do not receive this, you should contact {{provider}}`

benswannack:
Email now reads:
image

mnacos (approved):
Copy adjumstments work well and the new spec is super readable :rocket:

PR #3295, opened 2020-10-30
Add filtering to UCAS matches page

AgaDufrat:
## Context

We want to be able to filter by the recruitment cycle year.

- add filtering on the index page
Screenshot 2020-10-29 at 16 54 57

- display recruitment cycle year and first added date on the show page

| Before   |      After      |  
|----------|------------|
|Screenshot 2020-10-29 at 17 02 44| Screenshot 2020-10-29 at 17 02 24|

## Guidance to review

- visit `support/ucas-matches`

## Link to Trello card

## Things to check

### DIFF
--- a/app/controllers/support_interface/ucas_matches_controller.rb
+++ b/app/controllers/support_interface/ucas_matches_controller.rb
@@ -1,7 +1,10 @@
 module SupportInterface
   class UCASMatchesController < SupportInterfaceController
     def index
-      @matches = UCASMatch.includes(:candidate)
+      @filter = SupportInterface::UCASMatchesFilter.new(params: params)
+      @matches = @filter.filter_records(UCASMatch.includes(:candidate))
+        .order(updated_at: :desc)
+        .page(params[:page] || 1).per(15)
     end
 
     def show
--- a/app/views/support_interface/ucas_matches/index.html.erb
+++ b/app/views/support_interface/ucas_matches/index.html.erb
@@ -1,35 +1,3 @@
 <%= render 'support_interface/candidates/candidates_navigation', title: 'UCAS matches' %>
 
-<table class="govuk-table">
-  <thead class="govuk-table__head">
-    <tr class="govuk-table__row">
-      <th scope="col" class="govuk-table__header">Status</th>
-      <th scope="col" class="govuk-table__header">Candidate</th>
-      <th scope="col" class="govuk-table__header">Last update</th>
-    </tr>
-  </thead>
-
-  <tbody class="govuk-table__body">
-    <% @matches.each do |match| %>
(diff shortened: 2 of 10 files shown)


app/controllers/support_interface/ucas_matches_controller.rb
```diff
@@ -1,7 +1,13 @@
 module SupportInterface
   class UCASMatchesController < SupportInterfaceController
     def index
-      @matches = UCASMatch.includes(:candidate)
+      @filter = SupportInterface::MatchesFilter.new(params: params)
```
duncanjbrown, line 4: Maybe UCASMatchesFilter?

app/controllers/support_interface/ucas_matches_controller.rb
```diff
@@ -1,7 +1,13 @@
...
+      @matches = @filter.filter_records(
+        UCASMatch
+        .includes(:candidate)
+        .order(updated_at: :desc)
+        .page(params[:page] || 1).per(15),
```
duncanjbrown, line 9: Shouldn't we paginate after filtering? in this case we could get say 15 records then filter out 5?

app/views/support_interface/ucas_matches/index.html.erb
```diff
@@ -1,35 +1,5 @@
...
-        </td>
-
-        <td class="govuk-table__cell">
-          <%= match.updated_at.to_s(:govuk_date_and_time) %>
-        </td>
-      </tr>
-    <% end %>
-  </tbody>
-</table>
+<%= render PaginatedFilterComponent.new(filter: @filter, collection: @matches) do %>
```
duncanjbrown, line 3: I can't believe we have a reusable component that's this big AND is still useful! 🎉

app/models/support_interface/matches_filter.rb
```diff
@@ -0,0 +1,40 @@
...
+  class MatchesFilter
+    attr_reader :applied_filters
+
+    def initialize(params:)
+      @applied_filters = params
+    end
+
+    def filter_records(ucas_matches)
+      if applied_filters[:year]
+        ucas_matches = ucas_matches.where('recruitment_cycle_year IN (?)', applied_filters[:year])
```
duncanjbrown, line 11: Should this param be `years` as it can take multiple values?

duncanjbrown (commented):
Looks very good — one thing to check about the interaction between filtering and pagination.

- I think it would be friendlier to always be explicit that Matches are UCASMatches?
- should we default to current cycle only? (maybe the menu link should include ?year[]=2021 ?)

duncanjbrown (approved):

PR #3296, opened 2020-10-30
Export only applications in the current recruitment cycle for UCAS

AgaDufrat:
## Context

During the October 2020 Data Matching DSA Review we spoke with UCAS about two ways handling data in different recruitment cycles. 
The first option was to only send through candidates from the 2021 cycle going forward. 
The second was adding a 'recruitment cycle year' field in the data we send over and receive, which would require an amendment to the DSA. As discussed, both parties favoured the former.

Change UCAS export to only send candidate data from
the current recruitment cycle.

## Guidance to review

Makes sense?

## Link to Trello card

https://trello.com/c/E1Cjp1iK/2965-export-only-applications-in-the-current-recruitment-cycle

## Things to check

### DIFF
--- a/app/services/ucas_matching/matching_data_export.rb
+++ b/app/services/ucas_matching/matching_data_export.rb
@@ -54,6 +54,7 @@ module UCASMatching
 
     def relevant_applications
       ApplicationForm
+        .current_cycle
         .includes(
           :candidate,
         )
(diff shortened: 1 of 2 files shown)


benswannack (approved):
Nice :)

Not really for now, but will this cause issues at the end of the cycle? Could we miss applications close to the end?

AgaDufrat:
Hopefully not 🤞 The applications deadline (end of cycle) is usually a few days before the start of new recruitment cycle.
Also UCAS said take the approach that they only supply applicants in the current cycle, and they keep the data from the two cycles completely separate. That's why the first few datasets we got in the new cycle were empty because of that.

PR #3302, opened 2020-11-02
Update brakeman config to include UnscopedFind check

raamSoftwire:
## Context

We want to catch unscoped update actions and make sure they don't enter the codebase and cause security risks.

- Update the brakeman.yml config to include the `CheckUnscopedFind` check. 
- Fix the instances of this issue currently in the codebase (15 found)

## Link to Trello card

https://trello.com/c/5cJK6zDm/2394-investigate-ways-to-detect-un-scoped-queries

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/other_qualifications/type_controller.rb
+++ b/app/controllers/candidate_interface/other_qualifications/type_controller.rb
@@ -15,7 +15,7 @@ module CandidateInterface
     end
 
     def edit
-      @qualification_type = OtherQualificationTypeForm.build_from_qualification(ApplicationQualification.find(params[:id]))
+      @qualification_type = OtherQualificationTypeForm.build_from_qualification(current_qualification)
     end
 
     def update
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -1,0 +1,146 @@
+{
+  "ignored_warnings": [
+    {
+      "warning_type": "Unscoped Find",
+      "warning_code": 82,
+      "fingerprint": "07081676ffe0bf0ef753045ae9dcc86c848df5f35a9e48a22692159d5352fbd1",
+      "check_name": "UnscopedFind",
+      "message": "Unscoped call to `ProviderRelationshipPermissions#find`",
+      "file": "app/controllers/provider_interface/provider_relationship_permissions_controller.rb",
+      "line": 32,
+      "link": "https://brakemanscanner.org/docs/warning_types/unscoped_find/",
+      "code": "ProviderRelationshipPermissions.find(params[:id])",
+      "render_path": null,
+      "location": {
+        "type": "method",
+        "class": "ProviderInterface::ProviderRelationshipPermissionsController",
+        "method": "render_404_unless_permissions_found"
+      },
+      "user_input": "params[:id]",
+      "confidence": "Weak",
+      "note": ""
+    },
+    {
+      "warning_type": "Unscoped Find",
+      "warning_code": 82,
+      "fingerprint": "4c106936481ffa09c60383aed2a7b8e931b1672602d3f5a69ccaf9bd688cbf62",
+      "check_name": "UnscopedFind",
+      "message": "Unscoped call to `ProviderRelationshipPermissions#find`",
+      "file": "app/controllers/provider_interface/provider_relationship_permissions_setup_controller.rb",
+      "line": 40,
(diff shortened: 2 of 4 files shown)


tijmenb:
Nice! 

Some of the issues raised by Brakeman are not much of a problem (basically everything in app/controllers/support_interface is fine). Is there a way to  be more specific about where Brakeman looks, or to disable it in certain directories?

raamSoftwire:
Unfortunately there isn't functionality out of the box for skipping entire directories. But you can skip individual files.

`brakeman --skip-files file1,file2,etc`

Brakeman does “whole program” analysis, therefore skipping a file may affect warning results from more than just that one file.

The docs also warn that the inverse of only telling it which files to include is "dangerous" since Brakeman looks at the whole program, it is very likely not going to behave as expected when scanning a subset of files. Also, if certain files are excluded Brakeman may not function at all. 

I think we have 3 options 

- skip the two offending files in the support interface by naming them explicitly
- writing something custom to skip the entire directory eg. [ls app/some/dir/ | paste -s -d , - | xargs brakeman --skip-files](https://stackoverflow.com/questions/19188900/how-to-make-brakeman-ignore-certain-paths). I would need some help here to pipe this in to the yaml file
- fix the issues _properly_ by scoping the queries

raamSoftwire:
It looks like the docs I was reading were out of date. There is a neat way to exclude directories! 

I'll take a look at the remaining `provider_interface` cases tomorrow.

davidgisbey:
My vote would be to use the ignore file 👍

raamSoftwire:
I took at a look at the instances of unscoped finds in the provider interface and I _think_ it is reasonable to ignore these since they can't be scoped sensibly (from my initial understanding of the domain model and the parameters that are being passed).

config/brakeman.ignore
```diff
@@ -0,0 +1,146 @@
+{
+  "ignored_warnings": [
+    {
+      "warning_type": "Unscoped Find",
+      "warning_code": 82,
+      "fingerprint": "07081676ffe0bf0ef753045ae9dcc86c848df5f35a9e48a22692159d5352fbd1",
+      "check_name": "UnscopedFind",
+      "message": "Unscoped call to `ProviderRelationshipPermissions#find`",
+      "file": "app/controllers/provider_interface/provider_relationship_permissions_controller.rb",
+      "line": 32,
```
tvararu, line 10: What happens to this if the code gets moved around? Can you try it locally by shifting the `def` a few lines and see how Brakeman reacts, does it keep this file updated for us or is it a bit of a pain?

tvararu (approved):
LGTM assuming the provider interface ignores are good :+1:

config/brakeman.ignore
raamSoftwire: It looks pretty robust. I swapped two methods around and it was able to keep up. I think it actually looks at the `location` and `code` keys rather that the `line`. 

```    {
      "warning_type": "Unscoped Find",
      "warning_code": 82,
      "fingerprint": "4c106936481ffa09c60383aed2a7b8e931b1672602d3f5a69ccaf9bd688cbf62",
      "check_name": "UnscopedFind",
      "message": "Unscoped call to `ProviderRelationshipPermissions#find`",
      "file": "app/controllers/provider_interface/provider_relationship_permissions_setup_controller.rb",
      "line": 40,
      "link": "https://brakemanscanner.org/docs/warning_types/unscoped_find/",
      "code": "ProviderRelationshipPermissions.find(params[:id])",
      "render_path": null,
      "location": {
        "type": "method",
        "class": "ProviderInterface::ProviderRelationshipPermissionsSetupController",
        "method": "save_permissions"
      },
      "user_input": "params[:id]",
      "confidence": "Weak",
      "note": ""
    },

**Comment by raamSoftwire** [comment]
I've also just added a doc for how to update the `brakeman.ignore` file in future.

File: app/controllers/candidate_interface/other_qualifications/type_controller.rb [hunk 2]
```diff
@@ -15,7 +15,7 @@ def create
     end
 
     def edit
-      @qualification_type = OtherQualificationTypeForm.build_from_qualification(ApplicationQualification.find(params[:id]))
+      @qualification_type = OtherQualificationTypeForm.build_from_qualification(current_qualification)
```
tijmenb, line 18: 😮

PR #3303, opened 2020-11-02
Use govuk_markdown gem to render markdown

tijmenb:
## Context

We render some content that's written in Markdown. Some other systems in DfE do this as well (https://github.com/DFE-Digital/get-help-with-tech), so it made sense to push this functionality out into a gem. It hopefully will save service teams a little bit of time in the future. 

Replace our code with the gem. Everything should keep looking the same.

Unfortunately we can't yet delete all the `app-styled-content`, as some unrelated HTML still uses this. 

## Guidance to review

Does it render the same?


app/views/api_docs/reference/_properties_list.html.erb
```diff
@@ -29,7 +29,7 @@
         <% end %>
 
         <% if property.attributes.description %>
-          <div class="app-styled-content govuk-!-margin-top-4">
+          <div class="govuk-!-margin-top-4">
```
paulrobertlloyd, line 32: Looking at the API docs, I think you can get away with not having this extra `div` with the margin utility class on, but instead add padding bottom to `.app-api-metadata` instead. Saves having two nested `div`s.

Gemfile
```diff
@@ -10,10 +10,10 @@ gem 'pg', '~> 1.2.3'
 gem 'webpacker'
 gem 'govuk_design_system_formbuilder', '~> 2.0.0'
 
-# GovUK Notify
+# GOV.UK Notify
```
tvararu, line 13: Good branding fix

tvararu (approved):

Gemfile
tijmenb: That comment just hurt my soul.

PR #3306, opened 2020-11-02
Add action_taken column to UCAS matches table

AgaDufrat:
## Context

We need to be able to record what action was taken to resolve dual applications. The actions could be:
- 'initial_emails_sent'
- reminder_emails_sent'
- 'requested_removal_from_ucas'

We don't want to store it in the `matching_state` column as it will get overwritten if a match is updated eg with a new status on UCAS.

Add a new column

## Guidance to review

Is the column name ok?

## Link to Trello card

## Things to check


steventux (approved):
:+1:

PR #3308, opened 2020-11-02
Add permissions data for active provider users export

benswannack:
## Context
A bit of the work for the ticket I'm working on below. Add permission data about users for each provider for performance analysis

There is now one row per user/provider pair, rather than just per user. No info has been removed from the export.

## Guidance to review
Should be quite straightforward!

## Link to Trello card
https://trello.com/c/A5wsjz7z/2919-create-downloadable-csvs-for-access-control-performance-analysis

## Things to check

### DIFF
--- a/app/models/data_export.rb
+++ b/app/models/data_export.rb
@@ -40,6 +40,11 @@ class DataExport < ApplicationRecord
       description: 'The list of provider users that have signed in to apply at least once.',
       class: SupportInterface::ActiveProviderUsersExport,
     },
+    active_provider_user_permissions: {
+      name: 'Active provider user permissions',
+      description: 'The list of provider users with the permissions they have for each of their organisations.',
+      class: SupportInterface::ActiveProviderUserPermissionsExport,
+    },
     course_choice_withdrawal: {
       name: 'Candidate course choice withdrawal survey',
       description: 'A list of candidates explanations for withdrawing a course choice. Also includes contact details for candidates who have agreed to be contacted.',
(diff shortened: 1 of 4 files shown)


app/services/support_interface/active_provider_users_export.rb
```diff
@@ -4,13 +4,21 @@ def data_for_export
...
+            provider: provider.name,
+            last_signed_in_at: provider_user.last_signed_in_at,
+            has_make_decisions: permissions.make_decisions.exists?(provider: provider),
+            has_view_safeguarding: permissions.view_safeguarding_information.exists?(provider: provider),
+            has_view_diversity: permissions.view_diversity_information.exists?(provider: provider),
+            has_manage_users: permissions.manage_users.exists?(provider: provider),
+            has_manage_organisations: permissions.manage_organisations.exists?(provider: provider),
+          }
+        end
+      end.flatten
```
benswannack, line 21: I'm a bit sad that I have to flatten here. Any ideas for making this clearer?

duncanjbrown:
> There is not one row per user/provider pair, rather than just per user. No info has been removed from the export.

Very good point! Should we revisit this approach?

benswannack:
> > There is not one row per user/provider pair, rather than just per user. No info has been removed from the export.
> 
> Very good point! Should we revisit this approach?

Oh whoops that was a typo: there is *now one row per pair. Is that fine?

duncanjbrown (approved):
📐

PR #3312, opened 2020-11-02
Redirect back to reference review page if reference not found

davidgisbey:
## Context

If a candidate manually tries to edit a reference that is not associated with their application or doesn't exist they trigger the not found and error out respectively. The desired behaviour laid out by Paul in the Trello card is that in both situations they should be redirected to the review page. This is the same behaviour that occurs when a user tries to edit a reference in an invalid state.

- Redirect to the review page if a user tries to manually edit another candidates reference
- Redirect to the review page if a user inputs an id of a reference into the query string that doesn't exist in the db

## Link to Trello card

https://trello.com/c/TPXBtAdm/2385-%F0%9F%92%94-dont-allow-edits-to-references-that-have-already-been-requested

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/references/name_controller.rb
+++ b/app/controllers/candidate_interface/references/name_controller.rb
@@ -1,7 +1,7 @@
 module CandidateInterface
   module References
     class NameController < BaseController
-      before_action :set_reference, :redirect_to_review_page_unless_reference_is_not_requested_yet
+      before_action :set_reference, :redirect_to_review_page_unless_reference_is_editable
 
       def new
         @reference_name_form = Reference::RefereeNameForm.new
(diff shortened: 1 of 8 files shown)


app/controllers/candidate_interface/references/name_controller.rb
```diff
@@ -1,11 +1,10 @@
...
     class NameController < BaseController
-      before_action :set_reference, :redirect_to_review_page_unless_reference_is_not_requested_yet
-
-      def new
-        @reference_name_form = Reference::RefereeNameForm.new
-      end
+      before_action :set_reference, :redirect_to_review_page_unless_reference_is_editable,
+                    def new
+                      @reference_name_form = Reference::RefereeNameForm.new
+                    end
```
malcolmbaig, line 7: whitespace is a bit off here

davidgisbey: rubocop'd 👮‍♂️ thanks!

tvararu (approved):

paulrobertlloyd (approved):

PR #3316, opened 2020-11-03
Add a service to handle adding a course after application submission

steventux:
## Context

We have a support request to append a course to a submitted application, this isn't something trivial to do in the console.

- Adds a service class to handle creating a new application choice for the course option provided and associates with the submitted application.
- Notifies the provider of the application.

## Things to check

### DIFF
--- a/app/services/support_interface/add_course_choice_after_submission.rb
+++ b/app/services/support_interface/add_course_choice_after_submission.rb
@@ -1,0 +1,22 @@
+module SupportInterface
+  class AddCourseChoiceAfterSubmission
+    attr_reader :application_form, :course_option
+
+    def initialize(application_form:, course_option:)
+      @application_form = application_form
+      @course_option = course_option
+    end
+
+    def call
+      application_choice = ApplicationChoice.create!(
+        application_form: application_form,
+        course_option: course_option,
+        status: 'unsubmitted',
+      )
+
+      SendApplicationToProvider.call(application_choice)
+
+      application_choice
+    end
+  end
+end
--- a/spec/services/support_interface/add_course_choice_after_submission_spec.rb
+++ b/spec/services/support_interface/add_course_choice_after_submission_spec.rb
@@ -1,0 +1,21 @@
+require 'rails_helper'
+
+RSpec.describe SupportInterface::AddCourseChoiceAfterSubmission do
+  describe '#call' do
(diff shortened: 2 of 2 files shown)


app/services/support_interface/add_course_choice_after_submission.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+
+    def call
+      application_choice = ApplicationChoice.create!(
+        application_form: application_form,
+        course_option: course_option,
+        status: 'awaiting_provider_decision',
+      )
+
+      SendApplicationToProvider.call(application_choice) if send_to_provider
```
tijmenb, line 18: I think this shouldn't be optional:

```suggestion
      SendApplicationToProvider.call(application_choice)
```

app/services/support_interface/add_course_choice_after_submission.rb
```diff
@@ -0,0 +1,23 @@
...
+      @application_form = application_form
+      @course_option = course_option
+      @send_to_provider = send_to_provider
+    end
+
+    def call
+      application_choice = ApplicationChoice.create!(
+        application_form: application_form,
+        course_option: course_option,
+        status: 'awaiting_provider_decision',
```
tijmenb, line 15: `SendApplicationToProvider` changes the state, so perhaps this should be `unsubmitted` to avoid errors.

tijmenb (commented):
This looks good. Compared to the "normal" flow we 1) don't update the `submitted_at`, 2) send an email to the candidate, both of which are fine. For the rest it should all work the same.

app/services/support_interface/add_course_choice_after_submission.rb
steventux: My bad, read https://github.com/DFE-Digital/apply-for-teacher-training/blob/90d463513e64b35934bb84e19ace5afbff606b7c/app/services/send_application_to_provider.rb#L15 as an `if` :+1:

steventux:
@tijmenb thanks for review amended as advised: https://github.com/DFE-Digital/apply-for-teacher-training/pull/3316/commits/55cce7ecc7cbd6914c3e0b0ca06bd45eb811989c

tijmenb (approved):
👍

tvararu (approved):
Simples!

PR #3317, opened 2020-11-03
Add tzinfo-data gem

duncanjbrown:
Without this gem, `TZInfo`, which knows about which time zone we’re in,
will fall back to the system `zoneinfo` file. We’ve seen this file be in
a weird state inside containers eg where timezone data is incomplete, so
we want to insulate ourselves from that kind of inconsistency.

## Context

https://ukgovernmentdfe.slack.com/archives/CQA64BETU/p1604340589000400


tvararu (approved):
:crystal_ball:

tijmenb (approved):

PR #3318, opened 2020-11-03
Add automatic attribution to audited changes in console

tijmenb:
## Context

We sometimes make changes via the console. Half the time we remember to set the `audit_user` so the audit log has the correct attribution.

On development nothing changes:

Screenshot 2020-11-03 at 12 31 09

On deployed environments you'll have to identify yourself:

Screenshot 2020-11-03 at 12 34 06

With production being _RED_ to avoid confusion. 

Screenshot 2020-11-03 at 12 33 45

When you provide a name the audit log will automatically added to the log:

image

## Guidance to review

Is this a dumb idea? Hijacking the Rails console feels weird. I got the idea [from some random medium post](https://medium.com/@sulabhjain/how-to-extend-rails-console-on-initialization-14ac79011332), which is usually the quickest way to organise a production incident.

Unfortunately we don't have access to a username or similar inside the container (I've checked all vars in `ENV`), so hence we need to `gets` the name.

### DIFF
--- a/config/initializers/console.rb
+++ b/config/initializers/console.rb
@@ -1,0 +1,24 @@
+module ApplyConsole
+  def start
+    show_warning_message_about_environments
+    super
+  end
+
+  def show_warning_message_about_environments
+    if HostingEnvironment.production?
+      puts ("*" * 50).red
+      puts "** You are in the Rails console for PRODUCTION! **".red
+      puts ("*" * 50).red
+    else
+      puts ("-" * 65).blue
+      puts "-- This is the Rails console for the #{HostingEnvironment.environment_name} environment. --".blue
+      puts ("-" * 65).blue
+    end
+  end
+end
+
+module Rails
+  class Console
+    prepend ApplyConsole
+  end
+end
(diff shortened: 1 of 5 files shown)


config/initializers/console.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    if HostingEnvironment.development?
+      super
+    else
+      puts 'Hello! Who are you? This name will be used in the audit log for any changes you make.'
+      who_are_you = $stdin.gets
+      audited_user = "#{who_are_you.chomp} via the Rails console"
+      puts "Any updates to models will be attributed in the audit logs to #{audited_user.inspect}"
+
+      Audited.audit_class.as_user(audited_user) do
```
duncanjbrown, line 13: Very good!

duncanjbrown (approved):
I love it.

Our convention so far has been to use emails rather than names in the audit log — should we encourage that?

tvararu (approved):
Very nice!

malcolmbaig (approved):
nice :ok_hand:

PR #3319, opened 2020-11-03
Update data export labels for surveys/feedback

stevehook:
## Context

We are replacing the candidate survey with a new single-page feedback form. You will be able to export data from both for the foreseeable future in the support interface (even after the feature flag for this is switched on). This PR just updates the copy on the data export links in an attempt to make it a little clearer which is the old and which is the new one.

image

## Guidance to review

- Suggestions for better labels welcome.

Original card:
https://trello.com/c/lQdpvtmy/2281-dev-replace-multi-step-survey-with-single-page-feedback-form

## Things to check

### DIFF
--- a/app/models/data_export.rb
+++ b/app/models/data_export.rb
@@ -27,12 +27,12 @@ class DataExport < ApplicationRecord
     },
     candidate_survey: {
       name: 'Candidate survey',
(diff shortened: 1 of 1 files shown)


paulrobertlloyd (approved):
Labels look good to me. Noticed some inconsistent punctation across these labels though, so fixed that.

PR #3321, opened 2020-11-03
Move candidate survey link to the application submitted email

davidgisbey:
## Context

When candidates submit an application we already prompt them to complete the equality and diversity questionnaire. Prompting them to complete a satisfaction survey might be excessive. 

So we're adding a link to the [Application submitted](https://qa.apply-for-teacher-training.service.gov.uk/rails/mailers/candidate_mailer/application_submitted) notification prompting users to fill in the satisfaction survey.  

- Add a link to the application submitted email which has path=x in the query string. This allows us to retain their choice to complete the survey through the sign in process. This has to be done as going forward this will be the only way that a candidate can access the survey.

## Guidance to review

The final section of the card talks about making the link minimal as possible i.e. https://www.apply-for-teacher-training.service.gov.uk/candidate/survey

Should we be hiding the token & other query string params for the magic_link's in general?

If so is there an easy way to do this? I've had a bit of a google, but I've not seen a solution for ActionMailer.

## Link to Trello card

https://trello.com/c/lGQi3MZX/2401-dev-post-submit-satisfaction-survey-prompt

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/after_sign_in_controller.rb
+++ b/app/controllers/candidate_interface/after_sign_in_controller.rb
@@ -24,9 +24,7 @@ module CandidateInterface
   private
 
     def redirect_to_path_if_path_params_are_present_and_valid
-      return false if params[:path].blank? || valid_paths.exclude?(params[:path])
-
-      redirect_to path_params
+      redirect_to path_params if params[:path].present? || valid_paths.include?(params[:path])
     end
 
     def redirect_to_application_form_unless_course_from_find_is_present
--- a/app/controllers/candidate_interface/after_sign_in_controller.rb
+++ b/app/controllers/candidate_interface/after_sign_in_controller.rb
@@ -1,5 +1,6 @@
 module CandidateInterface
   class AfterSignInController < CandidateInterfaceController
+    before_action :redirect_to_path_if_path_params_are_present_and_valid
     before_action :redirect_to_application_form_unless_course_from_find_is_present
 
     def interstitial
@@ -22,6 +23,12 @@ module CandidateInterface
 
   private
 
+    def redirect_to_path_if_path_params_are_present_and_valid
+      return false if params[:path].blank? || valid_paths.exclude?(params[:path])
+
+      redirect_to path_params
+    end
+
     def redirect_to_application_form_unless_course_from_find_is_present
       return false unless course_from_find.nil?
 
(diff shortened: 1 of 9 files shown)


app/controllers/candidate_interface/after_sign_in_controller.rb
```diff
@@ -22,6 +23,12 @@ def interstitial
 
   private
 
+    def redirect_to_path_if_path_params_are_present_and_valid
+      return false if params[:path].blank? || valid_paths.exclude?(params[:path])
+
+      redirect_to path_params
```
davidgisbey, line 29: Brakeman is complaining about this method, but it is protected via the valid_paths logic above.

TobyRet: I wonder in that case if this logic needs to be reversed so that it whitelists valid paths?

TobyRet (commented):
LGMT - just one comment re: the brakeman warning

stevehook (approved):
Tried the review app and I was 'Very satisfied'.

I had a bit of a problem at first when I messed up the link and then I had to start again because it had expired. It just made me wonder about magic links that are use once and can't be re-requested. Seemed like a little bit of a loose end.

Did you consider using the raw path `/candidate/application/feedback-form` rather than the name of the helper in the `path` param?

davidgisbey:
I originally used to the raw path but ran into issues when I wanted to check that it was actually one of our endpoints. I couldn't find an easy way to check that is was without implementing something like this (except just with path)

image

and then doing something like
`
routes.map(&:path).include?(params['path'])
`
Is there an easier way to do this?

app/controllers/candidate_interface/after_sign_in_controller.rb
davidgisbey: Yup good point, must have been on autopilot 🤦

stevehook:
> seemed way messier. Is there an easier way to do this?

It’s a tricky one, if we need to validate the path then what you have makes total sense. I don’t have any issues with it, original comment was out of curiosity

I guess the more common way to do deep linking is to link to desired target URL (`/candidate/application/feedback-form?magic-token=xxxx`) and have it redirected through the authentication flow if needed. But I think we already discussed that and plumped for the approach you've taken.

PR #3322, opened 2020-11-03
Refactor work history and volunteering review components and localisation strings

paulrobertlloyd:
## Context

* Removes check icon from summary card header if job/role involved working with children, instead makes it explicit item in summary with ‘Yes’/‘No’ response
* Updates summary row ordering and labels to better match the ordering of fields and their labels on respective question pages
* Uses consistent localisation of strings and string names

| Before | After |
| - | - |
| job-before | job-after |
| role-before | role-after |

## Things to check

### DIFF
--- a/config/locales/application_form.yml
+++ b/config/locales/application_form.yml
@@ -4,8 +4,6 @@ en:
     completed_checkbox: I have completed this section
     continue: Continue
     complete_form_button: Save and continue
-    review:
-      role_involved_working_with_children: This role involved working with children
     courses:
       intro: You can apply for up to 3 courses.
       delete: Delete choice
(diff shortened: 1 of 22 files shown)


config/locales/application_form.yml
```diff
@@ -351,6 +349,9 @@ en:
         hint_text: For example, ‘I volunteer in the classroom every Friday morning’
           or ‘I spent 1 day observing in this school’ or ‘I am a Scout Leader involved
           in activities throughout the year’
+      review_working_with_children:
+        review_label: Did this role involve working in a school or with children?
+        change_action: if this role involve working in a school or with children
```
davidgisbey: involved

davidgisbey (approved):
Adding the date was smart 👍  I've tested it locally and it all seems to be behaving correctly.

config/locales/application_form.yml
paulrobertlloyd: Good catch! 😅

PR #3326, opened 2020-11-04
Improve search for application forms

tijmenb:
## Context

Our current "ILIKE" based search isn't very good, because it doesn't do fuzzy-like matching. For example, if the user is named "Sam Norman Seaborn", a search for "Sam Seaborn" will not return anything.

Luckily, Postgres is a beast and supports very cool full-text search out of the box. To make things even easier we can use the `pg_search` gem, which allows us to create a scope for nice searching.

## Guidance to review

Makes sense?

### DIFF
--- a/spec/services/support_interface/applications_filter_spec.rb
+++ b/spec/services/support_interface/applications_filter_spec.rb
@@ -1,0 +1,32 @@
+require 'rails_helper'
+
+RSpec.describe SupportInterface::ApplicationsFilter do
+  describe '#filter_records' do
+    it 'allows searching by a string' do
+      application_form_in = create(:application_form, first_name: 'Foo', last_name: 'Bar')
+      create(:application_form, first_name: 'Not', last_name: 'Bar')
+
+      expect(results_for(q: 'foo')).to match_array([application_form_in])
+    end
+
+    it 'allows searching by a partial matching string' do
+      application_form_in = create(:application_form, first_name: 'Foo Baz', last_name: 'Bar')
+      create(:application_form, first_name: 'Not', last_name: 'Bar')
+
+      expect(results_for(q: 'Foo Bar')).to match_array([application_form_in])
+    end
+
+    it 'allows searching by a candidate email' do
+      application_form_in = create(:application_form, candidate: create(:candidate, email_address: 'foo@example.com'))
+      create(:application_form)
+
+      expect(results_for(q: 'foo@example.com')).to match_array([application_form_in])
+    end
+  end
+
+  def results_for(params)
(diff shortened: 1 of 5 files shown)


spec/services/support_interface/applications_filter_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+      application_form_in = create(:application_form, first_name: "Foo Baz", last_name: "Bar")
+      application_form_out = create(:application_form, first_name: "Not", last_name: "Bar")
+
+      applications_filter = SupportInterface::ApplicationsFilter.new(params: { q: "Foo Bar" })
+
+      expect(applications_filter.filter_records(ApplicationForm)).to match_array([application_form_in])
+    end
+
+    it "allows searching by a candidate email" do
+      application_form_in = create(:application_form, first_name: "Foo Baz", last_name: "Bar")
```
duncanjbrown, line 24: Any need for these `first_name` and `last_name` here?

duncanjbrown (approved):

davidgisbey (approved):
Nice!

spec/services/support_interface/applications_filter_spec.rb
tijmenb: Good shout, I've updated the tests.

tvararu (approved):

stevehook (approved):
:+1: Great idea. It worked nicely on the review app. I spotted one glitch - if you filter then remove the filter and press the Filter button you get an empty result set. It think it's because there is a `q=` query param left over.
I guess might need to add free-text indexes to keep things speedy as we scale up.

peteryates:
Sorry to jump in here, but having done quite a lot of this in the past, @stevehook is bang on regarding indexing. A [generated column](https://www.postgresql.org/docs/13/ddl-generated-columns.html) that is a [tsvector](https://www.postgresql.org/docs/13/datatype-textsearch.html) generated from your concatenated search fields with a [GIN index](https://www.postgresql.org/docs/13/gin-intro.html) is probably the way to go if you have more than a bit of data.

duncanjbrown:
I'm not sure this is an improvement

Pre this PR if I had a candidate Yung Alterwerth with the email yung.altenwerth@example.com the following search strings worked:

yung.altenwerth
yung.
.altenwerth
example.com (returns a lot, including this candidate)
yung a

None of those strings returns the candidate under `pg_search`.

tijmenb:
Thanks for the feedback @stevehook, @peteryates & @duncanjbrown. I'll close this for now and put something on our backlog.

PR #3329, opened 2020-11-04
Don't crash on audit log when encountering deleted referees

tijmenb:
## Context

We allow deleting of references, and this interferes with audit log.

## Link to Trello card

https://sentry.io/organizations/dfe-bat/issues/2006562442

### DIFF
--- a/spec/components/support_interface/audit_trail_item_component_spec.rb
+++ b/spec/components/support_interface/audit_trail_item_component_spec.rb
@@ -91,6 +91,14 @@ RSpec.describe SupportInterface::AuditTrailItemComponent do
     expect(render_result.text).to include('Harry - harry@hogwarts.edu (Referee)')
   end
 
+  it 'renders an update on application form audit record with a deleted referee' do
+    audit.user_type = 'ApplicationReference'
+    audit.user_id = 412563789101
+    audit.action = 'update'
+
+    expect(render_result.text).to include('Deleted referee')
+  end
+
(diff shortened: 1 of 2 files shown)


stevehook (approved):
Makes sense. Just wondering now whether we delete other kinds of user...

spec/components/support_interface/audit_trail_item_component_spec.rb
```diff
@@ -91,6 +91,15 @@ def render_result
     expect(render_result.text).to include('Harry - harry@hogwarts.edu (Referee)')
   end
 
+  it 'renders an update on application form audit record with a referee' do
+    reference = build(:reference, name: 'Harry', email_address: 'harry@hogwarts.edu')
```
malcolmbaig, line 95: can remove this line

spec/components/support_interface/audit_trail_item_component_spec.rb
```diff
@@ -91,6 +91,15 @@ def render_result
     expect(render_result.text).to include('Harry - harry@hogwarts.edu (Referee)')
   end
 
+  it 'renders an update on application form audit record with a referee' do
```
malcolmbaig, line 94: `deleted referee`

malcolmbaig:
LG, couple of comments left inline

duncanjbrown:
> Makes sense. Just wondering now whether we delete other kinds of user...

Candidates never. Support users can be self-deleted, and Provider Users can't be deleted, only have their privileges removed. So I think we're ok. For users.

PR #3330, opened 2020-11-04
Update error message for other qualification type autocomplete

paulrobertlloyd:
Smol tweak to error message if error showing other UK qualification type autocomplete.


malcolmbaig (approved):
:+1:

PR #3331, opened 2020-11-04
Specify node and yarn versions

frankieroberto:
The current [README](./README) suggests that Node.js version `8.11.x` is being used, but this seems to be out-of-date, as `yarn install` for that version fails due to `@rails/webpacker@5.2.1` requiring a Node.js version of `>=10.17.0`.

This PR specifies required versions of Node.js and Yarn in the `package.json`, which will then throw an early error (with a helpful message) if you run `yarn install` with incompatible versions installed.  The README is updated with a link to this file.

Also adds a `.nvmrc` file for any developers using Node Version Manager (which is a common way to install and run multiple versions of Node.js locally).


tijmenb (approved):
👍 Thank you!

tijmenb:
Actually, this might not work, the Heroku deployment is failing:


```
         

       Bundle completed (69.21s)

       Cleaning up the bundler cache.

-----> Installing node-v12.16.2-linux-x64

-----> Installing yarn-v1.22.4

-----> Detecting rake tasks

-----> Preparing app for Rails asset pipeline

       Running: rake assets:precompile

       yarn install v1.22.4

       [1/5] Validating package.json...

       error apply-for-teacher-training@: The engine "node" is incompatible with this module. Expected version "^12.18.0". Got "12.16.2"

       error Found incompatible module.

       info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.

       W, [2020-11-04T13:39:16.373113 #1632]  WARN -- sentry: ** [Raven] You are running on Heroku but haven't enabled Dyno Metadata. For Sentry's release detection to work correctly, please run `heroku labs:enable runtime-dyno-metadata`

       E, [2020-11-04T13:39:16.520297 #1632] ERROR -- : [SKYLIGHT] [4.3.1] could not load config file; msg=key not found: "SKYLIGHT_AUTH_TOKEN"; disabling Skylight agent

       Compiling...

       Compilation failed:

       yarn run v1.22.4

       info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

       

       

       error Command "webpack" not found.

       

 !

 !     Precompiling assets failed.

 !

 !     Push rejected, failed to compile Ruby app.

 !     Push failed
```

frankieroberto:
@tijmenb ok, so I finally figured this out. The Heroku builds were failing, as the Ruby buildpack installs Node version `12.16.2` by default, and the . The [documentation suggests](https://devcenter.heroku.com/articles/ruby-support#installed-binaries) that to install any other specific version, you need to also specify the `node` buildpack, and make sure that this is specified _before_ the ruby buildpack in the list within `app.json`.

This now fixes that, which means that Heroku and the Azure pipelines are now installing the same version of Node.

Next step might be to bump them both to the current Long Term Support version of `14.*`, but that can wait for a separate pull request...

tvararu (approved):

tijmenb:
Great stuff, thanks @frankieroberto!

PR #3336, opened 2020-11-04
Remove hypenation from full time and part time

paulrobertlloyd:
## Context

Correctly and consistently not use hyphenation when using ‘full time’ and ‘part time’. Here’s one example of how things stand currently:

Screenshot 2020-11-04 at 18 09 40

## Things to check


duncanjbrown (approved):
Brilliant. For max consistency, there's also a spec name in `spec/system/candidate_interface/candidate_submits_application_with_full_course_study_mode_spec.rb` and a comment (!) in `app/services/reason_course_not_available.rb`

paulrobertlloyd:
@duncanjbrown Had noticed those, but I think the rules around hyphenating these words applies in these instances. I’ll let @gregknight1 be the judge – hyphenate or not:

> The location that the candidate picked has no full-time vacancies but does have part-time vacancies

> all part/full-time course options are full for the given course

gregknight1 (commented):
It can be quite tricky in this instance as we have a sentence case where the hyphen should ordinarily be used and the radios where it shouldn't.

I think the best thing to do is to remove the hyphen consistently throughout. As GDS say: 'Do not use a hyphen unless it's confusing without it' - It doesn't add any context here.

malcolmbaig (approved):

PR #3273, opened 2020-10-28
Explain lack of safeguarding access to provider users

mnacos:
### Context

Users who are unable to see safeguarding information due to it being forbidden by an org/org relationship need help to understand why they can't. This is a general need, but has become more pressing now we are onboarding greater numbers of providers, as their provider relationship permissions may not have been setup properly yet.

This PR introduces conditional text to explain the user's lack of access to safeguarding information for an application along the following lines (text provided by @adamsilver):

```
The candidate has disclosed sensitive material related to safeguarding.

{ if you don’t have permission but the relationship allows it }
   You cannot see safeguarding information.
   { if you can manage users }
       [Manage user permissions]
   { else you cannot manage users}
      [BLOCK FOR CONTACTS]
   { endif }
{ endif }

{ if you don’t have permission because the relationship forbids it }
  You have permission to see safeguarding information for the OAKS. However, the OAKS does not have permission to see safeguarding information for applications {ratified/run} by Blackfriars.
{ end }

{ if relationship not set }
    { if you can set them up }
        [Manage organisational permissions].
    { else if you can’t set them up }
        { if there are contacts }
            [BLOCK FOR CONTACTS]
        { else there aren’t any contacts }
             Email support at becomingateacher@digital.education.gov.uk
        { endif }
    { endif }
{ endif }
```

where BLOCK FOR CONTACTS is:

```
{% if just one contact %}
  Contact {{John Smith}} at {{john.smith@example.com}}.
{% else %}
  Contact one the following people:

  - {{John Smith}} at {{john.smith@example.com}}
  - {{Josh Brown}} at {{josh.brown@example.com}}
{% endif %}
```

UPDATE: After a lot of testing/discussion we have decided to avoid prompting the user to change anything if the org relationship has been set up. This includes dropping all 'Manage user permissions' links.

### Changes proposed in this pull request

Change `SafeguardingDeclarationComponent` to perform the relevant checks and display explanations according the current state of user-level permissions, org-level permissions and whether the provider relationship permissions have been set up.

Add a component preview for `SafeguardingDeclarationComponent` (`spec/components/previews/provider_interface/safeguarding_declaration_component_preview.rb`) to simplify development and testing for all these cases.

Add more providers and users to our `setup_local_dev_data` task and set up some of their permissions, so that the component previews have enough test data to work with.

### Guidance to review

[...]

PR #3338, opened 2020-11-05
Refactor all authorisation calls to populate errors array

mnacos:
## Context

This completes what started with https://github.com/DFE-Digital/apply-for-teacher-training/pull/3182 and addresses an edge case:

- Vendor API key makes decisions on applications to courses the relevant provider _ratifies_ (if the org relationship allows it).

This work is also required by https://github.com/DFE-Digital/apply-for-teacher-training/pull/3273 to make `ProviderAuthorisationAnalysis` usable for all permissions.

No UI changes, pure refactoring.

## Guidance to review

Does the refactoring make sense?

## Link to Trello card

No Trello card

## Things to check


app/services/provider_authorisation.rb
```diff
@@ -103,39 +97,72 @@ class NotAuthorisedError < StandardError; end
...
-      )
+  def full_authorisation?(permission:, course:)
+    return true if @actor.is_a?(SupportUser)
+
+    training_provider = course.provider
+    ratifying_provider = course.accredited_provider
+
+    # enforce user-level permissions
+    errors << :requires_provider_user_permission unless
+      @actor.is_a?(VendorApiUser) ||
```
duncanjbrown, line 131: Should we be checking the `ApiUser`'s permissions here too?

duncanjbrown (approved):

app/services/provider_authorisation.rb
mnacos: `VendorApiUser`s do not have user-level permissions, all that matters for them is their provider association and checking the `course_option_id` they supply for `make_decisions` belongs to them. Their provider association, of course, may bring in org-level restrictions.

PR #3340, opened 2020-11-05
Multiple English GCSEs

raamSoftwire:
## Context

English GCSEs currently use free text inputs. This is changing to a more structured approach.

More details:
https://bat-design-history.netlify.app/apply-for-teacher-training/structured-data-for-pre-degree-qualifications/#what-english-gcse-did-you-do
https://ukgovernmentdfe.slack.com/archives/CN1MCQCHZ/p1604478676227000

- [x] add a new feature flag
- [x] create a new multiple English gcse form to use when flag is on
- [x] convert response from form to a JSON blob to store in DB
- [x] update the review page to use new structured data, but fallback to old data if necessary
- [x] add autocomplete to new form when updating
- [x] new form validation and error handling
- [x] make sure old functionality is unaffected when feature flag is off
- [x] handle case where we need to invalidate the old GCSE value and create a new value
- [x] update validation on grades
- [x] ensure the international GCSEs feature is unaffected by changes
- [x] write tests

Also see https://trello.com/c/sLBZws9B/2548-refactor-english-gcses. Where I will refactor English GCSEs to be saved as multiple qualifications instead of as a JSON blob.

New multiple english GCSEs form

image

Review page 

image

Fallback review case

image

## Link to Trello card

https://trello.com/c/HLxaajFd/2428-%F0%9F%8F%88-allow-candidates-to-enter-multiple-english-gcses

## Things to check

### DIFF
--- a/app/components/candidate_interface/gcse_qualification_review_component.rb
+++ b/app/components/candidate_interface/gcse_qualification_review_component.rb
@@ -72,19 +72,12 @@ module CandidateInterface
           "#{grades['chemistry']} (Chemistry)",
           "#{grades['physics']} (Physics)",
         ]
-      elsif application_qualification.subject == 'english'
+      elsif application_qualification.subject == 'english' && application_qualification.grades
         grades = JSON.parse(application_qualification.grades)
         grades.map { |k, v,| "#{v} (#{k.humanize.titleize})" }
       else
         application_qualification.grade
       end
-    end
-
-    def formatted_grades
-      return if application_qualification.grades.nil?
-
-      grades = JSON.parse(application_qualification.grades)
-      grades.map { |k, v,| "#{v} (#{k.humanize.titleize})" }
     end
 
     def missing_qualification_row
--- a/app/components/candidate_interface/gcse_qualification_review_component.rb
+++ b/app/components/candidate_interface/gcse_qualification_review_component.rb
@@ -72,9 +72,19 @@ module CandidateInterface
           "#{grades['chemistry']} (Chemistry)",
           "#{grades['physics']} (Physics)",
         ]
+      elsif application_qualification.subject == 'english'
+        grades = JSON.parse(application_qualification.grades)
+        grades.map { |k, v,| "#{v} (#{k.humanize.titleize})" }
       else
         application_qualification.grade
       end
+    end
+
+    def formatted_grades
+      return if application_qualification.grades.nil?
+
+      grades = JSON.parse(application_qualification.grades)
+      grades.map { |k, v,| "#{v} (#{k.humanize.titleize})" }
     end
 
     def missing_qualification_row
--- a/app/controllers/candidate_interface/gcse/english/grade_controller.rb
+++ b/app/controllers/candidate_interface/gcse/english/grade_controller.rb
@@ -13,7 +13,8 @@ module CandidateInterface
     end
 
     def update
-      if multiple_gsces_are_active?
+      @qualification_type = gcse_english_qualification.qualification_type
+      if multiple_gsces_are_active? && @qualification_type == 'gcse'
         @gcse_grade_form = english_gcse_grade_form.assign_values(english_details_params)
       else
         @qualification_type = gcse_english_qualification.qualification_type
@@ -22,7 +23,7 @@ module CandidateInterface
         @gcse_grade_form.other_grade = english_details_params[:other_grade]
       end
 
-      save_successful = multiple_gsces_are_active? ? @gcse_grade_form.save_grades : @gcse_grade_form.save_grade
+      save_successful = multiple_gsces_are_active? && @qualification_type == 'gcse' ? @gcse_grade_form.save_grades : @gcse_grade_form.save_grade
 
       if save_successful
         update_gcse_completed(false)
--- a/app/controllers/candidate_interface/gcse/english/grade_controller.rb
+++ b/app/controllers/candidate_interface/gcse/english/grade_controller.rb
@@ -6,30 +6,85 @@ module CandidateInterface
     before_action :set_subject
 
     def edit
-      @application_qualification = details_form
-      @qualification_type = details_form.qualification.qualification_type
+      @gcse_grade_form = english_gcse_grade_form
+      @qualification_type = gcse_english_qualification.qualification_type
+
+      render view_path
     end
 
     def update
-      @qualification_type = details_form.qualification.qualification_type
+      if multiple_gsces_are_active?
+        @gcse_grade_form = english_gcse_grade_form.assign_values(english_details_params)
+      else
+        @qualification_type = gcse_english_qualification.qualification_type
+        @gcse_grade_form = english_gcse_grade_form
+        @gcse_grade_form.grade = english_details_params[:grade]
+        @gcse_grade_form.other_grade = english_details_params[:other_grade]
+      end
 
-      details_form.grade = details_params[:grade]
-      details_form.other_grade = details_params[:other_grade]
+      save_successful = multiple_gsces_are_active? ? @gcse_grade_form.save_grades : @gcse_grade_form.save_grade
 
-      @application_qualification = details_form.save_grade
-
-      if @application_qualification
+      if save_successful
         update_gcse_completed(false)
         redirect_to next_gcse_path
       else
-        @application_qualification = details_form
-        track_validation_error(@application_qualification)
-
-        render :edit
+        track_validation_error(@gcse_grade_form)
+        render view_path
       end
     end
 
   private
+
+    def english_details_params
+      params.require(:candidate_interface_english_gcse_grade_form).permit([
+        :english_single_award,
+        :grade_english_single,
+        :english_double_award,
+        :grade_english_double,
+        :english_language,
+        :grade_english_language,
+        :english_literature,
+        :grade_english_literature,
+        :english_studies_single_award,
+        :grade_english_studies_single,
+        :english_studies_double_award,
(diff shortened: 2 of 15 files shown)


george-bit (changes requested):
Looks good to me so far, Are you just focusing on removing the feature flag from the front end? There's a fair few files in the backend that it needs to be removed from too. Happy to show you 👍

app/components/candidate_interface/gcse_qualification_review_component.rb
```diff
@@ -58,12 +58,19 @@ def award_year_row
     def grade_row
       {
         key: 'Grade',
-        value: application_qualification.grade || t('gcse_summary.not_specified'),
+        value: formatted_grades || application_qualification.grade || t('gcse_summary.not_specified'),
```
raamSoftwire, line 61: This will change once I rebase on top of Toby's work.

app/forms/candidate_interface/english_gcse_grade_form.rb
```diff
@@ -0,0 +1,278 @@
...
+                  :english_studies_double_award,
+                  :grade_english_studies_double,
+                  :other_english_gcse,
+                  :other_english_gcse_name,
+                  :grade_other_english_gcse
+
+    validates :grade, presence: true, on: :grade
+    validates :other_grade, presence: true, if: :grade_is_other?
+    validate :validate_grade_format, on: :grade, unless: :is_multiple_gcse? || :new_record?
+    validate :validate_grades, unless: :new_record?, on: :grades, if: :is_multiple_gcse?
```
raamSoftwire, line 31: The validation here is not as clear as it could be. The next card will refactor this class so that English grades are saved in the `grade` column rather than using the `grades` column as a JSON blob.

That change will allow us to simplify this validation quite a lot.

app/forms/candidate_interface/english_gcse_grade_form.rb
```diff
@@ -0,0 +1,278 @@
...
+      qualification.update(grades: grades, grade: nil)
+    end
+
+    def save_grade
+      if !valid?(:grade)
+        log_validation_errors(:grade)
+        return false
+      end
+      qualification.update(grade: set_grade, award_year: award_year)
+    end
```
raamSoftwire, line 158: Again, this will be refactored as part of the next ticket so we an have just the one `save` method instead. Similar to the final state of Toby's `ScienceGcseGradeForm`

spec/forms/candidate_interface/english_gcse_grade_form_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+      let(:form) { subject }
+
+      before { allow(form).to receive(:grade_is_other?).and_return(true) }
+
+      it { is_expected.to validate_presence_of(:other_grade) }
+    end
+
+    context 'when qualification type is GCSE' do
+      context 'multiple GCSEs disabled' do
+        FeatureFlag.deactivate('multiple_english_gcses')
```
raamSoftwire, line 16: I have intentionally left out the activated case here. Since the `EnglishGcseGradeForm` is set to change quite a bit in the next ticket it doesn't seem worth adding tests here just yet. I made a start on these and fell in to difficulties with the validation rules with the `grade` and `grades` contexts, so it will also be easier to do after the refactor.

app/forms/candidate_interface/english_gcse_grade_form.rb
```diff
@@ -0,0 +1,274 @@
+module CandidateInterface
+  class EnglishGcseGradeForm
+    include ActiveModel::Model
+    include ValidationUtils
+
+    attr_accessor :grade,
+                  :award_year,
```
davidgisbey, line 7: I think this can come out.

app/forms/candidate_interface/english_gcse_grade_form.rb
```diff
@@ -0,0 +1,276 @@
...
+      end
+      qualification.update(grades: grades, grade: nil)
+    end
+
+    def save_grade
+      if !valid?(:grade)
+        log_validation_errors(:grade)
+        return false
+      end
+      qualification.update(grade: set_grade)
```
davidgisbey, line 155: Found a little bug while manual testing. I think you need to update grades to nil here otherwise if they can change from GCSE to any other type and see the following 

image

This might already be fixed by your follow up work though!

app/components/candidate_interface/gcse_qualification_review_component.rb
```diff
@@ -115,7 +115,7 @@ def country_row
     end
 
     def naric_statment_row
-      return nil unless FeatureFlag.active?('international_gcses') && application_qualification.qualification_type == 'non_uk'
```
davidgisbey, line 108: As you said Toby's pulling this out 👍

app/controllers/candidate_interface/gcse/english/grade_controller.rb
```diff
@@ -13,7 +13,8 @@ def edit
     end
 
     def update
-      if multiple_gsces_are_active?
+      @qualification_type = gcse_english_qualification.qualification_type
+      if multiple_gsces_are_active? && @qualification_type == 'gcse'
         @gcse_grade_form = english_gcse_grade_form.assign_values(english_details_params)
       else
         @qualification_type = gcse_english_qualification.qualification_type
```
davidgisbey, line 20: This was set on line 16

app/forms/candidate_interface/english_gcse_grade_form.rb
davidgisbey: This will fix the bug I mentioned too 👍

davidgisbey (approved):
There's going to be some issues with merge conflicts due to Toby's PR removing the flag. A couple of minor comments, but I think you're refactor should take care of them. 

Really solid work Raam! Looking forward to the refactor.

app/forms/candidate_interface/english_gcse_grade_form.rb
```diff
@@ -0,0 +1,276 @@
...
+            params[:english_literature] = true
+          end
+
+          if grades.keys.include? 'english_studies_single_award'
+            english_gcses << 'english_studies_single_award'
+            params[:grade_english_studies_single] = grades['english_studies_single_award']
+            params[:english_studies_single_award] = true
+          end
+
+          if grades.keys.include? 'english_studies_double_award'
```
TobyRet, line 86: The double award validation for English studies did not kick in for me.
english_studies_validation

app/components/candidate_interface/gcse_qualification_review_component.rb
```diff
@@ -72,6 +72,9 @@ def present_grades
           "#{grades['chemistry']} (Chemistry)",
           "#{grades['physics']} (Physics)",
         ]
+      elsif application_qualification.subject == 'english' && application_qualification.grades
```
TobyRet, line 75: 1. I switched the English GCSE flag on and updated my GCSE grades. 
2. I then turned the flag off and edited my English GCSE grade. 
3. When I went to the review component it was still showing the multiple grades from 1.

TobyRet: review_component

app/controllers/candidate_interface/gcse/english/grade_controller.rb
```diff
@@ -6,31 +6,87 @@ class Gcse::English::GradeController < Gcse::DetailsController
...
 
     def update
-      @qualification_type = details_form.qualification.qualification_type
-
-      details_form.grade = details_params[:grade]
-      details_form.other_grade = details_params[:other_grade]
+      @qualification_type = gcse_english_qualification.qualification_type
+      if multiple_gsces_are_active? && @qualification_type == 'gcse'
+        @gcse_grade_form = english_gcse_grade_form.assign_values(english_details_params)
+      else
```
TobyRet, line 19: Would it be possible to move the attribute assignment in lines 20 - 23 to the form as you have done on line 18?

app/forms/candidate_interface/english_gcse_grade_form.rb
raamSoftwire: Good spot, and that brings it in line with line 146 where we do a similar thing in `save_grades`.

app/components/candidate_interface/gcse_qualification_review_component.rb
raamSoftwire: I've rebased and resolved all of these changes now.

app/forms/candidate_interface/english_gcse_grade_form.rb
raamSoftwire: Thanks, I've found the typo and resolved that now

app/controllers/candidate_interface/gcse/english/grade_controller.rb
raamSoftwire: Great idea, I think this helps to tidy up the controller. I've pushed the change now

app/components/candidate_interface/gcse_qualification_review_component.rb
raamSoftwire: Thanks for spotting this Toby. This has been fixed by setting `grades` to `nil` when we set `grade`. 

I have just tested it again on the heroku app to verify that it is now working as expected.

davidgisbey: Yea this is a variation of the same bug i found earlier 👍

george-bit (approved):

PR #3343, opened 2020-11-05
Add export for provider access controls

benswannack:
## Context
This enables performance analysis for access controls

Add an export that details various metrics for a provider's permissions

## Guidance to review
Do the specs for the access control service make sense? Have I missed any more edge cases?

Is there a good place to write up what each field means? The headings are hard to name succinctly (in particular `total_org_permissions_changes` which is the number of times the org permissions have been updated where the provider is the trainer - corresponds to the number of changes this provider has made to the org permissions)

See https://docs.google.com/spreadsheets/d/1HeeuU5c-jhCHv01MEngS2wcOSbPIZUJdurKUP5ZV1S0/edit#gid=0 for the specification

## Link to Trello card
https://trello.com/c/A5wsjz7z/2919-create-downloadable-csvs-for-access-control-performance-analysis

## Things to check

### DIFF
--- a/app/services/support_interface/provider_access_controls.rb
+++ b/app/services/support_interface/provider_access_controls.rb
@@ -1,0 +1,76 @@
+module SupportInterface
+  class ProviderAccessControls
+    def initialize(provider)
+      @provider = provider
+    end
+
+    def dsa_signer_email
+      @provider
+        .provider_agreements
+        .data_sharing_agreements
+        .where.not(accepted_at: nil)
+        .order(accepted_at: :desc)
+        .first
+        &.provider_user
+        &.email_address
+    end
+
+    def user_permissions_audits
+      Audited::Audit.where(action: 'update', auditable: @provider.provider_permissions)
+    end
+
+    def user_permissions_last_changed_at
+      user_permissions_audits
+        .order(created_at: :desc)
+        .pick(:created_at)
+    end
+
+    def total_user_permissions_changes
+      user_permissions_audits.count
+    end
+
+    def user_permissions_changed_by
+      user_permissions_audits
+        .map { |audit| audit.user.email_address }
+        .uniq
+    end
+
+    def total_manage_users_users
+      @provider.provider_permissions.where(manage_users: true).count
+    end
+
+    def total_manage_orgs_users
+      @provider.provider_permissions.where(manage_organisations: true).count
+    end
+
+    def org_training_provider_permission_audits
+      Audited::Audit.where(action: 'update', auditable: @provider.training_provider_permissions)
+    end
+
+    def org_permissions_last_changed_at
+      org_training_provider_permission_audits
+        .order(created_at: :desc)
+        .pick(:created_at)
+    end
+
+    def total_org_permissions_changes
+      org_training_provider_permission_audits.count
+    end
+
+    def org_permissions_changed_by
+      org_training_provider_permission_audits
+        .map { |audit| audit.user.email_address }
+        .uniq
+    end
+
+    def total_org_relationships_as_trainer
+      @provider.training_provider_permissions.count
+    end
+
+    def total_org_relationships
+      (@provider.ratifying_provider_permissions + @provider.training_provider_permissions)
+        .uniq
+        .count
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


app/services/support_interface/provider_access_controls_export.rb
```diff
@@ -0,0 +1,96 @@
...
+        .data_sharing_agreements
+        .where.not(accepted_at: nil)
+        .order(accepted_at: :desc)
+        .first
+        .provider_user
+        .email_address
+    end
+
+    def user_permissions_audits(provider)
+      Audited::Audit.where(action: 'update', auditable: provider.provider_permissions)
```
duncanjbrown: We could memoize this for efficiency? `@_user_permissions_audits ||= Audited::Audit...`

app/services/support_interface/provider_access_controls_export.rb
```diff
@@ -0,0 +1,96 @@
...
+    def total_manage_users_users(provider)
+      provider.provider_permissions.where(manage_users: true).count
+    end
+
+    def total_manage_orgs_users(provider)
+      provider.provider_permissions.where(manage_organisations: true).count
+    end
+
+    def org_training_provider_permission_audits(provider)
+      Audited::Audit.where(action: 'update', auditable: provider.training_provider_permissions)
```
duncanjbrown: As above

app/services/support_interface/provider_access_controls_export.rb
```diff
@@ -0,0 +1,96 @@
...
+  private
+
+    def get_dsa_signer_email(provider)
+      provider
+        .provider_agreements
+        .data_sharing_agreements
+        .where.not(accepted_at: nil)
+        .order(accepted_at: :desc)
+        .first
+        .provider_user
```
duncanjbrown: I think this will blow up in production because there are providers without signed DSAs. I wonder if we should scope this export to only cover those with signed DSAs?

app/services/support_interface/provider_access_controls_export.rb
```diff
@@ -0,0 +1,96 @@
...
+    end
+
+    def org_training_provider_permission_audits(provider)
+      Audited::Audit.where(action: 'update', auditable: provider.training_provider_permissions)
+    end
+
+    def org_permissions_last_changed_at(provider)
+      org_training_provider_permission_audits(provider)
+        .order(created_at: :desc)
+        .pick(:created_at)
```
duncanjbrown: `pick`! TIL

spec/services/support_interface/provider_access_controls_export_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+        user_signs_dsa_for_provider(provider_user2, ratifying_provider, 3.months.ago)
+        user_signs_dsa_for_provider(provider_user3, ratifying_provider, 1.month.ago)
+
+        setup_org_permissions(training_provider, ratifying_provider)
+
+        provider_user_sets_their_view_diversity_information_to_true(provider_user1, 1.day.ago)
+        provider_user_sets_org_can_view_diversity_information_to_true(provider_user1, training_provider, 2.days.ago)
+
+        provider_user_sets_their_view_diversity_information_to_true(provider_user2, 10.days.ago)
+        provider_user_sets_their_view_safeguarding_information_to_true(provider_user3, 2.days.ago)
```
duncanjbrown, line 48: Very happy to see this style in service specs!

app/services/support_interface/provider_access_controls.rb
```diff
@@ -0,0 +1,76 @@
+module SupportInterface
+  class ProviderAccessControls
```
duncanjbrown, line 2: Good place to split! This class might have a more descriptive name — `ProviderAccessControlsStats` or sth?

app/services/support_interface/provider_access_controls_export.rb
duncanjbrown: Ah, I see fixed in following commit! On reflection I think it's right to do as you do and leave in all providers, not just DSA signers, as some may be on one side of a relationship with another.

duncanjbrown (approved):
Looks vg to me, nits inline!

PR #3344, opened 2020-11-05
Add a start_date attribute to the API Course object

duncanjbrown:
## Context

This is to support SRS vendors. It enables them to handle courses which
run in the same academic year as the recruitment cycle that feeds them —
for example, a course in January 2021 will be open for applications from
October 2020, ie the 2021 recruitment cycle.

Add the property to the `Course` object in the API. Update the docs.

## Link to Trello card

https://trello.com/c/LyspoB3J/2995-include-course-start-date-in-the-course-object-in-api-responses

## Things to check


mnacos (approved):
Looks good

PR #3345, opened 2020-11-05
Add the ApplicationFeedback table

davidgisbey:
## Context

We need a new model for feedback. 

- Migration to add the ApplicationFeedback table

## Guidance to review

The naming could probs be better. Any ideas?

## Link to Trello card

https://trello.com/c/DrnKcnnW/2466-%F0%9F%93%9D-dev-add-feedback-prompt-find

## Things to check


db/migrate/20201105145325_add_application_feedback_table.rb
```diff
@@ -0,0 +1,16 @@
+class AddApplicationFeedbackTable < ActiveRecord::Migration[6.0]
+  def change
+    create_table :application_feedback do |t|
+      t.string :section
+      t.string :path
+      t.string :page_title
+      t.boolean :issues
+      t.boolean :do_not_understand_section
```
frankieroberto, line 8: This might be worth changing to `understood_section` to avoid the double negative when set to false?

I'd also suggest having them all be in the past tense?

frankieroberto (commented):
Should any of the fields be non-nullable?

db/migrate/20201105145325_add_application_feedback_table.rb
stevehook: agree with @frankieroberto that `section`, `path`, `page_title` and all the booleans should be `null: false`. Normally it makes sense to add a `default: false` for the booleans too.

davidgisbey: Okay, I'll make the changes now. I was just thinking it might create some slightly strange data issues. If they hit 'Yes' there is an issue on the non-js version, then choose not to fill in the feedback, there will be some with `issues: true` and everything else set to false rather than nil. 

Now I'm thinking about it though maybe a GET request and hidden fields for the section, path and page title attributes are the way to go. With my current implementation, when they click `'yes`' that there is an issue it sends a post request with the attributes in the query string, creates the resource and redirects to an edit page.

stevehook (commented):
Looks good, except I'm not sure about the change to `understands_section` - I know it avoids the double negative but does this mean that the checkbox will default to checked? Would that look a bit odd?

davidgisbey:
Yea that's a good point 👍 Putting it back to a double negative

stevehook (approved):
👍

PR #3349, opened 2020-11-06
Allow user to enter multiple Science GCSEs

TobyRet:
## Context
As part of [🏈 Dev: Structured pre-degree qualifications](https://trello.com/c/ztwbkRly/1643-%F0%9F%8F%88-dev-structured-pre-degree-qualifications) we are updating pre-degree fields to use structure data.

- Add a new flow to enable users to enter multiple Science GCSEs
- Update the review component to display Science triple awards (where present)

## Guidance for review
- Welcome initial comments on the work so far, especially where more refactoring is deemed _essential_ before shipping the first iteration
- I would imagine a **general clean up PR** may be required once this new flow (and the one for GSCE English) is shipped.
- See the _Todos_ section below for ticket status
- If you are viewing via the deployed Heroku app then you will need to activate the feature flag 'science gcse awards'

## Gif-o-rama

multi_gcse_science

## Todos

- [x] User can enter single, double or triple award
- [x] Validation error occurs if fields are blank
- [x] Review component shows grades with correct format
- [x] New UI is populated with currently saved values when editing
- [x] Bug - some grade format validations no kicking in
- [x] Handle review and editing of Science GCSE created by original form
- [x] Add autocomplete - waiting to hear if this is still required
- [x] Agree API changes with ProVendor
- [x] Add feature specs and any missing test coverage - May want to create a new component and test validations here rather than within system spec
- [x] Refactor? - form is ugly, grade controller is still a bit fat. 

## Link to Trello card

https://trello.com/c/jJPtfRY3/1779-%F0%9F%8F%88-allow-candidates-to-enter-multiple-science-gcses

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/gcse/science/grade_controller.rb
+++ b/app/controllers/candidate_interface/gcse/science/grade_controller.rb
@@ -1,39 +1,65 @@
 module CandidateInterface
   class Gcse::Science::GradeController < Gcse::DetailsController
     include Gcse::GradeControllerConcern
-
     before_action :redirect_to_dashboard_if_submitted
     before_action :set_subject
 
     def edit
-      @application_qualification = details_form
-      @qualification_type = details_form.qualification.qualification_type
+      @gcse_grade_form = science_gcse_grade_form
+      @qualification_type = gcse_science_qualification.qualification_type
+
+      render view_path
     end
 
     def update
-      @qualification_type = details_form.qualification.qualification_type
+      @gcse_grade_form = science_gcse_grade_form.assign_values(science_details_params)
 
-      details_form.grade = details_params[:grade]
-      details_form.other_grade = details_params[:other_grade]
-
-      @application_qualification = details_form.save_grade
-
-      if @application_qualification
+      if @gcse_grade_form.save
         update_gcse_completed(false)
-        redirect_to next_gcse_path
+        redirect_to next_path
       else
-        @application_qualification = details_form
-        track_validation_error(@application_qualification)
-
-        render :edit
+        track_validation_error(@gcse_grade_form)
+        render view_path
       end
     end
 
   private
 
-    # Required by the GradeControllerConcern
+    def view_path
+      if gcse_qualification? && FeatureFlag.active?(:science_gcse_awards)
+        'candidate_interface/gcse/science/grade/awards_edit'
+      else
+        'candidate_interface/gcse/science/grade/edit'
+      end
+    end
+
+    def next_path
+      if science_gcse_grade_form.award_year.nil?
+        candidate_interface_gcse_details_edit_year_path(subject: @subject)
+      else
+        candidate_interface_gcse_review_path(subject: @subject)
+      end
+    end
+
+    def gcse_qualification?
+      gcse_science_qualification.qualification_type == 'gcse'
+    end
+
     def set_subject
-      @subject = 'science'
+      @subject = ApplicationQualification::SCIENCE
+    end
+
+    def science_details_params
+      params.require(:candidate_interface_science_gcse_grade_form)
+            .permit(%i[gcse_science grade single_award_grade double_award_grade biology_grade chemistry_grade physics_grade])
+    end
+
+    def gcse_science_qualification
+      @gcse_science_qualification ||= current_application.qualification_in_subject(:gcse, @subject)
+    end
+
+    def science_gcse_grade_form
+      @science_gcse_grade_form ||= ScienceGcseGradeForm.build_from_qualification(gcse_science_qualification)
     end
   end
 end
--- a/app/forms/candidate_interface/science_gcse_grade_form.rb
+++ b/app/forms/candidate_interface/science_gcse_grade_form.rb
@@ -1,0 +1,230 @@
+module CandidateInterface
+  class ScienceGcseGradeForm
+    include ActiveModel::Model
+    include ValidationUtils
+
+    attr_accessor :grade,
+                  :grades,
+                  :award_year,
+                  :qualification,
+                  :subject,
+                  :other_grade,
+                  :single_award_grade,
+                  :double_award_grade,
+                  :gcse_science,
+                  :biology_grade,
+                  :physics_grade,
+                  :chemistry_grade
+    validates :other_grade, presence: true, if: :grade_is_other?
+    validates :grade, length: { maximum: 6 }, unless: :international_gcses_flag_active?
+    validate :grade_length
+    validate :grade_format, unless: :new_record?
+    validate :triple_award_grade_format
+
+    class << self
+      def build_from_qualification(qualification)
+        if FeatureFlag.active?('international_gcses') && qualification.qualification_type == 'non_uk'
+          new(
+            grade: qualification.set_grade,
+            other_grade: qualification.set_other_grade,
+            qualification: qualification,
+          )
+        else
+          new(build_params_from(qualification))
+        end
+      end
+
+    private
+
+      def build_params_from(qualification)
+        params = {
+          gcse_science: qualification.subject,
+          subject: qualification.subject,
+          qualification: qualification,
+        }
+
+        case qualification.subject
+        when ApplicationQualification::SCIENCE_SINGLE_AWARD
+          params[:single_award_grade] = qualification.grade
+        when ApplicationQualification::SCIENCE_DOUBLE_AWARD
+          params[:double_award_grade] = qualification.grade
+        when ApplicationQualification::SCIENCE_TRIPLE_AWARD
+          grades = qualification.grades
+          return unless grades
+
+          params[:biology_grade] = grades['biology']
+          params[:chemistry_grade] = grades['chemistry']
+          params[:physics_grade] = grades['physics']
+        else
+          params[:grade] = qualification.grade
+        end
+
+        params
+      end
+    end
+
+    def save
+      unless valid?
+        log_validation_errors(errors.keys.first)
+        return false
+      end
+
+      qualification.update(
+        grade: set_grade,
+        grades: set_triple_award_grades,
+        subject: subject,
+      )
+    end
+
+    def assign_values(params)
+      self.gcse_science = params[:gcse_science]
+      self.grade = set_grade_from(params)
+      self.other_grade = params[:other_grade]
+      self.subject = params[:gcse_science] || ApplicationQualification::SCIENCE
+      self.biology_grade = params[:biology_grade]
+      self.chemistry_grade = params[:chemistry_grade]
+      self.physics_grade = params[:physics_grade]
+      self
+    end
+
+  private
+
+    def set_grade_from(params)
+      case params[:gcse_science]
+      when ApplicationQualification::SCIENCE_SINGLE_AWARD
+        params[:single_award_grade]
+      when ApplicationQualification::SCIENCE_DOUBLE_AWARD
+        params[:double_award_grade]
+      else
+        params[:grade]
+      end
+    end
+
+    def grade_format
+      return if
+        qualification.qualification_type.nil? ||
+          qualification.qualification_type == 'other_uk' ||
+          qualification.qualification_type == 'non_uk' ||
+          grade.nil? ||
+          triple_award?
+
+      if %w[gce_o_level scottish_national_5 gcse].include?(qualification.qualification_type) && subject == ApplicationQualification::SCIENCE
+        qualification_rexp = invalid_grades[qualification.qualification_type.to_sym]
+        errors.add(:grade, :invalid) if grade.match(qualification_rexp)
+      end
+
+      if gsce_qualification_type? && single_award?
+        self.single_award_grade = grade
+        errors.add(:single_award_grade, :invalid) unless SINGLE_GCSE_GRADES.include?(grade.strip.upcase)
+      end
+
+      if gsce_qualification_type? && double_award?
+        self.double_award_grade = grade
+        errors.add(:double_award_grade, :invalid) unless DOUBLE_GCSE_GRADES.include?(grade.delete(' ').upcase)
+      end
+    end
+
+    def triple_award_grade_format
+      return unless triple_award?
+
+      grade_hash = {
+        biology_grade: biology_grade,
+        chemistry_grade: chemistry_grade,
+        physics_grade: physics_grade,
+      }
+
+      grade_hash.each do |key, grade|
+        next if grade.blank?
+
+        public_send("#{key}=", grade)
+        errors.add(key, :invalid) unless SINGLE_GCSE_GRADES.include?(grade.strip.upcase)
+      end
+    end
+
+    def grade_length
+      errors.add(:grade, :blank) if grade.blank? && science?
+      errors.add(:single_award_grade, :blank) if grade.blank? && single_award?
+      errors.add(:double_award_grade, :blank) if grade.blank? && double_award?
+      errors.add(:biology_grade, :blank) if biology_grade.blank? && triple_award?
+      errors.add(:chemistry_grade, :blank) if chemistry_grade.blank? && triple_award?
+      errors.add(:physics_grade, :blank) if physics_grade.blank? && triple_award?
+    end
+
+    def invalid_grades
+      {
+        gcse: /[^1-9A-GU\*\s\-]/i,
+        gce_o_level: /[^A-EU\s\-]/i,
+        scottish_national_5: /[^A-D1-7\s\-]/i,
+      }
+    end
+
+    def log_validation_errors(field)
+      return unless errors.key?(field)
+
+      error_message = {
+        field: field.to_s,
+        error_messages: errors[field].join(' - '),
+        value: grade || grades,
+      }
+
+      Rails.logger.info("Validation error: #{error_message.inspect}")
+    end
+
+    def set_grade
+      return if triple_award?
+
+      case grade
+      when 'other'
+        other_grade
+      when 'not_applicable'
+        'N/A'
+      when 'unknown'
+        'Unknown'
+      else
+        grade
+      end
+    end
+
+    def set_triple_award_grades
+      if triple_award?
+        {
+          biology: biology_grade,
+          physics: physics_grade,
+          chemistry: chemistry_grade,
+        }
+      end
+    end
+
+    def international_gcses_flag_active?
+      FeatureFlag.active?('international_gcses')
+    end
+
+    def new_record?
+      qualification.nil?
+    end
+
+    def grade_is_other?
+      grade == 'other'
+    end
+
+    def triple_award?
+      subject == ApplicationQualification::SCIENCE_TRIPLE_AWARD
+    end
+
+    def double_award?
+      subject == ApplicationQualification::SCIENCE_DOUBLE_AWARD
+    end
+
+    def single_award?
+      subject == ApplicationQualification::SCIENCE_SINGLE_AWARD
+    end
+
+    def science?
+      subject == ApplicationQualification::SCIENCE
+    end
+
+    def gsce_qualification_type?
+      qualification.qualification_type == 'gcse'
+    end
+  end
+end
--- a/app/forms/candidate_interface/science_gcse_grade_form.rb
+++ b/app/forms/candidate_interface/science_gcse_grade_form.rb
@@ -115,12 +115,12 @@ module CandidateInterface
 
       if gsce_qualification_type? && single_award?
         self.single_award_grade = grade
-        errors.add(:single_award_grade, :invalid) unless SINGLE_GCSE_GRADES.include?(grade.strip.upcase)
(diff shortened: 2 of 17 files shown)


app/controllers/candidate_interface/gcse/science/grade_controller.rb
```diff
@@ -1,39 +1,108 @@
...
-
-      @application_qualification = details_form.save_grade
+      @gcse_grade_form = science_gcse_grade_form.tap do |form|
+        form.subject = science_details_params[:gcse_science] || @subject
+        form.grade = assign_grade
+        form.biology_grade = science_details_params[:biology_grade]
+        form.physics_grade = science_details_params[:physics_grade]
+        form.chemistry_grade = science_details_params[:chemistry_grade]
+        form.gcse_science = science_details_params[:gcse_science]
+      end
```
davidgisbey, line 22: Is there a reason we're not passing the params to the model and having it deal with munge all the data? I feel like we could pull quite a bit of logic out of the controller.

app/controllers/candidate_interface/gcse/science/grade_controller.rb
```diff
@@ -1,39 +1,108 @@
...
+      if single_award?
+        science_details_params[:single_award_grade]
+      elsif double_award?
+        science_details_params[:double_award_grade]
+      else
+        science_details_params[:grade]
+      end
+    end
+
+    def assign_subject
```
davidgisbey, line 88: Is this being used anywhere? It looks like the build_from_qualification method uses a case statement with the same logic inside the form.

app/forms/candidate_interface/science_gcse_grade_form.rb
```diff
@@ -0,0 +1,236 @@
...
+        qualification.update(grade: set_grade, award_year: award_year)
+        return true
+      end
+
+      false
+    end
+
+  private
+
+    def award_year_is_not_in_the_future
```
davidgisbey, line 110: It might be too early for this, but thoughts on pulling these validations into a separate module/lib? They will/should be reused throughout all qualification validations. It would also makes this file a bit easier to parse.

spec/forms/candidate_interface/science_gcse_grade_form_spec.rb
```diff
@@ -0,0 +1,276 @@
...
+        details_form.other_grade = 'D'
+
+        details_form.save_grades
+        qualification.reload
+
+        expect(qualification.grade).to eq('D')
+      end
+    end
+
+    describe '#save_year' do
```
davidgisbey, line 195: Should we have save_year logic in here? That should still currently be handled by the details from right? Which we should now change to year form (after this piece of work)

app/forms/candidate_interface/science_gcse_grade_form.rb
```diff
@@ -0,0 +1,236 @@
...
+                  :physics_grade,
+                  :chemistry_grade
+    validates :biology_grade, presence: true, on: :grades, if: :triple_award?
+    validates :chemistry_grade, presence: true, on: :grades, if: :triple_award?
+    validates :physics_grade, presence: true, on: :grades, if: :triple_award?
+    validates :other_grade, presence: true, if: :grade_is_other?
+    validates :award_year, presence: true, on: :award_year
+    validates :grade, length: { maximum: 6 }, unless: :international_gcses_flag_active?, on: :grade
+    validate :grade_length, on: :grade
+    validate :award_year_is_a_valid_date, if: :award_year, on: :award_year
```
davidgisbey, line 26: See comment below about award year stuff

davidgisbey:
I've tested it locally and it is looking really good  👍  I've left some comments about potential refactors and I'm not 100% convinced that the award_year related behaviour should be living in a grade form. I think that we should be decoupling grade and award year After this is done we should be refactoring the @details_from to @award_year_form.

app/forms/candidate_interface/science_gcse_grade_form.rb
TobyRet: Yep good spot - I'll remove

spec/forms/candidate_interface/science_gcse_grade_form_spec.rb
TobyRet: Yep I'll remove

app/forms/candidate_interface/science_gcse_grade_form.rb
```diff
@@ -0,0 +1,236 @@
...
+          errors.add(:physics_grade, :blank)
+        end
+      elsif grade.blank?
+        errors.add(:grade, :blank)
+      end
+    end
+
+    def invalid_grades
+      {
+        gcse: /[^1-9A-GU\*\s\-]/i,
```
raamSoftwire, line 156: Should we be using a set or enum for this instead? Especially if we move away from the autocomplete to text inputs.

I've tested locally and this would let through E*

config/locales/application_form.yml
```diff
@@ -803,6 +803,33 @@ en:
...
+              blank: Enter your grade
+            physics_grade:
+              invalid: Enter a real grade
+              blank: Enter your grade
+            award_year:
+              blank: Enter the year you gained your qualification
+              invalid: Enter a real year
+              in_future: Enter a year before %{date}
+            other_grade:
+              blank: Enter your grade
```
raamSoftwire, line 832: This file is very long. Would it be worth starting to split it out? Or are we intending to keep all the `application_form` strings in one place?

app/controllers/candidate_interface/gcse/science/grade_controller.rb
TobyRet: Yep that's old code that I forgot to delete - will delete

app/controllers/candidate_interface/gcse/science/grade_controller.rb
TobyRet: Yeah was thinking about that also ...
Possibly the way to go would be to have something like `ScienceGcseGradeForm.assign_attributes_from(params)`
I'll have a play ...

app/forms/candidate_interface/science_gcse_grade_form.rb
TobyRet: Yes that's a good point. I do think we need to check what the situation is with international GCSE equivalents and validation. It appears to use this regex as well. Even if we do have autocomplete, we will need validations in place for progressive enhancement (i.e no JS turned on)

tvararu: I think I prefer enum as it's a lot clearer what is and is not allowed :thinking:

config/locales/application_form.yml
tvararu: @raamSoftwire https://trello.com/c/g7vXu7sI/367-low-localisations-should-match-structure-of-application

app/forms/candidate_interface/science_gcse_grade_form.rb
TobyRet: I've removed the regex check and now checking that the submitted grade is included in the values defined here - https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/config/initializers/gcse_grades.rb

frankieroberto:
Can we make the text inputs smaller? eg perhaps "2 character width" for single grades, "4 character width" for double grades and "5 character width" for triple grades? See [Design System fixed width inputs](https://design-system.service.gov.uk/components/text-input/#fixed-width-inputs) for the classes.

config/locales/application_form.yml
```diff
@@ -803,6 +803,33 @@ en:
               in_future: Enter a year before %{date}
             other_grade:
               blank: Enter your grade
+        candidate_interface/science_gcse_grade_form:
+          attributes:
+            single_award_grade:
+              blank: Enter your grade
+              invalid: Enter a real grade
```
raamSoftwire, line 810: Has this copy from the designs? Should this be "Enter a valid grade"?

app/views/candidate_interface/gcse/science/grade/awards_edit.html.erb
```diff
@@ -0,0 +1,37 @@
...
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with model: @gcse_grade_form, url: candidate_interface_gcse_science_edit_grade_path, method: :patch do |f| %>
+      <%= f.govuk_error_summary %>
+
+      <h1 class="govuk-heading-xl"><%= t('gcse_edit_grade.page_title', subject: @subject) %></h1>
+
+      <%= render CandidateInterface::GcseGradeGuidanceComponent.new(@subject, @qualification_type) %>
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+          <%= f.govuk_radio_buttons_fieldset :gcse_science, legend: { text: "Select the GCSEs you did and include your grade"} do %>
```
raamSoftwire, line 14: Should these strings be pulled out in to a locales yml file?

app/controllers/candidate_interface/gcse/science/grade_controller.rb
```diff
@@ -1,39 +1,65 @@
...
+      render view_path
     end
 
     def update
-      @qualification_type = details_form.qualification.qualification_type
+      @gcse_grade_form = science_gcse_grade_form.assign_values(science_details_params)
 
-      details_form.grade = details_params[:grade]
-      details_form.other_grade = details_params[:other_grade]
+      if @gcse_grade_form.save_grades
```
malcolmbaig, line 17: Could just call this method `save`

app/forms/candidate_interface/science_gcse_grade_form.rb
```diff
@@ -0,0 +1,218 @@
...
+    validate :grade_length, on: :grade
+    validate :validate_grade_format, unless: :new_record?, on: :grade
+    validate :triple_award_grades_format, if: :triple_award?, on: :grade
+
+    class << self
+      def build_from_qualification(qualification)
+        if FeatureFlag.active?('international_gcses') && qualification.qualification_type == 'non_uk'
+          new(
+            grade: qualification.set_grade,
+            other_grade: qualification.set_other_grade,
```
malcolmbaig, line 32: Would be a good idea to move these `set_` methods into this form and give them a different name. They aren't used anywhere else, and neither of them set anything - they just map db values to radio button values as far as I can tell.

app/controllers/candidate_interface/gcse/science/grade_controller.rb
```diff
@@ -1,39 +1,65 @@
...
-      if @application_qualification
-        update_gcse_completed(false)
-        redirect_to next_gcse_path
+    def view_path
+      if gcse_qualification? && FeatureFlag.active?(:science_gcse_awards)
+        'candidate_interface/gcse/science/grade/awards_edit'
       else
-        @application_qualification = details_form
-        track_validation_error(@application_qualification)
+        'candidate_interface/gcse/science/grade/edit'
```
malcolmbaig, line 32: Any reason not to use path helpers here?

config/locales/application_form.yml
paulrobertlloyd: @tvararu Was just going to enthusiastically comment on that card – until I realised that I wrote it! 😂

config/locales/application_form.yml
TobyRet: This is the error message used in the current / old flow.  @paulrobertlloyd can you confirm?

app/views/candidate_interface/gcse/science/grade/awards_edit.html.erb
TobyRet: 👍

app/controllers/candidate_interface/gcse/science/grade_controller.rb
TobyRet: 👍

app/forms/candidate_interface/science_gcse_grade_form.rb
TobyRet: Yep good idea

TobyRet: 😞 These `set_` methods are in fact been called by the `GcseQualificationDetailsForm` which is still used in a number of places. I think we should deal with this in a ‘cleanup’ PR once the multi GCSE English and Science flow is shipped

app/controllers/candidate_interface/gcse/science/grade_controller.rb
TobyRet: These are just the paths to the appropriate view (not redirecting to a url path)

malcolmbaig: ah, skimming too quickly! :+1:

TobyRet:
> Design System fixed width inputs

Done!

config/locales/application_form.yml
paulrobertlloyd: Hmmm, ‘real’ does seem odd, but does seem more like what somebody would say in real life. I’d stick with real if that’s what we’re using elsewhere.

malcolmbaig:
Couple of bugs during testing:

1. I can't submit triple science grades, I get this validation error
Screenshot_20201112_145203

2. If I fill in grades for both triple science and one of the other two sciences, submitting the form persists grade values in both `grade` and `grades`.

3. I don't think we're clearing both grade fields prior to an update. I can't test this (because of issue 1), but I think this will result in the same side effect as issue 2, ie - changing from triple->single/double (and vice versa) will populate both fields.

TobyRet:
> A few issues found during testing:
> 
> 1. I can't submit triple science grades, I get this validation error
>    Screenshot_20201112_145203
> 2. If I fill in grades for both triple science and one of the other two sciences, submitting the form persists grade values in both `grade` and `grades`.
> 3. I don't think we're clearing both grade fields prior to an update. I can't test this (because of issue 1), but I think this will result in the same side effect as issue 2, ie - changing from triple->single/double (and vice versa) will populate both fields.

Hi @malcolmbaig 

Thanks for taking a look. Sorry I've just picked up on this (was away for part of yesterday). Yes you are right - will fix now.

frankieroberto:
Can we customise the error message displayed when leaving the grades for the triple science award blank please? Currently they all say "Enter your grade", which makes the error summary a bit confusing if there's 2 or more of them:

Screenshot 2020-11-13 at 10 35 50

I'd suggest using "Enter your biology/chemistry/physics grade"

app/views/candidate_interface/gcse/science/grade/awards_edit.html.erb
```diff
@@ -0,0 +1,32 @@
...
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with model: @gcse_grade_form, url: candidate_interface_gcse_science_edit_grade_path, method: :patch do |f| %>
+      <%= f.govuk_error_summary %>
+
+      <h1 class="govuk-heading-xl"><%= t('gcse_edit_grade.page_title', subject: @subject) %></h1>
+
+      <%= render CandidateInterface::GcseGradeGuidanceComponent.new(@subject, @qualification_type) %>
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
```
paulrobertlloyd, line 13: There needn’t be any grid layout on this part of the page, so this can come out. Currently this reduced the width of the form.

app/views/candidate_interface/gcse/science/grade/awards_edit.html.erb
```diff
@@ -0,0 +1,32 @@
...
+          <%= f.govuk_radio_buttons_fieldset :gcse_science, legend: { text: "Select the GCSEs you did and include your grade"} do %>
+            <%= f.govuk_radio_button(:gcse_science, ApplicationQualification::SCIENCE_SINGLE_AWARD, label: { text: "Single award" }, hint: { text: "You have one science grade" }, link_errors: true ) do %>
+              <%= f.govuk_text_field :single_award_grade, label: { text: "Grade"}, hint: { text: "For example, 'C' or '4'" }, width: 2 %>
```
paulrobertlloyd, line 16: These strings should be localised, and also, use smart quotes, i.e.

```suggestion
              <%= f.govuk_text_field :single_award_grade, label: { text: "Grade"}, hint: { text: "For example, ‘C’ or ‘4’" }, width: 2 %>
```

app/views/candidate_interface/gcse/science/grade/awards_edit.html.erb
TobyRet: 👍

app/views/candidate_interface/gcse/science/grade/awards_edit.html.erb
TobyRet: 👍

TobyRet:
History will be rebased before merging ...

davidgisbey (approved):

config/locales/application_form.yml
```diff
@@ -818,6 +818,33 @@ en:
...
             other_grade:
               blank: Enter your grade
+        candidate_interface/science_gcse_grade_form:
+          attributes:
+            single_award_grade:
+              blank: Enter your science single award grade
+              invalid: Enter a real science single award grade
+            double_award_grade:
+              blank: Enter your science double award grade
+              invalid: Enter a real science double award grade
```
paulrobertlloyd, line 828: Don’t think we need say science here as already saying science in page title and elsewhere, so clear what this relates to.

```suggestion
            single_award_grade:
              blank: Enter your single award grade
              invalid: Enter a real single award grade
            double_award_grade:
              blank: Enter your double award grade
              invalid: Enter a real double award grade
```

config/locales/application_form.yml
```diff
@@ -818,6 +818,33 @@ en:
...
+              blank: Enter your Chemistry grade
+            physics_grade:
+              invalid: Enter a real Physics grade
+              blank: Enter your Physics grade
+            award_year:
+              blank: Enter the year you gained your qualification
+              invalid: Enter a real year
+              in_future: Enter a year before %{date}
+            other_grade:
+              blank: Enter your grade
```
paulrobertlloyd, line 847: Can we be consistent in the order of validation strings? Sometimes `blank` comes before `invalid`, for example.

config/locales/application_form.yml
```diff
@@ -818,6 +818,33 @@ en:
...
+              too_long: Grade must be %{count} characters or fewer
+            biology_grade:
+              invalid: Enter a real Biology grade
+              blank: Enter your Biology grade
+            chemistry_grade:
+              invalid: Enter a real Chemistry grade
+              blank: Enter your Chemistry grade
+            physics_grade:
+              invalid: Enter a real Physics grade
+              blank: Enter your Physics grade
```
paulrobertlloyd, line 841: We only capitalise `English` subject name as it is a proper noun (or so the content gurus tell me)

```suggestion
            biology_grade:
              invalid: Enter a real biology grade
              blank: Enter your biology grade
            chemistry_grade:
              invalid: Enter a real chemistry grade
              blank: Enter your chemistry grade
            physics_grade:
              invalid: Enter a real physics grade
              blank: Enter your physics grade
```

paulrobertlloyd (changes requested):

config/locales/application_form.yml
TobyRet: 👍

config/locales/application_form.yml
```diff
@@ -818,6 +818,33 @@ en:
...
+            double_award_grade:
+              blank: Enter your double award grade
+              invalid: Enter a real double award grade
+            grade:
+              blank: Enter your science grade
+              invalid: Enter a real science grade
+              too_long: Grade must be %{count} characters or fewer
+            biology_grade:
+              blank: Enter your biology grade
+              invalid: Enter a real Biology grade
```
paulrobertlloyd, line 835:
```suggestion
              invalid: Enter a real biology grade
```

config/locales/application_form.yml
```diff
@@ -818,6 +818,33 @@ en:
...
+            chemistry_grade:
+              blank: Enter your chemistry grade
+              invalid: Enter a real Chemistry grade
```
paulrobertlloyd, line 838:
```suggestion
              invalid: Enter a real chemistry grade
```

config/locales/application_form.yml
```diff
@@ -818,6 +818,33 @@ en:
...
+            physics_grade:
+              blank: Enter your physics grade
+              invalid: Enter a real Physics grade
```
paulrobertlloyd, line 841:
```suggestion
              invalid: Enter a real physics grade
```

paulrobertlloyd (approved):
Wonder if we could/should localise all of the remaining unlocalised strings, but that’s not something that should prevent us from merging this PR, I don’t think. 🚢

TobyRet:
> Wonder if we could/should localise all of the remaining unlocalised strings, but that’s not something that should prevent us from merging this PR, I don’t think. 🚢

There has been some discussion about doing a locales refactor. @raamSoftwire  and I are planning to do a clean up PR once our work is merged.

PR #3358, opened 2020-11-06
Add missing validation for referee name

paulrobertlloyd:
Add missing validation string spotted while testing.


stevehook (approved):
:+1: Works for me

PR #3371, opened 2020-11-09
Remove remaining instances of styled content container

paulrobertlloyd:
## Context

The change in #3303 provided a means of automatically adding typography classes to markdown content, removing the need for the `app-styled-content` container in those instances. However, this container was still used in few other places.

* Removes remaining instances of `app-styled-content`, manually adding the typography classes to headings, paragraphs and lists it previously contained.
* Removes `.app-styled-content` styles
* Adds `.app-list--definition` styles, replacing `.app-styled-content dl` styles.
* Adds global styles for the `code` element, replacing `.app-styled-content code` styles.

## Guidance to review

The global nature of the styling for `code` may be controversial, however we only use this element within support and api-docs interfaces, and the styling is light touch.

## Things to check


app/frontend/styles/application.scss
```diff
@@ -6,9 +6,9 @@
   @return url("~assets/images/" + $filename);
 }
 
-$moj-images-path: "~@ministryofjustice/frontend/moj/assets/images/";
 $govuk-font-url-function: frontend-font-url;
 $govuk-image-url-function: frontend-image-url;
+$govuk-global-styles: true;
```
tijmenb, line 11: Do we need this? Everything should have classes now right?

tijmenb:
Should we propose to add `code` to GOV.UK Frontend?

app/frontend/styles/application.scss
paulrobertlloyd: Yes, for links. Otherwise you need to add `govuk-link` to all links, and even if did that in this PR, this class often gets missed. It also correctly styles any paragraphs that are missing `govuk-body`.

tijmenb: Hmm, but we've put in a looot of work to add `govuk-link` to everything. Should we remove it everywhere then?

paulrobertlloyd:
> Should we propose to add `code` to GOV.UK Frontend?

I’ll add something to the design system backlog. There are no code styles on GOV.UK, but several services implement styles for the element, so would be good for them to be consistent.

app/frontend/styles/application.scss
paulrobertlloyd: This is `govuk-frontend`’s original sin - so often you’ll see unstyled bright blue links because this class gets missed. But switching this flag on only effects paragraphs and links, so it becomes unclear to developers when they should/shouldn’t add typography classes, I think.

Here’s what the flag does: https://github.com/alphagov/govuk-frontend/blob/387c33d5280745523a9ee0fd1da961be584137e6/src/govuk/core/_global-styles.scss#L10

@fofr @edwardhorsford @frankieroberto @adamsilver In your travels, have you seen any issues with turning the `$govuk-global-styles` flag on?

fofr: I see and fix so many issues with "wrong-blue" links, I've seen no issues with `$govuk-global-styles` but then I've rarely used.

edwardhorsford: I've not seen any specific issues - though for 'completeness' I tend to *always* use the classes - as there's often unintended outcomes where you don't.

My gut is it's easier to *always* require the styles than to have rules for when you don't need them.

A similar issue to the bright blue links is to have links that are the wrong size - as they need to either be within a paragraph with `govuk-body` or explicitly given size - the `govuk-link` class doesn't set it.

edwardhorsford:
> > Should we propose to add `code` to GOV.UK Frontend?
> 
> I’ll add something to the design system backlog. There are no code styles on GOV.UK, but several services implement styles for the element, so would be good for them to be consistent.

I've seen the request come up before, so there's likely something already on the backlog if you have a look.

app/frontend/styles/application.scss
frankieroberto: Yeah, I think the “global styles” mode should probably be avoided now. The Design System team mainly added it for convenience when prototyping, but it’s really hard for them to support, and there have been a few bugs with the mode over the past couple of years (which I think have all been fixed now, but might creep in again).

My feeling is that we should aim to turn it off, and put the effort in to make sure that all the links, paragraphs, etc have the classes added.

edwardhorsford: You could set a default global style for links that makes the 'unstyled' ones more obvious - possibly only deployed to staging.

paulrobertlloyd: In the interests of moving this forward, and given that only a few links were affected by this change, have updated those links to use `govuk_link_to` helper, and removed the `$govuk-global-styles` setting.

tijmenb: Thank you!

tijmenb (approved):

PR #3378, opened 2020-11-10
Add feedback prompts - Non-JS solution

davidgisbey:
## Context

**The designs have now changed for this. This should be merged then the changes applied. It's behind a feature flag so it's safe to do so**

First of all apologies for the size of this PR. I realised it was going to be bigger than I thought fairly early on so I've tried to keep the commits clean.

```
As a service owner
I need to understand what issues users are having at specific points in the journey
So that I can address them

```
This PR adds the non-JavaScript solution. The desired behaviour is:

If a user clicks yes on the feedback prompt:
1. They are taken to a second page where they can provide details of their issue
2. Once submitted they see a thank you page with a link back to the section they were completing when they chose to provide feedback

if a user clicks no on the feedback prompt:
1. They are taken to the thank you page. The same as above re. the link

- Adds a component for the feedback prompt 

image

- Adds the feedback page 

image

- Adds the thank you page

image

## GIF section that I've shamelessly stolen from Toby (excuse the low quality)

feedback prompts

## Guidance to review

Def review per commit.

I've purposely only added the component to the references section for now for two reasons:
1. To reduce the size of the PR
2. So that any changes needed can be applied before expanding the component throughout the application

Once this PR is merged, then I'll add the component to the following sections:

* Degree
* GCSEs
* Other qualifications
* Personal statement
* Subject knowledge

I'll then add a recursive test that uses the relevant tests to checks the component is rendering where it should be.

## Link to Trello card

https://trello.com/c/AvxRVqld/2465-%F0%9F%93%9D-dev-add-feedback-prompt-apply

## Things to check

### DIFF
--- a/app/components/candidate_interface/application_feedback_component.html.erb
+++ b/app/components/candidate_interface/application_feedback_component.html.erb
@@ -2,7 +2,7 @@
   <div class="govuk-width-container govuk-!-margin-top-6">
     <div class="app-feedback__container">
       <h2 class="app-feedback__container-title">
-        Are there any issues with this section?
+        <%= t('application_feedback_component.title') %>
       </h2>
 
       <ul class="app-feedback__list">
@@ -16,7 +16,7 @@
             ),
             class: 'app-feedback__list-item app-feedback__button',
             method: :post do |f| %>
-              <%= f.govuk_submit 'Yes' %>
+              <%= f.govuk_submit t('application_feedback_component.issues.submit') %>
           <% end %>
 
           <%= form_with url: candidate_interface_application_feedback_path(
@@ -28,7 +28,7 @@
             ),
             class: 'app-feedback__list-item app-feedback__button govuk-!-margin-left-1',
             method: :post,  remote: true do |f| %>
-              <%= f.govuk_submit 'No' %>
+              <%= f.govuk_submit t('application_feedback_component.no_issues.submit') %>
           <% end %>
         </li>
       </ul>
--- a/app/components/candidate_interface/application_feedback_component.html.erb
+++ b/app/components/candidate_interface/application_feedback_component.html.erb
@@ -1,0 +1,37 @@
+<% if FeatureFlag.active?(:feedback_prompts) %>
+  <div class="govuk-width-container govuk-!-margin-top-6">
+    <div class="app-feedback__container">
+      <h2 class="app-feedback__container-title">
+        Are there any issues with this section?
+      </h2>
+
+      <ul class="app-feedback__list">
+        <li class="app-feedback__list-item">
+          <%= form_with url: candidate_interface_application_feedback_path(
+              issues: true,
+              section: section,
+              path: path,
+              page_title: page_title,
+              id_in_path: id_in_path
+            ),
+            class: 'app-feedback__list-item app-feedback__button',
+            method: :post do |f| %>
+              <%= f.govuk_submit 'Yes' %>
+          <% end %>
+
+          <%= form_with url: candidate_interface_application_feedback_path(
+              issues: false,
+              section: section,
+              path: path,
+              page_title: page_title,
+              id_in_path: id_in_path
+            ),
+            class: 'app-feedback__list-item app-feedback__button govuk-!-margin-left-1',
+            method: :post,  remote: true do |f| %>
+              <%= f.govuk_submit 'No' %>
+          <% end %>
+        </li>
+      </ul>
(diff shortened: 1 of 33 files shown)


app/controllers/candidate_interface/application_feedback_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+
+    def add_errors_to_flash_message
+      @application_feedback_form
+      .errors
+      .messages
+      .values
+      .flatten
+      .join('\n')
+    end
```
davidgisbey, line 78: this feels hacky. It's unlikely to be hit anyway as someone would have to POST without the necessary info in the query string.

app/controllers/candidate_interface/application_feedback_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+      else
+        track_validation_error(@application_feedback_form)
+        render :edit
+      end
+    end
+
+    def thank_you
+      @application_feedback = current_application.application_feedback.last
+      @previous_path = if @application_feedback.id_in_path
+                         Rails.application.routes.url_helpers.send(@application_feedback.path, @application_feedback.id_in_path)
```
davidgisbey, line 42: There's probably a better way to do this than persist the related id to the DB, but this was the first solution. thought of. I did think about just passing the raw path in, but I like testing that we're passing valid path int the DB which this implementation allows us to do. Open to suggestions.

app/frontend/styles/_application-feedback.scss
```diff
@@ -0,0 +1,65 @@
+.app-feedback__button .govuk-button {
```
paulrobertlloyd, line 1: We should call this file `_feedback.scss` to be consistent with how we name other stylesheets.

app/components/candidate_interface/application_feedback_component.html.erb
```diff
@@ -0,0 +1,37 @@
+<% if FeatureFlag.active?(:feedback_prompts) %>
+  <div class="govuk-width-container govuk-!-margin-top-6">
+    <div class="app-feedback__container">
+      <h2 class="app-feedback__container-title">
+        Are there any issues with this section?
```
paulrobertlloyd, line 5: Can this value be localised too?

app/components/candidate_interface/application_feedback_component.html.erb
```diff
@@ -0,0 +1,37 @@
...
+          <%= form_with url: candidate_interface_application_feedback_path(
+              issues: false,
+              section: section,
+              path: path,
+              page_title: page_title,
+              id_in_path: id_in_path
+            ),
+            class: 'app-feedback__list-item app-feedback__button govuk-!-margin-left-1',
+            method: :post,  remote: true do |f| %>
+              <%= f.govuk_submit 'No' %>
```
paulrobertlloyd, line 31: The two options for the buttons could be localised as well.

app/controllers/candidate_interface/application_feedback_controller.rb
davidgisbey: This will be pulled out with the new design

app/controllers/candidate_interface/application_feedback_controller.rb
davidgisbey: We won't need this is the new design as users won't be redirected back to the page they came from.

app/views/candidate_interface/references/type/new.html.erb
```diff
@@ -12,3 +12,12 @@
...
 </div>
+
+<%= render(
+    CandidateInterface::ApplicationFeedbackComponent.new(
+      section: 'application_references',
+      path: 'candidate_interface_references_type_path',
+      page_title: t('page_titles.references_type')
+    )
+  )
+%>
```
tvararu, line 23: Could we add this to the layout instead and only have it render on a whitelist of pages? Would save us some duplication.

tvararu (approved):
👍

app/controllers/candidate_interface/application_feedback_controller.rb
stevehook: Agree it's nice to be able to validate the paths that get stored, does `ActionController::Routing::Routes.recognize_path` go some way to achieving this?

stevehook (commented):
Looks good, just has a query about the way we store paths, sorry bit late giving feedback on this!

PR #3379, opened 2020-11-10
Fix test data generation for oracle

benswannack:
## Context
Oracle reported that some test data that was being generated in Sandbox had timestamps from the future (for e.g. updated_at), which is unrealistic and upsets things on their side.

Update the test application generator so that its time travel functionality stays in the past. (The issue was that the generator's time could go further in time than the current date, and set unrealistic timestamps)

## Guidance to review
Each commit separately

## Link to Trello card
https://trello.com/c/Ma0Erlkt/2944-fix-test-data-generation-for-oracle

## Things to check

### DIFF
--- a/app/services/end_of_cycle_timetable.rb
+++ b/app/services/end_of_cycle_timetable.rb
@@ -1,4 +1,7 @@
 class EndOfCycleTimetable
+  CURRENT_YEAR_FOR_SCHEDULE = 2021
+
+  # These dates are configuration for when the previous cycle ends and the next cycle starts
   # The 2020 dates are made up so we can generate sensible test data
   CYCLE_DATES = {
     2020 => {
@@ -92,7 +95,7 @@ class EndOfCycleTimetable
 
   def self.schedules
     {
-      real: CYCLE_DATES[2021],
+      real: CYCLE_DATES[CURRENT_YEAR_FOR_SCHEDULE],
 
       today_is_mid_cycle: {
         apply_1_deadline: 1.day.from_now.to_date,
--- a/app/services/end_of_cycle_timetable.rb
+++ b/app/services/end_of_cycle_timetable.rb
@@ -1,5 +1,4 @@
 class EndOfCycleTimetable
-
   # The 2020 dates are made up so we can generate sensible test data
   CYCLE_DATES = {
     2020 => {
@@ -18,7 +17,7 @@ class EndOfCycleTimetable
       find_reopens: Date.new(2020, 10, 6),
       apply_reopens: Date.new(2020, 10, 13),
     },
-  }
+  }.freeze
 
   def self.between_cycles?(phase)
     phase == 'apply_1' ? between_cycles_apply_1? : between_cycles_apply_2?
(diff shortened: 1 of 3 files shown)


spec/lib/test_applications_spec.rb
```diff
@@ -21,10 +21,10 @@
     candidate = application_form.candidate
 
     expect(candidate.created_at).to eq candidate.last_signed_in_at
-    expect(days_between_ignoring_time_of_day(application_choice.created_at, candidate.created_at)).to be >= 1
-    expect(days_between_ignoring_time_of_day(application_form.submitted_at, application_choice.created_at)).to be >= 1
-    expect(days_between_ignoring_time_of_day(application_choice.offered_at, application_choice.sent_to_provider_at)).to be >= 1
-    expect(days_between_ignoring_time_of_day(application_choice.accepted_at, application_choice.offered_at)).to be >= 1
+    expect(candidate.created_at <= application_choice.created_at).to be_truthy
```
duncanjbrown, line 24: Must we fudge it with `be_truthy` — would this not `be true`?

app/services/end_of_cycle_timetable.rb
```diff
@@ -72,14 +92,7 @@ def self.next_cycle_year
...
     {
-      real: {
-        apply_1_deadline: Date.new(2020, 8, 24),
-        stop_applications_to_unavailable_course_options: Date.new(2020, 9, 7),
-        apply_2_deadline: Date.new(2020, 9, 18),
-        find_closes: Date.new(2020, 10, 3),
-        find_reopens: Date.new(2020, 10, 6),
-        apply_reopens: Date.new(2020, 10, 13),
-      },
+      real: CYCLE_DATES[2021],
```
mnacos, line 95: `2021` here effectively controls `RecruitmentCycle.current_year`, give or take some adjustment for the transition before `find_reopens`. Would it be better to create and refer to a `RecruitmentCycle::CURRENT_YEAR` here, so that adjusting the year happens in one place, `RecruitmentCycle`?

mnacos (approved):
Time-travel FTW -- minor comment below about the source of truth for `current_year` (only indirectly related to this work, but still)

app/services/end_of_cycle_timetable.rb
benswannack: It feels like this whole thing needs some love. It looks like EndOfCycleTimetable is actually the source of truth, and RecruitmentCycle reads from it to determine which year we are in. I think that makes sense, especially if we rename EndOfCycleTimetable to CycleTimetable (since it's not especially related to the end of a cycle)

mnacos: That's fine but even if `CycleTimetable` remains the source of truth for which year we're in, having `2021` hardcoded inside the values of a hash called `.schedules` seems a bit hidden away. I'd rather have this defined at the top of the file and named according to its significance! Not a huge deal, I just don't like where this sits at the moment.

benswannack: Ah that's a good point - will extract to a named constant

PR #3381, opened 2020-11-10
Configure PORT and update README

despo:
## Context
Updates README to use foreman to run the server

- configure PORT in .env.example and .env.development (required by puma to run server at port 3000 and not default to puma configured PORT 5000)

## Things to check


tvararu:
> run yarn installation with --ignore-engines property to resolve incompatible engine error

The reason for the error is because [we recently started specifying a Node version in our package.json](https://github.com/DFE-Digital/apply-for-teacher-training/pull/3331).

Rather than using yarn with `--ignore-engines`, could you install and run the version of node LTS that we use in production? :+1: 

[We also removed foreman from the Gemfile because that's what the author recommends](https://github.com/DFE-Digital/apply-for-teacher-training/pull/1004).

despo:
@tvararu Thanks for the feedback. I wrongly assumed that any missing dependencies would have been install by `yarn` as my attempt to use brew to setup that specific version was not working. Sorted now.

.env.development
```diff
@@ -1,3 +1,4 @@
 FIND_BASE_URL=https://api.publish-teacher-training-courses.service.gov.uk/api/v3/
 HOSTING_ENVIRONMENT_NAME=development
 STATE_CHANGE_SLACK_URL=
+PORT=3000
```
duncanjbrown, line 4: This doesn't seem to make  a difference for me locally — possibly because `dotenv` loads after Puma?

duncanjbrown: It would be good to straighten this out, though, because local emails under foreman sometimes point to `localhost:5100`, which is odd. I wonder if that's the port Sidekiq thinks it's running on.

despo: I noticed that port 5000 is also mentioned on the instructions of this issue: https://github.com/DFE-Digital/apply-for-teacher-training/pull/3392 so other devs must be experiencing the same issue that I have. Could PORT already be configured to 3000 globally on your machine?

AgaDufrat: I normally run the app with `rails s` pointing to `localhost:3000`
Only if I need redis (for things like generating test applications or setting up organisational permissions), I run it with foreman (`localhost:5000`)

despo: As the [Procfile](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/Procfile#L1) points to the [puma configuration](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/config/puma.rb#L13) the expectation is that foreman runs the app on port 3000.

duncanjbrown: Looks like we're falling back to Foreman's default of `5000`. This statement explains why Sidekiq runs on 5100 I think https://github.com/ddollar/foreman/blob/c0b178c164396d6afe6fc3874abfd22f6cddbeae/lib/foreman/engine.rb#L271

duncanjbrown: I think setting `PORT` in the `Procfile` fixes everything actually! Setting for Sidekiq so email URLs are consistent

Screenshot 2020-11-12 at 16 35 21

duncanjbrown: ☝️ worth checking that doesn't break Heroku

despo: How about loading dotenv at the top of the Procfile as it would mean all configuration is kept at one place and also enable changing the port in a single place (for anyone who may have something else running at port 3000).

```
require 'dotenv'
Dotenv.load
```

duncanjbrown: This doesn't work — I don't think the `Procfile` is Ruby?

despo: You are right, it doesn't. 

There are a couple of other things that can be done to control when things are loaded, [look here](https://github.com/bkeepers/dotenv#note-on-load-order) but I think it's easy enough for anyone to find if they are having issues so I'll just remove this from .env.development and leave it in the example file for now.

despo:
Reopened as #3410 @duncanjbrown

PR #3384, opened 2020-11-10
Provider user gives feedback for applications rejected by default

mnacos:
### Context

We would like to give provider users the opportunity to add feedback for applications rejected by default. The aim of this PR is to capture this feedback and surface it in the right places within the provider interface. Displaying such feedback to candidates is not part of this work.

Add `give_feedback_for_rbd` section to `app/services/provider_interface/sort_application_choices.rb`, so that the 'Give feedback' section appears in the application list.

image

Add `@rejection_reason_required` block to `app/views/provider_interface/application_choices/_application_choice_header.html.erb` so that the following prompt appears (only for applications rejected by default).

image

Add routes/controller/views for:

- GET /feedback
- POST /feedback/check
- POST /feedback

Add a `FeedbackController` and a `RejectedByDefaultFeedbackForm` for validating `rejection_reason`.

Add a `RejectByDefaultFeedbackComponent` for displaying any unstructured reasons for rejection present for applications rejected by default. The 'Change' link allows you to go back and edit `rejection_reason` as many times as you want.

Store feedback within `rejection_reason` and set `reject_by_default_feedback_sent_at` (added in https://github.com/DFE-Digital/apply-for-teacher-training/pull/3376)

### Guidance to review

The new 'Give feedback: ...' group in the applications list will include both current cycle _and_ previous cycle applications, as long as they are missing a `rejection_reason` and were rejected by default. Should we only be prompting users to add feedback for current cycle applications?

Should the 'Give feedback' flow require `make_decisions` and what do we show users without it? The 'Give feedback: you did not respond in time' section within the applications list is shown to all.

The number of days quoted in the prompt correspond to `reject_by_default_days`. Is this correct?

I have named the component `RejectByDefaultFeedbackComponent` because I assume we'll only use it for applications rejected by default. Is this a valid assumption, or should we allow provider users to edit/change `rejection_reasons` for other types of rejection, as well? 

Are the error messages ok?

image
---
image

### Link to Trello card

[2920 - Giving feedback after Reject By Default](https://trello.com/c/6RZWDB9F)

### Things to check


app/views/provider_interface/application_choices/_application_choice_header.html.erb
```diff
@@ -44,6 +44,22 @@
       </div>
     </div>
   </div>
+<% elsif @rejection_reason_required && flash.empty? && @provider_can_respond -%>
```
steventux, line 47: Not necessarily in this PR but it looks like `_application_choice_header.html.erb` is getting complex enough to be a component or have some backing library to help with conditionals. The instance var setup in the controller could then be reduced.

steventux (approved):
:+1: Looks good to me.

app/views/provider_interface/application_choices/show.html.erb
```diff
@@ -6,6 +6,15 @@
   <div class="govuk-grid-column-two-thirds">
     <%= render ProviderInterface::StatusBoxComponent.new(application_choice: @application_choice, options: @status_box_options) unless @offer_present %>
 
+    <% if @application_choice.status == 'rejected' && @application_choice.reject_by_default_feedback_sent_at %>
+      <% feedback_at = @application_choice.reject_by_default_feedback_sent_at %>
+      <% intro = "You sent the following feedback to the candidate on #{feedback_at.strftime('%d %b %Y')}." %>
```
benswannack, line 11: Can we use the `to_s(:govuk_date)` here instead?

mnacos:
Plans for this feature have changed, I'm updating the design and also the normal rejection flow.

mnacos:
Design update:

Direct rejection flow has been updated to be consistent with the reject by default feedback flow, until structured reasons for rejection are merged.

Before:
image

After:
image

I've also added feedback events to the Timeline:
image

Code update:

Code now sends email to the candidate: see relevant preview here --> https://apply-for-te-2920-feedb-h0mf6t.herokuapp.com/rails/mailers/candidate_mailer/feedback_received_for_application_rejected_by_default

`RejectByDefaultFeedbackComponent` has been repurposed as `FeedbackPreviewComponent`. The status box component for 'rejected' applications has been updated. See https://apply-for-te-2920-feedb-h0mf6t.herokuapp.com/rails/view_components/provider_interface/feedback_preview_component and https://apply-for-te-2920-feedb-h0mf6t.herokuapp.com/rails/view_components/provider_interface/status_box_component/rejected (hit browser refresh for variations)

cc @adamsilver

app/views/provider_interface/application_choices/show.html.erb
mnacos: In the previous version we needed `10 Nov 2020` rather than `10 November 2020` but we've switched to `to_s(:govuk_date)` now.

app/services/provider_interface/sort_application_choices.rb
```diff
@@ -71,6 +71,7 @@ def self.give_feedback_for_rbd
           status = 'rejected'
             AND rejected_by_default
             AND rejection_reason IS NULL
+            AND rejected_at >= '2020-11-11T00:00:00+00:00'::TIMESTAMPTZ
```
duncanjbrown, line 74: Could we turn this into a named constant, perhaps?

spec/services/provider_interface/sort_application_choices_spec.rb
```diff
@@ -46,13 +46,21 @@
       described_class.call(application_choices: ApplicationChoice.all).first
     end
 
-    # TODO: groups 1, 3, 9 require the relevant EoC implementation first
+    it '#deferred_offers_pending_reconfirmation' do
+      create(:application_choice, :with_deferred_offer, :previous_year)
+      expect(application_choice.task_view_group).to eq(1)
```
duncanjbrown, line 51: Might this be a bit more readable (and maybe reliable) if we assigned the integers to constants?

app/components/provider_interface/application_timeline_component.rb
```diff
@@ -50,8 +50,24 @@ def note_events
       end
     end
 
+    def feedback_events
+      if application_choice.rejected_by_default
+        application_choice.audits.where(action: 'update').where(
+          'audited_changes ? \'reject_by_default_feedback_sent_at\'',
```
benswannack, line 56: Can we do something like `'audited_changes ? :key', key: :reject_by_default_feedback_sent_at` to avoid escaping quotes here?

duncanjbrown (commented):
Looking great — only nits about constants (!) inline. Going to take for one last spin before approving

app/components/provider_interface/application_timeline_component.rb
steventux: Is the `?` operator here a shorthand for a json function like `jsonb_exists`? 
It would be clearer to use the underlying function name if so.

steventux (commented):
One question about simplifying the audit changes lookup, otherwise looks good. There's going to be a bit of joining up around structured reasons for rejection but that's not for this PR.

adamsilver:
[ ] Having entered feedback, when clicking back from the check answers page the feedback is not blank.

app/components/provider_interface/application_timeline_component.rb
mnacos: No, it's a (jsonb-specific) built-in operator (see https://www.postgresql.org/docs/13/functions-json.html#FUNCTIONS-JSONB-OP-TABLE). I've added a comment to the code explaining it.

steventux: Understood, I was leaning towards using the longhand for this operator (though as you point out it's well documented):

`SELECT oprname, oprcode FROM pg_operator WHERE oprname = '?'` which returns `jsonb_exists`

So this could also be written as `.where("jsonb_exists('audited_changes', 'reject_by_default_feedback_sent_at'))"`
Not a blocker, just a style preference.

duncanjbrown (approved):

mnacos:
Updated email formatting (inset for `rejection_reason`):
image

spec/services/provider_interface/sort_application_choices_spec.rb
mnacos: These codes are still very local to `SortApplicationChoices`, so will postpone this for next time we make changes.

app/components/provider_interface/application_timeline_component.rb
mnacos: Ok, combining both your comments into:

```
        application_choice.audits.where(action: 'update').where(
          'jsonb_exists(audited_changes, :key)',
          key: :reject_by_default_feedback_sent_at,
        )
```

PR #3387, opened 2020-11-11
Revise flow for degree grades

malcolmbaig:
## Context

The UX of this part of the flow was a little confusing, so apply several structural and copy-related changes to make it easier.

- Add a separate step that confirms if the degree is completed or not (ie - awarded vs predicted grade)
- Remove handling of the degree `predicted_grades` attribute from the grade step
- Update copy on the grade step to reflect permutations of UK/International/Predicted/Completed
- Update the degrees review page to summarise these changes

See commits for detail.
 

Screenshot_20201111_094315

## Guidance to review

- Per-commit review highly recommended. I considered splitting the UK / International work into two separate PRs but I'm not sure that would've necessarily made reviews easier. Instead, the UK-related changes are done first in the commit history, with International changes in their own self-contained commit right towards to the end.
- There are four routes through this flow that are worth checking in the review app:
  - UK degree (completed)
  - UK degree (predicted)
  - Non-UK degree (completed)
  - Non-UK degree (predicted)

## Link to Trello card
https://trello.com/c/zeuemNFz

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/degrees/completion_status_controller.rb
+++ b/app/controllers/candidate_interface/degrees/completion_status_controller.rb
@@ -1,0 +1,42 @@
+module CandidateInterface
+  module Degrees
+    class CompletionStatusController < BaseController
+      def new
+        @completion_status_form = DegreeCompletionStatusForm.new
+      end
+
+      def create
+        @completion_status_form = DegreeCompletionStatusForm.new(completion_status_params)
+        if @completion_status_form.save(current_degree)
+          redirect_to candidate_interface_degree_grade_path
+        else
+          render :new
+        end
+      end
+
+      def edit
+        @completion_status_form = DegreeCompletionStatusForm.new(completion_status_params).fill(current_degree)
+      end
+
+      def update
+        @completion_status_form = DegreeCompletionStatusForm.new(completion_status_params)
+
+        if @completion_status_form.save(current_degree)
+          current_application.update!(degrees_completed: false)
+          redirect_to candidate_interface_degrees_review_path
+        else
+          track_validation_error(@completion_status_form)
+          render :new
+        end
+      end
+
+    private
+
+      def completion_status_params
+        params
+          .fetch(:candidate_interface_degree_completion_status_form, {})
+          .permit(:degree_completed)
+      end
+    end
+  end
+end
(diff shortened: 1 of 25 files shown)


app/controllers/candidate_interface/degrees/completion_status_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+      def new
+        @completion_status_form = DegreeCompletionStatusForm.new
+      end
+
+      def create
+        @completion_status_form = DegreeCompletionStatusForm.new(completion_status_params)
+        if @completion_status_form.save(current_degree)
+          redirect_to candidate_interface_degree_grade_path
+        else
+          render :new
```
davidgisbey, line 13: I think we should track the validation errors here and on update.

spec/system/candidate_interface/candidate_entering_degrees_spec.rb
```diff
@@ -379,4 +388,35 @@ def then_i_can_check_my_answers
...
+private
+
+  def click_change_link(row_description)
+    link_text = "Change #{row_description}"
+    page.all('.govuk-summary-list__actions')
+      .find { |row| row.has_link?(link_text) }
+      .click_link(link_text)
+  end
+
+  def expect_validation_error(message)
```
davidgisbey, line 418: Is it worth extracting this into a helper? It's really nice and would be good to use it more widely across the system specs.

davidgisbey (approved):
I tested all four flows locally and they all seem to be working really nicely. I've left a comment about tracking validation errors.

spec/system/candidate_interface/candidate_entering_degrees_spec.rb
malcolmbaig: Definitely! I have a followup PR in the works with some refactors, including this and another spec helper.

stevehook (approved):
Review app worked perfectly for me. One little question but no changes needed from me.

app/forms/candidate_interface/degree_completion_status_form.rb
```diff
@@ -0,0 +1,26 @@
...
+    validates :degree_completed, presence: true
+
+    def save(degree)
+      return false unless valid?
+
+      degree.update!(predicted_grade: grade_is_predicted?)
+    end
+
+    def fill(degree)
+      self.degree_completed = degree.completed? ? 'yes' : 'no'
```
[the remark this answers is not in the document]
malcolmbaig: Went with this because I thought it read clearer on that line, but I think you're right - symmetry between `fill` and `save` would be less confusing. Will update!

tvararu (approved):

PR #3389, opened 2020-11-11
 Cancel references for cancelled test applications

tijmenb:
## Context

The `DetectInvariants` checker checks if there are requested references that belong to application choices that haven't been submitted yet. This is impossible because you can't submit applications without all references being either cancelled, declined or with feedback.

However, in QA/Sandbox we also have application choices in the cancelled state. This state has been removed, but it still exists on prod and so we need it as test data too. This is currently raising an error:

https://sentry.io/organizations/dfe-bat/issues/1990257169

For this application:

https://qa.apply-for-teacher-training.service.gov.uk/support/applications/4240

I've manually cancelled the references on this test application.

Cancel references when cancelling the application.

## Guidance to review

Per commit.

## Link to Trello card

https://trello.com/c/9FgC13b6

### DIFF
--- a/app/lib/test_applications.rb
+++ b/app/lib/test_applications.rb
@@ -103,7 +103,7 @@ class TestApplications
         return application_choices
       end
 
-      if states.include? :cancelled
+      if states == [:cancelled]
         # The cancelled state doesn't exist anymore in the current system,
         # so we have to set this manually.
         @application_form.application_choices.each do |application_choice|
@@ -111,6 +111,11 @@ class TestApplications
             status: 'cancelled',
           )
         end
+
+        @application_form.application_references.each do |reference|
+          CancelReferee.new.call(reference: reference)
+        end
+
         return
       end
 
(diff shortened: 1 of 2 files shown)


app/services/test_application.rb
```diff
@@ -1,54 +1,43 @@
...
-        courses_to_apply_to: Course.current_cycle.includes(:course_options).joins(:course_options).distinct.open_on_apply.where(provider: provider),
-      )
-    end
+  def initialize(recruitment_cycle_year:, states:, courses_to_apply_to:, apply_again: false, course_full: false, candidate: nil)
+    @recruitment_cycle_year = recruitment_cycle_year
+    @states = states
+    @possible_courses_to_apply_to = courses_to_apply_to
+    @apply_again = apply_again
+    @course_full = course_full
+    @to_candidate = candidate
```
malcolmbaig, line 13: I think we should drop the `to_` prefix - not sure it helps clarify?

tvararu (approved):

malcolmbaig (approved):

PR #3392, opened 2020-11-11
Resolve dual applications flow

AgaDufrat:
## Context

As a support agent,
I need to be able to record when I take actions in the process of resolving dual applications
So I can see when I need to take the next action

Refactors existing "Send initial emails" action and adds the next actions until the dual application is resolved.
- when first flagged as "Action needed"
<kbd>Screenshot 2020-11-11 at 17 43 41</kbd>
- After confirming initial emails were sent
<kbd>Screenshot 2020-11-11 at 14 43 12</kbd>
- After 5 working days
<kbd>Screenshot 2020-11-11 at 17 45 36</kbd>
- After confirming reminder emails were sent
<kbd>Screenshot 2020-11-11 at 14 45 10</kbd>
- After 10 working days
<kbd>Screenshot 2020-11-11 at 17 51 16</kbd>
- After requesting withdrawal from UCAS
<kbd>Screenshot 2020-11-12 at 09 31 54</kbd>
- After confirming that the application was withdrawn from UCAS
<kbd>Screenshot 2020-11-11 at 14 48 23</kbd>

Note: When the candidate withdraws from one of the services as requested, the status on Apply is updated automatically and the status on UCAS is updated daily after receiving new matching data.
When there is no longer a dual application we display "No action required" along with the information about last action taken. This reduces support agents workload.
<kbd>Screenshot 2020-11-11 at 13 13 45</kbd>

## Guidance to review
- generate test applications `http://localhost:5000/support/settings/tasks` (start the app with foreman) 
- got to `/support/ucas-matches`
- click on a match without "Action needed" tag. "No action required" should be shown

- click on a match with "Action needed" tag. 
- click the action button
- change the time in the console eg `UCASMatch.find(ID).update(candidate_last_contacted_at: Time.zone.now - 6.days)`
🔁 repeat last two steps until you get to the end of the flow

Notes:
Not a massive fan of the `need_to_send_reminder_emails?` / `need_to_request_withdrawal_from_ucas` / `need_to_confirm_withdrawal_from_ucas?` methods, so any suggestions on improving them welcome! 
I considered replacing them with `need_to_perform_next_action_after?(last_action, days_wait)` but doesn't seem very readable and I don't like the idea of passing the number of days from different places. Also looked into using `TimeLimitConfig`.

## Link to Trello card

## Things to check

### DIFF
--- a/app/components/support_interface/ucas_match_action_component.rb
+++ b/app/components/support_interface/ucas_match_action_component.rb
@@ -13,6 +13,12 @@ module SupportInterface
         button_text: 'Confirm reminder emails were sent',
         form_path: :support_interface_record_reminder_emails_sent_path,
       },
+      request_withdrawal_from_ucas: {
+        description: 'Request withdrawal from UCAS',
+        button_text: 'Confirm withdrawal from UCAS was requested',
+        form_path: :support_interface_record_ucas_withdrawal_requested_path,
+        instructions: "We need to contact UCAS. Please send an encrypted file with the candidate's duplicate application details to Harry Haines (h.haines@ucas.ac.uk) and Lizzy Carter (l.carter@ucas.ac.uk) from UCAS to ask them to remove the candidate from UTT.",
+      },
     }.freeze
 
     def initialize(match)
@@ -45,6 +51,8 @@ module SupportInterface
         :send_initial_emails
       elsif @match.initial_emails_sent? && @match.need_to_send_reminder_emails?
         :send_reminder_emails
+      elsif @match.reminder_emails_sent? && @match.need_to_request_withdrawal_from_ucas?
+        :request_withdrawal_from_ucas
       end
     end
 
@@ -53,7 +61,11 @@ module SupportInterface
     end
 
     def required_action_details
-      "We need to contact the candidate and the provider. Use the appropriate Zendesk macro. Alternatively, you can use the appropriate template from <a class='govuk-link' href='https://docs.google.com/document/d/1s5ql4jNUUr3QDPUQYWImkZR6o8upvutrjOEuoZ0qqTE'>this document</a>.".html_safe
+      contact_info = ACTIONS[next_action][:instructions] ||
+        "We need to contact the candidate and the provider. Use the appropriate Zendesk macro. Alternatively, you can use the appropriate template from <a class='govuk-link' href='https://docs.google.com/document/d/1s5ql4jNUUr3QDPUQYWImkZR6o8upvutrjOEuoZ0qqTE'>this document</a>."
+      support_manual_info = "<br><br>Please refer to <a class='govuk-link' href='https://docs.google.com/document/d/1XvZiD8_ng_aG_7nvDGuJ9JIdPu6pFdCO2ujfKeFDOk4'>Dual-running user support manual</a> for more information about the current process."
+
+      contact_info.concat(support_manual_info).html_safe
     end
 
     def last_action_details
@@ -61,6 +73,8 @@ module SupportInterface
                       'sent the initial emails'
                     elsif @match.reminder_emails_sent?
                       'sent the reminder emails'
+                    elsif @match.ucas_withdrawal_requested?
+                      'requested for the candidate to be removed from UCAS'
                     end
 
       "We #{last_action} on the #{@match.candidate_last_contacted_at.to_s(:govuk_date_and_time)}"
--- a/app/components/support_interface/ucas_match_action_component.rb
+++ b/app/components/support_interface/ucas_match_action_component.rb
@@ -8,6 +8,11 @@ module SupportInterface
         button_text: 'Confirm initial emails were sent',
         form_path: :support_interface_record_initial_emails_sent_path,
       },
+      send_reminder_emails: {
+        description: 'Send reminder emails',
+        button_text: 'Confirm reminder emails were sent',
+        form_path: :support_interface_record_reminder_emails_sent_path,
+      },
     }.freeze
 
     def initialize(match)
@@ -38,6 +43,8 @@ module SupportInterface
     def next_action
       if @match.candidate_last_contacted_at.nil?
         :send_initial_emails
+      elsif @match.initial_emails_sent? && @match.need_to_send_reminder_emails?
+        :send_reminder_emails
       end
     end
 
@@ -52,6 +59,8 @@ module SupportInterface
     def last_action_details
       last_action = if @match.initial_emails_sent?
                       'sent the initial emails'
+                    elsif @match.reminder_emails_sent?
+                      'sent the reminder emails'
                     end
 
       "We #{last_action} on the #{@match.candidate_last_contacted_at.to_s(:govuk_date_and_time)}"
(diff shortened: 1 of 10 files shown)


app/controllers/support_interface/ucas_matches_controller.rb
```diff
@@ -15,26 +15,40 @@ def audit
...
+      match.update!(
         candidate_last_contacted_at: Time.zone.now,
+        action_taken: 'reminder_emails_sent',
       )
-        flash[:success] = 'The date of the initial emails was recorded'
-      else
-        flash[:warning] = 'There was a problem and the date of the initial emails was not recorded'
+      flash[:success] = 'The date of the reminder emails was recorded'
+      redirect_to support_interface_ucas_match_path(match)
+    end
```
despo, line 36: @AgaDufrat the card specifies that the `ChaserSent` model should be used when reminder emails are sent, so I think this should take place here, and record the interaction.

AgaDufrat: Sending emails automatically will be done in the next iteration. Sorry that the card is not accurate!
At the moment support agents send all the emails via Zendesk and only confirm/record that they performed the action by clicking the buttons in Apply service.

despo: 👍

despo (approved):
LGTM

app/components/support_interface/ucas_match_action_component.rb
```diff
@@ -0,0 +1,91 @@
...
+        description: 'Request withdrawal from UCAS',
+        button_text: 'Confirm withdrawal from UCAS was requested',
+        form_path: :support_interface_record_ucas_withdrawal_requested_path,
+        instructions: "We need to contact UCAS. Please send an encrypted file with the candidate's duplicate application details to Harry Haines (h.haines@ucas.ac.uk) and Lizzy Carter (l.carter@ucas.ac.uk) from UCAS to ask them to remove the candidate from UTT.",
+      },
+      confirm_withdrawal_from_ucas: {
+        description: 'Confirm withdrawal from UCAS',
+        button_text: 'Confirm the application was withdrawn from UCAS',
+        form_path: :support_interface_process_match_path,
+        instructions: "We need to ensure that UCAS have removed the candidate's duplicate application from UTT by the end of the day. Continue to contact UCAS until this has been confirmed.",
```
duncanjbrown, line 26: Does "by the end of the day" make sense here — what if they clicked the above button late in the afternoon?

app/components/support_interface/ucas_match_action_component.rb
```diff
@@ -0,0 +1,91 @@
...
+    def button
+      {
+        text: ACTIONS[next_action][:button_text],
+        path: Rails.application.routes.url_helpers.send(ACTIONS[next_action][:form_path], @match),
+      }
+    end
+
+  private
+
+    def next_action
```
duncanjbrown, line 55: I think this logic probably belongs on the model? It takes the keys further away from the hash, but it also keeps the logic about how matches behave tied up in one place

app/components/support_interface/ucas_match_action_component.rb
```diff
@@ -0,0 +1,91 @@
...
+        :confirm_withdrawal_from_ucas
+      end
+    end
+
+    def type_of_action
+      "<strong class='govuk-tag govuk-tag--yellow app-tag'>Action needed</strong> #{ACTIONS[next_action][:description]}".html_safe
+    end
+
+    def required_action_details
+      contact_info = ACTIONS[next_action][:instructions] ||
```
duncanjbrown, line 72: Better to turn this string into a variable, I think, and put it explicitly into the `ACTIONS` hash for the two actions that need it?

duncanjbrown: Is `contact_info` the right name for this variable — should it be `instructions`?

app/components/support_interface/ucas_match_action_component.rb
```diff
@@ -0,0 +1,91 @@
...
+        "We need to contact the candidate and the provider. Use the appropriate Zendesk macro. Alternatively, you can use the appropriate template from <a class='govuk-link' href='https://docs.google.com/document/d/1s5ql4jNUUr3QDPUQYWImkZR6o8upvutrjOEuoZ0qqTE'>this document</a>."
+      support_manual_info = "<br><br>Please refer to <a class='govuk-link' href='https://docs.google.com/document/d/1XvZiD8_ng_aG_7nvDGuJ9JIdPu6pFdCO2ujfKeFDOk4'>Dual-running user support manual</a> for more information about the current process."
+
+      contact_info.concat(support_manual_info).html_safe
+    end
+
+    def last_action_details
```
duncanjbrown, line 79: Again, this logic could live in the model? You could return a `last_action` symbol and include a `past_tense_description` in the hashes, so eg this method would be sth like

```
last_action = @match.last_action
"We #{ACTIONS[last_action][:past_tense_description]} on the #{@match.candidate_last_contacted_at.to_s(:govuk_date_and_time)}"
```

duncanjbrown (commented):
This is great — thanks @AgaDufrat ! I'd like to see if we go for one more refactor to push some logic back up into the model out of the component, which I think could simplify things a bit further.

app/components/support_interface/ucas_match_action_component.rb
AgaDufrat: This is the copy I got from Natasha 🤷‍♀️  She said that UCAS should remove them immediately.

app/components/support_interface/ucas_match_action_component.rb
```diff
@@ -0,0 +1,60 @@
...
+        form_path: :support_interface_record_initial_emails_sent_path,
+      },
+    }.freeze
+
+    def initialize(match)
+      @match = match
+    end
+
+    def inset_text_header
+      return 'No action required' if !@match.dual_application_or_dual_acceptance? || !@match.action_needed?
```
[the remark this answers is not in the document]
AgaDufrat: I wasn't sure about this. I could just do `unless @match.action_needed?` here if I changed the `action_needed?` to something like:

```
  def action_needed?
    return false if processed?

    return false unless dual_application_or_dual_acceptance?

    return true if ucas_withdrawal_requested?

    return need_to_send_reminder_emails? if initial_emails_sent?

    return need_to_request_withdrawal_from_ucas? if reminder_emails_sent?

    true
  end
```
We use `action_needed?` a lot on the index page, in the filters, for slack notifications, so with performance in mind I wanted to keep the `dual_application_or_dual_acceptance?` as the last check in this method since it goes through matching data of each UCAS match. 
But on that line we actually need to check `dual_application_or_dual_acceptance?` first.

Do you think I should change it as shown above or create some other method to capture `!@match.dual_application_or_dual_acceptance? || !@match.action_needed?`? 🤔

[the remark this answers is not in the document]
AgaDufrat: Added a commit d2d8ab7

duncanjbrown (approved):
🚀

PR #3394, opened 2020-11-11
Make sure autoselect is disabled on autocompletes using showAllValues option

paulrobertlloyd:
## Context

The service is using both `showAllValues` and `autoselect` - these should be mutually exclusive options as using them together will cause usability issues. `autoselect` should be false by default according to the docs, but in reality this is not the case.

* Add `autoselect: false` to all autocompletes that have `showAllValues: true`
* Aligns the implementations where we enhance selects to use autocompletes to use the same variable names (`id`, `selectElement`, etc.( and order options after `selectElement` option alphabetically, etc.

## Link to Trello card

https://trello.com/c/ozg6Qw8s/

## Things to check

### DIFF
--- a/app/frontend/packs/api-token-autocomplete.js
+++ b/app/frontend/packs/api-token-autocomplete.js
@@ -9,8 +9,9 @@ const initApiTokenProviderAutocomplete = () => {
 
     accessibleAutocomplete.enhanceSelectElement({
       selectElement: apiTokenProviderSelect,
-      showAllValues: true,
-      confirmOnBlur: false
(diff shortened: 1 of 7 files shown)


duncanjbrown:
Upstream bug from @edwardhorsford 🙏 https://github.com/alphagov/accessible-autocomplete/issues/466

tvararu (approved):

paulrobertlloyd:
Thank you @tvararu. Been an absolute pleasure working with you 💔

PR #3397, opened 2020-11-12
Degree Flow Refactorings

malcolmbaig:
## Context

Some small refactors identified during https://trello.com/c/zeuemNFz

See each commit - they're self-contained so can be individually discarded if there's any disagreement/objections around these changes.

## Guidance to review
View the diff, per-commit recommended.

## Link to Trello card
na

## Things to check

### DIFF
--- a/spec/system/candidate_interface/candidate_entering_degrees_spec.rb
+++ b/spec/system/candidate_interface/candidate_entering_degrees_spec.rb
@@ -423,9 +423,4 @@ private
       .find { |row| row.has_link?(link_text) }
       .click_link(link_text)
   end
-
-  def expect_validation_error(message)
-    errors = all('.govuk-error-message')
-    expect(errors.map(&:text).one? { |e| e.include? message }).to eq true
-  end
 end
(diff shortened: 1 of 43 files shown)


spec/system/candidate_interface/candidate_entering_degrees_spec.rb
```diff
@@ -287,7 +287,10 @@ def when_i_click_to_change_my_undergraduate_degree_type
   end
 
   def when_i_click_to_change_my_undergraduate_degree_year
-    click_change_link('year')
+    start_year_row = find('.govuk-summary-list__row', text: 'Start year')
+    within start_year_row do
+      click_change_link('year')
+    end
```
stevehook, line 293: :+1: Made me wonder, if this is a common pattern, could you go further with a helper that looks something like: `click_change_link('year', within: 'Start year')`? Maybe it's rare enough that it doesn't make sense.

stevehook (approved):
all really good stuff, no objections from me!

spec/system/candidate_interface/candidate_entering_degrees_spec.rb
malcolmbaig: Like that idea. Tested it by going through and using this helper across the candidate interface specs. I found the need to scope wasn't that common (and in a few cases, it was just as clear to leave the `within` call in the spec itself). 

Pushed those replacements in the newest commit. WDYT?

stevehook: Looks good to me, I don't think it's worth adding the extra complexity to the helper for a rare case.

stevehook (approved):

tvararu (approved):

PR #3400, opened 2020-11-12
Filter UCAS matches by Action needed

AgaDufrat:
## Context

With the increasing number of UCAS matches support agents need to be able to quickly see which ones require their actions.

Screenshot 2020-11-12 at 13 32 03

## Guidance to review

- visit `/support/ucas-matches`
- apply filters

## Link to Trello card

https://trello.com/c/B81XXk56/3010-filter-ucas-matches-by-action-needed-flag

## Things to check

### DIFF
--- a/app/models/support_interface/ucas_matches_filter.rb
+++ b/app/models/support_interface/ucas_matches_filter.rb
@@ -11,11 +11,16 @@ module SupportInterface
         ucas_matches = ucas_matches.where('recruitment_cycle_year IN (?)', applied_filters[:years])
       end
 
+      if applied_filters[:action_needed]&.include?('yes')
+        action_needed_ids = ucas_matches.select(&:action_needed?).map(&:id)
+        ucas_matches = ucas_matches.where(id: action_needed_ids)
+      end
+
       ucas_matches
     end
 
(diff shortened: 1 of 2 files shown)


app/models/support_interface/ucas_matches_filter.rb
```diff
@@ -36,5 +41,18 @@ def year_filter
...
+
+    def action_needed_filter
+      {
+        type: :checkboxes,
+        heading: 'Action needed',
+        name: 'action_needed',
+        options: [{
+          value: 'yes',
+          label: 'Yes',
+          checked: applied_filters[:action_needed]&.include?('yes'),
```
benswannack, line 53: Very minor, but you could change the URL to end with `&action_needed=anything` and have the filter apply but the box not be checked. Can we make the check for applying the filter and checking this be the same?

benswannack (approved):
Minor thing inline, otherwise looks good :)

app/models/support_interface/ucas_matches_filter.rb
AgaDufrat: Good spot, amended.

PR #3403, opened 2020-11-12
Update Application Feedback prompts to use the updated design

davidgisbey:
## Context

We're now simply having a link at the bottom of the specified pages which opens a feedback form in a new tab.

Application Feedback Component
|Before|After|
|--------------|--------------|
|image|image|

## Guidance to review

**This does rely on migrations. The not null on section was breaking the tests so I've just added it to this PR for ease**

This should render on all of the pages in the following sections:

 References
* Degree
* GCSEs
* Other qualifications
* Personal statement
* Subject knowledge

## Link to Trello card

https://trello.com/c/AvxRVqld/2465-%F0%9F%93%9D-dev-add-feedback-prompt-apply

## Things to check

- [] This code does not rely on migrations in the same Pull Request


frankieroberto:
Small bug:

If you leave the form blank, you get a validation error on the "Can we contact you about your feedback?" radio. But if you then fill out the form, and press submit feedback, it fails and just reloads the page.

I think this is because the re-rendered form from the validation error doesn't persist the hidden fields containing the `path` and `page_title` fields?

davidgisbey:
As discussed that was fixed in https://github.com/DFE-Digital/apply-for-teacher-training/pull/3403/commits/967e76b7a4ec46458e6207f11f1e71bb278d4ad4 👍

tvararu (approved):
:ship:

PR #3406, opened 2020-11-13
Add provider contact details to UCAS match

despo:
## Context

Exposing provider user contact details so that support agents can easily get in touch with them as part of resolving dual applications.

#### Before
before

#### After
after-collapsed
Collapsed

after-expanded-amended
Expanded

## Guidance to review

Navigate to http://localhost:3000/support/ucas-matches and select a UCAS match.
You should be able to see this change under Course Choices

## Link to Trello card

https://trello.com/c/sg3H1Wp0/3016-add-provider-contact-details-to-ucas-match-pages

## Things to check

### DIFF
--- a/app/components/support_interface/ucas_match_table_component.html.erb
+++ b/app/components/support_interface/ucas_match_table_component.html.erb
@@ -14,7 +14,28 @@
     <% table_rows.each do |table_row| %>
       <tr class="govuk-table__row">
         <td class="govuk-table__cell"><%= table_row[:course_code] %></td>
-        <td class="govuk-table__cell"><%= table_row[:course_details] %></td>
+        <td class="govuk-table__cell">
+          <%= table_row[:course_details] %>
+          <% if table_row[:course_provider_contacts].any? %>
+            <details class="govuk-details" data-module="govuk-details">
+              <summary class="govuk-details__summary">
+                <span class="govuk-details__summary-text">
+                  Contact details
+                </span>
+              </summary>
+              <div class="govuk-details__text govuk-!-padding-bottom-0">
+                <% table_row[:course_provider_contacts].each do |provider_contact| %>
+                  <dl class="app-provider_contacts_list govuk-!-margin-0 govuk-!-padding-bottom-2">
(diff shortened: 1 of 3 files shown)


app/components/support_interface/ucas_match_table_component.html.erb
```diff
@@ -14,7 +14,26 @@
...
+          <% if table_row[:provider_contacts].any? %>
+            <details class="govuk-details" data-module="govuk-details">
+              <summary class="govuk-details__summary">
+                <span class="govuk-details__summary-text">
+                  Contact details
+                </span>
+              </summary>
+              <div class="govuk-details__text">
+                <% table_row[:provider_contacts].each do |provider_contact| %>
+                  <%= render ProviderInterface::UserListCardComponent.new(
```
AgaDufrat, line 28: Now thinking about it, I don't think we should be using this component. Since the link goes to the provider interface, which will show the signed in support users "Access denied". 
Maybe we should just make it a really simple like a comma separated list the email addresses.

app/components/support_interface/ucas_match_table_component.rb
```diff
@@ -58,5 +59,9 @@ def course_code(course)
     def course_details(course)
       "#{course.name} – #{course.provider&.name || 'Provider not on Apply'}"
     end
+
+    def provider_contacts(course)
+      course.provider.present? ? course.provider.provider_users : []
```
AgaDufrat, line 64: I think we want to scope it to manage_users ProviderPermissions.

app/components/support_interface/ucas_match_table_component.html.erb
despo: @AgaDufrat Yes that makes sense, I am not very familiar with persmissions so I missed that. 

I do however think  that displaying a similar formatted list would make sense so they have both the name as well as the email of the person they are contacting. Maybe a non-bulleted list. I'll check the UI GDS components and see what fits best.

AgaDufrat (approved):
LGTM

PR #3411, opened 2020-11-13
Update export to include more detailed information about org permissions

benswannack:
## Context

We are interested in who is making changes to org permissions and what they are changing, and which providers are having their permissions changed and by whom

Updates to the existing export, better column names and more data

## Guidance to review
Any edge cases we should think about? Can these column names be made shorter while retaining the meaning?

## Link to Trello card
https://trello.com/c/cc3OZ3qd/3013-add-to-and-clarify-org-permissions-data-in-the-provider-access-controls-export

## Things to check


duncanjbrown (approved):
Very keen on these _extremely_ explicit variable names!

PR #3412, opened 2020-11-15
Slack: consolidate offer-accepted message, comma off thousands

duncanjbrown:
## Context

We get up to three messages per offer acceptance in Slack, as the candidate might automatically decline and withdraw at the same time.

Also, numbers in the thousands are difficult to read without commas, and we'll soon hit 10,000 candidates (!).

- Move the `without_slack_message_sending` method onto `StateChangeNotifier` where it belongs and give it a better name
- Consolidate the offer-accepted message
- Comma off thousands

I think `offer accepted` and `application sent to provider` are the only state changes entailing multiple messages, so I didn't make a generic `buffer_slack_messages` method. (That might be been usable in both cases but the messages would have been less readable without some fiddly extra work of the sort already in this PR but more unpleasantly generic).

## Guidance to review

Tests should prove it works 🤞

## Things to check

### DIFF
--- a/app/lib/state_change_notifier.rb
+++ b/app/lib/state_change_notifier.rb
@@ -23,6 +23,33 @@ class StateChangeNotifier
     send(message, url)
   end
 
+  def self.accept_offer(accepted:, withdrawn: [], declined: [])
+    accepted_msg = "#{accepted.application_form.first_name} has accepted #{accepted.offered_option.course.provider.name}’s offer for #{accepted.offered_option.course.name_and_code}"
+
+    if withdrawn.any?
+      withdrawn_msg = "withdrawn their #{'application'.pluralize(withdrawn.count)} for "
+      withdrawals = withdrawn.map do |ac|
+        "#{ac.offered_option.course.name_and_code} at #{ac.offered_option.course.provider.name}"
+      end
+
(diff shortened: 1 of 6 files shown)


app/lib/state_change_notifier.rb
```diff
@@ -23,6 +26,33 @@ def self.submit_application(application_form)
...
+        "#{ac.offered_option.course.name_and_code} at #{ac.offered_option.course.provider.name}"
+      end
+
+      withdrawn_msg += withdrawals.to_sentence
+    end
+
+    if declined.any?
+      declined_msg = 'declined '
+      declines = declined.map do |ac|
+        "#{ac.offered_option.course.provider.name}’s offer on #{ac.offered_option.course.name_and_code}"
```
benswannack, line 44: Is `offer on x` the right wording here? I'd expect it to read `offer for x`. (Same thing for accepted message above)

benswannack (approved):
Nice 👍

app/lib/state_change_notifier.rb
duncanjbrown: Good point — looks like we use `for` elsewhere. I'll update :+1:

duncanjbrown: Updated

PR #3419, opened 2020-11-16
Add Sidekiq job ID and logging to data exports

duncanjbrown:
Some data exports mysteriously failed on production. It's not clear whether this happened because the job wasn't enqueued properly, whether it silently died when it ran, or for some some other reason.

Couldn't reproduce locally, so will start adding instrumentation until we know more about what happened here. This PR: 

- adds a sidekiq job ID to the data export. This might help us track down the job if something goes wrong — or at least assure us that it was enqueued.
- adds logging inside the Sidekiq worker so we can catch any silent failures

## Context

https://ukgovernmentdfe.slack.com/archives/CN1MCQCHZ/p1605518196399900

### DIFF
--- a/app/workers/data_exporter.rb
+++ b/app/workers/data_exporter.rb
@@ -2,16 +2,23 @@ class DataExporter
   include Sidekiq::Worker
 
   def perform(importer_class, data_export_id)
+    RequestLocals.store[:debugging_info] = { data_export_id: data_export_id, importer_class: importer_class }
+
+    Rails.logger.info 'Sidekiq running. Loading data export record'
     data_export = DataExport.find(data_export_id)
 
+    Rails.logger.info 'Started CSV generation'
     csv_data = generate_csv(
       importer_class.constantize.new.data_for_export,
     )
+    Rails.logger.info 'Finished CSV generation'
 
+    Rails.logger.info 'Started writing CSV'
     data_export.update!(
       data: csv_data,
       completed_at: Time.zone.now,
     )
+    Rails.logger.info 'Finished writing CSV. Sidekiq done'
   end
 
 private


app/workers/data_exporter.rb
```diff
@@ -2,16 +2,21 @@ class DataExporter
   include Sidekiq::Worker
 
   def perform(importer_class, data_export_id)
+    Rails.logger.info "Data export ##{data_export_id} — #{importer_class}: Sidekiq running. Loading DataExport record"
```
tijmenb, line 5: Maybe picky but it feels wrong to use Sidekiq internals. I'm wondering if we can track the job without using the JID.

Perhaps could add structured logging in the job so you can look up Sidekiq logs by adding this at the start:

```
RequestLocals.store[:debugging_info] = { data_export_id: data_export_id, importer_class: importer_class }
```

This will be picked up by logstash: https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/lib/logstash_logging.rb#L70-L72

Would that prevent us from needing the `sidekiq_jid`? It'll also remove the need for prefixing the log lines with "Data export ##{data_export_id} — #{importer_class}".

duncanjbrown: Thank you, great — didn't know about `debugging _info`! I'll lose the JID reference. I guess we'll know by absence if the job was never enqueued.

tijmenb: Actually I'm now wondering if we can just add the params to the log by default, so we'll have this for all Sidekiq logs. Some middleware?

duncanjbrown: This is now ready to go I think — will return to that later if time

tijmenb (approved):

tijmenb:
Woopsie, I think this is leaking between sidekiq jobs:

https://kibana.logit.io/app/kibana#/doc/8ac115c0-aac1-11e8-88ea-0383c11b333a/logstash-2020.11.19?id=fhuA33UB2W7S_aTlHlYj&_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-4h,to:now))

I think we need to clear the variables we set before each job or something like that?

PR #3422, opened 2020-11-16
Do not include diversity information in HESA export if user does not have permission to view it

benswannack:
## Context
Currently, diversity info is included in this export regardless of whether or not the provider user who generated it has permission to view that information.

#### Added descriptive text about why `confidential` will show up in some reports
Content here updated as per discussion on card
image

## Link to Trello card
https://trello.com/c/p36aFgu9/2982-restrict-diversity-information-in-hesa-export

## Things to check

### DIFF
--- a/app/controllers/provider_interface/applications_exports_controller.rb
+++ b/app/controllers/provider_interface/applications_exports_controller.rb
@@ -5,10 +5,7 @@ module ProviderInterface
     def export
       respond_to do |format|
         format.csv do
-          csv_data = HesaDataExport.new(
-            provider_ids: current_provider_user.providers.map(&:id),
-            actor: current_provider_user,
-          ).call
+          csv_data = HesaDataExport.new(actor: current_provider_user).call
           send_data csv_data, disposition: 'attachment', filename: csv_filename
         end
       end
--- a/app/controllers/provider_interface/applications_exports_controller.rb
+++ b/app/controllers/provider_interface/applications_exports_controller.rb
@@ -5,7 +5,10 @@ module ProviderInterface
     def export
       respond_to do |format|
         format.csv do
-          csv_data = HesaDataExport.new(provider_ids: current_provider_user.providers.map(&:id)).call
+          csv_data = HesaDataExport.new(
+            provider_ids: current_provider_user.providers.map(&:id),
+            actor: current_provider_user,
+          ).call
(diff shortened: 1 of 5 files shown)


app/controllers/provider_interface/applications_exports_controller.rb
```diff
@@ -5,7 +5,10 @@ class ApplicationsExportsController < ProviderInterfaceController
     def export
       respond_to do |format|
         format.csv do
-          csv_data = HesaDataExport.new(provider_ids: current_provider_user.providers.map(&:id)).call
+          csv_data = HesaDataExport.new(
+            provider_ids: current_provider_user.providers.map(&:id),
+            actor: current_provider_user,
```
benswannack, line 10: We can just pass in actor here to simplify the interface.

When I did that and edited the tests accordingly, I realised that currently only members of the training provider can generate an export with applications to their courses (since we check the application choice `provider` and not the `accredited_provider`).

Do we want to expand to allow ratifiers to generate this export?

steventux (approved):
:+1: LGTM good to have permissions checks as part of this now.

app/services/provider_interface/hesa_data_export.rb
```diff
@@ -1,22 +1,26 @@
...
           :provider,
           :site,
           course_option: { course: %i[provider accredited_provider] },
           application_form: %i[candidate application_qualifications],
-        ).where(
+        )
+
+      applications =
+        applications_join.where('providers.id': @provider_ids)
+        .or(applications_join.where('courses_course_options.accredited_provider_id': @provider_ids))
```
duncanjbrown, line 20: Could we use `GetApplicationChoicesForProviders` here? We could pass in the `includes` if necessary

app/services/provider_interface/hesa_data_export.rb
```diff
@@ -1,22 +1,26 @@
 module ProviderInterface
   class HesaDataExport
-    def initialize(provider_ids:)
-      @provider_ids = provider_ids
+    def initialize(actor:)
+      @provider_ids = actor.providers.map(&:id)
+      @auth = ProviderAuthorisation.new(actor: actor)
```
duncanjbrown, line 5: I think `@actor.authorisation` will do the job here as this will always be a `ProviderUser`?

PR #3423, opened 2020-11-16
Extract logic about references into a single class

tijmenb:
## Context

The references logic is fairly complicated and we're seeing a number of bugs. One of these was fixed in https://github.com/DFE-Digital/apply-for-teacher-training/pull/3418. 

To start solving the issues, this is a big refactor PR. The goal of it is to create a single class, ReferenceActionsPolicy, which can answer all questions about whether a reference can be deleted, destroyed (yes), sent, cancelled, reminded, retried and more.

This is limited to refactoring. For manual testing it also exposes the policy booleans to the support UI.

Next steps:

- Find where we duplicate logic from the policy class and replace it
- Update the `editable?` to disallow edits when there are enough references
- Use the policy object consistently in controllers so that it's impossible to request a reference by fiddling with the URL (https://github.com/DFE-Digital/apply-for-teacher-training/pull/3418 only does the view)

## Guidance to review

Does the end result make sense?

## Link to Trello card

https://trello.com/c/9FgC13b6

## Things to check

### DIFF
--- a/spec/services/reference_actions_policy_spec.rb
+++ b/spec/services/reference_actions_policy_spec.rb
@@ -1,0 +1,90 @@
+require 'rails_helper'
+
+RSpec.describe ReferenceActionsPolicy do
+  describe '#editable?' do
+    it 'returns true for `not_requested_yet`' do
+      reference = build(:reference, :not_requested_yet)
+
+      expect(policy(reference).editable?).to eq true
+    end
+
+    it 'returns false for all other statuses' do
+      reference = build(:reference, :feedback_provided)
+
+      expect(policy(reference).editable?).to eq false
+    end
+  end
+
+  describe '#can_send_reminder?' do
+    it 'is true when state is feedback_requested and reminder_sent_at is nil' do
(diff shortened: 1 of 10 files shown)


spec/services/reference_actions_policy_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+    end
+
+    it 'returns false when the reference is in a state that cannot be deleted' do
+      invalid_states.each do |state|
+        reference = build_stubbed(:reference, state, application_form: unsubmitted_application_form)
+        expect(policy(reference).request_can_be_deleted?).to eq false
+      end
+    end
+
+    it 'is false when in a state that has been can be deleted, but the the application form has been submitted' do
```
davidgisbey, line 80: is false when in a state that can be deleted and the application form has been submitted

davidgisbey (approved):
Lovely stuff. Just one comment slamming my own work 👍

spec/services/reference_actions_policy_spec.rb
tijmenb: Will fix in follow up!

PR #3424, opened 2020-11-16
Introduce ApplicationChoice#self_and_siblings

duncanjbrown:
Dry up a common pattern where we need to go from a single application choice to that choice plus the other choices on the form


benswannack (approved):
Tidy 😁

PR #3427, opened 2020-11-16
Prevent editing of references in applications that have enough references

tijmenb:
##  Context

Similar to https://github.com/DFE-Digital/apply-for-teacher-training/pull/3418, we want to prevent candidates from editing references in applications that already have 2 references.

Updated `editable?`, which _should_ be used across the app to determine editability.

## Guidance to review

Before:

image

After:

image

## Link to Trello card

https://trello.com/c/9FgC13b6

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/references/email_address_controller.rb
+++ b/app/controllers/candidate_interface/references/email_address_controller.rb
@@ -1,7 +1,8 @@
 module CandidateInterface
   module References
     class EmailAddressController < BaseController
-      before_action :set_reference, :redirect_to_review_page_unless_reference_is_not_requested_or_email_bounced
+      before_action :set_reference
+      before_action :verify_email_is_editable
 
       def new
         @reference_email_address_form = Reference::RefereeEmailAddressForm.new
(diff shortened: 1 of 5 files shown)


app/controllers/candidate_interface/references/email_address_controller.rb
```diff
@@ -46,7 +46,7 @@ def referee_email_address_param
       end
 
       def redirect_to_review_page_unless_reference_is_not_requested_or_email_bounced
-        unless @reference.present? && (@reference.not_requested_yet? || @reference.email_bounced?)
+        unless @reference.present? && (@reference.editable? || @reference.email_bounced?)
```
malcolmbaig, line 49: Needs the policy instance here, looks like the cause of the test failures too.

spec/services/reference_actions_policy_spec.rb
```diff
@@ -2,13 +2,21 @@
 
 RSpec.describe ReferenceActionsPolicy do
```
malcolmbaig, line 3: This class is :ok_hand:

app/controllers/candidate_interface/references/email_address_controller.rb
tijmenb: Thanks - updated!

tijmenb:
I've removed the step from the review system test where we test the editing of references, because this isn't possible anymore. In a follow up PR I'll try to restructure the system tests a bit, because I suspect that we're missing some scenarios now.

malcolmbaig (approved):

PR #3428, opened 2020-11-16
Prevent unsendable references from being sent

tijmenb:
## Context

It's currently possible to trick the system into sending references requests for any existing reference, even if it's already sent or if the application already enough references.

Use the new policy object from #3423 to restrict the controller. 

## Guidance to review

Does the logic make sense?

## Link to Trello card

https://trello.com/c/9FgC13b6

### DIFF
--- a/app/components/support_interface/reference_with_feedback_component.rb
+++ b/app/components/support_interface/reference_with_feedback_component.rb
@@ -131,6 +131,10 @@ module SupportInterface
           value: policy.can_send_reminder? ? 'Yes' : 'No',
         },
         {
+          key: 'Can request?',
+          value: policy.can_request? ? 'Yes' : 'No',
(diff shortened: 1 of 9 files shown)


stevehook (approved):
:+1: Review app does what it's supposed to do.

There is a bit of overlap between `ReferenceActionsPolicy` and `RequestReference`. Maybe `RequestReference` should call `ReferenceActionsPolicy` rather than using it's own conditional logic?

tijmenb:
@stevehook nice catch. I've made it use the policy logic instead. Because it's checked elsewhere already, we're now raising an exception if we encounter a unrequestable reference here. This saves us from having to pass in the `flash`.

stevehook (approved):
:+1:

PR #3431, opened 2020-11-17
Add autocomplete to Other Qualifications grades

malcolmbaig:
## Context

As a candidate
I need to… provide the required information efficiently and accurately
So that… I can complete my application
Add a list of grades (extracted from the Apply prototype), and use these to enable autocomplete on the Other Qualifications grade field. This autocomplete only kicks in if the chosen qualification is one of the Other UK types.

Peek 2020-11-17 09-21

## Guidance to review
- Add one of the Other UK qualifications in the review app and start entering one of the grade values.

## Link to Trello card
https://trello.com/c/iTEdO0rL

## Things to check


davidgisbey (approved):
Beautiful stuff. This makes it so easy to add the GCSE/ level data in too

app/controllers/candidate_interface/other_qualifications/details_controller.rb
```diff
@@ -7,13 +7,15 @@ def new
         current_application.application_qualifications.other.order(:created_at),
       )
       @wizard.save_state!
-      set_subject_data_for_autocomplete
+      set_subject_autocomplete_data
+      set_grade_autocomplete_data
```
malcolmbaig, line 11: I think this data could live in the wizard or (more appropriately) a Details form object that we might extract. Because this code is currently in flux, I'm putting these methods in the controller for now. (fyi: @stevehook )

paulrobertlloyd:
Can we add `govuk-input--width-10` to this grade input (and the autocomplete that replaces it)?

stevehook (approved):
:+1: Review app works.

The list of possible values isn't meant to be comprehensive I guess? (It doesn't match the grades for 'GCE O level' for example, but candidates are not forced to pick from the autocomplete.)

malcolmbaig:
> Can we add `govuk-input--width-10` to this grade input (and the autocomplete that replaces it)?

@paulrobertlloyd added in latest commit.

paulrobertlloyd (approved):

PR #3441, opened 2020-11-17
Refactor OtherQualificationWizard

stevehook:
## Context

After converting the candidate interface other qualification forms to use a wizard that defers saving changes until the last step we identified a few areas that could be refactored to reduce complexity and make it easier to apply the same change to other parts of the candidate interface.

At present this is a DRAFT PR - it's not ready to merge but raised here to get feedback about whether the proposed changes are worthwhile.

The `OtherQualificationWizard` class is quite big despite being just a two step form. So our goals was to reduce the size and complexity of this class. To attempt to do that I've extracted `OtherQualificationsTypeForm` from the wizard to encapsulate the attributes and validations for the first of the two steps (with a view to adding an `OtherQualificationsDetailForm` for the second step).

The wizard responsibilities are:
- page flow rules (what's the current and next step)
- persisting the uncommitted state of the wizard (collecting form input and storing in Redis) until reaching the final step when it's stored in the database.

The form responsibilities are:
- declaring the attributes that are mapped to each form
- validation rules
  
## Guidance to review

- I've kept the interface between controller and wizard as it was because the controller needs to use the wizard to define the flow between pages so it made more sense to encapsulate the sub-forms within the wizard rather than vice versa. For now the wizard remains the object that the front-end views are mapped to.
- Both the form and wizard classes import `ActiveModel::Model` because they don't really have any purpose without having 'attributes'. This is the main reason I suppose that the original implementation didn't split the two. The form(s) are basically nested form objects in the wizard - this is similar(ish) to the pattern that we've adopted in https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/forms/provider_interface/provider_relationship_permissions_setup_wizard.rb

## Link to Trello card

https://trello.com/c/YWtsBdnu/2506-refactor-other-qualifications-wizard-implementation

## Things to check

### DIFF
--- a/app/forms/candidate_interface/other_qualification_type_form.rb
+++ b/app/forms/candidate_interface/other_qualification_type_form.rb
@@ -1,0 +1,15 @@
+module CandidateInterface
+  class OtherQualificationTypeForm
+    include ActiveModel::Model
+    include ActiveModel::Attributes
+
+    attribute :qualification_type
+    attribute :other_uk_qualification_type
+    attribute :non_uk_qualification_type
+
+    validates :qualification_type, presence: true
+    validates :qualification_type, inclusion: { in: ['A level', 'AS level', 'GCSE', 'Other', 'non_uk'], allow_blank: false }
+    validates :other_uk_qualification_type, presence: true, if: -> { qualification_type == 'Other' && FeatureFlag.active?('international_other_qualifications') }
+    validates :non_uk_qualification_type, presence: true, if: -> { qualification_type == 'non_uk' }
+  end
+end
--- a/app/forms/candidate_interface/other_qualification_wizard.rb
+++ b/app/forms/candidate_interface/other_qualification_wizard.rb
@@ -4,22 +4,21 @@ module CandidateInterface
     include ValidationUtils
 
     attr_accessor :current_step, :current_other_qualification_id, :checking_answers
-    attr_accessor :qualification_type, :other_uk_qualification_type, :non_uk_qualification_type
+
+    attr_accessor :type_form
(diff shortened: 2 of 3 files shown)


app/forms/candidate_interface/other_qualification_type_form.rb
```diff
@@ -0,0 +1,15 @@
...
+    include ActiveModel::Model
+    include ActiveModel::Attributes
+
+    attribute :qualification_type
+    attribute :other_uk_qualification_type
+    attribute :non_uk_qualification_type
+
+    validates :qualification_type, presence: true
+    validates :qualification_type, inclusion: { in: ['A level', 'AS level', 'GCSE', 'Other', 'non_uk'], allow_blank: false }
+    validates :other_uk_qualification_type, presence: true, if: -> { qualification_type == 'Other' && FeatureFlag.active?('international_other_qualifications') }
```
davidgisbey, line 12: I think this flag has gone now so it will need a rebase

app/forms/candidate_interface/other_qualification_wizard.rb
```diff
@@ -4,22 +4,21 @@ class OtherQualificationWizard
     include ValidationUtils
 
     attr_accessor :current_step, :current_other_qualification_id, :checking_answers
-    attr_accessor :qualification_type, :other_uk_qualification_type, :non_uk_qualification_type
+
+    attr_accessor :type_form
+    delegate :qualification_type, :qualification_type=, :other_uk_qualification_type, :other_uk_qualification_type=, :non_uk_qualification_type, :non_uk_qualification_type=, to: :type_form
```
davidgisbey, line 9: I like the idea of delegating validations to reduce cognitive complexity

stevehook:
Closing this for now following discussion on Slack. Trying a different approach with controllers interfacing with forms and the wizard (for want of a better name) being stripped back to be a simple intermediate data store service.

PR #3444, opened 2020-11-17
Fix org-level authorisation for users of both providers

mnacos:
## Context

If a user belongs to both the ratifying provider and the training provider, they should have permission to do something if either of their providers allows it. Currently, this is not the case if the course is ratified and only the ratifying provider is allowed to e.g. `make_decisions`.

The relevant code was recently refactored in https://github.com/DFE-Digital/apply-for-teacher-training/pull/3338.

Fix this.

## Guidance to review

The new code is a bit complicated, but more thorough. It even checks whether the user-level `make_decisions`, for example, is on the same side as the provider allowed org-level `make_decisions` when a user belongs to both providers. This scenario is extreme, but could happen, i.e. user belongs to both providers, ratifying provider is the one with org-level `make_decisions` but the user has `make_decisions` only for the training provider.

## Link to Trello card

No card.

## Things to check

### DIFF
--- a/app/services/provider_authorisation.rb
+++ b/app/services/provider_authorisation.rb
@@ -104,6 +104,10 @@ private
     )
   end
 
+  def user_level_can?(permission:, provider:)
+    @actor.provider_permissions.send(permission).exists?(provider: provider)
+  end
+
   def permission_as_training_provider_user?(permission:, course:)
     permission_name = "training_provider_can_#{permission}"
     relationship = relationship_for course: course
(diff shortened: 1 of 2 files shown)


duncanjbrown (approved):
My brain is now melting, but I think this is correct!

app/services/provider_authorisation.rb
```diff
@@ -129,23 +133,34 @@ def full_authorisation?(permission:, course:)
     # enforce user-level permissions
     errors << :requires_provider_user_permission unless
       @actor.is_a?(VendorApiUser) ||
-        @actor.provider_permissions.send(permission)
-          .exists?(provider: [training_provider, ratifying_provider].compact)
+        user_level_can?(permission: permission, provider: training_provider) ||
+        user_level_can?(permission: permission, provider: ratifying_provider)
 
     # enforce org-level permissions
     if ratifying_provider.present?
```
benswannack, line 140: What's the case where there is no ratifying provider - does that mean the course is ratified by the training provider? Why don't we need to enforce org-level permissions in this case?

app/services/provider_authorisation.rb
```diff
@@ -129,23 +133,34 @@ def full_authorisation?(permission:, course:)
...
+      # If user belongs to both providers, usually one of the two has org-level perm.
+      if @actor.providers.include?(ratifying_provider) && @actor.providers.include?(training_provider)
+        # If not, there is something wrong with the setup
+        if [training_provider_can, ratifying_provider_can].none?
+          errors << :requires_training_provider_permission
+          errors << :requires_ratifying_provider_permission
+        # Check org-level and user-level permissions match for ratifying provider
+        elsif !training_provider_can
+          errors << :requires_provider_user_permission unless
+            user_level_can?(permission: permission, provider: ratifying_provider)
```
benswannack, line 153: Do we need to check user permissions here again given we did it at the top of the method?

app/services/provider_authorisation.rb
```diff
@@ -129,23 +133,34 @@ def full_authorisation?(permission:, course:)
     # enforce user-level permissions
```
benswannack, line 133: I think it's worth breaking `full_authorisation?` down into multiple smaller methods with descriptive names. It's quite central to permissions so it'll be worth having some easy to digest code over something that's logically efficient.

app/services/provider_authorisation.rb
duncanjbrown: I thought the same thing, but I think this is necessary because the above check tests for permissions on either side, and we need to know that the user permissions align — eg above check could be true because you have permissions at the ratifier, but fail here because only the trainer has this permission

app/services/provider_authorisation.rb
duncanjbrown: In that case there is no relationship to consider

PR #3445, opened 2020-11-17
Support users can edit application phone numbers

duncanjbrown:
Add controller, form, route and view for editing applicant details. Only
editing phone number is supported (so far).

## Context

Easier for support staff and devs.

Screenshot 2020-11-17 at 17 54 15

Screenshot 2020-11-17 at 17 54 06

Screenshot 2020-11-17 at 18 03 48

## Guidance to review

a) Does it work properly?
b) I think it's ok for one field, but I don't really like adding the "Change" link to the shared component. Would rather decorate somehow? Ideas welcome.

## Link to Trello card

https://ukgovernmentdfe.slack.com/archives/CQA64BETU/p1605606666428400

## Things to check


app/components/personal_details_component.rb
```diff
@@ -51,10 +52,19 @@ def email_row
   end
 
   def phone_number_row
-    {
+    basic_row = {
       key: 'Phone number',
       value: phone_number || MISSING,
     }
+
+    if current_namespace == 'support_interface'
```
tijmenb, line 60: I think we should duplicate this component if we're making them namespace-specific. It's already a bit questionable that we share logic-heavy components between namespaces (see https://github.com/DFE-Digital/apply-for-teacher-training/pull/2071 for some considerations).

We could make the duplication less severe by pushing some stuff to the model.

app/controllers/support_interface/application_forms/applicant_details_controller.rb
```diff
@@ -0,0 +1,35 @@
+module SupportInterface
+  module ApplicationForms
+    class ApplicantDetailsController < SupportInterfaceController
```
tijmenb, line 3: Thoughts behind "Applicant"? Perhaps otherwise `CandidateDetailsController` or `EditApplicationFormController` would be more in line with the other code.

app/controllers/support_interface/application_forms/applicant_details_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+          ApplicationForm.find(params[:application_form_id]),
+        )
+      end
+
+      def update
+        @details = EditApplicantDetailsForm.new(
+          ApplicationForm.find(params[:application_form_id]),
+        )
+
+        @details.phone_number = edit_application_params[:phone_number]
```
tijmenb, line 15: This might work:

```suggestion
        @details.assign_attributes(edit_application_params)
```

It removes the duplication with `edit_application_params`, so that if we add new attributes they only need to be added once.

app/views/support_interface/application_forms/applicant_details/edit.html.erb
```diff
@@ -0,0 +1,16 @@
...
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with model: @details, url: support_interface_application_form_update_applicant_details_path do |f| %>
+      <%= f.govuk_error_summary %>
+
+      <%= f.govuk_fieldset legend: { text: 'Edit applicant details', size: 'l' } do %>
+        <%= f.govuk_text_field :phone_number, label: { text: 'Phone number' } %>
+      <% end %>
+
```
tijmenb, line 12: So I don't forget: we could add the audit comment field here where the support user can paste in a link to Zendesk, so that we have context on why the change was made.

config/locales/en.yml
```diff
@@ -212,6 +212,11 @@ en:
           attributes:
             further_conditions:
               too_long: "%{name} must be %{limit} characters or fewer"
+        support_interface/application_forms/edit_applicant_details_form:
```
tijmenb, line 215: Wondering if there's a way to dedupe this. Perhaps adding the text in `PhoneNumberValidator`? Not a blocker.

tijmenb (commented):
Very nice! This will be so useful.

app/components/personal_details_component.rb
duncanjbrown: Agreed!

app/controllers/support_interface/application_forms/applicant_details_controller.rb
duncanjbrown: I followed the UI here. No particularly strong feelings though.

app/views/support_interface/application_forms/applicant_details/edit.html.erb
duncanjbrown: Very good idea, will add!

duncanjbrown:
@tijmenb Now with added duplication!

app/views/support_interface/application_forms/applicant_details/edit.html.erb
duncanjbrown: Screenshot 2020-11-18 at 10 45 00

tijmenb (approved):
🚀 🚀 🚀

PR #3446, opened 2020-11-17
Support: Redirect sign in page to candidates page if already signed in

paulrobertlloyd:
## Context

If you are already signed in to the support console, the sign in page should redirect you to the default signed in page for this interface.

Adds a redirect to `support_interface_path` if user visits sign-in page and `current_support_user` is true.

## Guidance to review

* Is this redirect in the right place? Tried a `before_action` like used in other interfaces, but different requirements here. 
* Does this need a test? Can somebody help me write one if it does?

## Link to Trello card

https://trello.com/c/eGZpDgHK/

## Things to check

### DIFF
--- a/app/controllers/support_interface/sessions_controller.rb
+++ b/app/controllers/support_interface/sessions_controller.rb
@@ -3,6 +3,8 @@ module SupportInterface
     skip_before_action :authenticate_support_user!, except: :destroy
 
     def new
+      redirect_to support_interface_path and return if current_support_user
+
       session['post_dfe_sign_in_path'] ||= support_interface_path
       if FeatureFlag.active?('dfe_sign_in_fallback')
         render :authentication_fallback
--- a/spec/system/support_interface/authentication_spec.rb
+++ b/spec/system/support_interface/authentication_spec.rb
@@ -42,9 +42,11 @@ RSpec.describe 'A support user authenticates via DfE Sign-in' do
   end
 
   def and_i_should_not_see_support_menu
-    expect(page).not_to have_link 'Applications'
-    expect(page).not_to have_link 'APITokens'
-    expect(page).not_to have_link 'Vendors'
+    expect(page).not_to have_link 'Candidates'
+    expect(page).not_to have_link 'Providers'
+    expect(page).not_to have_link 'Performance'
+    expect(page).not_to have_link 'Settings'
+    expect(page).not_to have_link 'Documentation'
   end
 
   def when_i_sign_in_via_dfe_sign_in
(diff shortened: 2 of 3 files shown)


stevehook (commented):
Looks right to me. I think it's worth extending the existing authentication system spec to test it, a couple of extra steps here should do it - https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/spec/system/support_interface/authentication_spec.rb#L19

```
  when_i_visit_the_sign_in_page
  then_i_should_be_redirected_to_the_support_interface_users_path
```

spec/system/support_interface/authentication_spec.rb
```diff
@@ -22,6 +22,10 @@
 
     when_i_sign_in_via_dfe_sign_in
     and_i_should_not_have_received_an_email_about_the_new_login
+
+    when_i_visit_the_sign_in_page
+    and_i_should_see_my_email_address
+    then_i_should_be_redirected_to_the_support_interface_applications_path
```
stevehook, line 28: Swapping the above two lines would make the test read a little better.

app/controllers/support_interface/sessions_controller.rb
```diff
@@ -3,6 +3,8 @@ class SessionsController < SupportInterfaceController
     skip_before_action :authenticate_support_user!, except: :destroy
 
     def new
+      redirect_to support_interface_path if current_support_user
```
frankieroberto, line 6: Do we need to add an early return here?

```suggestion
      redirect_to support_interface_path and return if current_support_user
```

spec/system/support_interface/authentication_spec.rb
```diff
@@ -22,6 +22,10 @@
 
     when_i_sign_in_via_dfe_sign_in
     and_i_should_not_have_received_an_email_about_the_new_login
+
+    when_i_visit_the_sign_in_page
+    and_i_should_see_my_email_address
```
frankieroberto, line 27: is this line needed?

stevehook (commented):
Looks good, minor inline comment and agree with @frankieroberto 's suggestion about the return.

spec/system/support_interface/authentication_spec.rb
```diff
@@ -38,9 +42,11 @@ def then_i_should_be_redirected_to_login_page
...
   def and_i_should_not_see_support_menu
-    expect(page).not_to have_link 'Applications'
-    expect(page).not_to have_link 'APITokens'
-    expect(page).not_to have_link 'Vendors'
+    expect(page).not_to have_link 'Candidates'
+    expect(page).not_to have_link 'Providers'
+    expect(page).not_to have_link 'Performance'
+    expect(page).not_to have_link 'Settings'
+    expect(page).not_to have_link 'Documentation'
   end
```
davidgisbey, line 50: More of a stylistic thing, but this is the reason I tend to prefer testing that you're on the correct page i.e.

expect(page).to have_current_path candidate_interface_sign_in_page 

When you only test something is present on the page, you could just be on the wrong page and the test will still pass. 

p.s. I get you're fixing a pre-existing test and it wasn't your implementation

stevehook (approved):
:+1:

PR #3447, opened 2020-11-17
Make HTTP method and path helper usage consistent in the candidate namespace

davidgisbey:
## Context

At the moment our routes are a bit messy for two reasons:

1. We often POST when updating an object. We should use PATCH
2. We often two rails paths helpers for methods with the same raw path (new/create || edit/update) when we can just use one

- Update POSTs to PATCHs where appropriate
- Use one ails paths helper if endpoints share a raw path 

## Guidance to review

This isn't as painful as it looks to review. If you review by commit  I've broken it down by scope.

## Things to check

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -299,17 +299,17 @@ Rails.application.routes.draw do
       end
 
       scope '/other-qualifications' do
-        get '/' => 'other_qualifications/type#new', as: :new_other_qualification_type
-        post '/' => 'other_qualifications/type#create', as: :create_other_qualification_type
+        get '/' => 'other_qualifications/type#new', as: :other_qualification_type
+        post '/' => 'other_qualifications/type#create'
 
         get '/edit-type/:id' => 'other_qualifications/type#edit', as: :edit_other_qualification_type
-        post '/edit-type/:id' => 'other_qualifications/type#update'
+        patch '/edit-type/:id' => 'other_qualifications/type#update'
 
-        get '/new' => 'other_qualifications/details#new', as: :new_other_qualification_details
-        post '/new' => 'other_qualifications/details#create', as: :create_other_qualification_details
+        get '/new' => 'other_qualifications/details#new', as: :other_qualification_details
+        patch '/new' => 'other_qualifications/details#create'
 
         get '/edit-details/:id' => 'other_qualifications/details#edit', as: :edit_other_qualification_details
-        post '/edit-details/:id' => 'other_qualifications/details#update'
+        patch '/edit-details/:id' => 'other_qualifications/details#update'
 
         get '/review' => 'other_qualifications/review#show', as: :review_other_qualifications
         patch '/review' => 'other_qualifications/review#complete', as: :complete_other_qualifications
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -193,10 +193,10 @@ Rails.application.routes.draw do
         post '/' => 'volunteering/experience#submit'
 
         get '/new' => 'volunteering/base#new', as: :new_volunteering_role
-        post '/new' => 'volunteering/base#create', as: :create_volunteering_role
+        post '/new' => 'volunteering/base#create'
 
         get '/edit/:id' => 'volunteering/base#edit', as: :edit_volunteering_role
-        post '/edit/:id' => 'volunteering/base#update'
+        patch '/edit/:id' => 'volunteering/base#update'
 
         get '/review' => 'volunteering/review#show', as: :review_volunteering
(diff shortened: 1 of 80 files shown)


config/routes.rb
```diff
@@ -72,89 +72,90 @@
...
         get '/nationalities' => 'personal_details/nationalities#new', as: :nationalities
         post '/nationalities' => 'personal_details/nationalities#create'
         get '/nationalities/edit' => 'personal_details/nationalities#edit', as: :edit_nationalities
-        post '/nationalities/edit' => 'personal_details/nationalities#update'
+        patch '/nationalities/edit' => 'personal_details/nationalities#update'
         get '/languages' => 'personal_details/languages#edit', as: :languages
         post '/languages' => 'personal_details/languages#update'
         get '/languages/edit' => 'personal_details/languages#edit', as: :edit_languages
-        post '/languages/edit' => 'personal_details/languages#update'
+        patch '/languages/edit' => 'personal_details/languages#update'
```
TobyRet, line 86: Can line 84 be removed? I removed locally ran the specs and they all passed.

config/routes.rb
```diff
@@ -72,89 +72,90 @@
...
+        post '/' => 'personal_details/base#create'
+        get '/edit' => 'personal_details/base#edit', as: :edit_personal_details
+        patch '/edit' => 'personal_details/base#update'
         get '/nationalities' => 'personal_details/nationalities#new', as: :nationalities
         post '/nationalities' => 'personal_details/nationalities#create'
         get '/nationalities/edit' => 'personal_details/nationalities#edit', as: :edit_nationalities
-        post '/nationalities/edit' => 'personal_details/nationalities#update'
+        patch '/nationalities/edit' => 'personal_details/nationalities#update'
         get '/languages' => 'personal_details/languages#edit', as: :languages
         post '/languages' => 'personal_details/languages#update'
```
TobyRet, line 84: Where we are using POST should we be pointing to a #create action?

TobyRet (commented):
Nice clean up 👍 - couple of comments

config/routes.rb
TobyRet, line 84: Line 84 - should this be PATCH as well?

config/routes.rb
```diff
@@ -72,89 +72,90 @@
...
         get '/interview-preferences/review' => 'personal_statement/interview_preferences#show', as: :interview_preferences_show
-        post '/interview-preferences/complete' => 'personal_statement/interview_preferences#complete', as: :interview_preferences_complete
+        patch '/interview-preferences/complete' => 'personal_statement/interview_preferences#complete', as: :interview_preferences_complete
       end
 
       scope '/training-with-a-disability' do
-        get '/' => 'training_with_a_disability#edit', as: :training_with_a_disability_edit
-        post '/review' => 'training_with_a_disability#update', as: :training_with_a_disability_update
+        get '/' => 'training_with_a_disability#edit', as: :edit_training_with_a_disability
+        post '/' => 'training_with_a_disability#update'
```
TobyRet, line 114: Should this be PATCH as its pointing to an #update action?

config/routes.rb
```diff
@@ -167,15 +168,15 @@
         get '/explain-breaks' => 'work_history/breaks#edit', as: :work_history_breaks
         post '/explain-breaks' => 'work_history/breaks#update'
 
-        get '/explain-break/new' => 'work_history/break#new', as: :new_work_history_break
+        get '/explain-break/new' => 'work_history/break#new', as: :work_history_break
```
malcolmbaig, line 171: Might be missing something here, but `new_work_history_break` seems like the appropriate path helper?

malcolmbaig:
Nicely done! :ok_hand:  Just the one question inline.

config/routes.rb
davidgisbey: I should have changed those to new/create. I feel like using new/create edit/update is probably a better pattern to follow. Although the functionality in the actions is near identical so it might be overkill.

config/routes.rb
davidgisbey: Yes, you're right they should all be patches as they are all just attrs on the application form 🤦

config/routes.rb
davidgisbey: Yea you're right

config/routes.rb
davidgisbey: It should and there were some others too that I saw on second glance. Have made the changes 👍

davidgisbey:
I've gone back through and changed the personal statement, contact details and GCSE routes to use PATCH where appropriate.

TobyRet (approved):
LGMT - here's hoping we have good test coverage 🤞

PR #3093, opened 2020-10-08
Send reference request #1

stevehook:
## Context

Candidates will be able to send requests to referees as soon as they have filled in the details for that referee (or later from the review page).

## Changes proposed in this pull request

- Add `RequestController` and corresponding form to create new requests - it needs to send an email, progress the status and redirect to the review page.
- Add `CandidateNameController` and corresponding form to fill in missing names as required - this is triggered by the `RequestController`.
- System spec that covers the journey from the _Are you ready to send a reference request to ...?_ question through to the review page.
- `ReviewController` includes a `new` action that just renders the _Are you ready to send a reference request to ...?_ question to provide an entry point into this part of the flow. This will need to be integrated with the appropriate view/action maybe after https://github.com/DFE-Digital/apply-for-teacher-training/pull/3044 is merged.

To be covered in follow PR(s):
- Integration with the first part of the new reference journey including fixing the back button.
- Implement the links on the review page to send unsent reference requests.
- Automatic reference approval in Sandbox mode for `refbot1/2@example.com` - or is this handled elsewhere?

First screenshot is a temporary page that will need to be integrated to the rest of the journey:

image

image

image

## Guidance to review

- Are the controllers/actions/forms structured logically?
- Are the tests sufficient?
- Are there any issues you can foresee integrating with the rest of the new reference journey?
- To test on a review app or locally I think you will need to create a new reference, then navigate to `candidate/application/references/request/:id` to get started

## Link to Trello card

https://trello.com/c/50ncVpYS/2215-💔-dev-send-reference-request-mvp

## Things to check

- [x] This code does not rely on migrations in the same Pull Request
- [x] If this code includes a migration adding or changing columns, it also backfills existing records for consistency
- [x] API release notes have been updated if necessary
- [x] New environment variables have been [added to the Azure config](https://github.com/DFE-Digital/apply-for-teacher-training#azure-hosting-devops-pipeline)

PR #3453, opened 2020-11-18
Remove autocomplete from IELTS Grade field and update spec

george-bit:
#3093  Context

To comply with accessibility standards autocomplete drop down needs to be removed from the grade field on the IELTS page. 

Field type has been changed and method called to map array of grades has been  removed from form model. However, array still exists as it is used for back end validation. Users now  have free text field which will validate against correct values.

image
image
image
image

## Guidance to review

Check that solution matches expected behaviour.

## Link to Trello card

https://trello.com/c/Q9aOtnL0/2504-remove-autocomplete-for-ielts-band-score-and-use-same-approach-as-that-used-for-gcse-grades

## Things to check

### DIFF
--- a/app/views/candidate_interface/english_foreign_language/ielts/edit.html.erb
+++ b/app/views/candidate_interface/english_foreign_language/ielts/edit.html.erb
@@ -5,7 +5,7 @@
   <div class="govuk-grid-column-two-thirds">
     <%= form_with(model: @ielts_form, url: candidate_interface_edit_ielts_path, method: :patch) do |f| %>
       <%= f.govuk_error_summary %>
-      <h1 class='govuk-heading-xl class='govuk-heading-xl''>
+      <h1 class="govuk-heading-xl">
         <span class="govuk-caption-xl"><%= t('page_titles.efl.start') %></span>
         <%= t('page_titles.efl.ielts') %>
       </h1>
--- a/app/views/candidate_interface/english_foreign_language/ielts/edit.html.erb
+++ b/app/views/candidate_interface/english_foreign_language/ielts/edit.html.erb
@@ -3,12 +3,12 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class='govuk-heading-xl class='govuk-heading-xl''>
-      <span class="govuk-caption-xl"><%= t('page_titles.efl.start') %></span>
-      <%= t('page_titles.efl.ielts') %>
-    </h1>
-
     <%= form_with(model: @ielts_form, url: candidate_interface_edit_ielts_path, method: :patch) do |f| %>
+      <%= f.govuk_error_summary %>
+      <h1 class='govuk-heading-xl class='govuk-heading-xl''>
+        <span class="govuk-caption-xl"><%= t('page_titles.efl.start') %></span>
+        <%= t('page_titles.efl.ielts') %>
+      </h1>
       <%= render 'form_fields', { f: f } %>
     <% end %>
 
--- a/app/views/candidate_interface/english_foreign_language/ielts/new.html.erb
+++ b/app/views/candidate_interface/english_foreign_language/ielts/new.html.erb
@@ -5,7 +5,7 @@
   <div class="govuk-grid-column-two-thirds">
     <%= form_with(model: @ielts_form, url: candidate_interface_ielts_path) do |f| %>
       <%= f.govuk_error_summary %>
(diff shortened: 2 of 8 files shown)


TobyRet (changes requested):
I had a run through and picked up on some issues. Possibly some of this may be out of scope / require a new ticket.

1. Should there be an error summary shown at the top when validation errors occur? I noticed also that error summaries were missing in the prior steps of the flow.

missing_error_summary

2. A score submitted as a single digit is considered invalid. Is a more detailed error message required here?  CC: @frankieroberto @paulrobertlloyd 

single_digit_invalid

3. The year I was awarded the qualification can be in the future 🤯 

the_future

frankieroberto:
@TobyRet @george-bit yes, please add the error summary to any page where errors can appear. It should appear before the `<h1>` header, and the prefix `Error: ` should be added to the page `<title>` tag. It's also worth double checking that clicking the links in the error summary focus the relevant fields.

For the error messages, we should replace "is invalid" with something like "Enter an overall band score in the correct format".

Can we check that trailing and leading whitespace is stripped from all input prior to validation too (it's quite common on mobile for an extra space to get added)?

I also suggest making the width of the **overall band score** field smaller, to match its expected content. See https://design-system.service.gov.uk/components/text-input/#fixed-width-inputs for the relevant classes.

paulrobertlloyd:
> Should there be an error summary shown at the top when validation errors occur? I noticed also that error summaries were missing in the prior steps of the flow.

Yes! And these should appear above the page title.

> A score submitted as a single digit is considered invalid. Is a more detailed error message required here?

Think we can be a bit smarter here, and mutate the value before validating. So `6` becomes `6.0` which is then valid.

> The year I was awarded the qualification can be in the future

I think in this particular case, we shouldn’t allow future years. If you are expecting to complete this assessment in the future, you can provide that information earlier in the flow.

paulrobertlloyd:
> I also suggest making the width of the overall band score field smaller, to match its expected content

@frankieroberto Oh yes, good catch!

paulrobertlloyd:
@george-bit While we’re here, can we ensure that both IELTS ‘Test report form (TRF) number’ and ‘TOEFL registration number’ input fields use `govuk-input--width-20`?

app/views/candidate_interface/english_foreign_language/ielts/edit.html.erb
```diff
@@ -3,12 +3,12 @@
...
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class='govuk-heading-xl class='govuk-heading-xl''>
-      <span class="govuk-caption-xl"><%= t('page_titles.efl.start') %></span>
-      <%= t('page_titles.efl.ielts') %>
-    </h1>
-
     <%= form_with(model: @ielts_form, url: candidate_interface_edit_ielts_path, method: :patch) do |f| %>
+      <%= f.govuk_error_summary %>
+      <h1 class='govuk-heading-xl class='govuk-heading-xl''>
```
frankieroberto, line 8: Looks like this was an existing bug, but let's fix this too.

```suggestion
      <h1 class="govuk-heading-xl">
```

app/views/candidate_interface/english_foreign_language/ielts/new.html.erb
```diff
@@ -3,12 +3,12 @@
...
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class='govuk-heading-xl class='govuk-heading-xl''>
-      <span class="govuk-caption-xl"><%= t('page_titles.efl.start') %></span>
-      <%= t('page_titles.efl.ielts') %>
-    </h1>
-
     <%= form_with(model: @ielts_form, url: candidate_interface_ielts_path) do |f| %>
+      <%= f.govuk_error_summary %>
+      <h1 class='govuk-heading-xl class='govuk-heading-xl''>
```
frankieroberto, line 8:
```suggestion
      <h1 class="govuk-heading-xl">
```

frankieroberto (approved):
Looks good, thanks. 👍  Could you accept my minor HTML fixes, and then I think we can merge?

app/views/candidate_interface/english_foreign_language/ielts/new.html.erb
george-bit: For sure @frankieroberto, I think this is just legacy code so no problem changing!

TobyRet (approved):

davidgisbey (approved):
Looks good George!

PR #3459, opened 2020-11-18
Add UCAS holidays to calendar

benswannack:
## Context
UCAS have made us aware of their breaks in winter 2020 and spring 2021. Add these to the calendar to ensure RBD dates in apply are fair. Also fix RBD calculation tests

## Guidance to review
The ticket also mentions: 03 May 2021, 31 May 2021, 30 August 2021 (bank holidays) which I have checked are already in the calendar

## Link to Trello card
https://trello.com/c/kBLB4ot9/3043-fix-rbds-over-christmas-and-new-year

## Things to check

### DIFF
--- a/app/lib/time_limit_config.rb
+++ b/app/lib/time_limit_config.rb
@@ -33,24 +33,26 @@ class TimeLimitConfig
     28
   end
 
-  RULES = {
-    reject_by_default: [
-      Rule.new(nil, nil, 40),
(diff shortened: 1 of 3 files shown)


duncanjbrown (approved):
I guess at some point these things are far enough in the past to remove them altogether. I'm a bit superstitious about doing that now while we still have applications from last year in the system, and it seems wrong to remove facts that affected how they behaved. We should eventually cull them though — imagine seeing sth like `UCAS_WINTER_HOLIDAY_2010`

PR #3460, opened 2020-11-18
Consolidate icon components

paulrobertlloyd:
## Context

We had one icon component. Now we have two. A third may be on the way. But really these are just variations of a single component. Let’s consolidate things before this gets out of hand! 🤹‍♂️

* Removes `CheckIconComponent`
* Removes `CrossIcon` component
* Adds `IconComponent` which takes a `type` value
* Updates templates to use new `IconComponent`

## Guidance to review

* Is this a good idea?

## Things to check


steventux (approved):
Nice, didn't realise the `svg` element was identical for both.

PR #3463, opened 2020-11-19
Send initial emails to resolve dual applications

despo:
## Context

Replace marking initial emails to sent with sending initial emails to the candidate and provider in case of duplicate matches between UCAS and apply

#### Duplicate applicant 
Provider email
ucas_match--duplicate--provider-email

Candidate email  ❗ updated

updated-candidate-duplication-application-initial-email

### Multiple acceptances email (candidate)

ucas-multiple-acceptances-match

## Link to Trello card

https://trello.com/c/SL6IT49j/3018-send-initial-emails-to-resolve-dual-applications

## Things to check

### DIFF
--- a/app/lib/time_limit_config.rb
+++ b/app/lib/time_limit_config.rb
@@ -49,6 +49,15 @@ class TimeLimitConfig
       chase_candidate_before_dbd: [
         Rule.new(nil, nil, 5),
       ],
+      ucas_match_candidate_withdrawal_request: [
+        Rule.new(nil, nil, 10),
+      ],
+      ucas_match_candidate_withdrawal_request_reminder: [
+        Rule.new(nil, nil, 5),
+      ],
+      ucas_match_ucas_withdrawal_request: [
+        Rule.new(nil, nil, 10),
+      ],
     }
   end
 
--- a/app/lib/time_limit_config.rb
+++ b/app/lib/time_limit_config.rb
@@ -56,7 +56,7 @@ class TimeLimitConfig
         Rule.new(nil, nil, 5),
       ],
       ucas_match_ucas_withdrawal_request: [
-        Rule.new(nil, nil, 10),
+        Rule.new(nil, nil, 5),
       ],
     }
   end
--- a/app/mailers/candidate_mailer.rb
+++ b/app/mailers/candidate_mailer.rb
@@ -299,7 +299,7 @@ class CandidateMailer < ApplicationMailer
     @application_form = application_choice.application_form
     @course_name_and_code = application_choice.course_option.course.name_and_code
     @provider_name = application_choice.course_option.course.provider.name
-    @date_to_withdraw_application_by = 10.business_days.after(Time.zone.today).to_s(:govuk_date)
+    @date_to_withdraw_application_by = TimeLimitCalculator.new(rule: :ucas_match_candidate_withdrawal_request, effective_date: Time.zone.today).call.fetch(:time_in_future).to_s(:govuk_date)
 
     email_for_candidate(
       @application_form,
--- a/app/mailers/candidate_mailer.rb
+++ b/app/mailers/candidate_mailer.rb
@@ -307,6 +307,16 @@ class CandidateMailer < ApplicationMailer
     )
   end
 
+  def ucas_match_initial_email_multiple_acceptances(candidate)
+    @application_form = candidate.application_forms.first
+    @date_to_withdraw_application_by = TimeLimitCalculator.new(rule: :ucas_match_candidate_withdrawal_request, effective_date: Time.zone.today).call.fetch(:time_in_future).to_s(:govuk_date)
+
+    email_for_candidate(
+      @application_form,
+      subject: I18n.t!('candidate_mailer.ucas_match_initial_email.multiple_acceptances.subject'),
+    )
+  end
+
 private
 
   def new_offer(application_choice, template_name)
--- a/app/mailers/candidate_mailer.rb
+++ b/app/mailers/candidate_mailer.rb
@@ -295,6 +295,18 @@ class CandidateMailer < ApplicationMailer
     )
   end
 
+  def ucas_match_initial_email_duplicate_applications(application_choice)
+    @application_form = application_choice.application_form
+    @course_name_and_code = application_choice.course_option.course.name_and_code
+    @provider_name = application_choice.course_option.course.provider.name
+    @date_to_withdraw_application_by = 10.business_days.after(Time.zone.today).to_s(:govuk_date)
+
+    email_for_candidate(
+      @application_form,
+      subject: I18n.t!('candidate_mailer.ucas_match_initial_email.duplicate_applications.subject'),
+    )
+  end
+
 private
 
   def new_offer(application_choice, template_name)
--- a/app/models/ucas_match.rb
+++ b/app/models/ucas_match.rb
@@ -38,6 +38,11 @@ class UCASMatch < ApplicationRecord
       application_accepted_on_ucas_and_accepted_on_apply?
   end
 
+  def application_accepted_on_ucas_and_accepted_on_apply?
+    ucas_matched_applications.map(&:application_accepted_on_ucas?).any? &&
+      ucas_matched_applications.map(&:application_accepted_on_apply?).any?
+  end
+
   def invalid_matching_data?
     !ucas_matched_applications.all?(&:valid_matching_data?)
   end
@@ -93,10 +98,5 @@ private
 
     application_for_the_same_course_on_both_services.map(&:application_in_progress_on_ucas?).any? &&
       application_for_the_same_course_on_both_services.map(&:application_in_progress_on_apply?).any?
-  end
-
-  def application_accepted_on_ucas_and_accepted_on_apply?
-    ucas_matched_applications.map(&:application_accepted_on_ucas?).any? &&
-      ucas_matched_applications.map(&:application_accepted_on_apply?).any?
   end
 end
--- a/app/models/ucas_match.rb
+++ b/app/models/ucas_match.rb
@@ -82,6 +82,10 @@ class UCASMatch < ApplicationRecord
     action_taken.to_sym
   end
 
+  def application_choices_for_same_course_on_both_services
+    ucas_matched_applications.select(&:both_scheme?).map(&:application_choice)
+  end
+
 private
 
   def application_for_the_same_course_in_progress_on_both_services?
--- a/app/models/ucas_match.rb
+++ b/app/models/ucas_match.rb
@@ -56,15 +56,15 @@ class UCASMatch < ApplicationRecord
   def need_to_send_reminder_emails?
     return false unless initial_emails_sent?
 
-    send_reminder_emails_date = 5.business_days.after(candidate_last_contacted_at).to_date
-    Time.zone.today >= send_reminder_emails_date
+    candidate_withdrawal_request_reminder_date = calculate_action_date(:ucas_match_candidate_withdrawal_request_reminder, candidate_last_contacted_at)
+    Time.zone.today >= candidate_withdrawal_request_reminder_date
   end
 
   def need_to_request_withdrawal_from_ucas?
     return false unless reminder_emails_sent?
 
-    request_withdrawal_from_ucas_date = 10.business_days.after(candidate_last_contacted_at).to_date
-    Time.zone.today >= request_withdrawal_from_ucas_date
+    ucas_withdrawal_request_date = calculate_action_date(:ucas_match_ucas_withdrawal_request, candidate_last_contacted_at)
+    Time.zone.today >= ucas_withdrawal_request_date
   end
 
   def next_action
@@ -98,5 +98,9 @@ private
 
     application_for_the_same_course_on_both_services.map(&:application_in_progress_on_ucas?).any? &&
       application_for_the_same_course_on_both_services.map(&:application_in_progress_on_apply?).any?
+  end
+
+  def calculate_action_date(action, effective_date)
+    TimeLimitCalculator.new(rule: action, effective_date: effective_date).call.fetch(:time_in_future).to_date
   end
 end
--- a/app/services/support_interface/send_ucas_match_initial_emails.rb
+++ b/app/services/support_interface/send_ucas_match_initial_emails.rb
@@ -1,0 +1,15 @@
+module SupportInterface
+  class SendUCASMatchInitialEmails
+    attr_reader :ucas_match
+
+    def initialize(ucas_match)
+      @ucas_match = ucas_match
+    end
+
+    def call
+      return SupportInterface::SendUCASMatchInitialEmailsMultipleAcceptances.new(ucas_match).call if ucas_match.application_accepted_on_ucas_and_accepted_on_apply?
+
+      SupportInterface::SendUCASMatchInitialEmailsDuplicateApplications.new(ucas_match).call if ucas_match.dual_application_or_dual_acceptance?
+    end
+  end
+end
--- a/app/services/support_interface/send_ucas_match_initial_emails_duplicate_applications.rb
+++ b/app/services/support_interface/send_ucas_match_initial_emails_duplicate_applications.rb
@@ -1,0 +1,21 @@
+module SupportInterface
+  class SendUCASMatchInitialEmailsDuplicateApplications
+    attr_reader :ucas_match
+
+    def initialize(ucas_match)
+      @ucas_match = ucas_match
+    end
+
+    def call
+      raise 'UCAS Match initial emails already sent' if ucas_match.initial_emails_sent?
+
+      application_choice = ucas_match.application_choices_for_same_course_on_both_services.first
+
+      CandidateMailer.ucas_match_initial_email_duplicate_applications(application_choice).deliver_later
+
+      NotificationsList.for(application_choice).each do |provider_user|
+        ProviderMailer.ucas_match_initial_email_duplicate_applications(provider_user, application_choice).deliver_later
+      end
+    end
+  end
+end
--- a/app/services/support_interface/send_ucas_match_initial_emails_multiple_acceptances.rb
+++ b/app/services/support_interface/send_ucas_match_initial_emails_multiple_acceptances.rb
@@ -1,0 +1,15 @@
+module SupportInterface
+  class SendUCASMatchInitialEmailsMultipleAcceptances
+    attr_reader :ucas_match
+
+    def initialize(ucas_match)
+      @ucas_match = ucas_match
+    end
+
+    def call
+      raise 'UCAS Match initial emails already sent' if ucas_match.initial_emails_sent?
+
+      CandidateMailer.ucas_match_initial_email_multiple_acceptances(ucas_match.candidate).deliver_later
+    end
+  end
+end
--- a/app/views/candidate_mailer/ucas_match_initial_email_duplicate_applications.text.erb
+++ b/app/views/candidate_mailer/ucas_match_initial_email_duplicate_applications.text.erb
@@ -1,0 +1,22 @@
+Dear <%= @application_form.full_name %>
+
+It looks like you submitted an application for <%= @course_name_and_code %> at <%= @provider_name %> on both DfE Apply for teacher training and UCAS Teacher Training Apply.
+
+This has been flagged to us, as well as to <%= @provider_name %>, as applying for the same course on both services is not allowed, as outlined by the terms of use.
+
+# Action required – withdraw your duplicate application
+
+To ensure your application progresses effectively, you need to withdraw the course choice from either DfE Apply for teacher training or UCAS Teacher Training by <%= @date_to_withdraw_application_by %>.
+
+It’s up to you which system you choose to progress the choice on. If you have additional UCAS Teacher Training choices, you may want to keep the choice on UCAS Teacher Training, so all your choices are in one place.
+
+# What happens if you don’t do anything?
+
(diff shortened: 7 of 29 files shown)


app/mailers/candidate_mailer.rb
```diff
@@ -295,6 +295,30 @@ def offer_accepted(application_choice)
     )
   end
 
+  def ucas_match_initial_email_duplicate_applications(application_choice)
+    @application_form = application_choice.application_form
+    @course_name_and_code = application_choice.course_option.course.name_and_code
+    @provider_name = application_choice.course_option.course.provider.name
+    @date_to_withdraw_application_by = 10.business_days.after(Time.zone.today)
```
AgaDufrat, line 302:
```suggestion
    @date_to_withdraw_application_by = 10.business_days.after(Time.zone.today).to_s(:govuk_date)
```

app/views/candidate_mailer/ucas_match_initial_email_duplicate_applications.text.erb
```diff
@@ -0,0 +1,22 @@
...
+
+Action required – withdraw your duplicate application
+
+To ensure your application progresses effectively, you need to withdraw the course choice from either DfE Apply for teacher training or UCAS Teacher Training by <%= @date_to_withdraw_application_by %>.
+
+It’s up to you which system you choose to progress the choice on. If you have additional UCAS Teacher Training choices, you may want to keep the choice on UCAS Teacher Training, so all your choices are in one place.
+
+What happens if you don’t do anything?
+
+If you don’t withdraw your duplicate choice by <%= @date_to_withdraw_application_by %> (in 10 working days), we will automatically withdraw the choice removed from your UCAS Teacher Training application – meaning you’ll use DfE Apply for teacher training to progress this choice.
```
AgaDufrat, line 15: `we will automatically withdraw the choice removed from your UCAS Teacher Training application`
We may need some content work on this 😬

despo: @AgaDufrat 🤣

despo:
@duncanjbrown @mnacos can you review please with focus on the services relating to sending out the relevant emails.

app/views/candidate_mailer/ucas_match_initial_email_duplicate_applications.text.erb
```diff
@@ -0,0 +1,22 @@
...
+
+Action required – withdraw your duplicate application
+
+To ensure your application progresses effectively, you need to withdraw the course choice from either DfE Apply for teacher training or UCAS Teacher Training by <%= @date_to_withdraw_application_by %>.
+
+It’s up to you which system you choose to progress the choice on. If you have additional UCAS Teacher Training choices, you may want to keep the choice on UCAS Teacher Training, so all your choices are in one place.
+
+What happens if you don’t do anything?
+
+If you don’t withdraw your duplicate choice by <%= @date_to_withdraw_application_by %> (in 10 working days), we will withdraw your UCAS Teacher Training application – meaning you’ll use DfE Apply for teacher training to progress this choice.
```
despo, line 15: update to:

> ...we will withdraw the choice from your UCAS Teacher...

app/mailers/candidate_mailer.rb
```diff
@@ -295,6 +295,18 @@ def offer_accepted(application_choice)
     )
   end
 
+  def ucas_match_initial_email_duplicate_applications(application_choice)
+    @application_form = application_choice.application_form
+    @course_name_and_code = application_choice.course_option.course.name_and_code
+    @provider_name = application_choice.course_option.course.provider.name
+    @date_to_withdraw_application_by = 10.business_days.after(Time.zone.today).to_s(:govuk_date)
```
duncanjbrown, line 302: It would be good to put this rule somewhere closer to the data — is there a way we could push it up into the model? How do we track the date these initial emails were sent?

app/views/candidate_mailer/ucas_match_initial_email_duplicate_applications.text.erb
```diff
@@ -0,0 +1,22 @@
+Dear <%= @application_form.full_name %>
+
+It looks like you submitted an application for <%= @course_name_and_code %> at <%= @provider_name %> on both DfE Apply for teacher training and UCAS Teacher Training Apply.
+
+This has been flagged to us, as well as to <%= @provider_name %>, as applying for the same course on both services is not allowed, as outlined by the terms of use.
+
+Action required – withdraw your duplicate application
```
duncanjbrown, line 7: Does this need to be bold?

app/views/candidate_mailer/ucas_match_initial_email_duplicate_applications.text.erb
```diff
@@ -0,0 +1,22 @@
...
+
+To ensure your application progresses effectively, you need to withdraw the course choice from either DfE Apply for teacher training or UCAS Teacher Training by <%= @date_to_withdraw_application_by %>.
+
+It’s up to you which system you choose to progress the choice on. If you have additional UCAS Teacher Training choices, you may want to keep the choice on UCAS Teacher Training, so all your choices are in one place.
+
+What happens if you don’t do anything?
```
duncanjbrown, line 13: As above

config/locales/candidate_mailer.yml
```diff
@@ -81,3 +81,6 @@ en:
         course_full: "%{course_name} at %{provider_name} became full while %{candidate_name} was awaiting references"
         location_full: "%{course_name} at %{provider_name} became full at %{location} while %{candidate_name} was awaiting references"
         study_mode_full: "%{study_mode} places for %{course_name} at %{provider_name} became full while %{candidate_name} was awaiting references"
+    ucas_match_initial_email:
+      duplicate_applications:
+        subject: 'Action required: It looks like you submitted a duplicate application'
```
duncanjbrown, line 86: Should this `It` be capitalized?

spec/mailers/candidate_mailer_ucas_match_initial_email_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    it 'sends an email containing the provider name in the body' do
+      provider_name = application_choice.course.provider.name
+
+      expect(email.body).to include(provider_name)
+    end
+
+    it 'sends an email containing the date that the application needs to be withdrawn by' do
+      withdraw_application_date = 10.business_days.after(Time.zone.today).to_s(:govuk_date)
```
duncanjbrown, line 41: It would be good not to repeat the calculation out here — if we could ask `ucas_match` for the date that might be a bit tidier?

duncanjbrown (commented):
Looking good! Couple of minor copy things and a question about where the 10-days-to-respond rule lives

app/views/candidate_mailer/ucas_match_initial_email_duplicate_applications.text.erb
despo: All existing emails are in plain text format, so I followed the same pattern. There is also no formatting included in the email templates.

config/locales/candidate_mailer.yml
despo: Yes. All subject lines mentioned [in the email document](https://docs.google.com/document/d/1s5ql4jNUUr3QDPUQYWImkZR6o8upvutrjOEuoZ0qqTE/edit) follow this pattern.

app/mailers/candidate_mailer.rb
despo: We update `candidate_last contacted_at` and `action taken`. I'm not sure if this can be automated by using these fields, as `candidate_last_contacted_at` can potentially be mapped to other type of contacts as well, and I assume that the support agent verifies when the last set of emails were sent through the audit log before sending out the next round of emails. 

I can however add another field to map this. @AgaDufrat can you confirm if this is the case, and do you have any thoughts on maybe added another field?

app/services/support_interface/send_initial_emails.rb
```diff
@@ -0,0 +1,30 @@
...
+    attr_accessor :ucas_match
+
+    def initialize(ucas_match:)
+      self.ucas_match = ucas_match
+    end
+
+    def call
+      return false if ucas_match.initial_emails_sent?
+
+      return ::SendInitialEmailsMultipleAcceptances.new(ucas_match).call if ucas_match.application_accepted_on_ucas_and_accepted_on_apply?
```
mnacos, line 12: `SendInitialEmailsMultipleAcceptances` is not particularly complex, I'd say it's on a par with the `CandidateMailer` and `ProviderMailer` code below. Is there a reason we have `SendInitialEmailsMultipleAcceptances` as a service, while the code below isn't one?

app/services/support_interface/send_initial_emails.rb
```diff
@@ -0,0 +1,30 @@
...
+
+      application_choice = application_choice_for(ucas_match)
+
+      CandidateMailer.ucas_match_initial_email_duplicate_applications(application_choice).deliver_later
```
mnacos, line 16: Should this and the following few lines be turned into a service with a descriptive name and a spec?

mnacos (commented):
Looks good overall. I haven't followed this story very closely so cannot comment on the content, but I'm finding it a little hard to understand all the different states that `SendInitialEmails` caters for. Some renaming or comments could potentially help, a UCAS Matches for Dummies, if you will :)

I do wonder whether the entire UCAS matches code could benefit from some sort of state machine implementation.

app/services/support_interface/send_initial_emails.rb
despo: I started with `SendInitialEmails` but when I added the multiple acceptances part it became a bit too noisy and I wanted to seperate the code, but maintain the abstraction as the controller shouldn't care what type of initial email it is. But I agree, moving the duplicate acceptances code to it's own service makes sense and probably makes things a bit easier to understand.

app/views/candidate_mailer/ucas_match_initial_email_duplicate_applications.text.erb
duncanjbrown: There is: see https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/views/candidate_mailer/deferred_offer.text.erb which renders as https://qa.apply-for-teacher-training.service.gov.uk/rails/mailers/candidate_mailer/deferred_offer

config/locales/candidate_mailer.yml
duncanjbrown: Ah. Feels a bit odd to me — getting content design advice!

duncanjbrown: Lower case is correct: https://ukgovernmentdfe.slack.com/archives/CPH8J9G65/p1606217719066500?thread_ts=1606212747.059500&cid=CPH8J9G65

despo: @duncanjbrown updated.

spec/mailers/candidate_mailer_ucas_match_initial_email_spec.rb
despo: after discussing this with @AgaDufrat `TimeLimitConfig` looked like a more appropriate place so I moved all rules there and used `TimeLimitCalculator` to retrieve the calculated dates. The relevant changes are in my last commit:  https://github.com/DFE-Digital/apply-for-teacher-training/pull/3463/commits/00b7a0db5e2e5bf35fe85e3f17843bdc0c5ec624

app/services/support_interface/send_initial_emails.rb
despo: Done.

mnacos (approved):
:+1:  the extra service and `TimeLimitConfig`. Once all the content and formatting is finalised this looks good to go.

app/services/support_interface/send_initial_emails.rb
```diff
@@ -0,0 +1,15 @@
+module SupportInterface
+  class SendInitialEmails
```
duncanjbrown, line 2: Please could this class have a more descriptive name mentioning UCAS?

app/services/support_interface/send_initial_emails.rb
```diff
@@ -0,0 +1,15 @@
...
+    attr_accessor :ucas_match
+
+    def initialize(ucas_match:)
+      self.ucas_match = ucas_match
+    end
+
+    def call
+      return SupportInterface::SendInitialEmailsMultipleAcceptances.new(ucas_match: ucas_match).call if ucas_match.application_accepted_on_ucas_and_accepted_on_apply?
```
duncanjbrown, line 10: Likewise here and below!

app/models/ucas_match.rb
```diff
@@ -51,15 +56,15 @@ def ucas_matched_applications
   def need_to_send_reminder_emails?
     return false unless initial_emails_sent?
 
-    send_reminder_emails_date = 5.business_days.after(candidate_last_contacted_at).to_date
-    Time.zone.today >= send_reminder_emails_date
+    candidate_withdrawal_requst_reminder_date = calculate_action_date(:ucas_match_candidate_withdrawal_request_reminder, candidate_last_contacted_at)
```
duncanjbrown, line 59: Typo `requst`

app/mailers/candidate_mailer.rb
```diff
@@ -295,6 +295,30 @@ def offer_accepted(application_choice)
     )
   end
 
+  def ucas_match_initial_email_duplicate_applications(application_choice)
```
duncanjbrown, line 298: Would it make more sense for this method to take the `UCASMatch` itself? Then no need to depend on `TimeLimitCalculator` out here too — could just ask the match to figure it out

despo: I did initially pass in UCASMatch but then decided against it so it would follow the same pattern as the rests of the CandidateMailer methods. I would just argue that this seems to be the purpose of the TimeLimitCalculator anyway, and it is also reused in the CandidateMailer so it makes sense to keep things consistent.

We also don't use any other information found on the UCASMatch itself.

app/services/support_interface/send_initial_emails.rb
despo: Renamed to `SendUCASMatchInitialEmails..` to keep it consistent with the email names e.g. `ucas_match_initial_email`, `ucas_match_initial_email_duplicate_applications` etc

app/models/ucas_match.rb
despo: Resolved.

app/services/support_interface/send_ucas_match_initial_emails.rb
```diff
@@ -0,0 +1,15 @@
+module SupportInterface
+  class SendUCASMatchInitialEmails
+    attr_accessor :ucas_match
+
+    def initialize(ucas_match:)
+      self.ucas_match = ucas_match
+    end
+
+    def call
+      return SupportInterface::SendUCASMatchInitialEmailsMultipleAcceptances.new(ucas_match: ucas_match).call if ucas_match.application_accepted_on_ucas_and_accepted_on_apply?
```
duncanjbrown, line 10: Nit: this is a bit more difficult to read with the `if` at the end of a long line. Consider bringing the `if` up and turning the second into an `elsif`?

app/services/support_interface/send_ucas_match_initial_emails_duplicate_applications.rb
```diff
@@ -0,0 +1,28 @@
+module SupportInterface
+  class SendUCASMatchInitialEmailsDuplicateApplications
+    attr_accessor :ucas_match
+
+    def initialize(ucas_match:)
+      self.ucas_match = ucas_match
+    end
+
+    def call
+      return false if ucas_match.initial_emails_sent?
```
duncanjbrown, line 10: I'd go further and `raise` here — this should never happen?

app/services/support_interface/send_ucas_match_initial_emails_duplicate_applications.rb
```diff
@@ -0,0 +1,28 @@
...
+      CandidateMailer.ucas_match_initial_email_duplicate_applications(application_choice).deliver_later
+
+      NotificationsList.for(application_choice).each do |provider_user|
+        ProviderMailer.ucas_match_initial_email_duplicate_applications(provider_user, application_choice).deliver_later
+      end
+    end
+
+  private
+
+    def application_choice_for(ucas_match)
```
duncanjbrown, line 23: Will this always return the correct application choice? We're sending the specific course details to both provider and candidate — couldn't it just return some other, non-conflicting choice off the form?

duncanjbrown (commented):
Question — I think blocking, sorry! — about which application choice we talk about in the emails

app/lib/time_limit_config.rb
```diff
@@ -49,6 +49,15 @@ def self.rules
       chase_candidate_before_dbd: [
         Rule.new(nil, nil, 5),
       ],
+      ucas_match_candidate_withdrawal_request: [
+        Rule.new(nil, nil, 10),
+      ],
+      ucas_match_candidate_withdrawal_request_reminder: [
+        Rule.new(nil, nil, 5),
+      ],
+      ucas_match_ucas_withdrawal_request: [
```
AgaDufrat, line 58: When I was preparing the show and tell presentation realised I got the rules of the timings wrong 🤦‍♀️ 

It should be UCAS withdrawal request is sent 10 days after the initial email / 5 days after the reminder email / 5 days after candidate_last_contacted_at. 
We are currently doing 15 days after the initial email / 10 days after the initial email

I think this commit should do it https://github.com/DFE-Digital/apply-for-teacher-training/commit/b68fe077bbddf01bf60020960a2de3247019eea1
Feel free to cherry-cherry pick or I’m happy to raise a PR once yours is merged.

app/services/support_interface/send_ucas_match_initial_emails_duplicate_applications.rb
despo: Afetr discussing this with Aga, a candidate can currently have up to 2 UCAS matches - one for each year. So as long as the current recruitment year applies to the UCAS Match we should be able to retrieve the correct application.

So my approach is incorrect as I followed the pattern I saw in the candidate_mailer for retrieve the application choice when the `application_choice` should be exposed on `UCASMatchedApplication` and retrieved from there, rather than from the candidate (https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/models/ucas_matched_application.rb#L106)

cc @AgaDufrat

app/views/candidate_mailer/ucas_match_initial_email_duplicate_applications.text.erb
despo: I was looking for html emails 🙄

app/services/support_interface/send_ucas_match_initial_emails_duplicate_applications.rb
AgaDufrat: I think something like this
`ucas_match.ucas_matched_applications.select(&:both_scheme?)`
will get us the UCASMatchedApplication objects we are after.
Then we make this method public
https://github.com/DFE-Digital/apply-for-teacher-training/blob/aee8187135fe126fb13bca580606cfa21a4a9460/app/models/ucas_matched_application.rb#L106-L112

This should get us ApplicationChoices in the correct cycle already.

Also there is potentially a scenario when a candidate applied for more than one course on both services.

app/mailers/candidate_mailer.rb
```diff
@@ -295,6 +295,30 @@ def offer_accepted(application_choice)
...
+    @application_form = application_choice.application_form
+    @course_name_and_code = application_choice.course_option.course.name_and_code
+    @provider_name = application_choice.course_option.course.provider.name
```
AgaDufrat, line 301: For this email we only need candidate's name and date_to_withdraw_application_by

I think we could pass ucas_match or even candidate in the service:
`CandidateMailer.ucas_match_initial_email_duplicate_applications(ucas_match).deliver_later`

AgaDufrat (approved):
LGTM 💯 
Thank you for doing all the groundwork for sending the emails 🎉 🙇‍♀️

app/services/support_interface/send_ucas_match_initial_emails_multiple_acceptances.rb
```diff
@@ -0,0 +1,17 @@
...
+    attr_reader :ucas_match
+
+    def initialize(ucas_match)
+      @ucas_match = ucas_match
+    end
+
+    def call
+      raise 'UCAS Match initial emails already sent' if ucas_match.initial_emails_sent?
+
+      application_choice = ucas_match.application_choices.first
```
duncanjbrown, line 12: I don't think this will necessarily return the correct choice?

AgaDufrat, line 12: I think this may be a problem actually. If candidate applied on both services but not for the same course `ucas_match.application_choices` will be nil.

For the multiple acceptances email we don't really care about course/application choice details. We don't include this info in the email. If we want to pass `application_choice` for consistency maybe we could `ucas_match.candidate.application_form.application_choices.first`. 
Although, I would be tempted to just pass `application_form` (we do it in the mailer already). I'm worried that someone (myself in the future 😂 ) can incorrectly assume that we are passing the application from which has an acceptance.

despo: removed and using `ucas_match.candidate` instead as the email is not related to an application choice at all and all we need to know is the candidate's details (however I did retrieve `candidate.application_form.first` in the mailer so I could reuse `email_for_candidate`

duncanjbrown (approved):
🚀 Thanks @despo!

PR #3464, opened 2020-11-19
Fix test related to provider access controls

malcolmbaig:
Use match_array so that the expectation doesn't care about the order of
returned elements.

## Context

Flaky test.

See diff.

## Guidance to review
Is the new expectation ok? Do we care about the order of the array?

## Things to check


benswannack (approved):
Thank you!

PR #3467, opened 2020-11-19
Don’t capitalise 'candidate' on Sandbox start page

duncanjbrown:
## Context

This word isn't capitalised anywhere else


paulrobertlloyd (approved):
Good catch!

PR #3468, opened 2020-11-19
Include section name on application feedback form

paulrobertlloyd:
## Context

Reviewing design given decision to not use dynamic in-page feedback form, but instead open feedback form in a new tab. This means some context needs to be provided.

* Pass a `section` parameter over to the form, whose value is derived from the path
* Update page heading for form
* Tweak label for first structured response option
* Consolidate localisation strings for this feature

Screenshot

## Guidance to review

* Is this the right way to determine the section being referred to? The section a question appears in is not something we can infer directly from a path, so have to use a little manual intervention.
* I needed to add an exception for Rubocop to allow me to use an empty case statement – are we happy with this exception. Do we want to enable it more broadly (in a separate PR)?

## Link to Trello card

https://trello.com/c/p9sSZNfy/

## Things to check


gregknight1 (commented):
All good from a content point of view.

.rubocop.yml
```diff
@@ -122,6 +122,11 @@ Naming/MemoizedInstanceVariableName:
 Style/EachWithObject:
   Enabled: false
 
+# Using case with empty condition is less verbose that multiple if statements
```
davidgisbey, line 125: than

davidgisbey (approved):
LGTM!

tijmenb (commented):
Don't merge this just yet - I think we should approach this differently. I'll post something in a bit.

tijmenb:
There are 2 issues I want to avoid: 

- We shouldn't rely on parsing/understanding the URL. We don't do that anywhere else because we use Rails' path helpers, which makes it really easy to change the URLs. For example, this code would break if we were to say, remove `/candidate` from the URL (a prefix which makes a certain designer "quite ill", I've heard!)
- We should avoid displaying user-generated strings like this (sign in via https://apply-for-te-2560-revie-w6w36g.herokuapp.com/support/applications/12 and [click on this feedback link](https://apply-for-te-2560-revie-w6w36g.herokuapp.com/candidate/application/application-feedback?page_title=Degree&path=%2Fcandidate%2Fapplication%2Fdegrees%2Freview&section=way%20in%20which%20we%20do%20the%20BAT%20quizzes,%20for%20example%20by%20having%20a%20completely%20Paul%20Lloyd-free)).

Screenshot 2020-11-19 at 14 14 31

To avoid these issues we could pass in the `params[:action]` and `params[:controller]` in the URL, and then calculate the section in the feedback controller, so it can't be spoofed. Thoughts?

paulrobertlloyd:
> this code would break if we were to say, remove `/candidate` from the URL (a prefix which makes a certain designer "quite ill", I've heard!)

Hoisted by my own petard!! Well played @tijmenb, well played! 😛

paulrobertlloyd:
These all sound like valid concerns @tijmenb, though probably outside my wheelhouse. Given we now have the intended design, is this something you can help me with @davidgisbey?

tijmenb:
Fixed in https://github.com/DFE-Digital/apply-for-teacher-training/pull/3468/commits/3da77da2102b4ece7d15e50e3cb36938e4bbe1f5:

> This doesn't rely on the URLs, which can change. It also doesn't let users change the page title by fiddling with the URL - if this happens or if we don't know the section, we default to "this section".

> This also fixes a bug where the section would disappear when there are validation errors. And simplifies the passing through of `path` and `page_title`.

paulrobertlloyd:
Thanks @tijmenb!

PR #3469, opened 2020-11-19
Add guidance around additional reference requests being cancelled 

davidgisbey:
## Context

At the moment, when a candidate can request as many references as they want. However, when they receive 2 sets of feedback. We cancel all the other references with outstanding feedback(in both the awaiting reference and not requested yet states). 

We don't actually tell the candidate that this is going to happen.

This PR adds in this content (can be found in the Trello ticket)

- Before

image

- If **no** references are in the not_requested_yet, feedback_requested or feedback_provided state show

image

-  If **one** references are in the not_requested_yet, feedback_requested or feedback_provided state show

image

- If **more than one** 

image

- Does not render if 2 have been received 

## Guidance to review

Is the logic correct around what we consider a viable reference?

I understand that a cancelled reference can be re-requested so in that sense it could technically potentially be viable.

## Link to Trello card

https://trello.com/c/FOoQq16W/2532-%F0%9F%92%942%EF%B8%8F%E2%83%A3-dev-guidance-around-additional-reference-requests-being-cancelled-when-2-have-been-received

## Things to check

### DIFF
--- a/app/components/candidate_interface/add_reference_component.html.erb
+++ b/app/components/candidate_interface/add_reference_component.html.erb
@@ -1,6 +1,22 @@
 <% unless application_form.enough_references_have_been_provided? %>
   <div class="govuk-inset-text govuk-!-margin-top-0">
-    <% if enough_references_have_been_added? %>
+    <% if no_viable_references? %>
+
+      <h2 class="govuk-heading-m">Add a referee</h2>
+
+      <p class="govuk-body">You need 2 references before you can submit your application.</p>
+
+      <%= govuk_button_link_to 'Add a referee', candidate_interface_references_start_path %>
+
+    <% elsif one_viable_reference? %>
+
+      <h2 class="govuk-heading-m">Add a second referee</h2>
+
+      <p class="govuk-body">You need 2 references before you can submit your application.</p>
+
+      <%= govuk_button_link_to 'Add a second referee', candidate_interface_references_start_path %>
+
+    <% else  %>
 
       <h2 class="govuk-heading-m">Add another referee</h2>
 
@@ -9,22 +25,6 @@
       <p class="govuk-body">We’ll cancel any remaining requests when you’ve received 2 references.</p>
 
       <%= govuk_button_link_to 'Add another referee', candidate_interface_references_start_path %>
-
-    <% elsif no_viable_references? %>
-
-      <h2 class="govuk-heading-m">Add a referee</h2>
-
-      <p class="govuk-body">You need 2 references before you can submit your application.</p>
-
-      <%= govuk_button_link_to 'Add a referee', candidate_interface_references_start_path %>
-
-    <% else %>
-
-      <h2 class="govuk-heading-m">Add a second referee</h2>
-
-      <p class="govuk-body">You need 2 references before you can submit your application.</p>
-
-      <%= govuk_button_link_to 'Add a second referee', candidate_interface_references_start_path %>
 
     <% end %>
   </div>
--- a/app/components/candidate_interface/add_reference_component.html.erb
+++ b/app/components/candidate_interface/add_reference_component.html.erb
@@ -1,0 +1,31 @@
+<% unless application_form.enough_references_have_been_provided? %>
+  <div class="govuk-inset-text govuk-!-margin-top-0">
+    <% if enough_references_have_been_added? %>
+
+      <h2 class='govuk-heading-m'>Add another referee</h2>
+
+      <p class="govuk-body">You can add more referees to increase the chances of getting 2 references quickly.</p>
+
+      <p class="govuk-body">We’ll cancel any remaining requests when you’ve received 2 references.</p>
+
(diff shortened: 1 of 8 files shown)


davidgisbey:
Paul. I thought about localising the strings.

paulrobertlloyd:
> Paul. I can localise the strings, but wanted to get your thoughts as the differences between them are quite subtle and it might obfuscate them. Thought I'd open it up to PR and see what you think.

We currently localise page titles and form text, but there’s a whole other category which we don’t, such as this. Not sure what the solution is to that (or even if one is needed), but don’t think we need to localise the strings here.

app/components/candidate_interface/add_reference_component.html.erb
```diff
@@ -0,0 +1,31 @@
+<% unless application_form.enough_references_have_been_provided? %>
+  <div class="govuk-inset-text govuk-!-margin-top-0">
+    <% if enough_references_have_been_added? %>
+
+      <h2 class='govuk-heading-m'>Add another referee</h2>
```
paulrobertlloyd, line 5:
```suggestion
      <h2 class="govuk-heading-m">Add another referee</h2>
```

app/components/candidate_interface/add_reference_component.html.erb
```diff
@@ -0,0 +1,31 @@
...
+
+      <p class="govuk-body">You can add more referees to increase the chances of getting 2 references quickly.</p>
+
+      <p class="govuk-body">We’ll cancel any remaining requests when you’ve received 2 references.</p>
+
+      <%= link_to 'Add another referee', candidate_interface_references_start_path, class: 'govuk-button' %>
```
paulrobertlloyd, line 11: Think it’s possible to do this now:

```suggestion
      <%= govuk_button_link_to 'Add another referee', candidate_interface_references_start_path %>
```

app/components/candidate_interface/add_reference_component.html.erb
```diff
@@ -0,0 +1,31 @@
...
+
+    <% elsif no_viable_references? %>
+
+      <h2 class='govuk-heading-m'>Add a referee</h2>
```
paulrobertlloyd, line 15:
```suggestion
      <h2 class="govuk-heading-m">Add a referee</h2>
```

app/components/candidate_interface/add_reference_component.html.erb
```diff
@@ -0,0 +1,31 @@
...
+
+      <p class="govuk-body">You need 2 references before you can submit your application.</p>
+
+      <%= link_to 'Add a referee', candidate_interface_references_start_path, class: 'govuk-button' %>
```
paulrobertlloyd, line 19:
```suggestion
      <%= govuk_button_link_to 'Add a referee', candidate_interface_references_start_path %>
```

app/components/candidate_interface/add_reference_component.html.erb
```diff
@@ -0,0 +1,31 @@
...
+
+    <% else %>
+
+      <h2 class='govuk-heading-m'>Add a second referee</h2>
```
paulrobertlloyd, line 23:
```suggestion
      <h2 class="govuk-heading-m">Add a second referee</h2>
```

app/components/candidate_interface/add_reference_component.html.erb
```diff
@@ -0,0 +1,31 @@
...
+
+      <p class="govuk-body">You need 2 references before you can submit your application.</p>
+
+      <%= link_to 'Add a second referee', candidate_interface_references_start_path, class: 'govuk-button' %>
```
paulrobertlloyd, line 27:
```suggestion
      <%= govuk_button_link_to 'Add a second referee', candidate_interface_references_start_path %>
```

paulrobertlloyd (changes requested):
I need to test the logic; not sure if viable === given references only. 🤔

davidgisbey:
Made the changes. Sorry I was on autopilot while in a v important meeting. So i think the viable references in this case are:

image

Cancelled/declined/bounced emails shouldn't be counted into their two 'potentially successful' references, but not_requested_yet, received and requested references should be right?

gregknight1:
Content LGTM.

paulrobertlloyd:
This seems to work, just a bit hard to wrap my head around the logic as written. Wonder if it’d be a bit easier to parse if used same terminology across the different scenarios, so `minimum_viable_references` and `no_viable_references`?

davidgisbey:
would it be easier to have 

no_viable_references?
 do x 
one_viable_reference?
  do y 
else 
 do z

paulrobertlloyd:
> would it be easier to have
> 
> no_viable_references?
> do x
> one_viable_reference?
> do y
> else
> do z

Yeah, I think so. The ‘add a second referee’ state is a bit of an outlier.

paulrobertlloyd (approved):

paulrobertlloyd:
Much clearer – thanks @davidgisbey!

PR #3472, opened 2020-11-19
Fix test that is flaky locally

benswannack:
## Context
Some tests have been failing locally for me due to the timezone being changed and not reset in a case. Not sure why this doesn't affect deployments though.

## Guidance to review
Is this change in timezone necessary? Can we just remove that instead?

### DIFF
--- a/spec/system/candidate_interface/offers_and_withdrawls/candidate_receives_rejection_email_spec.rb
+++ b/spec/system/candidate_interface/offers_and_withdrawls/candidate_receives_rejection_email_spec.rb
@@ -4,11 +4,13 @@ RSpec.feature 'Receives rejection email' do
   include CandidateHelper
 
   around do |example|
+    original_time_zone = Time.zone
     Time.zone = 'Pacific Time (US & Canada)'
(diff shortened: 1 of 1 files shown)


duncanjbrown:
@stevehook do you remember anything about https://github.com/DFE-Digital/apply-for-teacher-training/commit/59054414e169c741301b518c1fa2350cb7d779ba ?

stevehook:
> @stevehook do you remember anything about [5905441](https://github.com/DFE-Digital/apply-for-teacher-training/commit/59054414e169c741301b518c1fa2350cb7d779ba) ?

I think I did that to reliably reproduce an error that was happening on CI but not locally but I made a mistake not resetting it. So I think the fix in this PR is good but I'm wondering whether we need the timezone switch at all now - I'd be tempted to just remove it because I don't think it's guarding against anything that is likely to reoccur.

benswannack:
Thanks @stevehook, I'll have a play without it and remove it if I can't see anything wrong

benswannack:
I've tested this a few times with the timezone change removed, so going with that

stevehook (approved):

PR #3478, opened 2020-11-20
Refactor other qualifications wizard

stevehook:
## Context

After converting the candidate interface other qualification forms to use a wizard that defers saving changes until the last step we identified a few areas that could be refactored to reduce complexity and make it easier to apply the same change to other parts of the candidate interface.

- Move validation and persistence back into per-page form objects
- Change the wizard to a simple `IntermediateDataService` that just deals with persistence to Redis.

## Guidance to review

- There may still be some overlap between `OtherQualificationTypeForm` and `OtherQualificationTypeForm` and there may be opportunities to factor out some common stuff.
- By dropping `OtherQualificationWizard` we lose a way to validate the 'whole' qualification in a single step. I've not tried to 'fix' this.

## Link to Trello card

https://trello.com/c/YWtsBdnu/2506-refactor-other-qualifications-wizard-implementation

## Things to check


app/controllers/candidate_interface/other_qualifications/details_controller.rb
```diff
@@ -1,106 +1,99 @@
...
-          current_qualification
-        else
-          current_application.application_qualifications.build(
-            level: ApplicationQualification.levels[:other],
-          )
-        end
-
-      application_qualification.assign_attributes(@wizard.attributes_for_persistence)
-      application_qualification.save!
+    def detail_attributes(application_qualification)
```
malcolmbaig, line 79: Is this method used anywhere?

app/forms/candidate_interface/other_qualification_detail_form.rb
```diff
@@ -0,0 +1,174 @@
...
+        end
+      when OtherQualificationTypeForm::AS_LEVEL_TYPE
+        unless grade.in?(AS_LEVEL_GRADES)
+          errors.add(:grade, :invalid)
+        end
+      end
+    end
+
+    def sanitize_grade_where_required
+      if qualification_type.in? [OtherQualificationTypeForm::A_LEVEL_TYPE, OtherQualificationTypeForm::AS_LEVEL_TYPE]
```
malcolmbaig, line 169: We now need GCSE in this list as well, following https://github.com/DFE-Digital/apply-for-teacher-training/pull/3477/files

app/services/candidate_interface/intermediate_data_service.rb
```diff
@@ -0,0 +1,33 @@
...
+      state = @state_store&.read
+
+      if state
+        JSON.parse(state)
+      else
+        {}
+      end
+    end
+  end
+end
```
malcolmbaig, line 33: :ok_hand:

app/controllers/candidate_interface/other_qualifications/details_controller.rb
```diff
@@ -1,106 +1,99 @@
 module CandidateInterface
   class OtherQualifications::DetailsController < OtherQualifications::BaseController
     def new
-      @wizard = wizard_for(current_step: :details)
-      @wizard.qualification_type ||= params[:qualification_type]
-      @wizard.initialize_new_qualification(
+      @form = form_for(current_step: :type)
```
malcolmbaig, line 4: I think it might be easier to follow the form initialization logic if each action instantiated the form directly in the state that it's after, rather than via the `form_for` method.  That might make it easier to spot further refinements to the form object's API too.

For example - is it possible to move from an options hash to keyword arguments?

app/controllers/candidate_interface/other_qualifications/destroy_controller.rb
```diff
@@ -2,7 +2,7 @@ module CandidateInterface
   class OtherQualifications::DestroyController < OtherQualifications::BaseController
     before_action :render_application_feedback_component, except: %i[confirm_destroy destroy]
     def confirm_destroy
-      @qualification = OtherQualificationWizard.build_from_qualification(current_qualification)
+      @qualification = OtherQualificationDetailForm.build_from_qualification(current_qualification)
```
malcolmbaig, line 5: tiny naming nitpick - `DetailsForm`

app/controllers/candidate_interface/other_qualifications/type_controller.rb
```diff
@@ -1,92 +1,92 @@
...
     end
 
     def edit
-      @wizard = wizard_for(current_step: :type)
-      @wizard.copy_attributes(current_qualification)
-      @wizard.save_state!
+      @form = form_for(
+        current_step: :type,
+        initialize_from_db: true,
+        checking_answers: true,
```
malcolmbaig, line 33: Naming suggestion - `editing`?

malcolmbaig:
Some comments left inline, but overall some really good changes here :+1:

app/controllers/candidate_interface/other_qualifications/type_controller.rb
stevehook: yes makes more sense. done

app/controllers/candidate_interface/other_qualifications/details_controller.rb
stevehook: `form_for` doesn't really reduce the LOC count much (if at all) and I think it's clearer

app/controllers/candidate_interface/other_qualifications/details_controller.rb
stevehook: nope, good spot

malcolmbaig (approved):
:+1:

PR #3480, opened 2020-11-20
Add logic to EAFL forms to validate against future dated awards

george-bit:
## Context

Whilst fixing another bug on the IELTs form it was noticed that the form allowed the submittal of future dates.
On searching it was found that the other eafl forms contained this bug. 

This fix updates the logic in the validation
method to include a year limit and throw a different validation error in accordance. Yml updated and new unit test
added to each form spec.
image
image
image

## Guidance to review

Dev - Check that Test coverage is sufficient.
UX/UI - Check that Error is displayed correctly, I notice that the TOEFL and Other EAFL don't seem to display error summaries should I also make this tweak?

## Link to Trello card

https://trello.com/c/27fRszKG/2568-bug-ielts-and-possibly-other-eafl-forms-validating-future-dates

## Things to check

### DIFF
--- a/app/forms/candidate_interface/english_foreign_language/ielts_form.rb
+++ b/app/forms/candidate_interface/english_foreign_language/ielts_form.rb
@@ -41,8 +41,12 @@ module CandidateInterface
     private
 
       def award_year_is_a_valid_year
+        year_limit = Time.zone.today.year.to_i + 1
+
         if !valid_year?(award_year)
           errors.add(:award_year, :invalid)
+        elsif award_year.to_i >= year_limit
+          errors.add(:award_year, :in_the_future, date: year_limit)
         end
       end
 
--- a/app/forms/candidate_interface/english_foreign_language/ielts_form.rb
+++ b/app/forms/candidate_interface/english_foreign_language/ielts_form.rb
@@ -41,12 +41,10 @@ module CandidateInterface
     private
 
       def award_year_is_a_valid_year
-        year_limit = Time.zone.today.year.to_i + 1
-
         if !valid_year?(award_year)
           errors.add(:award_year, :invalid)
-        elsif award_year.to_i >= year_limit
-          errors.add(:award_year, :in_the_future, date: year_limit)
+        elsif future_year?(award_year)
+          errors.add(:award_year, :in_the_future)
         end
       end
 
--- a/app/lib/validation_utils.rb
+++ b/app/lib/validation_utils.rb
@@ -4,6 +4,6 @@ module ValidationUtils
   end
 
   def future_year?(year)
-    year.to_i >= Time.zone.today.year
+    year.to_i > Time.zone.today.year
   end
 end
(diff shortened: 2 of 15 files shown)


app/forms/candidate_interface/english_foreign_language/ielts_form.rb
```diff
@@ -41,8 +41,12 @@ def fill(ielts:)
     private
 
       def award_year_is_a_valid_year
+        year_limit = Time.zone.today.year.to_i + 1
+
         if !valid_year?(award_year)
           errors.add(:award_year, :invalid)
+        elsif award_year.to_i >= year_limit
+          errors.add(:award_year, :in_the_future, date: year_limit)
```
davidgisbey, line 49: we're now using this same logic 4 times in different files. What do you think about extracting it into ValidationUtils like we do with valid_year?

davidgisbey:
Tested it locally and it's all working as expected 👍 I've just left one comment about potentially reducing some duplicaiton.

app/forms/candidate_interface/english_foreign_language/ielts_form.rb
george-bit: Not a bad shout lemme have a look 👍

davidgisbey (approved):
Looks good. I left an optional comment for a personal preference haha

app/lib/validation_utils.rb
```diff
@@ -2,4 +2,12 @@ module ValidationUtils
...
     year.to_s.match?(/^[12]\d{3}$/)
   end
+
+  def future_year?(year)
+    year.to_i >= year_limit
+  end
+
+  def year_limit
+    Time.zone.today.year.to_i + 1
+  end
```
frankieroberto, line 12: Suggest simplifying this as:

```suggestion
  def future_year?(year)
    year.to_i >= Time.zone.today.year
  end
```

george-bit: Hey Frankie, unfortunately this won't work because `year_limit` needs to be a separate method as it is called by the error message in the form's validation method as well. Happy to discuss.

config/locales/application_form.yml
```diff
@@ -705,6 +705,7 @@ en:
             award_year:
               blank: Enter year the assessment was done
               invalid: Enter a real year
+              in_the_future: Enter a year before %{date}
```
frankieroberto, line 708: I think the error message should reference what the year is ("assessment year"?), and perhaps doesn't need to include the next year?

```suggestion
              in_the_future: Assessment year must be this year or a previous year
```

or

```suggestion
              in_the_future: Assessment year must not be in the future
```

app/lib/validation_utils.rb
frankieroberto: Sorry, I forgot to batch up this comment with the next one...

Let's talk to a content designer.

george-bit: Have implemented in accordance with below change to error message

frankieroberto (approved):

PR #3482, opened 2020-11-20
Multiple English GCSE tidy up

raamSoftwire:
## Context

When adding multiple GCSEs I intentionally left out a few things as we thought we were going to refactor how we store English GCSEs in the database.

After recent conversations that refactor is no longer required, so I need to go back and finish off what is left.

This includes 

- adding a test for the activated case on the `EnglishGcseGradeForm`
- trying to simplify the validation on the `EnglishGcseGradeForm`

## Link to Trello card

https://trello.com/c/AbDpmFPY/2570-multiple-english-gcse-tidy-up

### DIFF
--- a/app/controllers/candidate_interface/gcse/english/grade_controller.rb
+++ b/app/controllers/candidate_interface/gcse/english/grade_controller.rb
@@ -16,9 +16,7 @@ module CandidateInterface
       @qualification_type = gcse_english_qualification.qualification_type
       @gcse_grade_form = english_gcse_grade_form.assign_values(english_details_params)
 
-      save_successful = multiple_gsces_are_active? && @qualification_type == 'gcse' ? @gcse_grade_form.save_grades : @gcse_grade_form.save_grade
-
-      if save_successful
+      if @gcse_grade_form.save
         update_gcse_completed(false)
         redirect_to next_gcse_path
       else
--- a/app/forms/candidate_interface/english_gcse_grade_form.rb
+++ b/app/forms/candidate_interface/english_gcse_grade_form.rb
@@ -144,20 +144,20 @@ module CandidateInterface
       self
     end
 
-    def save_grades
-      if !valid?(:structured_grades)
-        log_validation_errors(:structured_grades)
-        return false
+    def save
+      if is_multiple_gcse?
+        if !valid?(:structured_grades)
+          log_validation_errors(:structured_grades)
+          return false
+        end
+        qualification.update(structured_grades: build_grades_json, grade: nil)
+      else
+        if !valid?(:grade)
+          log_validation_errors(:grade)
+          return false
+        end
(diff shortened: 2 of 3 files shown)


app/forms/candidate_interface/english_gcse_grade_form.rb
```diff
@@ -144,20 +144,20 @@ def assign_values(params)
...
-      qualification.update(structured_grades: build_grades_json, grade: nil)
-    end
-
-    def save_grade
-      if !valid?(:grade)
-        log_validation_errors(:grade)
-        return false
+    def save
+      if is_multiple_gcse?
+        if !valid?(:structured_grades)
```
TobyRet, line 149: Worth extracting a little more for reaability?
You could have something like

```
def save
  is_multiple_gcse? ? save_grades : save_grade
end

```

spec/forms/candidate_interface/english_gcse_grade_form_spec.rb
```diff
@@ -11,6 +11,63 @@
...
+          form.grade_other_english_gcse = ''
+          form.validate(:structured_grades)
+
+          expect(form.errors[:other_english_gcse_name]).to include('Enter an English GCSE')
+          expect(form.errors[:grade_other_english_gcse]).to include('Enter your grade')
+        end
+      end
+
       context 'multiple GCSEs disabled' do
         FeatureFlag.deactivate('multiple_english_gcses')
```
TobyRet, line 72: This line won't get executed - should wrap in a `before`. Perhaps you do not require anyway as the default is false,  but might pay to be explicit.

app/controllers/candidate_interface/gcse/english/grade_controller.rb
```diff
@@ -16,9 +16,7 @@ def update
       @qualification_type = gcse_english_qualification.qualification_type
       @gcse_grade_form = english_gcse_grade_form.assign_values(english_details_params)
 
-      save_successful = multiple_gsces_are_active? && @qualification_type == 'gcse' ? @gcse_grade_form.save_grades : @gcse_grade_form.save_grade
-
-      if save_successful
+      if @gcse_grade_form.save
```
TobyRet, line 19: I think it's good that the interface to the form is simplified and the logic is now hidden 👍

TobyRet (changes requested):
One small change required here
https://github.com/DFE-Digital/apply-for-teacher-training/pull/3482/files#diff-6d33080ac7371db960bc7bd708de186251be476e82619a7af88b7c704cbeb032R72

One small suggested added as well but it's minor / not a blocker

app/forms/candidate_interface/english_gcse_grade_form.rb
raamSoftwire: Good idea, thanks

TobyRet (approved):

PR #3483, opened 2020-11-20
Update word limit

raamSoftwire:
## Context

Increasing word limit for references.

Simple change in the form and the view

## Guidance to review

Is there anywhere else we need to look for changes?
I have spoken to the Provendor team and we don't have any concerns about knock-on effects to the API. See https://ukgovernmentdfe.slack.com/archives/CN1MCQCHZ/p1605883511484200

## Link to Trello card

https://trello.com/c/QYDlevJV/2523-dev-increase-word-limit-for-references-to-500wds

### DIFF
--- a/app/forms/referee_interface/reference_feedback_form.rb
+++ b/app/forms/referee_interface/reference_feedback_form.rb
@@ -2,7 +2,7 @@ module RefereeInterface
   class ReferenceFeedbackForm
     include ActiveModel::Validations
     attr_reader :reference, :feedback
-    validates :feedback, presence: true, word_count: { maximum: 300 }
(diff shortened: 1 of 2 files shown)


raamSoftwire:
It would be nice to add a test to cover this change. Something similar to what we have in the `further_information_form_spec.rb`. But currently the `reference_feedback_form` doesn't have it's own test file and it looks like the validations are being carried out in `reference_review_form_spec` instead. I've spent some time trying to add a test but I can't make it stick.

If you think adding a test for this is worth it then I'd appreciate a bit of pairing to see what I'm missing here.

davidgisbey (approved):

PR #3485, opened 2020-11-20
Add govuk error summaries and track validation errors where missing

davidgisbey:
## Context

Recently, a few pages have been missing govuk error summaries so I thought I'd have a glance through our views. While doing this I thought I'd check the corresponding controllers and noticed that in some cases we aren't tracking validation errors

- Add govuk_error_summary to views where it's missing
- Track validation errors when missing in the corresponding controllers

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/english_foreign_language/ielts_controller.rb
+++ b/app/controllers/candidate_interface/english_foreign_language/ielts_controller.rb
@@ -15,6 +15,7 @@ module CandidateInterface
         if @ielts_form.save
           redirect_to candidate_interface_english_foreign_language_review_path
         else
+          track_validation_error(@ielts_form)
           render :new
         end
       end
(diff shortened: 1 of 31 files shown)


app/controllers/candidate_interface/english_foreign_language/ielts_controller.rb
```diff
@@ -32,6 +33,7 @@ def update
         if @ielts_form.save
           redirect_to candidate_interface_english_foreign_language_review_path
         else
+          track_validation_error(@ielts_form)
           render :new
```
stevehook, line 37:
```suggestion
          render :edit
```

stevehook (approved):
Changes look good. One extra one suggested.

app/controllers/candidate_interface/english_foreign_language/ielts_controller.rb
davidgisbey: Nice spot!

PR #3488, opened 2020-11-23
Try to fix issues with data export

tijmenb:
## Context

Some data exports are failing silently - they just stop doing their thing and aren't logging anything. My suspicion is that we're loading too many records into memory and that destroys the worker somehow.

Use `find_each` to load at most a 100 items into memory every time. Also adds some fixes for the logging.

I've tested this manually on production by replacing the class in the console and running that.

## Guidance to review

Does this make sense?

## Link to Trello card

https://trello.com/c/C7NyIXm9/2574-bug

## Things to check

### DIFF
--- a/app/workers/data_exporter.rb
+++ b/app/workers/data_exporter.rb
@@ -14,6 +14,7 @@ class DataExporter
         importer_class.constantize.new.data_for_export,
       )
     rescue StandardError => e
+      Rails.logger.info "Export generation failed: `#{e.message}`"
       data_export.update!(audit_comment: "Export generation failed: `#{e.message}`")
       raise
     end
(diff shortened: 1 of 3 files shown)


app/workers/data_exporter.rb
```diff
@@ -28,6 +28,8 @@ def perform(importer_class, data_export_id)
     )
 
     Rails.logger.info 'Finished writing CSV. Sidekiq done'
+  ensure
+    RequestLocals.clear_all!
```
duncanjbrown, line 32: Doesn't RequestStore::Sidekiq already do this? https://github.com/madebylotus/request_store-sidekiq/blob/master/lib/request_store/sidekiq/server_middleware.rb

It looks out of date with the guidance in https://github.com/ElMassimo/request_store_rails#usage-in-sidekiq

duncanjbrown (approved):
Q about `RequestStore` but not to block this fix

tijmenb:
@duncanjbrown removed the commit, I'll investigate further.

PR #3491, opened 2020-11-23
Add latitude and longitude to application_forms

malcolmbaig:
## Context
We want to analyse distances between candidates and the sites they
choose to train. 

Add columns to the application_forms table that will
allow us to start geocoding candidate addresses.

## Link to Trello card

https://trello.com/c/6A8t8vVY
## Things to check

  - It doesn't, but this isn't needed right now. Can (and will) be handled in a separate PR.


davidgisbey (approved):

TobyRet (changes requested):
Think it may be worth considering what library will be used for geocoding as it may specify the type you require for lat and long. 

E.g. Teacher Training API uses the Geokit gem
https://github.com/geokit/geokit-rails#how-to-find-all-stores-within-a-10-mile-radius-of-a-given-latlng

Possibly the rationale for floats may be as described here - https://stackoverflow.com/a/18182565

stevehook:
> Think it may be worth considering what library will be used for geocoding as it may specify the type you require for lat and long.
> 
> E.g. Teacher Training API uses the Geokit gem
> https://github.com/geokit/geokit-rails#how-to-find-all-stores-within-a-10-mile-radius-of-a-given-latlng
> 
> Possibly the rationale for floats may be as described here - https://stackoverflow.com/a/18182565

I was wondering about the data type for these too. Didn't see this earlier but I went with `float`s full the `sites` table. Will need to change those if we go with decimals here. https://github.com/DFE-Digital/apply-for-teacher-training/pull/3498

malcolmbaig:
Library:

I opted for the [geocoder](https://github.com/alexreisner/geocoder) gem as it's actively maintained and gives us the distance calculations we need for Locations work. Interestingly, Find uses both geocoder and geokit, although for our purposes I think we only need the one. Open to other suggestions if they're a better fit :+1: 

Data type:

Good shout, and definitely worth discussing. I just defaulted to decimal because of precision at the expense of storage. Do we understand the rationale for Find adopting floats?

davidgisbey:
Publish also use floats. Might be worth considering as we're going to pulling lat/long for Provider and Sites from them.

TobyRet:
> Library:
> 
> I opted for the [geocoder](https://github.com/alexreisner/geocoder) gem as it's actively maintained and gives us the distance calculations we need for Locations work. Interestingly, Find uses both geocoder and geokit, although for our purposes I think we only need the one. Open to other suggestions if they're a better fit 👍
> 
> Data type:
> 
> Good shout, and definitely worth discussing. I just defaulted to decimal because of precision at the expense of storage. Do we understand the rationale for Find adopting floats?

I think it may be worth checking if the library requires you to use a certain type (e.g. Geokit gem). I believe Floats are accurate up to 6 d.p and this will give still give you location accuracy to +- 11cm. You then also get the optmisation benefit over `decimal`

malcolmbaig:
Good to know, and the accuracy argument makes sense to me :+1: . We'll have more than enough precision in practice, especially as the default  column type created by a Rails migration using the `float` method appears to be `double precision` (looking at #3498 ), ie - 15 dp.

Will update this to use float.

(sources, for posterity):
https://www.postgresql.org/docs/9.6/datatype-numeric.html
https://en.wikipedia.org/wiki/Decimal_degrees#Precision

malcolmbaig:
> I think it may be worth checking if the library requires you to use a certain type

Geocoder supports both floats and decimals.

stevehook (approved):

TobyRet (approved):

PR #3493, opened 2020-11-23
Applying for the same course twice bug

raamSoftwire:
## Context

When a user arrives on the `confirm_selection` page from Find and confirms to add a course, they can at that point use the back button to go back and complete the journey again which lets them join the same course twice.

Add validation to the `application_choice` model and to the `find_course_selections_controller`.

## Link to Trello card

https://trello.com/c/RfVH10aE/2491-dev-bug-users-can-apply-for-the-same-course-twice-when-using-the-back-button

### DIFF
--- a/app/models/application_choice.rb
+++ b/app/models/application_choice.rb
@@ -12,6 +12,7 @@ class ApplicationChoice < ApplicationRecord
   has_one :accredited_provider, through: :course, class_name: 'Provider'
 
   has_many :notes, dependent: :destroy
+  validates :course_option, uniqueness: { scope: :application_form_id }
 
   audited associated_with: :application_form
 
--- a/app/controllers/candidate_interface/find_course_selections_controller.rb
+++ b/app/controllers/candidate_interface/find_course_selections_controller.rb
@@ -14,7 +14,7 @@ module CandidateInterface
       course = Course.find(params[:course_id])
 
       course_selection = CourseSelectionForm.new(course, course_selection_params[:confirm])
-      if !course_selection.confirm
+      if !course_selection.confirm || current_application.contains_course?(course)
         redirect_to candidate_interface_course_choices_index_path
         return
       end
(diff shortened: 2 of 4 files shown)


davidgisbey (approved):
Looks good 👍

app/models/application_choice.rb
```diff
@@ -12,6 +12,7 @@ class ApplicationChoice < ApplicationRecord
   has_one :accredited_provider, through: :course, class_name: 'Provider'
 
   has_many :notes, dependent: :destroy
+  validates :course_option, uniqueness: { scope: :application_form_id }
```
davidgisbey, line 15: Can we just add this as a test please 👍 

    it { is_expected.to validate_uniqueness_of(:course_option).scoped_to(:application_form_id) }

raamSoftwire: Sorted

raamSoftwire:
@davidgisbey I'm having trouble convincing Rubocop about the validation on the model.

```Rails/UniqueValidationWithoutIndex: Uniqueness validation should be with a unique index.
  validates :course_option_id, uniqueness: { scope: :application_form_id }
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Are you able to give me any pointers here? I've seen thgat we're doing something very similar in `course.rb` but can't see what I'm missing.

davidgisbey:
There needs to be an index on the table like this 

image

davidgisbey:
image

Got it working after more time than I'd care to admit haha

raamSoftwire:
> image
> 
> Got it working after more time than I'd care to admit haha

Thanks David, that was really helpful. I've ran it locally to make sure everything is working as expected. I've raised another PR for that migration. https://github.com/DFE-Digital/apply-for-teacher-training/pull/3497

PR #3496, opened 2020-11-24
Add Activity Log query service

mnacos:
## Context

We're planning to introduce an activity log within the provider interface which will show application events in reverse chronological order across all application choices visible to the user.

This list will be restricted to application choice events and exclude notes. Each item on that list will contain a short description of the event, a timestamp and a link to the corresponding application.

The main risk here is performance-related, especially as the number of application choices on the system will increase continuously. This PR is meant to spike/derisk that.

Introduce a `GetActivityLogEvents` service which accepts an application choices query and returns a reverse chronological list of their 'update' events. A previous version of this required a `provider_user` context and called `GetApplicationChoicesForProviders` behind the scenes, but asking for `application_choices` instead provides more flexibility. The controller will be responsible for using `GetApplicationChoicesForProviders` and passing its outcome to `GetActivityLogEvents` to ensure the user only sees events for applications they have access to. But we'll also be able to pass any `ApplicationChoice` query to the new service in the rails console to assess the query's performance against production data.

## Guidance to review

The relevant spec `spec/queries/get_provider_activity_log_events_spec.rb` is a good place to start. The service returns a collection of `Audited::Audit` objects from a single sql query, which means they can be paginated the usual kaminari way. Each of these objects responds to `created_at` (the timestamp of the event) and `auditable` (which, for these objects will always be an `ApplicationChoice` and can be used to generate a preview and link). What's missing is a component for generating a short description of the event based on `audited_changes`, which will be handled at the ruby side (per page).

There is also an optional `since:` argument to the service, which will limit results to all events after a certain point in time (e.g. `15.days.ago`)

The combination of `application_choices:` and `since:` arguments gives us many options, for example allowing the user to choose the time window or filtering events based on a specific provider or course.

Performance-wise, the query seems to complete in a reasonable timeframe (e.g. 50ms scanning ~21000 audits for 1000 applications) but we'll also look at production data once this code is deployed.

## Link to Trello card

https://trello.com/c/OOMAetc1

## Things to check

### DIFF
--- a/app/queries/get_activity_log_events.rb
+++ b/app/queries/get_activity_log_events.rb
@@ -1,23 +1,24 @@
-class GetProviderActivityLogEvents
-  attr_reader :provider_user
+class GetActivityLogEvents
+  attr_reader :application_choices
 
-  def initialize(provider_user:)
-    @provider_user = provider_user
+  def initialize(application_choices:)
+    @application_choices = application_choices
   end
 
-  def call
-    scope = GetApplicationChoicesForProviders.call(providers: provider_user.providers)
+  def call(since: nil)
+    since ||= Time.zone.local(2018, 1, 1)
 
     Audited::Audit.from <<~COMBINE_AUDITS_WITH_APPLICATION_CHOICES_SCOPE.squish
       (
-        SELECT a.*, '' AS description
+        SELECT a.*
           FROM audits a
           INNER JOIN application_choices ac
             ON auditable_id = ac.id
               AND auditable_type = 'ApplicationChoice'
               AND action = 'update'
-          INNER JOIN (#{scope.to_sql}) visible
-            ON ac.id = visible.t0_r0
+          INNER JOIN (#{application_choices.to_sql}) visible
+            ON ac.id = visible.id
+          WHERE a.created_at >= '#{since.iso8601}'::TIMESTAMPTZ
           ORDER BY a.created_at DESC
       ) AS audits
     COMBINE_AUDITS_WITH_APPLICATION_CHOICES_SCOPE
--- a/app/queries/get_activity_log_events.rb
+++ b/app/queries/get_activity_log_events.rb
@@ -1,12 +1,6 @@
 class GetActivityLogEvents
-  attr_reader :application_choices
-
-  def initialize(application_choices:)
-    @application_choices = application_choices
-  end
-
-  def call(since: nil)
-    since ||= Time.zone.local(2018, 1, 1)
+  def self.call(application_choices:, since: nil)
+    since ||= Time.zone.local(2018, 1, 1) # before the pilot began, i.e. all records
 
     Audited::Audit.from <<~COMBINE_AUDITS_WITH_APPLICATION_CHOICES_SCOPE.squish
       (
--- a/app/queries/get_application_choices_for_providers.rb
+++ b/app/queries/get_application_choices_for_providers.rb
@@ -21,5 +21,6 @@ class GetApplicationChoicesForProviders
       .or(ApplicationChoice.includes(*includes)
         .where('courses.accredited_provider_id' => providers, 'courses.recruitment_cycle_year' => RecruitmentCycle.years_visible_to_providers))
       .where('status IN (?)', statuses)
+      .select('"application_choices"."id" AS id')
   end
 end
--- a/app/queries/get_provider_activity_log_events.rb
+++ b/app/queries/get_provider_activity_log_events.rb
@@ -1,0 +1,25 @@
+class GetProviderActivityLogEvents
+  attr_reader :provider_user
+
+  def initialize(provider_user:)
+    @provider_user = provider_user
+  end
+
+  def call
+    scope = GetApplicationChoicesForProviders.call(providers: provider_user.providers)
+
+    Audited::Audit.from <<~COMBINE_AUDITS_WITH_APPLICATION_CHOICES_SCOPE.squish
+      (
+        SELECT a.*, '' AS description
+          FROM audits a
+          INNER JOIN application_choices ac
+            ON auditable_id = ac.id
+              AND auditable_type = 'ApplicationChoice'
+              AND action = 'update'
+          INNER JOIN (#{scope.to_sql}) visible
+            ON ac.id = visible.t0_r0
+          ORDER BY a.created_at DESC
+      ) AS audits
+    COMBINE_AUDITS_WITH_APPLICATION_CHOICES_SCOPE
+  end
+end
--- a/spec/queries/get_activity_log_events_spec.rb
+++ b/spec/queries/get_activity_log_events_spec.rb
@@ -1,13 +1,15 @@
 require 'rails_helper'
 
-RSpec.describe GetProviderActivityLogEvents, with_audited: true do
+RSpec.describe GetActivityLogEvents, with_audited: true do
   around do |example|
     @now = Time.zone.local(2020, 2, 11)
     Timecop.travel(@now) { example.run }
   end
 
-  let(:service) { GetProviderActivityLogEvents.new(provider_user: provider_user) }
   let(:provider_user) { create(:provider_user, :with_two_providers) }
+  let(:application_choices_for_provider_user) { GetApplicationChoicesForProviders.call(providers: provider_user.providers) }
+  let(:service) { GetActivityLogEvents.new(application_choices: application_choices_for_provider_user) }
+
   let(:course_provider_a) { create(:course, provider: provider_user.providers.first) }
   let(:course_provider_b) { create(:course, provider: provider_user.providers.second) }
   let(:course_unrelated) { create(:course) }
@@ -21,10 +23,15 @@ RSpec.describe GetProviderActivityLogEvents, with_audited: true do
   end
 
   def create_audit_for_application_choice(application_choice, changes: {})
-    create(:audit, user: provider_user, application_choice: application_choice, changes: changes)
+    create(:application_choice_audit, user: provider_user, application_choice: application_choice, changes: changes)
   end
 
   describe '#call' do
+    it 'raises an error unless application_choices responds_to #to_sql ' do
+      array = 2.times.map { create_application_choice_for_course(course_provider_a) }
+      expect { GetActivityLogEvents.new(application_choices: array).call }.to raise_error(NoMethodError)
+    end
+
     it 'returns an empty array if no audits are found' do
       expect(service.call).to eq([])
     end
@@ -37,9 +44,25 @@ RSpec.describe GetProviderActivityLogEvents, with_audited: true do
 
       expect(result.count).to eq(1)
 
-      %i[description created_at auditable].each do |attr|
+      %i[created_at auditable].each do |attr|
         expect(result.first).to respond_to(attr)
       end
+    end
+
+    it 'supports an since: argument for limiting number of events' do
+      choice = create_application_choice_for_course course_provider_a
+      create_audit_for_application_choice choice
+
+      expected = create(
+        :application_choice_audit,
+        user: provider_user,
+        application_choice: choice,
+        created_at: @now + 1.day,
+      )
+
+      result = service.call(since: @now + 6.hours)
+
+      expect(result.first).to eq(expected)
     end
   end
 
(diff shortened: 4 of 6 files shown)


app/queries/get_application_choices_for_providers.rb
```diff
@@ -21,5 +21,6 @@ def self.call(providers:, vendor_api: false, includes: DEFAULT_INCLUDES)
       .or(ApplicationChoice.includes(*includes)
         .where('courses.accredited_provider_id' => providers, 'courses.recruitment_cycle_year' => RecruitmentCycle.years_visible_to_providers))
       .where('status IN (?)', statuses)
+      .select('"application_choices"."id" AS id')
```
mnacos, line 24: This is an intersting one. The raw sql generated by `GetApplicationChoicesForProviders` labels the `id` column as `t0_r0`, probably because of the complexity of the query. This addition recasts `t0_r0` as `id`, which allows the resulting sql to be used in a join in a way consistent with any other `ApplicationChoice` Arel.

Because of this rename we can join on `id` as usual in the new service, which means you can write things like: `GetActivityLogEvents.new(application_choices: ApplicationChoice.where(....)).call` and expect them to work.

spec/queries/get_activity_log_events_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+
+    it 'returns objects responding to the required attributes' do
+      choice = create_application_choice_for_course course_provider_a
+      create_audit_for_application_choice choice
+
+      result = service.call
+
+      expect(result.count).to eq(1)
+
+      %i[created_at auditable].each do |attr|
```
benswannack, line 47: We should probably call this `auditable` `application_choice` instead, to avoid confusion

app/queries/get_activity_log_events.rb
```diff
@@ -0,0 +1,26 @@
+class GetActivityLogEvents
+  attr_reader :application_choices
+
+  def initialize(application_choices:)
+    @application_choices = application_choices
+  end
+
+  def call(since: nil)
```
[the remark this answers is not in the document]
benswannack: Now that this is sufficiently generic, we should be able to use this for individual application timelines :)

benswannack (approved):
Some discussion points inline, otherwise looks good

app/queries/get_activity_log_events.rb
```diff
@@ -0,0 +1,26 @@
...
+    since ||= Time.zone.local(2018, 1, 1)
```
despo, line 9: can this time be moved into a constant. What does it signify? Is this when the project begun?

spec/queries/get_activity_log_events_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+    course_option ||= create(:course_option, course: course)
+    create(:application_choice, :awaiting_provider_decision, course_option: course_option)
+  end
+
+  def create_audit_for_application_choice(application_choice, changes: {})
+    create(:application_choice_audit, user: provider_user, application_choice: application_choice, changes: changes)
+  end
+
+  describe '#call' do
+    it 'raises an error unless application_choices responds_to #to_sql ' do
```
despo, line 30: consider rewording this so it describes when something happens rather than when it doesn't e.g raises an error **when** application_choices **don'**t respond_to #to_sql

spec/queries/get_activity_log_events_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+
+    it 'supports an since: argument for limiting number of events' do
+      choice = create_application_choice_for_course course_provider_a
+      create_audit_for_application_choice choice
+
+      expected = create(
+        :application_choice_audit,
+        user: provider_user,
+        application_choice: choice,
+        created_at: @now + 1.day,
```
despo, line 60: I think you don't need an instance variable here as you should be able to use `1.day.from_now` to set the relevant creation time

spec/queries/get_activity_log_events_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+      )
+
+      result = service.call(since: @now + 6.hours)
```
despo, line 63: Same as above

spec/queries/get_activity_log_events_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+        %i[course_provider_a course_provider_b course_unrelated ratified_course_provider_b ratified_course_unrelated].each do |course|
+          choice = create_application_choice_for_course send(course)
+          20.times.map { create_audit_for_application_choice choice }
+        end
+      end
+
+      elapsed_time = Benchmark.measure { service.call }.real
+      puts "GetProviderActivityLogEvents #call completed in #{elapsed_time}s"
+
+      expect(elapsed_time).to be < 0.05
```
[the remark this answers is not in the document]
despo: I'm not sure of the benefit of running this as part of the test suite by default - especially taking into account how slow our test suite already issue.

Maybe it's worth adding a new performance check as part of the repository's GitHub actions, and tagging and excluding this from the default run, so in case there are any concerns the issue can be both noticed and also addressed as a separate issue.

app/queries/get_activity_log_events.rb
```diff
@@ -0,0 +1,26 @@
...
+          INNER JOIN application_choices ac
+            ON auditable_id = ac.id
+              AND auditable_type = 'ApplicationChoice'
+              AND action = 'update'
+          INNER JOIN (#{application_choices.to_sql}) visible
+            ON ac.id = visible.id
+          WHERE a.created_at >= '#{since.iso8601}'::TIMESTAMPTZ
+          ORDER BY a.created_at DESC
+      ) AS audits
+    COMBINE_AUDITS_WITH_APPLICATION_CHOICES_SCOPE
```
despo, line 24: Is this not achievable with AR, or at least with only partial SQL as Ben mentioned below?

app/queries/get_activity_log_events.rb
mnacos: It doesn't really mean anything... It's just a date before the pilot started so that the service returns all relevant records, effectively not limiting results based on time. I didn't want to add any conditionals in the code below. Suggestions for alternatives welcome.

spec/queries/get_activity_log_events_spec.rb
mnacos: `application_choice` would be more readable, yes, but the query returns `Audited::Audit` objects which already expose `.auditable` as method for referencing the polymorphic `auditable_type` and `auditable_id` data. For efficiency I've decided not to decorate or manipulate the results so that they have an `application_choice`.

The only code which will need to call `.auditable` will be inside the view or the view_component presenting the elements of the list (which are audit events), so I think this is a reasonable compromise.

spec/queries/get_activity_log_events_spec.rb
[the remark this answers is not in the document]
mnacos: True but we don't have support for automated tests against deployed environments yet (well, apart for smoke tests, but these won't help us in this case). I'd be reluctant to run this or similar code on one of our permanent deployed environments, because it will generate lots of junk. The advantage of running this within a test context is the database cleanup.

app/queries/get_activity_log_events.rb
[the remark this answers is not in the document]
mnacos: This would be lovely but we need to establish the activity log and the timeline should be similar things. Currently they are not, notes probably being the biggest difference between the two. But it could perhaps replace `FindStatusChangeAudits`. Some of its `audit.audited_changes.key?('status')` could be turned into scopes.

app/queries/get_application_choices_for_providers.rb
mnacos: This comment is for `GetActivityLogEvents`, not `GetApplicationChoicesForProviders`. Will post a reply on that file.

app/queries/get_activity_log_events.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    Audited::Audit.from <<~COMBINE_AUDITS_WITH_APPLICATION_CHOICES_SCOPE.squish
+      (
+        SELECT a.*
+          FROM audits a
+          INNER JOIN application_choices ac
+            ON auditable_id = ac.id
+              AND auditable_type = 'ApplicationChoice'
+              AND action = 'update'
+          INNER JOIN (#{application_choices.to_sql}) visible
```
mnacos, line 13: As Ben has already hinted at, the `INNER JOIN (#{application_choices.to_sql})` segment is the most important part of this query. It allows to search for `Audited::Audit` records while limiting them to the ones associated with application choices from the `GetApplicationChoicesForProviders` query.

There are multiple challenges with this query: a. `Audited::Audit` has a polymorphic association with `ApplicationChoice`, not a direct one, b. we need to somehow combine the query from `GetApplicationChoicesForProviders` into this one and c. this external query is `ApplicationChoice`-based but we need an `Audited::Audit`-based query overall.

Because the root of the query is audits, the multiplexing and ordering of events across applications is straightforward and fast. Considering the main aim of this story is speed, I'm happy to sacrifice some readability for performance.

app/queries/get_activity_log_events.rb
mnacos: I haven't found a way to get an equivalent result with ActiveRecord, so moving the `created_at` restriction into an `.order(...)` clause will not achieve much. Open to suggestions for the overall query, however.

spec/queries/get_activity_log_events_spec.rb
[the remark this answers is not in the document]
mnacos: Sorry Despo, missed your comment. This scenario is meant to be run manually and is actually skipped for normal rspec runs. Perhaps I should just remove it.

app/queries/get_provider_activity_log_events.rb
```diff
@@ -0,0 +1,25 @@
...
+    Audited::Audit.from <<~COMBINE_AUDITS_WITH_APPLICATION_CHOICES_SCOPE.squish
+      (
+        SELECT a.*, '' AS description
+          FROM audits a
+          INNER JOIN application_choices ac
+            ON auditable_id = ac.id
+              AND auditable_type = 'ApplicationChoice'
+              AND action = 'update'
+          INNER JOIN (#{scope.to_sql}) visible
+            ON ac.id = visible.t0_r0
```
duncanjbrown: Witchcraft

app/queries/get_activity_log_events.rb
duncanjbrown: Call `GetApplicationChoicesForProvider.call` seems to run that query whatever — ie it's a fixed cost we'll need to take anyway. Is there a way to not run that query in advance, and if not, could we replace this `JOIN` with a `where`?

duncanjbrown: Ah — no! I think `pry` is forcing it to realise when it prints the result. `GetApplicationChoicesForProviders.call(providers: pu.providers);` doesn't run the query.

duncanjbrown:
The log includes events initiated by the Candidate which technically happened before the application was sent to the provider, specifically the setting of the RBD date. Scoping to non-`Candidate` users could help? We'd lose the initial application-sent event but could include deliberately.

Screenshot 2020-11-25 at 09 56 40

mnacos:
> The log includes events technically initiated by the Candidate which happened before the application was sent to the provider, specifically the setting of the RBD date. Scoping to non-`Candidate` users could help? We'd lose the initial application-sent event but could include deliberately.

Thanks for spotting this but let's deal with it in the next PR. I've kept the query service as generic as possible for now, we can certainly filter further by chaining audit-based `.where(...)` clauses to main query. I wouldn't filter out candidate-driven events, however, as this may impact offer acceptance/decline events, withdrawals etc. down the line. Maybe we'll just choose to ignore the setting of RBD.

duncanjbrown (approved):

mnacos:
## Production data evaluation

TLDR; All events for all time for a single provider takes half a second. Pagination seems like a better way to increase performance than restricting the time window.

#### Gorse SCITT (all time)
```
irb(main):008:0> GetActivityLogEvents.call(application_choices: GetApplicationChoicesForProviders.call(providers: [gorse])).count
=> 109513
```
Query takes: __489ms__

#### Gorse SCITT (all time, first page)
```
irb(main):017:0> GetActivityLogEvents.call(application_choices: GetApplicationChoicesForProviders.call(providers: [p])).page(1).count
=> 25
```
Query takes: __15ms__

#### All providers (all time)
```
irb(main):013:0> GetActivityLogEvents.call(application_choices: GetApplicationChoicesForProviders.call(providers: Provider.all)).count
=> 3440619
```
Query takes: __6339ms__

#### All providers (all time, first page)
```
irb(main):013:0> GetActivityLogEvents.call(application_choices: GetApplicationChoicesForProviders.call(providers: Provider.all)).page(1).count
=> 25
```
Query takes: __234ms__

#### All providers (last 30 days)
```
irb(main):024:0> GetActivityLogEvents.call(application_choices: GetApplicationChoicesForProviders.call(providers: Provider.all), since: 30.days.ago).count
=> 1649734
```
Query takes: __3516ms__

#### All providers (last 5 days)
```
irb(main):023:0> GetActivityLogEvents.call(application_choices: GetApplicationChoicesForProviders.call(providers: Provider.all), since: 5.days.ago).count
=> 260251
```
Query takes: __2058ms__

PR #3497, opened 2020-11-24
Add unique index to application choices

raamSoftwire:
## Context

This index is required to allow us to add a new validation to the `application_choices` model. 

```
validates :course_option, uniqueness: { scope: :application_form_id }
```

This PR add the new index. [A separate PR](https://github.com/DFE-Digital/apply-for-teacher-training/pull/3493) adds the validation to the model. These two PRs should be released separately to ensure this migration is run before the code is deployed.

## Link to Trello card

https://trello.com/c/RfVH10aE/2491-dev-bug-users-can-apply-for-the-same-course-twice-when-using-the-back-button

## Things to check

### DIFF
--- a/spec/system/provider_interface/provider_applications_pagination_spec.rb
+++ b/spec/system/provider_interface/provider_applications_pagination_spec.rb
@@ -4,6 +4,8 @@ RSpec.feature 'Providers should be able to sort applications' do
   include CourseOptionHelpers
   include DfESignInHelpers
 
+  let(:provider) { create(:provider, :with_signed_agreement, code: 'ABC') }
+
   scenario 'viewing applications one page at a time' do
     given_i_am_a_provider_user_with_dfe_sign_in
     and_i_am_permitted_to_see_applications_for_my_provider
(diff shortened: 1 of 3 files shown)


stevehook (commented):
Migration looks good to me. Before merging/deploying we should check to find any duplicates on prod if not done already. They'll have to be resolved before deployment.

raamSoftwire:
> Migration looks good to me. Before merging/deploying we should check to find any duplicates on prod if not done already. They'll have to be resolved before deployment.

I've taken a look and resolved the 17 cases of duplicates on production.

spec/system/provider_interface/provider_applications_pagination_spec.rb
```diff
@@ -4,6 +4,8 @@
   include CourseOptionHelpers
   include DfESignInHelpers
 
+  let(:provider) { create(:provider, :with_signed_agreement, code: 'ABC') }
```
davidgisbey, line 7: can we remove this and add @provider = create(:provider, :with_signed_agreement, code: 'ABC') instead please

davidgisbey (approved):

PR #3498, opened 2020-11-24
Migration to add sites latitude & longitude

stevehook:
## Context

We want to add geographic location data to course sites. It's already available in the Find API. First step is this migration-only PR that makes the schema changes.

```
   bat_apply_development=# \d sites
                                            Table "public.sites"
   ┌───────────────┬────────────────────────────────┬───────────┬──────────┬───────────────────────────────────┐
   │    Column     │              Type              │ Collation │ Nullable │              Default              │
   ├───────────────┼────────────────────────────────┼───────────┼──────────┼───────────────────────────────────┤
   │ id            │ bigint                         │           │ not null │ nextval('sites_id_seq'::regclass) │
   │ code          │ character varying              │           │ not null │                                   │
   │ name          │ character varying              │           │ not null │                                   │
   │ provider_id   │ bigint                         │           │ not null │                                   │
   │ created_at    │ timestamp(6) without time zone │           │ not null │                                   │
   │ updated_at    │ timestamp(6) without time zone │           │ not null │                                   │
   │ address_line1 │ character varying              │           │          │                                   │
   │ address_line2 │ character varying              │           │          │                                   │
   │ address_line3 │ character varying              │           │          │                                   │
   │ address_line4 │ character varying              │           │          │                                   │
   │ postcode      │ character varying              │           │          │                                   │
-->│ latitude      │ double precision               │           │          │                                   │
-->│ longitude     │ double precision               │           │          │                                   │
   └───────────────┴────────────────────────────────┴───────────┴──────────┴───────────────────────────────────┘
```

## Guidance to review

- I've used the full names for these new columns rather than commonly used `lat`/`lng` abbreviations partly for consistency with the Find API.
- The new columns are nullable - we could in theory make them not null once they've been fully populated but initially there is no simple way to backfill them.

## Link to Trello card

https://trello.com/c/2uEjIOHm/2575-dev-use-the-v3-api-to-populate-the-latitude-and-longitude-for-sites

## Things to check


davidgisbey (approved):

db/schema.rb
```diff
@@ -405,8 +405,8 @@
   create_table "provider_users_providers", force: :cascade do |t|
     t.bigint "provider_id", null: false
     t.bigint "provider_user_id", null: false
-    t.datetime "created_at", default: -> { "now()" }, null: false
-    t.datetime "updated_at", default: -> { "now()" }, null: false
+    t.datetime "created_at", default: -> { "CURRENT_TIMESTAMP" }, null: false
+    t.datetime "updated_at", default: -> { "CURRENT_TIMESTAMP" }, null: false
```
malcolmbaig, line 409: Does this fluctuate because of version differences? Might be worth leaving this hunk out until we resolve this and standardise across our installs somehow.

malcolmbaig (approved):

db/schema.rb
stevehook: Yes, I think Postgres 9 uses `now`, later versions its `CURRENT_TIMESTAMP`. Will remove that hunk.

PR #3500, opened 2020-11-24
Add postcode to the providers table

davidgisbey:
## Context

I need providers' postcodes for a data export for Mike.

- Add a column for postcode to the providers' table

## Guidance to review

Unless there's a need for this column going forward I'll remove it after I've run the data export as it's a 1 timer.

## Link to Trello card

https://trello.com/c/yeZRKLYE/2530-dev-locations-persona-data

## Things to check


db/schema.rb
```diff
@@ -46,6 +46,7 @@
     t.datetime "offer_deferred_at"
     t.string "status_before_deferral"
     t.datetime "reject_by_default_feedback_sent_at"
+    t.index ["application_form_id", "course_option_id"], name: "index_course_option_to_application_form_id", unique: true
```
raamSoftwire, line 49: I don't think you want to be adding this change as part of this PR

raamSoftwire (changes requested):

db/schema.rb
davidgisbey: you're right sorry. Should've checked the schema 🤦

raamSoftwire (approved):

PR #3501, opened 2020-11-24
Add ability to run read queries on the database from support

tijmenb:
## Context

Analysists sometimes need data from the database for analysis that's not easy to get via one of the extracts.

Add the [Blazer gem](https://github.com/ankane/blazer/). 

It's secured via the same mechanism as Sidekiq web (so we don't have to trust Blazer's auth).

Blazer will wrap everything in a transaction, so you can't easily make UPDATE/DELETE queries on the database. But we're still looking to use [a read-only user](https://github.com/ankane/blazer/#postgresql) or read-replica for this.

## Guidance to review

I had to turn on Sprockets for this to work, is that going to cause problems?

## Link to Trello card

N/A


config/application.rb
```diff
@@ -11,6 +11,8 @@
 
 require 'view_component/engine'
 
+require 'sprockets/railtie'
```
duncanjbrown, line 14: CSS and JS still working on the review app. No notices on app boot. Nothing in our `assets` folder anyway. I think this is probably fine. 🤷

config/blazer.yml
```diff
@@ -0,0 +1,3 @@
+data_sources:
+  main:
+    url: <%= ENV["BLAZER_DATABASE_URL"] %>
```
duncanjbrown, line 3: Something in `.env.example` for this?

config/routes.rb
```diff
@@ -808,6 +808,9 @@
     mount Sidekiq::Web => '/sidekiq', constraints: SupportUserConstraint.new
     get '/sidekiq', to: redirect('/support/sign-in'), status: 302
 
+    mount Blazer::Engine => '/blazer', constraints: SupportUserConstraint.new
```
duncanjbrown, line 811: 👌

duncanjbrown (approved):
Should be interesting! Nit about `.env`

config/blazer.yml
tijmenb: Fixed it so that it'll work in development out of the box.

vigneshmsft:
Have merged #3581 which sets the `BLAZER_DATABASE_URL`

duncanjbrown:
Rebased @tijmenb

duncanjbrown:
It's good! It works!

Do we want to link from anywhere in Support?

PR #3082, opened 2020-10-07
Parse since param with Time#iso8601

steventux:
https://sentry.io/organizations/dfe-bat/issues/1939203076/?project=1765973&referrer=slack

## Context

Non iso-8601 dates in the 'since' parameter were not being caught by `Date.parse` and were being passed to ActiveRecord resulting in `PG::DatetimeFieldOverflow` as Postgresql is evidently configured with `datestyle='iso'`. 

## Changes proposed in this pull request

Ensure the `since` parameter is in iso-8601 format and responds with the appropriate JSON error if not.

## Link to Trello card

https://trello.com/c/FBINxp0u/2862-bug-pgdatetimefieldoverflow-sentry-error-on-sandbox

## Things to check

- [x] This code does not rely on migrations in the same Pull Request
- [x] If this code includes a migration adding or changing columns, it also backfills existing records for consistency
- [x] API release notes have been updated if necessary
- [x] New environment variables have been [added to the Azure config](https://github.com/DFE-Digital/apply-for-teacher-training#azure-hosting-devops-pipeline)

PR #3502, opened 2020-11-24
Add provider_type column to providers

davidgisbey:
#3082  Context

As part of the persona data work, Mike also needs to know if the provider/accrediting provider is a SCITT. We need to store this on the Provider object

- Add a type column to the providers table

## Link to Trello card

https://trello.com/c/yeZRKLYE/2530-dev-locations-persona-data

## Things to check


george-bit (approved):
Looks sound 👍

PR #3507, opened 2020-11-25
Add latitude and longitude to Providers

raamSoftwire:
## Context

We need to know where providers are so that we can calculate distances and range in search queries.

Add latitude and longitude, nullable, float columns to the providers table.

## Guidance to review

Follows the pattern from https://github.com/DFE-Digital/apply-for-teacher-training/pull/3498/files. NB. to other developers, there was a good discussion around why we choose to use a float type for these columns

## Link to Trello card

https://trello.com/c/YgubEbfX/2577-dev-use-the-v3-api-to-populate-the-latitude-and-longitude-for-providers

## Things to check


stevehook (approved):
:+1:

PR #3508, opened 2020-11-25
Review candidate routes

paulrobertlloyd:
## Context

Not only have we updated the ordering and labelling of items on a candidates application, but some of the paths we have refer to page titles we changed a long time ago, and no longer fully reflect their content or purpose.

This is a proposal for changes we make to user facing paths. Controller names and file locations have not been changed, but should be if these recommendations are accepted.

Comments inline, but some high level notes:

* Replaces any `_` in paths with the GDS recommendation `-`
* Uses `delete` in user-facing paths, not `destroy`
* Uses `/thing/edit` not `/edit-thing`
* Tries to logically and consistently group related paths
* Removes a few redirects – do we still need these?

## Guidance to review

Agree with these recommendations?

## Link to Trello card

https://trello.com/c/lVDr7xYM/

## Things to check

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -37,8 +37,8 @@ Rails.application.routes.draw do
     post '/sign-in/expired', to: 'sign_in#create_from_expired_token', as: :create_expired_sign_in
     get '/sign-in/check-email', to: 'sign_in#check_your_email', as: :check_email_sign_in
     get '/sign-in/expired', to: 'sign_in#expired', as: :expired_sign_in
-    get '/confirm_authentication', to: 'sign_in#confirm_authentication', as: :authenticate
-    post '/confirm_authentication', to: 'sign_in#authenticate'
+    get '/sign-in/confirm', to: 'sign_in#confirm_authentication', as: :authenticate
+    post '/sign-in/confirm', to: 'sign_in#authenticate'
     get '/authenticate', to: 'sign_in#expired'
 
     get '/apply', to: 'apply_from_find#show', as: :apply_from_find
@@ -68,16 +68,16 @@ Rails.application.routes.draw do
       get '/start-carry-over' => 'carry_over#start', as: :start_carry_over
       post '/carry-over' => 'carry_over#create', as: :carry_over
 
-      scope '/personal-details' do
+      scope '/personal-information' do
         get '/' => 'personal_details/base#new', as: :personal_details
         patch '/' => 'personal_details/base#create'
         get '/edit' => 'personal_details/base#edit', as: :edit_personal_details
         patch '/edit' => 'personal_details/base#update'
 
-        get '/nationalities' => 'personal_details/nationalities#new', as: :nationalities
-        patch '/nationalities' => 'personal_details/nationalities#create'
-        get '/nationalities/edit' => 'personal_details/nationalities#edit', as: :edit_nationalities
-        patch '/nationalities/edit' => 'personal_details/nationalities#update'
+        get '/nationality' => 'personal_details/nationalities#new', as: :nationalities
+        patch '/nationality' => 'personal_details/nationalities#create'
+        get '/nationality/edit' => 'personal_details/nationalities#edit', as: :edit_nationalities
+        patch '/nationality/edit' => 'personal_details/nationalities#update'
 
         get '/languages' => 'personal_details/languages#new', as: :languages
         patch '/languages' => 'personal_details/languages#create'
@@ -94,35 +94,39 @@ Rails.application.routes.draw do
       end
 
       scope '/personal-statement' do
-        get '/becoming-a-teacher' => 'personal_statement/becoming_a_teacher#edit', as: :edit_becoming_a_teacher
-        patch '/becoming-a-teacher' => 'personal_statement/becoming_a_teacher#update'
-        get '/becoming-a-teacher/review' => 'personal_statement/becoming_a_teacher#show', as: :becoming_a_teacher_show
-        patch '/becoming-a-teacher/complete' => 'personal_statement/becoming_a_teacher#complete', as: :becoming_a_teacher_complete
-
-        get '/subject-knowledge' => 'personal_statement/subject_knowledge#edit', as: :edit_subject_knowledge
-        patch '/subject-knowledge' => 'personal_statement/subject_knowledge#update'
-        get '/subject-knowledge/review' => 'personal_statement/subject_knowledge#show', as: :subject_knowledge_show
-        patch '/subject-knowledge/complete' => 'personal_statement/subject_knowledge#complete', as: :subject_knowledge_complete
-
-        get '/interview-preferences' => 'personal_statement/interview_preferences#edit', as: :edit_interview_preferences
-        patch '/interview-preferences' => 'personal_statement/interview_preferences#update'
-        get '/interview-preferences/review' => 'personal_statement/interview_preferences#show', as: :interview_preferences_show
-        patch '/interview-preferences/complete' => 'personal_statement/interview_preferences#complete', as: :interview_preferences_complete
+        get '/' => 'personal_statement/becoming_a_teacher#edit', as: :edit_becoming_a_teacher
+        patch '/' => 'personal_statement/becoming_a_teacher#update'
+        get '/review' => 'personal_statement/becoming_a_teacher#show', as: :becoming_a_teacher_show
+        patch '/complete' => 'personal_statement/becoming_a_teacher#complete', as: :becoming_a_teacher_complete
       end
 
-      scope '/training-with-a-disability' do
+      scope '/subject-knowledge' do
+        get '/' => 'personal_statement/subject_knowledge#edit', as: :edit_subject_knowledge
+        patch '/' => 'personal_statement/subject_knowledge#update'
+        get '/review' => 'personal_statement/subject_knowledge#show', as: :subject_knowledge_show
+        patch '/complete' => 'personal_statement/subject_knowledge#complete', as: :subject_knowledge_complete
+      end
+
+      scope '/interview-needs' do
+        get '/' => 'personal_statement/interview_preferences#edit', as: :edit_interview_preferences
+        patch '/' => 'personal_statement/interview_preferences#update'
+        get '/review' => 'personal_statement/interview_preferences#show', as: :interview_preferences_show
+        patch '/complete' => 'personal_statement/interview_preferences#complete', as: :interview_preferences_complete
+      end
+
+      scope '/additional-support' do
         get '/' => 'training_with_a_disability#edit', as: :edit_training_with_a_disability
         patch '/' => 'training_with_a_disability#update'
         get '/review' => 'training_with_a_disability#show', as: :training_with_a_disability_show
         patch '/complete' => 'training_with_a_disability#complete', as: :training_with_a_disability_complete
       end
 
-      scope '/contact-details' do
+      scope '/contact-information' do
         get '/' => 'contact_details/base#edit', as: :contact_details_edit_base
         patch '/' => 'contact_details/base#update'
 
-        get '/address_type' => 'contact_details/address_type#edit', as: :contact_details_edit_address_type
-        patch '/address_type' => 'contact_details/address_type#update'
+        get '/address-type' => 'contact_details/address_type#edit', as: :contact_details_edit_address_type
+        patch '/address-type' => 'contact_details/address_type#update'
 
         get '/address' => 'contact_details/address#edit', as: :contact_details_edit_address
         patch '/address' => 'contact_details/address#update'
@@ -189,7 +193,7 @@ Rails.application.routes.draw do
         delete '/delete/:id' => 'work_history/destroy#destroy'
       end
 
-      scope '/school-experience' do
+      scope '/unpaid-experience' do
         get '/' => 'volunteering/experience#show', as: :volunteering_experience
         post '/' => 'volunteering/experience#submit'
 
@@ -209,8 +213,8 @@ Rails.application.routes.draw do
       scope '/degrees' do
         get '/' => 'degrees/type#new', as: :new_degree
         post '/' => 'degrees/type#create'
-        get '/:id/type/edit' => 'degrees/type#edit', as: :edit_degree_type
-        patch '/:id/type/edit' => 'degrees/type#update'
+        get '/:id/edit' => 'degrees/type#edit', as: :edit_degree_type
+        patch '/:id/edit' => 'degrees/type#update'
 
         get '/:id/subject' => 'degrees/subject#new', as: :degree_subject
         post '/:id/subject' => 'degrees/subject#create'
@@ -222,10 +226,10 @@ Rails.application.routes.draw do
         get '/:id/institution/edit' => 'degrees/institution#edit', as: :edit_degree_institution
         patch '/:id/institution/edit' => 'degrees/institution#update'
 
-        get '/:id/completion_status' => 'degrees/completion_status#new', as: :degree_completion_status
-        post '/:id/completion_status' => 'degrees/completion_status#create'
-        get '/:id/completion_status/edit' => 'degrees/completion_status#edit', as: :edit_degree_completion_status
-        patch '/:id/completion_status/edit' => 'degrees/completion_status#update'
+        get '/:id/completion-status' => 'degrees/completion_status#new', as: :degree_completion_status
+        post '/:id/completion-status' => 'degrees/completion_status#create'
+        get '/:id/completion-status/edit' => 'degrees/completion_status#edit', as: :edit_degree_completion_status
+        patch '/:id/completion-status/edit' => 'degrees/completion_status#update'
 
         get '/:id/naric' => 'degrees/naric#new', as: :degree_naric
         post '/:id/naric' => 'degrees/naric#create'
@@ -255,7 +259,6 @@ Rails.application.routes.draw do
         get '/choose' => 'course_choices/have_you_chosen#ask', as: :course_choices_choose
         post '/choose' => 'course_choices/have_you_chosen#decide'
         get '/find-a-course' => 'course_choices/have_you_chosen#go_to_find', as: :go_to_find
-        get '/find_a_course', to: redirect('/candidate/application/courses/find-a-course')
 
         get '/provider' => 'course_choices/provider_selection#new', as: :course_choices_provider
         post '/provider' => 'course_choices/provider_selection#create'
@@ -274,9 +277,7 @@ Rails.application.routes.draw do
         get '/apply-on-ucas/provider/:provider_id/course/:course_id' => 'course_choices/ucas#with_course', as: :course_choices_ucas_with_course
 
         get '/confirm-selection/:course_id' => 'find_course_selections#confirm_selection', as: :course_confirm_selection
-        get '/confirm_selection/:course_id', to: redirect('/candidate/application/courses/confirm-selection/%{course_id}')
         post '/complete-selection/:course_id' => 'find_course_selections#complete_selection', as: :course_complete_selection
-        get '/complete_selection/:course_id', to: redirect('/candidate/application/courses/complete-selection/%{course_id}')
 
         get '/review' => 'application_choices#review', as: :course_choices_review
         patch '/review' => 'application_choices#complete', as: :course_choices_complete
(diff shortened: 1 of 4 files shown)


config/routes.rb
```diff
@@ -48,76 +47,85 @@
...
+      get '/' => 'unsubmitted_application_form#show', as: :application_form
+      get '/review' => 'unsubmitted_application_form#review', as: :application_review
       get '/submit' => 'unsubmitted_application_form#submit_show', as: :application_submit_show
       post '/submit' => 'unsubmitted_application_form#submit', as: :application_submit
 
+      get '/submit-success' => 'submitted_application_form#submit_success', as: :application_submit_success
       get '/complete' => 'submitted_application_form#complete', as: :application_complete
       get '/review/submitted' => 'submitted_application_form#review_submitted', as: :application_review_submitted
-      get '/submit-success' => 'submitted_application_form#submit_success', as: :application_submit_success
-      get '/edit' => 'submitted_application_form#edit', as: :application_edit
```
paulrobertlloyd, line 64: This functionality no longer exists

config/routes.rb
```diff
@@ -48,76 +47,85 @@
...
+      get '/review/submitted/:id' => 'application_form#review_previous_application', as: :review_previous_application
+
       get '/start-apply-again' => 'submitted_application_form#start_apply_again', as: :start_apply_again
       post '/apply-again' => 'submitted_application_form#apply_again', as: :apply_again
+
       get '/start-carry-over' => 'carry_over#start', as: :start_carry_over
       post '/carry-over' => 'carry_over#create', as: :carry_over
 
-      scope '/personal-details' do
+      scope '/personal-information' do
```
paulrobertlloyd, line 70: Changed to reflect new label for this section

config/routes.rb
```diff
@@ -48,76 +47,85 @@
...
-        get '/nationalities' => 'personal_details/nationalities#new', as: :nationalities
-        patch '/nationalities' => 'personal_details/nationalities#create'
-        get '/nationalities/edit' => 'personal_details/nationalities#edit', as: :edit_nationalities
-        patch '/nationalities/edit' => 'personal_details/nationalities#update'
+
+        get '/nationality' => 'personal_details/nationalities#new', as: :nationalities
+        patch '/nationality' => 'personal_details/nationalities#create'
+        get '/nationality/edit' => 'personal_details/nationalities#edit', as: :edit_nationalities
+        patch '/nationality/edit' => 'personal_details/nationalities#update'
+
```
paulrobertlloyd, line 80: The page title is ‘What is your nationality?’ and the review label is ‘nationality’.

config/routes.rb
```diff
@@ -48,76 +47,85 @@
...
+        get '/review' => 'personal_statement/subject_knowledge#show', as: :subject_knowledge_show
+        patch '/complete' => 'personal_statement/subject_knowledge#complete', as: :subject_knowledge_complete
+      end
+
+      scope '/interview-needs' do
+        get '/' => 'personal_statement/interview_preferences#edit', as: :edit_interview_preferences
+        patch '/' => 'personal_statement/interview_preferences#update'
+        get '/review' => 'personal_statement/interview_preferences#show', as: :interview_preferences_show
+        patch '/complete' => 'personal_statement/interview_preferences#complete', as: :interview_preferences_complete
+      end
```
paulrobertlloyd, line 114: Move `interview-preferences` from outside `personal-statement` - no other section has this sort of folder structure, and also this section no longer lives under the ‘Personal statement’ heading in the application.

Also, rename it `interview-needs` to reflect the section title.

config/routes.rb
```diff
@@ -48,76 +47,85 @@
...
+        patch '/complete' => 'personal_statement/becoming_a_teacher#complete', as: :becoming_a_teacher_complete
       end
 
-      scope '/training-with-a-disability' do
+      scope '/subject-knowledge' do
+        get '/' => 'personal_statement/subject_knowledge#edit', as: :edit_subject_knowledge
+        patch '/' => 'personal_statement/subject_knowledge#update'
+        get '/review' => 'personal_statement/subject_knowledge#show', as: :subject_knowledge_show
+        patch '/complete' => 'personal_statement/subject_knowledge#complete', as: :subject_knowledge_complete
+      end
```
paulrobertlloyd, line 107: Move `subject-knowledge` from outside `personal-statement` - no other section has this sort of folder structure.

config/routes.rb
```diff
@@ -48,76 +47,85 @@
...
-
-        get '/interview-preferences' => 'personal_statement/interview_preferences#edit', as: :edit_interview_preferences
-        patch '/interview-preferences' => 'personal_statement/interview_preferences#update'
-        get '/interview-preferences/review' => 'personal_statement/interview_preferences#show', as: :interview_preferences_show
-        patch '/interview-preferences/complete' => 'personal_statement/interview_preferences#complete', as: :interview_preferences_complete
+        get '/' => 'personal_statement/becoming_a_teacher#edit', as: :edit_becoming_a_teacher
+        patch '/' => 'personal_statement/becoming_a_teacher#update'
+        get '/review' => 'personal_statement/becoming_a_teacher#show', as: :becoming_a_teacher_show
+        patch '/complete' => 'personal_statement/becoming_a_teacher#complete', as: :becoming_a_teacher_complete
       end
```
paulrobertlloyd, line 100: Removing the sections previously nested below `personal-statement` means we can now use this for what was previously `/personal-statement/becoming-a-teacher`

config/routes.rb
```diff
@@ -48,76 +47,85 @@
...
+
+      scope '/additional-support' do
         get '/' => 'training_with_a_disability#edit', as: :edit_training_with_a_disability
         patch '/' => 'training_with_a_disability#update'
         get '/review' => 'training_with_a_disability#show', as: :training_with_a_disability_show
         patch '/complete' => 'training_with_a_disability#complete', as: :training_with_a_disability_complete
       end
 
-      scope '/contact-details' do
+      scope '/contact-information' do
```
paulrobertlloyd, line 123: Update to match new section title.

config/routes.rb
```diff
@@ -48,76 +47,85 @@
...
         get '/' => 'contact_details/base#edit', as: :contact_details_edit_base
         patch '/' => 'contact_details/base#update'
 
-        get '/address_type' => 'contact_details/address_type#edit', as: :contact_details_edit_address_type
-        patch '/address_type' => 'contact_details/address_type#update'
+        get '/address-type' => 'contact_details/address_type#edit', as: :contact_details_edit_address_type
+        patch '/address-type' => 'contact_details/address_type#update'
```
paulrobertlloyd, line 128: Use hyphens in paths, not underscores.

config/routes.rb
```diff
@@ -144,8 +152,8 @@
         get '/country' => 'gcse/institution_country#edit', as: :gcse_details_edit_institution_country
         patch '/country' => 'gcse/institution_country#update'
 
-        get '/naric-reference' => 'gcse/naric_reference#edit', as: :gcse_details_edit_naric_reference
-        patch '/naric-reference' => 'gcse/naric_reference#update'
+        get '/naric' => 'gcse/naric_reference#edit', as: :gcse_details_edit_naric_reference
+        patch '/naric' => 'gcse/naric_reference#update'
```
paulrobertlloyd, line 156: Use `naric` consistently in qualification sections.

config/routes.rb
```diff
@@ -184,7 +192,7 @@
         delete '/delete/:id' => 'work_history/destroy#destroy'
       end
 
-      scope '/school-experience' do
+      scope '/unpaid-experience' do
```
paulrobertlloyd, line 195: Update to reflect section title.

config/routes.rb
```diff
@@ -217,15 +225,15 @@
...
         patch '/:id/institution/edit' => 'degrees/institution#update'
 
-        get '/:id/completion_status' => 'degrees/completion_status#new', as: :degree_completion_status
-        post '/:id/completion_status' => 'degrees/completion_status#create'
-        get '/:id/completion_status/edit' => 'degrees/completion_status#edit', as: :edit_degree_completion_status
-        patch '/:id/completion_status/edit' => 'degrees/completion_status#update'
+        get '/:id/completion-status' => 'degrees/completion_status#new', as: :degree_completion_status
+        post '/:id/completion-status' => 'degrees/completion_status#create'
+        get '/:id/completion-status/edit' => 'degrees/completion_status#edit', as: :edit_degree_completion_status
+        patch '/:id/completion-status/edit' => 'degrees/completion_status#update'
```
paulrobertlloyd, line 231: Use hyphens, not underscores.

config/routes.rb
```diff
@@ -204,8 +212,8 @@
       scope '/degrees' do
         get '/' => 'degrees/type#new', as: :new_degree
         post '/' => 'degrees/type#create'
-        get '/:id/type/edit' => 'degrees/type#edit', as: :edit_degree_type
-        patch '/:id/type/edit' => 'degrees/type#update'
+        get '/:id/edit' => 'degrees/type#edit', as: :edit_degree_type
+        patch '/:id/edit' => 'degrees/type#update'
```
paulrobertlloyd, line 216: Be consistent in how we address editing the first question in a section.

config/routes.rb
```diff
@@ -217,15 +225,15 @@
...
+        patch '/:id/completion-status/edit' => 'degrees/completion_status#update'
 
-        get '/:id/naric-statement' => 'degrees/naric_statement#new', as: :degree_naric_statement
-        post '/:id/naric-statement' => 'degrees/naric_statement#create'
-        get '/:id/naric-statement/edit' => 'degrees/naric_statement#edit', as: :edit_degree_naric_statement
-        patch '/:id/naric-statement/edit' => 'degrees/naric_statement#update'
+        get '/:id/naric' => 'degrees/naric_statement#new', as: :degree_naric_statement
+        post '/:id/naric' => 'degrees/naric_statement#create'
+        get '/:id/naric/edit' => 'degrees/naric_statement#edit', as: :edit_degree_naric_statement
+        patch '/:id/naric/edit' => 'degrees/naric_statement#update'
```
paulrobertlloyd, line 236: Use `naric` consistently in qualification sections

config/routes.rb
```diff
@@ -254,7 +262,6 @@
         get '/choose' => 'course_choices/have_you_chosen#ask', as: :course_choices_choose
         post '/choose' => 'course_choices/have_you_chosen#decide'
         get '/find-a-course' => 'course_choices/have_you_chosen#go_to_find', as: :go_to_find
-        get '/find_a_course', to: redirect('/candidate/application/courses/find-a-course')
```
paulrobertlloyd, line 257: Do we still need this redirect?

config/routes.rb
```diff
@@ -272,9 +279,7 @@
         get '/apply-on-ucas/provider/:provider_id/course/:course_id' => 'course_choices/ucas#with_course', as: :course_choices_ucas_with_course
 
         get '/confirm-selection/:course_id' => 'find_course_selections#confirm_selection', as: :course_confirm_selection
-        get '/confirm_selection/:course_id', to: redirect('/candidate/application/courses/confirm-selection/%{course_id}')
         post '/complete-selection/:course_id' => 'find_course_selections#complete_selection', as: :course_complete_selection
-        get '/complete_selection/:course_id', to: redirect('/candidate/application/courses/complete-selection/%{course_id}')
```
paulrobertlloyd, line 277: Do we still need this 2 redirects?

config/routes.rb
```diff
@@ -317,11 +320,13 @@
       scope '/english-as-a-foreign-language' do
         get '/' => 'english_foreign_language/start#new', as: :english_foreign_language_start
         post '/' => 'english_foreign_language/start#create'
-        get '/proficiency/edit' => 'english_foreign_language/start#edit', as: :english_foreign_language_edit_start
-        patch '/proficiency/edit' => 'english_foreign_language/start#update'
+        get '/edit' => 'english_foreign_language/start#edit', as: :english_foreign_language_edit_start
+        patch '/edit' => 'english_foreign_language/start#update'
```
paulrobertlloyd, line 324: Be consistent in how we address editing the first question in a section.

config/routes.rb
```diff
@@ -317,11 +320,13 @@
...
 
         get '/type' => 'english_foreign_language/type#new', as: :english_foreign_language_type
         post '/type' => 'english_foreign_language/type#create'
+        # get '/type/edit' ?
+        # patch '/type/edit' ?
```
paulrobertlloyd, line 329: Are these 2 paths missing for a reason?

config/routes.rb
```diff
@@ -297,15 +302,13 @@
...
-
-        get '/new' => 'other_qualifications/details#new', as: :other_qualification_details
-        patch '/new' => 'other_qualifications/details#create'
-
-        get '/edit-details/:id' => 'other_qualifications/details#edit', as: :edit_other_qualification_details
-        patch '/edit-details/:id' => 'other_qualifications/details#update'
+        get '/details' => 'other_qualifications/details#new', as: :other_qualification_details
+        patch '/details' => 'other_qualifications/details#create'
+        get '/details/edit/:id' => 'other_qualifications/details#edit', as: :edit_other_qualification_details
+        patch '/details/edit/:id' => 'other_qualifications/details#update'
```
paulrobertlloyd, line 311: Use `/thing/edit` not `/edit-thing`

config/routes.rb
```diff
@@ -347,66 +352,68 @@
...
 
         get '/review' => 'references/review#show', as: :references_review
-        get '/review/destroy-referee/:id' => 'references/review#confirm_destroy_referee', as: :destroy_referee
-        get '/review/destroy-reference/:id' => 'references/review#confirm_destroy_reference', as: :destroy_reference
-        get '/review/destroy-reference-request/:id' => 'references/review#confirm_destroy_reference_request', as: :delete_reference_request
-        delete '/review/destroy-reference/:id' => 'references/review#destroy'
+        get '/review/delete-referee/:id' => 'references/review#confirm_destroy_referee', as: :destroy_referee
+        get '/review/delete-reference/:id' => 'references/review#confirm_destroy_reference', as: :destroy_reference
+        get '/review/delete-request/:id' => 'references/review#confirm_destroy_reference_request', as: :delete_reference_request
+        delete '/review/delete-reference/:id' => 'references/review#destroy'
```
paulrobertlloyd, line 380: Use `delete` in user facing paths, not `destroy`

config/routes.rb
```diff
@@ -347,66 +352,68 @@
...
         get 'review/cancel/:id' => 'references/review#confirm_cancel', as: :confirm_cancel_reference
         patch 'review/cancel/:id' => 'references/review#cancel', as: :cancel_reference
 
         get '/request/:id' => 'references/request#new', as: :references_new_request
         post '/request/:id' => 'references/request#create', as: :references_create_request
 
-        get '/retry_request/:id' => 'references/retry_request#new', as: :references_retry_request
-        post '/retry_request/:id' => 'references/retry_request#create'
+        get '/retry-request/:id' => 'references/retry_request#new', as: :references_retry_request
+        post '/retry-request/:id' => 'references/retry_request#create'
```
paulrobertlloyd, line 389: Use hyphens, not underscores.

config/routes.rb
```diff
@@ -347,66 +352,68 @@
...
 
         get '/reminder/:id' => 'references/reminder#new', as: :references_new_reminder
         post '/reminder/:id' => 'references/reminder#create'
 
-        get '/candidate_name/:id' => 'references/candidate_name#new', as: :references_new_candidate_name
-        post '/candidate_name/:id' => 'references/candidate_name#create', as: :references_create_candidate_name
+        get '/candidate-name/:id' => 'references/candidate_name#new', as: :references_new_candidate_name
+        post '/candidate-name/:id' => 'references/candidate_name#create', as: :references_create_candidate_name
```
paulrobertlloyd, line 395: Use hyphens, not underscores.

paulrobertlloyd (commented):
Comments inline about the various changes/recommendations.

paulrobertlloyd:
Testing this locally, links in the application menu no longer work, even though they should change based if using path helpers. Are the paths hard coded somewhere @TobyRet?

paulrobertlloyd:
If we make these changes, question is do we add redirects, and if so, to what extent? Maybe it’s safe to assume that these URLs are not bookmarked or saved somewhere (although browsers will have them in history).

@frankieroberto made the point that this could cause issues if the URLs change tries to save an answer as this change is delayed – though maybe that’d still be the case even with redirects?

Maybe add redirects, but only maintain them for a limited period?

config/routes.rb
```diff
@@ -297,15 +302,13 @@
       scope '/other-qualifications' do
         get '/' => 'other_qualifications/type#new', as: :other_qualification_type
         post '/' => 'other_qualifications/type#create'
+        get '/edit/:id' => 'other_qualifications/type#edit', as: :edit_other_qualification_type
+        patch '/edit/:id' => 'other_qualifications/type#update'
```
frankieroberto, line 306: I think `/edit` should always be the last bit of the URL?

```suggestion
        get '/:id/edit' => 'other_qualifications/type#edit', as: :edit_other_qualification_type
        patch '/:id/edit' => 'other_qualifications/type#update'
```

frankieroberto (commented):
All makes sense to me.

Agree we should avoid nesting URLs where possible - particularly as then it makes it easier to group and re-group sections without affecting the URLs.

I think we should use `/thing/edit` rather than `/edit/thing` – mainly to follow the convention that if you remove bits of the URL from the right up to the previous slash, then you get to an actual page rather than a 404.

For this reason, I have in the past sometimes used the `/thing` route for the "review" page, rather than `/thing/review` – but I think in our service most of those `/thing` pages are used for routing or guard questions?

For deletion confirmation pages, perhaps these should be at the end of the URL too - ie `/thing/delete` rather than `/delete-thing` or `/delete/thing`?

Personally, I prefer `_` to `-` in URLs, as I think it can make them easier to read, the underscores more closely resembling spaces. Plus then hyphens can be used for hyphenated words only, so that `/man-eating-shark` can be disambiguated into `/man-eating_shark`. 😄  But I'll accept this is a minority view, and underscores can be harder for people to type on a keyboard, so I'll let this go 🙊

paulrobertlloyd:
> Personally, I prefer _ to - in URLs, as I think it can make them easier to read, the underscores more closely resembling spaces.

I personally use underscores too - mainly as a small form of protest at Google who popularised their usage ✊🏻 But the GDS recommendation is hyphens, and we use hyphens everywhere else, so we should stick with that.

config/routes.rb
TobyRet: mmm interesting - the controller does not appear to have an edit or update action. 
https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/controllers/candidate_interface/english_foreign_language/type_controller.rb

@davidgisbey or @malcolmbaig can you shed some light on this?

TobyRet:
> Testing this locally, links in the application menu no longer work, even though they should have updated if using path helpers. Are the paths hard coded somewhere @TobyRet?

@paulrobertlloyd links in the application menu work for me. Did you restart your server?

paulrobertlloyd:
> Did you restart your server?

Ah, that’d be it! 🤦🏼‍♂️

config/routes.rb
paulrobertlloyd: So always `/[section-name]/:id/edit` or `/[section-name]/:id/edit/[question]`? That would effect all sections in the candidate interface, which might be a consideration if we were to change this.

frankieroberto: @paulrobertlloyd If it's one-page per question, I think you could do `/[section-name]/:id/[question]` – no need for the word "edit" at all?

config/routes.rb
```diff
@@ -68,16 +68,16 @@
...
         patch '/edit' => 'personal_details/base#update'
 
-        get '/nationalities' => 'personal_details/nationalities#new', as: :nationalities
-        patch '/nationalities' => 'personal_details/nationalities#create'
-        get '/nationalities/edit' => 'personal_details/nationalities#edit', as: :edit_nationalities
-        patch '/nationalities/edit' => 'personal_details/nationalities#update'
+        get '/nationality' => 'personal_details/nationalities#new', as: :nationalities
+        patch '/nationality' => 'personal_details/nationalities#create'
+        get '/nationality/edit' => 'personal_details/nationalities#edit', as: :edit_nationalities
+        patch '/nationality/edit' => 'personal_details/nationalities#update'
```
paulrobertlloyd, line 80: Use `nationality` as this is the word we use in the question title and on the review page.

config/routes.rb
malcolmbaig: The TypeController doesn't create a resource, it just handles POSTs from the type form and decides which EFL controller to send you to next. Therefore, there isn't really a need for edit/update actions.

When a candidate chooses to "edit" their type, nothing actually happens to their saved record until they submit the subsequent EFL qualification form.

PR #3510, opened 2020-11-25
Add geocoding of candidate addresses

malcolmbaig:
## Context

We want to analyse distances between candidates and the sites they choose to train at to improve location-related functionality in the service.

This change deals with geolocating candidate addresses.

- Set up `geocoder` gem
- Add a callback to application forms that invokes a geocoding worker when certain conditions are met
- Implement a worker that invokes the geocoder and persists the returned coordinates.

## Guidance to review
- Can be tested locally. Ensure Sidekiq is running and add/modify the address on an application form. 
- A valid UK address should be successfully geolocated. Inspect the form in the console - if a valid address was provided, latitude/longitude should be filled in.
- International addresses should not be geolocated (ie - blank lat/long).
- Changing a UK address to international address should clear the coordinates.
- Coordinates can be double checked by searching https://www.google.com/maps with a query in the form "(latitude),(longitude)"

## Link to Trello card
https://trello.com/c/6A8t8vVY

## Things to check

### DIFF
--- a/app/models/application_form.rb
+++ b/app/models/application_form.rb
@@ -333,7 +333,8 @@ private
       saved_change_to_address_line3? ||
       saved_change_to_address_line4? ||
       saved_change_to_country? ||
-      saved_change_to_postcode?
+      saved_change_to_postcode? ||
+      saved_change_to_address_type?
   end
 
   def withdrawn_course_choices
--- a/app/models/application_form.rb
+++ b/app/models/application_form.rb
@@ -3,6 +3,7 @@
 class ApplicationForm < ApplicationRecord
   audited
   has_associated_audits
+  geocoded_by :address_formatted_for_geocoding
 
   include Chased
 
@@ -70,13 +71,9 @@ class ApplicationForm < ApplicationRecord
 
   attribute :recruitment_cycle_year, :integer, default: -> { RecruitmentCycle.current_year }
 
-  before_create lambda {
-    self.support_reference ||= GenerateSupportRef.call
-  }
-
-  after_save lambda {
-    application_choices.update_all(updated_at: Time.zone.now)
-  }
+  before_create -> { self.support_reference ||= GenerateSupportRef.call }
+  after_save -> { application_choices.update_all(updated_at: Time.zone.now) }
+  after_commit :geocode_address_if_required
 
   def submitted?
     submitted_at.present?
@@ -314,7 +311,30 @@ class ApplicationForm < ApplicationRecord
     application_references.feedback_provided.count >= MINIMUM_COMPLETE_REFERENCES
   end
 
+  def address_formatted_for_geocoding
+    full_address.compact.join(', ')
+  end
+
 private
+
+  def geocode_address_if_required
+    return unless address_changed?
+
+    if international?
+      update!(latitude: nil, longitude: nil)
+    else
+      GeocodeApplicationAddressWorker.perform_async(id)
+    end
+  end
+
+  def address_changed?
+    saved_change_to_address_line1? ||
+      saved_change_to_address_line2? ||
+      saved_change_to_address_line3? ||
+      saved_change_to_address_line4? ||
+      saved_change_to_country? ||
+      saved_change_to_postcode?
+  end
 
   def withdrawn_course_choices
     application_choices.includes(%i[provider course]).select { |choice| choice.course.withdrawn == true }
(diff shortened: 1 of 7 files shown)


app/models/application_form.rb
```diff
@@ -314,8 +311,31 @@ def enough_references_have_been_provided?
...
+
 private
 
+  def geocode_address_if_required
+    return unless address_changed?
+
+    if international?
+      update!(latitude: nil, longitude: nil)
+    else
+      GeocodeApplicationAddressWorker.perform_async(id)
```
davidgisbey, line 326: Out of interest, why not just pass the application_form in?

malcolmbaig: Sidekiq [recommends](https://github.com/mperham/sidekiq/wiki/Best-Practices#1-make-your-job-parameters-small-and-simple) keeping job parameters simple. The docs encourage using simple identifiers rather than saving state.

malcolmbaig:
I've kept testing fairly light here given that candidate-facing functionality isn't dependent on the coordinates.

app/models/application_form.rb
davidgisbey: Okay thanks for the link. I'm just going to test it out locally, but the code looks good 👍

stevehook (approved):
Tested locally and triggered the call to the new worker (stopped there as I don't have an API key). So looks good. One minor query inline.

app/workers/geocode_application_address_worker.rb
```diff
@@ -0,0 +1,12 @@
+class GeocodeApplicationAddressWorker
+  include Sidekiq::Worker
+
+  def perform(application_form_id)
+    return unless Geocoder.config.api_key
+
+    application_form = ApplicationForm.find(application_form_id)
+    application_form.latitude, application_form.longitude = application_form.geocode
```
TobyRet, line 8: If the quality of the address data varies it may be wise to restrict searches to GB (or UK?) otherwise you may end up getting coordinates for other countries

app/workers/geocode_application_address_worker.rb
```diff
@@ -0,0 +1,12 @@
+class GeocodeApplicationAddressWorker
+  include Sidekiq::Worker
```
TobyRet, line 2: Should we define a new queue for geolocation?

TobyRet:
```
 queue_as :geocoding
```

TobyRet (commented):
Looking good - couple of suggestions / comments

app/workers/geocode_application_address_worker.rb
TobyRet: Actually forget that - I guess you may have candidates that are living in other countries right?

davidgisbey: I don't think we can use this for international addresses though. They don't provide their address in a structured format.

image

davidgisbey (approved):
I've tested it locally and it's all working nicely.

app/models/application_form.rb
```diff
@@ -314,8 +311,31 @@ def enough_references_have_been_provided?
...
+    end
+  end
+
+  def address_changed?
+    saved_change_to_address_line1? ||
+      saved_change_to_address_line2? ||
+      saved_change_to_address_line3? ||
+      saved_change_to_address_line4? ||
+      saved_change_to_country? ||
+      saved_change_to_postcode?
```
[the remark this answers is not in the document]
malcolmbaig: Good idea, decouples this code from the understanding that country/address_type are updated in lockstep. Will update.

app/workers/geocode_application_address_worker.rb
malcolmbaig: Good idea. Any objection to just using the existing `low_priority` queue? https://github.com/DFE-Digital/apply-for-teacher-training/blob/8b72523670e4e9f707f40d06d183a83ea5a0997d/config/sidekiq.yml#L5

app/workers/geocode_application_address_worker.rb
malcolmbaig: The worker doesn't get invoked for international addresses, and UK addresses don't surface country as an editable field (it defaults to GB). I think that's probably sufficient. WDYT?

app/workers/geocode_application_address_worker.rb
malcolmbaig: Update to use low_priority

app/workers/geocode_application_address_worker.rb
TobyRet: 👍

TobyRet (approved):

PR #3503, opened 2020-11-24
Populate latitude and longitude on Site sync from Find

stevehook:
## Context

We want to add geographic location data to course sites. Following on from https://github.com/DFE-Digital/apply-for-teacher-training/pull/3498 this PR adapts the code that sync sites to include latitude and longitude.

## Changes proposed in this pull request

- Extend `SyncCoursesFromFind`

## Guidance to review

- Do we need to extend our own API? Assuming not.
- The process of extending `FindAPIHelper` is a bit manual - would it make sense to use the VCR gem for this?

## Link to Trello card

https://trello.com/c/2uEjIOHm/2575-dev-use-the-v3-api-to-populate-the-latitude-and-longitude-for-sites

## Things to check

- [x] This code does not rely on migrations in the same Pull Request
- [x] If this code includes a migration adding or changing columns, it also backfills existing records for consistency
- [x] API release notes have been updated if necessary
- [x] New environment variables have been [added to the Azure config](https://github.com/DFE-Digital/apply-for-teacher-training#azure-hosting-devops-pipeline)

PR #3511, opened 2020-11-25
Populate latitude and longitude on providers when syncing from Find

raamSoftwire:
## Context

We want to add geographic location data to course providers. Following on from https://github.com/DFE-Digital/apply-for-teacher-training/pull/3507 this PR adapts the code that sync providers to include latitude and longitude.

- Extend `SyncProviderFromFind.`

## Guidance to review

This is very similar to https://github.com/DFE-Digital/apply-for-teacher-training/pull/3503

## Link to Trello card

https://trello.com/c/YgubEbfX/2577-dev-use-the-v3-api-to-populate-the-latitude-and-longitude-for-providers

## Things to check

### DIFF
--- a/spec/support/test_helpers/find_api_helper.rb
+++ b/spec/support/test_helpers/find_api_helper.rb
@@ -24,7 +24,9 @@ module FindAPIHelper
     funding_type: 'fee',
     age_range_in_years: '4 to 8',
     vac_status: 'full_time_vacancies',
-    content_status: 'published'
+    content_status: 'published',
+    latitude: '51.498024',
+    longitude: '0.129919'
   )
     stub_find_api_provider(provider_code)
       .to_return(
@@ -39,6 +41,8 @@ module FindAPIHelper
               'provider_code': provider_code,
               'region_code': region_code,
               'postcode': postcode,
+              'latitude': latitude,
+              'longitude': longitude,
             },
             'relationships': {
               'sites': {
@@ -158,7 +162,9 @@ module FindAPIHelper
     course_length: 'OneYear',
     region_code: 'north_west',
     age_range_in_years: '4 to 8',
-    content_status: 'published'
+    content_status: 'published',
+    latitude: '51.498024',
+    longitude: '0.129919'
   )
     stub_find_api_provider(provider_code)
       .to_return(
@@ -173,6 +179,8 @@ module FindAPIHelper
               'provider_code': provider_code,
               'region_code': region_code,
               'postcode': postcode,
+              'latitude': latitude,
+              'longitude': longitude,
             },
             'relationships': {
               'sites': {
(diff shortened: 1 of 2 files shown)


spec/support/test_helpers/find_api_helper.rb
```diff
@@ -22,7 +22,9 @@ def stub_find_api_provider_200(
     funding_type: 'fee',
     age_range_in_years: '4 to 8',
     vac_status: 'full_time_vacancies',
-    content_status: 'published'
+    content_status: 'published',
+    latitude: '51.498024',
+    longitude: '0.129919'
```
stevehook, line 27:
```suggestion
    longitude: '0.129919',
```

stevehook (commented):
Changes look good.

I tried locally and I got the follow error:

```
Syncing data from Find...
  SiteSetting Load (0.4ms)  SELECT "site_settings".* FROM "site_settings" WHERE "site_settings"."name" = $1 LIMIT $2  [["name", "cycle_schedule"], ["L
IMIT", 1]]
  ↳ app/models/site_setting.rb:7:in `cycle_schedule'
{:api_request_url=>"GET https://api.publish-teacher-training-courses.service.gov.uk/api/v3/recruitment_cycles/2020/providers/1JA?include=sites%2Ccours
es.sites%2Ccourses.subjects%2Ccourses.site_statuses.site", :response_status=>200, :runtime_ms=>610}
undefined method `latitude' for #<FindAPI::Provider:0x00007f8a45e9d960> excluded from capture: DSN not set
rails aborted!
NoMethodError: undefined method `latitude' for #<FindAPI::Provider:0x00007f8a45e9d960>
```

Is this because the FindAPI hasn't been updated / redeployed with the lat/lng attributes yet (@davidgisbey might know) or maybe the request we are making needs to specify those extra attributes?

spec/support/test_helpers/find_api_helper.rb
raamSoftwire: This is within a method call rather than JSON and so it doesn't accept the trailing comma.

raamSoftwire:
We should wait until the API is fixed (ie. it includes the latitude and longitude methods) before merging these changes.

It would stop the background sync working. It runs every 15 minutes to keep our provider, course and site data up to date.

duncanjbrown:
@raamSoftwire would you consider taking this from the new API, which already provides it? Building on #3513

raamSoftwire:
> @raamSoftwire would you consider taking this from the new API, which already provides it? Building on #3513

That's definitely something we're thinking about at the minute. But having spoken to @davidgisbey, we'll tackle that in a separate PR

davidgisbey (approved):
I've just added the try method to the job so that it doesn't break if there isn't a provider lat/long.

We can merge now.

We should def also add this to the public API job!

PR #3512, opened 2020-11-25
Remove banner style dependency from incomplete section component

paulrobertlloyd:
## Context

A [new banner component](https://design-system.service.gov.uk/components/notification-banner/) will surely be making its way to this codebase shortly, and with it our custom `.app-banner` styles will disappear. However, the incomplete messages we show on the application review page have a small dependency on these styles.

* Remove `.app-banner--missing-section` modifier, and instead add modifiers to `.govuk-inset-text`
* Rename the component from `SectionMissingBannerComponent` to `IncompleteSectionComponent`
* Be consistent with using `incomplete` keyword instead of using this as well as `missing`.

## Things to check


app/components/candidate_interface/incomplete_section_component.html.erb
```diff
@@ -0,0 +1,4 @@
+<div class="govuk-inset-text app-inset-text--<% if @error %>error<% else %>important<% end %>" id="<%= "incomplete-#{section}-error" %>">
```
frankieroberto, line 1: Minor possible tweak. More readable?

```suggestion
<div class="govuk-inset-text app-inset-text--<%= @error ? "error" : "important" %>" id="incomplete-<%= section %>-error">
```

frankieroberto (approved):

PR #3513, opened 2020-11-25
First integration with Teacher training API

duncanjbrown:
## Context

We need some data out of this API, and we also want to move away from the old v3 API in due course.

### Introduce a new parallel sync process

The data we want (`qualifications` and `program_type`) is on Courses in the new API. Put in a bare bones sync process for Providers, which we expect to add nothing new, and for each provider the API knows about, fetch its courses and update the corresponding courses on our side.

We can't move all the way to the new API because there isn't enough data about vacancies yet. We can build on the structure on this PR when it's time to go further.

Sync hourly instead of every 15 minutes.

### Namespace things more consistently

We're moving towards one namespace, `TeacherTrainingAPI`, instead of both `FindSync` and `FindAPI`. We can gradually remove references to the "Find API" which don't make sense any more.

### Deal with some pagination weirdness

The new API doesn't provide any information about available pages so we have to request and request until we get an error. I set the per_page count for courses to 500 - the most any provider has is [163](https://docs.google.com/spreadsheets/d/1aOVOQSHFz6m1MXgLm8QStAXdIrroWX-z/edit#gid=168081996).

### Hopefully improve the stubbing tools for the API

The old `FindAPIHelper` had weird global state (eg `stubbed_recruitment_cycle_year`) and method names `stub_find_provider_200` which wanted to be parameters on generic methods. It also relied on verbatim Ruby hashes instead of sample data from the API. Try to simplify this thing by relying on flat JSON files copied from the documentation, and provide (hopefully) reasonable interfaces for tailoring responses.

## Guidance to review

Run `TeacherTrainingAPI::SyncAllProvidersAndCourses` locally and see all your courses updated with the new fields. And read the code of course 🙂 .

## Link to Trello card

https://trello.com/c/Ywwwok6y/3059-move-to-public-publish-api

## Things to check

### DIFF
--- a/app/lib/api_request_logging_formatter.rb
+++ b/app/lib/api_request_logging_formatter.rb
@@ -1,0 +1,16 @@
+require 'faraday/logging/formatter'
+
+class APIRequestLoggingFormatter < Faraday::Logging::Formatter
+  # Don't do logging when the request starts
+  def request(env); end
+
+  def response(env)
+    info('Response') do
+      {
+        api_request_url: "#{env.method.upcase} #{env.url}",
+        response_status: env.status,
+        runtime_ms: (env.response_headers['x-runtime'].to_f * 1000).to_i,
+      }
+    end
+  end
+end
--- a/app/models/teacher_training_api/course.rb
+++ b/app/models/teacher_training_api/course.rb
@@ -1,28 +1,10 @@
-module FindAPI
-  class Course < FindAPI::Resource
+module TeacherTrainingAPI
+  class Course < TeacherTrainingAPI::Resource
     belongs_to :recruitment_cycle, through: :provider, param: :year
     belongs_to :provider, param: :provider_code
 
-    def provider
-      Provider.new(code: provider_code)
-    end
-
-    def code
-      course_code
-    end
-
-    def name_and_code
-      "#{name} (#{code})"
-    end
-
-    def subject_codes
-      subjects.to_a.map do |subject|
-        subject.try(:attributes).try(:[], 'subject_code')
-      end
-    end
-
     def self.fetch(provider_code, course_code)
-      where(year: RecruitmentCycle.current_year)
+      where(recruitment_cycle_year: RecruitmentCycle.current_year)
         .where(provider_code: provider_code)
         .find(course_code)
         .first
--- a/app/services/teacher_training_public_api/sync_all_providers_and_courses.rb
+++ b/app/services/teacher_training_public_api/sync_all_providers_and_courses.rb
@@ -1,24 +1,24 @@
-module TeacherTrainingAPI
+module TeacherTrainingPublicAPI
   class SyncAllProvidersAndCourses
     def self.call
       begin
         (1..).each do |page|
           sync_providers(
-            TeacherTrainingAPI::Provider.where(year: 2021).paginate(page: page, per_page: 500).all,
+            TeacherTrainingPublicAPI::Provider.where(year: 2021).paginate(page: page, per_page: 500).all,
           )
         end
       rescue JsonApiClient::Errors::ClientError
         # This is how the API responds when we run out of pages :/
       end
 
-      TeacherTrainingAPI::SyncCheck.set_last_sync(Time.zone.now)
+      TeacherTrainingPublicAPI::SyncCheck.set_last_sync(Time.zone.now)
     rescue JsonApiClient::Errors::ApiError
-      raise TeacherTrainingAPI::SyncError
+      raise TeacherTrainingPublicAPI::SyncError
     end
 
     def self.sync_providers(providers_from_api)
       providers_from_api.each do |provider_from_api|
-        TeacherTrainingAPI::SyncProvider.new(
+        TeacherTrainingPublicAPI::SyncProvider.new(
           provider_from_api: provider_from_api,
           recruitment_cycle_year: 2021,
         ).call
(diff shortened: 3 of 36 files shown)


app/models/teacher_training_api/course.rb
```diff
@@ -0,0 +1,18 @@
+module TeacherTrainingAPI
```
TobyRet, line 1: I wonder if it might pay to be more specific with the module name here (and any other references to the public API throughout the codebase)? 
i.e. `TeacherTrainingPublicAPI` 

Find is still talking to the old V3 API so might prevent a few brain somersaults? 
Publish talks to V2 and UCAS, V1

TobyRet (changes requested):
Looks good. One suggestion about naming in order to avoid confusion when referring to the numerous Teacher Training APIs that are currently available. 

Also, worth adding something to the README?

app/models/teacher_training_api/course.rb
duncanjbrown: Ha! Yes. I'll rename for maximum explicitness. Everything related to this API will be in this namespace so hopefully that'll give us a fairly strong guarantee we won't mix things up.

duncanjbrown:
Should be ready for another look @TobyRet

app/lib/api_request_logging_formatter.rb
```diff
@@ -0,0 +1,16 @@
+require 'faraday/logging/formatter'
+
+class APIRequestLoggingFormatter < Faraday::Logging::Formatter
```
mnacos, line 3: :+1: for moving this

app/models/find_api/course.rb
```diff
@@ -22,7 +22,7 @@ def subject_codes
     end
 
     def self.fetch(provider_code, course_code)
-      where(recruitment_cycle_year: RecruitmentCycle.current_year)
+      where(year: RecruitmentCycle.current_year)
```
mnacos, line 25: I'm a little confused about `year`... Could you explain a bit more where the year information comes from in the existing and the new public APIs?

app/services/teacher_training_public_api/sync_all_providers_and_courses.rb
```diff
@@ -0,0 +1,30 @@
+module TeacherTrainingPublicAPI
+  class SyncAllProvidersAndCourses
+    def self.call
+      begin
+        (1..).each do |page|
+          sync_providers(
+            TeacherTrainingPublicAPI::Provider.where(year: 2021).paginate(page: page, per_page: 500).all,
```
mnacos, line 7: I saw this here (and in Course below) and raised one eyebrow, then both, then actually read your PR description...

mnacos (approved):
It feels like there's a bit of code duplication going on and in an ideal world I would prefer the new sync code to live closer to and run at the same time as the old sync, but I get the impression we're not ready to commit to this transition yet and until then it makes sense to keep things a bit separate.

Otherwise, the code works well (tested locally), I can see 225 courses with `program_type` out of 450 for my provider setup.

app/services/teacher_training_public_api/sync_all_providers_and_courses.rb
```diff
@@ -0,0 +1,30 @@
...
+      TeacherTrainingPublicAPI::SyncCheck.set_last_sync(Time.zone.now)
+    rescue JsonApiClient::Errors::ApiError
+      raise TeacherTrainingPublicAPI::SyncError
+    end
+
+    def self.sync_providers(providers_from_api)
+      providers_from_api.each do |provider_from_api|
+        TeacherTrainingPublicAPI::SyncProvider.new(
+          provider_from_api: provider_from_api,
+          recruitment_cycle_year: 2021,
```
benswannack, line 23: Can we set this to `RecruitmentCycle.current_year`?

app/services/teacher_training_public_api/sync_provider.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    def create_or_update_provider(attrs)
+      # Prefer this to find_or_create_by as it results in 3x fewer audits
+      if existing_provider
+        existing_provider.update!(attrs)
+      else
+        new_provider = ::Provider.new(attrs.merge(code: @provider_from_api.code)).save!
+      end
+
+      existing_provider || new_provider
```
benswannack, line 48: Minor style question: is this better than explicitly `return`ing from inside the if block?

app/models/find_api/course.rb
duncanjbrown: `year` is a parameter name for the JSON API, eg https://api.publish-teacher-training-courses.service.gov.uk/api-reference.html#recruitment_cycles-year-providers-provider_code-courses-get

It's part of the URL — eg in the above example `/recruitment_cycles/{year}/providers/{provider_code}/courses`

app/services/teacher_training_public_api/sync_all_providers_and_courses.rb
duncanjbrown: Yes, why not

app/services/teacher_training_public_api/sync_provider.rb
duncanjbrown: No — copied blindly from `FindSync`, I'll shorten

duncanjbrown:
Thanks for reviews all

> It feels like there's a bit of code duplication going on and in an ideal world I would prefer the new sync code to live closer to and run at the same time as the old sync

I tried this approach, but it didn't feel right — I ended up preferring this approach because we can lift and shift from the old to the new until there's nothing left in `FindSync`, and we can refactor and improve the code as we move it. Also we can easily turn off the new sync if it goes wrong.

duncanjbrown:
Updated per @benswannack's suggestions

PR #3514, opened 2020-11-25
Detect and warn about database-level enum changes (requires Ruby 2.7.2)

mnacos:
## Context

When an enum definition in a model changes, any pre-existing enum values in the database not part of the new definition are not understood and are read as `nil` post-deploy. In these cases, it's important to include data migrations for these values. This PR adds a github action to remind us of this if any enums have changed.

image

## Guidance to review

This PR also upgrades Ruby from 2.7.1 to 2.7.2 because github actions no longer work with 2.7.1.

## Link to Trello card

No card, dev retro action

## Things to check


.github/workflows/enums.yml
```diff
@@ -0,0 +1,103 @@
...
+      - name: Capture changed enums
+        id: changed-enums
+        run: |
+          MULTILINE=$(cat changed_enums.log)
+          MULTILINE="${MULTILINE//'%'/'%25'}"
+          MULTILINE="${MULTILINE//$'\n'/'%0A'}"
+          MULTILINE="${MULTILINE//$'\r'/'%0D'}"
+          echo $MULTILINE
+          echo "::set-output name=data::$MULTILINE"
+        shell: bash
```
vigneshmsft, line 88: not required to specify `shell:` default will be bash on ubuntu

.github/workflows/enums.yml
```diff
@@ -0,0 +1,103 @@
+name: Custom checks
```
vigneshmsft, line 1:
```suggestion
name: Check For Enum Changes
```
or a similar name?

.github/workflows/enums.yml
```diff
@@ -0,0 +1,103 @@
...
+
+on:
+  pull_request:
+    branches: [ master ]
```
vigneshmsft, line 5:
```suggestion
    branches: [ master ]
    paths:
    - 'app/models/**'
```
add a path filter to run only when the content changes?

vigneshmsft (commented):
Github Actions, looks good. 👍🏽
Can we split the ruby update to a separate PR?
Also is is possible to run this as standalone rake task without db:setup ?

.github/workflows/enums.yml
mnacos: I've called the Github action 'Custom checks' because the same action file can specify multiple jobs. The job is called 'Detect enum changes' and what you see is: 

image

which may be descriptive enough? I was thinking we could add more jobs into 'Custom checks' in the future. Jobs can run in parallel and even depend on other jobs, such as a setup job.

.github/workflows/enums.yml
mnacos: Strange, I'd expect so, too, but the bash-specific `${MULTILINE//$'\r'/'%0D'}` constructs didn't work properly unless I explicitly asked for bash.

.github/workflows/enums.yml
mnacos: Great suggestion, I'll try it.

AgaDufrat (approved):
Tested the ruby upgrade on mac os x

mnacos:
> Github Actions, looks good. 👍🏽
> Can we split the ruby update to a separate PR?
> Also is is possible to run this as standalone rake task without db:setup ?

Thanks for reviewing! The script relies on loading Rails to get the db-level enum values, and Rails tries to access the database as part of that. The database needs no data but a database needs to be there.

Now we run this github action only when changes have been made to `app/models`, this PR is now green even though master is still on 2.7.1 -- I think we could just go ahead and upgrade.

.github/workflows/enums.yml
mnacos: Works well

vigneshmsft (approved):

.github/workflows/enums.yml
mnacos: Renamed 'Custom checks' to 'Pull request checks', as the overall action is PR-oriented.

malcolmbaig:
This is great. Thanks @mnacos !

PR #3516, opened 2020-11-25
Support: add filtering by candidate and application choice IDs

duncanjbrown:
## Context

It's long been a bit weird that you can't get directly to an application choice in support seeing as providers who contact us deal exclusively in application choice IDs.

It's also a bit weird that we send vendors candidate IDs in the form `CXXX` eg `C123` with no way to trace those back to applications except by hacking the `support/candidates/:id` URL.

It's *also* a shame we don't have an empty state for the support applications list. When the filters return nothing, we show a blank list instead of a message.

Fix all of the above.

### Search by application choice ID

"Provider application ID" because it's what providers use.

Screenshot 2020-11-25 at 22 56 09

### Search by candidate ID

This input treats `C123` and `123` the same

Screenshot 2020-11-25 at 22 55 22

### Empty state

Screenshot 2020-11-25 at 22 55 42

## Guidance to review

Code hopefully straightforward. Any improvement on field names?

## Link to Trello card

N/A, memories of support

## Things to check

### DIFF
--- a/app/components/filter_component.html.erb
+++ b/app/components/filter_component.html.erb
@@ -50,7 +50,7 @@
                   <%= filter[:heading] %>
                 </legend>
                 <% if filter[:type] == :search %>
-                  <input class="govuk-input" id="<%= filter[:name] %>" name="<%= filter[:name] %>" type="text" value="<%= filter[:value] %>" aria-labelledby="filter-legend-<%= filter[:name] %>" >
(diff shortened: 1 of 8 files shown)


paulrobertlloyd:
Just 2 things:

* I’d add `.govuk-input--width-5` class to the ID filters, to indicate the typical size of an ID
* Missing ‘No candidates found’ message when trying to filter by candidate ID

duncanjbrown:
Updated @paulrobertlloyd

paulrobertlloyd (approved):
🚢

PR #3517, opened 2020-11-26
Service to measure average distances

stevehook:
## Context

We've been asked to build two data exports. One that measures the distance between candidates and their selected choice and one that measures the distance between Provider's main site and other sites.

Because `Candidate`, `Provider` and `Site` all have the same location interface (`latitude` and `longitude` methods) this can be done with a service that takes a `start` model and an array of `destinations`. `start` and `destinations` can be any combination of `Candidate`, `Provider` or `Site`.

- [x] Create a new `MeasureDistances` service with an `average_distance` method.

I've added average distances for provider - sites and candidate - choices in the support interface in a follow-up PR https://github.com/DFE-Digital/apply-for-teacher-training/pull/3530.

## Guidance to review

- use the `geocoder` gem that we already using in https://github.com/DFE-Digital/apply-for-teacher-training/pull/3510 (needs rebasing after that PR merges). The calculation is done based on an established algorithm using the given coordinates (it doesn't depend on any external APIs).
- can you think of a better name for this service?
- should we expose the average distance of candidate address to course choices and provider main site to other sites in the support interface to make it easier to check?

## Link to Trello card

https://trello.com/c/FObU2tvZ/2576-dev-create-a-service-which-measures-the-average-distance-between-one-coordinate-and-a-group-of-coordinates

## Things to check

### DIFF
--- a/app/services/candidate_interface/measure_distances.rb
+++ b/app/services/candidate_interface/measure_distances.rb
@@ -3,15 +3,14 @@ module CandidateInterface
     def average_distance(start, destinations)
       return nil unless has_coordinates?(start)
 
-      distances = destinations.map do |destination|
-        distance(start, destination)
-      end.compact
+      distances = destinations.map { |destination| distance(start, destination) }.compact
       return nil if distances.blank?
 
       distances.sum(0.0) / distances.size
     end
 
   private
+
     def distance(from, to)
       return nil unless has_coordinates?(from) && has_coordinates?(to)
 
--- a/app/services/candidate_interface/measure_distances.rb
+++ b/app/services/candidate_interface/measure_distances.rb
@@ -1,0 +1,28 @@
+module CandidateInterface
+  class MeasureDistances
+    def average_distance(start, destinations)
+      return nil unless has_coordinates?(start)
+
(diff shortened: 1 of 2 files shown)


davidgisbey (commented):
Looks really good Steve. 

The naming is quite tricky actually, although I think it's actually fine. It's fairly intuitive what's going on.

davidgisbey (approved):

stevehook:
> Is it worth considering doing this calculation in Postgres instead? There's a built-in [earthdistance](https://www.postgresql.org/docs/current/earthdistance.html) extension which does this (doesn't require PostGIS). I think it uses the same Haversine formula as the geocoded gem.

It looks like it's possible and would likely improve performance especially for cases where the models objects are not already loaded into memory. I quite like the idea of offloading work to the DB but I'm not sure it would be as easy maintain in future - generally I reckon people would find it easier to work with geocoder than raw SQL (and extension functions).

davidgisbey:
I think it's definitely worth considering if we're struggling with performance, but since It's for extracts and we've already built a working implementation i think we should go for this for now. 

I also agree with Steve's point re. parsing ruby vs SQL blobs.

PR #3518, opened 2020-11-26
Fix bug with withdrawal feedback page not rendering

davidgisbey:
## Context

If a candidate chooses not to provide any withdrawal feedback then they get an error, as the view being called to render doesn't exist.

- Call the correct view in the controller action
- Display the correct error messages 
- Link errors correctly 
- Add a test

image

## Link to Trello card

https://trello.com/c/ka7QcU1A/2586-rails-error-shown-if-submit-withdrawal-feedback-form-without-answering-any-questions-template-is-missing

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/decisions_controller.rb
+++ b/app/controllers/candidate_interface/decisions_controller.rb
@@ -67,7 +67,7 @@ module CandidateInterface
         @provider = @application_choice.provider
         @course = @application_choice.course
 
-        render :feedback
+        render :withdrawal_feedback
       end
     end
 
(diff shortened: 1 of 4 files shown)


paulrobertlloyd:
Fixed the layout so that the `.govuk-grid-row` immediately wraps `.govuk-grid-column-two-thirds`. `.govuk-grid-row` adds a negative margin (that the column classes then add back in) so need to be careful where `<div>`s with this class appear as they can adversely affect the page layout.

davidgisbey:
Nice. Looks good.

paulrobertlloyd (approved):

PR #3519, opened 2020-11-26
Let candidates proceed if they decide not to add another job

TobyRet:
## Context

Currently, if you add a job, choose to add another, but then change your mind, options for exiting out of this state are non-optimal.

- Redirect to work history review page if candidate submits blank form
- ‘Add another job’ radio now defaults to false
- Horizontal line removed in new job form
- Adding another qualification radio button defaults to 'no'

## Guidance to review

- Create a job then choose to add another. When you submit the form (keeping all fields blank) you should be redirected back to the Work Experience review page.
- When adding another ‘Academic and other relevant qualification’ the ‘Do you want to add another qualification?’ radio button selection should now defaults to 'No, not at the moment' 

## Link to Trello card

https://trello.com/c/wbUacSal/1974-let-candidates-proceed-if-decide-not-to-add-another-qualification-or-job

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/work_history/edit_controller.rb
+++ b/app/controllers/candidate_interface/work_history/edit_controller.rb
@@ -14,18 +14,20 @@ module CandidateInterface
                                   start_date_year: start_date.year,
                                   end_date_month: end_date&.month || '',
                                   end_date_year: end_date&.year || '',
-                                  add_another_job: true,
+                                  add_another_job: 'no',
                                 )
                               else
-                                WorkExperienceForm.new(add_another_job: true)
+                                WorkExperienceForm.new(add_another_job: 'no')
                               end
     end
 
     def create
       @work_experience_form = WorkExperienceForm.new(work_experience_form_params)
 
-      if @work_experience_form.save(current_application)
-        if @work_experience_form.add_another_job == 'true'
+      if @work_experience_form.blank_form?
+        redirect_to candidate_interface_work_history_show_path
+      elsif @work_experience_form.save(current_application)
+        if @work_experience_form.add_another_job == 'yes'
           redirect_to candidate_interface_new_work_history_path
         else
           redirect_to candidate_interface_work_history_show_path
@@ -41,7 +43,7 @@ module CandidateInterface
       work_experience = current_application
         .application_work_experiences.find(work_experience_params[:id])
       @work_experience_form = WorkExperienceForm.build_from_experience(work_experience)
-      @work_experience_form.add_another_job = false
+      @work_experience_form.add_another_job = 'no'
     end
 
     def update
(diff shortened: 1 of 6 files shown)


paulrobertlloyd:
This looks good to me! I so slightly wonder if we need to have both fixes at the same time, but not sure how closely we are monitoring this to be able to make a decision on whether one or both are needed (I’m not entirely sure about defaulting to ‘No’, for example).

davidgisbey (approved):
Looks good. I tested it's working nicely. Not related to this, but this will get more complicated on other quals if we try and implement something similar as we prefill values in some contexts.

paulrobertlloyd (approved):

app/controllers/candidate_interface/work_history/edit_controller.rb
```diff
@@ -14,18 +14,20 @@ def new
                                   start_date_year: start_date.year,
                                   end_date_month: end_date&.month || '',
                                   end_date_year: end_date&.year || '',
-                                  add_another_job: true,
+                                  do_not_add_another_job: true,
```
frankieroberto, line 17: Any reason not to use `add_another_job: false`?

app/forms/candidate_interface/other_qualification_wizard.rb
```diff
@@ -0,0 +1,226 @@
+module CandidateInterface
+  class OtherQualificationWizard
```
frankieroberto, line 2: I can't see whether this new class is referenced from at all? 🤔

app/controllers/candidate_interface/work_history/edit_controller.rb
TobyRet: Yeah that would be ideal, however we now need to do reverse the logic in order to pre-select the correct radio button.
See https://github.com/DFE-Digital/apply-for-teacher-training/pull/3478

paulrobertlloyd: Not sure I follow… `add_another_job: false` seems much clearer. Can we not reverse the boolean on the form?

TobyRet: @paulrobertlloyd The radio button needs to resolve to `true` in order to be selected. In response to the question 'Do you want to add another job?' we want 'No' to be selected, not 'Yes' right?

app/forms/candidate_interface/other_qualification_wizard.rb
TobyRet: Sorry - this was the result of a bad rebase. Steve had done some refactoring in another PR. Fixed now.

app/controllers/candidate_interface/work_history/edit_controller.rb
frankieroberto: @TobyRet can't we just initialise the form with the `no` option selected by default, via `WorkExperienceForm.new(add_another_job: false)`?

TobyRet: 👍 Yep no problem @frankieroberto. I've made the changes

PR #3520, opened 2020-11-26
Add task to backfill geocoding of candidate addresses

malcolmbaig:
## Context

We want to analyse distances between candidates and the sites they choose to train at to improve location-related functionality in the service.

Building on https://github.com/DFE-Digital/apply-for-teacher-training/pull/3510, this change deals with back-filling address coordinates for existing application forms.

A rake task which:

- Retrieves application forms with addresses present, in batches of 1000
- Concurrently geocodes 10 at a time
- Sleeps for 0.25s between each set of 10

## Guidance to review
- Geocoding API request limit appears to be 50 requests per second (source: https://developers.google.com/maps/documentation/geocoding/usage-and-billing)
- I've used threads to speed this up a bit. I'm thinking here that 10 concurrent requests with a 0.25s sleep between each slice of ten should mean we stay under the 50 rps limit. Does that sound about right?
- I've run this locally and checked a sample of the coordinates with Google Maps. Coordinates appear to be consistent with addresses.

## Link to Trello card
https://trello.com/c/6A8t8vVY

## Things to check

### DIFF
--- a/lib/tasks/geocode.rake
+++ b/lib/tasks/geocode.rake
@@ -1,8 +1,15 @@
 namespace :geocode do
   desc 'Geocode candidate addresses'
   task candidate_addresses: :environment do
+    number_of_threads = 10
+    db_connection_pool_size = number_of_threads + 1
+
     if Geocoder.config.api_key.blank?
       raise 'You need to set a GOOGLE_MAPS_API_KEY'
+    end
+
+    if ENV['RAILS_MAX_THREADS'].to_i < db_connection_pool_size
+      raise "Set a connection pool size of at least #{db_connection_pool_size} by setting the RAILS_MAX_THREADS env var"
     end
 
     Benchmark.bm do |x|
--- a/lib/tasks/geocode.rake
+++ b/lib/tasks/geocode.rake
@@ -1,0 +1,25 @@
+namespace :geocode do
+  desc 'Geocode candidate addresses'
+  task candidate_addresses: :environment do
+    if Geocoder.config.api_key.blank?
+      raise 'You need to set a GOOGLE_MAPS_API_KEY'
+    end
+
+    Benchmark.bm do |x|
+      x.report do
+        ApplicationForm.where.not(address_line1: nil).find_in_batches do |application_batch|
+          application_batch.each_slice(10) do |slice|
+            threads = slice.map do |application|
+              Thread.new do
+                application.latitude, application.longitude = application.geocode
+                application.save!
+              end
+            end
+            threads.each(&:join)
+            sleep(0.25)
+          end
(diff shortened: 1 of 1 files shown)


lib/tasks/geocode.rake
```diff
@@ -0,0 +1,25 @@
+namespace :geocode do
+  desc 'Geocode candidate addresses'
+  task candidate_addresses: :environment do
+    if Geocoder.config.api_key.blank?
+      raise 'You need to set a GOOGLE_MAPS_API_KEY'
+    end
+
+    Benchmark.bm do |x|
```
TobyRet, line 8: LGTM ... Are there any implications for when it's run in prod in terms of performance or affecting users?

malcolmbaig: Shouldn't be anything too heavy, there's only ~4300 records in scope for his. Will result in about 4 minutes of increased database IO, if early benchmarks are any indication.

stevehook: I guess so long as we have 11 spare Postgres connections and it's run in a container other than the Web application  it would be ok.

stevehook (approved):
Tried locally and worked nicely. Throttling makes sense.

I had to prefix the command with `RAILS_MAX_THREADS=11 ` to change the default connection pool size up from the default 5. (10 worker threads plus one for the main thread?).

Is it worth moving it to `/support/settings/tasks`? Might make sense if a non-dev has to run it or its used more than once, otherwise not so much. It would mean switching it to a Sidekiq job.

malcolmbaig:
> I had to prefix the command with `RAILS_MAX_THREADS=11 ` to change the default connection pool size up from the default 5. (10 worker threads plus one for the main thread?).
> 
> Is it worth moving it to `/support/settings/tasks`? Might make sense if a non-dev has to run it or its used more than once, otherwise not so much. It would mean switching it to a Sidekiq job.

Increasing the connection pool size seems sensible :+1: . WDYT of the latest commit?

This will hopefully be a one-shot, but can make it available in support if needed later on.

stevehook:
> > I had to prefix the command with `RAILS_MAX_THREADS=11 ` to change the default connection pool size up from the default 5. (10 worker threads plus one for the main thread?).
> > Is it worth moving it to `/support/settings/tasks`? Might make sense if a non-dev has to run it or its used more than once, otherwise not so much. It would mean switching it to a Sidekiq job.
> 
> Increasing the connection pool size seems sensible 👍 . WDYT of the latest commit?
> 
> This will hopefully be a one-shot, but can make it available in support if needed later on.

Last commit looks good. :shipit:

stevehook (approved):

PR #3523, opened 2020-11-26
Fix provider permissions list component

benswannack:
## Context
The provider permissions list component has been confusing users because it displays information in the wrong direction. 

The correct behaviour is:
For a given provider, show information about what permissions this provider has, and with which organisations. It should not care about what the other provider in the relationship can do, only that there is a relationship. 

Switch around the check so that the correct side of permissions shows up.

Update the tests to clearly show that the component does not depend on any user permissions, and add contexts to split it up a bit.

## Guidance to review
Have a play with the review app to check the behaviour is what you'd expect

## Link to Trello card
https://trello.com/c/YBj2TDsT/3024-fix-does-not-apply-permissions-list-indicators

## Things to check

### DIFF
--- a/app/components/associated_providers_permissions_list_component.html.erb
+++ b/app/components/associated_providers_permissions_list_component.html.erb
@@ -1,49 +1,49 @@
-<% if training_providers_that_can(permission_name).any? %>
+<% if training_providers_for_which_this_provider_can(permission_name).any? %>
   <div class="govuk-!-margin-left-5 govuk-!-margin-top-2">
     <p class="govuk-body govuk-!-margin-bottom-1">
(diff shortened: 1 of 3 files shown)


app/components/associated_providers_permissions_list_component.html.erb
```diff
@@ -1,49 +1,49 @@
-<% if training_providers_that_can(permission_name).any? %>
+<% if training_providers_for_which_this_provider_can(permission_name).any? %>
```
duncanjbrown, line 1: Good wording

duncanjbrown (approved):
🎉 This is now correct! Thanks @benswannack

PR #3526, opened 2020-11-26
Add edit function to Support Int for candidate name

george-bit:
## Context

This is to cut down the reliance on developers for support.

This is built on top of the existing interface for changing candidate phone number, using the same logic.
image
image
image
image

## Guidance to review

Let me know if you believe this should be a separate view or if tests require more detail.

## Things to check


davidgisbey (approved):
Looks good George 👍

PR #3530, opened 2020-11-27
Add provider-sites & candidate-sites average distance to support UI

stevehook:
## Context

This is a follow-up to https://github.com/DFE-Digital/apply-for-teacher-training/pull/3517 to add the average distance between provider and their sites and the average distance between a candidate and the location of their chosen courses.

image

image

## Guidance to review

- Is this useful?

## Link to Trello card

https://trello.com/c/FObU2tvZ/2576-dev-create-a-service-which-measures-the-average-distance-between-one-coordinate-and-a-group-of-coordinates

## Things to check

### DIFF
--- a/app/components/support_interface/application_summary_component.rb
+++ b/app/components/support_interface/application_summary_component.rb
@@ -1,6 +1,7 @@
 module SupportInterface
   class ApplicationSummaryComponent < ViewComponent::Base
     include ViewHelper
+    include GeocodeHelper
 
     delegate :support_reference,
              :submitted_at,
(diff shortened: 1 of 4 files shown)


raamSoftwire (approved):
The code looks good Steve. 

I think there is a question about whether or not this is useful. I can see this information being useful when a BA looks across an entire dataset of candidates, but is this information useful to a support agent just looking at one candidate?

Did this come from some user need?

davidgisbey (approved):
lgtm

PR #3534, opened 2020-11-27
Update copy for organisational permissions page

benswannack:
## Context
There were some typos on this page, which triggered a general review of the content on this page

## Link to Trello card
https://trello.com/c/3afmoQzN/3007-fix-typos-on-the-change-organisational-permissions-page

## Things to check

### DIFF
--- a/config/locales/provider_relationship_permissions.yml
+++ b/config/locales/provider_relationship_permissions.yml
@@ -1,5 +1,5 @@
 en:
   provider_relationship_permissions:
-    make_decisions: This allows an organisation to make offers, amend offers and rejection applications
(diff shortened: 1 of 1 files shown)


benswannack:
This change correctly updates the content, however I'm not sure the content is being used in the correct place. This text is used on the setup permissions flow, but not in the change permissions (once they've been setup). It feels like they should be consistent – and they don't match the prototype at the moment

johndoates:
@benswannack is right that the setup and change pages are supposed to have the same content.

@adamsilver and I plan to review the whole organisational permissions journey, and can fix the inconsistency as part of that. We think that it's OK to make the change only to the setup page for now.

AgaDufrat (approved):

PR #3540, opened 2020-11-27
Structured Qualifications Clean Up

raamSoftwire:
## Context

While reviewing the structured qualification work, quite a few bugs were found. This card deals with cleaning up all of these bugs

Checklist documented on ticket

#1 Updated English GCSE error messages 

image

image

image

image

image

## Link to Trello card

https://trello.com/c/edEWlzAL/2599-dev-structured-qualifications-clean-up-work

## Things to check

### DIFF
--- a/app/components/candidate_interface/gcse_qualification_review_component.rb
+++ b/app/components/candidate_interface/gcse_qualification_review_component.rb
@@ -77,7 +77,7 @@ module CandidateInterface
         "#{application_qualification.grade} (Double award)"
       when ApplicationQualification::SCIENCE_SINGLE_AWARD
         "#{application_qualification.grade} (Single award)"
-      when ->(subject) { subject === 'english' && application_qualification.structured_grades }
+      when ->(subject) { subject == 'english' && application_qualification.structured_grades }
         grades = JSON.parse(application_qualification.structured_grades)
         grades.map { |k, v,| "#{v} (#{k.humanize.titleize})" }
       else
--- a/app/components/candidate_interface/gcse_qualification_review_component.rb
+++ b/app/components/candidate_interface/gcse_qualification_review_component.rb
@@ -77,11 +77,28 @@ module CandidateInterface
         "#{application_qualification.grade} (Double award)"
       when ApplicationQualification::SCIENCE_SINGLE_AWARD
         "#{application_qualification.grade} (Single award)"
-      when ->(subject) { subject == 'english' && application_qualification.structured_grades }
-        grades = JSON.parse(application_qualification.structured_grades)
-        grades.map { |k, v,| "#{v} (#{k.humanize.titleize})" }
+      when ->(_n) { application_qualification.structured_grades }
+        present_structured_grades
       else
         application_qualification.grade
+      end
+    end
+
+    def present_structured_grades
+      grades = JSON.parse(application_qualification.structured_grades)
+      grades.map do |k, v,|
+        case k
+        when "english_single_award"
+          "#{v} (English Single award)"
+        when "english_double_award"
+          "#{v} (English Double award)"
+        when "english_studies_single_award"
+          "#{v} (English Studies Single award)"
+        when "english_studies_double_award"
+          "#{v} (English Studies Double award)"
+        else
+          "#{v} (#{k.humanize.titleize})"
+        end
       end
     end
 
--- a/app/components/candidate_interface/gcse_qualification_review_component.rb
+++ b/app/components/candidate_interface/gcse_qualification_review_component.rb
@@ -88,13 +88,13 @@ module CandidateInterface
       grades = JSON.parse(application_qualification.structured_grades)
       grades.map do |k, v,|
         case k
-        when "english_single_award"
+        when 'english_single_award'
           "#{v} (English Single award)"
-        when "english_double_award"
+        when 'english_double_award'
           "#{v} (English Double award)"
-        when "english_studies_single_award"
+        when 'english_studies_single_award'
           "#{v} (English Studies Single award)"
-        when "english_studies_double_award"
+        when 'english_studies_double_award'
           "#{v} (English Studies Double award)"
         else
           "#{v} (#{k.humanize.titleize})"
--- a/app/components/candidate_interface/gcse_qualification_review_component.rb
+++ b/app/components/candidate_interface/gcse_qualification_review_component.rb
@@ -65,14 +65,19 @@ module CandidateInterface
     end
 
     def present_grades
-      if application_qualification.subject == ApplicationQualification::SCIENCE_TRIPLE_AWARD
+      case application_qualification.subject
+      when ApplicationQualification::SCIENCE_TRIPLE_AWARD
         grades = application_qualification.structured_grades
         [
           "#{grades['biology']} (Biology)",
           "#{grades['chemistry']} (Chemistry)",
           "#{grades['physics']} (Physics)",
         ]
-      elsif application_qualification.subject == 'english' && application_qualification.structured_grades
+      when ApplicationQualification::SCIENCE_DOUBLE_AWARD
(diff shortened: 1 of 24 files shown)


paulrobertlloyd (approved):

app/models/application_qualification.rb
```diff
@@ -74,7 +74,7 @@ def incomplete_other_qualification?
         send(field_name).blank?
       end
     else
-      return true if EXPECTED_OTHER_QUALIFICATION_DATA.any? do |field_name|
+      return true if EXPECTED_OTHER_QUALIFICATION_DATA.reject { |k| k == :grade }.any? do |field_name|
```
davidgisbey, line 77: Can we just remove `:grade` from the constant rather than reject it?

davidgisbey:
Looks good, just found this while testing locally.

image

davidgisbey:
One more. We should be using an autocomplete here. Just validating in the form
 
image

app/components/gcse_qualification_cards_component.rb
```diff
@@ -47,14 +53,19 @@ def naric_statement(qualification)
...
         "#{grades['physics']} (Physics)",
       ]
-    elsif qualification.structured_grades
+    when ApplicationQualification::SCIENCE_DOUBLE_AWARD
+      ["#{qualification.grade} (Double award)"]
+    when ApplicationQualification::SCIENCE_SINGLE_AWARD
+      ["#{qualification.grade} (Single award)"]
+    when ->(_n) { qualification.structured_grades }
       grades = JSON.parse(qualification.structured_grades)
       grades.map { |k, v,| "#{v} (#{k.humanize.titleize})" }
```
davidgisbey, line 70: This causes double award to upcase. We should probably have a bespoke bit of logic for when grade.key == english_double_award 

image

app/helpers/gcse_qualification_helper.rb
```diff
@@ -15,4 +15,24 @@ def hint_for_gcse_edit_grade(subject, qualification_type)
       t("gcse_edit_grade.hint.#{subject}.#{qualification_type}")
     end
   end
+
+  def grade_step_title(subject, qualification_type)
+    subject = subject.capitalize if subject == 'english'
```
davidgisbey, line 20: Please can we also capitalise Science and Maths here so it's consistent across the subjects

davidgisbey (changes requested):
Really great work Raam, I've found a few things while manual testing.

app/helpers/gcse_qualification_helper.rb
TobyRet: @paulrobertlloyd ☝️ - can you confirm? Thought it was only English that was supposed to be capitalised?

app/components/candidate_interface/gcse_qualification_review_component.rb
```diff
@@ -65,14 +65,19 @@ def grade_row
...
           "#{grades['biology']} (Biology)",
           "#{grades['chemistry']} (Chemistry)",
           "#{grades['physics']} (Physics)",
         ]
-      elsif application_qualification.subject == 'english' && application_qualification.structured_grades
+      when ApplicationQualification::SCIENCE_DOUBLE_AWARD
+        "#{application_qualification.grade} (Double award)"
+      when ApplicationQualification::SCIENCE_SINGLE_AWARD
+        "#{application_qualification.grade} (Single award)"
+      when ->(subject) { subject == 'english' && application_qualification.structured_grades }
```
TobyRet, line 80: Interesting, did not know you could use lambdas in case statements 👍

raamSoftwire: The reason I needed this was because it was the only way I could find to allow an AND statement. I don't think it's particularly readable, so please shout if there's a more standard way.

app/models/application_qualification.rb
TobyRet, line 77: I'm wondering the entire `incomplete_other_qualification` could be re-written to make it more readable or at the min, extract the if clauses to methods for readability?

e.g.

```
def incomplete_other_qualification?
   (qualification_type == 'non_uk' && incomplete_international_data?) || incomplete_other_data?
end

```

config/locales/application_form.yml
```diff
@@ -843,26 +843,26 @@ en:
             english_gcses:
               blank: Select at least one GCSE
             grade_english_single:
-              blank: Enter your grade
-              invalid: Enter a real grade
+              blank: Enter your English (Single award) grade
```
TobyRet, line 846: Probably need to ensure consistency with science validation messages. I think @paulrobertlloyd wanted ‘science’ dropped from each as this was deemed unnecessary.

Science validation messages - https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/config/locales/application_form.yml#L811

TobyRet (commented):
Looking good - some comments added

config/locales/application_form.yml
raamSoftwire: I wonder if English GCSEs are a bit different? 

I don't think people reference to Literature or Language GCSEs rather than English Literature or English Language. I'd be keen to see what Paul thinks here.

app/components/candidate_interface/gcse_qualification_review_component.rb
davidgisbey: +1

app/helpers/gcse_qualification_helper.rb
davidgisbey: You're probably right. My grasp of the English language is tenuous at best.

app/models/application_qualification.rb
raamSoftwire: This was a great suggestion Toby, thanks! I worked it through step by step and ended up at the same solution. This method is now much, much easier to read.

app/helpers/gcse_qualification_helper.rb
paulrobertlloyd: Yep, only English is capitalised.

app/models/application_qualification.rb
raamSoftwire: This actually broke some tests and led me to deep a bit deeper. For non UK qualifications `incomplete_other_data?` is true and the way the old way worked was by returning early so it didn't get that far. 

```  def incomplete_other_qualification?
    if qualification_type == 'non_uk'
      incomplete_international_data?
    elsif other?
      incomplete_other_qualification_data?
    else
      false
    end
  end

**Comment by raamSoftwire** [comment]
image

I've now updated the old English page to bring it more in line with how the newer page works 
- there is no autocomplete
- I've updated the hint to include double awards because we already accept them, but don't mention it
- made the width the right size for double award grades

File: app/views/candidate_interface/other_qualifications/details/_form.html.erb [hunk 6]
```diff
@@ -7,7 +7,7 @@
     <%= f.govuk_text_field :subject, label: { text: t('application_form.other_qualification.subject.label'), size: 'm' } %>
     <%= tag.div(id: 'subject-autocomplete-data', data: { source: subject_autocomplete_data }) if subject_autocomplete_data %>
     <%= f.govuk_text_field :grade,
-                           label: { text: t('application_form.other_qualification.grade.optional_label'), size: 'm' },
+                           label: { text: @form.qualification_type === "Other" ? t('application_form.other_qualification.grade.optional_label') : t('application_form.other_qualification.grade.label'), size: 'm' },
```
davidgisbey, line 10: == 'Other'

config/locales/application_form.yml
raamSoftwire: There is also English single and double as well as English Studies single and double, so I think it'd be clearer to leave the extra clarification in here.

davidgisbey (approved):
Great work 👍

PR #3542, opened 2020-11-27
Update the candidate tracker and application reference export to include references' requested_at times

davidgisbey:
## Context

The CandidateTrackerExport and the ApplicationReferenceExport need received references requested_at timings so analysis has can be done on the average amount of time it takes for references to be returned.

- Update the CandidateJourneyTracker to include reference_1_requested_at & reference_2_requested_at
- Update the ApplicationReferenceExport to include requested_at

## Guidance to review

The work on the CnadidateJourneyTracker is complicated slightly by the fact that reference_1_received and reference_1_requested_at should always be from the same reference, so sorting by the oldest requested_at_date doesn't work.
 
## Link to Trello card

https://trello.com/c/oTAqdpTF/2591-data-export-add-request-date-for-references-application-references-extract

## Things to check

### DIFF
--- a/app/services/support_interface/candidate_journey_tracker.rb
+++ b/app/services/support_interface/candidate_journey_tracker.rb
@@ -8,6 +8,8 @@ module SupportInterface
       form_not_started
       form_started_and_not_submitted
       submitted
+      reference_1_requested_at
+      reference_2_requested_at
       reference_1_received
       reference_2_received
       reference_reminder_email_sent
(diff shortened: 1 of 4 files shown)


app/services/support_interface/candidate_journey_tracker.rb
```diff
@@ -141,9 +151,16 @@ def received_references
...
-    def received_reference_times
+    def received_references_timings
       @received_reference_times ||=
-        received_references.map { |reference| earliest_update_audit_for(reference, feedback_status: 'feedback_provided') }.compact.sort
+        received_references.map do |reference|
+          {
+            received_at: earliest_update_audit_for(reference, feedback_status: 'feedback_provided'),
+            requested_at: reference.requested_at,
+          }
+        end
```
malcolmbaig, line 161: Nice :ok_hand: . Had a mind to make the references available in their own private method, but it's more than clear enough what's going on here.

app/services/support_interface/candidate_journey_tracker.rb
```diff
@@ -141,9 +151,16 @@ def received_references
       @received_references ||= all_references.select(&:feedback_provided?)
     end
 
-    def received_reference_times
+    def received_references_timings
       @received_reference_times ||=
```
malcolmbaig, line 155: Worth updating the instance var name as well for consistency.

malcolmbaig (approved):

app/services/support_interface/candidate_journey_tracker.rb
davidgisbey: Yea good point 👍

PR #3544, opened 2020-11-27
Review reference paths

paulrobertlloyd:
## Context

Update reference paths so that they follow conventions used elsewhere on the service. 

* Use hyphens instead of underscores
* Update ‘destroy’ paths so that they use the `delete` keyword.

## Link to Trello card

https://trello.com/c/EmpbwjXY/

## Things to check


davidgisbey (approved):
def cleaner 👌

PR #3552, opened 2020-11-30
Notify providers and candidates when a ucas match is resolved

despo:
## Context

Emails notifying candidate and provider about a UCAS Match being resolved.

UCAS resolved match email trigerred when processing matching data 
Apply resolved match email trigerred when an application is withdrawn

## Link to Trello card

https://trello.com/c/jLEgGDIv/3047-%F0%9F%8F%88-notify-providers-and-candidates-when-a-ucas-match-is-resolved

## Things to check

### DIFF
--- a/app/models/ucas_match.rb
+++ b/app/models/ucas_match.rb
@@ -13,6 +13,7 @@ class UCASMatch < ApplicationRecord
     initial_emails_sent: 'initial_emails_sent',
     reminder_emails_sent: 'reminder_emails_sent',
     ucas_withdrawal_requested: 'ucas_withdrawal_requested',
+    resolved_on_ucas: 'resolved_on_ucas',
   }
 
   def trackable_applicant_key
--- a/app/models/ucas_match.rb
+++ b/app/models/ucas_match.rb
@@ -32,13 +32,15 @@ class UCASMatch < ApplicationRecord
   def action_needed?
     return false if processed?
 
-    return false unless dual_application_or_dual_acceptance?
+    return false if resolved?
 
-    return true if ucas_withdrawal_requested?
+    return false unless dual_application_or_dual_acceptance?
 
     return need_to_send_reminder_emails? if initial_emails_sent?
 
     return need_to_request_withdrawal_from_ucas? if reminder_emails_sent?
+
+    return false if ucas_withdrawal_requested?
 
     true
   end
@@ -84,15 +86,11 @@ class UCASMatch < ApplicationRecord
       :reminder_emails_sent
     elsif reminder_emails_sent? && need_to_request_withdrawal_from_ucas?
       :ucas_withdrawal_requested
-    elsif ucas_withdrawal_requested?
-      :confirmed_withdrawal_from_ucas
     end
   end
 
   def last_action
     return nil if action_taken.nil?
-
-    return :confirmed_withdrawal_from_ucas if processed? && ucas_withdrawal_requested?
 
     action_taken.to_sym
   end
--- a/app/models/ucas_match.rb
+++ b/app/models/ucas_match.rb
@@ -98,27 +98,29 @@ class UCASMatch < ApplicationRecord
   end
 
   def application_choices_for_same_course_on_both_services
-    ucas_matched_applications.select(&:both_scheme?).map(&:application_choice)
+    duplicate_course_applications.map(&:application_choice)
   end
 
-  def ucas_withdrawal?
-    ucas_matched_applications.select(&:both_scheme?) &&
-      ucas_matched_applications.select(&:application_withdrawn_on_ucas?).any?
+  def duplicate_applications_withdrawn_from_ucas?
+    duplicate_course_applications.any? && duplicate_course_applications.all?(&:application_withdrawn_on_ucas?)
   end
 
-  def apply_withdrawal?
-    ucas_matched_applications.select(&:both_scheme?) &&
-      ucas_matched_applications.select(&:application_withdrawn_on_apply?).any?
+  def duplicate_applications_withdrawn_from_apply?
+    duplicate_course_applications.any? && duplicate_course_applications.all?(&:application_withdrawn_on_apply?)
   end
 
   def application_for_the_same_course_in_progress_on_both_services?
-    application_for_the_same_course_on_both_services = ucas_matched_applications.select(&:both_scheme?)
-
-    application_for_the_same_course_on_both_services.map(&:application_in_progress_on_ucas?).any? &&
-      application_for_the_same_course_on_both_services.map(&:application_in_progress_on_apply?).any?
+    duplicate_course_applications.map(&:application_in_progress_on_ucas?).any? &&
+      duplicate_course_applications.map(&:application_in_progress_on_apply?).any?
   end
 
   def calculate_action_date(action, effective_date)
     TimeLimitCalculator.new(rule: action, effective_date: effective_date).call.fetch(:time_in_future).to_date
+  end
+
+private
+
+  def duplicate_course_applications
+    ucas_matched_applications.select(&:both_scheme?)
   end
 end
--- a/app/models/ucas_match.rb
+++ b/app/models/ucas_match.rb
@@ -17,6 +17,14 @@ class UCASMatch < ApplicationRecord
     resolved_on_ucas: 'resolved_on_ucas',
   }
 
+  def ready_to_resolve?
+    action_taken.present? && !action_needed? && !resolved?
+  end
+
+  def resolved?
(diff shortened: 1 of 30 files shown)


app/models/ucas_match.rb
```diff
@@ -91,6 +101,16 @@ def application_choices_for_same_course_on_both_services
     ucas_matched_applications.select(&:both_scheme?).map(&:application_choice)
   end
 
+  def ucas_withdrawal?
+    ucas_matched_applications.select(&:both_scheme?) &&
+      ucas_matched_applications.select(&:application_withdrawn_on_ucas?).any?
```
AgaDufrat, line 106: I don’t think it explicitly checks if the duplicate application was withdrawn from UCAS. The candidate could have withdrawn from some other course they applied for on UCAS. 

I think it would return `true` for the following scenario:
Screenshot 2020-12-02 at 12 11 28

Maybe we could also call the method something more descriptive.

app/models/ucas_match.rb
```diff
@@ -91,6 +101,16 @@ def application_choices_for_same_course_on_both_services
...
+  end
+
+  def apply_withdrawal?
+    ucas_matched_applications.select(&:both_scheme?) &&
+      ucas_matched_applications.select(&:application_withdrawn_on_apply?).any?
```
AgaDufrat, line 111: similar to the comment above

AgaDufrat (commented):
Looking good. I left a couple of comments.

There is also a question for the Transition team: Do we need to notify providers and candidates when a multiple acceptances are resolved? There is currently no template for it.

app/views/candidate_mailer/ucas_match_resolved_on_apply_email.text.erb
```diff
@@ -0,0 +1,15 @@
+Dear <%= @application_form.full_name %>
+
+We’re writing to confirm your choice to study <%= @course_name_and_code %> at <%= @provider_name %> has been removed from your Apply for teacher training application.
+
+You should use [UCAS Teacher Training](https://teachertraining.track.ucas.com/track/Login.jsp) to continue your application for <%= @course_name_and_code %>.
+
+Any other courses you applied to have not been affected.
+
+We informed <%= @provider_name %> that the course choice was removed from your UCAS application.
```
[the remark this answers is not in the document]
despo: @AgaDufrat Apply of DfE Apply, similar to the email subject?

spec/models/ucas_matched_application_spec.rb
```diff
@@ -262,7 +262,7 @@
   end
 
   describe '#application_accepted_on_apply?' do
-    it 'retruns true if application is accepted on Apply' do
+    it 'retuns true if application is accepted on Apply' do
```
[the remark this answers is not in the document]
despo: ...and ironically making more typos while doing so :)

app/models/ucas_match.rb
despo:
```suggestion
      ucas_matched_applications.select(&:both_scheme?).select(&:application_withdrawn_on_ucas?).any?      
```


@AgaDufrat if I understand correctly should the condition then be updated to only check if the applications that are on both schemes have been withdrawn similar to my suggestions and rename it to something like `duplicate_application_withdrawn_from_ucas?`

app/services/ucas_matches/send_resolved_on_apply_emails.rb
```diff
@@ -0,0 +1,21 @@
...
+    attr_reader :ucas_match
+
+    def initialize(ucas_match)
+      @ucas_match = ucas_match
+    end
+
+    def call
+      raise 'The application has not been resolved on Apply' if !ucas_match.resolved_on_apply?
+
+      application_choice = ucas_match.application_choices_for_same_course_on_both_services.first
```
AgaDufrat, line 12: We are assuming that the candidate only applied for 1 course on both services but theoretically they could have up to 3 application choices for for same course on both services.

Maybe we could add a method called something like `duplicate_application_choices_withdrawn_from_ucas` to the UCASMatch model which would return ApplicationChoices for same course on both services which were withdrawn on UCAS and then send an email for each.
 _Not ideal to spam them with emails but it's an unlikely scenario._

Once we have done that we could change
`if match.ready_to_resolve? && match.ucas_withdrawal?`
to 
`if match.ready_to_resolve? && match.duplicate_application_choices_withdrawn_from_ucas.any?`

app/models/ucas_match.rb
AgaDufrat: Yes, that's what I was thinking and I like the name!

You may also want to consider dealing with more than one duplicate application choices withdrawn from UCAS. 
I described my idea here (we probably wouldn't need public `duplicate_application_withdrawn_from_ucas?`): https://github.com/DFE-Digital/apply-for-teacher-training/pull/3552/files#r534232874

app/services/ucas_matches/send_resolved_on_apply_emails.rb
despo: @AgaDufrat  can we catch up about this tomorrow morning. It seems like there are missing requirements around this functionality that we are trying to compensate for and maybe we should reconsider first reviewing our approach on the initial and reminder emails as well so that we have knowledge of which application we have already emailed the candidate about?

It also sounds like the UR should maybe get involved so we can figure out how to approach sending multiple emails out and if we can instead extend the existing email templates to handle cases of multiple applications.

app/services/ucas_matches/send_resolved_on_apply_emails.rb
```diff
@@ -0,0 +1,21 @@
+module UCASMatches
+  class SendResolvedOnApplyEmails
+    attr_reader :ucas_match
+
+    def initialize(ucas_match)
+      @ucas_match = ucas_match
+    end
+
+    def call
+      raise 'The application has not been resolved on Apply' if !ucas_match.resolved_on_apply?
```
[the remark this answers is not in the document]
despo: I think in this instance a negated if reads better, eg raise an error if the match has not been resolved on apply rather than raise an error unless it has been resolved.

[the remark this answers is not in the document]
AgaDufrat: yeah, I agree it reads nicer. I used to be told off by rubocop for `if !`

spec/components/support_interface/ucas_match_action_component_spec.rb
```diff
@@ -103,26 +103,40 @@
...
                             candidate_last_contacted_at: Time.zone.now - 1.day)
         allow(ucas_match).to receive(:dual_application_or_dual_acceptance?).and_return(true)
 
         result = render_inline(described_class.new(ucas_match))
 
         expect(result.text).to include('No action required')
-        expect(result.text).to include('We confirmed that the candidate was withdrawn from UCAS. We contacted UCAS to request removal from UTT on the 18 October 2020')
+        expect(result.text).to include('We confirmed that the candidate was withdrawn from Apply on the 18 October 2020')
       end
     end
```
AgaDufrat, line 142: I think we can delete the test below 👇 (it won't let me comment on it) 
`renders correct information after UCAS matching data was updated automatically after requesting withdrawal`

as it's checking that a resolved on UCAS match is no longer flagged as "Action needed". 
The new `renders correct information after withdrawal from UCAS was confirmed by a support user` covers the new scenario so we don't need it.

spec/components/support_interface/ucas_match_action_component_spec.rb
```diff
@@ -103,43 +103,40 @@
...
-                            candidate_last_contacted_at: Time.zone.now - 8.days)
-        allow(ucas_match).to receive(:dual_application_or_dual_acceptance?).and_return(false)
+                            action_taken: 'resolved_on_apply',
+                            candidate_last_contacted_at: Time.zone.now - 1.day)
+        allow(ucas_match).to receive(:dual_application_or_dual_acceptance?).and_return(true)
 
         result = render_inline(described_class.new(ucas_match))
 
         expect(result.text).to include('No action required')
-        expect(result.text).not_to include('Confirm withdrawal from UCAS')
```
despo, line 142: @AgaDufrat  Removed as suggested.

app/models/ucas_match.rb
```diff
@@ -24,12 +34,12 @@ def action_needed?
```
AgaDufrat, line 34:
```suggestion
return false if resolved?
```

as it's still flagged as "Action needed" even though it's resolved and it's breaking the component

Screenshot 2020-12-04 at 17 09 41

Otherwise, LGTM!

despo: @AgaDufrat ✔️

AgaDufrat (approved):

PR #3555, opened 2020-11-30
DRY up structured reasons for rejection validations

steventux:
## Context

The majority of questions are validated in the same way. 
We store the hierarchy of the complex questions in the `INITIAL_QUESTIONS` constant, so use this to generate the validation rules.

- Adds a validator class to handle the repetitive validation logic.
- Uses the existing question config to generate the validation rules where possible.

## Guidance to review

There are so many combinations of questions and answers - can you find a hole in the validation rules?

## Things to check

### DIFF
--- a/app/forms/provider_interface/reasons_for_rejection_wizard.rb
+++ b/app/forms/provider_interface/reasons_for_rejection_wizard.rb
@@ -50,6 +50,51 @@ module ProviderInterface
       },
     }.freeze
 
+    class NestedAnswerValidator < ActiveModel::EachValidator
+      def validate_each(record, attribute, value)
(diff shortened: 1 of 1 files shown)


duncanjbrown (approved):
I don't think this is any less readable, and it's nice to have the rules in one place

steventux:
@duncanjbrown I realised the presence validator was missing from the top level questions, the tests passed because the inclusion validator caught them but I've added this here https://github.com/DFE-Digital/apply-for-teacher-training/pull/3555/commits/f7aa26b5f462dbb75cc238bef263960928fa0502#diff-deabb9477ff8f413ef02f6dfe20f252f0f08c40e6a99435104396d04576dbb06R60

PR #3562, opened 2020-12-01
Activity log for application events across a provider user's providers

mnacos:
### Context

We've identified a need for provider users to view a summary of activity on applications relevant to them. In practice, this relevance is defined by the set of provider the user belongs to. Application choices are always sent to the provider running the corresponding course and the provider accrediting this course, if different.

This new 'activity log' is similar to the timeline, but is meant to show events in reverse chronological order from _every_ application visible to the user, as opposed to events for a single application, which is the current timeline. This introduces several challenges, such as performance concerns, reusing the same criteria for application visibility in an events context, dealing with large numbers of applications in the UI and showing text that's appropriate in this context. For example, 'Application submitted' and 'Offer made' are obvious within the context of a single application, but 'Susan made an offer to Darren' may be more useful when viewing events for all applications together.

A previous PR has already tackled query performance, see https://github.com/DFE-Digital/apply-for-teacher-training/pull/3496

For this iteration, the activity log should include all events that appear on timeline, excluding Notes.

This PR introduces a `ProviderInterface::ActivityLogEventComponent` for displaying the individual events returned by `GetActivityLogEvents` service. This component mostly re-uses the existing timeline styles and contains presentation logic for the overall event description, the associated course description and where its link should point (application vs. offer).

This new index of activity log events is made available at `/provider/activity`. The whole feature is behind a feature flag called `:provider_activity_log`. I have introduced pagination to this index, re-using existing patterns, as our performance investigation suggests pagination is more effective for keeping page load times low than restricting events by date.

To get the `GetActivityLogEvents` query to work properly without duplicates I had to refactor the `GetApplicationChoicesForProviders` query to produce SQL that's guaranteed to return one record per `ApplicationChoice`, before all the other includes are factored in (these generate a flood of LEFT OUTER JOINS).

In the process, I've also addressed a visibility enhancement previously discussed here https://trello.com/c/j5r7hC6r and https://docs.google.com/document/d/1RZdfvqk0aj3wNbncv9oExlsoh9nRCZUwRuicfGhGKbY/edit Specifically, if the provider changes as part of the 'make offer' or 'change offer' flows, both the original (very first) provider and the current provider can see the relevant application within the provider interface and the API, but only the current provider will be able to make decisions, see safeguarding information etc. (subject to permissions, of course). This is better than the current behaviour, in which the original provider retains visibility but cannot make decisions etc. and the new provider can make decisions but cannot see the application.

Because of pagination, I've had to restrict the query within `GetActivityLogEvents` to only the events the component deals with (see #event_description). My original plan was to just ignore events (e.g. render -> false) that the component does not know about, but this would have led to unpredictable behaviour when paginating (e.g. you could have a page with no events showing or the number of events reported would not tally up).

The current scope of events included is defined as:

- all `status` changes
- sending feedback for applications rejected by default

Finally, in order to generate component previews and test #event_description, I have added traits to the `application_choice_audits` factory mirroring the ones for `application_choices`.

### Guidance to review

Is the styling of the activity log correct? I have re-used the timeline styles, should I have cloned these instead?

You can inspect all the different event descriptions here https://github.com/DFE-Digital/apply-for-teacher-training/pull/3562/files#diff-091b0af8077c8ae62203ad52722593331e5c5812c29b2439b6f2c16da0ee7988R16 (spec) and here https://apply-for-te-3030-show--cmjzhz.herokuapp.com/rails/view_components/provider_interface/activity_log_event_component (previews)

It is also worth checking the logic within `ProviderInterface::ActivityLogEventComponent` deciding which course description to show and where the link links to.

Finally, are there any other significant events we should include in the activity log? 'Change offer' events, for example, after the initial offer, will not be shown -- then again, the timeline ignores these too.

### Link to Trello card

[3030 - Implement Activity log](https://trello.com/c/OOMAetc1)

## Things to check


app/components/provider_interface/activity_log_event_component.rb
```diff
@@ -0,0 +1,46 @@
...
+    attr_reader :event, :application_choice, :course_option
+
+    def initialize(activity_log_event:)
+      @event = activity_log_event
+      @application_choice = event.auditable
+      @course_option = @application_choice.offered_option
+    end
+
+    def event_description
+      user = event.user.try(:full_name) || event.user.try(:display_name)
```
duncanjbrown, line 14: Maybe push this down onto the `event` itself? Slightly more cohesive

app/components/provider_interface/activity_log_event_component.rb
```diff
@@ -0,0 +1,46 @@
...
+      candidate = application_choice.application_form.full_name
+      changes = event.audited_changes
+
+      case changes.key?('status') && changes['status'].second
```
duncanjbrown, line 18: Again, could this live on `event`?

app/helpers/navigation_items.rb
```diff
@@ -45,7 +45,11 @@ def for_provider_primary_nav(current_provider_user, current_controller, performi
       items = []
 
       if current_provider_user && !performing_setup
-        items << NavigationItem.new('Applications', provider_interface_applications_path, is_active(current_controller, %w[application_choices decisions offer_changes]))
+        items << NavigationItem.new('Applications', provider_interface_applications_path, is_active(current_controller, %w[application_choices decisions offer_changes notes feedback conditions reconfirm_deferred_offers]))
```
duncanjbrown, line 48: :+1:

duncanjbrown (commented):
Looking good so far. What would you say to wrapping the Audit records in an `ActivityLogEvent` class at this stage? Would keep the component simpler and there's surely more logic to come

spec/components/provider_interface/activity_log_event_component_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+    candidate = event.auditable.application_form.full_name
+    yield event, user, candidate
+  end
+
+  describe '#event_description' do
+    examples = {
+      awaiting_provider_decision: '<candidate> submitted an application',
+      withdrawn: '<candidate> withdrew their application',
+      with_rejection: '<user> rejected <candidate>’s application',
+      with_rejection_by_default: '<candidate>’s application was rejected automatically',
```
johndoates, line 20:
```suggestion
      with_rejection_by_default: '<candidate>’s application was automatically rejected',
```
This is to fit the way we write it in other parts of the service. Putting 'automatically' first places emphasis on it.

spec/components/provider_interface/activity_log_event_component_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+      with_rejection_by_default_and_feedback: '<user> sent feedback to <candidate>',
+      with_offer: '<user> made an offer to <candidate>',
+      with_modified_offer: '<user> made an offer to <candidate>',
+      with_accepted_offer: '<candidate> accepted an offer',
+      with_declined_offer: '<candidate> declined an offer',
+      with_declined_by_default_offer: '<candidate>’s offer was declined automatically',
```
johndoates, line 26:
```suggestion
      with_declined_by_default_offer: '<candidate>’s offer was automatically declined',
```
This is to match how we write 'automatically rejected' in the service.

spec/components/provider_interface/activity_log_event_component_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+      with_withdrawn_offer: '<user> withdrew <candidate>’s offer',
+      with_recruited: '<user> recruited <candidate>',
```
johndoates, line 28:
```suggestion
      with_recruited: '<candidate> met all offer conditions',
```
We don't use the term 'recruited' within the service'. Adam and I think this a more accurate description of what has happened.

app/components/provider_interface/activity_log_event_component.rb
```diff
@@ -0,0 +1,98 @@
+module ProviderInterface
+  class ActivityLogEventComponent < ViewComponent::Base
+    include ViewHelper
+    attr_reader :event, :application_choice
+
+    ORIGINAL_OPTION_STATUSES = %w[awaiting_provider_decision rejected].freeze
+
+    def initialize(activity_log_event:)
+      @event = activity_log_event
+      @application_choice = event.auditable
```
despo, line 10: Maybe consider using `activity_log_event` so it's more obvious that `auditable` is retrieved from the method parameter - I had to look around to figure out that event is the same object.

app/components/provider_interface/activity_log_event_component.rb
despo: I think safe navigation is a better option than a `try` block here.

```suggestion
      user = event.user&.full_name || event.user&.display_name
```

app/components/provider_interface/activity_log_event_component.rb
```diff
@@ -0,0 +1,98 @@
...
+                         end
+    end
+
+    def link
+      routes = Rails.application.routes.url_helpers
+
+      case application_status_at_event
+      when 'offer'
+        {
+          url: routes.provider_interface_application_choice_offer_path(event.auditable.id),
```
despo, line 75:
```suggestion
          url: routes.provider_interface_application_choice_offer_path(event.auditable),
```

Auditable should be able to convert itself to the correct object and call `to_param` to return the record's id (or any other specified url parameter) in order to construct the path.

app/components/provider_interface/activity_log_event_component.rb
mnacos: `user` here is polymorphic (support users have `.display_name` and provider users have `.full_name`) -- using `&.` here would result in `NoMethodError`. I think `.try` is appropriate.

adamsilver (approved):

app/models/activity_log_event.rb
```diff
@@ -0,0 +1,15 @@
+class ActivityLogEvent < Audited::Audit
```
duncanjbrown, line 1: Nice to see! I think it would be much cleaner to accept an `audit` in the constructor than to inherit here? Then we actually control the interface of this object.

app/queries/get_application_choices_for_providers.rb
```diff
@@ -16,11 +16,29 @@ def self.call(providers:, vendor_api: false, includes: DEFAULT_INCLUDES)
...
+          WHERE c.recruitment_cycle_year IN (#{RecruitmentCycle.years_visible_to_providers.join(', ')})
+            AND ac.status IN (#{statuses.map { |s| "'#{s}'" }.join(', ')})
+            AND (
+              c.provider_id IN (#{provider_ids_string})
+              OR c.accredited_provider_id IN (#{provider_ids_string})
+              OR original_c.provider_id IN (#{provider_ids_string})
+              OR original_c.accredited_provider_id IN (#{provider_ids_string})
+            )
+      ) AS application_choices
+    GET_APPLICATION_CHOICES_FOR_PROVIDERS
```
benswannack, line 42: As discussed, hopefully we can go SQL-free by using the AR statement from https://github.com/DFE-Digital/apply-for-teacher-training/compare/audits-for-application-choice-query-experiment . Moving the includes to the end of the statement keeps the `to_sql` cleaner. (Note that this broke a test, but the fix is on that branch too :) )

mnacos: Yes, having `applications.includes(*includes)` at the end of the chain makes a world of difference. I've revised  `GetApplicationChoicesForProviders` to reduce the amount of sql (but only just, to cater for the application visibility adjustments).

app/components/provider_interface/activity_log_event_component.rb
```diff
@@ -0,0 +1,91 @@
...
+  class ActivityLogEventComponent < ViewComponent::Base
+    include ViewHelper
+    attr_reader :event, :application_choice
+    delegate :changes, to: :event
+
+    ORIGINAL_OPTION_STATUSES = %w[awaiting_provider_decision rejected].freeze
+
+    def initialize(activity_log_event:)
+      @event = activity_log_event
+      @application_choice = activity_log_event.audit.auditable
```
duncanjbrown, line 11: Would you mind doing a Law of Demeter job on this? Clients shouldn't know about `audit`, ie expose `activity_log_event.application_choice`?

duncanjbrown (approved):
This is really great, thank you @mnacos!

One tiny refactor proposed, could be next PR

PR #3563, opened 2020-12-01
Add the find feedback page

davidgisbey:
## Context

In this PR https://github.com/DFE-Digital/find-teacher-training/pull/542 i added a feedback review component to the course show and results page on Find. They point to /candidate/find-feedback on Apply.

This PR adds the find feedback page where candidates can provide their feedback

- Add a page where candidates can give their feedback
- Add a hidden field which if completed, doesn't persist the feedback to the DB to stop bot spam.
- Adds a thank you page

image

image

## Guidance to review

It might be worth reviewing the two PRs so you can follow the flow end to end.

I think it's worth me going through this with @paulrobertlloyd as it may well need small tweaks.

## Link to Trello card

https://trello.com/c/DrnKcnnW/2466-%F0%9F%93%9D-%F0%9F%8F%88-dev-football-add-feedback-prompt-find

## Things to check

### DIFF
--- a/spec/forms/candidate_interface/find_feedback_form_spec.rb
+++ b/spec/forms/candidate_interface/find_feedback_form_spec.rb
@@ -4,7 +4,7 @@ RSpec.describe CandidateInterface::FindFeedbackForm, type: :model do
   let(:form) do
     described_class.new(
       path: '/course/T92/X130',
-      original_controller: 'courses',
+      find_controller: 'courses',
       feedback: 'Make it better.',
       email_address: 'email@gmail.com',
       hidden_feedback_field: nil,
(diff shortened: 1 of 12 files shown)


spec/forms/candidate_interface/find_feedback_form_spec.rb
```diff
@@ -0,0 +1,52 @@
+require 'rails_helper'
+
+RSpec.describe CandidateInterface::FindFeedbackForm, type: :model do
+  let(:form) do
+    described_class.new(
+      path: '/course/T92/X130',
+      original_controller: 'courses',
```
TobyRet, line 7: What is `original_controller`? Is this to do with with where the feedback request was initiated (i.e. Find)

TobyRet (commented):
Looking good - just slightly confused regarding the purpose of `original_controller`

spec/forms/candidate_interface/find_feedback_form_spec.rb
davidgisbey: Yea maybe it should be called find_controller

TobyRet (approved):

PR #3567, opened 2020-12-02
Improve the referee's journey when their references aren't required

TobyRet:
## Context

Currently, when a referee attempts to give a reference and the candidate already has 2 references they are taken to a page that just says thank you. It does not give them any context as to why they cannot give a reference.

- Copy added to the ‘Thank you’ page

## Guidance to review

- Sign in as a candidate and request three references
- Log into the support and click of the reference requested emails for that application
- When click on the link inside the third reference request email you should be redirected to the 'Thank you' page where you will see the additional guidance that was added

## Link to Trello card

https://trello.com/c/EHiwLtjp/2593-%F0%9F%92%942%EF%B8%8F%E2%83%A3-dev-improve-the-referees-journey-when-their-references-arent-required

## Things to check

### DIFF
--- a/spec/system/referee_interface/referee_cannot_provide_reference_if_two_already_received_spec.rb
+++ b/spec/system/referee_interface/referee_cannot_provide_reference_if_two_already_received_spec.rb
@@ -1,0 +1,61 @@
+require 'rails_helper'
+
+RSpec.feature 'Referee is not required to submit a reference' do
+  include CandidateHelper
+
+  scenario 'Candidate has already received the minimum number of references' do
+    given_the_candidate_has_requested_three_references_and_i_am_the_third_referee
+    and_the_first_referee_has_responded_with_a_reference
+    and_the_second_referee_has_responded_with_a_reference
+    when_i_receive_an_email_with_a_reference_request
+    and_i_click_on_the_link_within_the_email
+    then_i_should_see_that_i_am_not_required_to_give_a_reference
+  end
+
+  def given_the_candidate_has_requested_three_references_and_i_am_the_third_referee
+    @first_reference = create(:reference, :feedback_requested, email_address: 'terri@example.com', name: 'Terri Tudor')
+    @second_reference = create(:reference, :feedback_requested, email_address: 'Bruce@Wayne.com', name: 'Bat Man')
+    @third_reference = create(:reference, :feedback_requested, email_address: 'Clark@Kent.com', name: 'Super Man')
+    @application = create(:completed_application_form,
+                          application_references: [
+                            @first_reference,
+                            @second_reference,
+                            @third_reference,
+                          ],
+                          candidate: current_candidate)
+  end
+
+  def and_the_first_referee_has_responded_with_a_reference
+    submit_reference(@first_reference)
+  end
+
+  def and_the_second_referee_has_responded_with_a_reference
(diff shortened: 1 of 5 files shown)


george-bit (approved):
LGTM 👍, the solution makes sense

spec/system/referee_interface/referee_cannot_provide_reference_if_two_already_received_spec.rb
```diff
@@ -0,0 +1,64 @@
+require 'rails_helper'
+
+RSpec.feature 'Referee can submit reference', with_audited: true do
```
malcolmbaig, line 3: Is auditing needed for this spec?

spec/system/referee_interface/referee_cannot_provide_reference_if_two_already_received_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+  include CandidateHelper
+
+  scenario 'Referee submits a reference for a candidate with relationship, safeguarding and review page' do
+    given_the_candidate_has_requested_three_references_and_i_am_the_third_referee
+    and_the_first_referee_has_responded_with_a_reference
+    and_the_second_referee_has_responded_with_a_reference
+    then_i_receive_an_email_with_a_reference_request
```
malcolmbaig, line 10: I think this should be `when_i_receive..`

spec/system/referee_interface/referee_cannot_provide_reference_if_two_already_received_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def and_the_second_referee_has_responded_with_a_reference
+    submit_reference(@second_reference)
+  end
+
+  def submit_reference(reference)
+    # We are submitting references this way rather than via the UI because
+    # doing raises a Bullet eager loading error which is believed to be
+    # a false positive
```
malcolmbaig, line 39: I think there are a couple of ways we could be a little clearer here:

- Setting up database state explicitly for this spec is fine, so you could remove the comment altogether. At the moment it raises more questions than it answers because the relationship between UI interaction and a Bullet error isn't obvious.
- Alternatively, we could do the UI interaction and disable bullet at the top of the spec (with an explanatory comment). This sets us up for revisiting the issue at a later date for further investigation.

spec/system/referee_interface/referee_cannot_provide_reference_if_two_already_received_spec.rb
TobyRet: No, think this was a copy / paste error. I'll update the description as well.

spec/system/referee_interface/referee_cannot_provide_reference_if_two_already_received_spec.rb
TobyRet: 👍

spec/system/referee_interface/referee_cannot_provide_reference_if_two_already_received_spec.rb
TobyRet: Good point. I'll go for option 1

gregknight1 (commented):
LGTM

PR #3569, opened 2020-12-02
Fix for candidate journey tracker export

stevehook:
## Context

This report is broken in production.

https://sentry.io/organizations/dfe-bat/issues/2067746769/?environment=production&project=1765973&query=is%3Aunresolved

This is due to missing audit entries that this report depends on to work out the `received_at` value for provided references.

This is a workaround that adds some defensive code to handle missing audit entries. Follow up investigation needed to find out why they are missing.

## Guidance to review

Does this look like a reasonable workaround? Is there anything else we can do to handle missing audit entries?

## Link to Trello card

https://trello.com/c/ZmyAYz38/2652-fix-and-investigate-why-candidate-journey-tracker-export-is-broken

## Things to check

### DIFF
--- a/app/services/support_interface/candidate_journey_tracker.rb
+++ b/app/services/support_interface/candidate_journey_tracker.rb
@@ -160,7 +160,7 @@ module SupportInterface
           }
         end
 
-      @received_reference_timings.sort_by { |times| times[:received_at] }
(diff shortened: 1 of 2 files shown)


davidgisbey (approved):
my bad sorry about that

stevehook:
> my bad sorry about that!

Not at all. Was a reasonable assumption and may have flagged an issue with the audit log.

PR #3576, opened 2020-12-03
Add a 90% threshold to the word count message

frankieroberto:
This removes the "You have 400 words remaining" message, and only shows the number of words remaining once a 90% threshold (40 words remaining or less) has been reached.

This is to avoid candidates feeling like they need to give a long explanation of their reasons for the break in work history.

### Before

Screenshot 2020-12-03 at 14 36 03

### After

Screenshot 2020-12-03 at 14 36 18

### DIFF
--- a/app/views/candidate_interface/work_history/break/_form.html.erb
+++ b/app/views/candidate_interface/work_history/break/_form.html.erb
@@ -13,6 +13,6 @@
   <%= f.govuk_date_field :end_date, omit_day: true, legend: { text: 'End of break', size: 'm', tag: 'span' } %>
 </div>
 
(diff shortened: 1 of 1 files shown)


paulrobertlloyd:
Do we think 400 words is the right limit? Should it be less? It’s still quite a lot… but maybe it doesn’t matter so much if we don’t show it immediately?

frankieroberto:
@paulrobertlloyd I figured the limit was less important if you don't see it until you’re at 360 words?

The Design System notes for the Character count component says this:

> Screenshot 2020-12-03 at 17 06 37

...so maybe we could ship this, and then see if anyone ever writes anything close to the word limit?

My slight worry with lowering the word limit is that we might introduce problems if someone was still working on their application, and then went back to edit their work history after we'd lowered the limit, and so ended up with a validation error where there wasn't one before. Unlikely though!

paulrobertlloyd:
Fair enough. It’s a nice little improvement – 🚢 !

paulrobertlloyd (approved):

davidgisbey (approved):
This is smart. Nice one 👍

PR #3580, opened 2020-12-03
Fix link colour on app start page

paulrobertlloyd:
## Context

Imported overly aggressive link styling from Register, which turns all the links on the provider start page white – hiding most of them 😭

Scopes link styling only to phase banner link. 

## Things to check

### DIFF
--- a/app/frontend/styles/_phase-banner.scss
+++ b/app/frontend/styles/_phase-banner.scss
@@ -15,7 +15,7 @@
   }
 
   .govuk-phase-banner__text,
-  .govuk-link {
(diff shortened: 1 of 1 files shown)


duncanjbrown (approved):
😅 

Merge away — but while we're here, did this missing margin come in with the same changes?

Screenshot 2020-12-03 at 19 06 56

paulrobertlloyd:
@duncanjbrown That margin is correct, but will look broken in environments where the header border isn't blue. Think that's okay though.

paulrobertlloyd:
Might see if there's a quick fix that works regardless of header border colour.

PR #3583, opened 2020-12-04
Remove matchedstate from ucas matches

despo:
Pending merge of #3552 as it relies on `UCASMatch#resolved?` introduced on that PR.

## Context

Removes matched_state from UCASMatch

## Link to Trello card

https://trello.com/c/NQYTn8V5/3048-remove-matchedstate-from-ucas-matches

## Things to check

### DIFF
--- a/app/components/support_interface/ucas_matches_table_component.rb
+++ b/app/components/support_interface/ucas_matches_table_component.rb
@@ -11,7 +11,7 @@ module SupportInterface
     def table_rows
       matches.map do |match|
         {
-          status: render(TagComponent.new(text: match.matching_data_updated? ? 'Updated' : match.matching_state.humanize, type: match.processed? ? :green : :purple)),
+          status: render(TagComponent.new(text: match.matching_data_updated? ? 'Updated' : match.matching_state.humanize, type: match.resolved? ? :green : :purple)),
           action_needed: action_needed(match),
           match_link: govuk_link_to(match.candidate.email_address, support_interface_ucas_match_path(match)),
           last_update: match.updated_at.to_s(:govuk_date),
--- a/app/components/support_interface/ucas_matches_table_component.rb
+++ b/app/components/support_interface/ucas_matches_table_component.rb
@@ -11,7 +11,7 @@ module SupportInterface
     def table_rows
       matches.map do |match|
         {
-          status: render(TagComponent.new(text: match.matching_data_updated? ? 'Updated' : match.matching_state.humanize, type: match.resolved? ? :green : :purple)),
+          status: render(TagComponent.new(text: match.action_taken&.humanize || 'No action taken', type: match.resolved? ? :green : :purple)),
           action_needed: action_needed(match),
           match_link: govuk_link_to(match.candidate.email_address, support_interface_ucas_match_path(match)),
           last_update: match.updated_at.to_s(:govuk_date),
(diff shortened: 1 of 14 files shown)


app/components/support_interface/ucas_matches_table_component.rb
```diff
@@ -11,7 +11,7 @@ def initialize(matches:)
     def table_rows
       matches.map do |match|
         {
-          status: render(TagComponent.new(text: match.matching_data_updated? ? 'Updated' : match.matching_state.humanize, type: match.processed? ? :green : :purple)),
+          status: render(TagComponent.new(text: match.action_taken || 'No action taken', type: match.resolved? ? :green : :purple)),
```
AgaDufrat, line 14:
```suggestion
          status: render(TagComponent.new(text: match.action_taken&.humanize || 'No action taken', type: match.resolved? ? :green : :purple)),
```

db/migrate/20201204232014_remove_matching_state_from_ucas_matches.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveMatchingStateFromUCASMatches < ActiveRecord::Migration[6.0]
```
AgaDufrat, line 1: Do we also need to add a migration to update UCAS Matches which are currently in the processed state to either `resolved_on_apply` or `resolved_on_ucas`? Wondering as a lot of them will start flagging as "Action needed" and there we don't have a manual process to mark them as resolved.
I realise it will be difficult to determine where they got resolved 🤔 Especially as for instance we had matches in the previous cycle where statuses were
- RECRUITED on UCAS 
- PENDING CONDITIONS on Apply
They got marked "processed" as the candidates were recruited already there was no point in us contacting them at that stage.

However, how we resolve the matches can impact the statistics. Probably something to discuss with the Transition team.

db/schema.rb
```diff
@@ -526,7 +526,6 @@
   create_table "ucas_matches", force: :cascade do |t|
     t.bigint "candidate_id", null: false
     t.json "matching_data"
-    t.string "matching_state"
```
benswannack, line 529: Can we do this migration in a separate PR to be deployed later? We should change the code so it doesn't depend on the column in this PR, and remove the old column separately

despo: Seperate PR as suggested: https://github.com/DFE-Digital/apply-for-teacher-training/pull/3621

db/migrate/20201204232014_remove_matching_state_from_ucas_matches.rb
despo: PR that introduces `manually_processed` as per our discussion: 
https://github.com/DFE-Digital/apply-for-teacher-training/pull/3616

app/controllers/support_interface/ucas_matches_controller.rb
```diff
@@ -54,12 +54,5 @@ def record_ucas_withdrawal_requested
       flash[:success] = 'The date of requesting withdrawal from UCAS was recorded'
       redirect_to support_interface_ucas_match_path(match)
     end
-
-    def process_match
```
AgaDufrat, line 58: We can also remove the route 🚮 
https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/config/routes.rb#L701

AgaDufrat (approved):

PR #3584, opened 2020-12-04
Allow support users to add a course in support post-submission

TobyRet:
## Context

We've had some support queries come through where candidates have requested adding an additional course to their application post-submission.

This is something that ideally, should be able to be done through the support interface.

- Add button to add a course to candidate's application page 
- Add a new form and autocomplete to allow support users to search for and add a course to an application.

## Guidance to review

- This is a first pass so keen to get feedback on the design. Will add tests once design agreed.

add_a_course

## Link to Trello card

https://trello.com/c/yZ7Ahqvp/2621-dev-%F0%9F%8F%88-allow-support-users-to-add-a-course-in-support-post-submission

## Things to check

### DIFF
--- a/app/forms/support_interface/add_course_form.rb
+++ b/app/forms/support_interface/add_course_form.rb
@@ -1,0 +1,49 @@
+module SupportInterface
+  class AddCourseForm
+    include ActiveModel::Model
+
+    attr_accessor :course_option_id, :application_form_id, :candidate_id
+
+    validates :course_option_id, presence: true
+    validates :application_form_id, presence: true
+
+    DropdownOption = Struct.new(:id, :name)
+
+    def radio_available_courses
+      @radio_available_courses ||= begin
+        courses.order(:name)
+      end
+    end
+
+    def dropdown_available_courses
+      @dropdown_available_courses ||= courses.map { |course|
+        # YUCK
+
+        course.course_options.map do |course_option|
+          DropdownOption.new(course_option.id, "#{course.name} (#{course.code}) – #{course.provider&.name} - #{course_option.site.name}")
+        end
+      }.flatten.sort_by(&:name)
+    end
+
+    def courses
+      @courses ||= Course.current_cycle.includes(:provider, { course_options: [:site] }).open_on_apply
+    end
+
+    def save
+      return false unless valid?
+
+      SupportInterface::AddCourseChoiceAfterSubmission.new(
+        application_form: application_form,
+        course_option: course_option,
+      ).call
+    end
+
+    def application_form
+      @application_form ||= ApplicationForm.find(application_form_id)
+    end
+
+    def course_option
+      @course_option ||= CourseOption.find(course_option_id)
+    end
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -678,8 +678,10 @@ Rails.application.routes.draw do
 
     scope path: '/applications/:application_form_id' do
       get '/' => 'application_forms#show', as: :application_form
+      get '/add-course' => 'add_course#new', as: :add_course_for_candidate
       get '/audit' => 'application_forms#audit', as: :application_form_audit
       get '/comments/new' => 'application_forms/comments#new', as: :application_form_new_comment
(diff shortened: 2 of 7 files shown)


config/routes.rb
```diff
@@ -678,8 +678,10 @@
 
     scope path: '/applications/:application_form_id' do
       get '/' => 'application_forms#show', as: :application_form
+      get '/add-course' => 'add_course#new', as: :add_course_for_candidate
```
duncanjbrown, line 681: For least surprise, how about `GET courses/new` and `POST courses/`?

app/forms/support_interface/add_course_form.rb
```diff
@@ -0,0 +1,49 @@
...
+        # YUCK
+
+        course.course_options.map do |course_option|
+          DropdownOption.new(course_option.id, "#{course.name} (#{course.code}) – #{course.provider&.name} - #{course_option.site.name}")
+        end
+      }.flatten.sort_by(&:name)
+    end
+
+    def courses
+      @courses ||= Course.current_cycle.includes(:provider, { course_options: [:site] }).open_on_apply
```
duncanjbrown, line 29: We've got 4,810 courses in prod, and rising — does this work ok with that sort of volume?

duncanjbrown (commented):
Very excited to see this feature! Couple of minor qs inline

app/forms/support_interface/add_course_form.rb
TobyRet: This is only bringing back courses currently open on Apply but I would imagine we may want to rethink the design entirely if the volume grows significantly

config/routes.rb
TobyRet: Yep these routes need to be changed as you described 👍

app/forms/support_interface/add_course_form.rb
TobyRet: We would probably want to revert to the design / flow used on candidate side for adding a course but question is whether we do this later or now? @davidgisbey @paulrobertlloyd

duncanjbrown: All those 4,810 are open on Apply: https://www.apply-for-teacher-training.service.gov.uk/support/performance/course-stats, and I guess that's multiplied by locations here. We'll need to choose study mode too.

config/routes.rb
duncanjbrown: Lovely, thank you!

frankieroberto:
@TobyRet not bad at all!

A few thoughts:

* "Add a course" – maybe this should be a link instead of a button?
* "Add a course to a submitted application" - I think this ought to include the name of the candidate you're adding a course for, and a back link to their application.
* Autocomplete seems fine for a support UI. Could switch to a search box with a results listing (and radios to choose which one) if there's too many courses that are too similar, or if this was going to be used a lot. But probably overkill immediately?
* "Submit" button could be labelled something like "Add course to application"?
* Add a confirmation banner to the top of the application when you return to it?

I also wondered what the immediate consequence of this was - ie will it immediately trigger a notification email to the provider and/or candidate?  And if so, does it need a confirmation screen to make you double-check you've done the right one?

Would this only appear if the candidate selected less than 3 courses, or are support allowed to override that?

app/forms/support_interface/add_course_form.rb
davidgisbey: Yea I think this was a potential a reason for the idea of inputting a course code and having a list of available course options presented on a new page. Also +1 to the idea of a confirmation page as i hadn't really thought through the implication of incorrectly sending emails to providers.

davidgisbey: The idea of not rebuilding the whole course flow is it's quite a bit of work to add a whole new set of controllers and views. I had to do it before for the post submission journey and it took a substantial amount of time.

TobyRet:
Closing for now.
Unfortunately, the solution proposed in the PR will not scale well on production. @paulrobertlloyd  and @davidgisbey discussed and came up with some other options, however we believe the first step will be to consult Support on these before proceeding.

PR #3587, opened 2020-12-04
Remove remaining app-review styling

paulrobertlloyd:
## Context

In #3512 we replaced `.app-banner` that was used on review pages with a new component with `app-inset-text` (which inherits from `govuk-inset-text`).

Turns out we have another similar set of styles, `app-review`, used for messing around course choices that are no longer available. 

* Replace `app-review` with `app-inset-text`
* Use `data-qa=*` instead of `id="qa-*"`
* Add `app-inset-text__title` class (makes a paragraph bold, adjusts bottom paragraph margin) instead of misleading `govuk-heading-s`
* Use double quotes for HTML attributes 

## Things to check


app/frontend/styles/_inset-text.scss
```diff
@@ -1,17 +1,24 @@
+.app-inset-text {
+  border-left-width: $govuk-border-width;
+}
```
frankieroberto, line 3: I think we should use a modifier for this, rather than just swapping out the `govuk-` prefix for `app-`. Maybe `app-inset-text--narrow`?

app/components/candidate_interface/course_choices_review_component.html.erb
```diff
@@ -62,9 +62,9 @@
         </li>
       </ul>
     <% elsif application_choice.study_mode_full? %>
-      <p class='app-review-warning__primary-message'><%= application_choice.study_mode_full_error %>.</p>
-      <p class='govuk-body'>You can:</p>
-      <ul class='govuk-list govuk-list--bullet'>
+      <p class="govuk-body app-inset-text__title"><%= application_choice.study_mode_full_error %>.</p>
```
frankieroberto, line 65: Potentially we could add the relevant sizing / spacing css to `app-inset-text__title` to avoid the need to also add `govuk-body`? But it's probably fine as-is. I'm never sure about when to nest components...

frankieroberto (changes requested):

app/frontend/styles/_inset-text.scss
paulrobertlloyd: Nice! Have used `.app-inset-text--narrow-border` to be more explicit about what is narrow.

app/components/candidate_interface/course_choices_review_component.html.erb
paulrobertlloyd: Good point. Have updated `.app-inset-text__title` styles, and removed `govuk-body` from these paragraphs.

frankieroberto (approved):

PR #3588, opened 2020-12-04
Automate sending of initial and reminder emails about UCAS matches

AgaDufrat:
## Context

> As a support agent,
> I want emails about UCAS matches to be sent automatically, 
> so I can focus on other support tasks.

Automates the sending the initial and reminder email when a match is detected and records the action taken.

## Guidance to review

Decided to send emails after processing matches instead of doing it in the
clock because if there is a problem with our sync we don’t want to be chasing
candidates in case they have taken an action.

It will be still useful for see how many matches required an action, hence
sending emails after the stack notification.

## Link to Trello card

https://trello.com/c/gfw112Uk/3050-automate-sending-of-initial-emails-and-chaser-emails-about-ucas-matches

## Things to check

### DIFF
--- a/app/components/support_interface/ucas_match_action_component.html.erb
+++ b/app/components/support_interface/ucas_match_action_component.html.erb
@@ -7,7 +7,7 @@
       <p class="govuk-body">
         <%= action_details %>
       </p>
-      <% if @match.action_needed? %>
+      <% if @match.action_needed? && @match.requires_manual_action? %>
         <%= form_with url: button[:path], method: :post do |f| %>
           <%= f.submit button[:text], class: 'govuk-button' %>
         <% end %>
--- a/app/services/support_interface/send_ucas_match_initial_emails_duplicate_applications.rb
+++ b/app/services/support_interface/send_ucas_match_initial_emails_duplicate_applications.rb
@@ -7,7 +7,7 @@ module SupportInterface
     end
 
     def call
-      raise 'UCAS Match initial emails already sent' if ucas_match.initial_emails_sent?
+      raise "Initial emails for UCAS match ##{ucas_match.id} were already sent" if ucas_match.initial_emails_sent?
 
       application_choice = ucas_match.application_choices_for_same_course_on_both_services.first
 
--- a/app/components/support_interface/ucas_match_action_component.rb
+++ b/app/components/support_interface/ucas_match_action_component.rb
@@ -5,16 +5,12 @@ module SupportInterface
     ACTIONS = {
       initial_emails_sent: {
         description: 'Send initial emails',
-        button_text: 'Send emails',
-        form_path: :support_interface_send_initial_emails_path,
-        instructions: 'We need to contact the candidate and the provider.',
+        instructions: 'Emails to the candidate and the provider will be sent automatically.',
         past_tense_description: 'sent the initial emails',
       },
       reminder_emails_sent: {
         description: 'Send a reminder email',
-        button_text: 'Send a reminder email',
-        form_path: :support_interface_send_reminder_email_path,
-        instructions: 'We need to contact the candidate again to remind them they need to take action.',
+        instructions: 'A reminder email to the candidate will be sent automatically.',
         past_tense_description: 'sent the reminder emails',
       },
       ucas_withdrawal_requested: {
--- a/app/controllers/support_interface/ucas_matches_controller.rb
+++ b/app/controllers/support_interface/ucas_matches_controller.rb
@@ -15,36 +15,6 @@ module SupportInterface
       @match = UCASMatch.find(params[:id])
     end
 
-    def send_initial_emails
-      match = UCASMatch.find(params[:id])
-
-      if SupportInterface::SendUCASMatchInitialEmails.new(match).call
-        match.update!(action_taken: 'initial_emails_sent',
-                      candidate_last_contacted_at: Time.zone.now)
-
-        flash[:success] = 'Initial emails sent'
-      else
-        flash[:error] = 'Initial emails were not sent'
-      end
-      redirect_to support_interface_ucas_match_path(match)
-    end
-
-    def send_reminder_email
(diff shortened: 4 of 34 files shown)


spec/services/ucas_matching/send_ucas_match_emails_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+           matching_state: 'new_match',
+           scheme: 'B',
+           ucas_status: :awaiting_provider_decision,
+           action_taken: 'initial_emails_sent',
+           candidate_last_contacted_at: 8.business_days.before(Time.zone.now))
+  end
+
+  describe '#perform' do
+    before do
+      allow(ucas_match_no_action_needed).to receive(:action_needed?).and_return(false)
```
duncanjbrown, line 23: Is this stubbing belt-and-braces given we're specifying all the properties of these objects, above? Better to have a factory?

app/services/support_interface/send_ucas_match_initial_emails_duplicate_applications.rb
```diff
@@ -7,7 +7,7 @@ def initialize(ucas_match)
     end
 
     def call
-      raise 'UCAS Match initial emails already sent' if ucas_match.initial_emails_sent?
+      raise "Initial emails for UCAS match ##{ucas_match.id} was already sent" if ucas_match.initial_emails_sent?
```
duncanjbrown, line 10: Good idea :+1:

app/services/ucas_matching/send_ucas_match_emails.rb
```diff
@@ -16,13 +16,13 @@ def perform
   private
 
     def send_and_record_initial_emails(ucas_match)
-      if SupportInterface::SendUCASMatchInitialEmails.new(ucas_match).call
+      if UCASMatching::SendUCASMatchInitialEmails.new(ucas_match).call
```
duncanjbrown, line 19: Also good — thank you!

app/components/support_interface/ucas_match_action_component.html.erb
```diff
@@ -7,7 +7,7 @@
       <p class="govuk-body">
         <%= action_details %>
       </p>
-      <% if @match.action_needed? %>
+      <% if @match.action_needed? && @match.next_action == :ucas_withdrawal_requested %>
```
duncanjbrown, line 10: Maybe define this as a method on the match — `match.requires_manual_action?` or sth?

app/services/ucas_matching/record_ucas_match_action_taken.rb
```diff
@@ -0,0 +1,13 @@
+module UCASMatching
```
despo, line 1: I think these emails would fit better in `app/services/ucas_matches` as that is where the other emails live and they relate to actions we take in relation to ucas matches, rather than the ucas matching that takes place (despite the service being trigerred there).

In addition to this, we can then drop `ucas_match` from the filename as it's unecessary repetition and instead have e.g. `ucas_matches/record_action_taken.rb`

app/services/ucas_matching/send_ucas_match_initial_emails_duplicate_applications.rb
```diff
@@ -7,7 +7,7 @@ def initialize(ucas_match)
     end
 
     def call
-      raise 'UCAS Match initial emails already sent' if ucas_match.initial_emails_sent?
+      raise "Initial emails for UCAS match ##{ucas_match.id} was already sent" if ucas_match.initial_emails_sent?
```
despo, line 11: small typoe here `was` -> `were`

app/services/ucas_matching/send_ucas_match_initial_emails_multiple_acceptances.rb
```diff
@@ -7,7 +7,7 @@ def initialize(ucas_match)
     end
 
     def call
-      raise 'UCAS Match initial emails already sent' if ucas_match.initial_emails_sent?
+      raise "Initial emails for UCAS match ##{ucas_match.id} was already sent" if ucas_match.initial_emails_sent?
```
despo, line 10: small typo `was` -> `were`

app/components/support_interface/ucas_match_action_component.html.erb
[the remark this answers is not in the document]
AgaDufrat: This component is still used in the Support Interface. I think it could upset our Namespaces spec if I tried to use **UCASMatching**::UCASMatchActionComponent in app/views/**support_interface**/ucas_matches/show.html.erb view.

The services code that moved used to be executed in Support Interface by clicking a button but now it's automated and runs daily. So I think it's ok to keep the view component here.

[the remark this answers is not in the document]
steventux: Ah yes that makes sense :+1:

app/services/ucas_matching/record_ucas_match_action_taken.rb
```diff
@@ -0,0 +1,13 @@
...
+  class RecordUCASMatchActionTaken
+    include Sidekiq::Worker
```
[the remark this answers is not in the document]
AgaDufrat: Are you thinking of executing it immediately with `RecordUCASMatchActionTaken.new.perform` or not making it a Sidekiq Worker?

spec/services/ucas_matching/send_ucas_match_emails_spec.rb
AgaDufrat: Definitely belt-and-braces. I think one of them was flaky and just stubbed them all and forgot to tidy this up 😬 

Following our chat today about the ucas_match factory and updating process_matching_date_spec I think I will just stub it here for now and update the factory and process_matching_date_spec in a separate PR. I identified that we may want to have up to 11 traits.

duncanjbrown (approved):
Thanks @AgaDufrat !

PR #3591, opened 2020-12-04
Format API grade example as a string, not an array

duncanjbrown:
## Context

Some parts of OpenAPI [do support multiple examples](https://swagger.io/docs/specification/adding-examples/) under an "examples:" key, but properties unfortunately don’t. The old code rendered an array in the docs which implied that clients should expect an array, and it also made the autogenerated JSON example incorrect.

Put the extra information in the description and leave a valid example in the example field.

Before

Screenshot 2020-12-04 at 16 59 19

Screenshot 2020-12-04 at 20 46 20

After

Screenshot 2020-12-04 at 16 59 10

Screenshot 2020-12-04 at 20 47 44

### DIFF
--- a/app/views/api_docs/pages/release_notes.md
+++ b/app/views/api_docs/pages/release_notes.md
@@ -1,3 +1,9 @@
+### 7th December 2020
+
+Documentation:
+
+- Clarify the format of the `grade` field in the spec and documentation
+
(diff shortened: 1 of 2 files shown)


despo (approved):
👍

PR #3592, opened 2020-12-04
Remove banner style dependency from incomplete section component (follow up)

paulrobertlloyd:
## Context

In #3512 we updated the messages in the incomplete section component to use `app-inset-text` instead of a variation of `app-banner`. We also changed it so that `incomplete` keyword is used instead of `missing`.

However, the error messages in the summary at the top of the application review page were not updated, meaning they link to incomplete sections using `#missing-*`, not `#incomplete-*`.

* Update links in in error summary to use `#incomplete-*` not `#missing-*`
* Remove a left over `.app-banner--missing-section` style
* Fix test for absence of incomplete application section (not sure how useful this test is… passed even after I changed the value in previous commit 🤔)

## Things to check

### DIFF
--- a/app/frontend/styles/_banner.scss
+++ b/app/frontend/styles/_banner.scss
@@ -21,12 +21,6 @@
 
 .app-banner--warning {
   border-color: govuk-colour("red");
-
-  &.app-banner--missing-section {
-    .app-banner__message p {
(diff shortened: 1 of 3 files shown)


davidgisbey (approved):
lgtm

PR #3593, opened 2020-12-06
Bug fix - Reference is no longer required (guidance)

TobyRet:
## Context

Currently, the message saying a reference is ‘no longer required’ is shown to all referees 😱 

- Update the controller so that the guidance is only shown when the reference has been cancelled.
- Add additional test coverage
- Address (silence?) Bullet gem ‘eager loading’ warning

## Guidance to review

- Please see bug descriptions in the trello ticket and confirm that they are now resolved. 🤞 

## Link to Trello card

https://trello.com/c/EHiwLtjp/2593-%F0%9F%92%942%EF%B8%8F%E2%83%A3-dev-improve-the-referees-journey-when-their-references-arent-required

## Things to check

### DIFF
--- a/app/controllers/referee_interface/reference_controller.rb
+++ b/app/controllers/referee_interface/reference_controller.rb
@@ -131,10 +131,7 @@ module RefereeInterface
     end
 
     def finish
-      @min_feedback_provided = reference
-        .application_form
-        .application_references
-        .minimum_feedback_provided?
+      @reference_cancelled = reference.cancelled?
       @application_form = reference.application_form
     end
 
--- a/app/views/referee_interface/reference/finish.html.erb
+++ b/app/views/referee_interface/reference/finish.html.erb
@@ -6,7 +6,7 @@
       <%= t('page_titles.referee.finish') %>
     </h1>
 
-    <% if @min_feedback_provided %>
+    <% if @reference_cancelled %>
       <p class="govuk-body">
         You do not need to give a reference anymore.
       </p>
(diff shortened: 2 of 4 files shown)


george-bit (approved):
Looks sound 👍

app/services/cancel_referee.rb
```diff
@@ -1,6 +1,6 @@
 class CancelReferee
   def call(reference:)
-    reference.update!(feedback_status: 'cancelled')
+    reference.reload.update!(feedback_status: 'cancelled')
```
davidgisbey, line 3: Not sure if this feels a bit wrong? Bullet being bullet i guess

TobyRet: Yeah I think you're right. 
The Bullet gem is raising an `eager loading` error on the spec when it comes to cancelling a reference. Both @malcolmbaig  and I have tried to investigate the cause of this to no avail -  we think it's possibly a false positive. Reloading makes the Bullet error disappear but then probably not so good we are modifying source code in this way for the sake of a test. The alternative is to disable bullet in the spec concerned and add a comment detailing findings from our investigation. What do you think?

davidgisbey: Yea I had to something similar in this spec candidate_interface/offers_and_withdrawls/candidate_withdraws_spec.rb a while ago.

image

I think not touching source code is preferable.

paulrobertlloyd (approved):

app/services/cancel_referee.rb
TobyRet: Cool - I've disabled Bullet in the spec

PR #3594, opened 2020-12-07
Add region param to ApplicationForm geocoding

stevehook:
## Context

We are seeing some inaccurate geocoding results in the production system for candidate addresses. This seems to happen when the first line of an address matches a place name in North America. In such cases the rest of the address is sufficient to locate the correct coordinates but the API is biasing towards the incorrect match.

Adding the `region` param tells the Google Maps Geocoding API to bias results to the UK. This
makes sense because we only geocode candidates who are UK based.

We will need to re-run the script that geocodes candidate addresses again after this is deployed.

Docs: https://developers.google.com/maps/documentation/geocoding/overview#RegionCodes

## Guidance to review

- No tests for this one since we don't test the geocoder gem.
- I can point to some addresses that didn't work in the production system if needed.

## Link to Trello card

## Things to check

### DIFF
--- a/app/models/application_form.rb
+++ b/app/models/application_form.rb
@@ -3,7 +3,7 @@
 class ApplicationForm < ApplicationRecord
   audited
   has_associated_audits
-  geocoded_by :address_formatted_for_geocoding
+  geocoded_by :address_formatted_for_geocoding, params: { region: 'uk' }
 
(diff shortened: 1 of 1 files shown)


malcolmbaig (approved):
:+1:

PR #3595, opened 2020-12-07
Add Last action taken filter to UCAS Matches page

AgaDufrat:
## Context

Support users and UCAS engagement leads want to see the matches they acted on and what they did.

Screenshot 2020-12-07 at 09 58 22

## Guidance to review

Thoughts on 'No action taken' welcome. Should we exclude it from the filter? It would break in staging because all actions_taken are nil.

## Link to Trello card

https://trello.com/c/J5t3X5sB/3095-filter-ucas-matches-by-last-action-taken

## Things to check

### DIFF
--- a/app/models/support_interface/ucas_matches_filter.rb
+++ b/app/models/support_interface/ucas_matches_filter.rb
@@ -16,11 +16,15 @@ module SupportInterface
         ucas_matches = ucas_matches.where(id: action_needed_ids)
       end
 
+      if applied_filters[:action_taken]
+        ucas_matches = ucas_matches.where(action_taken: applied_filters[:action_taken])
+      end
+
       ucas_matches
     end
 
     def filters
-      @filters ||= [year_filter] + [action_needed_filter]
+      @filters ||= [year_filter] + [action_needed_filter] + [action_taken_filter]
     end
 
   private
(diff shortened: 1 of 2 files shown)


app/models/support_interface/ucas_matches_filter.rb
```diff
@@ -54,5 +58,22 @@ def action_needed_filter
         }],
       }
     end
+
+    def action_taken_filter
+      options = UCASMatch.distinct(:action_taken).pluck(:action_taken).uniq.map do |action|
```
despo, line 63: Do we need `uniq` in this instance, as `distinct` should remove any duplicate entries.

app/models/support_interface/ucas_matches_filter.rb
```diff
@@ -54,5 +58,22 @@ def action_needed_filter
         }],
       }
     end
+
+    def action_taken_filter
```
despo, line 62: Maybe consider extracting this code to its own private method `action_taken_filter_options`

despo (approved):
I've left some suggestions, but happy to approve as it looks good 💯

app/models/support_interface/ucas_matches_filter.rb
AgaDufrat: For some reason it doesn't quite work. I think I will remove the "No action taken" because it's not playing nicely in the filter and it's not that useful. 

<kbd>Screenshot 2020-12-07 at 10 53 37</kbd>

PR #3596, opened 2020-12-07
Fix bug - editing 'other' qualifications

TobyRet:
## Context

- When more than one non-UK qualification has been submitted, problems can occur when editing these. Editing one qualification actually edits the other.
- This is due to a type error and qualification id resolving to nil when reading from Redis

See bug report for full description https://trello.com/c/edEWlzAL/2599-%F0%9F%93%90-dev-structured-qualifications-clean-up-work

Check for qualification id type (String) when reading from Redis

## Guidance to review

- Try and replicate the bug as described in the ticket
- Possibly this is a quick, hacky solution. Should this be made more robust @stevehook ?

## Link to Trello card

https://trello.com/c/edEWlzAL/2599-%F0%9F%93%90-dev-structured-qualifications-clean-up-work

## Things to check

### DIFF
--- a/app/forms/candidate_interface/other_qualification_details_form.rb
+++ b/app/forms/candidate_interface/other_qualification_details_form.rb
@@ -48,7 +48,8 @@ module CandidateInterface
       @intermediate_data_service = intermediate_data_service
       options = @intermediate_data_service.read.merge(options) if @intermediate_data_service
 
-      self.id ||= options[:id]
+      self.id ||= options[:id] || options['id']
+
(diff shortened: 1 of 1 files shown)


davidgisbey (approved):
lgtm, i think we should probs ship this now and think about applying a more robust fix later?

stevehook (approved):
:+1:

Only suggestion I would have is whether `options` should be converted to `with_indifferent_access` somewhere.

But lets deploy as is because it's an important fix.

PR #3599, opened 2020-12-07
Only check for UCAS file download in production

despo:
Fix for [https://trello.com/c/ziLTnhqD/2960-fix-bug-that-stops-import-from-ucas](#3590)


AgaDufrat (approved):

duncanjbrown:
Will this break in Sandbox?

despo:
@duncanjbrown update to only run in production. On production deployment we need to mark the download as having taken place today `UCASMatching::FileDownloadCheck.set_last_sync(Time.zone.now)`.

This way the check will start working tomorrow and inform us if the download has not taken place again.

PR #3600, opened 2020-12-07
Refactor Apply Again checks

malcolmbaig:
We have code that checks that a candidate actually has an unsuccessful
current application before allowing them through the Apply Again
process. Move this code to from the controller into the actual service
that handles Apply Again, and add some tests for this behaviour.

- Move the logical check from controller to service object.
- Add unit tests for the service object.
- Update a system spec to exercise this behaviour.

## Guidance to review
- Is the testing adequate?

## Link to Trello card
https://trello.com/c/gYsowIoQ/2672-2x-apply-again-bug

## Things to check


stevehook (approved):
:+1: Refactor makes sense

PR #3601, opened 2020-12-07
Make the performance dashboard private

duncanjbrown:
## Context

The performance dashboard being public has sparked discussions inside DfE about which data we should publish. Until these are resolved, make the dashboard private to support users.

- the dashboard is no longer public
-  it now lives at `support/performance/dashboard` instead of `integrations/performance-dashboard`


config/routes.rb
```diff
@@ -670,7 +670,6 @@
   namespace :integrations, path: '/integrations' do
     post '/notify/callback' => 'notify#callback'
     get '/feature-flags' => 'feature_flags#index'
-    get '/performance-dashboard' => 'performance_dashboard#dashboard', as: :performance
```
tijmenb, line 673: Can you make this a redirect to the new URL?

spec/system/integrations/service_performance_spec.rb
```diff
@@ -1,19 +1,32 @@
 require 'rails_helper'
```
tijmenb, line 1: Perhaps move this file into `system/support_interface`

app/helpers/navigation_items.rb
```diff
@@ -17,7 +17,7 @@ def for_candidate_interface(current_candidate, _current_controller)
     end
 
     def for_support_primary_nav(current_support_user, current_controller)
-      if current_support_user
+      if current_support_user && !is_active(current_controller, %w[performance_dashboard])
```
tijmenb, line 20: Let's drop this conditional, it's not that important that there's no nav for this page since it's not displayed on TVs anymore (the nav is even quite useful).

fofr:
😞

paulrobertlloyd:
😔  At least we get a nicer URL…

tijmenb (approved):

tvararu:
😢  I was looking for this just now...

PR #3603, opened 2020-12-07
Avoid storing latitude/longitude for addresses outside UK

stevehook:
## Context

There are a few international candidate addresses in the system disguised as UK ones. They follow a pattern: `address_line1-4` specify an international address but they have a UK postcode. Such addresses skew location distance stats because they are typically a long way from the course they are applying to.

We support international addresses so they can be input but we need to know that they are so that we can avoid geocoding them.

Check geo-located coordinates to see whether they fall outside the bounding box for the UK. Store nil latitude/longitude values for such disguised international addresses.

Data source:
https://en.wikipedia.org/wiki/List_of_extreme_points_of_the_United_Kingdom

## Guidance to review

- Does this change make sense?
- Do we need to factor the logic out into a separate service?
- Is there a smarter way to do this check?

## Link to Trello card

## Things to check

### DIFF
--- a/app/workers/geocode_application_address_worker.rb
+++ b/app/workers/geocode_application_address_worker.rb
@@ -7,8 +7,24 @@ class GeocodeApplicationAddressWorker
     return unless Geocoder.config.api_key
 
     application_form = ApplicationForm.find(application_form_id)
-    application_form.latitude, application_form.longitude = application_form.geocode
+    coordinates = application_form.geocode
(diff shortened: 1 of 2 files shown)


davidgisbey (approved):
In Wikipedia we trust? 😅 

Looks good though. I'm guessing we will need to rerun the worker to set values outside of the UK to nil?

The approach seems reasonable to me

PR #3606, opened 2020-12-07
Format govuk_date_and_time and govuk_time according to the style guide

steventux:
## Context

According to the GOVUK style guide we should indicate midday or midnight when presenting dates.

Introduces conditional formatting of `:govuk_date_and_time` and `:govuk_time` to handle this.

## Link to Trello card

https://trello.com/c/9JYg9qnW/3103-govuk-helper-format-times-across-the-service-to-follow-govuk-styles

## Things to check

### DIFF
--- a/config/initializers/date_formats.rb
+++ b/config/initializers/date_formats.rb
@@ -5,5 +5,26 @@ Date::DATE_FORMATS[:govuk_date] = '%-d %B %Y'
 Time::DATE_FORMATS[:month_and_year] = '%B %Y'
 Date::DATE_FORMATS[:month_and_year] = '%B %Y'
 
-Time::DATE_FORMATS[:govuk_date_and_time] = '%e %B %Y at %l:%M%P'
-Time::DATE_FORMATS[:govuk_time] = '%l:%M%P'
+Time::DATE_FORMATS[:govuk_date_and_time] = lambda do |time|
+  format = if time >= time.midday && time <= time.midday.end_of_minute
+             '%e %B %Y at %l%P (midday)'
+           elsif time >= time.midnight && time <= time.midnight.end_of_minute
+             '%e %B %Y at %l%P (midnight)'
+           else
+             '%e %B %Y at %l:%M%P'
+           end
+
+  time.strftime(format)
+end
+
+Time::DATE_FORMATS[:govuk_time] = lambda do |time|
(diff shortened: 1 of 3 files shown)


config/initializers/date_formats.rb
```diff
@@ -5,5 +5,26 @@
 Time::DATE_FORMATS[:month_and_year] = '%B %Y'
 Date::DATE_FORMATS[:month_and_year] = '%B %Y'
 
-Time::DATE_FORMATS[:govuk_date_and_time] = '%e %B %Y at %l:%M%P'
-Time::DATE_FORMATS[:govuk_time] = '%l:%M%P'
+Time::DATE_FORMATS[:govuk_date_and_time] = lambda do |date_and_time|
+  format = if date_and_time == date_and_time.midday
```
mnacos, line 9: Comparing dates can always be a little problematic, as it depends on the precision of the objects compared. Are these down to the second or the millisecond etc. More importantly, these will seldom match, whatever the precision. I'd suggest either relaxing the precision or the comparison to allow 12:00:03 to be counted as midday etc.

spec/config/initializers/date_formats_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'rails_helper'
+
+RSpec.describe 'Time::DATE_FORMATS' do
+  describe ':govuk_date_and_time, :govuk_time formats' do
+    it 'formats time to indicate midday' do
+      date_and_time = Time.zone.local(2020, 12, 25, 12, 0, 0)
```
mnacos, line 6: Following on from my other comment, I'd prefer this to say

```suggestion
      date_and_time = Time.zone.local(2020, 12, 25, 12, 0, 3)
```

and see that the output still says '12pm (midday)'.

mnacos (commented):
Looks very good but will almost never display midday/midnight in its current form. Left suggestions in the comments.

config/initializers/date_formats.rb
steventux: @mnacos good point - amended to compare within the minute of midnight or midday eg. https://github.com/DFE-Digital/apply-for-teacher-training/pull/3606/files#diff-2cd7d9700ea89028447e936ab6d59ebd6187d3ee212db3c1c9791fdd309ff737R9

mnacos (approved):
Great

PR #3607, opened 2020-12-07
Store an offer_changed_at timestamp for ChangeOffer events

mnacos:
## Context

Currently neither the timeline nor the activity log show 'change offer' events, i.e. the modification of `offered_course_option_id` after an initial offer has been made (for either the original course_option_id or a modified one). Multple 'change offer' events can occur for a single application choice, but no such event modifies the application choice's `status`. Adding a timestamp for such events (uniquely associated with 'change offer'), will allow us to detect and track such events, as well as include them in timelines and the activity log.

No UI changes, only backend/database-level ones. I've modified the `ChangeOffer` service to populate `offer_changed_at`, grouping this change with `offered_course_option_id`.

## Guidance to review

Should we also backfill data by detecting such changes in our existing data and altering existing `Audited::Audit` events? _We've decided to do this manually, as we have few changed offers on the system_.

## Link to Trello card

https://trello.com/c/eES1T5BO

## Things to check

- [ ] This code does not rely on migrations in the same Pull Request -- IT DOES!
- [ ] If this code includes a migration adding or changing columns, it also backfills existing records for consistency -- IT WON'T


app/services/set_decline_by_default.rb
```diff
@@ -14,6 +14,7 @@ def call
     # applications have either an offer, been rejected by provider, withdrawn by
     # the candidate, or have the offer withdrawn.
     final_decision_date = [
```
benswannack, line 16: Can we add a spec that covers this behaviour?

benswannack (approved):
Nice :)

app/services/set_decline_by_default.rb
mnacos: We are already have some coverage for this in https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/spec/services/set_decline_by_default_spec.rb#L153 but I've modified the 'offer updated' scenario to use `offer_changed_at`.

PR #3609, opened 2020-12-08
Add edit functionality for date of birth to to support UI

george-bit:
## Context

Adding the ability to update the Date of birth through the support UI. This should allow the support team to change this without dev intervention.

Updated the 

## Guidance to review

Test in the review app to check functionality.

I have started with minimal validation, would be interested to know thoughts on this? Should we carry over all validation from the Candidate Interface or is this unnecessary in the Support UI? 

## Link to Trello card

https://trello.com/c/wMmxTDoE/2684-expand-the-personal-details-review-section-in-the-support-ui-date-of-birth

## Things to check

### DIFF
--- a/app/components/support_interface/personal_details_component.rb
+++ b/app/components/support_interface/personal_details_component.rb
@@ -40,8 +40,8 @@ module SupportInterface
       {
         key: 'Full name',
         value: "#{first_name} #{last_name}",
-        action: 'Change',
-        action_path: support_interface_application_form_edit_applicant_details_path(application_form),
+        action: 'name',
+        change_path: support_interface_application_form_edit_applicant_details_path(application_form),
       }
     end
 
@@ -56,8 +56,8 @@ module SupportInterface
       {
         key: 'Phone number',
         value: phone_number || MISSING,
-        action: 'Change',
-        action_path: support_interface_application_form_edit_applicant_details_path(application_form),
+        action: 'phone number',
+        change_path: support_interface_application_form_edit_applicant_details_path(application_form),
       }
     end
 
@@ -90,8 +90,8 @@ module SupportInterface
       {
         key: 'Date of birth',
         value: application_form.date_of_birth ? application_form.date_of_birth.to_s(:govuk_date) : MISSING,
-        action: 'Change',
-        action_path: support_interface_application_form_edit_applicant_details_path(application_form),
+        action: 'date of birth',
+        change_path: support_interface_application_form_edit_applicant_details_path(application_form),
       }
     end
 
--- a/app/components/support_interface/personal_details_component.rb
+++ b/app/components/support_interface/personal_details_component.rb
@@ -90,6 +90,8 @@ module SupportInterface
       {
         key: 'Date of birth',
         value: application_form.date_of_birth ? application_form.date_of_birth.to_s(:govuk_date) : MISSING,
+        action: 'Change',
+        action_path: support_interface_application_form_edit_applicant_details_path(application_form),
       }
     end
 
--- a/app/views/support_interface/application_forms/applicant_details/edit.html.erb
+++ b/app/views/support_interface/application_forms/applicant_details/edit.html.erb
@@ -7,8 +7,9 @@
       <%= f.govuk_error_summary %>
 
       <%= f.govuk_fieldset legend: { text: 'Edit applicant details', size: 'l' } do %>
-        <%= f.govuk_text_field :first_name, label: { text: 'First Name' } %>
-        <%= f.govuk_text_field :last_name, label: { text: 'Last Name' } %>
+        <%= f.govuk_text_field :first_name, label: { text: 'First name' } %>
+        <%= f.govuk_text_field :last_name, label: { text: 'Last name' } %>
+        <%= f.govuk_date_field :date_of_birth, date_of_birth: true, legend: { text: 'Date of birth', size: 's' } %>
         <%= f.govuk_text_field :phone_number, label: { text: 'Phone number' } %>
         <%= f.govuk_text_field :audit_comment, label: { text: 'Audit log comment' }, hint: { text: 'This will appear in the audit log alongside this change. If the change originated in a Zendesk ticket, paste the Zendesk URL here' } %>
       <% end %>
--- a/app/views/support_interface/application_forms/applicant_details/edit.html.erb
+++ b/app/views/support_interface/application_forms/applicant_details/edit.html.erb
@@ -7,11 +7,11 @@
       <%= f.govuk_error_summary %>
 
       <%= f.govuk_fieldset legend: { text: 'Edit applicant details', size: 'l' } do %>
-        <%= f.govuk_text_field :first_name, label: { text: 'First name' } %>
-        <%= f.govuk_text_field :last_name, label: { text: 'Last name' } %>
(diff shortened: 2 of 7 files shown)


app/views/support_interface/application_forms/applicant_details/edit.html.erb
```diff
@@ -9,6 +9,7 @@
       <%= f.govuk_fieldset legend: { text: 'Edit applicant details', size: 'l' } do %>
         <%= f.govuk_text_field :first_name, label: { text: 'First Name' } %>
         <%= f.govuk_text_field :last_name, label: { text: 'Last Name' } %>
+        <%= f.govuk_date_field :date_of_birth, date_of_birth: true, legend: { text: 'Date of Birth', size: 's' } %>
```
paulrobertlloyd, line 12:
```suggestion
        <%= f.govuk_date_field :date_of_birth, date_of_birth: true, legend: { text: 'Date of birth', size: 's' } %>
```

app/views/support_interface/application_forms/applicant_details/edit.html.erb
```diff
@@ -9,6 +9,7 @@
       <%= f.govuk_fieldset legend: { text: 'Edit applicant details', size: 'l' } do %>
         <%= f.govuk_text_field :first_name, label: { text: 'First Name' } %>
```
paulrobertlloyd, line 10:
```suggestion
        <%= f.govuk_text_field :first_name, label: { text: 'First name' } %>
```

app/views/support_interface/application_forms/applicant_details/edit.html.erb
```diff
@@ -9,6 +9,7 @@
...
         <%= f.govuk_text_field :last_name, label: { text: 'Last Name' } %>
```
paulrobertlloyd, line 11:
```suggestion
        <%= f.govuk_text_field :last_name, label: { text: 'Last name' } %>
```

paulrobertlloyd (changes requested):

paulrobertlloyd:
This is looking great @george-bit! Few comments inline around capitalisation, but I’d actually be tempted to use the same strings (are they localised?) as used in the candidate-facing pages.

We should also include the same hint text and same label sizes, too (I think it’s important support agents see the same information candidates do). All fields should be validated as well – though not sure if audit log one is required; maybe if no value is entered a default is used?

Screenshot 2020-12-08 at 13 07 29

Screenshot 2020-12-08 at 13 07 42

app/components/support_interface/personal_details_component.rb
```diff
@@ -90,6 +90,8 @@ def date_of_birth_row
       {
         key: 'Date of birth',
         value: application_form.date_of_birth ? application_form.date_of_birth.to_s(:govuk_date) : MISSING,
+        action: 'Change',
```
paulrobertlloyd, line 93: All of these change links need to include visually hidden text too. So when read aloud this says ‘Change date of birth’. Think there should be examples of how to do this elsewhere in the code base.

george-bit: Hey Paul, would appreciate if you could show me what you mean by this? had a quick look through the app and not 100% clear what you mean.

paulrobertlloyd: If you look in `SummaryListComponent`, you’ll see that by providing `change_path` instead of `action_path`, then the value for `action` gets shown in the hidden text, and ‘Change’ is shown by default.

https://github.com/DFE-Digital/apply-for-teacher-training/blob/06b0970b6f6157ddd55314cc2d930d08a75547ef/app/components/summary_list_component.html.erb#L19-L25

Here’s an example of this component being called from another view in the app:

```rb
def name_row(reference)
  {
    key: 'Name',
    value: reference.name,
    action: "name for #{reference.name}",
    change_path: candidate_interface_references_edit_name_path(reference.id),
  }
end
```

george-bit: Okay have done so, ready for a re-review

app/components/support_interface/personal_details_component.rb
```diff
@@ -40,8 +40,8 @@ def name_row
       {
         key: 'Full name',
         value: "#{first_name} #{last_name}",
-        action: 'Change',
-        action_path: support_interface_application_form_edit_applicant_details_path(application_form),
+        action: 'Change name',
```
paulrobertlloyd, line 43: Close! ‘Change’ is included in the component, so all you need to provide is a description of what can be changed.

```suggestion
        action: 'name',
```

app/components/support_interface/personal_details_component.rb
```diff
@@ -56,8 +56,8 @@ def phone_number_row
       {
         key: 'Phone number',
         value: phone_number || MISSING,
-        action: 'Change',
-        action_path: support_interface_application_form_edit_applicant_details_path(application_form),
+        action: 'Change phone number',
```
paulrobertlloyd, line 59:
```suggestion
        action: 'phone number',
```

app/components/support_interface/personal_details_component.rb
```diff
@@ -90,6 +90,8 @@ def date_of_birth_row
       {
         key: 'Date of birth',
         value: application_form.date_of_birth ? application_form.date_of_birth.to_s(:govuk_date) : MISSING,
+        action: 'Change date of birth',
```
paulrobertlloyd, line 93:
```suggestion
        action: 'date of birth',
```

paulrobertlloyd (changes requested):

paulrobertlloyd:
@george-bit Almost there! FWIW, I had exactly the same difficulty understanding how the change links in this component work, which is why I can help with an answer. It’s not the clearest or most obvious of things…

paulrobertlloyd:
Sorry, one more thing I noticed; need to provide custom validation messages (shows `can't be blank`):

Screenshot 2020-12-08 at 19 07 24

george-bit:
Hey @paulrobertlloyd , I noticed this while I was adding the validation. The reason I shied away is that the current error messages in the yaml are fairly contextual to the candidate interface and are addressed to the candidate.

Given that this is just the support UI and not user facing are we able to get away with it currently as is? If not should I just copy the error messages from the CI or do we have to tailor them to the support UI?

duncanjbrown:
> Hey @paulrobertlloyd , I noticed this while I was adding the validation. The reason I shied away is that the current error messages in the yaml are fairly contextual to the candidate interface and are addressed to the candidate.

@george-bit you should be able to change the validation errors on this particular form object by changing the keys under `        support_interface/application_forms/edit_applicant_details_form` in `en.yml`

duncanjbrown:
Screenshot 2020-12-08 at 20 46 53

☝️  Let's definitely add YAML keys!

Would be interesting to see if there's a nice way to share validations between this and the equivalent candidate form, but we needn't solve that problem in this PR. Maybe once we mop up Nationality and Address let's review `ApplicantDetailsForm` with a view to making it consistent?

george-bit:
Happy to proceed, @paulrobertlloyd am I okay to copy the existing error messages and modify them to be impersonal. i.e.
"Enter your email address" -> "Enter an email address"

george-bit:
I've updated the yaml file, making appropriate changes :)

paulrobertlloyd (approved):

PR #3610, opened 2020-12-08
Further fixes for inaccurate geocoding results

stevehook:
## Context

Follow up to https://github.com/DFE-Digital/apply-for-teacher-training/pull/3603 to resolve issues with international addresses being geocoded.

- Fix logic in `outside_uk?` method to handle `nil` geocode results.
- Extract `outside_uk?` to `GeocodeFilter` service class
- Use the filter in `geocode.rake`.

## Guidance to review

- Could anything else go wrong with the logic here?

## Link to Trello card

## Things to check

### DIFF
--- a/app/workers/geocode_application_address_worker.rb
+++ b/app/workers/geocode_application_address_worker.rb
@@ -1,7 +1,7 @@
 class GeocodeApplicationAddressWorker
   include Sidekiq::Worker
 
-  sidekiq_options queue: :low_priority
+  sidekiq_options queue: :low_priority, retry: 5
 
   def perform(application_form_id)
     return unless Geocoder.config.api_key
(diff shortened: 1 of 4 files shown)


TobyRet (approved):

app/workers/geocode_application_address_worker.rb
```diff
@@ -8,23 +8,10 @@ def perform(application_form_id)
```
malcolmbaig, line 8: Maybe we should reduce the retry count here as well? Down to 5 perhaps?

stevehook: agreed, I think even 5 is too many, generally if these things fail, they fail forever.

PR #3611, opened 2020-12-08
Allow candidates and providers to use old magic links

tijmenb:
## Context

Currently, the magic link authentication token that we send to candidates is stored as a column on the candidates table. This means they can only have a single valid token at a time. 

If the sign in email is slow to deliver, this can cause a cycle where the candidate requests a sign in email, gets bored, then requests a second magic link, receives the first email, clicks on the link and finds it expired, requests a third email, then clicks on the second email link, etc ad infinitum. The same for the provider fallback sign-in mechanism.
 
The issue was solved for the support sign-in fallback (https://github.com/DFE-Digital/apply-for-teacher-training/pull/2348).

This makes the candidate and providers magic link authentication use the new AuthenticationToken class, which means that if they request a sign-in email multiple times they will be able to use all of the links they've been sent (within an hour).

For candidates, we backfill the existing authentication tokens so the transition is seemless. Once this is merged we can remove the columns from the `candidates` and `provider_users` table.

## Guidance to review

It's quite a bit of code, so please read carefully and comment any doubts or double-takes you had to do. The standard for this code should be really high, so I'd like a lot of reviews and I expect to iterate on this a couple of times.

## Link to Trello card

https://trello.com/c/Gg3i292h/2625-dev-allow-candidates-to-use-old-magic-links-even-if-they-request-a-new-one

## Things to check


app/controllers/candidate_interface/sign_up_controller.rb
```diff
@@ -16,13 +16,13 @@ def create
       @sign_up_form = CandidateInterface::SignUpForm.new(candidate_sign_up_form_params)
 
       if @sign_up_form.existing_candidate?
-        MagicLinkSignIn.call(candidate: @sign_up_form.candidate)
+        CandidateInterface::RequestMagicLink.call(candidate: @sign_up_form.candidate)
```
duncanjbrown, line 19: The `call` convention makes this and `RequestMagicLinkSignup` a bit fiddly to read — maybe have an interface sth like this, and the namespace is bit more cohesive?

```
CandidateInterface::MagicLinks.send_sign_in_link
CandidateInterface::MagicLinks.send_sign_up_link
```

app/mailers/candidate_mailer.rb
```diff
@@ -402,12 +402,12 @@ def email_for_candidate(application_form, args = {})
   end
 
   def candidate_magic_link(candidate)
-    raw_token = candidate.refresh_magic_link_token!
+    raw_token = candidate.create_magic_link_token!
```
duncanjbrown, line 405: I still don't like creating this in the mailer but last time we concluded we didn't want it floating around as an arg in Sidekiq. Could we mitigate that?

app/models/authentication_token.rb
```diff
@@ -1,3 +1,19 @@
...
+  MAX_TOKEN_DURATION = 1.hour
+
   belongs_to :authenticable, polymorphic: true
+
+  def still_valid?
+    created_at > (Time.zone.now - MAX_TOKEN_DURATION)
+  end
+
+  def self.find_by_hashed_token(authenticable_type:, raw_token:)
+    hashed_token = MagicLinkToken.from_raw(raw_token)
```
duncanjbrown, line 11: What's the division of responsibilities between `AuthenticationToken` and `MagicLinkToken` and in practice do we ever have one without the other?

app/models/concerns/authenticable.rb
```diff
@@ -0,0 +1,13 @@
+module Authenticable
+  extend ActiveSupport::Concern
+
+  included do
+    has_many :authentication_tokens, as: :authenticable, dependent: :destroy
```
duncanjbrown, line 5: `:authenticatable`?

app/services/provider_interface/magic_link_authentication.rb
```diff
@@ -3,15 +3,19 @@ module MagicLinkAuthentication
...
-      ProviderUser.where('magic_link_token_sent_at > ?', TOKEN_DURATION.ago)
-        .find_by!(magic_link_token: magic_link_token)
+      authenticable_token = AuthenticationToken.find_by_hashed_token(
+        authenticable_type: 'ProviderUser',
+        raw_token: token,
+      )
+
+      raise ActiveRecord::RecordNotFound unless authenticable_token && authenticable_token.still_valid?
+
+      authenticable_token.authenticable
```
duncanjbrown, line 18: Could we rename this to `authenticatable` too?

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_12_03_144318) do
+ActiveRecord::Schema.define(version: 2020_12_08_105326) do
```
duncanjbrown, line 13: Think this belongs in the prior commit

app/services/provider_interface/magic_link_authentication.rb
```diff
@@ -3,15 +3,19 @@ module MagicLinkAuthentication
     TOKEN_DURATION = 1.hour
```
duncanjbrown, line 3: Do we still need this?

duncanjbrown (commented):
Approach looks sound, few qs inline!

app/controllers/candidate_interface/sign_in_controller.rb
```diff
@@ -12,38 +12,48 @@ def create
...
-      if token_not_expired
+      if authentication_token && authentication_token.still_valid?
+        candidate = authentication_token.authenticable
         flash[:success] = t('apply_from_find.account_created_message') if candidate.last_signed_in_at.nil?
         sign_in(candidate, scope: :candidate)
-        add_identity_to_log candidate.id
-        candidate.update_sign_in_fields!
+        add_identity_to_log(candidate.id)
+        candidate.update!(last_signed_in_at: Time.zone.now)
+        authentication_token.destroy!
```
davidgisbey, line 50: I'm wondering if it's worth just having an extra column for this model like 'used' which we set to true after candidate signs in, rather than deleting it. That way we can run queries really easily without having to use audits. For example, what % of links get clicked would just be:

`ApplicationToken.where(used: true).where(authenticable_type: 'Candidate') / ApplicationToken.token.where(authenticable_type: 'Candidate') `

tijmenb: Yes, good idea! I'll make a follow up PR for this because it will require the system tests to be updated, and I don't want to change both behaviour and tests in the same PR (so we can be sure this PR doesn't change core bahaviour).

malcolmbaig:
Approach looks good, nothing to add beyond what @duncanjbrown has commented on 👍🏽

app/models/concerns/authenticable.rb
```diff
@@ -0,0 +1,13 @@
+module Authenticable
```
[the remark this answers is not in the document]
tijmenb: Maybe `AuthenticatedUsingMagicLinks`?

app/controllers/candidate_interface/sign_up_controller.rb
tijmenb: Agree 👍

Perhaps this?

```
CandidateInterface::RequestMagicLink.for_sign_in
CandidateInterface::RequestMagicLink.for_sign_up
```

app/mailers/candidate_mailer.rb
tijmenb: I think we need to rethink that because it's in the email log now anyway, let's look at that next year.

Card: https://trello.com/c/GDGIVX65/399-draft-revisit-assumptions-about-security

app/models/concerns/authenticable.rb
tijmenb: Maybe just `user`?

app/controllers/candidate_interface/sign_in_controller.rb
```diff
@@ -12,38 +12,48 @@ def create
...
 
         redirect_to candidate_interface_interstitial_path(path: params[:path])
+      elsif params[:u]
```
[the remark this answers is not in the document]
davidgisbey: Yea it should only be hit with a u id in the params. Have updated.

app/models/authentication_token.rb
davidgisbey: As far as i can tell we won't ever have one without the other.

app/controllers/candidate_interface/sign_up_controller.rb
davidgisbey: Seems good to me. Have updated.

app/models/concerns/authenticable.rb
davidgisbey: have updated to use `user`

app/services/provider_interface/magic_link_authentication.rb
davidgisbey: is now user

app/models/concerns/authenticable.rb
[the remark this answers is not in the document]
davidgisbey: have updated

app/services/provider_interface/magic_link_authentication.rb
davidgisbey: Sorry, I think I'm missing something. Why don't we need this?

db/schema.rb
davidgisbey: Have sorted

db/migrate/20201215200431_change_authenticable_id_to_user_id.rb
```diff
@@ -0,0 +1,6 @@
+class ChangeAuthenticableIdToUserId < ActiveRecord::Migration[6.0]
+  def change
+    rename_column :authentication_tokens, :authenticable_id, :user_id
```
TobyRet, line 3: @davidgisbey We may get some issues (inserting into non existent column) if users are authenticating post migration but before before the rest of the logic is live?

app/controllers/candidate_interface/sign_in_controller.rb
```diff
@@ -12,38 +12,47 @@ def create
...
+        raw_token: params[:token],
+      )
 
-      redirect_to(action: :new) and return if candidate.nil?
-
-      if FindCandidateByToken.token_not_expired?(candidate)
+      if authentication_token && authentication_token.still_valid?
         render 'confirm_authentication'
-      else
+      elsif params[:u]
```
TobyRet, line 26: Not necessarily for this PR but might make code more readable if `:u` was renamed `:user_id` ?

spec/system/candidate_interface/signup_and_signin/candidate_signs_up_using_magic_link_with_an_invalid_token_spec.rb
```diff
@@ -54,7 +53,7 @@ def then_i_receive_an_email_inviting_me_to_sign_up
   end
 
   def when_the_magic_link_token_is_overwritten
-    Candidate.find_by(email_address: @email).update(magic_link_token: MagicLinkToken.new.raw)
+    Candidate.find_by(email_address: @email).authentication_tokens.delete_all
```
TobyRet, line 56: Do we need a system spec for signing in with old (but valid) magic links?

app/controllers/candidate_interface/sign_in_controller.rb
davidgisbey: Once this is merged we can remove all the user id related code as we will have associated the token with the user

db/migrate/20201215200431_change_authenticable_id_to_user_id.rb
davidgisbey: Yea i'll merge and deploy out the migration separately once we're happy that's the way to go. Good shout.

app/controllers/candidate_interface/sign_in_controller.rb
```diff
@@ -12,38 +12,47 @@ def create
...
+        # If the token is invalid or expired, redirect the user to a page
+        # with their encryped user id as a param where they can request
```
raamSoftwire, line 28: Typo. `Encrypted`

spec/system/candidate_interface/signup_and_signin/candidate_signs_up_using_magic_link_with_an_invalid_token_spec.rb
davidgisbey: Have added one 👍

app/controllers/candidate_interface/sign_in_controller.rb
davidgisbey: Good spot

davidgisbey (approved):
Toby and me have both agreed with think this is good to go 👍

PR #3612, opened 2020-12-08
Correctly display optional grades

raamSoftwire:
## Context

If a grade is shown as optional on the form, it should be shown as ‘Grade (optional)’ on the summary card, too. We should also show ‘Not entered’ if field was left blank.

- If a grade is not entered, make sure it is saved as `nil` rather than an empty string
- Make sure the "optional" label is shown when required
- Show "not entered" for optional grades if they are not present

## Guidance to review

I have updated the candidate review page and the support UI. Is there anywhere else that might be affected? 

By saving the grade as `nil` rather than an empty string, it meant we could avoid checking for `""` and made the code much neater. But will this change of how we store the data have repercussions  elsewhere?

## Link to Trello card

https://trello.com/c/edEWlzAL/2599-%F0%9F%93%90-dev-structured-qualifications-clean-up-work

## Things to check

### DIFF
--- a/app/components/candidate_interface/other_qualifications_review_component.rb
+++ b/app/components/candidate_interface/other_qualifications_review_component.rb
@@ -115,7 +115,7 @@ module CandidateInterface
     end
 
     def grade_set_key(qualification)
-      if qualification.qualification_type == 'non_uk'
+      if qualification.qualification_type == 'non_uk' || (qualification.qualification_type == 'Other' && qualification.qualification_type_name != 'BTEC')
         t('application_form.other_qualification.grade.optional_label')
       else
         t('application_form.other_qualification.grade.label')
--- a/app/components/qualification_grade_component.rb
+++ b/app/components/qualification_grade_component.rb
@@ -8,7 +8,7 @@ class QualificationGradeComponent < ViewComponent::Base
     if @qualification.predicted_grade?
       "#{@qualification.grade} (predicted)"
     else
-      @qualification.grade
+      @qualification.grade || 'Not entered'
     end
   end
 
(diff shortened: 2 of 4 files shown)


app/components/qualification_grade_component.rb
```diff
@@ -8,7 +8,7 @@ def grade
     if @qualification.predicted_grade?
       "#{@qualification.grade} (predicted)"
     else
-      @qualification.grade
+      @qualification.grade || 'Not entered'
```
stevehook, line 11:
```suggestion
      @qualification.grade.presence || 'Not entered'
```
because there will be some existing qualifications that have an empty string value?

stevehook (approved):
Tested and works well. Made one suggestion because I think we need to cater for empty non-nil values entered to date.

app/components/candidate_interface/other_qualifications_review_component.rb
```diff
@@ -115,7 +115,7 @@ def grade_row(qualification)
     end
 
     def grade_set_key(qualification)
-      if qualification.qualification_type == 'non_uk'
+      if qualification.qualification_type == 'non_uk' || (qualification.qualification_type == 'Other' && qualification.qualification_type_name != 'BTEC')
```
TobyRet, line 118: Should the corresponding spec be updated in this instance?

app/components/qualification_grade_component.rb
TobyRet: Was not aware of `#presence` - good to know 👍

raamSoftwire: Good thinking Steve, thanks for raising that.

app/components/candidate_interface/other_qualifications_review_component.rb
raamSoftwire: Yep, I've just added a test for this. Cheers Toby

PR #3613, opened 2020-12-08
Bug fix - editing  'other' qualifications

TobyRet:
## Context

When a user has more than one 'other' qualification and goes on to edit one of them, it's possible for old state stored in Redis to be merged into that qualification. This happens when clicking the 'back' link.

_"I have added 2 qualifications, an UK one (Scottish Advanced Higher Maths), and a non-UK one (German High school diploma Maths). When I edit the non-UK qualification, I am shown the edit page for the Scottish Advanced Higher Maths qualification, not the non-uk German qualification."_

1. Make sure that the Redis state is wiped clean before editing individual qualifications. If not then when we come to editing a
    qualification, it may have merged state from another qualification. A modified back link is now used on the qualification details form that ensures a hard refresh occurs, which then triggers clearing of the state by the controller.

2. This PR also fixes a bug that occurs when changing an 'other' qualification type from 'non_uk' to 'uk'. We now ensure that  non-relevant fields are cleared (set to `nil`) when changing a qualification type 'non-uk' to 'uk'.

## Guidance to review

- Add two or more 'other' qualifications to an application. Then go an edit the types, subjects, grades etc for each. Should now behave as expected.
- Hopefully this all makes sense? Do you think a feature spec or more documentation / comments are required in this instance? 

## Link to Trello card

https://trello.com/c/BqTT5ptc/2675-bug-edit-other-qualifications

## Things to check

### DIFF
--- a/app/forms/candidate_interface/other_qualification_type_form.rb
+++ b/app/forms/candidate_interface/other_qualification_type_form.rb
@@ -13,6 +13,13 @@ module CandidateInterface
     attr_reader :next_step
     attr_accessor :editing, :id, :current_step
 
+    attr_accessor :subject
+    attr_accessor :institution_country
+    attr_accessor :choice
+    attr_accessor :award_year
+    attr_accessor :predicted_grade
+    attr_accessor :grade
+
     attr_accessor :qualification_type
     attr_accessor :other_uk_qualification_type
     attr_accessor :non_uk_qualification_type
@@ -26,6 +33,12 @@ module CandidateInterface
       @current_application = current_application
       @intermediate_data_service = intermediate_data_service
       options = @intermediate_data_service.read.merge(options.select { |_, value| value.present? }) if @intermediate_data_service
+
+      if options && [A_LEVEL_TYPE, AS_LEVEL_TYPE, GCSE_TYPE].include?(options['qualification_type'])
+        options['non_uk_qualification_type'] = nil
+        options['institution_country'] = nil
+      end
+
       super(options)
     end
 
@@ -39,7 +52,7 @@ module CandidateInterface
       application_qualification.update!(attributes_for_persistence)
     end
 
-    PERSISTENT_ATTRIBUTES = %w[id current_step editing qualification_type other_uk_qualification_type non_uk_qualification_type].freeze
+    PERSISTENT_ATTRIBUTES = %w[id current_step editing qualification_type other_uk_qualification_type non_uk_qualification_type institution_country].freeze
 
   private
 
--- a/app/views/candidate_interface/other_qualifications/details/edit.html.erb
+++ b/app/views/candidate_interface/other_qualifications/details/edit.html.erb
@@ -1,5 +1,5 @@
 <% content_for :title, title_with_error_prefix(t('page_titles.qualification_details'), @form.errors.any?) %>
-<% content_for :before_content, govuk_back_link_to %>
+<% content_for :before_content, govuk_back_link_to(candidate_interface_review_other_qualifications_path) %>
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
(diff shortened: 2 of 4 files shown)


app/forms/candidate_interface/other_qualification_type_form.rb
```diff
@@ -58,5 +70,16 @@ def intermediate_state
     def qualification_type_changed?
       id && ApplicationQualification.find(id)&.qualification_type != qualification_type
     end
+
+    def prepare(options)
+      options.select { |_, value| value.present? }
```
stevehook, line 75: Does this line do anything? (it doesn't mutate `options`)

app/forms/candidate_interface/other_qualification_type_form.rb
```diff
@@ -25,7 +32,12 @@ class OtherQualificationTypeForm
     def initialize(current_application = nil, intermediate_data_service = nil, options = nil)
       @current_application = current_application
       @intermediate_data_service = intermediate_data_service
-      options = @intermediate_data_service.read.merge(options.select { |_, value| value.present? }) if @intermediate_data_service
+
+      options = if @intermediate_data_service
+                  prepared_options = prepare(options)
+                  @intermediate_data_service.read.merge(prepared_options)
+                end
```
stevehook, line 39: this will reset `options` to nil if `@intermediate_data_service` is nil. Is that intended?

stevehook (commented):
I tried hard to break this and it's all working much better. The only thing that caused problems was using the browser back button.

A couple of queries inline, otherwise I'm tempted to say we should merge this because it's a big improvement. What do you think?

davidgisbey:
Sorry Toby this has turned into such a can of worms.

image

image

image


Some attributes are being persisted that shouldn't due to the way we are merging attributes. I think we need to do something like this.

image

app/views/candidate_interface/other_qualifications/details/edit.html.erb
```diff
@@ -1,5 +1,5 @@
 <% content_for :title, title_with_error_prefix(t('page_titles.qualification_details'), @form.errors.any?) %>
-<% content_for :before_content, govuk_back_link_to %>
+<% content_for :before_content, govuk_back_link_to(refresh: true) %>
```
frankieroberto, line 2: Could we pass in the path to the previous page here, rather than relying on javascript?

app/forms/candidate_interface/other_qualification_type_form.rb
TobyRet: Yep this is an error on my part - have fixed

app/forms/candidate_interface/other_qualification_type_form.rb
TobyRet: Yep this is an error on my part - have fixed

app/views/candidate_interface/other_qualifications/details/edit.html.erb
TobyRet: Yep that seems to work ok - have made the changes 👍

TobyRet:
> Sorry Toby this has turned into such a can of worms.

> Some attributes are being persisted that shouldn't due to the way we are merging attributes. I think we need to do something like this.

Fixed. That should work ok for you now!

TobyRet:
> I tried hard to break this and it's all working much better. The only thing that caused problems was using the browser back button.

@frankieroberto - as @stevehook this bug still persists if the user users the browser back button. Any thoughts on what could be done here? I agree with Steve and think we should deploy what has been done so far. I can create a new ticket for the browser back button. This issue will come up again if we roll out the intermediate data service to other forms.

CC: @davidgisbey

frankieroberto:
One thing I've done when storing incomplete state for a multi-question form is to store it under a randomly-generated ID (as well as the section name), and persist that ID through the forms (either in the URL or as hidden fields). That way, you don't end up mixing up the data, eg if going through the same questions in different browser tabs, or pressing back.

stevehook (approved):
:+1:

PR #3616, opened 2020-12-09
Add manually resolved action_taken to UCASMatch

despo:
## Context

Introduced new action_taken so that previous recruitment cycle data can be migrated to it before `matching_state` is removed.

https://github.com/DFE-Digital/apply-for-teacher-training/pull/3583

After this PR is deployed to production we need to run the following code.

```ruby
UCASMatch.where(matching_state: 'processed', recruitment_cycle_year: 2020).each do |match|
   next unless match.dual_application_or_dual_acceptance?

   match.update(candidate_last_contacted_at: match.updated_at)
   match.manually_resolved!
end
```

### DIFF
--- a/app/models/ucas_match.rb
+++ b/app/models/ucas_match.rb
@@ -15,6 +15,7 @@ class UCASMatch < ApplicationRecord
     ucas_withdrawal_requested: 'ucas_withdrawal_requested',
     resolved_on_apply: 'resolved_on_apply',
     resolved_on_ucas: 'resolved_on_ucas',
+    manually_resolved: 'manually_resolved',
   }
 
   def ready_to_resolve?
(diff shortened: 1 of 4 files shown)


app/models/ucas_match.rb
```diff
@@ -30,6 +31,8 @@ def trackable_applicant_key
   end
 
   def action_needed?
+    return false if manually_resolved?
```
AgaDufrat, line 34: Could we do
```
  def resolved?
    %w[resolved_on_apply resolved_on_ucas manually_resolved].include?(action_taken)
  end
```
instead of adding a clause here.

spec/models/ucas_match_spec.rb
```diff
@@ -13,6 +13,12 @@
   end
 
   describe '#action_needed?' do
+    it 'returns false if ucas match has been manually resolved', wip: true do
```
duncanjbrown, line 16: What's `wip: true`?

despo: Oops I missed that! It's how I tag tests when developing (work in progress) so that I don't need to wait for seven minutes every time I do a test run.

duncanjbrown (approved):

AgaDufrat (approved):

PR #3617, opened 2020-12-09
Candidate feedback sus survey fix

raamSoftwire:
## Context

We want to move the satisfaction survey back out of emails and in to the service. 

The previous work to move it from the service and in to emails was done as part of https://github.com/DFE-Digital/apply-for-teacher-training/pull/3321.

- Add the link to the feedback form back in to the service
- Remove the link from the email
- Update tests and tidy up

## Guidance to review

There seem to be two forms `candidate_interface_feedback_form_path` and `candidate_interface_satisfaction_survey_recommendation_path`. From the current code it looks like the first one is the one we want, but I thought it worth checking here.

Should this be behind the `feedback_form` flag? Is that flag still useful?

The previous PR set up the ability to `redirect_to_path_if_path_params_are_present_and_valid`. That functionality seems more general and potentially useful and so I have left it in. Do you agree?

## Link to Trello card

https://trello.com/c/jrsGBC9X/2681-candidate-feedback-sus-survey-fix

## Things to check

### DIFF
--- a/app/views/candidate_interface/submitted_application_form/submit_success.html.erb
+++ b/app/views/candidate_interface/submitted_application_form/submit_success.html.erb
@@ -22,10 +22,12 @@
     <p class="govuk-body">You can track or withdraw your application from your application dashboard.</p>
     <p class="govuk-body"><%= govuk_link_to 'To view your application, return to your application dashboard', candidate_interface_application_complete_path %>.</p>
 
+    <h2 class="govuk-heading-m">Tell us what you think of this service</h2>
+    <p class="govuk-body">Your feedback will help us improve.</p>
     <% if FeatureFlag.active?('feedback_form') %>
-      <h2 class="govuk-heading-m">Tell us what you think of this service</h2>
-      <p class="govuk-body">Your feedback will help us improve.</p>
       <%= govuk_button_link_to 'Give feedback', candidate_interface_feedback_form_path %>
+      <%else %>
+      <%= govuk_button_link_to 'Give feedback', candidate_interface_satisfaction_survey_recommendation_path %>
     <% end %>
   </div>
 </div>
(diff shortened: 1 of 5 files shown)


app/views/candidate_interface/submitted_application_form/submit_success.html.erb
```diff
@@ -22,9 +22,11 @@
...
 
-    <% unless FeatureFlag.active?('feedback_form') %>
-      <h2 class="govuk-heading-m">Tell us what you think of this service</h2>
-      <p class="govuk-body">Your feedback will help us improve.</p>
+    <h2 class="govuk-heading-m">Tell us what you think of this service</h2>
+    <p class="govuk-body">Your feedback will help us improve.</p>
+    <% if FeatureFlag.active?('feedback_form') %>
+      <%= govuk_button_link_to 'Give feedback', candidate_interface_feedback_form_path %>
+      <%else %>
       <%= govuk_button_link_to 'Give feedback', candidate_interface_satisfaction_survey_recommendation_path %>
```
davidgisbey, line 30: I think this can come out now. We shouldn't be using this survey at all going forward. I'll card up something to remove it entirely.

raamSoftwire: Is this what you were thinking? Or should we remove the feature flag now too?

```    <% if FeatureFlag.active?('feedback_form') %>   
      <h2 class="govuk-heading-m">Tell us what you think of this service</h2>
      <p class="govuk-body">Your feedback will help us improve.</p>
      <%= govuk_button_link_to 'Give feedback', candidate_interface_feedback_form_path %>
    <% end %>
```

davidgisbey (approved):
lgtm.

Yes that's the correct form to use

Should this be behind the feedback_form flag? Is that flag still useful?

No i think this flag and the satisfaction survey can come out. I'll card it up

The previous PR set up the ability to redirect_to_path_if_path_params_are_present_and_valid. That functionality seems more general and potentially useful and so I have left it in. Do you agree?

Yea I think we should leave this for now 👍

app/views/candidate_interface/submitted_application_form/submit_success.html.erb
davidgisbey: I think we can remove the flag entirely. It's pretty old now and we know the form works as intended.

PR #3618, opened 2020-12-09
Display accredited body on applications in support

duncanjbrown:
## Context

This information is useful when you're chasing down an application following a support request by the accredited body.

- Turn the ApplicationChoice summary into a component
- Add a line to it

## Link to Trello card

Big enough to be annoying when you're on support but too small for a card?


george-bit (approved):
LGTM 👍

PR #3620, opened 2020-12-09
Add csv export for application data

benswannack:
Moved some logic into a common decorator to be used between the API and the new export

## Context
Providers sometimes want to take their applications off the service into their own system, and this export will allow them to get the relevant info

Adds new application data export.

This is not currently used, but in the interest of keeping PRs small, the UI work to hook this up will be done separately. This work will also include hooking up the CSV docs so providers can view them

## Link to Trello card
https://trello.com/c/9l90TE1b/2970-%F0%9F%8F%88let-users-export-application-data

## Things to check


app/services/application_data_service.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def self.missing_gcses_explanation(application_form:)
+    gcses = application_form.application_qualifications.select do |q|
+      q.level == 'gcse'
+    end
+
+    explanation = gcses
+                    .select(&:missing_qualification?)
+                    .map { |gcse| "#{gcse.subject.capitalize} GCSE or equivalent: #{gcse.missing_explanation}" }
+                    .join(',')
```
duncanjbrown, line 21: The `join` has changed from newlines to a comma here.

- we probably ought to retain newlines in the API for consistency
- do we want something harder separating them, given explanations may contain commas

app/services/application_data_service.rb
```diff
@@ -0,0 +1,45 @@
+class ApplicationDataService
```
duncanjbrown, line 1: I like this thing. Did you consider wrapping an ApplicationChoice in it, and extending that object's interface? See conversation with @tijmenb here: https://github.com/DFE-Digital/apply-for-teacher-training/pull/3156. Then instantiate one of these at the top of SingleApplicationPresenter, ApplicationDataExport etc?

app/exports/provider_application_data_export.yml
```diff
@@ -0,0 +1,185 @@
+common_columns:
```
benswannack, line 1: Files now removed. If we decide to use them, it will be in another PR

app/services/provider_interface/application_data_export.rb
```diff
@@ -0,0 +1,66 @@
...
+          'subject' => first_degree&.subject,
+          'grade' => first_degree&.grade,
+          'start_year' => first_degree&.start_year,
+          'award_year' => first_degree&.award_year,
+          'institution_details' => first_degree&.institution_name,
+          'equivalency_details' => first_degree&.composite_equivalency_details,
+          'awarding_body' => first_degree&.awarding_body,
+          'gcse_qualifications_summary' => application.gcse_qualifications_summary,
+          'missing_gcses_explanation' => application.missing_gcses_explanation,
+          'disability_disclosure' => application.application_form.disability_disclosure,
```
benswannack, line 58: Does this field (or any other while we're here) come under any permission restrictions?

duncanjbrown: No — just diversity and safeguarding, neither of which are present here

app/decorators/application_choice_export_decorator.rb
```diff
@@ -0,0 +1,47 @@
+class ApplicationChoiceExportDecorator < SimpleDelegator
```
steventux, line 1: I wonder about introducing `app/decorators/` (especially in the neutral namespace) to the project for the purpose of application exports, this seems like a provider-centric feature. 
I also wonder if it would be tidier to move this to `app/presenters/provider_interface/application_choice_export_presenter` or similar and adhere to the current pattern of services and presenters (and in this case sub-presenters). I like the delegation to `ApplicationChoice` but this could apply to a presenter too. Very much an opinion point, happy to discuss further.

app/decorators/application_choice_export_decorator.rb
```diff
@@ -0,0 +1,47 @@
...
+      .compact
+      .join(',')
+
+    summary_string.presence
+  end
+
+  def missing_gcses_explanation(separator_string: ',')
+    application_form
+      .application_qualifications
+      .select { |q| q.level == 'gcse' }
```
steventux, line 16: We could presumably do this with ActiveRecord, though you might have to mitigate lazy loading.

app/decorators/application_choice_export_decorator.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+private
+
+  def missing_gcse_explanation(gcse)
+    "#{gcse.subject.capitalize} GCSE or equivalent: #{gcse.missing_explanation}"
+  end
+
+  def summary_for_gcse(gcse)
+    "#{gcse.qualification_type.humanize} #{gcse.subject.capitalize}, #{gcse.grade}, #{gcse.start_year}-#{gcse.award_year}" if gcse.present?
```
steventux, line 45: Could short-circuit with `return if gcse.blank?` and reduce the line length.

steventux (changes requested):
Looks good, I've added a question about the decorator added in this PR as it seems a bit incongruous.

steventux (approved):

app/decorators/application_choice_export_decorator.rb
benswannack: Yeah that's true, the main reason for this is that the export was asking for things that were the same as in the API. It feels important to enforce that those things are exactly the same.

PR #3621, opened 2020-12-09
UCASMatch#matching state removal migration

despo:
Waiting on #3583 being merged


steventux:
@despo I think running this migration will also change https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/docs/domain-model.png via Graphviz, if so the change should be included in this PR.

AgaDufrat (approved):

steventux (approved):

PR #3622, opened 2020-12-09
Include paragraph detailing RBD date when a user cannot make decisions

steventux:
## Context

We would like to draw user's attention to the reject by default date even if they do not have the ability to make decisions on an application.

- Adds content to the inset text box above the application details which explains when the application will be rejected by default.
- Refactors some of the conditional logic in the `ApplicationChoiceHeaderComponent` markup.

image

## Link to Trello card

https://trello.com/c/w7zwEUju/3110-rbd-missing-from-applications-in-manage#comment-5fd0ae1460109a548d50204c

## Things to check

### DIFF
--- a/app/components/provider_interface/application_choice_header_component.rb
+++ b/app/components/provider_interface/application_choice_header_component.rb
@@ -56,5 +56,28 @@ module ProviderInterface
     def offer_present?
       ApplicationStateChange::OFFERED_STATES.include?(application_choice.status.to_sym)
     end
+
+    def respond_to_application?
+      provider_can_respond && application_choice.awaiting_provider_decision?
+    end
+
+    def deferred_offer_wizard_applicable?
+      provider_can_respond && deferred_offer_wizard_applicable
+    end
+
+    def rejection_reason_required?
+      provider_can_respond && rejection_reason_required
+    end
+
+    def provider_cannot_respond?
+      !provider_can_respond && application_choice.awaiting_provider_decision?
+    end
+
+    def show_inset_text?
+      flash.empty? && (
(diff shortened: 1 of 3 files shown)


app/components/provider_interface/application_choice_header_component.rb
```diff
@@ -56,5 +56,25 @@ def rejection_reason_required
...
+    def rejection_reason_required?
+      flash.empty? && provider_can_respond && rejection_reason_required
+    end
+
+    def show_reject_by_default_text?
+      flash.empty? && !provider_can_respond && application_choice.awaiting_provider_decision?
+    end
+
+    def show_inset_text?
+      respond_to_application? || deferred_offer_wizard_applicable? || rejection_reason_required? || show_reject_by_default_text?
```
duncanjbrown, line 77: Should we move `flash.empty? && ...` into this method?

app/components/provider_interface/application_choice_header_component.rb
```diff
@@ -56,5 +56,25 @@ def rejection_reason_required
...
+
+    def deferred_offer_wizard_applicable?
+      flash.empty? && provider_can_respond && deferred_offer_wizard_applicable
+    end
+
+    def rejection_reason_required?
+      flash.empty? && provider_can_respond && rejection_reason_required
+    end
+
+    def show_reject_by_default_text?
```
duncanjbrown, line 72: This method name is slightly inconsistent with the above — it describes the behaviour, not the condition. Maybe `provider_cannot_respond?` or sth?

duncanjbrown (approved):

PR #3627, opened 2020-12-10
Strip whitespace from candidate input

malcolmbaig:
## Context

We're seeing quite a few validation errors coming from candidates leaving whitespace on certain fields. 

For example, if a candidate copy and pastes a referee's email address and leaves some whitespace, it triggers a validation.

- Add the [strip_attributes](https://github.com/rmm5t/strip_attributes) gem 
- Expose its functionality in controllers via a `strip_whitespace` method
- Use this method in candidate controllers wherever it looks like we're handling text input from candidates

### Why this gem?

- It appears to be the most popular and well-maintained of the handful of gems that deal with stripping whitespace. 
- Its stripping functionality handles [several kinds](https://github.com/rmm5t/strip_attributes/blob/master/lib/strip_attributes.rb#L22) of whitespace in addition to regular spaces.
- Its stripping functionality is easy to invoke directly rather than relying on ActiveModel/ActiveRecord integration.

### Why strip whitespace in controllers?

The other alternative would have been to do this in form objects, in a `before_validation` callback. I opted for controllers as it's the simpler of the two options. Stripping in form objects would have required converting every form that needs this functionality over to the Attributes API (instead of using `attr_accessor`), which is a much more invasive change in comparison without being obviously better. We would also need to write our own macro for stripping attributes, as the gem only integrates well with ActiveRecord not plain ActiveModel objects.

With the approach outlined here, the default technique is to `strip_whitespace` on any params that we pass in to form objects. If we need to do something slightly more complex (like only strip a subset of the params), we have the option of handling that inside the form object by invoking the `StripWhitespace` helper class directly.

## Guidance to review
- Per-commit highly recommended. 
- The controller updates are in two distinct commits:
  - replacing cases where we were stripping regular spaces already
  -  other controllers where it looks like we need to strip candidate input

## Link to Trello card
https://trello.com/c/FHwkTgbp

## Things to check

### DIFF
--- a/app/lib/strip_whitespace.rb
+++ b/app/lib/strip_whitespace.rb
@@ -1,0 +1,9 @@
+class StripWhitespace
+  def self.from_hash(hash)
+    hash.transform_values { |v| from_string(v) }
+  end
+
+  def self.from_string(string)
+    StripAttributes.strip_string(string, allow_empty: true)
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -82,6 +82,8 @@ gem 'archive-zip'
 # Geocoding
 gem 'geocoder'
 
+gem 'strip_attributes'
+
 group :development do
   gem 'web-console', '>= 3.3.0'
   gem 'listen', '>= 3.0.5', '< 3.4'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -462,6 +462,8 @@ GEM
       actionpack (>= 4.0)
       activesupport (>= 4.0)
       sprockets (>= 3.0.0)
+    strip_attributes (1.11.0)
+      activemodel (>= 3.0, < 7.0)
     super_diff (0.5.2)
       attr_extras (>= 6.2.4)
       diff-lcs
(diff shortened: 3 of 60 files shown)


davidgisbey:
Looks good I'll have a quick glance through the controllers in a sec. We should add this to the support UI (in another card) too!

davidgisbey (approved):
Just had a look through the controllers and it looks like you've got everything 👍 

I've tested locally and it's working nicely.

app/lib/strip_whitespace.rb
```diff
@@ -0,0 +1,9 @@
+class StripWhitespace
+  def self.from_hash(hash)
+    hash.transform_values { |v| from_string(v) }
+  end
+
+  def self.from_string(string)
+    StripAttributes.strip_string(string, allow_empty: true)
```
malcolmbaig, line 7: re: `allow_empty` - by default, the gem converts empty strings into nils. This broke a couple of tests, including https://github.com/DFE-Digital/apply-for-teacher-training/blob/647d54311fb24d9ef16f9bc19787f7228985b857/spec/system/vendor_api/vendor_receives_application_spec.rb#L30

wherein we expect empty address lines to be empty strings in the API response.

Given this, I've disabled the nullifying behaviour by setting `allow_empty: true`. Is that reasonable, or should we try and preserve the `'' → nil` convention?

PR #3628, opened 2020-12-10
Present international addresses as separate fields in the API

steventux:
## Context

We've recently parsed and updated the single international address line into separate address lines. (See https://github.com/DFE-Digital/apply-for-teacher-training/pull/3597 for how)

- Present multiple address fields containing international address for a single application in the API.

## Guidance to review

I've included a fall-back to using `ApplicationForm#international_address` when there's an international address with no address individual fields, this is possible as we don't automatically parse the single line into multiple fields.
Once we start validating field length on multiple international addresses we can remove this fallback.

## Link to Trello card

https://trello.com/c/gz9VJ1wj/3093-resolve-international-non-uk-address-structures-for-api-integration

## Things to check

### DIFF
--- a/app/presenters/vendor_api/single_application_presenter.rb
+++ b/app/presenters/vendor_api/single_application_presenter.rb
@@ -286,9 +286,14 @@ module VendorAPI
 
     def contact_details
       if application_form.international?
+        address_line1 = application_form.address_line1 || application_form.international_address
+
         {
           phone_number: application_form.phone_number,
-          address_line1: application_form.international_address,
+          address_line1: address_line1,
+          address_line2: application_form.address_line2,
+          address_line3: application_form.address_line3,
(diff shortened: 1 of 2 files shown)


mnacos (approved):
Seems sensible to me 🚀 If there are any long international addresses entered in the meantime won't we end up with the same problem? (Tribal errors/reprocessing)

steventux:
@mnacos yes that's still a possibility until we validate input for international addresses, not sure of the best way to mitigate against this, we could just periodically clean up the data but this will mean periods of bad data being presented in the API.
In production there were 150 records with an international address, which seems like a lowish quantity, @george-bit is working on Support UI enhancements to allow us to amend addresses, this might be the best way to manage incoming data until we iterate how the addresses are saved.

PR #3630, opened 2020-12-10
Include change offer events in timeline and activity log

mnacos:
## Context

Changing offer details does not affect application status ('offer'), which means such events do not show up on the timeline or the activity log. This PR builds on https://github.com/DFE-Digital/apply-for-teacher-training/pull/3607 to include such events in the UI.

image

image

## Guidance to review

This code can handle multiple change events and display the selected provider/course details for that change correctly. However, a known issue involves making a modified offer in the first place, then changing offer details again. In that scenario, the 'offer made' event will display the most up-to-date offer details, which match the details of the 'change offer' event, not the 'offer made' event.

image

image

Solving this in a scalable way requires refactoring `MakeOffer` so that it groups changes in a single audit event (at least for `status` and `offered_course_option_id`) and could be done as a separate PR.

UPDATE: I have changed `MakeOffer` to group the `status` and `offered_course_option_id` changes in a single audit from now on. Once this is deployed, any new offers will display the right course details in the activity log at all times. Another PR may address historical data for 'made an offer' and 'change offer' events.

image

Art Rath, in this example, had originally applied for Computing (3BXF).

## Link to Trello card

https://trello.com/c/eES1T5BO

## Things to check


johndoates (changes requested):
In the activity log, I think it would be clearer if we keep the verb and noun together:

> Sabrina Kling changed the offer made to Earl Rice

This would also be more in line with 'set up an interview' and 'made an offer'.

The timeline entry is fine.

adamsilver (commented):
Same as John

app/components/provider_interface/application_timeline_component.rb
```diff
@@ -24,15 +26,22 @@ def initialize(application_choice:)
 
   private
 
+    def with_activity_log_events_for(attr)
+      filtered = @activity_log_events.select { |event| event.changes.key?(attr) }
+      filtered.map { |event| yield event }
```
benswannack, line 31: What does the `yield` on this line do?

app/components/provider_interface/activity_log_event_component.rb
```diff
@@ -45,14 +45,23 @@ def event_description
       else
```
benswannack, line 45: This case doesn't cover the `conditions_not_met` status, so we get blank titles in those blocks (there are a couple of examples in the review app). Is there a nice way to ensure that the specs cover all cases for the status and that were returning a sensible answer for each one?

benswannack (commented):
One thing to fix, otherwise looks good :)

app/components/provider_interface/activity_log_event_component.rb
mnacos: In theory, it'd be possible to write something based on `ApplicationStateChange.states_visible_to_provider`, but we have too many variations, e.g. multiple types of rejection, declining etc. so this remains a manual check for now.

app/components/provider_interface/application_timeline_component.rb
mnacos: Yields once per event, it's what enables
```
      with_activity_log_events_for('status') do |event|
        Event.new(
```
etc. below.

It abstracts away the filtering, while still letting you control the construction of Event structs.

johndoates (approved):
Looks good to me. Thanks!

app/components/provider_interface/application_timeline_component.rb
benswannack: Ah I see, missed that we weren't doing a `map` :)

benswannack (approved):

app/components/provider_interface/application_timeline_component.rb
mnacos: Yes, I guess it's easy to miss. Renaming the method to `map_activity_log_events_for`.

PR #3634, opened 2020-12-10
Enable 'add a course' in support interface

TobyRet:
## Context
We've had some support queries come through where candidates have requested adding an additional course to their application post-submission. This is something that ideally, should be able to be done through the support interface.

Adds the ability to search for course options by course code and add to an existing application. Includes basic validation.

## Guidance to review

To test it out, log into support and select an application _that has been submitted_. Under the course choice section you should see an 'add a course' button.

Git history will be rebased once approved 👍 

add_a_course_v2

## Link to Trello card

https://trello.com/c/yZ7Ahqvp/2621-dev-%F0%9F%8F%88-allow-support-users-to-add-a-course-in-support-post-submission

## Things to check

### DIFF
--- a/app/controllers/support_interface/application_forms/courses_controller.rb
+++ b/app/controllers/support_interface/application_forms/courses_controller.rb
@@ -1,0 +1,72 @@
+module SupportInterface
+  module ApplicationForms
+    class CoursesController < SupportInterfaceController
+      def new_search
+        @course_search = CourseSearchForm.new(
+          application_form_id: application_form_id,
+          course_code: course_code,
+        )
+      end
+
+      def search
+        @course_search = CourseSearchForm.new(
+          application_form_id: application_form_id,
+          course_code: course_search_params[:course_code],
+        )
+
+        if @course_search.valid?
+          redirect_to support_interface_application_form_new_course_path(
+            course_code: course_search_params[:course_code],
+            application_form_id: application_form_id,
+          )
+        else
+          render :new_search
+        end
+      end
+
+      def new
+        @pick_course = PickCourseForm.new(
+          course_code: course_code,
+          application_form_id: application_form_id,
+        )
+      end
+
+      def create
+        @pick_course = PickCourseForm.new(
+          course_option_id: course_option_id,
+          course_code: course_code,
+          application_form_id: application_form_id,
+        )
+
+        if @pick_course.save
+          redirect_to support_interface_application_form_path
+        else
+          render :new
+        end
+      end
+
+    private
+
+      def application_form
+        @application_form ||= ApplicationForm.find(application_form_id)
+      end
+
+      def course_option_id
+        params.dig(:support_interface_application_forms_pick_course_form, :course_option_id)
+      end
+
+      def course_search_params
+        params.require(:support_interface_application_forms_course_search_form)
+              .permit(:course_code)
+      end
+
+      def application_form_id
+        params[:application_form_id]
+      end
+
+      def course_code
+        params[:course_code]
+      end
+    end
+  end
+end
--- a/app/forms/support_interface/application_forms/course_search_form.rb
+++ b/app/forms/support_interface/application_forms/course_search_form.rb
@@ -1,0 +1,22 @@
+module SupportInterface
+  module ApplicationForms
+    class CourseSearchForm
+      include ActiveModel::Model
+      include ValidationUtils
+
+      attr_accessor :course_code, :application_form_id
+
+      validates :course_code, presence: true
+
+      def applicant_name
+        application_form.full_name
+      end
+
+    private
+
+      def application_form
+        @application_form ||= ApplicationForm.find(application_form_id)
+      end
+    end
+  end
+end
--- a/app/forms/support_interface/application_forms/pick_course_form.rb
+++ b/app/forms/support_interface/application_forms/pick_course_form.rb
@@ -1,0 +1,73 @@
+module SupportInterface
+  module ApplicationForms
+    class PickCourseForm
+      include ActiveModel::Model
+      include ValidationUtils
+
+      attr_accessor :course_code, :application_form_id, :candidate_id, :course_option_id
+
+      validates :course_option_id, presence: true
+      validates :application_form_id, presence: true
+      validates :course_code, presence: true
+
+      RadioOption = Struct.new(:course_option_id, :course_name, :course_code, :site_name)
+
+      def course_options
+        return @course_options if @course_options
+
+        course_options = courses.map { |course|
+          course.course_options
+                .available
+                .reject { |course_option| existing_course_ids.include?(course_option.course_id) }
+                .map { |course_option| RadioOption.new(course_option.id, course.name, course.code, course_option.site.name) }
+        }.flatten
+
+        sorted_course_options = course_options.sort_by(&:course_name)
+        @course_options = sorted_course_options
+      end
+
+      def save
+        return false unless valid?
+
+        SupportInterface::AddCourseChoiceAfterSubmission.new(
+          application_form: application_form,
+          course_option: course_option,
+        ).call
+      end
+
+      def applicant_name
+        application_form.full_name
+      end
+
+    private
+
+      def application_form
+        @application_form ||= ApplicationForm
+          .includes(application_choices: [:course_option])
+          .find(application_form_id)
+      end
+
+      def course_option
+        @course_option ||= CourseOption.find(course_option_id)
+      end
+
+      def courses
+        Course
+          .current_cycle
+          .includes(course_options: [:site])
+          .where(code: sanitize(course_code))
+      end
+
+      def sanitize(course_code)
+        course_code.strip.upcase
+      end
+
+      def existing_course_ids
+        application_form
+          .application_choices
+          .map(&:course_option)
+          .pluck(:course_id)
+      end
+    end
+  end
+end
--- a/app/views/support_interface/application_forms/courses/new.html.erb
+++ b/app/views/support_interface/application_forms/courses/new.html.erb
@@ -1,0 +1,22 @@
+<% content_for :title, title_with_error_prefix("Select a course to add to #{@pick_course.applicant_name}’s application", @pick_course.errors.any?) %>
+<% content_for :before_content, govuk_back_link_to(support_interface_application_form_search_course_new_path(course_code: @pick_course.course_code)) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <% if @pick_course.course_options.present? %>
+      <%= form_with model: @pick_course, url: support_interface_application_form_create_course_path(course_code: @pick_course.course_code) do |f| %>
+        <%= f.govuk_error_summary %>
+        <%= f.govuk_radio_buttons_fieldset :course_option_id, legend:  { text: "Which course should be added to the application?", size: "m" } do %>
+          <% @pick_course.course_options.each do |co| %>
+            <%= f.govuk_radio_button :course_option_id, co.course_option_id, label: { text: "#{co.course_name} (#{co.course_code}) - #{co.site_name}" } %>
+          <% end %>
+        <% end %>
+        <%= f.govuk_submit 'Add course to application' %>
+      <% end %>
+    <% else %>
+     <p class="govuk-body">No results</p>
+      <%= govuk_link_to( 'Search again', support_interface_application_form_search_course_new_path(course_code: @pick_course.course_code)) %>
+    <% end %>
+  </div>
+</div>
+
--- a/app/views/support_interface/application_forms/courses/new_search.html.erb
+++ b/app/views/support_interface/application_forms/courses/new_search.html.erb
@@ -1,0 +1,16 @@
+<% content_for :title, title_with_error_prefix("Search for a course to add to #{@course_search.applicant_name}’s application", @course_search.errors.any?) %>
+<% content_for :before_content, govuk_back_link_to(support_interface_application_form_path) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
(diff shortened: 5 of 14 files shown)


app/views/support_interface/application_forms/courses/new_search.html.erb
```diff
@@ -0,0 +1,16 @@
+<% content_for :title, 'Search for a course to add to an application' %>
```
frankieroberto, line 1: Would it be helpful to include the applicant’s name here?

```suggestion
<% content_for :title, "Search for a course to #{@applicant_name}’s application" %>
```

app/views/support_interface/application_forms/courses/new_search.html.erb
```diff
@@ -0,0 +1,16 @@
...
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with(
+            model: @course_search,
+            url: support_interface_application_form_search_course_path,
+        ) do |f| %>
+      <%= f.govuk_error_summary %>
+
+      <%= f.govuk_text_field :course_code, label: { text: 'Enter course code' }, width: 5 %>
```
frankieroberto, line 12: Don’t think we need the word "Enter" here.

```suggestion
      <%= f.govuk_text_field :course_code, label: { text: 'Course code' }, width: 5 %>
```

app/forms/support_interface/application_forms/course_search_form.rb
```diff
@@ -0,0 +1,12 @@
+module SupportInterface
+  module ApplicationForms
+    class CourseSearchForm
+      include ActiveModel::Model
+      include ValidationUtils
+
+      attr_accessor :course_code
+
+      validates :course_code, presence: true
```
frankieroberto, line 9: Could we add an error message for this of "Enter a course code" rather than the default "can't be blank"?

app/forms/support_interface/application_forms/pick_course_form.rb
```diff
@@ -0,0 +1,48 @@
+module SupportInterface
+  module ApplicationForms
+    class PickCourseForm
+      include ActiveModel::Model
+      include ValidationUtils
+
+      attr_accessor :course_code, :application_form_id, :candidate_id, :course_option_id
+
+      validates :course_option_id, presence: true
```
frankieroberto, line 9: Suggest an error message of something like "Select a course to add to the application" rather than "can't be blank"?

app/views/support_interface/application_forms/courses/new.html.erb
```diff
@@ -0,0 +1,16 @@
+<% content_for :title, 'Pick a course' %>
```
frankieroberto, line 1: Possible tweak to the page title?

```suggestion
<% content_for :title, "Select a course to add to #{@applicant_name}’s application" %>
```

app/views/support_interface/application_forms/courses/new.html.erb
```diff
@@ -0,0 +1,16 @@
...
+<% content_for :before_content, govuk_back_link_to(support_interface_application_form_search_course_new_path) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with model: @pick_course, url: support_interface_application_form_create_course_path(course_code: @course_code) do |f| %>
+      <%= f.govuk_error_summary %>
+      <%= f.govuk_radio_buttons_fieldset :course_option_id, legend:  { text: "Search results for course code '#{@course_code}'", size: "m" } do %>
```
frankieroberto, line 8: This legend might need to be more directly related to the purpose of the form, for accessibility reasons?

Could have a preceding heading which includes the search term?

```suggestion
      <%= f.govuk_radio_buttons_fieldset :course_option_id, legend:  { text: "Which course should be added to the application?", size: "m" } do %>
```

app/views/support_interface/application_forms/courses/new.html.erb
```diff
@@ -0,0 +1,16 @@
+<% content_for :title, 'Pick a course' %>
+<% content_for :before_content, govuk_back_link_to(support_interface_application_form_search_course_new_path) %>
```
frankieroberto, line 2: Could we feed in the course code to the back link as a query string, so that clicking back retains the search term?

```suggestion
<% content_for :before_content, govuk_back_link_to(support_interface_application_form_search_course_new_path(course_code: @course_code) %>
```

app/controllers/support_interface/application_forms/courses_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+        end
+      end
+
+    private
+
+      def search_courses
+        Course
+          .current_cycle
+          .includes(course_options: [:site])
+          .where(code: params[:course_code])
```
frankieroberto, line 51: It'd be good to trim leading and trailing whitespace here (quite common for it to get added on touch screen devices). And also upcase to avoid case-sensitivity?

```suggestion
          .where(code: params[:course_code].strip.upcase)
```

app/views/support_interface/application_forms/courses/new.html.erb
```diff
@@ -0,0 +1,16 @@
...
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with model: @pick_course, url: support_interface_application_form_create_course_path(course_code: @course_code) do |f| %>
+      <%= f.govuk_error_summary %>
+      <%= f.govuk_radio_buttons_fieldset :course_option_id, legend:  { text: "Search results for course code '#{@course_code}'", size: "m" } do %>
+        <% @pick_course.course_options.each do |co| %>
+          <%= f.govuk_radio_button :course_option_id, co.course_option_id, label: { text: "#{co.course_name} (#{co.course_code}) - #{co.site_name}" } %>
+        <% end %>
+      <% end %>
+      <%= f.govuk_submit 'Submit' %>
```
frankieroberto, line 13:
```suggestion
      <%= f.govuk_submit 'Add course to application' %>
```

frankieroberto (changes requested):
Looking good!

Left some suggestions for some minor tweaks.

The only other thing we might consider doing is handling the no-results-found page a bit better:

Screenshot 2020-12-10 at 17 20 48

Perhaps say "No results", remove the submit button, and have a link back to the search page?

app/views/support_interface/application_forms/courses/new.html.erb
```diff
@@ -0,0 +1,22 @@
+<% content_for :title, "Select a course to add to #{@pick_course.applicant_name}'s application" %>
```
paulrobertlloyd, line 1:
```suggestion
<% content_for :title, "Select a course to add to #{@pick_course.applicant_name}’s application" %>
```

paulrobertlloyd: Also need to use title with error prefix helper here

app/views/support_interface/application_forms/courses/new_search.html.erb
```diff
@@ -0,0 +1,16 @@
+<% content_for :title, "Search for a course to add to #{@course_search.applicant_name}'s application" %>
```
paulrobertlloyd, line 1:
```suggestion
<% content_for :title, "Search for a course to add to #{@course_search.applicant_name}’s application" %>
```

paulrobertlloyd: And again, use helper to prefix with ‘Error: ’ when validating page

TobyRet:
History will be rebased before merging

app/views/support_interface/application_forms/courses/new_search.html.erb
```diff
@@ -0,0 +1,16 @@
...
+<% content_for :before_content, govuk_back_link_to(support_interface_application_form_path) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with(
+            model: @course_search,
+            url: support_interface_application_form_search_course_path,
+        ) do |f| %>
+      <%= f.govuk_error_summary %>
+
```
paulrobertlloyd, line 11: Wonder if we should add a link to Find on this page (that opens in a new tab)

paulrobertlloyd (changes requested):
Does this enforce the business rules around adding choices? These are (that I know of):

* Candidates who have not used all three choices on Apply can add new courses within seven days of submission.
* A new course can then be added by the candidate on their application form, as long as it is within 7 days of the submission of their application.
* Candidates who have been rejected from a course within 7 days of the submission of their application cannot request to replace that course choice. This is applicable even in instances in which the reason for rejection is caused by course closure. They can apply to other courses when applying again, if they have resulted unsuccessful for all their course choices.

TobyRet:
> * 7 days of submission

@davidgisbey @paulrobertlloyd - no prob adding more business rules once confirmed. Will it be obvious to support users under what  conditions the 'add a course' button appears?

paulrobertlloyd:
> Will it be obvious to support users under what conditions the 'add a course' button appears?

Perhaps it should be something like this:

`[Add a course] 7 days remaining to amend course choices`

`Course choices can no longer be amended`

app/forms/support_interface/application_forms/pick_course_form.rb
```diff
@@ -0,0 +1,58 @@
...
+      validates :application_form_id, presence: true
+      validates :course_code, presence: true
+
+      RadioOption = Struct.new(:course_option_id, :course_name, :course_code, :site_name)
```
davidgisbey, line 13: Do you think we should also have provider_name for clarity? I have come across a course code that is 'Primary' for two different providers? It might be overkill though since they won't share a site.

davidgisbey: Kind fo just talked myself out of it haha

app/views/support_interface/application_forms/courses/new.html.erb
```diff
@@ -0,0 +1,22 @@
...
+<% content_for :before_content, govuk_back_link_to(support_interface_application_form_search_course_new_path(course_code: @pick_course.course_code)) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <% if @pick_course.course_options.present? %>
+      <%= form_with model: @pick_course, url: support_interface_application_form_create_course_path(course_code: @pick_course.course_code) do |f| %>
+        <%= f.govuk_error_summary %>
+        <%= f.govuk_radio_buttons_fieldset :course_option_id, legend:  { text: "Which course should be added to the application?", size: "m" } do %>
+          <% @pick_course.course_options.each do |co| %>
+            <%= f.govuk_radio_button :course_option_id, co.course_option_id, label: { text: "#{co.course_name} (#{co.course_code}) - #{co.site_name}" } %>
```
davidgisbey, line 11: I am wondering whether we should add provider_name to the struct and show it here. I have run into courses with the same name and course_code. It might be overkill as they won't share a site though.

davidgisbey:
The business rules discussion is an interesting one. I think we should probs sit down with Rachael to discuss. I think we should probs have some basic validations so you can't do stuff like this though at the very least.

image

app/views/support_interface/application_forms/courses/new.html.erb
TobyRet: Yep cool - I will add that in

TobyRet:
@davidgisbey @paulrobertlloyd - business rules added. 

Slack discussion:  https://ukgovernmentdfe.slack.com/archives/CP18YJXPY/p1607954153322200?thread_ts=1607952777.313300&cid=CP18YJXPY

app/views/support_interface/application_forms/courses/new.html.erb
```diff
@@ -0,0 +1,22 @@
+<% content_for :title, title_with_error_prefix("Select a course to add to #{@pick_course.applicant_name}‘s application", @pick_course.errors.any?) %>
```
paulrobertlloyd, line 1:
```suggestion
<% content_for :title, title_with_error_prefix("Select a course to add to #{@pick_course.applicant_name}’s application", @pick_course.errors.any?) %>
```

app/views/support_interface/application_forms/courses/new_search.html.erb
```diff
@@ -0,0 +1,16 @@
+<% content_for :title, title_with_error_prefix("Search for a course to add to #{@course_search.applicant_name}‘s application", @course_search.errors.any?) %>
```
paulrobertlloyd, line 1:
```suggestion
<% content_for :title, title_with_error_prefix("Search for a course to add to #{@course_search.applicant_name}’s application", @course_search.errors.any?) %>
```

paulrobertlloyd (commented):
Looks good to me, just using a single left quote when should be single right used for the apostrophe.

app/forms/support_interface/application_forms/pick_course_form.rb
```diff
@@ -53,6 +59,10 @@ def courses
       def sanitize(course_code)
         course_code.strip.upcase
       end
+
+      def existing_course_option_ids
+        application_form.application_choices.pluck(:course_option_id)
```
davidgisbey, line 64: I think what we're trying to do here is stop the same course from being added twice, rather than course_option (although that is covered too).

Maybe something like:

```
def existing_course_ids
  application_form.application_choices.map(:course_option).pluck(:course_id)
end
```

And then above on line 21:

`.reject { |course_option| existing_course_ids.include?(course_option.course_id) }`

davidgisbey (approved):
One minor comment from me, but all good other than that 👍

paulrobertlloyd (approved):

app/forms/support_interface/application_forms/pick_course_form.rb
TobyRet: Yup good spot. Have made the change

PR #3635, opened 2020-12-10
Remove the withdraw link from states with accepted offers

davidgisbey:
## Context

At the moment, a provider is presented with the option to withdraw offers for application choices in the deferred and pending conditions states. When they try and confirm the withdrawal they are presented the following error:

`The application is not ready for that action` 

This is because we do not allow a provider to reject applicants once they have accepted an offer 

image

☝️ FYI:  `withdraw! `is when a candidate withdraws their application and `reject! ` is for providers rejecting an application.

Pending conditions

|Before|After|
|------------|------------|
|image|image|

Deferred 

|Before|After|
|------------|------------|
|image|image| 

Confirm withdrawal error

image

## Guidance to review

I've looked into this as a result of a Zendesk ticket where a provider was getting the above error. 

## Things to check

### DIFF
--- a/app/components/provider_interface/status_box_components/pending_conditions_component.html.erb
+++ b/app/components/provider_interface/status_box_components/pending_conditions_component.html.erb
@@ -3,9 +3,7 @@
 
   <% if provider_can_respond %>
   <p class="govuk-body govuk-!-margin-bottom-4 govuk-!-display-none-print">
-    <%= govuk_link_to 'Withdraw offer', provider_interface_application_choice_new_withdraw_offer_path(application_choice.id) %>
-    or
-    <%= govuk_link_to 'defer offer', provider_interface_application_choice_new_defer_offer_path(application_choice.id) %>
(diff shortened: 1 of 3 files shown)


app/components/provider_interface/status_box_components/pending_conditions_component.html.erb
```diff
@@ -3,9 +3,7 @@
 
   <% if provider_can_respond %>
   <p class="govuk-body govuk-!-margin-bottom-4 govuk-!-display-none-print">
-    <%= govuk_link_to 'Withdraw offer', provider_interface_application_choice_new_withdraw_offer_path(application_choice.id) %>
-    or
-    <%= govuk_link_to 'defer offer', provider_interface_application_choice_new_defer_offer_path(application_choice.id) %>
+    <%= govuk_link_to 'Defer offer', provider_interface_application_choice_new_defer_offer_path(application_choice.id) %>
```
duncanjbrown, line 6: Thank you!

duncanjbrown (approved):
Thanks @davidgisbey!

PR #3637, opened 2020-12-10
Add sites data export to support

malcolmbaig:
## Context

We need an extract of sites data for analysis.

Add a data export to support that contains site/provider codes for synced providers, and the distance between site and provider.

## Guidance to review
- Code changes are fairly simple and follow the pattern of existing exports.
- Report can be accessed locally or on the review app via `/support/performance/data-exports/new`. Distance data dependent on geocodes existing for both sites and providers.

## Link to Trello card
https://trello.com/c/Mqz6Y4cO

## Things to check

### DIFF
--- a/app/models/data_export.rb
+++ b/app/models/data_export.rb
@@ -90,6 +90,11 @@ class DataExport < ApplicationRecord
       description: 'A list of application choices with the associated postcodes for the candidate, provider and site.',
       class: SupportInterface::LocationsExport,
     },
+    sites_export: {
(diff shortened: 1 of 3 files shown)


TobyRet (approved):
LGTM

duncanjbrown:
Should we be scoping these sites by cycle? Some go away between cycles

malcolmbaig:
> Should we be scoping these sites by cycle? Some go away between cycles

Good question! Will check with @Mikeemery2020

PR #3644, opened 2020-12-11
Allow support users to update references in support

raamSoftwire:
## Context

We want to make it easier for support users to input and edit references.

Add ability to edit references from support UI
- referee name
- referee email address
- referee relationship to candidate
- feedback

## Guidance to review

Are there other fields that should be editable?
At the moment, making these changes automatically sets the status of the reference to "feedback_provided". This is a first approximation, I think we will actually want to edit this too. But I'm not sure how this should work.

## Link to Trello card

https://trello.com/c/FS54ncwj/2622-dev-allow-support-users-to-update-references-in-support

## Things to check

image

image

image

### DIFF
--- a/app/components/support_interface/reference_with_feedback_component.rb
+++ b/app/components/support_interface/reference_with_feedback_component.rb
@@ -109,7 +109,7 @@ module SupportInterface
         key: 'Name',
         value: name,
         action: 'name',
-        change_path: support_interface_application_form_edit_reference_path(reference.application_form, reference),
+        change_path: support_interface_application_form_edit_reference_details_path(reference.application_form, reference),
       }
     end
 
@@ -118,7 +118,7 @@ module SupportInterface
         key: 'Email address',
         value: email_address,
         action: 'email_address',
-        change_path: support_interface_application_form_edit_reference_path(reference.application_form, reference),
+        change_path: support_interface_application_form_edit_reference_details_path(reference.application_form, reference),
       }
     end
 
@@ -127,7 +127,7 @@ module SupportInterface
         key: 'Relationship to candidate',
         value: relationship,
         action: 'relationship',
-        change_path: support_interface_application_form_edit_reference_path(reference.application_form, reference),
+        change_path: support_interface_application_form_edit_reference_details_path(reference.application_form, reference),
       }
     end
 
@@ -137,7 +137,7 @@ module SupportInterface
           key: 'Reference',
           value: feedback,
           action: 'feedback',
-          change_path: support_interface_application_form_edit_reference_path(reference.application_form, reference),
+          change_path: support_interface_application_form_edit_reference_feedback_path(reference.application_form, reference),
         }
       end
     end
--- a/app/components/support_interface/reference_with_feedback_component.rb
+++ b/app/components/support_interface/reference_with_feedback_component.rb
@@ -132,14 +132,12 @@ module SupportInterface
     end
 
     def feedback_row
-      if feedback
-        {
-          key: 'Reference',
-          value: feedback,
-          action: 'feedback',
-          change_path: support_interface_application_form_edit_reference_feedback_path(reference.application_form, reference),
-        }
-      end
+      {
+        key: 'Reference',
+        value: feedback,
+        action: 'feedback',
+        change_path: support_interface_application_form_edit_reference_feedback_path(reference.application_form, reference),
+      }
     end
 
     def consent_row
--- a/app/components/support_interface/reference_with_feedback_component.rb
+++ b/app/components/support_interface/reference_with_feedback_component.rb
@@ -11,9 +11,10 @@ module SupportInterface
              :consent_to_be_contacted,
              to: :reference
 
-    def initialize(reference:, reference_number:)
+    def initialize(reference:, reference_number:, editable: true)
       @reference = reference
       @reference_number = reference_number
+      @editable = editable
     end
 
     def rows
@@ -23,8 +24,8 @@ module SupportInterface
         name_row,
         email_address_row,
         relationship_row,
-        consent_row,
         feedback_row,
+        consent_row,
         history_row,
         possible_actions_row,
       ].flatten.compact
@@ -107,6 +108,8 @@ module SupportInterface
       {
         key: 'Name',
         value: name,
+        action: 'name',
+        change_path: support_interface_application_form_edit_reference_path(reference.application_form, reference),
       }
     end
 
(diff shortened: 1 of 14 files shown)


app/forms/support_interface/application_forms/edit_reference_form.rb
```diff
@@ -0,0 +1,45 @@
...
+      attr_accessor :name
+      attr_accessor :email_address
+      attr_accessor :feedback
+      attr_accessor :relationship
+      attr_accessor :audit_comment
+
+      attr_reader :reference
+      attr_reader :application_form
+
+      validates :name, presence: true
```
duncanjbrown: The validations here are different from those in the Candidate's form, eg `CandidateInterface::Reference::RefereeNameForm` has

```
    validates :name, presence: true, length: { minimum: 2, maximum: 200 }
```

Do we want to copy those?

duncanjbrown (commented):
This looks great — question about validations!

app/forms/support_interface/application_forms/edit_reference_form.rb
```diff
@@ -0,0 +1,46 @@
...
+      attr_accessor :feedback
+      attr_accessor :audit_comment
+
+      attr_reader :reference
+      attr_reader :application_form
+
+      validates :name, presence: true, length: { minimum: 2, maximum: 200 }
+      validates :email_address, presence: true, email_address: true, length: { maximum: 100 }
+      validates :relationship, presence: true, word_count: { maximum: 50 }
+      validates :feedback, presence: true, word_count: { maximum: 500 }
```
duncanjbrown, line 19: Is this always correct? We might want to edit a reference before feedback has been received

duncanjbrown:
I wonder if actually we want to have two forms here — one for editing the reference details, and one for providing feedback?

app/components/support_interface/reference_with_feedback_component.rb
```diff
@@ -11,9 +11,10 @@ class ReferenceWithFeedbackComponent < ViewComponent::Base
              :consent_to_be_contacted,
              to: :reference
 
-    def initialize(reference:, reference_number:)
+    def initialize(reference:, reference_number:, editable: true)
```
TobyRet, line 14: I can't see where `editable` is being used ... 🤔

app/forms/support_interface/application_forms/edit_reference_form.rb
raamSoftwire: Good point. I've also spoken to @paulrobertlloyd who agrees that we should split this out in to two separate forms. I've just pushed the changes

app/components/support_interface/reference_with_feedback_component.rb
raamSoftwire: It isn't called anywhere but without these changes we don't see the "change" links in the support UI. It's a bit of magic that I don't understand and would be very grateful if someone could explain! ✨🧙🏽‍♂️

app/components/support_interface/reference_with_feedback_component.rb
```diff
@@ -107,30 +108,36 @@ def name_row
...
       {
         key: 'Relationship to candidate',
         value: relationship,
+        action: 'relationship',
+        change_path: support_interface_application_form_edit_reference_details_path(reference.application_form, reference),
       }
     end
 
     def feedback_row
-      if feedback
```
raamSoftwire, line 128: I have removed this condition since a support user might want to add feedback as well as edit existing feedback.

app/forms/support_interface/application_forms/edit_reference_feedback_form.rb
```diff
@@ -0,0 +1,32 @@
...
+
+        super(feedback: @reference.feedback)
+      end
+
+      def save!
+        @reference.update!(
+          feedback: feedback,
+          audit_comment: audit_comment,
+        )
+        SubmitReference.new(reference: @reference).save!
```
raamSoftwire, line 28: By using this service it also triggers emails to the candidate and the referee and correctly changes the status of the reference

raamSoftwire: It feels like we should enable some way to reverse this process too, i.e. if a support user leaves a blank feedback it sets the status to `feedback_requested`

raamSoftwire: I spoke to @davidgisbey and we decided it best not to code for this for now. If we need to reverse the process this can still be done via the rails console.

app/forms/support_interface/application_forms/edit_reference_feedback_form.rb
```diff
@@ -2,30 +2,24 @@ module SupportInterface
...
-        @reference.update!(
+      def save(reference)
+        return false unless valid?
+
+        reference.update!(
           feedback: feedback,
           audit_comment: audit_comment,
         )
-        SubmitReference.new(reference: @reference).save!
+        SubmitReference.new(reference: reference).save!
```
davidgisbey, line 22: Do you think this should be called in the controller? It's not clear that this behaviour is going on from just calling save on the reference.

davidgisbey: Thinking about it. I don't think we should be calling this at all really. 

image

It does 3 things:

1. Update the reference to feedback_completed which is great
2. Emails the candidate to say received
3. Emails the referee

This would mean that if all we want to do is update the feedback, the emails still get sent.

I think we should just call r`eference.feedback_provided!`

davidgisbey: `  def initialize(reference:, edited_on_support: false)
    @reference = reference
  end

  def save!
    reference_feedback_provided!
    cancel_feedback_requested_references if enough_references_have_been_provided?
    CandidateMailer.reference_received(reference).deliver_later if edited_on_support
    RefereeMailer.reference_confirmation_email(application_form, reference).deliver_later if edited_on_support
  end
`

raamSoftwire: > Do you think this should be called in the controller? It's not clear that this behaviour is going on from just calling save on the reference.

I agree, it's worth pulling this out to make it clearer what is going on.

raamSoftwire: > Open to better suggestions though.

I've pushed a working implementation similar to your suggestion. Let me know what you think

davidgisbey (approved):
Nice work. Looks good to me. Let's :shipit:

PR #3645, opened 2020-12-11
Add form to select data for applications export

benswannack:
## Context
Adds an export for providers to get a (customisable) list of applications

Move the HESA export so that it is linked to from the new export page
Add a form to select which applications to show in the export

New form:
Screenshot 2020-12-14 at 17 20 13

## Link to Trello card
https://trello.com/c/9l90TE1b/2970-%F0%9F%8F%88let-users-export-application-data

## Things to check


benswannack:
Still need to add some system specs

duncanjbrown (approved):
Great stuff @benswannack — code looks good and it all works well through the front end.

I noticed while reviewing that we don't include the accredited provider code in the export, and also that we refer to it as the `accrediting` rather than `accredited` provider — please could you fix? (We try to avoid "accrediting" because the accreditation in this case comes from DfE and anoints the provider as a dispenser of NQT status. It's not the accreditation of teachers)

benswannack:
> I noticed while reviewing that we don't include the accredited provider code in the export, and also that we refer to it as the `accrediting` rather than `accredited` provider — please could you fix? (We try to avoid "accrediting" because the accreditation in this case comes from DfE and anoints the provider as a dispenser of NQT status. It's not the accreditation of teachers)

Ah, I copied this from the TAD export. Do we want to change accrediting -> accredited everywhere? @duncanjbrown

PR #3649, opened 2020-12-14
Add structured reasons for rejection to application dashboard

stevehook:
## Context

Structured reasons for rejection are now implemented in the provider interface so we need to bring the candidate interface inline with that to give the candidate more detailed feedback on unsuccessful applications. The PR displays the structured reasons on the application dashboard of the candidate's current application.

Further PR(s) are needed to cover the changes needed for previous applications.

- Add structured feedback to the application dashboard for rejected applications if `structured_reasons_for_rejection` feature flag is on and the provider has given structured feedback.
image

- Moved `ReasonsForRejectionComponent` and `ReasonsForRejection` (model) out of the `ProviderInterface` namespace to the top-level. So that they can used in `CandidateInterface` too.
- Minor tweaks to the content of the `ReasonsForRejectionComponent` that also appear in the provider interface:
image

- Assumes that we will need to display the legacy `rejection_reason` (unstructured reasons) for the foreseeable future for applications that pre-date this feature.

## Guidance to review

- Does the design work? My main concern is when you pick multiple answers under a given heading and each has suggestions to improve because it's not totally obvious which details belong with which item then.
- Is this change securely behind a feature flag?
- Are the tests adequate? (Just some additions to the component test and a new simple system spec).

## Link to Trello card

https://trello.com/c/Omq1ZtBK/1766-dev-feedback-on-previously-unsuccessful-applications-r4r

## Things to check


stevehook:
Made draft because it makes sense to add a few small changes here to handle the 'previous application' case too.

davidgisbey:
We're going to need to update the test_applications service to have structured rejection reasons.

davidgisbey:
image

I don't think we should be showing sections with no feedback. Do you agree?

spec/factories.rb
```diff
@@ -540,6 +540,56 @@
...
+          qualifications_which_qualifications: %w[no_english_gcse other],
+          quality_of_application_y_n: 'Yes',
+          quality_of_application_other_details: 'Lights on but nobody home',
+          quality_of_application_other_what_to_improve: 'Study harder',
+          quality_of_application_which_parts_needed_improvement: %w[personal_statement subject_knowledge other],
+          quality_of_application_subject_knowledge_what_to_improve: 'Claiming to be the \'world\'s leading expert\' seemed a bit strong',
+          quality_of_application_personal_statement_what_to_improve: 'Use a spellchecker',
+          safeguarding_y_n: 'Yes',
+          safeguarding_concerns: %w[other],
+          safeguarding_concerns_other_details: nil,
```
davidgisbey, line 573: I think we should make the factory consistent with the business rules. So safeguarding should be `No` if we're not going to assign one of the corresponding values. At the moment the UI is in an invalid state with a factory created object.

image

spec/factories.rb
```diff
@@ -540,6 +540,56 @@
...
+          candidate_behaviour_what_to_improve: 'Not scratch so much',
+          candidate_behaviour_what_did_the_candidate_do: %w[didnt_reply_to_interview_offer didnt_attend_interview other],
+          honesty_and_professionalism_y_n: 'Yes',
+          honesty_and_professionalism_concerns_other_details: nil,
+          honesty_and_professionalism_concerns: %w[information_false_or_inaccurate references],
+          honesty_and_professionalism_concerns_plagiarism_details: nil,
+          honesty_and_professionalism_concerns_references_details: 'Clearly not a popular student',
+          honesty_and_professionalism_concerns_information_false_or_inaccurate_details: 'Fake news',
+          offered_on_another_course_y_n: 'Yes',
+          offered_on_another_course_details: nil,
```
davidgisbey, line 559: Same as below 👇

davidgisbey (approved):
Tested it locally and it's working really nicely. 

No issues as far as i can see with the FF wrapping. I'm happy with the testing as the ProVendor team have covered it thoroughly. 

I've left a couple of minor nits on the factory.

app/components/reasons_for_rejection_component.html.erb
```diff
@@ -30,9 +30,10 @@
           <li>
             <% if reason == 'other' %>
               <p><%= reasons_for_rejection.quality_of_application_other_details %></p>
-              <p><%= reasons_for_rejection.quality_of_application_other_what_to_improve %></p>
+              <p>Details: <%= reasons_for_rejection.quality_of_application_other_what_to_improve %></p>
```
steventux, line 33: These `Details` prefixes don't appear on the Provider design - is this intended to change in both views? I know we iterated a similar `Advice:` prefix out of the provider designs...

spec/factories.rb
```diff
@@ -540,6 +540,56 @@
...
+          quality_of_application_subject_knowledge_what_to_improve: 'Claiming to be the \'world\'s leading expert\' seemed a bit strong',
+          quality_of_application_personal_statement_what_to_improve: 'Use a spellchecker',
+          safeguarding_y_n: 'Yes',
+          safeguarding_concerns: %w[other],
+          safeguarding_concerns_other_details: 'We need to run further checks',
+          safeguarding_concerns_vetting_disclosed_information_details: nil,
+          safeguarding_concerns_candidate_disclosed_information_details: nil,
+          interested_in_future_applications_y_n: nil,
+          why_are_you_rejecting_this_application: nil,
+          other_advice_or_feedback_y_n: nil,
```
steventux, line 578: I think @davidgisbey made a similar point, the `*_y_n` attributes will always be validated as `Yes` or `No`. It makes sense if the answer is `Yes` to have values in nested fields.

steventux (commented):
Looks good to me @stevehook , I've echoed what @davidgisbey said about factories and valid values. I can help with this if it's not obvious how the fields related to one another in the flat model attributes.

A question about `Details:` prefix and how this affects displaying the provider r4r component perhaps @adamsilver can advise?

app/models/reasons_for_rejection.rb
```diff
@@ -0,0 +1,71 @@
+class ReasonsForRejection
```
steventux, line 1: :+1: This helps with exposing the reasons in the API.

app/components/reasons_for_rejection_component.html.erb
stevehook: Yes it does update the provider design. I've done it that way mainly because the provider interface has words to the effect _this is what the candidate sees_ (maybe a bad assumption). But those _Details: _ prefixes can be removed when rendered in the provider interface if @adamsilver or others think they should be?

steventux: Yes my assumption was that apart from the `Change` links this should be the same rendering in Candidate and Provider UIs.

paulrobertlloyd: Yep, candidates and providers should see the same content.

The use of ‘Details: ’ prefix also appears to be used inconsistently; sometimes it appears for a free text answer, sometimes it doesn’t. Is that by design @adamsilver? I’d be tempted to remove it.

paulrobertlloyd (changes requested):
@stevehook The status row should appear before the feedback row on the course choice summary card – can’t see where this issue appears in the code.

steventux (approved):
Approving as this looks good to me, I'm assuming we get to the bottom of whether `Details:` prefix stays or goes, it's not significant enough to block this feature.

app/components/reasons_for_rejection_component.html.erb
adamsilver: @johndoates didn't we want to remove these prefixes?

johndoates: Yes we want to remove the prefixes. They don't add anything, given that there's a heading to give context. It's reasonable to assume that under a heading such as 'Something you did' you'll find details of what you did.

steventux: `Words:`

paulrobertlloyd (approved):

PR #3651, opened 2020-12-14
Handle deleted model files in 'Detect enum changes' checks

stevehook:
## Context

I'm getting build failures because I've moved a model file from one namespace to another and the `enum_in_models.rb` is erroring because it can't `require` the deleted file from the source tree.

- For now just handle the error

## Guidance to review

- I think this is the 'right thing to do' in cases where the model has genuinely been deleted. But what if the file has been moved? We shouldn't really ignore these cases because the file could have been moved _and_ modified.

## Link to Trello card

n/a

## Things to check

### DIFF
--- a/.github/scripts/enums_in_models.rb
+++ b/.github/scripts/enums_in_models.rb
@@ -7,16 +7,11 @@ if filename
     next unless line =~ /^<enum> /
     line.split('<enum> ')[1].split('.')[0]
   end
-  model_names.compact.uniq.map(&:constantize)
+  model_names.compact.uniq.map(&:constantize) rescue nil
 else
-  diff_tree_output = `git diff-tree -r --name-only --no-commit-id origin/master HEAD app/models`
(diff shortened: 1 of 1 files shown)


mnacos:
I've modified the diff detection step to a) treat git renames/moves as deletion/addition combinations and b) ignore deletion events. This way, the script will never fail to load any files, but it may encounter a constant that's no longer defined in the second stage, so I have moved the `rescue` there. If models have been moved/renamed, then the enums comment produced will only mention the new classes and their enum values, as if they are brand new additions (there will be no mention of the old classes/enum definitions). But there will still be an overall enums comment and I expect the human reviewers will be able to tell what's going on. Hope this helps @stevehook !

stevehook:
Thanks @mnacos - makes sense to me. Happy to merge if you are happy.

mnacos (approved):

PR #3653, opened 2020-12-14
Provide a list of possible GCSE/A-Level autocomplete values in the API

duncanjbrown:
## Context

Vendors have requested a copy of the possible values in subject autocompletes for GCSEs to help them consistently identify subjects.

Add a `reference-data` path to the API and underneath it add

- `gcse-subjects`
- `gcse-grades`
- `a-and-as-level-subjects`
- `a-and-as-level-grades`

## Guidance to review

Does this structure make sense?

## Link to Trello card

https://trello.com/c/OHGVySUu/3119-provide-possible-gcse-autocomplete-values-over-the-api

## Things to check

### DIFF
--- a/config/vendor-api-v1.yml
+++ b/config/vendor-api-v1.yml
@@ -257,6 +257,17 @@ paths:
           "$ref": "#/components/responses/NotFound"
         '422':
           "$ref": "#/components/responses/UnprocessableEntity"
+  "/reference-data/gcse-subjects":
+    get:
+      tags:
+      - Reference data
+      responses:
+        '200':
+          description: An array of possible GCSE subjects. Candidates are offered these in an autocomplete when filling in the form, but are also able to provide free-text responses.
+          content:
+            application/json:
+              schema:
+                "$ref": "#/components/schemas/ListResponse"
   "/test-data/regenerate":
     post:
       tags:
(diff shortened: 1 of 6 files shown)


config/vendor-api-v1.yml
```diff
@@ -257,6 +257,50 @@ paths:
...
+            application/json:
+              schema:
+                "$ref": "#/components/schemas/ListResponse"
+  "/reference-data/a-and-as-level-subjects":
+    get:
+      tags:
+      - Reference data
+      responses:
+        '200':
+          description: An array of possible A and AS level subjects. Candidates are offered these in an autocomplete when filling in the form, but are also able to provide free-text responses.
```
raamSoftwire, line 288: By grouping these together, we don't mention that a candidate can't have an A* grade at AS level. Should this be noted somewhere?

raamSoftwire (approved):
Looks good Duncan

config/vendor-api-v1.yml
duncanjbrown: Good spot — guilty as charged. I saw the two sets of possible grades were slightly different but decided that it was probably simpler for everyone if we had `as-and-a-level-grades` to go with `as-and-a-level-subjects`, rather than separate `as-level-grades` and `a-level-grades`. If anyone complains I'll split them!

PR #3656, opened 2020-12-15
Structured reasons for rejection for previous applications in candidate interface

stevehook:
## Context

Structured reasons for rejection are now implemented in the provider interface so we need to bring the candidate interface inline with that to give the candidate more detailed feedback on unsuccessful applications. This PR displays the structured rejection reasons for previous applications on the application dashboard of the candidate's current application.

It is a follow up to https://github.com/DFE-Digital/apply-for-teacher-training/pull/3649.

image

- [x] Show structured reasons for rejection on the application dashboard for apply again applications under _See feedback from your previous application_.
- [x] Don't show _See feedback from your previous application_ when there are no rejected applications with feedback.
- [x] Only show rejected applications that have feedback (structured or legacy).
- [x] Add status to the `ReasonsForRejectionComponent` used above.
- [x] Remove the link to Find from the `ReasonsForRejectionComponent` used above.

## Guidance to review

- Can be tested on the review app by enabling the feature flag `structured_reasons_for_rejection`, as a provider rejecting an application giving structured feedback and then logging in as that candidate to apply again.

## Link to Trello card

https://trello.com/c/Omq1ZtBK/1766-dev-feedback-on-previously-unsuccessful-applications-r4r

## Things to check

### DIFF
--- a/app/components/candidate_interface/rejection_reasons_component.rb
+++ b/app/components/candidate_interface/rejection_reasons_component.rb
@@ -29,7 +29,7 @@ module CandidateInterface
     end
 
     def course_details_row_value(application_choice)
-      if EndOfCycleTimetable.find_down? || FeatureFlag.active?(:structured_reasons_for_rejection)
+      if EndOfCycleTimetable.find_down?
         tag.p(application_choice.offered_course.name_and_code, class: 'govuk-!-margin-bottom-0') + tag.p(application_choice.course.description, class: 'govuk-body')
       else
         govuk_link_to(application_choice.offered_course.name_and_code,
--- a/app/components/candidate_interface/rejection_reasons_component.rb
+++ b/app/components/candidate_interface/rejection_reasons_component.rb
@@ -15,6 +15,10 @@ module CandidateInterface
       ].compact
     end
 
+    def render?
+      rejected_application_choices.present?
+    end
+
   private
 
     def course_details_row(application_choice)
@@ -60,6 +64,14 @@ module CandidateInterface
           key: 'Feedback',
           value: application_choice.rejection_reason,
         }
+      end
+    end
+
+    def rejected_application_choices
+      @rejected_application_choices ||= begin
+        rejected_applications = @application_form.application_choices.includes(:course, :provider, [:offered_course_option]).rejected
+        rejected_applications = rejected_applications.where('application_choices.rejection_reason IS NOT NULL OR application_choices.structured_rejection_reasons IS NOT NULL') if FeatureFlag.active?(:structured_reasons_for_rejection)
+        rejected_applications
       end
     end
   end
--- a/app/components/candidate_interface/rejection_reasons_component.rb
+++ b/app/components/candidate_interface/rejection_reasons_component.rb
@@ -10,6 +10,7 @@ module CandidateInterface
     def application_choice_rows(application_choice)
       [
         course_details_row(application_choice),
+        status_row(application_choice),
         rejection_reasons_row(application_choice),
       ].compact
     end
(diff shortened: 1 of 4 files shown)


app/components/candidate_interface/rejection_reasons_component.rb
```diff
@@ -10,10 +10,15 @@ def initialize(application_form:)
     def application_choice_rows(application_choice)
       [
         course_details_row(application_choice),
+        status_row(application_choice),
         rejection_reasons_row(application_choice),
       ].compact
     end
 
+    def render?
```
davidgisbey, line 18: Is this being called anywhere?

app/components/candidate_interface/rejection_reasons_component.rb
```diff
@@ -24,7 +29,7 @@ def course_details_row(application_choice)
     end
 
     def course_details_row_value(application_choice)
-      if EndOfCycleTimetable.find_down?
+      if EndOfCycleTimetable.find_down? || FeatureFlag.active?(:structured_reasons_for_rejection)
```
davidgisbey, line 32: Sorry if I'm being dense, but why don't we want to link to Find? I thought we were only concerned with that between cycles when Find is down.

davidgisbey:
A couple of questions, but other than that it's looking good.

app/components/candidate_interface/rejection_reasons_component.rb
TobyRet: ➕

app/components/candidate_interface/rejection_reasons_component.rb
```diff
@@ -52,5 +66,13 @@ def rejection_reasons_row(application_choice)
         }
       end
     end
+
+    def rejected_application_choices
+      @rejected_application_choices ||= begin
+        rejected_applications = @application_form.application_choices.includes(:course, :provider, [:offered_course_option]).rejected
```
TobyRet, line 72: Have `application_choices` been 'included' at this point?

TobyRet (dismissed):

app/components/candidate_interface/rejection_reasons_component.rb
stevehook: Not by our code but it's called by the `view_components` framework. It's just an easy way to do conditional rendering.

app/components/candidate_interface/rejection_reasons_component.rb
stevehook: @paulrobertlloyd did I interpret the requirement correctly here about removing the link to Find? What is the thinking behind this change?

app/components/candidate_interface/rejection_reasons_component.rb
stevehook: As far as I can see, no they've not been pre-loaded by an `include`. It looks like the object that ends up in `@application_form` is the `current_application.previous_application_form`. So it looks like the query that loads `application_choices` is a necessary one-off. (There is only a single instance of the `RejectionReasonsComponent`).

app/components/candidate_interface/rejection_reasons_component.rb
davidgisbey: Oh right okay, we explicitly call it in a few other components irrc. Thanks for the info

app/components/candidate_interface/rejection_reasons_component.rb
paulrobertlloyd: I have no idea what this is about! Where did this requirement surface?

stevehook: @paulrobertlloyd It was the third sub-task on the card
> **See feedback from your previous application** Only shown the **Course** (not linked to Find), **Status**, and **Feedback** in each summary card.

paulrobertlloyd: Huh. So funny reading my words back several months later! Also funny that I wrote this and yet it’s not reflected in the prototype. Funnier still that it took me a while to figure out what was wrong with seeing the link. Or maybe not that funny at all…

Given the above, let’s reduce complexity, remain consistent, and maintain the link. We can unlink it if it proves a problem (I suspect I may have been overthinking the solution).

stevehook: That link has been reinstated now.

TobyRet (approved):

paulrobertlloyd:
We shouldn’t show the feedback row if there is no feedback… and we shouldn’t show a choice in this list of previous applications if no feedback was provided either. 

Screenshot 2020-12-22 at 15 43 47

stevehook:
> We shouldn’t show the feedback row if there is no feedback… and we shouldn’t show a choice in this list of previous applications if no feedback was provided either.
> 
@paulrobertlloyd the `structured_reasons_for_rejection` feature flag was off on the review app so I think that's why you were seeing the old behaviour. Based on the course code I think this is the application you were looking at with the flag now on...

image

paulrobertlloyd:
Huzzah. Looks good to me… let’s merge this f**ker! :shipit:

paulrobertlloyd (approved):

PR #3660, opened 2020-12-15
Emails for structured reasons for rejection

despo:
## Context

Add new rejection emails that support structured rejection reasons

- Old emails are still in place as this functionality is only made available when the structured reasons for rejection feature flag is active

# New emails

### All applications rejected

all_applications_rejected

### All applications rejected - when rejected by default

all_applications_rejected_by_default

### Application rejected - othe applications awaiting decision only

application_rejected__awaiting_decision_only

### Application rejected by default - othe applications awaiting decision only

application_rejected_by_default__awaiting_decision_only

### Application rejected - one offer and one awaiting decision

application_rejected__one_offer_one_awaiting_decision

### Application rejected by default - one offer and one awaiting decision

application_rejected_by_default__one_offer_one_awaiting_decision

### Application rejected - other applications offers only

application_rejected_by_default__offers_only

### Application rejected by default - other applications offers only

application_rejected__offers_only

## Link to Trello card

https://trello.com/c/WPwIKSxe/3106-emails-for-structured-reasons-for-rejection

## Things to check


app/services/send_candidate_rejection_email.rb
```diff
@@ -6,16 +6,20 @@ def initialize(application_choice:)
   end
 
   def call
-    candidate_application_choices = application_choice.self_and_siblings
-    number_of_pending_decisions = candidate_application_choices.awaiting_provider_decision.count
-    number_of_offers = candidate_application_choices.offer.count
+    if FeatureFlag.active?(:structured_reasons_for_rejection)
+      SendRejectionEmailToCandidate.new(application_choice: application_choice).call
```
steventux, line 10: I wonder about breaking out into a new service to call back to the CandidateMailer, not a blocker, more of an opinion but it seems like we could reduce a bit of code if we had one method that could return the `CandidateMailer` method needed. The resulting call is always `CandidateMailer.send(:mailer_method, application_choice).deliver_later`

app/services/send_rejection_email_to_candidate.rb
```diff
@@ -0,0 +1,45 @@
+class SendRejectionEmailToCandidate
+  attr_reader :application_choice
+
+  def initialize(application_choice:)
+    @application_choice = application_choice
+  end
+
+  def call
+    if candidate_applications.all?(&:rejected?)
+      CandidateMailer.all_applications_rejected(application_choice).deliver_later
```
steventux, line 10: Is this deliberately different from the `application_rejected_all_rejected` email? Seems like the conditions are very similar for sending the mail but this method can handle structured reasons. I wonder if we need to consolidate the 3 possible formats somehow:

- structured reasons
- single reason (from API)
- rejected by default

This might just be me not understanding correctly, happy to chat about it.

steventux (changes requested):
Looks great, a few questions around how we cope with structured and simple reasons for rejection.

app/services/send_candidate_rejection_email.rb
despo: I considered this as a transitional step until the flag is removed and structured reasons for rejection takes over. Then the existing code can be canned. However, I now wonder if the flag is completely unecessary and instead I should be checking and including the correct reasons (structured or single reason as decline by default is already handled).

app/presenters/rejected_application_choice.rb
```diff
@@ -0,0 +1,119 @@
+class RejectedApplicationChoice < SimpleDelegator
+  def rejection_reasons
+    @rejection_reasons = [candidate_behaviour, quality_of_application, qualifications,
+                          interview_performance, full_course, offered_other_course,
+                          honesty_and_professionalism_reasons, safeguarding_issues, additional_advice,
+                          interested_in_future_applications].reduce({}, :merge)
+  end
+
+  def reasons
+    @reasons ||= OpenStruct.new(structured_rejection_reasons)
```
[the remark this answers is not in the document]
despo: Updated.

config/locales/reasons_for_rejection.yml
```diff
@@ -1,5 +1,27 @@
...
+      title: 'Quality of application'
+    qualifications:
+      title: 'Qualifications'
+    interview_performance:
+      title: 'Performance at interview'
+    full_course:
+      title: 'Course full'
+      reason: "We're sorry to tell you the course you applied to was full"
+    offered_on_another_course:
+      title: 'They offered you a place in another course'
```
steventux, line 15: This should be _They offered you a place on another course_

app/presenters/rejected_application_choice.rb
```diff
@@ -0,0 +1,119 @@
...
+    reason_details('additional_advice', [reasons.other_advice_or_feedback_details])
+  end
+
+  def interested_in_future_applications
+    return {} unless reasons.interested_in_future_applications_y_n.eql?('Yes')
+
+    reason_details(
+      'interested_in_future_applications',
+      [I18n.t('reasons_for_rejection.interested_in_future_applications.reason',
+              provider_name: course_option.course.provider.name)],
```
steventux, line 110: Not a blocker, this could be abbreviated to `provider.name` because https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/models/application_choice.rb#L11

spec/presenters/rejected_application_choice_presenter_spec.rb
```diff
@@ -0,0 +1,167 @@
+require 'rails_helper'
+
+RSpec.describe RejectedApplicationChoice do
```
steventux, line 3: The name of this spec implies we are testing `RejectedApplicationChoicePresenter` but the presenter class name is `RejectedApplicationChoice`. The convention seems to be to use `Presenter` in the class name.

app/presenters/rejected_application_choice.rb
despo: I tried that but I could not get `build_stubbed` object to work in the specs and then noticed a comment about it (here https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/mailers/candidate_mailer.rb#L64).

Any suggestions on how I can work around that?

steventux: No, you're right, it becomes a headache to stub, just ran into the same problem on the API changes.

app/services/send_candidate_rejection_email.rb
```diff
@@ -6,16 +6,20 @@ def initialize(application_choice:)
   end
 
   def call
-    candidate_application_choices = application_choice.self_and_siblings
-    number_of_pending_decisions = candidate_application_choices.awaiting_provider_decision.count
-    number_of_offers = candidate_application_choices.offer.count
+    if FeatureFlag.active?(:structured_reasons_for_rejection)
```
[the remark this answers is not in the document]
despo: I have now removed the feature flag completely and amended the presenter to support and format both structured_rejection_reasons and rejection_reason, making the assumption that when there is a rejection_reason set there shoud not be structured_rejection_reasons.

app/presenters/rejected_application_choice.rb
```diff
@@ -0,0 +1,127 @@
+class RejectedApplicationChoice < SimpleDelegator
+  def rejection_reasons
+    @rejection_reasons = single_reason_for_rejection if single_reason_for_rejection.present?
```
steventux, line 3: :+1:

app/mailers/candidate_mailer.rb
```diff
@@ -43,57 +43,45 @@ def new_referee_request(reference, reason:)
...
   end
 
-  def application_rejected_awaiting_decisions(application_choice)
-    @decisions = application_choice.self_and_siblings.select(&:awaiting_provider_decision?)
-    @application_choice = application_choice
-
-    # We cannot use `through:` associations with FactoryBot's `build_stubbed`. Using
-    # the association directly instead allows us to use `build_stubbed` in tests
-    # and mailer previews.
+  def application_rejected__one_offer_one_awaiting_decision(application_choice)
```
steventux, line 55: Not for this PR but this double underscore notation indicates that `CandidateMailer` is spread over a number of different concerns as we're namespacing via method names. At ~400 lines it's getting to the upper end of easy to work with, perhaps this is something to consider as tech debt as we could no doubt break this down into smaller themes.

app/presenters/rejected_application_choice.rb
```diff
@@ -0,0 +1,127 @@
+class RejectedApplicationChoice < SimpleDelegator
```
steventux, line 1: Minor style-police point: For better or for worse all our other presenters use `Presenter` in the class name. Not a blocker.

app/mailers/candidate_mailer.rb
```diff
@@ -43,57 +43,45 @@ def new_referee_request(reference, reason:)
...
-    )
+    email_for_candidate(application_choice.application_form)
   end
 
-  def application_rejected_offers_made(application_choice)
-    @offers = application_choice.self_and_siblings.select(&:offer?)
-    @decline_by_default_at = @offers.first.decline_by_default_at.to_s(:govuk_date)
-    @dbd_days = @offers.first.decline_by_default_days
-    @application_choice = application_choice
+  def application_rejected__awaiting_decision_only(application_choice)
```
steventux, line 66: Is there a reason for the double underscore? It's more noticeable when viewing the list of previewable email methods as we don't appear to use this elsewhere. Not a blocker as it's coding style but I don't understand how this method differs from other rejection email methods without a double underscore.

steventux (approved):
Looks good to me, preview emails look great.  
De-duping the all-rejected emails and having a reusable presenter for rejected applications is really useful. 
I've made a couple of very minor style points, feel free to ignore as they aren't intended to block this PR.

app/mailers/candidate_mailer.rb
despo: I agree. There are some quick fixes that can be done as well, like removing assignments to `@application_form` as `email_candidate` already does that. But I'll do a seperate PR as there is already another open PR that touches the candidate mailer.

app/mailers/candidate_mailer.rb
despo: The main reason I put this in place was so I could distinguish between the old and new emails, and it also made it a bit easier to understand the email context, but happy to update the names now that the old emails are out.

PR #3661, opened 2020-12-15
Add traits to ucas_match factory and use them in the tests

AgaDufrat:
## Context

It's difficult to create a match in action needed state, as it requires creating appropriate associations and populating `matching_data` accordingly.

Change how we generate UCAS matches so we don't have an Application Choice for an application on UCAS (for matching data scheme 'U').

Add traits:
- with_dual_application 
- with_multiple_acceptances do
- need_to_send_reminder_emails
- need_to_request_withdrawal_from_ucas

Also simplifies some of the test set up.

## Things to check


spec/components/support_interface/ucas_match_action_component_spec.rb
```diff
@@ -4,18 +4,16 @@
...
       result = render_inline(described_class.new(ucas_match_without_dual_applications))
       expect(result.text).to include('No action required')
     end
   end
 
   context 'when there is a dual application or dual acceptance' do
-    it 'renders correct information for a new match' do
-      ucas_match = create(:ucas_match, scheme: 'U', candidate_last_contacted_at: nil)
-      allow(ucas_match).to receive(:dual_application_or_dual_acceptance?).and_return(true)
+    let(:ucas_match) { create(:ucas_match, :with_dual_application) }
```
duncanjbrown, line 14: This is so much better!

spec/factories.rb
```diff
@@ -44,6 +48,31 @@
...
+      scheme { 'B' }
+      application_form { create(:completed_application_form, application_choices: [create(:submitted_application_choice)]) }
+      ucas_status { :awaiting_provider_decision }
+    end
+
+    trait :with_multiple_acceptances do
+      scheme { %w[D U] }
+      application_form do
+        create(:completed_application_form,
+               application_choices: [create(:application_choice, :with_accepted_offer), create(:application_choice, :with_accepted_offer)])
```
duncanjbrown, line 62: Do we need two application choices on this factory? Isn't the idea that there's one choice in Apply and one in UCAS?

duncanjbrown (commented):
🎉 This is a serious improvement!

One question about the way we assign `scheme` in the factories — should it be one per `application_choice`?

spec/factories.rb
AgaDufrat: Yeah, that's not quite right. I will amend the after build block.

spec/system/ucas_matching/matching_data_example.csv
```diff
@@ -2,5 +2,5 @@
 "U","99944","2FF","LMN","1","ABC99944DEF"
 "U","99944","D87","OPQ","1","ABC99944DEF"
 "U","99944","L06","RST","1","ABC99944DEF"
-"D","99957","1EP","UVW","0","ABC99957DEF�"
+"D","99957","1EP","UVW","0","ABC99957DEF"
```
duncanjbrown, line 5: Please leave this here as it's needed for a test: 56b5379

duncanjbrown (approved):
Very nice job @AgaDufrat — just one comment about the invisible char in the CSV!

spec/system/ucas_matching/matching_data_example.csv
AgaDufrat: Oops 🙊 Before I merge can you please double if it's ok for it to be saved in the db as `ABC99957DEFпїЅ`
That's the only way I can get that test to pass.

Our support agents have to copy the 'Trackable applicant key' if they need to contact UCAS to request withdrawal. That's the only thing we use it for.

duncanjbrown: Let's catch up in the morning!

duncanjbrown: We've decided not to worry about this as the offending field was a course name, which we don't use to look anything up by (it's effectively free text from UCAS). The siginificant fields — candidate ID etc — are strictly alphanumeric

PR #3663, opened 2020-12-15
Return domicile values consistent with HESA codes, not ISO 3166-2

mnacos:
### Context

Providers need to submit information to HESA about candidate's country of domicile. 

We currently return `domicile` as the country code of the student's address as the ISO 3166-2 country code, but these codes are not valid for particular countries of domicile.

For UK addresses:

The ISO country code GB is valid for nationality, but not domicile, where there are separate codes for the individual countries: `XF`, `XG`, `XH`, `XI` 

If the individual country isn't known for UK addresses, we can return code `XK`.

Add a `DomicileResolver` service, with a built-in table of known domicile codes.

This PR adds the following to our `COUNTRIES` list:

```
AQ Antarctica
XK Kosovo
```

The `DomicileResolver` service will pass country codes as HESA codes for all international addresses except for the following re-maps:

```
AQ (Antartica) to XX (Antartica and Oceania not otherwise specified)
CY (Cyprus) to XC (Cyprus not otherwise specified)
XK (Kosovo) to QO (Kosovo)
unresolved international addresses to ZZ (Not known)
```

For UK addresses, the challenge is specifying the the country (England, Wales...) from the address. `DomicileResolver` uses the UK postcode prefix against a lookup table to return:

```
XF England
XG Northern Ireland
XH Scotland
XI Wales
XL Channel Islands (not otherwise specified)
XK United Kingdom (not otherwise specified) [last resort]
```

`XK` is also returned when the postcode prefix spans two countries (e.g. England/Scotland).

### Guidance to review

No stored domicile field against an application for now, as the in-memory lookup is fast. AQ and XK are added to the `COUNTRIES` list because HESA re-maps them, so it's safer to add them now. Any other ISO country codes are valid HESA codes, so devs can add them to `COUNTRIES` without worrying about HESA codes.

I have changed our API responses and HESA export to return these new values instead of ISO codes for `domicile`. We don't seem to display `domicile` anywhere else. Should we also add it to application choices within the provider interface?

Finally, will specifying `domicile` in `ApplicationDataExport` break anything?

### Link to Trello card

[3117 - Correctly return country of domicile values](https://trello.com/c/LztnQHnq)

### Things to check

### DIFF
--- a/app/components/support_interface/personal_details_component.rb
+++ b/app/components/support_interface/personal_details_component.rb
@@ -26,6 +26,7 @@ module SupportInterface
         name_row,
         date_of_birth_row,
         nationality_row,
+        domicile_row,
         right_to_work_or_study_row,
         residency_details_row,
         phone_number_row,
@@ -67,6 +68,13 @@ module SupportInterface
       {
         key: 'Nationality',
         value: application_form.nationalities.to_sentence(last_word_connector: ' and '),
+      }
+    end
+
+    def domicile_row
+      {
+        key: 'Domicile',
+        value: application_form.domicile,
       }
     end
 
--- a/app/lib/domicile_resolver.rb
+++ b/app/lib/domicile_resolver.rb
@@ -1,0 +1,42 @@
+class DomicileResolver
+  def self.hesa_code_for_country(iso_country_code)
+    case iso_country_code
+    when nil then 'ZZ'
+    when 'AQ' then 'XX'
+    when 'CY' then 'XC'
+    when 'XK' then 'QO'
+    else
+      iso_country_code
+    end
+  end
+
+  def self.hesa_code_for_postcode(uk_postcode)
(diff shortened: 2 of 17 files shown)


duncanjbrown (commented):
Very nice! I think it's a good idea to add it to Manage UI while we're here. And support too.

mnacos:
> Very nice! I think it's a good idea to add it to Manage UI while we're here. And support too.

I've added the domicile code to the support interface but unsure if we should add to the provider interface yet, as all we have is the code, not the description of what it means e.g. `XF` instead of `England`.

I guess I could add all HESA data to the class and do a reverse lookup of code to country for the provider interface, but I guess we should chat a bit about it first and it could be its own story ('expose domicile in the provider interface').

duncanjbrown (approved):

PR #3665, opened 2020-12-15
Add provide and refuse feedback links in the support UI

davidgisbey:
## Context

As developers (and probably others!) we would like to be able easily and intuitively access the RefereeInterface. This change does this in the simplest way possible by adding `Provide feedback` and `Refuse feedback` links to the summary headers in the `ReferenceWithFeedbackComponent`. This component is rendered on the application show page.

These links only show on test environments and for references in the feedback_requested state.

- Add links to the summary header for references in the feedback_requested state in the Support UI

|Before|After|
|-------------|-------------|
|image|image|

## Guidance to review

Does this approach seem reasonable? My approach for this is to do the bare minimum to get it working. There might be a need later to build in a reference section/tab in the Support UI, but it seems a bit overkill for now. I can't really see the need for it tbh and we can always iterate later if needed.

Is the design okay? Should it be clearer that these only show on test envs?

Are the links only showing when they should be?
They def shouldn't be showing on prod & it seems to be they should only show for references that are awaiting feedback.

## Link to Trello card

https://trello.com/c/9XYQ3CEk/2624-dev-allow-signing-in-as-a-referee-from-support-not-in-production-though

## Things to check

### DIFF
--- a/app/components/support_interface/reference_with_feedback_component.html.erb
+++ b/app/components/support_interface/reference_with_feedback_component.html.erb
@@ -9,6 +9,10 @@
         </div>
       <% elsif reference.feedback_requested? %>
         <div class='app-summary-card__actions'>
+          <% if HostingEnvironment.test_environment? %>
+            <%= govuk_link_to 'Provide feedback', referee_interface_reference_relationship_path(token: reference.refresh_feedback_token!) %>
+            <%= govuk_link_to 'Refuse feedback', referee_interface_refuse_feedback_path(token: reference.refresh_feedback_token!) %>
+          <% end %>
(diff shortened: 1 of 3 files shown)


paulrobertlloyd:
@davidgisbey There are quite a number of features in support that can only be seen on test environments. If we think it’s important that this be highlighted, we can look at that in a separate issue (I began to have some thoughts, but it needs a more holistic approach)

Meanwhile, how about we do this: show the ‘Reference’ row if no reference has been given, and show the links there instead:

Screenshot 2020-12-17 at 10 47 27

raamSoftwire (approved):
LGTM David, cheers!

PR #3669, opened 2020-12-16
Surface structured reasons for rejection in the API

steventux:
## Context

The structured reasons for rejection feature means that more complex data will be stored on a rejection application, we should present this in the API where present.

Presents structured rejection reasons via the `SingleApplicationResponse` using the existing `Rejection.reason` attribute.

image

## Guidance to review

Unsure whether this needs documenting as the relevant response field and type have not changed but the contents may be more complex (and much longer) now.

The API docs state the `Rejection.reason` attribute is limited to 10240 characters. When using the limit on all text areas for structured reasons the API returned a string of around ~30000 characters. Though we are unlikely to see such wordy rejections this still might need addressing in the API documentation.

Presenter logic borrowed from https://github.com/DFE-Digital/apply-for-teacher-training/pull/3660 as this produces a hash of human-readable key/value(s) pairs representing section titles and answers.

## Things to check


despo (approved):
LGTM @steventux - I also updated the API field max Length in the docs one up to 65535 (TEXT) as well.

PR #3670, opened 2020-12-16
Add nationality editing to support interface

malcolmbaig:
## Context
We would like to make more areas of an application form editable in the support interface

Some work has already been completed to allow certain characteristics to be updated. This should be expanded to encompass all Personal Details, to assist with support queries.

- Add the nationality editing from the candidate interface to the support interface, including the right to work section.

## Guidance to review
- Should be easy to test in the review app, just need to navigate to an application form in support

## Link to Trello card
https://trello.com/c/vYKaG1or

## Things to check

### DIFF
--- a/app/controllers/support_interface/application_forms/nationalities_controller.rb
+++ b/app/controllers/support_interface/application_forms/nationalities_controller.rb
@@ -42,7 +42,8 @@ module SupportInterface
         StripWhitespace.from_hash params
           .require(:support_interface_application_forms_nationalities_form)
           .permit(
-            :first_nationality, :second_nationality, :other_nationality1, :other_nationality2, :other_nationality3, nationalities: []
+            :first_nationality, :second_nationality, :other_nationality1, :other_nationality2,
+            :other_nationality3, :audit_comment, nationalities: []
           )
       end
 
--- a/app/controllers/support_interface/application_forms/nationalities_controller.rb
+++ b/app/controllers/support_interface/application_forms/nationalities_controller.rb
@@ -11,10 +11,10 @@ module SupportInterface
         @nationalities_form = NationalitiesForm.new(prepare_nationalities_params)
 
         if @nationalities_form.save(@application_form)
-          if !british_or_irish?
-            redirect_to support_interface_application_form_edit_right_to_work_or_study_path
-          else
+          if british_or_irish?
             redirect_to support_interface_application_form_path(@application_form)
+          else
+            redirect_to support_interface_application_form_edit_right_to_work_or_study_path
           end
         else
           render :edit
--- a/app/controllers/support_interface/application_forms/nationalities_controller.rb
+++ b/app/controllers/support_interface/application_forms/nationalities_controller.rb
@@ -1,0 +1,54 @@
+module SupportInterface
+  module ApplicationForms
+    class NationalitiesController < SupportInterfaceController
+      before_action :set_application_form
+
+      def edit
+        @nationalities_form = NationalitiesForm.build_from_application(@application_form)
+      end
+
+      def update
+        @nationalities_form = NationalitiesForm.new(prepare_nationalities_params)
+
+        if @nationalities_form.save(@application_form)
+          if !british_or_irish?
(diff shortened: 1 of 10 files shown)


app/forms/support_interface/application_forms/nationalities_form.rb
```diff
@@ -0,0 +1,71 @@
...
+        inclusion: { in: NATIONALITY_DEMONYMS, allow_blank: true }
+      )
+      validates :audit_comment, presence: true
+      validate :candidate_provided_nationality
+
+      UK_AND_IRISH_NATIONALITIES = ['British', 'Welsh', 'Scottish', 'Northern Irish', 'Irish', 'English'].freeze
+
+      def self.build_from_application(application_form)
+        new(
+          application_form.build_nationalities_hash,
```
raamSoftwire, line 23: Where is `build_nationalities_hash` defined?

app/views/support_interface/application_forms/nationalities/edit.html.erb
```diff
@@ -0,0 +1,41 @@
+<% content_for :browser_title, title_with_error_prefix('Edit applicant nationality', @nationalities_form.errors.any?) %>
```
raamSoftwire, line 1: Could we make this `candidate` instead? 

I think they are synonymous within the service and so I think we should stick to just using one to reduce mental overheads.

app/views/support_interface/application_forms/nationalities/edit.html.erb
```diff
@@ -0,0 +1,41 @@
...
+    <%= form_with model: @nationalities_form, url: support_interface_application_form_edit_nationalities_path, class: 'app-nationality', method: :patch do |f| %>
+      <%= f.govuk_error_summary %>
+
+      <%= f.govuk_check_boxes_fieldset :nationalities, legend: { size: 'l', text: 'Edit applicant nationality' } do %>
+        <%= f.govuk_check_box(
+          :nationalities,
+          'British',
+          link_errors: true,
+          multiple: true,
+          label: { text: 'British' },
```
raamSoftwire, line 15: There's a mixture of using localizations and hardcoded strings for labels in this file, does that matter?

raamSoftwire (approved):
Looks great Malcolm. Just a few minor things from me

app/controllers/support_interface/application_forms/nationalities_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      def edit
+        @nationalities_form = NationalitiesForm.build_from_application(@application_form)
+      end
+
+      def update
+        @nationalities_form = NationalitiesForm.new(prepare_nationalities_params)
+
+        if @nationalities_form.save(@application_form)
+          if !british_or_irish?
```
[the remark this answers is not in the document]
malcolmbaig: Yep, good shout

app/forms/support_interface/application_forms/nationalities_form.rb
malcolmbaig: https://github.com/DFE-Digital/apply-for-teacher-training/blob/b82508ab2ec8a255c050f20ec5e6f11f5fb5023c/app/models/application_form.rb#L240

app/views/support_interface/application_forms/nationalities/edit.html.erb
malcolmbaig: Agree in principle, but this is a continuation of language already in use in the Support interface.

@paulrobertlloyd we're calling candidates "applicants" in the Support and Provider interfaces. Is that how we should continue to refer to them, or should we standardise on "candidate"?

app/views/support_interface/application_forms/nationalities/edit.html.erb
malcolmbaig: Not especially, IMHO

malcolmbaig:
Going to merge this. The candidate/applicant nomenclature question is being discussed here https://ukgovernmentdfe.slack.com/archives/CN9Q1VB2M/p1608289760147300. Can pick up any changes in a separate PR.

PR #3672, opened 2020-12-17
Multi-line international contact addresses in candidate interface

stevehook:
## Context

Some student record systems can’t accept the free text we currently allow candidates to enter if they have an international address. We want to update this so that they enter an address over 4 lines.

- [x] Add `international_addresses` feature flag.
- [x] Updated the international address page to use the same fields as UK addresses with (some) different labels as specified if `international_addresses` feature flag is active.
image
- [x] Update the review page to use the same address pattern as UK addresses but with the addition of _Country_ if `international_addresses` feature flag is active.
image

## Guidance to review

- Per commit recommended
- It should be possible to test this manually on the review app by toggling the `international_addresses` feature flag.
- Is the feature flag enforced correctly?

## Link to Trello card

https://trello.com/c/4kYVYB6J/2676-🌐-dev-split-international-contact-address-into-multiple-lines

## Things to check

### DIFF
--- a/app/forms/candidate_interface/contact_details_form.rb
+++ b/app/forms/candidate_interface/contact_details_form.rb
@@ -6,7 +6,8 @@ module CandidateInterface
                   :address_line4, :postcode, :address_type, :country, :international_address
 
     validates :address_line1, :address_line3, :postcode, presence: true, on: :address, if: :uk?
-    validates :international_address, presence: true, on: :address, if: :international?
+    validates :address_line1, presence: true, on: :address, if: ->(form){ form.international? && FeatureFlag.active?(:international_addresses) }
+    validates :international_address, presence: true, on: :address, if: ->(form){ form.international? && !FeatureFlag.active?(:international_addresses) }
     validates :address_type, presence: true, on: :address_type
     validates :country, presence: true, on: :address_type, if: :international?
 
@@ -15,7 +16,7 @@ module CandidateInterface
 
     validates :phone_number, length: { maximum: 50 }, phone_number: true, on: :base
 
-    validates :postcode, postcode: true, on: :address
+    validates :postcode, postcode: true, on: :address, if: :uk?
 
     def self.build_from_application(application_form)
       new(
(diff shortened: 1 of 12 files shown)


app/forms/candidate_interface/contact_details_form.rb
```diff
@@ -6,7 +6,8 @@ class ContactDetailsForm
                   :address_line4, :postcode, :address_type, :country, :international_address
 
     validates :address_line1, :address_line3, :postcode, presence: true, on: :address, if: :uk?
-    validates :international_address, presence: true, on: :address, if: :international?
+    validates :address_line1, presence: true, on: :address, if: ->(form){ form.international? && FeatureFlag.active?(:international_addresses) }
```
raamSoftwire, line 9: Do we intentionally only check the first line of international addresses? Presumably we're less strict because we can't be sure of the format?

spec/forms/candidate_interface/contact_details_form_spec.rb
```diff
@@ -61,20 +61,46 @@
...
+      before { FeatureFlag.activate(:international_addresses) }
+
+      it 'updates the provided ApplicationForm with the international address field if valid' do
+        form_data = {
+          address_line1: '123 Chandni Chowk',
+          address_line3: 'Old Delhi',
+          postcode: '110006',
+        }
+        application_form = build(:application_form)
+        contact_details = CandidateInterface::ContactDetailsForm.new(form_data)
```
raamSoftwire, line 95: Should we be using the `build_from_application` method here instead? And if not, are you able to explain why please?

raamSoftwire (approved):

app/forms/candidate_interface/contact_details_form.rb
davidgisbey: Yea we're really not sure what is and is not valid for each individual country so we decided this was probably the best approach.

paulrobertlloyd (approved):

spec/forms/candidate_interface/contact_details_form_spec.rb
stevehook: :+1: It's a bit simpler to just pass the form_data into set it up but it's a bit different to the way this would be used in real life so I've changed it to user `build_from_application` to initialise the form from a model, then user `assign_attributes` to input the form_data.

stevehook: Switched back to original following call. (`build_from_application` isn't used in `update` actions so the original was closer to real life usage).

PR #3673, opened 2020-12-17
Provider users can configure their notifications

despo:
## Context
Enable provider users to control their email notifications

Add a new notification settings page where provider users can check/uncheck a component to opt in or our of receiving email notifications.

account-page

notifications-page

## Link to Trello card

https://trello.com/c/hgYZ0Cqi/3026-provider-users-can-choose-whether-or-not-they-receive-emails-from-apply

## Things to check

### DIFF
--- a/app/views/provider_interface/notifications/show.html.erb
+++ b/app/views/provider_interface/notifications/show.html.erb
@@ -1,0 +1,35 @@
+<%= content_for :browser_title, t('page_titles.provider.notifications') %>
+
+<% content_for :before_content do %>
+  <%= render BreadcrumbComponent.new(items: [
+    {
+      text: t('page_titles.provider.account'),
+      path: provider_interface_account_path
+    },
+    {
+      text: t('page_titles.provider.notifications')
+    }
+  ]) %>
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l govuk-!-margin-bottom-2">
+      <%= t('page_titles.provider.notifications') %>
+      <span class="govuk-caption-m govuk-!-margin-top-1">These are sent when an application is submitted, withdrawn, automatically rejected, accepted or declined.</span>
+    </h1>
+
+    <%= form_with model: current_provider_user, url: provider_interface_notifications_path, method: :put do |f| %>
+      <div class='govuk-form-group'>
+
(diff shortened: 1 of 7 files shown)


duncanjbrown:
Now that we've got notifications do we need to move to the 4-up design @adamsilver?
 
Screenshot 2020-12-18 at 16 48 26

duncanjbrown:
UI looks like a radio in the prototype, rather than a checkbox.

(I don't think updating the flash styles is part of this PR, but we should do it soon!)

Screenshot 2020-12-18 at 16 49 21

despo:
@duncanjbrown Updated. I was not aware about these changes being part of the prototype as it was not mentioned in the card. Can you share the link to this please.

app/views/provider_interface/notifications/show.html.erb
```diff
@@ -0,0 +1,35 @@
...
+      path: provider_interface_account_path
+    },
+    {
+      text: t('page_titles.provider.notifications')
+    }
+  ]) %>
+<% end %>
+
+<div class="govuk-grid">
+  <div class="govuk-grid-column-two-third">
```
duncanjbrown, line 16:
```suggestion
  <div class="govuk-grid-column-two-thirds">
```

config/locales/en.yml
```diff
@@ -154,6 +154,7 @@ en:
       users: Users
       export_hesa_data: Export data for Higher Education Statistics Agency (HESA)
       export_application_data: Export data
+      notifications: Email notifications
```
duncanjbrown, line 157: :+1:

spec/system/provider_interface/account_page_spec.rb
```diff
@@ -5,9 +5,15 @@
 
   scenario 'Provider user visits their account page with various permissions' do
     given_i_am_a_provider_user_with_dfe_sign_in
-    and_i_can_manage_organisations_but_not_users
+    given_i_cannot_manage_users_or_organisations
```
duncanjbrown, line 8: Does this matter here?

duncanjbrown (approved):
Great stuff, thanks for changes. Sorry for not including prototype link on the card — my bad.

One typo to fix in a CSS class name and a minor question about a spec.

spec/system/provider_interface/account_page_spec.rb
despo: Because there is context around the other checks it made sense to have one for this as well.

I have now updated the notifications check to not only check for existence of notifications but also for non existence of user or organisation management links.

PR #3674, opened 2020-12-17
Add ability to edit candidate addresses

raamSoftwire:
## Context

We are increasing the support UI to enable support users to make changes more easily. This PR looks add editing address details.

image

image

image

image

image

image

## Guidance to review

@stevehook is currently working on refactoring international addresses. We will build on his work to offer both new and old forms on the international address details page to help support users migrate old addresses to the new format.

I've borrowed a lot from the existing flow in the candidate interface. It might be worth peeking at the `contact_details_form` and the associated system and form specs.

## Link to Trello card

https://trello.com/c/LrAuSaZs/2687-expand-the-personal-details-review-section-in-the-support-ui-address-incl-international

## Things to check

### DIFF
--- a/app/controllers/support_interface/application_forms/address_details_controller.rb
+++ b/app/controllers/support_interface/application_forms/address_details_controller.rb
@@ -4,13 +4,14 @@ module SupportInterface
       before_action :set_application_form
 
       def edit
-        @details = EditAddressDetailsForm.build_from_application_form(@application_form)
+        @details_form = EditAddressDetailsForm.build_from_application_form(@application_form)
       end
 
       def update
         @application_form.assign_attributes(address_details_params)
-        @details = EditAddressDetailsForm.build_from_application_form(@application_form)
-        if @details.save_address(@application_form)
+        @details_form = EditAddressDetailsForm.build_from_application_form(@application_form)
+
+        if @details_form.save_address(@application_form)
           flash[:success] = 'Address details updated'
           redirect_to support_interface_application_form_path(@application_form)
         else
--- a/app/controllers/support_interface/application_forms/address_details_controller.rb
+++ b/app/controllers/support_interface/application_forms/address_details_controller.rb
@@ -8,8 +8,9 @@ module SupportInterface
       end
 
       def update
-        @application_form.assign_attributes(address_details_params)
-        @details_form = EditAddressDetailsForm.build_from_application_form(@application_form)
+        @details_form = EditAddressDetailsForm.new(
+          address_details_params.merge(address_type: @application_form.address_type),
+        )
 
         if @details_form.save_address(@application_form)
           flash[:success] = 'Address details updated'
--- a/app/controllers/support_interface/application_forms/address_details_controller.rb
+++ b/app/controllers/support_interface/application_forms/address_details_controller.rb
@@ -1,0 +1,37 @@
+module SupportInterface
+  module ApplicationForms
+    class AddressDetailsController < SupportInterfaceController
+      def edit
+        @details = details_form
+      end
+
+      def update
+        application_form.assign_attributes(address_details_params)
+        @details = details_form
+        if @details.save_address(application_form)
+          flash[:success] = 'Address details updated'
+          redirect_to support_interface_application_form_path(application_form)
+        else
+          render :edit
+        end
+      end
+
+    private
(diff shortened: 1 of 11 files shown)


spec/system/support_interface/editing_address_details_spec.rb
```diff
@@ -0,0 +1,132 @@
+require 'rails_helper'
+
+RSpec.feature 'Editing reference' do
+  include DfESignInHelpers
+
+  scenario 'Support user edits reference', with_audited: true do
```
george-bit, line 6:
```suggestion
  scenario 'Support user edits address', with_audited: true do
```

spec/system/support_interface/editing_address_details_spec.rb
```diff
@@ -0,0 +1,132 @@
+require 'rails_helper'
+
+RSpec.feature 'Editing reference' do
```
george-bit, line 3:
```suggestion
RSpec.feature 'Editing address' do
```

george-bit (changes requested):
LGTM only minor name changes.

george-bit (approved):
👍

stevehook (approved):
Review app works really nicely.

My only thoughts were about the duplication of things like validations between forms in the candidate and support interfaces. I like the idea of having separate forms for separate interfaces and often they will be very different but I'd be interested in others thoughts on whether it's something we should try to DRY a bit and if so how best to go about it. In any case we should iterate on that in future (and we'll have to add the new international address structure here soon anyway).

raamSoftwire:
@malcolmbaig has done a similar PR (https://github.com/DFE-Digital/apply-for-teacher-training/pull/3670) and there are a few refactors I want to implement here before merging.

app/controllers/support_interface/application_forms/address_details_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    class AddressDetailsController < SupportInterfaceController
+      before_action :set_application_form
+
+      def edit
+        @details_form = EditAddressDetailsForm.build_from_application_form(@application_form)
+      end
+
+      def update
+        @details_form = EditAddressDetailsForm.new(
+          address_details_params.merge(address_type: @application_form.address_type),
```
raamSoftwire, line 12: @malcolmbaig after our chat I decided to go for the approach already taken on the Candidate side.

I think this is more explicit about passing in the extra `address_type` param, and allows us to stick to the pattern of using `Form.new` in the update action.

spec/forms/support_interface/application_forms/edit_address_details_form_spec.rb
```diff
@@ -0,0 +1,138 @@
+require 'rails_helper'
+
+RSpec.describe SupportInterface::ApplicationForms::EditAddressDetailsForm, type: :model do
+  describe '.build_from_application_form' do
+    it 'creates an object based on the provided ApplicationForm' do
+      data = {
+        address_line1: Faker::Address.street_name,
```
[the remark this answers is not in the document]
raamSoftwire: We already have other calls to `Faker::Address.street_name` in the codebase. Do we currently see this causing issues? 

If we did then it looks like [we can specify a max length](https://github.com/faker-ruby/faker/issues/249#issuecomment-54014800).

raamSoftwire:
> @malcolmbaig has done a similar PR (#3670) and there are a few refactors I want to implement here before merging.

I've pushed those refactors now.

raamSoftwire:
> Review app works really nicely.
> 
> My only thoughts were about the duplication of things like validations between forms in the candidate and support interfaces. I like the idea of having separate forms for separate interfaces and often they will be very different but I'd be interested in others thoughts on whether it's something we should try to DRY a bit and if so how best to go about it. In any case we should iterate on that in future (and we'll have to add the new international address structure here soon anyway).

I think this is a good point that we should think about. I did feel a little uneasy with the amount of copy and pasting. But I saw it as necessary within the namespaced approach we've taken. I had a good conversation with @malcolmbaig about it and he suggested we be guided by the [Rule of Three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)). I think this sounds about right to me but maybe also something to discuss with the other devs at the next dev retro/near miss session.

PR #3680, opened 2020-12-17
Add a support UI filter for provider users who have never signed in

steventux:
## Context

We'd like to be able to quickly identify provider users who have never signed in.

Adds a filter to easily identify provider users who have / have not signed in.

image

## Guidance to review

Amended to use dual checkboxes with the condition that if both are selected, the filter has no effect.

## Link to Trello card

https://trello.com/c/2UFg04ud/3130-add-a-filter-to-show-users-in-support-who-have-never-signed-into-manage

## Things to check

### DIFF
--- a/app/controllers/support_interface/provider_users_controller.rb
+++ b/app/controllers/support_interface/provider_users_controller.rb
@@ -5,6 +5,8 @@ module SupportInterface
         .includes(providers: %i[training_provider_permissions ratifying_provider_permissions])
         .page(params[:page] || 1).per(30)
 
+      @provider_users = scope_by_use_of_service
+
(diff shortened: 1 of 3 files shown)


AgaDufrat (approved):
Looking good.
I had the same dilemma of adding a filter with 1 checkbox https://github.com/DFE-Digital/apply-for-teacher-training/pull/3400 but went ahead with it and seems to work for the users.

steventux:
@AgaDufrat thanks for the review, I've amended to use 2 checkboxes as per @johndoates and @adamsilver 's advice.

AgaDufrat (approved):

PR #3682, opened 2020-12-18
Move UCASMatch file download check to worker

despo:
Move check for ucas match file download outside of OKComputer and follow DetectInvariant pattern

## Things to check

### DIFF
--- a/app/workers/ucas_integration_check.rb
+++ b/app/workers/ucas_integration_check.rb
@@ -1,0 +1,17 @@
+class UCASIntegrationCheck
+  include Sidekiq::Worker
+
+  def perform
+    detect_ucas_match_file_upload_failure
+  end
+
+  def detect_ucas_match_file_upload_failure
+    file_download = UCASMatching::FileDownloadCheck.new
+    file_download.check
+    return if file_download.success?
+
+    Raven.capture_exception(UCASMatchingFileDownloadFailure.new(file_download.message))
+  end
+
+  class UCASMatchingFileDownloadFailure < StandardError; end
+end
--- a/config/clock.rb
+++ b/config/clock.rb
@@ -37,6 +37,12 @@ class Clock
     end
   end
 
+  every(1.day, 'UCASIntegrationCheck', at: '10:00') do
+    return unless HostingEnvironment.productioun?
+
+    UCASIntegrationCheck.perform_async if Time.zone.yesterday.weekday?
+  end
+
   every(1.hour, 'SyncAllFromTeacherTrainingPublicAPI') do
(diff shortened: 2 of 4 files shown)


duncanjbrown:
Implementation looks good — I think we could do without the extra ceremony of `FailuresCheck` and this could probably go straight into `DetectInvariants` without a need for a separate service?

despo:
@duncanjbrown I am not sure I like the worker name but I think it going in DetectInvariants can be confusing as that worker is quite specific at checking for invalid application states, and in addition to that that check is quite regular (runs every hour) when we just want to be notified once a day for any missed downloads.

duncanjbrown:
> @duncanjbrown I am not sure I like the worker name but I think it going in DetectInvariants can be confusing as that worker is quite specific at checking for invalid application states, and in addition to that that check is quite regular (runs every hour) when we just want to be notified once a day for any missed downloads.

OK, fair enough. Name, then! There's nothing else in here yet...  `UCASIntegrationCheck`?

config/clock.rb
```diff
@@ -37,6 +37,12 @@ class Clock
     end
   end
 
+  every(1.day, 'UCASIntegrationCheck', at: '10:00') do
+    return unless HostingEnvironment.productioun?
+
+    UCASIntegrationCheck.perform_async if Time.zone.yesterday.weekday?
```
duncanjbrown, line 43: Should this `if` live inside `UCASMatching::FileDownloadCheck`?

app/workers/ucas_integration_check.rb
```diff
@@ -0,0 +1,18 @@
...
+  def perform
+    detect_ucas_match_file_upload_failure
+  end
+
+  def detect_ucas_match_file_upload_failure
+    file_download = UCASMatching::FileDownloadCheck.new
+    file_download.check
+    return if file_download.success?
+
+    message = "The UCAS data file has not been imported since #{UCASMatching::FileDownloadCheck.last_sync}."
```
duncanjbrown, line 13: We could return the `message` fro `FileDownloadCheck`?

config/clock.rb
despo: The condition for marking a failure in `UCASMatching::FileDownloadCheck` already takes the dates into account

> elsif Time.zone.now > Time.zone.parse(last_date).next_weekday

I just don't think it's necessary to run this check on a daily basis either and followed the [patten currently used](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/config/clock.rb#L31) in the clock configuration.

duncanjbrown (approved):

duncanjbrown:
Thanks for changes!

PR #3683, opened 2020-12-18
Remove duplicate assignments to @application_form

despo:
Remove assignments to @application_form as `email_for_candidate` [takes care of that](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/mailers/candidate_mailer.rb#L379).

### DIFF
--- a/app/mailers/candidate_mailer.rb
+++ b/app/mailers/candidate_mailer.rb
@@ -253,9 +253,8 @@ class CandidateMailer < ApplicationMailer
 
   def course_unavailable_notification(application_choice, reason)
     @application_choice = application_choice
-    @application_form = application_choice.application_form
     email_for_candidate(
-      @application_form,
+      application_choice.application_form,
       subject: I18n.t!(
         "candidate_mailer.course_unavailable_notification.subject.#{reason}",
         course_name: application_choice.course_option.course.name_and_code,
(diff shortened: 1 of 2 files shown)


app/mailers/candidate_mailer.rb
```diff
@@ -283,69 +281,63 @@ def offer_accepted(application_choice)
...
     )
   end
 
   def ucas_match_initial_email_multiple_acceptances(candidate)
-    @application_form = candidate.application_forms.first
     @date_to_withdraw_application_by = TimeLimitCalculator.new(rule: :ucas_match_candidate_withdrawal_request, effective_date: Time.zone.today).call.fetch(:time_in_future).to_s(:govuk_date)
 
     email_for_candidate(
-      @application_form,
+      candidate.application_forms.first,
```
malcolmbaig, line 298: Should this be `candidate.current_application`?

malcolmbaig (approved):
👍🏽 

Just the one question left inline.

app/mailers/candidate_mailer.rb
despo: It shouldn't make any difference as the application_form is only used for retrieving the candidate name, but I have updated as it makes sense to always use the latest application data.

PR #3684, opened 2020-12-18
Add international addresses to Support UI

raamSoftwire:
## Context

This ticket allows a support user to add/edit multi-line international addresses through the support UI.

As part of switching to multi-line international addresses, there is a bit of work to update existing international addresses in to the new format (https://trello.com/c/UVA1qhQJ/2754-migrate-international-addresses-in-production-to-new-multi-line-model). To aid this process the support UI will temporarily have both the old freetext input and the new multi-line input to enable support users to update these addresses. Once this work is complete the freetext input can be removed.

image

## Guidance to review

Is the feature flag implemented correctly? Do we need this functionality before we turn on the feature flag?

## Link to Trello card

https://trello.com/c/jJsZshMb/2751-add-multi-line-international-addresses-to-support-ui

## Things to check

### DIFF
--- a/app/forms/support_interface/application_forms/edit_address_details_form.rb
+++ b/app/forms/support_interface/application_forms/edit_address_details_form.rb
@@ -42,10 +42,10 @@ module SupportInterface
           address_line3: address_line3,
           address_line4: address_line4,
           postcode: postcode&.upcase,
-          international_address: nil,
           audit_comment: audit_comment,
         }
         attrs[:country] = 'GB' if uk?
+        attrs[:international_address] = international_address if international?
         application_form.update(attrs)
       end
 
--- a/app/forms/support_interface/application_forms/edit_address_details_form.rb
+++ b/app/forms/support_interface/application_forms/edit_address_details_form.rb
@@ -3,18 +3,20 @@ module SupportInterface
     class EditAddressDetailsForm
       include ActiveModel::Model
 
-      attr_accessor :address_line1, :address_line2, :address_line3,
-                    :address_line4, :postcode, :address_type, :country, :international_address, :audit_comment
+      attr_accessor :address_line1, :address_line2, :address_line3, :address_line4,
+                    :postcode, :address_type, :country, :international_address, :audit_comment
 
       validates :address_line1, :address_line3, :postcode, presence: true, on: :address, if: :uk?
-      validates :international_address, presence: true, on: :address, if: :international?
+
+      validates :address_line1, presence: true, on: :address, if: :international?
+
       validates :address_type, presence: true, on: :address_type
       validates :country, presence: true, on: :address_type, if: :international?
 
       validates :address_line1, :address_line2, :address_line3, :address_line4,
                 length: { maximum: 50 }, on: :address
 
-      validates :postcode, postcode: true, on: :address
+      validates :postcode, postcode: true, on: :address, if: :uk?
       validates :audit_comment, presence: true, on: :address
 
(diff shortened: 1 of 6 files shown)


app/forms/support_interface/application_forms/edit_address_details_form.rb
```diff
@@ -34,36 +36,25 @@ def self.build_from_application_form(application_form)
...
-            audit_comment: audit_comment,
-          )
-        end
+        attrs = {
+          address_line1: address_line1,
+          address_line2: address_line2,
+          address_line3: address_line3,
+          address_line4: address_line4,
+          postcode: postcode&.upcase,
+          international_address: nil,
```
stevehook, line 45: I'm wondering whether we should avoid overwriting this automatically so that we don't have to worry about what happens between the manual updates and activating the feature flag. I think this column can be dropped altogether when the dust settles.

app/views/support_interface/application_forms/address_details/edit.html.erb
```diff
@@ -7,20 +7,42 @@
...
-        <% if @details_form.uk? %>
-          <%= f.govuk_text_field :address_line1, label: -> { safe_join([t('application_form.contact_information.address_line1.uk.label'), ' ', tag.span(t('application_form.contact_information.address_line1.hidden'), class: 'govuk-visually-hidden')]) }, autocomplete: 'address-line1' %>
-          <%= f.govuk_text_field :address_line2, label: {text: t('application_form.contact_information.address_line2.uk.label'), hidden: true}, autocomplete: 'address-line2' %>
-          <%= f.govuk_text_field :address_line3, label: {text: t('application_form.contact_information.address_line3.uk.label')}, width: 'two-thirds', autocomplete: 'address-level2' %>
-          <%= f.govuk_text_field :address_line4, label: {text: t('application_form.contact_information.address_line4.uk.label')}, width: 'two-thirds', autocomplete: 'address-level1' %>
-          <%= f.govuk_text_field :postcode, label: {text: t('application_form.contact_information.postcode.uk.label')}, width: 10, autocomplete: 'postal-code' %>
-        <% else %>
+
+        <% if @details_form.international? && @details_form.international_address %>
```
stevehook, line 11: I don't think it would hurt to always display this field for international addresses, that way it'll be a bit more obvious to support users whether a given candidate has a legacy `international_address` or not. Longer term we remove it altogether.

stevehook (commented):
Tested the review app and works very nicely. Code looks good. I have a couple of queries about the way it works.

app/forms/support_interface/application_forms/edit_address_details_form.rb
raamSoftwire: I had the same thought but decided that since it's captured in the audit log that wouldn't be necessary.

app/views/support_interface/application_forms/address_details/edit.html.erb
raamSoftwire: Sure, I can see that might be useful. I'll make the other change as well then to keep the old address

davidgisbey:
Tested locally and it looks good to me. Good to go once you've applied Steve's suggestion. Have a great break!

davidgisbey (approved):

raamSoftwire:
I've pushed the changes relating to Steve's comments and tested them on the Heroku app. I'm happy for this to be merged once @stevehook has had a chance for another look.

stevehook (approved):
Tested review app again and all working well

PR #3686, opened 2020-12-18
Remove UCASMatch emails from Notifications check

despo:
## Context

Related to #3673 as we don't want to give provider users the ability to opt out of these emails.

## Link to Trello card
https://trello.com/c/h252hN4f/3145-remove-ucasmatch-emails-from-notifications-check

## Things to check

### DIFF
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -764,6 +764,7 @@ FactoryBot.define do
     email_address { "#{Faker::Name.first_name.downcase}-#{SecureRandom.hex}@example.com" }
     first_name { Faker::Name.first_name }
     last_name { Faker::Name.last_name }
+    send_notifications { Faker::Boolean.boolean(true_ratio: 0.5) }
 
     trait :with_provider do
       after(:create) do |user, _evaluator|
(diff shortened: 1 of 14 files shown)


duncanjbrown:
Code looks good! Could have a test?

despo:
@duncanjbrown instead of adding a test to verify that UCASMatches don't rely on this service, I have randomised the value of `send_notifications` in the provider_user factory and updated the tests that require provider_user to be able to receive notifications to set the data correctly. 

This way any test depending on the NotificationsList that does not explicitly set `send_notifications` to true should fail.

spec/factories.rb
```diff
@@ -764,6 +764,7 @@
     email_address { "#{Faker::Name.first_name.downcase}-#{SecureRandom.hex}@example.com" }
     first_name { Faker::Name.first_name }
     last_name { Faker::Name.last_name }
+    send_notifications { Faker::Boolean.boolean(true_ratio: 0.5) }
```
duncanjbrown, line 767: 🤯

duncanjbrown (approved):

PR #3687, opened 2020-12-18
Add missing h1 headings

frankieroberto:
This change adds `<h1>` elements to many form pages which currently only have a fieldset legend.

The `<h1>` heading is useful for users of assistive technologies, as it provides extra information about the page structure, and can be combined with the fieldset legend.

It looks like this is a regression caused by the removal of the `<h1>` being a default in [version 2.1.0 of the govuk_design_system_formbuilder implementation](https://github.com/DFE-Digital/govuk_design_system_formbuilder/releases/tag/v2.1.0).

### DIFF
--- a/app/components/provider_interface/edit_provider_user_permissions_component.html.erb
+++ b/app/components/provider_interface/edit_provider_user_permissions_component.html.erb
@@ -17,7 +17,7 @@
             label: { text: 'Extra permissions' },
             hint: { text: 'For example, being able to make offers and access sensitive content' } do %>
 
-      <%= f.govuk_check_boxes_fieldset :permissions, legend: { size: 'm', text: 'Select permissions' } do %>
+      <%= f.govuk_check_boxes_fieldset :permissions, legend: { text: 'Select permissions', size: 'm' } do %>
 
         <%= f.govuk_check_box :manage_organisations, true, multiple: false,
                               label: { text: 'Manage organisational permissions' },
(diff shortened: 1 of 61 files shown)


frankieroberto:
[Fix missing h1 tags in forms](https://trello.com/c/LyEJG52p/2763-fix-missing-h1-tags-in-forms)

duncanjbrown:
Nice. Would we prevent future accessibility issues if we configured `h1` as the default, per https://github.com/DFE-Digital/govuk_design_system_formbuilder/releases/tag/v2.1.0?

paulrobertlloyd:
Based on conversation with @frankieroberto before he went on leave for Christmas, have made these additional changes:

* Removed most occurrences of `legend: nil`, typically replacing a `h1` with using `h1` tag and content within the respective legend. The two remaining forms with `legend: nil` are where we have single checkboxes.
* Removed all occurrences of `tag: 'span'`. We don’t need to include span in legends. In removing these, in most cases I have replaced with adding `size: 'm'`. This is the default legend size, but have added to be explicit
* Removed occurrences of `tag: 'h2'`. In these cases a `legend` alone is enough
* Ensured consistent ordering of legend values where spotted: text, size, tag
* Consistent method names across the GCSE grade views
* Other quick fixes spotted while reviewing fieldsets

paulrobertlloyd (approved):

davidgisbey (approved):
Looks good. I'll look our for this when reviewing.

PR #3692, opened 2020-12-21
Return iso8601 values for DEGSTDT, DEGENDDT in HESA export

mnacos:
## Context

HESA requires dates to follow the ISO format, however we don't collect the day/month for degree start and end dates from candidates. We currently only return YYYY in the export. We should return what we have as YYYY-01-01 to help providers stay compliant when preparing their census submissions.

No UI changes.

## Guidance to review

Run the specs.

## Link to Trello card

https://trello.com/c/LPViuJit

## Things to check


app/decorators/application_choice_export_decorator.rb
```diff
@@ -24,6 +24,12 @@ def degrees_completed_flag
     application_form.degrees_completed ? 1 : 0
   end
 
+  def first_degree
```
duncanjbrown, line 27: :+1:

duncanjbrown (commented):
Looking good! Can we be more confident about the presence of a first degree (ie fewer `&`s) as it's a hard requirement for submitting the form?

mnacos:
> Looking good! Can we be more confident about the presence of a first degree (ie fewer `&`s) as it's a hard requirement for submitting the form?

Ok, le'ts do that and find out if any exports break. I've updated all `ApplicationChoice` factories to create applications with degrees.

app/services/provider_interface/hesa_data_export.rb
```diff
@@ -99,6 +98,14 @@ def qualification_aim(application)
 
       '021'
     end
+
+    def first_degree_year(application, year_type)
+      application.first_degree&.send year_type
```
duncanjbrown, line 103: Could lose the `&` here?

duncanjbrown (approved):
Nice — good to be bold!

PR #3693, opened 2020-12-21
Return prevalent nationality for HESA export

despo:
## Context

Exposes #nationality field in ApplicationChoiceDecorator satisfying conditions mentioned in notes section: https://www.hesa.ac.uk/collection/c18051/a/nation

> Where a student has dual nationality including British, they should be coded as United Kingdom (GB). If a dual nationality, not including British, but including non-UK EU country then use relevant EU country code. If neither British or non-UK EU country then code as either nationality.

Add`EU_COUNTRY_CODES` constant in nationalities initialiser and `nationality` field in `ApplicationChoice` decorator

## Link to Trello card

https://trello.com/c/428KZfgZ/3147-improve-nationality-value-calculation-in-hesa-export

## Things to check

### DIFF
--- a/config/initializers/nationalities.rb
+++ b/config/initializers/nationalities.rb
@@ -228,3 +228,44 @@ NATIONALITIES = [
 
 NATIONALITIES_BY_NAME = NATIONALITIES.map(&:reverse).to_h
 NATIONALITY_DEMONYMS = NATIONALITIES.map(&:second)
+
+# Last updated on 21/12/2020. To update run the following code in irb and update the EU_COUNTRY_CODES array with the result.
+# require 'net/http' # require 'json'
+# response = Net::HTTP.get('restcountries.eu', '/rest/v2/regionalbloc/eu?fields=alpha2Code;name')
+# JSON.parse(response).inject([]) { |response, country| response << [ country['alpha2Code'], country['name'] ] }
+
+EU_COUNTRY_CODES = [
+  'AX',
+  'AT',
+  'BE',
+  'BG',
+  'HR',
+  'CY',
+  'CZ',
+  'DK',
+  'EE',
+  'FO',
+  'FI',
+  'FR',
+  'GF',
+  'DE',
+  'GI',
+  'GR',
+  'HU',
+  'IE',
+  'IM',
+  'IT',
+  'LV',
+  'LT',
+  'LU',
+  'MT',
+  'NL',
+  'PL',
+  'PT',
+  'RO',
+  'SK',
+  'SI',
+  'ES',
+  'SE',
+  'GB'
+].freeze
--- a/app/decorators/application_choice_hesa_export_decorator.rb
+++ b/app/decorators/application_choice_hesa_export_decorator.rb
@@ -1,0 +1,8 @@
+class ApplicationChoiceHesaExportDecorator < ApplicationChoiceExportDecorator
+  def nationality
+    return 'GB' if nationalities.include?('GB')
+    return (nationalities & EU_COUNTRY_CODES).first if (nationalities & EU_COUNTRY_CODES).any?
+
+    nationalities.first
+  end
+end
--- a/app/services/provider_interface/hesa_data_export.rb
+++ b/app/services/provider_interface/hesa_data_export.rb
@@ -16,7 +16,7 @@ module ProviderInterface
       rows = []
 
       applications.each do |application_choice|
-        application = ApplicationChoiceExportDecorator.new(application_choice)
+        application = ApplicationChoiceHesaExportDecorator.new(application_choice)
 
         first_degree = application.application_form.application_qualifications
           .order(created_at: :asc)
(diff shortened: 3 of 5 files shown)


app/decorators/application_choice_export_decorator.rb
```diff
@@ -35,6 +35,13 @@ def nationalities
       .sort.partition { |e| %w[GB IE].include? e }.flatten
   end
 
+  def nationality
```
duncanjbrown, line 38: Logic looks good — but shouldn't this live in the `HESADataExport` as the rule is HESA-specific? I _think_ we still want to return multiple nationalities in the ordinary export.

config/initializers/nationalities.rb
```diff
@@ -228,3 +228,44 @@
 
 NATIONALITIES_BY_NAME = NATIONALITIES.map(&:reverse).to_h
 NATIONALITY_DEMONYMS = NATIONALITIES.map(&:second)
+
+# Last updated on 21/12/2020. To update run the following code in irb and update the EU_COUNTRY_CODES array with the result.
+# require 'net/http' # require 'json'
+# response = Net::HTTP.get('restcountries.eu', '/rest/v2/regionalbloc/eu?fields=alpha2Code;name')
+# JSON.parse(response).inject([]) { |response, country| response << [ country['alpha2Code'], country['name'] ] }
```
duncanjbrown, line 235: Useful — thank you!

spec/decorators/application_choice_export_decorator_spec.rb
```diff
@@ -111,4 +111,41 @@
...
+
+      it 'they should be coded as the EU country code' do
+        expect(decorated_application_form.nationality).to eq('CY')
+      end
+    end
+
+    context 'when dual nationality and both are non-UK EU countries' do
+      let(:application_form) { create(:application_form, first_nationality: 'Greek', second_nationality: 'Hungarian') }
+
+      it 'they should be coded as the EU country code' do
```
mnacos, line 138:
```suggestion
      it 'they should be coded as the first EU country code' do
```

spec/decorators/application_choice_export_decorator_spec.rb
```diff
@@ -111,4 +111,41 @@
...
+        expect(decorated_application_form.nationality).to eq('GR')
+      end
+    end
+
+    context 'when dual nationality and both are neither is British or EU' do
+      let(:application_form) { create(:application_form, first_nationality: 'Indian', second_nationality: 'American') }
+
+      it 'they can be coded as either country' do
```
mnacos, line 146:
```suggestion
      it 'they are coded as the first country' do
```

config/initializers/nationalities.rb
```diff
@@ -228,3 +228,44 @@
...
+  'MT',
+  'NL',
+  'PL',
+  'PT',
+  'RO',
+  'SK',
+  'SI',
+  'ES',
+  'SE',
+  'GB'
```
mnacos, line 270: With 11 days to go and the GB precedence rule it may be worth removing this last line :-/ 

There's also a gov.uk page at https://www.gov.uk/eu-eea but doesn't contain codes.

mnacos (commented):
Very minor comments from me. Looks good.

app/decorators/application_choice_export_decorator.rb
despo: We do. However as the decorator exposes fields that are specific to the ApplicationDataExport service only but is used both there and in the HESADataExport I assumed that it was purposely generic enough to be used by both services. 

HESADataExport does not seem like a good fit as only `call` is - and should be public - which means I cannot directly test the logic behind identifying the correct nationality.

I think a decorator specific to the HESA export would probably be the best approach.

config/initializers/nationalities.rb
despo: I did think that, however as this nationality would never be taken into account anyway I decided not to edit the returned service data.

spec/decorators/application_choice_export_decorator_spec.rb
despo: The condition that we need to satisfy is that they can be either country code, I just decided to return the first one as it seemed a bit pointless shuffling the data just to get a random result in this instance.

config/initializers/nationalities.rb
despo: We could work with names if we can get that JSON, but I don't think they provide that anywhere 🤷🏼‍♀️

spec/decorators/application_choice_export_decorator_spec.rb
mnacos: I guess you're stating what our business rule is, as a form of documentation, rather than focusing on the implementation? I think there is some value in preserving the nationality order specified by the candidate (which we do, unless we promote any of their other nationalities to assist them with funding), and being explicit about it in the specs. It's not a big thing, but saying 'either' makes me expect we're shuffling the array in the code, which we don't do.

config/initializers/nationalities.rb
mnacos: Yes, like you say it doesn't make much difference.

spec/decorators/application_choice_export_decorator_spec.rb
despo: 👍🏼 Reworded as part of moving to a new decorator

mnacos (approved):
Great! `ApplicationChoiceHesaExportDecorator` is exactly what was needed. 🚀

PR #3697, opened 2020-12-21
Populate ApplicationChoice.last_public_update_at

duncanjbrown:
## Context

We need the updated_at field in application API responses to accurately reflect when the data in the API has changed — not when just anything has changed upstream. This caused a problem when we eg migrated some enums, or geocoded some application forms: applications in the API had fresh updated_at dates, but the data hadn't changed.

The approach:

1. Cache a copy of the API response whenever we save an application choice or any other model which could change an application in the API — qualifications, work experience etc
2. When we save, generate a fresh API response and look at the old response in the cache
3. If the cache is different, we can touch `last_public_update_at` and update the cache with the new response

We need to keep track of `last_public_update_at` on the model itself because clients query the API using a `?since=` parameter, and it makes no sense to use `updated_at` for that and `last_public_update_at` for this.

Subsequent PRs will

- change API responses and query logic to use `last_public_update_at`
- use the cache to serve API responses!

We also need to populate `last_public_update_at` on all existing ApplicationChoices.

## Link to Trello card

https://trello.com/c/GIzhISxF/3118-clarify-and-enforce-rules-about-which-changes-to-an-application-form-should-mark-the-associated-application-choices-as-updated

## Things to check

### DIFF
--- a/app/models/application_experience.rb
+++ b/app/models/application_experience.rb
@@ -1,6 +1,11 @@
 class ApplicationExperience < ApplicationRecord
+  include AffectsApplicationAPIResponse
+
   validates :role, :organisation, :details, :start_date,
             presence: true
 
   validates :working_with_children, inclusion: { in: [true, false] }
+
+  belongs_to :application_form
+  has_many :application_choices, through: :application_form
 end
--- a/app/presenters/vendor_api/single_application_presenter.rb
+++ b/app/presenters/vendor_api/single_application_presenter.rb
@@ -16,7 +16,7 @@ module VendorAPI
           status: status,
           phase: application_form.phase,
           updated_at: application_choice.updated_at.iso8601,
-          submitted_at: application_form.submitted_at.iso8601,
+          submitted_at: application_form.submitted_at&.iso8601,
           personal_statement: personal_statement,
           interview_preferences: application_form.interview_preferences,
           reject_by_default_at: application_choice.reject_by_default_at&.iso8601,
--- a/app/services/initialise_application_response_cache_and_last_public_update_at.rb
+++ b/app/services/initialise_application_response_cache_and_last_public_update_at.rb
@@ -1,0 +1,11 @@
+class InitialiseApplicationResponseCacheAndLastPublicUpdateAt
+  def self.call
+    ApplicationChoice.where(last_public_update_at: nil)
+      .update_all('last_public_update_at = updated_at')
+
+    ApplicationChoice.all.each do |ac|
+      cache = ac.application_response_cache.presence || ac.build_application_response_cache
+      cache.refresh!
+    end
+  end
+end
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -47,6 +47,7 @@ Rails.application.configure do
     Bullet.unused_eager_loading_enable = false
     Bullet.counter_cache_enable = false
     Bullet.add_whitelist type: :n_plus_one_query, class_name: 'Audited::Audit', association: :user
+    Bullet.add_whitelist type: :n_plus_one_query, class_name: 'ApplicationChoice', association: :application_response_cache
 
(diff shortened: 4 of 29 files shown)


app/models/application_experience.rb
```diff
@@ -1,6 +1,11 @@
 class ApplicationExperience < ApplicationRecord
+  include AffectsApplicationAPIResponse
```
mnacos, line 2: Love this not least because it acts as documentation/reminder.

app/presenters/vendor_api/single_application_presenter.rb
```diff
@@ -16,7 +16,7 @@ def as_json
           status: status,
           phase: application_form.phase,
           updated_at: application_choice.updated_at.iso8601,
-          submitted_at: application_form.submitted_at.iso8601,
+          submitted_at: application_form.submitted_at&.iso8601,
```
mnacos, line 19: Perhaps we need a guard within `SingleApplicationPresenter` to return `nil` if the application choice is not in a state visible to providers, and therefore not worry about missing values. What do you think?

app/services/initialise_application_response_cache_and_last_public_update_at.rb
```diff
@@ -0,0 +1,11 @@
+class InitialiseApplicationResponseCacheAndLastPublicUpdateAt
```
mnacos, line 1: I thought the plan was to coordinate the population of `last_public_update_at` with the launch of the `:refresh_api_response_cache` callback by including a db migration. Is this missing? Do we still need this service if we do the migration?

config/environments/test.rb
```diff
@@ -47,6 +47,7 @@
     Bullet.unused_eager_loading_enable = false
     Bullet.counter_cache_enable = false
     Bullet.add_whitelist type: :n_plus_one_query, class_name: 'Audited::Audit', association: :user
+    Bullet.add_whitelist type: :n_plus_one_query, class_name: 'ApplicationChoice', association: :application_response_cache
```
mnacos, line 50: This seems a bit drastic, unless bullet also warns about writes/saves generating additional queries. If it's just the vendor api controllers/actions generating bullet warnings, perhaps we should be skipping bullet only for these specific actions, see https://github.com/flyerhzm/bullet#whitelist

spec/system/vendor_api/application_updates_spec.rb
```diff
@@ -0,0 +1,104 @@
+require 'rails_helper'
+
+RSpec.feature 'Application updated_at dates on the API' do
```
mnacos, line 3: 👍🏼

mnacos (commented):
Looks great. A few questions about a missing db migration (part of the rollout plan) and `Bullet`.

app/presenters/vendor_api/single_application_presenter.rb
duncanjbrown: I like the idea. I don't think the presenter should `nil`, though — might try skipping the cache?

app/services/initialise_application_response_cache_and_last_public_update_at.rb
duncanjbrown: The migration calls this service

config/environments/test.rb
duncanjbrown: Yeah, it's not good is it? I'll revisit this once I've fixed the specs

duncanjbrown:
Thanks for comments @mnacos  — I've broken some interesting bits and pieces in specs! Mostly to do with trying to generate API responses from ApplicationChoices in inconsistent states from factories. Cleaning that up — we might want to consider enforcing explicit `status` on choices we generate in tests because the code paths they end up going down can vary quite wildly

duncanjbrown:
Closing in favour of #3799

PR #3698, opened 2020-12-21
Fix illogical heading structures

paulrobertlloyd:
## Context

Our DAC December 2020 audit found 3 pages with illogical heading structures (i.e. a `h4` heading follows a `h2`). I found many, many more! ☺️

* Lots of adjusting heading levels (mostly `h3` to `h2`)
* Updated some components so that their default heading level can be overridden – and then overrode ’em
* Fixed a few more pages missing a `h1`

## Guidance to review

* Sorry, this is a big PR, but each change has been constrained to an individual commit.
* Should I update tests for those components where I’ve added a `heading_level` option? 

## Link to Trello card

https://trello.com/c/QitTcgV3/

## Things to check


app/views/content/sandbox.html.erb
```diff
@@ -4,46 +4,44 @@
...
 
-    <h2 class='govuk-heading-m'>You can:</h2>
+    <p class="govuk-body">You can:</p>
 
-    <ul class='govuk-list govuk-list--bullet'>
+    <ul class="govuk-list govuk-list--bullet">
       <li>sign in as a candidate and submit an application</li>
       <li>sign in as a training provider and process an application</li>
-      <li>test Apply’s <%= govuk_link_to 'API integrations', 'https://www.apply-for-teacher-training.service.gov.uk/api-docs' %></li>
+      <li>test Apply’s <%= govuk_link_to 'API integrations', api_docs_home_path %></li>
```
paulrobertlloyd, line 14: Is there a reason why we hard code the URL for the API docs on the sandbox start page? @duncanjbrown

duncanjbrown: Yes — we want providers to use the production docs as the source of truth, because that's ultimately where they're going to be connecting to. Not that production has ever differed from Sandbox, but it felt right.

duncanjbrown:
This looks very thorough — thank you! Can we get `axe` to "review" this PR?

app/views/content/sandbox.html.erb
paulrobertlloyd: Ah, that makes sense. Will revert.

paulrobertlloyd:
> This looks very thorough — thank you! Can we get `axe` to "review" this PR?

I don’t know much about how our accessibility tests work, or how much they cover. Going by [these files](https://github.com/DFE-Digital/apply-for-teacher-training-tests/tree/master/cypress/integration), perhaps not that much?

Can we have axe run tests in the CLI against generated code or something?

duncanjbrown:
> > This looks very thorough — thank you! Can we get `axe` to "review" this PR?
> 
> I don’t know much about how our accessibility tests work, or how much they cover. Going by [these files](https://github.com/DFE-Digital/apply-for-teacher-training-tests/tree/master/cypress/integration), perhaps not that much?
> 
> Can we have axe run tests in the CLI against generated code or something?

That would be really cool. We could add [something like this](https://dev.to/chrislewis60/introducing-automated-accessibility-testing-in-rails-apps-with-rspec-and-axe-4hch) and sprinkle assertions of `expect(page).to be_axe_clean` everywhere. Or I wonder if we could get Capybara to automatically dump every view it renders into a temporary folder and then unleash `axe` on that at the end of the test suite?

paulrobertlloyd:
> We could add [something like this](https://dev.to/chrislewis60/introducing-automated-accessibility-testing-in-rails-apps-with-rspec-and-axe-4hch) and sprinkle assertions of `expect(page).to be_axe_clean` everywhere.

Ooooh, now you’re talking! That’d be superb!

duncanjbrown (approved):
# This
## PR
### Is
#### An
##### Improvement!

PR #3699, opened 2020-12-21
Fix for logic DegreeType controller to mark section incomplete on create

george-bit:
## Context

Bug was uncovered that while adding an additional degree section remained completed if flow was exited
 after degree type only was given. This could cause problems as application form submittal validates on
on section completion, so an incomplete qualification could be submitted. 

Action added to controller so that when degree type is created section is marked as incomplete.

## Guidance to review

Please test in review app and advise RE: system spec if it is necessary.

## Link to Trello card

https://trello.com/c/lWXlNeUg/2729-buggy-ux-when-marking-degree-section-complete

## Things to check


davidgisbey (approved):
Nice. This should do it 👍

PR #3707, opened 2020-12-22
Migration to add `authentication_tokens.used_at` column

stevehook:
## Context

Following on from https://github.com/DFE-Digital/apply-for-teacher-training/pull/3611 we can simplify the URLs used to expired and invalid tokens to drop the `u` param (that contains an encrypted user id). To achieve this we need to retain used (and therefore invalid) tokens in the database so that we can identify a user by the token alone. This is a migration-only PR that adds the `used_at` column to the `authentication_tokens` table.

- [x] Migration to add an `authentication_tokens.used_at` column. Note that this column allows `null` values - `null` indicates an unused token.

## Guidance to review

- Does it make sense to make this a timestamp rather than a plain boolean and can you think of a better name?
- Backfill is not needed for this column because all existing authentication tokens will be unused as they are currently deleted immediately after first use. Therefore all values should be `null` initially.

## Link to Trello card

https://trello.com/c/zEo4y16R/2626-dev-change-the-way-we-detect-users-when-a-token-is-expired-and-make-the-url-nice

## Things to check


davidgisbey (approved):
Yea that's way better than a bool. Nice thinking 👍

PR #3709, opened 2020-12-22
Make user_id & user_not not nullable and add index to AuthenticationToken table

davidgisbey:
## Context

This PR covers a bit of clean up work from the magic link refactor.

It removes a couple of unused columns, makes user_type & user_id not nullable and creates an index.

- Migration to remove authenticable_id and authenticable_type
- Migration to make user_id and user_type not nullable & adds an index to auth tokens table  

## Things to check

### DIFF
--- a/db/migrate/20201222123211_remove_unused_sign_in_related_columns.rb
+++ b/db/migrate/20201222123211_remove_unused_sign_in_related_columns.rb
@@ -1,0 +1,6 @@
+class RemoveUnusedSignInRelatedColumns < ActiveRecord::Migration[6.0]
+  def change
+    remove_column :authentication_tokens, :authenticable_id, :bigint
+    remove_column :authentication_tokens, :authenticable_type, :string
+  end
+end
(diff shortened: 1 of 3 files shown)


db/migrate/20201222123211_remove_unused_sign_in_related_columns.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveUnusedSignInRelatedColumns < ActiveRecord::Migration[6.0]
+  def change
+    remove_column :authentication_tokens, :authenticable_id, :bigint
+    remove_column :authentication_tokens, :authenticable_type, :string
```
stevehook, line 4: Just curious, I guess the `remove_column` will remove the `index_authentication_tokens_on_id_and_type` index automatically? (It must do otherwise it would trigger a DB error).

db/migrate/20201222134529_make_user_id_and_user_type_not_nullable_and_add_index.rb
```diff
@@ -0,0 +1,7 @@
+class MakeUserIdAndUserTypeNotNullableAndAddIndex < ActiveRecord::Migration[6.0]
+  def change
+    change_column_null :authentication_tokens, :user_id, false
+    change_column_null :authentication_tokens, :user_type, false
```
stevehook, line 4: Might be worth double checking for `null` values in the production DB before deploying. (There shouldn't be any because there is a constraint at the active record level).

stevehook (approved):
All looks good to me.

db/migrate/20201222123211_remove_unused_sign_in_related_columns.rb
davidgisbey: Yea it does 👍

PR #3710, opened 2020-12-22
Remove ambiguous graphics from accessibility tree

paulrobertlloyd:
## Context

The DAC December 2020 audit highlighted a number of issues with our SVG diagrams, which are inaccessible – but fortunately have accessible alternatives presented on the same page they appear.

It also highlighted issues with using emoji in state descriptions. 

* Splits out `event_emoji` value from `StateDiagram`, and removes this from accessibility tree when shown in HTML text
* Removes the smaller state diagrams from the accessibility tree (their function is replicated in the text they sit besides)
* Replaces `app-relationship-diagram` with `app-diagram`. This consolidates the SVG styling that was embedded in various pages, and also fixes it so that SVG diagrams always appear with the correct aspect ratio.
* Creates a new `state_diagram` partial, that is used in support and the API docs. This consistently hides the full state diagram behind a details disclosure, which has a non-visual description to explain that an accessible version of the diagram appears below.

## Guidance to review

* ~I have saved the state_diagram partial in a new folder: `views/shared` – is this correct?~ This is now a component: `StateDiagramComponent`
* Are we happy with the new label for the details link: ‘► View diagram of application flow (accessible example presented after diagram)’

## Link to Trello card

## Things to check

### DIFF
--- a/app/components/state_diagram_component.html.erb
+++ b/app/components/state_diagram_component.html.erb
@@ -1,0 +1,13 @@
+<details class="govuk-details" data-module="govuk-details">
+  <summary class="govuk-details__summary">
+    <span class="govuk-details__summary-text">
+      View diagram of application flow
+      <span class="govuk-visually-hidden">(accessible example presented after diagram)</span>
+    </span>
+  </summary>
+  <div class="govuk-details__text">
+    <div class="app-diagram">
+      <%= StateDiagram.svg(only_from_state: nil, machine: @machine) %>
+    </div>
+  </div>
+</details>
--- a/app/lib/state_diagram.rb
+++ b/app/lib/state_diagram.rb
@@ -15,7 +15,7 @@ class StateDiagram
         graph.add_edges(
           state.name.to_s,
           event.transitions_to.to_s,
-          label: event_name(state.name, event, namespace),
+          label: event_emoji(state.name, event, namespace) + ' ' + event_name(state.name, event, namespace),
           fontname: 'Arial',
           color: '#0b0c0c',
           fontcolor: '#0b0c0c',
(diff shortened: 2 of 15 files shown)


app/lib/state_diagram.rb
```diff
@@ -58,15 +58,17 @@ def self.svg(only_from_state: nil, machine:)
   end
 
   def self.event_name(state, event, namespace)
+    I18n.t!("#{namespace}events.#{state}-#{event}.name")
+  end
+
+  def self.event_emoji(state, event, namespace)
```
duncanjbrown, line 64: :+1: good split

app/views/api_docs/pages/lifecycle.html.erb
```diff
@@ -7,11 +7,7 @@
   { name: 'When emails are sent', url: api_docs_when_emails_are_sent_path },
 ]) %>
 
-<%= StateDiagram.svg(only_from_state: nil, machine: ApplicationStateChange) %>
-
-<style>
-  svg { max-width: 100%; }
-</style>
+<%= render StateDiagramComponent.new(machine: ApplicationStateChange) %>
```
duncanjbrown, line 10: Nice to get the benefit of the shared markup here

app/views/support_interface/providers/users.html.erb
```diff
@@ -4,6 +4,6 @@
 
 <%= render(SupportInterface::ProviderUsersTableComponent.new(provider_users: @provider.provider_users)) %>
 
-<div class="app-relationship-diagram">
+<div class="app-diagram" aria-hidden="true">
```
duncanjbrown, line 7: :+1: — all the information is available on the relationships tab anyway

app/components/state_diagram_component.html.erb
```diff
@@ -0,0 +1,13 @@
+<details class="govuk-details" data-module="govuk-details">
+  <summary class="govuk-details__summary">
+    <span class="govuk-details__summary-text">
+      View diagram of application flow
+      <span class="govuk-visually-hidden">(accessible example presented after diagram)</span>
```
duncanjbrown, line 5: This is... quite classy!?

duncanjbrown (approved):
Nice work @paulrobertlloyd!

app/components/state_diagram_component.html.erb
paulrobertlloyd: GDS sure like their classes!

paulrobertlloyd: Once we get the GOVUK components gem implemented, this will all be hidden away 😉

duncanjbrown: No no — I meant the text! 🤵

paulrobertlloyd: Ooh, bit like “your feature presentation will start after these messages“ 😂

PR #3711, opened 2020-12-22
Indicate no permissions in organisation relationship diagram when no permissions setup

despo:
## Context

Display ‘Permissions not setup’ when none are setup between a training and a ratifying provider instead of unchecked permissions

example-permissions

## Link to Trello card

https://trello.com/c/m31Hqrtx/3152-show-on-org-org-relationship-diagram-when-permissions-have-not-been-set-up

## Things to check

### DIFF
--- a/app/presenters/support_interface/provider_relationships_diagram.rb
+++ b/app/presenters/support_interface/provider_relationships_diagram.rb
@@ -1,5 +1,11 @@
 module SupportInterface
   class ProviderRelationshipsDiagram
+    AVAILABLE_PERMISSIONS = {
+      view_safeguarding_information: 'view safeguarding',
+      view_diversity_information: 'view diversity',
+      make_decisions: 'make decisions',
+    }.freeze
+
     def initialize(provider:)
       @provider = provider
     end
@@ -33,7 +39,7 @@ module SupportInterface
           graph.add_edges(
             "Provider##{perm.training_provider_id}",
             "Provider##{perm.ratifying_provider_id}",
-            label: "can #{translate_training_permissions(perm)} for courses ratified by",
+            label: translate_training_permissions(perm),
             fontname: 'Arial',
             color: '#0b0c0c',
             fontcolor: '#0b0c0c',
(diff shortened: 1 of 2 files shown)


app/presenters/support_interface/provider_relationships_diagram.rb
```diff
@@ -78,19 +84,29 @@ def providers
...
+      return 'Permissions not setup' if permissions_not_setup(:ratifying_provider, permission)
+
+      "can #{build_permissions_string(:ratifying_provider, permission)} for courses run by"
     end
 
     def value_indicator(permission_value)
       permission_value ? '✅' : '❌'
     end
+
+    def permissions_not_setup(provider_type, permission)
```
mnacos, line 102: I think `permissions_not_setup` is not checking the right things. It's perfectly possible for either side of the relationship to have no permission to do anything (other than see the application), while the relationship has been properly set up. I think 'set up', in this instance, means that the training provider has been through the 'organisational permissions' flow.

There are other ways of detecting that this has not happened, such as checking if the relationship is `valid?` (which means at least one side has permission for each rule) but the simplest way is  checking for the presence of `ProviderRelationshipPermissions#setup_at`.

spec/system/support_interface/viewing_provider_permissions_spec.rb
```diff
@@ -37,6 +42,23 @@ def and_there_are_two_providers_in_a_partnership
     )
   end
 
+  def and_there_are_two_other_providers_in_a_partnership_with_no_permissions_configured
+    training = create(:provider, sync_courses: true, name: 'City Learning Trust')
+    ratifying = create(:provider, sync_courses: true, name: 'Staffordshire University (S72)')
+
+    create(
```
mnacos, line 49: This is a valid relationship (as far as permissions validation is concerned) and the only way any of these permissions would be true is if the training provider has completed the organisational permissions flow, except this would also have populated `setup_at`. Please see my other comment.

mnacos (commented):
Loving the updated code structure for this, but I think there's a misunderstanding about when 'Permissions not setup' should be displayed. My understanding is we should only show this text instead of individual crosses for the permissions a provider doesn't have if the training provider has not completed the 'organisational permissions' flow. In this case the red crosses are misleading, as all permissions are off by default.

mnacos (approved):
Great :shipit:

PR #3712, opened 2020-12-22
Improve sites data export

TobyRet:
## Context

In order for this export to be useful, we need to be able to filter out sites which are no longer available. 

Made it so! 🖖 

## Link to Trello card

https://trello.com/c/zRMqdpyq/2710-dev-improve-the-query-we-use-for-the-sites-data-export

## Things to check

### DIFF
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -1,5 +1,6 @@
 class Site < ApplicationRecord
   belongs_to :provider
+  has_many :course_options
 
   audited associated_with: :provider
 
--- a/app/services/support_interface/sites_export.rb
+++ b/app/services/support_interface/sites_export.rb
@@ -18,7 +18,10 @@ module SupportInterface
   private
 
     def relevant_sites
-      Site.includes(:provider).where(providers: { sync_courses: true }).order('providers.code ASC')
+      Site.joins(:course_options, :provider)
+          .where(course_options: { site_still_valid: true })
+          .where(providers: { sync_courses: true })
+          .order('providers.code ASC')
     end
   end
 end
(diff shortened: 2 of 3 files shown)


app/models/site.rb
```diff
@@ -1,5 +1,6 @@
 class Site < ApplicationRecord
   belongs_to :provider
+  has_one :course_option
```
stevehook, line 3: Isn't this a `has_many` association?

app/services/support_interface/sites_export.rb
```diff
@@ -18,7 +18,10 @@ def sites
   private
 
     def relevant_sites
-      Site.includes(:provider).where(providers: { sync_courses: true }).order('providers.code ASC')
+      Site.includes(:course_option, :provider)
```
stevehook, line 21: `includes` will generate a `LEFT OUTER JOIN` in this case but I think you need an `INNER JOIN` so I think you might need a `joins(:course_options, :provider)` if I understand correctly (a big 'if').

TobyRet: Yeah you're right - will make the change 👍

app/models/site.rb
TobyRet: Yep - I'll make the change

app/services/support_interface/sites_export.rb
TobyRet:
```
[5] pry(main)> Site.joins(:course_options, :provider).where(course_options: { site_still_valid: true }).where(providers: { sync_courses: true }).order('providers.code ASC')
  Site Load (5.2ms)  SELECT "sites".* FROM "sites" INNER JOIN "course_options" ON "course_options"."site_id" = "sites"."id" INNER JOIN "providers" ON "providers"."id" = "sites"."provider_id" WHERE "course_options"."site_still_valid" = $1 AND "providers"."sync_courses" = $2 ORDER BY providers.code ASC  [["site_still_valid", true], ["sync_courses", true]]
  ```

davidgisbey (approved):

PR #3716, opened 2020-12-22
Fix issues with validation errors and summaries

paulrobertlloyd:
## Context

The DAC December 2020 report spotted a number of issues with our validation errors.

* Adds a missing validation error for training with a disability page
* Consistent controllers for training with a disability and safeguarding questions
* Ensure error summaries link to the fieldset when it is causing an error on the page.

  Thanks to @peteryates for alerting me to needing to use `link_errors: true` on the first radio/checkbox in a fieldset. In quite a few cases this was missed off, in other cases it was added to the every option, and in a few cases not added to the first option. This is something we’ll need to keep an eye on when reviewing.

## Link to Trello card

## Things to check

### DIFF
--- a/app/views/candidate_interface/application_choices/review.html.erb
+++ b/app/views/candidate_interface/application_choices/review.html.erb
@@ -24,13 +24,14 @@
         <% if @application_choices.count >= 1 %>
           <%= f.govuk_check_boxes_fieldset :course_choices_completed, legend: nil do %>
             <%= f.hidden_field :course_choices_completed, value: false %>
-            <%= f.govuk_check_box :course_choices_completed,
+            <%= f.govuk_check_box(:course_choices_completed,
               true,
               multiple: false,
-              link_errors: true,
               label: {
                 text: t('application_form.courses.complete.completed_checkbox'),
-              } %>
+              },
+              link_errors: true,
+            ) %>
           <% end %>
 
           <%= f.govuk_submit 'Continue' %>
--- a/app/views/candidate_interface/degrees/grade/_form_fields.html.erb
+++ b/app/views/candidate_interface/degrees/grade/_form_fields.html.erb
@@ -3,7 +3,7 @@
   <%= f.govuk_radio_buttons_fieldset :grade, legend: { text: @page_title, size: 'xl', tag: 'h1' } do %>
     <p class="govuk-body"><%= t('application_form.degree.grade.international.grade_examples') %></p>
 
-    <%= f.govuk_radio_button :grade, 'other', label: { text: 'Yes' } do %>
+    <%= f.govuk_radio_button :grade, 'other', label: { text: 'Yes' }, link_errors: true do %>
       <% if degree.predicted_grade? %>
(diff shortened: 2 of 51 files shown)


config/locales/application_form.yml
```diff
@@ -962,6 +962,10 @@ en:
           attributes:
             ethnic_group:
               blank: Choose your ethnic group
+        candidate_interface/training_with_a_disability_form:
+          attributes:
+            disclose_disability:
+              inclusion: Choose if you want to ask for help to become a teacher
```
paulrobertlloyd, line 968: Any idea why `training_with_a_disability_form` validates on `inclusion`, but the practically similar `safeguarding_form` validates on `blank`?

app/views/candidate_interface/application_choices/review.html.erb
```diff
@@ -24,13 +24,14 @@
...
+            <%= f.govuk_check_box(:course_choices_completed,
               true,
               multiple: false,
-              link_errors: true,
               label: {
                 text: t('application_form.courses.complete.completed_checkbox'),
-              } %>
+              },
+              link_errors: true,
+            ) %>
```
paulrobertlloyd, line 34: Just changed the order so that `link_errors` is the last option

config/locales/application_form.yml
duncanjbrown: Ordinary inconsistency — inclusion is better though, I think?

app/views/candidate_interface/degrees/grade/_form_fields.html.erb
```diff
@@ -13,7 +13,7 @@
       <%= f.govuk_text_field(:other_grade, text_field_options) %>
     <% end %>
     <% @international_main_grades.each_with_index do |grade, i | %>
```
duncanjbrown, line 15: I think we can lose the `each_with_index` here now?

duncanjbrown (approved):
🚢  This is an improvement even if we've missed a few.

config/locales/application_form.yml
paulrobertlloyd: I mean, `safeguarding_form` has to validate on `blank`, while `training_with_a_disability_form` has to validate on `inclusion`, what’s different behind the scenes in how these values are stored (just curious)?

app/views/candidate_interface/degrees/grade/_form_fields.html.erb
paulrobertlloyd: Good catch!

config/locales/application_form.yml
duncanjbrown: The disability form stores a boolean 

Screenshot 2021-01-05 at 11 54 15

whereas the safeguarding form translates the answer to an enum

Screenshot 2021-01-05 at 11 53 42

This needn't make a difference to the validation logic, though — afaics it's still a required field and the acceptable answers are still 'yes' and 'no', I think?

paulrobertlloyd: Ah-ha, thanks! I’ll try and pretend I never saw this inconsistency 😆 Validation logic is still the same, just that the translation string keys need to be different.

PR #3717, opened 2020-12-23
Change the way we detect users when a token is expired

stevehook:
## Context

Following on from https://github.com/DFE-Digital/apply-for-teacher-training/pull/3611 we can now drop the `u` (encrypted user id) in URLs for regenerating expired or used tokens.

This PR should not be merged until https://github.com/DFE-Digital/apply-for-teacher-training/pull/3707 is deployed because it depends on a migration in that PR.

- Remove `u` param from `CandidateInterface::SigninController#confirm_authentication`, `authenticate` and `expired` actions. The `token` param is now sufficient to identify users.
- Rather than delete `AuthenticationToken`s after a successful login we now just set the `used_at` timestamp. This enables us to find any token in the DB for id purposes regardless of whether it's been used or not.
- Ensure that `AuthenticationToken#still_valid?` takes account of `used_at` (return false if it's set).

## Guidance to review

- Being a change to authentication this PR requires very thorough review and testing.

## Link to Trello card

https://trello.com/c/zEo4y16R/2626-dev-change-the-way-we-detect-users-when-a-token-is-expired-and-make-the-url-nice

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/sign_in_controller.rb
+++ b/app/controllers/candidate_interface/sign_in_controller.rb
@@ -21,7 +21,7 @@ module CandidateInterface
         raw_token: params[:token],
       )
 
-      if authentication_token && authentication_token.still_valid?
+      if authentication_token&.still_valid?
         render 'confirm_authentication'
       elsif authentication_token
         # If the token is expired, redirect the user to a page
--- a/app/controllers/candidate_interface/sign_in_controller.rb
+++ b/app/controllers/candidate_interface/sign_in_controller.rb
@@ -23,11 +23,11 @@ module CandidateInterface
 
       if authentication_token && authentication_token.still_valid?
         render 'confirm_authentication'
-      elsif params[:u]
-        # If the token is invalid or expired, redirect the user to a page
-        # with their encrypted user id as a param where they can request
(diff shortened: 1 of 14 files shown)


davidgisbey (commented):
Looks good. The factory will have to be changed to something like this to pass.

```
factory :authentication_token do
    user { create(:support_user) }
    hashed_token { '1234567890' }
 end
```

app/controllers/candidate_interface/sign_in_controller.rb
```diff
@@ -23,11 +23,11 @@ def confirm_authentication
 
       if authentication_token && authentication_token.still_valid?
         render 'confirm_authentication'
-      elsif params[:u]
-        # If the token is invalid or expired, redirect the user to a page
-        # with their encrypted user id as a param where they can request
+      elsif authentication_token
```
TobyRet, line 26: Not part of your work, but note line 24 can be changed to 
`authentication_token&.still_valid?`

TobyRet (approved):
Tested it out locally and LGTM

davidgisbey:
Still lgtm

davidgisbey (approved):

PR #3718, opened 2020-12-23
Reinstate missing label for deprecated work breaks page

paulrobertlloyd:
## Context

In 3857e0f458e2e02ddef5ee9edd2940231b63138f I updated the localised strings for work breaks, and in doing so, inadvertently broke this deprecated – but annoyingly still present – page for explaining work breaks. I regret the error.

* Adds the missing string
* Ordering of attributes 💅 
* Removal of line breaks in paragraphs 💅 

## Things to check

### DIFF
--- a/app/views/candidate_interface/work_history/breaks/edit.html.erb
+++ b/app/views/candidate_interface/work_history/breaks/edit.html.erb
@@ -6,11 +6,9 @@
     <%= form_with model: @work_breaks_form, url: candidate_interface_work_history_breaks_path do |f| %>
       <%= f.govuk_error_summary %>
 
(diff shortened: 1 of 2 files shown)


stevehook (approved):
All looks good to me. That page works on the review app (though had a bit of trouble finding it).

paulrobertlloyd:
Thanks Steve. Yeah, this page only appears for newly generated fake applications, so isn’t user facing. Theo wasn’t keen on us removing this page or the faker, but I’d like to revisit this decision as it no longer represents what we’re showing to candidates – and probably would mean we could remove a bit of complexity too.

PR #3719, opened 2020-12-23
Different namespaces for service feedback and application rejection styles

paulrobertlloyd:
## Context

I noticed a potential area of conflict in our CSS styles, only avoided due to these (currently) living in separate interfaces. We have `app-feedback__container` for the module requesting service feedback that appears above the footer in the candidate interface, and `app-feedback` for the rejection reasons in the provider interface.

### Rejection component
* Given the original SCSS file was called `reject.scss` and the markup appears in the `reasons_for_rejection_component.html.erb` file, updated the namespace for this component to `app-rejection` (and updated the CSS file name to match)
* Updated the class name for the change link:
  * Use BEM `__*` notation for a child class
  * Use `actions` suffix to match pattern used in design system components: `app-rejection__actions`
  * Only apply the class to the containing paragraph, not the link as well
* Removed an unnecessary containing `div`
* Fixing the positioning of the change link so that it appears in the top right of the parent container by removing the margin on the paragraph
* Used `govuk-spacing` helper for margin value

### Feedback component
* Renamed `app-feedback__container` to `app-feedback` since we no longer need the distinction between parent `div` and inner container element given the redesign of this component

## Things to check

### DIFF
--- a/app/components/candidate_interface/application_feedback_component.html.erb
+++ b/app/components/candidate_interface/application_feedback_component.html.erb
@@ -1,6 +1,6 @@
 <% if FeatureFlag.active?(:feedback_prompts) %>
   <div class="govuk-width-container">
-    <div class="app-feedback__container">
+    <div class="app-feedback">
       <p class="govuk-body govuk-!-margin-bottom-0 govuk-!-margin-top-4">
         <%= govuk_link_to(
           t('application_feedback.feedback_link'),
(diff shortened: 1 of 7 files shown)


steventux (changes requested):
Markup & CSS changes look good to me, some of our tests rely on the old `.app-feedback` class though.

paulrobertlloyd:
> Markup & CSS changes look good to me, some of our tests rely on the old `.app-feedback` class though.

Eek! All tests now fixed.

steventux (approved):
:+1:

PR #3720, opened 2020-12-23
Remove ambiguous form labelling on nationality forms

paulrobertlloyd:
## Context

The DAC December audit spotted a few issues with the way we enable users to add/remove additional nationalities. 

* Places the remove link _before_ the form label, not within it
* Fixes the title of the remove link (`RemoveSecond nationality` → `Remove second nationality`)
* Makes form labels unique by adding ordinal within hidden text (‘Nationality’ → ‘[Second] Nationality`)
* Add `event.preventDefault` to remove click handler so that clicking ‘Remove’ doesn’t take user to top of the page
* Add missing `h1` to nationalities form in support 

## Guidance to review

For adding the hidden label text, I’ve used the same `safe_join()` method we’re using on the address page.

## Link to Trello card

## Things to check

### DIFF
--- a/app/frontend/packs/nationalities-component.js
+++ b/app/frontend/packs/nationalities-component.js
@@ -30,20 +30,21 @@ const nationalitiesComponent = () => {
   hideSection(thirdSelectEl, thirdFormLabel);
 
   function addRemoveLink(labelEl, selectEl) {
+    const parentEl = labelEl.parentElement;
+
     const removeLink = document.createElement("a");
     removeLink.innerHTML = "Remove";
     removeLink.classList.add("govuk-link", "app-nationality__remove-link");
 
     // This has to be a link and not a button as the govuk-link class requires an
-    // href to apply  it's styling
+    // href to apply it's styling
     removeLink.href = "#";
-    labelEl.appendChild(removeLink);
+    parentEl.insertBefore(removeLink, labelEl);
 
     if (labelEl == secondFormLabel) {
       addNthNationalityHiddenSpan(removeLink, 'Second');
     } else {
       addNthNationalityHiddenSpan(removeLink, 'Third');
-
     }
 
     removeLink.addEventListener("click", function () {
--- a/app/frontend/packs/nationalities-component.js
+++ b/app/frontend/packs/nationalities-component.js
@@ -55,7 +55,7 @@ const nationalitiesComponent = () => {
   function addNthNationalityHiddenSpan(removeLink, nthNationality) {
     const nthNationalitySpan = document.createElement("span");
     nthNationalitySpan.classList.add("govuk-visually-hidden");
-    nthNationalitySpan.innerHTML = `${nthNationality} nationality`;
+    nthNationalitySpan.innerHTML = ` ${nthNationality.toLowerCase()} nationality`;
     removeLink.appendChild(nthNationalitySpan);
   }
 
--- a/app/frontend/packs/nationalities-component.js
+++ b/app/frontend/packs/nationalities-component.js
@@ -47,7 +47,8 @@ const nationalitiesComponent = () => {
       addNthNationalityHiddenSpan(removeLink, 'Third');
     }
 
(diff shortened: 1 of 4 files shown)


app/frontend/packs/nationalities-component.js
```diff
@@ -30,31 +30,33 @@ const nationalitiesComponent = () => {
...
   function addRemoveLink(labelEl, selectEl) {
+    const parentEl = labelEl.parentElement;
+
     const removeLink = document.createElement("a");
     removeLink.innerHTML = "Remove";
     removeLink.classList.add("govuk-link", "app-nationality__remove-link");
 
     // This has to be a link and not a button as the govuk-link class requires an
-    // href to apply  it's styling
+    // href to apply it’s styling
```
TobyRet, line 40: Apostrophe not req'd

app/views/candidate_interface/personal_details/nationalities/_shared_form.html.erb
```diff
@@ -22,9 +22,27 @@
...
   ) do %>
-    <%= f.govuk_collection_select :other_nationality1, select_nationality_options, :id, :name, label: { text: t('application_form.personal_details.nationality.label') } %>
-    <%= f.govuk_collection_select :other_nationality2, select_nationality_options, :id, :name, label: { text: t('application_form.personal_details.nationality.label') } %>
-    <%= f.govuk_collection_select :other_nationality3, select_nationality_options, :id, :name, label: { text: t('application_form.personal_details.nationality.label') } %>
+    <%= f.govuk_collection_select(
+      :other_nationality1,
+      select_nationality_options,
+      :id,
+      :name,
+      label: ->{ safe_join([tag.span('First ', class: 'govuk-visually-hidden'), t('application_form.personal_details.nationality.label')]) },
```
TobyRet, line 30: This seems like quite a fundamental change, but I don't really a sense from the commit message why this has occurred. A more descriptive commit in this instance might serve as useful documentation further down the line?

app/frontend/packs/nationalities-component.js
```diff
@@ -30,31 +30,33 @@ const nationalitiesComponent = () => {
...
     removeLink.href = "#";
-    labelEl.appendChild(removeLink);
+    parentEl.insertBefore(removeLink, labelEl);
```
TobyRet, line 42: Was this change the result of an accessibility issue? If so might be worth adding to the commit msg ...

TobyRet (changes requested):
All looks reasonable but I think it may be worth amending some commit messages so it is clear why the changes were made (i.e. what accessibility issue they address). Doing so will serve as useful documentation for the future.

app/frontend/packs/nationalities-component.js
paulrobertlloyd: Ha! Blinded by the horror of an seeing ‘unsexy’ quote, completely neglected to see that it was unneeded 😆

paulrobertlloyd:
@TobyRet Thanks for the feedback regarding the commit messages – never thought that much about adding additional detail before, but it makes sense. Do you have a rule when you do this?

Anyway, take a look, see what you think, hopefully the reason for these changes are a bit clearer now.

TobyRet (approved):
LGMT Paul - those commit msgs are a lot clearer now

TobyRet:
> @TobyRet Thanks for the feedback regarding the commit messages – never thought that much about adding additional detail before, but it makes sense. Do you have a rule when you do this?
> 
> Anyway, take a look, see what you think, hopefully the reason for these changes are a bit clearer now.

Generally I try and follow the [GDS guidance on commit messages](https://www.gov.uk/service-manual/technology/maintaining-version-control-in-coding#writing-commit-messages) which is inspired by other sources. I recall you saying there is a need to increase awareness amongst devs re: writing code that meets accessibility standards - descriptive commit messages is one tool at our disposal, alongside other things like Axe ...

PR #3722, opened 2020-12-23
Unique sub-section label for candidate accounts in support

paulrobertlloyd:
## Context

The DAC December 2020 audit spotted an issue in the support UI in that within the ‘Candidates’ section, we have a ‘Candidates’ section – duplicate links which may be confusing.

* Rename the ‘Candidates’ sub-section to ‘Candidate accounts’
* Update the breadcrumbs in this section so that the first item always returns you to the first page in this section, not the second. 

## Link to Trello card

## Things to check

### DIFF
--- a/app/components/support_interface/candidates_table_component.html.erb
+++ b/app/components/support_interface/candidates_table_component.html.erb
@@ -2,7 +2,7 @@
   <table class="govuk-table">
     <thead class="govuk-table__head">
       <tr class="govuk-table__row">
-        <th scope="col" class="govuk-table__header">Candidate</th>
+        <th scope="col" class="govuk-table__header">Email address</th>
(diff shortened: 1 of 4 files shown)


stevehook (approved):
One thing I find confusing (in the past and still on the review app) is that the breadcrumb 'top-level' is labelled _Candidate_ whether you are viewing an application or candidate account but that it links back to Applications (the first tab) - though _Candidates_ is the page title. Not a big issue and this is an improvement in any case.

paulrobertlloyd:
@stevehook I agree, it’s A bit unintuitive. The breadcrumbs make more sense if a candidate has multiple applications, in which case we show their account page with a list of those applications. But if they only have one application, we redirect to that instead. I’d be tempted to remove that redirect.

PR #3723, opened 2020-12-23
Add focus styles to links in SVG diagrams

paulrobertlloyd:
## Context

The DAC December 2020 audit spotted an issue with our SVG diagrams in that they a.) attract focus and b.) attract it in the wrong order (which is subjective). I’m not sure this PR really addresses the issue raised, but hopefully improves things a tad.

I wondered if focus should be removed from these diagrams entirely, however `focusable="false"` didn’t work, I suspect as that doesn’t affect any anchors that may be embedded within it.

There’s also a second issue that needs further investigation; links in the diagram that appears within the **closed** `details` element at the top of the page still attract focus. If these links weren’t embedded in the SVG, they wouldn’t while they were hidden, but somehow by being inside an SVG, they do. Browsers, eh! 🤪

* I’ve at least try to make the situation more obvious, by adding focus styles to the links in these diagrams — it wasn’t entirely obvious to me these had clickable elements in the first place!
* Also ensured the font stack in these diagrams matches that of the rest of the service – you can never have too much GDS Transport.

Screenshot of focused link in SVG

## Link to Trello card

## Things to check


paulrobertlloyd:
Come to think of it, a simpler option might be to remove the links from these diagrams… are they used/useful?

davidgisbey (approved):

tijmenb (approved):

PR #3728, opened 2020-12-24
Fix formatting of RecruitmentCycle.cycle_name

steventux:
## Context

Fixes a wrong assumption in https://github.com/DFE-Digital/apply-for-teacher-training/pull/3714

The year we pass to this method should be the end year of the cycle as this is how it's stored on Course records and surfaced via `ApplicationChoice.recruitment_cycle`

Treat the argument we pass to `RecruitmentCycle.cycle_name` as the end year of the cycle eg _2020 to **2021**_

## Guidance to review

View an application from previous cycle, navigate to offer tab and defer. The confirmation screen should show the next cycle from the one on the application (not the next cycle from our current one). This is the only place in the codebase we use this method.

## Link to Trello card

https://trello.com/c/98QAgAQX/2962-defer-offer-confirmation-screen-shows-incorrect-cycle-name

## Things to check

### DIFF
--- a/app/models/recruitment_cycle.rb
+++ b/app/models/recruitment_cycle.rb
@@ -29,6 +29,6 @@ module RecruitmentCycle
   end
 
   def self.cycle_name(year = current_year)
-    "#{year} to #{year + 1}"
+    "#{year - 1} to #{year}"
   end
 end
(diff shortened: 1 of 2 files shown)


despo (approved):
👍🏼

PR #3733, opened 2020-12-29
Add path column to the authentication tokens table

davidgisbey:
## Context

We previously had a mechanism to redirect a user to a specific page endpoint on signing in. We can now do this by adding a path to an authentication token.

This PR adds a `path` column to the authentication_tokens table

- migration to add a `path` column to the authentication_tokens table

## Link to Trello card

https://trello.com/c/dTjHZxp0/2682-use-authenticationtoken-to-redirect-a-user-to-the-correct-endpoint

## Things to check


stevehook (approved):
Migration looks good. Curious about how this will be populated in practice. I guess the initial unauthenticated request will be redirected to a 'request a new magic link' page so we'll have to pass the requested URL in a URL param to that page initially before we can store it in the `authentication_tokens` table (because the record won't exist there yet)?

davidgisbey:
Yea i think it's going to look at the object to see if there's a path present. If not, then default to the sign in page. Then ensuring we pass the path through for expired tokens.

PR #3737, opened 2020-12-29
Add sorting to validation error summary page

TobyRet:
## Context

```
As a... an analyst
I need to... order validation errors
So that... I can see what the most common errors in past month/week are
```

- Add a sorting dropdown to the page
- Update SQL query on back end to change sorting as required by the front end

## Guidance to review

- Pagination will be separate ticket
- The sorting is unit tested. Do we need to feature test this as well or is that overkill?
- Might need some design input on this - I [borrowed from Find](https://www.find-postgraduate-teacher-training.service.gov.uk/results?l=2&subjects%5B%5D=31&sortby=0) as inspiration.

sorting

## Link to Trello card

https://trello.com/c/0U8yaXyp/2725-add-ordering-pagination-to-validation-error-summary

## Things to check

### DIFF
--- a/app/helpers/select_options_helper.rb
+++ b/app/helpers/select_options_helper.rb
@@ -23,6 +23,15 @@ module SelectOptionsHelper
     ] + providers.map { |provider| OpenStruct.new(id: provider.id, name: "#{provider.name} (#{provider.code})") }
   end
 
+  def select_sort_options
+    sort_options = [
+      [ValidationErrorSummaryQuery::ALL_TIME, 'All time'],
+      [ValidationErrorSummaryQuery::LAST_WEEK, 'Last week'],
+      [ValidationErrorSummaryQuery::LAST_MONTH, 'Last month'],
+    ]
+    sort_options.map { |sort_option| OpenStruct.new(value: sort_option.first, text: sort_option.last) }
+  end
+
 private
 
   def nationality_options(include_british_and_irish:)
--- a/app/queries/validation_error_summary_query.rb
+++ b/app/queries/validation_error_summary_query.rb
@@ -1,5 +1,23 @@
 class ValidationErrorSummaryQuery
-  COUNT_SQL =
+  ALL_TIME = 'all_time'.freeze
+  LAST_WEEK = 'last_week'.freeze
+  LAST_MONTH = 'last_month'.freeze
+
+  def initialize(sort_param = ALL_TIME)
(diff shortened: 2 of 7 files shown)


davidgisbey (approved):
Tested it locally and it's working beautifully 👌

app/queries/validation_error_summary_query.rb
```diff
@@ -43,13 +61,17 @@ class ValidationErrorSummaryQuery
       COUNT(DISTINCT user_id) AS unique_users_all_time
     FROM validation_error_counts
     GROUP BY form_object, attribute
-    ORDER BY incidents_all_time DESC".freeze
+    #{order}".freeze
```
stevehook, line 64: Does `.freeze` make sense as this is dynamic now? (I'm not sure it made sense before).

app/helpers/select_options_helper.rb
```diff
@@ -23,6 +23,16 @@ def select_provider_options(providers)
     ] + providers.map { |provider| OpenStruct.new(id: provider.id, name: "#{provider.name} (#{provider.code})") }
   end
 
+  def select_sort_options
+    [
+      [0, 'All time'],
```
stevehook, line 28: Could the values here be strings, e.g. `all_time` to make the URL params more transparent?

app/helpers/select_options_helper.rb
```diff
@@ -23,6 +23,16 @@ def select_provider_options(providers)
...
+      [1, 'Last week'],
+      [2, 'Last month'],
+    ].map do |sort_option|
```
stevehook, line 31: Make the (frozen) array a constant?

stevehook (approved):
Looks good and works well locally. Some minor queries inline.

app/queries/validation_error_summary_query.rb
TobyRet: Yeah that can go - good spot

app/helpers/select_options_helper.rb
TobyRet: I've extracted to a var to make it clearer. Not so sure about extracting to a constant and putting into an initialiser right now? Maybe if it's used in another place perhaps?

app/helpers/select_options_helper.rb
TobyRet: Yep good idea

PR #3739, opened 2020-12-29
Allow support users to sign in as provider users

mnacos:
### Context

We already support signing in as the candidate from the support console, which enables support agents to investigate issues from the perspective of a candidate. We would like to do the same for provider users, as part of general support and troubleshooting.

For security & auditability, when this feature is used, the support user must be logged as extra variable in Sentry & Logit.

Add 'Sign in as this user' button to `SupportInterface::ProviderUsersController#show` and `#audits`. 

Add impersonation methods to `DfeSignInUser`, adding the relevant `provider_user_id` to the session object for the current support user.

Change support_interface header to highlight an active provider user impersonation and provide a link to stop it.

Make `ProviderUser#load_from_session` and `SupportUser#load_from_session` impersonation-aware. Add impersonation information to logs and Sentry errors for all `ProviderInterface` actions.

Ensure any actions taken by support users impersonating provider users are attributed to the support users in the audit log. This means the timeline and the activity log will also show these as performed by 'System' (timeline) or the relevant support user (activity log).

Preserve top-level `Audited.audited_class.as_user` value if set in new `AuditHelper.audit`.

Fix `VendorApiUser` details (e.g. email_address) updating bug in `VendorAPI::DecisionsController`.

New variables logged (logstash + sentry):

```
Real provider user with id 6:

"provider_user_admin_url":"http://localhost:3000/support/users/provider/6"
"dfe_sign_in_uid":"dev-1JB-admin-2"

Impersonated provider user with id 6:

"provider_user_admin_url":"http://localhost:3000/support/users/provider/6"
"dfe_sign_in_uid":"dev-support"
"support_user_email":"support@example.com"
"support_user_admin_url":"http://localhost:3000/support/users/support/1
```

Screenshots:

image

---

image

---

image

### Guidance to review

Impersonating a provider user is different to impersonating a candidate, because candidate interface and support interface use different authentication techniques, which allows a single user to maintain two different authenticated sessions at once. Provider interface uses the same authentication method as  support interface, namely DfE Sign-In, so it's not possible to just create two sessions in parallel. Instead, this PR requires a current support user and annotates their session with impersonation details. `ProviderUser` has been changed to respect these impersonation instructions, if present, returning the desired `current_provider_user` with extra `.impersonator` information.

Because this is all based on the `SupportUser` session, the same session timeouts apply and impersonating a provider user does not modify their `last_signed_in_at`. `current_provider_user` returns the desired provider user, so the provider interface behaves in exactly the same way it would for the real user, which is a good thing for end-to-end testing and troubleshooting. This includes authorisation calls.

However, I've modified all services that correspond to provider user actions with an `audit` block to ensure any changes made to applications in production are attributed to the support user, not the impersonated provider user, for audit trail purposes (both in support and provider interface, e.g. timeline/activity log). Have I missed any services?

The code preserving the original audited `.as_user` setting for these services is required for two reasons: a. test data generation (which passes a support user to bypass authorisation but sets the audited user to a provider user) and b. rails console use for support reasons sets its own `.as_user` information.

Finally, I've restricted data sharing agreements on production to disallow signing by impersonated users.

### Link to Trello card

[2902 - Implement "Sign in as provider user" in prod support](https://trello.com/c/pHP2dGVp)

### Things to check

### DIFF
--- a/app/controllers/provider_interface/provider_interface_controller.rb
+++ b/app/controllers/provider_interface/provider_interface_controller.rb
@@ -60,6 +60,12 @@ module ProviderInterface
         provider_user_admin_url: support_interface_provider_user_url(current_provider_user),
       }
 
+      if (impersonator = current_provider_user.impersonator)
+        useful_debugging_info[:dfe_sign_in_uid] = impersonator.dfe_sign_in_uid
+        useful_debugging_info[:support_user_email] = impersonator.email_address
+        useful_debugging_info[:support_user_admin_url] = support_interface_support_user_url(impersonator)
+      end
+
       RequestLocals.store[:identity] = useful_debugging_info
       Raven.user_context(useful_debugging_info)
     end
--- a/app/controllers/support_interface/provider_users_controller.rb
+++ b/app/controllers/support_interface/provider_users_controller.rb
@@ -81,13 +81,18 @@ module SupportInterface
     def impersonate
       @provider_user = ProviderUser.find(params[:provider_user_id])
       dfe_sign_in_user.begin_impersonate! session, @provider_user
-      flash[:success] = 'Provider user impersonated'
-      redirect_to support_interface_provider_user_path(@provider_user)
+      flash[:success] = "Impersonating #{@provider_user.email_address}"
+      redirect_to provider_interface_applications_path
     end
 
     def end_impersonate
-      dfe_sign_in_user.end_impersonate! session
-      redirect_to support_interface_provider_users_path
+      if (impersonated_user = current_support_user.impersonated_provider_user)
+        dfe_sign_in_user.end_impersonate! session
+        redirect_to support_interface_provider_user_path(impersonated_user)
+      else
+        flash[:success] = "No active provider user impersonation to stop"
+        redirect_to support_interface_provider_users_path
+      end
     end
 
   private
--- a/app/controllers/support_interface/provider_users_controller.rb
+++ b/app/controllers/support_interface/provider_users_controller.rb
@@ -80,8 +80,14 @@ module SupportInterface
 
     def impersonate
       @provider_user = ProviderUser.find(params[:provider_user_id])
+      dfe_sign_in_user.begin_impersonate! session, @provider_user
       flash[:success] = 'Provider user impersonated'
       redirect_to support_interface_provider_user_path(@provider_user)
+    end
+
+    def end_impersonate
+      dfe_sign_in_user.end_impersonate! session
+      redirect_to support_interface_provider_users_path
     end
 
   private
--- a/app/controllers/support_interface/provider_users_controller.rb
+++ b/app/controllers/support_interface/provider_users_controller.rb
@@ -78,6 +78,12 @@ module SupportInterface
       redirect_to support_interface_provider_user_path(provider_user)
     end
 
+    def impersonate
+      @provider_user = ProviderUser.find(params[:provider_user_id])
+      flash[:success] = 'Provider user impersonated'
+      redirect_to support_interface_provider_user_path(@provider_user)
+    end
+
   private
 
     def scope_by_use_of_service
--- a/app/controllers/support_interface/provider_users_controller.rb
+++ b/app/controllers/support_interface/provider_users_controller.rb
@@ -80,14 +80,13 @@ module SupportInterface
 
     def impersonate
       @provider_user = ProviderUser.find(params[:provider_user_id])
-      dfe_sign_in_user.begin_impersonate! session, @provider_user
-      flash[:success] = "Impersonating #{@provider_user.email_address}"
-      redirect_to provider_interface_applications_path
+      dfe_sign_in_user.begin_impersonation! session, @provider_user
+      redirect_to support_interface_provider_user_path(@provider_user)
     end
 
-    def end_impersonate
+    def end_impersonation
       if (impersonated_user = current_support_user.impersonated_provider_user)
-        dfe_sign_in_user.end_impersonate! session
+        dfe_sign_in_user.end_impersonation! session
         redirect_to support_interface_provider_user_path(impersonated_user)
       else
         flash[:success] = 'No active provider user impersonation to stop'
--- a/app/controllers/support_interface/provider_users_controller.rb
+++ b/app/controllers/support_interface/provider_users_controller.rb
@@ -90,7 +90,7 @@ module SupportInterface
         dfe_sign_in_user.end_impersonate! session
         redirect_to support_interface_provider_user_path(impersonated_user)
       else
-        flash[:success] = "No active provider user impersonation to stop"
+        flash[:success] = 'No active provider user impersonation to stop'
         redirect_to support_interface_provider_users_path
       end
     end
--- a/app/controllers/vendor_api/decisions_controller.rb
+++ b/app/controllers/vendor_api/decisions_controller.rb
@@ -1,6 +1,6 @@
 module VendorAPI
   class DecisionsController < VendorAPIController
-    before_action :validate_metadata!, :find_api_user
+    before_action :validate_metadata!
 
     def make_offer
       course_data = params.dig(:data, :course)
@@ -19,7 +19,7 @@ module VendorAPI
 
       if application_choice.offer?
         change_offer = ChangeOffer.new(
-          actor: @api_user,
+          actor: audit_user,
           application_choice: application_choice,
           course_option: course_option,
           offer_conditions: params.dig(:data, :conditions),
@@ -34,7 +34,7 @@ module VendorAPI
         end
       else
         make_offer = MakeAnOffer.new(
-          actor: @api_user,
+          actor: audit_user,
           application_choice: application_choice,
           course_option: course_option,
           offer_conditions: params.dig(:data, :conditions),
@@ -46,7 +46,7 @@ module VendorAPI
 
     def confirm_conditions_met
       decision = ConfirmOfferConditions.new(
-        actor: @api_user,
+        actor: audit_user,
         application_choice: application_choice,
       )
 
@@ -55,7 +55,7 @@ module VendorAPI
 
     def conditions_not_met
       decision = ConditionsNotMet.new(
-        actor: @api_user,
+        actor: audit_user,
         application_choice: application_choice,
       )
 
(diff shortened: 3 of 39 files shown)


app/controllers/support_interface/provider_users_controller.rb
```diff
@@ -78,6 +78,23 @@ def toggle_notifications
       redirect_to support_interface_provider_user_path(provider_user)
     end
 
+    def impersonate
+      @provider_user = ProviderUser.find(params[:provider_user_id])
+      dfe_sign_in_user.begin_impersonate! session, @provider_user
+      flash[:success] = "Impersonating #{@provider_user.email_address}"
```
paulrobertlloyd, line 84: I think this message should match what we say when signing in as a candidate.

Screenshot 2021-01-04 at 12 07 24

paulrobertlloyd:
Have longed for this feature, so useful. Nice one @mnacos!

app/lib/audit_helper.rb
```diff
@@ -0,0 +1,9 @@
+module AuditHelper
+  def audit(actor)
+    impersonator = actor.try(:impersonator)
+
+    Audited.audit_class.as_user(impersonator || actor) do
```
duncanjbrown: If there is an impersonator, do we want to audit this as something like `staff@digital.education.gov.uk on behalf of user@scitt.co.uk`?

app/lib/audit_helper.rb
```diff
@@ -1,8 +1,10 @@
 module AuditHelper
   def audit(actor)
-    impersonator = actor.try(:impersonator)
-
-    Audited.audit_class.as_user(impersonator || actor) do
+    if ::Audited.store[:audited_user].blank?
```
duncanjbrown, line 3: It's a bit tricky to understand what's going on here without a spec, imo — please could you add one?

app/controllers/vendor_api/decisions_controller.rb
```diff
@@ -95,12 +95,6 @@ def application_choice
       @application_choice ||= GetApplicationChoicesForProviders.call(providers: current_provider, vendor_api: true).find(params[:application_id])
     end
 
-    def find_api_user
-      @api_user = VendorApiUser.find_or_initialize_by(
-        vendor_api_token_id: @current_vendor_api_token.id,
```
duncanjbrown, line 100: 😬

app/lib/test_applications.rb
```diff
@@ -340,10 +340,10 @@ def confirm_offer_conditions(choice)
       fast_forward
       ConfirmOfferConditions.new(actor: actor, application_choice: choice).save
       choice.update_columns(recruited_at: time, updated_at: time)
+      # service generates two audit writes, one for status, one for timestamp
+      choice.audits.second_to_last&.update_columns(created_at: time)
+      choice.audits.last&.update_columns(created_at: time)
```
duncanjbrown, line 345: Does it matter that we're `as_provider_user` when calling `update_columns`?

duncanjbrown (commented):
Great stuff. `impersonated_provider_user` in the session seems like a nice straightforward way to do things. Few comments inline — I wonder if `ImpersonatedProviderUser` is a goer? — and I think it needs an end-to-end system spec?

duncanjbrown:
From discussion:

- change UI so that the impersonation is visible from the Provider, not the Support side
- a follow up PR will make sure that impersonated actions appear correctly in the Activity Log and Timeline
- it seems to be possible to sign the DSA when impersonating — eg as support@example.com I impersonated provider@example.com and was able to sign the DSA for Gorse SCITT
- we'll attempt `ImpersonatedProviderUser`
- remove `impersonated_at` but keep the hash structure for future proofing

app/models/dfe_sign_in_user.rb
```diff
@@ -54,12 +55,31 @@ def self.load_from_session(session)
       first_name: dfe_sign_in_session['first_name'],
       last_name: dfe_sign_in_session['last_name'],
       id_token: dfe_sign_in_session['id_token'],
+      impersonated_provider_user: impersonated_provider_user_from(session),
```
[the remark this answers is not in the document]
mnacos: Dropped `impersonated_at` but kept the hash structure.

app/helpers/navigation_items.rb
```diff
@@ -32,9 +32,9 @@ def for_support_primary_nav(current_support_user, current_controller)
 
     def for_support_account_nav(current_support_user)
       if current_support_user && current_support_user.impersonated_provider_user
-        impersonated_email = current_support_user.impersonated_provider_user.email_address
+        impersonated_user = current_support_user.impersonated_provider_user
         [
-          NavigationItem.new("🤿 #{impersonated_email} 🤿", provider_interface_applications_path, false),
+          NavigationItem.new("#{impersonated_user.email_address} 🤿", support_interface_provider_user_path(impersonated_user), false),
```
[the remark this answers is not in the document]
mnacos: Support navigation now remains unchanged.

app/controllers/provider_interface/provider_interface_controller.rb
```diff
@@ -60,6 +60,12 @@ def add_identity_to_log
         provider_user_admin_url: support_interface_provider_user_url(current_provider_user),
       }
 
+      if (impersonator = current_provider_user.impersonator)
+        useful_debugging_info[:dfe_sign_in_uid] = impersonator.dfe_sign_in_uid
```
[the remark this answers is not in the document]
mnacos: The are useful for people less familiar with our codebase/app inspecting these errors.

app/lib/audit_helper.rb
mnacos: The audit itself can only be associated with one user and I think it's safer if this user is the support one.

app/lib/audit_helper.rb
mnacos: There was one, in `spec/lib`. I've renamed it accordingly (`ImpersonationAuditHelper`).

app/controllers/vendor_api/decisions_controller.rb
mnacos: I know!

app/lib/test_applications.rb
mnacos: No, don't think so, update_columns bypasses the audit log.

app/controllers/support_interface/provider_users_controller.rb
mnacos: Thanks for reviewing! I made this change but have since removed the flash message entirely because of tweaks to this flow for support users. Which now is:

image
image
image

mnacos:
> From discussion:
> 
> * change UI so that the impersonation is visible from the Provider, not the Support side
> * a follow up PR will make sure that impersonated actions appear correctly in the Activity Log and Timeline
> * it seems to be possible to sign the DSA when impersonating — eg as [support@example.com](mailto:support@example.com) I impersonated [provider@example.com](mailto:provider@example.com) and was able to sign the DSA for Gorse SCITT
> * we'll attempt `ImpersonatedProviderUser`
> * remove `impersonated_at` but keep the hash structure for future proofing

I've addressed all these. The reason you were able to sign the DSA is I had restricted this check for `HostingEnvironment.production?`. This DSA restriction should now hold for all environments, which makes it more testable. If not being able to sign the DSA proves a big hindrance for testing we can re-introduce environment scoping.

app/controllers/support_interface/provider_users_controller.rb
mnacos: I've added a masks emoji in support, it will be useful as many of us have also created real provider users with the same DfE sign in uid as our support user on various environments.

mnacos: image

app/helpers/navigation_items.rb
```diff
@@ -31,7 +31,12 @@ def for_support_primary_nav(current_support_user, current_controller)
     end
 
     def for_support_account_nav(current_support_user)
-      if current_support_user
+      if (impersonated_user = current_support_user&.impersonated_provider_user)
+        [
+          NavigationItem.new("🎭 #{current_support_user.email_address}", support_interface_provider_user_path(impersonated_user), false),
```
paulrobertlloyd, line 36: Emoji can cause accessibility issues, as their (sometime non-intuitive) Unicode name is announced. We should output something like this:

`<span aria-hidden="true">🎭</span><span class="govuk-visually-hidden">Impersonating:</span>`

duncanjbrown:
Rebuilding the review app

app/helpers/navigation_items.rb
```diff
@@ -31,7 +31,12 @@ def for_support_primary_nav(current_support_user, current_controller)
     end
 
     def for_support_account_nav(current_support_user)
-      if current_support_user
+      if (impersonated_user = current_support_user&.impersonated_provider_user)
```
duncanjbrown, line 34: This breaks when not logged in because `current_support_user` returns `false`, not `nil`

duncanjbrown: I wonder if it should return `nil` anyway?

app/helpers/navigation_items.rb
mnacos: Noted, thanks. I've added such span elements.

app/helpers/navigation_items.rb
mnacos: Hmm yes. Thought about changing `current_support_user` to return `nil` (feels more natural) but this will break memoization. Fixed `navigation_items.rb` instead.

duncanjbrown:
This is a great feature... and feels long overdue!

The UX is now really nice imo — very easy to get back and forth between Manage and Support. I was able to change my current user's permissions with a couple of clicks, which was 👌 👌 .

One small inconsistency: notes don't audit as the support user (offers do)

Screenshot 2021-01-07 at 21 09 24

& two last qs:

- I found the fact that my support email linked to the provider user in support slightly confusing, though the link is v useful. I wonder if we should have the 🎭 as a distinct header element?
- It was a bit frustrating to impersonate a user and then find they hadn't signed the DSA. Should we show the button at all in that case?

mnacos:
As discussed, we're not going to fix notes for now because notes are directly associated with provider users (not polymorphic or audit-log based). The emoji is now its own element and I've brought back the `HostingEnvironment` check which allows support users to sign DSAs on all environments except production.

duncanjbrown (approved):
🎭

PR #3753, opened 2020-12-31
Redirect to the offer page after deferring an offer

AgaDufrat:
## Context

Currently, confirming the deferral takes the user to the application page, which makes it difficult for the user to see what offer they've actually deferred

Redirect to the offer page after deferring an offer

- <kbd>redirect</kbd>

## Link to Trello card

https://trello.com/c/ckrS3ru8/3175-only-export-the-current-cycles-data-for-the-hesa-export

## Things to check


benswannack (approved):
Just needs the change to be reflected in the specs 👍

PR #3754, opened 2020-12-31
Reduce number of notifications in apply support channel

despo:
## Context
Reduce number of notifications posted on Slack by only posting when the application has ended successfuly (on recruitment) or all applications have ended without success.

## Link to Trello card

https://trello.com/c/OFk25StK/3164-reduce-number-of-notifications-in-apply-support-channel

## Things to check


app/lib/state_change_notifier.rb
```diff
@@ -1,85 +1,77 @@
...
+    @application_choice = application_choice
   end
 
-  def self.submit_application(application_form)
-    message = ":rocket: #{application_form.first_name}’s application has been sent to #{application_form.application_choices.map(&:provider).map(&:name).to_sentence}"
-    url = helpers.support_interface_application_form_url(application_form)
-
-    send(message, url)
-  end
+  def application_outcome
```
steventux, line 11: Perhaps the method name should reflect the fact we're notifying something eg.
`notify_application_outcome`

app/lib/state_change_notifier.rb
```diff
@@ -1,85 +1,77 @@
 class StateChangeNotifier
-  def self.sign_up(candidate)
-    candidate_number = Candidate.where(hide_in_reporting: false).count
-
-    return unless (candidate_number % 25).zero?
-
-    candidate_number_is_significant = (candidate_number % 100).zero?
+  APPLICATION_OUTCOME = %i[rejected declined withdrawn recruited].freeze
```
steventux, line 2: I was confused by this constant name, that might just be me, but signifying these are event name symbols might clarify (it's also easy to confuse with application states) eg. `APPLICATION_OUTCOME_EVENTS`

app/lib/state_change_notifier.rb
```diff
@@ -1,85 +1,77 @@
...
+    candidate_name = application_choice.application_form.first_name
+    other_applications = application_choice.self_and_siblings.where.not(id: application_choice.id)
+    providers = [application_choice.course.provider.name]
+
+    if event == :rejected
+      providers += other_applications.select(&:rejected?).map(&:provider).map(&:name)
+
+      declined_message = declined_text_for(other_applications)
+      withdrawn_message = withdrawn_text_for(other_applications)
+      other_applications_outcome = [declined_message, withdrawn_message].compact.to_sentence
```
steventux, line 21: Would it be DRYer to have 2 methods like:

```ruby
def providers_for(event, other_applications)
  return [] if event == :recruited

  other_applications.select(&:"#{event}?") }.map(&:provider).map(&:name)
end

def messages_for(event, other_applications)
  [
    declined_text_for(other_applications),
    rejected_text_for(other_applications),
    withdrawn_text_for(other_applications),
  ].compact
end
```

and use these without the need for the repetition in the conditional clauses. eg.

```ruby
providers += providers_for(event, other_applications)
other_applications_outcome = messages_for(event, other_applications).to_sentence
```

Just a suggestion, not a blocker.

config/locales/slack_notifications.yml
```diff
@@ -0,0 +1,13 @@
+en:
+  slack_notifications:
```
steventux, line 2: :+1:

steventux (approved):
A couple of very minor suggestions about naming things.

app/lib/state_change_notifier.rb
despo: It's not as straighforward as that as you don't want, for example, to end up with 'the application has been rejected by all providers.'  and then also include the other rejected provider in the additional outcome message, but I'll take another look as I think there are some improvements that can be made.

steventux: Ah OK, I see, I wondered if I was missing some context here, it's not a blocker.

steventux (approved):
:+1: https://github.com/DFE-Digital/apply-for-teacher-training/pull/3754/commits/3da6e97315a9fd813af4156670a86a5ced189e9e is great.

PR #3757, opened 2020-12-31
Changes to support performance dashboard

benswannack:
## Context
The application performance dashboard is unclear and a few issues have cropped up with it

1. Rework the stats and names of things to make it clear that the dashboard is referring to application forms, not candidates
2. Add a separate candidate counter to the dashboard that matches the numbers displayed in the slack messages. Also move the never signed in counter to the candidate section
3. Rework the stats explainers so that they are grouped up with similar statuses

All years view:
Screenshot 2021-01-05 at 13 41 17

Explaining tables:
Screenshot 2020-12-31 at 16 17 48

## Guidance to review
Visit the dashboard in the review app

## Link to Trello card
https://trello.com/c/rzWhfLeT/3162-changes-to-support-performance-dashboard

## Things to check

### DIFF
--- a/app/models/performance_statistics.rb
+++ b/app/models/performance_statistics.rb
@@ -5,7 +5,7 @@ class PerformanceStatistics
     @year = year
   end
 
-  def candidate_query
+  def application_form_query
     year_clause = year ? "AND f.recruitment_cycle_year = #{year}" : ''
 
     <<-SQL
@@ -17,7 +17,6 @@ class PerformanceStatistics
               COUNT(f.id) FILTER (WHERE f.id IS NOT NULL) application_forms,
               COUNT(ch.id) FILTER (WHERE f.id IS NOT NULL) application_choices,
               CASE
-                WHEN f.id IS NULL THEN ARRAY['-1', 'never_signed_in']
                 WHEN ARRAY_AGG(DISTINCT ch.status) IN ('{NULL}', '{unsubmitted}') AND DATE_TRUNC('second', f.updated_at) = DATE_TRUNC('second', f.created_at) THEN ARRAY['0', 'unsubmitted_not_started_form']
                 WHEN ARRAY_AGG(DISTINCT ch.status) IN ('{NULL}', '{unsubmitted}') AND DATE_TRUNC('second', f.updated_at) <> DATE_TRUNC('second', f.created_at) THEN ARRAY['1', 'unsubmitted_in_progress']
                 WHEN 'awaiting_provider_decision' = ANY(ARRAY_AGG(ch.status)) THEN ARRAY['4', 'awaiting_provider_decisions']
@@ -30,7 +29,7 @@ class PerformanceStatistics
               END status
           FROM
               application_forms f
-          FULL OUTER JOIN
+          LEFT JOIN
               candidates c ON f.candidate_id = c.id
           LEFT JOIN
               application_choices ch ON ch.application_form_id = f.id
@@ -64,14 +63,14 @@ class PerformanceStatistics
   end
 
   def [](key)
-    candidate_status_counts
+    application_form_status_counts
       .select { |x| x['status'] == key.to_s }
       .map { |row| row['count'] }
       .sum
   end
 
   def total_form_count(only: nil, except: [], phase: nil)
-    candidate_status_counts
+    application_form_status_counts
       .select { |row| only.nil? || row['status'].to_sym.in?(only) }
       .select { |row| phase.nil? || row['phase']&.to_sym == phase.to_sym }
       .reject { |row| row['status'].to_sym.in?(except) }
@@ -79,15 +78,15 @@ class PerformanceStatistics
       .sum
   end
 
-  def candidate_status_counts
-    @candidate_status_counts ||= ActiveRecord::Base
+  def application_form_status_counts
+    @application_form_status_counts ||= ActiveRecord::Base
       .connection
-      .execute(candidate_query)
+      .execute(application_form_query)
       .to_a
   end
 
-  def candidate_status_total_counts
-    candidate_status_counts.group_by { |row| row['status'] }.map do |status, rows|
+  def application_form_status_total_counts
+    application_form_status_counts.group_by { |row| row['status'] }.map do |status, rows|
       { 'status' => status, 'count' => rows.map { |r| r['count'] }.sum }
     end
   end
--- a/app/models/performance_statistics.rb
+++ b/app/models/performance_statistics.rb
@@ -94,11 +94,15 @@ class PerformanceStatistics
   end
 
   def total_submitted_count
-    total_form_count(except: %i[never_signed_in unsubmitted_not_started_form unsubmitted_in_progress])
+    total_form_count(except: %i[unsubmitted_not_started_form unsubmitted_in_progress])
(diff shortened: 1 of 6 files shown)


app/views/support_interface/performance_dashboard/dashboard.html.erb
```diff
@@ -63,8 +73,13 @@
   </div>
 </div>
 
-<table class="govuk-table govuk-!-margin-top-8">
-  <caption class="govuk-table__caption govuk-heading-m">Candidate breakdown</caption>
+<h3 class="govuk-heading-m govuk-!-font-size-27">Application form breakdown</h3>
+
+<h4 class="govuk-heading-s">Unsubmitted applications</h4>
```
duncanjbrown, line 78: wdyt about the font size here @adamsilver?

app/views/support_interface/performance_dashboard/dashboard.html.erb
```diff
@@ -20,38 +20,48 @@
...
-      </div>
-      <div class="govuk-grid-column-one-third">
-        <%= render SupportInterface::TileComponent.new(count: @statistics[:unsubmitted_not_started_form], label: 'not started the form', size: :reduced) %>
+      <div class="govuk-grid-column-one-half">
+        <%= render SupportInterface::TileComponent.new(count: @statistics[:unsubmitted_not_started_form], label: 'forms not started', size: :reduced) %>
       </div>
-      <div class="govuk-grid-column-one-third">
-        <%= render SupportInterface::TileComponent.new(count: @statistics[:unsubmitted_in_progress], label: 'form in progress', size: :reduced) %>
+      <div class="govuk-grid-column-one-half">
+        <%= render SupportInterface::TileComponent.new(count: @statistics[:unsubmitted_in_progress], label: 'forms in progress', size: :reduced) %>
```
duncanjbrown, line 56: Could the tiles possibly anchor-link to the relevant parts of the glossary?

duncanjbrown (commented):
Were we going to break down forms into Apply 1, Apply 2 and Carried Over?

app/models/performance_statistics.rb
```diff
@@ -5,7 +5,24 @@ def initialize(year)
...
+    candidates = Candidate.all
+
+    if year.present?
+      candidates = candidates.joins(:application_forms).where('application_forms.recruitment_cycle_year': year)
+    end
+
+    candidates.where(hide_in_reporting: false).uniq.count
+  end
+
+  def candidate_never_signed_in_count
```
duncanjbrown, line 18: Were we going to remove this box altogether?

app/models/performance_statistics.rb
```diff
@@ -53,63 +67,81 @@ def candidate_query
...
-  end
-
   def total_submitted_count
-    total_candidate_count(except: %i[never_signed_in unsubmitted_not_started_form unsubmitted_in_progress])
+    total_form_count(except: %i[unsubmitted_not_started_form unsubmitted_in_progress])
   end
 
   def apply_again_submitted_count
-    total_candidate_count(except: %i[never_signed_in unsubmitted_not_started_form unsubmitted_in_progress], phase: :apply_2)
+    total_form_count(except: %i[unsubmitted_not_started_form unsubmitted_in_progress], phase: :apply_2)
```
duncanjbrown, line 112: Might be worth constantizing the `except` bit here?

app/models/performance_statistics.rb
```diff
@@ -135,4 +165,28 @@ def percentage_of_providers_onboarded
...
   end
+
+private
+
+  def date_range_query_for_recruitment_cycle_year(cycle_year)
+    start_date = EndOfCycleTimetable::CYCLE_DATES[cycle_year][:apply_reopens]
+
+    query = "created_at >= '#{start_date}'"
+
+    if EndOfCycleTimetable::CYCLE_DATES.to_a.map(&:first).max != cycle_year
```
duncanjbrown, line 176: Could this be expressed any more clearly? Are we effectively saying "if we're not looking at the current cycle year"?

app/models/performance_statistics.rb
```diff
@@ -30,7 +29,7 @@ def candidate_query
               END status
           FROM
               application_forms f
-          FULL OUTER JOIN
+          LEFT JOIN
```
duncanjbrown, line 45: Right — so this ignores candidates without forms?

app/models/performance_statistics.rb
```diff
@@ -52,14 +50,18 @@ def application_form_query
     SQL
   end
 
-  def ended_without_success_sql
+  def form_is_unsubmitted_sql
+    "('unsubmitted' = ANY(ARRAY_AGG(ch.status)) OR '{NULL}' = ARRAY_AGG(DISTINCT ch.status))"
```
duncanjbrown, line 69: Is this saying "any form with > 0 unsubmitted applications is considered to be unsubmitted", whereas it used to need to be all the applications?

app/models/performance_statistics.rb
```diff
@@ -14,17 +14,15 @@ def application_form_query
...
+                WHEN #{form_is_unsubmitted_sql} AND DATE_TRUNC('second', f.updated_at) = DATE_TRUNC('second', f.created_at) THEN ARRAY['0', 'unsubmitted_not_started_form']
+                WHEN #{form_is_unsubmitted_sql} AND DATE_TRUNC('second', f.updated_at) <> DATE_TRUNC('second', f.created_at) THEN ARRAY['1', 'unsubmitted_in_progress']
                 WHEN 'awaiting_provider_decision' = ANY(ARRAY_AGG(ch.status)) THEN ARRAY['4', 'awaiting_provider_decisions']
                 WHEN 'offer' = ANY(ARRAY_AGG(ch.status)) THEN ARRAY['6', 'awaiting_candidate_response']
                 WHEN 'recruited' = ANY(ARRAY_AGG(ch.status)) THEN ARRAY['8', 'recruited']
                 WHEN 'pending_conditions' = ANY(ARRAY_AGG(ch.status)) THEN ARRAY['7', 'pending_conditions']
                 WHEN 'offer_deferred' = ANY(ARRAY_AGG(ch.status)) THEN ARRAY['10', 'offer_deferred']
-                WHEN #{ended_without_success_sql} = '{}' THEN ARRAY['5', 'ended_without_success']
+                WHEN #{form_ended_without_success_sql} THEN ARRAY['5', 'ended_without_success']
```
duncanjbrown, line 40: This is better

duncanjbrown (approved):
Looking good. This page makes a lot more sense now!

Approving with one q about the date query which I think could be a bit more readable. Non-blocking tho.

One last style thing: the blue boxes aren't the same width. Maybe we can get Adam to come and review the page for beauty after this PR is in.

Screenshot 2021-01-05 at 17 08 48

app/models/performance_statistics.rb
benswannack: Yes, I guess just checking for explicitly what we want is better: `if EndOfCycleTimetable::CYCLE_DATES[cycle_year + 1].present?`

app/models/performance_statistics.rb
benswannack: Yes, it just gives a list of forms joined to their candidate

PR #3766, opened 2021-01-04
Standardise email validation

malcolmbaig:
## Context
We want to ensure we have a consistent way of validating emails.

- Use EmailAddressValidator for remaining email validations that weren't using it
- Make the error message a translation string with a more general-purpose namespace

## Guidance to review
- See diff
- Can be tested via support interface by adding a new provider user

## Link to Trello card
https://trello.com/c/VRJnsI1G

## Things to check

### DIFF
--- a/app/validators/email_address_validator.rb
+++ b/app/validators/email_address_validator.rb
@@ -10,7 +10,7 @@ class EmailAddressValidator < ActiveModel::EachValidator
 
   def validate_each(record, attribute, value)
     if value.blank? || !value.match?(EMAIL_REGEX)
-      record.errors[attribute] << I18n.t('activerecord.errors.models.candidate.attributes.email_address.invalid')
+      record.errors[attribute] << I18n.t('validation_errors.email_address_format')
     end
   end
 end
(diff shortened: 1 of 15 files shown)


app/validators/email_address_validator.rb
```diff
@@ -10,7 +10,7 @@ class EmailAddressValidator < ActiveModel::EachValidator
 
   def validate_each(record, attribute, value)
     if value.blank? || !value.match?(EMAIL_REGEX)
-      record.errors[attribute] << I18n.t('activerecord.errors.models.candidate.attributes.email_address.invalid')
```
tijmenb, line 13: Can we remove this string as well?

tijmenb (commented):
The difference between `validates email: true` and `validates email_address: true` is hard to spot. I wonder if we should rename it, perhaps to reflect the reason we have a custom validator, to [closely match GOV.UK Notify](https://github.com/DFE-Digital/apply-for-teacher-training/commit/3e090a8b564f1989af988101e552462866ae8b9e). Perhaps `ValidForNotifyValidator` would work?

```rb
validates :email_address, valid_for_notify: true
```

app/validators/email_address_validator.rb
malcolmbaig: Yes, definitely. Have removed repeated occurrences of it as well now that we validate all email fields with the custom validator.

tijmenb (approved):

PR #3770, opened 2021-01-04
Remove structured qualifications feature flags

davidgisbey:
## Context

The multiple science and Englsh GCSE's flags have been active for a while now and can be removed.

- Remove the multiple_english_gcses flag
- Remove the multiple_science_gcses flag

## Link to Trello card

https://trello.com/c/OrxCzhO2/2772-remove-the-structured-quals-feature-flags

## Things to check


spec/system/vendor_api/vendor_receives_application_spec.rb
```diff
@@ -71,10 +71,11 @@ def then_it_should_include_the_data_from_the_application_form
         qualifications: {
           gcses: [
             {
+
               id: @application.english_gcse.id,
```
davidgisbey, line 75: This is returning nil in the test which doesn't seem correct. I need to speak to @raamSoftwire and @duncanjbrown to confirm

raamSoftwire: I spoke to @duncanjbrown about this yesterday and I'm currently working on tidying this up.

For the API we want to return structured GCSEs as separate qualifications and so we split them out. But this means that the separated GCSEs don't have IDs (because they aren't stored in the DB). We decided to copy the ID from the original structured GCSE and to add a warning to the API that the IDs may not be unique.

davidgisbey: Great. That makes complete sense 👍

raamSoftwire (approved):
LGTM, thanks David

davidgisbey:
No worries! Thanks so much for picking that up @raamSoftwire!

PR #3772, opened 2021-01-05
Allow users to be pointed to a specific endpoint using the AuthenticationToken

davidgisbey:
## Context

Previously we built some functionality to be able to send users to a specific page on login. This worked by using a path param in the query string when arriving at the authentication actions.

After discussion with Rachael, Dan and Mike it seems likely that we will use this functionality going forward.

This PR changes the implementation to store the endpoint on an attribute on the AuthenticationToken.

- Use the path att on the AuthenticationToken to point candidates the desired endpoint
- Ensures that expired tokens behave correctly
 
## Guidance to review

Does this approach seem reasonable? I know this is a bit of a weird ticket as it's not actually being used for anything right now, but does the approach seem sensible. 

## Link to Trello card

https://trello.com/c/dTjHZxp0/2682-use-authenticationtoken-to-redirect-a-user-to-the-correct-endpoint

## Things to check

### DIFF
--- a/spec/system/candidate_interface/signup_and_signin/candidate_signs_in_with_an_authentication_token_with_a_path_spec.rb
+++ b/spec/system/candidate_interface/signup_and_signin/candidate_signs_in_with_an_authentication_token_with_a_path_spec.rb
@@ -1,0 +1,41 @@
+require 'rails_helper'
+
+RSpec.feature 'Candidates authenitcation token has the path attriute populated' do
+  include SignInHelper
+  include CandidateHelper
+
+  scenario 'Candidate is redirected to the appropriate page' do
+    given_the_pilot_is_open
+    and_i_am_a_candidate_with_an_account
+    and_i_have_received_a_token_associated_with_the_personal_statment_path
+
+    when_i_sign_in_using_the_token
+    and_i_confirm_the_sign_in
+    then_i_receive_am_redirected_to_the_personal_statement_page
+  end
+
+  def given_the_pilot_is_open
+    FeatureFlag.activate('pilot_open')
+  end
+
+  def and_i_am_a_candidate_with_an_account
+    create_and_sign_in_candidate
+  end
+
+  def and_i_have_received_a_token_associated_with_the_personal_statment_path
+    @magic_link_token = MagicLinkToken.new
+    create(:authentication_token, user: current_candidate, hashed_token: @magic_link_token.encrypted, path: 'candidate_interface_edit_becoming_a_teacher_path')
+  end
+
+  def when_i_sign_in_using_the_token
+    visit candidate_interface_authenticate_path(token: @magic_link_token.raw)
+  end
+
+  def and_i_confirm_the_sign_in
+    confirm_sign_in
+  end
+
+  def then_i_receive_am_redirected_to_the_personal_statement_page
+    expect(page).to have_current_path candidate_interface_edit_becoming_a_teacher_path
+  end
+end
--- a/spec/system/candidate_interface/signup_and_signin/candidate_signs_in_with_an_authentication_token_with_a_path_spec.rb
+++ b/spec/system/candidate_interface/signup_and_signin/candidate_signs_in_with_an_authentication_token_with_a_path_spec.rb
@@ -23,6 +23,20 @@ RSpec.feature 'Candidates authenitcation token has the path attriute populated' 
     when_i_click_on_the_link_in_my_email
     and_i_confirm_the_sign_in
     then_i_am_redirected_to_the_personal_statement_page
+
+    given_i_have_a_reference_in_the_not_requested_yet_state
+    and_i_have_received_a_token_associated_with_the_review_unsubmitted_reference_path
+
+    when_i_sign_in_using_the_token
+    and_i_confirm_the_sign_in
+    then_i_am_redirected_to_the_review_unsubmitted_reference_page
+
+    given_i_have_received_a_token_associated_with_the_review_unsubmitted_reference_path
+    and_i_have_subsequently_deleted_the_reference
+
+    when_i_sign_in_using_the_token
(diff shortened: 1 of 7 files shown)


stevehook (commented):
I think this a good feature. Does the value of the `path` param have to be a Rails path helper method name or can we just use the path itself? The method name is a bit less flexible because we can't include URL params.

davidgisbey:
Yea that's a really good point! I'll make the changes today.

stevehook:
> Yea that's a really good point! I'll make the changes today.
> 
> Edit: I was just thinking a bit about this. What if we link to a resource that has been deleted? I.e. a references review page. We could just as a rule of thumb link to the review/ start page for each section if someone tries to hit a resource that isn't associated with their application.

I think if the `path` is invalid for whatever reason then we should return a 404 or whatever the appropriate response would be had they hit that path directly. So I don't think we need to do anything different in this case.

spec/system/candidate_interface/signup_and_signin/candidate_signs_in_with_an_authentication_token_with_a_path_spec.rb
```diff
@@ -0,0 +1,130 @@
+require 'rails_helper'
+
+RSpec.feature 'Candidates authenitcation token has the path attriute populated' do
```
stevehook, line 3:
```suggestion
RSpec.feature 'Candidates authentication token has the path attribute populated' do
```

spec/system/candidate_interface/signup_and_signin/candidate_signs_in_with_an_authentication_token_with_a_path_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+    @reference = create(:reference, :not_requested_yet, application_form: @candidate.current_application)
+  end
+
+  def and_i_have_received_a_token_associated_with_the_review_unsubmitted_reference_path
+    @magic_link_token = MagicLinkToken.new
+    create(
+      :authentication_token,
+      user: @candidate,
+      hashed_token: @magic_link_token.encrypted,
+      path: "http://localhost:3000/candidate/application/references/review-unsubmitted/#{@reference.id}",
```
stevehook, line 111:
```suggestion
      path: "/candidate/application/references/review-unsubmitted/#{@reference.id}",
```

spec/system/candidate_interface/signup_and_signin/candidate_signs_in_with_an_authentication_token_with_a_path_spec.rb
```diff
@@ -0,0 +1,130 @@
+require 'rails_helper'
+
+RSpec.feature 'Candidates authenitcation token has the path attriute populated' do
+  include SignInHelper
+  include CandidateHelper
+
+  scenario 'Candidate is redirected to the appropriate page' do
+    given_the_pilot_is_open
+    and_i_am_a_candidate_with_an_account
+    and_i_have_received_a_token_associated_with_the_personal_statment_path
```
stevehook, line 10:
```suggestion
    and_i_have_received_a_token_associated_with_the_personal_statement_path
```

spec/system/candidate_interface/signup_and_signin/candidate_signs_in_with_an_authentication_token_with_a_path_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+
+    when_i_sign_in_using_the_token
+    and_i_confirm_the_sign_in
+    then_i_am_redirected_to_the_personal_statement_page
+
+    given_i_am_signed_out
+    and_i_have_an_expired_token_associated_with_the_personal_statment_path
```
stevehook, line 17:
```suggestion
    and_i_have_an_expired_token_associated_with_the_personal_statement_path
```

spec/system/candidate_interface/signup_and_signin/candidate_signs_in_with_an_authentication_token_with_a_path_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+  def given_the_pilot_is_open
+    FeatureFlag.activate('pilot_open')
+  end
+
+  def and_i_am_a_candidate_with_an_account
+    create_and_sign_in_candidate
+    @candidate = current_candidate
+  end
+
+  def and_i_have_received_a_token_associated_with_the_personal_statment_path
```
stevehook, line 51:
```suggestion
  def and_i_have_received_a_token_associated_with_the_personal_statement_path
```

spec/system/candidate_interface/signup_and_signin/candidate_signs_in_with_an_authentication_token_with_a_path_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+
+  def then_i_am_redirected_to_the_personal_statement_page
+    expect(page).to have_current_path candidate_interface_edit_becoming_a_teacher_path
+  end
+
+  def given_i_am_signed_out
+    click_link 'Sign out'
+  end
+
+  def and_i_have_an_expired_token_associated_with_the_personal_statment_path
```
stevehook, line 77:
```suggestion
  def and_i_have_an_expired_token_associated_with_the_personal_statement_path
```

stevehook (approved):
Looks good! A couple of inline suggestions.

spec/system/candidate_interface/signup_and_signin/candidate_signs_in_with_an_authentication_token_with_a_path_spec.rb
davidgisbey: Whoops 🤦 thanks Steve

PR #3774, opened 2021-01-05
Reinstate authentication links from emails that rely on the `u` (user id) param

stevehook:
## Context

https://github.com/DFE-Digital/apply-for-teacher-training/pull/3717 removed the uniquely identify users via the `u` (encrypted user id) param in favour of the `token` param that is being used going forward. This broke existing links in emails that have been sent in the past. This PR reinstates the old functionality alongside the new so that links such as `/candidate/sign-in/confirm?token=<token>&u=<encrypted-user-id>` where `token` is a missing token continue to work.

- New system spec to reproduce the issue.
- Reinstate the flow through `CandidateInterface::SignInController` that uses the `u` param to prompt the candidate to resend a new magic link without the candidate having to re-enter their email address.

## Guidance to review

- Does this completely restore the original feature?
- Does the new system spec cover the legacy use case?

## Link to Trello card

https://trello.com/c/sVFcG2EJ/2824-reinstate-authentication-links-from-emails-that-rely-on-the-u-user-id-param

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/sign_in_controller.rb
+++ b/app/controllers/candidate_interface/sign_in_controller.rb
@@ -70,13 +70,13 @@ module CandidateInterface
     end
 
     def create_from_expired_token
-      encrypted_user_id = params.fetch(:u)
+      encrypted_user_id = params[:u]
       authentication_token = AuthenticationToken.find_by_hashed_token(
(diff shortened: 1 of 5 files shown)


davidgisbey (approved):
lgtm steve

PR #3776, opened 2021-01-06
Add request specs for redirects (Candidate interface)

TobyRet:
## Context

We've added some redirects to the `routes.rb` file as part of the wider work to update routes.

As we plan to now keep these new redirects around indefinitely we should add request specs to ensure redirects are working as expected.

- Add request specs for redirects

## Guidance to review

NA

## Link to Trello card

https://trello.com/c/UiFVoJp8/2834-add-request-specs-for-redirects

## Things to check

### DIFF
--- a/spec/requests/candidate_interface/redirects_spec.rb
+++ b/spec/requests/candidate_interface/redirects_spec.rb
@@ -45,7 +45,15 @@ RSpec.describe 'Candidate Interface - Redirects', type: :request do
         get path
 
         expect(response.status).to eq(301)
+        expect(path_exists?(response.location)).to be true
       end
     end
+  end
+
+  def path_exists?(path)
+    controller_params = Rails.application.routes.recognize_path(path)
+    controller_params[:action] != 'not_found'
+  rescue ActionController::RoutingError
+    false
   end
 end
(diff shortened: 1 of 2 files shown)


spec/requests/candidate_interface/redirects_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      '/candidate/application/school-experience/edit/any_id',
+      '/candidate/application/school-experience/review',
+      '/candidate/application/school-experience/delete/any_id',
+      '/candidate/application/degrees/any_id/completion_status',
+      '/candidate/application/degrees/any_id/completion_status/edit',
+    ].each do |path|
+      it "returns status code 301 for #{path}" do
+        get path
+
+        expect(response.status).to eq(301)
```
frankieroberto, line 47: Is it worth also testing where they redirect to?

tijmenb (approved):

spec/requests/candidate_interface/redirects_spec.rb
tijmenb: I've added https://github.com/DFE-Digital/apply-for-teacher-training/pull/3776/commits/6c003e9f75cc129365c65be1696ae27acae97c99 to make sure that the URL we redirect actually exists. That's probably sufficient because explicitly testing all the destinations would duplicate routes.rb, and not guard against typos. @TobyRet WDYT?

TobyRet: @tijmenb I think the additional check makes sense. 
Do we need the `puts` on line 55 though?

PR #3777, opened 2021-01-06
Add validations to provider permissions setting forms

benswannack:
## Context
From the DAC review, it is confusing that when you select "Extra permissions", but don't actually choose any, you do not get a validation error. This change adds those validations

## Guidance to review
Test using the review app

## Link to Trello card
https://trello.com/c/3SsH4Fx0/3171-no-errors-on-setting-permissions-page-dac-review

## Things to check

### DIFF
--- a/app/forms/provider_interface/provider_user_invitation_wizard.rb
+++ b/app/forms/provider_interface/provider_user_invitation_wizard.rb
@@ -16,6 +16,7 @@ module ProviderInterface
 
     validates :providers, presence: true, on: :providers
     validates :view_applications_only, presence: { message: 'Choose whether this user has extra permissions' }, on: :permissions
+    validate :at_least_one_extra_permission_is_set_if_necessary, on: :permissions
 
     class PermissionsForm
       include ActiveModel::Model
@@ -123,6 +124,15 @@ module ProviderInterface
     end
 
   private
+
+    def at_least_one_extra_permission_is_set_if_necessary
+      return if ActiveModel::Type::Boolean.new.cast(view_applications_only)
+
+      selected_permissions = provider_permissions[current_provider_id].fetch('permissions', []).reject(&:blank?)
+      if selected_permissions.count == 0
+        errors[:provider_permissions] << 'Select extra permissions'
+      end
+    end
 
     def delete_permissions_if_view_applications_only(attrs)
       attrs.fetch(:provider_permissions, {}).each_key do |k|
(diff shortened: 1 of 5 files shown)


app/forms/provider_interface/provider_user_invitation_wizard.rb
```diff
@@ -124,6 +125,15 @@ def email_address=(raw_email_address)
 
   private
 
+    def at_least_one_extra_permission_is_set_if_necessary
+      return if ActiveModel::Type::Boolean.new.cast(view_applications_only)
+
+      selected_permissions = permissions_for(current_provider_id).fetch('permissions', []).reject(&:blank?)
+      if selected_permissions.count.zero?
```
despo, line 132:
```suggestion
      if selected_permissions.none?
```

app/forms/provider_interface/provider_user_permissions_form.rb
```diff
@@ -36,6 +37,7 @@ def self.from(permissions_model)
     def update_from_params(hash)
       view_applications_only = ActiveModel::Type::Boolean.new.cast(hash[:view_applications_only])
 
+      self.view_applications_only = view_applications_only
```
despo, line 40: Is there any benefit of doing this here rather than doing the assignment directly on line 38?

app/forms/provider_interface/provider_user_permissions_form.rb
```diff
@@ -51,5 +52,23 @@ def save
...
+
+      extra_permissions = [
+        manage_organisations,
+        manage_users,
+        make_decisions,
+        view_safeguarding_information,
+        view_diversity_information,
+      ]
+
+      if extra_permissions.all? { |value| value == false }
```
despo, line 69:
```suggestion
      if extra_permissions.none?
```

despo (approved):
Left some minor feedback but overall it looks great 💯

duncanjbrown:
I think `if ActiveModel::Type::Boolean.new.cast(var)` can usually be replaced by `var.present?`

duncanjbrown:
Other than that, looks good! :+1:

benswannack:
> I think `if ActiveModel::Type::Boolean.new.cast(var)` can usually be replaced by `var.present?`

I don't think that's the case here. The values this takes are `"true"` and `"false"`, which are both `present?`. I changed it on the other form though 🙂

mnacos (approved):
Looks good

PR #3778, opened 2021-01-06
Show rejected_at and rejected_by_default_at in Support

duncanjbrown:
Now that we link from Slack to applications once all rejections are in, include rejection dates in Support so we know when things happened


davidgisbey (approved):

duncanjbrown:
Thanks @davidgisbey

PR #3779, opened 2021-01-06
Feature metric dashboard part 1 - References

stevehook:
## Context

Product Owners I need to see in real time specific metrics for key features in dashboard so that they can see the impact of changes over time. This PR introduces the feature metrics dashboard and the first set of metrics for references.

- [x] A new link to _Feature metrics_ from the _Performance_ index page

image

- [x] New feature metric dashboard

image

- [ ] Tidy up front end
- [x] Minor correction to timestamps for references in `TestApplications` (mainly so that requested_at < time in audit trail when referee responds.)

## Guidance to review

- Does the implementation of the query in `FeatureMetrics` (that gets the wait times for reference responses in a given timeframe) make sense? Is there a simpler way? Would it be worthwhile adding an `ApplicationForm#references_complete_at` attribute (or similar) so it's easier to work out the wait times?
- Note that the review app is showing old test data for which `ApplicationReference#requested_at` is wrong (it's after the response) hence the negative numbers for average wait times.
- How should we style this dashboard given that it's going to have several more metrics added after these?

## Link to Trello card

https://trello.com/c/CBmwCoda/2770-feature-metric-dashboard

## Things to check

### DIFF
--- a/app/models/feature_metrics.rb
+++ b/app/models/feature_metrics.rb
@@ -8,11 +8,12 @@ class FeatureMetrics
     times_to_get = time_to_get_references(start_time, end_time)
     return 'n/a' if times_to_get.blank?
 
-    number_with_precision(
+    average_days = number_with_precision(
       times_to_get.sum.to_f / times_to_get.size,
       precision: 1,
       strip_insignificant_zeros: true,
     )
+    "#{average_days} #{'day'.pluralize(average_days)}"
   end
 
   def percentage_references_within(
--- a/app/models/feature_metrics.rb
+++ b/app/models/feature_metrics.rb
@@ -1,0 +1,51 @@
+class FeatureMetrics
+  # def time_to_get_references(start_time, end_time = Time.zone.now)
+  #   data ||= ActiveRecord::Base
+  #     .connection
+  #     .select_all(TIME_TO_GET_REFERENCES_SQL % [start_time.iso8601, end_time.iso8601])
+  #     .to_a
+  #   # times = data.map { |audits| time_to_get_reference_for(audits) }.compact
+  #   # times.sum.to_f / time.size
+  # end
+
+  def time_to_get_references(start_time, end_time = Time.zone.now)
+    references = ApplicationReference.where(
+      'id IN (:reference_ids)',
+      reference_ids: Audited::Audit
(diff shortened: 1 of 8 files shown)


paulrobertlloyd:
Only comment from me is not sure about using blue blocks for what are essentially headings, not stats. That said, I do like the presentation… I wonder if it’s something more like this:

Screenshot 2021-01-07 at 10 37 52

Have started a discussion on Slack.

app/views/support_interface/feature_metrics_dashboard/dashboard.html.erb
```diff
@@ -0,0 +1,68 @@
+<% content_for :title, 'Feature metrics' %>
+
+<h3 class="govuk-heading-m govuk-!-font-size-27 govuk-!-margin-top-8">References</h3>
```
paulrobertlloyd, line 3: Illogical heading structure, `h2` should follow `h1`…

```suggestion
<h2 class="govuk-heading-m govuk-!-font-size-27 govuk-!-margin-top-8">References</h2>
```

stevehook:
Thanks @paulrobertlloyd, I've updated the design and metrics as per your suggestion and discussion on Slack.

app/models/feature_metrics.rb
```diff
@@ -0,0 +1,62 @@
+class FeatureMetrics
+  include ActionView::Helpers::NumberHelper
+
+  def average_time_to_get_references(
+    start_time,
+    end_time = Time.zone.now.beginning_of_day
+  )
+    times_to_get = time_to_get_references(start_time, end_time)
```
davidgisbey, line 8: When I read this i wasn't completely sure what times_to_get_references was returning. I'm struggling to come up with a better name though.

`time_to_get_references_for_each_application` seems a bit clunky.

davidgisbey: Actually the above wouldn't really work as it is only returning times where both references are received.

I think it's probs fine as is 👍

davidgisbey (approved):
Looks great Steve. It's working really nicely locally!

paulrobertlloyd (approved):
Looks great. We can tweak the design as we add more metrics.

PR #3783, opened 2021-01-06
Fix two styling issues with the API docs

duncanjbrown:
- make the background colour of syntax-highlighted blocks of code match
the background we use for inline styling on `<code>` elements. This
prevents an ugly effect where the text bg is slightly darker than the
box bg
- remove some `<pre>` tags that smeared `<code>` background onto lines above
and below a code sample

| Before | After |
:-------------------------:|:-------------------------:
| before1 |  after1 |
| before2 | after2 |


benswannack (approved):
👌

PR #3784, opened 2021-01-06
Let providers search by candidate reference number

steventux:
## Context

Providers have requested the ability to search by 'candidate reference' (ie. support reference)

- Updates provider application filter to search by candidate name or support reference depending on input
- Adds ability to configure the filter component so that the search field can be treated as 'primary' and sit above results. (See screenshot)
- Adds support reference to application card caption

image

## Guidance to review

There are effectively 2 filtering forms now which have to maintain filtering params from one another, so combining filtering and searching should give expected results.
The support UI uses the same filtering mechanism, worth checking no support UI was harmed in the making of this PR.

## Link to Trello card

https://trello.com/c/XSZRMDg2/3138-implement-letting-providers-search-by-candidate-reference-number

## Things to check

### DIFF
--- a/app/services/filter_application_choices_for_providers.rb
+++ b/app/services/filter_application_choices_for_providers.rb
@@ -1,4 +1,6 @@
 class FilterApplicationChoicesForProviders
+  CANDIDATE_REFERENCE_REGEX = /^[a-zA-Z]{2}\d{4}$/.freeze
+
   def self.call(application_choices:, filters:)
     return application_choices if filters.empty?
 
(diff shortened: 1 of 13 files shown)


app/services/filter_application_choices_for_providers.rb
```diff
@@ -1,4 +1,6 @@
 class FilterApplicationChoicesForProviders
+  CANDIDATE_REFERENCE_REGEX = /^[aA-zZ]{2}\d{4}$/.freeze
```
benswannack, line 2: Really minor and pedantic, but`[aA-zZ]` should be `[A-Za-z]`

spec/services/filter_application_choices_for_providers_spec.rb
```diff
@@ -0,0 +1,68 @@
+require 'rails_helper'
+
+RSpec.describe FilterApplicationChoicesForProviders do
+  describe '.call' do
+    let(:application_choices) do
+      create_list(:application_choice, 2)
+      ApplicationChoice.all
```
benswannack, line 7: Do we need this line? By the looks of it, this could be a one line `let`

benswannack (approved):
Looks good to me!

spec/services/filter_application_choices_for_providers_spec.rb
steventux: We do need it yes, it's a side effect of the described class expecting all the joins or includes to be made before filtering.
`create_list` returns an array which won't respond to the `.joins` statements necessary in the tests.

benswannack: Ah yes, probably not one for this PR then. Maybe worth a tech debt card?

steventux: :+1: Will add a tech debt story

simonwhatley:
@steventux there’s a small change with the empty states.

If no applications are found, we show a message depending on the search and filter terms:

- Just the search was used: `There are no results for ‘search term’.`
- The search and a single filter was used: `There are no results for ‘search term’ and the selected filter.`
- The search and multiple filters/filter options were used: `There are no results for ‘search term’ and the selected filters.`

steventux:
@simonwhatley I've added contextual 'no results' messages here https://github.com/DFE-Digital/apply-for-teacher-training/pull/3784/commits/53a0c37492088ee9247c8c92f7b0d2dd31f5a783

simonwhatley (approved):

PR #3788, opened 2021-01-07
Send a slack notification when a feature flag is changed in production

AgaDufrat:
## Context

The feature flags can be changed by clicking a button. We have a daily notification
if they are not consistent between environments but we should be notified earlier
to prevent accidental changes in production.

Send a slack notification when a feature flag is changed in production

## Guidance to review

Should we be sending it out in all environments?
Other suggestions for the emoji welcome.

## Things to check

### DIFF
--- a/app/controllers/support_interface/settings_controller.rb
+++ b/app/controllers/support_interface/settings_controller.rb
@@ -2,13 +2,25 @@ module SupportInterface
   class SettingsController < SupportInterfaceController
     def activate_feature_flag
       FeatureFlag.activate(params[:feature_name])
-      flash[:success] = "Feature ‘#{params[:feature_name].humanize}’ activated"
+
+      SlackNotificationWorker.perform_async(
+        ":flags: Feature ‘#{feature_name}‘ was activated",
+        support_interface_feature_flags_url,
+      )
+
+      flash[:success] = "Feature ‘#{feature_name}’ activated"
       redirect_to support_interface_feature_flags_path
     end
 
     def deactivate_feature_flag
       FeatureFlag.deactivate(params[:feature_name])
-      flash[:success] = "Feature ‘#{params[:feature_name].humanize}’ deactivated"
+
+      SlackNotificationWorker.perform_async(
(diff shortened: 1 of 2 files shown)


app/controllers/support_interface/settings_controller.rb
```diff
@@ -2,13 +2,29 @@ module SupportInterface
   class SettingsController < SupportInterfaceController
     def activate_feature_flag
       FeatureFlag.activate(params[:feature_name])
-      flash[:success] = "Feature ‘#{params[:feature_name].humanize}’ activated"
+
+      if HostingEnvironment.production?
```
tijmenb, line 6: Yes, it's fine to send this in all envs because [the slack worker will send it to a test channel in other envs](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/workers/slack_notification_worker.rb#L26-L28).

tijmenb (commented):
👍 Looking good, but we should send it in all envs indeed.

app/controllers/support_interface/settings_controller.rb
```diff
@@ -2,13 +2,25 @@ module SupportInterface
   class SettingsController < SupportInterfaceController
     def activate_feature_flag
       FeatureFlag.activate(params[:feature_name])
-      flash[:success] = "Feature ‘#{params[:feature_name].humanize}’ activated"
+
+      SlackNotificationWorker.perform_async(
+        ":flags: Feature ‘#{feature_name}‘ was activated in production",
```
duncanjbrown, line 7: Do we need `in production` if this can be sent from any env?

duncanjbrown: If we want to mention production, maybe we should include the `HostingEnvironment.environment_name` instead?

AgaDufrat: Meant to remove that! The worked already includes that for env other than production eg `[QA] Message`

duncanjbrown (approved):

PR #3790, opened 2021-01-07
Improve Reasons for Rejection validation messaging

benswannack:
## Context
When rejecting an application either via the API or via the UI, if the rejection reason is left blank then two error messages show.

Only display one validation error if both structured and freetext reasons for rejection are left blank

Over the API:
Screenshot 2021-01-07 at 14 56 36

In the UI:
Screenshot 2021-01-07 at 14 54 41

## Guidance to review
Is there a recommended way to get a nice error message in the UI as well as over the API?
I like the look of it on manage, but the API message isn't perfect. It'd be nice to keep the code as common as possible though

## Link to Trello card
https://trello.com/c/26TUwK3C/3177-improve-reasons-for-rejection-error-handling

## Things to check

### DIFF
--- a/app/services/reject_application.rb
+++ b/app/services/reject_application.rb
@@ -3,8 +3,7 @@ class RejectApplication
 
   attr_accessor :rejection_reason, :structured_rejection_reasons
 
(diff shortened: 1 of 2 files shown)


AgaDufrat (approved):
👍

PR #3793, opened 2021-01-07
Remove checkboxes from feedback form, and make feedback mandatory

frankieroberto:
## Context

The accessibility audit highlighted that the checkboxes and the textarea on the feedback form did not indicate that they were both optional.

Whilst we could fix the narrow accessibility issue by adding the word "(optional)" to the legend/label, this would be a bit odd, and really it doesn't make sense that that the form can be submitted with no feedback or checkboxes ticked at all.

We did consider adding a forth catch-all checkbox, and making ticking at least one of them mandatory, but on further thought it seemed like having just a checkbox ticked with no other feedback given is unlikely to be that helpful (especially at low volumes), so a simpler approach would be to remove the checkboxes altogether, and just have a single mandatory feedback textarea.

In addition, the error message for the "Can we contact you about your feedback?" has been improved.

### Before

localhost_3000_candidate_application_application-feedback_original_controller=candidate_interface%2Freferences%2Freview page_title=References path=%2Fcandidate%2Fapplication%2Freferences%2Freview

localhost_3000_candidate_application_application-feedback

### After

localhost_3000_candidate_application_application-feedback_original_controller=candidate_interface%2Freferences%2Freview page_title=References path=%2Fcandidate%2Fapplication%2Freferences%2Freview (1)

localhost_3000_candidate_application_application-feedback (1)

## Guidance to review

Whilst this changes the form, the export is currently unaffected, and will still export the 3 boolean fields, even though for future feedback these will all be false. A follow-up Pull Request should remove these fields from the database and remove them from the export.

The feedback field is currently still named `other_feedback` in the database and the code. We should probably rename this to just `feedback` within a separate database migration.

## Link to Trello card

https://trello.com/c/eIt8mxNH/2805-dac-feedback-form-whats-optional

## Things to check

### DIFF
--- a/app/views/candidate_interface/application_feedback/new.html.erb
+++ b/app/views/candidate_interface/application_feedback/new.html.erb
@@ -5,22 +5,12 @@
     <%= form_with model: @application_feedback_form, url: candidate_interface_application_feedback_path, method: :post do |f| %>
       <%= f.govuk_error_summary %>
 
-      <h1 class="govuk-heading-xl">
-        <span class="govuk-caption-xl"><%= t('application_feedback.caption') %></span>
-        <%= t('page_titles.application_feedback', section: @application_feedback_form.section_name) %>
-      </h1>
-
       <%= f.hidden_field :path %>
       <%= f.hidden_field :page_title %>
       <%= f.hidden_field :original_controller %>
 
-      <%= f.govuk_check_boxes_fieldset :issues, multiple: false, legend: { text: t('application_feedback.issues.label'), size: 'm' } do %>
-        <%= f.govuk_check_box :does_not_understand_section, true, multiple: false, label: { text: t('application_feedback.issues.does_not_understand_section') }, link_errors: true %>
-        <%= f.govuk_check_box :need_more_information, true, multiple: false, label: { text: t('application_feedback.issues.need_more_information') } %>
-        <%= f.govuk_check_box :answer_does_not_fit_format, true, multiple: false, label: { text: t('application_feedback.issues.answer_does_not_fit_format') } %>
-      <% end %>
-
-      <%= f.govuk_text_area :other_feedback, label: { text: t('application_feedback.other_feedback.label'), size: 'm' }, rows: 5, max_words: 300 %>
+      <span class="govuk-caption-xl"><%= t('application_feedback.caption') %></span>
+      <%= f.govuk_text_area :other_feedback, label: { text: t('page_titles.application_feedback', section: @application_feedback_form.section_name), size: 'xl', tag: 'h1' }, rows: 5, max_words: 300, threshold: 90 %>
 
       <%= f.govuk_radio_buttons_fieldset :consent_to_be_contacted, inline: true, legend: { text: t('application_feedback.consent_to_be_contacted.label'), size: 'm' } do %>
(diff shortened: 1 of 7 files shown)


app/views/candidate_interface/application_feedback/new.html.erb
```diff
@@ -5,22 +5,12 @@
...
 
-      <%= f.govuk_check_boxes_fieldset :issues, multiple: false, legend: { text: t('application_feedback.issues.label'), size: 'm' } do %>
-        <%= f.govuk_check_box :does_not_understand_section, true, multiple: false, label: { text: t('application_feedback.issues.does_not_understand_section') }, link_errors: true %>
-        <%= f.govuk_check_box :need_more_information, true, multiple: false, label: { text: t('application_feedback.issues.need_more_information') } %>
-        <%= f.govuk_check_box :answer_does_not_fit_format, true, multiple: false, label: { text: t('application_feedback.issues.answer_does_not_fit_format') } %>
-      <% end %>
-
-      <%= f.govuk_text_area :other_feedback, label: { text: t('application_feedback.other_feedback.label'), size: 'm' }, rows: 5, max_words: 300 %>
+      <span class="govuk-caption-xl"><%= t('application_feedback.caption') %></span>
+      <%= f.govuk_text_area :other_feedback, label: { text: t('page_titles.application_feedback', section: @application_feedback_form.section_name), size: 'xl', tag: 'h1' }, rows: 5, max_words: 300, threshold: 90 %>
```
paulrobertlloyd, line 13: This can be simplified by using the form builder `caption` param:

```suggestion
      <%= f.govuk_text_area :other_feedback, caption: { text: t('application_feedback.caption'), size: 'xl' }, label: { text: t('page_titles.application_feedback', section: @application_feedback_form.section_name), size: 'xl', tag: 'h1' }, rows: 5, max_words: 300, threshold: 90 %>
```

frankieroberto: @paulrobertlloyd nice, although that puts the caption inside the `<label>`, which I'm not sure makes sense here (as the full label would then be read out as "Help us improve this service How can we improve the qualifications section?").

paulrobertlloyd: Ah yes, good point.

davidgisbey (approved):
Looks good. I think the next steps are:

1. Clean up the data from previously submitted feedback with something like this 
image

2. Remove the fields from the export
image

3. Migration to remove the unused columns

4. Rename other_feedback to feedback

frankieroberto:
@david-mears-dfe yes please! Are you happy for that to be a follow-up PR?

With the migration to rename `other_feedback` to `feedback`, how fussed are we about avoiding errors during migration? ie do we need to first add the column, then update the code to read/write to the new column, then copy across values from the old column, and then drop the old column? Or can we just rename the column in the same deploy as the code expecting the new column name, and accept the [likely small] risk of errors mid-deploy when the code and database don't align?

davidgisbey:
As a rule of thumb, we use the add, update, copy across and drop approach. We've been stung a few times recently with just renaming the column.

I'll card it up and pick the work up this afternoon. I also just noticed that we don't have an export for feedback from Find so I'll add that too.

PR #3795, opened 2021-01-08
Add affirmative flash messages for a candidate accepting and declining an offer

davidgisbey:
## Context

Currently, when a candidate accepts and declines (need to confirm with Paul) an offer. They should receive confirmation that their action has been successful.

- Accepting an offer 

image

- Declining an offer

image

## Guidance to review

The card doesn't actually mention anything about having a flash for a declined offer, but i think it probably should right?. If it shouldn't be there @paulrobertlloyd let me know and I'll take it out.

The card also mentions that it needs a content review @gregknight1 & @Tomcowan1. Thanks guys!

## Link to Trello card

https://trello.com/c/5cDegw3k/874-when-accepting-an-offer-theres-no-affirmative-state-eg-green-success-message-at-the-top-of-the-dashboard-or-a-panel

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/decisions_controller.rb
+++ b/app/controllers/candidate_interface/decisions_controller.rb
@@ -29,6 +29,7 @@ module CandidateInterface
     def confirm_decline
       decline = DeclineOffer.new(application_choice: @application_choice.reload)
       decline.save!
+      flash[:success] = "You have declined your offer for #{@application_choice.course.name_and_code} at #{@application_choice.provider.name}"
       redirect_to candidate_interface_application_complete_path
     end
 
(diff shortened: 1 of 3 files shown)


Tomcowan1:
Hi @davidgisbey sorry for the delay, just wanted to check "declined" was the wording we used in emails etc (it is) so happy with this.

Tomcowan1 (approved):

paulrobertlloyd (commented):
All looks good to me; the multiple banners are problematic, but that’s something we can look at when we come to review the design of the dashboard. Thanks for picking this one up @davidgisbey!

PR #3799, opened 2021-01-08
Experiment: use a whitelist of attributes to mark application choices as updated

duncanjbrown:
## Context

It would be nice to have correct updated_at dates in the API and on application choices generally.

We tried to do this in #3697 but ran into a problem with that approach, which was that if the structure of the JSON changed then the cache would be invalidated.

This PR pilots a simpler approach: list the fields that can affect an API response and only touch the application choices when those fields change. This would need to happen for other models (work experience etc) which affected the choice.

**Pros**

1. explicit list of significant attributes
2. no complicated cache

**Cons**

1. if we forget to mark a field as significant the choice's updated_at date will not update
2. if we accidentally leave a field as significant when it's no longer significant the choice's updated_at date _will_ update

**Mitigations for cons**

re: point 2, it would be nice to somehow be able to assert in a test that all the fields we care about are actually looked at when we build the choice. this addresses point 2 above. or we could use the caching strategy _in tests_ to assert that eg a choice renders in a certain way, then when any whitelisted property on the form changes, then that actually changes the JSON.

Potentially both these things are tricky because depending on the state of the form, we look at different fields. But in practice I think it's just international addresses which are handled differently, and those are going away soon anyway.

Not sure how we could address point 1 — any ideas?

### DIFF
--- a/adr/0017-application-choice-updated-at.md
+++ b/adr/0017-application-choice-updated-at.md
@@ -44,9 +44,9 @@ See [single_application_presenter_spec.rb](../spec/presenters/vendor_api/single_
 
 This solution will require that spec to be kept up to date with a variety of applications in different states so as to cover all the code paths the application presenter has.
 
-So far, this solution only addresses changes from the ApplicationForm model that affect API responses for ApplicationChoices.
+As described so far, this solution only addresses changes from the ApplicationForm model that affect API responses for ApplicationChoices.
 There are other models (e.g. ApplicationQualification or ApplicationWorkExperience) for which changes to their attributes affect their associated applications.
-For these smaller models, we will use a simple touch relationship between them and their application form, with touches on the application form also touching the application choices.
+For these smaller models, we have added a concern PublishedInAPI which touches the application_choices whenever the model is created, updated or deleted.
 
 ## Consequences
 
--- a/adr/0017-application-choice-updated-at.md
+++ b/adr/0017-application-choice-updated-at.md
@@ -1,0 +1,54 @@
+# 17. Application Choice Updated At
+
+Date: 2021-01-14
+
+## Status
+
+Accepted
+
+## Context
+
+Consumers of the Vendor API can filter applications by updated_at value, to get a list of applications that have changed since a certain time.
+They typically use this to avoid processing applications that have already been ingested into their services.
+If the updated_at values are inaccurate, there is a potential for vendors to process applications twice or miss application changes entirely.
+For that reason, we set up touch relationships from the application form to the application choice to ensure that changes to the application form touch the application choices.
+
+This approach resulted in false positives (where the updated_at value had changed but the data in the application API response had not). This is because there are fields on the application form which do not appear in the application API response.
+
+For example, when geocoding data was added to applications. Every application was touched, but the data was not included in the response so API consumers could not see any changes.
+
+## First approach
+
+The first idea we had to address this issue was to render the API response for an application whenever we were changing a model related to it and see if the change had affected the response.
+If a change was detected, we would change the update_at on the application choice.
+This intended to only change the updated_at value when the response changed.
+
+However, this approach is sensitive to changes to the shape of the API response, as well as changes to the data. For example, adding a new field to the response would indicate an update where there had not been one.
+
+To fix this drawback, we would have to add logic to ignore certain changes to API responses, at which point this simple idea becomes complex.
+
+## Decision
+
+For changes to the application form we will specify a whitelist of attributes that, when changed, affect the application API response. There are pros and cons to this method:
+
+Pros:
+ - We keep an explicit list of significant attributes
+ - This is a simple approach using standard Rails techniques, so is easy to follow
+
+Cons:
+ - Forgetting to mark a field as significant will result in the application updated_at not being changed (false negative)
+ - Forgetting to remove a field from the list of significant ones when it no longer affects an application response will result in the application updated_at being changed (false positive)
+
+In order to mitigate the cons, we have added specs to check that the list of significant attributes contains all attributes which affect an API response, and that there are no extra attributes included.
+See [single_application_presenter_spec.rb](../spec/presenters/vendor_api/single_application_presenter_spec.rb)
+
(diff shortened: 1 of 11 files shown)


tijmenb (approved):

app/models/application_form.rb
```diff
@@ -72,7 +72,25 @@ def equality_and_diversity_answers_provided?
...
+    disability_disclosure further_information safeguarding_issues_status
+    address_line1 address_line2 address_line3 address_line4
+    international_address country postcode equality_and_diversity
+    work_history_breaks first_nationality second_nationality third_nationality
+    fourth_nationality fifth_nationality phone_number
+  ].freeze
+
+  before_save do |form|
+    if (form.changed & PUBLISHED_FIELDS).any?
+      application_choices.update_all(updated_at: Time.zone.now)
```
benswannack, line 90: We can use touch_all for this :) 
```suggestion
      application_choices.touch_all
```

app/models/application_form.rb
```diff
@@ -72,7 +72,25 @@ def equality_and_diversity_answers_provided?
   attribute :recruitment_cycle_year, :integer, default: -> { RecruitmentCycle.current_year }
 
   before_create -> { self.support_reference ||= GenerateSupportRef.call }
-  after_save -> { application_choices.update_all(updated_at: Time.zone.now) }
```
benswannack, line 75: If we want touches on application form to also touch the application choices as we wanted before (e.g. change WorkHistory > Touch ApplicationForm > Touch ApplicationChoice), we need to put in an `after_touch` since `after_save` is not triggered on touch

spec/presenters/vendor_api/single_application_presenter_spec.rb
```diff
@@ -585,4 +585,30 @@
       expect(presenter.as_json[:attributes][:recruited_at]).to be_present
     end
   end
+
+  describe 'compliance with models that change updated_at' do
+    it 'looks at all fields which cause a touch' do
+      choice = create(:application_choice, :awaiting_provider_decision)
```
benswannack, line 591: We still need to add enough application choice types here to cover different cases in the application presenter

spec/presenters/vendor_api/single_application_presenter_spec.rb
```diff
@@ -585,4 +585,30 @@
...
+
+      # if there is a field on the form that causes a touch but isn't
+      # queried via the presenter on the API, fail this test
+      (ApplicationForm::PUBLISHED_FIELDS - %w[international_address right_to_work_or_study_details]).each do |field|
+        expect_any_instance_of(ApplicationForm).to receive(field).at_least(:once).and_call_original
```
benswannack, line 596: RuboCop doesn't like us doing this. I'm not sure what the cleanest way to get around this would be @duncanjbrown ?

spec/models/application_form_spec.rb
```diff
@@ -6,6 +6,22 @@
...
+    end
+
+    it 'does not touch the application choice when a field not affecting the application choice is changed' do
+      application_form = create(:completed_application_form, application_choices_count: 1)
+
+      expect { application_form.update(latitude: '0.12343') }
+        .not_to(change { application_form.application_choices.first.updated_at })
+    end
+  end
+
```
benswannack, line 24: Still need to add a spec that checks changes to e.g. ApplicationQualification touches the form and the choices.

spec/presenters/vendor_api/single_application_presenter_spec.rb
duncanjbrown: It is dirty! We can disable RuboCop for just these lines: https://docs.rubocop.org/rubocop/configuration.html#disabling-cops-within-source-code

app/models/application_form.rb
```diff
@@ -72,7 +72,32 @@ def equality_and_diversity_answers_provided?
...
+  before_save do |form|
+    if (form.changed & PUBLISHED_FIELDS).any?
+      application_choices.touch_all
+    end
+  end
+
+  # We want to touch the application_choices when models that touch application_form are changed
+  # e.g. ApplicationReference, ApplicationQualification
+  # This ensures that the updated_at for the application changes when the API response for the application changes
+  after_touch do
```
benswannack, line 97: Justo to challenge my own assumption here: Do we always want to touch choices when forms are touched? This covers the mentioned references and qualifications. Do we maybe want a more precise way of doing this so that we can touch application forms without touching the choices? What do you think @duncanjbrown @steventux ?

duncanjbrown: doesn't this line undo the work we do above? or is after_touch only called on explicit touches?

benswannack: This is only triggered on touch, either explicitly (`form.touch`) or via a belongs_to with a touch.

benswannack: I think the answer is that we do want to be more granular here. ApplicationFeedback has a `belongs_to :application form, touch: true`, but is not used in the API response for a choice. It will be clearer to add a concern that explicitly says "this model is published in the API"

adr/0017-application-choice-updated-at.md
```diff
@@ -0,0 +1,54 @@
...
+Cons:
+ - Forgetting to mark a field as significant will result in the application updated_at not being changed (false negative)
+ - Forgetting to remove a field from the list of significant ones when it no longer affects an application response will result in the application updated_at being changed (false positive)
+
+In order to mitigate the cons, we have added specs to check that the list of significant attributes contains all attributes which affect an API response, and that there are no extra attributes included.
+See [single_application_presenter_spec.rb](../spec/presenters/vendor_api/single_application_presenter_spec.rb)
+
+This solution will require that spec to be kept up to date with a variety of applications in different states so as to cover all the code paths the application presenter has.
+
+So far, this solution only addresses changes from the ApplicationForm model that affect API responses for ApplicationChoices.
```
duncanjbrown, line 47: We can update this now

duncanjbrown:
I can't approve my own PR, but this is good to go ⭐ @benswannack

PR #3803, opened 2021-01-08
Qualifications data extract

raamSoftwire:
## Context

As a... Data Analyst
I need to... understand what qualifications our user base has
So that... I can assess any issues/gaps etc.

Add a new qualifications export

## Guidance to review

We think we've tested the different cases for structured grades, but there could be something we've missed.

## Link to Trello card

https://trello.com/c/lWDdxqqr/2771-qualifications-data-extract

## Things to check

### DIFF
--- a/app/models/data_export.rb
+++ b/app/models/data_export.rb
@@ -86,7 +86,7 @@ class DataExport < ApplicationRecord
       class: SupportInterface::ProviderAccessControlsExport,
     },
     locations_export: {
-      name: 'Locations export',
+      name: 'Locations',
       description: 'A list of application choices with the associated postcodes for the candidate, provider and site.',
       class: SupportInterface::LocationsExport,
     },
@@ -94,6 +94,11 @@ class DataExport < ApplicationRecord
       name: 'Sites',
       description: 'A list of sites that are being synced from Find, along with distances to their respective providers',
       class: SupportInterface::SitesExport,
+    },
+    qualifications: {
+      name: 'Qualifications',
+      description: 'A list of qualifications for each application choice',
+      class: SupportInterface::QualificationsExport,
     },
   }.freeze
 
--- a/app/services/support_interface/qualifications_export.rb
+++ b/app/services/support_interface/qualifications_export.rb
@@ -1,0 +1,144 @@
+module SupportInterface
+  class QualificationsExport
+    def data_for_export
+      application_choices = ApplicationChoice
+                                .select(:id, :application_form_id, :rejection_reason, :status, :course_option_id)
+                                .includes(:course_option, :course, :provider)
+
+      application_choices.map do |application_choice|
+        application_form = application_choice.application_form
+        course = application_choice.course_option.course
+        qualifications = application_form.application_qualifications
+        a_levels = a_levels(qualifications)
+        degrees = degrees(qualifications)
+
+        output = {
+          'Candidate id' => application_form.candidate_id,
+          'Support ref' => application_form.support_reference,
+          'Phase' => application_form.phase,
+          'Cycle' => application_form.recruitment_cycle_year,
+
+          'Outcome (offer, rejected etc.)' => application_choice.status,
+          'Reason for Rejection' => application_choice.rejection_reason,
+          'Course Code' => course.code,
+          'Provider Code' => course.provider.code,
+
+          'GCSE maths grade' => maths_gcse_grade(qualifications),
+
+          'GCSE science single grade' => science_single_gcse_grade(qualifications),
+          'GCSE science double grade' => science_double_gcse_grade(qualifications),
+          'GCSE science triple grade' => science_triple_gcse_grade(qualifications),
+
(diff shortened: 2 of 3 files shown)


app/models/data_export.rb
```diff
@@ -86,7 +86,7 @@ class DataExport < ApplicationRecord
       class: SupportInterface::ProviderAccessControlsExport,
     },
     locations_export: {
-      name: 'Locations export',
+      name: 'Locations',
```
raamSoftwire, line 89: The word export is added later, so this was rendering as "Locations export export".

george-bit (approved):
Looking good Raam just spec just failing on that first candidate :)

app/services/support_interface/qualifications_export.rb
```diff
@@ -0,0 +1,144 @@
...
+        degrees = degrees(qualifications).sort_by(&:subject)
+
+        output = {
+          'Candidate id' => application_form.candidate_id,
+          'Support ref' => application_form.support_reference,
+          'Phase' => application_form.phase,
+          'Cycle' => application_form.recruitment_cycle_year,
+
+          'Outcome (offer, rejected etc.)' => application_choice.status,
+          'Reason for Rejection' => application_choice.rejection_reason,
```
davidgisbey, line 22: We need to think about adding structured rejection reasons in here too.

Maybe something like `application_choice.structured_rejection_reasons || application_choice.rejection_reason`

app/services/support_interface/qualifications_export.rb
```diff
@@ -0,0 +1,144 @@
...
+
+          'Number of other qualifications provided' => other_qualification_count(qualifications),
+        }
+        output
+      end
+    end
+
+  private
+
+    def maths_gcse_grade(qualifications)
```
davidgisbey, line 65: For English/science/maths GCSEs you can just pass the application form in and call

application_form.maths_gcse 
application_form.science_gcse
application_form.english_gcse

app/services/support_interface/qualifications_export.rb
```diff
@@ -0,0 +1,144 @@
...
+      maths_gcse = qualifications.where(level: :gcse, subject: :maths).first
+      maths_gcse.try(:grade)
+    end
+
+    def science_single_gcse_grade(qualifications)
```
davidgisbey, line 70: Do you think we need separate columns for each of these? 

Could we just have:

def science_single_gcse_grade(application_form)
  science_gcse = application_form.science_gcse
  science_gcse.structured_grade || science_gcse.unstructured_grade
end

Depending on the returned grades it should be fairly explicit whether it was a single, double or triple award. Might be worth asking Mike.

davidgisbey: Although your approach is far more filterable. So it may well be more appropriate.

app/services/support_interface/qualifications_export.rb
```diff
@@ -0,0 +1,144 @@
...
+      unstructured_grade = qualifications.where(level: :gcse, subject: :science).first.try(:grade)
+      return unstructured_grade if unstructured_grade.present? && not_single_or_double_award?(unstructured_grade)
+
+      science_triple_gcse = qualifications.where(level: :gcse, subject: ApplicationQualification::SCIENCE_TRIPLE_AWARD).first
+      return if science_triple_gcse.try(:structured_grades).blank?
+
+      science_triple_gcse[:structured_grades].values.join
+    end
+
+    def english_unstructured_gcse_grade(qualifications)
```
davidgisbey, line 96: Same with this in terms of granularity. 

I'd lean towards keeping it as is as you've already done a great job with all the legwork, but if Mike won't use it we can just do 

`def english_gcse_grade(application_form)
  english_gcse = application_form.english_gcse
  JSON.parse(english_gcse.structured_grades)&.values&.join || english_gcse.grade
end`

app/services/support_interface/qualifications_export.rb
raamSoftwire: Nice idea, I've added that in now

app/services/support_interface/qualifications_export.rb
raamSoftwire: We spoke to Mike quite a bit when we built this and he said that having the sciences in separate columns is really helpful.

app/services/support_interface/qualifications_export.rb
raamSoftwire: Similarly, this was a request from Mike

app/services/support_interface/qualifications_export.rb
raamSoftwire: I tried that but I'm seeing `undefined method `maths_gcse' for #<ActiveRecord::Associations::CollectionProxy []>`. 

The reason we wanted to pass in `qualifications` in the first place was to limit the times we call the DB, by getting a list of qualifications upfront.

davidgisbey (approved):
Thanks for the answers. All good on my end then 👍

PR #3805, opened 2021-01-08
Fix bug related to rejection emails

despo:
Referenced field should be `reject_by_default_at` rather than `decline_by_default_at`, as `decline_by_default_at` is the last day that the candidate, rather than the provider, can make a decision.


mnacos (approved):
🚀

PR #3807, opened 2021-01-10
ApplicationFeedback cleanup (part2)

davidgisbey:
## Context
**Do not merge until #3806 has been deployed**

PR that removed the need for several bools on the ApplicationFeedback model is #3793

First PR that removed references to the bools is #3806

This PR just drops the bools from the ApplicationFeedback table and adds a feedback column

## Changes proposed in this pull request

Drop the following columns  to the application_feedback table:

- does_not_understand_section
- need_more_information
- answer_does_not_fit_format

Adds the following columns to the application_feedback table:

- feedback 

## Guidance to review

Frankie has gone through the feedback and determined that we don't need to worry about backfilling the records with the dropped bools as candidates have left explanations in the other_feedback column.

## Link to Trello card

https://trello.com/c/UBt82tOr/2843-remove-unneeded-booleans-from-the-applicationfeedback-class

## Things to check

- [x] This code does not rely on migrations in the same Pull Request
- [x] If this code includes a migration adding or changing columns, it also backfills existing records for consistency
- [x] API release notes have been updated if necessary
- [x] New environment variables have been [added to the Azure config](https://github.com/DFE-Digital/apply-for-teacher-training#azure-hosting-devops-pipeline)

PR #3806, opened 2021-01-10
ApplicationFeedback cleanup (part1)

davidgisbey:
## Context

Changes were made to the feedback page in this PR #3793 

We no longer use the following bools:
- `does_not_understand_section`
- `need_more_information`
- `answer_does_not_fit_format`

We should also change the column name from `other_feedback` to `feedback`

There are going to be a few PRs that will follow this pattern:

1. Remove any code related to the above bools
2. Drop the above bool columns from the db and add a `feedback` column
3. Update the code to use the new `feedback` column
4. Backfill the `feedback` column with the data in the `other_feedback` column
5. drop the `other_feedback` column 

## Changes proposed in this pull request

Remove code related to the following booleans:

- `does_not_understand_section`
- `need_more_information`
- `answer_does_not_fit_format`

## Guidance to review

Does all the above sound sensible?

## Link to Trello card

https://trello.com/c/UBt82tOr/2843-remove-unneeded-booleans-from-the-applicationfeedback-class

## Things to check

- [x] This code does not rely on migrations in the same Pull Request
- [x] If this code includes a migration adding or changing columns, it also backfills existing records for consistency
- [x] API release notes have been updated if necessary
- [x] New environment variables have been [added to the Azure config](https://github.com/DFE-Digital/apply-for-teacher-training#azure-hosting-devops-pipeline)

PR #3808, opened 2021-01-10
ApplicationFeedback cleanup (part3)

davidgisbey:
## Context

**Do not merge until #3807 has been deployed**

PR that removed the need for several bools on the ApplicationFeedback model is #3793

 #3806 - First PR that removed references to the bools
 #3807 -Second PR that drops the unused bool columns and adds a feedback column

This PR backfills the new `feedback` column and updates the code to use it
 
- Update the code to use the new `feedback` column (previously used the `other_feedback` column).

## Link to Trello card

https://trello.com/c/UBt82tOr/2843-applicationfeedback-cleanup

## Things to check

### DIFF
--- a/db/migrate/20210110112042_backfill_feedback_column.rb
+++ b/db/migrate/20210110112042_backfill_feedback_column.rb
@@ -1,0 +1,5 @@
+class BackfillFeedbackColumn < ActiveRecord::Migration[6.0]
+  def change
+    ApplicationFeedback.all.each { |feedback| feedback.update!(feedback: feedback.other_feedback) }
+  end
+end
--- a/db/migrate/20210110112042_backfill_feedback_column.rb
+++ b/db/migrate/20210110112042_backfill_feedback_column.rb
@@ -1,5 +1,5 @@
 class BackfillFeedbackColumn < ActiveRecord::Migration[6.0]
   def change
-    ApplicationFeedback.all.each { |feedback| feedback.update!(feedback: feedback.other_feedback) }
+    ApplicationFeedback.all.each { |feedback| feedback.update_column(:feedback, feedback.other_feedback) }
   end
 end
--- a/app/controllers/candidate_interface/application_feedback_controller.rb
+++ b/app/controllers/candidate_interface/application_feedback_controller.rb
@@ -27,7 +27,7 @@ module CandidateInterface
     def feedback_params
       strip_whitespace params.require(:candidate_interface_application_feedback_form).permit(
         :path, :page_title,
-        :other_feedback, :consent_to_be_contacted,
+        :feedback, :consent_to_be_contacted,
         :original_controller
       )
     end
--- a/app/forms/candidate_interface/application_feedback_form.rb
+++ b/app/forms/candidate_interface/application_feedback_form.rb
@@ -2,10 +2,10 @@ module CandidateInterface
   class ApplicationFeedbackForm
     include ActiveModel::Model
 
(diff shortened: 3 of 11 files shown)


db/migrate/20210110100235_remove_unused_application_feedback_booleans_and_add_feedback_column.rb
```diff
@@ -0,0 +1,8 @@
+class RemoveUnusedApplicationFeedbackBooleansAndAddFeedbackColumn < ActiveRecord::Migration[6.0]
+  def change
+    remove_column :application_feedback, :does_not_understand_section, :boolean
+    remove_column :application_feedback, :need_more_information, :boolean
+    remove_column :application_feedback, :answer_does_not_fit_format, :boolean
+    add_column :application_feedback, :feedback, :string
```
raamSoftwire, line 6: I think this is the missing line from the previous PR and needs to be moved back, so that it's deployed before you make reference to the new `feedback` column

db/migrate/20210110100235_remove_unused_application_feedback_booleans_and_add_feedback_column.rb
```diff
@@ -0,0 +1,8 @@
+class RemoveUnusedApplicationFeedbackBooleansAndAddFeedbackColumn < ActiveRecord::Migration[6.0]
+  def change
+    remove_column :application_feedback, :does_not_understand_section, :boolean
+    remove_column :application_feedback, :need_more_information, :boolean
+    remove_column :application_feedback, :answer_does_not_fit_format, :boolean
```
raamSoftwire, line 5: I'm also not sure these lines should be here either.

db/migrate/20210110112042_backfill_feedback_column.rb
```diff
@@ -0,0 +1,5 @@
+class BackfillFeedbackColumn < ActiveRecord::Migration[6.0]
+  def change
+    ApplicationFeedback.all.each { |feedback| feedback.update!(feedback: feedback.other_feedback) }
```
raamSoftwire, line 3: This is really neat 👍🏽

raamSoftwire (commented):
Other than tidying up the migrations, this all looks good too.

db/migrate/20210110112042_backfill_feedback_column.rb
duncanjbrown: Could we use `update_column` so we don't touch `updated_at`? https://davidverhasselt.com/set-attributes-in-activerecord/

db/migrate/20210110112042_backfill_feedback_column.rb
```diff
@@ -0,0 +1,5 @@
+class BackfillFeedbackColumn < ActiveRecord::Migration[6.0]
+  def change
+    ApplicationFeedback.all.each { |feedback| feedback.update_column(feedback: feedback.other_feedback) }
```
duncanjbrown, line 3: `update_column` takes `(name, value)`, not `(name: value)`

davidgisbey: So it does. My bad

db/migrate/20210110100235_remove_unused_application_feedback_booleans_and_add_feedback_column.rb
davidgisbey: have sorted

raamSoftwire (approved):
Nice one, cheers David

PR #3815, opened 2021-01-11
Introduce "Interviewing" state

despo:
## Context

Introduces new interview state and transitions application_choice to that when an interview is created

- Introduce new interview state
- Add audit entry 
- Amend filtered list sorting to include application choices in interview state in deadline approaching section
- Include interviewing in filtetered list
- Rename interviewing state via locale yaml so it appears the same as 'awaiting_provider_decision' to the candidate
- Display `interviewing` state as `awaiting_provider_decision` at Vendor API

## Link to Trello card
https://trello.com/c/uLndN1CF/3225-introduce-interviewing-state

## Things to check


app/services/create_interview.rb
```diff
@@ -0,0 +1,37 @@
...
+  )
+    @auth = ProviderAuthorisation.new(actor: actor)
+    @application_choice = application_choice
+    @provider = provider
+    @date_and_time = date_and_time
+    @location = location
+    @additional_details = additional_details
+  end
+
+  def call
```
AgaDufrat, line 20:
```suggestion
  def save
```

We tend to used `save` or `save!` for the services that change the state e.g. AcceptOffer WithdrawApplication DeclineOffer MakeAnOffer etc

spec/presenters/vendor_api/single_application_presenter_spec.rb
```diff
@@ -585,4 +585,12 @@
       expect(presenter.as_json[:attributes][:recruited_at]).to be_present
     end
   end
+
+  describe 'attributes.status' do
+    it 'returns awaiting_provider_decision when status is interviewing' do
+      application_choice = build_stubbed(:application_choice, status: :interviewing)
```
benswannack, line 591: It's probably worth adding a trait that sets up a realistic interviewing application with interviews etc.

config/locales/application_states.yml
```diff
@@ -181,6 +191,12 @@ en:
       by: system
       description: Application choices that are unsubmitted are automatically rejected at the end of the cycle but can be carried over in the next cycle.
 
+    awaiting_provider_decision-interview:
+      name: Provider setups up an interview
```
benswannack, line 195:
```suggestion
      name: Provider sets up an interview
```

config/locales/application_states.yml
```diff
@@ -181,6 +191,12 @@ en:
...
+      by: provider
+      description: |
+        Providers sets up an interview with candidates.
```
benswannack, line 198: We could put a bit more detail in this one, maybe something like this?
```suggestion
        Providers can set up an interview with candidates before making a decision
```

spec/presenters/vendor_api/single_application_presenter_spec.rb
benswannack: Oh – I see that's in the next PR 👍

benswannack (approved):
LGTM

PR #3818, opened 2021-01-11
Change how we configure Redis

malcolmbaig:
## Context

For faster test runs, we want to be able to run our test suite in parallel using multiple cores. This is some of the configuration change required to support that.

The commits are a bit wordy but to summarize:

- We plan to use parallel_tests to split up our test suite and run each set of tests in a different process.
- Each process needs a unique Redis database to ensure test runs remain free of side effects.
- To support this, update Redis config so that it follows these conventions:
  - Use REDIS_URL wherever it's already set, meaning these changes should have no effect on any existing environment that sets this variable.
  - Use Redis database 0 in development (was previously 1, as defined by .env)
  - Use Redis database 9 when running tests normally (no change here)
  - Use Redis databases 1-n when running tests in multiple processes, where n is determined by how many processes are created.
- To ensure these conventions hold, access Redis via the `ApplyRedisConnection` class.

## Guidance to review
- Per commit recommended, commit messages give a bit more detail explaining the changes.
- Do the new conventions make sense?

## Link to Trello card
https://trello.com/c/ySrLogAF

## Things to check

### DIFF
--- a/app/lib/apply_redis_connection.rb
+++ b/app/lib/apply_redis_connection.rb
@@ -1,0 +1,19 @@
+class ApplyRedisConnection
+  def self.url
+    if ENV['REDIS_URL'].present?
+      # Always use the Redis database defined by the environment, if available.
+      ENV['REDIS_URL']
+    elsif ENV['TEST_ENV_NUMBER']
+      # If we're in the test environment and tests are being run in parallel,
+      # use a different database for each process. Add 1 to the environment
+      # number so that we don't clobber the development database at 0.
+      "redis://localhost:6379/#{ENV['TEST_ENV_NUMBER'].to_i + 1}"
+    elsif Rails.env.test?
+      # If we're in the test environment and tests are being run in a single
+      # process, default to database 9 (this could be any number except 0).
+      'redis://localhost:6379/9'
+    else
+      'redis://localhost:6379/0'
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


app/lib/apply_redis_connection.rb
```diff
@@ -0,0 +1,23 @@
+class ApplyRedisConnection
```
tijmenb, line 1: Have you considered setting `Redis.current` in an initializer? If that works it would save some extra config, and new Redis-dependent dependencies (like Sidekiq) would automatically be able to fetch the connection.

tijmenb (commented):
Looking good! Question about `Redis.current`

app/lib/apply_redis_connection.rb
malcolmbaig: Good idea! Something like this? https://github.com/DFE-Digital/apply-for-teacher-training/pull/3818/commits/b1458eb28fddd28167bac822a2d538118aa9e293

duncanjbrown:
This looks good to me and I like the `Redis.current` approach. Does it work with `parallel_tests`?

malcolmbaig:
> Does it work with `parallel_tests`?

Yes, just about to raise the PR adding that in.

duncanjbrown (approved):

tijmenb (approved):
LGTM!

davidgisbey (approved):

PR #3819, opened 2021-01-11
Generate applications in Interviewing state

AgaDufrat:
## Context

We are adding a new state.

Create an Interview for an application choice in an `interviewing` state when generating test data.

## Guidance to review

Dependent on #3815 being merged first.

## Link to Trello card

https://trello.com/c/GapoDi9e/3214-implement-set-up-an-interview-flow

## Things to check

### DIFF
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -543,6 +543,15 @@ FactoryBot.define do
       reject_by_default_at { 40.business_days.from_now }
     end
 
+    trait :with_scheduled_interview do
+      awaiting_provider_decision
+
+      after(:build) do |choice, _evaluator|
+        choice.status = :interviewing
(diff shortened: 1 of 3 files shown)


spec/factories.rb
```diff
@@ -536,6 +536,14 @@
       reject_by_default_days { 40 }
     end
 
+    trait :with_scheduled_interview do
```
benswannack, line 539: We should make sure this looks like it's been through the awaiting_provider decision state first: Can we just add
```
awaiting_provider_decision
```
at the top to handle that for us?

benswannack (approved):
Looks good! One last thought: can we create an interview in the `:with_scheduled_interview` trait?

PR #3820, opened 2021-01-11
Add another button on review pages

paulrobertlloyd:
## Context

* @adamsilver suggested moving buttons to add another item to the top of the review page, below the heading, making this function easier to find and more clearly differentiated from the ‘mark as complete’ checkbox.

* The recently redesigned reference flow moved the add another functionality to the top of its review page:
  
  references-add-another

* We have observed an issue with candidates not adding all their relevant qualifications. **Might moving the add another button up and alongside guidance, help address this?**

### Course choices

The button is moved to the top of the page, but remains grey as you may only want to add one course. Additional guidance shows how many more choices you have left to add. Uses intended text style like that used for very similar messaging for references.

add-another-course

### References

Remains largely the same as before but:

* uses grey button once 2 references have been given
* removes headings which duplicated that in button text

add-second-referee

add-another-reference

### Degrees

The button is moved to the top of the page, but remains grey as most candidates will likely only have 1 degree.

add-another-degree

### Work history

The button is moved to the top of the page. When there are no jobs it says ‘Add a job’. When there are already jobs it says ‘Add another job’.

add-another-job

### Volunteering

The button is moved to the top of the page. When there are no roles it says ‘Add a role’. When there are already roles it says ‘Add another role’.

add-another-role

### Other qualifications

The button is moved to the top of the page.

add-another-qualification

## Guidance to review

* The design has been quite possibly over-reviewed 😅 We ended at this point because:
  * moving the button to the top makes it more visible, and separates it from the other controls on this page.
  * we considered adding additional guidance on more of these pages, but realised it repeated that which had already been shown or will be shown again.
  * tried alternative designs, but these are untested. This pattern is used throughout Apply and Manage applications already.

The rest of this PR tidies up some code on these review pages, ensuring their layouts are consistent.

## Link to Trello card

https://trello.com/c/ij7yjN5x

## Things to check

### DIFF
--- a/app/views/candidate_interface/application_choices/review.html.erb
+++ b/app/views/candidate_interface/application_choices/review.html.erb
@@ -1,13 +1,24 @@
 <% content_for :title, t('page_titles.courses') %>
 <% content_for :before_content, govuk_back_link_to(candidate_interface_application_form_path, 'Back to application') %>
 
-<h1 class="govuk-heading-xl">
-  <% if @application_form.candidate_can_choose_single_course? %>
-    <%= t('page_titles.course_choice') %>
-  <% else %>
-    <%= t('page_titles.course_choices') %>
-  <% end %>
-</h1>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">
+      <% if @application_form.candidate_can_choose_single_course? %>
+        <%= t('page_titles.course_choice') %>
+      <% else %>
+        <%= t('page_titles.course_choices') %>
+      <% end %>
+    </h1>
+
+    <% if @application_choices.present? && @application_form.can_add_more_choices? %>
+      <div class="govuk-inset-text">
+        <p class="govuk-body">You can choose <%= pluralize(current_application.choices_left_to_make, "more course") %>.</p>
+        <%= govuk_button_link_to t('application_form.courses.another.button'), candidate_interface_course_choices_choose_path, class: 'govuk-button--secondary' %>
+      </div>
+    <% end %>
+  </div>
+</div>
 
 <%= form_with model: @application_form, url: candidate_interface_course_choices_complete_path do |f| %>
   <%= f.govuk_error_summary %>
@@ -17,10 +28,6 @@
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
       <% if @application_choices.present? %>
-        <% if @application_form.can_add_more_choices? %>
-          <%= govuk_button_link_to 'Add another course', candidate_interface_course_choices_choose_path, class: 'govuk-button--secondary govuk-!-margin-bottom-9' %>
-        <% end %>
-
         <% if @application_choices.count >= 1 %>
           <%= f.govuk_check_boxes_fieldset :course_choices_completed, legend: nil do %>
             <%= f.hidden_field :course_choices_completed, value: false %>
--- a/app/views/candidate_interface/volunteering/review/show.html.erb
+++ b/app/views/candidate_interface/volunteering/review/show.html.erb
@@ -2,22 +2,24 @@
 <% content_for :before_content, govuk_back_link_to(candidate_interface_application_form_path, 'Back to application') %>
 
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-full">
+  <div class="govuk-grid-column-two-thirds">
     <h1 class="govuk-heading-xl">
       <%= t('page_titles.volunteering.short') %>
     </h1>
 
-    <%= render(CandidateInterface::VolunteeringReviewComponent.new(application_form: @application_form)) %>
-
-    <%= govuk_button_link_to t('application_form.volunteering.another.button'), candidate_interface_new_volunteering_role_path, class: 'govuk-button--secondary' %>
-
-    <%= form_with model: @application_form, url: candidate_interface_complete_volunteering_path do |f| %>
-      <div class='govuk-form-group'>
-        <%= f.hidden_field :volunteering_completed, value: false %>
-        <%= f.govuk_check_box :volunteering_completed, true, multiple: false, label: { text: t('application_form.volunteering.review.completed_checkbox') } %>
-      </div>
-
-      <%= f.govuk_submit t('continue') %>
+    <%= govuk_button_link_to nil, candidate_interface_course_choices_choose_path, class: 'govuk-button--secondary' do %>
+      <%= @application_form.application_volunteering_experiences.any? ? t('application_form.volunteering.another.button') : t('application_form.volunteering.add.button') %>
     <% end %>
   </div>
 </div>
+
+<%= render(CandidateInterface::VolunteeringReviewComponent.new(application_form: @application_form)) %>
+
+<%= form_with model: @application_form, url: candidate_interface_complete_volunteering_path do |f| %>
+  <div class="govuk-form-group">
+    <%= f.hidden_field :volunteering_completed, value: false %>
+    <%= f.govuk_check_box :volunteering_completed, true, multiple: false, label: { text: t('application_form.volunteering.review.completed_checkbox') } %>
+  </div>
+
+  <%= f.govuk_submit t('continue') %>
+<% end %>
--- a/app/components/candidate_interface/add_reference_component.html.erb
+++ b/app/components/candidate_interface/add_reference_component.html.erb
@@ -1,31 +1,15 @@
 <% unless application_form.enough_references_have_been_provided? %>
-  <div class="govuk-inset-text govuk-!-margin-top-0">
+  <div class="govuk-inset-text">
     <% if no_viable_references? %>
-
-      <h2 class="govuk-heading-m">Add a referee</h2>
-
       <p class="govuk-body">You need 2 references before you can submit your application.</p>
-
       <%= govuk_button_link_to 'Add a referee', candidate_interface_references_start_path %>
-
     <% elsif one_viable_reference? %>
-
-      <h2 class="govuk-heading-m">Add a second referee</h2>
-
       <p class="govuk-body">You need 2 references before you can submit your application.</p>
-
       <%= govuk_button_link_to 'Add a second referee', candidate_interface_references_start_path %>
-
     <% else  %>
-
-      <h2 class="govuk-heading-m">Add another referee</h2>
-
       <p class="govuk-body">You can add more referees to increase the chances of getting 2 references quickly.</p>
-
       <p class="govuk-body">We’ll cancel any remaining requests when you’ve received 2 references.</p>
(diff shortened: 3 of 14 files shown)


adamsilver (commented):
Nice one @paulrobertlloyd.

Some of the questions you have asked are tricky to me because maybe it's more than just colour that plays a part here.

## Course choices

For example, should the ‘add another course’ button be green? 

I am inclined to say ‘yes and no’. 

No, because making it green means there's 2 green buttons drawing the user's attention in 2 opposing directions; and the candidate may consciously just want to apply for 1.

Yes, because it has advantages.

I wonder if radio buttons are better (to avoid 2 primary calls to action)...but I can imagine that's beyond the scope of this work...

...so if you want to do this iteratively, then I’d say keep it grey and maybe add guidance around the benefits of doing so.

(The guidance should not be ‘You can add further choices’ as that much is clear by the presence of the button.)

## Work history

In the first state, I think you need a button because of the way the content is laid out in the summary card. 

But I wouldn't include a button - instead I would have a row for ‘Do you have any work history?’ with a change link in order to change their answer and trigger that flow. This solves the other problem that we are not properly playing back the user's answers.

In the second state, there’s 3 ‘Add another job’ links/buttons and it’s not clear which one should be clicked. Making it green exacerbates the issue I think.

## Other qualifications

I think again, this should be grey - the guidance, call out and positioning is good.

duncanjbrown:
> No, because making it green means there's 2 green buttons drawing the user's attention in 2 opposing directions; and the candidate may consciously just want to apply for 1.

We see over 50% of candidates applying for a single course — I don't know if we want to be trying to move that in the direction of more multiple applications. Maybe a question for Rachael/Mike E?

adamsilver:
> > No, because making it green means there's 2 green buttons drawing the user's attention in 2 opposing directions; and the candidate may consciously just want to apply for 1.
> 
> We see over 50% of candidates applying for a single course — I don't know if we want to be trying to move that in the direction of more multiple applications. Maybe a question for Rachael/Mike E?

Yep, so now I’m wondering _why_ that is. If it's a conscious decision then that's fine. If it's because they don't realise they can or they don't realise the advantages of doing so, then that's more of a problem. 

Though in any case, the grey button (with guidance) solves that.

frankieroberto:
One quick thought: is it confusing to use the word “another” *before* the things it is ‘another’ of? Especially perhaps for users of screen readers who may be hearing the page read linearly?

gregknight1:
> One quick thought: is it confusing to use the word “another” _before_ the things it is ‘another’ of? Especially perhaps for users of screen readers who may be hearing the page read linearly?

My instinct is that this would be OK as the last action they will have done is to add something so I feel "another" is OK. Be good to test that, though.

frankieroberto:
> My instinct is that this would be OK as the last action they will have done is to add something so I feel "another" is OK. Be good to test that, though.

Good point. Although it's possible that the user will add something, and then come back to this page and some point later. But perhaps it's not that big a deal, as the confusion might get resolved fairly quickly. Although we could just reword "another" to "a" anyway?

adamsilver:
@frankieroberto even if they come back later, they have already added (for example) a degree so I think the label is not only fine, but it's good.

frankieroberto:
Looking at this again, I wonder about the inset text? The [Design System guidance](https://design-system.service.gov.uk/components/inset-text/) suggests using it "very sparingly", and I've definitely seen users skip over it in the past. I think it works for the referees, as it _only_ appears when you've not yet added 2 referees.  

Does the inset text and the heading imply that the user is missing something and _needs_ to add more information? Sadly I don't think we have any reasonable heuristic we could use to remove the inset text (and just keep the button) when "enough" qualifications or work history has been added? Although we could remove it once the "I have completed this section" box has been ticked.

paulrobertlloyd:
Thanks for the feedback. I’ve had a second look at this, and find myself going into interesting territory, that being ‘using blue accents when we give advise to candidates on how to improve their application’.

I quite like this approach, though I suspect we’d want to look at the content for each scenario to ensure the guidance uses similar tone (note how the headings are different; some repeat the button text, others are arguable better as they say why you should add another item).

### Course choices

We can use the same (or similar) content we show after you add a course and suggest you can add another N choices.

add-another-course-choice

### References

add-second-referee

add-another-referee

### Work history

(@adamsilver I take your point about the scenario in which no jobs have been added, and the action needing to be elsewhere, can look at that separately, I think)

add-another-job

### Volunteering

No need to emphasis this action (?), so sits as a grey button.

add-another-role

### Degrees

No need to emphasis this action, so sits as a grey button.

add-another-degree

### Other qualifications

add-another-qualification

adamsilver:
I am worried about blue buttons in their own right but also: GDS did AB tests on this (granted in a different context than this) and found users click on them far less.

I also think in the case of, for example, referees, using a green button is correct because the user really must add another one.

frankieroberto:
Would a "more prominent" link style work instead of the blue buttons? eg maybe blue and bold, possibly with a an arrow icon to the left or a (+) icon? See NHS [action links](https://service-manual.nhs.uk/design-system/components/action-link).

Another option might be to make the buttons green, but to remove the Continue buttons (and turn those into links) – but that'd mean removing the "I have completed this section" feature... 🤔

paulrobertlloyd:
Updated design

* Styles link as a link
* Uses action link component (based on [NHS Design System](https://service-manual.nhs.uk/design-system/components/action-link)/[GOV.UK Publishing](https://components.publishing.service.gov.uk/component-guide/action_link) component
* Where required, uses a second line of the link to add further hint/reasoning why you should add another. Again, needs some word love if we decide to go in this direction.

Worth pointing out how much more visible this link is compared to the secondary grey button, which is visually quite similar to that of the summary cards.

add-another-course-choice
add-second-referee
add-another-referee
add-another-degree
add-another-job
add-another-role
add-another-qualification

adamsilver:
Nice one for trying so many things @paulrobertlloyd.

I am worried about this because I don't think it's immediately obvious what the user should be doing as we have 2 prominent buttons fighting for attention. 

I am also worried about introducing a brand new button/link without a much larger body of evidence as to why we'd use it above design system components (including the secondary button).

paulrobertlloyd:
Updated initial description with the final design.

app/views/candidate_interface/application_choices/review.html.erb
```diff
@@ -1,13 +1,24 @@
...
+      <% if @application_form.candidate_can_choose_single_course? %>
+        <%= t('page_titles.course_choice') %>
+      <% else %>
+        <%= t('page_titles.course_choices') %>
+      <% end %>
+    </h1>
+
+    <% if @application_choices.present? && @application_form.can_add_more_choices? %>
+      <div class="govuk-inset-text">
+        <p class="govuk-body"><%= "You can choose #{current_application.choices_left_to_make} more #{'course'.pluralize(current_application.choices_left_to_make)}." %></p>
```
frankieroberto, line 16: I think you could use the [`pluralize`](https://api.rubyonrails.org/v6.1.0/classes/ActionView/Helpers/TextHelper.html#method-i-pluralize) helper here to simplify this, by pluralizing "more course" / "more courses".

```suggestion
        <p class="govuk-body">You can choose <%= pluralize(current_application.choices_left_to_make), "more course" %>.</p>
```

app/views/candidate_interface/volunteering/review/show.html.erb
```diff
@@ -2,22 +2,24 @@
...
-    <%= govuk_button_link_to t('application_form.volunteering.another.button'), candidate_interface_new_volunteering_role_path, class: 'govuk-button--secondary' %>
-
-    <%= form_with model: @application_form, url: candidate_interface_complete_volunteering_path do |f| %>
-      <div class='govuk-form-group'>
-        <%= f.hidden_field :volunteering_completed, value: false %>
-        <%= f.govuk_check_box :volunteering_completed, true, multiple: false, label: { text: t('application_form.volunteering.review.completed_checkbox') } %>
-      </div>
-
-      <%= f.govuk_submit t('continue') %>
+    <%= govuk_button_link_to nil, candidate_interface_course_choices_choose_path, class: 'govuk-button--secondary' do %>
```
frankieroberto, line 10: It'd be nice if `govuk_button_link_to` behaved the same as Rails’s [`link_to`](https://api.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to) helper, where you can avoid passing in `nil` if you're specifying the link text as a block. Maybe a dev could help us with this?

app/views/candidate_interface/volunteering/review/show.html.erb
```diff
@@ -2,22 +2,24 @@
...
+      <%= @application_form.application_volunteering_experiences.any? ? t('application_form.volunteering.another.button') : t('application_form.volunteering.add.button') %>
```
frankieroberto, line 11: There's a neat trick my last team used to handle these kind of conditions within the locale file itself by doing:

```suggestion
      <%= t('application_form.volunteering.add_button', @application_form.application_volunteering_experiences.any? %>
```

And then in the locale file doing:

```yaml
add_button:
  false: Add a role
  true: Add another role
```

Worth exploring?

frankieroberto (approved):
Thanks for exploring all the options!

Left some minor comments/suggestions, but nothing blocking a merge.

gregknight1 (approved):
LGTM

app/views/candidate_interface/volunteering/review/show.html.erb
paulrobertlloyd: Yeah, I’ve come across this same issue with the `govuk_link_to` helper, to which I added a `TODO` comment. Will add a `TODO` comment for this helper too.

app/views/candidate_interface/application_choices/review.html.erb
paulrobertlloyd: 👌

app/views/candidate_interface/volunteering/review/show.html.erb
paulrobertlloyd: It’s a nice idea… however tempted to leave this as really there shouldn’t ever be a ‘Add a job’ button on this page, (as @adamsilver noted, the playback of a page with no work history/no volunteering should show the answer to the question, which you can then change)

frankieroberto: I'm not 100% convinced by never needing an "Add a job" button, even though in the prototype I've now implemented this by redirecting back to the first branching question if you have no jobs left after deleting one. It feels a bit weird though, and I think the more expected thing would be to return to the "You've said you have some work history, but you've not yet added any jobs" state, with some other mechanism to change your answer to the branching question.

Equally perhaps you ought to be able to change your answer to the branching question without having to delete all the jobs you've added first? But that feels even less likely.

Gah, it's so tricky!

PR #3821, opened 2021-01-11
Add received at columns to Application Reference Data Extract

george-bit:
## Context

As a requirement on the current Application Reference Data Extract, business analysts require
the capacity to view the time at which references have been completed.

This solution builds on the
existing data extract model and adds the columns along with a method to find find the received_at
time in the audit log.

## Guidance to review

Test in the review app and review test coverage.

## Link to Trello card

https://trello.com/c/utHH62RD/2755-dev-update-application-references-data-extract

## Things to check

### DIFF
--- a/app/services/support_interface/application_references_export.rb
+++ b/app/services/support_interface/application_references_export.rb
@@ -15,6 +15,7 @@ module SupportInterface
           output["Ref #{index} type"] = reference.referee_type
           output["Ref #{index} state"] = reference.feedback_status
           output["Ref #{index} requested at"] = reference.requested_at
+          output["Ref #{index} received at"] = received_at(reference)
         end
 
         output
@@ -23,6 +24,12 @@ module SupportInterface
       # The DataExport class creates the header row for us so we need to ensure
       # we sort by longest hash length to ensure all headers appear
       data_for_export.sort_by(&:length).reverse
+    end
+
+  private
+
+    def received_at(reference)
+      reference.audits.where("audited_changes#>>'{feedback_status, 1}' = 'feedback_provided'").last&.created_at
     end
   end
 end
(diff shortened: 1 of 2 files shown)


app/services/support_interface/application_references_export.rb
```diff
@@ -24,5 +25,11 @@ def data_for_export
       # we sort by longest hash length to ensure all headers appear
       data_for_export.sort_by(&:length).reverse
     end
+
+    private
+
+    def received_at(reference)
+      reference.audits.where("audited_changes#>>'{feedback_status, 1}' = 'feedback_provided'").last&.created_at
```
raamSoftwire, line 32: I've not come across this syntax (`#>>`) before, are you able to explain it to me, please?

raamSoftwire (approved):
Looks good to me, cheers ~~Toby~~ George!

app/services/support_interface/application_references_export.rb
george-bit: Hey @raamSoftwire so this was a little tip from David. Basically, we're passing in an sql query based off of the work that Steve did on the candidate_journey_tracker.rb. This check the audit object associated with the reference to find the time of the last audit that occurred when feeback_status was updated to "feedback_provided". This is kind of a work around as 'created_at' does not exist currently as a column in the database. David has added this as a separate card (as we;re now using the workaround in multiple places. (Also, is the below approval for a different PR?)

raamSoftwire:
> Looks good to me, cheers ~Toby~ George!

That approval was for this PR 🤦🏽‍♂️ I can see it works from the tests, I just wanted to understand that one line

app/services/support_interface/application_references_export.rb
davidgisbey: @stevehook  can correct me if I'm wrong, but I believe the functionality used to query JSON blobs can also be applied to hashes. 

It checks nested values. So it looks for audited changes. Which returns this.

`{ "feedback_status" => ["feedback_requested", "feedback_provided"] }`

It then checks that the key is `feedback status` and the value of the second element of the array is `feedback_provided`

https://www.postgresql.org/docs/9.3/functions-json.html

PR #3823, opened 2021-01-12
Break down `application_form.yml` file

raamSoftwire:
## Context

At the moment the application_form.yml file is too long, so we want to split it out.  We should break this down so that it is reflective of the folder structure we use (see the structure inside `app/views/candidate_interface` for an example)

Extracting big chunks of the `application_form.yml` to separate files, including the `activemodel.errors` code.

## Guidance to review

- Review commit by commit
- There is a lot of copypasting here and so a higher than likely chance for human error. Have I dropped a line somewhere?
- ~At the moment, everything is still scoped by `application_form`. Is that necessary?~ Yes, see Duncan's comment
- ~The locales directory is starting to grow, is it worth collecting all of these new files and the `application_form.yml` into a nested directory?~ Yes, I've now done this
- It might be worth pulling the code and having a look at the file structure more naturally, to get a feel for it after these changes

## Link to Trello card

https://trello.com/c/jviWkNFf/2830-break-down-applicationformyml-file

## Things to check


duncanjbrown:
> There is a lot of copypasting here and so a higher than likely chance for human error. Have I dropped a line somewhere?

I believe we have `raise_on_missing_translations` enabled, which should cause errors where things have been missed out. If you're anxious you could probably parse and combine the YAML from all the separated files, sort it somehow, and compare that with the parsed YAML. 

> At the moment, everything is still scoped by application_form. Is that necessary?
> The locales directory is starting to grow, is it worth collecting all of these new files and the application_form.yml into a nested directory?

IMO this namespace is useful as the nested objects also live elsewhere — eg there are `contact_details` in support, in Manage etc. If they're all going to have that key, putting them all into an `application_form` directory seems logical.

raamSoftwire:
I've created an `emails` directory inside `locales` since that seems like a clear distinction to make. I feel like there's another that could be pulled out, but I'm struggling for a name. It's for strings that aren't used for views (take a look in `locales/reference.yml` to see what I mean). There's less of them so I'm less sure than I was for the emails.

config/locales/application_form/application_form.yml
```diff
@@ -0,0 +1,112 @@
...
+            provider_id:
+              blank: Select a training provider
+        candidate_interface/pick_study_mode_form:
+          attributes:
+            study_mode:
+              blank: Select if the course is full time or part time
+        candidate_interface/pick_site_form:
+          attributes:
+            course_option_id:
+              blank: Select which location you’re applying to
```
paulrobertlloyd, line 22: Can these be moved `courses.yml`?

config/locales/application_form/application_form.yml
```diff
@@ -0,0 +1,112 @@
...
+        candidate_interface/add_another_referee_form:
+          attributes:
+            add_another_referee:
+              blank: Select if you want to add another referee
```
paulrobertlloyd, line 26: Can this be moved to `references.yml`?

config/locales/application_form/application_form.yml
```diff
@@ -0,0 +1,112 @@
...
+          attributes:
+            registration_number:
+              blank: Enter your registration number
+            total_score:
+              blank: Enter your total score
+              not_a_number: Enter your total score (this should be a number)
+            award_year:
+              blank: Enter year the assessment was done
+              invalid: Enter a real year
+              in_the_future: Assessment year must be this year or a previous year
```
paulrobertlloyd, line 70: Can these move to a separate locale file for the EFL section?

config/locales/application_form/application_form.yml
```diff
@@ -0,0 +1,112 @@
...
+        candidate_interface/safeguarding_issues_declaration_form:
+          attributes:
+            share_safeguarding_issues:
+              blank: Choose if you want to share any safeguarding issues
+            safeguarding_issues:
+              too_many_words: Safeguarding issues must be %{count} words or fewer
```
paulrobertlloyd, line 76: Can these move to a separate safeguarding locale file?

paulrobertlloyd (commented):
Probably a separate commit, but looking at the emails, they use these keys:

* `chase_candidate_decision_email`
* `candidate_mailer`
* `reference_request`
* `candidate_offer`
* `provider_mailer`

Not only do the file names not always reflect the key name, I’m not sure if these keys reflect the architecture of the app; should they all fall under the following keys:

* `candidate_mailer`
* `referee_mailer`
* `provider_mailer`

raamSoftwire:
> Probably a separate commit, but looking at the emails, they use these keys:
> 
> * `chase_candidate_decision_email`
> * `candidate_mailer`
> * `reference_request`
> * `candidate_offer`
> * `provider_mailer`
> 
> Not only do the file names not always reflect the key name, I’m not sure if these keys reflect the architecture of the app; should they fall under the following keys:
> 
> * `candidate_mailer`
> * `referee_mailer`
> * `provider_mailer`

Good spot, I've tidied those up now.

config/locales/emails/candidate_mailer.yml
```diff
@@ -6,6 +6,16 @@ en:
...
     application_submitted_apply_again:
       subject: "You’ve submitted your teacher training application"
+    candidate_offer:
+      single_offer:
+        subject: "Make a decision: successful application for %{provider_name}"
+      multiple_offers:
+        subject: "Make a decision: successful application for %{provider_name}"
+      decisions_pending:
+        subject: "Successful application for %{provider_name}"
+    chase_candidate_decision_email:
```
paulrobertlloyd, line 16: `email` seems redundant here

```suggestion
    chase_candidate_decision:
```

davidgisbey (approved):
Amazing job Raam. This is so much better. I'll make sure I'm keeping this in mind when I'm reviewing code so we can keep it this orderly!

Once we've figured out the shared strings that are used across interfaces and we can put the remaining ones into namespaces this is going to be so intuitive to use!

PR #3826, opened 2021-01-12
Add support for parallel tests

malcolmbaig:
## Context
For faster test runs, we want to be able to run our test suite in parallel using multiple cores.

- Add the [parallel_tests](https://github.com/grosser/parallel_tests) gem
- Update config for Postgres, Capybara, and Redis
- Document how to run the test suite in parallel locally

## Guidance to review
- Per-commit gives a good breakdown of the changes
- Does the doc make sense?
- Run it locally! Are you seeing any issues?

## Link to Trello card
https://trello.com/c/ySrLogAF

## Things to check

### DIFF
--- a/docs/parallel_tests.md
+++ b/docs/parallel_tests.md
@@ -1,0 +1,54 @@
+# Parallel Tests
+
+This document explains how to run the test suite across multiple cores in development.
+
+## Prerequisites
+
+### Check environment
+
+Ensure that REDIS_URL is not present in .env.test or .env.
+
+In order to successfully run tests in parallel, each process needs its own Redis database. We determine the Redis database id in [ApplyRedisConnection](app/lib/apply_redis_connection.rb). This class always uses REDIS_URL if present in the environment. In order for it to return the appropriate URL for test runs, we need to ensure REDIS_URL is not set in the test environment.
+
+### Prepare Postgres databases
+
+Run the following commands.
+
+Create additional databases:
+`bundle exec rake parallel:create`
+
+Copy development schema (repeat this after migrations)
+`bundle exec rake parallel:prepare`
+
+### Update Redis database count if needed
+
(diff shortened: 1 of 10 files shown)


docs/parallel_tests.md
```diff
@@ -0,0 +1,54 @@
+# Parallel Tests
+
+This document explains how to run the test suite across multiple cores in development.
+
+## Prerequisites
+
+### Check environment
+
+Ensure that REDIS_URL is not present in .env.test or .env.
```
davidgisbey, line 9: Why .env?

davidgisbey, line 9: why in .env?

davidgisbey (approved):
Just set it up and ran the tests locally. They were significantly quicker. Like wayyyy quicker. Thanks for all your work on this Malcolm!

I only have 8 cores btw so if someone has 16 it might be worth them following that section of the doc.

duncanjbrown (approved):
🏎️ This is great!  `1m46s` for the whole suite on my machine. And docs too 👌👌👌.

I liked having the same output in and out of `parallel_tests` so I copied the `--format` line in `.rspec` to `.rspec_parallel` — would you consider adopting that?

(Only 8 cores here too).

malcolmbaig:
@davidgisbey @duncanjbrown Excellent, thank you both for testing.

> I only have 8 cores btw so if someone has 16 it might be worth them following that section of the doc.

Had to do this locally, can confirm the fix works nicely.

> I liked having the same output in and out of parallel_tests so I copied the --format line in .rspec to .rspec_parallel — would you consider adopting that?

I haven't quite figured out how to collate test failures from each process and print them all at the end. With `progress` formatting, there's much less scrolling back to identify what's failed. Happy to change over to `documentation` if we can find a nicer way to output errors.

PR #3831, opened 2021-01-12
Specify healthCheckPath to determine if app service instance is healthy

vigneshmsft:
Specify a healthCheckPath to determine if an instance
is healthy and can be included in the load balancer pool.

## Context

When running more than 1 instance of the app, 
the healthCheckPath is used to determine the health of the specific instance and if it can service traffic.
If the path returns a non 2xx status the instance is removed from the azure load balancer pool.

Configure '/check' as healthCheckPath for each instance
Drop the instance from the pool after 5 failed pings.
## Guidance to review

To be merged after https://github.com/DFE-Digital/bat-platform-building-blocks/pull/82
Tested in [DevOps ](https://dev.azure.com/dfe-ssp/Become-A-Teacher/_build/results?buildId=137232&view=results)environment

## Things to check

### DIFF
--- a/azure/template.json
+++ b/azure/template.json
@@ -688,6 +688,10 @@
                                 "value": "0"
                             },
                             {
+                                "name": "WEBSITE_HEALTHCHECK_MAXPINGFAILURES",
+                                "value": "5" //At 1 ping per minute
+                            },
+                            {
                                 "name": "RAILS_SERVE_STATIC_FILES",
                                 "value": "[parameters('railsServeStaticFiles')]"
                             },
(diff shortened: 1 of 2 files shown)


azure/template.json
```diff
@@ -687,6 +687,10 @@
                                 "name": "WEBSITE_SLOT_POLL_WORKER_FOR_CHANGE_NOTIFICATION",
                                 "value": "0"
                             },
+                            {
+                                "name": "WEBSITE_HEALTHCHECK_MAXPINGFAILURES",
+                                "value": "5" //At 1 ping per minute
```
saliceti, line 692: I don't think you can add a comment in json

saliceti: Apparently in ARM templates you can: https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-syntax#comments

saliceti (approved):

PR #3835, opened 2021-01-13
Avoid inadvertent rejection of applications

tijmenb:
## Context

We've got [an application](https://www.apply-for-teacher-training.service.gov.uk/support/applications/8459/audit) that was [inadvertently rejected by a provider](https://becomingateacher.zendesk.com/agent/tickets/11274). I think it's some kind of back-button issue.

Normally, this wouldn't be a problem, because our state machine protects against weird transitions. However, the state transition to reject an offer and the transition to withdraw an existing offer is the same (#reject!). This means that the `RejectApplication` service was invoked, and not the `WithdrawOffer` service. Because of this we don't set the `offer_withdrawn_at` attribute, which in turn crashes the offer page.

This changes the transition so any inadvertent invocation of the `RejectApplication` service for offered applications will result in a crash rather than an inconsistent state. Hopefully it will allow us to find the scenario where this occurred so we can fix it.

## Link to Trello card

https://trello.com/c/4IVO2PzH/111-ref-an2514

### DIFF
--- a/app/services/application_state_change.rb
+++ b/app/services/application_state_change.rb
@@ -53,7 +53,7 @@ class ApplicationStateChange
 
     state :offer do
       event :make_offer, transitions_to: :offer
-      event :reject, transitions_to: :offer_withdrawn
+      event :withdraw_offer, transitions_to: :offer_withdrawn
       event :accept, transitions_to: :pending_conditions
       event :decline, transitions_to: :declined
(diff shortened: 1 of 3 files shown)


mnacos (approved):
Yes, `withdraw_offer` is a much better transition name for this case.

duncanjbrown (approved):

PR #3836, opened 2021-01-13
Speed up UCAS spec

malcolmbaig:
## Context

This is one of the slower specs in the test suite. Running the [RSpecDissect](https://test-prof.evilmartians.io/#/profilers/rspec_dissect) profiler against it indicates that this is largely because of the amount of time it spends in `let` (ie - creating records in the database).

The approach recommended by the test-prof docs for dealing with this is using the `let_it_be`/`before_all` methods provided by the gem. These methods hook into RSpec's `before(:context)` to allow database records to be set up once before executing examples in the group (rather than recreating them for each example). Any modifications to these records are automatically rolled back between examples, and an option is available to automatically reload the ActiveRecord object as well.

The unfamiliar syntax and new behaviour can potentially make the spec more difficult to follow. Raising this PR for discussion to weigh up the performance improvement vs impact on clarity.

Use the `let_it_be`/`before_all` methods in ucas_matched_application_spec. This change reduces the runtime from 6.72 seconds to 0.5 seconds (tested on a machine with an AMD Ryzen 7 3700X).

## Guidance to review

- Is it still reasonably clear what's going on in the spec?
- We can specify an alias for `let_it_be` if we think that would help clarify things.

## Link to Trello card
https://trello.com/c/ySrLogAF

## Things to check

### DIFF
--- a/spec/models/ucas_matched_application_spec.rb
+++ b/spec/models/ucas_matched_application_spec.rb
@@ -1,25 +1,25 @@
 require 'rails_helper'
 
 RSpec.describe UCASMatchedApplication do
-  let(:course) { create(:course, recruitment_cycle_year: 2020) }
-  let(:candidate) { create(:candidate) }
-  let(:course_option) { create(:course_option, course: course) }
-  let(:application_choice) { create(:application_choice, course_option: course_option) }
-  let(:application_form) { create(:completed_application_form, candidate_id: candidate.id, application_choices: [application_choice]) }
-  let(:apply_again_application_form) { create(:application_form, candidate_id: candidate.id) }
-  let(:recruitment_cycle_year) { 2020 }
-  let(:candidate1) { create(:candidate) }
-  let(:application_choice1) { create(:application_choice, :with_accepted_offer, course_option: course_option) }
-  let(:application_form1) { create(:completed_application_form, candidate_id: candidate1.id, application_choices: [application_choice1]) }
-  let(:candidate2) { create(:candidate) }
-  let(:application_choice2) { create(:application_choice, :with_rejection, course_option: course_option) }
-  let(:application_form2) { create(:completed_application_form, candidate_id: candidate2.id, application_choices: [application_choice2]) }
+  # Reduce database interaction by using the let_it_be/before_all methods. This
+  # creates each record once before running the examples in this group.
+  let_it_be(:course) { create(:course, recruitment_cycle_year: 2020) }
+  let_it_be(:candidate) { create(:candidate) }
+  let_it_be(:course_option) { create(:course_option, course: course) }
+  # Use the reload: true option to reload this object after we roll back any
+  # changes made in the examples below.
+  let_it_be(:application_choice, reload: true) { create(:application_choice, course_option: course_option) }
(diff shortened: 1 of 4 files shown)


spec/models/ucas_matched_application_spec.rb
```diff
@@ -1,25 +1,25 @@
...
-    application_form1
-    application_form2
+  # Reduce database interaction by using the let_it_be/before_all methods. This
+  # creates each record once before running the examples in this group.
+  let_it_be(:course) { create(:course, recruitment_cycle_year: 2020) }
+  let_it_be(:candidate) { create(:candidate) }
+  let_it_be(:course_option) { create(:course_option, course: course) }
+  # Use the reload: true option to reload this object after we roll back any
+  # changes made in the examples below.
+  let_it_be(:application_choice, reload: true) { create(:application_choice, course_option: course_option) }
```
davidgisbey, line 11: Personally, I think this is fine and worth the trade-off. However, if we're going to use this we should probably update the testing styleguide  https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/docs/testing-styleguide.md with an explanation of how it works and why we chose to use it.

Interested to hear what others think.

TobyRet: Nice improvement! Luckily, this shouldn't cause issues with running specs in parallel as this is done at the file level (?)
I think the 'let_it_be` syntax is a good option where you want to refactor an existing slow spec with minimal changes (i.e. rather than converting to before(:all) and writing instance vars). With regards to new and potentially slow specs - do we use the `let_it_be` syntax or use `before(:all)`

despo: That's an interesting improvement and a quick win. However, I find that this just masks the problem as what we should really be tackling is figuring out how we can right more efficient specs rather than introducing new dependencies that we also need to document and learn how to use.

We can also potentially avoid having to use before blocks by disabling the RSpec/LetSetup check so we can use `let!()` instead, which is a lot cleaner and maybe educate the team in our documentation about when to use this feature instead of preventing usage.

Maybe we could give refactoring ucas_matched_application soec a shot before considering merging this pull request?
(similar to here https://github.com/DFE-Digital/apply-for-teacher-training/pull/3863/files#diff-4f894049af3375c2bd4e608f546f8d4a0eed95464efcdea850993200db9fef5c)

malcolmbaig: > Maybe we could give refactoring ucas_matched_application soec a shot before considering merging this pull request?

Yeah I think that's sensible. 👍🏽 No rush getting this in, and if we can reduce the run time using existing syntax, all the better.

despo: @malcolmbaig  I may have already spend sometime on this last night: https://github.com/DFE-Digital/apply-for-teacher-training/pull/3871

malcolmbaig:
Closing in favour of https://github.com/DFE-Digital/apply-for-teacher-training/pull/3871. That gets us most of the way in terms of performance without needing to introduce new syntax. Can revisit this technique if there are specs that need heavy, one-off DB setup but are difficult to stub without affecting clarity.

PR #3839, opened 2021-01-13
Support: improve how we display the course when a different course is offered

tijmenb:
## Context

Providers can offer a different course than the one the candidate applied to. It's not super clear in the support UI that this has happened, and what the changes are. Making it clearer in support might make it easier to debug issues like this one https://becomingateacher.zendesk.com/agent/tickets/11268, in which a provider is confused about which course was offered.

Before, when the offered course is not different (or no offer):

Screenshot 2021-01-13 at 14 38 40

After:

Screenshot 2021-01-13 at 14 38 06

Before, when the offered course is different from the applied course:

Screenshot 2021-01-13 at 14 43 16

After:

Screenshot 2021-01-13 at 14 42 46

## Link to Trello card

https://trello.com/c/nkRuL1iT/109-provider-has-changed-the-course-they-are-offering-a-candidate-however-they-are-not-seeing-this-mirrored-on-the-system-ticket-num

### DIFF
--- a/app/components/support_interface/application_choice_component.rb
+++ b/app/components/support_interface/application_choice_component.rb
@@ -10,14 +10,17 @@ module SupportInterface
 
     def rows
       rows = [
-        { key: 'Recruitment cycle', value: application_choice.offered_option.course.recruitment_cycle_year },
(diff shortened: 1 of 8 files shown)


davidgisbey (approved):
I'm not sure if someone from ProVendor needs to approve too, but it looks pretty uncontroversial It's much clearer what's going on now 👍

PR #3844, opened 2021-01-13
Remove Course#qualifications & program_type from Find sync

stevehook:
## Context

The Find sync was overwriting the values of `Course#qualifications` & `program_type` that are synched from the TTAPI with `nil`. This sets up a perpetual cycle that toggles the values from one to the other twice an hour. As a result the audit trail for `Course`s is badly bloated. 

- [x] Just remove these attributes from the find sync.

## Guidance to review

- Is there anything else that could cause audit bloat?
- If necessary we can log another card to clean up some of the redundant entries in the audit trail.

## Link to Trello card

https://trello.com/c/qfdMyoFJ/2850-investigate-audit-trail-bloat

## Things to check

### DIFF
--- a/app/services/find_sync/sync_courses_from_find.rb
+++ b/app/services/find_sync/sync_courses_from_find.rb
@@ -96,6 +96,8 @@ module FindSync
       course.exposed_in_find = find_course.findable?
       course.subject_codes = find_course.subject_codes
       course.funding_type = find_course.funding_type
+      # this field is in the API docs, but not actually in the API yet
+      course.program_type = find_course.try(:program_type)
       course.age_range = find_course.age_range_in_years&.humanize
       course.withdrawn = find_course.content_status == 'withdrawn'
     end
--- a/app/services/find_sync/sync_courses_from_find.rb
+++ b/app/services/find_sync/sync_courses_from_find.rb
@@ -96,9 +96,6 @@ module FindSync
       course.exposed_in_find = find_course.findable?
       course.subject_codes = find_course.subject_codes
       course.funding_type = find_course.funding_type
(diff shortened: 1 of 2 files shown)


app/services/find_sync/sync_courses_from_find.rb
```diff
@@ -96,9 +96,6 @@ def assign_course_attributes_from_find(course, find_course)
       course.exposed_in_find = find_course.findable?
       course.subject_codes = find_course.subject_codes
       course.funding_type = find_course.funding_type
-      # these two fields are in the API docs, but not actually in the API yet
-      course.program_type = find_course.try(:program_type)
```
davidgisbey, line 100: I added program type to the v3 a while ago as Mike needed it for analysis.

https://github.com/DFE-Digital/teacher-training-api/blob/master/app/serializers/api/v3/serializable_course.rb#L24

stevehook: I'll reinstate the `program_type` sync; it doesn't seem to be the cause of any extra audit entries. I was originally concerned that it might conflict with https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/services/teacher_training_public_api/sync_courses.rb#L33

davidgisbey (approved):
All good to go 👍

app/services/find_sync/sync_courses_from_find.rb
davidgisbey: Yea I'd missed that. Do we actually need the quals? If so we can just add it to the public API. We've gone for this far without is so i'd assume not?

PR #3846, opened 2021-01-13
Fix issue with referee sign in creating excess tokens

davidgisbey:
## Context

 Currently, when you visit the application show page on the support interface, a new reference token is created for each reference in the feedback requested state.

This PR adds the `impersonate_and_give` and `impersonate_and_decline` endpoints to the reference controller. These endpoints do 2 things.

 1. Call the reference.refresh_feedback_token! method
 2. Redirect to the appropriate page of reference flow

## Guidance to review

Does this all seem reasonable?

## Things to check


stevehook (approved):
This looks good to me, tested locally and only the required tokens were generated.

PR #3848, opened 2021-01-14
Implement set up an interview flow

despo:
~~Waiting on for PR of https://trello.com/c/AyfUzn5t/3226-implement-create-interview-form and https://github.com/DFE-Digital/apply-for-teacher-training/pull/3819 to be merged.~~

## Context

Enables provider users to view, create and cancel application choice interviews

- Introduce interviews Feature flag
- Ability to filter by Interviewing state
- Enable provider users to create new interviews if application choice is awaiting_provider_decision
- Enable providers users to cancel an interview
- List interviews in interviewing tab is application_choice is in interviewing state
- Add interviews system test

## Link to Trello card

https://trello.com/c/GapoDi9e/3214-implement-set-up-an-interview-flow

## Things to check


app/controllers/provider_interface/interviews_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    def new
+      CreateInterview.new(
+        actor: current_provider_user,
+        application_choice: @application_choice,
+        provider: @application_choice.offered_course.provider,
+        date_and_time: Time.zone.now,
+        location: 'Zoom',
+        additional_details: 'Some preferences',
+      ).save!
```
despo, line 24: To be removed. As it enables testing on the review app so will leave for now.

app/services/cancel_interview.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def save!
+    auth.assert_can_make_decisions!(
+      application_choice: application_choice,
+      course_option_id: application_choice.offered_option.id,
+    )
+
+    if ApplicationStateChange.new(application_choice).can_cancel_interview?
+      interview.update!(cancellation_reason: cancellation_reason,
+                        cancelled_at: Time.zone.now)
```
despo, line 24: @steventux in relation to deleted interviews

app/services/cancel_interview.rb
```diff
@@ -0,0 +1,31 @@
...
+    @cancellation_reason = cancellation_reason
+  end
+
+  def save!
+    auth.assert_can_make_decisions!(
+      application_choice: application_choice,
+      course_option_id: application_choice.offered_option.id,
+    )
+
+    if ApplicationStateChange.new(application_choice).can_cancel_interview?
```
benswannack, line 22: Currently, this will fail silently if you have all the permissions, but the application is not in the right state. Can we do something to let the user know that the cancellation failed?

(I think the same is true for the CreateInterview service)

app/views/provider_interface/interviews/review_cancel.html.erb
```diff
@@ -0,0 +1,32 @@
...
+        Check and send interview cancellation
+      </label>
+    </h1>
+
+    <%= render SummaryCardComponent.new(editable: true, border: false, rows: [
+    {
+      key: 'Reason for cancelling interview',
+      value: @interview.cancellation_reason
+    },
+    {}
```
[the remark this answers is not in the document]
despo: Yes as it's the only way to get the bottom border displayed.

app/components/provider_interface/application_choice_header_component.rb
```diff
@@ -77,7 +93,7 @@ def show_inset_text?
       flash.empty? && (
         respond_to_application? || deferred_offer_wizard_applicable? ||
         rejection_reason_required? || provider_cannot_respond?
-      )
+      ) || waiting_for_interview?
```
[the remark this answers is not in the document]
despo: I don't think so as this is not the final state of the application and they should still be able to make decisions and see how many days are left until RBD.

@adamsilver when the provider succesfuly sets up an interview and is taken back to the main application page, should the notification message be displayed **as well as the box with the days left until rejection and the make decision button?**

If I remember correctly this is how the prototype seems to behave as well.

app/services/cancel_interview.rb
despo: Added else condition that throws exception. Under other circumstances I would have changed the order of the actions in a transaction and let `cancel_interview!` to throw the exception but in this instance as we can only transition the state if there are no other interviews it's not possible.

Also updated `CreateInterview`.

app/components/provider_interface/application_choice_header_component.html.erb
```diff
@@ -9,17 +9,40 @@
       <div class="govuk-inset-text govuk-!-margin-top-0">
         <% if respond_to_application? -%>
           <h2 class="govuk-heading-m govuk-!-margin-bottom-2">
-            Respond to application
+            <% if FeatureFlag.active?(:interviews) %>
+              Set up an interview or make a decision
+            <% else %>
+                Make a decision
```
mnacos, line 15:
```suggestion
              Make a decision
```

I guess

app/components/provider_interface/application_choice_header_component.rb
```diff
@@ -57,10 +63,20 @@ def offer_present?
       ApplicationStateChange::OFFERED_STATES.include?(application_choice.status.to_sym)
     end
 
+    def interviews_present?
+      return false unless FeatureFlag.active?(:interviews)
+
+      application_choice.interviewing? && application_choice.interviews.any?
```
mnacos, line 69: Does `interviews.any?` take into account the possibility of cancelled interviews?

mnacos (approved):
Works well. Just a question around the default scope for associations when using the discard gem.

Also, based on the PR title I was expecting to see the create flow implemented, too. Perhaps rename to 'Implement interview cancellation flow' or equivalent?

Finally, do we have factories/traits for these things (e.g. cancelled interview) Perhaps we could do this as a separate PR.

app/components/provider_interface/application_choice_header_component.rb
despo: Good catch. I assumed that the gem works like by adding a default scope (like [paranoia](https://github.com/rubysherpas/paranoia) but that doesnt seem to be the case.

I can either update this check of implement a default scope on the interviews model to do that. I think the **former is probably the best idea** as we don't want to potential impact other areas where we might want to access deleted interviews e.g. support

mnacos: 👍🏼

app/controllers/provider_interface/interviews_controller.rb
mnacos: How about we introduce a `before_action` for the entire controller requiring the feature flag to be on? Along the lines of https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/controllers/provider_interface/reasons_for_rejection_controller.rb#L3

Then we don't need to worry about what `new` or any other action does etc.

despo: There i a feature flag. I just dont want to merge anything which hardcodes informatino (even though I am aware that we should never get here unless the flag is enabled).

app/controllers/provider_interface/interviews_controller.rb
```diff
@@ -0,0 +1,79 @@
+module ProviderInterface
+  class InterviewsController < ProviderInterfaceController
+    before_action :set_application_choice
+    before_action :interview_flag_enabled?
```
despo, line 4: @mnacos

app/controllers/provider_interface/interviews_controller.rb
mnacos: In that case, great!

PR #3849, opened 2021-01-14
Extend GetActivityLog to retrieve associated audits for interviews

steventux:
## Context

Activity log and application time line need to able to render interview audit events

Updates `GetActivityLogEvents` query to also retrieve interview events associated with the given applications.

### Application timeline
Updates the application timeline to allow interview audit items to appear. The corresponding application state change won't appear in this view.

image

### Activity log
Updates the activity log to render interview audit events, no events are filtered in this view.
image

## Guidance to review

There's an update necessary to links to the interview tab, for the moment I've linked to the application details. This will need to be updated once we have the tab route in master.

## Link to Trello card

https://trello.com/c/DrOghuNb/3235-extend-getactivitylog-to-retrieve-associatedaudits

## Things to check

### DIFF
--- a/app/components/provider_interface/activity_log_event_component.rb
+++ b/app/components/provider_interface/activity_log_event_component.rb
@@ -77,10 +77,7 @@ module ProviderInterface
         "#{course_option.course.name_and_code} at #{course_option.site.name}"
       elsif event.audit.auditable.is_a?(Interview)
         interview = event.audit.auditable
-        title = "Interviewing on #{interview.date_and_time.to_s(:govuk_date_and_time)}"
-        title << ", #{interview.location}" if interview.location.present?
-
-        title
+        "Interviewing on #{interview.date_and_time.to_s(:govuk_date_and_time)}, #{interview.location}"
       end
     end
 
--- a/app/components/provider_interface/activity_log_event_component.rb
+++ b/app/components/provider_interface/activity_log_event_component.rb
@@ -72,7 +72,7 @@ module ProviderInterface
                          end
     end
 
-    def event_note_title
+    def event_title
       if application_choice.present?
         "#{course_option.course.name_and_code} at #{course_option.site.name}"
       elsif event.audit.auditable.is_a?(Interview)
@@ -81,7 +81,7 @@ module ProviderInterface
       end
     end
 
-    def event_note_content
+    def event_content
       if application_choice.present?
         content = course_option.provider.name
 
--- a/app/components/provider_interface/activity_log_event_component.rb
+++ b/app/components/provider_interface/activity_log_event_component.rb
@@ -120,7 +120,7 @@ module ProviderInterface
           }
         elsif event.audit.auditable.is_a?(Interview)
           {
-            url: routes.provider_interface_application_choice_path(event.audit.associated),
+            url: routes.provider_interface_application_choice_interviews_path(event.audit.associated, anchor: "interview-#{event.audit.auditable.id}"),
             text: 'View interview',
           }
         else
--- a/app/components/provider_interface/activity_log_event_component.rb
+++ b/app/components/provider_interface/activity_log_event_component.rb
@@ -15,6 +15,8 @@ module ProviderInterface
       user = event.user_full_name
       candidate = event.candidate_full_name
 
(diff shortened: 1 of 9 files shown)


app/components/provider_interface/activity_log_event_component.rb
```diff
@@ -70,6 +72,30 @@ def course_option
                          end
     end
 
+    def event_note_title
+      if application_choice.present?
+        "#{course_option.course.name_and_code} at #{course_option.site.name}"
+      elsif event.audit.auditable.is_a?(Interview)
+        interview = event.audit.auditable
+        title = "Interviewing on #{interview.date_and_time.to_s(:govuk_date_and_time)}"
+        title << ", #{interview.location}" if interview.location.present?
```
despo, line 81: the location is not optional so we can skip the condition here.

app/components/provider_interface/application_timeline_component.rb
```diff
@@ -87,10 +87,32 @@ def change_offer_events
...
     end
 
+    def interview_events
+      @activity_log_events.select { |e| e.audit.auditable_type == 'Interview' }.map do |event|
+        Event.new(
+          interview_title_for(event.audit),
+          actor_for(event),
+          event.created_at,
+          'View interview',
+          provider_interface_application_choice_path(application_choice),
```
despo, line 97: Here we should only link to the interview conditionally. As I'm utilising the discard gem you should be able to check if the interview has been cancelled by using `interview.discarded?`. In that instance there shouldn't be a link.

We can update the path to link to the interviews path later as that code is stil waiting to be reviewed

mnacos (approved):
Looking good! In fact, I think we should probably do something similar for notes in the future. Added a minor naming suggestion but I think this could be merged once links are updated to point to the the newly merged interviews tab.

app/components/provider_interface/activity_log_event_component.rb
```diff
@@ -95,6 +118,11 @@ def link
             url: routes.provider_interface_application_choice_offer_path(application_choice),
             text: 'View offer',
           }
+        elsif event.audit.auditable.is_a?(Interview)
+          {
+            url: routes.provider_interface_application_choice_interviews_path(event.audit.associated, anchor: "interview-#{event.audit.auditable.id}"),
```
steventux, line 123: @despo I've added an anchor using the interview ID for now, though it would be possible to use the interview date and time if this isn't suitable.

spec/components/provider_interface/application_timeline_component_spec.rb
```diff
@@ -118,6 +118,48 @@ def provider_user
...
+      expect(rendered.text).to include 'Interview set up'
+      expect(rendered.text).to include '11 February 2020 at 10:00pm'
+      expect(rendered.css('a').text).to eq 'View interview'
+      expect(rendered.css('a').attr('href').value).to eq "/provider/applications/#{application_choice.id}/interviews#interview-#{interview.id}"
+    end
+
+    it 'renders the interview updated event' do
+      application_choice = create(:application_choice, status: 'interviewing')
+      interview = create(:interview)
+      application_choice_audit = create(:application_choice_audit, application_choice: application_choice)
```
despo, line 138: Can se use `build_stubbed` and method stubs so we can avoid having to create the application_choice in this instance. As we are already testing the audit retrieval as part of the query service specs I think we can skip it here.

spec/factories.rb
```diff
@@ -1214,4 +1224,25 @@
...
+
+  factory :interview_audit, class: 'Audited::Audit' do
+    action { 'create' }
+    user { create(:provider_user) }
+    version { 1 }
+    request_uuid { SecureRandom.uuid }
+    created_at { Time.zone.now }
+
+    transient do
+      application_choice { create(:application_choice, :awaiting_provider_decision) }
```
despo, line 1236: Can we change this to `build` instead of `create`. I think we should in general avoid creating object in the factories, but building them should enable creation anyway if `create` is called.

steventux: @despo https://github.com/DFE-Digital/apply-for-teacher-training/pull/3849/commits/ef88aab93675aa9dfe0557eea80c7894d98c383b was simple enough to stub everything

despo (approved):
💯

PR #3850, opened 2021-01-14
Improve undocumented response-changing

tijmenb:
## Context

The `/respond` page is available to providers even if they've already made a decision. This is a kind of bug (https://trello.com/c/ahz5YKbC/1121-do-not-show-the-respond-page-if-the-application-is-not-awaiting-provider-decision-or-offered) but also a feature! Namely, it allows providers to change their response from a rejection into an offer, as is allowed in our state machine.

This PR does 2 things:

- Adds a test that verifies that everything still works if a provider makes an offer on a rejected application
- Hides the "reject" option from the respond page if it's not possible to reject (for example, when the application is already rejected)
- Hides the "offer" option if an offer can't be made (when the offer was declined for example)

## Guidance to review

Does this make sense.

## Link to Trello card

https://trello.com/c/SwJ1jNhF/118-br8429-change-course-offer

### DIFF
--- a/spec/system/provider_interface/provider_makes_an_offer_on_rejected_application_spec.rb
+++ b/spec/system/provider_interface/provider_makes_an_offer_on_rejected_application_spec.rb
@@ -1,0 +1,117 @@
+require 'rails_helper'
+
+RSpec.feature 'Provider makes an offer' do
+  include CourseOptionHelpers
+  include DfESignInHelpers
+  include ProviderUserPermissionsHelper
+
+  scenario 'Provider makes an offer' do
+    given_i_am_a_provider_user_with_dfe_sign_in
+    and_application_choices_exist_for_my_provider
+    and_i_am_permitted_to_see_applications_for_my_provider
+    and_i_am_permitted_to_make_decisions_for_my_provider
+    and_i_sign_in_to_the_provider_interface
+
+    when_i_visit_a_rejected_application
+    and_i_visit_the_secret_respond_url
+
+    and_i_choose_to_make_an_offer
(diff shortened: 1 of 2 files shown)


spec/system/provider_interface/provider_makes_an_offer_on_rejected_application_spec.rb
```diff
@@ -0,0 +1,117 @@
+require 'rails_helper'
+
+RSpec.feature 'Provider makes an offer' do
+  include CourseOptionHelpers
+  include DfESignInHelpers
+  include ProviderUserPermissionsHelper
+
+  scenario 'Provider makes an offer' do
```
tijmenb, line 8:
```suggestion
  scenario 'Provider makes an offer on a rejected application' do
```

spec/system/provider_interface/provider_makes_an_offer_on_rejected_application_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+    given_i_am_a_provider_user_with_dfe_sign_in
+    and_application_choices_exist_for_my_provider
+    and_i_am_permitted_to_see_applications_for_my_provider
+    and_i_am_permitted_to_make_decisions_for_my_provider
+    and_i_sign_in_to_the_provider_interface
+
+    when_i_visit_a_rejected_application
+    and_i_visit_the_secret_respond_url
```
duncanjbrown, line 16: 😂

duncanjbrown (approved):

PR #3854, opened 2021-01-14
Update clock for ProcessMatchingData

despo:
## Context

Move UCAS file processing forward by an hour to allow for delays
Also ensures that any errors are handled and posted to Raven so we are notified about any future failures

## Trello card

https://trello.com/c/HAkF8bg3/3237-update-clock-for-processmatchingdata

### DIFF
--- a/config/clock.rb
+++ b/config/clock.rb
@@ -6,6 +6,8 @@ require 'clockwork'
 class Clock
   include Clockwork
 
+  error_handler { |error| Raven.capture_exception(error) if defined? Raven }
+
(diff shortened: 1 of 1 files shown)


config/clock.rb
```diff
@@ -6,6 +6,8 @@
 class Clock
   include Clockwork
 
+  error_handler { |error| Raven.capture_exception(error) if defined? Raven }
```
duncanjbrown, line 9: That's nice!

duncanjbrown (approved):

PR #3855, opened 2021-01-14
Prevent candidates from entering an invalid year for o levels

JR-G:
## Context

Users are selecting the wrong type of GCSE qualification. They're able to add an O level with an invalid award year.

If candidate enters a GCE O level, validate if awarded date is any year after 1988.
Screenshot

## Guidance to review

Was 1988 the last year that O levels could be awarded _or_ the first year they couldn't be?

I've implemented it via a nested if statement in the `award_year_is_a_valid_date` method - open to suggestions on this as I'm sure there must be a better way.

## Link to Trello card

https://trello.com/c/0lka2ppD/1113-prevent-users-from-selecting-gce-o-levels-for-years-when-they-werent-available

## Things to check

### DIFF
--- a/spec/forms/candidate_interface/gcse_qualification_details_form_spec.rb
+++ b/spec/forms/candidate_interface/gcse_qualification_details_form_spec.rb
@@ -118,6 +118,24 @@ RSpec.describe CandidateInterface::GcseQualificationDetailsForm, type: :model do
           expect(form.errors[:grade]).to include('Enter a real grade')
         end
       end
+
+      it 'returns validation error if award year is after 1988' do
+        form.award_year = '2012'
+        form.validate(:award_year)
+
+        expect(form.errors[:award_year]).to include('Enter a year before 1989 - GSCEs replaced O levels in 1988')
+      end
+
+      it 'returns no error if award year is valid' do
+        valid_years = (1951..1988)
+
+        valid_years.each do |year|
+          form.award_year = year
+          form.validate(:award_year)
+
+          expect(form.errors[:award_year]).to be_empty
+        end
+      end
     end
 
     context 'when qualification type is Scottish National 5' do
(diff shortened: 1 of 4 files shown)


spec/forms/candidate_interface/gcse_qualification_details_form_spec.rb
```diff
@@ -118,6 +118,24 @@
...
+
+      it 'returns validation error if award year is after 1988' do
+        form.award_year = '2012'
+        form.validate(:award_year)
+
+        expect(form.errors[:award_year]).to include('Enter a year before 1989')
+      end
+
+      it 'returns no error if award year is valid' do
+        valid_years = (1951..1988)
```
davidgisbey, line 130: i had no idea you could do this. I've always used  [*1951..1988]. That's cool to know 👍

davidgisbey:
I'm happy if @paulrobertlloyd is happy.

paulrobertlloyd:
Not sure if we need to say _why_ a date needs to be before 1989? This is probably enough to make you reconsider using this section to enter a GCSE if you’ve done so by mistake. Thoughts @gregknight1?

davidgisbey:
Yea good point. It's probs worth being explicit with this particular error message

gregknight1:
> Not sure if we need to say _why_ a date needs to be before 1989? This is probably enough to make you reconsider using this section to enter a GCSE if you’ve done so by mistake. Thoughts @gregknight1?

I'm fresh in on this so hopefully, I understand. Agree with @paulrobertlloyd - a little more detail would be helpful. So the error message and summary could read:

_Enter a year before 1989 - GSCEs replaced O levels in 1988. _

Also, the hint text needs changing to a year before 1989

paulrobertlloyd:
One last thing ‘O level’ or ‘O Level’ – I think we always use a lowercase l for level (see also: A level)

gregknight1:
> One last thing ‘O level’ or ‘O Level’ – I think we always use a lowercase l for level (see also: A level)

Yep, should always be lower case - O level.

JR-G:
> > One last thing ‘O level’ or ‘O Level’ – I think we always use a lowercase l for level (see also: A level)
> 
> Yep, should always be lower case - O level.

@paulrobertlloyd , @gregknight1 Updated. Please let me know if you're happy from a content point of view

paulrobertlloyd:
🚢

PR #3862, opened 2021-01-18
Shared values should share same localised strings

raamSoftwire:
## Context

There are a lot of strings that are unlikely to change and are unnecessarily duplicated in the strings files.

We analyzed all the strings and counted the number of duplications to find the best candidates for refactoring, you can find the results here, [locales.xlsx](https://github.com/DFE-Digital/apply-for-teacher-training/files/5829583/locales.xlsx). This PR currently refactors
- "Continue"
- "Save and continue" 

I'm not sure how far down that list to go, but this feels like a good start.

## Guidance to review

- Review commit by commit

## Link to Trello card

https://trello.com/c/S1sZ3qdy/366-low-shared-values-should-share-same-localised-strings

## Things to check


george-bit (approved):
LGTM @raamSoftwire, Much neater! (Appreciate the structured commits too, easy to review)

spec/support/test_helpers/candidate_helper.rb
```diff
@@ -151,41 +151,41 @@ def candidate_fills_in_personal_details
...
     check 'British'
     check 'Citizen of a different country'
     within('#candidate-interface-nationalities-form-other-nationality1-field') do
       select 'American'
     end
-    click_button t('complete_form_button', scope: scope)
+    click_button t('application_form.complete_form_button')
 
     check t('application_form.completed_checkbox')
     click_button 'Continue'
```
paulrobertlloyd, line 164: Should this be localised?

paulrobertlloyd:
Two things:

* Should these strings live in a more global location. Currently they exist per localisation
* Have caught any occurrences of unlocalised strings that use these values?

davidgisbey:
Good work Raam. I agree with Paul that something like 'Continue' should live in a top-level file that can be shared across the interfaces. 

This is way neater though and is looking great 👍

spec/support/test_helpers/candidate_helper.rb
raamSoftwire: Done

raamSoftwire:
> Two things:
> 
> * Should these strings live in a more global location. Currently they exist per localisation
> * Have you checked for any occurrences of unlocalised strings that use these same values?

I've done both of these now.

paulrobertlloyd (approved):
❤️ it!

PR #3863, opened 2021-01-18
Refactor and speed up Mailer specs

despo:
## Context

- Extract reusable shared_example and group checks (this significantly reduces the spec running time as it **reduces the mailer tests from 221 to 55** just by grouping the checks) 
- Refactor specs to enable usage of the shared example (and make sure we only use stubbed data)
- Disable rubocop RSpec/LetSetup to make it easier to setup data using let! (This enables using the parameter directly instead of having to for example, go through the application form to access the references `application_form.application_references.first` to ensure that the application_form object is created)
- includes a fix for the resolved_ucas_email content (remove space, display course name and code 

## Before

> Finished in 19.63 seconds (files took 11.98 seconds to load)
> 221 examples, 0 failures

## After

> Finished in 4.29 seconds (files took 11.97 seconds to load)
> 55 examples, 0 failures

### DIFF
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -557,6 +557,14 @@ FactoryBot.define do
       withdrawn_at { Time.zone.now }
     end
 
+    trait :dbd do
+      with_offer
+
+      status { :declined }
+      declined_by_default { true }
+      decline_by_default_days { 10 }
+    end
+
     trait :withdrawn_with_survey_completed do
       status { :withdrawn }
       withdrawn_at { Time.zone.now }
--- a/spec/mailers/referee_mailer_spec.rb
+++ b/spec/mailers/referee_mailer_spec.rb
@@ -1,163 +1,82 @@
 require 'rails_helper'
 RSpec.describe RefereeMailer, type: :mailer do
-  include TestHelpers::MailerSetupHelper
-
   subject(:mailer) { described_class }
 
+  let(:application_form) { build_stubbed(:application_form, first_name: 'Elliot', last_name: 'Alderson') }
+  let(:reference) do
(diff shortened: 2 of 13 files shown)


spec/mailers/referee_mailer_spec.rb
```diff
@@ -1,163 +1,82 @@
 require 'rails_helper'
 RSpec.describe RefereeMailer, type: :mailer do
-  include TestHelpers::MailerSetupHelper
-
   subject(:mailer) { described_class }
 
+  before do
+    allow(reference).to receive(:refresh_feedback_token!).and_return('raw_token')
+  end
```
duncanjbrown, line 7: I'd switch this with the below `let`s just for readability?

spec/support/shared_examples/mailers.rb
```diff
@@ -0,0 +1,10 @@
+RSpec.shared_examples 'a mail with subject and content' do |email_subject, content|
```
duncanjbrown, line 1: :star: great to see this reused across the board

spec/factories.rb
```diff
@@ -557,6 +557,12 @@
       withdrawn_at { Time.zone.now }
     end
 
+    trait :dbd do
+      status { :declined }
+      declined_by_default { true }
+      decline_by_default_days { 10 }
```
duncanjbrown, line 563: We should maybe add offer and offered_at?

duncanjbrown (approved):
:+1: This is great — much less noisy and more readable

spec/factories.rb
despo: amended to use the `with_offer` trait to make sure all prerequisites are satisfied.

PR #3864, opened 2021-01-18
Configure Redis for parallel tests

malcolmbaig:
## Context

A prior commit dealing with this was reverted as part of an investigation into performance issues in production. Although seemingly unrelated to the issue, it did prompt a review of how Redis was being configured. This PR takes a simpler approach with a smaller footprint - the changes only affect parallel test runs, and no other form of Redis usage should be impacted.

This PR once again adds support for each parallel process to use a
different Redis database.

- Rather than configuring connections directly, we continue to rely on
  the REDIS_URL set in a given environment.
- Use ClimateControl to stub the value of REDIS_URL when running tests
  in parallel, ensuring we have uniqueness without affecting the
  configuration of anything else in the codebase.

In addition, this PR updates .env.example and .env.development to
encourage the convention of using Redis database 0 in development,
leaving databases 1 through n available for the test environment.

## Guidance to review
- Run parallel tests locally, should work without issue if you've run them before (see the doc if trying for the first time).
- Is the config update clear enough?

## Things to check

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -1,3 +1,4 @@
+REDIS_URL=redis://localhost:6379/0
 FIND_BASE_URL=https://api.publish-teacher-training-courses.service.gov.uk/api/v3/
 TEACHER_TRAINING_API_BASE_URL=https://api.publish-teacher-training-courses.service.gov.uk/api/public/v1
 HOSTING_ENVIRONMENT_NAME=development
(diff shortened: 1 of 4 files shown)


duncanjbrown (approved):
:+1: It's so good to see it become radically simpler second time around

PR #3865, opened 2021-01-18
Remove Received At column from References table and refactor

george-bit:
## Context

After review of the code base it was found that concise methods that can
used to call this information from the audit log are present on the application
reference model. 

This PR removes the column and refactors export files touching
feedback provided at.

## Guidance to review

Check logic is sound

## Link to Trello card

https://trello.com/c/LLEPTmu1/2854-add-receivedat-to-the-applicationreference-model

## Things to check

### DIFF
--- a/app/services/support_interface/candidate_journey_tracker.rb
+++ b/app/services/support_interface/candidate_journey_tracker.rb
@@ -155,7 +155,7 @@ module SupportInterface
       @received_reference_timings ||=
         received_references.map do |reference|
           {
-            received_at: earliest_update_audit_for(reference, feedback_status: 'feedback_provided'),
+            received_at: reference.feedback_provided_at,
             requested_at: reference.requested_at,
           }
         end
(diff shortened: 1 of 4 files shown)


app/services/support_interface/candidate_journey_tracker.rb
```diff
@@ -155,7 +155,7 @@ def received_references_timings
       @received_reference_timings ||=
         received_references.map do |reference|
           {
-            received_at: earliest_update_audit_for(reference, feedback_status: 'feedback_provided'),
+            received_at: reference.feedback_provided_at,
```
stevehook, line 158: :+1: together with the extra `includes` this is likely to be a bit faster (because it avoids a 1+n query). It could lead to a higher memory usage.

stevehook (approved):
I still think the `received_at` column would be useful as an explicit column. The helper methods on `ApplicationReference` are very handy but not as efficient (and not always so convenient) as just putting a column into a `SQL WHERE`. The column would have really helped for the feature metrics though that's done now.

davidgisbey:
I'll do an audit of all the columns using those helper methods and see if moving them all over to individual columns and backfilling them is worth it. 

This is quick win for now anyway 👍

davidgisbey (approved):

PR #3869, opened 2021-01-18
Fix feature metrics crash

stevehook:
## Context

There are a couple of bad assumptions about data in the feature metrics query logic.

Handle the following cases:
- [x] `ApplicationReference#requested_at` can be `nil` even if the `feedback_status` is `feedback_provided`.
- [x] `ApplicationForm#work_history_completed` can be `true` with no work history items (presumably this is valid if a person has never worked).

## Guidance to review

- Per commit (just two of them)
- Does it make sense to simply discard the figures that have nil values from the calculation of averages? e.g. if there is no work history is it reasonable to ignore such applications rather than count them as having a zero time to complete?
- Anything else missing?

## Link to Trello card

https://trello.com/c/sdK6C9kB/2894-feature-metrics-crashing-on-production

## Things to check

### DIFF
--- a/app/models/reference_feature_metrics.rb
+++ b/app/models/reference_feature_metrics.rb
@@ -55,8 +55,10 @@ private
       provided_audit = reference.audits.where("audited_changes#>>'{feedback_status, 1}' = 'feedback_provided'").last
       [reference.requested_at, provided_audit&.created_at]
     end
-    requested_at_times = times.map(&:first)
-    provided_at_times = times.map(&:second)
+    requested_at_times = times.map(&:first).compact
+    provided_at_times = times.map(&:second).compact
+    return nil if requested_at_times.blank? || provided_at_times.blank?
+
     (provided_at_times.min - requested_at_times.max).to_i / 1.day
   end
 end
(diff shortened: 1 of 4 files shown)


davidgisbey (approved):
Makes sense Steve. I think this is something we're going to run into from time to time as some columns which are updated in services. I'm wondering if, on some occasions, people have forgotten to update them when working in the console. 

We could potentially add some hooks to the model which ensure the columns are updated when the state enum is changed. So received at when feedback_requested -> feedback_provided, but it might be a bit much.

stevehook:
> Makes sense Steve. I think this is something we're going to run into from time to time as some columns which are updated in services. I'm wondering if, on some occasions, people have forgotten to update them when working in the console.
> 
> We could potentially add some hooks to the model which ensure the columns are updated when the feedback_status enum is changed. So received at when feedback_requested -> feedback_provided, but it might be a bit much.
> 
> What do you think?

thanks @davidgisbey, I think generally if you use the services as they should be there are no nasty surprises. Trouble is we have some data that pre-dates those services and like you say we may need to make a data change for support purposes. I'm in two minds about enforcing these with model level validations or callbacks. Something we may want to discuss with the wider team.

PR #3870, opened 2021-01-19
Prevent LocalJump error in config/clock.rb

mnacos:
## Context

Using `return` in blocks within the clockwork config results in `LocalJump Error: unexpected return`.

Use `if` rather than `return`.

## Guidance to review

Easy peasy.

## Link to Trello card

General maintenance

## Things to check


despo (approved):
👍🏼

PR #3871, opened 2021-01-19
Optimise ucas_matched_application spec

despo:
## Context

Refactors and optimises ucas_matched_application spec. 
This clashes with @malcolmbaig 's [speed up UCAS PR](https://github.com/DFE-Digital/apply-for-teacher-training/pull/3836) but I wanted to see if we could easily improve the specs without having to require an external library by refactoring the data setup.

I think it's a slightly slower than Malcom's PR, however the data setup is a lot clearer and focused and I was able to get away with using `build_stubbed` across the spec with only a couple of exceptions where we were testing the retrieval of application_choice and provider.

_Test run comparison_

> master
Finished in 10.96 seconds (files took 21.9 seconds to load)
26 examples, 0 failures

> 2832-speed-up-ucas-spec
Finished in 1.02 seconds (files took 22.06 seconds to load)
26 examples, 0 failures

**current PR**
>optimise-ucas_matched_application-spec 
Finished in 1.11 seconds (files took 13.38 seconds to load)
26 examples, 0 failures

### In addition to that

- fixes spec referencing wrong method (application_accepted_on_apply)
- change `find_by` to `exists?` in ucas_matched_application as object is unnecessarily retrieved from the database.

## Things to check

### DIFF
--- a/spec/models/ucas_matched_application_spec.rb
+++ b/spec/models/ucas_matched_application_spec.rb
@@ -1,95 +1,99 @@
 require 'rails_helper'
 
 RSpec.describe UCASMatchedApplication do
-  let(:course) { create(:course, recruitment_cycle_year: 2020) }
-  let(:candidate) { create(:candidate) }
-  let(:course_option) { create(:course_option, course: course) }
-  let(:application_choice) { create(:application_choice, course_option: course_option) }
-  let(:application_form) { create(:completed_application_form, candidate_id: candidate.id, application_choices: [application_choice]) }
-  let(:apply_again_application_form) { create(:application_form, candidate_id: candidate.id) }
   let(:recruitment_cycle_year) { 2020 }
-  let(:candidate1) { create(:candidate) }
-  let(:application_choice1) { create(:application_choice, :with_accepted_offer, course_option: course_option) }
-  let(:application_form1) { create(:completed_application_form, candidate_id: candidate1.id, application_choices: [application_choice1]) }
-  let(:candidate2) { create(:candidate) }
-  let(:application_choice2) { create(:application_choice, :with_rejection, course_option: course_option) }
-  let(:application_form2) { create(:completed_application_form, candidate_id: candidate2.id, application_choices: [application_choice2]) }
-
-  before do
-    apply_again_application_form
-    application_form
-    application_form1
-    application_form2
-    create(:course, code: course.code, provider: course.provider, recruitment_cycle_year: 2021)
-  end
+  let(:candidate) { build_stubbed(:candidate) }
+  let(:provider) { build_stubbed(:provider) }
+  let(:course) { build_stubbed(:course, recruitment_cycle_year: recruitment_cycle_year, provider: provider) }
+  let(:course_option) { build_stubbed(:course_option, course: course) }
 
   describe '#course' do
+    let(:course) { create(:course, recruitment_cycle_year: 2020) }
+    let(:ucas_matching_data) do
+      {
+        'Course code' => course.code.to_s,
+        'Provider code' => course.provider.code.to_s,
+        'Apply candidate ID' => candidate.id.to_s,
+      }
+    end
+
     it 'returns the course for the correct recruitment cycle' do
-      ucas_matching_data =
-        { 'Course code' => course.code.to_s,
-          'Provider code' => course.provider.code.to_s,
-          'Apply candidate ID' => candidate.id.to_s }
-      ucas_matching_application = UCASMatchedApplication.new(ucas_matching_data, recruitment_cycle_year)
+      ucas_matching_application = UCASMatchedApplication.new(ucas_matching_data, course.recruitment_cycle_year)
 
       expect(ucas_matching_application.course).to eq(course)
     end
 
-    it 'returns the course details for a course which is not on Apply' do
-      ucas_matching_data =
-        { 'Scheme' => 'U',
+    context 'when a course is not on Apply' do
+      let(:provider) { create(:provider) }
+      let(:ucas_matching_data) do
+        {
           'Course code' => '123',
           'Course name' => 'Not on Apply',
           'Provider code' => course.provider.code.to_s,
-          'Apply candidate ID' => candidate.id.to_s }
-      ucas_matching_application = UCASMatchedApplication.new(ucas_matching_data, recruitment_cycle_year)
+        }
+      end
 
-      expect(ucas_matching_application.course.code).to eq('123')
-      expect(ucas_matching_application.course.name).to eq('Not on Apply')
-      expect(ucas_matching_application.course.provider).to eq(course.provider)
-    end
+      it 'returns the correct course details' do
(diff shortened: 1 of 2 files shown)


spec/models/ucas_matched_application_spec.rb
```diff
@@ -1,77 +1,70 @@
...
 
-    it 'returns the course details for a course which is not on Apply' do
-      ucas_matching_data =
-        { 'Scheme' => 'U',
+    context 'when a course is not on Apply' do
+      it 'returns the correct course details' do
+        allow(Provider).to receive(:find_by).with(code: course.provider.code.to_s).and_return(course.provider)
+
+        ucas_matching_data = {
+          'Scheme' => 'U',
```
malcolmbaig, line 28: The scheme isn't relevant to the behaviour of this method, so potentially serves to confuse if included here? Might be worth omitting.

spec/models/ucas_matched_application_spec.rb
```diff
@@ -1,77 +1,70 @@
...
-      ucas_matching_application = UCASMatchedApplication.new(ucas_matching_data, recruitment_cycle_year)
+      ucas_matching_application = UCASMatchedApplication.new(ucas_matching_data, course.recruitment_cycle_year)
 
       expect(ucas_matching_application.course).to eq(course)
     end
 
-    it 'returns the course details for a course which is not on Apply' do
-      ucas_matching_data =
-        { 'Scheme' => 'U',
+    context 'when a course is not on Apply' do
```
malcolmbaig, line 23: Suggestion - hoisting the `ucas_matching_data` up here in a let makes it a bit clearer what's changed in this context.

spec/models/ucas_matched_application_spec.rb
```diff
@@ -1,77 +1,70 @@
...
+      it 'returns the correct course details' do
+        allow(Provider).to receive(:find_by).with(code: course.provider.code.to_s).and_return(course.provider)
```
malcolmbaig, line 25: Suggestion—as we're creating db records anyway, can remove this for clarity in exchange for a minor performance hit.

spec/models/ucas_matched_application_spec.rb
```diff
@@ -1,77 +1,70 @@
...
+      let(:application_choice) { build_stubbed(:application_choice, course_option: course_option) }
+
       it 'returns the applications status' do
         dfe_matching_data =
           { 'Scheme' => 'D',
             'Course code' => course.code.to_s,
             'Provider code' => course.provider.code.to_s,
             'Apply candidate ID' => candidate.id.to_s }
         ucas_matching_application = UCASMatchedApplication.new(dfe_matching_data, recruitment_cycle_year)
+        allow(ucas_matching_application).to receive(:application_choice).and_return(application_choice)
```
malcolmbaig, line 67: This looks like a hint that `#application_choice` should probably live somewhere else.

Rather than stub a method on the object under test, we could:

- Move the implementation of the `#application_choice` method to something appropriately named on the ApplicationChoice model.
- Stub that new method instead.
- Replace the testing we're doing for `#application_choice` in this spec file with something equivalent in `application_choice_spec.rb`.

spec/models/ucas_matched_application_spec.rb
```diff
@@ -114,58 +109,63 @@
...
-          'Applicant withdrawn from scheme while offer awaiting applicant reply' => '.',
-          'Applicant withdrawn from scheme after firmly accepting a conditional offer' => '.',
-          'Applicant withdrawn from scheme after firmly accepting an unconditional offer' => '.' }
-      ucas_matching_application = UCASMatchedApplication.new(ucas_matching_data, recruitment_cycle_year)
+        expect(ucas_matching_application.application_in_progress_on_ucas?).to eq(true)
+      end
 
-      expect(ucas_matching_application.application_in_progress_on_ucas?).to eq(false)
-    end
+      it 'returns false if application is in unsucesfull state on UCAS' do
```
malcolmbaig, line 143: unsucesfull -> unsuccessful

spec/models/ucas_matched_application_spec.rb
```diff
@@ -114,58 +109,63 @@
...
+        ucas_matching_application = UCASMatchedApplication.new(ucas_matching_data, recruitment_cycle_year)
 
-    it 'returns false the application is DfE scheme' do
-      ucas_matching_data = { 'Scheme' => 'D' }
-      ucas_matching_application = UCASMatchedApplication.new(ucas_matching_data, recruitment_cycle_year)
+        expect(ucas_matching_application.application_in_progress_on_ucas?).to eq(false)
+      end
 
-      expect(ucas_matching_application.application_in_progress_on_ucas?).to eq(false)
+      it 'returns false the application is DfE scheme' do
```
malcolmbaig, line 163: `if the application..`

spec/models/ucas_matched_application_spec.rb
```diff
@@ -262,26 +276,36 @@
...
 
   describe '#application_accepted_on_apply?' do
-    it 'returns true if application is accepted on Apply' do
-      ucas_matching_data =
-        { 'Scheme' => 'D',
-          'Course code' => course.code.to_s,
-          'Provider code' => course.provider.code.to_s,
-          'Apply candidate ID' => candidate1.id.to_s }
-      ucas_matching_application = UCASMatchedApplication.new(ucas_matching_data, recruitment_cycle_year)
+    describe 'accepted' do
```
malcolmbaig, line 279: Could these be context blocks instead, for consistency? Also, could do something like the following and simplify the it block descriptions:

```
context 'when the application is accepted on Apply' do
  it 'returns true'
...
```

malcolmbaig:
Some comments left inline, but overall some really nice improvements here 👍🏽

spec/models/ucas_matched_application_spec.rb
```diff
@@ -114,58 +109,63 @@
...
-          'Applicant withdrawn from scheme while offer awaiting applicant reply' => '.',
-          'Applicant withdrawn from scheme after firmly accepting a conditional offer' => '.',
-          'Applicant withdrawn from scheme after firmly accepting an unconditional offer' => '.' }
-      ucas_matching_application = UCASMatchedApplication.new(ucas_matching_data, recruitment_cycle_year)
+    context 'when there is a provider' do
+      let(:provider) { create(:provider) }
 
-      expect(ucas_matching_application.application_in_progress_on_ucas?).to eq(true)
-    end
+      it 'returns true if application is not in an unsucesfull state on UCAS' do
```
steventux, line 123: `s/unsucesfull/unsuccessful/`

steventux (approved):
:+1: Great optimisating!

AgaDufrat (approved):
This sparks joy ✨

spec/models/ucas_matched_application_spec.rb
despo: As the intention of this PR is to improve optimise the spec this is something we should address separately so I will add a card for it.

malcolmbaig (approved):
🚤

PR #3872, opened 2021-01-19
SKYLIGHT_ENABLE as string parameter

vigneshmsft:
## Context

container instance is setting boolean param as `True` or `False`
and rails ENV is loading SKYLIGHT_ENABLE as string
requiring case insensitive compare in application code.

change SKYLIGHT_ENABLE as string parameter

## Things to check


mnacos (approved):
Thanks Vignesh 🚀

PR #3874, opened 2021-01-19
Add request_method and response_headers columns to vendor_api_requests

steventux:
## Context

We've seen a number of 302 responses in the API requests logs and we're not sure of their origin.
We'd also like to know the request method used.

Adds `request_method` and `response_headers` columns to `vendor_api_requests` table.
These will be used by UI filters.

## Link to Trello card

https://trello.com/c/LBHHTB9x/3248-single-api-request-to-qa-produces-three-vendorapirequest-entries
https://trello.com/c/Qyzx7zmE/3266-filter-vendor-api-requests-by-get-and-post-requests-in-support

## Things to check


db/migrate/20210119154820_add_request_method_and_response_headers_to_vendor_api_requests.rb
```diff
@@ -0,0 +1,6 @@
+class AddRequestMethodAndResponseHeadersToVendorAPIRequests < ActiveRecord::Migration[6.0]
+  def change
+    add_column :vendor_api_requests, :request_method, :string
```
mnacos, line 3: May be worth back-filling this information, should be possible via the request path.

steventux: @mnacos I've attempted this here https://github.com/DFE-Digital/apply-for-teacher-training/pull/3874/commits/97f5e75f237dbd2b5f449fe7de6cad4a861ef8cc#diff-5f48943dcb8a0ea9c0bbc153df01f259d328b5747eb11eae7b75627abb724c64R7

db/migrate/20210119154820_add_request_method_and_response_headers_to_vendor_api_requests.rb
```diff
@@ -0,0 +1,16 @@
...
+    add_column :vendor_api_requests, :response_headers, :jsonb
+
+    VendorAPIRequest.where(request_method: nil).find_each do |vendor_api_request|
+      if vendor_api_request.request_path =~ /\/(offer|conditions-not-met|confirm-conditions-met|reject|confirm-enrolment|regenerate|generate|clear)$/
```
duncanjbrown, line 7: A client could have made a `GET` to one of these endpoints in error. I'd almost rather have no data than unreliable data?

steventux: @duncanjbrown in this case I think we'd serve a 404, we could omit any vendor api requests with 404 status from update?

duncanjbrown: :+1: Good idea!

duncanjbrown (approved):

PR #3877, opened 2021-01-19
Show offered_at and DBD dates in support

duncanjbrown:
## Context

These dates are useful but missing in support.

Screenshot 2021-01-19 at 20 58 38

Also fix a bug where we rendered unstructured reasons for rejection twice.

## Guidance to review

Should be straightforward 🤞

### DIFF
--- a/app/components/support_interface/application_choice_component.rb
+++ b/app/components/support_interface/application_choice_component.rb
@@ -13,6 +13,11 @@ module SupportInterface
         { key: 'Status', value: render(SupportInterface::ApplicationStatusTagComponent.new(status: application_choice.status)) },
       ]
 
+      if application_choice.offer?
+        rows << { key: 'Offer made at', value: application_choice.offered_at.to_s(:govuk_date_and_time) }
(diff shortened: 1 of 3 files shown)


app/components/support_interface/application_choice_component.rb
```diff
@@ -33,7 +38,6 @@ def rows
         end
       end
 
-      rows << { key: 'Feedback', value: application_choice.rejection_reason } if application_choice.rejection_reason.present?
```
steventux, line 36: Assume we are removing this as it's duplicated.

steventux (approved):
:+1:

app/components/support_interface/application_choice_component.rb
duncanjbrown: Yes — the other field falls back correctly: https://github.com/DFE-Digital/apply-for-teacher-training/pull/3877/commits/d61fd89050ba4114e3eb1e2cd8c74db4ccbf5c3d

PR #3879, opened 2021-01-20
Fix state-mixing bug in "Other qualifications" section

tijmenb:
## Context

The "Other qualifications" section uses our wizardry functionality. This means that we have multiple pages with forms, and save the candidate input into an "intermediate data service" first. Once the candidate is finished with all the pages we persist the data to the database. This allows us to ask the user for less information per page but avoid invalid and inconsistent data in the database.

The initial implementation of this feature saved the data provided by the user into a singular Redis field
(`candidate_user_other_qualification_flow-#{current_candidate.id}`). This means that every time we load a qualification form page, we have to clear the state to make sure that if we visit a different form page the wrong data isn't displayed. This is very hard to do, and we currently are seeing the following bug:

- Create a application with 2 "Other Qualifications"
- Click on "Change" for the first qualification
- Use the back button in your browser
- Click on "Change" for the second qualification
- The form is erroneously populated with the data for the first qualification, not the second

Previously we fixed this issue by adding more state clearing: https://github.com/DFE-Digital/apply-for-teacher-training/pull/3613.

This PR changes the approach, to avoid the state clearing in most cases (this is proof of the lesson is that cache invalidation is hard so best avoided). By using key per record, it's not possible to share data between the forms.

The only place where we need to explicitly clear the state is when the candidate visits the "Add other qualification" page (the one that lets you choose the type). Without the reset, the second time a candidate clicks the "Add" button they'll see the data they've entered the first time. I've considered avoiding this by generating an ID every time you visit the page, persisting this in the session, and using this as the key suffix instead of "new". I suspect that's more trouble than it's
worth though.

## Guidance to review

- What to do with the [code added to fail gracefully when clicking "back"](https://github.com/DFE-Digital/apply-for-teacher-training/commit/36980444ec15e4d51c330095fbe2e85fe08e972e)? This doesn't fail in the same way anymore, because we haven't cleared the state. The specs are now commented out. @stevehook 
- Have I missed anything in the flow that would make this not work? Does not clearing state break anything?
- Where in the spec should I test for this? The file [already has 100 lines of steps](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/spec/system/candidate_interface/entering_details/candidate_entering_other_qualification_spec.rb).

## Link to Trello card

https://trello.com/c/P5pagu0l

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/other_qualifications/base_controller.rb
+++ b/app/controllers/candidate_interface/other_qualifications/base_controller.rb
@@ -9,20 +9,12 @@ module CandidateInterface
 
   private
 
-    def reset_intermediate_state!
-      intermediate_data_service.clear_state!
-    end
-
     def intermediate_data_service
       @intermediate_data_service ||= IntermediateDataService.new(
         WizardStateStores::RedisStore.new(
-          key: persistence_key_for_current_user,
+          key: "candidate_user_other_qualification_flow-#{current_candidate.id}-#{params[:id] || 'new'}",
         ),
       )
-    end
-
-    def persistence_key_for_current_user
-      "candidate_user_other_qualification_flow-#{current_candidate.id}"
     end
   end
 end
--- a/app/controllers/candidate_interface/other_qualifications/details_controller.rb
+++ b/app/controllers/candidate_interface/other_qualifications/details_controller.rb
@@ -7,11 +7,14 @@ module CandidateInterface
         current_step: :details,
       )
 
-      @form.qualification_type ||= params[:qualification_type]
+      unless @form.qualification_type
+        @form.qualification_type = params[:qualification_type]
+        @form.save_intermediate!
+      end
+
       @form.initialize_from_last_qualification(
         current_application.application_qualifications.other.order(:created_at),
       )
-      @form.save_intermediate!
     end
 
     def create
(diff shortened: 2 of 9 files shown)


stevehook (commented):
I think this is a good change. We should have been scoping the keys like this from the outset.

I had one issue when I tried it locally where the subject wasn't getting populated when editing an existing qualification, steps to reproduce:

1. Add 3 A levels.
2. From the review page use the _Change_ link to edit the grade of one of the A levels and save changes.
3. Again hit the same _Change_ link then use the back button.

I am having trouble reproducing this consistently but editing and using the back button is somehow causing the stored state to be messed up somehow.

duncanjbrown:
Argh, I'm having trouble understanding this code, as we seem to persist things as soon as we submit the form (I'm looking at `DetailsController#create` and `DetailsController#update`). All the possible bits of data entry seem to be single-screen. What value is the wizard adding here?

tijmenb:
That's a good question. It's a 2 page flow currently (type => details). Perhaps the plan was to add even more pages to the flow? In that case it's good to have the wizardry set up already, so that adding things is easier.

duncanjbrown:
> Yes it's just a two page wizard. The original problem that it was trying to solve was that we were seeing incomplete qualifications created, with just a type. These were picked up later during pre-submit validation but catching the issue early is a better UX.

Could it be simplified by having the radios GET rather than POST — ie `qualifications/new?type=a_level` and then only persist on posting that form?

stevehook (approved):

PR #3886, opened 2021-01-21
Feature metric dashboard - fixes reference stats

stevehook:
## Context

When deployed to production we noticed that there were a few oddities about the reference feature metrics. The numbers were too low.

Two issues are addressed by this PR:

- [x] There was a mistake in the logic, we compared the maximum reference requested at date with the minimum received at date for the two provided references. This should have been minimum reference requested at date and the maximum received
at date. This mistake was compounded by some rounding errors. This was a straightforward fix.
- [x] The query that finds applications with references within a given date range was only including those for which both references were received in that date range. It should be considering applications for which only the second or both references fall within the data range.

## Guidance to review

- Per commit
- I don't think the fix for the second issue listed above is good from a maintainability point of view. I couldn't see a way to write the logic for the date range in a single database query. Instead I've added some post query filtering to ensure that only the correct applications are counted. This would be much easier given an `ApplicationForm#references_provided_at` timestamp. I wasn't 100% convinced before but I think we really need this now.

## Link to Trello card

https://trello.com/c/CBmwCoda/2770-feature-metric-dashboard

## Things to check

### DIFF
--- a/app/models/reference_feature_metrics.rb
+++ b/app/models/reference_feature_metrics.rb
@@ -59,6 +59,6 @@ private
     provided_at_times = times.map(&:second).compact
     return nil if requested_at_times.blank? || provided_at_times.blank?
 
-    (provided_at_times.min - requested_at_times.max).to_i / 1.day
+    (provided_at_times.max - requested_at_times.min) / 1.day
   end
 end
--- a/app/models/reference_feature_metrics.rb
+++ b/app/models/reference_feature_metrics.rb
@@ -46,19 +46,20 @@ private
           .where('created_at BETWEEN ? AND ?', start_time, end_time),
       )
       .group('application_forms.id')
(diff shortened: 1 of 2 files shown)


app/models/reference_feature_metrics.rb
```diff
@@ -59,6 +59,6 @@ def time_to_get_for(application)
     provided_at_times = times.map(&:second).compact
     return nil if requested_at_times.blank? || provided_at_times.blank?
 
-    (provided_at_times.min - requested_at_times.max).to_i / 1.day
+    (provided_at_times.max - requested_at_times.min) / 1.day
```
davidgisbey: This is 100% on me. I noticed this when i reviewed the last PR and then got distracted before i left a comment. Sorry about that!

davidgisbey (approved):
Looks good Steve

PR #3887, opened 2021-01-21
Prevent users from submitting gcses as completed with an empty award year

JR-G:
## Context

Users are able to add a qualification in the 'GCSE or equivalent' section and mark it as completed without entering a value in the 'Year awarded' or 'Grade' section.

For consistency with the incomplete degree logic, a flash message will now appear on the qualification review page, and prevent the session being redirected back to the application page, if they try to mark the qualification as complete and press continue:

|Before|After|
|------------|------------|
|Before|After|

It will also present a message on the application review page informing them it is incomplete, with a link back to the qualification review page:
|Before|After|
|------------|------------|
|Before |After |

## Guidance to review

Considering this is an edge case, is it acceptable for these messages to be generic rather than specifying which fields are incomplete?

## Link to Trello card

https://trello.com/c/iJKKl9KS

## Things to check

### DIFF
--- a/app/components/candidate_interface/gcse_qualification_review_component.rb
+++ b/app/components/candidate_interface/gcse_qualification_review_component.rb
@@ -33,6 +33,13 @@ module CandidateInterface
       end
     end
 
+    def show_values_missing_banner?
+      if @submitting_application && @application_qualification.grade.nil? || @submitting_application && @application_qualification.award_year.nil?
+        gcse_completed = "#{@subject}_gcse_completed"
+        @application_form.send(gcse_completed) && @editable
+      end
+    end
+
   private
 
     attr_reader :application_qualification, :subject
--- a/app/components/candidate_interface/gcse_qualification_review_component.rb
+++ b/app/components/candidate_interface/gcse_qualification_review_component.rb
@@ -34,15 +34,19 @@ module CandidateInterface
     end
 
     def show_values_missing_banner?
-      if @submitting_application && @application_qualification.grade.nil? || @submitting_application && @application_qualification.award_year.nil?
-        gcse_completed = "#{@subject}_gcse_completed"
-        @application_form.send(gcse_completed) && @editable
+      if @submitting_application
+        section_or_gcse_incomplete? && @editable
       end
     end
 
   private
 
     attr_reader :application_qualification, :subject
+
+    def section_or_gcse_incomplete?
+      gcse_completed = "#{@subject}_gcse_completed"
+      (!@application_form.send(gcse_completed) || @application_form.send("#{@subject}_gcse").incomplete_gcse_information?) && !@application_qualification.missing_qualification?
+    end
 
     def qualification_row
       {
--- a/app/models/application_qualification.rb
+++ b/app/models/application_qualification.rb
@@ -8,6 +8,12 @@ class ApplicationQualification < ApplicationRecord
     award_year
   ].freeze
 
+  EXPECTED_GCSE_DATA = %i[
+    qualification_type
+    subject
+    award_year
+  ].freeze
+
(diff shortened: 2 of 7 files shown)


gregknight1 (commented):
Not sure what the numbers are for this edge case but it's preferable for us to be specific about the fields that need completing. Might be not worth it if it rarely happens?

davidgisbey:
Hey Greg, i think it's a good point that this could be more granular. I've been thinking about a solution that we can use throughout the sections and am going to add a card to the tech debt column for the devs to discuss at the next tech debt meeting.

app/components/candidate_interface/gcse_qualification_review_component.rb
```diff
@@ -33,6 +33,13 @@ def show_missing_banner?
       end
     end
 
+    def show_values_missing_banner?
+      if @submitting_application
+        gcse_completed = "#{@subject}_gcse_completed"
+        (!@application_form.send(gcse_completed) || @application_form.send("#{@subject}_gcse").incomplete_gcse_information?) && @editable && !@application_qualification.missing_qualification?
```
davidgisbey, line 39: Looking at this with fresh eyes. Can we pull this into a separate method so it's a bit clearer what is going on?

We can put this block

`(!@application_form.send(gcse_completed) || @application_form.send("#{@subject}_gcse").incomplete_gcse_information?)`

into a private method called `section_or_gcse_is_incomplete?` or something like that

app/models/application_qualification.rb
```diff
@@ -65,6 +65,13 @@ def incomplete_degree_information?
     false
   end
 
+  def incomplete_gcse_information?
+    return true if award_year.nil?
+    return true if grade.nil? && structured_grades.nil?
```
davidgisbey, line 70: I think we can just change combine these two lines, it's pretty clear what's going on.

So `return true if award_year.nil? || ( grade.nil? && structured_grades.nil?)`

davidgisbey: I think we can also use the established pattern here of creating a constant 

EXPECTED_GCSE_DATA = %i[
    qualification_type
    subject
    award_year
  ].freeze

and then doing something like

`
  def incomplete_gcse_information?
     return true if grade.nil? && structured_grades.nil?
     return true if EXPECTED_GCSE_DATA.any? do |field_name|
      send(field_name).blank?
    end

    false
  end
`

app/components/candidate_interface/gcse_qualification_review_component.rb
davidgisbey: On second thought we can also add this here too `!@application_qualification.missing_qualification?` as the method name covers the behaviour.

So 

```
private

def section_or_gcse_is_incomplete?
  (!@application_form.send(gcse_completed) || 
    @application_form.send("#{@subject}_gcse").incomplete_gcse_information?) && 
    !@application_qualification.missing_qualification?
end 

```

gregknight1:
> Hey Greg, i think it's a good point that this could be more granular. I've been thinking about a solution that we can use throughout the sections and am going to add a card to the tech debt column for the devs to discuss at the next tech debt meeting.

Cool, that sounds like a plan @davidgisbey. Think we could probably improve the wording a bit too but will hold off until we know the way forward. Thanks.

app/components/candidate_interface/gcse_qualification_review_component.rb
```diff
@@ -33,10 +33,21 @@ def show_missing_banner?
       end
     end
 
+    def show_values_missing_banner?
+      if @submitting_application
+        section_or_gcse_incomplete?
```
davidgisbey, line 38: can we move @editable to this line please?

davidgisbey (approved):
LGTM

PR #3889, opened 2021-01-21
Spike out the new work history flow

davidgisbey:
## Context

This is an initial stab at how we're going to implement the new Work History flow.

I'll add some comments throughout to document my thought process.

- Add a new bool to the application form that essentially functions as a feature flag/redirection tool
- Backfill existing applications to ensure they don't use the new flow
- Add in the basic plumbing for redirecting to the new flow

## Guidance to review

1. We can establish a pattern going forward for redesigns on an existing section by prefixing an attr with 'feature_' then the feature flags name
2. It makes it easier to determine what state the application is in and can be used throughout the application
3. We can add scopes easily

Does this actually work? Have I missed something?

## Link to Trello card

https://trello.com/c/ULGeU2Pp/2907-work-history-spike-how-were-going-to-integrate-and-move-users-through-the-new-and-old-flows

## Things to check


db/schema.rb
```diff
@@ -145,6 +145,7 @@
     t.text "feedback_suggestions"
     t.float "latitude"
     t.float "longitude"
+    t.boolean "feature_restructured_work_history", default: true
```
davidgisbey, line 148: Defaulting this to true means that all new application forms will by default use the new flow. New applications being redirected to the old flow prior to the flag being switched on is dealt with in a later commit.

app/controllers/candidate_interface/work_history/length_controller.rb
```diff
@@ -10,6 +10,8 @@ def submit
       @work_details_form = WorkHistoryForm.new(work_history_form_params)
 
       if @work_details_form.valid?
+        current_application.update!(feature_restructured_work_history: false)
```
davidgisbey, line 13: Prior to the flag being switched on, all candidates will be redirected to the existing flow. This allows us to ensure applicants that add jobs post between the feature_restructured_work_history column being deployed and us switching the flag use the old flow.

config/routes.rb
```diff
@@ -240,6 +240,11 @@
         delete '/delete/:id' => 'work_history/destroy#destroy'
       end
 
+      scope '/restructured-work-history' do
+        get '/' => 'restructured_work_history/base#show', as: :restructured_work_history
+        get '/' => 'restructured_work_history/review#show', as: :restructured_work_history_review
```
raamSoftwire, line 245: Should this be `/review`?

app/controllers/candidate_interface/restructured_work_history/base_controller.rb
```diff
@@ -0,0 +1,17 @@
+module CandidateInterface
+  class RestructuredWorkHistory::BaseController < CandidateInterfaceController
+    before_action :redirect_to_dashboard_if_submitted, :render_404_if_candidate_has_used_existing_flow_or_flag_is_active
+
+    def show; end
+
+  private
+
+    def render_404_if_candidate_has_used_existing_flow_or_flag_is_active
+      render_404 if candidate_has_used_existing_flow || FeatureFlag.active?(:restructured_work_history)
```
malcolmbaig, line 10: Should this be `!FeatureFlag.active?` ?

app/presenters/candidate_interface/application_form_presenter.rb
```diff
@@ -145,7 +145,13 @@ def references_path
     end
 
     def work_experience_path
-      if @application_form.application_work_experiences.any? || @application_form.work_history_explanation.present?
+      if @application_form.feature_restructured_work_history && FeatureFlag.active?(:restructured_work_history)
+        if @application_form.application_work_experiences.any? || @application_form.work_history_explanation.present?
+          Rails.application.routes.url_helpers.candidate_interface_restructured_work_history_path
+        else
+          Rails.application.routes.url_helpers.candidate_interface_restructured_work_history_review_path
```
malcolmbaig, line 152: Should these branches be swapped? As in, if you've got any work experience inputted, go to the review. Otherwise, go  add some.

app/controllers/candidate_interface/restructured_work_history/base_controller.rb
```diff
@@ -0,0 +1,17 @@
+module CandidateInterface
+  class RestructuredWorkHistory::BaseController < CandidateInterfaceController
+    before_action :redirect_to_dashboard_if_submitted, :render_404_if_candidate_has_used_existing_flow_or_flag_is_active
+
+    def show; end
```
malcolmbaig, line 5: In the interest of establishing conventions, I think Base should be for common behaviours, and something like StartController for the first step. Clearer separation of concerns.

malcolmbaig:
Liking the explicit field on ApplicationForm and the naming convention for it 👍🏽

config/routes.rb
davidgisbey: Yea it should. Good spot!

app/presenters/candidate_interface/application_form_presenter.rb
davidgisbey: Yes, you're correct. Changing now

app/controllers/candidate_interface/restructured_work_history/base_controller.rb
davidgisbey: I agree. That's a good point. I will update now.

davidgisbey:
Okay i'm going to leave this for a little bit longer and add a few extra things in like scopes. Then I'll start opening up some actual PRs if people are generally happy with the approach.

PR #3891, opened 2021-01-21
R4R Support Dashboard

george-bit:
## Context

Our team requires the ability to view Reasons for Rejection statistics in a palatable format, in order to make business decisions. As such, we are building a Reasons for rejection Dashboard in the support UI to achieve this. This PR covers the addition of the first main page with high level overview statistics.

We've added a controller that pulls the required info from the database via a DB query. This is then interpreted and displayed in the front end via a view component.

Dashboard
Dashboard with data

## Guidance to review

Please test in review app and comment on code.

**Questions**

@paulrobertlloyd - from a Design perspective would it be better to have 2 adjacent columns rather than one single column. We struggled to achieve so if you think it's necessary some front end govuk guidance would be appreciated!

@devs - We've gone ahead with a system spec only, would it be worthwhile to do a spec for the component also? From our POV the system spec covers a lot of the component functionality.

## Link to Trello card

https://trello.com/c/X7UDBypO/2887-dev-r4r-support-dashboard

## Things to check


stevehook (commented):
Good stuff! I've left a couple of comments inline.

app/components/support_interface/reasons_for_rejection_dashboard_component.html.erb
```diff
@@ -0,0 +1,299 @@
...
+          count: current_month_rejection_count("course_full_y_n"),
+          label: 'this month',
+          size: :reduced,
+        ) %>
+      </div>
+      <div class="govuk-grid-column-one-half">
+        <%= render SupportInterface::TileComponent.new(
+          count: previous_rejection_count("course_full_y_n"),
+          label: 'before this month',
+          size: :reduced,
```
davidgisbey, line 24: I hadn't envisioned that we would only have total & last month, but I guess this can't hurt for people who struggle with 'quick maths'.

app/components/support_interface/reasons_for_rejection_dashboard_component.html.erb
```diff
@@ -0,0 +1,299 @@
+<h3 class="govuk-heading-m govuk-!-font-size-27 govuk-!-margin-top-8">Course full</h3>
```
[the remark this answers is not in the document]
davidgisbey: I feel like this could be a component in itself?

Maybe `RejectionReasonDashboardComponent` or something similar.

So you could just pass in what you need.

`RejectionReasonDashboardComponent.new(
   rejection_reason: 'course_full_y_n', 
   total_count:  total_rejection_count("course_full_y_n")
   last_month: current_month_rejection_count("course_full_y_n")
   before_this_month: "previous_rejection_count("course_full_y_n")
)`

You can then get the remaining stuff you need for IDs etc. by adding a few methods like

```
def section_id
  "#{split_and_remove_y_n}_section"
end

def header
  split_and_remove_y_n.join(' ').capitalise
end 

def split_and_remove_y_n
  rejection_reason.split('_').reject { |string| string == 'y' || string == 'n' }
end  
```
This ☝️ can be cleaner but hopefully, you get the idea 👍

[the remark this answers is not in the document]
davidgisbey: Sorry that looks really ugly in markdown

spec/system/support_interface/reasons_for_rejection_dashboard_spec.rb
```diff
@@ -0,0 +1,192 @@
...
+  end
+
+  def given_i_am_a_support_user
+    sign_in_as_support_user
+  end
+
+  def create_application_form_with_references
+    application_form = create(:application_form)
+    create(:application_choice, application_form: application_form)
+    references = create_list(:reference, 2, application_form: application_form)
```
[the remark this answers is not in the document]
davidgisbey: Yea +1 on all of this. We can just create the applications in the correct state before you go through the process of rejecting them 👍

app/views/support_interface/performance/index.html.erb
```diff
@@ -6,5 +6,6 @@
   <li><%= govuk_link_to 'Course options without vacancies', support_interface_course_options_path %></li>
   <li><%= govuk_link_to 'Applications with unavailable choices', support_interface_unavailable_choices_path %></li>
   <li><%= govuk_link_to 'Feature metrics', support_interface_feature_metrics_dashboard_path %></li>
+  <li><%= govuk_link_to 'Reasons for rejection', support_interface_reasons_for_rejection_dashboard_path %></li>
```
davidgisbey, line 9: I had envisioned this being in a separate tab like validations, but tbh it is a dashboard so it probably should go here. I think @stevehook's card should go into a separate tab though.

george-bit: Yeah we were a little, unsure on this one. We kind of placed it here for the time being. not against putting it into a separate tab though!

app/components/support_interface/reasons_for_rejection_dashboard_component.html.erb
george-bit: Yeah, this is due to us following the layout of the Feature metrics dashboard and also having the value available so we just thought it wouldn't hurt to display it. We can remove it if it feels a bit too incongruous.

spec/system/support_interface/reasons_for_rejection_dashboard_spec.rb
```diff
@@ -0,0 +1,192 @@
...
+  def when_i_visit_the_performance_page_in_support
+    visit support_interface_performance_path
+  end
+
+  def and_i_click_on_the_reasons_for_rejection_metrics_link
+    click_on 'Reasons for rejection'
+  end
+
+  def then_i_should_see_reasons_for_rejection_course_full
+    within('#course_full_section') do
```
davidgisbey, line 101: We could think about using something like within `find('. govuk-grid-row)[0]` through to `find('. govuk-grid-row)[7]`

app/components/support_interface/reasons_for_rejection_dashboard_component.html.erb
paulrobertlloyd: I’d be inclined to go with the following layout:

```
[Reason for rejection name]

67%                                 12                16
of all rejections this cycle        last month        this month
```

And have each row stretch the full width of the page (1/2, 1/4, 1/4).

My guess is that having the main number be a percentage of total rejections might be more useful than raw numbers, but I may be wrong.

davidgisbey: Good idea. The only thing is that it would have to be `% of all structured rejections this cycle` 👍

paulrobertlloyd: That’s fine, as long as we have a row which captures the remaining unstructured rejections, so the total equals 100%.

AgaDufrat: There is some prior art on percentage for dashboards you may want to refactor/extract
https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/components/support_interface/ucas_matches_statistics_component.rb#L73-L85

app/components/support_interface/reasons_for_rejection_dashboard_component.rb
```diff
@@ -0,0 +1,45 @@
...
+      values['count']
+    end
+
+    def total_rejection_count(reason)
+      current_month_rejection_count(reason) + previous_rejection_count(reason)
+    end
+
+    def percentage_rejected_for_reason(reason)
+      count = current_month_rejection_count(reason) + previous_rejection_count(reason)
+      total = @rejection_reasons.inject(0) {|sum, hash| sum + hash['count']}
```
george-bit: @davidgisbey We came a cross a problem with the previous way of doing this when we were updating the spec. Essentially, by running `ApplicationChoice.where.not(structured_rejection_reasons: nil).count` this just counts the number of db records with structured reasons. However, each db record can contain multiple reasons as its stored in JSON. Therefore, it made sense to revert to querying the splat and totalling these values.

app/components/support_interface/reasons_for_rejection_dashboard_component.html.erb
george-bit: Thanks @AgaDufrat!

paulrobertlloyd:
Pushed a few changes:

* Made layout full width, with stats in a row
* Changed the label for the first stat from ‘Overall percentage’ to ‘of all rejections’
* I also made this view and it’s related controller a child of `performance` controller. The same should probably be true for the feature metrics page, but left that alone as beyond the scope of this PR.

app/queries/get_reasons_for_rejection_from_application_choices.rb
```diff
@@ -0,0 +1,17 @@
+class GetReasonsForRejectionFromApplicationChoices
```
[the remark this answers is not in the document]
stevehook: I still think it's worth adding the method to run the query in this class rather than the controller. Also it would be worth using SQL params rather than string interpolation for the date range, if only because it would avoid Brakeman complaining about possible SQL injection.

spec/system/support_interface/reasons_for_rejection_dashboard_spec.rb
```diff
@@ -0,0 +1,161 @@
...
+
+  def then_i_should_see_reasons_for_rejection_course_full
+    expect(all('.govuk-heading-m')[0]).to have_content('Course full')
+    expect(all('.govuk-grid-row')[0]).to have_content('10% of all rejections')
+    expect(all('.govuk-grid-row')[1]).to have_content('2 total')
+    expect(all('.govuk-grid-row')[1]).to have_content('1 this month')
+  end
+
+  def and_i_should_see_reasons_for_rejection_candidate_behaviour
+    expect(all('.govuk-heading-m')[1]).to have_content('Candidate behaviour')
```
stevehook, line 100: The index numbers here might make the tests a bit brittle, alternative would be to put a unique `id` attribute on each section and use something like `within('#id').do ... end`

app/components/support_interface/reasons_for_rejection_dashboard_component.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    def formatted_percentage(count, total)
+      percentage = count.percent_of(total)
+      precision = (percentage % 1).zero? ? 0 : 2
+      number_to_percentage(percentage, precision: precision)
+    end
+  end
+end
+
+class Numeric
```
stevehook, line 41: We probably shouldn't be monkey-patching standard lib classes (even though Rails does it!) Maybe this would be better as a helper method if there's nothing else built-in to do the job. (Helpers are accessible from components if you mix in `ViewHelper` I think).

stevehook (commented):
I've left a couple of inline comments. The only thing that I'd be reluctant to merge is the change to `Numeric`. So pretty much ready to go. Good stuff!

app/queries/get_reasons_for_rejection_from_application_choices.rb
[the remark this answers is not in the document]
george-bit: We managed to extract the db query into the the Query object and get rid of interpolation 👍

spec/system/support_interface/reasons_for_rejection_dashboard_spec.rb
george-bit: @stevehook We initially were using `id` as an identifier in our tests, but after a discussion with @davidgisbey we decided that these tests should be more granular. Granted Paul's recent changes to the front end broke the tests (which are now fixed)because of this, but arguably this shouldn't happen again. @davidgisbey is it worth reverting to `id` here?

app/components/support_interface/reasons_for_rejection_dashboard_component.rb
george-bit: Thanks Steve refactored accordingly

spec/system/support_interface/reasons_for_rejection_dashboard_spec.rb
stevehook: I don't think this should hold up this PR.

app/helpers/view_helper.rb
```diff
@@ -131,6 +131,10 @@ def provider_user_name(provider_user)
     end
   end
 
+  def percent_of(number)
```
stevehook, line 134: Do you need to pass two args into this method now it's a view helper? e.g.

```
def percent_of(numerator, denominator)
    numerator.to_f / denominator * 100.0
end
```

george-bit: Apologies, it worked implicitly by calling it in our private method `count.percent_of(total)`which calls `.to_f` on the count and divides it by `total *100`. Therefore, I didn't think to change it.

I think that your way of writing it though is more explicit and is probably more useful for future applications. Have updated now.

stevehook (approved):
Looks good for merge to me if @paulrobertlloyd is happy.

paulrobertlloyd:
Just checking with Mike to see that the labels we’re using are correct (i.e. % of all rejections)

app/components/support_interface/reason_for_rejection_dashboard_section_component.html.erb
```diff
@@ -0,0 +1,27 @@
+<h3 class="govuk-heading-m govuk-!-font-size-27"><%=@heading%></h3>
+<div class="govuk-grid-row govuk-!-margin-bottom-8">
+  <div class="govuk-grid-column-one-half">
+    <%= render SupportInterface::TileComponent.new(
+      count: @percentage_rejected,
+      label: 'of all rejections',
```
davidgisbey, line 6: 'of all structured rejections'?

davidgisbey (approved):
lgtm guys. I just left one comment as I think it should probably be clear this is only factoring in structured rejections.

app/components/support_interface/reason_for_rejection_dashboard_section_component.html.erb
paulrobertlloyd: I'm possibly overthinking this, but I think your suggestion is a little bit clearer, so let's go with that. Then we can :shipit:

PR #3892, opened 2021-01-21
Remove devise localisation file

tijmenb:
## Context

I suspected that this file isn't used anywhere, and the tests run fine without it. We don't any UI from Devise, so this does make sense.

🔥 

## Guidance to review

Does this make sense?

## Link to Trello card

https://trello.com/c/Cot6OIyo/396-medium-locales-review-and-refactoring

## Things to check


docs/parallel_tests.md
```diff
@@ -36,7 +36,7 @@ If you have 16 cores you'll start seeing the following error:
 
 You can fix this as follows:
 
-- Find your Redis configuration file (eg - /etc/redis.conf)
+- Find your Redis configuration file (eg - `/etc/redis.conf`, you can find it running `redis-cli INFO | grep 'config_file'`)
```
despo, line 39: 👍🏼 

I had to use`brew info redis` but this works for everyone, so a better idea.

PR #3894, opened 2021-01-21
Add columns for timestamps for feedback_status changes for references

davidgisbey:
## Context

We're using audits to determine when the feedback status of references. This is causing us to hit the db with a bunch of unnecessary taxing queries.

It would be far easier to just add some columns to the object and update the attr when as part of the services that transition references between states.

This PR adds these columns

Adds the following columns
- provided_at (we've been using received_at, but i think we should probably use the enum value)
- refused_at
- email_bounced
- cancelled_at
- cancelled_at_the_end_of_cycle_at

## Guidance to review

All seem reasonable?

## Link to Trello card

https://trello.com/c/KvzwfGqD/2914-add-receivedat-refusedat-cancelledat-emailbouncedat-and-cancelledattheendofcycleat-to-the-applicationreference-model

## Things to check


stevehook (approved):
👍

PR #3896, opened 2021-01-22
Add feature_restructured_work_history boolean to the application form

davidgisbey:
## Context

This is the first PR from the spike found here #3889 

It adds the feature_restructured_work_history boolean to the application_forms table

- Migration to add the feature_restructured_work_history bool to the application_forms table

## Link to Trello card

https://trello.com/c/ULGeU2Pp/2907-work-history-spike-how-were-going-to-integrate-and-move-users-through-the-new-and-old-flows

## Things to check


db/migrate/20210122102149_add_feature_restructured_work_history_bool_to_application_form.rb
```diff
@@ -0,0 +1,5 @@
+class AddFeatureRestructuredWorkHistoryBoolToApplicationForm < ActiveRecord::Migration[6.0]
+  def change
+    add_column :application_forms, :feature_restructured_work_history, :boolean, default: true
```
stevehook, line 3: I think you need the change in https://github.com/DFE-Digital/apply-for-teacher-training/pull/3889 to set this attribute to false until the feature flag is on. Otherwise we'll have `true` values for any application forms created between the migration and the code change deployments. 

Bit of a pain but could this be `default: false` for now and changed in a later migration?

davidgisbey: Hey Steve, in the draft this is dealt with here in this commit https://github.com/DFE-Digital/apply-for-teacher-training/pull/3889/commits/de9c5fda120a2f5fa95563f404d9bd32539ff4f2

Basically, the new flow will be behind a feature flag. Prior to this being flipped we update the bool to false when a candidate uses the new flow

stevehook: Yes I think so. All I was thinking of was what happens between this deploy and the follow up that adds that code change. I am guessing you could get into situation where a candidate completes work history between the two deploys and would end up with this new attribute set to `true` even though they had used the 'old flow'. It's very much an edge case.

davidgisbey: Oh right sorry. I should also have said that the next PR will include a backfill as the first commit to ensure that it is set to false for courses with any work history or an explanation for why they are not providing one. https://github.com/DFE-Digital/apply-for-teacher-training/pull/3889/commits/e1b5d126cb87acdc376f1da57db110bf6e4db883

davidgisbey: should have linked that earlier too!

stevehook (approved):

PR #3897, opened 2021-01-22
Email accredited bodies about changes

AgaDufrat:
## Context

We currently only notify the training provider when an application arrives. In keeping with the principle that providers are assumed to be interested in all applications they can see in Manage, we should also send emails to users of the accredited body.

Providers should receive emails for all applications they can view.

- extend `NotificationsList` to conditionally return list including ratifying provider users (accredited body)
- Email ratifying providers when 
  - Application is received
  - Offer is accepted
  - Offer is declined
  - Candidate withdraws

## Link to Trello card

https://trello.com/c/AoTxywFa/3269-email-accredited-bodies-as-well-as-training-providers-about-changes-to-applications

## Things to check

### DIFF
--- a/app/services/decline_offer.rb
+++ b/app/services/decline_offer.rb
@@ -12,7 +12,7 @@ class DeclineOffer
       StateChangeNotifier.new(:declined, @application_choice).application_outcome_notification
     end
 
-    NotificationsList.for(@application_choice).each do |provider_user|
+    NotificationsList.for(@application_choice, include_ratifying_provider: true).each do |provider_user|
       ProviderMailer.declined(provider_user, @application_choice).deliver_later
     end
   end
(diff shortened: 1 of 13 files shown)


app/services/decline_offer.rb
```diff
@@ -12,7 +12,7 @@ def save!
       StateChangeNotifier.new(:declined, @application_choice).application_outcome_notification
     end
 
-    NotificationsList.for(@application_choice).each do |provider_user|
+    NotificationsList.for(@application_choice, include_ratifying_provider: true).each do |provider_user|
```
benswannack, line 15: The card mentions we care about declined offers – do we care about decline_by_default ones?

benswannack (approved):
Looks good!

PR #3898, opened 2021-01-22
DO NOT MERGE Unique API IDs for multi-grade GCSEs

duncanjbrown:
## Context

We broke SITS because the API duplicated IDs across multiple GCSEs, eg when a candidate had english language and literature qualifications.

When constructing these qualifications in the API, sort them so they always appear in the same order and add numbered suffixes to the IDs. See comment inline.

## Link to Trello card

## Things to check

### DIFF
--- a/app/presenters/vendor_api/single_application_presenter.rb
+++ b/app/presenters/vendor_api/single_application_presenter.rb
@@ -204,15 +204,39 @@ module VendorAPI
 
     def structured_gcse_to_hashes(gcse)
       structured_grades = JSON.parse(gcse[:structured_grades])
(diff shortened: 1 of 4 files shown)


duncanjbrown:
This shouldn't be merged until we've checked with vendors that the string id won't break their systems

duncanjbrown:
This breaks Ellucian 😞  https://ukgovernmentdfe.slack.com/archives/CTHFLRFHB/p1611568384001400?thread_ts=1611566309.001300&cid=CTHFLRFHB

duncanjbrown:
Closing this to try something less breaky

PR #3900, opened 2021-01-22
Don't call SetDeclineByDefault inside a DB transaction

duncanjbrown:
This service looks at the database to see if there are any offers before
proceeding. If the transaction hasn't committed yet, the offer in
progress will be missing.

### DIFF
--- a/app/services/make_an_offer.rb
+++ b/app/services/make_an_offer.rb
@@ -45,8 +45,9 @@ class MakeAnOffer
           application_choice.offer = { 'conditions' => offer_conditions }
           application_choice.offered_at = Time.zone.now
           application_choice.save!
-          SetDeclineByDefault.new(application_form: application_choice.application_form).call
         end
+
+        SetDeclineByDefault.new(application_form: application_choice.application_form).call
 
         SendNewOfferEmailToCandidate.new(application_choice: @application_choice).call
       else


despo (approved):
👍🏼

app/services/make_an_offer.rb
```diff
@@ -45,9 +45,10 @@ def save
           application_choice.offer = { 'conditions' => offer_conditions }
           application_choice.offered_at = Time.zone.now
           application_choice.save!
-          SetDeclineByDefault.new(application_form: application_choice.application_form).call
```
tijmenb, line 48: Can we do away with the whole transaction here, as there's only a one database update in the statement now?

mnacos:
Yet the tests pass... The idea behind the transaction is to roll back the state transition if DBD cannot be set. In theory, if all statements go via the same postgres db connection/session, pending changes should be visible to other statements within the same transaction. Could we try to reproduce the suspected behaviour in a test?

app/services/make_an_offer.rb
duncanjbrown: :+1: Will update

duncanjbrown:
> Yet the tests pass... The idea behind the transaction is to roll back the state transition if DBD cannot be set. In theory, if all statements go via the same postgres db connection/session, pending changes should be visible to other statements within the same transaction. Could we try to reproduce the suspected behaviour in a test?

Fair challenge @mnacos! And accepted

mnacos:
> Yet the tests pass... The idea behind the transaction is to roll back the state transition if DBD cannot be set. In theory, if all statements go via the same postgres db connection/session, pending changes should be visible to other statements within the same transaction. Could we try to reproduce the suspected behaviour in a test?

Oops, I guess this depends on the isolation level Rails uses: https://alxibra.medium.com/isolation-level-in-rails-847edc9e347d

I guess we could get rid of the transaction, like Tijmen says.

duncanjbrown:
> > Yet the tests pass... The idea behind the transaction is to roll back the state transition if DBD cannot be set. In theory, if all statements go via the same postgres db connection/session, pending changes should be visible to other statements within the same transaction. Could we try to reproduce the suspected behaviour in a test?
> 
> Oops, I guess this depends on the isolation level Rails uses: https://alxibra.medium.com/isolation-level-in-rails-847edc9e347d
> 
> I guess we could get rid of the transaction, like Tijmen says.

"Read committed is the default isolation level in PostgreSQL" — I think this means no dirty reads. It says a dirty read is "Dirty read happens when transaction reads other uncommited transaction". But what about the same uncommitted transaction? Might do an experiment

duncanjbrown:
I'm going to merge this as it's not going to make things worse, then come back to work on suggestions

PR #3901, opened 2021-01-24
Add timestamps when a references feedback_status is updated

davidgisbey:
## Context

At the moment we don't have columnist that indicate when the feedback status was updated to feedback_provided, feedback_refused, cancelled email_bounced or cancelled_at_the_end_of_cycle.

To compensate for this we are using audits to establish when these state transitions occur.

It's not optimal, slow and has recently stung us with the feature metric dashboard, where Steve has had to go full beautiful mind to query the db for something that should be pretty straight forward.

- Backfill the feedback_provided_at, feedback_refused_at, cancelled_at email_bounced_at and cancelled_at_the_end_of_cycle_at columns
- Update the services that deal with transitions between feedback_status states to update the above columns
- Update the codebase to use the above columns rather than using audits

## Guidance to review

I've broken down updating each of the above into a commit, so per commit is recommended.

Have I missed anything?

## Link to Trello card

https://trello.com/c/KvzwfGqD/2914-add-providedat-refusedat-cancelledat-emailbouncedat-and-cancelledattheendofcycleat-to-the-applicationreference-model

## Things to check


app/services/support_interface/candidate_journey_tracker.rb
```diff
@@ -48,19 +48,19 @@ def submitted
     end
 
     def completed_reference_1_requested_at
-      received_references_timings.dig(0, :requested_at)
+      received_references[0]&.requested_at
```
davidgisbey, line 51: I was going to order this like before we previously were and then thought, does it actually matter? Happy to change this there's a reason there should be.

db/migrate/20210121183741_backfill_new_application_reference_datetime_columns.rb
```diff
@@ -0,0 +1,35 @@
...
+    .all
+    .each do |reference|
+      feedback_provided_at = reference.audits.where("audited_changes#>>'{feedback_status, 1}' = 'feedback_provided'").last&.created_at
+      feedback_refused_at = reference.audits.where("audited_changes#>>'{feedback_status, 1}' = 'feedback_refused'").last&.created_at
+      cancelled_at = reference.audits.where("audited_changes#>>'{feedback_status, 1}' = 'cancelled'").last&.created_at
+      cancelled_at_eoc_at = reference.audits.where("audited_changes#>>'{feedback_status, 1}' = 'cancelled_at_end_of_cycle'").last&.created_at
+      email_bounced_at = reference.audits.where("audited_changes#>>'{feedback_status, 1}' = 'email_bounced'").last&.created_at
+
+      reference.update!(
+        feeback_provided_at: feedback_provided_at,
```
tijmenb, line 13: Maybe an `audit_comment` here would be good for posterity.

tijmenb (approved):
Looking good! Good cleanups along the way as well 👍

db/migrate/20210121183741_backfill_new_application_reference_datetime_columns.rb
davidgisbey: 💯  good shout

PR #3906, opened 2021-01-25
R4R support - sub-reason page

stevehook:
## Context

This PR adds a detailed view of structured reasons for rejection in Support UI so that support staff can observe trends.

- [x] Add `ReasonsForRejectionCountQuery` class to encapsulate queries that return R4R reason and sub-reason counts.
- [x] Add sub-reasons page to support interface
- [x] System spec

image

- [x] Integrate with R4R dashboard
- [x] Sorting by count descending
- [x] Totals for each section
- [x] Fix the design

## Guidance for review

- How can the design be improved?
- I've included all the sub-reasons (including that that haven't been used - zero counts) for clarity.

## Link to Trello card

https://trello.com/c/8f1auqR3/2895-dev-r4r-support-sub-reason-page

## Things to check


davidgisbey:
I'll have a look at this either later tonight or first thing tomorrow. I just finished reviewing Toby's find one and my brains not really working at the moment! It definitely needs to be for this one.

stevehook:
> I'll have a look at this either later tonight or first thing tomorrow. I just finished reviewing Toby's find one and my brains not really working at the moment! It definitely needs to be for this one.

No problem waiting until tomorrow. I've made a few changes this afternoon, will get those pushed up later this evening.

app/queries/reasons_for_rejection_count_query.rb
```diff
@@ -0,0 +1,99 @@
...
+          'this_month'
+        ELSE
+          'before_this_month'
+      END AS time_period,
+      count(*)
+    FROM application_choices,
+      jsonb_each(structured_rejection_reasons) AS reasons,
+      jsonb_array_elements_text(reasons.value) AS sub_reasons
+    WHERE structured_rejection_reasons IS NOT NULL
+      AND (reasons.value)::text LIKE '[%'
```
malcolmbaig, line 95: Is this checking the value is an array? Is there a different check that can make this clearer?

app/queries/reasons_for_rejection_count_query.rb
```diff
@@ -0,0 +1,99 @@
...
+    end
+    results
+  end
+
+  def reason_counts_sql
+    "
+    SELECT reasons.key AS key,
+      CASE
+        WHEN rejected_at > $1 THEN
+          'this_month'
```
malcolmbaig, line 66: Potentially worth interpolating the constants `THIS_MONTH`/`BEFORE_THIS_MONTH` into this query to make the connection a little clearer.

app/queries/reasons_for_rejection_count_query.rb
```diff
@@ -0,0 +1,99 @@
...
+    "
+  end
+
+  def sub_reason_counts_sql
+    "
+    SELECT reasons.key AS key,
+      sub_reasons.value AS value,
+      CASE
+        WHEN rejected_at > $1 THEN
+          'this_month'
```
malcolmbaig, line 86: As above, re: constant usage.

app/components/support_interface/reasons_for_rejection_table_component.rb
```diff
@@ -0,0 +1,30 @@
+module SupportInterface
+  class ReasonsForRejectionTableComponent < ViewComponent::Base
+    include ViewHelper
+    attr_accessor :reasons
+
+    TOP_LEVEL_REASONS_I18N_KEYS = {
+      candidate_behaviour_y_n: :candidate_behaviour_what_did_the_candidate_do,
+      quality_of_application_y_n: :quality_of_application,
+      qualifications_y_n: :qualifications_which_qualifications,
```
malcolmbaig, line 9: Should this be `qualifications`?

app/components/support_interface/reasons_for_rejection_table_component.rb
```diff
@@ -0,0 +1,30 @@
+module SupportInterface
+  class ReasonsForRejectionTableComponent < ViewComponent::Base
+    include ViewHelper
+    attr_accessor :reasons
+
+    TOP_LEVEL_REASONS_I18N_KEYS = {
```
malcolmbaig, line 6: It might be worth making the direction of the mapping a bit more obvious, something like `TOP_LEVEL_REASONS_TO_I18N_KEYS`. Had to double check a few times which one was which.

app/queries/reasons_for_rejection_count_query.rb
stevehook: Yes, its pretty horrible though, I'll try to work out a better way to do it.

app/queries/reasons_for_rejection_count_query.rb
```diff
@@ -0,0 +1,99 @@
...
+      'SQL',
+      [[nil, Time.zone.now.beginning_of_month]],
+    ).to_a
+
+    add_sub_results(results, rows)
+  end
+
+private
+
+  MAPPING = {
```
malcolmbaig, line 31: Is this subreasons mapped to top-level reasons? Are some missing, eg - `candidate_behaviour_what_to_improve`?

app/queries/reasons_for_rejection_count_query.rb
```diff
@@ -0,0 +1,99 @@
...
+  }.with_indifferent_access
+
+  def to_results(rows)
+    rows.reduce(ActiveSupport::HashWithIndifferentAccess.new { |hash, reason| hash[reason] = Result.new(0, 0, ActiveSupport::HashWithIndifferentAccess.new { |sub_hash, sub_reason| sub_hash[sub_reason] = Result.new(0, 0, nil) }) }) { |results, row|
+      if row['time_period'] == THIS_MONTH
+        results[row['key']].this_month += row['count'].to_i
+      end
+      results[row['key']].all_time += row['count'].to_i
+      results
+    }.with_indifferent_access
```
malcolmbaig, line 46: A few suggestions for clarity:

Maybe something like
```
    empty_results_hash =
      ActiveSupport::HashWithIndifferentAccess.new do |hash, reason|
        hash[reason] = Result.new(
          0, 0, ActiveSupport::HashWithIndifferentAccess.new do |sub_hash, sub_reason|
            sub_hash[sub_reason] = Result.new(0, 0, nil)
          end
        )
      end
```
which can then be passed to `reduce`.

Also, perhaps `each_with_object` instead of `reduce`?

Finally, is the last call to `with_indifference_access` required?

malcolmbaig:
Shaping up nicely 👍🏽 . Some comments/questions left inline, although some of these are almost definitely because of my limited understanding of reasons for rejection. 😄

app/queries/reasons_for_rejection_count_query.rb
stevehook: I don't think we need to include `candidate_behaviour_what_to_improve` here because it's the free text question that used to candidates to elaborate if they select _other_ for `candidate_behaviour_what_did_the_candidate_do`. This list is limited to the questions that have sub-reasons as answers.

app/components/support_interface/reasons_for_rejection_table_component.rb
stevehook: I'm a bit confused about these names to be honest but the sub-reason labels were put under the `qualifications_which_qualifications` key so this works even if it seems a bit inconsistent.

app/queries/reasons_for_rejection_count_query.rb
stevehook: Much better!

app/components/support_interface/reasons_for_rejection_table_component.rb
```diff
@@ -0,0 +1,30 @@
...
+      performance_at_interview_y_n: :interview_performance,
+      interested_in_future_applications_y_n: :interested_in_future_applications,
+    }.with_indifferent_access
+
+    def initialize(reasons)
+      @reasons = reasons
+    end
+
+    def reason_label(reason)
+      I18n.t("reasons_for_rejection.#{TOP_LEVEL_REASONS_I18N_KEYS[reason]}.title")
```
davidgisbey, line 23: This needs to be updated to use the new `TOP_LEVEL_REASONS_TO_I18N_KEYS` value

app/components/support_interface/reasons_for_rejection_table_component.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+
+    def sub_reason_label(reason, sub_reason)
+      I18n.t("reasons_for_rejection.#{TOP_LEVEL_REASONS_I18N_KEYS[reason]}.#{sub_reason}")
```
davidgisbey, line 27: as does this

spec/system/support_interface/reasons_for_rejection_sub_reasons_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+    Timecop.freeze(@today) do
+      example.run
+    end
+  end
+
+  scenario 'View reasons for rejection sub-reasons' do
+    given_i_am_a_support_user
+    and_there_are_candidates_and_rejected_applications_in_the_system
+
+    when_i_visit_the_reasons_for_rejection_sub_reasons_page
```
davidgisbey, line 17: We need to add a tab to the performance page that links to this page. Maybe we could add clicking it to this test?

app/queries/reasons_for_rejection_count_query.rb
```diff
@@ -0,0 +1,106 @@
...
+  def to_results(rows)
+    results_hash = ActiveSupport::HashWithIndifferentAccess.new do |hash, reason|
+      hash[reason] = Result.new(
+        0, 0, ActiveSupport::HashWithIndifferentAccess.new do |sub_hash, sub_reason|
+          sub_hash[sub_reason] = Result.new(0, 0, nil)
+        end
+      )
+    end
+
+    rows.each_with_object(results_hash) do |row, results|
```
davidgisbey, line 48: This was new to me. Very nice 👌

app/queries/reasons_for_rejection_count_query.rb
```diff
@@ -0,0 +1,106 @@
...
+      if row['time_period'] == THIS_MONTH
+        results[row['key']].this_month += row['count'].to_i
+      end
+      results[row['key']].all_time += row['count'].to_i
+    end
+  end
+
+  def add_sub_results(results, rows)
+    rows.each do |row|
+      result = results[MAPPING[row['key']]]
```
davidgisbey, line 58: Optional: I wonder if it's worth extracting this logic into some explicit methods. It might just be me, but it took me a while to figure out exactly what was going on here. 

Maybe something like 

```
def get_result_object_from_rows_key
end

def get_nested_sub_result_from_rows_value
end 

def populate_sub_result_counts
end 
```

davidgisbey (approved):
I gave this a pretty thorough manual test and it behaved as expected. Really great work Steve. I left a couple of comments inline, but it's looking good to go 👍

app/queries/reasons_for_rejection_count_query.rb
stevehook: that's better, it's not just you!

app/queries/reasons_for_rejection_count_query.rb
malcolmbaig: Would something like `SUBREASONS_TO_TOP_LEVEL_REASONS` be a little clearer?

app/queries/reasons_for_rejection_count_query.rb
stevehook: Turns out I can use `jsonb_typeof(reasons.value) = 'array'`

stevehook:
I've updated this PR to merge the sub-reasons into the dashboard as discussed on Slack.

malcolmbaig (approved):

paulrobertlloyd:
Updated design so that:

* Using the correct heading levels
* Using accessible markup for table (column and row headings with `scope`)

Screenshot 2021-01-29 at 15 06 04

paulrobertlloyd (approved):
Let me know if you are happy with my changes @stevehook 😀

stevehook:
> Let me know if you are happy with my changes @stevehook 😀

Looks much nicer. Just a spec to update.

PR #3907, opened 2021-01-25
Update rake rspec to use format --progress

despo:
Update rake rspec configuration to use `progress` format as there isn't much value in display documentation when running through all the tests, the performance and unit tests because of the amount of text printed.

As the default is set to `--documentation` in [.rspec](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/.rspec), developers should still see a nice readable format when running invidual specs.

### DIFF
--- a/lib/tasks/testing.rake
+++ b/lib/tasks/testing.rake
@@ -3,7 +3,7 @@ begin
 
   performance_test_pattern = 'spec/performance/**/*.rb'
   RSpec::Core::RakeTask.new(:spec_without_performance) do |t|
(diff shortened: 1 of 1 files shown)


duncanjbrown:
I prefer `--documentation` but I don't mind about this change as I never use `rake` — anyone else have a view?

mnacos (approved):
Likewise, I don't use `rake` much, either, calling `rspec` directly most of the time. Since `.rspec` is unchanged, I won't notice anything different. But `--format progress` definitely reduces noise if you're running the whole test suite.

PR #3912, opened 2021-01-25
Fix Support email log search by email address

mnacos:
## Context

Searching the email log in support with an email address doesn't work. It appears to be working sometimes because names are often part of email addresses, and people's names are mentioned in email bodies. 

Currently:

image

but

image

In postgres, column names matching reserved SQL keywords can be quoted with double quotes, but we are using single quotes instead. Switching to double quotes fixes the issue.  

## Guidance to review

Search using a full/exact email address from the log.

## Link to Trello card

https://trello.com/c/qGHgFa3S/3296-some-provider-invites-do-not-appear-in-the-support-email-log

## Things to check

### DIFF
--- a/app/controllers/support_interface/email_log_controller.rb
+++ b/app/controllers/support_interface/email_log_controller.rb
@@ -9,7 +9,7 @@ module SupportInterface
         .page(params[:page] || 1).per(30)
 
       if params[:q]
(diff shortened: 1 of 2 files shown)


duncanjbrown (approved):
🤪

PR #3919, opened 2021-01-26
Show interview details on candidate dashboard

malcolmbaig:
## Context

Providers can arrange interviews with candidates. We want to summarise interview details for each course choice on the candidate dashboard.

Introduce a component to summarize interview details as part of CourseChoicesReviewComponent.

Notable behaviour:

- It shows interviews in a numbered list if there is more than one for a
  given application choice.
- It shows newlines and creates proper hyperlinks if either is present
  in the text of `Interview#location` or `Interview#additional_details`.
Screenshot_20210126_094157

## Guidance to review
- Might need to test this locally to create interviews on the console, as it doesn't look like the Provider-side UI for creating interviews is finished yet?

## Link to Trello card
https://trello.com/c/DikOswoE

## Things to check


TobyRet (approved):
Tested locally and LGTM

app/components/interview_bookings_component.rb
```diff
@@ -0,0 +1,48 @@
...
+  def initialize(application_choice)
+    @application_choice = application_choice
+  end
+
+  def interviews
+    @application_choice.interviews.includes(:provider).order(date_and_time: :asc)
+  end
+
+  def formatted_time(interview)
+    interview.date_and_time.to_s(:govuk_date_and_time).squish
```
despo, line 13: Is there any value in using `squish` here? The formatted date and time should never be a multiline string.

malcolmbaig: This was added to handle the fact that `:govuk_date_and_time` leaves extra spaces in the string if dealing with single-digit day or hour. It's not especially clear what's going on though - have removed and just added an extra space to the spec. Will raise a separate PR to fix the date format.

app/components/interview_bookings_component.html.erb
```diff
@@ -0,0 +1,20 @@
...
+      <li>
+        <p class='govuk-body'><%= formatted_time(interview) %></p>
+        <p class='govuk-body govuk-!-margin-bottom-0'><%= provider_name(interview) %></p>
+        <%= location(interview) %>
+        <%= additional_details(interview) %>
+      </li>
+    <% end %>
+  </ul>
+<% else %>
+  <% interviews.first.tap do |interview| %>
```
davidgisbey, line 13: 👍

davidgisbey (approved):
looks good 👍

app/components/interview_bookings_component.rb
```diff
@@ -0,0 +1,48 @@
...
+
+private
+
+  def show_newlines(text)
+    simple_format(text, class: 'govuk-body')
+  end
+
+  def make_urls_clickable(text)
+    text
+      .yield_self { |t| t.gsub(URI::DEFAULT_PARSER.make_regexp, '<a href="\0">\0</a>') }
```
frankieroberto, line 44: Do we need to set the allowable schemes here (ie only `http` and `https`) to avoid cross-site-scripting security issues (eg `javascript:alert('test')`)?

malcolmbaig: Good question! The sanitize call underneath handles that case, but I'm going to add these anyway because:

- It's more explicit
- It allows for the refactor in commit https://github.com/DFE-Digital/apply-for-teacher-training/pull/3919/commits/eef9799d85c40ac377a79f2b7d15f9fddea8ff9a

PR #3920, opened 2021-01-26
Specify withdraw by date in the subject for multiple acceptances email

AgaDufrat:
## Context

It was missed in the last PR.

## Guidance to review

Preview path `/rails/mailers/candidate_mailer/ucas_match_initial_email_multiple_acceptances`

## Things to check

### DIFF
--- a/spec/mailers/candidate_mailer_ucas_match_initial_email_spec.rb
+++ b/spec/mailers/candidate_mailer_ucas_match_initial_email_spec.rb
@@ -33,7 +33,7 @@ RSpec.describe CandidateMailer, type: :mailer do
 
     it_behaves_like(
       'a mail with subject and content',
-      I18n.t!('candidate_mailer.ucas_match.multiple_acceptances.subject'),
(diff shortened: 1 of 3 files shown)


spec/mailers/candidate_mailer_ucas_match_initial_email_spec.rb
```diff
@@ -33,7 +33,7 @@
 
     it_behaves_like(
       'a mail with subject and content',
-      I18n.t!('candidate_mailer.ucas_match.multiple_acceptances.subject'),
+      I18n.t!('candidate_mailer.ucas_match.multiple_acceptances.subject', withdraw_by_date: '1 February 2021'),
```
despo, line 36: I wonder if we should refactor email specs and not use the locale to test them, as we are hiding what the actual subject is meant to be. Maybe on another PR though.

despo (approved):
LGTM 👍🏼

spec/mailers/candidate_mailer_ucas_match_initial_email_spec.rb
AgaDufrat: Yes, good point. I will add a commit in the next PR

PR #3921, opened 2021-01-26
Extract wizard to reusable module

despo:
## Context

Extracts reusable wizard functionality to module

## Link to Trello card

https://trello.com/c/HfJPOWEM/3299-extract-wizard-to-reusable-module

## Things to check

### DIFF
--- a/app/forms/concerns/wizard.rb
+++ b/app/forms/concerns/wizard.rb
@@ -1,0 +1,57 @@
+module Wizard
+  extend ActiveSupport::Concern
+
+  included do
+    attr_writer :state_store
+    attr_accessor :current_step, :review_mode
+  end
+
+  def initialize(state_store, params = {})
+    @state_store = state_store
+    sanitize_attributes!(params)
+
+    super(last_saved_state.deep_merge(params))
+
+    enter_review_mode! if current_step.eql?('check')
+  end
+
+  # placeholder for multiple step wizards
+  def next_step; end
+
+  # placeholder for multiple step wizards
+  def previous_step; end
+
+  def clear_state!
+    @state_store.delete
+  end
+
+  def save_state!
+    @state_store.write(state)
+  end
+
+  def valid?
+    current_step.present? ? super(current_step.to_sym) : super
+  end
+
+private
+
+  def state
+    as_json(except: params_to_exclude_from_saved_state).to_json
+  end
+
+  def last_saved_state
+    saved_state = @state_store.read
+    saved_state ? JSON.parse(saved_state) : {}
+  end
+
+  def params_to_exclude_from_saved_state
+    %w[state_store errors validation_context current_step]
+  end
+
+  # placeholder to enable sanitizing of input parameters if required
+  def sanitize_attributes!(params); end
+
+  def enter_review_mode!
+    @review_mode = true
+  end
+end
--- a/app/forms/provider_interface/provider_relationship_permissions_setup_wizard.rb
+++ b/app/forms/provider_interface/provider_relationship_permissions_setup_wizard.rb
@@ -1,8 +1,9 @@
 module ProviderInterface
   class ProviderRelationshipPermissionsSetupWizard
     include ActiveModel::Model
+    include Wizard
 
-    attr_accessor :current_step, :current_provider_relationship_id, :checking_answers
+    attr_accessor :current_provider_relationship_id, :checking_answers
     attr_writer :provider_relationships, :provider_relationship_permissions, :state_store
     validate :validate_permissions_form, on: :permissions
 
@@ -19,14 +20,6 @@ module ProviderInterface
           end
         end
       end
-    end
-
-    def initialize(state_store, attrs = {})
-      @state_store = state_store
-
-      super(last_saved_state.deep_merge(attrs))
-
-      self.checking_answers = false if current_step == 'check'
     end
 
     def provider_relationships
@@ -95,14 +88,6 @@ module ProviderInterface
       end
     end
 
-    def save_state!
-      @state_store.write(state)
-    end
-
-    def clear_state!
-      @state_store.delete
-    end
-
     def current_permissions_form
       @_current_permissions_form ||= PermissionsForm.new(
         permissions_for_relationship(current_provider_relationship_id).merge(id: current_provider_relationship_id),
@@ -110,20 +95,6 @@ module ProviderInterface
     end
 
   private
-
-    def state
-      as_json(except: %w[state_store errors validation_context _current_permissions_form current_step]).to_json
-    end
-
-    def last_saved_state
-      state = @state_store.read
-
-      if state
-        JSON.parse(state)
-      else
-        {}
-      end
-    end
 
     def next_provider_relationship_id
       if current_provider_relationship_id.blank?
@@ -155,6 +126,14 @@ module ProviderInterface
       current_permissions_form.errors.map do |key, message|
         errors.add("provider_relationship_permissions[#{current_provider_relationship_id}][#{key}]", message)
       end
+    end
+
+    def enter_review_mode!
+      @checking_answers = false
+    end
+
+    def params_to_exclude_from_saved_state
+      super + %w[_current_permissions_form]
     end
   end
 end
--- a/app/forms/provider_interface/provider_user_invitation_wizard.rb
+++ b/app/forms/provider_interface/provider_user_invitation_wizard.rb
@@ -1,17 +1,15 @@
 module ProviderInterface
   class ProviderUserInvitationWizard
     include ActiveModel::Model
-    STATE_STORE_KEY = :provider_user_invitation_wizard
+    include Wizard
 
-    attr_accessor :current_step, :current_provider_id, :first_name, :last_name, :checking_answers, :single_provider, :view_applications_only
+    attr_accessor :current_provider_id, :first_name, :last_name, :checking_answers, :single_provider, :view_applications_only
     attr_reader :email_address
     attr_writer :providers, :provider_permissions, :state_store
 
     with_options(on: :details) do
-      validates :email_address, presence: true
+      validates :email_address, :first_name, :last_name, presence: true
       validates :email_address, valid_for_notify: true
-      validates :first_name, presence: true
-      validates :last_name, presence: true
     end
 
     validates :providers, presence: true, on: :providers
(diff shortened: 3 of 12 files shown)


app/forms/concerns/wizard.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  def ignored_params
+    %w[state_store errors validation_context current_step]
+  end
+
+  # placeholder to enable sanitizing of input parameters if required
+  def sanitize_attributes(attrs); end
+
+  def update_review_mode
```
duncanjbrown, line 58: could this be a `!` method as it has a side effect? `enter_review_mode!`?

app/forms/concerns/wizard.rb
```diff
@@ -0,0 +1,61 @@
...
+  def state
+    as_json(except: ignored_params).to_json
+  end
+
+  def last_saved_state
+    saved_state = @state_store.read
+    saved_state ? JSON.parse(saved_state) : {}
+  end
+
+  def ignored_params
```
duncanjbrown, line 51: Could this also have a clearer name — as it's a form it almost implies that we're rejecting params coming from the front end? How about `params_to_exclude_from_stored_state`?

app/forms/provider_interface/provider_user_invitation_wizard.rb
```diff
@@ -1,17 +1,15 @@
 module ProviderInterface
   class ProviderUserInvitationWizard
     include ActiveModel::Model
-    STATE_STORE_KEY = :provider_user_invitation_wizard
+    include Wizard
 
-    attr_accessor :current_step, :current_provider_id, :first_name, :last_name, :checking_answers, :single_provider, :view_applications_only
+    attr_accessor :current_provider_id, :first_name, :last_name, :checking_answers, :single_provider, :view_applications_only
```
duncanjbrown, line 6: I think `:checking_answers` can go here?

app/forms/provider_interface/provider_user_invitation_wizard.rb
```diff
@@ -177,5 +141,17 @@ def next_provider_needing_permissions_setup
     def any_provider_needs_permissions_setup?
       next_provider_needing_permissions_setup.present?
     end
+
+    def sanitize_attributes(params)
+      delete_permissions_if_view_applications_only(params)
+    end
+
+    def update_review_mode
```
duncanjbrown, line 149: I find this a bit confusing — why does it do one thing in the module and the opposite out here?

app/forms/concerns/wizard.rb
```diff
@@ -0,0 +1,61 @@
...
+    @state_store.delete
+  end
+
+  alias_method :save, :clear_state!
+
+  def save_state!
+    @state_store.write(state)
+  end
+
+  def valid?
```
duncanjbrown, line 36: fwiw I prefer `valid_for_current_step?` but happy to hear other opinions!

duncanjbrown (commented):
Happy to see this formalised a bit! Left some comments, would like to take it for a spin too

app/forms/provider_interface/provider_user_invitation_wizard.rb
despo: I used `review_mode` in the wizard (as this is what it does in the interview form, enters the review page) but here its' called checking answers so not. We could tackle this by additional refactoring of the provider user invitation wizard but I think it's better to tackle it separately.

app/forms/provider_interface/provider_user_invitation_wizard.rb
despo: This is how the provider_user_invitaiton_form behaves and I'm not entirely sure why it's true in one instance and not in the other, which is why I also extracted this out of the initializer to enable developers to configure it as they need to. Maybe something to look at separately?

app/forms/concerns/wizard.rb
despo: This overrides active record's `valid?` which is what we call to check for the validity of any object. As the form knows what step we are currently are at, I think it makes sense sto be consistent across the app in our validation calls rather than have to call another method that does the same thing in certain places.

duncanjbrown: I accept that argument. I offer a counterargument, which is that the controller has to accommodate the wizard in various ways too - eg implementing the mapping from next_step and previous_step to routes. I think there might be a case that it's a useful sign that we're dealing with a wizard?

app/forms/provider_interface/provider_user_invitation_wizard.rb
benswannack: It doesn't look like it would be too big to remove checking_answers. I think it's worth doing this now since the check_answers variable seems to do the opposite of what you'd expect

app/forms/concerns/wizard.rb
```diff
@@ -0,0 +1,61 @@
...
+    super(last_saved_state.deep_merge(params))
+
+    enter_review_mode! if current_step.eql?('check')
+  end
+
+  def next_step
+    :check
+  end
+
+  def previous_step
```
benswannack, line 22: It would be really cool if we could commonize the logic around review_mode and previous/next steps in this module. Ideally, no-one else needs to know about review_mode. Something like:
```
def previous_step
  if review_mode?
    :check 
  else 
    get the step from the wizard
  end
end
```

spec/forms/concerns/wizard_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+
+      expect(wizard.name).to eq('Fiona')
+      expect(wizard.email).to eq('fiona@dfe.co.uk')
+      expect(wizard.other_property).to eq('_to_remove')
+    end
+
+    context 'when parameter sanitisation is required' do
+      before do
+        described_class.send(:define_method, :sanitize_attributes) do |params|
+          params.delete(:other_property)
```
benswannack, line 39: This spec is a little confusing because it tests logic that it implements itself. A check that the sanitize_attributes method is called would be equivalent and simpler.

benswannack (commented):
I'm going to have a more detailed look, but here are some initial comments

app/forms/concerns/wizard.rb
```diff
@@ -0,0 +1,61 @@
...
+    :new
+  end
+
+  def clear_state!
+    @state_store.delete
+  end
+
+  alias_method :save, :clear_state!
```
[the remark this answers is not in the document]
despo: It was part of the `reasons_for_rejection_wizard.rb` but as I made a typo (it should have been `save!` rather than `save`) and all the tests pass succesfuly I'll just double check to make sure it's not used and remove it.

spec/forms/concerns/wizard_spec.rb
despo: I don't think there is any way I can do that as I can't setup an expectation on an instance of a class without initialising it first. However I have simplified the spec so it's easier to understand.

app/forms/concerns/wizard.rb
despo: There is no common pattern between the different wizards (besides that they tend to go to :check if their conditions are not satisfied, and even that doesn't apply to all of them) so I think trying to achieve this right now will be overengineering.

app/forms/concerns/wizard.rb
despo: Maybe enable both then. In two step wizards where you only have one input step and one verification step (e.g. interviews) there is only one step that can be validated, so in that context `valid_for_current_step` can be confusing.

app/forms/concerns/wizard.rb
benswannack: I think there should be a common pattern between wizards that use a review mode. My understanding was that the existing checking answers flag is used when a user presses a change link from the check page, they go back to the relevant step. Then once they click continue, they jump right back to the check page. It seems like it would be useful to have that logic in one place, even if it were an optional thing for wizards with a check page. I might have misunderstood what review_mode means though?

If not, does that suggest that the logic around entering review mode doesn't belong in this module either?

app/forms/concerns/wizard.rb
steventux: My 2p: while just having a `valid?` method is neater, it reads like the entire wizard instance is valid (ie. either for all steps or all completed steps) rather than for a particular step, so I think it introduces ambiguity from the caller's perspective. 
It also seems like we can't ever now ask if a different step to the current one is valid (not sure we have a use case for this though).

steventux (commented):
I think this is a good place to start with making common parts of the wizard available in a module.
I've added a few comments and questions.

app/forms/provider_interface/provider_relationship_permissions_setup_wizard.rb
```diff
@@ -156,5 +127,13 @@ def validate_permissions_form
         errors.add("provider_relationship_permissions[#{current_provider_relationship_id}][#{key}]", message)
       end
     end
+
+    def enter_review_mode!
+      @checking_answers = false
+    end
+
+    def params_to_exclude_from_saved_state
+      super + %w[_current_permissions_form]
```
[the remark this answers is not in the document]
despo: Not something I've seen used in any other projects and I don't think it adds any value. We should bring this up discuss and address in the dev retro on Tuesday.

spec/forms/concerns/wizard_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+      expect(store).to have_received(:write).with(included_params_to_json)
+    end
+  end
+
+  describe '#clear_state' do
+    it 'calls the data store\'s delete method' do
+      wizard.clear_state!
+
+      expect(store).to have_received(:delete)
+      expect(store.read).to be_nil
```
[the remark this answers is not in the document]
despo: Yes, I moved the check here as well in error from `provider_relationship_permissions_setup_wizard_spec.rb` even though as you mentioned it does nothing!

app/forms/concerns/wizard.rb
```diff
@@ -0,0 +1,59 @@
...
+    @state_store = state_store
+    sanitize_attributes!(params)
+
+    super(last_saved_state.deep_merge(params))
+
+    enter_review_mode! if current_step.eql?('check')
+  end
+
+  def next_step
+    :check
```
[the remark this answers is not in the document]
despo: As the steps are only required on wizards with multiple input steps, I think it's best to just leave the method stubs but no implementation which can cause confusion. Amended to reflect that.

duncanjbrown:
@despo what do you want to do with this PR — would you consider closing it until we're ready to come back to it?

PR #3923, opened 2021-01-26
Squish extra whitespace in custom date formats

malcolmbaig:
Update these formats so that single digit hours and days aren't padded
with an extra space.

```
time = Time.zone.local(2020, 6, 6, 6, 30)
```
## Before
```
time.to_s(:govuk_date_and_time)
=> " 6 June 2020 at  6:30 am"
```

## After
```
time.to_s(:govuk_date_and_time)
=> "6 June 2020 at 6:30am"
```

## Guidance to review
Any reason not to?

## Link to Trello card

na
## Things to check

### DIFF
--- a/config/initializers/date_formats.rb
+++ b/config/initializers/date_formats.rb
@@ -14,7 +14,7 @@ Time::DATE_FORMATS[:govuk_date_and_time] = lambda do |time|
              '%e %B %Y at %l:%M%P'
            end
 
-  time.strftime(format)
+  time.strftime(format).squish
 end
 
 Time::DATE_FORMATS[:govuk_time] = lambda do |time|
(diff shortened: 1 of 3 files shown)


steventux (approved):
Tidy!

davidgisbey (approved):

despo:
@malcolmbaig as we are already using the the no padding flag (`-`) to achieve the `govuk_date` format maybe we can update the existing directives so they don't add the extra padding instead.

eg. change `'%e %B %Y at %l:%M%P'` to `%-e %B %Y at %-l:%M%P`

malcolmbaig:
@despo Makes sense, but seems simpler to handle whitespace once in each of those lambdas, rather than add multiple flags.

PR #3926, opened 2021-01-27
Other qualifications: fix issue with grades when changing BTEC to GCSE

stevehook:
## Context

After creating a BTEC other qualification and then changing the qualification type to GCSE the details form still shows the grade options for BTEC (radio buttons). This is a simple change to ensure that these buttons are only shown when the qualification_type is 'Other' and the other_uk_qualification_type is 'BTEC'.

- [x] Extend existing system spec to reproduce the issue
- [x] Fix in `OtherQualificationDetailsForm` 
- [x] Reset the `other_uk_qualification` attribute so that the qualification isn't labelled as _BTEC_

Before:
image

After:
image

## Guidance to review

- Is there a better way to do this?
- Should we be testing for anything else?

## Link to Trello card

Found when testing (but not the same issue as):
https://trello.com/c/P5pagu0l/2688-clicking-browser-back-button-causes-other-qualifications-to-be-loaded-with-incorrect-state

## Things to check

### DIFF
--- a/app/forms/candidate_interface/other_qualification_details_form.rb
+++ b/app/forms/candidate_interface/other_qualification_details_form.rb
@@ -122,7 +122,8 @@ module CandidateInterface
     end
 
     def btec?
-      other_uk_qualification_type == 'BTEC'
+      qualification_type == OtherQualificationTypeForm::OTHER_TYPE &&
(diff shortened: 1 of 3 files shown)


davidgisbey (approved):
Looks good 👍

PR #3927, opened 2021-01-27
Fix duplication and formatting issues with the support reference

tijmenb:
## Context

The support reference shouldn't have the numbers `0` and `1` because they can be misread as O and L/I. They currently do though. We're also not preventing support reference from being the same between applications.

Fix the bug, do some refactoring, and guarantee uniqueness.

## Guidance to review

Per commit.

## Link to Trello card

https://trello.com/c/qDpMchNO/2926-avoid-duplicate-support-references-and-fix-formatting

## Things to check

### DIFF
--- a/app/models/application_form.rb
+++ b/app/models/application_form.rb
@@ -77,7 +77,7 @@ class ApplicationForm < ApplicationRecord
 
   attribute :recruitment_cycle_year, :integer, default: -> { RecruitmentCycle.current_year }
 
-  before_create -> { self.support_reference ||= GenerateSupportRef.call }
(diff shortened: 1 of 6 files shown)


benswannack (approved):
This all LGTM!

duncanjbrown (approved):
Looks good, pending deduping in prod!

duncanjbrown:
Putting a DO NOT MERGE on this pending confirmation from @tijmenb that he's fixed all dupes in prod, as the index will fall over if not

tijmenb:
I've removed the index for now, because there are a few duplicates left that we're not sure of how to handle.

PR #3931, opened 2021-01-27
"Open all courses" now opens ratified courses too

duncanjbrown:
This makes onboarding HEIs much more straightforward as we don't have to visit each ratified org and pick out the exact courses we need to open.

## Context

We're onboarding more HEIs and this is a painful part of the process.

Make the "Open all courses" button open ratified courses too.

Screenshot 2021-01-27 at 13 08 38

## Guidance to review

Does it do the right thing?

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -48,6 +48,14 @@ class Provider < ApplicationRecord
     Course.where(accredited_provider: self)
   end
 
+  def acts_as_accredited_body_in_current_cycle?
+    accredited_courses.current_cycle.exposed_in_find.any?
+  end
+
+  def all_courses_open_in_current_cycle?
+    accredited_courses.or(courses).current_cycle.exposed_in_find.all?(&:open_on_apply?)
+  end
+
   def application_forms
     ApplicationForm.where(id: application_choices.select(:application_form_id))
   end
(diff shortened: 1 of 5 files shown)


app/models/provider.rb
```diff
@@ -48,6 +48,14 @@ def accredited_courses
     Course.where(accredited_provider: self)
   end
 
+  def acts_as_accredited_body_in_current_cycle?
+    accredited_courses.current_cycle.exposed_in_find.any?
+  end
+
+  def all_courses_open_in_current_cycle?
+    (accredited_courses.current_cycle.exposed_in_find + courses.current_cycle.exposed_in_find).all?(&:open_on_apply?)
```
benswannack, line 56:
```suggestion
    (accredited_courses.or(courses)).current_cycle.exposed_in_find.all?(&:open_on_apply?)
```
Is a bit more concise and seems to do the same thing

benswannack (approved):
Nice!
Is it always the case they want to open every course that a provider is involved with at once? Is there a way to do this more granularly?

duncanjbrown:
> Nice!
> Is it always the case they want to open every course that a provider is involved with at once? Is there a way to do this more granularly?

Yes, this is the requirement — all HEIs from now on. More granular method is to click into each course and individually open, which is what they're doing right now.

app/models/provider.rb
duncanjbrown: Tidy, thank you!

PR #3932, opened 2021-01-27
Ensure incomplete a levels are excluded from the qualifications export

davidgisbey:
## Context

In the qualifications export we order A levels base on subject name. At the moment we have some A levels in incomplete states causing the export to crash.

- Exclude incomplete a levels from the export 

## Guidance to review

These incomplete quals were added prior to use adding the wizard.

## Things to check


stevehook (approved):
This change makes sense. Not tested on `qa` yet because it doesn't look like there are any A/AS levels in the test data.

PR #3935, opened 2021-01-27
Show API-formatted applications in Support

duncanjbrown:
## Context

We want to make the API more accessible and less mysterious. This might also be handy for debugging one day.

This will support James's and my talk about the API tomorrow!

Screenshot 2021-01-27 at 15 55 11

## Guidance to review

Does the wording work? Is the dev-design sufficient?

### DIFF
--- a/app/components/support_interface/application_choice_component.rb
+++ b/app/components/support_interface/application_choice_component.rb
@@ -42,7 +42,13 @@ module SupportInterface
       rows << { key: 'Reject by default at', value: application_choice.reject_by_default_at.to_s(:govuk_date_and_time) } if application_choice.reject_by_default_at && application_choice.awaiting_provider_decision?
       rows << { key: 'Decline by default at', value: application_choice.decline_by_default_at.to_s(:govuk_date_and_time) } if application_choice.decline_by_default_at && application_choice.offer?
 
-      rows.flatten
+      if ApplicationStateChange::STATES_VISIBLE_TO_PROVIDER.include?(application_choice.status.to_sym)
+        rows << { key: 'API', value: render(SupportInterface::ApplicationAPIRepresentationComponent.new(application_choice: application_choice)) }
+      else
+        rows << { key: 'API', value: 'This application hasn’t been sent to the provider yet, so it isn’t available over the API' }
+      end
+
+      rows
     end
 
   private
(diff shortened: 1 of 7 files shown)


app/components/support_interface/application_choice_component.rb
```diff
@@ -42,7 +42,11 @@ def rows
       rows << { key: 'Reject by default at', value: application_choice.reject_by_default_at.to_s(:govuk_date_and_time) } if application_choice.reject_by_default_at && application_choice.awaiting_provider_decision?
       rows << { key: 'Decline by default at', value: application_choice.decline_by_default_at.to_s(:govuk_date_and_time) } if application_choice.decline_by_default_at && application_choice.offer?
 
-      rows.flatten
+      if ApplicationStateChange::STATES_VISIBLE_TO_PROVIDER.include?(application_choice.status.to_sym)
```
tijmenb, line 45: This could use a `ApplicationChoice#visible_to_provider?` method

app/components/support_interface/application_choice_component.rb
```diff
@@ -42,7 +42,11 @@ def rows
...
+        rows << { key: 'API', value: render(SupportInterface::ApplicationAPIRepresentationComponent.new(application_choice: application_choice)) }
+      end
```
tijmenb, line 47: For extra unmysteriousing, we could add a row here to explain that this isn't visible to the API because it's not in a visible state.

tijmenb (approved):
Love it!

app/components/support_interface/application_choice_component.rb
duncanjbrown: Better yet!

app/components/support_interface/application_choice_component.rb
duncanjbrown: We definitely want that elsewhere too, I'll do a follow-up

duncanjbrown:
We've got a failing namespace spec here because we're using the `VendorAPI::SingleApplicationPresenter` in the `SupportInterface`. What do you think we ought to do about that? Move it out of that namespace? Somehow make it an exception in the spec?

cc @tijmenb

tijmenb:
@duncanjbrown We could add an initializer that does this so you can use `AllowedCrossNamespaceUsage::SingleApplicationPresenter`?

```
module AllowedCrossNamespaceUsage
  SingleApplicationPresenter = VendorAPI::SingleApplicationPresenter
end
```

duncanjbrown:
@tijmenb Thanks — I'll take it!

duncanjbrown:
Added to app/lib as autoloading in initialisers isn't allowed

PR #3936, opened 2021-01-27
Don't add HESA data to every API response

TobyRet:
## Context

We only want to return HESA data when an application choice has a status 'recruited'

- Remove the hard-coded hesa attribute from SingleApplicationPresenter. This was overriding the conditional logic responsible for determining whether HESA data should be returned. 😢 
- Only return HESA data if the application choice has a status 'recruited' ('enrolled' is no longer used)
- Add test coverage 

## Guidance to review

Make sense?

## Link to Trello card

https://trello.com/c/uGBOx0F8/2933-dont-add-hesa-data-to-every-api-response

## Things to check

### DIFF
--- a/spec/presenters/vendor_api/single_application_presenter_spec.rb
+++ b/spec/presenters/vendor_api/single_application_presenter_spec.rb
@@ -44,19 +44,36 @@ RSpec.describe VendorAPI::SingleApplicationPresenter do
   end
 
   describe 'attributes.hesa_itt_data' do
-    let(:application_choice) do
-      application_form = create(:completed_application_form, :with_completed_references, :with_equality_and_diversity_data)
-      create(:application_choice, status: 'awaiting_provider_decision', application_form: application_form)
+    context "when an application choice has status 'recruited'" do
+      let(:application_choice) do
+        application_form = create(:completed_application_form, :with_completed_references, :with_equality_and_diversity_data)
+        create(:application_choice, status: 'recruited', application_form: application_form)
+      end
+
+      it 'returns the hesa_itt_data attribute of an application' do
+        equality_and_diversity_data = application_choice.application_form.equality_and_diversity
+
+        response = VendorAPI::SingleApplicationPresenter.new(application_choice).as_json
+
+        expect(response.dig(:attributes, :hesa_itt_data)).to eq(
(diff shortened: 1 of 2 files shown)


spec/presenters/vendor_api/single_application_presenter_spec.rb
```diff
@@ -44,19 +44,33 @@
...
-      )
+    context "when an application choice does not have status 'recruited'" do
+      let(:application_choice) do
+        application_form = create(:completed_application_form, :with_completed_references, :with_equality_and_diversity_data)
+        create(:application_choice, status: 'offer', application_form: application_form)
+      end
+
+      it 'does not return the hesa_itt_data attribute of an application' do
+        response = VendorAPI::SingleApplicationPresenter.new(application_choice).as_json
+        expect(response.dig(:attributes).keys).not_to include(:hesa_itt_data)
```
duncanjbrown, line 72: What's the value of a `dig` when we only go one layer?

duncanjbrown (approved):
This is great, thank you @TobyRet!

One very minor style q inline.

And I wonder if for consistency we should set `hesa_itt_data` explicitly to `nil` in line with the other top-level attrs in the API? e.g. we always include `rejection` and `withdrawal` but they can simply be `nil` if those things haven't happened. I quite like that pattern because it avoids surprises later.

spec/presenters/vendor_api/single_application_presenter_spec.rb
TobyRet: Woops that `dig` should not be I've updates the assertion in this spec. Given we want to assign `nil` to the `:hesa_itt_data` in all other status scenarios it now reads:

` expect(response[:attributes][:hesa_itt_data]).to be_nil`

TobyRet:
> This is great, thank you @TobyRet!
> 
> One very minor style q inline.
> 
> And I wonder if for consistency we should set `hesa_itt_data` explicitly to `nil` in line with the other top-level attrs in the API? e.g. we always include `rejection` and `withdrawal` but they can simply be `nil` if those things haven't happened. I quite like that pattern because it avoids surprises later.

Good point - have made changes to reflect the above

duncanjbrown (approved):
Nice and explicit — thank you!

PR #3937, opened 2021-01-27
Update the R4R percentage tile 

davidgisbey:
## Context
 
The previous implementation worked as follows:

If there are 5 application choices with the following rejection reasons:
- 5 of 5 were rejected for candidates_behaviour
- 3 of 5 were rejected on quals,
- 2 of 5 were rejected on safeguarding
- 0 of 5were rejected for all other reasons

It would show the following %s
- candidates behaviour: 50% (5 of the 10 rejection reasons)
- quals: 30%  (3 of the 10 rejection reasons)
- safeguarding: 20%  (2 of the 10 rejection reasons)
- all_others: 0%

The cumulative total of all the rejection reasons will always add up to 100%.

After some discussion with James, Steve and Rachael, using a different approach which treats each rejection reason as distinct and doing the `total_rejections_for_x  /total_number_of_choices_with_structured_rejection_reasons * 100` seems to be the way to go. 

Firstly, while is indicates rejection reasons are more common, this can be ascertained by just looking at the total number of rejection for each section.

Secondly, most of the tiles on the dashboard are for rejection reasons. However, the `Interested in future applications `and` Other advice or feedback `sections are not. Due to this the actual %s of the reasons the providers gave for rejecting applicants will be distorted.

For example, there were 5 total, rejections all of which were for qualifications, but for each rejection, the provider said they would be interested in the applicant applying again and gave feedback telling the candidate to reapply, you would get the following %s on the dashboard:

33.3% rejected for quals
33.3% for Interested in future applications
33.3% Other advice or feedback

Update the components to:

1. Calculate the percentage by taking the amount_rejected_for_reason / application_choices_with_restructured_rejection_reasons* 100
2. Be a bit more explicit what is g
Before 

image

After

image

## Things to check

### DIFF
--- a/app/components/support_interface/reason_for_rejection_dashboard_section_component.html.erb
+++ b/app/components/support_interface/reason_for_rejection_dashboard_section_component.html.erb
@@ -1,9 +1,9 @@
 <h3 class="govuk-heading-m govuk-!-font-size-27"><%=@heading%></h3>
-<div class="govuk-grid-row govuk-!-margin-bottom-8">
+<div id="<%= set_div_id_from_heading %>"class="govuk-grid-row govuk-!-margin-bottom-8">
   <div class="govuk-grid-column-one-half">
     <%= render SupportInterface::TileComponent.new(
       count: @percentage_rejected,
-      label: 'of all structured rejections',
+      label: number_of_rejections_out_of_total_rejections,
       colour: :blue,
       size: :reduced,
     ) %>
--- a/app/components/support_interface/reasons_for_rejection_dashboard_component.rb
+++ b/app/components/support_interface/reasons_for_rejection_dashboard_component.rb
@@ -19,12 +19,11 @@ module SupportInterface
     end
 
     def percentage_rejected_for_reason(reason)
-      count = current_month_rejection_count(reason) + previous_rejection_count(reason)
-      formatted_percentage(count, total_structured_rejection_reasons_count)
+      formatted_percentage(total_rejection_count(reason), total_structured_rejection_reasons_count)
     end
 
     def total_structured_rejection_reasons_count
-      ApplicationChoice.where.not(structured_rejection_reasons: nil).count
+      @total_structured_rejection_reasons_count ||= ApplicationChoice.where.not(structured_rejection_reasons: nil).count
     end
 
     def previous_rejection_count(reason)
--- a/app/components/support_interface/reasons_for_rejection_dashboard_component.rb
+++ b/app/components/support_interface/reasons_for_rejection_dashboard_component.rb
@@ -5,6 +5,8 @@ module SupportInterface
       @rejection_reasons = rejection_reasons
     end
 
+  private
+
     def current_month_rejection_count(reason)
       values = @rejection_reasons.find { |h| h['key'] == reason && h['time_period'] == 'this_month' }
       return 0 if values.nil?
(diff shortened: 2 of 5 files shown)


spec/system/support_interface/reasons_for_rejection_dashboard_spec.rb
```diff
@@ -90,72 +134,82 @@ def reject_application_without_structured_reasons(application_choice)
   end
 
   def then_i_should_see_reasons_for_rejection_course_full
-    expect(all('.govuk-heading-m')[0]).to have_content('Course full')
-    expect(all('.govuk-grid-row')[0]).to have_content('10% of all structured rejections')
-    expect(all('.govuk-grid-row')[0]).to have_content('2 total')
-    expect(all('.govuk-grid-row')[0]).to have_content('1 this month')
+    within '#course_full' do
```
george-bit, line 137: Would we want to keep an `expect` here for the percentage? It will still be displayed on the page so would make sense to cover this functionality.

george-bit (commented):
LGTM, just see my comment RE: test coverage :)

spec/system/support_interface/reasons_for_rejection_dashboard_spec.rb
davidgisbey: Yea really good point George 👍 . Will add it back in!

app/components/support_interface/reasons_for_rejection_dashboard_component.rb
```diff
@@ -18,11 +20,12 @@ def total_rejection_count(reason)
 
     def percentage_rejected_for_reason(reason)
       count = current_month_rejection_count(reason) + previous_rejection_count(reason)
-      total = @rejection_reasons.inject(0) { |sum, hash| sum + hash['count'] }
-      formatted_percentage(count, total)
+      formatted_percentage(count, total_structured_rejection_reasons_count)
```
JR-G, line 23: Everything looks good - only a minor question - can the `total_rejection_count` method be passed here in place of count?

davidgisbey: 💯 good spot

app/components/support_interface/reason_for_rejection_dashboard_section_component.html.erb
```diff
@@ -1,9 +1,9 @@
 <h3 class="govuk-heading-m govuk-!-font-size-27"><%=@heading%></h3>
-<div class="govuk-grid-row govuk-!-margin-bottom-8">
+<div id="<%= set_div_id_from_heading %>"class="govuk-grid-row govuk-!-margin-bottom-8">
```
stevehook, line 2: :+1:

stevehook (approved):
Looks good. One suggestion inline.

app/components/support_interface/reasons_for_rejection_dashboard_component.rb
```diff
@@ -17,12 +19,12 @@ def total_rejection_count(reason)
...
     def percentage_rejected_for_reason(reason)
-      count = current_month_rejection_count(reason) + previous_rejection_count(reason)
-      total = @rejection_reasons.inject(0) { |sum, hash| sum + hash['count'] }
-      formatted_percentage(count, total)
+      formatted_percentage(total_rejection_count(reason), total_structured_rejection_reasons_count)
     end
 
-  private
+    def total_structured_rejection_reasons_count
+      ApplicationChoice.where.not(structured_rejection_reasons: nil).count
```
stevehook, line 26:
```suggestion
      @ total_structured_rejection_reasons_count ||= ApplicationChoice.where.not(structured_rejection_reasons: nil).count
```

davidgisbey: 💯 yea good point!

PR #3941, opened 2021-01-28
Support dashboards

paulrobertlloyd:
## Context

Tidies up some of the different dashboard names in the Performance section of Support.

* Replace ‘Course stats’ with ‘Course statistics’
* Replace ‘UCAS matches stats’ with ‘UCAS matches’
* Changes `/support/performance/ucas-matches-stats` to `/support/performance/ucas-matches` 
* Replace ‘Reasons for rejection’ with ‘Structured reasons for rejection’
* Move feature metrics into performance namespace (also fixes main navigation item not being correctly highlighted)
* Move service performance dashboard into performance namespace
* Use `dashboard` keyword in views that use dashboard design/layout

## Things to check


app/controllers/support_interface/performance_controller.rb
```diff
@@ -10,6 +10,12 @@ def course_options
       @course_options = CourseOption.where('vacancy_status != ?', 'vacancies').includes(:course, :site)
     end
 
+    def dashboard
+      raise ArgumentError unless params[:year].in?([nil, '2020', '2021'])
```
duncanjbrown, line 14: While we're here, this seems a bit overdramatic! Maybe we should ignore non-year values not in that list? sth like

```ruby
year = params[:year] if ['2020', '2021'].include(params[:year])

@statistics = PerformanceStatistics.new(year)
```

config/routes.rb
```diff
@@ -816,23 +816,22 @@
 
     scope '/performance' do
       get '/' => 'performance#index', as: :performance
-      get '/dashboard' => 'performance_dashboard#dashboard', as: :performance_dashboard
 
-      get '/course-stats', to: 'performance#course_stats', as: :course_stats
+      get '/dashboard' => 'performance#performance_dashboard', as: :performance_dashboard
```
duncanjbrown, line 820: A dashboard among dashboards — could this be `service-performance` and `service_performance_dashboard` maybe?

duncanjbrown (approved):
👌 Tidy and consistent — thanks @paulrobertlloyd! Minor comments inline

app/controllers/support_interface/performance_controller.rb
paulrobertlloyd: Done! Seemed like it actually need to be this:

```ruby
year = params[:year] if ['2020', '2021'].include?(params[:year])
```

(`include?`)

config/routes.rb
paulrobertlloyd: How does `/support/performance/service` grab you?

duncanjbrown: 🙆‍♂️ !

app/controllers/support_interface/performance_controller.rb
```diff
@@ -10,10 +10,10 @@ def course_options
 
     def courses_dashboard; end
 
-    def performance_dashboard
-      raise ArgumentError unless params[:year].in?([nil, '2020', '2021'])
+    def service_performance_dashboard
+      year = params[:year] if ['2020', '2021'].include?(params[:year])
```
AgaDufrat, line 14:
```suggestion
      year = params[:year] if %w[2020 2021].include?(params[:year])
```
I think rubocop may complain about that

paulrobertlloyd: It did! Fixed 😀

PR #3942, opened 2021-01-28
Add activerecord-pg_enum & use it to create a new enum column for work_history_status

davidgisbey:
## Context

I got GitHub action'ed in my other PR: 

image

Due to this, I've added the Activerecord-pg_enum which allows us to use enum types in our schema. This should ensure data integrity and will complain if fail loudly if someone renames an enum in the codebase.

- Add the Activerecord-pg_enum gem
- Use it to create the work_history_status enum

## Guidance to review

Does this all look correct?

## Link to Trello card

https://trello.com/c/byVRnLCS/2896-work-history-add-no-i-have-always-been-in-full-time-education-option-to-initial-page

## Things to check


db/migrate/20210128094853_add_work_history_status_enum.rb
```diff
@@ -0,0 +1,13 @@
+class AddWorkHistoryStatusEnum < ActiveRecord::Migration[6.0]
+  def up
+    remove_column :application_forms, :work_history_status, :string
+    create_enum "work_history_type", %w[can_complete full_time_education can_not_complete]
```
frankieroberto, line 4: I’d be tempted to name the enum `work_history_status` as it’s not really a type? Or alternatively something reflecting the UI?

```suggestion
    create_enum "work_history_answer", %w[yes no_full_time_education no_work_history]
```

davidgisbey:
@mnacos want's to discuss using these at the near misses at 11:30. So I'm going to hold off on making any changes until then. 👍

db/migrate/20210128094853_add_work_history_status_enum.rb
davidgisbey: Yea that makes sense. I'll make that change, but hold off on thinking about merging until 👇 that discussion is had.

davidgisbey:
Closing after discussion with Michael

PR #3943, opened 2021-01-28
Spec enhancements -  stubbing in factories

despo:
## Context

Enables stubbing in factories, as discussed in last week's dev retro, and updates ucas_match spec to utilise stubbed factory where data retrieval is not required.

- Update rspec configuration to enable stubs in factories
- Use new stubeed factory in UCASMatch spec
## Things to check

### DIFF
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -1,4 +1,20 @@
 FactoryBot.define do
+  factory :dual_application_ucas_match, class: 'UCASMatch' do
+    trait :need_to_send_reminder_emails do
+      action_taken { 'initial_emails_sent' }
+      candidate_last_contacted_at { 5.business_days.before(Time.zone.now) }
+    end
+
+    trait :need_to_request_withdrawal_from_ucas do
+      action_taken { 'reminder_emails_sent' }
+      candidate_last_contacted_at { 5.business_days.before(Time.zone.now) }
+    end
+
+    after(:stub) do |ucas_match, _|
+      allow(ucas_match).to receive(:dual_application_or_dual_acceptance?).and_return(true)
+    end
+  end
+
   factory :ucas_match do
     candidate { application_form.candidate }
     matching_data { nil }
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -1,20 +1,4 @@
 FactoryBot.define do
-  factory :dual_application_ucas_match, class: 'UCASMatch' do
-    trait :need_to_send_reminder_emails do
-      action_taken { 'initial_emails_sent' }
-      candidate_last_contacted_at { 5.business_days.before(Time.zone.now) }
-    end
-
-    trait :need_to_request_withdrawal_from_ucas do
-      action_taken { 'reminder_emails_sent' }
-      candidate_last_contacted_at { 5.business_days.before(Time.zone.now) }
-    end
-
-    after(:stub) do |ucas_match, _|
-      allow(ucas_match).to receive(:dual_application_or_dual_acceptance?).and_return(true)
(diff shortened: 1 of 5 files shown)


duncanjbrown (commented):
This mocking in factories is powerful stuff. I'm on the fence. If we used it everywhere we'd be lying to ourselves, but I can see that it simplifies things here. Welcome views from others.

despo:
I did consider using a more descriptive name for the factory to make it more obvious that this is a stub, which may be a pattern worth adopting if we move forward with this. 

At the moment we seem to be moving away from testing everything thoroughly because of the complexity and time it takes to setup data, so I think it can help us aim for better test coverage.

tijmenb (commented):
I think this is excellent, and a good way to allow us to use `build_stubbed` in cases where it wouldn't be possible before.

But because it's slightly unorthodox and might raise some questions, I would prefer the factory definition to be [moved to the test file like we do here](https://github.com/DFE-Digital/apply-for-teacher-training/blob/spec-enhancements--stubbing-in-factories/spec/presenters/candidate_interface/personal_details_review_presenter_spec.rb#L4). That way it's close to the consuming code, which may make it easier to understand. It also solves our problem of ever-expanding factories. WDYT?

mnacos:
I think this approach works well for `:dual_application_ucas_match` but I'd be interested to see how/if it works for application forms and choices, which have greater dependencies on associations between themselves and other models. Yes, our current default of _creating_ applications leads to tests that are slower to run but these are faster to write (because in most cases we just reference a trait and overwrite one or two values, if any). There is, of course, the cost of maintaining these traits, but I think maintaining traits and aligning them with business rules, process states etc. is a good thing™️. But this is an aside. 

The other thing to consider is our general approach to testing. We tend to favour service-level tests because a) they immediately correspond to business value and processes and b) they allow us to refactor without constant revising of tests. I don't think using stubbed models would be appropriate for such tests, but perhaps many presentation layer scenaria (e.g. component specs, previews) could benefit from a more stubbed approach. But even then, some components fetch data from multiple associations etc. so we'd need to do a fair amount of stubbing.

To summarise, I'm happy with this specific PR but I'm not convinced we could take the same approach across the board, with the aim of speeding up the test suite. Is it worth trying this approach for different types of spec involving application choices in different states next time?

duncanjbrown:
I agree with @tijmenb's comment. I think we should make that change, merge this, and stick to that (stub factories should be in the spec file) as our practice in the short term, then review in due course? Wdyt @despo?

despo:
@tijmenb @duncanjbrown  Happy to move this out of the factories, however I am not a fan of noisy specs (having to scroll to line 62 to find the actual tests as in the example provided by @tijmenb) and would rather to have a well named helper module that is included in the specs, which should also make visibility of this functionality easier across the project as it won't rely on other team members having to specifically open the ucas_match_spec in order to see this implementation. E.g. UCASMatchStubbedFactory in `/spec/support/stub_factories/`

tijmenb (approved):
Nice. In the interest of moving forward, let's merge this in no!

Not 100% sold yet on the new way of creating the factory, we could revisit at some point.

despo:
@mnacos 

The intention is not to use these across all specs, but to instead replace some of the **already** stubbed factories generated using `build_stubbed` and overcome some of FactoryBot's limitations, such as being able to access the course of a stubbed application choice. I think the example demonstrated in this PR is unfortunately not the best one but due to the necessary identified refactoring to the application choice factories (to improve trait usage) I didnt want to rush and create any stub factories for ApplicationChoice just yet!

PR #3944, opened 2021-01-28
Update spec for adding volunteering roles

paulrobertlloyd:
## Context

When I updated the add another volunteering role link in #3911, copy and pasting a link from another view resulted in me shipping a change that broke this journey (the link pointed to adding another course instead).

* Test that clicking ‘No’ shows the review page with a link to the *Get school experience* service
* Test that in this state the links say ‘Add a role’
* Test that when there are roles added, the links says ‘Add another role’
* Test that when clicking on either of these links, the user is actually taken to the ‘Add role’ form

## Guidance to review

* I inserted these additional tests in the existing journey; do these revised test journeys work?
* I ended up adding the new methods to the bottom of the file; is that correct?

## Link to Trello card

https://trello.com/c/pOlhOCkF/

## Things to check

### DIFF
--- a/spec/system/candidate_interface/entering_details/candidate_entering_volunteering_and_school_experience_spec.rb
+++ b/spec/system/candidate_interface/entering_details/candidate_entering_volunteering_and_school_experience_spec.rb
@@ -29,12 +29,19 @@ RSpec.feature 'Entering volunteering and school experience' do
     when_i_click_on_volunteering_with_children_and_young_people
     then_i_can_check_my_answers
 
-    when_i_delete_my_volunteering_role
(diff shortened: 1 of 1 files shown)


stevehook (approved):
Makes sense to me. To answer the second question, normally we declare the methods in the same order as they are appear in the scenario block though it's not always possible if methods are reused.

paulrobertlloyd:
> normally we declare the methods in the same order as they are appear in the scenario block though it's not always possible if methods are reused.

Yeah, I was unsure of this, as they all follow the method that made sense to go after `then_i_can_check_my_answers`.

PR #3945, opened 2021-01-28
Issues with pdf application downloads

SamJamCul:
## Context
https://trello.com/c/AZSwZwcx/2568-issues-with-pdf-application-downloads

Applications can be downloaded as PDFs, but they were being formatted as American Letter pages, rather than A4 pages. Additionally, they took a relatively unmodified version of the page, which included a lot of 'webby' bits that weren't very high value, and took up page space unecessarily. 

* Change application PDF downloads to A4 sizing
* Added no print flags to a selection of elements on the page, so that they don't show up in the PDF.

### Before
image

### After
image

## Guidance to review

You can check the PDF page sizing by downloading an application as a PDF using the link on the application page here 

image

If you're on a Mac, you can open the PDF from the finder, and from the top menu under Tools, use the inspector to view the page dimensions. A4 pages should be 210mm x 297mm. 

You can check the components that have been removed by downloading the PDF for an application as well, but if you want to inspect the code itself, the class `govuk-!-display-none-print` has been added to any elements that are deliberately being ignored

## Link to Trello card
https://trello.com/c/AZSwZwcx/2568-issues-with-pdf-application-downloads

## Things to check

### DIFF
--- a/app/components/header_component.html.erb
+++ b/app/components/header_component.html.erb
@@ -1,4 +1,4 @@
-<header class="govuk-header <%= classes %>" role="banner" data-module="govuk-header">
+<header class="govuk-header govuk-!-display-none-print <%= classes %>" role="banner" data-module="govuk-header">
   <div class="govuk-header__container govuk-width-container">
     <div class="govuk-header__logo">
       <%= link_to 'https://gov.uk', class: 'govuk-header__link govuk-header__link--homepage' do %>
--- a/app/components/interview_preferences_component.html.erb
+++ b/app/components/interview_preferences_component.html.erb
@@ -1,6 +1,6 @@
 <h2 class="govuk-heading-m govuk-!-font-size-27 govuk-!-margin-top-8" id="interview">Interview</h2>
 <details class="govuk-details" data-module="govuk-details">
-  <summary class="govuk-details__summary">
+  <summary class="govuk-details__summary govuk-!-display-none-print">
     <span class="govuk-details__summary-text">
       View guidance given to candidate
     </span>
--- a/app/components/personal_statement_component.html.erb
+++ b/app/components/personal_statement_component.html.erb
@@ -1,6 +1,6 @@
 <h2 class="govuk-heading-m govuk-!-font-size-27 govuk-!-margin-top-8" id="personal-statement">Personal statement</h2>
 <details class="govuk-details" data-module="govuk-details">
-  <summary class="govuk-details__summary">
+  <summary class="govuk-details__summary govuk-!-display-none-print">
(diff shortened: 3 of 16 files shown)


app/views/layouts/_header.html.erb
```diff
@@ -1,12 +1,12 @@
 <% case try(:current_namespace) %>
 <% when 'candidate_interface' %>
-  <%= render(HeaderComponent.new(classes: "app-header--#{HostingEnvironment.environment_name}",
+  <%= render(HeaderComponent.new(classes: "app-header--#{HostingEnvironment.environment_name} govuk-!-display-none-print",
```
duncanjbrown, line 3: I wonder if we should just add this to the `HeaderComponent`? It's now `govuk-!-display-none-print` everywhere it's used

app/views/layouts/_header.html.erb
```diff
@@ -1,12 +1,12 @@
...
                                  service_name: service_name,
                                  service_url: service_link,
                                  navigation_items: NavigationItems.for_candidate_interface(try(:current_candidate), controller))) %>
                                <%= render PhaseBannerComponent.new %>
 <% when 'support_interface' %>
-  <%= render(ProductHeaderComponent.new(classes: "app-header--#{HostingEnvironment.environment_name} app-header--full-border",
+  <%= render(ProductHeaderComponent.new(classes: "app-header--#{HostingEnvironment.environment_name} app-header--full-border govuk-!-display-none-print",
```
duncanjbrown, line 9: As above!

duncanjbrown (commented):
Looking good! Would you consider also excluding

- the "SANDBOX FEATURE" block
- the expanding "View guidance given to Candidate" links

?

app/views/layouts/_header.html.erb
SamJamCul: I did try to add it directly to the header component template, but it wasn't working. Was hoping that someone could help me understand why, seeing as it worked here when injected as the arg

SamJamCul: Sorted it out, turns out there's more than one type of header component, so I've flagged the correct ones.

SamJamCul:
> 
> 
> Looking good! Would you consider also excluding
> 
>     * the "SANDBOX FEATURE" block
> 
>     * the expanding "View guidance given to Candidate" links
> 
> 
> ?

I've excluded these, and updated the preview image accordingly!

duncanjbrown (approved):
Thanks @SamJamCul!

I've just realised this won't include the offer any more as it's now on a separate tab. I'll ask URs whether they might want us to the offer back into the exported PDF.

PR #3946, opened 2021-01-28
Initial migration for addition of public ids for qualifications

benswannack:
## Context
See the ADR included here

Initial migration for work coming up, should not change any behaviour!

## Guidance to review
Read the ADR, check it makes sense :) 

## Link to Trello card

## Things to check

This one is interesting, I don't yet want to backfill data. I have added a rake task to backfill the data which may need to be run multiple times during the crossover


benswannack:
My planned steps for releasing the change are as follows (each is a separate PR):
1. This PR: Add columns to application_qualification for storing public_id
2. Change the grade forms to start populating the constituent_grades rather than the structured_grades. Also migrate data from the structured_grades to constituent_grades, leaving the public_ids nil for now. (structured_grades should no longer be used)
3. Final PR: Migration to populate all public_ids
4. Start using public_id in API

How does that sound?

adr/0018-public-ids-for-qualifications.md
```diff
@@ -0,0 +1,70 @@
...
+API consumers expect one qualification entity per exam/grade.
+However, because of the structured grades introduced above, some exams/grades share the same database id.
+Therefore, not all the ids for qualifications coming through the API are unique.
+
+## Decision
+
+We initially considered changing the data type of the id field in the API response to a string, to support ids of the form `123`, `123_1`, `123_2`.
+Vendors were not able to support that change, so the field had to remain an integer.
+
+The solution for this will be split into two: one for qualifications with a simple grade, and one for qualifications with structured grades.
```
benswannack, line 27: This isn't true!

Triple science GCSEs are treated like simple grade qualifications in the API even though they have structured grades. This type of GCSE should get a top level public_id, rather than separate public_ids for each grade

adr/0018-public-ids-for-qualifications.md
```diff
@@ -0,0 +1,70 @@
...
+    public_id: nil,
+    grade: nil,
+    structured_grades: {
+      'English Language': { grade: 'E', public_id: '123' },
+      'English Literature': { grade: 'B', public_id: '1093' },
+      'Cockney Rhyming Slang': { grade: 'A*', public_id: '1094' }
+    },
+    level: 'gcse'
+}
+```
```
benswannack, line 45: It is possible to edit qualifications between structured and simple grade ones. E.g. I can start with
```
{
    id: 123
    subject: 'english',
    public_id: 1092,
    grade: 'A',
    structured_grades: nil,
    level: 'gcse'
}
```
Then edit it to be a multiple part GCSE:
```
{
    id: 123
    subject: 'english',
    public_id: nil,
    grade: nil,
    structured_grades: {
      'English Language': { grade: 'E', public_id: '123' },
      'English Literature': { grade: 'B', public_id: '1093' },
      'Cockney Rhyming Slang': { grade: 'A*', public_id: '1094' }
    },
    level: 'gcse'
}
```
Then what happens if I change my mind back again?
```
{
    id: 123
    subject: 'english',
    public_id: ????,
    grade: 'A',
    structured_grades: nil,
    level: 'gcse'
}
```
This isn't really a new qualification object, so giving it a new public_id doesn't seem correct.

Do we want to try to remember the old public_id? Is this even an issue?

adr/0018-public-ids-for-qualifications.md
raamSoftwire: This was one of the sticking points I wanted to revisit. I'm glad you picked it up, cheers Ben!

adr/0018-public-ids-for-qualifications.md
duncanjbrown: From an API point of view you'll change the set from 

`[ { id: 123...}, { id: 1093 ...}, { id: 1094...} ]`

to 

`[{ id: 123...}]`

I don't know how SRSs will cope will removed IDs — hopefully they'll tolerate them.

In practice, though, this will hardly ever happen because providers don't see applications until they're submitted, so the candidate should get it right first. If they did need to change after submission we'd handle the change through support and probably advise the provider too as we'd assume they'd already looked at it.

raamSoftwire: To answer the initial question, this can't happen with English GCSEs, because they are always stored as structured grades. But it could happen with science GCSEs where single and double awards are stored _normally_, but triple awards are structured as json.

I agree with Duncan on the next part. This should be set once the applications are submitted.

duncanjbrown (approved):

adr/0018-public-ids-for-qualifications.md
```diff
@@ -0,0 +1,71 @@
...
+For qualifications with structured grades (built from multiple exams/grades), the public_id column will be set to `nil`, and the public id will be set inside the structured_grades json, with one id per grade.
+A qualification with structured grades could look like this:
+```
+{
+    id: 123
+    subject: 'english',
+    public_id: nil,
+    grade: nil,
+    structured_grades: {
+      'English Language': { grade: 'E', public_id: '123' },
```
mnacos, line 40: Minor thing, the nested `public_id` type should probably match the global `public_id` type, i.e. they should probably both be integers (json supports ints).

db/migrate/20210201142920_add_public_id_to_qualifications.rb
```diff
@@ -0,0 +1,6 @@
+class AddPublicIdToQualifications < ActiveRecord::Migration[6.0]
+  def change
+    add_column :application_qualifications, :constituent_grades, :jsonb
+    add_column :application_qualifications, :public_id, :bigint
```
mnacos, line 4: Following on from our chat, if you'd like to use postgres sequences to ensure you always have unique ids for each qualification, you could define a sequence-based default here:

```suggestion
    add_column :application_qualifications, :public_id, :bigint, default: -> { "nextval('qualifications_public_id_seq')" }
```

You may have to define this sequence first in your db migration, along the lines of:
```
  execute "CREATE SEQUENCE qualifications_public_id_seq;"
```
and you could also set its start to take into account any existing `public_id`s, if any:

```
  execute "SELECT setval('qualifications_public_id_seq', max(public_id)) FROM application_qualifications;"
```
This should go at the end of the migration. Setting a default will also populate all `public_id`s with unique ascending numbers at the point of migration. If you'd like to use different ranges for `public_id` there are ways to achieve this (e.g. you could set a starting point for the sequence with START when defining it).

For setting nested `public_id`s in the jsonb column post-migration, you'll need to be calling ActiveRecord::Base.connection.execute("SELECT nextval('qualifications_public_id_seq')") in the ruby code for each new public_id you set. But this way you get unique `public_id` numbers for both the top-level column and the nested `public_id` values.

Hope this helps!

mnacos (commented):
Added some suggestions around the use of postgres sequences to achieve what you want.

adr/0018-public-ids-for-qualifications.md
benswannack: True, will update and ensure we do this!

db/migrate/20210201142920_add_public_id_to_qualifications.rb
benswannack: Thanks, this is really useful and will make the public_id setting way more performant!

- We don't want to use the default here, because we don't necessarily want to set the value for public_id in the column (for structured qualifications it will be nil, and set inside the JSON instead)
- I'm going to add a service to handle setting the public_ids. The initial migration will call this on existing qualifications, and from then on, when a qualification is created it will be called. I can use the sequence there – it will just be less baked-in to the DB

PR #3948, opened 2021-01-28
Show all interviews against applications in support

mnacos:
## Context

We are about to launch the interviews feature and we'd like to see any interviews scheduled against applications in the support console for support reasons.

Add a `SupportInterface::InterviewDetailsComponent`. Show interviews at the end of each application choice, if present. This is not restricted to applications in the `interviewing` state on purpose.

image

## Guidance to review

Use the review app. This component has been dev-designed!

## Link to Trello card

https://trello.com/c/1fkhJzrW

## Things to check

### DIFF
--- a/app/components/support_interface/application_choice_component.rb
+++ b/app/components/support_interface/application_choice_component.rb
@@ -56,7 +56,7 @@ module SupportInterface
         rows << { key: 'Interviews', value: interview_blocks }
       end
 
-      rows.flatten
+      rows
     end
 
   private
--- a/app/components/support_interface/application_choice_component.rb
+++ b/app/components/support_interface/application_choice_component.rb
@@ -48,7 +48,15 @@ module SupportInterface
         rows << { key: 'API', value: 'This application hasn’t been sent to the provider yet, so it isn’t available over the API' }
       end
 
-      rows
+      if application_choice.interviews.present?
+        interview_blocks = application_choice.interviews.order('created_at').map do |interview|
+          render(SupportInterface::InterviewDetailsComponent.new(interview: interview))
+        end
+
+        rows << { key: 'Interviews', value: interview_blocks }
+      end
+
+      rows.flatten
     end
 
   private
--- a/app/components/support_interface/interview_details_component.rb
+++ b/app/components/support_interface/interview_details_component.rb
@@ -15,10 +15,6 @@ module SupportInterface
           value: interview.provider.name,
         },
         {
-          key: 'Meeting details',
-          value: interview.location,
-        },
-        {
           key: 'Created at',
           value: interview.created_at.to_s(:govuk_date_and_time),
         },
@@ -32,7 +28,7 @@ module SupportInterface
   private
 
     def interview_status
-      interview.cancelled_at.present? ? 'Cancelled' : 'Active'
+      interview.cancelled? ? 'Cancelled' : 'Active'
     end
   end
 end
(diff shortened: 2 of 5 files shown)


app/components/support_interface/interview_details_component.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+
+    def rows
+      [
+        {
+          key: 'Provider',
+          value: interview.provider.name,
+        },
+        {
+          key: 'Meeting details',
```
despo, line 18: Interviews have both a `location` and an `additional_information` field so using the title Meeting details to render the location can cause confusion. 

Both of these text fields could also contain new lines, paragraph breaks and links that should be preserved and  linked to as well, so we should  use the markdown renderer to display their content.

app/components/support_interface/interview_details_component.rb
```diff
@@ -0,0 +1,38 @@
...
+          key: 'Updated at',
+          value: interview.created_at.to_s(:govuk_date_and_time),
+        },
+      ]
+    end
+
+  private
+
+    def interview_status
+      interview.cancelled_at.present? ? 'Cancelled' : 'Active'
```
despo, line 35:
```suggestion
      interview.discarded? ? 'Cancelled' : 'Active'
```
We should use the [discard gem's API](https://github.com/jhawthorn/discard#discard-a-record) to check if an interview has been cancelled or not. Maybe we can alias the method on the interview model for clarity.

app/components/support_interface/interview_details_component.rb
mnacos: Good point, 'Meeting details' could be confusing. I think I'd prefer removing it altogether, as, from a support perspective, the key information is whether interviews have been scheduled for this application or not, whether some of them are cancelled etc. I'd go even further and say the support operator shouldn't have access to meeting details, e.g. zoom links, because links could be clicked and we definitely don't want that.

I tried to gather some requirements for this work and the key questions are: a. have interviews been scheduled, and b. do provider users choose to discard existing interviews and create new ones rather than modify existing ones. With 'Status', 'Created At' and 'Updated At' we should be able to tell.

app/components/support_interface/interview_details_component.rb
mnacos: Ok, I'll add a `cancelled?` alias to the model.

app/components/support_interface/application_choice_component.rb
```diff
@@ -48,7 +48,15 @@ def rows
...
-      rows
+      if application_choice.interviews.present?
+        interview_blocks = application_choice.interviews.order('created_at').map do |interview|
+          render(SupportInterface::InterviewDetailsComponent.new(interview: interview))
+        end
+
+        rows << { key: 'Interviews', value: interview_blocks }
+      end
+
+      rows.flatten
```
despo, line 59:
```suggestion
      rows
```

There was an issue on line 22 resolved [here](https://github.com/DFE-Digital/apply-for-teacher-training/commit/98c16ed80bb9b398c08360a37f6ff7f8f3ecf021) so we shouldn't need to flatten the rows.

app/models/interview.rb
```diff
@@ -2,6 +2,7 @@ class Interview < ApplicationRecord
   include Discard::Model
 
   self.discard_column = :cancelled_at
+  alias_method :cancelled?, :discarded?
```
despo, line 5: 👍🏼

despo (approved):
Looks good 👍🏼

app/components/support_interface/application_choice_component.rb
mnacos: Cool, removed

PR #3952, opened 2021-01-29
Remove null constraints from application_experiences table

TobyRet:
## Context

In order to implement the new restructured work history work flow we need to remove the null constraints on the `working_with_children` and `description` columns.

- Add db migration

# Guidance for review
The new restructured work history flow is behind a feature flag, so I think its ok to update the API in a future PR

## Link to Trello card

See https://trello.com/c/Vc9Hj802/2919-dev-add-the-add-a-job-page

## Things to check


db/schema.rb
```diff
@@ -57,9 +57,9 @@
     t.string "type", null: false
     t.string "role", null: false
     t.string "organisation", null: false
-    t.text "details", null: false
-    t.boolean "working_with_children", null: false
-    t.datetime "start_date", null: false
+    t.text "details"
+    t.boolean "working_with_children"
+    t.datetime "start_date"
```
TobyRet, line 62: mmm ... start_date should not have changed?

db/migrate/20210129122017_remove_null_constraints_from_application_experiences.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveNullConstraintsFromApplicationExperiences < ActiveRecord::Migration[6.0]
+  def change
+    change_column :application_experiences, :details, :text, null: true
```
tijmenb, line 3: `change_column_null` here could be preferable because you can't accidentally change a different attribute (like type). And I think it's reversible as well?

https://edgeapi.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-change_column_null

TobyRet: Yup - I've made the change

PR #3954, opened 2021-01-29
Restructured work experience - add a job

TobyRet:
## Context

As part of adding the new work history flow, we need to add the new 'Add a job' page.

- Create 'add a job' page.
- Allow users to create and update a job

## Guidance to review

- Make sure the 'restructured work history' feature flag is turned on
- Log into a candidate's application and go to `/candidate/application/restructured-work-history/new`
- Note that the new work history review page still needs to be wired up (separate ticket)
- deleting a job to be a separate ticket

## Screenshot

add_a_job

## Link to Trello card

https://trello.com/c/Vc9Hj802/2919-dev-add-the-add-a-job-page

## Things to check


app/views/candidate_interface/restructured_work_history/job/new.html.erb
```diff
@@ -0,0 +1,23 @@
+<% content_for :title, title_with_error_prefix(t('page_titles.add_job'), @job_form.errors.any?) %>
+<% if current_application.application_work_experiences.present? %>
+  <% content_for :before_content, govuk_back_link_to(candidate_interface_work_history_show_path) %>
```
JR-G, line 3: Just noticed on my local env - this is taking me back to the old work history review

app/controllers/candidate_interface/restructured_work_history/job_controller.rb
```diff
@@ -0,0 +1,81 @@
+module CandidateInterface
+  class RestructuredWorkHistory::JobController < RestructuredWorkHistory::BaseController
+    before_action :redirect_to_dashboard_if_submitted
```
davidgisbey, line 3: I think this can come out as base controller is already calling this method.

app/controllers/candidate_interface/restructured_work_history/job_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    def new
+      @job_form = RestructuredWorkHistory::JobForm.new
+    end
+
+    def create
+      @job_form = RestructuredWorkHistory::JobForm.new(job_form_params)
+
+      if @job_form.blank_form?
```
davidgisbey, line 12: I don't see any harm in redirecting blank forms, but I'm wondering if this is still needed as it was put in place to deal with the previous `add another` mechanism (people were clicking yes on add another job and had no way to get back to the review page)

config/locales/candidate_interface/restructured_work_history.yml
```diff
@@ -11,7 +11,35 @@ en:
...
+      commitment:
+        label: Was this job full time or part time?
+        full_time:
+          label: Full time (usually at least 35 hours a week)
+        part_time:
+          label: Part time
+      start_date:
+        label: When did you start this job?
+        hint_text: For example, 5 2018. If you cannot remember the exact month or year, enter an estimate.
+      start_date_unknown_checkbox: I'm not sure the exact month or year I started
```
davidgisbey, line 33: Channelling my inner Paul Lloyd here. 

`I’m not sure the exact month or year I started`

config/locales/candidate_interface/restructured_work_history.yml
```diff
@@ -11,7 +11,35 @@ en:
...
+      end_date:
+        label: When did you leave this job?
+        hint_text: For example, 5 2018. If you cannot remember the exact month or year, enter an estimate.
+      end_date_unknown_checkbox: I'm not sure the exact month or year I left
```
davidgisbey, line 37: `I’m not sure the exact month or year I left`

config/locales/candidate_interface/restructured_work_history.yml
```diff
@@ -21,3 +49,26 @@ en:
...
+              invalid: Enter a start date in the correct format
+              before: Enter a start date that is before the end date
+              in_the_future: Enter a start date that is not in the future
+            currently_working:
+              blank: Select if you are still working in this job
+            end_date:
+              invalid: Enter an end date in the correct format
+              in_the_future: Enter an end date that is not in the future
+            relevant_skills:
+              blank: Select if you use skills relevant to teaching in this job
```
davidgisbey, line 74: The error messaging for the bottom two fields are slightly off 

image

I think the radios also need to be validated for presence.

`  validates :role,
                :organisation,
                :commitment,
                :relevant_skills,
                :currently_working,
                presence: true`

app/forms/candidate_interface/restructured_work_history/job_form.rb
```diff
@@ -0,0 +1,116 @@
...
+          start_date_month: job.start_date&.month,
+          start_date_year: job.start_date&.year,
+          start_date_unknown: job.start_date_unknown,
+          end_date_day: job.end_date&.day || '',
+          end_date_month: job.end_date&.month || '',
+          end_date_year: job.end_date&.year || '',
+          end_date_unknown: job.end_date_unknown,
+          currently_working: job.currently_working,
+          relevant_skills: job.relevant_skills,
+        )
```
davidgisbey, line 52: i tested locally and this is working nicely 👌

davidgisbey:
Really great work Toby, just a few minor comments and it's good to go 👍

spec/forms/candidate_interface/restructured_work_history/job_form_spec.rb
```diff
@@ -0,0 +1,202 @@
...
+
+      it 'is invalid if not well-formed' do
+        form = described_class.new(end_date_month: '99', end_date_year: '2019', currently_working: false)
+
+        form.validate
+
+        expect(form.errors.full_messages_for(:end_date)).to eq(
+          ["End date #{t('activemodel.errors.models.candidate_interface/restructured_work_history/job_form.attributes.end_date.invalid')}"],
+        )
+      end
```
davidgisbey, line 83: Really thorough testing of the validations Toby 🙌 

I don't think it's a blocker for this, but we're now calling the methods from the `DateValidationHelper` in a few places in the application.

I wonder if we should really just be poritng the tests we have into a `date_validation_helper_spec.rb` file. At the moment i think we're calling it in a few places.

Also this exists 

image

I'll add a tech debt card though. No need to block this.

app/controllers/candidate_interface/restructured_work_history/job_controller.rb
TobyRet: Ah yes - that correct. I'll remove.

config/locales/candidate_interface/restructured_work_history.yml
TobyRet: Slightly scary but I see the error of my ways ...

config/locales/candidate_interface/restructured_work_history.yml
TobyRet: 👍

config/locales/candidate_interface/restructured_work_history.yml
TobyRet: Ah my bad. Have fixed this now

spec/forms/candidate_interface/restructured_work_history/job_form_spec.rb
TobyRet: Yeah I was not able to add `include_examples 'validation for an end date'` because in the new flow we only validate the end date if the user selects that they are no longer working in that role. Given this is a unique case I did not think it would be appropriate to update the includes specs to reflect this (e.g. via a flag)

davidgisbey (approved):
LGTM 🚀

PR #3957, opened 2021-01-31
Add structured rejection reasons to the ApplicationChoicesExport

davidgisbey:
## Context

Recently, we've given providers the opportunity to provide their rejection feedback in a structured way. This PR adds a structured rejection reasons column to the ApplicationChoicesExport. 

I discussed the granularity of details Mike wanted for this column with him and he would like the high level reasons (qualifications, candidate behaviour etc.). Later on, we will create a more detailed reasons for rejections export.

- Select the high-level rejection reasons 
- Put these rejection reasons into a new structured_rejection_reasons column

This is what the new column looks like.

image

## Guidance to review

I was interested to see how performant this was. I created 500 choices with structured rejection reasons. It took 2 seconds with and without the new column.

## Link to Trello card

https://trello.com/c/x8aB4hnp/2949-update-applicationchoicesexport-to-include-structured-rejection-reasons

## Things to check

### DIFF
--- a/app/services/support_interface/application_choices_export.rb
+++ b/app/services/support_interface/application_choices_export.rb
@@ -23,6 +23,7 @@ module SupportInterface
             offer_response: offer_response_interpretation(choice: choice),
             offer_response_at: choice.accepted_at || choice.declined_at,
             rejection_reason: choice.rejection_reason,
+            structured_rejection_reasons: format_structured_rejection_reasons(choice.structured_rejection_reasons),
           }
         end
       end
@@ -67,6 +68,25 @@ module SupportInterface
         .where('candidates.hide_in_reporting' => false)
         .where.not(submitted_at: nil)
         .order('submitted_at asc')
+    end
+
+    def format_structured_rejection_reasons(structured_rejection_reasons)
+      return nil if structured_rejection_reasons.blank?
+
+      select_high_level_rejection_reasons(structured_rejection_reasons)
+      .keys
+      .map { |reason| format_reason(reason) }
+      .join("\n")
+    end
+
+    def select_high_level_rejection_reasons(structured_rejection_reasons)
+      structured_rejection_reasons.select { |reason, value| value == 'Yes' && reason.include?('_y_n') }
+    end
+
+    def format_reason(reason)
+      reason
+      .delete_suffix('_y_n')
+      .humanize
     end
   end
 end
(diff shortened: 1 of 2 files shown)


app/services/support_interface/application_choices_export.rb
```diff
@@ -68,5 +69,27 @@ def relevant_applications
...
+      structured_rejection_reasons.select { |reason, value| value == 'Yes' && reason.include?('_y_n') }
+    end
+
+    def format_reason(reason)
+      reason
+      .split('_')
+      .reject { |string| %w[y n].include?(string) }
+      .join(' ')
+      .humanize
+    end
```
davidgisbey, line 93: Is it clear what's going on in these methods? There's quite a bit of munging going on. Happy to extract some logic out if needed.

app/services/support_interface/application_choices_export.rb
```diff
@@ -68,5 +69,24 @@ def relevant_applications
...
+      return nil if structured_rejection_reasons.blank?
+
+      select_high_level_rejection_reasons(structured_rejection_reasons)
+      .keys
+      .map { |reason| format_reason(reason) }
+      .join("\n")
+    end
+
+    def select_high_level_rejection_reasons(structured_rejection_reasons)
+      structured_rejection_reasons.select { |reason, value| value == 'Yes' && reason.include?('_y_n') }
```
davidgisbey, line 83: All the high-level rejection reasons have the `reason_y_n` format, so this allows us to filter based on that pattern

stevehook: I think it would be nice to have a `top_level_reason?` helper method somewhere. Then we could start to tidy up some of these 'metadata lookups'. You use `SupportInterface::SubReasonsForRejectionTableComponent::TOP_LEVEL_REASONS_TO_I18N_KEYS.keys` but that's not really the right places for the R4R model either.

app/services/support_interface/application_choices_export.rb
```diff
@@ -68,5 +69,24 @@ def relevant_applications
...
+    end
+
+    def format_reason(reason)
+      reason
+      .delete_suffix('_y_n')
+      .humanize
```
stevehook, line 89: There are labels for these in `reasons_for_rejection.yml` but not very logically organised (at least not to me). So again this is the best way to do this for now but something we might want to revisit.

stevehook (approved):
Looks good. Makes me even more convinced that we need a core R4R 'meta model' and a decent set of labels in the `config/locales`.

davidgisbey:
Agreed on everything. I'm going to merge this in because Mike needs it for research soon, but we should sort this 👍

PR #3961, opened 2021-02-01
Add 'Has interviews' filter to support console

mnacos:
## Context

For supporting and tracking usage of interviews feature.

Add a 'Has interviews' filter to support console.

image

Add support for `interviewing` state to `ProcessState`. This adds the following segment to our process docs:

image

## Guidance to review

Navigate to support, use the filter.

## Link to Trello card

https://trello.com/c/1fkhJzrW

## Things to check

### DIFF
--- a/app/models/support_interface/applications_filter.rb
+++ b/app/models/support_interface/applications_filter.rb
@@ -25,6 +25,10 @@ module SupportInterface
         application_forms = application_forms.where('phase IN (?)', applied_filters[:phase])
       end
 
+      if applied_filters[:interviews]
+        application_forms = application_forms.joins(application_choices: [:interviews])
+      end
+
       if applied_filters[:year]
         application_forms = application_forms.where('recruitment_cycle_year IN (?)', applied_filters[:year])
       end
@@ -33,7 +37,7 @@ module SupportInterface
     end
 
     def filters
-      @filters ||= [search_filter, search_by_application_choice_filter, year_filter, phase_filter]
+      @filters ||= [search_filter, search_by_application_choice_filter, year_filter, phase_filter, interviews_filter]
     end
 
   private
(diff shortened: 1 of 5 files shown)


app/models/support_interface/applications_filter.rb
```diff
@@ -25,6 +25,10 @@ def filter_records(application_forms)
         application_forms = application_forms.where('phase IN (?)', applied_filters[:phase])
       end
 
+      if applied_filters[:interviews]
+        application_forms = application_forms.joins(application_choices: [:interviews])
```
duncanjbrown, line 29: Could we perhaps have a test for this one? Just a unit test for the filter. I added a similar one in #3967

duncanjbrown (commented):
LGTM, could have 1 small test!

app/models/support_interface/applications_filter.rb
mnacos: Ok, added

duncanjbrown (approved):
Thank you!

AgaDufrat (approved):
The only non-dev question I have is about having a filter with just 1 checkbox. 
I've done it myself #3400 but later on @steventux had a chat with the content/design team and went with 2 checkboxes: https://github.com/DFE-Digital/apply-for-teacher-training/pull/3680#issuecomment-749519784

Nothing blocking but I'd add a card to Tech debt to make this consistent across the app.

mnacos:
> The only non-dev question I have is about having a filter with just 1 checkbox.
> I've done it myself #3400 but later on @steventux had a chat with the content/design team and went with 2 checkboxes: [#3680 (comment)](https://github.com/DFE-Digital/apply-for-teacher-training/pull/3680#issuecomment-749519784)
> 
> Nothing blocking but I'd add a card to Tech debt to make this consistent across the app.

Thanks for pointing this out Aga! I'll merge as is and we could revisit. In my mind, this is a temporary filter for tracking feature usage post-launch and we could decide to remove it after a while.

PR #3962, opened 2021-02-01
Ensure conditions status is correct for deferred offers

steventux:
## Context

Deferring an offer with pending conditions currently results in the conditions on the deferred offer conditions being marked as 'Met'. This is a display/rendering bug as there is no state saved on the conditions themselves.

- Update conditions component to inspect `status_before_deferral` when labelling conditions 'Pending' or 'Met'
- Add component spec.

## Guidance to review

Steps to reproduce original bug:

 -  View an 'Accepted' application with pending condition(s)
 -  Click on the 'Offer' tab
 -  Click 'Defer offer' link and confirm deferral
 -  View same application offer tab, conditions are now marked as 'Met' rather than pending.

## Link to Trello card

https://trello.com/c/a5spqa06/3309-conditions-pending-met-labelling-wrong-for-deferred-offers

## Things to check

### DIFF
--- a/app/components/provider_interface/conditions_component.rb
+++ b/app/components/provider_interface/conditions_component.rb
@@ -26,6 +26,8 @@ module ProviderInterface
     end
 
     def conditions_met?
+      return application_choice.status_before_deferral == 'recruited' if application_choice.status == 'offer_deferred'
+
       application_state.current_state >= :recruited
     end
 
--- a/spec/components/provider_interface/conditions_component_spec.rb
+++ b/spec/components/provider_interface/conditions_component_spec.rb
@@ -1,0 +1,50 @@
+require 'rails_helper'
+
+RSpec.describe ProviderInterface::ConditionsComponent do
+  describe 'rendered component' do
+    let(:conditions) { ['Fitness to teach check'] }
+
+    it 'renders the conditions' do
+      application_with_conditions_met = build(:application_choice, status: 'recruited', offer: { 'conditions' => conditions })
+      result = render_inline(described_class.new(application_choice: application_with_conditions_met))
+
+      expect(result.to_html).to include('Fitness to teach check')
+    end
+
+    it 'indicates whether conditions are met' do
+      application_with_conditions_met = build(:application_choice, status: 'recruited', offer: { 'conditions' => conditions })
+      result = render_inline(described_class.new(application_choice: application_with_conditions_met))
+
+      expect(result.css('.app-tag').text).to eq('Met')
+    end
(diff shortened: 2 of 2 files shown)


spec/components/provider_interface/conditions_component_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    let(:conditions) { ['Fitness to teach check'] }
+
+    it 'renders the conditions' do
+      application_with_conditions_met = build(:application_choice, status: 'recruited', offer: { 'conditions' => conditions })
+      result = render_inline(described_class.new(application_choice: application_with_conditions_met))
+
+      expect(result.to_html).to include('Fitness to teach check')
+    end
+
+    it 'indicates whether conditions are met' do
```
mnacos, line 14: How about one more scenario, 'indicates whether conditions are pending'? For non-deferred offers.

app/components/provider_interface/conditions_component.rb
```diff
@@ -26,6 +26,8 @@ def application_state
     end
 
     def conditions_met?
+      return false if application_choice.status_before_deferral&.eql?('pending_conditions')
```
mnacos, line 29: This is much better but how about also checking that current status is 'deferred'? This is because `status_before_deferral` is not reset when offers are re-instated and such offers will run into similar inconsistencies.

mnacos (commented):
This is a good catch but we also need to make the fix work for later stages, too. Suggestions below.

app/components/provider_interface/conditions_component.rb
mnacos: A candidate may have satisfied offer conditions before the deferral, in which case `status_before_deferral` will be `recruited`, but the provider may choose to re-instate the offer as `pending_conditions` because the candidate may require e.g. a new DBS check. In this case, the current code will actually return the right thing, but this feels slightly accidental.

In the somewhat less likely scenario of a `pending_conditions` state before the deferral but a `recruited` state when the provider reinstates the offer, I'd expect the current code to say `pending_conditions`.

steventux: Ah yes, good point, I didn't spot that the `status_before_deferral` field was not unset/reset when an offer was reconfirmed.
I've amended here https://github.com/DFE-Digital/apply-for-teacher-training/pull/3962/commits/a0db628d2a887385abfc964e55b7e7c3f9488193 along with an additional test.

mnacos (approved):
Great stuff

PR #3965, opened 2021-02-01
Migrate to use constituent grades

benswannack:
## Context
We need to migrate structured grades so that there is space in the data model to include public ids for each qualification.

Change all forms to use the new constituent_grades column for an application, rather than the structured_grades column

The structure will change from:
```
{ "English Lit": "A", "English Lang": "B" }
```
to:
```
{ "English Lit": { "grade": "A" },  "English Lang": { "grade": "B" } }
```

There will be a follow up PR to actually populate the public_id section of this JSON

## Guidance to review
There should be no change in behaviour!

## Link to Trello card

## Things to check

### DIFF
--- a/app/components/candidate_interface/gcse_qualification_review_component.rb
+++ b/app/components/candidate_interface/gcse_qualification_review_component.rb
@@ -96,7 +96,7 @@ module CandidateInterface
     end
 
     def present_constituent_grades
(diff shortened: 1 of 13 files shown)


benswannack:
This code depends on the rake task from the previous PR having been run

duncanjbrown (approved):
LGTM!

PR #3969, opened 2021-02-02
Reasons for rejection support search page

stevehook:
## Context

The Delivery team needs a summary view of structured R4R in Support UI so that they can observe trends and adjust/intervene accordingly

- [x] Add links to the rejection reasons dashboard to a new page that displays the matching applications

image

- [x] Use `SummaryListComponent` for results
- [x] Include free-text reasons (a) under top-level reasons and (b) sub-reasons
- [x] Use `mark` element to highlight the search reason
- [x] Add pagination
- [x] Extend breadcrumbs to show top-level reason when viewing sub-reasons

## Guidance to review

- The search page only displays the bare minimum, application choice and link, top-level reasons and sub-reasons. We can add more details about each application.
- Do the links on the dashboard (including the one in the `h2`) make sense? Is the layout of the search results logical?
- I'm getting a spec failing due to a cross module reference (lookup from a ProviderInterface data structure). Ideas about how to resolve this? Longer term we want to extract all the R4R 'metadata' to somewhere common.

## Link to Trello card

https://trello.com/c/aWsSjWyQ/2888-dev-r4r-support-search-page

## Things to check

### DIFF
--- a/app/queries/reasons_for_rejection_applications_query.rb
+++ b/app/queries/reasons_for_rejection_applications_query.rb
@@ -1,0 +1,31 @@
+class ReasonsForRejectionApplicationsQuery
+  attr_accessor :filters
+
+  def initialize(filters)
+    self.filters = filters
+  end
+
+  def call
+    application_choices = ApplicationChoice.where.not(structured_rejection_reasons: nil)
+
+    apply_filters(application_choices)
+  end
+
+private
+
+  def apply_filters(application_choices)
+    filters.each do |key, value|
+      if key =~ /_y_n$/
+        application_choices = application_choices.where(
+          "application_choices.structured_rejection_reasons->>'#{key}' = '#{value}'",
+        )
+      else
+        application_choices = application_choices.where(
+          "application_choices.structured_rejection_reasons->'#{key}' ? '#{value}'",
+        )
+      end
+    end
+
+    application_choices
+  end
+end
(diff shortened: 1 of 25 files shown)


paulrobertlloyd:
@stevehook Can we try the following:

* Use Summary List Component for the body of the summary card, rather than a table
* Pull out the title of the reason so that it’s the page title (I’ve currently updated this to say ‘Search’. Something more like this:

Screenshot 2021-02-03 at 11 24 01

app/queries/reasons_for_rejection_applications_query.rb
```diff
@@ -0,0 +1,31 @@
...
+    apply_filters(application_choices)
+  end
+
+private
+
+  def apply_filters(application_choices)
+    filters.each do |key, value|
+      if key =~ /_y_n$/
+        application_choices = application_choices.where(
+          "application_choices.structured_rejection_reasons->>'#{key}' = '#{value}'",
```
davidgisbey, line 20: This just accesses the blob vs accessing the attr right?

stevehook: It's a raw SQL condition on a property inside the JSONB column so it should translate to something like `WHERE application_choices.structured_rejection_reasons->>'qualifications_y_n' = 'Yes'` rather than loading the JSONB value into the model and then filtering in memory. Postgresql syntax for this stuff is a bit odd but it's quite powerful.

davidgisbey:
This is looking great. I've played around with it quite a bit and it's behaving really nicely. Once pagination is in place i think it's good to go 👍

davidgisbey (approved):
This if fantastic Steve. I gave it a good test locally and it's working beautifully 👌

PR #3978, opened 2021-02-03
Fix vulnerability in fallback support authentication

tijmenb:
## Context

Recently we've started using the `AuthenticationToken` that's used in support for for candidate authentication.

The method to look up the token doesn't scope the lookup to support users. This means that an attacker could request a sign in link on the candidate interface, and craft a URL to use the token in the link to gain access to the support console.

Properly scope the lookup.

## Guidance to review

Any new issues?

## Link to Trello card

## Things to check


davidgisbey (approved):
good catch

PR #3981, opened 2021-02-03
Hesa codes int to string

richardpattinson:
## Context

Ellucian have reported that as per API documentation, HESA data are all of data type string, in production currently they are coming through as integers. In order to ensure the documentation we have sent to HEI's is correct hesa codes need to be all strings. 

Hesa codes converted to strings, tests updated and DB backfilled. 

## Link to Trello card

https://trello.com/c/oH3KGPio/3308-hesa-data-types-are-string-but-passing-integer-values

## Things to check


malcolmbaig:
Code lgtm 👍🏽 . Will let a Provendor dev review as well and give it the final ✅.

Also, worth squashing the commits down. The context you've provided in the PR description would make a good commit message.

Some useful guidelines around commit messages here: https://thoughtbot.com/blog/5-useful-tips-for-a-better-commit-message

db/migrate/20210203130641_backfill_hesa_sex_and_ethnicity_codes.rb
```diff
@@ -0,0 +1,10 @@
+class BackfillHesaSexAndEthnicityCodes < ActiveRecord::Migration[6.0]
+  def change
+    not_nil = ApplicationForm.all.where.not(equality_and_diversity: nil)
+    not_nil.each do |application_form|
+      application_form.equality_and_diversity['hesa_sex'] = application_form.equality_and_diversity['hesa_sex'].to_s
+      application_form.equality_and_diversity['hesa_ethnicity'] = application_form.equality_and_diversity['hesa_ethnicity'].to_s
+      application_form.save
```
duncanjbrown, line 7: Just noticed that using `save` will trigger callbacks on the application form, which will in turn touch the application choices' `updated_at`, which will in turn cause them to appear as "changed" in the API, which we don't want.

We can avoid this by using `update_column` — suggest pulling out the whole `equality_and_diversity` hash, updating it, and then writing back? eg (untested)

```
equality_and_diversity = application_form.equality_and_diversity
equality_and_diversity['hesa_sex'] = equality_and_diversity['hesa_sex'].to_s
equality_and_diversity['hesa_ethnicity'] = equality_and_diversity['hesa_ethnicity'].to_s

application_form.update_column(:equality_and_diversity, equality_and_diversity)
```

duncanjbrown:
Change is great, one blocking comment! Might be worth writing a test for that transformation, just to be confident

db/schema.rb
```diff
@@ -10,6 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
+
```
AgaDufrat, line 13: we probably don't need this line here

AgaDufrat (approved):
Run locally and it's looking good 👌

PR #3982, opened 2021-02-03
Fix structured_grades data migration rake task

benswannack:
Update rake data migration to account for both hashes and JSON strings being stored in the structured_grades column

The constituent_grades will now be populated by hashes rather than JSON strings.

## Guidance to review
I think we should prefer hashes over JSON strings and more importantly be consistent, happy to hear counterpoints to that though!

## Link to Trello card

## Things to check

### DIFF
--- a/lib/tasks/backfill_constituent_grades.rake
+++ b/lib/tasks/backfill_constituent_grades.rake
@@ -3,10 +3,18 @@ task backfill_constituent_grades: :environment do
   structured_qualifications = ApplicationQualification.where(constituent_grades: nil).where.not(structured_grades: nil)
 
   structured_qualifications.each do |qualification|
-    grades = JSON.parse(qualification.structured_grades)
+    grades = get_structured_grades_hash(qualification)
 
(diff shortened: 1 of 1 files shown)


lib/tasks/backfill_constituent_grades.rake
```diff
@@ -3,10 +3,18 @@ task backfill_constituent_grades: :environment do
   structured_qualifications = ApplicationQualification.where(constituent_grades: nil).where.not(structured_grades: nil)
 
   structured_qualifications.each do |qualification|
-    grades = JSON.parse(qualification.structured_grades)
+    grades = get_structured_grades_hash(qualification)
 
-    constituent_grades = grades.transform_values { |grade| { grade: grade } }.to_json
```
benswannack, line 8: This line now means we store hashes and not JSON strings.
So we don't have to parse it when reading, and don't have to convert to json when writing.

(This will require me to change how constituent_grades are read/written, but that's all in a future PR anyway)

duncanjbrown (approved):

PR #3984, opened 2021-02-03
Add interview emails

AgaDufrat:
## Context

Add candidate email templates for new, edited and cancelled interviews and amend the relevant services to trigger emails

https://bat-design-history.netlify.app/manage-teacher-training-applications/interviews-iteration-3

Add all templates and emails. 
Sending emails when an interview is updated will need to be implemented once the service is added.

## Guidance to review

<kbd>Screenshot 2021-02-03 at 16 16 51</kbd>

<kbd>Screenshot 2021-02-03 at 16 16 59</kbd>

<kbd>Screenshot 2021-02-03 at 16 17 05</kbd>

## Link to Trello card

https://trello.com/c/hEhNASC5/3242-interview-emails

## Things to check

### DIFF
--- a/spec/mailers/candidate_mailer_spec.rb
+++ b/spec/mailers/candidate_mailer_spec.rb
@@ -242,4 +242,37 @@ RSpec.describe CandidateMailer, type: :mailer do
       'course start' => 'September 2021',
     )
   end
+
+  context 'Interview emails' do
+    let(:provider)  { create(:provider, name: 'Hogwards') }
+    let(:interview) do
+      create(:interview,
+             date_and_time: Time.zone.local(2021, 1, 15, 9, 30),
+             location: 'Hogwarts Castle',
+             additional_details: 'Bring your magic wand for the spells test',
+             provider: provider)
+    end
+    let(:application_choice_with_interview) { interview.application_choice }
+
+    before do
+      build_stubbed(:application_form,
+                    first_name: 'Fred',
+                    candidate: candidate,
+                    application_choices: [application_choice_with_interview])
+    end
+
+    describe '.new_interview' do
+      let(:email) { mailer.new_interview(application_choice_with_interview, interview) }
+
+      it_behaves_like(
+        'a mail with subject and content',
+        'Interview arranged - Hogwards',
+        'greeting' => 'Dear Fred,',
+        'details' => 'You have an interview with Hogwards',
+        'interview date and time' => '15 January 2021 at 9:30am',
+        'interview location' => 'Hogwarts Castle',
+        'additional interview details' => 'Bring your magic wand for the spells test',
+      )
+    end
+  end
 end
--- a/spec/mailers/candidate_mailer_spec.rb
+++ b/spec/mailers/candidate_mailer_spec.rb
@@ -288,5 +288,17 @@ RSpec.describe CandidateMailer, type: :mailer do
         'additional interview details' => 'Bring your magic wand for the spells test',
       )
     end
+
+    describe '.interview_cancelled' do
+      let(:email) { mailer.interview_cancelled(application_choice_with_interview, interview, 'We recruited someone else') }
+
+      it_behaves_like(
+        'a mail with subject and content',
+        'Interview cancelled - Hogwards',
+        'greeting' => 'Dear Fred,',
+        'details' => 'Hogwards has cancelled the interview on 15 January 2021 at 9:30am',
+        'cancellation reason' => 'We recruited someone else',
+      )
+    end
   end
 end
--- a/spec/mailers/candidate_mailer_spec.rb
+++ b/spec/mailers/candidate_mailer_spec.rb
@@ -274,5 +274,19 @@ RSpec.describe CandidateMailer, type: :mailer do
         'additional interview details' => 'Bring your magic wand for the spells test',
       )
     end
+
+    describe '.interview_updated' do
+      let(:email) { mailer.interview_updated(application_choice_with_interview, interview) }
+
+      it_behaves_like(
+        'a mail with subject and content',
+        'Interview details updated - Hogwards',
+        'greeting' => 'Dear Fred,',
+        'details' => 'Hogwards has updated the details of the interview',
+        'interview date and time' => '15 January 2021 at 9:30am',
+        'interview location' => 'Hogwarts Castle',
+        'additional interview details' => 'Bring your magic wand for the spells test',
+      )
+    end
   end
 end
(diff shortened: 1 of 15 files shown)


benswannack (approved):
Looking good!

spec/mailers/candidate_mailer_spec.rb
```diff
@@ -242,4 +242,93 @@
       'course start' => 'September 2021',
     )
   end
+
+  describe '.new_interview' do
+    let(:email) { mailer.new_interview(@application_choice, @interview) }
+
+    before do
+      @interview = create(:interview,
```
despo, line 250: can we define interview and application choice using let instead.

spec/mailers/candidate_mailer_spec.rb
```diff
@@ -242,4 +242,93 @@
...
+      @application_choice = @interview.application_choice
+      build_stubbed(
+        :application_form,
+        first_name: 'Fred',
+        candidate: candidate,
+        application_choices: [@application_choice],
+      )
+    end
+
+    it 'has the correct subject and content' do
```
despo, line 264: Can we reuse the shared example similar to the other specs for consistency.

AgaDufrat: good point, will update!

spec/services/cancel_interview_spec.rb
```diff
@@ -45,5 +45,11 @@
       ])
       expect(associated_audit.audited_changes['cancellation_reason']).to eq([nil, 'There is a global pandemic going on'])
     end
+
+    it 'sends an email to the candidate', sidekiq: true do
+      CancelInterview.new(service_params).save!
+
+      expect(ActionMailer::Base.deliveries.first['rails-mail-template'].value).to eq('interview_cancelled')
+    end
```
despo, line 53: Can we merge this with the previous spec. 

Reason being the time it takes to create an application choice. As they are both mandatory when cancelling an interview, we can just test them as part of the same spec.

spec/mailers/candidate_mailer_spec.rb
```diff
@@ -14,9 +14,20 @@
   let(:candidate) { build_stubbed(:candidate) }
   let(:application_choices) { [build_stubbed(:application_choice)] }
   let(:dbd_application) { build_stubbed(:application_choice, :dbd) }
+  let(:interview) do
+    create(:interview,
+           date_and_time: Time.zone.local(2021, 1, 15, 9, 30),
+           location: 'Hogwarts Castle',
+           additional_details: 'Bring your magic wand for the spells test')
+  end
+  let(:application_choice_with_interview) { interview.application_choice }
```
despo, line 23: I think to make it easier to see this setup we should create a new context for all interview related specs and move this setup in it.

spec/mailers/candidate_mailer_spec.rb
```diff
@@ -14,9 +14,20 @@
...
 
   before do
     magic_link_stubbing(candidate)
+    build_stubbed(:application_form,
+                  first_name: 'Fred',
+                  candidate: candidate,
+                  application_choices: [application_choice_with_interview])
```
despo, line 30: Similar to above, This should only happen within an interview email related context, as alternatively this will create a stub application form before running each specs, which is not necessary.

spec/mailers/candidate_mailer_spec.rb
```diff
@@ -242,4 +253,56 @@
       'course start' => 'September 2021',
     )
   end
+
+  describe '.new_interview' do
+    let(:email) { mailer.new_interview(application_choice_with_interview, interview) }
+
+    before do
+      @provider_name = application_choice_with_interview.offered_course.provider.name
```
despo, line 261: This can be moved to a let within the interview block, which should also enable to remove the duplicate entries specified below.

spec/mailers/candidate_mailer_spec.rb
```diff
@@ -242,4 +253,56 @@
...
+      'interview location' => 'Hogwarts Castle',
+      'additional interview details' => 'Bring your magic wand for the spells test',
+    )
+  end
+
+  describe '.interview_updated' do
+    let(:email) { mailer.interview_updated(application_choice_with_interview, interview) }
+
+    before do
+      @provider_name = application_choice_with_interview.offered_course.provider.name
```
despo, line 279: Same as above.

spec/mailers/candidate_mailer_spec.rb
```diff
@@ -242,4 +253,56 @@
...
+      'interview location' => 'Hogwarts Castle',
+      'additional interview details' => 'Bring your magic wand for the spells test',
+    )
+  end
+
+  describe '.interview_cancelled' do
+    let(:email) { mailer.interview_cancelled(application_choice_with_interview, interview, 'We recruited someone else') }
+
+    before do
+      @provider_name = application_choice_with_interview.offered_course.provider.name
```
despo, line 297: Same as above.

despo (approved):
Thanks for spending time to improve the specs!
I left some minor feedback about some additional things that could be improved, but overall it looks great 💯

spec/mailers/candidate_mailer_spec.rb
AgaDufrat: I tried that before but for some reason I'm getting
```
'provider_name' is not available on an example group (e.g. a `describe` or `context` block). 
It is only available from within individual examples (e.g. `it` blocks) or from constructs that 
run in the scope of an example (e.g. `before`, `let`, etc).
```
when I defined it with `let` underneath `let(:application_choice_with_interview)`

despo: I didn't notice that the variable was being referenced from inside the shared examples when I reviewed! let declarations are not available from within shared examples due to potential issues that may be caused, such as records are created outside example transactions.

However I think we can still avoid requiring the variable by just defining the provider and not doing any evaluations within the examples below.

```
 let(:provider)  { create(:provider, name: 'Brighthurst Technical College') }
 let(:interview) do
    create(:interview,
           date_and_time: Time.zone.local(2021, 1, 15, 9, 30),
           location: 'Hogwarts Castle',
           additional_details: 'Bring your magic wand for the spells test',
           provider: provider)
 end
```

This will also require updating the interview factory (and also removing the invalid condition that I added, as there will always be an application choice!!) 

```
 factory :interview do
    application_choice

    date_and_time { 7.business_days.from_now }
    location { [Faker::Address.full_address, 'Link to video conference'].sample }
    additional_details { [nil, 'Use staff entrance', 'Ask for John at the reception'].sample }

    after(:build) do |interview|
      interview.application_choice.status = 'interviewing'
      interview.provider ||= interview.application_choice.offered_course.provider
    end
  end

```

PR #3985, opened 2021-02-03
Enable GCSE year editing via support

duncanjbrown:
## Context

A few support requests have come in for this recently

Create an edit-GCSE form in support and link from the view.

Slightly icky `if` in the component but I don't think it's _too_ bad? In future could maybe improve by passing down a flag through the QualificationsComponent into the GcseQualificationCardsComponent to put it in `support mode` or sth. Or split and duplicate the components.

Screenshot 2021-02-03 at 20 00 50
Screenshot 2021-02-03 at 20 00 29

## Link to Trello card

Support tickets

## Things to check

### DIFF
--- a/app/components/gcse_qualification_cards_component.html.erb
+++ b/app/components/gcse_qualification_cards_component.html.erb
@@ -50,6 +50,10 @@
                 <% end %>
               </dl>
             <% end %>
+            <% if in_support_console? %>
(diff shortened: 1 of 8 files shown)


malcolmbaig (approved):
Fine with the if statement, one to refactor later if needed 👍🏽

PR #3989, opened 2021-02-04
Set public IDs for qualifications

benswannack:
Migrate to include a sequence to generate unique public ids quickly
Add a service that fills in public ids for qualifications
Add a rake task to backfill qualification public IDs

## Guidance to review
Per commit

## Link to Trello card

## Things to check


app/services/set_public_id_for_qualification.rb
```diff
@@ -0,0 +1,34 @@
+class SetPublicIdForQualification
+
+  def initialize(qualification:)
+    @qualification = qualification
+  end
+
+  def call(preferred_id_value: nil)
+    top_level_public_id = @qualification.public_id.presence || preferred_id_value.presence || next_available_id
```
benswannack, line 8: This line means:
- If a public_id has already been set, don't touch it – this ensures consistency for the API
- We can set a value for the ID manually (as the rake task does, so we can enforce the id is the same in the API after the change)
- Otherwise, generate a new unique ID

db/migrate/20210203160529_add_qualification_public_id_sequence.rb
```diff
@@ -0,0 +1,10 @@
+class AddQualificationPublicIdSequence < ActiveRecord::Migration[6.0]
+  def up
+    max_qualification_id = ApplicationQualification.maximum(:id)
+    create_sequence :qualifications_public_id_seq, start: max_qualification_id * 2
```
benswannack, line 4: Starting the sequence high enough to avoid duplicating and qualification IDs from the db. Twice the current max value seemed safe

lib/tasks/backfill_qualification_public_ids.rake
```diff
@@ -0,0 +1,6 @@
+desc 'Sets the public_id for qualifications that do not yet have one'
+task backfill_qualification_public_ids: :environment do
+  ApplicationQualification.where(public_id: nil).each do |qualification|
```
benswannack, line 3: We will have to run this multiple times, to ensure everything is backfilled. Second run should be very short

app/services/set_public_id_for_qualification.rb
```diff
@@ -0,0 +1,33 @@
+class SetPublicIdForQualification
```
benswannack, line 1: Second thoughts about putting this in a service, it feels like it belongs on the model. That way I can call it from the model whenever a qualification is updated

app/models/application_qualification.rb
```diff
@@ -141,4 +141,30 @@ def composite_equivalency_details
...
+
+private
+
+  def set_public_ids_for_constituent_grades(top_level_public_id)
+    return if constituent_grades.blank?
+
+    grades_with_ids = constituent_grades.transform_values.with_index do |grade, index|
+      next grade if grade['public_id'].present?
+
+      public_id = index.zero? ? top_level_public_id : next_available_public_id
```
benswannack, line 160: This is a bit unclear and prone to issues we've not thought of.

It's here because the method is trying to do two things:
- Account for the initial migration, where we need to copy across the db ids
- Generate new ids for new qualifications

I'm going to split up the above

tijmenb (approved):

app/services/backfill_qualification_public_id.rb
```diff
@@ -0,0 +1,26 @@
+class BackfillQualificationPublicId
+  def initialize(qualification)
+    @qualification = qualification
+  end
+
+  def call
+    if @qualification.constituent_grades.present?
```
benswannack, line 7: Ah this is wrong, since triple science GCSEs will be treated differently.

We should set the top level public_id on triple science ones since the grades for that subject are just glued together in the API like `AAB`

duncanjbrown (approved):
Looks good, I'm happy!

PR #3993, opened 2021-02-04
Edit an interview

steventux:
## Context

Providers need to be able to change the details of saved interviews.

- Adds edit, check and update actions which use the existing `InterviewWizard`.
- Allows the normal wizard 'check' step to reroute the user depending on whether we are creating a new interview or editing.

image

## Guidance to review

TODO: Integrate with interview email PR https://github.com/DFE-Digital/apply-for-teacher-training/pull/3984

## Link to Trello card

https://trello.com/c/aqr0qBqt/3240-editing-an-interview

## Things to check


config/locales/interview.yml
```diff
@@ -5,13 +5,23 @@ en:
...
+      edit:
+        title: Change interview details
+        interview_preferences:
+          title:  Candidate interview preferences
       check:
         page_title: Set up Interview
         title: Check and send interview details
         confirm: Send interview details
         cancel: Cancel
+        update:
```
despo, line 17: I think the update block is a level too deep. However, update is an action, so do we even need this? Line 23 should cover it.

config/routes.rb
```diff
@@ -635,10 +635,12 @@
 
       resources :notes, only: %i[index show new create], as: :application_choice_notes
 
-      resources :interviews, only: %i[new index], as: :application_choice_interviews do
+      resources :interviews, only: %i[new edit index], as: :application_choice_interviews do
         collection do
           get '/new/check', to: 'interviews#check'
           get '/confirm', to: 'interviews#commit'
+          get '/check/:id', to: 'interviews#check', as: :check
```
despo, line 642: These should be in the member block as they apply to a single interview, and the update should be a POST

```suggestion
          get :check
          post :update
```

config/routes.rb
```diff
@@ -635,10 +635,12 @@
...
+          get '/update/:id', to: 'interviews#update', as: :update
```
despo, line 643:
```suggestion
           put :update
```

spec/system/provider_interface/update_interview_spec.rb
```diff
@@ -0,0 +1,91 @@
+require 'rails_helper'
+
+RSpec.describe 'A provider user changes interview details' do
```
despo, line 3: Consider merging these steps in the [manage interview spec](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/spec/system/provider_interface/provider_manages_application_interviews_spec.rb#L36) after the creation and before the cancellation steps as we will then be testing all aspects of this feature via the interface and ensure everything works well together.

app/views/provider_interface/interviews/_form.html.erb
```diff
@@ -0,0 +1,19 @@
...
+      <%= f.govuk_radio_button :provider_id, provider.id, label: { text: provider.name } %>
+    <% end %>
+  <% end %>
+<% end %>
+
+<%= f.govuk_text_area :location, label: { size: 'm' } %>
+
+<%= f.govuk_text_area :additional_details, label: { size: 'm' } %>
+
+<%= f.govuk_submit 'Continue' %>
```
despo, line 19:
```suggestion
<%= f.govuk_submit t('continue') %>
```

app/controllers/provider_interface/interviews_controller.rb
```diff
@@ -20,7 +20,16 @@ def new
       @wizard.save_state!
     end
 
+    def edit
+      @interview = @application_choice.interviews.find(params[:id])
+
+      @wizard = InterviewWizard.from_model(store, @interview, 'input')
```
despo, line 26: Do we need `from_model`? Can we not just do
```
@wizard = InterviewWizard.new(store, @interview.attributes.merge(current_step: 'input', application_choice: interview.application_choice, provider: interview.provider))
```

app/forms/provider_interface/interview_wizard.rb
```diff
@@ -104,6 +104,21 @@ def clear_state!
       @state_store.delete
     end
 
+    def self.from_model(store, interview, step = 'input')
+      wizard = new(store, { current_step: step })
```
despo, line 108: I was wondering if we could simplify this somehow so that we don't need to worry if we make any additions to the model, but I don't think there is any straightforward way to do it unless we switch `attr_accessor` to `attributes` and then grab `attribute_names`. 

Maybe something to keep in mind and do as part of the wizard work.

despo (approved):
Left some feedback. It looks good, I think it raises some good questions about how we use form models and if we should instead be using `attributes` to specify any parameters but nothing that needs to be addressed in the PR.

I think it would be good however if we merge the two interview system specs.

config/routes.rb
steventux: :+1: Should `/confirm` also be moved to as a POST? I didn't think to question why this was a GET at the time.

despo: Yes, it should be a post!

spec/system/provider_interface/update_interview_spec.rb
steventux: :+1: https://github.com/DFE-Digital/apply-for-teacher-training/pull/3993/commits/e09a5048bd9652981dd9cec3ef3c39162515339b

app/views/provider_interface/interviews/_form.html.erb
steventux: :+1: https://github.com/DFE-Digital/apply-for-teacher-training/pull/3993/commits/555105222e11aecf8ac7e144ea254a76419cc838

config/routes.rb
steventux: :+1: https://github.com/DFE-Digital/apply-for-teacher-training/pull/3993/commits/00bd67a555c31f91758853fd2daf472b767f1923

config/routes.rb
steventux: :+1: https://github.com/DFE-Digital/apply-for-teacher-training/pull/3993/commits/00bd67a555c31f91758853fd2daf472b767f1923

config/routes.rb
steventux: :+1: https://github.com/DFE-Digital/apply-for-teacher-training/pull/3993/commits/00bd67a555c31f91758853fd2daf472b767f1923

app/helpers/view_helper.rb
```diff
@@ -49,6 +49,17 @@ def govuk_button_link_to(body, url, html_options = {}, &_block)
...
+    html_options = {
+      class: prepend_css_class('govuk-button', html_options[:class]),
+      role: 'button',
+      data: { module: 'govuk-button' },
+      draggable: false,
+    }.merge(html_options)
+
+    button_to(name, options, html_options)
+  end
+
```
despo, line 62: 👍🏼 I hadn't seen this helper before. It seems like it could do with some refactoring so that govuk components are moved into their own helper. I'll add a card on the tech debt boart.

app/views/provider_interface/interviews/check.html.erb
```diff
@@ -1,16 +1,21 @@
-<% content_for :browser_title,  t('.page_title')  %>
+<% translation_prefix = @interview.present? ? '.update' : '' %>
```
despo, line 1: Consider moving this on the controller, maybe on a helper method. It's easy to miss variables that are setup in views.

spec/system/provider_interface/update_interview_spec.rb
despo: Can we remove this spec now?

despo (approved):
Other than removing spec/system/provider_interface/update_interview_spec.rb it looks great :shipit:

steventux:
@despo update spec removed here https://github.com/DFE-Digital/apply-for-teacher-training/pull/3993/commits/72569c6b1179f5c5bc9ccf0c1542c9b8172e069d
Also moved the translation prefix setup here https://github.com/DFE-Digital/apply-for-teacher-training/pull/3993/commits/2f60909ece602c4e4f157e3441405d9b25da876a

despo:
@steventux great work. Thanks :shipit:

PR #3995, opened 2021-02-04
Add constants for GIT contact numbers and URLs

paulrobertlloyd:
## Context

We reference a number of contact values that relate to Get into teaching, which we should store in one place.

* Adds constants for GIT’s beta website URL and phone number, opening times and other GIT URLs we use
* Uses those constants throughout the application
* Removes guidance around right to work for the form that’s shown in support

## Guidance to review

This PR was reviewed once when it was in draft, but I’ve made a few more changes since, so worth another gander, I reckon.

## Link to Trello card

https://trello.com/c/wpgn9P98

## Things to check

### DIFF
--- a/app/components/personal_statement_component.html.erb
+++ b/app/components/personal_statement_component.html.erb
@@ -9,7 +9,7 @@
     <h2 class="govuk-heading-m">Why do you want to be a teacher?</h2>
     <p class="govuk-body">Use this section to showcase your motivation, commitment and teaching potential, backing up your answer with specific examples.</p>
     <p class="govuk-body">Give providers an insight into your personality by writing honestly and thoughtfully. Avoid cliché and write in clear, correct, concise English.</p>
-    <p class="govuk-body">You can ask your training provider for examples written by successful applicants, or register with <a href="https://beta-getintoteaching.education.gov.uk/" class="govuk-body">Get into Teaching</a> for help from a teacher training adviser.</p>
+    <p class="govuk-body">You can ask your training provider for examples written by successful applicants, or register with <%= govuk_link_to t('service_name.get_into_teaching'), t('get_into_teaching.url') %> for help from a teacher training adviser.</p>
     <p class="govuk-body govuk-!-font-weight-bold">Suggested topics to cover include:</p>
     <ul class="govuk-list govuk-list--bullet">
       <li>why you want to be a teacher</li>
--- a/app/views/support_interface/application_forms/right_to_work_or_study/edit.html.erb
+++ b/app/views/support_interface/application_forms/right_to_work_or_study/edit.html.erb
@@ -12,18 +12,18 @@
         <% end %>
         <%= f.govuk_radio_button :right_to_work_or_study, 'no', label: { text: 'Not yet – I will need to apply for permission to work or study in the UK' } do %>
           <p class="govuk-body">
-            Visit <%= govuk_link_to 'Get Into Teaching', 'https://beta-getintoteaching.education.gov.uk/international-candidates' %> for guidance on immigration.
+            Visit <%= govuk_link_to t('service_name.get_into_teaching'), t('get_into_teaching.url_international_candidates') %> for guidance on immigration.
           </p>
           <p class="govuk-body">
-            You can also <%= govuk_link_to 'register to talk to an adviser', 'https://beta-getintoteaching.education.gov.uk/helping-you-become-a-teacher' %> if you need help.
+            You can also <%= govuk_link_to 'register to talk to an adviser', t('get_into_teaching.url_get_an_advisor') %> if you need help.
           </p>
         <% end %>
         <%= f.govuk_radio_button :right_to_work_or_study, 'decide_later', label: { text: 'I do not know' } do %>
           <p class="govuk-body">
-            Visit <%= govuk_link_to 'Get Into Teaching', 'https://beta-getintoteaching.education.gov.uk/international-candidates' %> for guidance on immigration.
+            Visit <%= govuk_link_to t('service_name.get_into_teaching'), t('get_into_teaching.url_international_candidates') %> for guidance on immigration.
           </p>
           <p class="govuk-body">
-            You can also <%= govuk_link_to 'register to talk to an adviser', 'https://beta-getintoteaching.education.gov.uk/helping-you-become-a-teacher' %> if you need help.
+            You can also <%= govuk_link_to 'register to talk to an adviser', t('get_into_teaching.url_get_an_advisor') %> if you need help.
           </p>
         <% end %>
       <% end %>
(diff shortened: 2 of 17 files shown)


despo:
@paulrobertlloyd maybe consider putting these details in `config/locales/en.yml` as it makes them easier to find and update, and they shouldn't really be part of the configuration.

malcolmbaig:
Good idea 👍🏽 

Agree with @despo on using i18n strings.

> There’s no way of using constants in Markdown files, or is there? I haven’t updated those, so these values are still not entirely centralised

We could potentially update [this helper](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/helpers/content_helper.rb) to process ERB, thereby allowing calls to `I18n` in the markdown, eg - https://github.com/DFE-Digital/apply-for-teacher-training/pull/3999

paulrobertlloyd:
Thanks @despo, seems like localised strings is the way to go!

app/views/candidate_mailer/_get_help.text.erb
```diff
@@ -1,7 +1,7 @@
 # Get help
 
-For advice on becoming a teacher, call Get Into Teaching for free on 0800 389 2500 or chat to an adviser online:
+For advice on becoming a teacher, call <%= service_name.get_into_teaching %> for free on <%= get_into_teaching.tel %> or chat to an adviser online:
 
-https://beta-getintoteaching.education.gov.uk/#talk-to-us
+<%= get_into_teaching.url %><%= get_into_teaching.path_online_chat %>
```
duncanjbrown, line 5: I think I'd have a single key for `get_into_teaching.online_chat_url`? If there's duplication across multiple GIT urls then it'll be adjacent in the YAML, and not knowing whether to including a slash when sticking paths on domains makes me anxious

malcolmbaig:
https://github.com/DFE-Digital/apply-for-teacher-training/pull/3999 now merged, can use erb tags in .md files.

paulrobertlloyd:
🎉

duncanjbrown (approved):
💯

paulrobertlloyd:
Just going to review some of the URLs with @gregknight1, then this’ll be good to go - thanks all!

app/views/support_interface/application_forms/right_to_work_or_study/edit.html.erb
```diff
@@ -10,22 +10,8 @@
...
-          </p>
-        <% end %>
-        <%= f.govuk_radio_button :right_to_work_or_study, 'decide_later', label: { text: 'I do not know' } do %>
-          <p class="govuk-body">
-            Visit <%= govuk_link_to 'Get Into Teaching', 'https://beta-getintoteaching.education.gov.uk/international-candidates' %> for guidance on immigration.
-          </p>
-          <p class="govuk-body">
-            You can also <%= govuk_link_to 'register to talk to an adviser', 'https://beta-getintoteaching.education.gov.uk/helping-you-become-a-teacher' %> if you need help.
-          </p>
-        <% end %>
```
paulrobertlloyd, line 28: Note: have removed GIT content from this form as it’s seen by support agents and this guidance is intended for candidates. Leaving it here means it’ll probably fall out of sync with what’s shown in the candidate interface.

app/components/personal_statement_component.html.erb
```diff
@@ -6,30 +6,17 @@
...
+    <p class="govuk-body">You can include:</p>
+      <ul class="govuk-list govuk-list--bullet">
+      <li>your interest in the subject or age group</li>
+      <li>the demands and rewards of teaching</li>
+      <li>the personal qualities that would make you a good teacher</li>
+      <li>how you could contribute to a school outside of the classroom</li>
+      <li>any past experience working with children or young people, and what you learnt</li>
+      <li>your thoughts on welfare and education</li>
     </ul>
+    <p class="govuk-body">If you need help, you can speak to a <%= govuk_link_to t('service_name.get_into_teaching'), t('get_into_teaching.url_get_an_advisor') %> adviser. They’re all experienced teachers who can give you advice on applying for teacher training.</p>
```
paulrobertlloyd, line 19: The content shown in the provider interface has fallen out of step with what’s shown to candidates. Not the first time this has happened, either. Surely there must be a better way to keep these in sync… maybe for starters we can add some comments in the code? @duncanjbrown

duncanjbrown: This is a fact of life rn unfortunately, unless we keep track of all previous versions — applications submitted in the past ought to have the guidance from then, rather than now. We accepted this as a drawback when we did this in the first place, but we should def update this in the provider interface too now.

paulrobertlloyd: Ah, the whole historical aspect. Gah!

> we should def update this in the provider interface too now

Just to be clear, that’s what this PR does.

paulrobertlloyd: Having spoken to the ProVendor design team, we’re actually remove this guidance. So in this PR I’ll just limit the change to pointing to the new GIT constants. Much better.

config/locales/en.yml
```diff
@@ -8,6 +8,18 @@ en:
     manage: Manage teacher training applications
     support: Support for Apply
     api: Apply for teacher training API
+    get_into_teaching: Get Into Teaching
+  get_into_teaching:
+    # TODO: Update to 0800 389 2500 when GIT leaves beta
+    tel: 0800 389 2501
+    opening_times: "Monday to Friday, 8.30am to 5pm (except public\u00A0holidays)"
```
duncanjbrown, line 15: Why `\00A0` instead of a space?

config/locales/en.yml
```diff
@@ -8,6 +8,18 @@ en:
...
+    # TODO: Update domain names when GIT leaves beta
+    url: https://beta-getintoteaching.education.gov.uk
+    url_applying: https://beta-getintoteaching.education.gov.uk/steps-to-become-a-teacher#step-5
+    url_get_an_advisor: https://beta-adviser-getintoteaching.education.gov.uk
+    url_international_candidates: https://beta-getintoteaching.education.gov.uk/international-candidates
+    url_training_with_a_disability: https://beta-getintoteaching.education.gov.uk/guidance/become-a-teacher-in-england#training-to-teach-if-you-have-a-disability
+    url_online_chat: https://beta-getintoteaching.education.gov.uk/#talk-to-us
```
duncanjbrown, line 22: ✅ all URLs checked and correct

duncanjbrown (approved):
Approving (pending @Tomcowan1's content sign-off) with one last q!

config/locales/en.yml
paulrobertlloyd: Prevents an orphan appearing in the footer (which is coming in a follow-up PR).

So instead of this:

```
Monday to Friday, 8.30am to 5pm (except public
holidays)
```

we get this:

```
Monday to Friday, 8.30am to 5pm (except
public holidays)
```

duncanjbrown: Nice. Orphan! I haven't heard that expression since subediting class. Skills I never used, obviously

PR #3999, opened 2021-02-05
Update render_content_page to support ERB

malcolmbaig:
## Context
Allows use of translation strings in content that's currently defined in
markdown.

Prompted by https://github.com/DFE-Digital/apply-for-teacher-training/pull/3995

Add an ERB processing step just prior to rendering the markdown.

## Guidance to review

- Any issues with this approach?
- Change to accessibility.md is just for example, will be removed if this PR is merged. Primary motivation is for things like contact details which are constant throughout the app.

## Things to check

### DIFF
--- a/app/helpers/content_helper.rb
+++ b/app/helpers/content_helper.rb
@@ -1,6 +1,8 @@
 module ContentHelper
   def render_content_page(page_name)
-    @converted_markdown = GovukMarkdown.render(File.read("app/views/content/#{page_name}.md")).html_safe
+    raw_content = File.read("app/views/content/#{page_name}.md")
+    content_with_erb_tags_replaced = ApplicationController.renderer.render(inline: raw_content)
+    @converted_markdown = GovukMarkdown.render(content_with_erb_tags_replaced).html_safe
     @page_name = page_name
     render 'rendered_markdown_template'
   end


paulrobertlloyd:
This is awesome! So many strings in Markdown files would benefit from this change 😀

app/helpers/content_helper.rb
```diff
@@ -1,6 +1,8 @@
 module ContentHelper
   def render_content_page(page_name)
-    @converted_markdown = GovukMarkdown.render(File.read("app/views/content/#{page_name}.md")).html_safe
+    raw_content = File.read("app/views/content/#{page_name}.md")
+    content_with_erb_tags_replaced = ERB.new(raw_content).result
```
duncanjbrown, line 4: How about

```suggestion
    content_with_erb_tags_replaced = ERB.new(raw_content).result(binding)
```

This lets us use `t` directly in the template without needing to reference `I18n`?

duncanjbrown: Ah, this breaks `t!` but plain `t` works. I wonder what's going on there. It's good to raise on missing translations (as we do in `application.rb`) but that doesn't seem to affect this manual ERB rendering. Anyway, would it be nice not to have to remember to call `I18n.t!` instead of `I18n.t`, at risk of letting through "Translation missing..." strings into rendered templates?

duncanjbrown: OK, how about this?

```suggestion
    content_with_erb_tags_replaced = ApplicationController.renderer.render(inline: raw_content)
```

and then plain `t` in the template

From https://api.rubyonrails.org/classes/ActionController/Renderer.html

malcolmbaig: @duncanjbrown missed your last message before pushing a new commit! Let's go with your one, much clearer.

duncanjbrown (approved):

PR #4001, opened 2021-02-05
Fix bug: Correctly change name of all variable references

benswannack:
## Context
Missed some mentions of a variable. Will follow up with a system test


davidgisbey (approved):
makes sense 👍

PR #4003, opened 2021-02-05
Get support content in candidate emails

raamSoftwire:
## Context

To make guidance around getting support consistent.

I have replaced the inconsistent text blocks. On the ticket, it suggested creating a template of some sort so that this block was automatically added to each email by default. But at the moment it is not present in every email, so I'm not sure yet. Very happy for input here.

New guidance to include:

image

Example email with new text:

image

## Guidance to review

1. Review by commit
2. Check for typos in "_get_support.text.erb". I couldn't copy the text so I had to type it out
3. Have I missed any other templates? 

## Link to Trello card

https://trello.com/c/23av0zGE/2973-get-support-content-in-candidate-emails

## Things to check


paulrobertlloyd (approved):
This makes me so unbelievably happy 😅

Once this is merged I will rebase my #3995 branch and update the common include so that it references the new GIT constants.

raamSoftwire:
It might be worth waiting until https://github.com/DFE-Digital/apply-for-teacher-training/pull/3995 is merged before merging this. That way I can resolve the conflicts in a few of the email templates.

app/views/candidate_mailer/_get_support.text.erb
```diff
@@ -1,3 +1,7 @@
 # Get support
 
-Contact [becomingateacher@digital.education.gov.uk](mailto:becomingateacher@digital.education.gov.uk) if you have problems applying online or want to give feedback.
+You can chat to a Get Into Teaching adviser online for help and advice:
+
+[https://beta-getintoteaching.education.gov.uk/#talk-to-us](https://beta-getintoteaching.education.gov.uk/#talk-to-us)
```
davidgisbey, line 5: Sorry i hadn't picked this up by looking at the ticket before, but does it seem right that we are sending everyone to GIT rather than giving linking them to our services support? Let's check on Monday.

davidgisbey:
Looks good, but I just have one question. Why are we linking everyone to GIT? Shouldn't issues be coming directly to us?

raamSoftwire:
> Why are we linking everyone to GIT? Shouldn't issues be coming directly to us?

From the Trello card 

> In short, yes, all emails should point to GIT. It's following a decision by Rachael and Sarah. GIT already have processes in place to point candidates to our Support team when they need to.

raamSoftwire:
Tom is just having a look at the 3 outstanding UCAS emails to see what should be done in those cases.

davidgisbey:
Great. Thanks for clearing that up 👍

raamSoftwire:
I've now pushed the changes for those last 3 emails. Now every candidate email contains the `get_support` footer so I will now refactor the code so that it is added by default.

raamSoftwire:
Are we missing an `offer_accepted` candidate mailer preview?

raamSoftwire:
> Are we missing an `offer_accepted` candidate mailer preview?

Yes we were, I have now added it in.

app/mailers/candidate_mailer.rb
```diff
@@ -1,4 +1,6 @@
 class CandidateMailer < ApplicationMailer
+  layout 'candidate_email_with_footer'
```
paulrobertlloyd, line 2: Does this need to be a tiny bit clearer?

```suggestion
  layout 'candidate_email_with_support_footer'
```

paulrobertlloyd: Not even sure ‘footer’ is the right word here… but can’t think of a good alternative.

Tomcowan1 (approved):

PR #4005, opened 2021-02-05
Add more filtering to the Support /providers view

duncanjbrown:
## Context

As requested on the Support board filter by what sort of body accredits courses.
Also introduce filtering by provider type.

Add those two filters:

Screenshot 2021-02-05 at 14 12 50

Make two improvements to existing behaviour:

- if there's a search term and some of the results have been filtered out, show a message saying "n search results were filtered out"
- add an 'any' option to the "onboarding stage" to avoid having to test various combinations of those checkboxes

## Link to Trello card

https://trello.com/c/n5Wzfm2U/157-ability-to-filter-courses-by-accredited-body-ie-scitt-or-hei

### DIFF
--- a/app/models/support_interface/providers_filter.rb
+++ b/app/models/support_interface/providers_filter.rb
@@ -2,23 +2,19 @@ module SupportInterface
   class ProvidersFilter
     attr_reader :applied_filters
     ONBOARDING_STAGES = {
-      'synced' => 'Courses synced',
-      'dsa_signed' => 'DSA signed',
+      'synced_only' => 'With synced courses',
+      'dsa_signed_only' => 'With signed DSAs',
     }.freeze
 
     PROVIDER_TYPES = {
       'lead_school' => 'School Direct',
       'scitt' => 'SCITT',
       'university' => 'HEI',
-    }
+    }.freeze
 
     RATIFIED_BY = {
       'scitt' => 'SCITT',
       'university' => 'HEI',
-    }
-
-    DEFAULT_STATE = {
-      onboarding_stages: ONBOARDING_STAGES.keys,
     }.freeze
 
     def initialize(params:)
@@ -28,7 +24,7 @@ module SupportInterface
         :ratified_by,
         :onboarding_stages,
         :q,
-      ).presence || DEFAULT_STATE
+      )
     end
 
     def filters
@@ -68,16 +64,12 @@ module SupportInterface
         @search_count = 0
       end
 
-      if applied_filters[:onboarding_stages]&.include?('synced')
+      if applied_filters[:onboarding_stages]&.include?('synced_only')
         providers = providers.where(sync_courses: true)
-      else
-        providers = providers.where(sync_courses: false)
       end
 
-      if applied_filters[:onboarding_stages]&.include?('dsa_signed')
+      if applied_filters[:onboarding_stages]&.include?('dsa_signed_only')
         providers = providers.joins(:provider_agreements)
-      else
-        providers = providers.includes(:provider_agreements).where(provider_agreements: { provider_id: nil })
       end
 
       if applied_filters[:provider_types].present?
--- a/app/models/support_interface/providers_filter.rb
+++ b/app/models/support_interface/providers_filter.rb
@@ -12,6 +12,11 @@ module SupportInterface
       'university' => 'HEI',
     }
 
+    RATIFIED_BY = {
+      'scitt' => 'SCITT',
+      'university' => 'HEI',
+    }
+
     DEFAULT_STATE = {
       onboarding_stages: ONBOARDING_STAGES.keys,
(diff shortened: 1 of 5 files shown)


app/views/support_interface/providers/index.html.erb
```diff
@@ -10,6 +10,12 @@
       </tr>
     </thead>
 
+    <% if @filter.search_results_filtered_out_count.positive? %>
+      <p class="govuk-body">
+      <%= pluralize(@filter.search_results_filtered_out_count, 'search result') %> were filtered out.
```
benswannack, line 15: This pluralise doesn't read well if it's singular, maybe drop the "were"?

benswannack (approved):
Nice, one minor thing inline :)

app/models/support_interface/providers_filter.rb
```diff
@@ -4,59 +4,114 @@ class ProvidersFilter
...
-        providers = providers.joins(:provider_agreements)
-      else
-        providers = providers.includes(:provider_agreements).where(provider_agreements: { provider_id: nil })
+      if applied_filters[:provider_types].present?
+        providers = providers.where(provider_type: applied_filters[:provider_types])
+      end
+
+      if applied_filters[:ratified_by].present?
+        ratifiers = Provider.where(provider_type: applied_filters[:ratified_by])
+        providers = providers.joins(:training_provider_permissions)
```
benswannack, line 92: This currently filters out anyone who falls into the category, but ratifies their own courses. I'm not sure how this is going to be used, is that intentional?

spec/models/support_interface/providers_filter_spec.rb
```diff
@@ -1,31 +1,96 @@
...
+      create(:provider)
+
       providers = Provider.all
       filter = described_class.new(params: { onboarding_stages: %w[dsa_signed] })
 
       expect(filter.filter_records(providers)).to eq [provider_with_signed_dsa]
     end
 
-    it 'excludes providers when the boxes aren’t ticked' do
+    it 'can ignore DSA and synced status' do
```
benswannack, line 25: It's a bit unexpected that ticking no boxes in the onboarding status is actually filtering out results. I don't think any of our other filters do that. I guess the need is for a user to be able to filter to see who has done no setup. Would it be weird to change the wording to be "providers who haven't done the thing" since that is what is interesting here?

app/models/support_interface/providers_filter.rb
duncanjbrown: As I understand it this shouldn't show any providers who aren't ratified by a separate accredited body.

spec/models/support_interface/providers_filter_spec.rb
duncanjbrown: 100% agree! I find these checkboxes a bit painful, but they are useful. And `Any` is definitely a hack.

Unchecked checkboxes would still be filtering out results, even if we reversed their meaning.

I thought radio buttons might be better — `Any`, `Courses synced and DSA signed`, and `Courses synced but DSA not signed`?

Soon we'll sync all courses so this will become a bit simpler.

benswannack: If we reversed the meaning of the checkboxes, couldn't it work like: Nothing checked – show all, Check "Haven't signed the DSA" – Filter down to providers who haven't signed it, Check "Haven't synced courses" – Filter down to providers who do not have their courses synced. I think the any is implicit when nothing is checked (at least in our implementation of filters)

Not too important, but that feels more consistent with other filter use to me.

app/views/support_interface/providers/index.html.erb
duncanjbrown: oops!

spec/models/support_interface/providers_filter_spec.rb
duncanjbrown: We used to have a "synced" and "unsynced" providers tab. That was a bit annoying as you had to flip between them to find providers. It was good, thought, not to have to look at the unsynced ones all the time though as there were lots of them, they didn't tell us much, and they made the list very long.

re DSAs, the original need from the Support board was

```
we need to know when a provider has signed the DSA so that we can open their courses. At the moment, we have to manually check which providers have signed the DSA by going into each individual provider. This is time consuming. 
```

So I wanted to enable filtering by having-DSA too, which is what led me to the current implementation.

How about we do as you say — ie filters behave more like elsewhere — but we keep them positive but use the word "only", ie:

- synced courses only
- signed DSA only

benswannack: Oh I see, that solution makes sense then!

duncanjbrown:
Updated and I hope improved @benswannack

benswannack:
This looks good now! 😄

PR #4006, opened 2021-02-05
Allow bulk creation of test applications for ratified courses via API

mnacos:
## Context

We already allow API calls to generate test data for a given provider. Providers that ratify many courses would also like to test visibility of applications to courses they ratify, not run, via the API.

Allow providers to specify `for_ratified_courses` when calling `/test-data/generate`. If this param is passed, then `count` and `courses_per_application` are still honoured, but test applications are created against random courses the provider ratifies, not runs. Note: API keys are scoped to a single provider.

image

## Guidance to review

Inspect code changes. Is the documentation easy to understand?

If a provider tries to use `/test-data/clear`, this won't delete applications generated `for_ratified_courses`.

## Link to Trello card

https://trello.com/c/9HLwGYID

## Things to check

### DIFF
--- a/app/lib/test_applications.rb
+++ b/app/lib/test_applications.rb
@@ -4,12 +4,19 @@ class TestApplications
 
   attr_reader :time
 
-  def generate_for_provider(provider:, courses_per_application:, count:)
+  def generate_for_provider(provider:, courses_per_application:, count:, for_ratified_courses: false)
+    courses_to_apply_to = if for_ratified_courses
+                            GetCoursesRatifiedByProvider.call(provider: provider)
+                          else
+                            Course.current_cycle.includes(:course_options)
+                              .joins(:course_options).distinct.open_on_apply
+                              .where(provider: provider)
+                          end
     1.upto(count).flat_map do
       create_application(
-        recruitment_cycle_year: 2021,
+        recruitment_cycle_year: RecruitmentCycle.current_year,
         states: [:awaiting_provider_decision] * courses_per_application,
-        courses_to_apply_to: Course.current_cycle.includes(:course_options).joins(:course_options).distinct.open_on_apply.where(provider: provider),
+        courses_to_apply_to: courses_to_apply_to,
       )
     end
   end
--- a/app/views/api_docs/pages/release_notes.md
+++ b/app/views/api_docs/pages/release_notes.md
@@ -1,3 +1,9 @@
+## 5th February
+
+The following experimental/sandbox endpoint has been updated:
+
+- `/test-data/generate` now accepts an optional `for_ratified_courses` query param. If this parameter is supplied and set to a non-empty string, applications will be generated for courses the organisation awards, not runs. This means subsequent calls to `/test-data/clear` will NOT delete these applications.
+
 ## 29th January
 
 - The documented enum values for `Reference.referee_type` have been corrected to remove commas and replace `school-based` with `school_based`.
--- a/config/vendor-api-experimental.yml
+++ b/config/vendor-api-experimental.yml
@@ -25,6 +25,11 @@ paths:
           default: 1
           minimum: 1
           maximum: 3
+      - name: for_ratified_courses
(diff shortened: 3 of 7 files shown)


app/views/api_docs/pages/release_notes.md
```diff
@@ -1,3 +1,9 @@
+## 5th February
+
+The following experimental/sandbox endpoint has been updated:
+
+- `/test-data/generate` now accepts an optional `for_ratified_courses` query param. If this parameter is supplied and set to a non-empty string, applications will be generated for courses the organisation awards, not runs. This means subsequent calls to `/test-data/clear` will NOT delete these applications.
```
benswannack, line 5: Is the word "awards" used elsewhere to describe courses that an org ratifies? For consistency, maybe using the word "ratifies" is better?

The last point is interesting. This sort of leaves them with no way to remove applications that they generated themselves. Not sure what the best way to get around that would be

config/vendor-api-experimental.yml
```diff
@@ -25,6 +25,11 @@ paths:
           default: 1
           minimum: 1
           maximum: 3
+      - name: for_ratified_courses
+        description: Generate applications for courses your organisation awards, not runs. Use a non-empty string value, e.g. 't'
```
benswannack, line 29: Is it worth writing the example here like this to avoid any confusion?
```suggestion
        description: Generate applications for courses your organisation awards, not runs. Use a non-empty string value, e.g. 'for_ratified_courses=true'
```

benswannack: I think it would be clearer to force a value like `true` for this (since we don't yet have any other flag parameters like this to copy). It may be slightly confusing that `for_ratified_courses=false` will activate the setting.

benswannack (approved):
I've made a few improvement suggestions, but looks good overall :)

app/lib/test_applications.rb
```diff
@@ -4,12 +4,19 @@ class ZeroCoursesPerApplicationError < RuntimeError; end
 
   attr_reader :time
 
-  def generate_for_provider(provider:, courses_per_application:, count:)
+  def generate_for_provider(provider:, courses_per_application:, count:, for_ratified_courses: false)
+    courses_to_apply_to = if for_ratified_courses
+                            GetCoursesRatifiedByProvider.call(provider: provider)
```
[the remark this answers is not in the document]
mnacos: Yes, it'd be nice but I'm guessing that many, if not all, of the ratified courses on sandbox may not be `open_on_apply` so creating a uniform `GetCoursesByProvider` to cover all cases may actually get in the way of testing. But we could revisit this as part of the 'change offer' work.

app/lib/test_applications.rb
```diff
@@ -4,12 +4,19 @@ class ZeroCoursesPerApplicationError < RuntimeError; end
...
+                          else
+                            Course.current_cycle.includes(:course_options)
+                              .joins(:course_options).distinct.open_on_apply
+                              .where(provider: provider)
+                          end
     1.upto(count).flat_map do
       create_application(
-        recruitment_cycle_year: 2021,
```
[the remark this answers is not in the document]
mnacos: I actually like passing a list of courses to sample from. But, by doing this, you could pass a list of courses spanning multiple recruitment cycles. Furthermore, we have a `recruitment_cycle_year` as an attribute of `ApplicationForm`, which could be derived by the course perhaps, but may not be obvious in cases of `apply_2` or deferrals.

config/vendor-api-experimental.yml
mnacos: Yes, `for_ratified_courses=false` is problematic. I'll change this to require `true`.

app/views/api_docs/pages/release_notes.md
mnacos: Adding more text to explain the use-case better (HEIs ratifying School Direct courses).

PR #4007, opened 2021-02-05
Conform to Sentry's user_context schema

duncanjbrown:
## Context

We're not able to search or filter on the values we set in `Raven.user_context` because it doesn't allow unlisted params: https://docs.sentry.io/product/sentry-basics/guides/enrich-data/#types-of-data.

Only set `id` param in `user_context`. Send everything else in `extra_context`.

## Link to Trello card

Noticed while doing support trying to track down an error seen by a candidate.


tijmenb (approved):

tijmenb (commented):
Very minor: maybe something like `Candidate#12312` and `ProviderUser#981236` instead of `candidate_18237`?

PR #4009, opened 2021-02-05
Use path helpers in Markdown files

paulrobertlloyd:
## Context

Now that we can use ERB in Markdown files (#3999), we should use path helpers to ensure links in these files reference the paths they relate to in the application’s routes.

* Use `bat_contact_mail_to` helper in Markdown files
* Use path helpers in Markdown files
* Fix incorrect link to sign up form on QA 😱
* Use RedCarpet’s `with_toc_data` option to auto-generate heading ids

## Guidance to review

* ~~I’m using `candidate_interface_path` to link to Apply/candidate path (this route either points to GOV.UK start page or the application route depending on the environment). However, this appears to be rendering a link to the Find feedback page, oddly? Any idea why this might be @davidgisbey~~ Fixed in #4018
* ~~I was going to update the Markdown in API docs in a similar fashion, but looks like those pages use a different Markdown rendering method; should this be a globally available helper, used by the different interfaces @duncanjbrown?~~ Investigated this, and while it is possible to use the same helper with a little modification, doesn’t deliver much value as most of the links in the API docs wouldn’t benefit from using path helpers.

## Things to check

### DIFF
--- a/app/views/content/accessibility.md
+++ b/app/views/content/accessibility.md
@@ -10,7 +10,7 @@ We have taken steps to make this service accessible:
 
 ## What to do if you cannot access parts of this website
 
-If you need information on this website in a different format, such as accessible PDF, large print, easy read, audio recording or braille, please email us on <becomingateacher@digital.education.gov.uk>.
+If you need information on this website in a different format, such as accessible PDF, large print, easy read, audio recording or braille, please email us on <%= bat_contact_mail_to %>.
 
 We’ll consider your request and respond to you within 5 working days.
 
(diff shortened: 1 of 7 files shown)


app/helpers/content_helper.rb
```diff
@@ -1,7 +1,7 @@
 module ContentHelper
   def render_content_page(page_name)
     raw_content = File.read("app/views/content/#{page_name}.md")
-    content_with_erb_tags_replaced = ApplicationController.renderer.render(inline: raw_content)
+    content_with_erb_tags_replaced = ApplicationController.renderer.render(inline: raw_content, with_toc_data: true)
```
duncanjbrown, line 4: This is `ApplicationController::Renderer` rather than `RedCarpet` — I don't think `with_toc_data` will do anything here? https://api.rubyonrails.org/classes/ActionController/Renderer.html#method-i-render

paulrobertlloyd: Well, it worked! Or maybe it was working already… I can double check.

paulrobertlloyd: Ha! Turns out this was enabled anyway… just didn’t need to add manual ids in the markdown document.

duncanjbrown (approved):

PR #4010, opened 2021-02-06
Require confirmation when flipping a feature flag

duncanjbrown:
## Context

We accidentally turned on a feature in production: https://ukgovernmentdfe.slack.com/archives/CQA64BETU/p1612522574165000

When flipping a feature flag, require confirmation via a JS popup which prominently displays the current `HostingEnvironment`.

Requires `rails/ujs`, which this PR adds to Support.

Not tested because we don't have JS testing machinery in our system specs and it's not worth the cost and complexity of adding it for a stock Rails feature we're using for progressive enhancement in Support.

## Guidance to review

Try it on the review app.

### DIFF
--- a/app/frontend/packs/application-support.js
+++ b/app/frontend/packs/application-support.js
@@ -1,5 +1,6 @@
 require.context("govuk-frontend/govuk/assets");
 
+import Rails from '@rails/ujs';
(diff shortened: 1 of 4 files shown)


davidgisbey (approved):
lgtm! Works nicely locally

PR #4012, opened 2021-02-07
Remove international address feature flag

davidgisbey:
## Context

A while ago we went from using a free text area to a structured series on inputs for international candidates addresses. 

While pulling this flag out I realised that the implementation was not in the shape agreed by Duncan and me. I should have picked this up at review/filled the card in properly so my bad.

The shape it should is at the moment:

```
address_line1
address_line2
address_line3
address_line4
postcode
```
the shape it should be (and is being exposed as on the API)

```
address_line1
address_line2
address_line3
address_line4
```
We decided on the above as we really have no way of knowing what international addresses look like throughout the world. Populating postcode with values that may well not actually be postcodes felt wrong so we went with as generic values as possible.

This PR does a few things.

1. Removes the feature flag
2. Updates the views to use the new design
3. Cleans up the data for international candidates who have populated the postcode value

Candidate UI
|Before|After|
|------------|------------|
|image|image|

Support UI
|Before|After|
|------------|------------|
|image|image|

## Guidance to review

Does this all make sense?

## Link to Trello card

https://trello.com/c/SUdMaIUt/2982-dev-remove-international-addresses-feature-flag

## Things to check


app/models/application_form.rb
```diff
@@ -277,7 +277,10 @@ def nationalities
   def full_address
     if international?
       [
-        international_address,
+        address_line1,
+        address_line2,
+        address_line3,
+        address_line4,
```
stevehook, line 283: Why is postcode not included here? It was on the form for international addresses but are we removing it now?

stevehook (commented):
Just had one question about the `postcode` for international addresses. It looks like this PR removes it in a few places though it's still there on the form. Should international addresses use the `postcode` attribute?

app/forms/candidate_interface/contact_details_form.rb
```diff
@@ -3,11 +3,10 @@ class ContactDetailsForm
     include ActiveModel::Model
 
     attr_accessor :phone_number, :address_line1, :address_line2, :address_line3,
-                  :address_line4, :postcode, :address_type, :country, :international_address
+                  :address_line4, :postcode, :address_type, :country
 
     validates :address_line1, :address_line3, :postcode, presence: true, on: :address, if: :uk?
-    validates :address_line1, presence: true, on: :address, if: ->(form) { form.international? && FeatureFlag.active?(:international_addresses) }
-    validates :international_address, presence: true, on: :address, if: ->(form) { form.international? && !FeatureFlag.active?(:international_addresses) }
+    validates :address_line1, presence: true, on: :address, if: ->(form) { form.international? }
```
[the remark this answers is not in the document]
davidgisbey: It does 💯

app/models/application_form.rb
davidgisbey: In the discussion i had with Duncan, we said we would just use 4 inputs with no labels for international addresses as we weren't sure what shape the data would come back in. Just applying the changes now.

davidgisbey: image

app/views/candidate_interface/contact_details/address/edit.html.erb
```diff
@@ -6,8 +6,9 @@
     <%= form_with model: @contact_details_form, url: candidate_interface_contact_information_edit_address_path, method: :patch do |f| %>
       <%= f.govuk_error_summary %>
 
-      <% if @contact_details_form.uk? || FeatureFlag.active?(:international_addresses) %>
         <%= f.govuk_fieldset legend: { text: t('page_titles.address'), size: 'xl', tag: 'h1' } do %>
```
stevehook, line 9: indentation if off from here.

db/migrate/20210208105018_clean_up_international_addresses_data.rb
```diff
@@ -0,0 +1,26 @@
+class CleanUpInternationalAddressesData < ActiveRecord::Migration[6.0]
+  def change
+    application_forms = ApplicationForm
+                        .international
+                        .where
+                        .not(postcode: nil)
+
+    audit_comment = 'Backfilled while removing the international addresses feature flag. International addresses should not be able to provide a postcode.'
+
+    application_forms.each do |application_form|
```
stevehook, line 10:
```suggestion
    application_forms.find_each do |application_form|
```
though we may not yet have enough international addresses to be worried about this.

stevehook (approved):
Looks good to me. I think it makes sense to tackle dropping the `postcode` and otherwise simplifying international addresses as part of this PR. A couple of nits but nothing to stop a merge. :shipit:

db/migrate/20210208105018_clean_up_international_addresses_data.rb
davidgisbey: still good to get in the habit. Thanks 👍

app/views/candidate_interface/contact_details/address/edit.html.erb
```diff
@@ -36,9 +37,28 @@
             width: 10,
             autocomplete: 'postal-code',
           ) %>
-        <% end %>
       <% else %>
-        <%= f.govuk_text_area :international_address, label: { text: t('page_titles.address'), size: 'xl' }, autocomplete: 'address' %>
+        <%= f.govuk_text_field(
+          :address_line1,
+          label: { text: t('application_form.contact_details.address_line1.international.label') },
```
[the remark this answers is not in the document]
davidgisbey: nice 👌

PR #4017, opened 2021-02-08
Use public IDs for qualifications in the API

benswannack:
## Context
The final PR for this change!

Qualifications need a public id to display in the API. This cannot be the db id because some qualifications in the API share the same row in the db due to structured qualifications. 

Set public ids for qualifications that are new, or being edited. 
Use public_id rather than id when showing qualifications on the API

## Guidance to review
Are there any edge cases I missed?

## Link to Trello card

## Things to check

### DIFF
--- a/app/models/application_qualification.rb
+++ b/app/models/application_qualification.rb
@@ -58,6 +58,8 @@ class ApplicationQualification < ApplicationRecord
 
   audited associated_with: :application_form
 
+  after_save :set_public_id
+
   def missing_qualification?
     qualification_type == 'missing'
   end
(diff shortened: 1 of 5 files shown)


app/models/application_qualification.rb
```diff
@@ -58,6 +58,8 @@ class ApplicationQualification < ApplicationRecord
 
   audited associated_with: :application_form
 
+  after_save :set_public_id
```
tijmenb, line 61: Perhaps try to do this with `before_save`.

duncanjbrown (approved):

PR #4019, opened 2021-02-08
Revert "Require confirmation when flipping a feature flag"

duncanjbrown:
Reverts DFE-Digital/apply-for-teacher-training#4010

This caused all forms in support to be remote=true, which breaks them


cwrw (commented):
~is it possible to set `config.action_view.form_with_generates_remote_forms` to false to avoid the default behaviour or rails ujs?~ actually ignore this as we don't want a global setting since this is used elsewhere :D

tijmenb (approved):

PR #4020, opened 2021-02-08
Exclude interviews from activity log for now

mnacos:
## Context

Adding associated models (for interviews) seems to have killed the activity log events query, leading to statement timeouts. The original query takes about 100ms so this PR reverts changes to the activity log, which is fine because the interviews feature has not been launched yet.

Revert activity log events query to previous version. Mark associated spec as pending.

## Guidance to review

Easy-peasy

## Link to Trello card

https://trello.com/c/B79CEysQ

## Things to check


duncanjbrown (approved):
👌 thank you @mnacos !!

steventux (approved):
:+1: It seems like a lot of what is going on in this query is `ApplicationChoice` specific which makes the polymorphism of `Audited::Audit` redundant. It would be good to look at how we can refactor some of this to deal with any audit instance.

PR #4023, opened 2021-02-08
Remove 'Structured reasons for rejection' feature flag

AgaDufrat:
## Context

Feature is live in production

Remove a feature flag

## Guidance to review

Is there a better way of handling `rejection_reason` coming from the API and `structured_rejection_reasons` from Manage in the RejectApplication class?

Have I missed something? 

## Things to check


app/forms/provider_interface/pick_response_form.rb
```diff
@@ -15,6 +15,9 @@ def redirect_attrs
         attrs[:controller] = 'offer_changes'
         attrs[:action] = 'edit_offer'
         attrs[:step] = decision.delete_prefix('edit_')
+      elsif decision == 'new_reject'
```
steventux, line 18: Assume this is necessary because we remove the feature flag dependant redirect?

steventux (approved):
:+1: All seems to work fine on preview app.

app/forms/provider_interface/pick_response_form.rb
AgaDufrat: yea, I could redirect in the controller with
```
def new_reject
  redirect_to provider_interface_reasons_for_rejection_initial_questions_path(@application_choice)
end
```

but I though it may be better to do it here and avoid an extra step.

steventux: :+1: I think this is the more obvious way to do it given the rest of the code uses the same method.

PR #4024, opened 2021-02-08
Update API docs with field lengths summary

SamJamCul:
## Context

Request from a provider for us to show the character limits on various fields

Added a line to the api docs about the field lengths summary 

Before
image

After
image

And added a table at the bottom of the docs page which shows the Fields and their corresponding limits, where applicable
image

## Guidance to review

Head to the /api-docs/reference page

## Link to Trello card

https://trello.com/c/5oTJXnEe/3328-update-api-docs-add-field-lengths-summary

## Things to check

### DIFF
--- a/app/views/api_docs/reference/reference.html.erb
+++ b/app/views/api_docs/reference/reference.html.erb
@@ -26,6 +26,9 @@
       <% end %>
     </ol>
   </li>
+  <li class="app-contents-list__list-item app-contents-list__list-item--parent">
+    <%= govuk_link_to 'Field lengths summary', '#field-lengths', class: 'app-contents-list__link' %>
+  </li>
 </ol>
 
 <hr class="govuk-section-break govuk-section-break--visible govuk-!-margin-top-6 govuk-!-margin-bottom-6">
@@ -34,6 +37,10 @@
 
 <p class="govuk-body">
   The OpenAPI spec from which this documentation is generated is <%= govuk_link_to 'available in YAML format', api_docs_spec_url %>.
+</p>
+
+<p class="govuk-body">
+  You can find a  <%= govuk_link_to 'summary of the field length limits', '#field-lengths' %> at the foot of this page.
 </p>
 
 <h3 class="govuk-heading-m">Environments</h3>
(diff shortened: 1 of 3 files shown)


app/views/api_docs/reference/reference.html.erb
```diff
@@ -113,3 +117,22 @@
 <% @api_reference.schemas.each do |schema| %>
   <%= render 'schema', schema: schema %>
 <% end %>
+
+<h2 class="govuk-heading-l" id="field-lengths">Field lengths summary</h2>
```
benswannack, line 121: Can a link to this be added to the contents at the top of the page?

app/views/api_docs/reference/reference.html.erb
```diff
@@ -113,3 +117,22 @@
...
+
+<table class="govuk-table">
+  <thead class="govuk-table__head">
+    <tr class="govuk-table__row">
+      <th scope="col" class="govuk-table__header">Rule</th>
```
benswannack, line 126: What does rule mean here? Is there a possibility for different rules to be included here, or will it only ever be maxLength ones? Is it worth changing the heading to "Field" and trimming off the `maxLength` from all the values

Screenshot 2021-02-09 at 08 46 25

benswannack (approved):
Looking good! Couple of minor things to look at inline

app/views/api_docs/reference/reference.html.erb
SamJamCul: Yeah, the rule is what the limit is referring to - and it would account for a minLength if one does show up later on. We currently don't have any mininmums though, so I can go ahead and trim off the the tail ends

benswannack: In that case maybe it's safer to leave it as-is. I'd be worried we might add a minLength one and start broadcasting it as a max length because of this table. Could we put the max/min part of the information in the Limit column maybe? Like:

Property | Limit
name | 2 min, 100 max
id | 10 max

app/views/api_docs/reference/reference.html.erb
SamJamCul: Done!

app/views/api_docs/reference/reference.html.erb
SamJamCul: I might leave it as is, but add a ticket that lets us know this is a bit of a landmine, and at the point where we do have minimum lengths showing up, this will be a bit odd

PR #4028, opened 2021-02-09
Add DateValidator & YearValidator

despo:
## Context

Currently year and date validations are not consistent throughout the app, and every time we introduce a new feature that relies on them we seem to be re-implementing and testing basic validations.

This PR 
- introduces a DateValidator and a YearValidator.
- Updates forms to use new validators for date of birth, date and year checks
- Update candidate specs to use new short locales for verifying displaying erros
- Update in some candidate specs to test for expected error e.g. input and invalid date rather than a date missing information

#### TODO
- Use new validators in interview forms

### Default checks `validate :field, date: true`
- missing field values if any fields are set e.g where date => 01/nil/2010 adds missing field error (as described in [Design system date input component page](https://design-system.service.gov.uk/components/date-input/)
- invalid date e.g. a/b/2010

#### Presence check `validate :field, date: { presence: true }`
_Adding this check as part of the DateValidator was required as a date field should not be nil but instead a Struct with empty values, so we need to check that all fields (day, month year) are empty_
- checks if all fields are empty

#### Future check `validate :field, date: { future: true }`
- checks that date is not in the future

#### Date of birth `validate :field, date: { date_of_birth: true }`
- checks that date is not in the future 
- checks that age is above the minimum specified age (default is 16)

#### Month and year only checks `validate :field, date: { month_and_year: true }`
- checks if data is valid (displays error with correct example if data is invalid e.g. _Enter %{article} %{attribute} in the correct format, for example 5 2019_
- when `presence: true` checks if any fields are empty (ignore day field)

### Additional improvements

- Check that date year is in an acceptable range  (-100years to 2999) - previously any value between 1000-2999 was accepted
- Error message reuse

### Default checks `validate :field, year: true`

- Check if year is valid (100 year ago to 2999)

### Optional checks `validate :field, year: { future: true }`

- Checks that year is not in the future

## Link to Trello card

Related card - removing DateUtils
https://trello.com/c/Pc6EszXt/407-remove-validationutils-from-app-lib

## Things to check


despo:
@raamSoftwire there are still a couple of forms that I need to look at (interview related forms in particular as they were merged after I started this PR) but any feedback would be appreciated

app/validators/date_validator.rb
```diff
@@ -0,0 +1,77 @@
...
+  def is_invalid?(value)
+    !value.is_a?(Date) || outside_acceptable_age_range(value.year)
+  end
+
+  def outside_acceptable_age_range(value)
+    !(100.years.ago.year..2999).cover?(value)
+  end
+
+  def article(attribute)
+    %w[a e i o u].include?(attribute.to_s.first) ? 'an' : 'a'
```
raamSoftwire, line 53: This is not always going to work (but it will most of the time). Do we care about any of those cases? If so it looks like there are gems for this.

https://stackoverflow.com/questions/5381738/rails-article-helper-a-or-an

raamSoftwire: I think we are only using this for attributes (eg. day, month, year) so this should be fine.

raamSoftwire:
> @raamSoftwire there are still a couple of forms that I need to look at (interview related forms in particular as they were merged after I started this PR) but any feedback would be appreciated

This is a really neat refactor. I've been through everything so far and it looks great. Thanks, Despo!

app/validators/date_validator.rb
despo: attribute in this instance is the name of the field e.g. `start_date`, `end_date` etc. 

As all of our use cases our satisfied by this check (plus the gem is not very commonly used and has not been updated in six years) I think this check is sufficient. Maybe I could look into refactoring this in a common module reused by the Date and Year validator though so that if we feel the need to improve this both validators can benefit from it.

despo:
@raamSoftwire thanks for taking the time to review. I will let you know as soon as the rest of the changes are in.

app/validators/date_validator.rb
raamSoftwire: That sounds like a good idea. There were some other bits of logic (eg. the `current year - 100`) that could also move to that module.

raamSoftwire (approved):
LGTM. That concern nicely keeps all the logic in one place. Cheers Despo!

PR #4030, opened 2021-02-09
Dup the middleware object for thread safety

steventux:
## Context

We suspect that the instance variables in the `call` method are not thread safe and this is leading to unexpected log entries. 

Calls `dup` on the middleware object so that instance variables within the `call` method are thread-safe.

## Guidance to review

See https://github.com/sinatra/sinatra/blob/master/lib/sinatra/base.rb#L929 for an example of how Sinatra does this.

## Link to Trello card

https://trello.com/c/LBHHTB9x/3248-single-api-request-to-qa-produces-two-three-vendorapirequest-entries

## Things to check

### DIFF
--- a/app/middlewares/vendor_api_request_middleware.rb
+++ b/app/middlewares/vendor_api_request_middleware.rb
@@ -18,6 +18,10 @@ class VendorAPIRequestMiddleware
   end
 
   def call(env)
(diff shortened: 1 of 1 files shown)


duncanjbrown (approved):
Definitely worth a try!

PR #4031, opened 2021-02-09
Handle empty POST data in Vendor API request log

steventux:
## Context

Some POST requests to the API do not require data, so there's no need to parse as JSON, currently this shows an error that the requests data was not valid JSON.

Handle empty request data for POST requests. 

## Link to Trello card

https://trello.com/c/EtKnhZQM/3352-support-vendor-api-requests-screen-post-requests-with-no-request-body-display-incorrect-data

## Things to check


benswannack (approved):
Nice 😄 Could we also add a field that shows the query parameters from the URL?

steventux:
@benswannack I've added the `QUERY_STRING` value from Rack as a method-agnostic way of seeing the original querystring.

PR #4032, opened 2021-02-09
Link to Get Into Teaching from footer and sidebar

paulrobertlloyd:
## Context

~~**NOTE:** Dependent on #3995~~

We are now pointing candidates to Get Into Teaching as the first point of contact for support.

Uses the same footer as that shown on Find, and uses this same content in the sidebar.

| Before | After |
| - | - |
| footer-before | footer-after |
| sidebar-before | sidebar-after |

## Link to Trello card

https://trello.com/c/ycLBGKQA/2971-use-find-footer-on-apply

## Things to check


davidgisbey (approved):
👌

PR #3960, opened 2021-02-01
Restructured work history - update the design of review page and candidate flow

JR-G:
## Context

Now that we're asking candidates to enter their full work history, the summary cards take up too much room, making it harder for candidates to scan to see if their work history is complete.

The new design takes up less vertical height by omitting the labels, combining lines, and omitting "full time" and any text when a role was not relevant to teaching.

This PR is for the new review page, connecting these previous PRs:

- [3954](https://github.com/DFE-Digital/apply-for-teacher-training/pull/3954) - The new add a job page
- [3908](https://github.com/DFE-Digital/apply-for-teacher-training/pull/3908) - The new work history start page

## Changes proposed in this pull request

The review page has been redesigned:

|Current work history review page|Restructured work history review page|
|------------|------------|
|Before Screenshot|After Screenshot| 

The review page will now also display '(estimate)' after the start/end month if the candidate selects "I’m not sure the exact month or year I started/left" when they enter a job:

Estimate Screenshot

- Three new components have been added that will render conditionally on the review page:

  `RestructuredWorkHistory::JobComponent`
  `RestructuredWorkHistory::GapComponent`
  `RestructuredWorkHistory::WorkBreakComponent`

- Deleting a job functionality has been included in this ticket

## Guidance to review

- Make sure the 'restructured work history' feature flag is turned on - this flow will only appear to candidates who are not currently using the previous design.

Prototype: https://apply-beta-prototype.herokuapp.com/application/12345/work-history/review

## Link to Trello card

https://trello.com/c/TodQ3QZ9/2904-work-history-update-design-of-review-page-to-use-simple-headings-and-lines-rather-than-summary-card

## Things to check

- [ ] This code does not rely on migrations in the same Pull Request
- [ ] If this code includes a migration adding or changing columns, it also backfills existing records for consistency
- [ ] API release notes have been updated if necessary
- [ ] New environment variables have been [added to the Azure config](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/docs/environment-variables.md#azure-hosting-devops-pipeline)

PR #4033, opened 2021-02-09
Update the application review page to show the new work history design

JR-G:
## Context

The work history section & flow has been redesigned. This change also needs to be reflected within the application review page.

it is an extension of [3960](https://github.com/DFE-Digital/apply-for-teacher-training/pull/3960).

The application review page will now show the new design:
|Previous design|Restructured design|
|-----|-----|
|Screenshot 2021-02-09 at 20 41 14|Screenshot 2021-02-09 at 20 36 05|

## Guidance to review

- Make sure the Restructured work history feature flag is on
- The candidate needs to have a fresh application in order to be presented with the new flow

Prototype: https://apply-beta-prototype.herokuapp.com/application/12345/review

## Link to Trello card

https://trello.com/c/43SrdCXy/3002-dev-update-the-application-review-page-to-show-the-new-work-history-design

## Things to check

### DIFF
--- a/app/components/candidate_interface/restructured_work_history/job_component.rb
+++ b/app/components/candidate_interface/restructured_work_history/job_component.rb
@@ -21,12 +21,14 @@ module CandidateInterface
       end
 
       def formatted_end_date
-        return 'Present' if @work_experience.currently_working
-
-        if @work_experience.end_date_unknown
-          "#{@work_experience.end_date.to_s(:short_month_and_year)} (estimate)"
+        if @work_experience.currently_working
+          'to Present'
+        elsif @work_experience.start_date == @work_experience.end_date
+          nil
+        elsif @work_experience.end_date_unknown
+          "to #{@work_experience.end_date.to_s(:short_month_and_year)} (estimate)"
         else
-          @work_experience.end_date.to_s(:short_month_and_year)
+          "to #{@work_experience.end_date.to_s(:short_month_and_year)}"
         end
       end
     end
--- a/app/components/candidate_interface/restructured_work_history/job_component.html.erb
+++ b/app/components/candidate_interface/restructured_work_history/job_component.html.erb
@@ -9,7 +9,7 @@
     <% end %>
 
     <p class="govuk-body">
-      <%= formatted_start_date %> to <%= formatted_end_date %>
+      <%= formatted_start_date %> <%= formatted_end_date %>
     </p>
 
     <% if @work_experience.relevant_skills %>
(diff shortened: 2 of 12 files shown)


frankieroberto (dismissed):
Looking good!

A few issues:

1. please remove this validation, as this isn't needed for now (as unexplained breaks aren't a show-stopper):

Screenshot 2021-02-10 at 15 16 16

2. Remove prompt to add a break explanation or job _before_ the first item in the list (as we can’t reliably calculate when the candidate was expected to have "entered the workplace").

Screenshot 2021-02-10 at 15 17 52

3. Minor bug: if you save a job with an end date, and then go back and edit it to say that you're still working in that job, the end date isn't removed upon save. This causes a prompt for a break explanation to appear after the end date in the review, even though this has been overridden by the "Present" flag.

4. Minor: Could we drop the "to [Month Year]" string if the start date and end date are the same?

Screenshot 2021-02-10 at 15 24 33

app/components/candidate_interface/restructured_work_history/job_component.rb
```diff
@@ -21,12 +21,13 @@ def formatted_start_date
...
-        return 'Present' if @work_experience.currently_working
+        return 'to Present' if @work_experience.currently_working
+        return nil if @work_experience.start_date == @work_experience.end_date
 
         if @work_experience.end_date_unknown
-          "#{@work_experience.end_date.to_s(:short_month_and_year)} (estimate)"
+          "to #{@work_experience.end_date.to_s(:short_month_and_year)} (estimate)"
         else
-          @work_experience.end_date.to_s(:short_month_and_year)
+          "to #{@work_experience.end_date.to_s(:short_month_and_year)}"
```
davidgisbey, line 30:
```
if @work_experience.currently_working
  'to Present'
elsif @work_experience.start_date == @work_experience.end_date
  nil
elsif @work_experience.end_date_unknown
  "to #{@work_experience.end_date.to_s(:short_month_and_year)} (estimate)"
else 
   "to #{@work_experience.end_date.to_s(:short_month_and_year)}"
end
```

spec/components/candidate_interface/work_history_review_component_spec.rb
```diff
@@ -150,27 +150,10 @@
...
       end
     end
-
-    context 'when individual breaks are editable' do
-      it 'renders component without break placeholders' do
-        result = render(editable: true)
-
-        expect(result.css('.app-summary-card__title').text).to include('You have a break in your work history in the last 5 years')
-      end
-    end
```
davidgisbey, line 165: I think we need to add two tests here:

1. one that checks that the placeholder renders when there is a break between two jobs
2. one that checks that a placeholder doesn't render if they're currently in a job role (i.e. end date is nil on their latest job)

davidgisbey (approved):
Looks good. I can see you've addressed Frankies points. Just needs a couple of tests added

davidgisbey (approved):
lgtm. Let's merge this in and do an audit of what still needs to be done for us to switch this on, on  Monday am 👍

PR #4034, opened 2021-02-09
Fix errors on support interface R4R search page

stevehook:
## Context

Errors occurred on the support interface R4R search page because we didn't have a key for `other_advice_or_feedback_y_n` in the hash of top-level reasons. This change fills in the missing key and adds tests and defensive code to prevent similar issues that happen due to data that doesn't match our assumptions.

https://sentry.io/organizations/dfe-bat/issues/2208115277/?environment=qa&project=1765973&query=is%3Aunresolved#grouping-info

- Add the missing metadata for `other_advice_or_feedback_y_n`
- Handle any unknown R4R attributes/values that may exist in the database or be used as params to this page.

## Guidance to review

- Is it reasonable to ignore attributes and values that we don't know about? (rather than crash)

## Link to Trello card

n/a - see https://sentry.io/organizations/dfe-bat/issues/2208115277/?environment=qa&project=1765973&query=is%3Aunresolved#grouping-info

## Things to check

### DIFF
--- a/app/components/support_interface/reasons_for_rejection_search_results_component.rb
+++ b/app/components/support_interface/reasons_for_rejection_search_results_component.rb
@@ -22,13 +22,13 @@ module SupportInterface
     def search_title_text
       if @search_value == 'Yes'
         i18n_key = SupportInterface::SubReasonsForRejectionTableComponent::TOP_LEVEL_REASONS_TO_I18N_KEYS[@search_attribute].to_s
-        t("reasons_for_rejection.#{i18n_key}.title")
+        t("reasons_for_rejection.#{i18n_key}.title", default: '')
       else
         top_level_reason = ReasonsForRejectionCountQuery::SUBREASONS_TO_TOP_LEVEL_REASONS[@search_attribute.to_sym]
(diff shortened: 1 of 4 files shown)


davidgisbey (approved):
Interesting. Do we know how the data got into this state? I think this fine, but it's something to bear in mind going forward. I was under the assumption that this was a fairly rigid structure. 

We might need to rethink this is the structured rejection reasons design changes at some point to.

stevehook:
> Interesting. Do we know how the data got into this state? I think this fine, but it's something to bear in mind going forward. I was under the assumption that this was a fairly rigid structure.
> 
> We might need to rethink this is the structured rejection reasons design changes at some point too.

The crash happened because I just hadn't added `other_advice_or_feedback_y_n` to the hash of top-level reasons, so we couldn't find the i18n title. So it was valid data (and an oversight on my part) that triggered it. I added the defensive code because (a) the search params are just on the query string so could be invalid (b) we might add keys to the structure in future and I just wanted to make sure it doesn't break if we don't update the support interface.

PR #4038, opened 2021-02-10
Make support and provider fallback authentication consistent

tijmenb:
## Context

We've got a lot of duplication in the app around the authentication by token. 

- Use the `AuthenticationToken` approach for provider fallback auth
- Make the tokens usable only once, like the candidate tokens
- Refactor bits to avoid duplication

## Guidance to review

Any holes?

## Link to Trello card

## Things to check


davidgisbey (approved):
Great work. This is much cleaner. I think we can mark this card as done now https://trello.com/c/VohxOZiV/309-high-magic-link-token-has-a-lot-of-duplication

PR #4039, opened 2021-02-10
Hard code sequence start value so schema.rb does not change based on which db it is run on

benswannack:
With the addition of public ids to qualifications, we inadvertently added a minor annoyance: People running this migration locally will get different values for the start of the sequence. Then the schema file changes to reflect that

## Guidance to review
- This migration has already been run on all environments, each environment's sequence has a different starting value
- Are there any risks to this change? My understanding is that only test and dev environments consult the schema.rb to generate the database, so this change in number will have no effect unless we rollback and reapply the migration.
- To properly fix this, people who have already run the migration locally will need to roll back and reapply the migration.

## Things to check

### DIFF
--- a/db/migrate/20210203160529_add_qualification_public_id_sequence.rb
+++ b/db/migrate/20210203160529_add_qualification_public_id_sequence.rb
@@ -3,7 +3,7 @@ class AddQualificationPublicIdSequence < ActiveRecord::Migration[6.0]
     # We set the start of the sequence to be twice the max id for qualifications that currently exist.
     # It is important that the public_id does not clash with any existing ids from the database
     # See adr/0018-public-ids-for-qualifications.md for more details
-    max_qualification_id = ApplicationQualification.maximum(:id)
+    max_qualification_id = 60000
     create_sequence :qualifications_public_id_seq, start: max_qualification_id * 2
   end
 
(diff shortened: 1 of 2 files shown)


db/migrate/20210203160529_add_qualification_public_id_sequence.rb
```diff
@@ -3,7 +3,7 @@ def up
     # We set the start of the sequence to be twice the max id for qualifications that currently exist.
     # It is important that the public_id does not clash with any existing ids from the database
     # See adr/0018-public-ids-for-qualifications.md for more details
-    max_qualification_id = ApplicationQualification.maximum(:id)
+    max_qualification_id = 60000
```
benswannack, line 6: This is a rough value from production currently

db/schema.rb
```diff
@@ -38,7 +38,7 @@
   create_sequence "provider_users_id_seq"
   create_sequence "provider_users_providers_id_seq"
   create_sequence "providers_id_seq"
-  create_sequence "qualifications_public_id_seq", start: 196
+  create_sequence "qualifications_public_id_seq", start: 120000
```
steventux, line 41: Docs say the default postgresql sequence max value is 2,147,483,647 so this seems a safe starting point.

steventux (approved):
Would be good to get more eyes on this but it seems like a sensible set of values to use.

db/schema.rb
benswannack: Ok! This has already run on production with a value of 110000 or so, the value here is just what it was when it was run locally

tijmenb (approved):

PR #4047, opened 2021-02-10
Restore interviews in the provider activity log

mnacos:
## Context

Our last attempt at including interviews (reverted in https://github.com/DFE-Digital/apply-for-teacher-training/pull/4020) produced SQL which specified the OR logic needed for an .own_and_associated_audits query in the WHERE clause of a query and included the application_choices visibility
rules twice. By making this OR logic a part of the INNER JOIN between audits and the application_choices from GetApplicationChoicesForProviders, we can scan the audits table more efficiently and take advantage of the
existing [auditable_type, auditable_id] and [associated_type, associated_id] indexes.

Besides speed, this approach should work well for any other ApplicationChoice-associated audits we add in the future, similar to interviews. We filter so much on ApplicationChoice statuses and other fields because not all state transitions are relevant for Manage, but associated models (such as interviews) should either be included in the feed for all their changes or not at all.

The SQL produced changes from

```sql
SELECT 
  "audits".* 
FROM 
  "audits" 
WHERE 
  (
    "audits"."auditable_type" = 'ApplicationChoice' 
    AND "audits"."auditable_id" IN (
      SELECT 
        "application_choices"."id" 
      FROM 
        "application_choices" 
        INNER JOIN course_options AS current_option ON COALESCE(
          offered_course_option_id, course_option_id
        ) = current_option.id 
        INNER JOIN course_options AS original_option ON course_option_id = original_option.id 
        INNER JOIN courses AS current_course ON current_option.course_id = current_course.id 
        INNER JOIN courses AS original_course ON original_option.course_id = original_course.id 
      WHERE 
        (
          (
            "original_course"."recruitment_cycle_year" IN (2021, 2020) 
            AND (
              "original_course"."provider_id" IN (1593, 77, 404, 1538) 
              OR "original_course"."accredited_provider_id" IN (1593, 77, 404, 1538)
            ) 
            OR "current_course"."provider_id" IN (1593, 77, 404, 1538) 
            AND "current_course"."recruitment_cycle_year" IN (2021, 2020)
          ) 
          OR "current_course"."accredited_provider_id" IN (1593, 77, 404, 1538) 
          AND "current_course"."recruitment_cycle_year" IN (2021, 2020)
        ) 
        AND (
          status IN (
            'awaiting_provider_decision', 'offer', 
            'pending_conditions', 'recruited', 
            'rejected', 'declined', 'withdrawn', 
            'conditions_not_met', 'offer_withdrawn', 
            'offer_deferred'
          )
        )
    ) 
    AND "audits"."action" = 'update' 
    AND (
      jsonb_exists(
        audited_changes, 'reject_by_default_feedback_sent_at'
      ) 
      OR jsonb_exists(
        audited_changes, 'offer_changed_at'
      ) 
      OR audited_changes :: json #>>'{status, 1}' IN ('awaiting_provider_decision', 'interviewing', 'offer', 'pending_conditions', 'recruited', 'rejected', 'declined', 'withdrawn', 'conditions_not_met', 'offer_withdrawn', 'offer_deferred'))
      OR "audits"."associated_type" = 'ApplicationChoice'
      AND "audits"."associated_id" IN (
        SELECT
          "application_choices"."id"
        FROM
          "application_choices"
          INNER JOIN course_options AS current_option ON COALESCE(
            offered_course_option_id, course_option_id
          ) = current_option.id
          INNER JOIN course_options AS original_option ON course_option_id = original_option.id
          INNER JOIN courses AS current_course ON current_option.course_id = current_course.id
          INNER JOIN courses AS original_courseON original_option.course_id = original_course.id
        WHERE
          (
            (
              "original_course"."recruitment_cycle_year" IN (2021, 2020)
              AND (
                "original_course"."provider_id" IN (1593, 77, 404, 1538)
                OR "original_course"."accredited_provider_id" IN (1593, 77, 404, 1538)
              )
              OR "current_course"."provider_id" IN(1593, 77, 404, 1538)
              AND "current_course"."recruitment_cycle_year" IN (2021, 2020)
            )
            OR "current_course"."accredited_provider_id" IN (1593, 77, 404, 1538)
            AND "current_course"."recruitment_cycle_year" IN (2021, 2020)
          )
          AND (
            status IN (
              'awaiting_provider_decision', 'offer',
              'pending_conditions', 'recruited',
              'rejected', 'declined', 'withdrawn',
              'conditions_not_met', 'offer_withdrawn',
              'offer_deferred'
            )
          )
      )
      AND "audits"."action" IN ('create', 'update')
  )
  AND (created_at > '2018-01-01 00:00:00') ORDER BY "audits"."created_at" DESC LIMIT 50 OFFSET 0
```

which doesn't complete, to

```sql
SELECT
  "audits".*
FROM
  (
    SELECT
      a.*
    FROM
      audits a
      INNER JOIN (
        SELECT
          "application_choices".*
        FROM
          "application_choices"
          INNER JOIN course_options AS current_option ON COALESCE(
            offered_course_option_id, course_option_id
          ) = current_option.id
          INNER JOIN course_options AS original_option ON course_option_id = original_option.id
          INNER JOIN courses AS current_course ON current_option.course_id = current_course.id
          INNER JOIN courses AS original_course ON original_option.course_id = original_course.id
        WHERE
          (
            (
              "original_course"."recruitment_cycle_year" IN (2021, 2020)
              AND (
                "original_course"."provider_id" IN (1593, 77, 404, 1538)
                OR "original_course"."accredited_provider_id" IN (1593, 77, 404, 1538)
              )
              OR "current_course"."provider_id" IN (1593, 77, 404, 1538)
              AND "current_course"."recruitment_cycle_year" IN (2021, 2020)
            )
            OR "current_course"."accredited_provider_id" IN (1593, 77, 404, 1538)
            AND "current_course"."recruitment_cycle_year" IN (2021, 2020)
          )
          AND (
            status IN (
              'awaiting_provider_decision', 'offer',
              'pending_conditions', 'recruited',
              'rejected', 'declined', 'withdrawn',
              'conditions_not_met', 'offer_withdrawn',
              'offer_deferred'
            )
          )
      ) ac ON
        (
          auditable_type = 'ApplicationChoice'
          AND auditable_id = ac.id
          AND action = 'update'
          AND (
            jsonb_exists(
              audited_changes, 'reject_by_default_feedback_sent_at'
            )
            OR jsonb_exists(
              audited_changes, 'offer_changed_at'
            )
	    OR audited_changes :: json #>>'{status, 1}' IN
	      ('awaiting_provider_decision', 'interviewing', 'offer', 'pending_conditions', 'recruited', 'rejected', 'declined', 'withdrawn', 'conditions_not_met', 'offer_withdrawn', 'offer_deferred')
          )
        ) OR (
	  associated_type = 'ApplicationChoice'
	  AND associated_id = ac.id
        )
    WHERE a.created_at >= '2018-01-01T00:00:00+00:00'::TIMESTAMPTZ
    ORDER BY a.created_at DESC
  ) AS audits
```

which completes in 75ms (faster than the 100ms the version in https://github.com/DFE-Digital/apply-for-teacher-training/pull/4020, without associated audits, takes)

## Guidance to review

The multiple INNER JOINs on courses and course options are coming from the `GetApplicationChoicesForProviders` service. In the controller, we call `GetActivityLogEvents` like this:

```ruby
      @events = GetActivityLogEvents.call(
        application_choices: GetApplicationChoicesForProviders.call(
          providers: current_provider_user.providers,
        ),
      ).page(params[:page] || 1).per(50)
```

Yes, we've gone back to hand-crafted/embedded SQL in the service, but I don't know how to specify such a join using ActiveRecord syntax. Any suggestions welcome!

UPDATE: I've moved the raw SQL to a `.joins` expression, because this is the bit ActiveRecord can't do, made everything else standard AR.

## Link to Trello card

https://trello.com/c/B79CEysQ

## Things to check

### DIFF
--- a/app/queries/get_activity_log_events.rb
+++ b/app/queries/get_activity_log_events.rb
@@ -25,6 +25,26 @@ class GetActivityLogEvents
   def self.call(application_choices:, since: nil)
     since ||= Time.zone.local(2018, 1, 1) # before the pilot began, i.e. all records
 
+    application_choices_join_sql = <<~COMBINE_AUDITS_WITH_APPLICATION_CHOICES_SCOPE_AND_FILTER.squish
+      INNER JOIN (#{application_choices.to_sql}) ac
+        ON (
+          auditable_type = 'ApplicationChoice'
+          AND auditable_id = ac.id
+          AND action = 'update'
+          AND ( #{application_choice_audits_filter_sql} )
+        ) OR (
+          associated_type = 'ApplicationChoice'
+          AND associated_id = ac.id
+        )
+    COMBINE_AUDITS_WITH_APPLICATION_CHOICES_SCOPE_AND_FILTER
+
+    Audited::Audit.includes(INCLUDES)
+                  .joins(application_choices_join_sql)
+                  .where('audits.created_at >= ?', since)
+                  .order('audits.created_at DESC')
+  end
+
+  def self.application_choice_audits_filter_sql
     application_choice_audits_filter = INCLUDE_APPLICATION_CHOICE_CHANGES_TO.map { |f|
       "jsonb_exists(audited_changes, '#{f}')"
     }.join(' OR ')
@@ -32,27 +52,6 @@ class GetActivityLogEvents
     filtered_statuses = ApplicationStateChange::STATES_VISIBLE_TO_PROVIDER - IGNORE_STATUS
     status_transitions_to_include = filtered_statuses.map { |status| "'#{status}'" }.join(', ')
 
-    application_choice_audits_filter += " OR audited_changes::json#>>'{status, 1}' IN (#{status_transitions_to_include})"
-
-    Audited::Audit.includes(INCLUDES).from <<~COMBINE_AUDITS_WITH_APPLICATION_CHOICES_SCOPE_AND_FILTER.squish
-      (
-        SELECT a.*
-          FROM audits a
-
-          INNER JOIN (#{application_choices.to_sql}) ac
-            ON (
-              auditable_type = 'ApplicationChoice'
-              AND auditable_id = ac.id
-              AND action = 'update'
-              AND ( #{application_choice_audits_filter} )
-            ) OR (
-              associated_type = 'ApplicationChoice'
-              AND associated_id = ac.id
-            )
-
-          WHERE a.created_at >= '#{since.iso8601}'::TIMESTAMPTZ
-          ORDER BY a.created_at DESC
-      ) AS audits
-    COMBINE_AUDITS_WITH_APPLICATION_CHOICES_SCOPE_AND_FILTER
+    application_choice_audits_filter + " OR audited_changes::json#>>'{status, 1}' IN (#{status_transitions_to_include})"
   end
 end
--- a/app/queries/get_activity_log_events.rb
+++ b/app/queries/get_activity_log_events.rb
@@ -15,21 +15,44 @@ class GetActivityLogEvents
     ],
   }.freeze
 
-  CHANGES_TO = %w[
+  INCLUDE_APPLICATION_CHOICE_CHANGES_TO = %w[
     reject_by_default_feedback_sent_at
     offer_changed_at
   ].freeze
 
+  IGNORE_STATUS = %i[interviewing].freeze
+
   def self.call(application_choices:, since: nil)
     since ||= Time.zone.local(2018, 1, 1) # before the pilot began, i.e. all records
 
-    filter_query = "audited_changes ?| array[:changes] OR audited_changes #>> '{status, 1}' IN (:statuses)"
+    application_choice_audits_filter = INCLUDE_APPLICATION_CHOICE_CHANGES_TO.map { |f|
+      "jsonb_exists(audited_changes, '#{f}')"
+    }.join(' OR ')
 
-    Audited::Audit.includes(INCLUDES)
-                  .where(auditable: application_choices)
-                  .where(action: :update)
-                  .where(filter_query, changes: CHANGES_TO, statuses: ApplicationStateChange::STATES_VISIBLE_TO_PROVIDER)
-                  .where('created_at > ?', since)
-                  .order(created_at: :desc)
+    filtered_statuses = ApplicationStateChange::STATES_VISIBLE_TO_PROVIDER - IGNORE_STATUS
+    status_transitions_to_include = filtered_statuses.map { |status| "'#{status}'" }.join(', ')
+
+    application_choice_audits_filter += " OR audited_changes::json#>>'{status, 1}' IN (#{status_transitions_to_include})"
+
+    Audited::Audit.includes(INCLUDES).from <<~COMBINE_AUDITS_WITH_APPLICATION_CHOICES_SCOPE_AND_FILTER.squish
+      (
+        SELECT a.*
+          FROM audits a
+
+          INNER JOIN (#{application_choices.to_sql}) ac
+            ON (
+              auditable_type = 'ApplicationChoice'
+              AND auditable_id = ac.id
+              AND action = 'update'
+              AND ( #{application_choice_audits_filter} )
+            ) OR (
+              associated_type = 'ApplicationChoice'
+              AND associated_id = ac.id
+            )
+
+          WHERE a.created_at >= '#{since.iso8601}'::TIMESTAMPTZ
+          ORDER BY a.created_at DESC
+      ) AS audits
+    COMBINE_AUDITS_WITH_APPLICATION_CHOICES_SCOPE_AND_FILTER
   end
 end
--- a/app/components/provider_interface/application_timeline_component.rb
+++ b/app/components/provider_interface/application_timeline_component.rb
@@ -5,7 +5,7 @@ module ProviderInterface
 
     def initialize(application_choice:)
       @application_choice = application_choice
-      audits = GetActivityLogEvents.call(application_choices: [application_choice])
+      audits = GetActivityLogEvents.call(application_choices: ApplicationChoice.where(id: application_choice.id))
       @activity_log_events = audits.map { |audit| ActivityLogEvent.new(audit: audit) }
     end
 
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -101,6 +101,26 @@
       "note": ""
     },
     {
+      "warning_type": "SQL Injection",
+      "warning_code": 0,
+      "fingerprint": "53c8ba33ea76005585e7380e66bbc196385862111c7cb27fe2f0e6793c1ed630",
+      "check_name": "SQL",
+      "message": "Possible SQL injection",
+      "file": "app/queries/get_activity_log_events.rb",
+      "line": 42,
+      "link": "https://brakemanscanner.org/docs/warning_types/sql_injection/",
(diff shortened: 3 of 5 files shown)


app/queries/get_activity_log_events.rb
```diff
@@ -15,25 +15,44 @@ class GetActivityLogEvents
...
-    Audited::Audit.includes(INCLUDES)
-                  .where(auditable: application_choices)
-                  .where(action: :update)
-                  .where(filter)
-                  .where('created_at > ?', since)
-                  .order(created_at: :desc)
+          WHERE a.created_at >= '#{since.iso8601}'::TIMESTAMPTZ
+          ORDER BY a.created_at DESC
+      ) AS audits
+    COMBINE_AUDITS_WITH_APPLICATION_CHOICES_SCOPE_AND_FILTER
```
stevehook, line 56: I wonder if the join could lead to duplicates. Thinking about it probably not assuming there are no audit records with `associated` values the same as `auditable`.

stevehook (commented):
This seems sensible to me. The performance is impressive.

I don't know how you could express this join in ActiveRecord either. You can probably do it with Arel though I'm not sure that would really make it any clearer than it is now (as raw SQL).

app/queries/get_activity_log_events.rb
cwrw: good question, might be worth dropping a distinct in there just incase

app/queries/get_activity_log_events.rb
```diff
@@ -15,25 +15,44 @@ class GetActivityLogEvents
...
-    filter += " OR audited_changes::json#>>'{status, 1}' IN (#{statuses_visible_to_providers})"
+          INNER JOIN (#{application_choices.to_sql}) ac
+            ON (
+              auditable_type = 'ApplicationChoice'
+              AND auditable_id = ac.id
+              AND action = 'update'
+              AND ( #{application_choice_audits_filter} )
+            ) OR (
+              associated_type = 'ApplicationChoice'
+              AND associated_id = ac.id
```
cwrw, line 50: I noticed previously we were filtering by `action` being `create` or `update`, is it ok to include all actions for associated type?

despo:
@mnacos I spend sometime yesteday in trying to convert the root sql into AR and ended up with this so I think it's doable (didn't have time to look into the activity log one):
https://github.com/DFE-Digital/apply-for-teacher-training/compare/convert-application-retrieval-to-AR?expand=1

I get one failure in a system spec so it might need a little bit more work, but other than that everything seems to work as expected.

app/queries/get_activity_log_events.rb
steventux: I think this omission is OK here, at least for interviews as it's a soft-deleted model, all events should implicitly be `create` or `update`.

mnacos:
> @mnacos I spend sometime yesteday in trying to convert the root sql into AR and ended up with this, so I think it's possible to also do the same with this part of the SQL (didn't have time to look into the activity log one):
> https://github.com/DFE-Digital/apply-for-teacher-training/compare/convert-application-retrieval-to-AR?expand=1

Hey Despo, this is great but I don't see why we need to tackle both things at once. There are two reasons I can see why we would want to wait and combine the two PRs:

Argument 1: if simplifying the `GetApplicationChoicesForProviders` query (as in making it more ActiveRecord, which does not necessarily mean less complex at the SQL level) is that easy we may be able to do something similar for the `GetActivityLogEvents` query, which would make it easier to maintain.

Argument 2: if the `GetApplicationChoicesForProviders` query is more efficient, perhaps the `GetActivityLogEvents` query which includes it would also perform better and maybe we don't need to change it this drastically.

Considering how performance sensitive the activity log query is (as a feed of activity across all providers, and a number of events that will always be increasing) my preference is performance over readability (within reason) and I'm suggesting a JOIN-based approach, which also combines the filters in one go. I don't think this is supported by ActiveRecord syntax, which means there will always be a bit of raw SQL in this service taking this direction. Besides, the complexity in `GetApplicationChoicesForProviders` is of a different type. I can move a few things around like @cwrw suggested, of course, and we could try to express similar things using Arel, like you suggested, but whether chaining things together with Arel constructs would be more readable, this remains to be seen.

For argument 2, I don't think the complexity of `GetApplicationChoicesForProviders` was the main issue with the activity log query performance, other than we absolutely need its results to filter activity log events. I blame the fact that it's executed twice in the two subqueries ActiveRecord produces for the `OR` and that the query structure stops postgres from using the `audits` indexes.

app/queries/get_activity_log_events.rb
mnacos: Yes, that's intentional. We exclude `create` events when the event directly related to the ApplicationChoice because this corresponds to the inital model creation, when the candidate adds courses to their application. At this stage application choices are at the `:unsubmitted` stage and should not be visible to providers. The first event a provider user should see is application submission, which is an `update`.

For interviews, on the other hand, and any such associated model, a `create` corresponds to the initial scheduling of an interview, which should be visible.

app/queries/get_activity_log_events.rb
mnacos: This is a good thing to consider, but the answer is duplicates are not possible here, because:

1. Application choice ids are unique (unless there's a problem with the `GetApplicationChoicesForProviders` query)

2. All `audits` considered will have either `auditable_type = 'ApplicationChoice'` or `associated_type = 'ApplicationChoice` from normal use of the app.

3. Even if you artificially constructed an `Audit` record with both `auditable` and `associated` pointing to the same _application choice_ (which makes no sense), you wouldn't get duplicates because the join evaluates records against each other, so this `Audit` would only be considered once.

4. Yes, you could get duplicates by artificially crafting an `Audit` with `auditable` pointing to one application choice and `associated` pointing to another application choice, both of which are part of the `GetApplicationChoicesForProviders` result set.

Including a `DISTINCT` is, in theory, a good idea, but it can have a serious impact on performance, so I'll leave it out and make sure that `GetApplicationChoicesForProviders` returns no duplicates when we change it.

mnacos: By duplicates, I assume you mean the same audit event being shown more than once. See also: https://stackoverflow.com/questions/40218138/will-inner-join-allow-duplicates

steventux:
The performance improvement here is great and I think this gets us out of difficulty but at a maintenance cost. There will undoubtedly be amends needed to this query over time and I can see the sense in using Arel over raw SQL. Whether we iterate to Arel or do it now is a question of how urgent this issue is, but for readability and maintainability my preference would be a Ruby DSL over SQL.
A bit tangental to this PR but @mnacos did you consider something like https://github.com/scenic-views/scenic to create views of pre-filtered data which we can run standardised `Auditable::Audit` scopes / clauses on? I appreciate views come with their own pitfalls and there'sa fair bit of dev overhead to this approach but I wonder if it would simplify how we retrieve audit data.

mnacos:
> Whether we iterate to Arel or do it now is a question of how urgent this issue is, but for readability and maintainability my preference would be a Ruby DSL over SQL.

@cwrw has already suggested we minimise the SQL to the bit that cannot be reproduced with AR, i.e. joining with the subquery from `GetApplicationChoicesForProviders` while specifying conditions, similar to https://stackoverflow.com/questions/44283756/rails-activerecord-subquery-on-inner-join so we'll probably be able to take it this far.

steventux (approved):
@mnacos I think this is pretty legible considering the SQL it produces. It feels like it's fairly minimal in raw SQL and neat as a query class which helps with understanding what's going on.

PR #4054, opened 2021-02-11
Small fixes to interviews

benswannack:
## Context
There are a few obvious bugs in the interviews feature that we've noticed already

- Display interview preferences in the interview summary card
- Capitalise fee status
- Validate presence of cancellation reason when cancelling interview
- Add change links to cancellation check page
- Hide links that provider users do not have permission to view

## Guidance to review
Per commit

## Link to Trello card
https://trello.com/c/TgL7U1Tr/3305-review-interview-flow-with-designers-and-implement-fixes-where-necessary

## Things to check

### DIFF
--- a/app/views/provider_interface/interviews/index.html.erb
+++ b/app/views/provider_interface/interviews/index.html.erb
@@ -12,46 +12,7 @@
         </h2>
 
         <% @interviews.each do |interview| %>
-          <div class="app-interviews__interview">
-
-            <%= render SummaryCardComponent.new(editable: true, border: true, rows: [
-              {
-                key: 'Course',
-                value: interview.offered_course.name
-              },
-              {
-                key: 'Funding type',
-                value: interview.offered_course.funding_type
-              },
-              {
-                key: 'Organisation carrying out interview',
-                value: interview.provider.name
-              },
-              {
-                key: 'Address or online meeting details',
-                value: interview.location
-              },
-              {
-                key: 'Additional details',
-                value: interview.additional_details
-              }
-            ].compact
-             ) do %>
-             <%= render SummaryCardHeaderComponent.new(title: interview.date_and_time.to_s(:govuk_date_and_time), heading_level: 2) do %>
-               <div class="app-summary-card__actions">
-               <ul class= "app-summary-card__actions-list">
-                 <li class="app-summary-card__actions-list-item">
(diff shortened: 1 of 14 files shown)


config/locales/interview.yml
```diff
@@ -49,14 +49,20 @@ en:
         provider_interface/interview_wizard:
           attributes:
             date:
-              invalid: The interview date must be a real date
-              missing_values: The interview date must include a %{missing_details}
-              blank: Enter the interview date
-              past: The interview date must be in the future
-              after_rdb: The interview date must be before the application closing date
+              invalid: Date must be a real date
```
despo, line 52: `invalid`, `missing_values` and `blank` can be removed (as they no longer map to the date errors anyway, but should also default to the same errors used across the project which more or less maps the the amended error messages)

spec/forms/provider_interface/cancel_interview_wizard_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'rails_helper'
+
+RSpec.describe ProviderInterface::CancelInterviewWizard do
+  let(:store) do
+    s = instance_double(WizardStateStores::RedisStore)
```
despo, line 5:
```suggestion
instance_double(WizardStateStores::RedisStore, read: nil)
```

app/views/provider_interface/interviews/index.html.erb
```diff
@@ -6,52 +6,16 @@
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
       <div class="app-interviews">
-        <%= govuk_link_to 'Set up interview', new_provider_interface_application_choice_interview_path(@application_choice), class: 'govuk-button govuk-button--secondary' %>
+        <% if @provider_can_respond %>
```
despo, line 9: Maybe update this to `@user_can_change_interview` so it's easier to understand how this variable is used here and in the `InterviewAndCourseSummaryComponent`.

benswannack: I've gone with `@provider_can_make_decisions` which actually describes what this variable means (rather than how it will be used inside a component)

AgaDufrat (approved):
LGTM

_Git commit history police 👮_ not sure if `Appraise rubocop` deserves a separate commit (it could fixup `Refactor interview and course details UI into component`) but good to merge as it is.

PR #4055, opened 2021-02-11
Use new `govuk-button-group` component

paulrobertlloyd:
## Context

`govuk-frontend` now has styles for a button group: `govuk-button-group`; when two or more buttons or links appear next to each other.

* Adds a button group to candidate application form and provider users page in the Support interface (only minor spacing change)
* Adds a button group to ‘Make a decision’ inset text on Manage interface (only minor spacing change)
* Uses button group in cookie banner (which is part of the reason why this new style was introduced). Possible a whole other bit of design work to bring our cookie banner inline with the new guidance, ~~but for now updated our cookie banner to use this new button group style~~ updated the code so that the button and link appear within `govuk-button-group` (thanks to the form builder magic), adding a modifier to ensure the buttons are the same width:

| Before | After |
| - | - |
| cookies-before | cookies-after

Honestly, I thought this might be needed in more places… 😆

## Things to check

### DIFF
--- a/app/views/shared/_cookies_banner.html.erb
+++ b/app/views/shared/_cookies_banner.html.erb
@@ -1,22 +1,15 @@
 <div class="govuk-width-container">
-  <div class="govuk-grid-row">
-    <div class="govuk-grid-column-two-thirds">
-      <div class="govuk-width-container govuk-!-padding-top-4">
-        <h1 class="govuk-heading-m govuk-!-margin-bottom-4">
-          Tell us whether you accept cookies
-        </h1>
-        <p class="govuk-body">
-        We use <%= govuk_link_to 'cookies to collect information', cookies_path %> about how you use ‘<%= service_name %>’. We use this information to make the website work as well as possible and improve government services.
-        </p>
-        <div class="cookies_banner__button govuk-grid-column-full govuk-grid-column-one-half-from-desktop">
-          <%= form_with model: CookiePreferencesForm.new(consent: 'yes'), url: cookie_preferences_path, method: :post do |f| %>
-            <%= f.govuk_submit 'Accept all cookies' %>
-          <% end %>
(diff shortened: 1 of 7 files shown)


app/views/shared/_cookies_banner.html.erb
```diff
@@ -6,16 +6,14 @@
...
-          <% end %>
-        </div>
-        <div class="cookies_banner__button govuk-grid-column-full govuk-grid-column-one-half-from-desktop">
-          <a class="govuk-button" href=<%= cookies_path %>>Set cookie preferences</a>
-        </div>
+        <%= form_with model: CookiePreferencesForm.new(consent: 'yes'), url: cookie_preferences_path, method: :post do |f| %>
+          <div class="govuk-button-group">
+            <%= f.govuk_submit 'Accept all cookies', classes: 'app-button--cookies' %>
+            <a class="govuk-button app-button--cookies" href=<%= cookies_path %>>Set cookie preferences</a>
+          </div>
```
peteryates, line 15: 👋🏽  hey, I [updated the formbuilder](https://github.com/DFE-Digital/govuk_design_system_formbuilder/pull/237) yesterday to make this easier; you should be able to just do this for the same result.

```erb
<%= f.govuk_submit('Accept all cookies', classes: 'app-button--cookies') do %>
  <a class="govuk-button app-button--cookies" href=<%= cookies_path %>>Set cookie preferences</a>
<% end %>
```

paulrobertlloyd: Ooooooh, nice one!

paulrobertlloyd: Done! Also realise can use our `govuk_button_link_to` helper here as well.

adamsilver (approved):

PR #4057, opened 2021-02-11
Add line to PR template about data migrations

benswannack:
## Context
Encourage developers to put data migration code in a fully tested service to avoid pushing untested code in migrations

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -18,5 +18,6 @@
 
 - [ ] This code does not rely on migrations in the same Pull Request
 - [ ] If this code includes a migration adding or changing columns, it also backfills existing records for consistency
+- [ ] If this code includes a data migration/backfill, then the code that does the backfill is in a fully tested service called by the migration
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -18,5 +18,6 @@
 
 - [ ] This code does not rely on migrations in the same Pull Request
 - [ ] If this code includes a migration adding or changing columns, it also backfills existing records for consistency
+- [ ] If this code includes a data migration/backfill, then the code that does the backfill is in a fully tested service called by the migration
```
cwrw, line 21: I think this might be superseded by this approach, which is adding a rake task to run data migrations after a migration instead: https://github.com/DFE-Digital/apply-for-teacher-training/pull/4061
After we agree on the approach, definitely an good idea to add to the template as well

benswannack: Ok, I'll close this PR and leave it up to #4061 to decide what to put in the PR template (@despo)

PR #4058, opened 2021-02-11
Change govuk_link_to/govuk_button_link_to be consistent with Rails

stevehook:
## Context

The `govuk_link_to` and `govuk_button_link_to` helpers should follow the principle of least surprise by being consistent with Rails. The `body` param should be 'optional' when a block is given (as an alternative way to specify the content of the link). 

- Because we cannot make a first parameter optional in the true sense we shift the params by interpreting the first as the `url`, second as `html_options` etc. when a block is given just the way Rails does. https://github.com/rails/rails/blob/b9ca94caea2ca6a6cc09abaffaad67b447134079/actionview/lib/action_view/helpers/url_helper.rb#L196-L206
- Update references to these helpers that specify a block.

## Guidance to review

- Anything missing?

## Link to Trello card

https://trello.com/c/Qw9Q80Bq/2983-dev-refactor-govuklinkto-and-govukbuttonlinkto-to-make-body-param-optional

## Things to check

### DIFF
--- a/app/helpers/view_helper.rb
+++ b/app/helpers/view_helper.rb
@@ -1,5 +1,12 @@
 module ViewHelper
-  def govuk_link_to(body, url, html_options = {}, &_block)
+  def govuk_link_to(body = nil, url = nil, html_options = nil, &block)
+    if block_given?
+      html_options = url
+      url = body
+      body = block
+    end
+    html_options ||= {}
+
     html_options[:class] = prepend_css_class('govuk-link', html_options[:class])
 
     return link_to(url, html_options) { yield } if block_given?
@@ -36,7 +43,14 @@ module ViewHelper
     mail_to('becomingateacher@digital.education.gov.uk', name.html_safe, html_options)
   end
 
-  def govuk_button_link_to(body, url, html_options = {}, &_block)
+  def govuk_button_link_to(body = nil, url = nil, html_options = nil, &block)
+    if block_given?
+      html_options = url
+      url = body
+      body = block
+    end
+    html_options ||= {}
+
     html_options = {
       class: prepend_css_class('govuk-button', html_options[:class]),
       role: 'button',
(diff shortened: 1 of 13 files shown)


app/helpers/view_helper.rb
```diff
@@ -36,7 +43,14 @@ def bat_contact_mail_to(name = 'becomingateacher<wbr>@digital.education.gov.uk',
     mail_to('becomingateacher@digital.education.gov.uk', name.html_safe, html_options)
   end
 
-  def govuk_button_link_to(body, url, html_options = {}, &_block)
+  def govuk_button_link_to(body = nil, url = nil, html_options = nil, &block)
+    if block_given?
+      body = block
```
cwrw, line 48: I think we will have to move setting the `block` till the end since setting it in the beginning is setting `url` and `html_options` to the `block` as well, also might need to switch the order to avoid `html_options` being set to the body as so: 
```
if block_given?
  html_options = url
  url = body
  body = block
end
```

app/helpers/view_helper.rb
```diff
@@ -1,5 +1,12 @@
 module ViewHelper
-  def govuk_link_to(body, url, html_options = {}, &_block)
+  def govuk_link_to(body = nil, url = nil, html_options = nil, &block)
+    if block_given?
```
cwrw, line 3: same here: 
```
if block_given?
  html_options = url
  url = body
  body = block
end
```

cwrw (commented):
looks good, just an ordering of setting the variables

app/helpers/view_helper.rb
stevehook: thanks @crrw, good spot, this was a last minute change because Rubocop doesn't like parallel assignment and I was just wondering why all the tests had broken.

cwrw (approved):

malcolmbaig (approved):

PR #4067, opened 2021-02-12
Fix sentry error 2214497591

despo:
## Context
Fix for sentry error: https://sentry.io/organizations/dfe-bat/issues/2214497591/?project=1765973&referrer=slack

- Update DateValidationHelper to only run `start_date_before_end_date` check if both start and end date are valid dates
- Adds specs to verify updates on DateValidator
- Adds specs to verify `start_date_before_end_date`

### DIFF
--- a/app/helpers/date_validation_helper.rb
+++ b/app/helpers/date_validation_helper.rb
@@ -18,6 +18,8 @@ module DateValidationHelper
   end
 
   def start_date_before_end_date
+    return unless start_date.is_a?(Date) && end_date.is_a?(Date)
+
     errors.add(:start_date, :before) unless start_date <= end_date
   end
 end
--- a/app/validators/date_validator.rb
+++ b/app/validators/date_validator.rb
@@ -4,9 +4,7 @@ class DateValidator < ActiveModel::EachValidator
   MIN_AGE = 16
 
   def validate_each(record, attribute, value)
-    return if value.blank?
-
-    month_and_year_validations(record, attribute, value) if options[:month_and_year]
+    return if value.blank? || (blank?(value) && !options[:presence])
 
     date_validations(record, attribute, value) if !record.errors.keys.include?(attribute)
 
@@ -22,15 +20,11 @@ class DateValidator < ActiveModel::EachValidator
   def date_validations(record, attribute, value)
     return record.errors.add(attribute, :blank_date, article: article(attribute), attribute: humanize(attribute)) if options[:presence] && blank?(value)
 
-    return record.errors.add(attribute, :blank_date_fields, attribute: humanize(attribute), fields: blank_fields(value).to_sentence) if blank_fields(value).any?
+    return record.errors.add(attribute, :blank_date_fields, attribute: humanize(attribute), fields: blank_fields(value).to_sentence) if !blank?(value) && blank_fields(value).any?
 
     return record.errors.add(attribute, invalid_date_locale(options), article: article(attribute), attribute: humanize(attribute)) if is_invalid?(value)
 
     record.errors.add(attribute, :future, article: article(attribute), attribute: humanize(attribute)) if value > Time.zone.today && options[:future]
-  end
-
-  def month_and_year_validations(record, attribute, value)
-    record.errors.add(attribute, :blank_date, article: article(attribute), attribute: humanize(attribute)) if options[:presence] && empty_date_and_month?(value)
   end
 
 private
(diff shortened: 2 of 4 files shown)


app/validators/date_validator.rb
```diff
@@ -22,17 +20,13 @@ def date_of_birth_validations(record, attribute, value)
   def date_validations(record, attribute, value)
     return record.errors.add(attribute, :blank_date, article: article(attribute), attribute: humanize(attribute)) if options[:presence] && blank?(value)
 
-    return record.errors.add(attribute, :blank_date_fields, attribute: humanize(attribute), fields: blank_fields(value).to_sentence) if blank_fields(value).any?
+    return record.errors.add(attribute, :blank_date_fields, attribute: humanize(attribute), fields: blank_fields(value).to_sentence) if !blank?(value) && blank_fields(value).any?
 
     return record.errors.add(attribute, invalid_date_locale(options), article: article(attribute), attribute: humanize(attribute)) if is_invalid?(value)
 
     record.errors.add(attribute, :future, article: article(attribute), attribute: humanize(attribute)) if value > Time.zone.today && options[:future]
```
duncanjbrown, line 27: By the by, but does this one belong here with the other generic "is this a well-formed date" validations? And should there be a corresponding one to assert that a date should be in the past?

app/validators/date_validator.rb
```diff
@@ -22,17 +20,13 @@ def date_of_birth_validations(record, attribute, value)
   def date_validations(record, attribute, value)
     return record.errors.add(attribute, :blank_date, article: article(attribute), attribute: humanize(attribute)) if options[:presence] && blank?(value)
 
-    return record.errors.add(attribute, :blank_date_fields, attribute: humanize(attribute), fields: blank_fields(value).to_sentence) if blank_fields(value).any?
+    return record.errors.add(attribute, :blank_date_fields, attribute: humanize(attribute), fields: blank_fields(value).to_sentence) if !blank?(value) && blank_fields(value).any?
```
duncanjbrown, line 23: So we were accidentally marking missing date fields if nothing at all was submitted, which is basically the same as enforcing presence?

app/validators/date_validator.rb
```diff
@@ -52,13 +46,10 @@ def blank_fields(value)
...
-    value.to_h.slice(:day, :month, :year).all? { |_, v| v.blank? }
-  end
-
-  def empty_date_and_month?(value)
-    return false if value.is_a?(Date)
+    date_fields = %i[day month year]
+    date_fields -= [:day] if options[:month_and_year]
 
-    value.to_h.slice(:month, :year).all? { |_, v| v.blank? }
+    value.to_h.slice(*date_fields).all? { |_, v| v.blank? }
```
duncanjbrown, line 52: Yes — so we only check for blanks in the fields we ought to have received

app/helpers/date_validation_helper.rb
```diff
@@ -18,6 +18,8 @@ def valid_or_invalid_date(year, month)
   end
 
   def start_date_before_end_date
+    return unless start_date.is_a?(Date) && end_date.is_a?(Date)
```
duncanjbrown, line 21: Right, so if this is a `Date` rather than a `Struct` it means we've successfully parsed it? I'd almost go so far as to name the Struct `IncompleteDate` or sth  in `DateValidationHelper`?

duncanjbrown (approved):
Looks good — couple of thinking-out-louds inline

app/validators/date_validator.rb
despo: Exactly.

app/validators/date_validator.rb
despo: Yes. Before I was attempting to do this separately for month_and_date validations which is what caused the problem as it would end up running this validation as part of the overall date checks when that did not fail.

app/helpers/date_validation_helper.rb
despo: The models only run this check if there are no errors on the `start_date` or `end_date` fields already. As we can potentially have an empty start_date struct and no errors due to the start_date not being mandatory. The same applies for end_date. 

https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/forms/candidate_interface/work_experience_form.rb#L19

app/validators/date_validator.rb
despo: It is optional however still part of the main date object validations, so yes I think this is a good place for it.

The validator was created based on existing validations that could be found in the code. As the only validation checks relate to a date having to be in the past were quite specific to the models they weren't a good fit for here. Maybe something to look into a bit more now that the interview wizard is in, as it is the one place that has a past check.

PR #4069, opened 2021-02-12
Temporarily add personal details to E&D export

raamSoftwire:
## Context

This is a temporary adjustment of the E&D export to include personal details of candidates so that the user researchers on the Impact Squad can contact candidates for interviews.

Because this is a large export we decided not to go through the rails console as we have done previously. This means we will need to release this change, take a download, and then revert the changes.

I've added name, email address, phone number, and whether or not they are happy to be contacted. `Consent to be contacted`  comes from the `application_feedback` table. We thought it might be worth being aware of those candidates who would prefer not to be contacted.

## Guidance to review

Are there data security reasons why this is a really bad idea? Is there a better way?

## Link to Trello card

https://trello.com/c/pUUyeFDN/84-adding-contact-details-to-ed-data-extract (This is on the Impact Squad board)

## Things to check

### DIFF
--- a/app/services/support_interface/equality_and_diversity_export.rb
+++ b/app/services/support_interface/equality_and_diversity_export.rb
@@ -1,9 +1,13 @@
 module SupportInterface
   class EqualityAndDiversityExport
     def data_for_export
-      data_for_export = application_forms.includes(:application_choices).map do |application_form|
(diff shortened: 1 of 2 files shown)


davidgisbey (approved):
lgtm

raamSoftwire:
I'll merge this then create a separate PR for reverting the changes so that we can have it ready to go once we've taken the export.

PR #4070, opened 2021-02-12
Interview wizard fixes

benswannack:
## Context
Initial review of the interviews feature resulted in a few bugs being spotted

1. The validation error summary appears above the h1 in the forms
2. Validate the length of the freetext fields.
3. Add visually hidden text for screen reader users to the change links to go back to the form
4. Validation messages in the error summary appear in the correct order. They are in the same order as they appear in the form now
5. Fix error when leaving the time blank in the form (and nothing else)
6. The check answers page displays times in the correct format. 8 in the morning is now displayed as 8am rather than 8:00am
7. The check answers page now displays the text "None" in the "Additional details" row if the user doesn't enter a value

Still to fix:
- The change links should anchor link to the first element (e.g first radio button) in the field for which it is changing.
## Guidance to review
Per commit
Test in the review app!

## Link to Trello card
https://trello.com/c/DZ1Dz10H/3369-interview-create-edit-form-bugs

## Things to check

### DIFF
--- a/app/components/provider_interface/interview_details_component.rb
+++ b/app/components/provider_interface/interview_details_component.rb
@@ -21,6 +21,7 @@ module ProviderInterface
       {
         key: key,
         value: value,
+        action: "Change #{key.downcase}",
         change_path: change_path,
       }
     end
--- a/app/components/provider_interface/interview_details_component.rb
+++ b/app/components/provider_interface/interview_details_component.rb
@@ -12,7 +12,7 @@ module ProviderInterface
       time_row = build_row('Time', @interview_form.date_and_time.to_s(:govuk_time))
       organisation_row = build_row('Organisation carrying out interview', @interview_form.provider.name)
       location_row = build_row('Address or online meeting details', @interview_form.location)
-      details_row = build_row('Additional details', @interview_form.additional_details)
+      details_row = build_row('Additional details', @interview_form.additional_details.presence || 'None')
 
       [date_row, time_row, organisation_row, location_row, details_row]
     end
--- a/app/components/provider_interface/interview_details_component.rb
+++ b/app/components/provider_interface/interview_details_component.rb
@@ -8,29 +8,72 @@ module ProviderInterface
     end
 
     def rows
-      date_row = build_row('Date', @interview_form.date_and_time.to_s(:govuk_date))
-      time_row = build_row('Time', @interview_form.date_and_time.to_s(:govuk_time))
-      organisation_row = build_row('Organisation carrying out interview', @interview_form.provider.name)
-      location_row = build_row('Address or online meeting details', @interview_form.location)
-      details_row = build_row('Additional details', @interview_form.additional_details.presence || 'None')
-
-      [date_row, time_row, organisation_row, location_row, details_row]
+      [
+        date_row,
+        time_row,
+        organisation_row,
+        location_row,
+        details_row,
+      ]
     end
 
-    def build_row(key, value)
+  private
+
+    def date_row
+      build_row(:date, @interview_form.date_and_time.to_s(:govuk_date))
+    end
+
+    def time_row
+      build_row(:time, @interview_form.date_and_time.to_s(:govuk_time))
+    end
+
+    def organisation_row
+      build_row(:provider_id, @interview_form.provider.name)
+    end
+
+    def location_row
+      build_row(:location, @interview_form.location)
+    end
+
+    def details_row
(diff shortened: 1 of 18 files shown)


config/initializers/date_formats.rb
```diff
@@ -13,6 +13,8 @@
              '%e %B %Y at %l%P (midday)'
            elsif time >= time.midnight && time <= time.midnight.end_of_minute
              '%e %B %Y at %l%P (midnight)'
+           elsif time.min.zero?
+             '%e %B %Y at %l%P'
```
benswannack, line 17: @adamsilver this updates the time displays everywhere in the application (Apply & Manage). So anything that said `1st January 2021 at 11:00am` will now say `1st January 2021 at 11am` and `1:00pm` will change to just `1pm`. Does that sound OK?

app/forms/provider_interface/interview_wizard.rb
```diff
@@ -20,7 +20,8 @@ class InterviewWizard
                                        unless: ->(c) { %i[date time].any? { |d| c.errors.keys.include?(d) } }
     validate :date_after_rbd_date, if: %i[date_and_time date_and_time_in_future]
     validates :provider_id, presence: true, if: %i[application_choice provider_user multiple_application_providers?]
-    validates :location, presence: true
+    validates :location, presence: true, length: { maximum: 10240 }
```
cwrw, line 23: Although the maximum has been set, whilst testing locally this isn't mitigating the following error: 
```
2021-02-15 17:04:44 +0200 HTTP parse error, malformed request ("GET /provider/applications/17/interviews/new/check" - (-)): #<Puma::HttpParserError: HTTP element QUERY_STRING is longer than the (1024 * 10) allowed length (was 15164)>
```
As it isn't even reaching the validation stage yet, and falls over before that. I was wondering around us using POST vs GET which bypasses this issue? Let me know your thoughts

benswannack: Yes you're right, and the validation wouldn't even prevent that – it's possible to be under that number and still cause the error. The number here is chosen to match other free text fields in the API.

I'll card up and we can fix this bug separately

app/components/provider_interface/interview_details_component.rb
```diff
@@ -8,29 +8,72 @@ def initialize(interview_form, interview = nil)
...
-        action: "Change #{key.downcase}",
-        change_path: change_path,
+        action: key.downcase,
+        change_path: change_path(field),
       }
     end
 
-    def change_path
+    def key_for_field(field)
+      if %i[date provider_id].include?(field)
```
cwrw, line 53: We will need to include the provider id in the anchor as the form fields include those ids, example:
```
<input id="provider-interface-interview-wizard-provider-id-319-field" class="govuk-radios__input" type="radio" value="319" checked="checked" name="provider_interface_interview_wizard[provider_id]">
```

app/forms/provider_interface/interview_wizard.rb
```diff
@@ -39,7 +42,7 @@ def date
     end
 
     def date_and_time
-      Time.zone.local(date.year, date.month, date.day, parsed_time.hour, parsed_time.min) if date.is_a?(Date)
+      Time.zone.local(date.year, date.month, date.day, parsed_time.hour, parsed_time.min) if date.is_a?(Date) && parsed_time.is_a?(Time)
```
cwrw, line 45: If I enter a time as `6767` I am getting the following error: `argument out of range` when it attempts to parse the time. Maybe we should also ensure that the time is within a parseable range?

cwrw (commented):
Looks good, just a few minor comments that cropped up while testing

app/components/provider_interface/interview_details_component.rb
benswannack: Good spot!

cwrw: Ah I actually noticed that the error messages as well don't have the correct anchor to this section, should we tackle that in  this PR or add a new ticket?
Screenshot 2021-02-16 at 14 11 32

benswannack: Easy to fix the latter issue – I'll do it here.

app/forms/provider_interface/interview_wizard.rb
benswannack: This should be addressed now – we now check that the time is in the right format before trying to parse it

cwrw (approved):
LGTM! Also can see the length validation errors now in the form since the POST/GET work has also been merged 👍

PR #4071, opened 2021-02-12
Use simple past in Slack updates

duncanjbrown:
eg "withdrew" means the candidate withdrew at some point in the past (which will always be true) but "has withdrawn" implies it happened just now (which won’t).


AgaDufrat (approved):
English and it's 12 tenses 🤯

duncanjbrown:
God is it that many 😬

PR #4072, opened 2021-02-12
Enforce `start_date` presence on candidate work history

despo:
## Context

Enforce `start_date` presence check on work experience and restructured work history job.
The check was missed when switching to the DateValidator as it was previously done as part of checking for the validity of the date object.

## Things to check


davidgisbey (approved):
lgtm once rubocop is happy 👍

PR #4078, opened 2021-02-14
Enforce use of govuk_link_to et al

duncanjbrown:
## Context

Follow up to #4060, where @paulrobertlloyd came up with the ingenious idea of using `erblint` to check that we're using the proper `govuk_link_to` helper for links.

Add Rubocop cops for `govuk_link_to`, `govuk_button_link_to`, `govuk_back_link_to`, `govuk_button_to` and `govuk_submit`. Introduce `govuk_footer_link_to` and check for that as well. Then fix all the violations.

## Guidance to review

What do you think? We've ignored a tiny set of files where it's not appropriate to use these helpers, but apart from that it seems that it's more or less always right to use the `govuk_` helper.   

## Link to Trello card

N/A

## Things to check

### DIFF
--- a/app/helpers/view_helper.rb
+++ b/app/helpers/view_helper.rb
@@ -69,6 +69,11 @@ module ViewHelper
   end
 
   def govuk_button_to(name, options = {}, html_options = {})
+    if block_given?
+      html_options = options
+      options = name
+    end
+
     html_options = {
       class: prepend_css_class('govuk-button', html_options[:class]),
       role: 'button',
@@ -76,7 +81,7 @@ module ViewHelper
       draggable: false,
     }.merge(html_options)
 
-    return button_to(name, options, html_options) { yield } if block_given?
+    return button_to(options, html_options) { yield } if block_given?
 
     button_to(name, options, html_options)
   end
--- a/app/helpers/view_helper.rb
+++ b/app/helpers/view_helper.rb
@@ -76,6 +76,8 @@ module ViewHelper
       draggable: false,
     }.merge(html_options)
 
+    return button_to(name, options, html_options) { yield } if block_given?
+
     button_to(name, options, html_options)
   end
 
--- a/app/helpers/view_helper.rb
+++ b/app/helpers/view_helper.rb
@@ -33,16 +33,6 @@ module ViewHelper
     link_to(body, url, class: classes)
   end
 
-  def break_email_address(email_address)
-    email_address.gsub(/@/, '<wbr>@').html_safe
-  end
-
-  def bat_contact_mail_to(name = 'becomingateacher<wbr>@digital.education.gov.uk', html_options: {})
-    html_options[:class] = prepend_css_class('govuk-link', html_options[:class])
-
-    mail_to('becomingateacher@digital.education.gov.uk', name.html_safe, html_options)
-  end
-
   def govuk_button_link_to(body = nil, url = nil, html_options = nil, &block)
     if block_given?
       html_options = url
@@ -79,6 +69,16 @@ module ViewHelper
     return button_to(options, html_options) { yield } if block_given?
 
     button_to(name, options, html_options)
+  end
+
+  def break_email_address(email_address)
+    email_address.gsub(/@/, '<wbr>@').html_safe
+  end
+
+  def bat_contact_mail_to(name = 'becomingateacher<wbr>@digital.education.gov.uk', html_options: {})
+    html_options[:class] = prepend_css_class('govuk-link', html_options[:class])
+
+    mail_to('becomingateacher@digital.education.gov.uk', name.html_safe, html_options)
   end
 
   def submitted_at_date
--- a/app/helpers/view_helper.rb
+++ b/app/helpers/view_helper.rb
@@ -63,7 +63,7 @@ module ViewHelper
     link_to(body, url, html_options)
   end
 
-  def govuk_button_to(name, options = {}, html_options = {})
+  def govuk_button_to(name, options = {}, html_options = {}, &_block)
     if block_given?
       html_options = options
       options = name
--- a/app/helpers/view_helper.rb
+++ b/app/helpers/view_helper.rb
@@ -33,6 +33,11 @@ module ViewHelper
     link_to(body, url, class: classes)
   end
 
+  def govuk_footer_link_to(body = nil, url = nil, html_options = {}, &block)
+    html_options[:class] = prepend_css_class('govuk-footer__link', html_options[:class])
+    govuk_link_to(body, url, html_options, &block)
+  end
+
   def break_email_address(email_address)
     email_address.gsub(/@/, '<wbr>@').html_safe
   end
--- a/app/helpers/view_helper.rb
+++ b/app/helpers/view_helper.rb
@@ -33,11 +33,6 @@ module ViewHelper
     link_to(body, url, class: classes)
   end
 
-  def govuk_footer_link_to(body = nil, url = nil, html_options = {}, &block)
-    html_options[:class] = prepend_css_class('govuk-footer__link', html_options[:class])
-    govuk_link_to(body, url, html_options, &block)
-  end
-
   def break_email_address(email_address)
     email_address.gsub(/@/, '<wbr>@').html_safe
   end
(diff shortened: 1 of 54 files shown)


paulrobertlloyd:
Awesome! I don’t see a helper for `govuk_back_link_to` though?

duncanjbrown:
> Awesome! I don’t see a helper for `govuk_back_link_to` though?

We already had one!

paulrobertlloyd:
@duncanjbrown I’d just double-check that one… that helper enables a certain type of behaviour (and uses JavaScript) and I'm not sure if recommending that helper is always the right thing to do. Whereas the other helpers are essentially shorthand's that add a CSS class only.

paulrobertlloyd:
@duncanjbrown Fixed the failing test in e2c76041cbf2868f30c49138e414ad70c5c562b0 – it was an issue with the `govuk_button_to` helper not accepting block content.

I thought it might be the `govuk_back_link_to` helpers, but even though not, might still want to double check that we want to always use this helper for back links. This was something Theo @george-bit worked on, and it still confuses me to be honest as to how it actually works 😅

paulrobertlloyd:
Hmmm, still failing…

george-bit:
> @duncanjbrown Fixed the failing test in [e2c7604](https://github.com/DFE-Digital/apply-for-teacher-training/commit/e2c76041cbf2868f30c49138e414ad70c5c562b0) – it was an issue with the `govuk_button_to` helper not accepting block content.
> 
> I thought it might be the `govuk_back_link_to` helpers, but even though not, might still want to double check that we want to always use this helper for back links. This was something Theo @george-bit worked on, and it still confuses me to be honest as to how it actually works 😅

@paulrobertlloyd we designed the `govuk_back_link_to` be applicable across the entire app. Basically, it assigns the path (unless overwritten) first and foremost from the previous value HTTP_REFFER in the first instance. If this doesn't exist it then assigns the path to `javascript:history.back()`. We also assign a 'no js' class to hide the link from view. We then remove this class in the JS initialiser to ensure that if it's a JS link it will only show when JS is enabled. 

Happy to talk it through further if you aren't sure :)

app/views/candidate_interface/submitted_application_form/start_apply_again.html.erb
```diff
@@ -16,10 +16,12 @@
     </p>
     <p class="govuk-body">You can only apply to one course at a time at this stage of your application.</p>
 
-    <%= button_to 'Start now', candidate_interface_apply_again_path, class: 'govuk-button govuk-button govuk-!-margin-bottom-5' %>
```
cwrw, line 19: can we keep the class `govuk-!-margin-bottom-5`?

app/helpers/view_helper.rb
```diff
@@ -76,6 +76,8 @@ def govuk_button_to(name, options = {}, html_options = {})
       draggable: false,
     }.merge(html_options)
 
+    return button_to(name, options, html_options) { yield } if block_given?
```
cwrw, line 79: I think we might need something similar to https://github.com/rails/rails/blob/b9ca94caea2ca6a6cc09abaffaad67b447134079/actionview/lib/action_view/helpers/url_helper.rb#L300 or https://github.com/DFE-Digital/apply-for-teacher-training/pull/4078/commits/61f33f20eff4e3e9ecb8800df4af8189683f5f8d#diff-20b0835bb2cec2d019f0fc54cbaa1457a740c573dba553c3333d7521de832092L52 where we need to check if a block is given, and then set the variables accordingly. We could also allow a block to be passed into the method. Might look something like this:

```
def govuk_button_to(name, options = {}, html_options = {}, &block)
  if block_given?
    html_options = options
    options = name
  end

  html_options = {
      class: prepend_css_class('govuk-button', html_options[:class]),
      role: 'button',
      data: { module: 'govuk-button' },
      draggable: false,
    }.merge(html_options)

  return button_to(options, html_options) { yield } if block_given?
  button_to(name, options, html_options)
end
```

app/views/candidate_interface/submitted_application_form/start_apply_again.html.erb
paulrobertlloyd: I looked at this and it only adds (or was it removes) an additional 5px, so I'd rather defer to the standard margin.

cwrw:
Looks good, just the comment around the helper. I was wondering if we were going to extend this to `- 'app/components/**/*'` next?

paulrobertlloyd:
> Looks good, just the comment around the helper. I was wondering if we were going to extend this to `- 'app/components/**/*'` next?

I’ve been using the `--lint-all` flag, so components should have been covered by this. But can double-check!

cwrw:
@paulrobertlloyd I think you might need to include them in the `.rubocop` file as so:
```
Govuk:
  Include:
    - 'app/views/**/*'
    - 'app/components/**/*'
```
when I did I got `36 error(s) were found in ERB files`

paulrobertlloyd:
> when I did I got `36 error(s) were found in ERB files`

Oh ’eek! We run ERB Lint on those, but not the custom cops! Will address now. Thanks!

paulrobertlloyd:
@cwrw @duncanjbrown 

* Updated Rubocop’s config to include components, and specifically exclude checks for `govuk_link_to` for a limited number of ERB files. And, bugger, the change to the `govuk_button_to` helpers is in this commit too: 77e2e4e2d2b7d1bf847a6e353bcc4d4c083cfec0

* Fixed 36 violations in components in this commit: aca542668fb4d6076781c8fbdfda891e2eb4c058

app/helpers/view_helper.rb
```diff
@@ -64,13 +69,20 @@ def govuk_button_link_to(body = nil, url = nil, html_options = nil, &block)
   end
 
   def govuk_button_to(name, options = {}, html_options = {})
```
cwrw, line 71: can we include `&block` as so:
```
def govuk_button_to(name, options = {}, html_options = {}, &block)
```
to help indicate that this method accepts a block?

app/helpers/view_helper.rb
```diff
@@ -33,6 +33,11 @@ def govuk_back_link_to(url = :back, body = 'Back')
     link_to(body, url, class: classes)
   end
 
+  def govuk_footer_link_to(body = nil, url = nil, html_options = {}, &block)
+    html_options[:class] = prepend_css_class('govuk-footer__link', html_options[:class])
+    govuk_link_to(body, url, html_options, &block)
```
cwrw, line 38: I think we need something similar to other methods here as well, to make sure we set the correct class if a block is given:
```
def govuk_footer_link_to(body = nil, url = nil, html_options = {}, &block)
    if block_given?
      html_options = url
      url = body
    end

    html_options[:class] = prepend_css_class('govuk-footer__link', html_options[:class])

    return govuk_link_to(url, html_options) { yield } if block_given?

    govuk_link_to(body, url, html_options)
  end
```

cwrw (commented):
Looks good! just some fixes to the view helper, happy to approve after

app/helpers/view_helper.rb
paulrobertlloyd: I originally added this, but Rubocop complained:

```
app/helpers/view_helper.rb:71:63: W: Lint/UnusedMethodArgument: Unused method argument - block. If it's necessary, use _ or _block as an argument name to indicate that it won't be used. (https://rubystyle.guide#underscore-unused-vars)
  def govuk_button_to(name, options = {}, html_options = {}, &block)
                                                              ^^^^^
```

Is there a way to make the method explicit, but also appease our linting overlords?

paulrobertlloyd: Oh, maybe I should just follow it’s advice and prefix this value with `_`…

paulrobertlloyd:
@cwrw Have updated helper in line with yours (and Rubocop’s) suggestion.

@duncanjbrown Have removed `govuk_footer_link_to` helper, and instead excluding testing these links like we do for other bespoke link classes within `govuk-frontend`.

Have also grouped the different link helpers together.

cwrw (approved):
LGTM 👏  thanks for addressing all feedback

PR #4079, opened 2021-02-15
Add notifications export with daily schedule

SamJamCul:
## Context

https://trello.com/c/oyKg6VVy/3329-create-csv-extract-for-minimal-notifications-analysis

Creating a new export, which breaks down how many users have notifications on for each provider, and how many notifiable events are taking place.

Added a new export in the export list

image

## Guidance to review

Try and run the export from support/performance/data-exports/new (it'll be at the bottom of the page)

## Link to Trello card

https://trello.com/c/oyKg6VVy/3329-create-csv-extract-for-minimal-notifications-analysis

## Things to check

### DIFF
--- a/app/services/support_interface/notifications_export.rb
+++ b/app/services/support_interface/notifications_export.rb
@@ -1,0 +1,57 @@
+module SupportInterface
+  class NotificationsExport
+    def data_for_export
+      courses = Course.where(recruitment_cycle_year: RecruitmentCycle.current_year, open_on_apply: true)
+      providers = Provider.includes(:provider_users, :application_choices).joins(:courses).merge(courses).group('providers.id')
+      providers.flat_map do |provider|
+        data_for_provider(provider, provider_users_with_make_decisions(provider), visible_applications(provider))
+      end
+    end
+
+  private
+
+    def data_for_provider(provider, users_with_make_decisions, applications)
+      {
+        provider_code: provider.code,
+        provider_name: provider.name,
+        applications_received: applications.count,
+        applications_awaiting_decisions: awaiting_decisions_count(applications),
+        applications_receiving_decisions: receiving_decisions_count(applications) - rbd_count(applications) - applications.count(&:withdrawn?),
+        applications_rbd: rbd_count(applications),
+        applications_withdrawn: applications.count(&:withdrawn?),
+        number_of_provider_users: provider.provider_users.count,
+        users_with_make_decisions: users_with_make_decisions.count,
+        users_with_make_decisions_and_notifications_disabled: users_with_make_decisions.count { |u| !u.send_notifications },
+        users_with_make_decisions_and_notifications_enabled: users_with_make_decisions.count(&:send_notifications),
+      }
+    end
+
+    def provider_users_with_make_decisions(provider)
+      provider.provider_permissions.includes(:provider_user).where(make_decisions: true).map(&:provider_user)
+    end
+
+    def visible_applications(provider)
+      provider.application_choices.where(status: ApplicationStateChange::STATES_VISIBLE_TO_PROVIDER)
+    end
+
+    def awaiting_decisions_count(applications)
+      applications.count do |application|
+        ApplicationStateChange::DECISION_PENDING_STATUSES.include?(application.status.to_sym)
+      end
+    end
+
+    def receiving_decisions_count(applications)
+      applications.count do |application|
+        decision_receiving_states.include?(application.status.to_sym)
+      end
+    end
+
+    def rbd_count(applications)
+      applications.count { |a| a.rejected? && a.rejected_by_default? }
+    end
+
+    def decision_receiving_states
+      ApplicationStateChange::STATES_VISIBLE_TO_PROVIDER - ApplicationStateChange::DECISION_PENDING_STATUSES
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


app/services/support_interface/notifications_export.rb
```diff
@@ -0,0 +1,35 @@
+module SupportInterface
+  class NotificationsExport
+    def data_for_export
+      providers = Provider.all.includes(:provider_users, :application_choices)
```
despo, line 4: This should only return retrieving providers who have open courses in the current recruitment cycle.
To do that you need to check for `open_on_apply` and `recruitment_cycle_year` on the Course model. I have sometime today if you want to pair on this.

SamJamCul: Sorted, added a condition, thanks for the help!

despo:
Looks good but will let someone else approve as we paired on it!

config/clock.rb
```diff
@@ -50,4 +50,9 @@ class Clock
       TeacherTrainingPublicAPI::SyncAllProvidersAndCoursesWorker.perform_async
     end
   end
+
+  every(1.day, 'Generate export for Notifications', at: '23:59') do
```
cwrw, line 54: Although this is specified to run the same time as the TAD export, maybe we leave a few minutes between them for performance purposes?

spec/services/support_interface/notifications_export_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    let(:course_options) { courses.map { |course| create(:course_option, course: course) } }
+    let(:courses) { create_list(:course, 3, :open_on_apply, provider: provider) }
+    let(:provider) { create(:provider) }
+
+    before do
+      create(:provider_user, :with_make_decisions, providers: [provider], send_notifications: false)
+      create(:provider_user, :with_make_decisions, providers: [provider], send_notifications: true)
+      create(:provider_user, providers: [provider], send_notifications: true)
+    end
+
```
cwrw, line 17: Should we also add the shared example `it_behaves_like 'a data export'` which tests documentation?

app/services/support_interface/notifications_export.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    def data_for_provider(provider, users_with_make_decisions, applications)
+      {
+        'Provider Code' => provider.code,
+        'Provider' => provider.name,
+        'No. Applications Received' => applications.count,
+        'No. Applications awaiting decisions' => applications.count(&:awaiting_provider_decision?),
+        'No. applications receiving decisions' => receiving_decisions_count(applications) - rbd_count(applications),
+        "No. applications RBD'd" => rbd_count(applications),
+        'No. provider users' => provider.provider_users.count,
```
cwrw, line 21: I noticed we are missing the following column "No. applications withdrawn" from the spreadsheet, is that intentional?

app/services/support_interface/notifications_export.rb
```diff
@@ -0,0 +1,50 @@
...
+        'No. users with make_decisions' => users_with_make_decisions.count,
+        'No. of users with make_decisions and notifications disabled' => users_with_make_decisions.count(&:send_notifications),
```
cwrw, line 23: Should this be the opposite with the below as it's notifications disabled 🤔 ?

app/services/support_interface/notifications_export.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+  private
+
+    def data_for_provider(provider, users_with_make_decisions, applications)
+      {
+        'Provider Code' => provider.code,
+        'Provider' => provider.name,
+        'No. Applications Received' => applications.count,
+        'No. Applications awaiting decisions' => applications.count(&:awaiting_provider_decision?),
```
mnacos, line 18: Shouldn't 'applications awaiting decisions' include applications at the `interviewing` state?

mnacos (commented):
Looks good to me. Added a suggestion around the `interviewing` state. Maybe add a few more cases to the spec?

config/clock.rb
SamJamCul: Shifted it to 23.57

app/services/support_interface/notifications_export.rb
SamJamCul: Good spot, was not intentional! I've also excluded the withdrawn applications from the set of ones receiving decisions.

app/services/support_interface/notifications_export.rb
SamJamCul: good spot, flipped them around

spec/services/support_interface/notifications_export_spec.rb
SamJamCul: Done, took a bit of refactoring! I'm not completely sure if the outputs for most of the rows are integers or strings. I've assumed that they're integers because they're all the result of a `.count`. The behaviour test doesn't seem to mind what typing I put!

app/services/support_interface/notifications_export.rb
SamJamCul: Good spot - I've modified it to use the awaiting decisions grouping instead, and I've added a couple of extra application choices to the test which help tease out the differences

cwrw (approved):
👍  looks good and can see the documentation on the FE as well

PR #4081, opened 2021-02-15
Make sure aria label is on NAV landmark element and export data item is aligned right

adamsilver:

- make sure NAV landmark labels are in the correct place and read nicely
- make sure ‘export data‘ is aligned right on big screens

## Link to Trello card

https://trello.com/c/IsoTnh7q/3376-fix-navigation-landmarks-and-layout

### DIFF
--- a/app/components/primary_navigation_component.html.erb
+++ b/app/components/primary_navigation_component.html.erb
@@ -1,11 +1,11 @@
-<div class="moj-primary-navigation govuk-!-display-none-print">
-  <div class="moj-primary-navigation__container">
-    <div class="moj-primary-navigation__nav">
-      <nav class="moj-primary-navigation" aria-label="Primary navigation">
-        <ul class="moj-primary-navigation__list">
+<div class="app-primary-navigation govuk-!-display-none-print">
+  <div class="app-primary-navigation__container">
+    <div class="app-primary-navigation__nav">
+      <nav class="app-primary-navigation" aria-label="main menu">
+        <ul class="app-primary-navigation__list">
           <% @navigation_items.each do |nav_item| %>
-            <li class="moj-primary-navigation__item">
-              <a class="moj-primary-navigation__link"
+            <li class="app-primary-navigation__item <%= nav_item.classes %>">
+              <a class="app-primary-navigation__link"
                 <%= nav_item.active ? 'aria-current=page' : '' %>
                  href="<%= nav_item.href %>">
                  <%= nav_item.text %>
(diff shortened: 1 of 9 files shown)


app/components/primary_navigation_component.html.erb
```diff
@@ -1,11 +1,11 @@
-<div class="moj-primary-navigation govuk-!-display-none-print">
-  <div class="moj-primary-navigation__container">
-    <div class="moj-primary-navigation__nav">
-      <nav class="moj-primary-navigation" aria-label="Primary navigation">
-        <ul class="moj-primary-navigation__list">
+<div class="app-primary-navigation govuk-!-display-none-print">
+  <div class="app-primary-navigation__container">
+    <div class="app-primary-navigation__nav">
```
cwrw, line 3: I noticed that we aren't defining the following class: `app-primary-navigation__nav` in `_primary-navigation.scss 
 ` as opposed to `moj-primary-navigation__nav`, is this because of it forcing the left alignment of all tabs? Are there any repercussions of leaving it out of the new primary navigation?
old class definition for reference:
 ```
 .moj-primary-navigation__nav {
  text-align: left;
  @include govuk-media-query($from: desktop) {
    display: inline-block;
    vertical-align: middle;
  }

}
 ```

app/frontend/styles/application.scss
```diff
@@ -45,6 +45,7 @@ $moj-images-path: "~@ministryofjustice/frontend/moj/assets/images/";
 @import "_diagram";
 @import "_flex-grid";
 @import "_grid";
+@import "_primary-navigation";
```
AgaDufrat, line 48: Does it mean we can remove
https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/frontend/styles/provider/_all.scss#L6 ?

app/components/primary_navigation_component.html.erb
adamsilver: Eagle eye - yep the MOJ primary navigation component has 2 parts to it. The nav items aligned left, and the search box aligned right. We don’t use the search box aligned right.

So as I didn't want to cause bloat and bring in code we aren't using, I have just renamed all the classes to `app-` etc

app/frontend/styles/application.scss
adamsilver: Yep, I'll kill - thanks.

AgaDufrat (approved):
LGTM, you may want to squash those commits into one with `git reset --soft HEAD~3 &&
git commit` _(will prompt to write a new commit message for the squashed commit)_

cwrw (approved):

PR #3922, opened 2021-01-26
Move syncing courses to public V1 TTAPI

raamSoftwire:
## Context

The SyncCourses job currently uses the old Find API to pull in courses, but then it calls the new TTAPI to update them to add extra parameters.

This PR moves the syncing logic over to only use the TTAPI.

## Changes proposed in this pull request

- Add a `create_or_update_course` method to the SyncCourses job
- Add a job for getting sites and site statuses

## Guidance to review

It would be worth having the old Find API code open (`sync_courses_from_find.rb`). A lot of this code has been _borrowed_ from there.

We can no longer use an includes statement on courses to get the location data so that has been pulled out into a separate API call.

## Link to Trello card

https://trello.com/c/KcsgAOtE/2861-finish-fully-syncing-courses-from-the-new-public-teacher-training-api

## Things to check

- [x] This code does not rely on migrations in the same Pull Request
- [x] If this code includes a migration adding or changing columns, it also backfills existing records for consistency
- [x] API release notes have been updated if necessary
- [x] New environment variables have been [added to the Azure config](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/docs/environment-variables.md#azure-hosting-devops-pipeline)

PR #4084, opened 2021-02-15
Fix setup_local_dev_data task by syncing courses and sites synchronously

mnacos:
## Context

Our recent refactor of syncing courses (https://github.com/DFE-Digital/apply-for-teacher-training/pull/3922) broke the `setup_local_dev_data` task by not making any courses available when the system attempts to create test applications.

This is because:
- Providers added as part of `setup_local_dev_data` do not have `sync_courses` set to `true`
- `SyncCourses` and `SyncSites` run parallelised in the background by default, but no sidekiq is running when `setup_local_dev_data` is invoked locally

Add a `force_sync_courses:` parameter to `SyncProvider`.

Add a `run_in_background:` parameter to `SyncCourses` to control whether `SyncSites` are synchronous or not 

## Guidance to review

Drop your database and re-create it, `rake db:migrate`, then `rake setup_local_dev_data`. The setup should complete without errors.

## Link to Trello card

https://trello.com/c/NCkdrmel

## Things to check

### DIFF
--- a/app/services/teacher_training_public_api/sync_courses.rb
+++ b/app/services/teacher_training_public_api/sync_courses.rb
@@ -1,12 +1,13 @@
 module TeacherTrainingPublicAPI
   class SyncCourses
-    attr_reader :provider
+    attr_reader :provider, :run_in_background
 
     include Sidekiq::Worker
     sidekiq_options retry: 3, queue: :low_priority
 
-    def perform(provider_id, recruitment_cycle_year)
+    def perform(provider_id, recruitment_cycle_year, run_in_background: true)
       @provider = ::Provider.find(provider_id)
+      @run_in_background = run_in_background
 
       TeacherTrainingPublicAPI::Course.where(
         year: recruitment_cycle_year,
@@ -32,7 +33,11 @@ module TeacherTrainingPublicAPI
       add_accredited_provider(course, course_from_api[:accredited_body_code], recruitment_cycle_year)
       course.save!
 
-      TeacherTrainingPublicAPI::SyncSites.perform_async(provider.id, recruitment_cycle_year, course.id)
+      if run_in_background
+        TeacherTrainingPublicAPI::SyncSites.perform_async(provider.id, recruitment_cycle_year, course.id)
+      else
+        TeacherTrainingPublicAPI::SyncSites.new.perform(provider.id, recruitment_cycle_year, course.id)
+      end
     end
 
     def assign_course_attributes(course, course_from_api, recruitment_cycle_year)
--- a/app/services/teacher_training_public_api/sync_provider.rb
+++ b/app/services/teacher_training_public_api/sync_provider.rb
@@ -5,11 +5,18 @@ module TeacherTrainingPublicAPI
       @recruitment_cycle_year = recruitment_cycle_year
     end
 
-    def call(run_in_background: true)
-      provider = create_or_update_provider(
-        provider_attrs_from(@provider_from_api),
-      )
+    def call(run_in_background: true, force_sync_courses: false)
+      @force_sync_courses = force_sync_courses
 
+      provider_attrs = if existing_provider
+                         provider_attrs_from(@provider_from_api)
+                       else
+                         provider_attrs_from(@provider_from_api).merge(
+                           sync_courses: force_sync_courses,
+                         )
+                       end
+
+      provider = create_or_update_provider(provider_attrs)
       sync_courses(run_in_background, provider)
     end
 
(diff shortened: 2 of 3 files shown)


app/services/teacher_training_public_api/sync_sites.rb
```diff
@@ -53,7 +53,7 @@ def study_modes(course)
     end
 
     def create_course_options(site, study_mode, site_status)
-      course_option = CourseOption.find_or_initialize_by(
+      course_option = CourseOption.find_or_create_by(
```
mnacos, line 56: What do you think about this change?

app/services/teacher_training_public_api/sync_courses.rb
```diff
@@ -32,7 +33,11 @@ def create_or_update_course(course_from_api, recruitment_cycle_year)
       add_accredited_provider(course, course_from_api[:accredited_body_code], recruitment_cycle_year)
       course.save!
 
-      TeacherTrainingPublicAPI::SyncSites.perform_async(provider.id, recruitment_cycle_year, course.id)
+      if @run_in_background
```
cwrw, line 36: can you add `run_in_background` to `attr_reader`?

app/services/teacher_training_public_api/sync_sites.rb
cwrw: Since we are updating attributes on `CourseOption` below, I would suggest we keep this as `find_or_initialize_by` to avoid writing to the record twice

app/services/teacher_training_public_api/sync_provider.rb
```diff
@@ -5,11 +5,14 @@ def initialize(provider_from_api:, recruitment_cycle_year:)
       @recruitment_cycle_year = recruitment_cycle_year
     end
 
-    def call(run_in_background: true)
-      provider = create_or_update_provider(
-        provider_attrs_from(@provider_from_api),
-      )
-
+    def call(run_in_background: true, force_sync_courses: false)
```
cwrw, line 8: Im wondering about the need to force syncing courses at this stage, since this is already set here: https://github.com/DFE-Digital/apply-for-teacher-training/pull/4084/files#diff-f48eb2c3a6563c8b75c612393c0862a248b03c7276ae21832337476c1f688317R63
Let me know your thoughts

app/services/teacher_training_public_api/sync_sites.rb
mnacos: Could it be that valid course options are not created because of the `if course_option.vacancy_status != vacancy_status` conditional below? @raamSoftwire

app/services/teacher_training_public_api/sync_provider.rb
mnacos: It's needed here because `TeacherTrainingPublicApi::SyncProvider` uses the `sync_courses` value to decided whether to sync courses itself (via `TeachTrainingPublicApi::SyncCourses` and its own network of classes). Ordinarily, `sync_courses` is set per provider as we keep migrating training/ratifying providers to our system. But in this context, it needs to be forced.

`FindSync::SyncProviderFromFind` and its related classes, from what I understand, are on their way out and will be removed when the API transition is complete. Setting `sync_courses` here is too late for any of the `TeacherTrainingPublicApi` updates.

app/services/teacher_training_public_api/sync_sites.rb
raamSoftwire: Yes Michael, I think you're correct. That conditional logic was added as a review markup and I hadn't considered that scenario. So I think `find_or_create_by` is what we want here.

app/services/teacher_training_public_api/sync_courses.rb
mnacos: sure

app/services/teacher_training_public_api/sync_provider.rb
cwrw: Ah I missed that bit in this class🤦  It seems though we have a variable that isn't being set anywhere and might have been left behind: https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/services/teacher_training_public_api/sync_provider.rb#L29 maybe we can tidy up the logic there with the new flag being passed in i.e. either set it on initialising this class and removing the force flag, or just removing this left over variable?

raamSoftwire: Either of those options sounds like a good improvement

raamSoftwire (approved):
I've tested this locally and it works as expected. Other than the refactoring comments I think this is OK to be merged. Thanks Michael!

app/services/teacher_training_public_api/sync_provider.rb
mnacos: Yes, I'll remove `@sync_courses`, it's no longer needed.

I think think setting `sync_courses` to `true` for each provider is better in the context of local dev data, as these courses would only be there if `sync_courses` was set to `true` in the first place, therefore closer to reality.

mnacos: Then again, `sync_courses` needs to be set to something when creating a provider. I've merged the logic of the force flag and the instance variable so that the value of the force flag is used as `sync_courses:` when creating new providers but may also be used to import courses for existing providers with `sync_courses: false`, if needed.

PR #4086, opened 2021-02-15
Create new exports for longitudinal permissions analysis

steventux:

Adds two support interface exports:

- User permissions changes
- Organisational permissions changes

image

## Link to Trello card

I've attempted to make the necessary database joins to retrieve the information without multiple queries. Do they make sense in terms of the audit user belonging to multiple providers?

https://trello.com/c/mAgTgskS/3327-create-new-export-for-longitudinal-permissions-analysis

## Things to check

### DIFF
--- a/app/services/support_interface/organisation_permissions_export.rb
+++ b/app/services/support_interface/organisation_permissions_export.rb
@@ -13,7 +13,12 @@ module SupportInterface
         .pluck(
           'audits.created_at',
           'audits.user_id',
-          'audits.username',
+          Arel.sql(
+            "CASE audits.user_type
+              WHEN 'ProviderUser' THEN (SELECT CONCAT(first_name, ' ', last_name) FROM provider_users WHERE id = audits.user_id)
+              WHEN 'SupportUser' THEN (SELECT CONCAT(first_name, ' ', last_name) FROM support_users WHERE id = audits.user_id)
+            END".squish,
+          ),
           'ups.code',
           'ups.name',
           'tp.code',
--- a/app/services/support_interface/organisation_permissions_export.rb
+++ b/app/services/support_interface/organisation_permissions_export.rb
@@ -1,0 +1,78 @@
+module SupportInterface
+  class OrganisationPermissionsExport
+    def data_for_export
+      data = Audited::Audit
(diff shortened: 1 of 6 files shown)


app/services/support_interface/organisation_permissions_export.rb
```diff
@@ -0,0 +1,78 @@
...
+        .joins('INNER JOIN providers rp ON prp.ratifying_provider_id = rp.id')
+        .joins('LEFT JOIN provider_users_providers pup ON audits.user_id = pup.provider_user_id')
+        .joins('LEFT JOIN providers ups ON pup.provider_id = ups.id')
+        .where(auditable_type: 'ProviderRelationshipPermissions')
+        .where(action: %w[create update])
+        .where(permissions_clauses)
+        .pluck(
+          'audits.created_at',
+          'audits.user_id',
+          'audits.username',
```
AgaDufrat, line 16: I think we don’t set `username` on the Audits. All my local ones are `username: nil` so this column in the CSV is empty for me. We may need to use the `user_id` and `user_type` to figure that out.

app/services/support_interface/user_permissions_export.rb
```diff
@@ -0,0 +1,61 @@
+module SupportInterface
+  class UserPermissionsExport
```
AgaDufrat, line 2: As a side note as it's out of scope of this PR and I know it originated from the info on the Trello cards,
but I think a “nice to have” would be to unify the names of those provider permissions exports to make them a bit more descriptive. We now have:
- Active provider users
- Active provider user permissions
- Provider access controls
- User permissions
- Organisation permissions

I will add a card to the backlog.

app/services/support_interface/organisation_permissions_export.rb
steventux: I've attempted to fix this here https://github.com/DFE-Digital/apply-for-teacher-training/pull/4086/commits/4dbeb14f93693dfe0811959979afcaa7ab861df9 I've gone for the approach that doing as much db work as possible will help with performance, but this is undoubtedly resulting in a SQL-heavy export.

AgaDufrat: yea, that is unfortunate :/

AgaDufrat (approved):
LGTM

PR #4087, opened 2021-02-15
Interview cancellation flow bugfixes

benswannack:
## Context
Initial design review uncovered these small UI bugs

1. Add a GOVUK error summary to the form, above the H1 of the page

2. Add a validation on the length of the cancellation reason. It should be fewer than 10240 characters

3. Add visually hidden text to the cancellation reason change link on the check answers page. It should read Change cancellation reason

4. Add Cancel links to the cancel form and its check answers pages to abort the process of cancelling the interview and go back to the interviews tab (as per the prototype)

5. Add the "Error - " prefix to page/browser tab title when validation errors occur (as per some other forms)

## Guidance to review
Per commit

## Link to Trello card
https://trello.com/c/IKhgMo2w/3368-interview-cancellation-form-bugs

## Things to check

### DIFF
--- a/app/forms/provider_interface/cancel_interview_wizard.rb
+++ b/app/forms/provider_interface/cancel_interview_wizard.rb
@@ -4,7 +4,7 @@ module ProviderInterface
 
     attr_accessor :cancellation_reason
 
-    validates :cancellation_reason, presence: true
+    validates :cancellation_reason, presence: true, length: { maximum: 10240 }
 
     def initialize(state_store, attrs = {})
       @state_store = state_store
--- a/config/locales/interview.yml
+++ b/config/locales/interview.yml
@@ -13,7 +13,6 @@ en:
         page_title: Set up Interview
         title: Check and send interview details
         confirm: Send interview details
-        cancel: Cancel
         update:
           page_title: Change interview details
           title: Check and send new interview details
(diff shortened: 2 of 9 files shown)


config/locales/interview.yml
```diff
@@ -26,10 +26,12 @@ en:
         page_title: Cancel Interview
         title: Reason for cancelling interview
         success: Interview cancelled
+        cancel: Cancel
       review_cancel:
         page_title: Cancel interview
         title: Check and send interview cancellation
         cancellation_reason: Reason for cancelling interview
+        cancel: Cancel
```
despo, line 34: Can we move this in `en.yml` as there are cancel components across the entire app and this will also enble reuse. And update any related I18n calls `t('.cancel')` -> `t('cancel')`

despo (approved):
Looks good 👍🏼 
Left some minor feedback

app/forms/provider_interface/cancel_interview_wizard.rb
```diff
@@ -4,7 +4,7 @@ class CancelInterviewWizard
 
     attr_accessor :cancellation_reason
 
-    validates :cancellation_reason, presence: true
+    validates :cancellation_reason, presence: true, length: { maximum: 10240 }
```
despo, line 7: We seem to be using `word_count` to restrict content across the app. Shouldn't the same apply in this instance?

benswannack: It is specified in this doc as a character count rather than a word count: https://docs.google.com/document/d/1r7ANbveFFcgkRyiiyvCqlpfufNRz5UdHLL17n6FDx0E

I've tweaked the number to be consistent with other free text fields in the API, but I'm open to changing it if there's a reason to. Let's bring this up when we review the feature

PR #4089, opened 2021-02-16
Partition interviews into past and upcoming

benswannack:
## Context
Something we missed when originally implementing interviews

Split interviews on an application page into past and upcoming, also sorting by date and time

## Guidance to review
Hopefully simple!

## Link to Trello card
https://trello.com/c/eI5OACOF/3367-interviews-tab-group-upcoming-and-past-interviews

## Things to check

### DIFF
--- a/app/forms/provider_interface/interview_wizard.rb
+++ b/app/forms/provider_interface/interview_wizard.rb
@@ -102,7 +102,7 @@ module ProviderInterface
       wizard.send('date(3i)=', interview.date_and_time.day) if wizard.send('date(3i)').blank?
       wizard.location ||= interview.location
       wizard.provider_id ||= interview.provider_id
-      wizard.time ||= interview.date_and_time.strftime('%l:%M%P')
+      wizard.time ||= interview.date_and_time.strftime('%-l:%M%P')
 
       wizard
     end
--- a/app/views/provider_interface/interviews/index.html.erb
+++ b/app/views/provider_interface/interviews/index.html.erb
@@ -12,7 +12,7 @@
 
       <% if @upcoming_interviews.any? %>
         <h2 class="govuk-heading-m">
-          Upcoming interviews
+          <%= t('provider_interface.interviews.upcoming') %>
         </h2>
 
         <% @upcoming_interviews.each do |interview| %>
@@ -22,7 +22,7 @@
 
       <% if @past_interviews.any? %>
         <h2 class="govuk-heading-m">
-          Past interviews
+          <%= t('provider_interface.interviews.past') %>
         </h2>
 
         <% @past_interviews.each do |interview| %>
--- a/app/views/provider_interface/interviews/index.html.erb
+++ b/app/views/provider_interface/interviews/index.html.erb
@@ -2,22 +2,33 @@
 
 <%= render ProviderInterface::ApplicationChoiceHeaderComponent.new(application_choice: @application_choice, provider_can_respond: @provider_can_make_decisions) %>
 
-<% if @interviews.any? %>
-  <div class="govuk-grid-row">
-    <div class="govuk-grid-column-two-thirds">
-      <div class="app-interviews">
-        <% if @provider_can_make_decisions %>
-          <%= govuk_link_to 'Set up interview', new_provider_interface_application_choice_interview_path(@application_choice), class: 'govuk-button govuk-button--secondary' %>
-        <% end %>
 
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <div class="app-interviews">
+      <% if @provider_can_make_decisions %>
+        <%= govuk_link_to 'Set up interview', new_provider_interface_application_choice_interview_path(@application_choice), class: 'govuk-button govuk-button--secondary' %>
+      <% end %>
+
+      <% if @upcoming_interviews.any? %>
         <h2 class="govuk-heading-m">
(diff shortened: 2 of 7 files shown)


app/forms/provider_interface/interview_wizard.rb
```diff
@@ -102,7 +102,7 @@ def self.from_model(store, interview, step = 'input')
       wizard.send('date(3i)=', interview.date_and_time.day) if wizard.send('date(3i)').blank?
       wizard.location ||= interview.location
       wizard.provider_id ||= interview.provider_id
-      wizard.time ||= interview.date_and_time.strftime('%l:%M%P')
+      wizard.time ||= interview.date_and_time.strftime('%l:%M%P').strip
```
despo, line 105:
```suggestion
      wizard.time ||= interview.date_and_time.strftime('%-l:%M%P')
```

app/views/provider_interface/interviews/index.html.erb
```diff
@@ -2,22 +2,33 @@
...
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <div class="app-interviews">
+      <% if @provider_can_make_decisions %>
+        <%= govuk_link_to 'Set up interview', new_provider_interface_application_choice_interview_path(@application_choice), class: 'govuk-button govuk-button--secondary' %>
+      <% end %>
+
+      <% if @upcoming_interviews.any? %>
         <h2 class="govuk-heading-m">
           Upcoming interviews
```
despo, line 15: Can we move this in the interviews locale file.

app/views/provider_interface/interviews/index.html.erb
```diff
@@ -2,22 +2,33 @@
...
 
-        <% @interviews.each do |interview| %>
+        <% @upcoming_interviews.each do |interview| %>
+          <%= render ProviderInterface::InterviewAndCourseSummaryComponent.new(interview: interview, user_can_change_interview: @provider_can_make_decisions) %>
+        <% end %>
+      <% end %>
+
+      <% if @past_interviews.any? %>
+        <h2 class="govuk-heading-m">
+          Past interviews
```
despo: Can we move this in the interviews locale file.

spec/system/provider_interface/provider_manages_application_interviews_spec.rb
```diff
@@ -26,14 +26,18 @@
 
     when_i_visit_that_application_in_the_provider_interface
     and_i_click_set_up_an_interview
-    and_i_fill_out_the_interview_form(days_in_future: 1, time: '12pm')
+    and_i_fill_out_the_interview_form(days_in_future: 0, time: '12:30pm')
```
despo, line 29: I think it would be clearer to use a future day as it's not easy to understand at a glance that the created interview is actually in the future.

Consider extracting this line in a step with a more descriptive name.

spec/system/provider_interface/provider_manages_application_interviews_spec.rb
```diff
@@ -114,13 +118,28 @@ def and_an_interview_has_been_created(date)
...
   alias_method :and_another_interview_has_been_created, :and_an_interview_has_been_created
 
+  def when_the_first_interview_has_happened
+    Timecop.travel(30.minutes)
+    visit current_path
+  end
+
+  def then_i_see_the_upcoming_interview_under_the_correct_heading
+    expect(page).to have_css('.app-interviews > :nth-child(2)', text: 'Upcoming interviews')
+    expect(page).to have_css('.app-interviews > :nth-child(3)', text: '3 March 2020')
```
despo, line 128: Can we instead use the `data-qa` tag with a unique descriptive value, maybe `interview-#{interview.id}` to ensure that the spec can never fail as we've has a couple of issues with selectors that expect things to be in a certain order in order to work.

spec/system/provider_interface/provider_manages_application_interviews_spec.rb
```diff
@@ -114,13 +118,28 @@ def and_an_interview_has_been_created(date)
...
+  end
+
+  def and_i_see_the_past_interview_under_the_correct_heading
+    expect(page).to have_css('.app-interviews > :nth-child(4)', text: 'Past interviews')
+    expect(page).to have_css('.app-interviews > :nth-child(5)', text: '1 March 2020 at 12:30pm')
```
despo, line 133: same as above

despo (approved):
Looks good so happy to approve. I left some minor feedback that should be addressed before merging.

spec/system/provider_interface/provider_manages_application_interviews_spec.rb
benswannack: This spec is trying to check that the page is in the correct order. I want it to fail if we don't get the heading followed by the interview. Let me know if I've missed your point though!

despo: We had some flaky specs using this type of selector which were caused due to additional data being generated. Probably because our factories tend to be extended without taking into consideration existing usage. I guess we can keep an eye on it for now.

spec/system/provider_interface/provider_manages_application_interviews_spec.rb
benswannack: I've reverted it to how it was and put more logic in the "When the interview has happened" method.

PR #4090, opened 2021-02-16
Restructured Work History cleanup 

davidgisbey:
## Context

There are a few loose ends to clean up for the new Work History flow. There's quite a bit of old unused code to come out. There have now been 3 different versions of the Work History section. The first one (WorkHistoryBreaks) is completely unused and can come out.

We also need to redirect away from the existing flow if a candidate should see the new flow so they can't manually visit the endpoint and we need to pass a value into the details field on the new flow so we don't break the API.
 
1. Redirect away from the old flow if a candidate should be seeing the new flow
2. Remove the unused first implementation (work history breaks)
3. Pass a value into the details field when a candidate creates/updates a job (so we don't break the API)

## Guidance to review

The above 3 things are broken out into different commits. Per commit definitely recommended. 

TODO: give the whole flow a thorough manual test and try and break it.

## Link to Trello card

https://trello.com/c/6pH5Wy1U/3004-work-history-cleanup

## Things to check


app/queries/get_application_choices_for_providers.rb
```diff
@@ -4,7 +4,7 @@ class GetApplicationChoicesForProviders
     :offered_course_option,
     :provider,
     :site,
-    application_form: %i[candidate application_references application_work_experiences application_volunteering_experiences application_qualifications],
+    application_form: %i[candidate application_references application_work_experiences application_volunteering_experiences application_qualifications application_work_history_breaks],
```
malcolmbaig, line 7: Should this go in a different commit, with an explanation for why we're adding it?

app/forms/candidate_interface/restructured_work_history/job_form.rb
```diff
@@ -81,25 +82,10 @@ def save(application_form)
...
-          start_date_month,
-          start_date_year,
-          end_date_day,
-          end_date_month,
-          end_date_year,
-          currently_working,
-          relevant_skills,
-        ].all?(&:blank?)
-      end
-
```
malcolmbaig, line 102: Unrelated to the commit message, suggest putting this deletion into its own commit.

malcolmbaig (approved):
Couple of comments left inline, but otherwise LG 👍🏼

config/routes.rb
```diff
@@ -218,9 +218,6 @@
         get '/missing' => 'work_history/explanation#show', as: :work_history_explanation
         post '/missing' => 'work_history/explanation#submit'
 
-        get '/explain-breaks' => 'work_history/breaks#edit', as: :work_history_breaks
```
malcolmbaig, line 221: Spotted a bug while testing https://github.com/DFE-Digital/apply-for-teacher-training/pull/4090, but the comment probably fits better in this PR.

This line still uses the path helper deleted above. https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/components/candidate_interface/work_history_review_component.rb#L45

malcolmbaig (changes requested):
Small bug related to the work history breaks path helper

config/routes.rb
davidgisbey: Good spot. So this only surfaces when an application that used the first implementation is accessed via the review previous applications feature. So it still needs to exist in terms of being able to render/view, but will never need the change links. It would have broken the page though.

app/forms/candidate_interface/restructured_work_history/job_form.rb
davidgisbey: Yea good shout. Was being lazy!

app/queries/get_application_choices_for_providers.rb
davidgisbey: Have done 👍

malcolmbaig (approved):

PR #4093, opened 2021-02-16
Always show provider interface calls to action when applicable

mnacos:
## Context

... regardless of any flash messages present

One-liner to change the current weird behaviour.

## Guidance to review

Easiest review ever.

Thought about adding a component test for `show_inset_text?` but it's not particularly worth it, especially if we're going to be stubbing methods like `:awaiting_provider_decision?` etc.

## Link to Trello card

https://trello.com/c/dI0l2hyl

## Things to check

### DIFF
--- a/app/components/provider_interface/application_choice_header_component.rb
+++ b/app/components/provider_interface/application_choice_header_component.rb
@@ -90,10 +90,9 @@ module ProviderInterface
     end
 
     def show_inset_text?
-      flash.empty? && (
-        respond_to_application? || deferred_offer_wizard_applicable? ||
-        rejection_reason_required? || provider_cannot_respond?
-      ) || waiting_for_interview?
+      respond_to_application? || deferred_offer_wizard_applicable? ||
+        rejection_reason_required? || provider_cannot_respond? ||
+        waiting_for_interview?
     end
   end
 end


app/components/provider_interface/application_choice_header_component.rb
```diff
@@ -90,10 +90,9 @@ def provider_cannot_respond?
     end
 
     def show_inset_text?
-      flash.empty? && (
-        respond_to_application? || deferred_offer_wizard_applicable? ||
-        rejection_reason_required? || provider_cannot_respond?
-      ) || waiting_for_interview?
+      respond_to_application? || deferred_offer_wizard_applicable? ||
+        rejection_reason_required? || provider_cannot_respond? ||
+        waiting_for_interview?
```
despo, line 95: Can we juts remove `show_inset_text?` entirely?
Are there any other possible states?

mnacos: Not really, because it controls whether the `govuk-inset-text` surrounding structure is created. If you remove this check, you'll have extra whitespace and a nice short grey bar where the inset text would appear.

despo: So there are other states as well that this doesn't apply to.

despo (approved):
👍🏼

app/components/provider_interface/application_choice_header_component.rb
mnacos: Correct

PR #4094, opened 2021-02-16
Add interview emails to 'When emails are sent' doc

mnacos:
## Context

We've added these candidate emails and they are present as previews, but they don't show up in 'When emails are sent' doc in support.

### Before:

image
image

### After:

image
image

## Guidance to review

Is the doc correct?

## Link to Trello card

No Trello card

## Things to check


benswannack (approved):
Nice, thanks for doing this!

PR #4099, opened 2021-02-16
[Spike] Accept offers unconditionally

steventux:
## Context

We need to understand the technical complexities of making unconditional offers.

I found that you can legitimately make an offer via the API and not include conditions. No validation seems to prevent this.
So this spike adds a way for a candidate to accept an offer with no conditions, transitioning it from `offered` to `recruited` skipping any conditions related states and provider workflow.

## Guidance to review

To test:

1. Using a 'Received' application make an offer via curl, Postman with data like:
```
{
  "data": {
    "conditions": [],
    "course": null
  },
  "meta": {
    "attribution": {
      "full_name": "Jane Smith",
      "email": "jane.smith@example.com",
      "user_id": "12345"
    },
    "timestamp": "2019-03-01T15:33:49.216Z"
  }
}
```

2. Check the application (this can be done in the provider UI) now has an offer with no conditions.
3. Sign in to support and find the same application, login as the applying candidate
4. Go to the application offer tab and accept the offer

## TODO:

`AcceptUnconditionalOffer` Service spec
CandidateMailer method/template for unconditional offers.

## Link to Trello card

https://trello.com/c/uNuG9vGi/3397-tech-spike-supporting-unconditional-offers-via-the-api

## Things to check

### DIFF
--- a/app/services/accept_unconditional_offer.rb
+++ b/app/services/accept_unconditional_offer.rb
@@ -1,0 +1,20 @@
+class AcceptUnconditionalOffer
+  def initialize(application_choice:)
+    @application_choice = application_choice
+  end
+
+  def save!
+    ActiveRecord::Base.transaction do
+      ApplicationStateChange.new(@application_choice).accept!
+      @application_choice.update!(accepted_at: Time.zone.now)
+
+      ApplicationStateChange.new(@application_choice).confirm_conditions_met!
+      @application_choice.update!(recruited_at: Time.zone.now)
+    end
+
+    # TODO: New mailer method for unconditional offers.
+    CandidateMailer.offer_accepted(@application_choice).deliver_later
+
+    StateChangeNotifier.new(:recruited, @application_choice).application_outcome_notification
+  end
+end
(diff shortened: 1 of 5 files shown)


duncanjbrown (commented):
Nice and simple! I wonder if we want to start the `if` branch earlier and accept a bit of duplication (which we could then factor out) between conditional and unconditional acceptance? eg the emails probably won't be the same, we should probably have a dedicated state transition from offer -> recruited etc

steventux:
@duncanjbrown This draft PR should be up to date with your suggestions, it should be fairly simple to use this as a basis for enabling this feature.

app/services/accept_unconditional_offer.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  def save!
+    ActiveRecord::Base.transaction do
+      ApplicationStateChange.new(@application_choice).accept_unconditional_offer!
+      @application_choice.update!(accepted_at: Time.zone.now, recruited_at: Time.zone.now)
+    end
+
+    # TODO: Mailer method for unconditional offers.
+    CandidateMailer.offer_accepted(@application_choice).deliver_later
```
duncanjbrown, line 13: Should we have a provider email here too?

steventux:
Closing as we're implementing here https://github.com/DFE-Digital/apply-for-teacher-training/pull/4154

PR #4100, opened 2021-02-16
Add NotificationMessageComponent

despo:
## Context

This PR introduces a `NotificationMessageComponent` used to render flash messages and other information messages across the app.

**Notification box with message**
banner-with-message-only

**Notification box with link**
banner-with-link

**Notification box with secondary message**
banner-with-message-and-secondary-message

## Link to Trello card

https://trello.com/c/n5sL0Ysb/3278-update-notification-banners-to-the-gds-compliant-pattern

## Things to check


app/components/flash_message_component.html.erb
```diff
@@ -1,32 +1,4 @@
 <% if !flash.empty? %>
-  <% if flash[:warning] %>
-    <div class="app-banner app-banner--warning" aria-labelledby="warning-message">
```
cwrw, line 3: We've got a few banners which could potentially be moved into the new structure, example: 
https://github.com/DFE-Digital/apply-for-teacher-training/blob/0e14c10c27bb5a694e6eec1af92f9e26a9a14834/app/components/candidate_interface/ucas_downtime_component.html.erb
Are these good candidates/possible to move into the new `NotificationMessageComponent` components or are they different?

app/frontend/styles/application.scss
```diff
@@ -40,6 +40,7 @@ $moj-images-path: "~@ministryofjustice/frontend/moj/assets/images/";
 @import "_application-card";
 @import "_autocomplete";
 @import "_banner";
+@import "_notification_banner";
```
cwrw, line 43: Is the intention of this to replace the banner styles (or at least the similar ones) or are we intending of keeping both?

spec/components/notification_message_component_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      it 'renders the link' do
+        component_css = component.css('.govuk-notification-banner__content .govuk-body a')
+
+        expect(component_css.text.squish).to eql(link[:text])
+        expect(component_css[0].attr('href')).to eql(link[:url])
+      end
+    end
+
+    context 'when both an secondary_message and a link are present' do
```
cwrw, line 47: ✂️  an/a

app/components/notification_message_component.html.erb
```diff
@@ -0,0 +1,19 @@
+<div class="govuk-notification-banner govuk-notification-banner--<%= notification_type %>" aria-labelledby="govuk-notification-banner-title" data-module="govuk-notification-banner">
```
cwrw, line 1: can we also include the role here as well?
```
role="alert"
```

app/components/notification_message_component.html.erb
```diff
@@ -0,0 +1,19 @@
...
+      <%= title %>
+    </h2>
+  </div>
+  <div class="govuk-notification-banner__content">
+    <p class="govuk-notification-banner__heading">
+      <%= heading %>
+    </p>
+    <p class="govuk-body">
+      <% if message.present? %>
+        <%= message %>
```
cwrw, line 13: I was wondering how we would include a message which has a nested link embedded similar to https://design-system.service.gov.uk/components/notification-banner/success/index.html? (maybe we don't have that use case)

app/components/notification_message_component.html.erb
```diff
@@ -0,0 +1,19 @@
...
+    <h2 class="govuk-notification-banner__title" id="govuk-notification-banner-title">
+      <%= title %>
+    </h2>
+  </div>
+  <div class="govuk-notification-banner__content">
+    <p class="govuk-notification-banner__heading">
+      <%= heading %>
+    </p>
+    <p class="govuk-body">
+      <% if message.present? %>
```
cwrw, line 12: if we don't have a message/link we are rendering the following `<p class="govuk-body"></p>`, maybe we should move this check above the section to avoid rendering an empty section?

cwrw (commented):
This is great, and good step in moving towards the standardised styles proposed for notifications. I have a few questions and suggestions but otherwise 👍

paulrobertlloyd:
At some point I think we’re planning to move to use `govuk-components` gem, which has this component: https://github.com/DFE-Digital/govuk-components/blob/master/app/components/govuk_component/notification_banner.rb

Should this follow the same structure, or just ignore this for now until we move over?

app/components/flash_message_component.html.erb
```diff
@@ -1,32 +1,4 @@
...
-        </h2>
-        <% if flash[:link] %>
-          <p class="govuk-link">
-            <%= link_to flash[:link]['text'], flash[:link]['url'], class: 'govuk-link' %>
-          </p>
-        <% end %>
-      </div>
-    </div>
-  <% end %>
+  <% flash_key = flash.keys.detect { |key| %[warning info success].include?(key) } %>
```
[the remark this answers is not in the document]
despo: We already check that on line 1. As flash is only used in combination with `success`, `warning` or `info` in our views there should never be a case when one of those keys does not exist. The reason for the mapping here, is to make sure that `:link` is never used as the primary key.

However, I should probably move the key selection to the FlashMessageComponent

app/frontend/styles/application.scss
despo: We need to keep both as we currently use banners to render form details and the error summary (which the GDS guidelines specifically indicate that should not be replaced by this component). I did consider renaming it, however we already have `summary` components in the app, so I thought it's probably best to leave as it is to avoid confusion.

app/components/flash_message_component.html.erb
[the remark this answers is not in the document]
cwrw: My main worry was someone using a flash combination we don't support by accident, for instance `alert`, but if that is something unlikely then it's ok.

app/components/flash_message_component.html.erb
despo: I missed this one. I wonder why this is using a warning rather than an info box 🤔

app/frontend/styles/application.scss
despo: We still need the ordinary banner as we use it to display [form related information](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/views/provider_interface/interviews/new.html.erb#L18).

app/components/notification_message_component.html.erb
despo: We don't currently have that use case. An easy fix would potentially be to use `raw` or `html_safe` but I think we should not jump into this until required and then try and implement a more thought of solution.

despo:
@paulrobertlloyd I was not aware of this. However it also seems to allow for configuration that we don't require, is quite specific as to how it handles `success` but not info or warning, and also does not cater for all our needs, such as including links. I think we should just ignore for now and if required, revisit when we do begin using `govuk-components`

Also, should we add a card on the `Apply Tech` board  about `govuk-components`, so both the candidate and the provdendor teams are aware.

paulrobertlloyd:
> I think we should just ignore for now and if required, revisit when we do begin using `govuk-components`

Sounds good!

> Should we add a card on the Apply Tech board about `govuk-components`, so both the candidate and the provdendor teams are aware.

Guess so. This is a conversation that I’m only vaguely aware of – I know that we’ve started used this gem on Find as a way of dipping our toes in the water.

app/components/flash_message_component.html.erb
despo: Addressed. Also updated another three entries related to displaying information to the user when the DFE Sign in has problems/is down and removed the `warning`, `info` and `success` states from banner, since we no longer use them.

app/components/flash_message_component.html.erb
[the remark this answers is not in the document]
despo: Extracted that line to the component and update the component to only render when flash is set && notification type is `info`, `warning` or `success`, which matches what we had in place before. If we decide that we need the component to be more generic we can easily extend it.

app/components/notification_message_component.html.erb
despo: I completely missed this, thanks for pointing it out.

Added a bit of logic behind this so we render `role="alert"` or `role="region"` as mentioned on the GDS component page system. More specific information can be found on the relevant commit message.

cwrw (commented):
Looks great, thank you for addressing all the comments! Only thing (and I might be totally off about this and might just be a local issue) is when testing locally I can't get the warning flash message red, its coming back as blue (the default flash message colour). After a bit of digging I think its being overridden by the `node_modules/govuk-frontend/govuk/components/notification-banner/_index.scss`. Could you double check please? (ignore content for this one as just a test and should be a success banner)
Screenshot 2021-02-17 at 21 52 33
If all is well with that happy to approve otherwise

paulrobertlloyd:
Why are we including styles for `giving-notification-banner`? Are these not provided by the standard `gov-frontend` styles?

app/components/notification_message_component.rb
```diff
@@ -0,0 +1,40 @@
+class NotificationMessageComponent < ViewComponent::Base
+  include ViewHelper
+
+  def initialize(notification_type, message:, secondary_message: nil, message_link: nil)
```
richardpattinson, line 4: Should it be notification_type: here in order to be consistent with the other arguments?

richardpattinson (commented):
Looks good to me, Just one comment around arguments for the notification_message_component

app/components/notification_message_component.rb
despo: I think it's not necessary and it reads better this way.
I used the hash format for the other attributes just so it's clear what their purpose is.

despo:
@paulrobertlloyd it's required to setup the warning version of the banner, which does not seem to be available as part of the `govuk-notification-banner`. There is no mention of the card in moving away from displaying warning messages so if necessary, it's something we should address separately.

Also, to tweak the banner to be full width:

**Before**

before-full-width-change

**After**

after-full-width-change-2

despo:
@cwrw I had a mistake in the structure of my css. After I had a look at the warning version of the component I decided to nest the content inside `.govuk-notification-banner` and missed the fact that the warning class was on the same html tag. Addressed now.

paulrobertlloyd:
@despo I think I may have seen an earlier commit which had more styles in it?

As to warnings… I think ultimately we’ll want to refactor out these instances, as there are likely better way of dealing with the issues that cause warning messages to be shown (I’ve seen some cases, and they are really icky!) But for now this is a mighty fine solution.

Thanks for implementing this – you’ve made every BAT designer’s day 😃

app/frontend/styles/_notification_banner.scss
```diff
@@ -0,0 +1,25 @@
...
+  &.govuk-notification-banner--warning {
+    border-color: govuk-colour("red");
+    background-color: govuk-colour("red");
+    color: govuk-colour("red");
+
+    .govuk-notification-banner__link:link {
+      color: govuk-colour("red");
+    }
+
+    .govuk-notification-banner__link:hover, .govuk-notification-banner__link:visited {
```
paulrobertlloyd, line 16:
```suggestion
    .govuk-notification-banner__link:hover,
    .govuk-notification-banner__link:visited {
```

app/frontend/styles/_notification_banner.scss
```diff
@@ -0,0 +1,25 @@
+/* ==========================================================================
+   # NOTIFICATION BANNER
+   ========================================================================== */
+
+.govuk-notification-banner {
+
```
paulrobertlloyd, line 6:
```suggestion
```

app/views/provider_interface/application_choices/index.html.erb
```diff
@@ -3,16 +3,8 @@
...
-        <div class="app-banner__message">
-          <h2 class="govuk-heading-m" id="change-to-working-days-message">
-            20 December 2020 to 1 January 2021 do not count as working days
-          </h2>
-          <p class="govuk-body">
-            The Christmas period will not be included when calculating dates for applications to be automatically rejected. This is to match the dates used by UCAS.
-          </p>
-        </div>
-      </div>
+      <%= render NotificationMessageComponent.new(:info, message: '20 December 2020 to 1 January 2021 do not count as working days',
```
paulrobertlloyd, line 6: Just a note that ERB Lint trips up on indention like this, but fine to leave as for now.

paulrobertlloyd (approved):

despo:
@paulrobertlloyd I did indeed attempt to set the warning and success colors as well in a previous commit 👀  but addressed after having another look and realising that I didn't have to

app/views/provider_interface/application_choices/index.html.erb
despo: I'll take a look. There seem to be a couple of complains as well that I should address. Hoping to get some time so we can investigate how to make this run a bit faster or in parallel as you suggested on the PR.

cwrw (approved):
Looks great especially now we've swapped many of the old banners for the new notifications👏  Thanks for addressing all the feedback

PR #4102, opened 2021-02-17
Reinstate styles on buttons in the call to action to match the prototype

benswannack:
## Context
This button styling was removed in a refactor, but it's necessary in order to match the prototype

## Things to check


cwrw (approved):
oops

PR #4105, opened 2021-02-17
Expose API for TAD data exports

tijmenb:
## Context

TAD are currently downloading data exports manually. We'd like an API.

We need API keys for authentication. I wanted to avoid modifying the database for this, and reuse as much as possible. To do this I've added a static `DataAPIUser` model, which has hard coded records. This model is used to generate an AuthenticationToken, which we're currently using for magic links. This means we can generate multiple API keys for TAD.

In the future we can more easily add API authentication schemes. For example, I'm thinking that we might expose an "API" for data analysts to download other exports.

## Guidance to review

- What do you think of the auth?
- What do you think of the data export code?

## Things to check

### DIFF
--- a/app/controllers/data_api/tad_data_exports_controller.rb
+++ b/app/controllers/data_api/tad_data_exports_controller.rb
@@ -1,0 +1,38 @@
+module DataAPI
+  class TADDataExportsController < ActionController::API
+    include ActionController::HttpAuthentication::Token::ControllerMethods
+
+    before_action :verify_token!
+
+    def latest
+      data_export = DataAPI::TADExport.latest
+      data_export.update!(audit_comment: "File downloaded via API using token ID #{@authenticating_token.id}")
+      send_data data_export.data, filename: data_export.filename
+    end
+
+  private
+
+    def verify_token!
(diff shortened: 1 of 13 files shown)


app/controllers/data_api/tad_data_exports_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+          message: 'Please provide a valid API token',
+          status: :unauthorized,
+        )
+      end
+    end
+
+    def authorized?
+      authenticate_with_http_token do |token|
+        @authenticating_token = AuthenticationToken.find_by_hashed_token(user_type: 'DataAPIUser', raw_token: token)
+        @authenticating_token.user_id == 2
```
duncanjbrown, line 28: Shall we constantize this?

duncanjbrown (commented):
I guess we'll have to shell in and create the API tokens? Should we have an idempotent rake task/button to do that?

tijmenb:
@duncanjbrown updated with some clarifications and documentation updates.

duncanjbrown (approved):

PR #4106, opened 2021-02-17
Add code review guidelines

stevehook:
## Context

Based on discussions at the dev retro (16/2/2021) about better code reviews this PR adds a new guideline doc with ideas about how we can improve the code review process.

- Add initial code review guidelines

## Guidance to review

- Is there anything we need to do to import this into the new BAT dev guide?
- Should we have a checklist of more specific things that reviewers should look for (partly based on lessons learnt from bugs that have got through in the past)? (There's a lot more that could go into this guide).
- Does this overlap with any existing docs? (I think there was some advice in the old Apply dev team guide but couldn't find that any longer).

## Link to Trello card

n/a

## Things to check

### DIFF
--- a/docs/code-review-guidelines.md
+++ b/docs/code-review-guidelines.md
@@ -23,14 +23,19 @@ benefits like knowledge sharing.
     choices.
   - Step-by-step instructions on how to manually test the change, whether
     running locally or using a review app.
-- Consider extending test application data if it would make the PR easier to test.
+- We strongly recommend extending test application data if it would make the PR
+  easier to test. We rely on test data to provide a solid test platform so it
+  should aim to replicate the features of production data as possible.
 - If a PR is going to be complex to review, affects core functionality, touches
   on application security or is deemed to be risky for some other reason tag
   it as requiring two approvals rather than the normal one.
 - Follow general best practice for raising PR e.g. [How to raise a good pull request](https://www.annashipman.co.uk/jfdi/good-pull-requests.html).
   - Aim for each commit to be atomic, introducing a non-breaking change with all tests and linter passing. Consider arranging your commits into appropriate logical chunks with [git's history rewriting features](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History). This can make the PR easier to review and revert.
   - Try to avoid merge commits and use `git rebase master` instead.
+
 ### Reviewing PRs
+
+#### Process
 
 - Anybody can review a PR. Code reviews from a different team are encouraged as
   well as those from people more familiar with the code being changed. Everyone
@@ -42,3 +47,9 @@ benefits like knowledge sharing.
 - If as a reviewer you have any concerns about being the only reviewer you
   should tag the PR as requiring two approvals so that at least one other
   person checks it.
+
+#### Checklist
+
+These are some of the things that as a reviewer you may wish to check:
+
+- [] Are tests sufficient and robust? It is reasonable to ask for more tests.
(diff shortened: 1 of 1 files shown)


docs/code-review-guidelines.md
```diff
@@ -0,0 +1,42 @@
...
+  helpful. In particular consider adding:
+  - A list of particular questions that you would like feedback on, e.g. design
+    choices.
+  - Step-by-step instructions on how to manually test the change, whether
+    running locally or using a review app.
+- Consider extending test application data if it would make the PR easier to test.
+- If a PR is going to be complex to review, affects core functionality, touches
+  on application security or is deemed to be risky for some other reason tag
+  it as requiring two approvals rather than the normal one.
+
```
AgaDufrat, line 30: I was wondering if it would be useful to include guidance about commit history.
Some recent PRs on the project had a commit consisting of a single breaking test. This can make it difficult to  find when a breaking change was introduced with `git bisect`.  Also normally I don't think correcting linting  errors picked by rubocop deserves a separate commit if the errors were introduced in the same PR. 

```suggestion
- Follow general best practice for raising PR e.g. [How to raise a good pull request](https://www.annashipman.co.uk/jfdi/good-pull-requests.html).
  - Aim for each commit to be atomic, introducing a non-breaking change with all tests and linter passing. Consider arranging your commits into appropriate logical chunks, which can make the PR easier to review and revert, with [git's history rewriting features](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History).
  - Try to avoid merge commits and use `git rebase master` instead.
```

docs/code-review-guidelines.md
```diff
@@ -0,0 +1,42 @@
...
+- Create a draft PR if you want early feedback for work in progress.
+- We have a template for the PR descriptions. The _Guidance to
+  review_ section should contain any information that a reviewer might find
+  helpful. In particular consider adding:
+  - A list of particular questions that you would like feedback on, e.g. design
+    choices.
+  - Step-by-step instructions on how to manually test the change, whether
+    running locally or using a review app.
+- Consider extending test application data if it would make the PR easier to test.
```
AgaDufrat, line 26: I would probably make this point sound a bit stronger. 

Given we don't copy production data into other environments I think it's really important to update the factories and the GenerateTestApplications worker. 
This will not only make it easier to review the PR but also to product review in QA and staging. 
It may also speed up the development if we can assume that the data generated locally is a good representation of production data. I certainly waisted some time by not checking prod data early and discovering that the factories are not up to date 🙈

docs/code-review-guidelines.md
```diff
@@ -0,0 +1,69 @@
...
+  should tag the PR as requiring two approvals so that at least one other
+  person checks it.
+- Use positive language and be humble in review comments. Nits are OK but flag
+  them as such.
+
+#### Checklist
+
+These are some of the things that as a reviewer you may wish to check:
+
+- [] Are tests sufficient and robust? It is reasonable to ask for more tests.
```
duncanjbrown, line 57: Nit: these checkboxes don't render correctly for me on the GitHub markdown preview — do they need spaces?

duncanjbrown (approved):
I think this is a good start — later on we can add eg lists of things to check. There is some overlap with the PR guidance here, 
but I think this is a better place for the guidance as nobody looks at that doc and parts of it are outdated. https://docs.google.com/document/d/1eLj_KWInSmBmL3Tuep8ixAF28f-uh1JmrpBz4q1wbq0/edit#

PR #4107, opened 2021-02-17
Fix bugs on work history flow

davidgisbey:
## Context

As part of the work history cleanup card, I tried to break the new flow. This flushed out a few small bugs.

1. If you edit or add a work break then while the section is complete, then the section does not become incomplete
2. Errors are not linked on the full-time/part-time radios
3. The `I’m not sure the exact month or year I started` and `I’m not sure the exact month or year I left` checkboxes aren't retaining values if validations are hit
4.  If you don't add a start date on a work history break it blows up 
5. The explanation field on the choice page should have a 400 word limit and a 90% threshold (also the UI needs tweaking)

For the above:

1. Added  `current_application.update!(work_history_completed: false)` to the work history breaks controller edit/update/destroy actions
2. Link the errors on the commitment radio buttons
3. Cast the booleans on the jobs form before trying to save/update the job. This ensures the checkbox values are retained when validations are hit
4. add a presence validation to start date on work history breaks form 
5. Updated the explanation field on the choice page should have a 400 word limit and a 90% threshold and updated the UI

## Guidance to review

Per commit will be easiest. The errors are all easy to replicate, so it's probably just worth having a quick check locally. 

## Link to Trello card

https://trello.com/c/6pH5Wy1U/3004-work-history-cleanup

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/restructured_work_history/break_controller.rb
+++ b/app/controllers/candidate_interface/restructured_work_history/break_controller.rb
@@ -8,6 +8,8 @@ module CandidateInterface
       @work_break = RestructuredWorkHistory::WorkHistoryBreakForm.new(work_history_break_params)
 
       if @work_break.save(current_application)
+        current_application.update!(work_history_completed: false)
+
         redirect_to candidate_interface_restructured_work_history_review_path
       else
         track_validation_error(@work_break)
@@ -23,6 +25,8 @@ module CandidateInterface
       @work_break = RestructuredWorkHistory::WorkHistoryBreakForm.new(work_history_break_params)
 
       if @work_break.update(current_work_history_break)
+        current_application.update!(work_history_completed: false)
+
         redirect_to candidate_interface_restructured_work_history_review_path
       else
         track_validation_error(@work_break)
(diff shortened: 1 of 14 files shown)


app/controllers/candidate_interface/restructured_work_history/break_controller.rb
```diff
@@ -8,6 +8,8 @@ def create
       @work_break = RestructuredWorkHistory::WorkHistoryBreakForm.new(work_history_break_params)
 
       if @work_break.save(current_application)
+        current_application.update!(work_history_completed: false)
```
stevehook, line 11: These work for me. Should they be covered in the system spec? (optional - not sure myself)

stevehook (commented):
Nice tidy up!. It all worked for me apart from point 3. When I checked either of those boxes the values were lost when the form was displayed.

A couple of minor inline comments.

app/controllers/candidate_interface/restructured_work_history/break_controller.rb
davidgisbey: Yea I'm never really sure about this one. I'll add it though.

davidgisbey:
`Nice tidy up!. It all worked for me apart from point 3`

Yea i forgot to cast the booleans before re-rendering the page on the create action 🤦 

Good spot!

stevehook (approved):
:+1:

PR #4109, opened 2021-02-17
Change offered course option column type to BIGINT

cwrw:
## Context

Some columns referring to the same tables are defined interchangeably as `integer` and `BIGINT`, this is due to us referencing the same table under different ids.

Course option ids are defined as `BIGINT` in our schema, make sure the offered course option id is also of type `BIGINT` as well, as it also refers to the same `course_options` table.

## Guidance to review

Making sure that running the migration and rolling back works.

## Things to check


cwrw:
I just realised we have multiple inconsistent IDs past this which need fixing, I'll raise a separate PR for those

PR #4111, opened 2021-02-17
Fix validation for SupportUser#email_address uniqueness

stevehook:
## Context

Downcasing of email address in the model should happen before validation. Otherwise a mixed case email address could still pass validation, then be downcased and violate the unique index when it hits the database.

Just changes the `SupportUser#downcase_email_address` callback to be to be called `before_validation` rather than `before_save` and add a test to verify.

## Guidance to review

- Does this make sense?

## Link to Trello card

https://trello.com/c/3gKAM4de/3054-add-new-support-users-with-duplicate-email-causes-crash

## Things to check


davidgisbey (approved):
Makes sense to me 👍

PR #4119, opened 2021-02-18
Use EFL qualification description string only in the english_language_qualifications API field

benswannack:
## Context
Tribal need to get data about a candidate's EFL qualifications if they are provided

## Guidance to review
Is this worthy of a release note? Or would that be confusing

Do not expose the deprecated `english_language_details` field over the API if it is present

## Link to Trello card
https://trello.com/c/nb0tCACC/3401-only-include-esl-qualification-data-in-the-api-when-it-is-provided-as-a-qualification-by-the-candidate
## Things to check

### DIFF
--- a/app/models/english_proficiency.rb
+++ b/app/models/english_proficiency.rb
@@ -1,4 +1,6 @@
 class EnglishProficiency < ApplicationRecord
+  include PublishedInAPI
+
   audited associated_with: :application_form
 
   belongs_to :application_form
--- a/app/models/ielts_qualification.rb
+++ b/app/models/ielts_qualification.rb
@@ -21,7 +21,7 @@ class IeltsQualification < ApplicationRecord
     9.0
   ].freeze
 
-  has_one :english_proficiency, as: :efl_qualification
+  has_one :english_proficiency, as: :efl_qualification, touch: true
 
   def name
     'IELTS'
--- a/config/vendor-api-v1.yml
+++ b/config/vendor-api-v1.yml
@@ -601,10 +601,8 @@ components:
           type: string
           maxLength: 10240
           nullable: true
-          description: About this candidate’s English language qualifications, if
-            English is not their main language
-          example: I have a degree in English Language and Literature from the University
-            of Munich
+          description: Details of this candidate's English language qualification, if English is not their main language
+          example: 'Name: TOEFL, Grade: 20, Awarded: 1999'
         other_languages:
           type: string
           maxLength: 10240
--- a/app/controllers/vendor_api/applications_controller.rb
+++ b/app/controllers/vendor_api/applications_controller.rb
@@ -9,16 +9,28 @@ module VendorAPI
     end
 
     def show
-      application_choice = GetApplicationChoicesForProviders.call(providers: current_provider, vendor_api: true)
-        .find(params[:application_id])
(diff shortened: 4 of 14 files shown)


duncanjbrown (commented):
Do we need to hook up the IELTS and TOEFL qualifications to `PublishedInAPI`?

app/models/ielts_qualification.rb
```diff
@@ -21,7 +21,7 @@ class IeltsQualification < ApplicationRecord
     9.0
   ].freeze
 
-  has_one :english_proficiency, as: :efl_qualification
+  has_one :english_proficiency, as: :efl_qualification, touch: true
```
benswannack, line 24: Not sure that this actually works, adding a spec for it now

config/vendor-api-v1.yml
```diff
@@ -601,10 +601,8 @@ components:
           type: string
           maxLength: 10240
           nullable: true
-          description: About this candidate’s English language qualifications, if
-            English is not their main language
-          example: I have a degree in English Language and Literature from the University
-            of Munich
+          description: Details of this candidate's English language qualification, if English is not their main language
+          example: 'Name: TOEFL, Grade: 20, Awarded: 1999'
```
cwrw, line 605: I was wondering if this constitutes as a breaking change, as in we were initially returning free text, now we are still returning a string but it is actually structured data in the text 🤔 ?

benswannack: I did look into what this would change in practice. It previously relied on the english_qualification_details free text, and fell back on this new structured text. That means the only changes will be:
- Applications with the old free text (and no EFL qualification added) will no longer return a value here. 
- Things with the old free text and the new will now start returning the new structured EFL qualification text.

I checked and there are only 5 applications with both the old and new fields set (the latter case above), and they are not HEI applications so won't have been ingested on the API ever before.

Now I think about it, maybe we should switch around the fallback rather than turning it off completely, so that things with the old text and no EFL qualification set will still return something. The reason I didn't do this before it that Tribal are interested in the EFL qualification in particular, but we should probably favour backwards compatibility.

What do you think?

app/models/english_proficiency.rb
```diff
@@ -1,4 +1,6 @@
 class EnglishProficiency < ApplicationRecord
+  include PublishedInAPI
```
benswannack, line 2: Included this here (rather than in the actual EFL models) for multiple reasons:
- This cannot be included in the EFL models, since it requires the application form to be always available. However, you can have a standalone EFL qualification object (with no associated application_form when it's created). To make this work, then we'd have to enforce weird rules about the order of creation of `english_proficiency` and `efl_qualifications`.
- This model also affects the API response for an application choice. It decides on the value of `english_main_language`

This works as intended because the touch from the EFL qualifications triggers the after_commit on the english_proficiency and hence the touch on the application choices

config/vendor-api-v1.yml
benswannack: I've now gone with the slightly more backwards compatible change, so that we aren't left with a bunch of old applications returning `nil` for the field where they didn't before.

I now don't think this is a breaking change because we have just added specificity to a field. That is, where there used to be data that could be in any format, we now return something more predictable.

cwrw (approved):
LGTM

PR #4120, opened 2021-02-18
Add data export for candidate email counts

malcolmbaig:
## Context

> As a... data analyst
I need to... have a list of all email notifications and how frequently they are sent that we currently send to candidates
In order to... get a better understanding of the 'as is' landscape of email notifications and make an informed decision on the scale of work required to develop data driven email notification tracking
So that... we can understand how effective email notifications are in prompting candidate to take action

A data export that provides:

- A list of all email types sent by the service
- The total number of each type that have been logged by the
  EmailLogInterceptor
- The most recent send date of each type

> Screenshot_20210218_153635

> email-counts-csv

## Guidance to review

- We probably have legacy templates that we don't use anymore in production. Rather than provide a hardcoded list of email templates that this export should pick up, for simplicity I've opted to just include the most recent send date of each email type. Anything older than *n* months can be filtered out.
- Does this export meet the goals of the card?

## Link to Trello card
https://trello.com/c/K7FdnVOg

## Things to check

### DIFF
--- a/app/models/data_export.rb
+++ b/app/models/data_export.rb
@@ -25,6 +25,11 @@ class DataExport < ApplicationRecord
       description: 'This provides the compiled results of all feedback received from prompts throughout the application form.',
       class: SupportInterface::CandidateApplicationFeedbackExport,
     },
+    candidate_email_send_counts: {
+      name: 'Candidate email send counts',
+      description: "A list of all emails sent by the service, and how many we've sent to date",
+      class: SupportInterface::CandidateEmailSendCountsExport,
+    },
     candidate_feedback: {
       name: 'Candidate feedback',
       description: 'This provides the compiled results of all the single-page candidate feedback forms.',
--- a/app/models/data_export.rb
+++ b/app/models/data_export.rb
@@ -1,29 +1,24 @@
 class DataExport < ApplicationRecord
   EXPORT_TYPES = {
+    active_provider_user_permissions: {
+      name: 'Active provider user permissions',
+      description: 'The list of provider users with the permissions they have for each of their organisations.',
+      class: SupportInterface::ActiveProviderUserPermissionsExport,
+    },
+    active_provider_users: {
+      name: 'Active provider users',
+      description: 'The list of provider users that have signed in to apply at least once.',
+      class: SupportInterface::ActiveProviderUsersExport,
+    },
+    application_references: {
+      name: 'Application references',
+      description: 'A list of all application references which have been selected by candidates to date.',
+      class: SupportInterface::ApplicationReferencesExport,
+    },
     application_timings: {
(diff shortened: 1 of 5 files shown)


app/models/data_export.rb
```diff
@@ -1,109 +1,114 @@
...
+    },
+    tad_provider_performance: {
+      name: 'TAD provider performance',
+      description: 'A list of all application/offered/accepted counts for all courses in Apply.',
+      class: SupportInterface::TADProviderStatsExport,
+    },
+    unexplained_breaks_in_work_history: {
+      name: 'Unexplained breaks in work history',
+      description: 'A list of candidates with unexplained breaks in their work history.',
+      class: SupportInterface::UnexplainedBreaksInWorkHistoryExport,
```
stevehook, line 111: 👍

app/services/support_interface/candidate_email_send_counts_export.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  private
+
+    def emails_with_counts
+      Email
+        .select(
+          'mail_template, COUNT(mail_template) AS send_count, MAX(created_at) AS last_sent_at',
+        )
+        .group('mail_template')
+        .order('mail_template ASC')
```
stevehook, line 21: Makes sense, I guess the only thing this won't reveal is email templates that have never been sent (or rather not sent since we started tracking) but I don't think that matters.

app/services/support_interface/candidate_email_send_counts_export.rb
```diff
@@ -0,0 +1,24 @@
+module SupportInterface
+  class CandidateEmailSendCountsExport
+    def data_for_export
+      emails_with_counts.map do |email_hash|
+        {
+          'Email' => email_hash['mail_template'],
+          'Send count' => email_hash['send_count'],
+          'Last sent at' => email_hash['last_sent_at'],
```
stevehook, line 8: Do we specify the time format?

stevehook (approved):
Looks really good and review app works. Couple of non-blocking queries inline.

app/services/support_interface/candidate_email_send_counts_export.rb
malcolmbaig: That's correct, we've only been logging emails in this table since 21 February 2020.

app/services/support_interface/candidate_email_send_counts_export.rb
malcolmbaig: Just using the UTC timestamp straight from the db at the moment.

malcolmbaig:
Working on a followup PR that adds how many unique candidates have been sent each email type.

PR #4121, opened 2021-02-18
Add test that we correctly document the status for TAD

tijmenb:
## Context

We've recently introduced new states, which breaks the analysis by TAD.

Having this test will make sure that if we add another status, we also are required to update the documentation for the export. In turn, this should prompt us to warn TAD of the changes.

## Things to check

### DIFF
--- a/app/services/application_state_change.rb
+++ b/app/services/application_state_change.rb
@@ -111,6 +111,12 @@ class ApplicationStateChange
     STATES_VISIBLE_TO_PROVIDER - [:interviewing]
   end
 
+  # TAD receive the same applications as providers, but the `rejected` and `declined`
+  # states are more specific when they are rejected or declined by default.
+  def self.states_visible_to_tad
+    states_visible_to_provider + %i[rejected_by_default declined_by_default]
+  end
+
   def self.states_visible_to_provider_without_deferred
     states_visible_to_provider - [:offer_deferred]
   end
--- a/spec/services/data_api/tad_export_spec.rb
+++ b/spec/services/data_api/tad_export_spec.rb
@@ -17,4 +17,13 @@ RSpec.describe DataAPI::TADExport do
       expect(result.map { |r| r[:status] }).to match_array(%w[rejected_by_default declined_by_default rejected declined])
     end
   end
+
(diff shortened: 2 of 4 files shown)


spec/services/data_api/tad_export_spec.rb
```diff
@@ -17,4 +17,13 @@
       expect(result.map { |r| r[:status] }).to match_array(%w[rejected_by_default declined_by_default rejected declined])
     end
   end
+
+  describe '#status' do
+    it 'only includes states that are documented' do
+      documented_columns = DataSetDocumentation.for(described_class)['status']['enum']
+      possible_states_in_export = ApplicationStateChange.states_visible_to_provider.map(&:to_s) + %w[rejected_by_default declined_by_default]
```
duncanjbrown, line 24: Should we define a new set on `ApplicationStateChange`?

tijmenb: Hmm, not sure - we aren't using it anywhere else I think?

tijmenb: Done now!

app/services/application_state_change.rb
```diff
@@ -111,6 +111,12 @@ def self.states_visible_to_provider
     STATES_VISIBLE_TO_PROVIDER - [:interviewing]
   end
 
+  # TAD receive the same applications as providers, but the `rejected` and `declined`
+  # states are more specific when they are rejected or declined by default.
+  def self.states_visible_to_tad
```
duncanjbrown, line 116: Should this be part of the scope in `DataAPI::TADExport`?

tijmenb: No, that would be a bit weird I think. The states here aren't all real states, rejected_by_default and declined_by_default are kind of virtual, I wouldn't want them in a query perhaps.

tijmenb:
Superseded by https://github.com/DFE-Digital/apply-for-teacher-training/pull/4167.

app/services/application_state_change.rb
duncanjbrown: Got it

PR #4124, opened 2021-02-18
Refactor tile component

paulrobertlloyd:
## Context

To support linking cards in the performance dashboard, the `AnchorLinkComponent` was created. However, it may be more elegant to allow the `TileComponent` to accept an `href` param, which if present, links the card.

And, with the possibility of a card having a link, we can then style linked cards properly, giving them hover and focused states, too.

* Add `href` param to `SupportInterface::TileComponent`
* Add `app-card__link` styles
* Update performance dashboard page to use new `param` on this component value instead of using separate `SupportInterface::AnchorLinkComponent`
* Remove `SupportInterface::AnchorLinkComponent`

## Guidance to review

Preview: https://apply-for-te-refactor-t-auuc6g.herokuapp.com/rails/view_components/support_interface/tile_component/linked_tile

## Things to check


app/components/support_interface/tile_component.html.erb
```diff
@@ -1,4 +1,6 @@
-<div class="app-card<%= colour == :default ? '' : " app-card--#{colour}" %>">
-  <span class="<%= count_class %>"><%= number_with_delimiter(count) %></span>
-  <%= label %>
+<div class="<%= card_classes %>">
+  <%= href ? "<a class=\"app-card__link\" href=\"#{href}\">".html_safe : '' %>
+    <span class="<%= count_class %>"><%= number_with_delimiter(count) %></span>
+    <%= label %>
+  <%= href ? "</a>".html_safe : '' %>
```
benswannack, line 5: Either something like:
```
<% if href.present? %>
    <a class="app-card__link" href="<%= href %>">
      <span class="<%= count_class %>"><%= number_with_delimiter(count) %></span>
      <%= label %>
    </a>
  <% else %>
    <span class="<%= count_class %>"><%= number_with_delimiter(count) %></span>
    <%= label %>
  <% end %>
```
which is explicit and readable.

Or you could go with a method like:
```
def wrap_in_link_if_href_present(&block)
      if href.blank?
        concat capture(&block)
      else
        concat content_tag(:a, capture(&block), class: 'app-card__link', href: href)
      end
    end
```
and then
```
 <% wrap_in_link_if_href_present do %>
    <span class="<%= count_class %>"><%= number_with_delimiter(count) %></span>
    <%= label %>
  <% end %>
```
(very much open to better method name suggestions)

spec/components/support_interface/ucas_matches_statistics_component_spec.rb
```diff
@@ -42,48 +42,48 @@
   it 'renders number of candidates with dual application or multiple acceptances' do
     result = render_inline(described_class.new(ucas_matches))
 
-    expect(result.css('.app-card--blue')[0].text.strip).to include("7\n  candidates with dual application or multiple acceptances")
+    expect(result.css('.app-card--blue')[0].text.strip).to include("7\n    candidates with dual application or multiple acceptances")
```
despo, line 45:
```suggestion
    expect(result.css('.app-card--blue')[0].text.squish).to include("7  candidates with dual application or multiple acceptances")
```
Using `squish` instead of `trim` enables us to have cleaner specs as what we are testing in this instance is the number of candidates and the correct text not not how the content is rendered.

app/components/support_interface/tile_component.html.erb
paulrobertlloyd: Thanks @benswannack… your first solution I think is the most straight–forward. Updated!

app/components/support_interface/tile_component.rb
```diff
@@ -1,12 +1,17 @@
...
+    def initialize(count:, label:, colour: :default, size: :regular, href: nil)
       @count = count
       @label = label
       @colour = colour
       @size = size
+      @href = href
+    end
+
+    def card_classes
+      @colour == :default ? 'app-card' : "app-card app-card--#{colour}"
```
despo, line 14:
```suggestion
      colour == :default ? 'app-card' : "app-card app-card--#{colour}"
```

app/components/support_interface/tile_component.rb
```diff
@@ -1,12 +1,17 @@
...
-    attr_reader :count, :label, :colour
+    attr_reader :count, :label, :colour, :href
 
-    def initialize(count:, label:, colour: :default, size: :regular)
+    def initialize(count:, label:, colour: :default, size: :regular, href: nil)
       @count = count
       @label = label
       @colour = colour
       @size = size
+      @href = href
```
despo, line 10: Consider changing this to `anchor` so it's more indicative as to how it's used.

app/components/support_interface/tile_component.html.erb
```diff
@@ -1,4 +1,11 @@
-<div class="app-card<%= colour == :default ? '' : " app-card--#{colour}" %>">
-  <span class="<%= count_class %>"><%= number_with_delimiter(count) %></span>
-  <%= label %>
+<div class="<%= card_classes %>">
+  <% if href.present? %>
+    <a class="app-card__link" href="<%= href %>">
```
despo, line 3:
```suggestion
    <%= link_to href, class: 'app-card__link' do %>
```

app/components/support_interface/tile_component.html.erb
```diff
@@ -1,4 +1,11 @@
...
+      <span class="<%= count_class %>"><%= number_with_delimiter(count) %></span>
+      <%= label %>
+    </a>
```
despo, line 6:
```suggestion
    <% end %>
```

despo:
@paulrobertlloyd can you also extend https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/spec/components/previews/support_interface/tile_component_preview.rb to include an example of a component using a link please.

app/components/support_interface/tile_component.rb
paulrobertlloyd: Hmmm, it’s used as an anchor on the performance dashboard, but can take any given URL, so maybe `href` is clearer? Or is `anchor` a Rails convention?

despo: No it's not. I simply did not see any cases where it's used to link to anything but anchors so it feels like premature optimisation. Up to you.

app/components/support_interface/tile_component.html.erb
paulrobertlloyd: Ah, of course, Much neater.

app/components/support_interface/tile_component.rb
paulrobertlloyd: I’ll leave it for now… I think it’s a bit clearer, but that’s only with my HTML hat on!

paulrobertlloyd:
@despo Added a test for a linked card in 85fcd0de0c60b03fa6e340d391fc5be8696dbc39. Is this correct?

spec/components/support_interface/tile_component_spec.rb
```diff
@@ -19,6 +19,15 @@
     end
   end
 
+  describe 'a tile with a href' do
+    subject { described_class.new(count: 3, label: 'blind mice', href: '#cheese') }
+
+    it 'adds an anchor element with that href' do
```
despo, line 25: To keep it consistent, this should be referring to a link and not an anchor.

despo:
@paulrobertlloyd have you seen the component previews? You can access them locally and makes reviewing a lot easier as we can see how the component behaves on the heroku review app. e.g.  https://apply-for-te-refactor-t-auuc6g.herokuapp.com/rails/view_components/support_interface/tile_component

paulrobertlloyd:
> @paulrobertlloyd have you seen the component previews?

I had, but never thought to investigate how they are generated. I found out! So here’s a preview of a linked tile: https://apply-for-te-refactor-t-auuc6g.herokuapp.com/rails/view_components/support_interface/tile_component/linked_tile

despo (approved):
@paulrobertlloyd great work 🚀  :shipit:

PR #4135, opened 2021-02-22
Fix focus states for primary navigation, sub navigation and filter links

adamsilver:
## Context

The focus styles are either not there or not very good.

Add focus states copying the fixes from https://github.com/DFE-Digital/register-trainee-teachers/pull/546

## Link to Trello card

https://trello.com/c/SS0ZVmKm/3413-bug-on-primary-navigation-link-focus

### DIFF
--- a/app/frontend/styles/_primary-navigation.scss
+++ b/app/frontend/styles/_primary-navigation.scss
@@ -60,7 +60,20 @@
     color: govuk-colour("black"); // Focus colour on yellow should really be black.
     position: relative; // Ensure focus sits above everything else.
     z-index: 1;
-    box-shadow: none;
+    outline: 3px solid transparent;
+    background-color: govuk-colour("yellow");
+    text-decoration: none;
+  }
+
+  &:focus:before {
+    background-color: govuk-colour("black");
+    content: "";
+    display: block;
+    height: 5px;
+    position: absolute;
+    bottom: 0;
+    left: 0;
+    width: 100%;
   }
 
   &:hover {
@@ -88,6 +101,7 @@
 
       &:before {
         background-color: govuk-colour("black");
+        height: 7px;
       }
 
     }
--- a/app/components/filter_component.html.erb
+++ b/app/components/filter_component.html.erb
@@ -9,7 +9,7 @@
         </div>
       </div>
 
-      <div class="moj-filter__content">
+      <div class="moj-filter__content" tabindex="-1">
         <% if active_filters.any? %>
           <div class="moj-filter__selected">
             <div class="moj-filter__selected-heading">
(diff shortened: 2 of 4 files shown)


edwardhorsford:
Will you also add the extra styles for the selected filters?

app/frontend/styles/application.scss
```diff
@@ -9,6 +9,8 @@
 $govuk-font-url-function: frontend-font-url;
 $govuk-image-url-function: frontend-image-url;
 
+$govuk-global-styles: true;
```
despo, line 12: What does setting this property do?

app/frontend/styles/_primary-navigation.scss
```diff
@@ -88,6 +99,7 @@
 
       &:before {
         background-color: govuk-colour("black");
+        height: 7px;
```
despo, line 102: Shouldn't we be using `rem` instead of pixels so font size is scaled correctly across browsers?

despo (approved):
Verified fix on review app using Chrome. I have some questions so I can understand why we are doing things a certain way and maybe consider addressing in future PRs, but happy to approve as we can merge this fix.

app/frontend/styles/application.scss
adamsilver: It turns on global styles for P tags and A tags. This means you don't have to, for example add `govuk-body` and ‘govuk-link’ respectively onto every link to get the right styles (including focus styles).

It also makes it easy to copy styles from the prototype into production (which we sometimes do) which works the same way.

app/frontend/styles/_primary-navigation.scss
adamsilver: I think it's fine - we use pixels in many places that do not impact the font size such as margin, padding and borders.

app/frontend/styles/application.scss
edwardhorsford: I might be wary of turning that on. It does catch some things, but in general I've been trying to always use the right classes.

paulrobertlloyd: I suggested turning this on once too… there aspects of the design system that really bug me, and while I totally understand the reasoning, having to add a class to every element to get standard styling is up there on my list of pet peeves!

paulrobertlloyd: In fact, turning on these styles breaks the design of the tile component… link styles and focus state are broken.

Screenshot 2021-02-22 at 10 34 52

frankieroberto: I'm wary of turning it on. The global styles mode has had some bugs in the past (as I don't think it’s routinely tested by the Design System team), and was apparently mainly done to make prototyping easier.

Maybe we can find other ways (like the rubocop rule) to catch elements which don't use the helpers that add the classes?

(I also dislike having to add a class to every element though. Especially as there are some which don't require it, like `<li>`...)

app/frontend/styles/_primary-navigation.scss
edwardhorsford: The design system uses `px` for most units - I know for a good reason, but I don't remember them any more. I think all modern browsers will scale them fine.

app/frontend/styles/application.scss
adamsilver: I have removed.

paulrobertlloyd (approved):

despo (approved):
Thanks for addressing all feedback and also sorting out your commit log!

PR #4136, opened 2021-02-22
Show interviews that happened today in upcoming interviews sections

benswannack:
## Context

This means that a provider can still see interviews that have just happened in the upcoming interviews sections

Add scopes to select interviews in the past, or interviews that are upcoming

## Guidance to review
Does it make sense?

## Link to Trello card
https://trello.com/c/CGPoJnQM/3418-interviews-should-only-be-marked-as-in-the-past-the-next-day

## Things to check

### DIFF
--- a/app/components/provider_interface/application_choice_interviews_list_component.html.erb
+++ b/app/components/provider_interface/application_choice_interviews_list_component.html.erb
@@ -1,0 +1,25 @@
+<div class="app-interviews">
+  <% if user_can_create_or_change_interviews %>
+    <%= govuk_button_link_to 'Set up interview', new_provider_interface_application_choice_interview_path(application_choice), class: 'govuk-button--secondary' %>
+  <% end %>
+
+  <% if upcoming_interviews.any? %>
+    <h2 class="govuk-heading-m">
+      <%= t('provider_interface.interviews.upcoming') %>
+    </h2>
+
+    <% upcoming_interviews.each do |interview| %>
+      <%= render ProviderInterface::InterviewAndCourseSummaryComponent.new(interview: interview, user_can_change_interview: user_can_create_or_change_interviews) %>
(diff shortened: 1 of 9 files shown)


spec/system/provider_interface/provider_manages_application_interviews_spec.rb
```diff
@@ -156,12 +156,24 @@ def and_an_interview_has_been_created(date)
...
+    expect(page).to have_css('.app-interviews > :nth-child(3)', text: '2 March 2020')
+    expect(page).to have_css('.app-interviews > :nth-child(4)', text: '3 March 2020')
+  end
+
+  def when_the_first_interview_happened_yesterday
+    Timecop.travel(2020, 3, 3, 9, 0, 0)
+    provider_signs_in_using_dfe_sign_in
+    visit provider_interface_application_choice_interviews_path(application_choice)
+  end
+
```
despo, line 176: Is it worth moving the interview view logic in a component and verifying that the correct forms are rendered at that level than attempting to expand the system spec so it caters for all possible scenarios.

app/components/provider_interface/application_choice_interviews_list_component.html.erb
```diff
@@ -0,0 +1,25 @@
+<div class="app-interviews">
+  <% if user_can_create_or_change_interviews %>
+    <%= govuk_link_to 'Set up interview', new_provider_interface_application_choice_interview_path(application_choice), class: 'govuk-button govuk-button--secondary' %>
```
despo, line 3:
```suggestion
    <%= govuk_button_link_to 'Set up interview', new_provider_interface_application_choice_interview_path(application_choice), class: 'govuk-button--secondary' %>
```

despo (approved):
👍🏼

PR #4137, opened 2021-02-22
Only send email if edit interview actually changes the interview

benswannack:
## Context
Currently, an email will be sent even when there are no change to the interview details

Only save the interview and send an email if the updates actually change the interview

## Guidance to review
Have I missed any edge cases?

## Link to Trello card
https://trello.com/c/UMuoDuKF/3417-dont-send-an-email-following-the-change-interview-flow-if-no-details-were-changed

## Things to check

### DIFF
--- a/app/services/update_interview.rb
+++ b/app/services/update_interview.rb
@@ -23,12 +23,14 @@ class UpdateInterview
       course_option_id: interview.application_choice.offered_option.id,
     )
 
-    interview.update!(
-      provider: provider,
-      date_and_time: date_and_time,
-      location: location,
-      additional_details: additional_details,
-    )
+    interview.provider = provider
+    interview.date_and_time = date_and_time
+    interview.location = location
+    interview.additional_details = additional_details
+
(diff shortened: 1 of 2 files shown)


app/services/update_interview.rb
```diff
@@ -23,12 +23,14 @@ def save!
...
-      additional_details: additional_details,
-    )
+    interview.provider = provider
+    interview.date_and_time = date_and_time
+    interview.location = location
+    interview.additional_details = additional_details
+
+    return unless interview.changed?
+
+    interview.save!
```
despo, line 33: an other way of achieving this is still running the `update!` as it will only be executed if there are changes in the fields, and  returning unless `interview.saved_change_to_updated_at?`.

despo (approved):

app/services/update_interview.rb
benswannack: I'll go with what we've got now – I find it a bit more readable

PR #4138, opened 2021-02-22
Enable support agents to replace withdrawn application choices

stevehook:
## Context

We have had a few support requests to manually withdraw an application choice for a submitted application and replace it with a new course. This isn't possible in the support interface if the total number of choices (including withdrawn ones) is already at the upper limit (3 for apply 1, 1 for apply 2).

This PR permits additional application choices to be added if there are less than the maximum number of live choices.

- [x] Refine the logic around whether or not to display the _Add new course_ button to take withdrawn applications into account.
- [x] Fix 1+n query highlighted by Bullet after adding multiple application choices to an application and viewing it in the support interface.
- [ ] Update support guide

## Guidance to review

- Is this a safe thing to allow support agents to do?
- Not sure about the 1+n query fix as it's nothing directly to do with this change.

## Link to Trello card

https://trello.com/c/rYuPHyiQ/3068-enable-support-agents-to-switch-candidates-course-options

## Things to check

### DIFF
--- a/app/components/support_interface/application_add_course_component.html.erb
+++ b/app/components/support_interface/application_add_course_component.html.erb
@@ -4,7 +4,7 @@
   </p>
 <% elsif application_form.support_cannot_add_course_choice? %>
   <p class='govuk-body'>
-    This application already has the maximum number of active course choices (<%= application_form.maximum_number_of_course_choices %>), so we can't add any more.
+    This application already has the maximum number of active course choices (<%= application_form.maximum_number_of_course_choices %>), so we can't add any more until the candidate withdraws one or more course choices.
   </p>
 <% else %>
   <%= govuk_button_link_to 'Add a course', support_interface_application_form_search_course_new_path(application_form_id: application_form.id), form_class: 'govuk-!-display-inline-block' %>
--- a/app/components/support_interface/application_add_course_component.html.erb
+++ b/app/components/support_interface/application_add_course_component.html.erb
@@ -2,9 +2,9 @@
   <p class='govuk-body'>
     This candidate can add <%= application_form.maximum_number_of_course_choices %> course choices to this application.
   </p>
-<% elsif application_form.has_the_maximum_number_of_course_choices? %>
+<% elsif application_form.support_cannot_add_course_choice? %>
   <p class='govuk-body'>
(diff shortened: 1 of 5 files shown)


app/components/support_interface/application_add_course_component.html.erb
```diff
@@ -2,9 +2,9 @@
   <p class='govuk-body'>
     This candidate can add <%= application_form.maximum_number_of_course_choices %> course choices to this application.
   </p>
-<% elsif application_form.has_the_maximum_number_of_course_choices? %>
+<% elsif application_form.support_cannot_add_course_choice? %>
   <p class='govuk-body'>
-    This application already has the maximum number of course choices (<%= application_form.maximum_number_of_course_choices %>), so we can't add any more.
+    This application already has the maximum number of active course choices (<%= application_form.maximum_number_of_course_choices %>), so we can't add any more.
```
despo, line 7: Is it worth adding a note about requesting candidate to withdraw one of the three options if required, in order for this feature to become available; to avoid any relevant support requests

spec/system/support_interface/add_course_to_submitted_application_spec.rb
```diff
@@ -32,6 +32,14 @@
...
     and_i_click_add_course_to_application
     then_i_should_see_the_application_with_the_course_added
+
+    when_there_are_two_more_courses_added
+    and_i_visit_the_application_page
+    then_i_should_not_be_able_to_add_further_courses
+
+    when_one_course_is_withdrawn
+    and_i_visit_the_application_page
+    then_i_should_be_able_to_add_further_courses
```
despo, line 42: Consider moving these tests to the AddCourseComponent, where we should be verifying all possible scenarios.

app/components/support_interface/application_add_course_component.html.erb
stevehook: That's a good idea I think. In the support doc we give instructions to withdraw applications as well so maybe there's a case for making 'Withdraw application' a feature of the support interface too. What do you think?

despo: I think it's better to ask candidates to do that, as this option is already available to them (and the support agent would have to double check and confirm withdrawing any of their available options anyway as well as explain why they are doing it).

despo (approved):
👍🏼

PR #4139, opened 2021-02-22
Exclude the application status transition from cancelling an interview

steventux:
## Context

When an interview is cancelled the application reverts to `awaiting_provider_decision` status and this update appears in the timeline and activity log, the audit entry for the cancelled interview is more accurate in describing what has happened so we should filter out this status change.

Filter application choice status update from `interviewing` to `awaiting_provider_decision` when cancelling an interview. 
This is consistent with filtering the application status change when creating an interview.

image

## Guidance to review

Create an interview on an application and then cancel it, you should only see the setup and cancel events for the interview in the activity log.

Is there a more concise way to compare the status change array from `audited_changes` ? 

## Link to Trello card

https://trello.com/c/Bk1MQA1L/3419-timeline-activity-log-show-an-application-received-event-after-an-interview-was-cancelled

## Things to check

### DIFF
--- a/app/queries/get_activity_log_events.rb
+++ b/app/queries/get_activity_log_events.rb
@@ -52,6 +52,11 @@ class GetActivityLogEvents
     filtered_statuses = ApplicationStateChange::STATES_VISIBLE_TO_PROVIDER - IGNORE_STATUS
     status_transitions_to_include = filtered_statuses.map { |status| "'#{status}'" }.join(', ')
 
-    application_choice_audits_filter + " OR audited_changes::json#>>'{status, 1}' IN (#{status_transitions_to_include})"
+    application_choice_audits_filter += " OR audited_changes::json#>>'{status, 1}' IN (#{status_transitions_to_include})"
+    application_choice_audits_filter + ignore_interview_cancelled_application_choice_status_change_sql
+  end
+
+  def self.ignore_interview_cancelled_application_choice_status_change_sql
+    %( AND NOT audited_changes @> '{"status" : ["interviewing", "awaiting_provider_decision"]}')
   end
 end
--- a/spec/queries/get_activity_log_events_spec.rb
+++ b/spec/queries/get_activity_log_events_spec.rb
@@ -192,4 +192,30 @@ RSpec.describe GetActivityLogEvents, with_audited: true do
       expect(result.first.associated).to eq(application_choice)
     end
   end
+
+  context 'with a cancelled interview' do
+    let(:application_choice) { create(:application_choice, :with_scheduled_interview) }
+    let(:auth) { instance_double(ProviderAuthorisation, assert_can_make_decisions!: true) }
+
+    before do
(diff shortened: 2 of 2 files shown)


AgaDufrat (approved):
LGTM. 
I can't think of a better way of comparing the status change array myself.

app/queries/get_activity_log_events.rb
```diff
@@ -52,6 +54,15 @@ def self.application_choice_audits_filter_sql
...
+    application_choice_audits_filter += " OR audited_changes::json#>>'{status, 1}' IN (#{status_transitions_to_include})"
+    application_choice_audits_filter + interview_cancelled_application_choice_status_change_sql
+  end
+
+  def self.interview_cancelled_application_choice_status_change_sql
+    jsonb_status_clauses = EXCLUDE_APPLICATION_CHOICE_STATUS_CHANGE.map.with_index do |status, idx|
+      "audited_changes::json#>>'{status,#{idx}}' = '#{status}'"
+    end
+
+    " AND NOT(#{jsonb_status_clauses.join(' AND ')})"
```
mnacos, line 66: This method could perhaps be reduced to:

```suggestion
    " AND NOT audited_changes @> '{\"status\": [\"interviewing\", \"awaiting_provider_decision\"]}'"
```

Maybe we could also drop the new constant and the map.

app/queries/get_activity_log_events.rb
```diff
@@ -52,6 +54,15 @@ def self.application_choice_audits_filter_sql
     filtered_statuses = ApplicationStateChange::STATES_VISIBLE_TO_PROVIDER - IGNORE_STATUS
     status_transitions_to_include = filtered_statuses.map { |status| "'#{status}'" }.join(', ')
 
-    application_choice_audits_filter + " OR audited_changes::json#>>'{status, 1}' IN (#{status_transitions_to_include})"
+    application_choice_audits_filter += " OR audited_changes::json#>>'{status, 1}' IN (#{status_transitions_to_include})"
+    application_choice_audits_filter + interview_cancelled_application_choice_status_change_sql
+  end
+
+  def self.interview_cancelled_application_choice_status_change_sql
```
mnacos, line 61:
```suggestion
  def self.ignore_interview_cancelled_application_choice_status_change_sql
```
just as a form of documentation

mnacos (commented):
It works well and I agree with the approach, because it keeps the queries fast.

I've left a minor suggestion for brevity using the containment operator `@>`.

app/queries/get_activity_log_events.rb
steventux: Ah yes that's much simpler, I was looking for a way to do an array comparison but hadn't thought of this - thanks :+1:

steventux:
@mnacos amended here https://github.com/DFE-Digital/apply-for-teacher-training/pull/4139/commits/db4b7eb572d6cb4210f7871702b74a3f913213fe

mnacos (approved):
great, 🚢 it

PR #4140, opened 2021-02-22
Attempt to fix E&D export

tijmenb:
## Context

We're seeing this on prod:

https://sentry.io/organizations/dfe-bat/issues/2200848096

I assume that some records may not have `equality_and_diversity['disabilities']` specified, which this should
fix.

## Link to Trello card

https://sentry.io/organizations/dfe-bat/issues/2200848096

## Things to check


stevehook (approved):
Makes sense to me.

There is one record that matches:
```
select id, equality_and_diversity
from application_forms
where equality_and_diversity is not null
  and equality_and_diversity->'disabilities' is null;
```
By the way, thanks for adding Blazer!

PR #4145, opened 2021-02-23
Proof of concept - persisted feature metrics dashboard

malcolmbaig:
## Context
The feature metrics dashboard takes ~1.5 minutes to load. We need to make this faster.

This is a proof of concept which illustrates:

- Using a dedicated table to hold the statistics we display on the dashboard.
- Updating this table hourly via a Sidekiq worker.
- Moving retrieval of all metrics into the FeatureMetricsDashboard ActiveRecord model.
- Calling methods on an instance of this model when rendering the view.

## Guidance to review

- Does this work? Does it make sense?
- Alternatives considered:
  - Optimise the existing SQL queries, especially for work history and magic links. I prioritised the approach in this PR instead, as the existing queries are fairly complex as is. Rather than additional SQL complexity, I've opted for more complexity around the update mechanism for the dashboard. This is probably easier to reason about and gives us the biggest speed increase.

## Link to Trello card
https://trello.com/c/NAENgBR6

## Things to check

### DIFF
--- a/app/views/support_interface/performance/feature_metrics_dashboard.html.erb
+++ b/app/views/support_interface/performance/feature_metrics_dashboard.html.erb
@@ -14,9 +14,7 @@
   <div class="govuk-grid-column-one-half">
     <div id="headline-stat" class="govuk-!-margin-bottom-4">
       <%= render SupportInterface::TileComponent.new(
-        count: @reference_statistics.average_time_to_get_references(
-          EndOfCycleTimetable.apply_reopens.beginning_of_day,
-        ),
+        count: @dashboard.avg_time_to_get_references,
         label: 'average time to get reference back',
         colour: :blue,
         size: :reduced,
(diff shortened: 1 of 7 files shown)


app/workers/update_feature_metrics_dashboard.rb
```diff
@@ -0,0 +1,11 @@
+class UpdateFeatureMetricsDashboard
+  include Sidekiq::Worker
+
+  sidekiq_options retry: 0, queue: :low_priority
+
+  def perform
+    dashboard = FeatureMetricsDashboard.new
+    dashboard.load_updated_metrics
+    dashboard.save!
+  end
```
malcolmbaig, line 10: Should we delete the old record as we go? One interesting (but not immediately useful) side effect of keeping these records is the historical view of these metrics over time.

stevehook: I would say keep the records, they are not costing us much.

stevehook (commented):
I think the overall approach is sound. Having data that is stale by at most an hour is perfectly fine I would have thought. It's going to put a bit of load on the worker process, but I don't think that's an issue because there would be only one of the worker threads blocked while it runs.

The only thing I'm a bit unsure about is the schema for storing this kind of report. I know there are plans for a version 2 of these metrics so what happens when the schema changes? I just wonder whether it would be better going with a flexible data model e.g. by storing the data payload in a JSONB column. That has downsides but it would mean that we don't ever have to worry about the schema changing.

I think it's worth keeping the historical records, hourly updates would generate ~9K records per year. Perhaps we'd never use them though. The value of the historical data would be questionable though as soon as you extend the schema.

app/views/support_interface/performance/feature_metrics_dashboard.html.erb
```diff
@@ -14,9 +14,7 @@
   <div class="govuk-grid-column-one-half">
```
stevehook, line 14: probably should add a _last generated at ..._ message on this view somewhere.

malcolmbaig: Yeah definitely, was thinking a 'last generated at' and 'next update expected' would be useful.

malcolmbaig:
Thanks for the feedback @stevehook .

Closing this PR for now while I work on the actual implementation. Will be proceeding with the approach outlined here, with the likely change of using a JSONB column for the data.

PR #4146, opened 2021-02-23
Add new guidance to course search page on Support UI

george-bit:
## Context

In order to assist the support team in using the support UI more effectively, further guidance needs
to be added to the search course view.

image

## Guidance to review

Is it worth extracting some of this text into a partial?

## Link to Trello card

https://trello.com/c/GeEcKIP2/3025-dev-support-ui-guidance-to-the-add-course-flow

## Things to check

### DIFF
--- a/app/views/support_interface/application_forms/courses/new_search.html.erb
+++ b/app/views/support_interface/application_forms/courses/new_search.html.erb
@@ -4,6 +4,9 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <p class="govuk-body">A candidate can only add up to 3 courses. They cannot add the same course twice.</p>
+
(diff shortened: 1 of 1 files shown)


despo:
When possible we should not be asking developers to withdraw candidate options, but instead the support agents should ask the candidate to withdraw their least wanted/added in error course option (as that is what the developers are doing, but manually which is more error prone). That should enable the agents to manage the rest of the process on their own (look at #4138)

george-bit:
Good point @despo, I'm uncertain who wrote the copy for this content change as it was carded up by David based on a document from the support team. It's a bit of a grey area as to who is responsible for the content. I'll ask in the support channel. 👍

george-bit:
@despo - I've spoken to support and they agree with the above comment, I have updated the copy to be more reflective please let me know what you think.

@paulrobertlloyd - feel free to pull apart my grammar 😂

paulrobertlloyd (approved):

PR #4147, opened 2021-02-23
Stub state diagram generation

despo:
Spec takes ~10 seconds to run and not testing the generated diagram
Added expectations to ensure that calls to service still take place.

### DIFF
--- a/spec/system/support_interface/docs_spec.rb
+++ b/spec/system/support_interface/docs_spec.rb
@@ -6,11 +6,13 @@ RSpec.feature 'Docs' do
   scenario 'Support user visits process documentation' do
     given_i_am_a_support_user
     when_i_visit_the_process_documentation
-    then_i_see_the_provider_flow_documentation
+    then_the_application_state_diagram_is_generated
+    and_i_see_the_provider_flow_documentation
     and_it_contains_documentation_for_all_emails
 
     when_i_click_on_candidate_flow_documentation
-    then_i_see_the_candidate_flow_documentation
+    then_the_process_state_diagram_is_generated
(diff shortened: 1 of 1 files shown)


spec/system/support_interface/docs_spec.rb
```diff
@@ -18,7 +18,10 @@ def given_i_am_a_support_user
   end
 
   def when_i_visit_the_process_documentation
+    allow(ApplicationStateChange).to receive(:workflow_spec).and_return(Struct.new(:states).new([]))
+
     visit support_interface_docs_provider_flow_path
+    expect(ApplicationStateChange).to have_received(:workflow_spec).exactly(3).times
```
cwrw, line 24: Maybe we can move this to its own `then` step, as I think usually we should avoid any expectations in `when` steps, what do you think?

cwrw (commented):
This is great, spec went from 10 sec to 0.2 sec, just a comment around the stpes

spec/system/support_interface/docs_spec.rb
despo: Good point. I have moved the expectations to their own `then` steps.

cwrw (approved):

tijmenb (commented):
I wonder if we should stub the SVG generation at a higher level, would stubbing `allow(StateDiagram).to receive(:svg)` work as well? That avoids stubbing the internals.

tijmenb:
Hmm, on second thought, this change removes all content from the flow pages, so the code is excercised at all. Is that the intention? In that case, we should remove all tests except the emails, because they aren't doing anything.

PR #4148, opened 2021-02-23
Small change to QA email view

raamSoftwire:
## Context

As a content designer, I can't see the reference name of the email so I have to jump back and forth. Also, without the name, if you're looking at a number of different emails it gets confusing.

Add the mailer and email reference names to the email preview header

Before

image

After

image

## Guidance to review

- commit by commit 
- the first commit is a straight copy from the Rails repo
- do we need tests here?
- this change isn't scoped by environment. Is that a problem?

## Link to Trello card

https://trello.com/c/5Lf2AxrW/3033-not-a-small-change-to-qa-email-view

## Things to check

### DIFF
--- a/spec/mailers/previews/email.html.erb
+++ b/spec/mailers/previews/email.html.erb
@@ -1,0 +1,164 @@
+<!--from https://github.com/rails/rails/blob/6-0-stable/railties/lib/rails/templates/rails/mailers/email.html.erb-->
+
+<!DOCTYPE html>
+<html><head>
+  <meta name="viewport" content="width=device-width" />
+  <style type="text/css">
+    html, body, iframe {
+      height: 100%;
+    }
+
+    body {
+      margin: 0;
+    }
+
+    header {
+      width: 100%;
+      padding: 10px 0 0 0;
+      margin: 0;
+      background: white;
+      font: 12px "Lucida Grande", sans-serif;
+      border-bottom: 1px solid #dedede;
+      overflow: hidden;
+    }
+
+    dl {
+      margin: 0 0 10px 0;
+      padding: 0;
+    }
+
+    dt {
+      width: 80px;
+      padding: 1px;
+      float: left;
+      clear: left;
+      text-align: right;
+      color: #7f7f7f;
+    }
+
+    dd {
+      margin-left: 90px; /* 80px + 10px */
+      padding: 1px;
+    }
+
+    dd:empty:before {
+      content: "\00a0"; // &nbsp;
+    }
+
+    iframe {
+      border: 0;
+      width: 100%;
+    }
+  </style>
+</head>
+
+<body>
+<header>
+  <dl>
+    <% if @email.respond_to?(:smtp_envelope_from) && Array(@email.from) != Array(@email.smtp_envelope_from) %>
+      <dt>SMTP-From:</dt>
+      <dd><%= @email.smtp_envelope_from %></dd>
+    <% end %>
+
+    <% if @email.respond_to?(:smtp_envelope_to) && @email.to != @email.smtp_envelope_to %>
+      <dt>SMTP-To:</dt>
+      <dd><%= @email.smtp_envelope_to %></dd>
+    <% end %>
+
+    <dt>From:</dt>
+    <dd><%= @email.header['from'] %></dd>
+
+    <% if @email.reply_to %>
+      <dt>Reply-To:</dt>
+      <dd><%= @email.header['reply-to'] %></dd>
+    <% end %>
+
+    <dt>To:</dt>
+    <dd><%= @email.header['to'] %></dd>
+
+    <% if @email.cc %>
+      <dt>CC:</dt>
+      <dd><%= @email.header['cc'] %></dd>
+    <% end %>
+
+    <dt>Date:</dt>
+    <dd><%= Time.current.rfc2822 %></dd>
+
+    <dt>Subject:</dt>
+    <dd><strong><%= @email.subject %></strong></dd>
+
+    <% unless @email.attachments.nil? || @email.attachments.empty? %>
+      <dt>Attachments:</dt>
+      <dd>
+        <% @email.attachments.each do |a| %>
+          <% filename = a.respond_to?(:original_filename) ? a.original_filename : a.filename %>
+          <%= link_to filename, "data:application/octet-stream;charset=utf-8;base64,#{Base64.encode64(a.body.to_s)}", download: filename %>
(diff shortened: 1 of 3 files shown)


raamSoftwire:
@malcolmbaig the `render_preview_wrapper` method in the `mail-notify` gem is `private`. Should we be able to do this? Does this surprise you?

Tomcowan1:
Looks good to me, thanks Raam!

Tomcowan1 (approved):

spec/mailers/previews/email.html.erb
```diff
@@ -0,0 +1,170 @@
...
+      width: 120px;
+      padding: 1px;
+      float: left;
+      clear: left;
+      text-align: right;
+      color: #7f7f7f;
+    }
+
+    dd {
+      margin-left: 90px; /* 80px + 10px */
```
paulrobertlloyd, line 40:
```suggestion
      margin-left: 125px; /* 120px + 5px */
```

spec/mailers/previews/email.html.erb
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    dl {
+      margin: 0 0 10px 0;
+      padding: 0;
+    }
+
+    dt {
+      width: 120px;
+      padding: 1px;
```
paulrobertlloyd, line 32: Let’s increase the padding a tiny bit 🤏

```suggestion
      padding: 2px;
```

spec/mailers/previews/email.html.erb
```diff
@@ -0,0 +1,170 @@
...
+      float: left;
+      clear: left;
+      text-align: right;
+      color: #7f7f7f;
+    }
+
+    dd {
+      margin-left: 90px; /* 80px + 10px */
+      padding: 1px;
```
paulrobertlloyd, line 41: Let’s increase the padding a tiny bit 🤏

```suggestion
      padding: 2px;
```

spec/mailers/previews/email.html.erb
```diff
@@ -0,0 +1,170 @@
...
+    <% if @email.respond_to?(:smtp_envelope_to) && @email.to != @email.smtp_envelope_to %>
+      <dt>SMTP-To:</dt>
+      <dd><%= @email.smtp_envelope_to %></dd>
+    <% end %>
+
+    <dt>Mailer:</dt>
+    <dd><%= @email.instance_variable_get(:@mailer_class).instance_variable_get(:@mailer_name).humanize %></dd>
+
+    <dt>Reference name:</dt>
+    <dd><%= @email.instance_variable_get(:@action) %></dd>
```
paulrobertlloyd, line 72: Can we combine these two fields to reduce the overall height of this header?

```suggestion
    <dt>Template:</dt>
    <dd><%= @email.instance_variable_get(:@action) %> (<%= @email.instance_variable_get(:@mailer_class).instance_variable_get(:@mailer_name).humanize %>)</dd>
```

spec/mailers/previews/email.html.erb
paulrobertlloyd: Or leave as is if you decide to accept the other recommendation which combines the two new fields under a single ‘Template’ key

spec/mailers/previews/email.html.erb
```diff
@@ -0,0 +1,170 @@
...
+
+    <dt>From:</dt>
+    <dd><%= @email.header['from'] %></dd>
```
paulrobertlloyd, line 75: Can re remove this given it’s always empty?

spec/mailers/previews/email.html.erb
```diff
@@ -0,0 +1,170 @@
...
+    <% if I18n.available_locales.count > 1 %>
+      <dt>Locale:</dt>
+      <dd>
+        <select id="locale" onchange="refreshBody(true);">
+          <% I18n.available_locales.each do |locale| %>
+            <option <%= I18n.locale == locale ? 'selected' : '' %> value="<%= locale_query(locale) %>"><%= locale %></option>
+          <% end %>
+        </select>
+      </dd>
+    <% end %>
```
paulrobertlloyd, line 127: Do we need to show these values?

paulrobertlloyd:
Suggested a few tweaks, after which the header will look more like this:

Screenshot 2021-02-23 at 18 32 17

spec/mailers/previews/email.html.erb
raamSoftwire: I've accepted the other change and left this one.

raamSoftwire:
@paulrobertlloyd Cheers for the styling updates, it looks much neater now

spec/mailers/previews/email.html.erb
raamSoftwire: Nope. I've removed them too

malcolmbaig:
> @malcolmbaig the `render_preview_wrapper` method in the `mail-notify` gem is `private`. Should we be able to do this? Does this surprise you?

Ideally the gem would expose some public mechanism for us to declare which template is used for the preview wrapper. In the absence of that, I think overriding the private method is fine - especially as the worst that can happen is that previews start breaking after a mail-notify version update. 

An alternative option with a smaller footprint—and worth considering applying here—is the following:

- Add `prepend_view_path('spec/mailers/previews')` just before this line https://github.com/DFE-Digital/apply-for-teacher-training/commit/e901fbad662db5b08faa18652b03e582c4b7e5f4#diff-e530c57fe094b50f394218dfd0b938dda863d461c8a1d61301dece5925367048R4
- Rename our custom template to `spec/mailers/previews/rails/mailers/email.html.erb`
- Remove the `render_preview_wrapper` override.

The `render_preview_wrapper` in `mail-notify` will then look for the email template in our directory before searching the default view paths.

spec/mailers/previews/email.html.erb
```diff
@@ -0,0 +1,141 @@
...
+      overflow: hidden;
+    }
+
+    dl {
+      margin: 0 0 10px 0;
+      padding: 0;
+    }
+
+    dt {
+      width: 120px;
```
paulrobertlloyd, line 31: Just need to revert this to the width used by the original Rails component.

```suggestion
      width: 80px;
```

raamSoftwire: Done

raamSoftwire:
@malcolmbaig thanks again for the help with this. I've made those changes now

paulrobertlloyd (approved):

malcolmbaig (approved):

PR #4153, opened 2021-02-24
Make timeline links unique to address DAC recommendations

AgaDufrat:
## Context

Multiple links with the same name could be found on the Timeline page without a
context. This could be confusing for users who want to navigate outside of the
context. We are adding visually hidden date and time information so the purpose
of each link can be identified.
Screenshot 2021-02-23 at 10 58 23
Screenshot 2021-02-23 at 10 58 33

We are adding visually hidden date and time information so the purpose
of each link can be identified.

Screenshot 2021-02-23 at 11 24 07

## Things to check

### DIFF
--- a/app/components/provider_interface/application_timeline_component.html.erb
+++ b/app/components/provider_interface/application_timeline_component.html.erb
@@ -11,7 +11,9 @@
       </p>
       <div class="app-timeline__event_link">
         <% if event.link_name && event.link_path %>
-          <%= govuk_link_to event.link_name, event.link_path, class: 'govuk-link--no-visited-state' %>
+          <%= govuk_link_to event.link_path, class: 'govuk-link--no-visited-state' do %>
+            <%= event.link_name %> <span class="govuk-visually-hidden"><%= event.date.to_s(:govuk_date_and_time) %></span>
+          <% end %>
         <% end %>
       </div>
     </div>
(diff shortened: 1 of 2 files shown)


app/components/provider_interface/application_timeline_component.html.erb
```diff
@@ -11,7 +11,9 @@
       </p>
       <div class="app-timeline__event_link">
         <% if event.link_name && event.link_path %>
-          <%= govuk_link_to event.link_name, event.link_path, class: 'govuk-link--no-visited-state' %>
+          <%= link_to event.link_path, class: 'govuk-link govuk-link--no-visited-state' do %>
```
cwrw, line 14: could we use `govuk_link_to` instead of `link_to`?

cwrw (approved):
Looks good, just a minor comment 👍

app/components/provider_interface/application_timeline_component.html.erb
```diff
@@ -11,7 +11,9 @@
       </p>
       <div class="app-timeline__event_link">
         <% if event.link_name && event.link_path %>
-          <%= govuk_link_to event.link_name, event.link_path, class: 'govuk-link--no-visited-state' %>
+          <%= govuk_link_to event.link_path, class: 'govuk-link govuk-link--no-visited-state' do %>
```
cwrw, line 14:
```suggestion
          <%= govuk_link_to event.link_path, class: 'govuk-link--no-visited-state' do %>
```

PR #4143, opened 2021-02-23
Fix daily run of TAD reports

tijmenb:
## Context

I broke this in https://github.com/DFE-Digital/apply-for-teacher-training/pull/4105.

## Link to Trello card

https://sentry.io/organizations/dfe-bat/issues/2232004786

PR #4155, opened 2021-02-24
Really fix the TAD export clockwork job

tijmenb:
## Context

I tried to fix this in https://github.com/DFE-Digital/apply-for-teacher-training/pull/4143, and @duncanjbrown suggested in the PR to add a test - which I didn't do (actually I looked at it but it seemed complicated).

In the least surprising event of the year, my fix didn't fix the issue - and a test would have totally caught this.

Add a test that loops through all clockwork jobs, and see if they raise any error. And fix the original issue.

## Guidance to review

Ok?

## Link to Trello card

https://sentry.io/organizations/dfe-bat/issues/2234062504

## Things to check

### DIFF
--- a/app/services/data_api/tad_export.rb
+++ b/app/services/data_api/tad_export.rb
@@ -4,7 +4,7 @@ module DataAPI
 
     def self.run_daily
       data_export = DataExport.create!(name: EXPORT_NAME)
-      DataExporter.perform_async(TADExport::TADExport, data_export.id)
+      DataExporter.perform_async(DataAPI::TADExport, data_export.id)
     end
 
     def self.latest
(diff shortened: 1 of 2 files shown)


despo:
@tijmenb this is reverting with the change I merged this morning. You can just add `clockwork: true` to make sure the clock is cleared

https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/spec/spec_helper.rb#L121
https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/spec/support_specs/clockwork_spec.rb#L20

tijmenb:
@despo Nice! Rebased.

despo (approved):
👍🏼

PR #3184, opened 2020-10-19
Use provider application patterns on support application page

paulrobertlloyd:
## Context

Use provider application review components to display work history and volunteering items in support and generally tidy up the relationship between these two similar pages and their respective components.

## Changes proposed in this pull request

* Moves headings for application review components into these components (DRYer)
* Adjusts size of headings in these components from large to 27px (partly addressing [this change](https://bat-design-history.netlify.app/manage-teacher-training-applications/smaller-headings/))
* Remove application related rows from personal details review component
* Add a provider-facing application summary component
* Include recruitment cycle information in this new application summary (matching prototype)
* Update name spacing for work history and volunteering components:
  * `ProviderInterface::WorkHistoryComponent` → `WorkHistoryComponent` 
  * `ProviderInterface::VolunteeringHistoryComponent` → `VolunteeringHistoryComponent`
  * `ProviderInterface::WorkHistoryItemComponent` → `WorkHistoryItemComponent `
  * `WorkHistoryReviewComponent` → `CandidateInterface::WorkHistoryReviewComponent`
  * `VolunteeringReviewComponent` → `CandidateInterface::VolunteeringReviewComponent`
* Add context regarding cycle year (apply again or carried over) to application summary in support)

| Before | After |
| - | - |
| support-application-old | support-application-new |
| provider-application-old | provider-application-new |

## Link to Trello card

https://trello.com/c/qfGonFDe

## Things to check

- [ ] This code does not rely on migrations in the same Pull Request
- [ ] If this code includes a migration adding or changing columns, it also backfills existing records for consistency
- [ ] API release notes have been updated if necessary
- [ ] New environment variables have been [added to the Azure config](https://github.com/DFE-Digital/apply-for-teacher-training#azure-hosting-devops-pipeline)

PR #4156, opened 2021-02-24
Fix layout of application in Support and Manage

paulrobertlloyd:
## Context

Layout of application in support is incorrect; application components now specify their own widths individually, so shouldn’t appear within a two-thirds column as they fo now, which further reduces the width of answers shown.

Thought I had addressed this in #3184, but evidently not.

| Before | After |
| - | - |
| Screenshot Before | Screenshot After |

## Guidance to review

* Need to double-check this doesn’t break layout on Manage

## Things to check

### DIFF
--- a/app/components/efl_qualification_card_component.html.erb
+++ b/app/components/efl_qualification_card_component.html.erb
@@ -1,29 +1,28 @@
-<div class="govuk-grid-column-full">
-  <h2 class="govuk-heading-m govuk-!-margin-top-8" id="english-as-a-foreign-language">English as a foreign language</h2>
-  <p class="govuk-body"><%= qualification_status %></p>
+<h3 class="govuk-heading-m govuk-!-margin-top-6" id="english-as-a-foreign-language">English as a foreign language</h3>
 
-  <% if english_proficiency.has_qualification? %>
-    <div class="govuk-grid">
-      <div class="govuk-grid-row app-grid-row--flex">
-        <div class="govuk-grid-column-one-third">
-          <div class="app-card app-card--outline" data-qa="english-proficiency-qualification">
-            <h3 class="govuk-heading-s govuk-!-margin-bottom-1"><%= name %></h3>
-            <dl class="app-qualification">
-              <dt class="app-qualification__key govuk-visually-hidden">Awarded</dt>
-              <dd class="app-qualification__value app-qualification__value--caption"><%= award_year %></dd>
-              <dt class="app-qualification__key"><%= grade_title %></dt>
-              <dd class="app-qualification__value"><%= grade %></dd>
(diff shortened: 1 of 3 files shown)


app/components/efl_qualification_card_component.html.erb
```diff
@@ -1,29 +1,28 @@
-<div class="govuk-grid-column-full">
```
paulrobertlloyd, line 1: I’ve removed this grid container, and thus reduced the indentation of each line

app/components/efl_qualification_card_component.html.erb
```diff
@@ -1,29 +1,28 @@
...
-  <h2 class="govuk-heading-m govuk-!-margin-top-8" id="english-as-a-foreign-language">English as a foreign language</h2>
-  <p class="govuk-body"><%= qualification_status %></p>
+<h3 class="govuk-heading-m govuk-!-margin-top-6" id="english-as-a-foreign-language">English as a foreign language</h3>
```
paulrobertlloyd, line 1: This should be `h3` as sits within the `QualificationsComponent` which has a `h2` heading

app/components/qualifications_table_component.html.erb
```diff
@@ -1,19 +1,18 @@
 <% if qualifications.any? %>
-  <div class="govuk-grid-column-two-thirds">
```
paulrobertlloyd, line 2: I’ve removed this grid container, and thus reduced the indentation of each line.

app/components/qualifications_table_component.html.erb
```diff
@@ -1,19 +1,18 @@
...
-      <tbody class="govuk-table__body">
-        <% qualifications.each do |qualification| %>
-          <%= render QualificationRowComponent.new(qualification: qualification) %>
-        <% end %>
-      </tbody>
-    </table>
-  </div>
+  <h3 class="govuk-heading-m govuk-!-margin-top-6" id="other-qualifications"><%= header %></h3>
+
+  <table class="govuk-table govuk-!-width-two-thirds" data-qa="qualifications-table-<%= header.parameterize %>">
```
paulrobertlloyd, line 4: Applied a width to the table instead.

app/views/support_interface/application_forms/show.html.erb
```diff
@@ -39,18 +39,14 @@
 
 <%= render QualificationsComponent.new(application_form: @application_form, show_hesa_codes: true) %>
 
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
```
paulrobertlloyd, line 43: Removed this grid container, as no longer needed given each component dictates its own width.

malcolmbaig (approved):

PR #4157, opened 2021-02-24
Handle non-integer values for app choice id filter

stevehook:
## Context

Inputting a non-integer value for the _Provider application id_ filter in the Support interface application forms page causes a crash.

https://sentry.io/organizations/dfe-bat/issues/2171471392/?project=1765973&referrer=slack

Convert the input value to an integer in the filter so that it's guaranteed to be valid when it reaches the database.

image

## Guidance to review

There is no validation in the filter so I think the correct thing to do is to treat this as a non-matching value by coercing it to an integer before it gets to the database. Because `'ABC'.to_i` evaluates to `0` it leads to no matching application choices which is the 'expected' result. Seem reasonable? Is there a better way?

## Link to Trello card

https://trello.com/c/GwgOI9qU/2915-crash-with-non-integer-value-for-provider-application-id-filter-in-support-interface

## Things to check

### DIFF
--- a/app/models/support_interface/applications_filter.rb
+++ b/app/models/support_interface/applications_filter.rb
@@ -23,7 +23,7 @@ module SupportInterface
       end
 
       if applied_filters[:application_choice_id].present?
-        application_forms = application_forms.joins(:application_choices).where('application_choices.id = ?', applied_filters[:application_choice_id])
+        application_forms = application_forms.joins(:application_choices).where('application_choices.id = ?', applied_filters[:application_choice_id].to_i)
       end
 
       if applied_filters[:phase]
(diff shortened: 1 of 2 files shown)


tijmenb (approved):
This looks good 👍 . I wonder if we could avoid it altogether by merging the 2 search fields?

stevehook:
> This looks good 👍 . I wonder if we could avoid it altogether by merging the 2 search fields?

Yes I think that would work, it would make the whole thing a bit tidier...

duncanjbrown:
One concern: what if a user confidently types a support ref into this and then clicks on the application, not realising it's not what they wanted at all?

stevehook:
@duncanjbrown @tijmenb I've tried out Tijmen's suggestion and and it's generating quite a few false positives so I think on balance it's best left as separate fields.

Application choice ids tend to match support refs and each other when compared with an `ILIKE`. Admittedly that's made worse running locally because my ids are just 2/3 digits but I'm pretty sure there will be false matches with 'real-life' ids too.

PR #4158, opened 2021-02-24
Support: Show useful links relating to an application

paulrobertlloyd:
## Context

There are a number of places/services where useful information about an application can be found. It can help support agents to have these links available to them from a candidate’s application page.

* Show additional application links to right of page title
* Move sign in as candidate below header to match design used for provider users

application

## Guidance to review

* Need to add paths for Vendor API requests, Sentry and Logit (as shown in Trello card) – not sure how to do this.

## Link to Trello card

https://trello.com/c/4zc8EbNK/

## Things to check

### DIFF
--- a/app/components/support_interface/application_navigation_component.rb
+++ b/app/components/support_interface/application_navigation_component.rb
@@ -1,0 +1,50 @@
+module SupportInterface
+  class ApplicationNavigationComponent < ViewComponent::Base
+    include ViewHelper
+
+    def initialize(application_form)
+      @application_form = application_form
+    end
+
+    def links
+      [
+        email_log_link,
+        logit_link,
+        sentry_link,
+      ]
+    end
+
+  private
+
+    def email_log_link
+      {
+        title: 'Emails about this application',
+        href: paths.support_interface_email_log_path(application_form_id: @application_form.id),
+      }
+    end
+
+    def sentry_link
+      link = "https://sentry.io/organizations/dfe-bat/issues/?project=1765973&query=user.id%3A%22candidate_#{@application_form.candidate_id}%22"
+
+      {
+        title: 'Sentry errors for this candidate',
+        href: link,
+      }
+    end
+
+    def logit_link
+      environment = HostingEnvironment.environment_name
+      candidate_id = @application_form.candidate_id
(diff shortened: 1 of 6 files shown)


app/views/support_interface/application_forms/_application_navigation.html.erb
```diff
@@ -15,9 +15,27 @@
...
+    <% end %>
+  </div>
+
+  <div class="govuk-grid-column-one-third">
+    <div class="govuk-inset-text app-inset-text--important app-inset-text--narrow-border govuk-!-margin-top-0">
+      <h2 class="govuk-heading-s govuk-!-margin-bottom-2">Useful links for this application</h2>
+      <ul class="govuk-list govuk-!-font-size-16">
+        <li><%= govuk_link_to 'Emails sent', support_interface_email_log_path(application_form_id: @application_form.id), class: 'govuk-link--no-visited-state' %></li>
+        <li><%= govuk_link_to 'Logit logs', '#', class: 'govuk-link--no-visited-state' %></li>
+        <li><%= govuk_link_to 'Sentry errors', '#', class: 'govuk-link--no-visited-state' %></li>
```
despo, line 33: Support agents do not have access to neither sentry or logit, and they should not be expected to understand those errors either, so I don't think there is value in adding these two links.

paulrobertlloyd: These links are in the trello card in the support feature wish list… although I can’t see who originally wrote the card.

paulrobertlloyd: Good point though… maybe we should only include internal links?

duncanjbrown: I wrote it! The idea was to have links that were useful for developers in a "Useful links for developers" section

duncanjbrown:
I'll add the `href`s!

duncanjbrown:
I removed the vendor API requests as those are per-choice — we can find another place to put them

app/components/support_interface/application_navigation_component.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    def initialize(application_form)
+      @application_form = application_form
+    end
+
+    def links
+      [
+        email_log_link,
+        sentry_link,
+        logit_link,
```
paulrobertlloyd, line 13: Can we have these in alphabetical order?

duncanjbrown: Your wish is my command 🙇

paulrobertlloyd: :heart:

duncanjbrown (approved):

PR #4159, opened 2021-02-24
Refactor primary navigation component

paulrobertlloyd:
## Context

Inadvertently found myself looking at the markup and styles for this component and finding a lot of extraneous markup and CSS – possibly die to the design being inherited from the MoJ pattern which had other requirements.

* Remove extraneous `div`s on markup
* Use `govuk-width-container` instead of custom container styles (where this all started – we may want to override this within support, and having this use a different class was causing issues)
* Refactor CSS; rather than use generated content to draw the underline, use inset box shadow instead (this is what’s used elsewhere in the design system). This massively reduced the complexity and verbosity of the CSS for this component.
* Also use mixins and variables from `govuk-frontend` which means styles are self documented 

## Guidance to review

@adamsilver @simonwhatley No visual changes – anything in this change look problematic?

@edwardhorsford You may want to see if these changes are relevant to the same component in Register

## Things to check

### DIFF
--- a/app/components/primary_navigation_component.html.erb
+++ b/app/components/primary_navigation_component.html.erb
@@ -4,7 +4,7 @@
       <ul class="app-primary-navigation__list">
         <% @navigation_items.each do |nav_item| %>
           <li class="app-primary-navigation__item <%= nav_item.classes %>">
-            <a class="app-primary-navigation__link"<%= nav_item.active ? ' aria-current=page' : ' ' %>href="<%= nav_item.href %>">
+            <a class="app-primary-navigation__link"<%= nav_item.active ? 'aria-current=page' : '' %> href="<%= nav_item.href %>">
               <%= nav_item.text %>
             </a>
           </li>
--- a/app/components/primary_navigation_component.html.erb
+++ b/app/components/primary_navigation_component.html.erb
@@ -2,11 +2,13 @@
   <div class="govuk-width-container">
     <nav class="app-primary-navigation__nav" aria-label="main menu">
       <ul class="app-primary-navigation__list">
-        <% @navigation_items.each do |nav_item| %>
-          <li class="app-primary-navigation__item <%= nav_item.classes %>">
-            <a class="app-primary-navigation__link"<%= nav_item.active ? 'aria-current=page' : '' %> href="<%= nav_item.href %>">
-              <%= nav_item.text %>
-            </a>
+        <% items.each do |item| %>
+          <li class="app-primary-navigation__item <%= item.classes %>">
+            <% if item.active %>
+              <%= govuk_link_to item.text, item.href, class: 'app-primary-navigation__link', aria: { current: 'page' } %>
+            <% else %>
+              <%= govuk_link_to item.text, item.href, class: 'app-primary-navigation__link' %>
+            <% end %>
           </li>
         <% end %>
       </ul>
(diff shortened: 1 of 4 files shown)


app/components/primary_navigation_component.html.erb
```diff
@@ -1,19 +1,15 @@
...
-          <% end %>
-        </ul>
-      </nav>
-    </div>
+  <div class="govuk-width-container">
+    <nav class="app-primary-navigation__nav" aria-label="main menu">
+      <ul class="app-primary-navigation__list">
+        <% @navigation_items.each do |nav_item| %>
+          <li class="app-primary-navigation__item <%= nav_item.classes %>">
+            <a class="app-primary-navigation__link"<%= nav_item.active ? 'aria-current=page' : '' %> href="<%= nav_item.href %>">
```
edwardhorsford, line 7:
```suggestion
            <a class="app-primary-navigation__link"<%= nav_item.active ? 'aria-current=page' : '' %> href="<%= nav_item.href %>">
```
Shouldn't `page` be in quotes?

app/components/primary_navigation_component.html.erb
```diff
@@ -1,19 +1,15 @@
...
-                <%= nav_item.active ? 'aria-current=page' : '' %>
-                 href="<%= nav_item.href %>">
-                 <%= nav_item.text %>
-              </a>
-            </li>
-          <% end %>
-        </ul>
-      </nav>
-    </div>
+  <div class="govuk-width-container">
```
edwardhorsford, line 2: Should this have `app-primary-navigation` on it?

app/frontend/styles/_primary-navigation.scss
```diff
@@ -2,30 +2,17 @@
...
-.app-primary-navigation__container {
-  @include govuk-width-container;
-  font-size: 0; // Hide whitespace between elements
-  text-align: justify; // Trick to remove the need for floats
-
-  &:after {
-    content: '';
-    display: inline-block;
-    width: 100%;
-  }
```
edwardhorsford, line 14: What's this for?

app/frontend/styles/_primary-navigation.scss
```diff
@@ -2,30 +2,17 @@
...
   list-style: none;
   margin: 0;
   padding: 0;
-  @include govuk-clearfix;
 }
 
 .app-primary-navigation__item {
   @include govuk-font($size: 19);
-  margin-right: govuk-spacing(3);
   float: left;
```
edwardhorsford, line 14: Comment above says that `justify` is used to remove the need for floats.

app/components/primary_navigation_component.html.erb
paulrobertlloyd: `app-primary-navigation` is the parent `div` that stretches full width and adds the background colour.
`govuk-width-container` fixes the width of the content within this `div`.

Each thing does one thing, and this follows the pattern of other common components, like the header and footer.

app/frontend/styles/_primary-navigation.scss
paulrobertlloyd: @adamsilver thinks its related to the MoJ component which can contain a search field. We don’t have this, so isn’t needed. https://moj-design-system.herokuapp.com/components/primary-navigation/examples/inline-search

app/frontend/styles/_primary-navigation.scss
paulrobertlloyd: Having removed that, this is needed to allow for right floated navigation items.

app/components/primary_navigation_component.html.erb
edwardhorsford: Sorry - I didn't see the parent in the code snippet.

edwardhorsford:
@paulrobertlloyd my bad, I was commenting on code you removed rather than added! 🤦

adamsilver (approved):

PR #4163, opened 2021-02-24
Add filter and validations to ensure course to be added is open on Apply

steventux:
## Context

It's possible to add a course to an application which is not open on Apply or visible in Find. This causes emails to be sent which won't make sense to some providers (eg. those yet to be onboarded).

- Filters `CourseOption`s which relate to courses not open on Apply.
- Validates that the course is open on Apply when attempting to save the `PickCourseForm`.

## Link to Trello card

https://trello.com/c/G63lZVXw/3425-prevent-support-users-from-adding-courses-to-an-application-that-are-either-hidden-in-find-or-that-are-not-for-the-current-cycle

## Things to check

### DIFF
--- a/app/forms/support_interface/application_forms/pick_course_form.rb
+++ b/app/forms/support_interface/application_forms/pick_course_form.rb
@@ -8,6 +8,7 @@ module SupportInterface
       validates :course_option_id, presence: true
       validates :application_form_id, presence: true
       validates :course_code, presence: true
+      validate :course_is_open_on_apply, on: :save
 
       RadioOption = Struct.new(:course_option_id, :course_name, :course_code, :site_name)
 
@@ -26,7 +27,7 @@ module SupportInterface
       end
 
       def save
-        return false unless valid?
+        return false unless valid?(:save)
 
         SupportInterface::AddCourseChoiceAfterSubmission.new(
           application_form: application_form,
@@ -53,12 +54,13 @@ module SupportInterface
       def courses
         Course
           .current_cycle
+          .open_on_apply
           .includes(course_options: [:site])
           .where(code: sanitize(course_code))
       end
 
       def sanitize(course_code)
-        course_code.strip.upcase
+        course_code&.strip&.upcase
       end
 
       def existing_course_ids
@@ -66,6 +68,12 @@ module SupportInterface
           .application_choices
           .map(&:course_option)
           .pluck(:course_id)
+      end
+
+      def course_is_open_on_apply
+        return if Course.open_on_apply.exists?(code: sanitize(course_code))
+
+        errors.add(:course_option_id, 'This course is not open on the Apply service')
       end
     end
   end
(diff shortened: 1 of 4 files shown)


app/forms/support_interface/application_forms/pick_course_form.rb
```diff
@@ -54,11 +56,11 @@ def courses
         Course
           .current_cycle
           .includes(course_options: [:site])
-          .where(code: sanitize(course_code))
+          .where(code: sanitize(course_code), open_on_apply: true, exposed_in_find: true)
```
cwrw, line 59: can we use the [scopes](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/models/course.rb#L13) `exposed_in_find` and `open_on_apply` instead?

app/forms/support_interface/application_forms/pick_course_form.rb
```diff
@@ -53,12 +55,14 @@ def course_option
       def courses
         Course
           .current_cycle
+          .open_on_apply
+          .exposed_in_find
```
duncanjbrown, line 59: `.open_on_apply` implies `exposed_in_find`: https://github.com/DFE-Digital/apply-for-teacher-training/blob/a9384c00d2c9da838d838dc0c32ef12c15aa611d/app/models/course.rb#L13

spec/forms/support_interface/application_forms/pick_course_form_spec.rb
```diff
@@ -62,5 +77,15 @@
   describe 'validations' do
     it { is_expected.to validate_presence_of(:course_option_id) }
     it { is_expected.to validate_presence_of(:application_form_id) }
+
+    it 'checks that the course is open on Apply' do
+      course_scope = class_double(Course)
```
cwrw, line 82: I was initially wondering if we can create a course here instead of using doubles, but then with testing found that running it as so without any courses still passes:
```
it 'checks that the course is open on Apply' do
    form = described_class.new(course_option_id: '123', course_code: 'ABC')
    expect(form.valid?(:save)).to be false
    expect(form.errors[:course_option_id]).to eq(['This course is not open on the Apply service'])
end
```

I was wondering if the intention was to also include non existing course ids being passed in with this error, or if we should handle that separately with a different error, let me know your thoughts

steventux: Ah yes, so we're not currently validating the existence of the course for the given `course_option_id` param in the update action presumably on the assumption that a support user is not going to craft data here.
If we think this validation is useful (I'm wondering if it is given it's the support UI) we could just make the validation error message more generic, or have 2 validations. either way it's very unlikely to ever be seen given we are excluding unopen courses from the choices that can be submitted here.

cwrw: Thats a good point, I think we can keep it as is. 👍 

My only comment now is around preferring the creation of the course rather than using doubles, as that is tying our test to the internal implementation, and I think creation of the course isn't very expensive, what do you think?

cwrw (approved):
LGTM 👍  just some minor comments around the spec

spec/forms/support_interface/application_forms/pick_course_form_spec.rb
steventux: I intentionally avoided creating a db record in the interests of speed, I'm happy to use `create` if that's preferred. We've typically tried to reduce db record creation for the sake of the overall speed of the test suite.

steventux: @cwrw I've removed the double as this wasn't necessary. The spec works fine using `FactoryBot#build` for the course model so we still save a trip to the db.

PR #4166, opened 2021-02-25
Expose docs for the TAD data export

tijmenb:
## Context

This is a page that describes the columns and values for the TAD data export. Because TAD will start consuming our API, they won't have access to the docs in the support section of the site anymore.

Add a new page, reusing the existing docs component.

## Guidance to review

Does it make sense? URL okay?

## Things to check

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -753,6 +753,7 @@ Rails.application.routes.draw do
   end
 
   namespace :data_api, path: '/data-api' do
+    get '/tad-data-exports/docs' => 'tad_docs#docs'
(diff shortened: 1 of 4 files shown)


config/routes.rb
```diff
@@ -753,6 +753,7 @@
   end
 
   namespace :data_api, path: '/data-api' do
+    get '/tad-data-exports/docs' => 'tad_docs#docs'
```
duncanjbrown, line 756: How about `docs/tad-data-export`?

duncanjbrown (approved):
Looks good, maybe tweak URL?

PR #4169, opened 2021-02-25
Refactor provider user permissions forms

benswannack:
## Context
There was a bug on the provider user invitation wizard whereby the error summary did not link to the correct place.
There is lots of duplication between the wizard there and in the single provider user permission editing form.

Move the edit provider user permission form to use `fields_for`
Move the wizard so it uses the same form html and ruby model
This commonises the erb and the validations

To test:
In the review app:
Invite a new provider user (Your account > Users > Invite User). Go through the flow and ensure checking different things has the desired effect.
Edit a provider user's permissions. Go to Your account> Users > click a user > click Change next to one of their permissions for a provider. Make sure that checking different boxes and submitting has the desired effect

## Guidance to review
Per commit
Are there any edge cases I've missed from the specs?
(Note that this should result in no change in behaviour apart from that the error is correctly linked to when selecting "Extra permissions" and then not checking any boxes)

## Link to Trello card
https://trello.com/c/EW62DVmA/3421-dac-change-no-errors-when-selecting-permissions

## Things to check


app/views/provider_interface/_fields_for_permissions.html.erb
```diff
@@ -1,100 +1,85 @@
-<%= form_with(
```
benswannack, line 1: This is a rather unhelpful diff... I've removed the component and moved the fields_for section out into a partial.

I would have liked to put the fields_for bit into another component, but this doesn't seem to work. (Just replacing the partial with a `<%= render MyComponent.new(...) %>` is what I've tried)

benswannack: Looks like the issue I ran into is a known one for view_component: https://github.com/github/view_component/issues/241

cwrw (commented):
I've had a quick look and some initial thoughts: the component issue and rendering is quite odd, I agree that this fits better into a component. Will have a closer look into this shortly.

I think the rest looks ok, and gets rid of the duplication quite nicely. Although I found it slightly confusing having another form for `FieldsForProviderUserPermissionsForm`, but not sure we can get around that since we need to use this in both the wizard and existing form 🤔

cwrw (approved):
I think the refactor looks really good and moving the shared code to a partial whilst validating against it in a separate form cleans it up nicely. I also tested it locally and all works as expected.

I only have one question which might not be an issue on production:
whilst testing and switching branches from master to the current branch I encountered the following error:
```
unknown attribute 'view_applications_only' for ProviderInterface::ProviderUserInvitationWizard.
```
This is because we removed the `view_applications_only` attribute from the wizard, but the last saved state for a user is restoring this value. Is this something we should worry about or do we usually really reset the states on deploy?

benswannack:
> This is because we removed the `view_applications_only` attribute from the wizard, but the last saved state for a user is restoring this value. Is this something we should worry about or do we usually reset the states on deploy?
That's a good spot, I don't think we do clear the caches on deploy.

I think the easiest way around that is to change the cache key to ensure the old caches are not loaded

PR #4171, opened 2021-02-25
Remove owners for "settings" feature flags

tijmenb:
Normal feature flags need owners because we need to know who's responsible for removing them. The settings don't need to have owners, because they'll never be removed.

### DIFF
--- a/app/services/feature_flag.rb
+++ b/app/services/feature_flag.rb
@@ -12,14 +12,14 @@ class FeatureFlag
   end
 
   PERMANENT_SETTINGS = [
-    [:banner_about_problems_with_dfe_sign_in, 'Displays a banner to notify users that DfE-Sign is having problems', 'Jake Benilov'],
-    [:banner_for_ucas_downtime, 'Displays a banner to notify users that UCAS is having problems', 'Theodor Vararu'],
-    [:dfe_sign_in_fallback, 'Use this when DfE Sign-in is down', 'Tijmen Brommet'],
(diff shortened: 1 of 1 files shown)


app/services/feature_flag.rb
```diff
@@ -12,20 +12,20 @@ def feature
...
+    [:pilot_open, 'Enables the Apply for Teacher Training service', 'Apply team'],
+    [:provider_information_banner, 'Displays an information banner for providers on the start page and applications page', 'Apply team'],
+    [:deadline_notices, 'Show candidates copy related to end of cycle deadlines', 'Apply team'],
   ].freeze
 
   TEMPORARY_FEATURE_FLAGS = [
     [:sync_from_public_teacher_training_api, 'Pull data from the public Teacher training API as well as the old "Find" API', 'Duncan Brown'],
     [:provider_activity_log, 'Show provider users a log of all application activity', 'Michael Nacos'],
     [:export_application_data, 'Providers can export a customised selection of application data', 'Ben Swannack'],
+    [:export_hesa_data, 'Providers can export applications including HESA data.', 'Steve Laing'],
```
cwrw, line 28: I think this was moved to permanent settings intentionally? https://github.com/DFE-Digital/apply-for-teacher-training/commit/c94b7ec0bf3d314a92df27b535c8dc3add2962ca

tijmenb: Good spot!

cwrw (approved):

PR #4174, opened 2021-02-25
Add Model to Reverse Map Structured Rejection Reasons

george-bit:
## Context

In order to facilitate the creation of new reports we need a reliable model to isolate all of the specific values in the JSON blob stored in the database.

Model and proto-report added

## Guidance to review

We've proceeded at this point to create a model that maps the first several R4Rs and sub-R4Rs. 

We have a functioning report and will continue to build on this but would like some feedback RE: our approach and any ways we could refactor as the model is already very big.

We did have the thought to perhaps have several smaller models instead of 1 large one but this could defeat the point of doing things in this way.

## Link to Trello card

https://trello.com/c/fqI5eAhg/3031-dev-create-schema-for-reverse-mapping-structured-reasons-for-rejection

## Things to check

### DIFF
--- a/app/models/flat_reasons_for_rejection_extract.rb
+++ b/app/models/flat_reasons_for_rejection_extract.rb
@@ -1,10 +1,53 @@
 class FlatReasonsForRejectionExtract
   include ActiveModel::Model
+  include ReasonsForRejection
 
-  def initialize(application_choice)
+  def initialize(structured_rejection_reasons)
     # @structured_rejection_reasons = application_form.application_choices.structured_rejection_reasons
-    @structured_rejection_reasons = application_choice
+    @structured_rejection_reasons = structured_rejection_reasons
   end
+
+  def separate_high_level_rejection_reasons(structured_rejection_reasons)
+    select_high_level_rejection_reasons(structured_rejection_reasons)
+    .keys
+    .map { |reason| format_reason(reason) }
+  end
+
+  def candidate_behaviour
+    return nil if @structured_rejection_reasons.blank?
+    @structured_rejection_reasons.select { |reason, value| value == 'candidate_behaviour_y_n' }
+
+    select_high_level_rejection_reasons(structured_rejection_reasons)
+    .keys
+    .map { |reason| format_reason(reason) }
+  end
+
+  def candidate_behaviour_sub_reasons
+
+
+  end
+
+  def quality_of_application
+
+
+  end
+
+  def quality_of_application_sub_reasons
+
+
+  end
+
+  def candidate_behaviour
+
+
+  end
+
+  def candidate_behaviour_sub_reasons
+
+
+  end
+
+
 
   # These three methods are copied from the application_choices_export.rb .... currently doesn't provide enough granularity
   def format_structured_rejection_reasons
--- a/app/models/flat_reasons_for_rejection_extract.rb
+++ b/app/models/flat_reasons_for_rejection_extract.rb
@@ -5,66 +5,70 @@ class FlatReasonsForRejectionExtract
     @structured_rejection_reasons = structured_rejection_reasons
   end
 
-  def top_level_reasons
-    I18n.t('reasons_for_rejection_copy.top_level_reasons').keys
-  end
-
-  def top_level_reason_title(reason)
-    I18n.t("reasons_for_rejection_copy.top_level_reasons.#{reason}.title")
-  end
-
-  def separate_high_level_rejection_reasons(structured_rejection_reasons)
-    select_high_level_rejection_reasons(structured_rejection_reasons)
-    .keys
-    .map { |reason| format_reason(reason) }
-  end
-
-  def candidate_behaviour
-    return nil if @structured_rejection_reasons.blank?
-    @structured_rejection_reasons.select { |reason, value| value == 'candidate_behaviour_y_n' }
-
-    select_high_level_rejection_reasons(structured_rejection_reasons)
-    .keys
-    .map { |reason| format_reason(reason) }
-  end
-
-  def qualifications
-
-  end
-
-  def quality_of_application
-
-
-  end
-
-  def quality_of_application_sub_reasons
-
-
-  end
-
   def candidate_behaviour?
-    return nil if @structured_rejection_reasons['candidate_behaviour_y_n'].blank?
-
-    @structured_rejection_reasons.select { |reason, value| value == 'Yes' && reason == 'candidate_behaviour_y_n' }.present?
+    top_level_reasons('candidate_behaviour_y_n')
   end
 
   def didnt_reply_to_interview_offer?
-    return nil if @structured_rejection_reasons['candidate_behaviour_what_did_the_candidate_do'].blank?
-
-    @structured_rejection_reasons['candidate_behaviour_what_did_the_candidate_do'].include?("didnt_reply_to_interview_offer")
+    sub_level_reasons('candidate_behaviour_what_did_the_candidate_do', 'didnt_reply_to_interview_offer')
   end
 
   def didnt_attend_interview?
-    return nil if @structured_rejection_reasons['candidate_behaviour_what_did_the_candidate_do'].blank?
-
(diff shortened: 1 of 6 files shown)


app/models/flat_reasons_for_rejection_extract.rb
```diff
@@ -0,0 +1,183 @@
+class FlatReasonsForRejectionExtract
```
stevehook, line 1: I wonder if this class could replace / merge with the existing `ReasonsForRejection` model?

app/models/flat_reasons_for_rejection_extract.rb
```diff
@@ -0,0 +1,183 @@
...
+
+    @structured_rejection_reasons["#{key}"].include?("#{value}")
+  end
+
+  def other_detail_or_what_to_improve(key)
+    return nil if @structured_rejection_reasons["#{key}"].blank?
+
+    @structured_rejection_reasons["#{key}"]
+  end
+end
```
despo, line 183: I think this belongs in a presenter and not a model, similar to the [RejectedApplicationChoicePresenter](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/presenters/rejected_application_choice_presenter.rb)

app/models/flat_reasons_for_rejection_extract.rb
```diff
@@ -0,0 +1,183 @@
+class FlatReasonsForRejectionExtract
+  include ActiveModel::Model
+
+  def initialize(structured_rejection_reasons)
+    @structured_rejection_reasons = structured_rejection_reasons
+  end
+
+  def candidate_behaviour?
```
stevehook, line 8: Not sure whether this type of method returns a `'Yes'` or `'No'` string or a bool. It kind of breaks convention if its a string.

stevehook (commented):
This is looking good. I've added a few inline mostly minor comments.

duncanjbrown:
@steventux do you have any thoughts on this?

app/models/flat_reasons_for_rejection_extract.rb
steventux: Echoing what @despo has said: 
Given this is a unpersisted/transient set of attributes on a model would we be better to turn this into an export-centric [Presenter](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/presenters/rejected_application_choice_presenter.rb) which delegates to the underlying [`ReasonsForRejection`](https://github.com/DFE-Digital/apply-for-teacher-training/blob/ebb0d63a0a32507206cf3a29ee30c7bdc66492e7/app/models/reasons_for_rejection.rb) model attributes?
It seems the main concern here is to extract/translate/present what we store in `ApplicationChoice#structured_reasons_for_rejection` but there's a lot of repetition of attributes which are already defined in `ReasonsForRejection` and presentation seems to be the goal here.
I don't fully understand the need for another model-like class as opposed to something which presents the data in a desireable format.

steventux (commented):
I'd agree with @despo that a presenter would be better suited for this export, we're close to repeating a lot of attributes already present in another transient model and maintanence of both attribute sets along with the existing presenter would also be a concern.

app/models/flat_reasons_for_rejection_extract.rb
stevehook: @steventux I think the intention here is to make something that's a bit more generic than what's needed just for the export in this PR because this kind of requirement has come up a few times. But a presenter makes sense to me too.

davidgisbey: I'll have a proper look at this over the weekend, but yea a presenter does seem to make sense.

davidgisbey: Yea my main thoughts is that the naming of this file is a bit misleading. as Steve said the purpose of this is that we use the nested data in the structure in multiple exports and throughout the dashboard. 

We need an easy way to replicate each individual checkbox that the provider selects. I think having an easy way to return nested values as a flat object would be pretty useful so we don't have to use stuff like `application_choice.structured_rejection_reasons.dig('qualifications_which_qualifications'], 'no_maths_gcse')` in multiple places in the export (for each attr).

I agree though that it is essentially taking the data we've got and presenting it in a different way though and could/should live in a presenter.

steventux: It's worth noting you can pass the data from the application choice into the `ReasonsForRejection` pseudo-model like this https://github.com/DFE-Digital/apply-for-teacher-training/blob/3dcd4625615b959488b7c5043453a32b1936d1ac/app/components/provider_interface/application_rejection_feedback_component.html.erb#L13 and then interrogate/present the attributes in a more OO way

app/components/support_interface/reasons_for_rejection_search_breadcrumb_component.rb
```diff
@@ -19,13 +19,13 @@ def items
         },
       ]
       if top_level_reason?
-        i18n_key = SupportInterface::SubReasonsForRejectionTableComponent::TOP_LEVEL_REASONS_TO_I18N_KEYS[@search_attribute].to_s
+        i18n_key = ReasonsForRejection::TOP_LEVEL_REASONS_TO_I18N_KEYS[@search_attribute].to_s
```
stevehook, line 22: 👍

app/presenters/flat_reasons_for_rejection_presenter.rb
```diff
@@ -0,0 +1,72 @@
...
+      label_for_subreason_detail('safeguarding_y_n', 'candidate_disclosed_information_details') => reasons_for_rejection.safeguarding_concerns_candidate_disclosed_information_details,
+      label_for_subreason('safeguarding_y_n', 'vetting_disclosed_information') => subreason_exists?(reasons_for_rejection.safeguarding_concerns, 'vetting_disclosed_information'),
+      label_for_subreason_detail('safeguarding_y_n', 'vetting_disclosed_information_details') => reasons_for_rejection.safeguarding_concerns_vetting_disclosed_information_details,
+      label_for_subreason_detail('safeguarding_y_n', 'other_details') => reasons_for_rejection.safeguarding_concerns_other_details,
+      title_for_top_level_reason('other_advice_or_feedback_y_n') => y_n_to_boolean(reasons_for_rejection.other_advice_or_feedback_y_n),
+      title_for_top_level_reason('interested_in_future_applications_y_n') => y_n_to_boolean(reasons_for_rejection.interested_in_future_applications_y_n),
+      'why are you rejecting this application details' => reasons_for_rejection.why_are_you_rejecting_this_application,
```
stevehook, line 42: This is a bit of a handful and would need work if we ever extend the model. I'd like to see a data declaration somewhere that allows this sort of thing to be generated, but that's for a follow-up PR I reckon.

stevehook (approved):
This looks good to me. I've got one totally optional suggestion inline. Would be good to get one more 👍 on this before merging.

steventux (approved):
It seems much more understandable now refactor to use a presenter. I think there's scope for reducing some of the repetition in places but it generally looks good to me.

config/locales/provider_interface/reasons_for_rejection.yml
```diff
@@ -7,12 +7,17 @@ en:
     quality_of_application:
       title: 'Quality of application'
       personal_statement: Personal statement
+      personal_statement_what_to_improve_details: Personal statement - what to improve
```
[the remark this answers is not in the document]
JR-G: Yeah, I was on the fence with this one myself.. in the end I opted to go for the additional translation in to an attempt to minimise hardcoded values in the presenter - should there be requests to change column headers to something less repetitive, once the export starts being used.

With the [comment](https://github.com/DFE-Digital/apply-for-teacher-training/pull/4174#discussion_r590297199) @stevehook left I think it's fair to say there will need to be a follow up PR, at a later point, to clean up the presenter and also revisit how we generate the key/values in the presenter cc @davidgisbey

PR #4178, opened 2021-02-26
Display offer withdrawal reasons in manage

benswannack:
## Context
Providers will find it useful to view reasons they withdrew offers in manage UI.

Display reasons for offer withdrawal in manage UI:

Screenshot 2021-02-26 at 12 10 23

## Guidance to review
Per commit

## Link to Trello card
https://trello.com/c/JBcKcdvY/3431-show-reasons-why-offer-was-withdrawn-in-the-ui

## Things to check

### DIFF
--- a/app/components/provider_interface/application_rejection_feedback_component.rb
+++ b/app/components/provider_interface/application_rejection_feedback_component.rb
@@ -1,0 +1,49 @@
+module ProviderInterface
+  class ApplicationRejectionFeedbackComponent < ViewComponent::Base
+    attr_reader :application_choice
+
+    def initialize(application_choice:)
+      @application_choice = application_choice
+    end
+
+    def render?
+      application_choice.rejected? && !application_choice.no_feedback?
+    end
+
+    def rejected_rows
+      if application_choice.rejected_by_default
+        rows = [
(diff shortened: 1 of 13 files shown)


app/components/provider_interface/application_rejection_feedback_component.rb
```diff
@@ -0,0 +1,49 @@
+module ProviderInterface
+  class ApplicationRejectionFeedbackComponent < ViewComponent::Base
+    attr_reader :application_choice
+
+    def initialize(application_choice:)
+      @application_choice = application_choice
+    end
+
+    def render?
+      application_choice.rejected?
```
AgaDufrat, line 10: I don't think it will be an issue because we have the redirect in place and a condition on displaying the 'Feedback' tab but an application rejected by default may not have any feedback for a while.

So something like the below would be more accurate.
```suggestion
      application_choice.rejected? && (application_choice.rejection_reason.present? || application_choice.structured_rejection_reasons.present?)
```
or `application_choice.rejected? && !application_choice.no_feedback?`

AgaDufrat (approved):
LGTM

PR #4180, opened 2021-02-26
Remove `ApplicationQualification.awarding_body`

duncanjbrown:
## Context

This field was added but never used. Nevertheless we added it to the Manage CSV export and the API.

`SELECT COUNT(DISTINCT awarding_body) from application_qualifications;` on Blazer in prod returns 0 results.

Remove the field and hardcode `null`s in those places 😔

## Link to Trello card

Following feedback from Oracle, who found this field populated in test data and got confused as to where it came from.

## Things to check

### DIFF
--- a/app/services/provider_interface/application_data_export.rb
+++ b/app/services/provider_interface/application_data_export.rb
@@ -50,7 +50,7 @@ module ProviderInterface
           'award_year' => application.first_degree.award_year,
           'institution_details' => application.first_degree.institution_name,
           'equivalency_details' => replace_smart_quotes(application.first_degree.composite_equivalency_details),
-          'awarding_body' => application.first_degree.awarding_body,
+          'awarding_body' => nil,
           'gcse_qualifications_summary' => replace_smart_quotes(application.gcse_qualifications_summary),
           'missing_gcses_explanation' => replace_smart_quotes(application.missing_gcses_explanation),
           'disability_disclosure' => application.application_form.disability_disclosure,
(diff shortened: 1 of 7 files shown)


db/migrate/20210226142457_remove_awarding_body_from_application_qualifications.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveAwardingBodyFromApplicationQualifications < ActiveRecord::Migration[6.0]
+  def change
+    remove_column :application_qualifications, :awarding_body, :string
```
benswannack, line 3: Including this here may result in some undefined method errors during deploy time. Can we split this into a subsequent PR?

benswannack (approved):

db/migrate/20210226142457_remove_awarding_body_from_application_qualifications.rb
duncanjbrown: Good point, I'll do

duncanjbrown:
Updated

app/services/provider_interface/application_data_export.rb
```diff
@@ -50,7 +50,7 @@ def self.call(application_choices:)
           'award_year' => application.first_degree.award_year,
           'institution_details' => application.first_degree.institution_name,
           'equivalency_details' => replace_smart_quotes(application.first_degree.composite_equivalency_details),
-          'awarding_body' => application.first_degree.awarding_body,
+          'awarding_body' => nil,
```
benswannack, line 53: Sorry, only just thought that this is potentially worth a comment explaining why this and the above look odd

PR #4181, opened 2021-02-26
Validate presence of candidate date of birth

despo:
Add missing date of birth presence check to personal details form.

## Context

Resolves sentry error: https://sentry.io/organizations/dfe-bat/issues/2239406103/?project=1765973&referrer=slack

- Add shared examples to verify Date of birth validations and presence (optionally)
- Remove tests duplicating DateValidator checks

## Guidance to review

Attempt to submit the personal details form without filling in the date of bith.

## Things to check

### DIFF
--- a/spec/forms/candidate_interface/personal_details_form_spec.rb
+++ b/spec/forms/candidate_interface/personal_details_form_spec.rb
@@ -54,30 +54,6 @@ RSpec.describe CandidateInterface::PersonalDetailsForm, type: :model do
     end
   end
 
-  describe '#date_of_birth' do
-    it 'return a nil for nil day/month/year' do
-      personal_details = CandidateInterface::PersonalDetailsForm.new(day: nil, month: nil, year: nil)
-
-      expect(personal_details.date_of_birth.day).to be_nil
-      expect(personal_details.date_of_birth.month).to be_nil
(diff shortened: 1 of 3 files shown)


spec/forms/candidate_interface/personal_details_form_spec.rb
```diff
@@ -86,57 +62,7 @@
...
-      it 'is valid if the candidate is older than 16' do
-        personal_details = CandidateInterface::PersonalDetailsForm.new(
-          day: '31', month: '12', year: '2002',
-        )
-
-        personal_details.validate
-
-        expect(personal_details.errors.full_messages_for(:date_of_birth)).to be_empty
-      end
+      include_examples 'date_of_birth validations', verify_presence: true
```
duncanjbrown, line 65: Merging this to fix the bug, but I think we should probably keep this inlined if it's only used in one place?

duncanjbrown (approved):

spec/forms/candidate_interface/personal_details_form_spec.rb
despo: @duncanjbrown the `date_of_birth` is also used in the edit forms so the plan is to reuse these validations there and tidy up the current spec!

PR #4185, opened 2021-03-01
Substitute smart quotes for straight quotes on free text fields

steventux:
## Context

Smart quotes entered in text fields are appearing as odd characters to some provider users when viewing exports.

Attempt to substitute any smart-quotes for straight quotes on text output in the applications export.
The issue was reported by a provider viewing the applications export, though I think this is OS/Application dependent as the same example works fine in LibreOffice, Vim and other text editors.

## Link to Trello card

https://trello.com/c/eXqfXMCt/3360-csv-export-shows-%C3%A2%E2%82%AC%C5%93-instead-of-in-free-text-fields-where-quotation-marks-are-used

## Things to check


benswannack (approved):
Looks good to me!

PR #4186, opened 2021-03-01
Delete redundant ViewComponent code

duncanjbrown:
## Context

We can simplify our `ViewComponent` code by

- inheriting from `SummaryListComponent`, which lets us delete a bunch of templates which only rendered `SummaryListComponent`s
- removing validations, which don't do anything

Do both those things!


despo (approved):
💯

PR #4191, opened 2021-03-01
Add a new make offer service which wraps the old one

despo:
## Context
We would like to refactor the legacy make an offer service. To avoid a total rewrite we would like to wrap the old service with a new one which will exercise the practices adopted in newer code.

## Link to Trello card
https://trello.com/c/1b6vJh6N/3432-add-a-new-make-offer-service-which-wraps-the-old-one

## Things to check

### DIFF
--- a/app/services/make_offer.rb
+++ b/app/services/make_offer.rb
@@ -18,7 +18,7 @@ class MakeOffer
                                     offer_conditions: conditions)
     make_an_offer.save
 
-    if make_an_offer.errors[:base].include?(I18n.t('activerecord.errors.models.application_choice.attributes.status.invalid_transition'))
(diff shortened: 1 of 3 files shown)


app/services/make_offer.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def save!
+    make_an_offer = MakeAnOffer.new(actor: actor,
+                                    application_choice: application_choice,
+                                    course_option: course_option,
+                                    offer_conditions: conditions)
+    make_an_offer.save
+
+    if make_an_offer.errors[:base].include?(MakeAnOffer::STATE_TRANSITION_ERROR)
+      ApplicationStateChange.new(application_choice).make_offer!
```
cwrw, line 22: 👍 for forcing an exception

cwrw (approved):

PR #4196, opened 2021-03-02
Add course summary box to make an offer page

cwrw:
## Context

When making an offer, a provider sees the radio buttons without seeing what the existing offer looks like. Adding a course summary above the select decision will allow them to see the current course details that a user has selected.

- Add a new course summary component which wraps the summary list component
- Surface the component on the make offer page
- Amend page title content and question legend on make offer page

## Guidance to review

- Compare the styles with the prototype https://manage-applications-beta.herokuapp.com/applications/P6RGOZC/decision
- Did I miss any common patterns with the proposed styles/classes in the component?

## Link to Trello card

https://trello.com/c/sHdPhNdl/3393-implement-course-details-box-on-make-offer-page-in-manage

## Things to check

### DIFF
--- a/app/components/provider_interface/course_summary_component.rb
+++ b/app/components/provider_interface/course_summary_component.rb
@@ -1,0 +1,35 @@
+module ProviderInterface
+  class CourseSummaryComponent < ViewComponent::Base
+    attr_reader :course_option, :provider_name, :course_name_and_code,
+                :location_name_and_address, :study_mode
+
+    def initialize(course_option:)
+      @course_option = course_option
+      @provider_name = course_option.provider.name
+      @course_name_and_code = course_option.course.name_and_code
+      @location_name_and_address = course_option.site.name_and_address
+      @study_mode = course_option.study_mode.humanize
+    end
+
+    def rows
+      [
+        {
+          key: 'Provider',
+          value: provider_name,
+        },
+        {
+          key: 'Course',
+          value: course_name_and_code,
+        },
+        {
+          key: 'Location',
+          value: location_name_and_address,
+        },
+        {
+          key: 'Full time or part time',
+          value: study_mode,
+        },
+      ]
+    end
+  end
+end
--- a/app/views/provider_interface/decisions/respond.html.erb
+++ b/app/views/provider_interface/decisions/respond.html.erb
@@ -3,6 +3,11 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l">
+      <span class="govuk-caption-l"><%= @application_choice.application_form.full_name %></span>
+      <%= t('page_titles.provider.respond') %>
+    </h1>
+    <%= render ProviderInterface::CourseSummaryComponent.new(course_option: @application_choice.course_option) %>
     <%= form_with model: @pick_response_form, url: provider_interface_application_choice_submit_response_path(@application_choice.id), method: :post do |f| %>
       <%- offer_text = 'Make an offer' %>
       <%- reject_text = 'Reject application' %>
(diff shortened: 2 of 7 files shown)


app/components/provider_interface/course_summary_component.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    def initialize(application_choice:)
+      @application_choice = application_choice
+      @provider_name = application_choice.provider.name
+      @course_name_and_code = application_choice.course.name_and_code
+      @location_name_and_address = application_choice.course_option.site.name_and_address
+      @study_mode = application_choice.course_option.study_mode.humanize
+    end
+
+    def course_summary_rows
```
despo, line 14: Consider renaming this to rows. As the component name indicates its purpose I think we can simplify this method name.

app/components/provider_interface/course_summary_component.rb
```diff
@@ -0,0 +1,35 @@
+module ProviderInterface
+  class CourseSummaryComponent < ViewComponent::Base
+    attr_reader :application_choice, :provider_name, :course_name_and_code,
+                :location_name_and_address, :study_mode
+
+    def initialize(application_choice:)
```
despo, line 6: Can we make this more specific and use course instead? `application_choice.provider` is retrieved through the course model anyway, similar to course_option.

app/views/provider_interface/decisions/respond.html.erb
```diff
@@ -3,15 +3,19 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l">
+      <span class="govuk-caption-l"><%= @application_choice.application_form.full_name %></span>
+      <%= t('page_titles.provider.respond') %>
+    </h1>
+    <%= render ProviderInterface::CourseSummaryComponent.new(application_choice: @application_choice) %>
```
despo, line 10: Just a question here, shouldn't we wait until the updated make offer is in place or use the new feature flag to control when the component is displayed?

despo (approved):
@cwrw Thank you for your PR! I left some comments inline.
Can you also add a preview for this component please!

app/views/provider_interface/decisions/respond.html.erb
cwrw: On the ticket it says: 
> It should be possible to implement this without waiting for the new Make/Change Offer flow
I also double checked with James for posterity and he confirmed its ok to show the box without a feature flag

app/components/provider_interface/course_summary_component.rb
cwrw: I was attempting to pass in the course instead, but realised because we are surfacing the study mode we do need the course option as using the study mode on the course itself won't give us the definitive result we need (it come back as both full time or part time if those exist on the course options). This means we will need to pass the course option itself here, but it feels a bit disjoined without the application choice and might be better to go with the application choice, let me know your thoughts around keeping it as is or using the `course_option` instead.

despo: I would still go for the object closer to matching all dependencies, `course_option` as you correctly mentioned rather than `course`. I always get confused about which level they live on, course and course_option! 
Which makes the component reusable for the offered course (or rather offered_course_option) as well.

spec/components/previews/provider_interface/course_summary_component_preview.rb
```diff
@@ -0,0 +1,7 @@
+module ProviderInterface
+  class CourseSummaryComponentPreview < ViewComponent::Preview
+    def course_details
+      render CourseSummaryComponent.new(course_option: CourseOption.all.sample)
```
despo, line 4: I did not realise component previews were connecting to a prepopulated database! Is it worth limiting this? e.g. `CourseOption.limit(10).sample` ?

despo (approved):
Looks great, thank you for addressing all my feedback.

I left a comment however I'm happy for it to be merged with or without that change :shipit:

PR #4197, opened 2021-03-02
Update restructured work history error messaging

malcolmbaig:
## Context
The error messages in the work history need to be updated to match the new design.

- Most of the changes described by the Google doc linked in the ticket
- Does not include error messages specifically for the month or specifically for the year, as agreed with @gregknight1 (reason—not just a copy change, needs some non-trivial changes to how we handle date validations).
- Some spec setup changes, mostly related to locale strings and the shared examples we're using for dates.

## Guidance to review
- Running locally or in the review app, ensure restructured work history feature flag is active, and the current application has `feature_restructured_work_history` set to true. Easiest way to ensure the latter is to sign up as a new user (rather than log in as an existing one).
- Go through the Add Job and Explain Break flows, testing the scenarios outlined in the doc (except the ones where a month-only or year-only error is expected).
- Regarding the missing locale string mentioned in one of the commits - I've written up this card to deal with similar cases https://trello.com/c/ouZ2DExb

## Link to Trello card

https://trello.com/c/pArii32e/3053-error-messaging-in-the-work-history-section
## Things to check

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -415,6 +415,7 @@ en:
       blank_date: Enter %{article} %{attribute}
       dob_future: Enter a %{attribute} that is in the past, for example 31 3 1980
       dob_below_min_age: Enter a date of birth before %{date} – you must be over %{min_age} years old to Apply for teacher training
+      before: Enter a start date that is before the end date
     application_choices:
       course_not_available: You cannot apply to ‘%{descriptor}’ because it is not running
       course_full: You cannot apply to ‘%{descriptor}’ because it has no vacancies
--- a/spec/forms/candidate_interface/restructured_work_history/job_form_spec.rb
+++ b/spec/forms/candidate_interface/restructured_work_history/job_form_spec.rb
@@ -39,7 +39,7 @@ RSpec.describe CandidateInterface::RestructuredWorkHistory::JobForm, type: :mode
     it { is_expected.to validate_length_of(:role).is_at_most(60) }
     it { is_expected.to validate_length_of(:organisation).is_at_most(60) }
 
-    include_examples 'validation for a start date', 'restructured_work_history/job_form', verify_presence: true
+    include_examples 'validation for a start date', true, 'restructured_work_history/job_form'
 
     context "'currently working' is 'false'" do
       it 'is invalid if left completely blank' do
--- a/spec/forms/candidate_interface/restructured_work_history/job_form_spec.rb
+++ b/spec/forms/candidate_interface/restructured_work_history/job_form_spec.rb
@@ -47,7 +47,7 @@ RSpec.describe CandidateInterface::RestructuredWorkHistory::JobForm, type: :mode
 
         form.validate
 
-        expect(form.errors[:end_date]).to contain_exactly('Enter an end date')
+        expect(form.errors[:end_date]).to contain_exactly('Enter the month and year you left this job')
       end
 
       it 'is invalid if month left blank' do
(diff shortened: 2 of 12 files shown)


malcolmbaig:
Unit tests need a rethink, closing for now.

config/locales/en.yml
```diff
@@ -415,6 +415,7 @@ en:
       blank_date: Enter %{article} %{attribute}
       dob_future: Enter a %{attribute} that is in the past, for example 31 3 1980
       dob_below_min_age: Enter a date of birth before %{date} – you must be over %{min_age} years old to Apply for teacher training
+      before: Enter a start date that is before the end date
```
despo, line 418: This is quite a specific validation to the start and end date fields so if it is to be moved in the shared validation errors I think the name should be more descriptive, e.g. `:start_date_before_end_date`

spec/forms/candidate_interface/restructured_work_history/job_form_spec.rb
```diff
@@ -39,15 +39,15 @@
     it { is_expected.to validate_length_of(:role).is_at_most(60) }
     it { is_expected.to validate_length_of(:organisation).is_at_most(60) }
 
-    include_examples 'validation for a start date', 'restructured_work_history/job_form', verify_presence: true
```
despo, line 42: why is ths presence check skipped - as it's not reflected on the form, so I expect that we still want to verify it through the specs.

despo:
@malcolmbaig this is clashing with some work I did last night to create some reusable shared examples that enable easier and more consistent date validation. Can we somehow fit both together? I think it will also remove the need for having to have conditions specific to certain forms in the shared validations, as it checks for the locale key rather than the exact error message.

https://github.com/DFE-Digital/apply-for-teacher-training/pull/4207

I also couldn't identify how it skips the month and year specific error, as I was expecting `blank_date_fields` to be overriden in the form specific locales.

config/locales/en.yml
malcolmbaig: Yep, makes sense. Will update.

spec/forms/candidate_interface/restructured_work_history/job_form_spec.rb
malcolmbaig: Removed as part of https://github.com/DFE-Digital/apply-for-teacher-training/pull/4197/commits/bea0b6a5bc5c60f1c335e029f946d2644cfba0ab

Rationale - we don't (AFAICT) have any test cases where we don't verify presence, making an option redundant. Can drop this commit altogether if we want to retain the option.

malcolmbaig:
> Can we somehow fit both together?

Sounds good, reviewing now.

> I also couldn't identify how it skips the month and year specific error, as I was expecting blank_date_fields to be overriden in the form specific locales.

Not sure I follow, but restructured work history forms are falling back on `errors.messages.blank_date_fields`.

despo:
> > Can we somehow fit both together?
> 
> Sounds good, reviewing now.
> 
> > I also couldn't identify how it skips the month and year specific error, as I was expecting blank_date_fields to be overriden in the form specific locales.
> 
> Not sure I follow, but restructured work history forms are falling back on `errors.messages.blank_date_fields`.

In the description of the PR there is a note about not including error messages specifically for the month, or specifically for the year (as quoted below).  I assume that means that you don't want to display the default `invalid_date_month_and_year` error which  indicates if the month or year are missing, but instead replace it with a more generic one? Unless I have mirsunderstood the description!

> Does not include error messages specifically for the month or specifically for the year, as agreed with @gregknight1 (reason—not just a copy change, needs some non-trivial changes to how we handle date validations).

malcolmbaig:
Closing to rebase on https://github.com/DFE-Digital/apply-for-teacher-training/pull/4207

PR #4202, opened 2021-03-02
Fix flaky spec

benswannack:
Sometimes too many withdrawn application choices are created, which takes the application count lower than the threshold in the component


cwrw (approved):
👍

PR #4204, opened 2021-03-02
(don't merge) Spike on DfEComponent::PrimaryNavigation

duncanjbrown:
## Context

Following @paulrobertlloyd's excellent [Google Doc](https://docs.google.com/document/d/1RbT11P_75TD2BV27VZKGfBb1tbbaUg8HdXnj-qUrhuY/edit) on component maturity and proposals for how we might share components across BAT services, this branch experiments with pulling the `PrimaryNavigation` component out into a standalone Rails component folder that could possibly be copy-pasted into another project.

New folder structure:

```
components/dfe_component
├── _all.scss
└── primary_navigation
    ├── _component.scss
    ├── component.html.erb
    └── component.rb
```

## Some notes

1. Our component depends on an array of `NavigationItem` structs. Register would need to make some changes to support that interface. They do _almost_ reflect the [MoJ nunjucks hash](https://moj-design-system.herokuapp.com/components/primary-navigation) - they just have an extra `classes` attr. Maybe it would be better to use the `hash[:style]` way of addressing structs inside the component so we support hashes too?
2. The responsibilities of the component also differ from app to app. The [Register version of this component](https://github.com/DFE-Digital/register-trainee-teachers/blob/master/app/components/navigation_bar/view.rb) has some business logic in about signed in users etc, whereas Apply (which uses it in various contexts) [prefers to pass in a decided array of `NavigationItems`](https://github.com/DFE-Digital/apply-for-teacher-training/blob/7ed5a105a00569d72589db2cd21e68a7d451a3f6/app/helpers/navigation_items.rb#L19-L31).
3. Supporting this folder structure in the app is pretty simple — a change to `webpacker.yml` and the proposed `dfe_component/_all` SCSS file to load everything from the `dfe_components` namespace.
4. I didn't add docs or anything but in due course we could copy a bit from [govuk-components](https://github.com/DFE-Digital/govuk-components/blob/master/spec/dummy/app/views/demos/show.html.erb)?


app/frontend/styles/application.scss
```diff
@@ -36,6 +36,8 @@ $moj-images-path: "~@ministryofjustice/frontend/moj/assets/images/";
 @import "_tag";
 @import "_back_link";
 
+@import "dfe_component/_all";
+
 // Shared (custom components)
```
paulrobertlloyd, line 41: Might want to have some additional commenting to describe the differences between BAT-wide and application-wide styles

```suggestion
// Shared (common components replicated across BAT services)
@import "dfe_component/_all";

// Shared (components used by all application interfaces)
```
```suggestion
@import "dfe_component/_all";

// Shared (custom components)
```

paulrobertlloyd:
I like the Register version of this component, in that it moves the logic about the active navigation link into the class, and out of the template file. `:items` and `:current_path` seem like reasonable params; having `:items` accept optional params seems like a good shout (i.e. classes, other HTML attributes). The way the form builder handles this stuff is a good model I think.

AbigailMcP:
Hi! 👋 

> 2. The responsibilities of the component also differ from app to app. The Register version of this component has some business logic in about signed in users etc, whereas Apply (which uses it in various contexts) prefers to pass in a decided array of NavigationItems.

I prefer Apply's approach here, taking the business logic surrounding `current_user` out of the component itself - actually I recently refactored a similar component (header) to remove this type of logic and accept a simple array of pre-determined `items`.

But re @paulrobertlloyd 's comment, I think it's neat that the component retains the 'active link' logic.

duncanjbrown:
Thanks for the feedback! I hear you both. At the moment the logic for whether menu item links are active, irrespective of what kind of menu they're in, is in https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/helpers/navigation_items.rb. To push that down into the component we'd have to duplicate the is-it-active logic into both kinds of menu (the header and the primary nav) and the component would have to have knowledge of the controller etc. I think I prefer the [design system approach](https://design-system.service.gov.uk/components/header/) where we pass in a true or false, and the component is pretty dumb?

paulrobertlloyd:
> I think I prefer the design system approach where we pass in a true or false, and the component is pretty dumb?

I agree. The component should know how to treat an active link, but not know why something is deemed active.

duncanjbrown:
OK I'm going to close this and leave as a museum piece in the Slack channel. We can reconvene the common components gang to review findings

duncanjbrown:
ugh it's not a Slack channel, it's a private chat. Going to post in #twd_bat_design

PR #4207, opened 2021-03-03
Add shared examples for month year validations

despo:
## Context

A new set of configurable and reusable shared examples that can be used to verify both the start and end date validations.

- [x] Remove `valid_date_or_nil` and handle empty end_date fields
- [x] Check that validations relying on a start_date only take place when that is set
- [x] Replace and remove existing start_date and end_date shared examples
- [x] ~Move combination checks (e.g. start_date being before end_date - :before error) in specs~
- [x] Use locale keys to verify errror so examples work for both models using the default error messages and models using custom error messages e.g `errors.messages.invalid_date` vs `activemodel.errors.models.candidate_interface/form_model.attributes.start_date.invalid_date`
- [x] Extend DateValidator to include `:before` validation e.g. `start_date: date: { before: :end_date}` and add shared example

## Things to check


app/helpers/date_validation_helper.rb
```diff
@@ -22,4 +13,10 @@ def start_date_before_end_date
 
     errors.add(:start_date, :before) unless start_date <= end_date
   end
```
malcolmbaig, line 15: General observation, unrelated to this PR - `start_date_before_end_date` doesn't seem like it should live here. It's an actual validation in its own right rather than just a helper. Probably worth co-locating with the logic in DateValidator.

spec/forms/candidate_interface/restructured_work_history/work_history_break_form_spec.rb
```diff
@@ -28,8 +28,53 @@
...
+      include_examples 'month and year date validations', :start_date, verify_presence: true, future: true
+
+      describe 'when after the end date' do
+        it 'returns a :before error' do
+          model = described_class.new(start_date_month: '5', start_date_year: '2018',
+                                      end_date_month: '5', end_date_year: '2017')
+
+          expect(model).to be_invalid
+
+          expect(model.errors.added?(:start_date, :before)).to eq(true)
```
malcolmbaig, line 47: The `before` validation seems fairly generic - any way to add this to the shared examples? Could maybe have a spec that only runs if validating :start_date and both date fields have values. Otherwise we have to remember to write the inline spec if we want to test this behaviour. Case in point - job_form_spec above isn't currently testing this validation.

spec/support/shared_examples/month_and_year_date_validations.rb
```diff
@@ -0,0 +1,59 @@
...
+  let("#{date_field}_month".to_sym) { date.month }
+  let("#{date_field}_year".to_sym) { date.year }
+
+  describe 'when in the future', if: validations[:future] do
+    let(:date) { 40.years.from_now }
+
+    it 'returns :future error' do
+      expect(model).to be_invalid
+
+      expect(model.errors.added?(date_field,
```
malcolmbaig, line 14: Nice 👌🏼 . Once https://trello.com/c/ouZ2DExb is in we'll pick up missing keys too, giving us pretty decent test coverage without the verbosity.

spec/forms/candidate_interface/restructured_work_history/work_history_break_form_spec.rb
despo: Good point. Let me look more into this, as I think that would be ideal.

app/helpers/date_validation_helper.rb
```diff
@@ -8,12 +8,6 @@ def valid_or_invalid_date(year, month)
     end
   end
 
-  def start_date_before_end_date
-    return unless start_date.is_a?(Date) && end_date.is_a?(Date)
-
-    errors.add(:start_date, :before) unless start_date <= end_date
-  end
-
```
malcolmbaig: 👍🏼

malcolmbaig (approved):
LGTM 👍🏼

PR #4208, opened 2021-03-03
Add shared examples for year validations

despo:
## Context

Add shared examples for year validations to remove duplication and enable consistent testing of fields using the YearValidator.

## Things to check

### DIFF
--- a/spec/support/shared_examples/year_validations.rb
+++ b/spec/support/shared_examples/year_validations.rb
@@ -1,0 +1,37 @@
+RSpec.shared_examples 'year validations' do |year_field, validations|
+  include DateAndYearConcerns
+
+  let(year_field.to_sym) { year }
+
+  describe 'when invalid' do
+    context 'when year is not a number' do
+      let(:year) { 'A950' }
+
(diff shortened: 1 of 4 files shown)


spec/support/shared_examples/year_validations.rb
```diff
@@ -0,0 +1,37 @@
+RSpec.shared_examples 'year validations' do |year_field, validations|
+  include DateAndYearConcerns
+
+  let(year_field.to_sym) { year }
+
+  describe 'when invalid' do
```
cwrw, line 6: Might be superfluous, but maybe we should add a when valid describe block as well as some of those are custom validations

cwrw (approved):
Looks good, just the one comment

spec/support/shared_examples/year_validations.rb
despo: Good idae, but maybe in the year validator specs rather than the shared examples, as then I can easily verify the validity of the object without relying on additional data setup.

cwrw (approved):

PR #4210, opened 2021-03-03
Support: Increase container width

paulrobertlloyd:
## Context

Many pages in Support are information dense (tables with filters etc). We can improve things for users of desktop computers slightly by increasing the maximum width of the layout container.

Bit of a blunt instrument this, but adds a class to the support layout template (`app-template--support`), and this modifies any child `govuk-width-container` classes to have the maximum width of 1100px. This matches the maximum width of that used on the [design system](https://design-system.service.gov.uk) site.

## Guidance to review

This doesn’t effect many pages, but there might be cases where we want to reduce the overall width of components.

## Things to check

### DIFF
--- a/app/frontend/styles/support/_all.scss
+++ b/app/frontend/styles/support/_all.scss
@@ -12,3 +12,4 @@
 @import "_details";
 @import "_sort";
 @import "_syntax-highlighting";
+@import "_template";
--- a/app/frontend/styles/support/_template.scss
+++ b/app/frontend/styles/support/_template.scss
@@ -1,0 +1,5 @@
+.app-template--support {
+  .govuk-width-container {
+    @include govuk-width-container(1100px);
+  }
+}
--- a/app/views/layouts/support_layout.html.erb
+++ b/app/views/layouts/support_layout.html.erb
@@ -1,3 +1,5 @@
+<% content_for :body_class, 'app-template--support' %>
+
 <% content_for :content do %>
   <% if yield(:context).present? %>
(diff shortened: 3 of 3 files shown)


app/frontend/styles/support/_all.scss
```diff
@@ -11,4 +11,4 @@
 @import "_checkboxes-scroll";
 @import "_details";
 @import "_sort";
-@import "_syntax-highlighting";
```
paulrobertlloyd, line 14: `_syntax-highlighting.scss` no longer exists in this location, and is no longer used by support anyway (I think).

fofr:
Do you have a before and after screenshot?

app/frontend/styles/support/_all.scss
duncanjbrown: It is — for the API responses!

duncanjbrown:
## Before
Screenshot 2021-03-03 at 13 59 57

## After
Screenshot 2021-03-03 at 13 59 46

app/frontend/styles/support/_all.scss
paulrobertlloyd: Ah, well then that’s a bug – won’t work as that file doesn’t exist in this folder!

paulrobertlloyd: Oh, actually, scratch that – I was wrong. Was relying to much on VS Code’s highlighting…

duncanjbrown (approved):

PR #4214, opened 2021-03-04
Remove interview info from dashboard when cancelled

TobyRet:
## Context

When an interview is cancelled there isn't any indication of this on the dashboard, as we're using the current dashboard design at the moment (but they do get an email). The new dashboard design will address this but in the meantime displaying the cancelled interview details with no indication that it's been cancelled is very confusing,

- Remove interview info from dashboard when an interview is cancelled
- Create a new FactoryBot trait to make it easier to create an  application choice with a cancelled interview.  

## Guidance to review

- We use the [discard](https://github.com/jhawthorn/discard) gem to soft delete interviews when they are cancelled. When `#kept` is called on the interviews it should only retrieve the active interviews

## Link to Trello card

https://trello.com/c/2OKsz36L/3123-remove-interview-from-dashboard-for-cancelled-interviews

## Things to check

### DIFF
--- a/app/components/candidate_interface/course_choices_review_component.rb
+++ b/app/components/candidate_interface/course_choices_review_component.rb
@@ -151,7 +151,7 @@ module CandidateInterface
     end
 
     def interview_row(application_choice)
-      if application_choice.interviews.present?
+      if application_choice.interviews.kept.any?
         {
           key: 'Interview'.pluralize(application_choice.interviews.size),
           value: render(InterviewBookingsComponent.new(application_choice)),
--- a/spec/components/candidate_interface/course_choices_review_component_spec.rb
+++ b/spec/components/candidate_interface/course_choices_review_component_spec.rb
@@ -353,6 +353,28 @@ RSpec.describe CandidateInterface::CourseChoicesReviewComponent do
     end
   end
 
+  context 'when an interview has been scheduled' do
+    it 'renders the component with interview details' do
+      application_choice = create(:application_choice, :with_completed_application_form, :with_scheduled_interview)
+      application_form = application_choice.application_form
+
+      result = render_inline(described_class.new(application_form: application_form, editable: false, show_status: true))
+
+      expect(result.css('.govuk-summary-list__key').text).to include('Interview')
+    end
+  end
+
+  context 'when an interview has been cancelled' do
+    it 'renders the component without interview details' do
+      application_choice = create(:application_choice, :with_completed_application_form, :with_cancelled_interview)
+      application_form = application_choice.application_form
+
+      result = render_inline(described_class.new(application_form: application_form, editable: false, show_status: true))
+
+      expect(result.css('.govuk-summary-list__key').text).not_to include('Interview')
+    end
+  end
+
   def create_application_form_with_course_choices(statuses:)
     application_form = create(:application_form)
 
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -546,6 +546,16 @@ FactoryBot.define do
       end
     end
 
+    trait :with_cancelled_interview do
+      awaiting_provider_decision
+
+      after(:build) do |application_choice, _evaluator|
+        application_choice.status = :awaiting_provider_decision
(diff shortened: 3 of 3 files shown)


app/components/candidate_interface/course_choices_review_component.rb
```diff
@@ -151,7 +151,7 @@ def start_date_row(application_choice)
     end
 
     def interview_row(application_choice)
-      if application_choice.interviews.present?
+      if application_choice.interviews.kept.any?
```
stevehook, line 154: :+1: `any?` is better (faster) than `present?`. Apparently `exists?` is even better but probably won't make a measurable difference here because interviews is a small result set.

spec/factories.rb
```diff
@@ -546,6 +546,16 @@
       end
     end
 
+    trait :with_cancelled_interview do
+      awaiting_provider_decision
+
+      after(:build) do |application_choice, _evaluator|
+        application_choice.status = :awaiting_provider_decision
```
stevehook, line 553: is this line needed? Isn't it covered by line 550 above?

stevehook (approved):
This makes sense to me, one tiny nit.

spec/factories.rb
despo: Line 550 will create a new application choice with that state. But if one is passed in to the factory, this line should ensure that its status is updated to `awaiting_provider_decision`

spec/components/candidate_interface/course_choices_review_component_spec.rb
```diff
@@ -353,6 +353,28 @@
...
+      result = render_inline(described_class.new(application_form: application_form, editable: false, show_status: true))
+
+      expect(result.css('.govuk-summary-list__key').text).to include('Interview')
+    end
+  end
+
+  context 'when an interview has been cancelled' do
+    it 'renders the component without interview details' do
+      application_choice = create(:application_choice, :with_completed_application_form, :with_cancelled_interview)
+      application_form = application_choice.application_form
```
despo, line 370:
```suggestion
    let(:application_choice) { create(:application_choice, :with_completed_application_form, :with_cancelled_interview) }
    let(:application_form) { application_choice.application_form }
    
    it 'renders the component without interview details' do    
```

spec/components/candidate_interface/course_choices_review_component_spec.rb
```diff
@@ -353,6 +353,28 @@
     end
   end
 
+  context 'when an interview has been scheduled' do
+    it 'renders the component with interview details' do
+      application_choice = create(:application_choice, :with_completed_application_form, :with_scheduled_interview)
+      application_form = application_choice.application_form
```
despo, line 359:
```suggestion
    let(:application_choice) { create(:application_choice, :with_completed_application_form, :with_scheduled_interview) }
    let(:application_form) { application_choice.application_form }
    
    it 'renders the component without interview details' do    
```

despo (approved):
Left some minor feedback but I'm happy to approve!

💯

spec/components/candidate_interface/course_choices_review_component_spec.rb
stevehook: @despo we've followed the 'let's not' approach on the project (https://thoughtbot.com/blog/lets-not). I must admit I've used `let` everywhere on all the other projects I've worked on so I'm happy either way - but for consistency we should probably stick with the explicit local variables. What do you think?

despo: I have not seen this article before. However I'm not sure I agree with it. I find that in a lot of our specs it's very hard to understand what's going on because of the amount of setup that is included in them.

I understand how it can be an issue if all the data setup happens at the beginning of the spec, but doing that within the context block means that they are easy to see, understand and keep up to date if the spec changes in the future. Also, I'm not sure I get this point `It causes Slow Tests by creating more data than is necessary in each test.` as anything defined with let is not evaluated until the first time it's called.

PR #4215, opened 2021-03-04
Remove unused views

despo:
The check icon was used by one of the provider permission partials, but this is not the case anymore, and the `new_edit_response` view was mapped to an action on the `DecisionController` who was at some point removed.


mnacos (approved):
Great

PR #4216, opened 2021-03-04
SPIKE: Find changed apply again applications

stevehook:
## Context

We need to work out a way to find the apply again applications that the candidate changed (in a significant way) since the original application.

This PR is a spike to explore ways to work this out.

- [x] Add a `FindChangedApplyAgainApplications` query service class.

## Guidance to review

- This is a fairly straightforward (naive) implementation that relies on a Ruby `select` to filter the list of apply again applications. I don't think it would be easy to do this in a database query so some of the logic is happening in memory. This may be fine given that the numbers of apply again applications isn't huge. Using `lazy` enumerator may help.
- Is there an alternative approach? I didn't see how we could use the API to serialise and then diff the two results because we are only interested in a narrow subset of changes.
- The set of changes we regard as significant is very much hard-coded into the filter code here. Is there a way/need to generalise it a bit?

## Link to Trello card

https://trello.com/c/0fMYPksT/2862-dev-use-govuk-view-components-in-find-to-replace-accordion-markup

## Things to check

### DIFF
--- a/app/services/candidate_interface/find_changed_apply_again_applications.rb
+++ b/app/services/candidate_interface/find_changed_apply_again_applications.rb
@@ -1,9 +1,21 @@
 module CandidateInterface
   class FindChangedApplyAgainApplications
-    def call
+    def all_forms
+      apply_again_forms
+    end
+
+    def changed_forms
       apply_again_forms.find_each.lazy.select do |application_form|
         changed?(application_form)
       end
+    end
+
+    def all_candidate_count
+      all_forms.map(&:candidate_id).uniq.count
+    end
+
+    def changed_candidate_count
+      changed_forms.map(&:candidate_id).uniq.count
     end
 
   private
--- a/app/services/candidate_interface/find_changed_apply_again_applications.rb
+++ b/app/services/candidate_interface/find_changed_apply_again_applications.rb
@@ -14,7 +14,7 @@ module CandidateInterface
           phase: 'apply_2',
           recruitment_cycle_year: RecruitmentCycle.current_year,
         )
-        .includes(previous_application_form: [:application_qualifications], application_qualifications: [])
+        .includes(:application_qualifications, previous_application_form: [:application_qualifications])
     end
 
     def changed?(application_form)
@@ -43,11 +43,9 @@ module CandidateInterface
       new_qualifications.all? { |new_qualification| !qualifications_match?(original_qualification, new_qualification) }
     end
 
-    IGNORED_QUALIFICATION_ATTRIBUTES = DuplicateApplication::IGNORED_ATTRIBUTES + %i[application_form_id public_id]
-
     def qualifications_match?(original_qualification, new_qualifications)
-      original_attributes = original_qualification.attributes.reject { |k, _| IGNORED_QUALIFICATION_ATTRIBUTES.include?(k) }
-      new_attributes = new_qualifications.attributes.reject { |k, _| IGNORED_QUALIFICATION_ATTRIBUTES.include?(k) }
+      original_attributes = original_qualification.attributes.reject { |k, _| DuplicateApplication::IGNORED_CHILD_ATTRIBUTES.include?(k) }
+      new_attributes = new_qualifications.attributes.reject { |k, _| DuplicateApplication::IGNORED_CHILD_ATTRIBUTES.include?(k) }
       original_attributes == new_attributes
     end
   end
--- a/app/services/candidate_interface/find_changed_apply_again_applications.rb
+++ b/app/services/candidate_interface/find_changed_apply_again_applications.rb
@@ -1,0 +1,54 @@
+module CandidateInterface
+  class FindChangedApplyAgainApplications
+    def call
+      apply_again_forms.find_each.lazy.select do |application_form|
+        changed?(application_form)
+      end
+    end
+
+  private
+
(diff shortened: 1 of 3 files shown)


davidgisbey:
This looks like a really sound approach to me @stevehook. Might be worth someone else having a look, but i'm pretty confident that this is solid and the way to go 👌

app/services/candidate_interface/find_changed_apply_again_applications.rb
```diff
@@ -0,0 +1,54 @@
...
+      return true if original_qualifications.size != new_qualifications.size
+
+      original_qualifications.any? { |qualification| any_qualification_differ?(qualification, new_qualifications) }
+    end
+
+    def any_qualification_differ?(original_qualification, new_qualifications)
+      new_qualifications.all? { |new_qualification| !qualifications_match?(original_qualification, new_qualification) }
+    end
+
+    IGNORED_QUALIFICATION_ATTRIBUTES = DuplicateApplication::IGNORED_ATTRIBUTES + %i[application_form_id public_id]
```
malcolmbaig, line 46: Do we need `id` here as well, to account for identical quals with a different primary key?

app/services/candidate_interface/find_changed_apply_again_applications.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  private
+
+    def apply_again_forms
+      ApplicationForm
+        .where(
+          phase: 'apply_2',
+          recruitment_cycle_year: RecruitmentCycle.current_year,
+        )
+        .includes(previous_application_form: [:application_qualifications], application_qualifications: [])
```
malcolmbaig, line 17: Nit, but could this alternatively be:

`.includes(:application_qualifications, previous_application_form: [:application_qualifications])`

app/services/candidate_interface/find_changed_apply_again_applications.rb
stevehook: `id` is already in the `DuplicateApplication::IGNORED_ATTRIBUTES`. Perhaps I should just add `application_form_id` and `public_id` to that...

malcolmbaig:
Overall approach looks good to me 👍🏼 . Couple of comments left inline.

> The set of changes we regard as significant is very much hard-coded into the filter code here. Is there a way/need to generalise it a bit?

I think this is fine, as long as it's clear to consumers of this data what the precise definition of "changed" is. If we needed to generalise, it might be worth experimenting with:

- Retrieving the records in the same way as this spike
- Creating a flattened hash representation of each application form and its qualifications
- Blacklisting attributes we definitely don't care about (eg—`id`, `application_form_id`)
- Comparing the two hashes

Another observation— is the proposed way of scoping application forms at risk of double counting the same candidate? Does that matter? I'm thinking of a candidate whose application history goes something like:

```
Initial application ==> Apply Again 1 ==> Apply Again 2
```

My understanding is that both of those Apply Again forms will get picked up and treated as a separate comparison (ie — `Initial application` vs `Apply Again 1`, and `Apply Again 1` vs `Apply Again 2`.

app/services/candidate_interface/find_changed_apply_again_applications.rb
malcolmbaig: Ah, should've checked! 🤦🏼‍♂️ 

Might make sense to add `public_id` to `IGNORED_CHILD_ATTRIBUTES` and use that instead?

app/services/candidate_interface/find_changed_apply_again_applications.rb
stevehook: :+1: that's much better

app/services/candidate_interface/find_changed_apply_again_applications.rb
stevehook: yes that's what I should have done

stevehook:
> Overall approach looks good to me 👍🏼 . Couple of comments left inline.
> 
> > The set of changes we regard as significant is very much hard-coded into the filter code here. Is there a way/need to generalise it a bit?
> 
> I think this is fine, as long as it's clear to consumers of this data what the precise definition of "changed" is. If we needed to generalise, it might be worth experimenting with:
> 
> * Retrieving the records in the same way as this spike
> * Creating a flattened hash representation of each application form and its qualifications
> * Blacklisting attributes we definitely don't care about (eg—`id`, `application_form_id`)
> * Comparing the two hashes

I think sounds like a good approach, it would make sense to have some way of declaring the list of significant changes that isn't muddled up with the comparison logic. I'll give that a try.

> 
> Another observation— is the proposed way of scoping application forms at risk of double counting the same candidate? Does that matter? I'm thinking of a candidate whose application history goes something like:
> 
> ```
> Initial application ==> Apply Again 1 ==> Apply Again 2
> ```
> 
> My understanding is that both of those Apply Again forms will get picked up and treated as a separate comparison (ie — `Initial application` vs `Apply Again 1`, and `Apply Again 1` vs `Apply Again 2`.

Good question. I think we would be counting both apply again applications if they changed relative to their immediate predecessor. Need to check on the requirement there I think.

stevehook:
Closing this spike in favour of a tidied up version https://github.com/DFE-Digital/apply-for-teacher-training/pull/4249

PR #4217, opened 2021-03-04
Add new permissions based query service for courses and providers

mnacos:
## Context

When modifying course details when making an offer or changing an offer, we allow provider users to change all aspects of the offer, including the provider. This is subject to their list of providers and `make_decisions` permissions, of course. But the current logic of presenting all providers you're associated with does not work very well when the user represents a ratifying provider (e.g. an HEI), in which case there may be no direct association between the training provider and the user.

To address this, we have come up with a different set of rules for selecting providers to show as part of the change offer flow. The main idea is we work out which courses the current user has permissions to `make_decisions` on first, then work backwards to obtain the corresponding list of training providers.

Requirements:
- I can only see courses that are open on apply
- I can only see courses that I have permission to make_decisions on
- Having chosen a provider in the “Select provider” step, I should only see courses that provider runs
- I can only see courses that are run or ratified by the accredited body to which the application was originally submitted

This PR adds a query service which obtains a list of training providers based on the new rules.

Add a `GetOfferChangeOptions` query service which returns a list of training providers for a given user and course.

No UI changes and no need for a feature flag, as this service is not yet used anywhere.

## Guidance to review

The new service combines user-level and org-level permissions information for the user with courses to generate a list of `make_decisions_courses`, which is user specific and serves as the foundation for all the other checks.

Another quirk of the new rules is this concept of preserving the ratifying provider, which is why the service accepts a `original_course:` argument. If the current course is self-ratified, we use the training provider as the ratifying providers for this constraint, otherwise we use the ratifying provider of the current course.

The corresponding rspec file provides some interesting examples.

The `activerecord-cte` gem provides an elegant way to combine active record queries from different steps or services. `Provider.with(offerable_courses: offerable_courses)` is a good example of this.

## Link to Trello card

https://trello.com/c/huYAUhqp

## Things to check

### DIFF
--- a/app/queries/get_change_offer_options.rb
+++ b/app/queries/get_change_offer_options.rb
@@ -1,31 +1,32 @@
 class GetChangeOfferOptions
-  attr_accessor :user, :ratifying_provider, :recruitment_cycle_year
+  attr_accessor :user, :for_provider, :recruitment_cycle_year
 
-  def initialize(user:, ratifying_provider:, recruitment_cycle_year:)
+  def initialize(user:, for_provider:, recruitment_cycle_year:)
     @user = user
-    @ratifying_provider = ratifying_provider
+    @for_provider = for_provider
     @recruitment_cycle_year = recruitment_cycle_year
   end
 
-  def actionable_courses
-    courses_with_org_permission_joins = Course
+  def make_decisions_courses
+    Course
       .joins(training_provider_make_decisions)
       .joins(ratifying_provider_make_decisions)
-
-    same_year_open_courses_with_make_decisions = courses_with_org_permission_joins
       .where(combine_user_and_provider_permissions)
+  end
+
+  def offerable_courses
+    make_decisions_courses
       .where(
         open_on_apply: true,
         recruitment_cycle_year: recruitment_cycle_year,
       )
-
-    same_year_open_courses_with_make_decisions.where(accredited_provider: ratifying_provider)
+      .where(ratifying_provider_is_preserved)
   end
 
   def available_providers
     Provider
-      .with(actionable_courses: actionable_courses)
-      .joins('INNER JOIN actionable_courses ON providers.id = actionable_courses.provider_id')
+      .with(offerable_courses: offerable_courses)
+      .joins('INNER JOIN offerable_courses ON providers.id = offerable_courses.provider_id')
       .distinct
   end
 
@@ -68,5 +69,13 @@ private
         ratifying_provider_make_decisions.id IS NOT NULL
       )
     COMBINE_USER_AND_PROVIDER_PERMISSIONS
+  end
+
+  def ratifying_provider_is_preserved
+    # accredited_provider_id is nil for self-ratified courses
+    <<~RATIFYING_PROVIDER_IS_PRESERVED
+      (provider_id = #{for_provider.id} AND accredited_provider_id IS NULL)
+      OR accredited_provider_id = #{for_provider.id}
+    RATIFYING_PROVIDER_IS_PRESERVED
   end
 end
--- a/app/queries/get_change_offer_options.rb
+++ b/app/queries/get_change_offer_options.rb
@@ -6,35 +6,23 @@ class GetChangeOfferOptions
     @user = user
   end
 
+  def actionable_courses
+    courses_with_org_permission_joins = Course
+      .joins('INNER JOIN providers AS training_provider ON provider_id = training_provider.id')
+      .joins('LEFT OUTER JOIN providers AS ratifying_provider ON accredited_provider_id = ratifying_provider.id')
+      .joins(training_provider_make_decisions)
+      .joins(ratifying_provider_make_decisions)
+
+    courses_with_org_permission_joins
+      .where(
+        open_on_apply: true,
+        recruitment_cycle_year: application_choice.offered_course.recruitment_cycle_year,
+      )
+      .where(combine_user_and_provider_permissions)
+  end
+
   def available_providers
-    make_decision_providers = user.provider_permissions.where(make_decisions: true).map(&:provider)
-
-# course -> provider            -> provider_relationship_permissions (training side) -> providers
-#        -> accredited_provider -> provider_relationship_permissions (ratifying side) -> providers
-
-    # provider_id IN (1, 2, 3, 5)
-
-    # .with(
-      # provider_ids: user.provider_permissions.where(make_decisions: true).pluck(:provider_id)
-    # )
-    # .with(
-      # training_providers:
-        # Provider.joins('INNER JOIN courses ON courses.provider_id = providers.id AND courses.provider_id IN provider_ids')
-    # )
-
-    # .with(ratifying_providers: Course.where(accredited_provider: make_decision_providers))
-
-    courses = Course.where(open_on_apply: true,
-                          provider: make_decision_providers,
-                          recruitment_cycle_year: application_choice.offered_course.recruitment_cycle_year,
-                           )
-                          .or(
-                              Course.where(open_on_apply: true,
-                              accredited_provider: make_decision_providers,
-                              recruitment_cycle_year: application_choice.offered_course.recruitment_cycle_year,)
-                            )
-
-    courses.map(&:provider)
+    actionable_courses.map(&:provider)
   end
 
   # GetChangeOfferOptions.new(application_choice: ..., user: ...).available_courses(provider: ...)
@@ -47,23 +35,42 @@ class GetChangeOfferOptions
   
   # GetChangeOfferOptions.new(application_choice: ..., user: ...).available_sites(course: ..., study_mode: ...)
 
-  #   @available_courses = Course.where(
-  #     open_on_apply: true,
-  #     provider: application_choice.offered_course.provider,
-  #     study_mode: study_mode_for_other_courses,
-  #     recruitment_cycle_year: application_choice.offered_course.recruitment_cycle_year,
-  #   ).order(:name)
-  #
-  #   @available_study_modes = CourseOption.where(
-  #     course: application_choice.offered_course,
-  #   ).pluck(:study_mode).uniq
-  #
-  #   @available_course_options = CourseOption.where(
-  #     course: application_choice.offered_course,
-  #     study_mode: application_choice.offered_option.study_mode, # preserving study_mode
-  #   ).includes(:site).order('sites.name')
-  # end
-
 private
 
+  def provider_relationship_permissions_sql(join_name, additional_checks)
+    <<~PROVIDER_RELATIONSHIP_PERMISSIONS.squish
+      LEFT OUTER JOIN provider_relationship_permissions AS #{join_name}
+        ON training_provider.id = #{join_name}.training_provider_id
+        AND ratifying_provider.id = #{join_name}.ratifying_provider_id
+        AND #{additional_checks}
+    PROVIDER_RELATIONSHIP_PERMISSIONS
+  end
+
+  def training_provider_make_decisions
+    check = 'training_provider_make_decisions.training_provider_can_make_decisions IS TRUE'
+    provider_relationship_permissions_sql('training_provider_make_decisions', check)
+  end
+
+  def ratifying_provider_make_decisions
+    check = 'ratifying_provider_make_decisions.ratifying_provider_can_make_decisions IS TRUE'
+    provider_relationship_permissions_sql('ratifying_provider_make_decisions', check)
+  end
+
+  def combine_user_and_provider_permissions
+    provider_ids = user.provider_permissions.where(make_decisions: true).pluck(:provider_id)
+
+    <<~COMBINE_USER_AND_PROVIDER_PERMISSIONS
+      (
+        training_provider.id IN (#{provider_ids.join(',')}) AND
+        (
+          ratifying_provider.id IS NULL
+          OR training_provider_make_decisions.id IS NOT NULL
+        )
+      ) OR
+      (
+        ratifying_provider.id IN (#{provider_ids.join(',')}) AND
+        ratifying_provider_make_decisions.id IS NOT NULL
+      )
+    COMBINE_USER_AND_PROVIDER_PERMISSIONS
+  end
 end
--- a/app/queries/get_change_offer_options.rb
+++ b/app/queries/get_change_offer_options.rb
@@ -1,0 +1,69 @@
+class GetChangeOfferOptions
+  attr_accessor :application_choice, :user
+
+  def initialize(application_choice:, user:)
+    @application_choice = application_choice
+    @user = user
+  end
+
+  def available_providers
+    make_decision_providers = user.provider_permissions.where(make_decisions: true).map(&:provider)
+
+# course -> provider            -> provider_relationship_permissions (training side) -> providers
+#        -> accredited_provider -> provider_relationship_permissions (ratifying side) -> providers
+
+    # provider_id IN (1, 2, 3, 5)
+
+    # .with(
+      # provider_ids: user.provider_permissions.where(make_decisions: true).pluck(:provider_id)
+    # )
+    # .with(
+      # training_providers:
+        # Provider.joins('INNER JOIN courses ON courses.provider_id = providers.id AND courses.provider_id IN provider_ids')
+    # )
+
+    # .with(ratifying_providers: Course.where(accredited_provider: make_decision_providers))
+
+    courses = Course.where(open_on_apply: true,
+                          provider: make_decision_providers,
+                          recruitment_cycle_year: application_choice.offered_course.recruitment_cycle_year,
+                           )
+                          .or(
+                              Course.where(open_on_apply: true,
+                              accredited_provider: make_decision_providers,
+                              recruitment_cycle_year: application_choice.offered_course.recruitment_cycle_year,)
+                            )
+
+    courses.map(&:provider)
+  end
+
+  # GetChangeOfferOptions.new(application_choice: ..., user: ...).available_courses(provider: ...)
+  def available_courses(provider:)
+  end
+
+  # GetChangeOfferOptions.new(application_choice: ..., user: ...).available_study_modes(course: ...)
+  def available_study_modes(course:)
+  end
+  
+  # GetChangeOfferOptions.new(application_choice: ..., user: ...).available_sites(course: ..., study_mode: ...)
+
+  #   @available_courses = Course.where(
+  #     open_on_apply: true,
+  #     provider: application_choice.offered_course.provider,
+  #     study_mode: study_mode_for_other_courses,
+  #     recruitment_cycle_year: application_choice.offered_course.recruitment_cycle_year,
+  #   ).order(:name)
+  #
+  #   @available_study_modes = CourseOption.where(
+  #     course: application_choice.offered_course,
+  #   ).pluck(:study_mode).uniq
+  #
+  #   @available_course_options = CourseOption.where(
+  #     course: application_choice.offered_course,
+  #     study_mode: application_choice.offered_option.study_mode, # preserving study_mode
+  #   ).includes(:site).order('sites.name')
+  # end
+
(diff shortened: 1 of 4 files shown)


app/queries/get_change_offer_options.rb
```diff
@@ -0,0 +1,84 @@
+class GetChangeOfferOptions
+  attr_accessor :user, :for_provider, :recruitment_cycle_year
+
+  # application_choice.offered_course (either current or the original)
```
despo, line 4: Can this comment be removed

app/queries/get_change_offer_options.rb
```diff
@@ -0,0 +1,84 @@
...
+  def initialize(user:, current_course:)
+    @user = user
+    @for_provider = current_course.accredited_provider || current_course.provider
```
despo, line 7:
```suggestion
    @provider = current_course.accredited_provider || current_course.provider
```
I think `for_provider` is a bit confusing. I would have expected that to be used in an action if anywhere, but as a variable I don't see any readability benefits.

app/queries/get_change_offer_options.rb
```diff
@@ -0,0 +1,84 @@
...
+    @recruitment_cycle_year = current_course.recruitment_cycle_year
+  end
+
+  def make_decisions_courses
```
despo, line 11: I'm a bit confused as to why we have two different methods retrieving courses. Within the context of `GetChangeOfferOption` don't we just need one? Also can the one we need to use be called `courses` instead?

app/queries/get_change_offer_options.rb
```diff
@@ -0,0 +1,84 @@
...
+    provider_relationship_permissions_sql('training_provider_make_decisions', check)
+  end
+
+  def ratifying_provider_make_decisions
+    check = 'ratifying_provider_make_decisions.ratifying_provider_can_make_decisions IS TRUE'
+    provider_relationship_permissions_sql('ratifying_provider_make_decisions', check)
+  end
+
+  def combine_user_and_provider_permissions
+    return '1 != 1' if permitted_provider_ids.blank?
```
despo, line 60: What does `1 != 1` mean? 
Can it be moved to a well named constant so this line is easy to understand.

spec/queries/get_change_offer_options_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+  include CourseOptionHelpers
+
+  let(:ratifying_provider) { create(:provider) }
+  let(:self_ratified_course) { create(:course, :open_on_apply, provider: ratifying_provider) }
+  let(:externally_ratified_course) { create(:course, :open_on_apply, accredited_provider: ratifying_provider) }
+  let(:provider_user) { create(:provider_user) }
+
+  def service(provider_user, course)
+    GetChangeOfferOptions.new(user: provider_user, current_course: course)
+  end
```
despo, line 13:
```suggestion
  let(:service) { described_class.new(user: provider_user, current_course: course)
```    
Can we use a`let` and context blocks around the examples to setup the required data.

spec/queries/get_change_offer_options_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+
+  def set_user_permission(provider, make_decisions)
+    provider_user.provider_permissions << ProviderPermissions.new(
+      provider: provider,
+      make_decisions: make_decisions,
+    )
+  end
```
despo, line 20: Can we utilise the `provider` factory instead as all examples set `make_decisions` to true

```ruby
let(:ratifying_provider) { create(:provider, :with_user) }
let(:provider_user) { ratifying_provider.provider_users.first }

File: spec/queries/get_change_offer_options_spec.rb [hunk 7]
```diff
@@ -0,0 +1,126 @@
...
+
+  def set_provider_permissions(training_provider, make_d1, ratifying_provider, make_d2)
+    create(
+      :provider_relationship_permissions,
+      training_provider: training_provider,
+      ratifying_provider: ratifying_provider,
+      training_provider_can_make_decisions: make_d1,
+      ratifying_provider_can_make_decisions: make_d2,
+    )
+  end
```
despo, line 30: Can this be moved to a `let` an setup the required data within a context wrapping the example isntead.

app/queries/get_change_offer_options.rb
```diff
@@ -0,0 +1,84 @@
...
+        recruitment_cycle_year: recruitment_cycle_year,
+      )
+      .where(ratifying_provider_is_preserved)
+  end
+
+  def available_providers
+    Provider
+      .with(offerable_courses: offerable_courses)
+      .joins('INNER JOIN offerable_courses ON providers.id = offerable_courses.provider_id')
+      .distinct
```
despo, line 31:
```suggestion
      .group('providers.id')
```

can we avoid using distinct if possible.

despo (commented):
Left some feedback and some questions inline.

Also can the commit log and messages be amended so this is easier to follow.

app/queries/get_change_offer_options.rb
mnacos: The one you need for the wizard is `offerable_courses`, but you'll probably restrict it to the chosen provider, e.g. `service.offerable_courses.where(provider: ...)`

The `current_course:` argument to the service determines the ratifying provider to perserve. In most cases, this will be `application_choice.offered_course`, hence the comment we'll remove.

We could rename `offerable_courses` to `courses`, but isn't this a bit less descriptive? You won't be just calling `.courses` in the wizard.

As for the existence `make_decisions_courses`, we're trying to isolate this part of the query in order to test it properly, as it's a database-level re-implementation of `can_make_decisions?` from `ProviderAuthorisation`. Once we establish this works correctly in the rspec examples, we can stop worrying about permissions and test the other stuff (like preserving the ratifying body) without worrying about which complicated permissions setup we're bundling into the other tests.

The `make_decisions_courses` aspect could eventually be moved to a different part of the code, such as `User` or `ProviderAuthorisation`.

app/queries/get_change_offer_options.rb
mnacos: If your user has no providers, by definition you are not authorised to make decisions on anything. By adding `1 != 1` to the where clause we guarantee no courses are returned while still supplying valid SQL for the `where`. Without it, an exception is thrown because `provider_id IN ()` is not valid SQL.

app/queries/get_change_offer_options.rb
mnacos: Will remove

app/queries/get_change_offer_options.rb
mnacos: Would `@accredited_provider` be better? We use this to select courses ratified by this provider, which, for self-ratified courses, is the training provider. So just `@provider` seems simple/deceptive to me.

spec/queries/get_change_offer_options_spec.rb
mnacos: You're right that we seem to always set `make_decisions` to true, so I'll make this the default and override it in the scenario that needs it.

However, beyond `make_decisions`, a big factor in these tests is which provider your user is associated with. Using a factory would only make sense if `let(:provider_user)` was re-defined in context blocks, which you've suggested, but I don't think is appropriate for these tests.

spec/queries/get_change_offer_options_spec.rb
mnacos: No, I would rather avoid context blocks. Every scenario is unique.

mnacos:
> Left some feedback and some questions inline.
> 
> Also can the commit log and messages be amended so this is easier to follow.

Thanks @despo, I've tried to answer your questions. Happy to discuss more if you want. We'll make sure to rebase and squash the commits.

app/queries/get_change_offer_options.rb
mnacos: `distinct` performs badly in some cases, but no sure it will here. I'll have a look at the SQL produced and do some testing.

app/queries/get_change_offer_options.rb
despo: Yes, `@accredited_provider` works too!

spec/queries/get_change_offer_options_spec.rb
despo: I mean an individual context around each block, similar to how we do in other specs that require data setup.

app/queries/get_change_offer_options.rb
mnacos: Added a comment

spec/queries/get_change_offer_options_spec.rb
mnacos: I tried to introduce a `let(:service)` in the later sections of the spec but it clashes with the mocks. In the `make_decisions_courses` section it's important to re-instantiate, otherwise we need to do a lot of `.reload`.

app/queries/get_change_offer_options.rb
mnacos: Switched it to `.group` just to be on the safe side. Will have another look once it's merged.

spec/queries/get_change_offer_options_spec.rb
mnacos: We only use this when it's needed and when we use it we need full control of which side has `make_decisions`.

spec/queries/get_change_offer_options_spec.rb
mnacos: I've renamed the method to `set_provider` to make it clear a big part of using this method is setting which provider the user is associated with (training vs. ratifying).

app/queries/get_change_offer_options.rb
```diff
@@ -0,0 +1,83 @@
+class GetChangeOfferOptions
+  attr_accessor :user, :accredited_provider, :recruitment_cycle_year
+
+  def initialize(user:, current_course:)
```
despo, line 4: In order to be able to implement the change link for Changing a course, we expect to be able to initialise the service with the selected provider, and be able to get back the list of courses.

app/queries/get_change_offer_options.rb
despo: Maybe it would make more sense then to move this method in it's own service, so it's descriptive enough and can be tested properly. Then it should also be reusable if required by other services/code that is not part of the make offer funcitonality.


> We could rename offerable_courses to courses, but isn't this a bit less descriptive? You won't be just calling .courses in the wizard.

We would be calling `GetChangeOfferOption.new(...).courses` which I think is sufficiently descriptive.

app/queries/get_change_offer_options.rb
```diff
@@ -0,0 +1,83 @@
...
+    provider_relationship_permissions_sql('training_provider_make_decisions', check)
+  end
+
+  def ratifying_provider_make_decisions
+    check = 'ratifying_provider_make_decisions.ratifying_provider_can_make_decisions IS TRUE'
+    provider_relationship_permissions_sql('ratifying_provider_make_decisions', check)
+  end
+
+  def combine_user_and_provider_permissions
+    return '1 != 1' if permitted_provider_ids.blank? # valid SQL which returns no results
```
despo, line 59:
```ruby
EMPTY_QUERY = '1 != 1'
```
```suggestion
    return EMPTY_QUERY if permitted_provider_ids.blank?
```

app/queries/get_change_offer_options.rb
mnacos: This service focuses on getting the list of `offerable_courses` for a user, which is a function of the current user and a course context (for preserving the ratifying provider). In this sense it's independent of any choices the user makes in the wizard.

For getting the list of courses for a given provider you could do `.offerable_courses.where(provider: ...)`, as all public methods in this service return active record queries. Likewise, for verifying the user has make_decisions on the chosen course, you carry on scoping e.g. `.offerable_courses.find_by(id: course_id)`

Once you reach the course level, and you verify the course is appropriate for this user, you don't necessarily need to scope anymore.

We could add a `def courses(provider:)` to the service, if it helps.

spec/queries/get_change_offer_options_spec.rb
despo: I'm not sure I understand. Using a `let` will create a new instance of `service` for every example it's referenced from, so not sure why reloading would be required.

app/queries/get_change_offer_options.rb
despo: Did we miss a card then? We were expecting to get a service that we would just plug in, and it would give us back anything we needed depending on what we requested from it. Also the card name is a bit deceiving then `Add new permissions-based query service for courses and **providers**`

mnacos: I don't think so. As far as I know, these are the queries you'll need:

- A list of training providers based on courses the user has make_decisions on, already available via `.available_providers`. This needs the context of a user and a ratifying provider, which is extracted from a course (e.g. application_choice.offered_course). This list is not influenced by any choices the user makes in the wizard
- A list of courses based on the provider the user has chosen in the wizard. This needs to be a subset of courses the user has make_decisions on, not all courses the chosen provider runs. Therefore, it's `offerable_courses` (or just `.courses`, if you prefer), scoped to a provider id.
- A list of study modes for the chosen course, available via either `.study_mode` on the selected course or by plucking `.study_mode` from the course's course_options.
- A list of locations for the chosen course, available via plucking `site` from the course's course_options.

Do we really need another service for the last queries? If it helps, we can wrap these in their own methods. Let's discuss on Monday.

mnacos: See also https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/queries/get_all_change_options_from_offered_option.rb#L21

app/queries/get_change_offer_options.rb
```diff
@@ -0,0 +1,83 @@
...
+      .where(combine_user_and_provider_permissions)
+  end
+
+  def offerable_courses
+    make_decisions_courses
+      .where(
+        open_on_apply: true,
+        recruitment_cycle_year: recruitment_cycle_year,
+      )
+      .where(ratifying_provider_is_preserved)
```
duncanjbrown, line 23: This is a very good, explicit method — props @mnacos and @SamJamCul !

spec/queries/get_change_offer_options_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+
+  def set_provider(provider, make_decisions: true)
+    provider_user
+      .provider_permissions
+      .find_or_create_by(provider: provider)
+      .update(make_decisions: make_decisions)
+  end
+
+  def set_provider_permissions(training_provider, make_d1, ratifying_provider, make_d2)
```
duncanjbrown, line 22: Nits: maybe `set_make_decisions_permissions` for clarity?

spec/queries/get_change_offer_options_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+  let(:ratifying_provider) { create(:provider) }
+  let(:self_ratified_course) { create(:course, :open_on_apply, provider: ratifying_provider) }
+  let(:externally_ratified_course) { create(:course, :open_on_apply, accredited_provider: ratifying_provider) }
+  let(:provider_user) { create(:provider_user) }
+
+  def service(provider_user, course)
+    described_class.new(user: provider_user, current_course: course)
+  end
+
+  def set_provider(provider, make_decisions: true)
```
duncanjbrown, line 15: Nit2: maybe `set_provider_user_make_decisions`?

duncanjbrown: And maybe remove the default value for `make_decisions`? Is nice to be explicit in spec body

spec/queries/get_change_offer_options_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+    it 'returns externally and self-ratified courses based on an self-ratified course' do
+      service = service(provider_user, self_ratified_course)
+      create(:course, :open_on_apply)
+      allow(service).to receive(:make_decisions_courses).and_return(Course.all)
+      expect(service.offerable_courses).to match_array([externally_ratified_course, self_ratified_course])
+    end
+  end
+
+  describe '#available_providers' do
+    it 'only returns training providers, even if user is not associated with them' do
```
duncanjbrown, line 112: Maybe

```suggestion
    it 'only returns training providers, even if the user is associated with an accredited provider' do
```

?

spec/queries/get_change_offer_options_spec.rb
SamJamCul: Updated!

spec/queries/get_change_offer_options_spec.rb
SamJamCul: Changed!

spec/queries/get_change_offer_options_spec.rb
SamJamCul: Changed for clarity, thanks

app/queries/get_change_offer_options.rb
mnacos: We've changed this to just 'FALSE'

app/queries/get_change_offer_options.rb
mnacos: In line with our earlier chat, we now have `available_courses`, `available_study_modes`, `available_sites` and `available_course_options` as well as `available_providers`, but these new methods require additional context which is only relevant as you progress in the wizard.

I've added both `available_sites` and `available_course_options`, it's up to you which one you use during the 'location' part of the journey (using course options instead of sites for location radio buttons would allow you to pass a course_option_id, instead of a site_id, with the POST).

cwrw (approved):
I joined the party a bit late but seems most comments have been addressed and the interface has been agreed. I left some minor comments, but I think there is a bit to think about whilst integrating this with our make offer service so I do envision this might change a bit, but I think its fair to leave that till then.

I think regarding an older question around clarity on how we should handle self-ratified courses, I'm chasing up a confirmation of an official answer but I don't think this will block this work, and if any changes come in we can pick them up in a subsequent PR.

I think something noted here is that there is a bit of complex logic around permissions which we need to revisit later (as touched on previously). Otherwise well done both!

spec/queries/get_change_offer_options_spec.rb
mnacos: Reloading was required because the specs would start with no `make_decisions` permissions, verify no courses were returned, then set the permissions and try again. I've since changes the specs in line with  @cwrw 's suggestion to separate expectations into their own `it` blocks but I'm trying to avoid using too many contexts.

spec/queries/get_change_offer_options_spec.rb
mnacos: It's now `set_provider_with_make_decisions`.

spec/queries/get_change_offer_options_spec.rb
```diff
@@ -0,0 +1,178 @@
...
+    )
+  end
+
+  describe '#make_decisions_courses' do
+    it 'returns no courses if user is not associated with the course via a provider' do
+      expect(service(provider_user, self_ratified_course).make_decisions_courses).to be_empty
+    end
+
+    it 'returns a self-ratified course when a user has user-level make decisions' do
+      set_provider_user_make_decisions(self_ratified_course.provider, make_decisions: false)
```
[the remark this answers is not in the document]
mnacos: Yes, I've just done this. The specs were just testing the negative case first before assigning permissions and testing again, but the same can be accomplished with explicit `it` blocks.

app/queries/get_change_offer_options.rb
```diff
@@ -0,0 +1,111 @@
...
+      .group('providers.id')
+  end
+
+  def available_courses(provider:)
+    offerable_courses.where(provider: provider)
+  end
+
+  def available_study_modes(course:)
+    CourseOption
+      .where(course: offerable_courses.find(course.id))
```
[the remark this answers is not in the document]
mnacos: Yes, this is exactly it.

despo (approved):

PR #4218, opened 2021-03-04
Happy path for create the new make offer flow

richardpattinson:
## Context

Implement a new offer flow (happy path only), including basic validations for condition length. 

This feature is done when: 
> A provider can make a new offer, enter the conditions required and send the offer to the candidate using the new flow.
> Add new views/controllers and routes, add a feature flag to the main start page for make an offer.

## Guidance to review

Go to support page and enable the updated make offer flow flag. Then log in as a provider and select an application choice that is in either the received or interviewing status and make a decision on it. 

## Link to Trello card

https://trello.com/c/6Ay7XJYu/3458-happy-path-for-create-the-new-make-offer-flow

## Things to check

### DIFF
--- a/app/controllers/provider_interface/decisions_controller.rb
+++ b/app/controllers/provider_interface/decisions_controller.rb
@@ -1,6 +1,7 @@
 module ProviderInterface
   class DecisionsController < ProviderInterfaceController
     before_action :set_application_choice
+    before_action :confirm_application_is_in_decision_pending_state, only: %i[new create]
     before_action :requires_make_decisions_permission
 
     def new
@@ -165,6 +166,12 @@ module ProviderInterface
         decision: :default,
         standard_conditions: MakeAnOffer::STANDARD_CONDITIONS,
       }
+    end
+
+    def confirm_application_is_in_decision_pending_state
+      return if ApplicationStateChange::DECISION_PENDING_STATUSES.include?(@application_choice.status.to_sym)
+
+      redirect_back(fallback_location: provider_interface_application_choice_path(@application_choice))
     end
 
     def provider_interface_offer_params
--- a/app/controllers/provider_interface/decisions_controller.rb
+++ b/app/controllers/provider_interface/decisions_controller.rb
@@ -157,7 +157,7 @@ module ProviderInterface
         study_mode: course_option.study_mode,
         location_id: course_option.site.id,
         current_context: :default,
-        conditions: MakeAnOffer::STANDARD_CONDITIONS,
+        standard_conditions: MakeAnOffer::STANDARD_CONDITIONS,
       }
     end
 
--- a/app/controllers/provider_interface/decisions_controller.rb
+++ b/app/controllers/provider_interface/decisions_controller.rb
@@ -18,8 +18,6 @@ module ProviderInterface
         @wizard.save_state!
 
         if @wizard.decision == 'rejection'
-          @wizard.clear_state!
-
           redirect_to provider_interface_reasons_for_rejection_initial_questions_path(@application_choice)
         else
           redirect_to [:new, :provider_interface, @application_choice, :offer, @wizard.next_step]
--- a/app/controllers/provider_interface/decisions_controller.rb
+++ b/app/controllers/provider_interface/decisions_controller.rb
@@ -3,7 +3,17 @@ module ProviderInterface
     before_action :set_application_choice
     before_action :requires_make_decisions_permission
 
+    def new
+      @wizard = OfferWizard.new(offer_store,
+                                offer_context_params(@application_choice.course_option).merge!(current_step: 'select_option'))
+      @wizard.save_state!
+    end
+
     def respond
+      if FeatureFlag.active?(:updated_offer_flow)
+        redirect_to new_provider_interface_application_choice_decision_path(@application_choice) and return
+      end
+
       @pick_response_form = PickResponseForm.new
       @alternative_study_mode = @application_choice.offered_option.alternative_study_mode
     end
@@ -124,6 +134,23 @@ module ProviderInterface
 
     def make_an_offer_params
       params.require(:make_an_offer)
+    end
+
+    def offer_context_params(course_option)
+      {
+        course_id: course_option.course.id,
+        course_option_id: course_option.id,
+        provider_id: course_option.provider.id,
+        study_mode: course_option.study_mode,
+        location_id: course_option.site.id,
+        current_context: :default,
+        conditions: MakeAnOffer::STANDARD_CONDITIONS,
+      }
+    end
+
+    def offer_store
+      key = "offer_wizard_store_#{current_provider_user.id}_#{@application_choice.id}"
+      WizardStateStores::RedisStore.new(key: key)
     end
   end
 end
--- a/app/controllers/provider_interface/decisions_controller.rb
+++ b/app/controllers/provider_interface/decisions_controller.rb
@@ -10,15 +10,21 @@ module ProviderInterface
     end
 
     def create
-      @wizard = OfferWizard.new(offer_store, { current_context: selected_decision })
-      @wizard.save_state!
+      @wizard = OfferWizard.new(offer_store, { decision: selected_decision })
 
-      if @wizard.current_context == 'rejection'
-        @wizard.clear_state!
+      if @wizard.valid?
 
-        redirect_to provider_interface_reasons_for_rejection_initial_questions_path(@application_choice)
+        @wizard.save_state!
+
+        if @wizard.decision == 'rejection'
+          @wizard.clear_state!
+
+          redirect_to provider_interface_reasons_for_rejection_initial_questions_path(@application_choice)
+        else
+          redirect_to [:new, :provider_interface, @application_choice, :offer, @wizard.next_step]
+        end
       else
-        redirect_to [:new, :provider_interface, @application_choice, :offer, @wizard.next_step]
+        render 'new'
       end
     end
 
@@ -156,13 +162,17 @@ module ProviderInterface
         provider_id: course_option.provider.id,
         study_mode: course_option.study_mode,
         location_id: course_option.site.id,
-        current_context: :default,
+        decision: :default,
         standard_conditions: MakeAnOffer::STANDARD_CONDITIONS,
       }
     end
 
+    def provider_interface_offer_params
+      params[:provider_interface_offer_wizard] || ActionController::Parameters.new
+    end
+
     def selected_decision
-      params.require(:provider_interface_offer_wizard).permit(:decision)[:decision]
+      provider_interface_offer_params.permit(:decision)[:decision]
     end
 
     def offer_store
--- a/app/controllers/provider_interface/decisions_controller.rb
+++ b/app/controllers/provider_interface/decisions_controller.rb
@@ -12,7 +12,7 @@ module ProviderInterface
     def create
       @wizard = OfferWizard.new(offer_store, { decision: selected_decision })
 
-      if @wizard.valid?
+      if @wizard.valid_for_current_step?
 
         @wizard.save_state!
 
--- a/app/controllers/provider_interface/decisions_controller.rb
+++ b/app/controllers/provider_interface/decisions_controller.rb
@@ -18,7 +18,7 @@ module ProviderInterface
 
         redirect_to provider_interface_reasons_for_rejection_initial_questions_path(@application_choice)
       else
-        redirect_to [:new, :provider_interface, :offer, @wizard.next_step]
+        redirect_to [:new, :provider_interface, @application_choice, :offer, @wizard.next_step]
       end
     end
 
--- a/app/controllers/provider_interface/decisions_controller.rb
+++ b/app/controllers/provider_interface/decisions_controller.rb
@@ -9,6 +9,19 @@ module ProviderInterface
       @wizard.save_state!
     end
 
+    def create
+      @wizard = OfferWizard.new(offer_store, { current_context: selected_decision })
+      @wizard.save_state!
+
+      if @wizard.current_context == 'rejection'
+        @wizard.clear_state!
+
+        redirect_to provider_interface_reasons_for_rejection_initial_questions_path(@application_choice)
+      else
+        redirect_to [:new, :provider_interface, :offer, @wizard.next_step]
+      end
+    end
+
     def respond
       if FeatureFlag.active?(:updated_offer_flow)
         redirect_to new_provider_interface_application_choice_decision_path(@application_choice) and return
@@ -146,6 +159,10 @@ module ProviderInterface
         current_context: :default,
         conditions: MakeAnOffer::STANDARD_CONDITIONS,
       }
+    end
+
+    def selected_decision
+      params.require(:provider_interface_offer_wizard).permit(:decision)[:decision]
     end
 
     def offer_store
--- a/app/controllers/provider_interface/offer/conditions_controller.rb
+++ b/app/controllers/provider_interface/offer/conditions_controller.rb
@@ -8,17 +8,24 @@ module ProviderInterface
         @wizard.save_state!
       end
 
-      def create; end
+      def create
+        @wizard = OfferWizard.new(offer_store, conditions_params.to_h)
+        @wizard.save_state!
+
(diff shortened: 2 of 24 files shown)


app/views/provider_interface/decisions/new.html.erb
```diff
@@ -0,0 +1,34 @@
+<% content_for :browser_title, title_with_error_prefix(t('page_titles.provider.respond'), @wizard.errors.any?) %>
```
richardpattinson, line 1: Update to use new locale title.

app/views/provider_interface/decisions/new.html.erb
```diff
@@ -0,0 +1,34 @@
...
+<% content_for :before_content, govuk_back_link_to(provider_interface_application_choice_path(@application_choice.id)) %>
```
richardpattinson, line 2:
```suggestion
<% content_for :before_content, govuk_back_link_to(provider_interface_application_choice_path(@application_choice)) %>
```

app/views/provider_interface/decisions/new.html.erb
```diff
@@ -0,0 +1,34 @@
...
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l">
+      <span class="govuk-caption-l"><%= @application_choice.application_form.full_name %></span>
+      <%= t('page_titles.provider.respond') %>
```
richardpattinson, line 8: Use the updated locale.

app/views/provider_interface/offer/checks/new.html.erb
```diff
@@ -0,0 +1,84 @@
...
+          { key: 'Location',
+            value: @wizard.course_option.site.name_and_address,
+            action: 'Change',
+            change_path: nil,
+          },
+          { key: 'Full time or part time',
+            value: @wizard.course_option.study_mode.humanize,
+            action: 'Change',
+            change_path: nil,
+          }] %>
```
richardpattinson, line 38: Consider extracting outside of a view. Either moving to its own component or other clever solution.

app/views/provider_interface/offer/conditions/new.html.erb
```diff
@@ -0,0 +1,40 @@
+<% content_for :browser_title, title_with_error_prefix(t('.title'), @wizard.errors.any?) %>
+<% content_for :before_content, govuk_back_link_to(provider_interface_application_choice_respond_path(@application_choice.id)) %>
```
richardpattinson, line 2:
```suggestion
<% content_for :before_content, govuk_back_link_to(provider_interface_application_choice_respond_path(@application_choice)) %>
```
Apply the same to all views.

app/controllers/provider_interface/decisions_controller.rb
```diff
@@ -125,5 +155,36 @@ def defer_offer
...
+        course_option_id: course_option.id,
+        provider_id: course_option.provider.id,
+        study_mode: course_option.study_mode,
+        location_id: course_option.site.id,
+        decision: :default,
+        standard_conditions: MakeAnOffer::STANDARD_CONDITIONS,
+      }
+    end
+
+    def application_choice_allowed_to_make_decision
```
benswannack, line 171: This method could be more descriptively named as `redirect_back_if_application_in_decision_pending_state`

benswannack: Having said that, this check looks like it would be the wrong way around. Also, the `status` is a string where the array contains symbols only.

Can we add a spec to cover this behaviour?

app/controllers/provider_interface/offer/conditions_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+        @wizard.save_state!
+      end
+
+      def create
+        @wizard = OfferWizard.new(offer_store, conditions_params.to_h)
+
+        if @wizard.valid_for_current_step?
+          @wizard.save_state!
+
+          redirect_to [:new, :provider_interface, @application_choice, :offer, @wizard.next_step]
```
benswannack, line 15: Is there a way to write this in a more readable way? It's not immediately clear to me where we are redirecting to (because of the array)

app/forms/provider_interface/offer_wizard.rb
```diff
@@ -0,0 +1,58 @@
+module ProviderInterface
+  class OfferWizard
+    include ActiveModel::Model
+
+    STEPS = { make_offer: %i[select_option conditions check] }.freeze
```
benswannack, line 5: How come this appears to be dependent on the decision?

app/controllers/provider_interface/decisions_controller.rb
```diff
@@ -125,5 +155,36 @@ def defer_offer
...
+      return unless ApplicationStateChange::DECISION_PENDING_STATUSES.include?(@application_choice.status)
+
+      redirect_back(fallback_location: provider_interface_application_choice_path(@application_choice))
```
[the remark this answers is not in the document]
despo: They would have to be on the application_choice page to get here (assuming they pressed the Make decision button) so it made sense to take them back to the specific tab that they were in e.g. application choice notes, or timeline.

app/controllers/provider_interface/offer/conditions_controller.rb
despo: The alternative was to use metaprogramming but I think that makes things even harder to undertand, whilst this format is not uncommon in rails.

app/forms/provider_interface/offer_wizard.rb
despo: Because the selected decision controls which steps are displayed in the original flow.

app/controllers/provider_interface/decisions_controller.rb
```diff
@@ -1,9 +1,39 @@
...
+
+    def create
+      @wizard = OfferWizard.new(offer_store, { decision: selected_decision })
+
+      if @wizard.valid_for_current_step?
+
+        @wizard.save_state!
+
+        if @wizard.decision == 'rejection'
+          @wizard.clear_state!
```
cwrw, line 21: Clearing the state here is causing errors if I go back in the browser and try to make an offer again/choose rejection, is there a way we could avoid this?

app/forms/provider_interface/offer_wizard.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    attr_accessor :provider_id, :course_id, :course_option_id, :study_mode, :location_id,
+                  :standard_conditions, :further_condition_1, :further_condition_2,
+                  :further_condition_3, :further_condition_4, :current_step, :decision
+
+    validates :decision, presence: true
+    validates :further_condition_1, :further_condition_2, :further_condition_3, :further_condition_4, length: { maximum: 255 }
```
cwrw, line 12: I was wondering around making `further_condition` here dynamically accept as many further conditions as passed in, we would still hard code it in the view because we need to, but here we allow as many as possible, what do you think?

app/views/provider_interface/decisions/new.html.erb
```diff
@@ -0,0 +1,34 @@
...
+      <span class="govuk-caption-l"><%= @application_choice.application_form.full_name %></span>
+      <%= t('.title') %>
+    </h1>
+    <%= render ProviderInterface::CourseSummaryComponent.new(course_option: @application_choice.course_option) %>
+
+    <%= form_with model: @wizard, url: provider_interface_application_choice_decision_path(@application_choice), method: :post do |f| %>
+      <%= f.govuk_error_summary %>
+
+      <%= f.govuk_radio_buttons_fieldset :decision, legend: { text: t('.select'), size: 'm' } do %>
+        <%= f.govuk_radio_button :decision,
```
cwrw, line 16: Do we need something like this here as well?
```
if ApplicationStateChange.new(@application_choice).can_make_offer?
```

app/views/provider_interface/decisions/new.html.erb
```diff
@@ -0,0 +1,34 @@
...
+          'make_offer',
+          label: { text: 'Make an offer' },
```
cwrw, line 18: should we also make use of locales here and below?

app/views/provider_interface/decisions/new.html.erb
```diff
@@ -0,0 +1,34 @@
...
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l">
+      <span class="govuk-caption-l"><%= @application_choice.application_form.full_name %></span>
+      <%= t('.title') %>
+    </h1>
+    <%= render ProviderInterface::CourseSummaryComponent.new(course_option: @application_choice.course_option) %>
+
+    <%= form_with model: @wizard, url: provider_interface_application_choice_decision_path(@application_choice), method: :post do |f| %>
+      <%= f.govuk_error_summary %>
```
cwrw, line 13: I wonder if this should be at the top of the summary component? 🤔

app/controllers/provider_interface/decisions_controller.rb
despo: This is interesting. I guess we can move clearing the wizard to the end of the rejection process, as that would be the end of the decision flow when rejection is picked.

app/forms/provider_interface/offer_wizard.rb
despo: As we already plan to tackle conditions separately I don't think there is much value in doing that right now. It would also make it unecessarily complex to display/enforce the model validations.

app/views/provider_interface/decisions/new.html.erb
despo: I don't think we do as there are pre-existing checks in the application_choice_header_component, ensuring that the Make Decision button is only displayed if the application_choice is in a decision making state (either [awaiting_provider_decision](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/components/provider_interface/application_choice_header_component.html.erb#L34) or [interviewing](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/components/provider_interface/application_choice_header_component.html.erb#L10). 
 
As part of this PR we are also enforcing this in the DecisionsController.

Based on this, is seems like the condition that we moved across from the old Make offer flow, around displaying the Rejection radio button, is the one not required, so I will remove it.

app/views/provider_interface/decisions/new.html.erb
despo: Yes I think it should as that's the case on the interview page as well.

app/views/provider_interface/offer/conditions/new.html.erb
```diff
@@ -0,0 +1,38 @@
...
+  <%= f.govuk_error_summary %>
+
+  <h1 class="govuk-heading-xl">
+    <span class="govuk-caption-xl"><%= @application_choice.application_form.full_name %></span>
+    <%= t('.title') %>
+  </h1>
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <%= f.govuk_collection_check_boxes(:standard_conditions,
```
cwrw, line 16: There are a few style differences and missing text ("We’ll email the candidate to tell them that they’ll need to provide ID
") when comparing with the prototype, (ignoring the add another as that's out of scope) could we make them similar?

app/views/provider_interface/offer/checks/new.html.erb
```diff
@@ -0,0 +1,80 @@
...
+    <span class="govuk-caption-xl"><%= @application_choice.application_form.full_name %></span>
+    <%= t('.title') %>
+  </h1>
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <%= f.govuk_error_summary %>
+      <div class="app-offer-panel app-banner app-banner--details">
+        <h2 class="govuk-heading-m">Course details</h2>
+        <% rows = [
```
cwrw, line 17: can the row definition move to the controller?

app/controllers/provider_interface/decisions_controller.rb
```diff
@@ -1,9 +1,39 @@
...
+
+          redirect_to provider_interface_reasons_for_rejection_initial_questions_path(@application_choice)
+        else
+          redirect_to [:new, :provider_interface, @application_choice, :offer, @wizard.next_step]
```
cwrw, line 25: It seems that sometimes we lose the `@wizard.next_step` and we get into an inconsistent state and I get the following error:
```
undefined method `new_provider_interface_application_choice_offer_url'
```
when it tries to build this path, maybe we can have a fallback if the next step is missing? this might be resolved by the clearing the wizard state above

app/views/provider_interface/offer/conditions/new.html.erb
despo: As we decided to not tackle conditions as part of this piece of work we felt that we should ignore this, as it is included in the new piece of work.

app/views/provider_interface/offer/checks/new.html.erb
despo: Maybe a new component to wrap this instead, so we can also test it.

app/controllers/provider_interface/decisions_controller.rb
despo: Can you point out a scenario when this is happenning? Is this after going back from the Rejections controller or are there any other scenarios as well?

app/views/provider_interface/offer/checks/new.html.erb
```diff
@@ -0,0 +1,80 @@
+<% content_for :browser_title, title_with_error_prefix(t('.title'), @wizard.errors.any?) %>
+<% content_for :before_content, govuk_back_link_to(provider_interface_application_choice_new_offer_path(@application_choice.id)) %>
+
+<%= render(FlashMessageComponent.new(flash: flash)) %>
```
cwrw, line 4: do we need a flash message here?

cwrw (commented):
Well done, a lot of good work here, just a few comments mostly around some errors that come up from different wizard state.

I also noticed that we introduced many more controllers here for each check/condition etc., although its following a more railsy way, I wonder if they can live under the same controller instead, or do we envision they will become more unweildy/complicated with the next iteration on this to introduce changing things?

app/controllers/provider_interface/decisions_controller.rb
despo: @cwrw So we attempted to amend the form object's `next_step` so it defaults to the one at the beginning of the flow. However, as the decisions page does not follow our path convention, that didn't work.

The temporary solution that we have put in place for now, so that the error can be handle gracefully, is on commit https://github.com/DFE-Digital/apply-for-teacher-training/pull/4218/commits/bf04a9b0174d012865705a61d4f886e21c66d18b. Do you have any thoughts on this? 

I think we can look more into it as part of change offer.

app/views/provider_interface/offer/checks/new.html.erb
despo: No, we should not need this in any of the forms as it's rendered in the application layout. I have also removed from the conditions/new page. The old offer pages appear to use it, but as we will be removing them soon I don't think there is any value in revisiting and removing now.

app/controllers/provider_interface/decisions_controller.rb
cwrw: I think that handling the error gracefully is much better, although still surprising to be redirected to the main new offer page, but happy to wait for the change offer work to tackle this later 👍

app/controllers/provider_interface/decisions_controller.rb
despo: https://github.com/DFE-Digital/apply-for-teacher-training/pull/4218/files#diff-437c905cf8db3c525696f24fdd6c586ba729cbe555aa9eeb6a0838484478726b

app/controllers/provider_interface/decisions_controller.rb
despo: Moved to the end of the rejection process, which seems to have resolved this issue.

cwrw (approved):
Looks great 👍  thanks for addressing comments. Myself and @despo had a chat around the wizard behaviour and we noted that if we go back in the browser we might miss a step, but this is something that exists in other wizards and is worth us revisiting generally in further work.

PR #4219, opened 2021-03-04
Prevent application choices being added to application forms from previous recruitment cycles

steventux:
## Context

This PR attempts to fix a bug where a candidate starts an application in the previous recruitment cycle but does not add a course.
They then attempt to **add a course from Find** in the current recruitment cycle.
Currently this will add the course to the application form from the previous recruitment cycle, then redirect the user to the carry over start page.

Modifies `EndOfCyclePolicy.can_add_course_choice?` to check that if the `current_application` (ie the old `ApplicationForm`) needs to be carried over. This has the effect of short-cutting the journey and preventing the candidate from adding the course to the old application form.

## Guidance to review

There's potential that changing `EndOfCyclePolicy` will have side effects.
It's a bit of a bumpy candidate journey as they have selected a course and will be re-routed to choose again for the new application form.
Is there a better way? eg. Should `Candidate#current_application` be smarter?
 

## Link to Trello card

https://trello.com/c/psRD7rKb/3212-application-choices-being-added-to-application-forms-from-previous-recruitment-cycles

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/apply_from_find_controller.rb
+++ b/app/controllers/candidate_interface/apply_from_find_controller.rb
@@ -9,7 +9,9 @@ module CandidateInterface
     def show
       set_service_and_course(params[:providerCode], params[:courseCode])
 
-      if @service.can_apply_on_apply? && current_candidate.present?
+      if candidate_has_application_in_different_recruitment_cycle_year?
+        redirect_to candidate_interface_application_form_path
+      elsif @service.can_apply_on_apply? && current_candidate.present?
         current_candidate.update!(course_from_find_id: @course.id)
 
         redirect_to candidate_interface_interstitial_path
@@ -80,6 +82,13 @@ module CandidateInterface
     def apply_on_ucas_or_apply_params
       params.require(:candidate_interface_apply_on_ucas_or_apply_form)
         .permit(:service, :provider_code, :course_code)
+    end
+
+    def candidate_has_application_in_different_recruitment_cycle_year?
+      return false if @course.blank?
+      return false if current_candidate.blank?
+
+      current_candidate.current_application.recruitment_cycle_year != @course.recruitment_cycle_year
     end
   end
 end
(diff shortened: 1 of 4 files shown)


davidgisbey (approved):
Yea I think going light touch is the way to go for now. I think we will also probably want to ensure that the candidates course_from_find_id is set to nil when we carry over or dupe their application. I'll have a bit of a dig to see at what stage the course_from_find_id was set because it seems like it should be set to nil earlier.

steventux:
@davidgisbey Ah I see, so if we worked out if the course was current here https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/controllers/candidate_interface/apply_from_find_controller.rb#L10 we could short circuit the whole thing earlier? ie. do not set the course id and the candidate will be redirected to the new application form? I can amend this PR to do that instead.

steventux:
@davidgisbey see https://github.com/DFE-Digital/apply-for-teacher-training/pull/4219/commits/559a16fc563b9325b7a0849ca697ac274ed63257 for an earlier short-circuit if the course / current application recruitment cycles do not match, I can remove the initial commit in this PR if you think it's unnecessary now. I think either approach works but the latter seems to keep things simpler in terms of redirect hops, what do you think?

davidgisbey (approved):
100%! Thanks for your work on this @steventux. This whole flow is crying out for a refactor.

app/controllers/candidate_interface/apply_from_find_controller.rb
```diff
@@ -81,5 +83,13 @@ def apply_on_ucas_or_apply_params
       params.require(:candidate_interface_apply_on_ucas_or_apply_form)
         .permit(:service, :provider_code, :course_code)
     end
+
+    def candidate_has_application_in_different_recruitment_cycle_year?
+      return false if @course.blank?
+      return false if current_candidate.blank?
+      return false if current_candidate.current_application.blank?
```
duncanjbrown, line 90: Will it ever be blank? The method creates one if it's not present

steventux: @duncanjbrown no you're right it can't be blank, will remove this line.

steventux: Removed/amended https://github.com/DFE-Digital/apply-for-teacher-training/pull/4219/commits/b64223e733f91b9d93fcd02677e31c056e835954

PR #4220, opened 2021-03-04
Add custom I18n exception handler in application.rb 

malcolmbaig:
## Context
We're already raising exceptions for missing locale strings when using
the `t` helper in views. Adding this additional config here will ensure
we also raise exceptions when locale strings for model attributes can't
be resolved.

- Add the exception handler in application.rb 
- Fix a few missing locale string errors

## Guidance to review
- Ok for this to be applied to all environments?
- Do the fixes make sense? 

## Link to Trello card
https://trello.com/c/ouZ2DExb

## Things to check

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -51,6 +51,7 @@ module ApplyForPostgraduateTeacherTraining
 
     config.action_view.default_form_builder = GOVUKDesignSystemFormBuilder::FormBuilder
 
+    config.i18n.exception_handler = Proc.new { |exception| raise exception.to_exception }
     config.action_view.raise_on_missing_translations = true
     config.action_view.form_with_generates_remote_forms = false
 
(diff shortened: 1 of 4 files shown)


stevehook (approved):
This makes sense. I'm not sure about switching this on in production, would be interested to know what others think. If we do so best to do a morning deploy and monitor any errors that may not be covered by tests.

malcolmbaig:
> This makes sense. I'm not sure about switching this on in production, would be interested to know what others think. If we do so best to do a morning deploy and monitor any errors that may not be covered by tests.

Another option might be to just switch this on in QA and monitor for a few days before rolling out to prod.

@duncanjbrown @davidgisbey @cwrw - thoughts on enabling this exception handler in prod?

steventux (approved):
:+1: I'd agree with @stevehook this might need us to pay close attention to Sentry on deployment.

duncanjbrown (approved):
Makes sense to me!

cwrw (approved):
LGTM, I attempted a _questionable_ missing key scan for locale strings and it came back ok, so happy for this to be merged as long as we keep a watchful eye or two on Sentry as suggested.

PR #4221, opened 2021-03-04
Work history error messaging

malcolmbaig:
## Context
Replacement for https://github.com/DFE-Digital/apply-for-teacher-training/pull/4197, accounting for recent changes to how date validations are tested.

From the Trello card:

> The error messages in the work history need to be updated to match the new design.

- Most of the changes described by the Google doc linked in the ticket, [with the exception of](https://docs.google.com/document/d/1zTENV4UewP5jJeRl_jWhto4NEaHWviYyMvetfsR5KQc/edit?disco=AAAAH7fnRS8) the most granular changes that relate solely to the month or solely to the year field.

## Guidance to review
- Running locally or in the review app, ensure restructured work history feature flag is active, and the current application has feature_restructured_work_history set to true. Easiest way to ensure the latter is to sign up as a new user (rather than log in as an existing one).

- Go through the Add Job and Explain Break flows, testing the scenarios outlined in the doc (except the ones where a month-only or year-only error is expected).

## Link to Trello card
https://trello.com/c/pArii32e

## Things to check

### DIFF
--- a/app/validators/date_validator.rb
+++ b/app/validators/date_validator.rb
@@ -7,7 +7,6 @@ class DateValidator < ActiveModel::EachValidator
     return if value.blank? || (blank?(value) && !options[:presence])
 
     date_validations(record, attribute, value) if !record.errors.keys.include?(attribute)
-
     date_of_birth_validations(record, attribute, value) if options[:date_of_birth] && !record.errors.keys.include?(attribute)
   end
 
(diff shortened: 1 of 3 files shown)


app/validators/date_validator.rb
```diff
@@ -19,9 +18,7 @@ def date_of_birth_validations(record, attribute, value)
 
   def date_validations(record, attribute, value)
     return record.errors.add(attribute, :blank_date, article: article(attribute), attribute: humanize(attribute)) if options[:presence] && blank?(value)
-
     return record.errors.add(attribute, :blank_date_fields, attribute: humanize(attribute), fields: blank_fields(value).to_sentence) if !blank?(value) && blank_fields(value).any?
-
```
despo, line 24: I think this makes it easier to read because the lines are quite busy. If you are going to remove them though, you should also do the same on line 23!

despo (approved):
LGTM 👍🏼

PR #4222, opened 2021-03-04
Remove duplicate references from the referee survey export and cleanse data

davidgisbey:
## Context

When we duplicate an application form for apply again or at the end of the cycle, we also duplicate the references. This means that we have quite a few duplicate references in this export.

Additionally, we've currently got invalid data in our system as when I backfilled the `feedback_provided_at` as carried over applications did not have the required audits to populate the values.

This PR does two things. 
- Marks carried over references as duplicates
- Populates missing feedback_provided_at values

## Guidance to review

Per commit is probably the easiest.

## Link to Trello card

https://trello.com/c/CL1ZO70r/3124-investigate-why-there-are-missing-dates-in-referee-survey-extract

## Things to check

### DIFF
--- a/app/models/application_reference.rb
+++ b/app/models/application_reference.rb
@@ -8,6 +8,7 @@ class ApplicationReference < ApplicationRecord
 
   belongs_to :application_form, touch: true
   has_many :reference_tokens, dependent: :destroy
+  has_one :candidate, through: :application_form
 
   audited associated_with: :application_form
 
--- a/app/models/candidate.rb
+++ b/app/models/candidate.rb
@@ -15,6 +15,7 @@ class Candidate < ApplicationRecord
   has_one :ucas_match
   has_many :application_forms
   has_many :application_choices, through: :application_forms
+  has_many :application_references, through: :application_forms
   belongs_to :course_from_find, class_name: 'Course', optional: true
 
   def self.for_email(email)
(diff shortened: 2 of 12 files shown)


stevehook (commented):
Code makes sense but I've got a couple of questions about the approach...

I'm a bit worried about setting `feedback_provided_at` to `nil`. I can't see any code that obviously depends on this attribute being present so maybe it's fine. But does this mean there is data missing on apply again applications? (i.e. do providers/support/candidates need to see when the reference was provided?) Would it be possible to use some other attribute (or even a new dedicated attribute) to indicate that a reference was a duplicate and therefore not to be included in reporting?

davidgisbey:
Hey Steve, the only place it's exposed at the moment is in the support UI. We don't actually show providers or candidates when it was provided. I did think about possibly adding a `duplicated` boolean to the form. I'll have a look into it today.

db/migrate/20210308172728_add_duplicate_boolean_to_references.rb
```diff
@@ -0,0 +1,5 @@
+class AddDuplicateBooleanToReferences < ActiveRecord::Migration[6.0]
+  def change
+    add_column :references, :duplicate, :boolean, default: false
+  end
+end
```
stevehook, line 5: I'm guessing this will go away after rebasing as it's in a separate PR?

stevehook (approved):
This looks great. I've got a few minor suggestions inline.

PR #4223, opened 2021-03-04
Add interview changes export

benswannack:
## Context
To enable analysis on:
- Who is scheduling interviews
- How many interviews are being scheduled
- Are interviews being changed?
- What kind of changes are being made?
- Are interviews being cancelled? 
- Why are they being cancelled?

Add an interview changes export and documentation for it to support

## Guidance to review
Are there any gaps in the export that will make the analysis difficult?

## Link to Trello card
https://trello.com/c/7eXWJiJd/3453-create-interviews-csv-extracts-to-be-accessed-in-support-performance

## Things to check

### DIFF
--- a/app/services/support_interface/interview_changes_export.rb
+++ b/app/services/support_interface/interview_changes_export.rb
@@ -1,0 +1,103 @@
+module SupportInterface
+  class InterviewChangesExport
+    def data_for_export
+      interview_audits.map do |audit|
+        row_for_audit(audit)
+      end
+    end
+
+    def row_for_audit(audit)
+      interview = audit.auditable
+      application_choice = interview.application_choice
+      {
+        audit_id: audit.id,
+        audit_created_at: audit.created_at,
+        audit_type: audit.action,
+        interview_id: interview.id,
+        candidate_id: application_choice.candidate.id,
+        application_form_id: application_choice.application_form.id,
+        provider_code: application_choice.provider.code,
+        provider_user: audit_user(audit),
+        interview_status: interview_status(interview, audit),
+        interview_preferences: application_choice.application_form.interview_preferences,
+        application_submitted_at: application_choice.application_form.submitted_at,
+        course_code: application_choice.course.code,
+        course_location: application_choice.site.name_and_code,
+      }.merge(interview_change(audit))
+    end
+
+  private
+
+    def interview_audits
+      Audited::Audit.where(auditable_type: 'Interview').includes(
+        :user,
+        auditable: {
+          application_choice: %i[application_form course site candidate provider],
+        },
+      )
+    end
+
+    def interview_status(interview, audit)
+      if audit.created_at < interview_time_when_change_made(interview, audit)
+        'upcoming'
+      else
+        'past'
+      end
+    end
+
+    def interview_time_when_change_made(interview, audit)
+      # We want to return the time the interview was scheduled for at the time of the audit.
+      # If the time was changed, then return the original time from the audited_changes.
+      # If the time was not changed, then it is either an upcoming interview (since the validation means you can't make it in the past)
+      # or it is a cancellation and the interview time on the interview model is representative.
+      interview_time_from_audit(audit).presence || interview.date_and_time
+    end
+
+    def interview_time_from_audit(audit)
+      audited_time = audit.audited_changes['date_and_time']
+      if audited_time.is_a?(Array)
+        Time.zone.parse(audited_time.first)
+      elsif audited_time.present?
+        Time.zone.parse(audited_time)
+      end
+    end
+
+    def interview_change(audit)
+      changes = changed_attributes(audit)
+      {
+        date_and_time: change_to_attr(changes, 'date_and_time'),
+        cancelled_at: change_to_attr(changes, 'cancelled_at'),
+        cancellation_reason: change_to_attr(changes, 'cancellation_reason'),
+        provider_id: change_to_attr(changes, 'provider_id'),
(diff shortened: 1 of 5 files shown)


app/services/support_interface/interview_changes_export.rb
```diff
@@ -0,0 +1,93 @@
...
+    end
+
+  private
+
+    def interview_audits
+      Audited::Audit.where(auditable_type: 'Interview')
+    end
+
+    def interview_status(interview, audit)
+      if audit.created_at < interview_time_when_change_made(interview, audit)
```
cwrw, line 41: I found out today that we have access to `audit.revision` which will give us the interview at it's state at that time, so maybe we can use this instead:
```
if audit.created_at < audit.revision.date_and_time
```

benswannack: Oh this is a nice feature, TIL!

benswannack: Looking at the source for this, it will go through the entire audit history up to the point of this audit to reconstruct the object. That is less efficient than my method, but perhaps more readable. Given this is an export and any performance issues would only affect support users, shall we go with the readable one?

cwrw: spoke in person but for posterity: we're going to avoid revision for now as there is more work to be done to investigate building a performant solution.

app/services/support_interface/interview_changes_export.rb
```diff
@@ -0,0 +1,93 @@
...
+    def interview_change(audit)
+      changes = changed_attributes(audit)
+      {
+        'date_and_time' => '',
+        'cancelled_at' => '',
+        'cancellation_reason' => '',
+        'provider_id' => '',
+        'location' => '',
+        'additional_details' => '',
+      }.map { |key, value| [key, (changes.keys.include?(key) ? changes[key] : value)] }.to_h
```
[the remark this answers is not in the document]
benswannack: The intention of this was to distinguish between "changed to have a value of `nil`" vs, "not changed". We could return something in place of the `''` like a `-` if that's preferable?

app/services/support_interface/interview_changes_export.rb
```diff
@@ -0,0 +1,93 @@
...
+      end
+    end
+
+    def interview_change(audit)
+      changes = changed_attributes(audit)
+      {
+        'date_and_time' => '',
+        'cancelled_at' => '',
+        'cancellation_reason' => '',
+        'provider_id' => '',
```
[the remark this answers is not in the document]
benswannack: This is what is in the spec, since this is the column name in the interview model and what is actually being changed. Since we are interested in what has changed, I don't think we really need the code. This also avoids having to load the provider code by ID for each audit.

I'm open to changing it, but can't see a strong reason to yet

app/services/support_interface/interview_changes_export.rb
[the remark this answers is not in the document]
benswannack: Updated without changing the behaviour, what do you think now?

app/services/support_interface/interview_changes_export.rb
```diff
@@ -0,0 +1,93 @@
+module SupportInterface
+  class InterviewChangesExport
+    def data_for_export
+      interview_audits.map do |audit|
```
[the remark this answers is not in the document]
benswannack: I have added the includes, but I missed the batching comment. I'll add that now too

app/services/support_interface/interview_changes_export.rb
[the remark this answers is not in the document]
cwrw: Fair enough, as long as it's clear and there is documentation I think we can keep as is.

app/services/support_interface/interview_changes_export.rb
```diff
@@ -0,0 +1,100 @@
...
+        course_location: application_choice.site.name_and_code,
+      }.merge(interview_change(audit))
+    end
+
+  private
+
+    def interview_audits
+      Audited::Audit.where(auditable_type: 'Interview').includes(
+        auditable: {
+          application_choice: %i[application_form course site candidate],
```
cwrw, line 34: Could you also add `user: :provider_user` to the includes and also `provider` to `application_choice`? I think this way we cover all n+1s

app/services/support_interface/interview_changes_export.rb
```diff
@@ -0,0 +1,100 @@
...
+      else
+        ''
+      end
+    end
+
+    def audit_user(audit)
+      if audit.user_type == 'SupportUser'
+        'Support'
+      else
+        audit.user.email_address
```
cwrw, line 96: Is it realistic we can get a `nil` user in production on an audit? As locally I have many examples of `nil` users on the audit

cwrw (approved):
Looks good, just some extra includes and guarding against `nil` users if possible

app/services/support_interface/interview_changes_export.rb
benswannack: I think it's much less likely in prod (if it's possible at all). I'll cover this with a `user&.email_address`

duncanjbrown: Just going to have a look on blazer...

duncanjbrown: Yes — we have `nil` users whenever something's updated by an automated process. Possibly worth handling that here?

benswannack: I don't think there are any automated processes which edit interviews at the moment, but the case is now covered

PR #4225, opened 2021-03-05
Make OpenProviderCourses audit the changes it makes

benswannack:
## Context
Currently we do not have a record for when courses were opened on apply using this service.

Change from using update_all which skips callbacks, to looping through and calling update on each course record.

## Guidance to review
Have I missed a nicer way of doing this?

## Link to Trello card
https://trello.com/c/r9aaU0Jm/3466-openprovidercourses-does-not-audit-its-changes-to-openonapply

## Things to check

### DIFF
--- a/app/services/open_provider_courses.rb
+++ b/app/services/open_provider_courses.rb
@@ -6,7 +6,7 @@ class OpenProviderCourses
   end
 
   def call
-    if run_courses.or(ratified_courses).update_all(open_on_apply: true).positive?
+    if run_courses.or(ratified_courses).update(open_on_apply: true).any?
       provider.provider_users.each do |provider_user|
         ProviderMailer.courses_open_on_apply(provider_user)
       end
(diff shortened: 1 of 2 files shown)


app/services/open_provider_courses.rb
```diff
@@ -6,7 +6,12 @@ def initialize(provider:)
   end
 
   def call
-    if run_courses.or(ratified_courses).update_all(open_on_apply: true).positive?
+    courses_to_open = run_courses.or(ratified_courses)
```
cwrw, line 9: What are your thoughts around using `run_courses.or(ratified_courses).update(name: nil, open_on_apply: true)` instead, which does run callbacks/validations, to avoid having to update each course in turn?

benswannack: Ah, thanks! Did not realise that was a thing!

cwrw (approved):

duncanjbrown (approved):
:+1:

PR #4227, opened 2021-03-05
Update Before You Start view link

george-bit:
## Context

In the case that user proceeds from the before you start page there is a duplication of information. This is not the case if proceeding from the application form. Therefore, this PR updates the link on that to redirect in this situation.

This PR updates the link.

View to be removed:

image

## Guidance to review

Test in review app, let me know if there's any unintended knock on to this approach or any refactor that could happen controller wise.

## Link to Trello card

https://trello.com/c/b22m649k/2947-add-a-course-first-flow-shows-the-same-guidance-twice

## Things to check

### DIFF
--- a/app/views/candidate_interface/unsubmitted_application_form/before_you_start.html.erb
+++ b/app/views/candidate_interface/unsubmitted_application_form/before_you_start.html.erb
@@ -9,7 +9,7 @@
 
     <hr class="govuk-section-break govuk-section-break--m">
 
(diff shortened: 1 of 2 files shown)


paulrobertlloyd (changes requested):
Sorry, perhaps the Trello card wasn’t clear enough. We should keep the all the guidance that we currently have (no views should be deleted) but, change some of the logic around when we show them.

* Dual running interstitial – logic stays the same
* Course choices guidance – don’t show if candidate clicks on ‘Choose a course’ from the dual running interstitial

Therefore, I think this issue requires only changing the path on the ‘Choose a course’ link.

paulrobertlloyd (approved):
👌

PR #4228, opened 2021-03-05
Set up models and services for granular provider notification settings

AgaDufrat:
## Context

As a provider user 
I need the ability to choose the type of email notification I receive
So that I only get notified about the things I need

- add a feature flag
- add ProviderUserNotification model
- update NotificationsList service
- add a service to update all notification events
- update email sending services to account for different type of preferences

No UI changes.

## Guidance to review

Initially wanted to call the table 'provider_user_notification_preferences' but index name was too long.

To test:
- turn on the feature flag `FeatureFlag.activate(:configurable_provider_notifications)`
- create ProviderUserNotification  in the rails console and change it via provider interface and support interface
- check the values in the rails console

## Link to Trello card

https://trello.com/c/AtNkf165/3438-%F0%9F%8F%88-set-up-models-and-services-for-granular-provider-notification-settings

## Things to check

### DIFF
--- a/app/models/provider_user.rb
+++ b/app/models/provider_user.rb
@@ -4,6 +4,7 @@ class ProviderUser < ActiveRecord::Base
   has_many :provider_permissions, dependent: :destroy
   has_many :providers, through: :provider_permissions
   has_many :notes, dependent: :destroy
+  has_one :notification_preferences, class_name: 'ProviderUserNotificationPreferences'
   attr_accessor :impersonator
 
   validates :dfe_sign_in_uid, uniqueness: true, allow_nil: true
--- a/app/services/notifications_list.rb
+++ b/app/services/notifications_list.rb
@@ -1,9 +1,27 @@
 class NotificationsList
-  def self.for(application_choice, include_ratifying_provider: false)
-    return application_choice.provider.provider_users.where(send_notifications: true) if application_choice.accredited_provider.nil? || !include_ratifying_provider
+  NOTIFICATION_PREFERENCE_NAMES_FOR_EVENTS = {
+    application_received: %i[application_submitted chase_provider_decision],
+    application_withdrawn: %i[application_withdrawn],
+    application_rejected_by_default: %i[application_rejected_by_default],
+    offer_accepted: %i[offer_accepted unconditional_offer_accepted],
+    offer_declined: %i[declined declined_by_default],
+  }.freeze
 
-    application_choice.provider.provider_users.where(send_notifications: true).or(
-      application_choice.accredited_provider.provider_users.where(send_notifications: true),
-    )
+  def self.for(application_choice, include_ratifying_provider: false, event: nil)
+    if FeatureFlag.active?(:configurable_provider_notifications)
+      notification_name = NOTIFICATION_PREFERENCE_NAMES_FOR_EVENTS.select { |k, v| k if event.in? v }.keys.first
+      raise 'Undefined type of notification event' unless ProviderUserNotificationPreferences.notification_preference_exists?(notification_name)
+
+      return application_choice.provider.provider_users.joins(:notification_preferences).where("#{notification_name} IS true") if application_choice.accredited_provider.nil? || !include_ratifying_provider
+
+      application_choice.provider.provider_users.or(application_choice.accredited_provider.provider_users)
+        .joins(:notification_preferences).where("#{notification_name} IS true")
+    else
+      return application_choice.provider.provider_users.where(send_notifications: true) if application_choice.accredited_provider.nil? || !include_ratifying_provider
+
+      application_choice.provider.provider_users.where(send_notifications: true).or(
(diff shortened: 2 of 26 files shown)


app/services/notifications_list.rb
```diff
@@ -1,9 +1,20 @@
...
-  def self.for(application_choice, include_ratifying_provider: false)
-    return application_choice.provider.provider_users.where(send_notifications: true) if application_choice.accredited_provider.nil? || !include_ratifying_provider
+  attr_reader :event
 
-    application_choice.provider.provider_users.where(send_notifications: true).or(
-      application_choice.accredited_provider.provider_users.where(send_notifications: true),
-    )
+  def self.for(application_choice, include_ratifying_provider: false, event: nil)
+    if FeatureFlag.active?(:configurable_provider_notifications)
+      raise 'Undefined type of notification event' if event.nil? || !ProviderUserNotification.method_defined?(event)
```
duncanjbrown, line 6: This binds us 1:1 event->notification type — is that ok? (if so, probably no need to change this for an explicit mapping at this stage)

app/models/provider_user_notification.rb
```diff
@@ -0,0 +1,3 @@
+class ProviderUserNotification < ActiveRecord::Base
```
duncanjbrown, line 1: Would be good to have `ProviderUserNotificationPreferences`! It reads a bit like it's a notification itself. I think you can do something like `self.table_name = :provider_user_notifications`. Or alternatively manually create the index in the migration and give it a shorter name there.

app/services/update_all_notification_preferences.rb
```diff
@@ -0,0 +1,20 @@
+class UpdateAllNotificationPreferences
```
duncanjbrown, line 1: This is such a thin class, I'd consider adding methods straight to the model?

```ruby
class ProviderUserNotificationPreferences
  belongs_to :provider_user
  NOTIFICATION_PREFERENCES = [:application_received...].freeze

  def update_all_preferences(value)
    NOTIFICATION_PREFERENCES.each { |n| update(n => value) }
  end

  def self.notification_preference_exists?(notification_name)
    NOTIFICATION_PREFERENCES.include? notification_name
  end
end
```

app/models/provider_user.rb
```diff
@@ -4,6 +4,7 @@ class ProviderUser < ActiveRecord::Base
   has_many :provider_permissions, dependent: :destroy
   has_many :providers, through: :provider_permissions
   has_many :notes, dependent: :destroy
+  has_one :provider_user_notification
```
duncanjbrown, line 7: Maybe call this association `notification_preferences`?

spec/services/withdraw_application_spec.rb
```diff
@@ -41,6 +41,54 @@
       expect(ratifying_provider_email['rails-mail-template'].value).to eq('application_withdrawn')
     end
 
+    context 'when the configurable provider notifications feature flag is off' do
```
duncanjbrown, line 44: Can we lose the above spec as this effectively duplicates it?

duncanjbrown (commented):
Looking good!

app/services/notifications_list.rb
AgaDufrat: I think `notification_type` may be a better name for this arg. 
For users with `application_received: true` we send up to 2 emails (application received and reminder email to the training providers), similarly for `offer_declined: true` there are 2 possible events (offer declined by the candidate and offer decline by default).

duncanjbrown: Sure. Do we want to map in here or have the callers effectively do the mapping when they pass the `notification_type`?

spec/services/accept_offer_spec.rb
```diff
@@ -27,20 +27,44 @@
...
-      application_choice = create(:application_choice, :with_offer, course_option: course_option)
+        ratifying_provider = create(:provider)
+        ratifying_provider_user = create(:provider_user, send_notifications: true, providers: [ratifying_provider])
 
-      expect { described_class.new(application_choice: application_choice).save! }.to change { ActionMailer::Base.deliveries.count }.by(3)
-      expect(ActionMailer::Base.deliveries.first.subject).to match(/has accepted your offer/)
-      expect(ActionMailer::Base.deliveries.first.to).to eq [training_provider_user.email_address]
-      expect(ActionMailer::Base.deliveries.second.to).to eq [ratifying_provider_user.email_address]
+        course_option = course_option_for_accredited_provider(provider: training_provider, accredited_provider: ratifying_provider)
+        application_choice = create(:application_choice, status: :offer, course_option: course_option)
```
duncanjbrown, line 41: Why does this need to change? `with_offer` ought to be more realistic?

duncanjbrown (approved):
Great! One last (minor) q inline

duncanjbrown (approved):

PR #4229, opened 2021-03-05
Remove redundant view component configuration

despo:
Followup to #4186


AgaDufrat (approved):

duncanjbrown:
Interestingly, I tried to cause validation errors on the models when I was removing the `validates` directives and it made no difference. So this code must not have been working anyway 🤷

duncanjbrown (approved):

PR #4236, opened 2021-03-08
Remove V3 API (1/4) — Apply from Find uses public API

duncanjbrown:
### Removing V3 from Apply

**1. Apply from Find uses public API** 👈
2. Consume subject codes from the public API and remove remaining references to the v3 API
3. Delete all TTAPI v3 code
4. Clean up after removing v3 API

## Context

The only place in the app we use the v3 Teacher Training API is the [apply-from-find page](https://qa.apply-for-teacher-training.service.gov.uk/candidate/apply?providerCode=1CS&courseCode=2DGY). After this PR, that uses the new public/v1 API.

- Bring the `TeacherTrainingPublicAPI` models up to scratch to support this, porting code where necessary from V3, ie methods on `Course`.
- update stubs and tests to support this new API call
- hook up `ApplyFromFindPage` to use the new API

## Guidance to review

As the `Course` response from the Public API doesn't include the provider code, I added it as an `attr` on the model. Technically we ought to `include=provider` on the request but that made the stubbing more complicated — enough so that I preferred to do this.

## Link to Trello card

https://trello.com/c/io8acUsQ/291-sync-all-courses

## Things to check

### DIFF
--- a/app/models/teacher_training_public_api/course.rb
+++ b/app/models/teacher_training_public_api/course.rb
@@ -21,11 +21,11 @@ module TeacherTrainingPublicAPI
     end
 
     def sites
-      Location.where(
+      @sites_cache ||= Location.where(
         year: ::RecruitmentCycle.current_year,
         provider_code: provider_code,
         course_code: code,
-      ).includes(:location_status).paginate(per_page: 500)
+      ).includes(:location_status).paginate(per_page: 500).all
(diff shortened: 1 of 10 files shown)


raamSoftwire (approved):
Thanks, this looks good Duncan. The comments are really helpful

PR #4238, opened 2021-03-08
Create export for autosuggest usage

TobyRet:
## Context

```
As a... Designer
I need to... understand how candidates use the autosuggest options within APPLY
So that... we can understand how the structured data is captured
```

The candidate application uses two types of autofill:
- autocomplete (user must choose from a range of options)
- autosuggest (user has the option to choose from a range of options or submit their own value)

We're interested in the latter, particularly values submitted by the user as 'freetext'

This PR add a downloadable export that breaks down the values submitted via each autosuggest field in the candidate application.

## Guidance to review
- Go the exports dashboard `/support/performance/data-exports/new` and select 'Candidate autosuggest usage'
- I've tried to memoise the db calls where possible. Are there any further optimisations required?

## Link to Trello card

https://trello.com/c/69AeFZwg/3116-data-request-autofill-usage

## Things to check

### DIFF
--- a/app/services/support_interface/candidate_autofill_usage_export.rb
+++ b/app/services/support_interface/candidate_autofill_usage_export.rb
@@ -1,0 +1,138 @@
+module SupportInterface
+  class CandidateAutofillUsageExport
+    def data_for_export
+      degree_grade_output +
+        degree_institution_output +
+        degree_subject_output +
+        degree_type_output +
+        other_grade_output +
+        other_type_output
+    end
+
+  private
+
+    def degree_grade_output
+      create_output(
+        level: :degree,
+        attribute: :grade,
+        field_name: 'Degree grade',
+        set_to_check: degree_grades,
+      )
+    end
+
+    def degree_institution_output
+      create_output(
+        level: :degree,
+        attribute: :institution_name,
+        field_name: 'Degree institution',
+        set_to_check: degree_institutions,
+      )
+    end
+
+    def degree_subject_output
+      create_output(
+        level: :degree,
+        attribute: :subject,
+        field_name: 'Degree subject',
+        set_to_check: degree_subjects,
+      )
+    end
+
+    def degree_type_output
+      create_output(
+        level: :degree,
+        attribute: :qualification_type,
+        field_name: 'Degree type',
+        set_to_check: degree_types,
+      )
+    end
+
+    def other_grade_output
+      create_output(
+        level: :other,
+        attribute: :grade,
+        field_name: 'Other grade',
+        set_to_check: other_grades,
+      )
+    end
+
+    def other_type_output
+      create_output(
+        level: :other,
+        attribute: :qualification_type,
+        field_name: 'Other type',
+        set_to_check: other_types,
+      )
+    end
+
+    def create_output(level:, attribute:, field_name:, set_to_check:)
+      counts_for(level: level, attribute: attribute)
+        .map do |value, count|
+        {
+          'Field' => field_name,
+          'Value entered' => value,
+          'Frequency' => count,
+          'Free text?' => !set_to_check.include?(value),
+        }
+      end
+    end
+
+    def counts_for(level:, attribute:)
+      case level
+      when :degree
+        degree_qualifications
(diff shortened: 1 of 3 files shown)


app/services/support_interface/candidate_autofill_usage_export.rb
```diff
@@ -0,0 +1,157 @@
...
+    def other_type_output
+      counts_for(level: :other, attribute: :qualification_type)
+        .map do |type, count|
+        {
+          'Field' => 'Other type',
+          'Value entered' => type,
+          'Frequency' => count,
+          'Free text?' => !other_types.include?(type),
+        }
+      end
```
raamSoftwire, line 23: This block is repeated 6 times. I wonder if we could pull out a method here 

```
    def output(level, attribute, field_name, set_to_check)
      counts_for(level: level, attribute: attribute)
          .map do |attribute, count|
        {
            'Field' => field_name,
            'Value entered' => attribute,
            'Frequency' => count,
            'Free text?' => !set_to_check.include?(attribute),
        }
      end
    end

    def other_type_output
      output(:other, :qualification_type, 'Other type', other_types)
    end

    def other_grade_output
      output(:other, :grade, 'Other grade', other_grades)
    end
    
    ```

File: app/services/support_interface/candidate_autofill_usage_export.rb [hunk 2]
```diff
@@ -0,0 +1,157 @@
...
+          level: 'other',
+          application_forms: {
+            phase: 'apply_1',
+            recruitment_cycle_year: RecruitmentCycle.current_year,
+          },
+        )
+        .all
+    end
+
+    def free_text?(grade)
```
raamSoftwire, line 125: Is this ever used?

app/services/support_interface/candidate_autofill_usage_export.rb
```diff
@@ -0,0 +1,157 @@
...
+      !all_possible_grades.include?(grade)
+    end
+
+    def all_possible_grades
```
raamSoftwire, line 129: If `free_text?` is not used or required, then I think we can remove this too.

app/services/support_interface/candidate_autofill_usage_export.rb
```diff
@@ -0,0 +1,157 @@
...
+        degree_institution_output +
+        degree_subject_output +
+        degree_type_output +
+        other_grade_output +
+        other_type_output
+    end
+
+  private
+
+    def other_type_output
```
raamSoftwire, line 14: I think it would read better if these methods followed the same order that they are called in above. Or reverse order even, but something systematic.

app/services/support_interface/candidate_autofill_usage_export.rb
TobyRet: Yup good idea 👍

app/services/support_interface/candidate_autofill_usage_export.rb
TobyRet: Nope - I'll remove

app/services/support_interface/candidate_autofill_usage_export.rb
TobyRet: I'll remove

app/services/support_interface/candidate_autofill_usage_export.rb
TobyRet: 👍

spec/services/support_interface/candidate_autofill_usage_export_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+            degree_qualification: degree_qualification,
+            other_qualification: other_qualification,
+            free_text: false,
+          ),
+        )
+      end
+    end
+
+    context 'Phase 2 applications' do
+      it "does not return a has of candidates' autofill info" do
```
raamSoftwire, line 27: hash*

raamSoftwire (approved):
Just a few small refactors, but it all looks great. Cheers Toby!

PR #4241, opened 2021-03-08
Add reusable component for selecting a course.

richardpattinson:
## Context

Adding a new component for selecting courses as per the reusable pattern.

## Guidance to review

None

## Link to Trello card

https://trello.com/c/0b3x0EvO/3472-select-course-component
## Things to check

### DIFF
--- a/app/components/provider_interface/select_course_component.html.erb
+++ b/app/components/provider_interface/select_course_component.html.erb
@@ -1,0 +1,8 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with model: form_object, url: form_path, method: :post do |f| %>
+      <%= f.govuk_collection_radio_buttons :course_id, courses, :id, :name_and_code, :description, legend: { text: page_title, size: 'l' } %>
+      <%= f.govuk_submit t('continue') %>
+    <% end %>
+  </div>
+</div>
--- a/app/components/provider_interface/select_course_component.rb
+++ b/app/components/provider_interface/select_course_component.rb
@@ -1,0 +1,14 @@
+module ProviderInterface
+  class SelectCourseComponent < ViewComponent::Base
+    include ViewHelper
+
+    attr_reader :form_object, :form_path, :page_title, :courses
+
+    def initialize(form_object:, form_path:, courses:, page_title: 'Select course')
+      @form_object = form_object
+      @form_path = form_path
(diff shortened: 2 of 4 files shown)


app/components/provider_interface/select_course_component.html.erb
```diff
@@ -0,0 +1,8 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with model: form_object, url: form_path, method: :post do |f| %>
```
benswannack, line 3: What is the plan for validating that at least one option is selected (and displaying an error summary)?

app/components/provider_interface/select_course_component.rb
```diff
@@ -0,0 +1,14 @@
+module ProviderInterface
+  class SelectCourseComponent < ViewComponent::Base
+    include ViewHelper
+
+    attr_reader :form_object, :form_path, :page_title, :courses, :course_id
```
benswannack, line 5: Is the course_id attr_reader used anywhere?

benswannack (commented):
Looking good, a couple of questions inline

app/components/provider_interface/select_course_component.html.erb
despo: We plan to test that on the form object, but not until we start working on the next card.

app/components/provider_interface/select_course_component.rb
despo: It's used https://github.com/DFE-Digital/apply-for-teacher-training/pull/4241/files#diff-58c92cf3b3dacb6de8b03a18d15bef170341967100187863c5bbfdbd4944e793R4 to set and read the selected course.

benswannack: Ah is that not done inside the form_object (with the attr_accessor)?

despo: Ah you are right. We'll remove

cwrw (commented):
I noticed that a pattern has evolved here, we would like to render a collection of radio buttons for an object of either locations/providers/courses etc. Could we have one component instead that accepts a collections, what are your thoughts?

despo:
@cwrw good point. New PR introducing reusable collection select component and removing SelectProvider one:
https://github.com/DFE-Digital/apply-for-teacher-training/pull/4256

PR #4245, opened 2021-03-09
Don’t attempt to sync courses when no provider can be found

despo:
### Why? 
We are getting around 1K false errors per week posted on sentry, making it extremely difficult to identify and resolve actual issues.

Resolves sentry error:
https://sentry.io/organizations/dfe-bat/issues/2220591200/?project=1765973&query=is%3Aunresolved

> JsonApiClient::Errors::NotFound
> Couldn't find resource at: https://api.publish-teacher-training-courses.service.gov.uk/api/public/v1/recruitment_cycles/2021/providers/nil

### DIFF
--- a/app/services/teacher_training_public_api/sync_courses.rb
+++ b/app/services/teacher_training_public_api/sync_courses.rb
@@ -6,7 +6,7 @@ module TeacherTrainingPublicAPI
     sidekiq_options retry: 3, queue: :low_priority
 
     def perform(provider_id, recruitment_cycle_year, run_in_background: true)
-      @provider = ::Provider.find(provider_id)
+      @provider = ::Provider.find_by!(id: provider_id)
       @run_in_background = run_in_background
 
       TeacherTrainingPublicAPI::Course.where(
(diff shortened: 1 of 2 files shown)


app/services/teacher_training_public_api/sync_courses.rb
```diff
@@ -6,7 +6,7 @@ class SyncCourses
     sidekiq_options retry: 3, queue: :low_priority
 
     def perform(provider_id, recruitment_cycle_year, run_in_background: true)
-      @provider = ::Provider.find(provider_id)
+      @provider = ::Provider.find_by!(id: provider_id)
```
duncanjbrown, line 9: `find` will throw a `RecordNotFound` exception — no need to change to `find_by!`

duncanjbrown:
Why are we getting invalid `provider_id`s here? Shouldn't `SyncProvider#create_or_update_provider` reliably give us providers in the database?

despo:
@duncanjbrown as you correctly pointed out find would have thrown an exception if the provider did no exist for the give id, so we would be seeing RecordNotFound errors rather than API sync ones. Also I'm now realising that the failure is on another endpoint 

> Couldn't find resource at: https://api.publish-teacher-training-courses.service.gov.uk/api/public/v1/recruitment_cycles/2021/providers/nil

so not caused by the call on line 12. 

Back to the drawing board...

PR #4249, opened 2021-03-09
Find changed apply again applications

stevehook:
## Context

We need to work out a way to determine (add count) which applications have been changed in significant ways when a candidate uses apply again. This PR introduces a service to do this based on the spike https://github.com/DFE-Digital/apply-for-teacher-training/pull/4216/files 

Add `FindChangedApplyAgainApplications` service. This can be used to
count the number of candidates that have applied again and changed their
applications in a significant way. For our purposes significant means
any of the following:

- Updated `becoming_a_teacher` (personal statements)
- Updated `subject_knowledge`
- Updated/added/deleted any qualifications

## Guidance to review

- We just use `as_json` to extract the attributes and associations that are considered significant when changed. This makes the definition of those 'significant things' more declarative and easier to change.
- This is a fairly naive implementation that relies on a Ruby `select` to filter the list of apply again applications. I don't think it would be easy to do this in a database query so some of the logic is happening in memory. We think this will be fine given that the numbers of apply again applications isn't huge. Using `lazy` enumerator helps.

## Link to Trello card

https://trello.com/c/0fMYPksT/2862-dev-use-govuk-view-components-in-find-to-replace-accordion-markup

## Things to check

### DIFF
--- a/app/services/candidate_interface/find_changed_apply_again_applications.rb
+++ b/app/services/candidate_interface/find_changed_apply_again_applications.rb
@@ -11,7 +11,7 @@ module CandidateInterface
     end
 
     def all_candidate_count
-      all_forms.map(&:candidate_id).uniq.count
+      all_forms.select(:candidate_id).distinct.count
     end
 
     def changed_candidate_count


app/services/candidate_interface/find_changed_apply_again_applications.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+
+    def changed_forms
+      apply_again_forms.find_each.lazy.select do |application_form|
+        changed?(application_form)
+      end
+    end
+
+    def all_candidate_count
+      all_forms.map(&:candidate_id).uniq.count
```
duncanjbrown, line 14: As this is a relation, could we `pluck` rather than `map`?

duncanjbrown (commented):
Looks good, and I think the specs are convincing!

app/services/candidate_interface/find_changed_apply_again_applications.rb
```diff
@@ -0,0 +1,48 @@
...
+    def changed?(application_form)
+      significant_values(application_form) !=
+        significant_values(application_form.previous_application_form)
+    end
+
+    def significant_values(application_form)
+      application_form.as_json(
+        only: %i[becoming_a_teacher subject_knowledge],
+        include: {
+          application_qualifications: {
```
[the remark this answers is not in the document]
stevehook: Good question. I think it's OK because `Hash#==` ignores order (even though hashes are ordered in Ruby). If they were arrays we'd be in trouble.

```
pry(main)> { b: 'B', a: 'A', c: { d: 'D', e: 'E' } } == { a: 'A', c: { e: 'E', d: 'D' }, b: 'B' }
=> true
```

app/services/candidate_interface/find_changed_apply_again_applications.rb
stevehook: Yes good point. I've gone with `select(:candidate_id).distinct.count` so that we don't load anything into memory other than the end result. Does that make sense?

duncanjbrown (approved):
Looks good to me!

PR #4254, opened 2021-03-09
Refactor APIDocs namespace so we can reuse it for new APIs

duncanjbrown:
## Context

We have an APIDocs namespace which so far has been used only for Vendor API docs.

We've already added the Data API for TAD and it's likely we'll add an API for Register in the near future. 

This PR lays groundwork for reusing our API documentation tooling for those APIs. A follow-up will add the Data API documentation itself.

- Convert all the APIDocs partials to components in the APIDocs namespace so we can reuse them (and if we like, test them)
- Namespace all the Vendor API docs under APIDocs

## Guidance to review

It's quite a big PR, but this shouldn't change behaviour at all 🤞 . Everything under `/api-docs` should still work as before.

## Things to check


cwrw (approved):
LGTM 👍  
Could you add specs where relevant and previews for the new components?

duncanjbrown:
> LGTM 👍
> Could you add specs where relevant and previews for the new components?

Thanks! Will add :1:

PR #4257, opened 2021-03-09
Qualifications export: Don't load all application choices in memory at once

TobyRet:
## Context

- The qualifications export appears to hang in production, and remains "in progress" without completing.
- The qualifications export is loading all application choices in memory and we suspect this is causing the background worker to run out of memory on production when the export is generated

- Load application choices in batches (via `#find_each` ) 

## Guidance to review

- Unfortunately, we need to deploy out to prod to see if this has worked or not!
- Thanks @stevehook for helping to debug this!

## Link to Trello card

https://trello.com/c/LkbuvHIv/3137-qualifications-export-not-completing-on-prod

## Things to check

### DIFF
--- a/app/services/support_interface/qualifications_export.rb
+++ b/app/services/support_interface/qualifications_export.rb
@@ -2,10 +2,10 @@ module SupportInterface
   class QualificationsExport
     def data_for_export
       application_choices = ApplicationChoice
-                                .select(:id, :application_form_id, :rejection_reason, :structured_rejection_reasons, :status, :course_option_id)
-                                .includes(:course_option, :course, :provider)
+        .select(:id, :application_form_id, :rejection_reason, :structured_rejection_reasons, :status, :course_option_id)
+        .includes(:course_option, :course, :provider)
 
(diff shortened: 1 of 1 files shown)


app/services/support_interface/qualifications_export.rb
```diff
@@ -5,7 +5,9 @@ def data_for_export
                                 .select(:id, :application_form_id, :rejection_reason, :structured_rejection_reasons, :status, :course_option_id)
                                 .includes(:course_option, :course, :provider)
 
-      application_choices.map do |application_choice|
+      data_for_export = []
+
+      application_choices.find_each(batch_size: 100) do |application_choice|
```
stevehook, line 10: I think you can use `find_each(batch_size: 100).lazy.map` if you want to avoid the array concatenation.

stevehook (approved):
🚢

app/services/support_interface/qualifications_export.rb
TobyRet: 💯 Yep that works!

PR #4259, opened 2021-03-09
Modify application filters in Manage to use the offered course

mnacos:
## Context

Application filters in Manage currently use the original course details when filtering by provider, recruitment cycle or site, which means the filters will not work as they should when the offered course details are different (e.g. the provider has modified the location, or has even transferred the application to a different provider).

Besides, course/provider info presented alongside each application for applications in `offer` state and beyond is only partially accurate when the provider has modified the course details.

image

This PR creates a `application_choices_with_courses` view which encapsulates the two-way joins previously done by `GetApplicationChoicesForProviders` and can be used to retrieve application choices annotated with original vs. current course information spanning providers, courses, sites and year information. These extra columns can then be referenced by the filters to ensure we filter on information derived by the current course, i.e. either the original course or the course currently offered.

[TO DO] Update the list item component to show course details from the currently offered course everywhere (e.g. show the correct ratifying provider).

## Guidance to review

Make or modify an offer in Manage, then use the filters to filter by provider, ratifying provider or location based on the new course details.

Should filtering by ratifying provider also include self-ratified courses run by the selected provider?

## Link to Trello card

https://trello.com/c/mwF7TeXg

## Things to check


cwrw:
After discussion we have moved towards a different approach: adding a new current course column which will help resolve issues at the root, closing this for now.

PR #4261, opened 2021-03-09
Refactor ApplyFromFindController and ApplyFromFindPage, fix bug

duncanjbrown:
## Context

We caused exceptions in `ApplyFromFindController` because of the difference between courses from the database and courses from the API.

Restore logic reverted in #4260.

This logic was difficult to test, being in the controller, so pull it into `ApplyFromFindPage` and test it there, then fix the bug.

The params for this controller are still a bit weird and inconsistent, as are params elsewhere in the onboarding flow — subsequent PR will fix!

## Link to Trello card

https://ukgovernmentdfe.slack.com/archives/CN1MCQCHZ/p1615304965080900

## Things to check

### DIFF
--- a/app/services/candidate_interface/apply_from_find_page.rb
+++ b/app/services/candidate_interface/apply_from_find_page.rb
@@ -9,7 +9,7 @@ module CandidateInterface
     def candidate_has_application_in_wrong_cycle?
       return false if course.blank? || @current_candidate.blank?
 
-      @current_candidate.current_application.recruitment_cycle_year != course.recruitment_cycle_year
+      @current_candidate.current_application.recruitment_cycle_year != RecruitmentCycle.current_year
     end
 
     def course_available_on_apply_and_candidate_signed_in?
--- a/app/services/candidate_interface/apply_from_find_page.rb
+++ b/app/services/candidate_interface/apply_from_find_page.rb
@@ -12,9 +12,20 @@ module CandidateInterface
       @current_candidate.current_application.recruitment_cycle_year != course.recruitment_cycle_year
     end
 
+    def course_available_on_apply_and_candidate_signed_in?
+      course_available_on_apply? && @current_candidate.present?
+    end
+
+    def course_available_on_apply_and_candidate_not_signed_in?
+      course_available_on_apply? && @current_candidate.blank?
+    end
+
+    def course_available_on_apply_and_provider_not_on_ucas?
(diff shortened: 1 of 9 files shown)


davidgisbey (approved):
This is way better. Thanks for sorting this. It's been on my todo list forever

app/services/candidate_interface/apply_from_find_page.rb
```diff
@@ -1,43 +1,65 @@
...
+    def course
+      @_course ||= load_course
     end
 
   private
 
-    def pilot_open?
-      FeatureFlag.active?('pilot_open')
+    def course_in_apply_database?
+      course.present? && !(course.is_a? TeacherTrainingPublicAPI::Course)
```
malcolmbaig, line 48:
```suggestion
      course.present? && course.is_a? Course
```

spec/services/candidate_interface/apply_from_find_page_spec.rb
```diff
@@ -3,26 +3,116 @@
...
-        expect(service.course_on_find?).to be_truthy
-        expect(service.course).to eq(course)
+  before do
+    FeatureFlag.activate(:pilot_open)
+    create(:course, :open_on_apply, code: 'ABC1', provider: create(:provider, code: 'ABC'))
+  end
+
+  describe '#candidate_has_application_in_wrong_cycle?' do
+    context 'when the course is in the Apply database already' do
+      it 'is correct' do
```
malcolmbaig, line 13: `is true` might be clearer here

spec/services/candidate_interface/apply_from_find_page_spec.rb
```diff
@@ -3,26 +3,116 @@
...
-
-        expect(service.can_apply_on_apply?).to be_truthy
-        expect(service.course_on_find?).to be_truthy
-        expect(service.course).to eq(course)
+  before do
+    FeatureFlag.activate(:pilot_open)
+    create(:course, :open_on_apply, code: 'ABC1', provider: create(:provider, code: 'ABC'))
+  end
+
+  describe '#candidate_has_application_in_wrong_cycle?' do
```
malcolmbaig, line 11: Worth adding a spec here for the `false` scenario too?

malcolmbaig (approved):
Nice! 👌🏼 

A few suggestions for clarity left inline, but otherwise LGTM.

duncanjbrown:
Thanks both — have adopted all your suggestions @malcolmbaig

PR #4266, opened 2021-03-10
Apply again metrics for Feature Metrics dashboard 2 (part 1)

stevehook:
## Context

We are adding a new set of metrics to the feature metrics dashboard. The first batch in this PR will be for Apply again success rates and change rates. More PRs to follow.

- [x] Add _Apply again_ section to feature metrics dashboard
- [x] `ApplyAgainFeatureMetrics` service to fetch success rates and change rates
- [x] `ApplyAgainFeatureMetrics` filters by date range. For success rates it uses the last status change date to determine when the success happened. For change rates we use the `submitted_at` date to determine when the change happened. (see discussion on card).
- [x] Update system spec
- [x] Changed the default behaviour for missing `FeatureMetricDashboard` keys to just show a _n/a_ rather than throw an exception. This is so that the dashboard runs without errors even if we are waiting for the background job to populate new keys.

image

## Guidance to review

- Success is defined as the ratio of `apply again applications that have at least one successful choice / apply again applications for which all the choices have an outcome`. `successful choice` means having status `offer_deferred`, `pending_conditions` or `recruited`. `have an outcome` means having status `withdrawn`, `offer_withdrawn`, `rejected`, `conditions_not_met`, `declined`, `cancelled`, `offer`, `offer_deferred`, `pending_conditions` or `recruited`
- Date filtering for success rate is a bit messy because it requires a join to the audits table. Is there a simpler way?
- There are lots of model classes for feature metrics now, all top-level. There is also a `CandidateInterface::FindChangedApplyAgainApplications` service. I think these should be re-organised. Should we put make them all services? stick them under the `SupportInterface` or maybe even a `SupportInterface::FeatureMetrics` namespace? (this can be done as a separate PR but interested in what people think).
- Does the layout and content make sense? (Probably best to review this in the whole when the full set of changes is available but any input welcome).
- Is it OK to not put this behind a feature flag? Will shipping bits as they are ready confuse people?

## Link to Trello card

## Things to check

### DIFF
--- a/app/models/magic_link_feature_metrics.rb
+++ b/app/models/magic_link_feature_metrics.rb
@@ -7,6 +7,7 @@ class MagicLinkFeatureMetrics
     end_time = Time.zone.now.beginning_of_day
   )
     records = ApplicationForm
+      .apply_1
       .select(
         'count(DISTINCT audits.id) as audit_count',
         'count(DISTINCT authentication_tokens.id) as token_count',
--- a/app/models/reference_feature_metrics.rb
+++ b/app/models/reference_feature_metrics.rb
@@ -36,6 +36,7 @@ private
 
   def time_to_get_references(start_time, end_time = Time.zone.now)
     applications = ApplicationForm
+      .apply_1
       .joins(:application_references)
       .where('"references".feedback_provided_at BETWEEN ? AND ? AND "references".duplicate = ?', start_time, end_time, false)
       .group('application_forms.id')
--- a/app/models/reference_feature_metrics.rb
+++ b/app/models/reference_feature_metrics.rb
@@ -36,7 +36,6 @@ private
 
   def time_to_get_references(start_time, end_time = Time.zone.now)
     applications = ApplicationForm
-      .apply_1
       .joins(:application_references)
       .where('"references".feedback_provided_at BETWEEN ? AND ? AND "references".duplicate = ?', start_time, end_time, false)
       .group('application_forms.id')
@@ -46,7 +45,7 @@ private
   def time_to_get_for(application, end_time)
     return nil unless application.enough_references_have_been_provided?
 
-    times = application.application_references.feedback_provided.map do |reference|
+    times = application.application_references.feedback_provided.where(duplicate: false).map do |reference|
       [reference.requested_at, reference.feedback_provided_at]
     end
     requested_at_time = times.map(&:first).compact.min
--- a/app/views/support_interface/performance/feature_metrics_dashboard.html.erb
+++ b/app/views/support_interface/performance/feature_metrics_dashboard.html.erb
@@ -213,4 +213,34 @@
       </div>
     </div>
   </div>
+
+  <h3 class="govuk-heading-m govuk-!-font-size-27 govuk-!-margin-top-8">Apply again</h3>
+  <div id="apply_again_dashboard_section" class="govuk-grid-row govuk-!-margin-bottom-4">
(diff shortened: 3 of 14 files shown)


app/models/reference_feature_metrics.rb
```diff
@@ -36,6 +36,7 @@ def percentage_references_within(
 
   def time_to_get_references(start_time, end_time = Time.zone.now)
     applications = ApplicationForm
+      .apply_1
```
davidgisbey, line 39: I think we should be checking for `duplicate: false` instead here as some people request new references as part of their apply_again application. Sorry i should have picked this up as part of my earlier PR

davidgisbey: where('"references".feedback_provided_at BETWEEN ? AND ? AND duplicate = ? ', start_time, end_time, false)``

spec/system/support_interface/feature_metrics_dashboard_spec.rb
```diff
@@ -56,6 +57,34 @@ def reject_application(application_choice)
...
+    application_choice = create(
+      :application_choice,
+      :awaiting_provider_decision,
+      application_form: apply_again_application_form,
+    )
+    ApplicationStateChange.new(application_choice).make_offer!
+    apply_again_application_form
+  end
+
+  def apply_again_and_offer_application(application_form)
```
davidgisbey, line 74: i think these two methods are the wrong way round. This one rejects the application and the other one makes an offer.

app/models/magic_link_feature_metrics.rb
```diff
@@ -7,6 +7,7 @@ def average_magic_link_requests_upto(
     end_time = Time.zone.now.beginning_of_day
   )
     records = ApplicationForm
+      .apply_1
```
davidgisbey, line 10: I'm probs missing something, but does this have to be scoped to apply_1?

app/views/support_interface/performance/feature_metrics_dashboard.html.erb
```diff
@@ -242,5 +242,31 @@
         </div>
       </div>
     </div>
+    <div class="govuk-grid-column-one-half">
+      <div id="headline-stat" class="govuk-!-margin-bottom-4">
+        <%= render SupportInterface::TileComponent.new(
+          count: @dashboard.read_metric(:apply_again_change_rate),
+          label: 'candidates that make changes to application when applying again',
```
davidgisbey, line 249: super nitty, but probably should be 'made changes to their application'. Sorry to be that guy haha

davidgisbey (approved):
A few minor comments, but i've tested it locally and it's working nicely 👍

app/models/reference_feature_metrics.rb
stevehook: I vaguely remember talking to someone about whether including apply again applications would skew the stats. The problem is that some apply again applications will have no new references, some might update one, others update both and to compare like with like you only really want to consider the ones that are replacing both references. I think your suggestion achieves that though.

app/models/magic_link_feature_metrics.rb
stevehook: I think I was concerned about the two getting muddled up because the magic link tokens are linked to the candidate rather than the application form. So it's a very good question though I don't feel confident about just changing it right now. To open it up to apply again applications I basically need to make sure we have a reliable method to avoid counting sign-ins from the previous application.

PR #4267, opened 2021-03-10
Hide interview link in activity log if it has been cancelled

benswannack:
## Context
Interviews can be cancelled and will no longer be displayed in the UI. The link to interviews was still present in the activity log however, which meant that we were linking to something that does not exist

Hide interview link when associated interview has been cancelled

## Guidance to review
Per commit

## Link to Trello card
https://trello.com/c/Z3qSD93D/3476-hide-view-interview-link-from-interview-activity-log-when-cancelled

## Things to check

### DIFF
--- a/app/components/provider_interface/activity_log_event_component.rb
+++ b/app/components/provider_interface/activity_log_event_component.rb
@@ -147,6 +147,8 @@ module ProviderInterface
     end
 
     def interview_link(audit)
+      return if audit.auditable.cancelled?
+
       {
         url: routes.provider_interface_application_choice_interviews_path(audit.associated, anchor: "interview-#{audit.auditable.id}"),
         text: 'View interview',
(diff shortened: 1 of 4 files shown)


app/components/provider_interface/activity_log_event_component.rb
```diff
@@ -138,5 +118,41 @@ def interview_event_description(user, candidate)
...
+
+    def feedback_link
+      {
+        url: routes.provider_interface_application_choice_feedback_path(application_choice),
+        text: 'View feedback',
+      }
+    end
+
+    def interview_link(audit)
+      return if audit.auditable.cancelled?
```
AgaDufrat, line 150: Since this can now return `nil`, I think we need to update the conditional in the view to
`<% if link.present? %>`
https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/components/provider_interface/activity_log_event_component.html.erb#L25

AgaDufrat (approved):
Looking good. Left one comment we should address before merging.

app/components/provider_interface/activity_log_event_component.rb
benswannack: That's a good point and one I thought I had tested!

PR #4268, opened 2021-03-10
Display valid date in interview date hint

benswannack:
## Context
Noticed during product review, the date hint for interviews displays an invalid time.

Display a dynamic date that is the minimum of tomorrow and the application RBD

## Guidance to review
Theoretically the rbd date could be in the past and we end up displaying a past date. This wouldn't happen in practice though since the application would have been rejected.

## Link to Trello card
https://trello.com/c/CxPyFUvk/3477-change-hint-text-for-interview-time-to-a-date-in-the-future

## Things to check


app/views/provider_interface/interviews/_form.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= f.govuk_date_field :date %>
+<%= f.govuk_date_field :date, hint: { text: t('helpers.hint.provider_interface_interview_wizard.date', example_date: @wizard.example_date) } %>
```
benswannack, line 1: This form should really be a component and then we can test that it's rendering sensible things. Looking into that now

spec/components/provider_interface/interview_form_component_spec.rb
```diff
@@ -0,0 +1,103 @@
+require 'rails_helper'
+
+RSpec.describe ProviderInterface::InterviewFormComponent do
+  let(:interview_preferences) { nil }
+  let(:application_form) { build_stubbed(:application_form, interview_preferences: interview_preferences) }
+  let(:application_choice) { build_stubbed(:submitted_application_choice, application_form: application_form) }
+
+  let(:form_object_class) do
```
benswannack, line 8: This makes me think the component and wizard are too tightly coupled. I experimented with moving the provider list logic into the component, but really the wizard needs to know about this too. Any suggestions for a better place for this?

app/components/provider_interface/interview_form_component.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    def initialize(application_choice:, form_model:, form_url:, form_heading:)
+      @application_choice = application_choice
+      @form_model = form_model
+      @form_url = form_url
+      @form_heading = form_heading
+    end
+
+    def example_date
+      valid_future_date_for_form = [application_choice.reject_by_default_at, 1.day.from_now].min
```
benswannack, line 13: There was a question about whether it is useful to display today’s date in the case the application is going to be rejected that day.

Given that it is the only date that would be accepted by the form, it feels like the only choice here

spec/components/provider_interface/interview_form_component_spec.rb
cwrw: Hmm good question, maybe if we move the logic for `application_providers` and `multiple_application_providers` into the `application_choice` as it does feel like its that model's concern, then we just use it straight in the view via `application_choice`, what do you think?

benswannack: That sounds good! 👍

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_03_08_223028) do
+ActiveRecord::Schema.define(version: 2021_03_08_172728) do
```
cwrw, line 13: Can you remove this change as seems unrelated to this PR

cwrw (approved):
Looks good, just the one comment.

I noticed that we still show a change link on the check page for interview changes next a provider even if there is only one provider present, I'll add a bug ticket to fix this.

PR #4271, opened 2021-03-10
Rename service to change_an_offer

SamJamCul:
## Context

To bring the change_an_offer service name more in line with others in the same scope

Change all instances of ChangeOffer to ChangeAnOffer.

One locale was set as change_offer, and that resulted in a failing test - this has been updated to change_an_offer as well. 

## Things to check

### DIFF
--- a/app/controllers/provider_interface/offer_changes_controller.rb
+++ b/app/controllers/provider_interface/offer_changes_controller.rb
@@ -37,7 +37,7 @@ module ProviderInterface
       change_offer_form.step = :update
 
       if change_offer_form.valid?
-        ::ChangeOffer.new(
+        ::ChangeAnOffer.new(
           actor: current_provider_user,
           application_choice: @application_choice,
           course_option: change_offer_form.selected_course_option,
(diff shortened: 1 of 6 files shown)


duncanjbrown:
Very good to be consistent! I would rather go the other way so `MakeOffer` is consistent with `ChangeOffer`. We don't bother with `a` or `an` elsewhere. Keen to hear what others think,

despo:
@duncanjbrown the goal is to move away from both MakeAnOffer and ChangeAnOffer (after we are done with the new Make offer flow work), so this rename is done in order to enable the creation of a wrapper service 'CreateOffer' similar to the one we've already put in place for `MakeOffer`, that can be used by the offer flow and raises an error if any of the validations fail.

despo (approved):

duncanjbrown:
> @duncanjbrown the goal is to move away from both MakeAnOffer and ChangeAnOffer (after we are done with the new Make offer flow work), so this rename is done in order to enable the creation of a wrapper service `CreateOffer` similar to the one we've already put in place for `MakeOffer`, that can be used by the offer flow and raises an error if any of the validations fail.

Ah, great. Thanks for explaining!

PR #4273, opened 2021-03-10
Prevent users from being able to add new users to providers they don't have the right permissions for

steventux:
## Context

The update action was for the provider user invitation controller is using the user's list of providers regardless of whether the user could manage other users for that organisation.

Ensures the list of providers throughout the wizard flow is comprised of organisations the current provider user has permissions to manage users for.

## Guidance to review

The system spec already contains a test for the correct list of providers, so I added the necessary failing validation step to ensure the bug is fixed.

Steps to reproduce bug:

- Make sure that you're a user associated with at least two organisations, with manage user permissions for one provider, and not the other
- Log into Manage, and select invite user (enter contact details)
- Don't select an organisation on the next page
- The error page allows the user to select any organisation that they are associated with, not just the orgs they have manage user permissions for

## Link to Trello card

https://trello.com/c/kgbH65A2/3480-prevent-users-from-being-able-to-add-new-users-to-providers-they-dont-have-the-right-permissions-for

## Things to check

### DIFF
--- a/app/controllers/provider_interface/provider_users_invitations_controller.rb
+++ b/app/controllers/provider_interface/provider_users_invitations_controller.rb
@@ -27,7 +27,7 @@ module ProviderInterface
 
     def update_providers
       @wizard = wizard_for(providers_params.merge(current_step: 'providers'))
-      @available_providers = current_provider_user.providers
+      @available_providers = available_providers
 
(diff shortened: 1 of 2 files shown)


despo (approved):
LGTM 👍🏼

PR #4274, opened 2021-03-11
Rails route helpers included in ViewComponent::Base

despo:
Rails route helpers are already included in ViewComponent::Base

https://github.com/github/view_component/blob/main/lib/view_component/base.rb#L225


TobyRet (approved):
LGTM!

PR #4276, opened 2021-03-11
Remove V3 API (3/4) — delete all V3 code

duncanjbrown:
### Removing V3 from Apply
1. Apply from Find uses public API
1. Consume subject codes from the public API and remove remaining references to the v3 API
1. **Delete all TTAPI v3 code** 👈
1. Clean up after removing v3 API

🚮


TobyRet (approved):
LGTM

PR #4277, opened 2021-03-11
UK NARIC name change to UK ENIC

TobyRet:
## Context

UK NARIC, the agency for equivalent qualifications has changed its name to (the more catchy?) _UK ENIC_. We point users in their direction quite a few times in the Apply journey so we need to update the content and URL

- Rename all references to UK NARIC (and its other forms) in the codebase

## Guidance to review

- In order to not push breaking changes to users, values will be written to both the old `naric_reference` column and the new `enic_reference` column.
- A follow up PR will need to be deployed at the same time which backfills the `enic_reference` column and deletes the `naric_reference` column - *please also review* https://github.com/DFE-Digital/apply-for-teacher-training/pull/4281

## Link to Trello card

## Things to check


db/schema.rb
```diff
@@ -10,9 +10,9 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_03_08_223028) do
+ActiveRecord::Schema.define(version: 2021_03_11_161050) do
 
-  create_sequence "application_choices_id_seq"
+  create_sequence "application_choices_id_seq1"
```
TobyRet, line 15: This is strange?

davidgisbey (approved):
lgtm Toby 👍

PR #4278, opened 2021-03-11
Redirect back from conditions actions if the application is already r…

steventux:
…ecruited

## Context

Conditions should not be editable or updateable if the application has reached the recruited state.

Adds a filter to redirect back (falling back to `/provider`) from the conditions `edit/confirm_update/update` actions if the application has reached `recruited` state.

## Link to Trello card

https://trello.com/c/Q6IY6Dk8/3464-prevent-access-to-provider-conditions-controller-views-for-recruited-applications

## Things to check

### DIFF
--- a/app/controllers/provider_interface/conditions_controller.rb
+++ b/app/controllers/provider_interface/conditions_controller.rb
@@ -1,6 +1,7 @@
 module ProviderInterface
   class ConditionsController < ProviderInterfaceController
     before_action :set_application_choice
+    before_action :redirect_back_if_application_terminated
     before_action :requires_make_decisions_permission
 
     def edit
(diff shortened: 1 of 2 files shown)


app/controllers/provider_interface/conditions_controller.rb
```diff
@@ -40,5 +41,9 @@ def update
 
       redirect_to provider_interface_application_choice_path(@application_choice.id)
     end
+
+    def redirect_back_if_application_recruited
+      redirect_back(fallback_location: provider_interface_path) if @application_choice.recruited?
```
cwrw, line 46: Is it safe to use `TERMINAL_STATES` instead to guard against all those or should it only be the recruited state?
Also I was wondering if we should use `provider_interface_application_choice_path(@application_choice.id)` here instead rather than `/provider`?

steventux: :+1: https://github.com/DFE-Digital/apply-for-teacher-training/pull/4278/commits/c54844635a3fc8794e4ca1ad2166e6a17d65be34

cwrw (approved):

PR #4280, opened 2021-03-11
Humanize compared attribute

raamSoftwire:
## Context

We need to humanize the error message for date validations that compare two dates.

Before
image

After
image

## Guidance to review

Will this have any negative side effects?

## Link to Trello card

https://trello.com/c/iGPeDu77/3142-date-validation-error-message-bug

## Things to check


despo (approved):
Looks great. Thanks for fixing this!

PR #4285, opened 2021-03-12
Edit a new offer from the review page via the change links

despo:
*Note that we are not retrieving the correct set of data, as we will be done on a separate card by integrating the new `GetChangeOfferOptions` query service.

## Link to Trello card

https://trello.com/c/TbL2H8ck/3459-edit-a-new-offer-from-the-review-page-via-the-change-links

## Things to check

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -42,6 +42,8 @@ Metrics/BlockLength:
 RSpec/NestedGroups:
   Enabled: true
   Max: 4
+  Exclude:
+    - 'spec/forms/provider_interface/offer_wizard_spec.rb'
 
 RSpec/ExampleLength:
   Enabled: false
--- a/app/components/provider_interface/offer_summary_component.rb
+++ b/app/components/provider_interface/offer_summary_component.rb
@@ -15,7 +15,7 @@ module ProviderInterface
         { key: 'Provider',
           value: course_option.provider.name_and_code,
           action: 'Change',
-          change_path: nil },
+          change_path: new_provider_interface_application_choice_offer_providers_path(application_choice) },
         { key: 'Course',
           value: course_option.course.name_and_code,
           action: 'Change',
--- a/app/components/provider_interface/offer_summary_component.rb
+++ b/app/components/provider_interface/offer_summary_component.rb
@@ -23,7 +23,7 @@ module ProviderInterface
         { key: 'Location',
           value: course_option.site.name_and_address,
           action: 'Change',
-          change_path: nil },
+          change_path: new_provider_interface_application_choice_offer_locations_path(application_choice) },
         { key: 'Full time or part time',
           value: course_option.study_mode.humanize,
           action: 'Change',
--- a/app/components/provider_interface/offer_summary_component.rb
+++ b/app/components/provider_interface/offer_summary_component.rb
@@ -2,12 +2,16 @@ module ProviderInterface
   class OfferSummaryComponent < ViewComponent::Base
     include ViewHelper
 
-    attr_accessor :application_choice, :course_option, :conditions
+    attr_accessor :application_choice, :course, :course_option, :conditions, :available_providers, :available_courses, :available_course_options
 
-    def initialize(application_choice:, course_option:, conditions:)
+    def initialize(application_choice:, course:, course_option:, conditions:, available_providers: [], available_courses: [], available_course_options: [])
       @application_choice = application_choice
       @course_option = course_option
       @conditions = conditions
+      @available_providers = available_providers
+      @available_courses = available_courses
+      @available_course_options = available_course_options
+      @course = course
     end
 
     def rows
@@ -15,20 +19,36 @@ module ProviderInterface
         { key: 'Provider',
           value: course_option.provider.name_and_code,
           action: 'Change',
-          change_path: new_provider_interface_application_choice_offer_providers_path(application_choice) },
+          change_path: change_provider_path },
         { key: 'Course',
           value: course_option.course.name_and_code,
           action: 'Change',
-          change_path: new_provider_interface_application_choice_offer_courses_path(application_choice) },
+          change_path: change_course_path },
         { key: 'Location',
           value: course_option.site.name_and_address,
           action: 'Change',
-          change_path: new_provider_interface_application_choice_offer_locations_path(application_choice) },
+          change_path: change_location_path },
         { key: 'Full time or part time',
           value: course_option.study_mode.humanize,
           action: 'Change',
-          change_path: new_provider_interface_application_choice_offer_study_modes_path(application_choice) },
+          change_path: change_study_mode_path },
       ]
+    end
+
+    def change_provider_path
+      available_providers.many? ? new_provider_interface_application_choice_offer_providers_path(application_choice) : nil
+    end
+
+    def change_course_path
+      available_courses.many? ? new_provider_interface_application_choice_offer_courses_path(application_choice) : nil
+    end
+
+    def change_location_path
+      available_course_options.many? ? new_provider_interface_application_choice_offer_locations_path(application_choice) : nil
+    end
+
+    def change_study_mode_path
+      course.full_time_or_part_time? ? new_provider_interface_application_choice_offer_study_modes_path(application_choice) : nil
     end
   end
 end
--- a/app/components/provider_interface/offer_summary_component.rb
+++ b/app/components/provider_interface/offer_summary_component.rb
@@ -27,7 +27,7 @@ module ProviderInterface
         { key: 'Full time or part time',
           value: course_option.study_mode.humanize,
           action: 'Change',
-          change_path: nil },
+          change_path: new_provider_interface_application_choice_offer_study_modes_path(application_choice) },
       ]
     end
   end
--- a/app/components/provider_interface/offer_summary_component.rb
+++ b/app/components/provider_interface/offer_summary_component.rb
@@ -19,7 +19,7 @@ module ProviderInterface
         { key: 'Course',
           value: course_option.course.name_and_code,
           action: 'Change',
-          change_path: nil },
+          change_path: new_provider_interface_application_choice_offer_courses_path(application_choice) },
         { key: 'Location',
           value: course_option.site.name_and_address,
           action: 'Change',
--- a/app/controllers/provider_interface/offer/study_modes_controller.rb
+++ b/app/controllers/provider_interface/offer/study_modes_controller.rb
@@ -1,0 +1,41 @@
+module ProviderInterface
+  module Offer
+    class StudyModesController < OffersController
+      def new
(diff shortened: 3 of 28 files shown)


app/forms/provider_interface/offer_wizard.rb
```diff
@@ -38,14 +45,55 @@ def valid_for_current_step?
...
+      attrs = { study_mode: available_study_modes.first } if step.eql?(:study_modes)
+      attrs = { course_option_id: available_course_options.first.id } if step.eql?(:locations)
+
+      assign_attributes(last_saved_state.deep_merge(attrs))
+      save_state!
+
+      next_step(step)
+    end
+
+    def available_study_modes
```
cwrw, line 80: we seem to be defining those values twice, once in the controller and once here, I was wondering if we only defined them once in the controller, and passed them in as attributes to wizard instead as well as the component?

app/forms/provider_interface/offer_wizard.rb
```diff
@@ -54,5 +102,15 @@ def last_saved_state
     def state
       as_json(except: %w[state_store errors validation_context course_option]).to_json
     end
+
+    def update_path_history(attrs)
```
cwrw, line 106: maybe this can be extracted outside this wizard and into its own class to deal with the back link functionality that can be reused/tested in isolation?

app/forms/provider_interface/offer_wizard.rb
```diff
@@ -2,19 +2,26 @@ module ProviderInterface
   class OfferWizard
     include ActiveModel::Model
 
-    STEPS = { make_offer: %i[select_option conditions check] }.freeze
+    STEPS = { make_offer: %i[select_option conditions check],
+              change_offer: %i[select_option providers courses study_modes locations conditions check] }.freeze
```
cwrw, line 6: Isn't locations supposed to be the step before `study_modes`? 🤔

app/forms/provider_interface/offer_wizard.rb
```diff
@@ -38,14 +45,55 @@ def valid_for_current_step?
...
+    def next_step(step = current_step)
+      index = STEPS[decision.to_sym].index(step.to_sym)
+      return unless index
+
+      current_step = STEPS[decision.to_sym][index + 1]
+
+      return save_and_go_to_next_step(current_step) if current_step.eql?(:providers) && available_providers.one?
+      return save_and_go_to_next_step(current_step) if current_step.eql?(:courses) && available_courses.one?
+      return save_and_go_to_next_step(current_step) if current_step.eql?(:study_modes) && available_study_modes.one?
+      return save_and_go_to_next_step(current_step) if current_step.eql?(:locations) && available_course_options.one?
```
cwrw, line 57: I'm running into an issue locally where a course does not have course options and so we're getting an empty locations page as this is checking if its one and not if any exists, I checked on production and we do have some of those, is it safer to check if any exist as well? Also I'm not sure if this is actually a real issue as those courses will probably not be valid when we plug in the real query service?
Screenshot 2021-03-18 at 10 35 16

.rubocop.yml
```diff
@@ -42,6 +42,8 @@ Metrics/BlockLength:
 RSpec/NestedGroups:
   Enabled: true
   Max: 4
+  Exclude:
```
cwrw, line 45: The nesting feels like it can be avoided, could we attempt that rather than excluding the spec here, maybe if we pass in the different courses/locations/providers into the wizard then we can avoid the extra setup/nesting?

spec/system/provider_interface/make_offer_spec.rb
```diff
@@ -38,6 +43,86 @@
     then_the_review_page_is_loaded
     and_i_can_confirm_my_answers
 
+    # locations
```
cwrw, line 46: Maybe we should move the change link tests to a new system spec as this one has become quite large?

app/helpers/offer_path_helper.rb
```diff
@@ -0,0 +1,9 @@
+module OfferPathHelper
+  def offer_path_for(application_choice, step, params = {})
```
cwrw, line 2: can you add a spec for this?

config/locales/provider_interface/make_offer.yml
```diff
@@ -43,3 +56,5 @@ en:
           attributes:
             decision:
               blank: Select if you want to make an offer or reject the application
+            course_option_id:
```
cwrw, line 59: can we add a custom error for the course selection as well?

cwrw (commented):
Looks good, left a few comments, happy to discuss over a call

app/forms/provider_interface/offer_wizard.rb
despo: The problem with that is that we don't always need them, neither in the controller or in the wizard. That would force us to retrieve all the values on all steps.

app/forms/provider_interface/offer_wizard.rb
despo: No. The study mode is required in order to limit the course options. The alternative would be the candidates picking a course based on where it's located and then potentially that course not having a full time option available. We discussed this with both Michael and Adam as we also thought it would be the other way around.

Actually as part of this we also need to reverse the order of Study mode and Location in all places where offers are displayed, e.g. make offer summary page, offer page etc, but we decided to do that separately.

app/forms/provider_interface/offer_wizard.rb
despo: We raised this to Michael and expect the service retrieving the available data to take this into account and not return any options that will result in a dead-end.

spec/system/provider_interface/make_offer_spec.rb
despo: Won't it cause exactly the same issue, as the new system spec does more than the make offer one? We added the code here as we wanted it to be clear that the change links are available via the make offer process as well.

I don't know if we are a bit unecessarily overtesting here however, as we are testing the next step behavior in the wizard spec as well, so maybe we can just leave one step and remove the rest?

app/helpers/offer_path_helper.rb
despo: Spec added.

config/locales/provider_interface/make_offer.yml
richardpattinson: added =)

app/forms/provider_interface/offer_wizard.rb
despo: We looked a little bit into this and it feels very unnatural to do it in the controller. Besides having to load upfront information, we also rely on the wizard to validate the different information we provide e.g.  the wizard needs to be initialised with a different set of parameters, then we need to check for the validity and then, any of these details can be retrieved. We also need to know the selected option in order to retrieve the next set of information, e.g. if a provider has a single course then we need to check for that before we can retrieve the course options. That moves a lot of the logic to the controller making it do a lot of extra things we can't properly test for.

spec/system/provider_interface/make_offer_spec.rb
cwrw: That's a good question, I think if we test the following two scenarios we don't have to cover every change link and just check that the change links are visible instead:
1) Going through all the steps again when clicking a change link eg. provider
2) Going through some steps that skip a step because there is only one option

cwrw (approved):
looks great, just the one system spec comment

app/components/provider_interface/offer_summary_component.rb
```diff
@@ -2,33 +2,53 @@ module ProviderInterface
   class OfferSummaryComponent < ViewComponent::Base
     include ViewHelper
 
-    attr_accessor :application_choice, :course_option, :conditions
+    attr_accessor :application_choice, :course, :course_option, :conditions, :available_providers, :available_courses, :available_course_options
```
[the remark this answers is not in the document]
richardpattinson: They need to be accessors because we assign values to them.

app/forms/provider_interface/offer_wizard.rb
```diff
@@ -38,21 +44,68 @@ def valid_for_current_step?
       valid?(current_step.to_sym)
     end
 
-    def next_step
-      index = STEPS[decision.to_sym].index(current_step.to_sym)
+    def next_step(step = current_step)
```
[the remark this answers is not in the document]
richardpattinson: > Can this be `next_step!` instead, since it changes the `current_step`?

It cant be next_step! because even though it sets the value of the next step, it doesn't save it to the data store. 

> Also, do we need the parameter here? Can we just refer to instance's current_step everywhere inside these methods?

The intention of using the name current_step is not to override the value of the instance varaible current_step as we dont know yet if the user will click next or not. This is why we want to specify the parameter when we want to calculate the step (in case any steps are missed). However this does raise a question about the choice of the variable name that we use within next step. We will rename current_step to next_step for clarity.

app/forms/provider_interface/offer_wizard.rb
```diff
@@ -38,21 +44,68 @@ def valid_for_current_step?
...
+      return save_and_go_to_next_step(current_step) if current_step.eql?(:locations) && available_course_options.one?
+
+      current_step
     end
 
+    delegate :previous_step, to: :wizard_path_history
+
   private
 
+    def save_and_go_to_next_step(step)
```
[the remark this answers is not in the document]
richardpattinson: As per the thread above we renamed the variable in order to avoid confusion with current step. We don’t want to change anything because we are not saving current step, we dont want to persist it either.

spec/system/provider_interface/make_offer_spec.rb
```diff
@@ -103,6 +140,137 @@ def and_i_can_confirm_my_answers
     end
   end
 
+  def given_the_course_has_multiple_course_options
```
cwrw, line 143: Can you clean up the now unused methods from this spec?

cwrw (approved):
Looks great, just the one comment otherwise 👍

app/forms/provider_interface/offer_wizard.rb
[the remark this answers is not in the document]
benswannack: I think my confusion is because the next_step method has this side effect where it sets some state on the wizard. It's a bit unexpected, but also minor

app/controllers/provider_interface/offer/study_modes_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+      end
+
+      def create
+        @wizard = OfferWizard.new(offer_store, study_mode_params.to_h)
+        @wizard.course_option_id = nil if @wizard.course_option_id && @wizard.course_option.study_mode != @wizard.study_mode
+
+        if @wizard.valid_for_current_step?
+          @wizard.save_state!
+
+          redirect_to [:new, :provider_interface, @application_choice, :offer, @wizard.next_step]
```
benswannack, line 19: Can these use the OfferPathHelper?

PR #4286, opened 2021-03-12
Let users manage the type of email notifications they receive

AgaDufrat:
## Context

> As a provider user 
> I need the ability to choose the type of email notification I receive
> So that I only get notified about the things I need

Screenshot 2021-03-12 at 18 03 27

## Guidance to review

- test the provider interface `/provider/account/notification-settings`
- test the support interface `/support/users/provider/:id`

## Link to Trello card

https://trello.com/c/YzNo7PT3/3448-let-users-manage-the-type-of-email-notifications-they-receive

## Things to check

### DIFF
--- a/app/components/provider_user_notification_preferences_component.html.erb
+++ b/app/components/provider_user_notification_preferences_component.html.erb
@@ -1,0 +1,14 @@
+<%= form_with model: notification_preferences, url: form_path, method: :put do |f| %>
+  <div class='govuk-form-group'>
+    <% ProviderUserNotificationPreferences::NOTIFICATION_PREFERENCES.each do |notification_preference| %>
+      <%= f.govuk_collection_radio_buttons notification_preference,
+        on_off_options,
+        :value,
+        :name,
+        inline: true,
+        legend: { text: t("provider_user_notification_preferences.#{notification_preference}.legend") } %>
+    <% end %>
+  </div>
+
+  <%= f.govuk_submit 'Save settings' %>
+<% end %>
--- a/app/components/provider_user_notification_preferences_component.rb
+++ b/app/components/provider_user_notification_preferences_component.rb
@@ -1,0 +1,16 @@
+class ProviderUserNotificationPreferencesComponent < ViewComponent::Base
+  include ViewHelper
+  attr_reader :notification_preferences, :form_path
+
+  def initialize(notification_preferences, form_path:)
+    @notification_preferences = notification_preferences
+    @form_path = form_path
+  end
+
+  def on_off_options
+    [
+      OpenStruct.new(value: 'true', name: 'On'),
+      OpenStruct.new(value: 'false', name: 'Off'),
+    ]
+  end
+end
--- a/app/components/support_interface/provider_user_summary_component.rb
+++ b/app/components/support_interface/provider_user_summary_component.rb
@@ -1,0 +1,34 @@
+module SupportInterface
+  class ProviderUserSummaryComponent < SummaryListComponent
+    include ViewHelper
+
+    attr_reader :provider_user
+
+    def initialize(provider_user)
+      @provider_user = provider_user
+    end
+
+    def rows
+      [
+        { key: 'First name', value: provider_user.first_name },
+        { key: 'Last name', value: provider_user.last_name },
+        { key: 'Email address', value: provider_user.email_address },
+        { key: 'DfE Sign-in UID', value: provider_user.dfe_sign_in_uid },
+        { key: 'Account created at', value: provider_user.created_at.to_s(:govuk_date_and_time) },
+        { key: 'Last sign in at', value: provider_user.last_signed_in_at&.to_s(:govuk_date_and_time) || 'Not signed in yet' },
+        {
+          key: 'Email notifications',
+          value: boolean_to_word(provider_user.send_notifications?),
+          action: 'notifications',
+          change_path: edit_support_interface_provider_user_path(provider_user, anchor: 'update-email-notifications'),
+        },
+        {
+          key: 'Permissions',
+          value: render(SupportInterface::ProviderUserPermissionsComponent.new(provider_user: provider_user)),
+          action: 'permissions',
+          change_path: edit_support_interface_provider_user_path(provider_user),
+        },
+      ]
+    end
+  end
+end
--- a/app/components/support_interface/provider_user_summary_component.rb
+++ b/app/components/support_interface/provider_user_summary_component.rb
@@ -18,7 +18,7 @@ module SupportInterface
         { key: 'Last sign in at', value: provider_user.last_signed_in_at&.to_s(:govuk_date_and_time) || 'Not signed in yet' },
         {
           key: 'Email notifications',
-          value: boolean_to_word(provider_user.send_notifications?),
+          value: email_notifications_value,
           action: 'notifications',
           change_path: edit_support_interface_provider_user_path(provider_user, anchor: 'update-email-notifications'),
         },
@@ -29,6 +29,19 @@ module SupportInterface
           change_path: edit_support_interface_provider_user_path(provider_user),
         },
       ]
+    end
+
+  private
+
+    def email_notifications_value
+      FeatureFlag.active?(:configurable_provider_notifications) ? render(SummaryCardComponent.new(rows: notification_preferences_rows, border: false)) : boolean_to_word(provider_user.send_notifications?)
+    end
+
+    def notification_preferences_rows
+      rows = ProviderUserNotificationPreferences::NOTIFICATION_PREFERENCES.map do |type|
+        [t("provider_user_notification_preferences.#{type}.legend"), provider_user.notification_preferences.send(type) ? 'On' : 'Off']
+      end
+      rows.to_h
     end
   end
 end
--- a/app/controllers/provider_interface/notifications_controller.rb
+++ b/app/controllers/provider_interface/notifications_controller.rb
@@ -1,9 +1,13 @@
 module ProviderInterface
   class NotificationsController < ProviderInterfaceController
     def update
-      SaveProviderUserNotificationPreferences
-        .new(provider_user: current_provider_user)
-        .backfill_notification_preferences!(send_notifications: notification_params[:send_notifications])
+      provider_user_notifications_service = SaveProviderUserNotificationPreferences.new(provider_user: current_provider_user)
+
+      if FeatureFlag.active?(:configurable_provider_notifications)
+        provider_user_notifications_service.update_all_notification_preferences!(notification_preferences_params: notification_preferences_params)
+      else
+        provider_user_notifications_service.backfill_notification_preferences!(send_notifications: notification_params[:send_notifications])
+      end
 
       flash[:success] = 'Email notification settings saved'
       redirect_to provider_interface_notifications_path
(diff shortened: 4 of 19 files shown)


app/components/provider_user_notification_preferences_component.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def on_off_options
+    [
+      OpenStruct.new(id: 'true', name: 'On'),
+      OpenStruct.new(id: 'false', name: 'Off'),
+    ]
+  end
+
+  def form_path
```
cwrw, line 21: can this be an attribute we pass into this component to avoid setting the path depending on where we call it?

app/controllers/support_interface/provider_users_controller.rb
```diff
@@ -84,6 +84,19 @@ def toggle_notifications
       redirect_to support_interface_provider_user_path(provider_user)
     end
 
+    def update_notifications
+      render_404 unless FeatureFlag.active?(:configurable_provider_notifications)
+
+      provider_user = ProviderUser.find(params[:provider_user_id])
+      notification_preferences = provider_user .notification_preferences
```
cwrw, line 91: ✂️ `provider_user . notification_preferences  | provider_user.notification_preferences `

app/components/provider_user_notification_preferences_component.rb
```diff
@@ -0,0 +1,28 @@
+class ProviderUserNotificationPreferencesComponent < ViewComponent::Base
+  include ViewHelper
+  attr_reader :notification_preferences
+
+  CUSTOM_LEGENDS = {
+    application_withdrawn: 'Application withdrawn by candidate',
```
cwrw, line 6: I think we might need to define this for all notifications to avoid relying on the column name incase it might be renamed, also what do you think about moving those to locales instead?

app/components/provider_user_notification_preferences_component.rb
```diff
@@ -0,0 +1,28 @@
...
+    application_rejected_by_default: 'Application automatically rejected',
+  }.freeze
+
+  def initialize(notification_preferences)
+    @notification_preferences = notification_preferences
+  end
+
+  def on_off_options
+    [
+      OpenStruct.new(id: 'true', name: 'On'),
```
cwrw, line 16: Nit: Maybe we can use `value` instead of `id` as it's more descriptive?

app/controllers/provider_interface/notifications_controller.rb
```diff
@@ -1,13 +1,15 @@
 module ProviderInterface
   class NotificationsController < ProviderInterfaceController
     def update
-      current_provider_user.update!(
-        send_notifications: notification_params[:send_notifications],
-      )
-
       if FeatureFlag.active?(:configurable_provider_notifications)
-        current_provider_user.notification_preferences
```
cwrw, line 9: Should we be updating all preferences to true/false by default outside of the feature flag block to make sure we update the new values whilst being backward compatible until the feature flag is turned on?

app/views/support_interface/provider_users/show.html.erb
```diff
@@ -4,6 +4,9 @@
 
 <%= render 'provider_user_navigation', title: 'Details', provider_user: @provider_user %>
 
+<% notification_preferences_rows = ProviderUserNotificationPreferences::NOTIFICATION_PREFERENCES.map { |type| [type.to_s.humanize, boolean_to_word(@provider_user.notification_preferences.send(type))] }.to_h if FeatureFlag.active?(:configurable_provider_notifications) %>
```
cwrw, line 7: could this move to the controller or a component?

app/controllers/provider_interface/notifications_controller.rb
AgaDufrat: Yes, we should do that. We just need to make sure we run the migration #4269 before merging this PR.

app/controllers/support_interface/provider_users_controller.rb
```diff
@@ -123,5 +136,12 @@ def provider_permissions_params
             .fetch(:provider_permissions_forms, {})
             .to_h
     end
+
+    def notification_preferences_params
+      return {} unless params.key?(:provider_user_notification_preferences)
```
despo, line 141:
```suggestion
      return ActionController::Parameters.new unless params.key?(:provider_user_notification_preferences)
```

app/controllers/provider_interface/notifications_controller.rb
```diff
@@ -1,11 +1,15 @@
...
     def update
-      current_provider_user.update!(
-        send_notifications: notification_params[:send_notifications],
-      )
-
       if FeatureFlag.active?(:configurable_provider_notifications)
+        notification_preferences = current_provider_user.notification_preferences
+        notification_preferences.assign_attributes(notification_preferences_params)
+
+        notification_preferences.save! if notification_preferences.valid?
```
despo, line 8: if this is invalid we won't save, but will display the success message. So we can either assume everything is valid and `save!` which will end up throwing an exception and indicating there is an issue, or handle the invalid model by re-rendering the update page and allowing for the errors to be rendered.

app/components/support_interface/provider_user_summary_component.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    def notification_preferences_rows
+      rows = ProviderUserNotificationPreferences::NOTIFICATION_PREFERENCES.map do |type|
+        [t("provider_user_notification_preferences.#{type}.legend"), provider_user.notification_preferences.send(type) ? 'On' : 'Off']
+      end
+      rows.to_h
+    end
+  end
+end
```
despo, line 47: Can we add a preview for this component please.

app/components/provider_user_notification_preferences_component.html.erb
```diff
@@ -0,0 +1,14 @@
...
+        on_off_options,
+        :value,
+        :name,
+        inline: true,
+        legend: { text: t("provider_user_notification_preferences.#{notification_preference}.legend") } %>
+    <% end %>
+
+    <%= f.govuk_submit 'Save settings' %>
+  <% end %>
+</div>
```
despo, line 14: Can we add a preview for this component please.

app/components/support_interface/provider_user_summary_component.rb
```diff
@@ -0,0 +1,47 @@
+module SupportInterface
+  class ProviderUserSummaryComponent < ViewComponent::Base
```
despo, line 2: Can we change this to inherit from the SummaryListComponent and remove the template?

I think similar to this...
```suggestion
  class ProviderUserSummaryComponent < SummaryListComponent
  
    def initialize(provider_user)
      @provider_user = provider_user
      super(table_rows)
    end
 
```

app/components/support_interface/provider_user_summary_component.rb
```diff
@@ -0,0 +1,47 @@
...
+    include ViewHelper
+
+    attr_reader :provider_user
+
+    def initialize(provider_user)
+      @provider_user = provider_user
+    end
```
despo, line 9: Can we change this to inherit from the SummaryListComponent and remove the template?

I think similar to this...
```suggestion
  class ProviderUserSummaryComponent < SummaryListComponent
    include ViewHelper

    attr_reader :provider_user
    
    def initialize(provider_user)
      @provider_user = provider_user
      super(table_rows)
    end
 
```

AgaDufrat:
Dependent on #4304 being merged first

app/controllers/provider_interface/notifications_controller.rb
```diff
@@ -1,11 +1,14 @@
...
   class NotificationsController < ProviderInterfaceController
     def update
-      current_provider_user.update!(
-        send_notifications: notification_params[:send_notifications],
-      )
-
       if FeatureFlag.active?(:configurable_provider_notifications)
+        notification_preferences = current_provider_user.notification_preferences
+        notification_preferences.assign_attributes(notification_preferences_params)
+        notification_preferences.save!
```
despo, line 7: I think that should also work?

```suggestion
        notification_preferences.update!(notification_preferences_params)
```

app/controllers/provider_interface/notifications_controller.rb
```diff
@@ -17,7 +20,16 @@ def update
   private
 
     def notification_params
+      return {} unless params.key?(:provider_user)
```
despo, line 23:
```suggestion
       return ActionController::Parameters.new unless params.key?(:provider_user)
```

app/controllers/support_interface/provider_users_controller.rb
```diff
@@ -84,6 +84,19 @@ def toggle_notifications
...
     end
 
+    def update_notifications
+      render_404 unless FeatureFlag.active?(:configurable_provider_notifications)
+
+      provider_user = ProviderUser.find(params[:provider_user_id])
+      notification_preferences = provider_user.notification_preferences
+      notification_preferences.assign_attributes(notification_preferences_params)
+
+      if notification_preferences.save!
```
despo, line 94: Similar to above, I think we can merge these two steps to an `update!` as it will throw a `ActiveRecord::RecordInvalid` similar to `save!`

app/controllers/support_interface/provider_users_controller.rb
```diff
@@ -84,6 +84,19 @@ def toggle_notifications
       redirect_to support_interface_provider_user_path(provider_user)
     end
 
+    def update_notifications
```
despo, line 87: This controller seems to be quickly expanding, and I think we could benefit from splitting it up into more specific controllers (and handle things in a more railsy way) e.g. `app/controllers/support_interface/provider_user/notifications_controller.rb` with an `update` action, and its own update route rather than plugging another action on this controller.

spec/components/previews/support_interface/provider_user_summary_component_preview.rb
```diff
@@ -0,0 +1,11 @@
+module SupportInterface
+  class ProviderUserSummaryComponentPreview < ViewComponent::Preview
+    def provider_user_summary
+      provider_user = FactoryBot.create(:provider_user,
+                                        created_at: Time.zone.now - 2.days)
+      FactoryBot.create(:provider_permissions, provider_user: provider_user)
```
despo, line 6: I found out recently that previews have access to dev data, so you should be able to do ```
```ruby
 provider_user = ProviderPermissions.limit(10).sample.provider_user
```

despo (approved):
@AgaDufrat looks good 💯.

The main thing that I think can be improved is creating a new `SupportInterface/ProviderUser/NotificationsController` as the existing support provider_user controller seems to already be doing too much.

app/components/provider_user_notification_preferences_component.html.erb
```diff
@@ -0,0 +1,14 @@
...
+    <% ProviderUserNotificationPreferences::NOTIFICATION_PREFERENCES.each do |notification_preference| %>
+      <%= f.govuk_collection_radio_buttons notification_preference,
+        on_off_options,
+        :value,
+        :name,
+        inline: true,
+        legend: { text: t("provider_user_notification_preferences.#{notification_preference}.legend") } %>
+    <% end %>
+
+    <%= f.govuk_submit 'Save settings' %>
```
cwrw, line 12: Can you add some padding around this submit button as per the [protoype](https://manage-applications-beta.herokuapp.com/account/notifications), as when in mobile view there isn't any spacing

cwrw (commented):
Looks great, just the one comment and also as discussed in person we just need to wait for @steventux PR to go in first then we can always update both `send_notifications` and `notification_preferences` in both controllers regardless of the feature flag to be backward compatible

cwrw (approved):
👏

PR #4290, opened 2021-03-15
Write to new 'enic reference' column in 'application_qualifications' table

TobyRet:
## Context

UK NARIC, the agency for equivalent qualifications has changed its name to UK ENIC. We point users in their direction quite a few times in the Apply journey so we need to all references to UK NARIC

- Write to 'enic_reference' column when creating / updating application qualifications
- Create data migration to backfill the 'enic_reference' column

# Guidance for review
- Backfill created via @despo 's new data migration rake task 😍. This task will run automatically on deploy. Will do a separate PR to clean up laterz ...

## Link to Trello card

## Things to check


davidgisbey (approved):

app/services/data_migrations/backfill_enic_reference_data.rb
```diff
@@ -0,0 +1,15 @@
+module DataMigrations
+  class BackfillEnicReferenceData
+    TIMESTAMP = 20210315122316
+    MANUAL_RUN = false
```
malcolmbaig, line 4: If this is run automatically after db:migrate, there's a non-zero chance that some records remain where naric_reference is present but enic_reference is nil. Not a huge deal, just worth making a note to check and running this service again manually if so.

malcolmbaig (approved):

app/services/data_migrations/backfill_enic_reference_data.rb
TobyRet: Yep agree @malcolmbaig - will check the data once deployed

PR #4295, opened 2021-03-15
Export documentation

raamSoftwire:
## Context

This is part of a wider effort to improve the visibility and documentation of these exports and is related to https://trello.com/c/VOjhcHfX/3159-data-directory.

Only a small number of exports currently have documentation. This ticket is to start adding documentation for the other exports. Depending on how long this takes, we might spin some more tech debt tickets out of this.

I think it worth someone taking a look now that I have done a few. Mainly in terms of the general approach. If there are any things we want to change or do differently then it's better to know sooner rather than later.

As part of a future PR I'm thinking about splitting out the common columns file as it is already getting a bit big. What do you think?

For each doc added

## Guidance to review

- commit by commit

## Link to Trello card

## Things to check


app/exports/_common_columns.yml
```diff
@@ -2,7 +2,7 @@ extract_date:
   type: string
   format: date-time
   description: Time when the report ran
-  example: 2019-06-13T10:44:31Z
+  example: 2020-11-01T00:00:00+00:00
```
raamSoftwire, line 5: This is how ISO 8601 strings are rendered in the exports, so I thought it worth updating these.

app/exports/candidate_journey_tracking_export.yml
```diff
@@ -0,0 +1,157 @@
...
+  offer_accepted:
+    type: string
+    format: date-time
+    description: Date the candidate accepted an offer
+    example: 2020-11-01T00:00:00+00:00
+
+  pending_conditions:
+    type: string
+    format: date-time
+    description: Date the candidate accepted an offer
```
raamSoftwire, line 128: This feels like a bug. For now I have documented the actual behaviour (in case I have misunderstood something). But this is exactly the same as `offer_accepted`. 

https://github.com/DFE-Digital/apply-for-teacher-training/blob/f9b37c806e57cac171f7f74a10c6176bd696686e/app/services/support_interface/candidate_journey_tracker.rb#L92

app/services/support_interface/work_history_break_export.rb
```diff
@@ -27,12 +27,12 @@ def data_for_export
...
 
         output = {
-          'Candidate id' => application_form.candidate_id,
-          'Application id' => application_form.id,
-          'Application submitted' => submitted_at(application_form).strftime('%d/%m/%Y'),
-          'Course choice statuses' => application_form.application_choices.map(&:status).sort,
-          'Start of working life' => start_of_working_life(application_form),
-          'Total time in employment (months)' => total_time_in_employment(application_form),
+          candidate_id: application_form.candidate_id.to_s,
+          application_form_id: application_form.id.to_s,
```
raamSoftwire, line 31: I will remove these `to_s`

raamSoftwire:
> This is looking great so far. Should we make the tests also check that none of the custom columns are shadowing common columns?

Interesting, yes I think that's a good idea

app/exports/work_history_break_export.yml
```diff
@@ -0,0 +1,90 @@
+common_columns:
+  - candidate_id
+  - application_form_id
+  - submitted_at
+
+custom_columns:
+  course_choice_statuses:
+    type: array
+    description: An array of course choice statuses
+    enum:
```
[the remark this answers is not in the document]
raamSoftwire: I had a short poke around but didn't see any way of doing it. But I don't have loads of yamlFoo™

app/exports/candidate_application_feedback_export.yml
```diff
@@ -0,0 +1,26 @@
...
+  page_title:
+    type: string
+    description: Page title
+
+  feedback:
+    type: string
+    description: Feedback left by candidate
+
+  consent_to_be_contacted:
+    type: string
```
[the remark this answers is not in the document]
raamSoftwire: Yes, we should change that

app/exports/candidate_feedback_export.yml
```diff
@@ -0,0 +1,25 @@
+common_columns:
+  - full_name
+  - recruitment_cycle_year
+  - email
+  - phone_number
+  - submitted_at
+
+custom_columns:
+  satisfaction_level:
+    type: integer
```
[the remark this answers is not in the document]
raamSoftwire: I think it is a string of an integer, eg. "5". I was going by looking at the downloaded export.

`Minerva Koelpin,2021,minerva.koelpin@example.com,07770 220030,2020-12-14T09:42:33+00:00,very_satisfied,5,test`

But you're right it's a string. Good spot!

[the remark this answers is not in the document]
raamSoftwire: ⚠ 👆🏽 FYI that's fake data from QA.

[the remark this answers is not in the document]
raamSoftwire: Ignore me, I was looking at the wrong column. Yes that is a string 🤦🏽‍♂️

app/services/data_set_documentation.rb
```diff
@@ -12,4 +12,19 @@ def self.for(klass)
     used_common_columns = common_columns.slice(*spec['common_columns'])
     used_common_columns.merge(spec['custom_columns'] || {})
   end
+
```
raamSoftwire, line 15: I'm in two minds with this class and would appreciate some feedback. There are a few different options

- use instance variables, which means that we only need to load the yml files once. But this class is used in a few different places and currently has a neat concise notation in this format 

```<% DataSetDocumentation.for(DataAPI::TADExport).each do |column_name, column| %>
  <%= render DataSetAttributeComponent.new(column_name: column_name, column: column) %>
<% end %>```

- stick with using "static" methods but use something like `private_class_method` or `class << self` so that we can pull out some of the repeated logic into private methods. This helps to dry out the class, but it would still load the files twice.

It's only in our export test that it would load the files twice. For the rest of the uses it only uses the `self.for` method and so loads them once. So I'm leaning towards the second option https://stackoverflow.com/questions/29737435/ruby-class-with-static-method-calling-a-private-method

Comment (raamSoftwire) [reply app/services/data_set_documentation.rb:15]: I've just pushed those changes so you can see what they look like

File: app/services/data_set_documentation.rb [hunk 8]
```diff
@@ -1,30 +1,34 @@
...
+  end
+
+  def self.load_export_documentation(klass)
     name = klass.name.demodulize.underscore
 
     begin
-      spec = YAML.load_file(Rails.root.join("app/exports/#{name}.yml"))
+      return YAML.load_file(Rails.root.join("app/exports/#{name}.yml"))
     rescue Errno::ENOENT
       return false
```
raamSoftwire: I missed that this `false` was being relied upon to exit early from this method and this refactor broke that mechanism. I've reverted those changes now and I plan to leave it as it is. 

Useless there are any other suggestions?

raamSoftwire: See other comment for context

app/services/data_set_documentation.rb
duncanjbrown: :+1: I think loading twice is fine.

app/services/data_set_documentation.rb
duncanjbrown: If we expect everything to have documentation, should this throw an exception if the file doesn't exist?

app/services/data_set_documentation.rb
```diff
@@ -12,4 +12,19 @@ def self.for(klass)
...
+    begin
+      spec = YAML.load_file(Rails.root.join("app/exports/#{name}.yml"))
+    rescue Errno::ENOENT
+      return false
+    end
+
+    common_columns = YAML.load_file(Rails.root.join('app/exports/_common_columns.yml')).keys.map(&:to_sym)
+    custom_columns = spec['custom_columns'].keys.map(&:to_sym)
+
+    custom_columns & common_columns
```
duncanjbrown, line 28: :+1:

duncanjbrown:
This is good!

It's a little bit odd that the code which tests for shadowed columns effectively has to live in the production code rather than the test suite, but I don't think that's worth losing any sleep over.  I suppose you could throw an exception in DataSetDocumentation.for if any of the columns are shadowed?

app/services/data_set_documentation.rb
raamSoftwire: Yes, I think that is a nice solution, once everything has documentation. I will add that in as part of a later PR.

raamSoftwire:
>  you could throw an exception in DataSetDocumentation.for if any of the columns are shadowed?

I think that's a neat idea and removes the need for this extra method in the production code. I'll include that in the other refactor once all the documentation is done.

duncanjbrown (approved):

PR #4297, opened 2021-03-15
Update TAD export documentation to include deferred status

AgaDufrat:
When getting relevant applications we are checking `.where.not(submitted_at: nil)` which means that deferred applications are also included.

For more details see recent analysis: https://docs.google.com/document/d/14q7WsvV-_Bn6oophwIHcD0utZrFSzO9DkwIzM2Ky7A4/edit#heading=h.7heei0m7t6nz

### DIFF
--- a/app/exports/tad_export.yml
+++ b/app/exports/tad_export.yml
@@ -40,6 +40,7 @@ custom_columns:
     - conditions_not_met
     - declined
     - declined_by_default
+    - deferred
     - interviewing
     - offer
     - offer_deferred


app/exports/tad_export.yml
```diff
@@ -40,6 +40,7 @@ custom_columns:
     - conditions_not_met
     - declined
     - declined_by_default
+    - deferred
```
cwrw, line 43: We currently include `offer_deferred` (on line 46), I was wondering how/if `deferred` was different?

AgaDufrat:
Good point! It was on my TO-DO list but someone has done it already.

PR #4299, opened 2021-03-16
Support - replace 'Reinstate reference' with 'Undo refusal'

malcolmbaig:
## Context
The 'Reinstate reference' feature is intended to allow support agents to
put cancelled and refused references back into a `feedback_requested`
state. It currently has a few issues:

- Candidates can now re-request cancelled references on their own,
  making that aspect of the feature mostly redundant.
- An error is triggered for refused references because the
  ReferenceActionsPolicy doesn't currently permit the move from
  `feedback_refused` to `feedback_requested`.
- The copy on the confirmation page states that no email will be sent,
  however the current implementation invokes a service object that sends
  email.

Rather than address these issues independently, this commit reworks
'Reinstate reference' into an 'Undo refusal' feature. This discards
handling of cancelled references and only appears for references that
are `feedback_refused`. The feature is useful in scenarios where a
referee has accidentally declined a reference request and wants Support
to make it active again.

Screenshot_20210316_095810

## Guidance to review
- In the support interface, find an application with a reference in the `feedback_refused` state (or create this situation via the console)
- Click 'Undo refusal' and follow the flow through.

## Link to Trello card

## Things to check

### DIFF
--- a/app/controllers/support_interface/references_controller.rb
+++ b/app/controllers/support_interface/references_controller.rb
@@ -15,7 +15,7 @@ module SupportInterface
     def reinstate; end
 
     def confirm_reinstate
-      RequestReference.new.call(@reference)
+      @reference.update!(feedback_status: 'feedback_requested', audit_comment: 'Reversing reference refusal using Support interface')
       flash[:success] = 'Reference was reinstated'
       redirect_to support_interface_application_form_path(@reference.application_form)
     end
--- a/app/controllers/support_interface/references_controller.rb
+++ b/app/controllers/support_interface/references_controller.rb
@@ -15,7 +15,7 @@ module SupportInterface
     def reinstate; end
 
     def confirm_reinstate
-      @reference.update!(feedback_status: 'feedback_requested', audit_comment: 'Reversing reference refusal using Support interface')
+      UndoReferenceRefusal.new(@reference).call
       flash[:success] = 'Reference was reinstated'
       redirect_to support_interface_application_form_path(@reference.application_form)
     end
--- a/app/views/support_interface/references/reinstate.html.erb
+++ b/app/views/support_interface/references/reinstate.html.erb
@@ -4,20 +4,15 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <h1 class="govuk-heading-l">
-      Are you sure you want to reinstate this reference?
+      Are you sure you want to undo the referee's refusal to provide a reference?
     </h1>
 
     <p class="govuk-body">
-      Reinstating a reference will:
+      This will allow this referee to provide a reference
     </p>
 
-    <ul class="govuk-list govuk-list--bullet">
-      <li>allow this referee to provide a reference</li>
-      <li>prevent the candidate from providing a replacement referee</li>
-    </ul>
-
     <p class="govuk-body">
-      Reinstating a reference does <strong>not send email notifications</strong>.
+      Undoing a refusal does <strong>not send email notifications</strong>.
     </p>
 
     <%= render(SummaryCardComponent.new(rows: [
@@ -32,7 +27,7 @@
     ])) %>
 
     <%= form_with model: @reference, url: support_interface_reinstate_reference_path(@reference), method: :post do |f| %>
-      <%= f.govuk_submit 'Reinstate reference' %>
+      <%= f.govuk_submit 'Undo refusal' %>
     <% end %>
   </div>
 </div>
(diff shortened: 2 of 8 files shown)


app/controllers/support_interface/references_controller.rb
```diff
@@ -15,7 +15,7 @@ def confirm_cancel
     def reinstate; end
 
     def confirm_reinstate
-      RequestReference.new.call(@reference)
+      @reference.update!(feedback_status: 'feedback_requested', audit_comment: 'Reversing reference refusal using Support interface')
```
davidgisbey, line 18: we also need to set the `feedback_requested_at` timestamp to `nil` here

app/views/support_interface/references/reinstate.html.erb
```diff
@@ -4,20 +4,19 @@
...
     </p>
 
     <ul class="govuk-list govuk-list--bullet">
       <li>allow this referee to provide a reference</li>
-      <li>prevent the candidate from providing a replacement referee</li>
     </ul>
 
     <p class="govuk-body">
-      Reinstating a reference does <strong>not send email notifications</strong>.
+      Undoing a refusal does <strong>not send email notifications</strong>.
```
davidgisbey, line 19: Might be out of scope for this card, but I do wonder whether we should be sending an email here. Mainly just so the referee has something to click on. When I've done this via console I've tended to send a chaser.
 
The other option is that we just tell support colleagues that they need to go back and click on the original link.

I guess there are situations where we might not want to send an email though.

What do you think?

app/controllers/support_interface/references_controller.rb
duncanjbrown: Why's that? (Just curious)

app/views/support_interface/references/reinstate.html.erb
```diff
@@ -4,20 +4,19 @@
...
     </h1>
 
     <p class="govuk-body">
-      Reinstating a reference will:
+      This will:
     </p>
 
     <ul class="govuk-list govuk-list--bullet">
       <li>allow this referee to provide a reference</li>
-      <li>prevent the candidate from providing a replacement referee</li>
```
duncanjbrown, line 16: Maybe we could lose the bullets now there's only one item?

duncanjbrown (commented):
This looks good to me. Deferring to @davidgisbey re `feedback_requested_at`!

app/controllers/support_interface/references_controller.rb
davidgisbey: So we added some timestamp columns for data reporting purposes that track change in state. Previously we were having to dig through audits to determine how long references took to be provided/refused etc.

app/controllers/support_interface/references_controller.rb
```diff
@@ -15,7 +15,7 @@ def confirm_cancel
...
       flash[:success] = 'Reference was reinstated'
```
duncanjbrown, line 19: Maybe this could read "The referee is able to provide a reference again" or sth?

app/controllers/support_interface/references_controller.rb
duncanjbrown: OK. Why set it to `nil`, though — isn't that just discarding information?

davidgisbey: The idea of this is to revert the state change. I wouldn't have thought we would want a reference in the feedback_provided state with a feedback_refused_at timestamp?

I mean it doesn't hurt i guess, but if someone does something like

`ApplicationReference.where.not(feedback_refused_at: nil)` for reporting it's going to get picked up (although using the `ApplicationReferece.feedback_refused` enum scope would make more sense tbh)

app/views/support_interface/references/reinstate.html.erb
malcolmbaig: Good question. I'd tend towards producing the minimal set of side effects possible from this action. I think this happens infrequently enough that asking support colleagues to tell referees to use the original email is sustainable. Can always manually send the email if for some reason it's asked for especially.

app/controllers/support_interface/references_controller.rb
malcolmbaig: I like the idea of keeping as much info as possible - the fact that the timestamp was populated at some point is relevant info about the history of a given reference.

Fortunately the audit history will reflect that, so I think we're good to nullify the timestamp on the record itself.

davidgisbey: Yea i certainly think we should be retaining the audit 👍

davidgisbey (approved):
lgtm!

PR #4301, opened 2021-03-16
Interview csv fixes

benswannack:
## Context
There are a couple of issues with the interviews export in prod and QA
Account for interviews that have been hard deleted on QA.
Lazy-get the audits to avoid a timeout

## Guidance to review
Make sense?

## Link to Trello card

https://trello.com/c/3mIxkctF/3525-interview-changes-csv-export-query-times-out-on-production

## Things to check

### DIFF
--- a/app/services/support_interface/interview_changes_export.rb
+++ b/app/services/support_interface/interview_changes_export.rb
@@ -1,5 +1,7 @@
 module SupportInterface
   class InterviewChangesExport
+    include AuditHelper
+
     def data_for_export
       rows = interview_audits.find_each(batch_size: 100).lazy.map do |audit|
         row_for_audit(audit)
@@ -95,7 +97,7 @@ module SupportInterface
     end
 
     def audit_user(audit)
-      if audit.user_type == 'SupportUser'
+      if change_by_support?(audit)
         'Support'
       elsif audit.user.present?
         audit.user.email_address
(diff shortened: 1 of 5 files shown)


app/services/support_interface/interview_changes_export.rb
```diff
@@ -1,13 +1,17 @@
 module SupportInterface
   class InterviewChangesExport
     def data_for_export
-      interview_audits.find_each(batch_size: 100).map do |audit|
+      rows = interview_audits.find_each(batch_size: 100).lazy.map do |audit|
         row_for_audit(audit)
       end
+
+      rows.compact
```
cwrw, line 8: I think we need `to_a` here as we can't call compact on a `Enumerator::Lazy`

cwrw (commented):
Unrelated to this PR, something I ran into when running this locally:
If a user does something from the console `audit.user.email_address` isn't working, we might need to add another check for those type of audits

benswannack:
> Unrelated to this PR, something I ran into when running this locally:
> If a user does something from the console `audit.user.email_address` isn't working, we might need to add another check for those type of audits

I've attempted to address this in a consistent way here, since it's a question we seem to ask in many places. LMK what you think

spec/helpers/audit_helper_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+
+    context 'change was made in the rails console' do
+      let(:username) { 'Developer via the Rails console' }
+
+      it { is_expected.to eq(true) }
+    end
+
+    context 'change was made by a different user' do
+      let(:username) { 'Ghost man' }
```
cwrw, line 30: 👻

cwrw (approved):
looks good with the helper, thanks for addressing 👍

PR #4302, opened 2021-03-16
Add JS specs to autosuggests

TobyRet:
## Context

Some of the autosuggests that feature on Apply have a fair degree of complexity that should be covered by specs. If an autosuggest is not working, we should try and know about this prior to deploy.  

- Refactor autosuggests to make them easier to test. This has also reduced a lot of duplicated code! 😹 
- Add snapshot testing for autosuggests. 

## Guidance to review
- This is a 'proof of concept' - if we're happy with this I can apply to the remainder of the autocompletes
- I think it may be worth hooking up Raven JS / Sentry to notify us if an autocomplete / autosuggest fails to load. Currently we only capture the error and print to the console. Can do this in a separate PR

To run the JS specs:
```sh
$ npm run test
```

## Link to Trello card
https://trello.com/c/cZ7QpsD4/2995-high-add-js-specs-for-autocomplete

## Things to check


george-bit (approved):
Hang on, sorry mate not what you were asking :/

vigneshmsft:
we run `yarn jest` as part of the docker image build.. 
are these tests different?

TobyRet:
> we run `yarn jest` as part of the docker image build..
> are these tests different?

Good to know - so no need to amend the build. I'll remove the commit

frankieroberto:
@TobyRet whilst we're looking at this, is it worth rethinking the init process to avoid the hardcoded list of IDs? We could do something more [similar to `govuk-frontend`](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/all.js#L21-L24) and use a `data-module` attribute to trigger the initialisation?

frankieroberto:
> @frankieroberto - yeah I agree hardcoding IDs feels a bit wrong but there are some cases where some customisable behaviour needs to be passed in to the initialiser. E.g. https://github.com/DFE-Digital/apply-for-teacher-training/pull/4302/files#diff-c817e85983eccc3aab451e5a7340519176fd9b0205b0beb333661341eb939919R31

Ah, I didn't spot that. Maybe for now we could just create a card in the backlog to come back to this. Feels like there should be a way to create a generic initialiser that covers both behaviours – but that might be scope creep for this PR?

TobyRet:
> > @frankieroberto - yeah I agree hardcoding IDs feels a bit wrong but there are some cases where some customisable behaviour needs to be passed in to the initialiser. E.g. https://github.com/DFE-Digital/apply-for-teacher-training/pull/4302/files#diff-c817e85983eccc3aab451e5a7340519176fd9b0205b0beb333661341eb939919R31
> 
> Ah, I didn't spot that. Maybe for now we could just create a card in the backlog to come back to this. Feels like there should be a way to create a generic initialiser that covers both behaviours – but that might be scope creep for this PR?

Yes in terms of scope, I believe this card is mainly about putting the missing specs in place, in part so that we can safely make functional improvements in future 👍  Yes to creating a backlog card for your suggestion though @frankieroberto 
Do you agree @davidgisbey ?

frankieroberto (approved):

george-bit (approved):
looking good Toby! just ran the tests locally and noticed that `degree-grade-autosuggest.spec.js.snap` is marked as obsolete. Presumably this has been supplanted by your new snapshots? If so can it just be deleted?

TobyRet:
> looking good Toby! just ran the tests locally and noticed that `degree-grade-autosuggest.spec.js.snap` is marked as obsolete. Presumably this has been supplanted by your new snapshots? If so can it just be deleted?

Yep!

davidgisbey (approved):

PR #4303, opened 2021-03-16
Use the offered_option in the ApplicationDataExport

davidgisbey:
## Context

This support ticket https://becomingateacher.zendesk.com/agent/tickets/12511 noted that the application export is using the course the candidate applied too, not the offered course.

They noted that this could cause issues down the line in terms of finance/bursaries etc. 

- Update the export to use the offered_course_option for the course, site and provider columns

## Guidance to review

Have I got everything?

## Link to Trello card

https://trello.com/c/QQKinp9r/3498-ensure-offered-course-data-is-surfaced-in-csv-exports

## Things to check


app/controllers/provider_interface/application_data_export_controller.rb
```diff
@@ -21,6 +21,7 @@ def export
               :provider,
               :accredited_provider,
               :site,
+              :offered_course_option,
```
cwrw, line 24: do we need the same includes as the `course_option` here? `%i[course site]`

davidgisbey: Yup. Good shout

cwrw (approved):

PR #4304, opened 2021-03-16
Update notification_preferences when updating send_notifications

steventux:
## Context

Notification preferences should be created and/or updated when #send_notifications is modified to ensure we have notification preference parity for feature launch.

Adds a namespace-neutral service which handles saving notification preferences. For the time being this updates `ProviderUser#send_notifications` and creates and/or updates `ProviderUserNotificationPreferences` for the same user.

The service is called from both provider interface and support interface controllers where the user preferences can be changed.

## Guidance to review

Does this PR miss any other places where the notification preferences can be written?

## Link to Trello card

https://trello.com/c/Wm7NAOrk/3449-migrate-provider-user-notification-settings-to-new-structure

## Things to check

### DIFF
--- a/app/services/save_provider_user_notification_preferences.rb
+++ b/app/services/save_provider_user_notification_preferences.rb
@@ -1,22 +1,47 @@
 class SaveProviderUserNotificationPreferences
-  attr_reader :notification_params, :provider_user
+  attr_reader :provider_user
 
-  def initialize(provider_user:, notification_params: {})
+  def initialize(provider_user:)
     @provider_user = provider_user
-    @notification_params = notification_params
   end
 
-  def call!
-    return false unless notification_params.key?(:send_notifications)
+  def backfill_notification_preferences!(send_notifications:)
+    return false if send_notifications.nil?
 
-    provider_user.update!(send_notifications: notification_params[:send_notifications])
-    provider_user_notification_preferences.update_all_preferences(notification_params[:send_notifications])
+    update_provider_user!(send_notifications)
+    provider_user_notification_preferences.update_all_preferences(send_notifications)
+  end
+
+  def update_all_notification_preferences!(notification_preferences_params: {})
+    return false if notification_preferences_params.empty?
+
+    provider_user_notification_preferences.update!(notification_preferences_params)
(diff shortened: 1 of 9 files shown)


AgaDufrat (approved):
👌 I like this approach because it'll handle creating notification preferences when ProviderUsers are created via the SaveProviderUser service and in the rails console.

app/models/provider_user.rb
```diff
@@ -66,4 +67,11 @@ def can_manage_organisations?
   def downcase_email_address
     self.email_address = email_address.downcase
   end
+
+  def update_notification_preferences
+    if saved_change_to_attribute?(:send_notifications)
+      self.notification_preferences ||= ProviderUserNotificationPreferences.create!(provider_user: self)
+      notification_preferences.update_all_preferences(send_notifications)
```
cwrw, line 74: unrelated to this PR but might impact performance: can we amend the implementation of `update_all_preferences` in `ProviderUserNotificationPreferences` to do one update instead of the number of notification preferences?

steventux: This is outdated now as I've gone for a discreet service called from relevant controllers.

cwrw (commented):
Looks great, and works with updating a provider user's preferences from the support and provider interface. While testing I noticed that we are missing creating the new notification preferences when creating a provider user through the support interface or inviting a provider user through the provider interface. Could we invoke creating the new preferences there as well? Let me know your thoughts

app/services/save_provider_user_notification_preferences.rb
```diff
@@ -0,0 +1,47 @@
...
+  def provider_user_notification_preferences
+    @provider_user_notification_preferences ||= provider_user.notification_preferences ||
+      ProviderUserNotificationPreferences.create!(provider_user: provider_user)
+  end
+
+  def send_notifications_from_notificaion_prefernces
+    values =
+      provider_user_notification_preferences
+        .attributes
+        .with_indifferent_access
```
AgaDufrat, line 39: 🤩

AgaDufrat (approved):
LGTM

PR #4311, opened 2021-03-17
Change messaging when candidate adds already chosen course

malcolmbaig:
## Context
We have a validation in place which ensures that a candidate can't add the
same course to their application twice.

Rather than style this as an error message in the UI, handle the
situation more gracefully by taking the candidate to the course choices
review page and display an info banner informing them they've already
added that course.

**Before**
image

**After**
image

## Guidance to review
- Add a course to an application
- Click back button and add same course

## Link to Trello card

https://trello.com/c/UxJswxbz
## Things to check

### DIFF
--- a/app/forms/candidate_interface/pick_course_form.rb
+++ b/app/forms/candidate_interface/pick_course_form.rb
@@ -65,6 +65,14 @@ module CandidateInterface
       @course ||= provider.courses.find(course_id)
     end
 
+    def already_picked_this_one?
+      application_form.application_choices.includes([:course]).any? { |application_choice| application_choice.course == course }
+    end
+
+    def already_added_error
+      I18n.t!('errors.application_choices.already_added', course_name_and_code: course.name_and_code)
+    end
+
   private
 
     def provider
@@ -74,8 +82,8 @@ module CandidateInterface
     def user_cant_apply_to_same_course_twice
       return if course_id.blank?
 
-      if application_form.application_choices.includes([:course]).any? { |application_choice| application_choice.course == course }
-        errors[:base] << I18n.t!('errors.application_choices.already_added', course_name_and_code: course.name_and_code)
+      if already_picked_this_one?
+        errors[:base] << already_added_error
       end
     end
   end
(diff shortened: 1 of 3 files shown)


spec/system/candidate_interface/course_selection/candidate_adds_a_course_theyve_already_chosen_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def then_i_am_taken_to_the_review_page
+    expect(page).to have_current_path candidate_interface_course_choices_review_path
+  end
+
+  def and_i_am_informed_that_ive_already_added_that_course
+    within '.govuk-notification-banner--info' do
+      expect(page).to have_content "You have already added #{@choice.course.name_and_code}"
+    end
+    expect(current_candidate.current_application.application_choices.size).to eq 1
```
stevehook, line 42: Do you need a `reload` in here to pick up any newly added application choice? (not that there will be one!)

app/forms/candidate_interface/pick_course_form.rb
```diff
@@ -65,6 +65,14 @@ def course
       @course ||= provider.courses.find(course_id)
     end
 
+    def already_picked_this_one?
+      application_form.application_choices.includes([:course]).any? { |application_choice| application_choice.course == course }
```
stevehook, line 69: Could you do the comparison on `application_choice.course_id` to avoid needing to `include` course?

stevehook (commented):
Should we update the site selection action too? When I tested it all works nicely but I get a red Warning rather than the blue one if I add a course, hit back once, pick a different site and then proceed.

image

app/forms/candidate_interface/pick_course_form.rb
malcolmbaig: Not a column on the `application_choice` record unfortunately, the id is obtained via the associated `course_option`.

spec/system/candidate_interface/course_selection/candidate_adds_a_course_theyve_already_chosen_spec.rb
malcolmbaig: Yes, I think so - have replaced with a simpler query

malcolmbaig:
> Should we update the site selection action too? When I tested it all works nicely but I get a red Warning rather than the blue one if I add a course, hit back once, pick a different site and then proceed.

Yes, definitely. It's revealed some refactoring that could make potentially make this easier to do across the course/site/study mode controllers. I'm going to create a separate PR for this, branching off this one.

stevehook (approved):
:+1:

PR #4313, opened 2021-03-17
Redirect to the interviews index if the store has been cleared

steventux:
## Context

Clicking back after creating or editing an interview takes the user to the 'check' step of the interviews wizard.
The store will be cleared on creating or editing the interview so any attempt to resubmit will cause a 500 error.

Check to see if the store is empty on check and commit steps as an indication that there's no interview to check or save and that the user should be redirected to the interviews index.

## Link to Trello card

https://trello.com/c/rTjsws8z/3493-pressing-back-button-after-creating-an-interview-returns-a-500

## Things to check

### DIFF
--- a/app/controllers/provider_interface/interviews_controller.rb
+++ b/app/controllers/provider_interface/interviews_controller.rb
@@ -4,6 +4,7 @@ module ProviderInterface
     before_action :interview_flag_enabled?
     before_action :requires_make_decisions_permission, except: %i[index]
     before_action :confirm_application_is_in_decision_pending_state, except: %i[index]
+    before_action :redirect_to_index_if_store_cleared, only: %i[check commit]
 
     def index
       application_at_interviewable_stage = ApplicationStateChange::INTERVIEWABLE_STATES.include?(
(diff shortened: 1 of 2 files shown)


despo:
We encountered a similar issue when working on the make offer flow. However, when the user clicks back the action won't be re-issued, which will still cause a 500 as it will attempt to call the service without having the required details..

What we did however was to add a `save` scope on all mandatory validations to ensure that they all take place before calling the service again (in this commit https://github.com/DFE-Digital/apply-for-teacher-training/pull/4285/commits/9cdc1edec017268715508fd33681840068193acb)

steventux:
@despo In this scenario the 500 doesn't occur when the user clicks back, the error is thrown when they click back and then attempt to resubmit from the (now stale) confirmation page.
The problem lies when the wizard is invalid in the commit action, this attempts to render the check page with missing dates, provider etc. Arguably this should also be rectified.

despo (approved):
As discussed I will add a separate card for the validation issue identified in make offer that I mentioned earlier, as we need to resolve it on the other wizards as well!

PR #4293, opened 2021-03-15
Feature metric dashboard 2.0 part 3 - Carry Over

stevehook:
## Context

We are adding a new set of metrics to the feature metrics dashboard. This 3rd PR follows on from https://github.com/DFE-Digital/apply-for-teacher-training/pull/4288 and adds _Carry over_ counts. One or more PRs to follow.

## Changes proposed in this pull request

image

## Guidance to review

- 

## Link to Trello card

https://trello.com/c/1X9y1JPl/2981-feature-metric-dashboard-20

## Things to check

- [x] This code does not rely on migrations in the same Pull Request
- [x] If this code includes a migration adding or changing columns, it also backfills existing records for consistency
- [x] API release notes have been updated if necessary
- [x] New environment variables have been [added to the Azure config](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/docs/environment-variables.md#azure-hosting-devops-pipeline)

PR #4314, opened 2021-03-17
Qualifications (A level) metrics for Feature Metrics dashboard 2.0 (part 4/4)

stevehook:
## Context

We are adding a new set of metrics to the feature metrics dashboard. This 3rd PR follows on from https://github.com/DFE-Digital/apply-for-teacher-training/pull/4293 and adds qualification metrics, specifically for A and A/S levels. This is the last of the PRs that adds new functionality for the dashboard.

- [x] Add new dashboard section for the percentage of applications (submitted within the time ranges) that have specified at least 1 and at least 3 A or A/S levels.

image

- [x] Rebase onto master after https://github.com/DFE-Digital/apply-for-teacher-training/pull/4293 gets merged 

## Guidance to review

- Does the query in `QualificationsFeatureMetrics` make sense?
- Is the copy right?
- With this PR the new dashboard is now feature complete (I think) so should we be refactoring the implementation at this stage? e.g. there are lots of model classes for feature metrics now, all top-level. There is also a CandidateInterface::FindChangedApplyAgainApplications service. I think these should be re-organised. Should we put make them all services? stick them under the SupportInterface or maybe even a SupportInterface::FeatureMetrics namespace? (this can be done as a separate PR but interested in what people think).

Here's a screenshot of all the Feature Metrics 2.0 additions:

image

## Link to Trello card

## Things to check

### DIFF
--- a/app/models/qualifications_feature_metrics.rb
+++ b/app/models/qualifications_feature_metrics.rb
@@ -1,0 +1,51 @@
+class QualificationsFeatureMetrics
+  include ActionView::Helpers::NumberHelper
+
+  def formatted_a_level_percentage(
+    minimum_count,
+    start_time,
+    end_time = Time.zone.now.end_of_day
+  )
+    applications_with_a_levels_count = applications_with_a_levels_grouped(
+      minimum_count,
+      start_time,
+      end_time,
+    ).count.keys.count
+    all_applications_count = all_applications(start_time, end_time).count
+
+    return 'n/a' if all_applications_count.zero?
+
+    format_as_percentage(applications_with_a_levels_count.to_f / all_applications_count)
+  end
+
+private
+
+  def all_applications(start_time, end_time)
+    ApplicationForm
+      .where(
+        recruitment_cycle_year: RecruitmentCycle.current_year,
+      )
+      .where('application_forms.submitted_at BETWEEN ? AND ?', start_time, end_time)
+      .where(recruitment_cycle_year: RecruitmentCycle.current_year)
+  end
+
+  def applications_with_a_levels_grouped(minimum_count, start_time, end_time)
+    all_applications(start_time, end_time)
+      .joins(:application_qualifications)
+      .where('application_qualifications.level': 'other')
+      .where("application_qualifications.qualification_type IN ('A level', 'A/S level')")
+      .having('count(application_qualifications.id) >= ?', minimum_count)
+      .group('application_forms.id')
+  end
+
+  def format_as_percentage(ratio)
+    return 'n/a' if ratio.nil?
+
+    percentage = number_with_precision(
+      100.0 * ratio,
+      precision: 1,
+      strip_insignificant_zeros: true,
+    )
+    "#{percentage}%"
+  end
+end
--- a/app/models/qualifications_feature_metrics.rb
+++ b/app/models/qualifications_feature_metrics.rb
@@ -6,11 +6,11 @@ class QualificationsFeatureMetrics
     start_time,
     end_time = Time.zone.now.end_of_day
   )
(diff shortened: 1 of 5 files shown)


app/models/qualifications_feature_metrics.rb
```diff
@@ -0,0 +1,51 @@
...
+  def formatted_a_level_percentage(
+    minimum_count,
+    start_time,
+    end_time = Time.zone.now.end_of_day
+  )
+    applications_with_a_levels_count = applications_with_a_levels_grouped(
+      minimum_count,
+      start_time,
+      end_time,
+    ).count.keys.count
```
davidgisbey, line 13: Is this right? Just going to test locally.

spec/models/qualifications_feature_metrics_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      end
+    end
+
+    context 'with applications with A levels' do
+      it 'returns 0 when there are no applications with A levels' do
+        create_application
+        expect(feature_metrics.formatted_a_level_percentage(1, 1.month.ago)).to eq('0%')
+      end
+
+      it 'returns 50% when there one of two applications has A levels' do
```
davidgisbey, line 31: While true, this isn't descriptive of what's in the test

app/models/qualifications_feature_metrics.rb
stevehook: This is a bit messy. The first count is a SQL `SELECT count(...) ...` for the `GROUP BY` query and returns a hash, the second count is a Ruby `count` on the hash. Maybe I need to refactor this to make it a bit clearer (it confused me).

davidgisbey (approved):
Looks great Steve. A couple of minor nits

app/models/qualifications_feature_metrics.rb
davidgisbey: Yea i assumed it was something to with group by. I just couldn't see how the method calls were interacting with the objects in my head.

app/views/support_interface/performance/feature_metrics_dashboard.html.erb
```diff
@@ -325,4 +325,60 @@
...
     </div>
   </div>
+
+  <h3 class="govuk-heading-m govuk-!-font-size-27 govuk-!-margin-top-8">Qualifications</h3>
+  <div id="qualifications_dashboard_section" class="govuk-grid-row govuk-!-margin-bottom-4">
+    <div class="govuk-grid-column-one-half">
+      <div id="headline-stat" class="govuk-!-margin-bottom-4">
+        <%= render SupportInterface::TileComponent.new(
+          count: @dashboard.read_metric(:pct_applications_with_one_a_level),
+          label: 'candidates input 1 or more AS & A-levels',
```
paulrobertlloyd, line 335:
```suggestion
          label: 'candidates added 1 or more AS or A level',
```

app/views/support_interface/performance/feature_metrics_dashboard.html.erb
```diff
@@ -325,4 +325,60 @@
...
+            size: :reduced,
+          ) %>
+        </div>
+      </div>
+    </div>
+    <div class="govuk-grid-column-one-half">
+      <div id="headline-stat" class="govuk-!-margin-bottom-4">
+        <%= render SupportInterface::TileComponent.new(
+          count: @dashboard.read_metric(:pct_applications_with_three_a_levels),
+          label: 'candidates input 3 or more AS & A-levels',
```
paulrobertlloyd, line 361:
```suggestion
          label: 'candidates added 3 or more AS or A level',
```

paulrobertlloyd (approved):
Just a couple of copy tweaks – we say ‘A levels’ throughout the service.

PR #4315, opened 2021-03-18
Expect exact matches on api request paths when filtering

steventux:
## Context

This test was failing intermittently when one of the factory generated applications had a path which was a substring of another.
The `have_content` matcher was not precise enough to distinguish the substring eg. `/api/v1/applications/99` and `/api/v1/applications/9999/offer`

Switch to `have_selector` with an exact text match to fix.

## Link to Trello card

Broke master.

## Things to check

### DIFF
--- a/spec/system/support_interface/view_vendor_api_requests_spec.rb
+++ b/spec/system/support_interface/view_vendor_api_requests_spec.rb
@@ -68,9 +68,9 @@ RSpec.feature 'Vendor API Requests' do
   end
 
   def then_i_see_the_api_request
-    expect(page).to have_content('/api/v1/applications/9999/offer')
-    expect(page).to have_content(vendor_api_path(@first_application_choice))
-    expect(page).to have_content(vendor_api_path(@last_application_choice))
+    expect(page).to have_selector('p.govuk-body', exact_text: '/api/v1/applications/9999/offer')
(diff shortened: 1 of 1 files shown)


spec/system/support_interface/view_vendor_api_requests_spec.rb
```diff
@@ -69,8 +69,8 @@ def when_i_visit_the_vendor_api_requests_page
 
   def then_i_see_the_api_request
     expect(page).to have_content('/api/v1/applications/9999/offer')
```
AgaDufrat, line 71: Do we want to update this one as well?
```suggestion
    expect(page).to have_selector('p.govuk-body', exact_text: '/api/v1/applications/9999/offer')
```

AgaDufrat (approved):

spec/system/support_interface/view_vendor_api_requests_spec.rb
steventux: Oops yes missed that will update :+1:

steventux: @AgaDufrat updated https://github.com/DFE-Digital/apply-for-teacher-training/pull/4315/commits/d279affce4b41ef82f16f4aaa89d3d7e8da29ca9

PR #4316, opened 2021-03-18
Add service to validate course detail combination.

richardpattinson:
## Context

Add service to check whether course details are valid.

## Link to Trello card

https://trello.com/c/tr4toE9e/3530-add-offer-validation-service

## Things to check

### DIFF
--- a/spec/services/offered_course_option_details_check_spec.rb
+++ b/spec/services/offered_course_option_details_check_spec.rb
@@ -1,0 +1,44 @@
+require 'rails_helper'
+
+RSpec.describe OfferedCourseOptionDetailsCheck do
+  let(:course_option) { create(:course_option, study_mode: :full_time) }
+
+  it 'is successful when all the provided details correspond to the course option' do
+    service = described_class.new(provider_id: course_option.provider.id,
+                                  course_id: course_option.course.id,
+                                  course_option_id: course_option.id,
+                                  study_mode: course_option.study_mode)
+
+    expect { service.validate! }.not_to raise_error
+  end
+
+  it 'throws an InvalidProviderError when the provider does not correspond to the course option' do
+    provider = create(:provider)
+    service = described_class.new(provider_id: provider.id,
+                                  course_id: course_option.course.id,
+                                  course_option_id: course_option.id,
+                                  study_mode: course_option.study_mode)
+
+    expect { service.validate! }.to raise_error(OfferedCourseOptionDetailsCheck::InvalidStateError, 'Invalid provider for CourseOption')
+  end
+
+  it 'throws an InvalidCourseError when the course does not correspond to the course option' do
+    course = create(:course)
+    service = described_class.new(provider_id: course_option.provider.id,
+                                  course_id: course.id,
+                                  course_option_id: course_option.id,
+                                  study_mode: course_option.study_mode)
+
+    expect { service.validate! }.to raise_error(OfferedCourseOptionDetailsCheck::InvalidStateError, 'Invalid course for CourseOption')
+  end
+
+  it 'throws an InvalidStudyModeError when the study mode does not correspond to the course option' do
+    study_mode = :part_time
+    service = described_class.new(provider_id: course_option.provider.id,
+                                  course_id: course_option.course.id,
+                                  course_option_id: course_option.id,
+                                  study_mode: study_mode)
+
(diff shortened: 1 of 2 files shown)


duncanjbrown (commented):
This looks like it will definitely throw exceptions reliably. Sorry to ask this question so late (!), but I'd like to understand the reasoning behind throwing a series of individual exceptions rather than calling `CourseOption.find_by!(the params)`. Unless we're going to handle all the exceptions differently?

despo:
@duncanjbrown our understanding is that eventually this check will also be used by the VendorAPI Decisions controller, so this enables us to move away from using active model validation errors yet providing more details about what is wrong in the provided data when calling the service.

app/services/course_option_check.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  attr_reader :provider_id, :course_id, :course_option_id, :study_mode, :location_id, :recruitment_cycle_year
+
+  def initialize(
+    provider_id:,
+    course_id:,
+    course_option_id:,
+    study_mode:,
+    location_id:,
+    recruitment_cycle_year: RecruitmentCycle.current_year
```
cwrw, line 31: we don't seem to be using this attribute?

duncanjbrown:
> @duncanjbrown our understanding is that eventually this check will also be used by the VendorAPI Decisions controller, so this enables us to move away from using active model validation errors yet providing more details about what is wrong in the provided data when calling the service.

This feels a bit like using exceptions for control flow. In the current design wouldn't it be:

```ruby
begin
  CourseOptionCheck.new(args)
rescue CourseOptionCheck:: InvalidProviderError
  # format the error for our client
rescue CourseOptionCheck::InvalidCourseError
  # format the error for our client
rescue CourseOptionCheck::InvalidSiteError
  # format the error for our client
rescue CourseOptionCheck::InvalidStudyModeError
  # format the error for our client
end

# carry on
```

In this design we can catch at most one exception, and it's quite verbose to handle them. We would also need to enumerate all the exceptions here and in any other clients of this service.

With an errors-style interface you could be more like we are in `ProviderAuthorisation`

```ruby
check = CourseOptionCheck.new(args)
if check.valid?
  # carry on
else
  # pass check.errors (could be a plain hash or ActiveRecord::Errors or sth else) back to our client who knows what to do with it
end
```

This way the catching and the formatting of the errors is encapsulated in the check operation and it can pass back a plain data structure which is consistent across all clients.

app/services/course_option_check.rb
despo: We are indeed not using it. We were thinking that that would ensure that the provided course option will be in the current recruitment cycle but I guess if we go down that route then we need to check if the course is open on apply etc which is a bit excessive I think.

app/services/course_option_check.rb
```diff
@@ -0,0 +1,35 @@
+class CourseOptionCheck
```
mnacos, line 1: I think `CourseOptionCheck` is a little too general and implies this is general-purpose course option validation to be re-used, when, in fact, it is very specific to the new wizard flow.

Everything else in the app expects just a `course_option`, not a combination of attributes, so I'd rename this to `OfferedCourseDetailsValidator` or similar. It should probably also be a validator rather than a service, with support for raising exceptions rather than returning errors if required.

mnacos (commented):
Left some suggestions around renaming and moving this validation service into validators. If exceptions are required, we could expose a `validate!` method, in the same vein as https://api.rubyonrails.org/classes/ActiveModel/Validations.html#method-i-validate-21

app/services/course_option_check.rb
despo: Agreed. We could do with a better name!

spec/services/offered_course_option_details_check_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'rails_helper'
+
+RSpec.describe OfferedCourseOptionDetailsCheck do
+  let(:course_option) { create(:course_option, study_mode: :full_time) }
+
+  it 'throws an InvalidProviderError when the provider does not correspond to the course option' do
```
cwrw, line 6: can you add a test for a valid course option?

cwrw (approved):
Looks good and as discussed we can plug this into the wizard in the next step, just the one comment

spec/services/offered_course_option_details_check_spec.rb
despo: Done ✅

PR #4318, opened 2021-03-18
Adding a Data Directory to the Performance tab

JR-G:
## Context

Currently it's not easy for BAs to see what data our exports can provide. This PR builds a centralised data directory, listing all exports. 

Each export will also include a link that takes the user to an 'export information page' detailing all the columns contained within that export (as part of the work from [#4295](https://github.com/DFE-Digital/apply-for-teacher-training/pull/4295)), a button to generate the export and also another link to view the history.

This PR introduces a data directory section:
Screenshot 2021-03-22 at 22 57 33

In addition, each export now has its own page, providing more detail on what the export contains and its history:
Screenshot 2021-03-22 at 22 57 44

## Guidance to review

- Can the existing tab - now named `Export history` - be removed?
- Some of the methods feel a bit "hacky" in how I'm grabbing the export ids.. any obvious signs for improvement here?

## Link to Trello card

https://trello.com/c/VOjhcHfX/3159-data-directory

## Things to check


app/controllers/support_interface/data_exports_controller.rb
```diff
@@ -8,6 +8,18 @@ def index
...
+    def directory
+      @export_types = DataExport::EXPORT_TYPES
+    end
+
+    def view_export_information
+      @data_export = DataExport::EXPORT_TYPES[params[:data_export_id].to_sym]
+    end
+
+    def view_download_history
+      @data_exports = DataExport.includes(:initiator).where(name: params[:data_export_id].humanize).order(id: :desc).page(params[:page] || 1).per(30)
```
duncanjbrown, line 20: Maybe not for this PR, but it's a bit wonky to map back and forth from the name to the symbol here — would it be better to annotate DataExport records with the `type`?

app/controllers/support_interface/data_exports_controller.rb
```diff
@@ -8,6 +8,18 @@ def index
       @data_exports = DataExport.includes(:initiator).order(id: :desc).page(params[:page] || 1).per(30)
     end
 
+    def directory
+      @export_types = DataExport::EXPORT_TYPES
+    end
+
+    def view_export_information
+      @data_export = DataExport::EXPORT_TYPES[params[:data_export_id].to_sym]
```
duncanjbrown, line 16: Is this more of a `type` than an `id`? (Individual CSV exports have an `id`)

app/views/support_interface/data_exports/view_export_information.html.erb
```diff
@@ -0,0 +1,34 @@
...
+  }]) %>
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <p class="govuk-body"><%= @data_export.fetch(:description) %></p>
+
+    <%= govuk_button_to 'Generate new export', support_interface_data_exports_path(export_type_id: params[:data_export_id]), class: 'govuk-button--secondary' %>
+
+    <p class="govuk-body"><%= govuk_link_to 'View download history', support_interface_view_download_history_path(data_export_id: params[:data_export_id]) %></p>
```
duncanjbrown, line 21: Is it download history or just "History"? (We don't record what's actually been downloaded)

config/routes.rb
```diff
@@ -887,6 +887,10 @@
       get '/validation-errors/summary' => 'validation_errors#summary', as: :validation_error_summary
 
       get '/data-export/documentation/:export_type_id' => 'data_exports#data_set_documentation', as: :data_set_documentation
+      get '/data-exports/history' => 'data_exports#index', as: :data_exports_history
+      get '/data-directory' => 'data_exports#directory', as: :data_directory
+      get '/data-directoy/export/:data_export_id' => 'data_exports#view_export_information', as: :view_export_information
```
duncanjbrown, line 892: Typo here and below: `directoy`

config/routes.rb
```diff
@@ -887,6 +887,10 @@
...
+      get '/data-directoy/export/:data_export_id/history/:data_export_id' => 'data_exports#view_download_history', as: :view_download_history
```
duncanjbrown, line 893: This URL could lose the second `data_export_id`

duncanjbrown (commented):
Looking good so far!

app/views/support_interface/data_exports/view_export_information.html.erb
JR-G: Good point - I suppose it's more so a history of when an export has been generated. The more detailed log is within the individual export page.

Originally I did have it as just "history" but i didn't know if this was explicit enough as to what type of history it was referring to. Perhaps overthinking, so happy to revert it back!

app/controllers/support_interface/data_exports_controller.rb
```diff
@@ -8,6 +8,18 @@ def index
...
+    def directory
+      @export_types = DataExport::EXPORT_TYPES
+    end
+
+    def view_export_information
+      @data_export = DataExport::EXPORT_TYPES[params[:data_export_type].to_sym]
+    end
+
+    def view_history
+      @data_exports = DataExport.includes(:initiator).where(name: params[:data_export_type].humanize).order(id: :desc).page(params[:page] || 1).per(30)
```
davidgisbey, line 20: It's not a massive deal, but the conversion that we do going back and forward using humanize or dasherize rather than just having a type attr feels a bit wrong.

It might be worth adding a column and backfilling wdyt?

davidgisbey: This would be done in a separate or though

davidgisbey:
I'll let Ducan have sign off since i know he wants to take another look, but this is looking really good James

app/controllers/support_interface/data_exports_controller.rb
duncanjbrown: I agree with @davidgisbey — & this would be useful in other places too! (eg the TAD API where we currently look up relevant exports by human-readable name)

paulrobertlloyd:
@JR-G Made a few changes:

* Removed the repeated GDPR guidance from the docs section, given it now appears in context alongside data exports. This is now shown in a grey box to the right of the directory listing.
* Relatedly, moved the guidance about adding new support users to that form, using the warning pattern. Note the TODO comment I’ve added here, probably for a separate PR but we should use a component for this warning (nudge @duncanjbrown sure would be nice if we could just use the respective component from `govuk-components` gem for this…)
* Fixed `DataSetAttributeComponent`; was outputting the wrong markup, empty paragraphs that were creating needed space, and added unnecessary inline styles (think this was introduced some time ago?)
* Following the pattern used elsewhere for sections within support, where pages have a common navigation component, showing ‘Details’ and ‘History’ tabs.

Still a bit of clean up needed around all this – wouldn’t mind taking another look once other issues addressed and deprecated views have been removed.

app/views/support_interface/data_exports/directory.html.erb
```diff
@@ -0,0 +1,27 @@
+<%= render 'support_interface/performance/performance_navigation', title: 'Data directory' %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <% @export_types.each do |id, export| %>
+      <%# TODO: If first section, don’t add `app-section--with-top-border` class %>
```
paulrobertlloyd, line 6: @JR-G Might you be able to help me solve this one… basically for the first section, don't want to show a top border.

app/views/support_interface/support_users/new.html.erb
```diff
@@ -8,6 +8,15 @@
...
 
+      <%# TODO: Use a component for warning text %>
+      <div class="govuk-warning-text">
+        <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
+        <strong class="govuk-warning-text__text">
+          <span class="govuk-warning-text__assistive">Warning</span>
+          Only give access to the support console to approved users.
+        </strong>
+      </div>
+
```
paulrobertlloyd, line 19: Sure would be nice if I could use the component from `govuk-components`… cough @duncanjbrown cough.

app/controllers/support_interface/data_exports_controller.rb
JR-G: @duncanjbrown & @davidgisbey I attempted the above - I don't think it's quite there but also hopefully not too far off from what you imaged? It'd be good to catch up with either of you today though and confirm this

app/views/support_interface/support_users/new.html.erb
duncanjbrown: Absolutely no objection to you adding it in this PR!

paulrobertlloyd: Addressed in #4347

app/views/support_interface/data_exports/directory.html.erb
```diff
@@ -0,0 +1,28 @@
...
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <% @export_types.each do |id, export| %>
+
+      <% if @export_types.first.second[:name] != export[:name] %>
+        <section class="app-section app-section--with-top-border">
+      <% end %>
+        <h3 class="govuk-heading-m govuk-!-margin-bottom-2"><%= govuk_link_to export.fetch(:name), support_interface_view_export_information_path(data_export_type: id) %></h3>
+        <p class="govuk-body govuk-!-margin-bottom-2"><%= export.fetch(:description) %></p>
+      </section>
```
paulrobertlloyd, line 12: @JR-G Each item should still be wrapped in a section, but the first section shouldn’t have the `app-section--with-top-border` class.

app/controllers/support_interface/data_exports_controller.rb
```diff
@@ -30,7 +30,7 @@ def create
 
     def show
       @data_export = DataExport.find(params[:id])
-      @data_export_type = @data_export.name.parameterize.underscore
+      @data_export.update!(export_type: @data_export.export_type_value)
```
davidgisbey, line 33: I don't think this should be here. It's just setting a value that is used is the view right?

JR-G: Yeah this doesn't feel like the right place but it would fall over without it here. A Data Export now requires an `export_type` attribute to be populated each time it is generated

davidgisbey: Line 25 of the export should be 
```
  data_export = DataExport.create!(name: export_type.fetch(:name), initiator: current_support_user, export_type: export_type.fetch(:type) though i think
```

duncanjbrown:
Very happy to see `export_type` added. Shall we get that in as a separate PR first? This one feels quite big now. HTH with rollback/rebase if required

JR-G:
Merge after [4364](https://github.com/DFE-Digital/apply-for-teacher-training/pull/4364)

davidgisbey (approved):
Tested locally and it's all working nicely 👍

PR #4323, opened 2021-03-19
Test application generator improvements

benswannack:
## Context
The test-data/generate endpoint times out very easily. This is because it is synchronous and has to wait until all applications are generated in order to return a list of IDs

Make the endpoint asynchronous

Touch test applications after they have been generated so that API users still have a way of getting the applications that have just been generated

Audit the touch for transparency in support about how applications are generated

## Guidance to review
Do all the changes make sense? 

There is some duplication of error detection, but this is so that providers get an instant error if they supply bad params. Maybe worth a refactor to extract the errors out of the generator?

## Link to Trello card
https://trello.com/c/PLTc5qtd/3485-investigate-issues-when-generating-20-test-applications-over-the-api

## Things to check

### DIFF
--- a/app/lib/test_applications.rb
+++ b/app/lib/test_applications.rb
@@ -18,23 +18,6 @@ class TestApplications
 
   attr_reader :time
 
-  def generate_for_provider(provider:, courses_per_application:, count:, for_ratified_courses: false)
-    courses_to_apply_to = if for_ratified_courses
-                            GetCoursesRatifiedByProvider.call(provider: provider)
-                          else
-                            Course.current_cycle.includes(:course_options)
-                              .joins(:course_options).distinct.open_on_apply
-                              .where(provider: provider)
-                          end
-    1.upto(count).flat_map do
-      create_application(
-        recruitment_cycle_year: RecruitmentCycle.current_year,
-        states: [:awaiting_provider_decision] * courses_per_application,
-        courses_to_apply_to: courses_to_apply_to,
-      )
-    end
-  end
-
   def create_application(
     recruitment_cycle_year:,
     states:,
@@ -58,6 +41,8 @@ class TestApplications
       candidate: candidate,
     )
   end
+
+private
 
   def courses_to_apply_to(
     states:,
--- a/app/lib/test_applications.rb
+++ b/app/lib/test_applications.rb
@@ -1,6 +1,20 @@
 class TestApplications
   class NotEnoughCoursesError < RuntimeError; end
-  class ZeroCoursesPerApplicationError < RuntimeError; end
+  class ZeroCoursesPerApplicationError < RuntimeError
+    def message
+      'You cannot have zero courses per application'
+    end
+  end
+  class CourseAndStateNumbersDoNotMatchError < RuntimeError
+    def message
+      'The number of states and courses must be equal'
+    end
+  end
+  class OnlyOneCourseWhenApplyingAgainError < RuntimeError
+    def message
+      'You can only apply to one course when applying again'
+    end
+  end
 
   attr_reader :time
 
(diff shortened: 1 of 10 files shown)


config/vendor-api-experimental.yml
```diff
@@ -32,7 +32,7 @@ paths:
           type: string
       responses:
         '201':
-          "$ref": "#/components/responses/TestDataGenerated"
+          "$ref": "#/components/responses/OK"
```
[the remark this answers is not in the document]
cwrw: We also might need a different message if we go with something different than the current`OK`:
```
{
  "data": {
    "message": "OK"
  }
}
```

app/lib/test_applications.rb
```diff
@@ -179,6 +162,8 @@ def create_application(recruitment_cycle_year:, states:, courses_to_apply_to:, a
         FactoryBot.create(:ucas_match, application_form: @application_form) if rand < 0.5
       end
 
+      application_choices.each { |a| a.update!(updated_at: Time.zone.now, audit_comment: 'This application was automatically generated') }
```
cwrw, line 165: Just to confirm the audit functionality doesn't support bulk operations with an `audit_comment`?

cwrw (commented):
looks good, just some minor comments. Is this worth a release note, or since its experimental do we avoid that usually?

Regarding the duplication of validations I think we can attempt a refactor to avoid the duplication of logic by either extracting to a shared validator, however that doesn't stop it going through the logic twice, or restructure of the code. What are your thoughts

app/lib/test_applications.rb
benswannack: I'll double check this. The reason it's like this is because application choices is an array rather than a relation. Maybe I can do the update via `@application_form` though

config/vendor-api-experimental.yml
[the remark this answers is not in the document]
benswannack: I think we can go with anything of the form:
```{
  "data": {
    "message": "Message here indicating success"
  }
}```
and meet the definition of the OkResponse.

Maybe the message could be "Request submitted. Applications will appear once they have been generated" or similar? Suggestions welcome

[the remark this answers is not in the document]
Comment (benswannack) [reply config/vendor-api-experimental.yml:35]: Also 201 is probably the wrong status code now, since nothing has been created at the time of response

**Comment by benswannack** [comment]
> looks good, just some minor comments. Is this worth a release note, or since its experimental do we avoid that usually?
> 
> Regarding the duplication of validations I think we can attempt a refactor to avoid the duplication of logic by either extracting to a shared validator, however that doesn't stop it going through the logic twice, or restructure of the code. What are your thoughts

Looking into restructuring the code for this now

**Review by duncanjbrown — COMMENTED** [review]
Oops, forgot to actually submit the below comments a couple of days ago!

File: spec/lib/test_applications_spec.rb [hunk 3]
```diff
@@ -150,26 +150,13 @@
     end
   end
 
-  describe 'generate_for_provider' do
-    it 'can generate applications only for courses the provider ratifies' do
```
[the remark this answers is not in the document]
benswannack: Reinstated

app/lib/test_applications.rb
```diff
@@ -1,56 +1,108 @@
...
+      courses: courses,
+      apply_again: apply_again,
+      carry_over: carry_over,
+      candidate: candidate,
+    )
+  end
+
+private
+
+  def courses_to_apply_to(
```
benswannack, line 47: I did this refactor and then realised it didn't help with my error handling. It seems ok to leave it in though, as a first step towards breaking the huge method down?

app/services/vendor_api/generate_test_applications_for_provider.rb
```diff
@@ -0,0 +1,27 @@
+module VendorAPI
+  class GenerateTestApplicationsForProvider
+    def call(provider:, courses_per_application:, count:, for_ratified_courses: false)
+      raise ParameterInvalid, 'Parameter is invalid (cannot be zero): courses_per_application' if courses_per_application.zero?
```
benswannack, line 4: After fiddling around refactoring, I think these errors are distinct from the ones inside the TestApplication generator (even if the checks are similar).

app/lib/test_applications.rb
```diff
@@ -1,56 +1,108 @@
...
+
   def create_application(
     recruitment_cycle_year:,
     states:,
     courses_to_apply_to:,
     apply_again: false,
     carry_over: false,
     course_full: false,
-    candidate: nil,
-    decline_by_default_at: 3.days.from_now
```
duncanjbrown, line 32: I wonder why we had this in the first place

app/services/vendor_api/generate_test_applications_for_provider.rb
duncanjbrown: I think I agree — this is an interface concern and that's a business logic concern

app/views/api_docs/vendor_api_docs/pages/release_notes.md
```diff
@@ -1,3 +1,12 @@
+## 22nd March
+
+The following experimental/sandbox endpoint has been updated:
+
+- `/test-data/generate` is now asynchronous. A POST request to this endpoint will queue a job to generate the specified test applications.
+The list of created application IDs will no longer be returned.
+Any applications will become available as soon as they have been generated.
```
duncanjbrown, line 7:
```suggestion
The new applications will become available as soon as they have been generated.
```

app/views/api_docs/vendor_api_docs/pages/release_notes.md
```diff
@@ -1,3 +1,12 @@
...
+Applications generated in this way will now have their `updated_at` set to the current time, so they can be retrieved using the `GET /applications` endpoint with the `since` parameter.
```
duncanjbrown, line 8: Maybe bullet these? Markdown collapses them all onto one line, which is less readable

spec/services/vendor_api/generate_test_applications_for_provider_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+
+    it 'can generate applications to courses the provider ratifies' do
+      provider = create(:provider)
+      create(:course_option)
+      create(:course_option, course: create(:course, provider: provider))
+      # rubocop:disable FactoryBot/CreateList
+      3.times do
+        create(:course_option, course: create(:course, accredited_provider: provider))
+      end
+      # rubocop:enable FactoryBot/CreateList
```
duncanjbrown, line 33: Why not use `create_list`?

app/lib/test_applications.rb
duncanjbrown: :+1: This is definitely better!

duncanjbrown (approved):
Looks good!

app/lib/test_applications.rb
cwrw: Agree 👍

cwrw (approved):
Looks good, and happy for this to go in past the other comments are addressed.

spec/services/vendor_api/generate_test_applications_for_provider_spec.rb
benswannack: With `create_list`, it will create three course options with the same course associated. It doesn't look there's a way using `create_list` to set a different course for each of the options (at least not one that is clearer than this)

PR #4326, opened 2021-03-19
Revert application choices updated_at for stale applications

davidgisbey:
## Context

Quite a few old applications were touched as part of a data cleanup. When the second nationality attribute was updated, it touched all the associated application choices which has triggered them to be passed through the API.

See https://www.apply-for-teacher-training.service.gov.uk/support/applications/2393/audit for an example.

This PR reverts the application choices updated_at to the previous audits created_at time. It will only do this if the latest audit on the application form is the one that updated second_nationality from '' to nil.

Adds a service that gets all the audits that updated `second_nationality` from `''` to `nil` between 12:30 - 1pm on the 17/03/2020 and reverts the updated_at time to the previous audit on the application choices created_at.

It will only do this if the latest audit on the application choice is the one that updated second_nationality so live applications will be unaffected

## Guidance to review

This isn't a perfect solution as it uses the created_at for the latest audit on the application choice and a subsequent update on the application form could have been made for an attribute that triggers a touch on its application choices.

I think this probably falls under 'close enough' though, as a complete solution that also collects audits for those attrs being updated would be quite a bit more work. Would you agree?
 
## Link to Trello card

https://trello.com/c/NYBG2nb9/3203-some-nationalities-being-rendered-incorrectly-on-production

## Things to check

### DIFF
--- a/app/services/data_migrations/revert_application_choice_updated_at_timestamps.rb
+++ b/app/services/data_migrations/revert_application_choice_updated_at_timestamps.rb
@@ -27,9 +27,9 @@ module DataMigrations
         )
 
         application_form.application_choices.each do |application_choice|
-          latest_application_choices_audits_created_at = application_choice.audits.last.created_at
+          latest_application_choices_audits_created_at = application_choice.audits.order(:created_at).last.created_at
 
-          if application_form.audits.last == audit && latest_application_choice_audits_created_at < audit.created_at
+          if application_form.audits.order(:created_at).last == audit && latest_application_choice_audits_created_at < audit.created_at
             application_choice.update_column('updated_at', latest_application_choices_audits_created_at)
           end
         end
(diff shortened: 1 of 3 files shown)


app/services/data_migrations/revert_application_choice_updated_at_timestamps.rb
```diff
@@ -0,0 +1,39 @@
...
+          "action = 'update' AND
+          audited_changes#>>'{second_nationality, 0}' = '' AND
+          audited_changes#>>'{second_nationality, 1}' is null
+          AND audits.created_at between ? AND ?",
+          Time.zone.local(2021, 3, 17, 12, 30),
+          Time.zone.local(2021, 3, 17, 13),
+        )
+
+        application_form.application_choices.each do |application_choice|
+          latest_application_choices_audits_created_at = application_choice.audits.last.created_at
```
duncanjbrown, line 30: Does last definitely return the last chronological audit?

duncanjbrown (commented):
This looks good, I think. One question inline

duncanjbrown (approved):

PR #4327, opened 2021-03-19
Remove Rails/ApplicationRecord Cop

despo:
- We are now using ApplicationRecord across all models. 

The models updated as part of this PR were implemented after this Cop was ignored, and thus look like they unintentionally inherited ActiveRecord::Base

## Link to Trello card

https://trello.com/c/XANNLBur/432-should-we-continue-using-rubocop-govuk

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -104,10 +104,6 @@ Lint/AmbiguousBlockAssociation:
 Rails/ApplicationController:
   Enabled: false
 
-# We do not use ApplicationRecord
-Rails/ApplicationRecord:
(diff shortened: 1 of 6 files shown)


duncanjbrown (approved):
Tidy — thank you! I think we do sth similar with ApplicationController, to the point of including things in individual controllers that are also included in ApplicationController (which they don't inherit from!)

despo:
@duncanjbrown I'll take a look. I'm trying to verify that we do need the current setup before attempting the spike for moving away from rubocop-govuk

PR #4333, opened 2021-03-22
Export documentation 2

raamSoftwire:
## Context

This is part of a wider effort to improve the visibility and documentation of the exports and is a continuation of https://github.com/DFE-Digital/apply-for-teacher-training/pull/4295.

There are still a number of exports to document, so I have raised this PR for 8 exports to avoid creating one really long PR at the end of this work.

For each doc added

This PR also adds a new "Qualifications" tab in the Support UI documentation page. 

image

We can then link to this page from the export documentation

image

## Guidance to review

- commit by commit

## Link to Trello card

## Things to check


app/exports/qualifications_export.yml
```diff
@@ -0,0 +1,118 @@
...
+  - recruitment_cycle_year
+  - choice_status
+  - rejection_reason
+  - course_code
+  - provider_code
+
+custom_columns:
+  gcse_maths_grade:
+    type: string
+    description: GCSE maths grade
```
raamSoftwire, line 14: We _could_ list the enum of possible grades here. But they would be repeated a lot and would make the doc harder to read and less useful IMHO. What do you think? Is there a link we can provide to a list? Similar to how we describe `sex` by providing a link to HESA.

app/exports/qualifications_export.yml
```diff
@@ -0,0 +1,118 @@
...
+
+  a_level_5_subject:
+    type: string
+    description: A-level 5 subject
+
+  a_level_5_grade:
+    type: string
+    description: A-level 5 grade
+
+  degree_1_type:
```
raamSoftwire, line 100: Similarly, the list of possible degree types is really long and I'm not sure if this is the best place to document them all.

app/exports/qualifications_export.yml
duncanjbrown: I suppose we could link to the API's reference data endpoint for GCSE grades: `/api/v1/reference-data/gcse-grades`? We'd need to `skip_before_action :require_valid_api_token!` in the `VendorAPI::ReferenceDataController` but no reason not to do that.

duncanjbrown: Or I guess we could have a page which listed each set in `config/initializers/qualifications`?

raamSoftwire: I think I'd prefer to make a separate page for these details. It would be laid out better and more user friendly, and from what I understand the API only includes data on GCSE and Alevels. If we have a separate page, we could also include the lists of degrees data and other qualifications

raamSoftwire: Prototype idea

image

app/components/support_interface/qualification_documentation_component.html.erb
```diff
@@ -0,0 +1,14 @@
+<details class="govuk-details" data-module="govuk-details">
```
raamSoftwire, line 1: This is just the `govuk-details` component and could be refactored once we have govuk-components

app/exports/qualifications_export.yml
```diff
@@ -11,108 +11,108 @@ common_columns:
 custom_columns:
   gcse_maths_grade:
     type: string
-    description: GCSE maths grade
+    description: GCSE maths grade. [See list of GCSE grades](/support/docs/qualifications/#gcses)
```
raamSoftwire, line 14: This works locally, and the internet seems to think that this will work across the different environments. But I wanted to call it out for a quick sanity check.

app/exports/qualifications_export.yml
davidgisbey: I really like this 👌

app/views/support_interface/docs/qualifications.html.erb
```diff
@@ -0,0 +1,29 @@
...
+    <h2 class="govuk-heading-m govuk-!-margin-bottom-2" id="a-levels">A levels</h2>
+    <%= render SupportInterface::QualificationDocumentationComponent.new(list: A_AND_AS_LEVEL_SUBJECTS, label: 'A and AS level subjects') %>
+    <%= render SupportInterface::QualificationDocumentationComponent.new(list: A_LEVEL_GRADES, label: 'A level grades') %>
+    <%= render SupportInterface::QualificationDocumentationComponent.new(list: AS_LEVEL_GRADES, label: 'AS level grades') %>
+  </div>
+</div><div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h2 class="govuk-heading-m govuk-!-margin-bottom-2" id="degrees">Degrees</h2>
+    <%= render SupportInterface::QualificationDocumentationComponent.new(list: HESA_DEGREE_TYPES.map(&:third), label: 'HESA degree types') %>
+    <%= render SupportInterface::QualificationDocumentationComponent.new(list: HESA_DEGREE_GRADES.map(&:second), label: 'HESA degree grades') %>
```
davidgisbey, line 21: I'm not sure what the use cases for this are, but I'm wondering if it's worth having the associated HESA codes here too?

`
SupportInterface::QualificationDocumentationComponent.new(list: HESA_DEGREE_GRADES.map { |hesa_grade| "#{hesa_grade.second}  - #{hesa_grade.first"), label: 'HESA degree grades') 
`
There's also a definite chance that there's no need for this though.

davidgisbey:
This is looking fantastic Raam. Thanks for all your work on it.

app/views/support_interface/docs/qualifications.html.erb
raamSoftwire: Sure thing. I've just added that in

image

app/exports/qualifications_export.yml
duncanjbrown: +10!

duncanjbrown:
The quals page is super. Maybe add a bit of text at the top explaining that this is what goes into the autocompletes/validation rules? And also consider a `max-height` plus `overflow-y` on the opened lists as some are quite long? (@paulrobertlloyd probably has views!)

app/components/support_interface/qualification_documentation_component.html.erb
duncanjbrown: We've got it now!

duncanjbrown: Possibly one to come back to in the next PR — we should apply consistently across the app?

paulrobertlloyd:
@raamSoftwire I think I have an idea for the quals page… will just have a quick tinker…

app/views/support_interface/docs/qualifications.html.erb
```diff
@@ -1,29 +1,34 @@
...
+
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
-    <h2 class="govuk-heading-m govuk-!-margin-bottom-2" id="gcses">GCSEs</h2>
-    <%= render SupportInterface::QualificationDocumentationComponent.new(list: GCSE_SUBJECTS, label: 'GCSE subjects') %>
-    <%= render SupportInterface::QualificationDocumentationComponent.new(list: SINGLE_GCSE_GRADES, label: 'Single GCSE grades') %>
-    <%= render SupportInterface::QualificationDocumentationComponent.new(list: DOUBLE_GCSE_GRADES, label: 'Double GCSE grades') %>
+  <div class="govuk-grid-column-one-quarter">
+    <%# TODO: Update content describing purpose of this page %>
+    <p class="govuk-body govuk-!-margin-top-4">I’m not sure how you would describe the purpose of this page, but you would describe it here.</p>
```
paulrobertlloyd, line 6: Needs words! @raamSoftwire

app/views/support_interface/docs/qualifications.html.erb
davidgisbey: looks great 👍

raamSoftwire:
I've now made all of the suggested changes and added a screenshot to the description of this PR. Hopefully, it should be good to go 🤞🏽

app/views/support_interface/docs/qualifications.html.erb
```diff
@@ -0,0 +1,56 @@
+<%= render 'docs_navigation', title: 'Qualifications' %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-one-quarter">
+    <p class="govuk-body govuk-!-margin-top-4">
+      These lists are a reference for the values used throughout the service.
+    </p>
+    <p class="govuk-body govuk-!-margin-top-4">
+      They are used to populate the autocomplete fields and the form validation.
```
paulrobertlloyd, line 9:
```suggestion
      They are used to populate autocomplete fields and for form validation.
```

paulrobertlloyd (approved):
One tiny copy change, but otherwise looks good to me.

app/views/support_interface/docs/qualifications.html.erb
```diff
@@ -0,0 +1,56 @@
...
+    </p>
+  </div>
+
+  <div class="govuk-grid-column-three-quarters">
+    <%= render GovukComponent::Accordion.new do |component| %>
```
duncanjbrown, line 14: Great!

duncanjbrown (approved):
Beautiful 👌

app/views/support_interface/docs/qualifications.html.erb
raamSoftwire: Done

PR #4334, opened 2021-03-22
Add additional details to the add course page on the support UI

davidgisbey:
## Context

This PR tackles two defects 

1. Courses that have no vacancies are not returned by design. However, this is really confusing for support staff as there is no indication that the course is full
2. Some courses share a code and name (Physics etc.). This can lead to the page occasionally looking like this.

image

1. Add a list of full courses for the provided code
2. Expose the provider and study mode to the course options struct 

|Before|After|
|------------|------------|
|image|image|
## Guidance to review

Does this approach seem reasonable?

## Link to Trello card

https://trello.com/c/g6ygCMtV/3055-support-interface-when-attempting-to-switch-to-a-course-that-has-no-vacancies-its-not-obvious-why-you-cannot-do-so

https://trello.com/c/FIi30t2q/3212-show-additional-detail-on-the-add-course-page-on-the-support-ui

## Things to check

### DIFF
--- a/app/forms/support_interface/application_forms/pick_course_form.rb
+++ b/app/forms/support_interface/application_forms/pick_course_form.rb
@@ -10,7 +10,7 @@ module SupportInterface
       validates :course_code, presence: true
       validate :course_is_open_on_apply, on: :save
 
-      RadioOption = Struct.new(:course_option_id, :course_name, :course_code, :site_name)
+      RadioOption = Struct.new(:course_option_id, :provider_name, :provider_code, :course_name, :course_code, :site_name, :study_mode)
 
       def course_options
         return @course_options if @course_options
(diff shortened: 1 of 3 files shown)


app/forms/support_interface/application_forms/pick_course_form.rb
```diff
@@ -10,7 +10,7 @@ class PickCourseForm
       validates :course_code, presence: true
       validate :course_is_open_on_apply, on: :save
 
-      RadioOption = Struct.new(:course_option_id, :course_name, :course_code, :site_name)
+      RadioOption = Struct.new(:course_option_id, :provider_name, :provider_code, :course_name, :course_code, :site_name, :study_mode)
```
stevehook, line 13: regular args get a bit error prone when you have this many. Not a big problem here because this struct is only used in it's parent class. Alternative (had to look this up) is to use `keyword_init: true` to switch the struct initializer over to keyword args.

stevehook (approved):
Tested the review app and it works really nicely. This is a big improvement. :+1:

app/forms/support_interface/application_forms/pick_course_form.rb
davidgisbey: Yea good shout. Have sorted it

PR #4335, opened 2021-03-22
Refactor API User Authentication

cwrw:
## Context

We'd like to introduce a new API to be used by the Register team. To allow authenticated users access to the API, reuse and refactor the authentication logic from the TAD data API.

- Move the `DataApiUser` logic to a shared `APIUser` class and introduce the new register user
- Move the verification logic from the `DataAPI` controller into a shared concern
- Add the new register skeleton controller which reuses the new authentication concern

## Guidance to review

This is the first PR to introduce the register API, the next step will be to introduce the different endpoints/fields.
To allow the authentication to be shared, I introduced an `api_type` attribute to the user data to allow the authenticator access to the User depending on the API namespace, let me know if there anyone has other/better suggestions for naming and getting access to those users.

## Link to Trello card

## Things to check

### DIFF
--- a/app/controllers/concerns/api_user_authentication.rb
+++ b/app/controllers/concerns/api_user_authentication.rb
@@ -1,4 +1,5 @@
 module APIUserAuthentication
+  include ActionController::HttpAuthentication::Token::ControllerMethods
   extend ActiveSupport::Concern
 
   included do
--- a/app/controllers/concerns/api_user_authentication.rb
+++ b/app/controllers/concerns/api_user_authentication.rb
@@ -1,0 +1,30 @@
+module APIUserAuthentication
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :verify_token!
+  end
+
+  def verify_token!
+    unless authorized?
+      render_error(
+        name: 'Unauthorized',
+        message: 'Please provide a valid API token',
+        status: :unauthorized,
+      )
+    end
+  end
+
+  def authorized?
+    authenticate_with_http_token do |token|
+      @authenticating_token = AuthenticationToken.find_by_hashed_token(user_type: 'APIUser', raw_token: token)
+      @authenticating_token.user_id == APIUser.find_by(api_type: self.class.name.deconstantize).id
+    end
+  end
+
+  def render_error(name:, message:, status:)
+    response = { errors: [{ error: name, message: message }] }
+
+    render json: response, status: status
+  end
+end
--- a/app/controllers/data_api/tad_data_exports_controller.rb
+++ b/app/controllers/data_api/tad_data_exports_controller.rb
@@ -24,8 +24,8 @@ module DataAPI
 
     def authorized?
       authenticate_with_http_token do |token|
-        @authenticating_token = AuthenticationToken.find_by_hashed_token(user_type: 'DataAPIUser', raw_token: token)
-        @authenticating_token.user_id == DataAPIUser.tad_user.id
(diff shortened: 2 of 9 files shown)


app/controllers/data_api/tad_data_exports_controller.rb
```diff
@@ -1,38 +1,12 @@
 module DataAPI
   class TADDataExportsController < ActionController::API
     include ActionController::HttpAuthentication::Token::ControllerMethods
```
despo, line 3: Can we move this to the reusable module, as it relies on it?

despo (approved):
:shipit:

app/controllers/concerns/api_user_authentication.rb
```diff
@@ -0,0 +1,30 @@
+module APIUserAuthentication
```
duncanjbrown, line 1: It's slightly confusing that this is one of two authentication methods for APIs. Could this maybe be `ServiceAPIUserAuthentication` or something? Is a "service API" a thing?

app/models/api_user.rb
```diff
@@ -4,14 +4,15 @@
 #
 # To obtain an API token for a user, run this in the console:
 #
-#   DataAPIUser.tad_user.create_magic_link_token!
-class DataAPIUser < ActiveHash::Base
+#   APIUser.tad_user.create_magic_link_token!
+class APIUser < ActiveHash::Base
```
duncanjbrown, line 8: I mean.... `ServiceAPIUser`!?

app/models/api_user.rb
```diff
@@ -4,14 +4,15 @@
...
   include ActiveHash::Associations
   include AuthenticatedUsingMagicLinks
 
   self.data = [
-    { id: 1, name: 'User for testing, not used in production' },
-    { id: 2, name: 'DfE TAD' },
+    { id: 1, name: 'User for testing, not used in production', api_type: 'TestAPI' },
```
duncanjbrown, line 13: - `api_type` is odd because it implies there's more than one API under the type. Maybe just `api`? Or `authorized_api` or equivalent?
- do we need a testing user for each API? Maybe we should give them high IDs starting at 100 or sth?

spec/requests/register_api/applications_controller_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  it 'verifies the API token' do
+    get '/register-api/applications', headers: {}
+
+    expect(response).to have_http_status(:unauthorized)
+  end
+
+  it 'does not allow access to the API from other data users' do
+    api_token = APIUser.test_data_user.create_magic_link_token!
+
+    headers = { 'Authorization' => "Bearer #{api_token}" }
```
duncanjbrown, line 13: :+1: We can eventually fit this out with an `APISpecHelper` along the lines of https://github.com/DFE-Digital/apply-for-teacher-training/commit/8bbcc11ca74f4f022b2025667aa64453c8fc9656

app/models/api_user.rb
cwrw: > api_type is odd because it implies there's more than one API under the type. Maybe just api? Or authorized_api or equivalent?

I renamed to `authorized_api` for better clarity 👍 

> do we need a testing user for each API? Maybe we should give them high IDs starting at 100 or sth?

I don't think we do right now, the test user in our specs now is just to ensure that access is not given across different authorised APIs. Since there is only one now I think it's ok to leave with the current id. Let me know your thoughts

duncanjbrown (approved):
LGTM!

PR #4344, opened 2021-03-22
Run rspec in parallel by default

AgaDufrat:
## Context

Run rspec in parallel by default as agreed in a dev retro to speed up development.

## Guidance to review

- set up https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/docs/parallel_tests.md
- run `bundle exec rake`

## Things to check

### DIFF
--- a/lib/tasks/testing.rake
+++ b/lib/tasks/testing.rake
@@ -54,5 +54,10 @@ task :compile_assets do
   sh 'RAILS_ENV=test rails webpacker:compile'
 end
 
+desc 'Run rspec in parallel without performance'
+task :parallel_rspec_without_performance do
+  sh "bundle exec parallel_rspec --exclude-pattern=#{performance_test_pattern} spec"
+end
+
 desc 'Run all the tests'
-task run_tests: %i[compile_assets linting spec_without_performance brakeman jest]
+task run_tests: %i[compile_assets linting parallel_rspec_without_performance brakeman jest]


lib/tasks/testing.rake
```diff
@@ -54,5 +54,10 @@ task :compile_assets do
...
 end
 
+desc 'Run rspec in parallel without performance'
+task :parallel_rspec_without_performance do
+  sh "bundle exec parallel_rspec --exclude-pattern=#{performance_test_pattern} spec"
+end
+
 desc 'Run all the tests'
-task run_tests: %i[compile_assets linting spec_without_performance brakeman jest]
+task run_tests: %i[compile_assets linting parallel_rspec_without_performance brakeman jest]
```
despo, line 63: Is it possible to get all tasks to run in parallel? As alternatively it will be difficult to drop the total run time to under 4-5 minutes.

duncanjbrown: Rubocop runs in parallel. I don't think erblint supports it, but potentially we could have a rake `multitask` for erblint, brakeman and just? https://www.rubydoc.info/gems/rake/12.0.0/Rake/MultiTask

despo: rspec_parallel supports ruinnning any task in parallel https://github.com/grosser/parallel_tests#run-an-arbitrary-task-in-parallel. However, rake multitask seems to be an easy way around it that saves us time as well. Whatever is easiest would work I think.

duncanjbrown (approved):
This works and is shippable! We can see about parallelising other bits in a subsequent PR

PR #4345, opened 2021-03-22
Update the candidate offer page

davidgisbey:
## Context

As part of the work updating the dashboard, the offer page was also reviewed. This PR covers changing the guidance on the page and updating the design so it is in line with the prototype.

- Update the offer page

|Before|After|
|------------|------------|
|image|image|

## Guidance to review

I haven't put this behind a feature flag because it really doesn't seem like any of the changes need to be? Would you agree?

I've added further requested changes in a second commit

## Link to Trello card

https://trello.com/c/dSny6dy0/3194-content-changes-to-offer-page

## Things to check

### DIFF
--- a/app/components/candidate_interface/offer_review_component.rb
+++ b/app/components/candidate_interface/offer_review_component.rb
@@ -27,14 +27,14 @@ module CandidateInterface
     def course_row
       {
         key: 'Course',
-        value: @course_choice.offered_course.name_and_code,
+        value: course_row_value,
       }
     end
 
     def location_row
       {
         key: 'Location',
-        value: @course_choice.offered_option.site.name,
+        value: @course_choice.offered_option.site.name_and_address,
       }
     end
 
@@ -43,6 +43,21 @@ module CandidateInterface
         key: 'Conditions',
         value: render(OfferConditionsReviewComponent.new(conditions: @course_choice.offer['conditions'], provider: @course_choice.offered_course.provider.name)),
       }
+    end
+
+    def course_row_value
+      if EndOfCycleTimetable.find_down?
+        tag.p(@course_choice.offered_course.name_and_code, class: 'govuk-!-margin-bottom-0') +
+          tag.p(@course_choice.offered_course.description, class: 'govuk-body')
+      else
+        govuk_link_to(
+          @course_choice.offered_course.name_and_code,
+          @course_choice.offered_course.find_url,
+          target: '_blank',
+          rel: 'noopener',
+        ) +
+          tag.p(@course_choice.offered_course.description, class: 'govuk-body')
+      end
     end
   end
 end
--- a/app/components/candidate_interface/offer_review_component.rb
+++ b/app/components/candidate_interface/offer_review_component.rb
@@ -5,12 +5,13 @@ module CandidateInterface
     end
 
     def rows
-      [
+      rows = [
         provider_row,
         course_row,
         location_row,
-        conditions_row,
       ]
+      rows << conditions_row if @course_choice.offer&.dig('conditions')&.present?
+      rows
     end
 
   private
@@ -34,7 +35,7 @@ module CandidateInterface
     def location_row
       {
         key: 'Location',
-        value: @course_choice.offered_option.site.name_and_address,
+        value: @course_choice.offered_option.site.name,
       }
     end
 
--- a/app/controllers/candidate_interface/decisions_controller.rb
+++ b/app/controllers/candidate_interface/decisions_controller.rb
@@ -8,6 +8,7 @@ module CandidateInterface
 
     def offer
       @respond_to_offer = CandidateInterface::RespondToOfferForm.new
+      @offer_count = @application_choice.self_and_siblings.offer.count
     end
 
     def respond_to_offer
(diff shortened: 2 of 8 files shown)


app/controllers/candidate_interface/decisions_controller.rb
```diff
@@ -8,6 +8,7 @@ class DecisionsController < CandidateInterfaceController
 
     def offer
       @respond_to_offer = CandidateInterface::RespondToOfferForm.new
+      @offer_count = @application_choice.application_form.application_choices.offer.count
```
despo, line 11:
```suggestion
      @offer_count = @application_choice.self_and_siblings.offer.count
```

davidgisbey: Amazing. I've not come across that before. Thanks!

malcolmbaig (approved):

app/views/candidate_interface/decisions/offer.html.erb
```diff
@@ -12,20 +12,22 @@
       <%= f.govuk_error_summary %>
 
       <h1 class="govuk-heading-xl">
+        <span class="govuk-caption-xl"><%= @application_choice.course.provider_and_name_code %></span>
```
frankieroberto, line 15: Sorry, can we drop this please? I think it was left in the prototype as a legacy, rather than something we deliberately added back in.
```suggestion
```

app/views/candidate_interface/decisions/offer.html.erb
```diff
@@ -12,20 +12,22 @@
...
-        <li>if you fail to respond, or you decline all your offers, you can still apply to other courses this year through <%= govuk_link_to 'UCAS Teacher Training', UCAS.apply_url %></li>
+        <li>if you accept this offer, you cannot also accept another offer through UCAS teacher training</li>
+        <% if @offer_count > 1 %>
+          <li>if you accept this offer, your other <%= 'offer'.pluralize(@offer_count - 1) %> will be automatically declined</li>
+        <% end %>
+        <li>if you decline all of your offers, or do not respond in time, you can still apply for courses that start this academic year</li>
       </ul>
 
-      <p class="govuk-body">If you need help, you can speak to a <%= govuk_link_to t('service_name.get_into_teaching'), t('get_into_teaching.url_get_an_advisor') %> adviser. They’re all experienced teachers who can give you advice on applying for teacher training.</p>
-
```
frankieroberto, line 28: Opps, we didn't mean to drop this.

Could we reword it though, to be a bit shorter (and avoid referring to GIT by name):

```erb
<p class="govuk-body">If you need help, you can speak to our <%= govuk_link_to 'teacher training advisers', t('get_into_teaching.url_get_an_advisor') %>.</p>
```

app/components/candidate_interface/offer_review_component.rb
```diff
@@ -27,14 +27,14 @@ def provider_row
...
-        value: @course_choice.offered_course.name_and_code,
+        value: course_row_value,
       }
     end
 
     def location_row
       {
         key: 'Location',
-        value: @course_choice.offered_option.site.name,
+        value: @course_choice.offered_option.site.name_and_address,
```
frankieroberto, line 37: We didn't deliberately change this - I think it was an existing difference between the prototype and production. Possibly it's still worth doing though?

frankieroberto (changes requested):
Thanks, looking good!

This highlighted a couple of inadvertent changes we made. I've left comments, and updated the prototype here: https://github.com/DFE-Digital/apply-for-teacher-training-prototype/pull/482

Would you mind also checking that the "Conditions" row is completely dropped from the summary if there were no conditions?  (This could be a follow-up PR).

app/components/candidate_interface/offer_review_component.rb
davidgisbey: I'll revert it for now since it was an accidental change, but happy to revert it if you'd like me too.

app/views/candidate_interface/decisions/offer.html.erb
davidgisbey: Have done

app/views/candidate_interface/decisions/offer.html.erb
davidgisbey: Have done 👍

app/components/candidate_interface/offer_review_component.rb
```diff
@@ -5,12 +5,13 @@ def initialize(course_choice:)
...
 
     def rows
-      [
+      rows = [
         provider_row,
         course_row,
         location_row,
-        conditions_row,
       ]
+      rows << conditions_row if @course_choice.offer&.dig('conditions')&.present?
```
davidgisbey, line 13: this is a bit clunky but i'm not sure if this is going to return nil or an empty array so it guards against both

config/locales/provider_interface/decisions.yml
```diff
@@ -4,6 +4,8 @@ en:
       legend: How do you want to respond to this offer?
       accept:
         label: Accept offer and conditions
+      accept_no_conditions:
+        label: Accept offer
```
frankieroberto, line 8: Nice, thanks for doing this! (even though it was technically out of scope 😉  )

frankieroberto (approved):

PR #4347, opened 2021-03-23
Replace warning component with one from govuk-components

paulrobertlloyd:
## Context

(This PR is being spun out of #4318. This PR should be merged before that one)

We have a `WarningComponent`. It has the same API as the respective component in `govuk-components`. We should make this the first component we move over to using that gem.

And with that component in place, we can move the warning about only adding approved users to the ‘Add a new support user’ form.

| Before | After |
| - | - |
| before | after |

## Things to check


Gemfile.lock
```diff
@@ -496,6 +500,13 @@ GEM
     simplecov_json_formatter (0.1.2)
     skylight (5.0.1)
       activesupport (>= 5.2.0)
+    slim (4.1.0)
+      temple (>= 0.7.6, < 0.9)
+      tilt (>= 2.0.6, < 2.1)
+    slim-rails (3.2.0)
+      actionpack (>= 3.1)
+      railties (>= 3.1)
+      slim (>= 3.0, < 5.0)
```
paulrobertlloyd, line 509: Bit annoying that `govuk-components` adds `slim` as a dependency – is this needed? @duncanjbrown

duncanjbrown: Maybe not for us, but not our choice either

Gemfile.lock
```diff
@@ -183,6 +183,10 @@ GEM
     geocoder (1.6.6)
     globalid (0.4.2)
       activesupport (>= 4.2.0)
+    govuk-components (0.1.0)
```
peteryates, line 186: 🕰️

Gemfile.lock
paulrobertlloyd: Will update this PR with the correct version of this gem when https://github.com/DFE-Digital/govuk-components/pull/123 is merged.

duncanjbrown (approved):

PR #4349, opened 2021-03-23
Add ChangeOffer service and new checks in wrapper services

mnacos:
### Context

The new `MakeOffer`/`ChangeOffer` flow introduces new restrictions around which courses can be offered. In particular, it is important the ratifying provider is preserved, at least for now.

The `MakeOffer`/`ChangeOffer` services (and their underlying `MakeAnOffer`/`ChangeAnOffer` services which will be removed eventually) are the last defence against problematic offers, as the UI is meant to suppress any options which could be problematic. When it comes to the API, however, these services are the only checks for the course options and conditions specified.

This PR introduces a `ChangeOffer` service, which wraps `ChangeAnOffer` and exposes a `save!` method, which raises exceptions if any errors are detected.

The wrapper performs additional checks. In particular:

- Checks the current ratifying provider doesn't change in the updated offer
- Validates the length of conditions, for consistency with `MakeOffer`

This PR also adds more checks to the `MakeOffer` service:

- Checks the ratifying provider doesn't change, compared to the original application choice
- Checks whether the application is already in the 'offer' state (which wouldn't raise an invalid transition)

### Guidance to review

We are adding a wrapper rather than refactoring `ChangeAnOffer` because our launch plan for the new offer wizard/flow involves preserving the existing functionality unchanged until the feature flag is activated. When this happens, Manage users will be taken through the new flow, which calls the wrapper service containing the additional rules.

The API is still using `MakeAnOffer` and `ChangeAnOffer`. The plan is to adapt the relevant controller, when we are ready, to work with the new interface and rescue the errors in order to pass them to the API consumers.

There is some duplication between the tests for `ChangeOffer` and `ChangeAnOffer`. This is intentional, as our plan is to remove `ChangeAnOffer` completely in the future, at which point we'll just delete the  `ChangeAnOffer` spec.

Do the exceptions raised make sense? There are six different types across the two services:

- CourseValidationError
- ConditionsValidationError
- RatifyingProviderChangeError
- IdenticalOfferError
- AlreadyOfferedError
- NoTransitionAllowedError

These are defined within `MakeOffer` and `ChangeOffer`, but could be extracted to an `Offer` set of validations when we remodel offers/conditions, so that they are reused. `IdenticalOfferError`, however, is specific to `ChangeOffer`, in the same way as `AlreadyOfferedError` and `NoTransitionAllowedError` are specific to `MakeOffer`.

### Link to Trello card

[Update MakeOffer/ChangeOffer services with new rules](https://trello.com/c/9I0RTTts)

### Things to check

- [x] API release notes have been updated if necessary -> _API has not been updated_

### DIFF
--- a/app/services/change_offer.rb
+++ b/app/services/change_offer.rb
@@ -25,16 +25,16 @@ class ChangeOffer
 
     unless change_an_offer.save
       if change_an_offer.errors[:base].include?('The new offer is identical to the current offer')
-        raise IdenticalOffer, 'The new offer is identical to the current offer'
+        raise IdenticalOfferError, 'The new offer is identical to the current offer'
       elsif change_an_offer.errors[:course_option].present?
         raise CourseValidationError, change_an_offer.errors[:course_option].join
       end
     end
   end
 
-  class IdenticalOffer < StandardError; end
+  class IdenticalOfferError < StandardError; end
   class CourseValidationError < StandardError; end
-  class RatifyingProviderChange < StandardError; end
+  class RatifyingProviderChangeError < StandardError; end
   class ConditionsValidationError < StandardError; end
 private
 
@@ -42,7 +42,7 @@ private
     previous_ratifying_provider = application_choice.offered_course.accredited_provider || application_choice.offered_course.provider
     new_ratifiying_provider = course_option.course.accredited_provider || course_option.course.provider
     if previous_ratifying_provider != new_ratifiying_provider
-      raise RatifyingProviderChange, 'The new offer has a different ratifying provider to the current offer'
+      raise RatifyingProviderChangeError, 'The new offer has a different ratifying provider to the current offer'
     end
   end
 
--- a/app/services/change_offer.rb
+++ b/app/services/change_offer.rb
@@ -47,7 +47,7 @@ private
   end
 
   def check_conditions!
-    if conditions&.count > MAX_CONDITIONS_COUNT
+    if conditions && conditions.count > MAX_CONDITIONS_COUNT
       raise ConditionsValidationError, 'Too many conditions specified (20 or fewer required)'
     elsif conditions.any? { |c| c.length > MAX_CONDITION_LENGTH }
       raise ConditionsValidationError, 'Condition exceeds length limit (255 characters or fewer required)'
--- a/app/services/change_offer.rb
+++ b/app/services/change_offer.rb
@@ -12,14 +12,32 @@ class ChangeOffer
   end
 
   def save!
+    if offer_changes_ratifying_provider?
+      raise RatifyingProviderChange, 'The new offer has a different ratifying provider to the current offer'
+    end
+
     change_an_offer = ChangeAnOffer.new(actor: actor,
                                         application_choice: application_choice,
                                         course_option: course_option,
                                         offer_conditions: conditions)
-    change_an_offer.save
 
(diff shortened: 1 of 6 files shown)


despo:
I'm a bit confused about this piece of work. 

I thought we discussed that currently we just need a wrapper around MakeOffer. As the offer wizard handles all required validations related to an offer, the goal is to eventually separate any validations out of the service and add a separate validator used by the VendorAPI and not to make sure that MakeOffer fails for the same reasons as MakeAnOffer.. The same applies to ChangeOffer.

mnacos:
> I thought we discussed that currently we just need a wrapper around MakeOffer. As the offer wizard handles all required validations related to an offer, the goal is to eventually separate any validations out of the service and add a separate validator used by the VendorAPI and not to make sure that MakeOffer fails for the same reasons as MakeAnOffer.. The same applies to ChangeOffer.

Fyi, we've just pushed some changes to the `MakeOffer` wrapper to introduce additional checks.

There are two main goals for this PR:

1. Provide service wrappers for both making and updating offers which throw exceptions if any validations fail. This way, the wizard controllers know something has gone wrong and can redirect users back to the start of the flow, as discussed. The controllers won't worry too much about what type of exception is thrown.
2. Move towards services that raise errors for different types of failure for `MakeOffer`/`ChangeOffer` without affecting the API immediately. We can probably ship changes to the API offer endpoint separately, i.e. after the feature flag is activated on production and after any communications with vendors.

After launch, we could have a go at re-modelling offers and conditions, create the relevant validators and modify `MakeOffer`/`ChangeOffer` to use them.

cwrw:
The addition of the two new services wrappers `MakeOffer` and `ChangeOffer` which wrap the old flow services which themselves hold some validations, was intentional in us avoiding a major refactor to the existing services to reduce the scope and complexity of the make offer flow as it stands. 

Last week we agreed to add a new validation service, which initially was going to be integrated into the new `MakeOffer` and `ChangeOffer` services. However after reconvening we agreed that all validations will live in the wizard for now, so the new validation service will be integrated into the wizard as well. The new validation service was merged today with some validations that can be extended as required  and can be used in the API later.

I think we should avoid duplicating the validation logic found in the `MakeAnOffer` and `ChangeAnOffer` services themselves in the wrappers for now, and keep the wrappers as simple as possible, and keep the tests the same as well. 

When we get to the point of refactor then we can think about extending those services as we still have many moving parts and it is preferable to avoid any duplication as well as extra logic as possible until we have fully shipped the current piece of work, and have a better understanding of the final picture.

My proposition for now would be to add the new `ChangeOffer` and keep as close to what we have in `MakeOffer`, with the same tests. The additional validations already live in the existing old services, so there is no risk in us not covering those in the new make offer flow. Let me know your thoughts.

mnacos:
Let's discuss because the real value of this work stems from the new test files, which combine expectations around the new exceptions behaviour with verification of what each service does. These put us in a very good position to refactor what we want while still delivering the updated flows. The original services remain unchanged, down to the lack of conditions validation in `ChangeAnOffer`.

mnacos:
Closing for now, as we have decided to introduce the least code possible, until we have clear decision on how and when MakeOffer/ChangeOffer behaviour will change for the API. We could then resurrect this or use it as a starting point for further refactoring.

Looking at all the different error modes in this PR, and assuming the wizard does its own validation of conditions, the only extra validation we need to worry about is `IdenticalOfferError`, which is a special edge case when editing/changing an offer.

PR #4352, opened 2021-03-23
Rework how PickCourseOption deals with repeated course choices 

malcolmbaig:
## Context
Currently, if the candidate repeats submission of the course selection
form (eg - by navigating back to a stale form with the back button),
they see an info message telling them they've already added the course.
If they do the same with the study mode and site forms, we want
different behaviour - in those cases, we want the existing course choice
updated with the new study mode/site selections.

The final step of any journey through the course choice flow is the
PickCourseOption service. To achieve the requirement above, update the
service so that:

- It looks for existing application choices that match the course being
  added
- If one is found, replace it with a new application choice reflecting
  the candidate-submitted values

Note that this logic doesn't get executed for course selections, as in
that controller we return early if we detect the course is already
present (so that we can display the info message).

One of the side effects of this change is that it fixes two bugs in the
current implementation that occur when navigating back:

- Going back to the study mode page
  * Form submission can create two choices with the same course.
- Going back to the site page
  * Form submission results in confusing UX, ie - a silent redirect to
    the review, and upon hitting back again, seeing an error about the
    course already being added.

## Guidance to review
- Per-commit highly recommended. There are three refactoring commits that precede the primary change, added in so as to make the final code change easier.
- Manual testing can be done locally or on the review app. Adding various course options which provide a choice of study mode and/or site would be useful. Testing that the two bug scenarios above have been resolved would also be helpful.

## Link to Trello card
https://trello.com/c/eP3BTlVI

## Things to check


malcolmbaig:
Closing to use correct branch name.

PR #4356, opened 2021-03-23
Use GovukComponent::Accordion

paulrobertlloyd:
## Context

Replace HTML markup with that generated by `GovukComponent::Accordion` component.

We only use the accordion component in one place: the list of providers shown to candidates within Apply.

## Things to check

### DIFF
--- a/app/components/candidate_interface/grouped_provider_courses_component.html.erb
+++ b/app/components/candidate_interface/grouped_provider_courses_component.html.erb
@@ -2,28 +2,19 @@
 
 <% courses_by_provider_and_region.each do |region_code, courses_by_provider| %>
   <h2 class="govuk-heading-m govuk-!-margin-bottom-2 govuk-!-margin-top-8"><%= label_for(region_code) %></h2>
-  <div class="govuk-accordion" data-module="govuk-accordion" id="accordion-default">
+  <%= render GovukComponent::Accordion.new(id: region_code) do |component| %>
     <% courses_by_provider.each_with_index do |provider, index| %>
-      <section class="govuk-accordion__section">
-        <header class="govuk-accordion__section-header">
-          <h2 class="govuk-accordion__section-heading">
-            <span class="govuk-accordion__section-button" id="accordion-default-heading-<%= index %>"><%= provider.provider_name %></span>
-          </h2>
-        </header>
-        <div class="govuk-accordion__section-content" id="accordion-default-content-<%= index %>" aria-labelledby="accordion-default-heading-<%= index %>">
-          <ul class="govuk-list govuk-list--bullet">
+      <% component.slot(:section, title: provider.provider_name) do %>
+        <ul class="govuk-list govuk-list--bullet">
+          <% provider.courses.sort_by(&:name).each do |course| %>
             <% if find_down %>
-             <% provider.courses.sort_by(&:name).each do |course| %>
-               <li><%= course.name_and_code %></li>
-             <% end %>
-           <% else %>
-             <% provider.courses.sort_by(&:name).each do |course| %>
-               <li><%= govuk_link_to course.name_and_code, course.find_url %></li>
-             <% end %>
-           <% end %>
-          </ul>
-        </div>
-      </section>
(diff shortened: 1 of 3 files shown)


app/components/support_interface/audit_trail_item_component.html.erb
```diff
@@ -1,4 +1,4 @@
-<tr class="govuk-table__row gov uk-accordion__section" data-qa="audit-trail-item">
```
paulrobertlloyd, line 1: This looks like something leftover from when/if audit trails used an accordion. Not quite sure how this got here, to be honest.

app/components/candidate_interface/grouped_provider_courses_component.html.erb
```diff
@@ -2,28 +2,19 @@
...
-           <% end %>
-          </ul>
-        </div>
-      </section>
+              <li><%= course.name_and_code %></li>
+            <% else %>
+              <li><%= govuk_link_to course.name_and_code, course.find_url %></li>
+            <% end %>
+          <% end %>
+        </ul>
```
paulrobertlloyd, line 16: Changed how we test for `find_down`; bit simpler than before, hopefully.

app/components/candidate_interface/grouped_provider_courses_component.html.erb
```diff
@@ -2,28 +2,19 @@
...
     <% courses_by_provider.each_with_index do |provider, index| %>
-      <section class="govuk-accordion__section">
-        <header class="govuk-accordion__section-header">
-          <h2 class="govuk-accordion__section-heading">
-            <span class="govuk-accordion__section-button" id="accordion-default-heading-<%= index %>"><%= provider.provider_name %></span>
-          </h2>
-        </header>
-        <div class="govuk-accordion__section-content" id="accordion-default-content-<%= index %>" aria-labelledby="accordion-default-heading-<%= index %>">
-          <ul class="govuk-list govuk-list--bullet">
+      <% component.slot(:section, title: "#{provider.provider_name}") do %>
```
paulrobertlloyd, line 7: `title: "#{provider.provider_name}"` is odd… but found that doing this was the only way I could get the test to pass. 

Here was the error:

```
Failure/Error: <%= render GovukComponent::Accordion.new(id: region_code) do |component| %>

     NoMethodError:
       undefined method `parameterize' for #<Provider:0x00007f7f3a3cf918>
```

This relates to this definition in `govuk-components`: https://github.com/DFE-Digital/govuk-components/blob/1066064af162cafb58caf1d82264de593c3adbd0/app/components/govuk_component/accordion.rb#L35

Is there a better way of fixing/addressing this?

stevehook: I reckon `title: provider.provider_name` should just work, unless `provider` is `nil` or doesn't have a `provider_name` method (which it isn't if `title: "#{provider.provider_name}"` works).

stevehook: `title: provider` would cause the error above. `parameterize` is a method on `String` (from Rails) so the `title` argument does need to be a string (or something else that implements `parameterize`).

app/components/candidate_interface/grouped_provider_courses_component.html.erb
```diff
@@ -2,28 +2,19 @@
 
 <% courses_by_provider_and_region.each do |region_code, courses_by_provider| %>
   <h2 class="govuk-heading-m govuk-!-margin-bottom-2 govuk-!-margin-top-8"><%= label_for(region_code) %></h2>
-  <div class="govuk-accordion" data-module="govuk-accordion" id="accordion-default">
+  <%= render GovukComponent::Accordion.new(id: region_code) do |component| %>
```
stevehook, line 5: I guess using `id: region_code` fixes duplicate `accordion-default` ids? :+1:

stevehook (approved):
Looks good. I think there is a redundant string interpolation that you've noted.

Review app works!

PR #4359, opened 2021-03-24
Delete spurious lat/long setting from the audit log

duncanjbrown:
## Context

We have 1.3 million audit entries that consist of the lat/long flipping back and forth:

Screenshot 2021-03-24 at 07 39 59

- Find all the entries that consist only of lat/long changes
- for each provider, ascertain whether one of these entries originally set the lat/long
- if so, delete all the lat/long-only audit logs for that provider but that one
- if not, delete all the lat/long audit logs for that provider

## Guidance to review

The SQL could definitely do with another pair of eyes! And could we dry-run this somehow?

## Link to Trello card

https://trello.com/c/6vbtx9Nm/427-audit-trial-is-full-of-flip-flops

## Things to check

### DIFF
--- a/app/services/data_migrations/fix_lat_long_flip_flops.rb
+++ b/app/services/data_migrations/fix_lat_long_flip_flops.rb
@@ -14,7 +14,7 @@ module DataMigrations
       deleted_audit_count = 0
       total_audits = audits.count
 
-      Provider.find_each do |p|
+      providers.find_each do |p|
         log("Cleaning up provider #{p.id} (#{p.name_and_code})")
 
         audits_for_provider = audits.where(auditable_id: p.id)
@@ -48,6 +48,14 @@ module DataMigrations
     end
 
   private
+
+    def providers
+      if dry_run?
+        Provider.where(sync_courses: true).limit(50)
+      else
+        Provider.all
+      end
+    end
 
     def delete_audits(description, audits)
       log("Deleting #{audits.count} #{description}")
--- a/app/services/data_migrations/fix_lat_long_flip_flops.rb
+++ b/app/services/data_migrations/fix_lat_long_flip_flops.rb
@@ -1,0 +1,37 @@
+module DataMigrations
+  class FixLatLongFlipFlops
+    TIMESTAMP = 20210323203521
+    MANUAL_RUN = true
+
+    def change
+      audits = Audited::Audit.where(
+        "auditable_type = 'Provider' AND
+        action = 'update' AND
+        audited_changes ?& array['latitude', 'longitude'] AND
+        (audited_changes - 'latitude') - 'longitude' = '{}'",
+      ).order(created_at: :asc)
+
+      audit_count = 0
+      total_audits = audits.count
+
+      Provider.find_each do |p|
+        audits_for_provider = audits.where(auditable_id: p.id)
+
+        # if we already had lat/lng when we made the first spurious update, delete every audit
+        if audits_for_provider.first.audited_changes['latitude'].last.nil?
+          audits_for_provider.destroy_all
+
+        # otherwise, delete all the audits except the one that set it for the first time
+        else
+
+          audits_for_provider.where("audited_changes#>>'{longitude, 1}' is null").destroy_all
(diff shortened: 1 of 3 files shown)


app/services/data_migrations/fix_lat_long_flip_flops.rb
```diff
@@ -0,0 +1,36 @@
...
+      ).order(created_at: :asc)
+
+      audit_count = 0
+      total_audits = audits.count
+
+      Provider.find_each do |p|
+        audits_for_provider = audits.where(auditable_id: p.id)
+
+        # if we already had lat/lng when we made the first spurious update, delete every audit
+        if audits.first.audited_changes['latitude'].last.nil?
```
stevehook, line 21:
```suggestion
        if audits_for_provider.first.audited_changes['latitude'].last.nil?
```
Not sure about this, but `audits.first` would always be the first of the audits for all providers I think?

app/services/data_migrations/fix_lat_long_flip_flops.rb
```diff
@@ -0,0 +1,36 @@
...
+          audits_for_provider.destroy_all
+
+        # otherwise, delete all the audits except the one that set it for the first time
+        else
+          audits_for_provider.where("audited_changes#>>'{longitude, 1}' is null").destroy_all
+          audits_for_provider.where("audited_changes#>>'{longitude, 0}' is null").offset(1).destroy_all
+        end
```
stevehook, line 28: If the last flip was setting the values to `null` the data would be in a inconsistent state with the audit trail. I guess the data itself would get fixed up the next time the sync happens so it's not a big problem.

stevehook: I'm not sure the `is null` condition works as intended here. The problem is that `audited_changes#>>'{longitude, 0}'` is `null` for records that don't have the `longitude` key at all.

Maybe something like `audited_changes#>>'{longitude, 1}' is not null AND audited_changes#>>'{longitude, 0}' is null` etc. is needed.

stevehook (commented):
I've not been able to test thoroughly but have a couple of inline comments.

app/services/data_migrations/fix_lat_long_flip_flops.rb
duncanjbrown: I _think_ the original query guards against that by checking that the audited_changes includes both latitude and longitude and it also contains no other keys?

stevehook:
For the dry run the only thing I can think of is replacing the actual deletes with log statements for say the first 10 providers. I guess we could then spot check the dry run changes with the actual data. I think it's worth doing.

app/services/data_migrations/fix_lat_long_flip_flops.rb
duncanjbrown: Wouldn't line 26 delete all flips to null?

duncanjbrown:
> For the dry run the only thing I can think of is replacing the actual deletes with log statements for say the first 10 providers. I guess we could then spot check the dry run changes with the actual data. I think it's worth doing.

Good idea — I'll add

app/services/data_migrations/fix_lat_long_flip_flops.rb
duncanjbrown: Good catch!

app/services/data_migrations/fix_lat_long_flip_flops.rb
stevehook: Ah yes I think you're right about the conditions, that's makes sense now!

For the data consistency thing, my only concern was that the lat/lng values would be null immediately after the migration but that's dumb - we fixed the flips to null a while back so all the current data should be set to non-null values (where they ever existed).

duncanjbrown:
Thanks for comments @stevehook — I've updated accordingly.

stevehook (approved):
:+1: This looks good to me now. The only question I had was whether it's worth limiting the dry run providers to those that have courses that are synched with apply (so that the 50 sample represents the most useful test cases).

duncanjbrown:
Thanks @stevehook. I can't argue with that, will amend before merging!

PR #4363, opened 2021-03-24
Further refinements to course choice flow

malcolmbaig:
## Context

A set of potential improvements identified while working on https://trello.com/c/eP3BTlVI

- Fix 'too many course choices' error handling
- Rename PickCourseOption service to AddOrUpdateCourseChoice
- Simplify initialization of PickSiteForm and AddOrUpdateCourseChoice
- Use named parameters for AddOrUpdateCourseChoice service

## Guidance to review

- Per-commit is helpful as the changes are only loosely related.
- Each commit is standalone, so anything controversial can be dropped.
- Is the reduced scoping OK? Normally this is something we want to check carefully. In this case I don't think we're introducing a security issue or anything which could affect the data integrity of a candidate's app.

## Link to Trello card
na

## Things to check


stevehook (approved):
Each change makes sense to me. Much tidier!

PR #4364, opened 2021-03-24
Migration to add export_type to the Data Export table

JR-G:
## Context

As part of PR [#4318](https://github.com/DFE-Digital/apply-for-teacher-training/pull/4318) it was decided that a new export_type column should be added to the Data Export table.

- Migration to add the export_type column
- Update the `create` method in the `DataExportsController` to write values to this new column

## Link to Trello card

https://trello.com/c/VOjhcHfX/3159-data-directory

## Things to check


raamSoftwire (approved):
LGTM. But might be best for @duncanjbrown to sign off before you merge

duncanjbrown:
Adding the column looks great. I'm not sure about the backfill — what about records that are added between this and the next PR being merged? It's also good to do backfills as a data migration if possible.

One approach could be 
- add the column here
- add the code which writes values to the column for all newly created records
- backfill everything else
- start depending on the column

PR #4366, opened 2021-03-24
Remove "Future applications" question from rejection flow

benswannack:
## Context
Lots of providers were saying "no" to this question, which is demotivating for candidates

- Remove the question from the flow
- Change the other questions page in the flow to replace the h1 with the legend for the form when there is only one question to ask

## Guidance to review
First commit: Remove the question
Second commit: UI tweaks to the whole flow, but mainly the work to combine the h1 and legend

If the question has been answered for applications already, we want to show this in the UI and emails still. I think this should be enough to achieve that, but I'm double checking all the places this value is used just in case

## Link to Trello card
https://trello.com/c/PLwr7vas/3513-remove-would-you-be-interested-in-future-applications-question-from-structured-reasons-for-rejection-wizard

## Things to check


app/forms/provider_interface/reasons_for_rejection_wizard.rb
```diff
@@ -277,7 +270,7 @@ def last_saved_state
       saved_state = @state_store.read
 
       if saved_state
-        JSON.parse(saved_state)
+        JSON.parse(saved_state).except('interested_in_future_applications_y_n')
```
benswannack, line 273: This should avoid issues where the saved state has the old structure cached

app/views/provider_interface/reasons_for_rejection/edit_other_reasons.html.erb
```diff
@@ -1,33 +1,33 @@
...
-  </div>
   <div class="govuk-grid-column-two-thirds">
     <%= form_with model: @wizard, url: provider_interface_reasons_for_rejection_update_other_reasons_path do |f| %>
 
       <%= f.govuk_error_summary %>
 
+      <span class="govuk-caption-l"><%= @application_choice.application_form.full_name %></span>
+
       <% if @wizard.reason_not_captured_by_initial_questions? %>
+        <h1 class="govuk-heading-l">Other reasons for rejecting this application</h1>
```
benswannack, line 12: This copy change (and others in the flow) have come from the prototype, which I assume is what we want

app/components/support_interface/reasons_for_rejection_dashboard_component.html.erb
```diff
@@ -75,15 +75,6 @@
   sub_reasons_result: sub_reasons_for(:safeguarding_y_n),
 ) %>
 
-<%= render SupportInterface::ReasonForRejectionDashboardSectionComponent.new(
```
benswannack, line 78: It feels like any stats we show about this will be misleading, so I've removed it

app/presenters/flat_reasons_for_rejection_presenter.rb
```diff
@@ -54,6 +54,8 @@ def self.build_top_level_reasons(structured_rejection_reasons)
   end
 
   def self.y_n_to_boolean(string)
+    return nil unless %w[Yes No].include?(string)
```
benswannack, line 57: This better represents the value, rather than false.

This presenter seems to only be used in support, is there a reason I shouldn't move it into the support interface?

steventux (approved):
:+1: This works well for me locally. Support UI and API rendering worked for both legacy and new rejections.

app/presenters/flat_reasons_for_rejection_presenter.rb
cwrw: Agree with the above, I think it makes sense to move it into support

PR #4362, opened 2021-03-24
Post submission dashboard updates

stevehook:
## Context

Following a design review and 2 rounds of user testing, we've updated the design and layout of the post submission candidate _Application Dashboard_ to give better guidance to candidates.

## Changes proposed in this pull request

- [x] h1 title changes from "Application Dashboard" to "Your applications" - should be pluralised if the candidate has applied more than once OR applied to multiple courses in a single application.
- [x] Drop caption beneath h1
- [x] Add contextual guidance between h1 and the course summary boxes. 
image

- [x] Display "Apply again" call to action where appropriate, using Inset Text component.
image

- [x] Display a "Congratulations on your offers" Inset text component when the candidate has received at least one offer AND all providers have responded.
image

- [x] For the course summary boxes, move the course name (and code) into the header, using new CSS styles.
- [x] Add a "View application" link top right of the course box.
- [x] Drop course length, type, date starts and location rows.
- [x] Add an "Interview" row to the course summary boxes when the course is awaiting decision but no interviews are scheduled saying "The provider will be in touch if they want to invite you to an interview."
image

- [x] Copy changes to the interview details when the interview is in future.
- [x] When an interview is in the past (or more than 1 hour in the past?), remove the details and just say "You had an interview on [date]"
- [x] Move "Withdraw this application" to the bottom of the course summary box.
- [x] Add a "Respond to offer" link at the bottom of course summary box, and either add deadline or text saying they can wait for all providers to make a decision.
- [x] Drop right hand column ("Get support", etc")
- [x] Add "Updating your contact details" content to the bottom of the main content.
image

## Guidance to review

- Per commit review highly recommended. This PR is big but the each commit corresponds pretty closely with the list above.
- There is no feature flag for these changes. So it's marked as DO NOT MERGE until the other parts of the Dashboard are ready and we've done sufficient testing.

### Questions

- Should the carry over banner be replaced with an inset component like we've done for apply again?

## Link to Trello card

https://trello.com/c/saDEvkun/3191-update-content-and-layout-of-dashboard-post-submission

## Things to check

- [x] This code does not rely on migrations in the same Pull Request
- [x] If this code includes a migration adding or changing columns, it also backfills existing records for consistency
- [x] API release notes have been updated if necessary
- [x] This code does not rely on the addition/removal of Azure config environment variables in the same Pull Request
[...]

PR #4368, opened 2021-03-24
Add "Previous applications" table to bottom of dashboard view

TobyRet:
## Context

To enable candidates to see feedback or content from previous applications, this adds a summary table to the bottom of the dashboard.

The table displays 1 row per course applied to, not per "round", so 3 courses applied to in one application should display as 3 rows.

In addition, when a candidate has accepted an offer, any other courses they applied to at the same time should appear within this table, rather than further up the page as a more detailed summary.

- Create PreviousApplications component
- Any application choices from previous application forms **and** any application choices from current application forms that do not have an [ACCEPTED](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/services/application_state_change.rb#L8) state will appear in the PreviousApplications component
- Where a current application has an application choice in an ACCEPTED state, all other application choices (not in the ACCEPTED state) will no longer be displayed in the CourseChoicesReview component. These will instead be displayed in the PreviousApplications component that appears below

# Screenshot 

application_choices_v2

## Link to Trello card

https://trello.com/c/5oEXieEN/3193-add-previous-applications-table-to-bottom-of-dashboard-view

## Things to check

### DIFF
--- a/app/components/candidate_interface/course_choices_review_component.rb
+++ b/app/components/candidate_interface/course_choices_review_component.rb
@@ -10,10 +10,10 @@ module CandidateInterface
       show_incomplete: false,
       missing_error: false,
       application_choice_error: false,
-      render_link_to_find_when_rejected_on_qualifications: false
+      render_link_to_find_when_rejected_on_qualifications: false,
+      display_accepted_application_choices: false
     )
       @application_form = application_form
-      @application_choices = @application_form.application_choices.includes(:course, :site, :provider, :offered_course_option, :interviews).order(id: :asc)
       @editable = editable
       @heading_level = heading_level
       @show_status = show_status
@@ -21,6 +21,7 @@ module CandidateInterface
       @missing_error = missing_error
       @application_choice_error = application_choice_error
       @render_link_to_find_when_rejected_on_qualifications = render_link_to_find_when_rejected_on_qualifications
+      @display_accepted_application_choices = display_accepted_application_choices
     end
 
     def course_choice_rows(application_choice)
@@ -43,7 +44,7 @@ module CandidateInterface
     end
 
     def any_withdrawable?
-      @application_form.application_choices.any? do |application_choice|
+      application_choices.any? do |application_choice|
         withdrawable?(application_choice)
       end
     end
@@ -78,6 +79,16 @@ module CandidateInterface
       if application_choice.course_option_availability_error?
         "govuk-inset-text app-inset-text--narrow-border app-inset-text--#{@application_choice_error ? 'error' : 'important'}"
       end
+    end
+
+    def application_choices
+      @application_choices ||= if @display_accepted_application_choices && application_choice_with_accepted_state_present?
+                                 # Reject all applications that do not have an ACCEPTED_STATE
+                                 # These will appear in the CandidateInterface::PreviousApplications component
+                                 application_choices_with_accepted_states
+                               else
+                                 all_application_choices
+                               end
     end
 
   private
@@ -206,6 +217,25 @@ module CandidateInterface
 
     def has_multiple_courses?(application_choice)
       Course.current_cycle.where(provider: application_choice.provider).many?
+    end
+
+    def application_choices_with_accepted_states
+      @application_form
+        .application_choices
+        .includes(:course, :site, :provider, :offered_course_option, :interviews)
+        .order(id: :asc)
+        .select { |ac| ac.status.to_sym.in?(ApplicationStateChange::ACCEPTED_STATES) }
+    end
+
+    def all_application_choices
+      @application_form
+        .application_choices
+        .includes(:course, :site, :provider, :offered_course_option, :interviews)
+        .order(id: :asc)
+    end
+
+    def application_choice_with_accepted_state_present?
+      @application_form.application_choices.any? { |ac| ApplicationStateChange::ACCEPTED_STATES.include?(ac.status.to_sym) }
     end
   end
 end
--- a/app/components/candidate_interface/previous_applications_component.html.erb
+++ b/app/components/candidate_interface/previous_applications_component.html.erb
@@ -16,7 +16,9 @@
         <span class="app-course-choice__course-name">
           <%= course_name_and_code_for(application_choice) %>
         </span>
-        <%= view_application_url_for(application_choice) %>
+        <%= govuk_link_to candidate_interface_review_previous_application_path(application_choice.application_form.id) do %>
+          View application <span class="govuk-visually-hidden"> for <%= application_choice.course.name %> at <%= application_choice.course.provider.name %></span>
+        <% end %>
       </td>
       <td class="govuk-table__cell">
         <%= render CandidateInterface::ApplicationStatusTagComponent.new(application_choice: application_choice) %>
--- a/app/components/candidate_interface/previous_applications_component.html.erb
+++ b/app/components/candidate_interface/previous_applications_component.html.erb
@@ -1,3 +1,4 @@
+<hr class="govuk-section-break govuk-section-break--xl govuk-section-break--visible">
 <table class="govuk-table govuk-!-margin-top-6">
   <caption class="govuk-table__caption govuk-table__caption--m">Your previous applications</caption>
   <thead class="govuk-table__head">
--- a/app/components/candidate_interface/previous_applications_component.html.erb
+++ b/app/components/candidate_interface/previous_applications_component.html.erb
@@ -1,0 +1,27 @@
+<table class="govuk-table govuk-!-margin-top-6">
+  <caption class="govuk-table__caption govuk-table__caption--m">Your previous applications</caption>
+  <thead class="govuk-table__head">
+  <tr class="govuk-table__row">
+    <th scope="col" class="govuk-table__header">Course</th>
+    <th scope="col" class="govuk-table__header">Status</th>
+  </tr>
+  </thead>
+  <tbody class="govuk-table__body">
+  <% application_choices.each do |application_choice| %>
+    <tr class="govuk-table__row" data-qa="application-choice-<%= application_choice.id %>">
+      <td class="govuk-table__cell">
+        <span class="app-course-choice__provider-name">
+          <%= provider_name_for(application_choice) %>
+        </span>
+        <span class="app-course-choice__course-name">
+          <%= course_name_and_code_for(application_choice) %>
+        </span>
+        <%= view_application_url_for(application_choice) %>
+      </td>
+      <td class="govuk-table__cell">
+        <%= render CandidateInterface::ApplicationStatusTagComponent.new(application_choice: application_choice) %>
+      </td>
+    </tr>
+  <% end %>
+  </tbody>
+</table>
--- a/app/components/candidate_interface/previous_applications_component.rb
+++ b/app/components/candidate_interface/previous_applications_component.rb
@@ -27,10 +27,6 @@ module CandidateInterface
       application_choice.course.name_and_code
     end
 
-    def view_application_url_for(application_choice)
-      govuk_link_to('View application', candidate_interface_review_previous_application_path(application_choice.application_form.id))
-    end
-
   private
 
     def eligible_current_application_choices
--- a/app/components/candidate_interface/previous_applications_component.rb
+++ b/app/components/candidate_interface/previous_applications_component.rb
@@ -1,0 +1,75 @@
+module CandidateInterface
+  class PreviousApplicationsComponent < ViewComponent::Base
+    include ViewHelper
+
+    def initialize(candidate:)
+      @candidate = candidate
+    end
+
+    def render?
+      application_choices.present?
+    end
+
+    def application_choices
+      (previous_application_choices + eligible_current_application_choices)
+        .compact
+        .sort_by(&:id)
+        .reverse
+    end
+
+    def provider_name_for(application_choice)
+      application_choice
+        .provider
+        .name
+    end
+
+    def course_name_and_code_for(application_choice)
+      application_choice.course.name_and_code
+    end
+
+    def view_application_url_for(application_choice)
+      govuk_link_to('View application', candidate_interface_review_previous_application_path(application_choice.application_form.id))
+    end
+
+  private
+
+    def eligible_current_application_choices
+      if application_choice_has_accepted_state_present?
+        # Reject all applications that have an ACCEPTED_STATE
+        # These will appear in the CandidateInterface::CourseChoicesReview component
+        application_choices_without_accepted_states
+      else
+        []
+      end
+    end
+
+    def application_choices_without_accepted_states
+      current_application_choices
+        .reject { |ac| ApplicationStateChange::ACCEPTED_STATES.include?(ac.status.to_sym) }
+    end
+
+    def application_choice_has_accepted_state_present?
+      current_application_choices.any? { |ac| ApplicationStateChange::ACCEPTED_STATES.include?(ac.status.to_sym) }
+    end
+
+    def current_application_choices
+      @candidate
+        .current_application
+        .application_choices
+        .includes(:course, :provider)
+        .all
+    end
+
+    def previous_application_choices
+      previous_application_forms = @candidate.application_forms.all - [@candidate.current_application]
+
+      if previous_application_forms
+        previous_application_forms
+          .map { |af| af.application_choices.includes(:course, :provider) }
+          .flatten
+      else
+        []
+      end
+    end
+  end
+end
--- a/spec/components/candidate_interface/course_choices_review_component_spec.rb
+++ b/spec/components/candidate_interface/course_choices_review_component_spec.rb
@@ -392,6 +392,19 @@ RSpec.describe CandidateInterface::CourseChoicesReviewComponent do
     end
   end
 
+  describe '#application_choices' do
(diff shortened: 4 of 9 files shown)


app/components/candidate_interface/course_choices_review_component.rb
```diff
@@ -207,5 +206,17 @@ def has_multiple_sites?(application_choice)
     def has_multiple_courses?(application_choice)
       Course.current_cycle.where(provider: application_choice.provider).many?
     end
+
+    def application_choices
```
TobyRet, line 210: Woops need to add a spec for this

frankieroberto:
@TobyRet hiya. I'm a bit confused by the screenshot - the "Offer accepted" choice should never appear within "Previous applications" - only choices which were unsuccessful (either rejected, declined or withdrawn).

> In the PreviousApplications component, should additional guidance under status be displayed (ie replicating the behaviour in the CourseChoicesReviewComponent at the top of the page)?

No, I think we should only display the status tag in this context. Although I'm not sure what other guidance there would be for unsuccessful applications?

> If a candidate has no previous applications, but has a current application with an application choice which is successful or unsuccessful, then this then gets moved to the 'previous applications' table. Is this the desired behaviour?

No, the intention was that they'd only get moved to "Previous applications" when you either go through the Apply again flow, or you accept one of the offers (in which case the others move to the bottom).

> Generally I'm finding the domain language confusing and this possibly being reflected in the code - i.e having a 'previous applications' table which actually displays 'application choices' 🤷‍♂️ - let me know if I have missed something here!

This is intentional. My feeling (from instinct and a bit of the research) is that candidates often perceive each course choice as a separate "application", even if you complete one application form and then it gets sent to 3 different providers.  Listing each choice as a separate application also avoids having to communicate the parent-child relationship between applications and choices, and lets us iterate in future to potentially allowing a candidate to apply for another course even whilst a current application is awaiting a decision.

It does mean that our interface language has moved away from the domain language in the code a bit though...

app/components/candidate_interface/course_choices_review_component.rb
```diff
@@ -207,5 +206,17 @@ def has_multiple_sites?(application_choice)
...
+      @application_choices ||= @application_form
+        .application_choices
+        .includes(:course, :site, :provider, :offered_course_option, :interviews)
+        .order(id: :asc)
+        .reject { |application_choice| ineligible_application_choice_states.include?(application_choice.status.to_sym) }
+    end
+
+    def ineligible_application_choice_states
+      ApplicationStateChange::SUCCESSFUL_STATES + ApplicationStateChange::UNSUCCESSFUL_END_STATES - %w[offer]
```
davidgisbey, line 219: Having looked at the constants again i think this should reject the application choices that aren't in the  `ACCEPTED_STATES` if an application choice in the accepted states is present.

When a candidate accepts an offer, they transition from the `offer` to `pending_conditions` state. At this point, all other application choices are rejected.

This means that is should never be possible to have more than 1 application choice in the `%i[pending_conditions conditions_not_met recruited offer_deferred]` states.

davidgisbey: I think this 
```

def application_choices
      @application_choices ||= @application_form
        .application_choices
        .includes(:course, :site, :provider, :offered_course_option, :interviews)
        .order(id: :asc)
        .reject { |application_choice| ineligible_application_choice_states.include?(application_choice.status.to_sym) }
    end

    def ineligible_application_choice_states
      ApplicationStateChange::SUCCESSFUL_STATES + ApplicationStateChange::UNSUCCESSFUL_END_STATES - %w[offer]
    end
```

should be something like this 

```
def application_choices
      @application_choices ||= reject_ineligible_application_choices
    end

    def reject_ineligible_application_choice_states
      if @application_form.application_choices.any? { |ac| ApplicationStateChange::ACCEPTED_STATES.include?(ac) }
        @application_form
         .application_choices
         .includes(:course, :site, :provider, :offered_course_option, :interviews)
         .order(id: :asc)
         .reject { |ac| ApplicationStateChange::ACCEPTED_STATES.exclude?(ac)}
       else
         @application_form
          .application_choices
          .includes(:course, :site, :provider, :offered_course_option, :interviews)
          .order(id: :asc)
       end
     end
```

This can def be cleaned up but you get the idea

davidgisbey: i think we can also add a method to the ApplicationChoice like 

```
def successful?
  ApplicationStateChange::ACCEPTED_STATES.include?(self)
end
```

then we can just do 


```

    def reject_ineligible_application_choice_states
      if @application_form.application_choices.any?(&:successful?)
        @application_form
         .application_choices
         .includes(:course, :site, :provider, :offered_course_option, :interviews)
         .order(id: :asc)
         .select(&:successful?)
       else
         @application_form
          .application_choices
          .includes(:course, :site, :provider, :offered_course_option, :interviews)
          .order(id: :asc)
       end
     end
```

frankieroberto: @davidgisbey hiya. I'm not quite following the ruby code, but your examples look good. Although in most cases, if there are 3 previous application choices at the bottom, the candidate will only see a single choice at the top, as you can only apply for 1 at a time during "apply again" at the moment.  The exception would be if the previous applications were from previous "cycles".

The intent here is mainly:

* To hide irrelevant application choices once an applicant has accepted one, so that they can focus on meeting those conditions (and be constantly reminded of any negative feedback)
* To give candidates who are applying again access to their previous application forms and feedback from providers (currently we don't link to these at all once you've submitted).

In both cases, only the minimum of information is needed, as they're unlikely to be looked at, and candidates can click through for the full detail.

TobyRet: @davidgisbey @frankieroberto - thanks for clarifying. That all makes sense now 😁

frankieroberto: @TobyRet no probs. Let me know if you'd like me to test this at all, when it's ready. Cheers!

spec/components/candidate_interface/course_choices_review_component_spec.rb
```diff
@@ -375,6 +375,19 @@
     end
   end
 
+  describe '#application_choices' do
```
TobyRet, line 378: Keen to know how people feel about this spec. To me it felt easier to test the logic in the `#application_choices` method rather than assert against what is and what is not rendered. Is this a code smell though?

spec/components/candidate_interface/previous_applications_component_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+
+      expect(result.css('.app-table__caption--m').text).to include('Your previous applications')
+      expect(result.css('.app-course-choice__provider-name').text).to include(unsuccessful_application_choice.course.provider.name)
+      expect(result.css('.app-course-choice__course-name').text).to include("#{unsuccessful_application_choice.course.name} (#{unsuccessful_application_choice.course.code})")
+      expect(result.css('.app-tag').text).to include('Unsuccessful')
+      expect(result.css('.govuk-link').text).to include('View application')
+    end
+  end
+
+  describe '#application_choices' do
```
TobyRet, line 51: Same here re: https://github.com/DFE-Digital/apply-for-teacher-training/pull/4368/files#r601568795

app/components/candidate_interface/previous_applications_component.html.erb
```diff
@@ -0,0 +1,27 @@
+<table class="govuk-table govuk-!-margin-top-6">
+  <caption class="govuk-table__caption app-table__caption--m">Your previous applications</caption>
```
frankieroberto, line 2: This can be changed to the govuk-frontend class now, I managed to get it added upstream. 😄 

```suggestion
  <caption class="govuk-table__caption govuk-table__caption--m">Your previous applications</caption>
```

app/components/candidate_interface/previous_applications_component.rb
```diff
@@ -0,0 +1,77 @@
...
+
+    def course_name_and_code_for(application_choice)
+      name = application_choice.course.name
+      code = application_choice.course.code
+
+      "#{name} (#{code})"
+    end
+
+    def view_application_url_for(application_choice)
+      govuk_link_to('View application', candidate_interface_review_previous_application_path(application_choice.application_form.id))
```
frankieroberto, line 34: We should probably add some visually-hidden text to the link text here, for screen reader users. Maybe something like:

```
<a href="#">View application<span class="govuk-visually-hidden"> for <%= application_choice.course.name %> at <%= application_choice.course.provider.name %></span></a>
```

app/frontend/styles/_table.scss
```diff
@@ -0,0 +1,33 @@
...
+}
+
+.app-table__caption--m {
+  @include govuk-font($size: 24, $weight: bold);
+  margin-bottom: govuk-spacing(3);
+}
+
+.app-table__caption--s {
+  @include govuk-font($size: 19, $weight: bold);
+}
```
frankieroberto, line 33: These can be dropped now, as they've been added upstream.

```suggestion
```

spec/system/candidate_interface/submitting/candidate_submitting_application_spec.rb
```diff
@@ -212,7 +212,9 @@ def then_i_can_see_my_application_dashboard
   end
 
   def when_i_click_view_application
-    click_link 'View application'
+    within '.govuk-hint' do
+      click_link 'View application'
+    end
```
frankieroberto, line 217: Is this really within hint text somewhere?

app/components/candidate_interface/course_choices_review_component.rb
```diff
@@ -207,5 +216,24 @@ def has_multiple_sites?(application_choice)
     def has_multiple_courses?(application_choice)
       Course.current_cycle.where(provider: application_choice.provider).many?
     end
+
+    def application_choices_with_accepted_states
+      @application_form
+        .application_choices
+        .includes(:course, :site, :provider, :offered_course_option, :interviews)
+        .order(id: :asc)
+        .reject { |ac| ApplicationStateChange::ACCEPTED_STATES.exclude?(ac.status.to_sym) }
```
frankieroberto, line 225: Would this be slightly easier to follow by reversing the double-negative?

```suggestion
        .select { |ac| ApplicationStateChange::ACCEPTED_STATES.include?(ac.status.to_sym) }
```

or even

```suggestion
        .select { |ac| ac.status.to_sym.in? ApplicationStateChange::ACCEPTED_STATES }
```

frankieroberto (changes requested):
Looking good! 👍 

Some of the CSS got upstreamed into govuk-frontend since I wrote it. Forgot to update the prototype, apologies!

app/components/candidate_interface/previous_applications_component.rb
TobyRet: 👍

app/components/candidate_interface/previous_applications_component.html.erb
TobyRet: 👍

app/frontend/styles/_table.scss
TobyRet: 👍

app/components/candidate_interface/course_choices_review_component.rb
TobyRet: Yep good call!

spec/system/candidate_interface/submitting/candidate_submitting_application_spec.rb
TobyRet: Yep. See https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/views/candidate_interface/submitted_application_form/complete.html.erb#L9

Let me know if you want changes here @frankieroberto 👍

frankieroberto: @TobyRet ah yes! That will get changed with #4362, so I think you can leave this as is for now.

frankieroberto (approved):

TobyRet:
Seem to be having an issue with the review app. I will close and reopen this PR

app/components/candidate_interface/previous_applications_component.rb
```diff
@@ -0,0 +1,73 @@
...
+        .course
+        .provider
+        .name
+    end
+
+    def course_name_and_code_for(application_choice)
+      name = application_choice.course.name
+      code = application_choice.course.code
+
+      "#{name} (#{code})"
```
davidgisbey, line 30: `application_choice.course.name_and_code`

app/components/candidate_interface/previous_applications_component.rb
```diff
@@ -0,0 +1,73 @@
...
+
+    def current_application_choices
+      @application_choices ||= @application_form
+        .application_choices
+        .includes(:course, :site, :provider, :offered_course_option, :interviews)
+        .order(id: :asc)
+    end
+
+    def previous_application_choices
+      previous_application_form = @application_form.previous_application_form
```
davidgisbey, line 62: The candidate could have multiple previous application_forms so I think this probably needs to be something like

```
candidate =  @application_form.candidate
previous_application_forms = candidate.application_forms - [candidate.current_application]

if previous_application_forms
    previous application_forms.map(&:application_choices).flatten.includes(course: provider)
end 

```

spec/components/candidate_interface/course_choices_review_component_spec.rb
davidgisbey: Personally, this makes sense to em.

spec/components/candidate_interface/previous_applications_component_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+    context "with an application choice 'pending_conditions' and another that has been 'rejected'" do
+      let(:current_application_form) { create_application_form_with_course_choices(statuses: %w[pending_conditions rejected]) }
+
+      it 'renders component with rejected application choice in a table' do
+        rejected_application_choice = current_application_form.application_choices.last
+        result = render_inline(described_class.new(application_form: current_application_form))
+
+        expect(result.css('.govuk-table__caption--m').text).to include('Your previous applications')
+        expect(result.css('.app-course-choice__provider-name').text).to include(rejected_application_choice.course.provider.name)
+        expect(result.css('.app-course-choice__course-name').text).to include("#{rejected_application_choice.course.name} (#{rejected_application_choice.course.code})")
```
davidgisbey, line 28:
```
expect(result.css('.app-course-choice__course-name').text).to include(rejected_application_choice.course.name_and_code)
```

spec/components/candidate_interface/previous_applications_component_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+    let(:current_application_form) { create(:completed_application_form, submitted_at: 3.days.ago, previous_application_form: previous_application_form) }
+    let(:previous_application_form) { create(:completed_application_form, submitted_at: 10.days.ago) }
+    let!(:unsuccessful_application_choice) { create(:application_choice, :with_rejection, application_form: previous_application_form) }
+
+    it 'renders component with rejected application choice in a table' do
+      result = render_inline(described_class.new(application_form: current_application_form))
+
+      expect(result.css('.govuk-table__caption--m').text).to include('Your previous applications')
+      expect(result.css('.app-course-choice__provider-name').text).to include(unsuccessful_application_choice.course.provider.name)
+      expect(result.css('.app-course-choice__course-name').text).to include("#{unsuccessful_application_choice.course.name} (#{unsuccessful_application_choice.course.code})")
```
davidgisbey, line 45: same as above

spec/components/candidate_interface/previous_applications_component_spec.rb
davidgisbey: As above this makes sense to me. I guess we could test that page doesn't include 'Your previous applications', but i'm not convinced it's necessary

app/components/candidate_interface/previous_applications_component.rb
TobyRet: Ah ok. I'll make the changes - thanks

app/components/candidate_interface/previous_applications_component.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+
+    def render?
+      application_choices.present?
+    end
+
+    def application_choices
+      (previous_application_choices + eligible_current_application_choices)
+        .compact
+        .sort_by(&:id)
```
davidgisbey, line 16: `
.sort_by(&:id)
.reverse
`

I think the most recently submitted application choices should be at the top.

app/components/candidate_interface/previous_applications_component.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+
+    def provider_name_for(application_choice)
+      application_choice
+        .course
+        .provider
+        .name
+    end
```
davidgisbey, line 24:
```
application_choice
        .provider
        .name
```

will work here as it has through relationship with course 

image

davidgisbey (approved):

app/components/candidate_interface/previous_applications_component.rb
TobyRet: Yep no prob - now sorting by `updated_at` field

PR #4369, opened 2021-03-24
Add change_offer wrapper and refactor make offer wrapper

SamJamCul:
## Context

Adding wrapper services for make_an_offer and change_an_offer for the Offer workflow stuff

Added a change_offer wrapper to the change_an_offer service, and refactored the make_offer wrapper to simplify the validation. 

## Trello card

https://trello.com/c/9I0RTTts

### DIFF
--- a/app/services/make_offer.rb
+++ b/app/services/make_offer.rb
@@ -16,10 +16,13 @@ class MakeOffer
                                     application_choice: application_choice,
                                     course_option: course_option,
                                     offer_conditions: conditions)
-    make_an_offer.save
 
-    if make_an_offer.errors[:base].include?(MakeAnOffer::STATE_TRANSITION_ERROR)
-      ApplicationStateChange.new(application_choice).make_offer!
+    unless make_an_offer.save
+      if make_an_offer.errors[:base].include?(MakeAnOffer::STATE_TRANSITION_ERROR)
+        ApplicationStateChange.new(application_choice).make_offer!
+      else
+        raise 'Unable to complete save on make_an_offer'
+      end
     end
   end
 end
--- a/spec/services/change_offer_spec.rb
+++ b/spec/services/change_offer_spec.rb
@@ -1,0 +1,114 @@
+require 'rails_helper'
+
+RSpec.describe ChangeOffer do
+  include CourseOptionHelpers
+
+  let(:change_offer) do
+    ChangeOffer.new(actor: provider_user,
+                    application_choice: application_choice,
+                    course_option: course_option,
+                    conditions: conditions)
+  end
+
+  describe '#save!' do
+    let(:application_choice) { create(:application_choice) }
+    let(:course_option) { course_option_for_provider(provider: application_choice.course_option.provider) }
+    let(:provider_user) { create(:provider_user, providers: [build(:provider)]) }
+    let(:conditions) { [Faker::Lorem.sentence] }
+
+    describe 'if the actor is not authorised to perform this action' do
+      it 'throws an exception' do
+        expect {
+          change_offer.save!
+        }.to raise_error(
+          ProviderAuthorisation::NotAuthorisedError,
+          'You are not permitted to view this application. The specified course is not associated with any of your organisations. You do not have the required user level permissions to make decisions on applications for this provider.',
+        )
+      end
+    end
+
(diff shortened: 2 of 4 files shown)


app/services/make_offer.rb
```diff
@@ -16,10 +16,7 @@ def save!
                                     application_choice: application_choice,
                                     course_option: course_option,
                                     offer_conditions: conditions)
-    make_an_offer.save
 
-    if make_an_offer.errors[:base].include?(MakeAnOffer::STATE_TRANSITION_ERROR)
-      ApplicationStateChange.new(application_choice).make_offer!
-    end
+    raise 'Unable to complete save on make_an_offer' unless make_an_offer.save
```
despo, line 20: As the only error added to MakeAnOffer is the state transition error, and this behavior (raising state transition errors) is replicated across other services I don't think this change is required as it enables a better understanding of why this exception occured..

spec/services/make_offer_spec.rb
```diff
@@ -11,60 +16,27 @@
...
-        make_offer.save!
+    it 'raises an error when the save is unsuccessful' do
+      make_an_offer = instance_double(MakeAnOffer)
+      allow(make_an_offer).to receive(:save).and_return(false)
+      allow(MakeAnOffer).to receive(:new) { make_an_offer }
 
-        expect(SetDeclineByDefault).to have_received(:new).with(application_form: application_choice.application_form)
-        expect(SendNewOfferEmailToCandidate).to have_received(:new).with(application_choice: application_choice)
-      end
+      expect { make_offer.save! }.to raise_error('Unable to complete save on make_an_offer')
```
despo, line 39: The purpose of this spec is to verify that the correct set of services is called. As the goal is to eventually transition to only using the MakeOffer service, which should ensure that the correct set of services are called, these changes are not required.

spec/services/change_offer_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    it 'raises an error when the save is unsuccessful' do
+      change_an_offer = instance_double(ChangeAnOffer)
+      allow(change_an_offer).to receive(:save).and_return(false)
+      allow(ChangeAnOffer).to receive(:new) { change_an_offer }
+
+      expect { change_offer.save! }.to raise_error('Unable to complete save on change_an_offer')
+    end
+  end
+end
```
despo, line 42: This looks ok, however I think that similar to the make offer spec - prior to the changes included in this PR - it should ensure that the correct set of services is called (e.g. it should set some expectations for the calls to SetDeclineByDefault, CandidateMailer and StateChangeNotifier). This way if any further changes take place in ChangeAnOffer for any reason we will be aware and none of the expected ChangeOffer functionality will change.

spec/services/make_offer_spec.rb
mnacos: The spec for `MakeAnOffer` already verifies these services are called and there is another scenario in the `MakeOffer` spec which verifies that `MakeAnOffer` is called. If we want a spec that verifies the full range of actions `MakeOffer` performs, we have a comprehensive one included in the previous PR. With this PR's approach, the only checks that feel appropriate are the ones proposed.

app/services/make_offer.rb
mnacos: You are right no other error should normally occur, based on the current code, but I think what we have here is safer. Checking for the inclusion of a specific error and raising an exception only if found feels too specific, as we essentially ignore the return value of `save`.

cwrw: I've had a think about this and I think we can support both cases:
1) if we have the ability to surface a granular error we should do so
2) We should also guard against unexpected errors that might occur in the wrapped class past the validations.

Maybe something similar to this will work, where we surface a transition error if it has occurred, but also raise the other error otherwise:

```
unless make_an_offer.save
  if make_an_offer.errors[:base].include?(MakeAnOffer::STATE_TRANSITION_ERROR)
    ApplicationStateChange.new(application_choice).make_offer!
  else
    raise 'Unable to complete save on make_an_offer'
  end
end
```

cwrw (approved):
Looks good, thanks for addressing the previous feedback, just the one comment for the errors

cwrw (approved):

PR #4361, opened 2021-03-24
Export documentation 3

raamSoftwire:
## Context

This is part of a wider effort to improve the visibility and documentation of the exports and is a continuation of https://github.com/DFE-Digital/apply-for-teacher-training/pull/4295 and https://github.com/DFE-Digital/apply-for-teacher-training/pull/4333.

There are still a number of exports to document, so I have raised this PR for 5 exports to avoid creating one really long PR at the end of this work.

## Changes proposed in this pull request

For each doc added
- create a yml file for the documentation
- decide which columns are required and de-dup any in the common columns file
- update the export file to use symbols instead of strings in the hash
- add a `it_behaves_like 'a data export'` test to the spec along with any setup required
- update existing tests to use symbols instead of strings in the hash

## Guidance to review

- commit by commit

## Link to Trello card

https://trello.com/c/s2pOqGYM/3162-export-documentation

## Things to check

- [x] This code does not rely on migrations in the same Pull Request
- [x] If this code includes a migration adding or changing columns, it also backfills existing records for consistency
- [x] API release notes have been updated if necessary
- [x] This code does not rely on the addition/removal of Azure config environment variables in the same Pull Request
- [x] New environment variables have been [added to the Azure config](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/docs/environment-variables.md#azure-hosting-devops-pipeline)

PR #4370, opened 2021-03-24
3162 export documentation 4

raamSoftwire:
## Context

This is part of a wider effort to improve the visibility and documentation of the exports and is a continuation of https://github.com/DFE-Digital/apply-for-teacher-training/pull/4295, https://github.com/DFE-Digital/apply-for-teacher-training/pull/4333 and https://github.com/DFE-Digital/apply-for-teacher-training/pull/4361.

This PR will document the final, outstanding exports. This is the last one 🎉!

For each doc added

## Guidance to review

- commit by commit

## Link to Trello card

## Things to check


app/exports/common_columns/_common_columns.yml
```diff
@@ -89,3 +89,10 @@ rejection_reason:
   type: string
   description: The reason the provider gave for rejecting the candidate
   example: Absence of English GCSE
+
+structured_rejection_reasons:
+  type: string
+  description: A list of structured reasons for rejection, split by new lines (\n)
```
raamSoftwire, line 95: I've seen a few different approaches to lists and it would be good to commonise these. But I think that is something to do as a separate piece of work.

app/exports/common_columns/_provider_information.yml
```diff
@@ -39,7 +39,13 @@ course_level:
...
 program_type:
   type: string
-  description: Not provided yet
+  description: The program type
+  enum:
+    - scitt_programme
+    - school_direct_training_programme
+    - higher_education_programme
+    - school_direct_salaried_training_programme
+    - pg_teaching_apprenticeship
```
raamSoftwire, line 48: I couldn't find a definitive list for these, so took my best guess. Could someone from the ProVendor team confirm these please?

app/exports/common_columns/_provider_information.yml
```diff
@@ -60,3 +66,15 @@ course_code:
...
+  type: string
+  description: The provider type
+  enum:
+    - scitt
+    - lead_school
+    - university
+    - hei
+    - ratified_by_scitt
+    - ratified_by_university
+    - ratified_by_hei
```
raamSoftwire, line 80: Same here.

app/exports/persona_export.yml
```diff
@@ -0,0 +1,88 @@
...
+      - west_midlands
+      - yorkshire_and_the_humber
+
+  accrediting_provider_type:
+    type: string
+    description: The accrediting provider type
+    enum:
+      - scitt
+      - university
+      - hei
```
raamSoftwire, line 56: And these please

app/exports/persona_export.yml
```diff
@@ -0,0 +1,88 @@
...
+
+  degree_award_year:
+    type: integer
+    description: The award year for the candidate’s most recent degree
+    example: 2019
+
+  degree_type:
+    type: string
+    description: The candidate’s most recent degree type. [See list of degree types](/support/docs/qualifications)
```
raamSoftwire, line 65: 🎉

app/services/support_interface/tad_provider_stats_export.rb
```diff
@@ -1,42 +1,47 @@
...
-        .map { |c| course_to_row(c) }
-        .sort_by { |r| r[:provider_code] }
-    end
+    def data_for_export
+      data_for_export = courses.map do |course|
+        choice_statuses = choice_statuses(course)
 
-    alias_method :data_for_export, :call
+        {
+          provider_id: nil,
```
raamSoftwire, line 8: I assumed these `nil` columns are here for a reason. But if not can we remove them? @duncanjbrown

app/presenters/flat_reasons_for_rejection_presenter.rb
```diff
@@ -61,12 +61,16 @@ def self.subreason_exists?(reason, subreason)
     reason.include?(subreason)
   end
 
-  def self.title_for_top_level_reason(reason)
+  def self.humanized_title_for_top_level_reason(reason)
```
raamSoftwire, line 64: This is a bit of a hack. I didn't want to change the methods on lines 68 and 72 as it would have made the lines at the top of the file even longer than they are now.

app/services/data_set_documentation.rb
```diff
@@ -2,35 +2,27 @@ class DataSetDocumentation
   def self.for(klass)
     name = klass.name.demodulize.underscore
 
-    begin
-      spec = YAML.load_file(Rails.root.join("app/exports/#{name}.yml"))
-    rescue Errno::ENOENT
-      return false
```
raamSoftwire, line 8: This will now raise the `Errno::ENOENT` error instead if there is a doc file missing

app/services/data_set_documentation.rb
```diff
@@ -2,35 +2,27 @@ class DataSetDocumentation
...
-  def self.shadowed_common_columns(klass)
-    name = klass.name.demodulize.underscore
-
-    begin
-      spec = YAML.load_file(Rails.root.join("app/exports/#{name}.yml"))
-    rescue Errno::ENOENT
-      return false
+  def self.check_for_shadowed_columns(common_columns, custom_columns)
+    if (custom_columns.keys.map(&:to_sym) & common_columns.keys.map(&:to_sym)).any?
+      raise 'There are columns in the export documentation that shadow the shared common columns'
```
raamSoftwire, line 17: I wasn't sure if this was the correct way to raise exceptions. But I've seen similar examples in the codebase.

spec/support/data_export_shared_example.rb
```diff
@@ -9,10 +9,4 @@
 
     expect(exported_columns).to match_array(documented_columns)
   end
-
-  it 'doesn’t shadow any common columns' do
-    shadowed_columns = DataSetDocumentation.shadowed_common_columns(described_class)
-
-    expect(shadowed_columns.count).to eq(0)
-  end
```
raamSoftwire, line 17: If there are any shadowed columns the class will raise an exception during the first test. 

It would still be nice to add a test in if we could, but I'm struggling to see how you'd do it. Feedback very welcome!

app/exports/common_columns/_provider_information.yml
raamSoftwire: Cheers Ben! 

https://github.com/DFE-Digital/apply-for-teacher-training/blob/34ff311899125761b3bbeac3d7735540bbf3f503/app/models/course.rb#L43

benswannack: These are defined in the enum in the course object. More generally, the values for this come from Find when we sync

app/exports/common_columns/_common_columns.yml
raamSoftwire: I fixed this as part of this PR in the end

app/exports/common_columns/_provider_information.yml
davidgisbey: So this is the list from the TTAPI

```
  enum provider_type: {
    scitt: "B",
    lead_school: "Y",
    university: "O",
    unknown: "",
    invalid_value: "0", # there is only one of these in the data
  }
```
I think ratified_by_univeristy and ratified_by_hei are the same thing.

I'm just going to put a PIM in to check the data, but i think the ratified values are to do with the provider_type for the accredited_provider

app/exports/persona_export.yml
davidgisbey: This should always be university or scitt.

app/services/support_interface/offer_conditions_export.rb
```diff
@@ -36,31 +36,31 @@ def offers
   private
 
     def qualification_type(form)
-      qualifications(form).map(&:level).join(',')
+      qualifications(form).map(&:level).join(', ')
```
davidgisbey, line 39: Should this be 

```
 qualifications(form).map(&:qualification_type).join(', ')
```

If we use level it will only ever return one of these

`  enum level: {
    degree: 'degree',
    gcse: 'gcse',
    other: 'other',
  }`

app/services/support_interface/tad_provider_stats_export.rb
davidgisbey: Yea i don't have any context on this, It does seem strange, but maybe they match and populate these columns on their end with their id's?

spec/support/data_export_shared_example.rb
davidgisbey: Yea this is tricky. I guess you could test the DataSetDocumentation service directly?

Something like

```
  it 'raises an error if there are shared columns' do
    common_columns = ['provider_code']
    custom_columns = ['provider_code']

    expect(DataSetDocumentation.check_for_shadowed_columns(common_columns, 
      custom_columns)).raise_error Errno::ENOENT 
  end
```

davidgisbey: The first test should catch it anyway though so i think that's fine.

davidgisbey (approved):
LGTM Raam. I'm happy for it to be merged once the tests go green. @duncanjbrown might be able to think of a good way to test the raised exception.

spec/support/data_export_shared_example.rb
raamSoftwire: Neat solution David, thanks! But I also agree that we probably don't need it here.

PR #4371, opened 2021-03-24
Create export for notes analysis

steventux:

Adds a support interface export of notes data relating to candidates, their course provider and the provider relationships for the provider.

image

## Guidance to review

Is there a consideration around offered course vs. desired course? I've not accommodated offered course but this might be necessary.
I've assumed the provider relationship counts relate to the provider of the course from the application and not the providers the provider user belongs to as this can be > 1.

## Link to Trello card

https://trello.com/c/69Zrsy7u/3462-create-csv-specification-for-notes-analysis

## Things to check


app/services/support_interface/notes_export.rb
```diff
@@ -0,0 +1,32 @@
...
+      ).pluck(
+        :subject,
+        :message,
+        :created_at,
+        :"candidates.id",
+        :"providers.code",
+        :"providers.name",
+        :"provider_users.id",
+        'COUNT(tpp.id)',
+        Arel.sql('COUNT(tpp.id) + COUNT(rpp.id)'),
```
benswannack, line 26: These seem odd to include in a notes export. Can we check the need for this and make sure we're meeting it?

app/services/support_interface/notes_export.rb
```diff
@@ -0,0 +1,32 @@
...
+        'LEFT JOIN provider_relationship_permissions rpp ON rpp.ratifying_provider_id = providers.id',
+      ).group(
+        :subject, :message, :created_at, :"candidates.id", :"providers.code", :"providers.name", :"provider_users.id"
+      ).pluck(
+        :subject,
+        :message,
+        :created_at,
+        :"candidates.id",
+        :"providers.code",
+        :"providers.name",
```
benswannack, line 23: I suspect the intention of the column here is to capture which provider a user is from when making the note. I guess that means we do need to include the offered_option too. Worth checking though

benswannack (commented):
Looking good. Needs clarification on some of the columns

app/services/support_interface/notes_export.rb
steventux: I'll check this, the value can be multiple provider names and codes if coming from the join to the provider user.

app/services/support_interface/notes_export.rb
steventux: Checked up about this and it is necessary for the analysis. I've since changed the query though as the way we derive the exported provider details and subsequent counts has changed.

app/services/support_interface/notes_export.rb
steventux: The desire here is to export the course provider assuming it's a provider belonging to the provider user making the note, otherwise we can safely assume the note was made by someone on the accredited provider side and export that provider instead.

app/exports/notes_export.yml
```diff
@@ -0,0 +1,31 @@
...
+  note_message:
+    type: string
+    description: Body of the note
+    example: This application was withdrawn because reasons.
+  note_created_at:
+    type: string
+    format: datetime
+    description: The date and time when the note was created
+    example: 2020-11-01T00:00:00+00:00
+  provider_user_id:
```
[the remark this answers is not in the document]
steventux: https://github.com/DFE-Digital/apply-for-teacher-training/pull/4371/commits/02b3d887eec1541923aac8287c6bf76c827b076f

app/services/support_interface/notes_export.rb
```diff
@@ -0,0 +1,30 @@
...
+                .includes(
+                  { provider_user: { provider_permissions: :provider } },
+                  { application_choice: [{ application_form: :candidate }, { course_option: { course: %i[provider accredited_provider] } }] },
+                )
+
+      notes.map do |note|
+        providers = note.provider_user.providers
+        training_provider = note.application_choice.course_option.course.provider
+        ratifying_provider = note.application_choice.course_option.course.accredited_provider
+        provider = providers.include?(training_provider) ? training_provider : ratifying_provider
```
benswannack, line 14: Is there a case we have missed here? What if the offered_course was with a different training provider and the user making the note is from the new training provider?
More generally, it might be nice to guarantee we return something from `provider_user.providers` by choosing the first match between that and the list of providers that comes from `course` and `offered_course`

steventux: I've amended here https://github.com/DFE-Digital/apply-for-teacher-training/pull/4371/commits/6ecfe77267cc0d1528e0b5bea44a03c23eaba27f and added a more complex test scenario where the ratifying provider user is associated to an offered course option.
I don't think we need to sanitise whether the various permeations of providers belong to the note-making provider user as this should always be the case given the nature of the default 'view applications' permission.

benswannack: I was thinking that sometimes providers lose the view applications permission when an application is offered a course that is not run or ratified by their organisation

spec/services/support_interface/notes_export_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'rails_helper'
+
+RSpec.describe SupportInterface::NotesExport do
```
raamSoftwire, line 3: All exports should now include `it_behaves_like 'a data export'` in the specs. See the other export specs for an example of how this is done. 

https://github.com/DFE-Digital/apply-for-teacher-training/blob/2677f58aaaa9679494616156d9b60a2ec8190900/spec/services/support_interface/sites_export_spec.rb#L13

app/services/support_interface/notes_export.rb
benswannack: Yeah, it may get confusing which provider is doing what. I think this situation is quite possible and we should at least avoid returning incorrect data here. e.g.

```
training_provider1 = create(:provider)
training_provider2 = create(:provider)
ratifying_provider = create(:provider)

course = create(:course, provider: training_provider1, accredited_provider: ratifying_provider)
offered_course = create(:course, provider: training_provider2, accredited_provider: ratifying_provider)

application_choice = create(:application_choice, candidate: candidate, course_option: create(:course_option, course: course), offered_course_option: create(:course_option, course: offered_course))

provider_user1 = create(:provider_user, providers: [training_provider1])
provider_user2 = create(:provider_user, providers: [training_provider2])

create(:provider_relationship_permissions, training_provider: training_provider1, ratifying_provider: ratifying_provider)
create(:provider_relationship_permissions, training_provider: training_provider2, ratifying_provider: ratifying_provider)

note1 = create(:note, application_choice: application_choice, provider_user: provider_user1)
note2 = create(:note, application_choice: application_choice, provider_user: provider_user2)
```

And then I'd expect the export to do something like:
```
        {
          note_subject: note1.subject,
          note_message: note1.message,
          note_created_at: note1.created_at,
          candidate_id: candidate.id,
          provider_code: training_provider1.code,
          provider_name: training_provider1.name,
          provider_user_id: provider_user1.id,
          number_of_training_provider_relationships: ...,
          total_number_of_provider_relationships: ...,
        },
        {
          note_subject: note2.subject,
          note_message: note2.message,
          note_created_at: note2.created_at,
          candidate_id: candidate.id,
          provider_code: training_provider2.code,
          provider_name: training_provider2.name,
          provider_user_id: provider_user2.id,
          number_of_training_provider_relationships: ...,
          total_number_of_provider_relationships: ...,
        },
```
at the moment the first row will have the ratifying provider details, even though none of the users making notes are from the ratifying org

steventux: Thanks @benswannack for the test scenario, I've amended the export and included the test case here https://github.com/DFE-Digital/apply-for-teacher-training/pull/4371/commits/066d30bbe8833a488ea5bbd923da7df18570c746 so now we return no provider or org permissions count data in the event of the application becoming disassociated with the note-making user.

benswannack (approved):
Looks good!

PR #4373, opened 2021-03-24
Use GovukComponent::BackLink

paulrobertlloyd:
## Context

We already have the `govuk_back_link_to` helper for back links, so arguably we don’t need to use this component.

However, calling `GovukComponent::BackLink` from within this helper means we can refactor our back links to stop serving additional and duplicated CSS.

Calls `GovukComponent::BackLink` from within `govuk_back_link_to`. As the component uses `govuk-back-link` class, we no longer need to add duplicate styles for `app-back-link`.

## Guidance to review

Does this make sense? I think we could do similarly for our button helpers.

## Things to check

### DIFF
--- a/app/frontend/styles/_back_link.scss
+++ b/app/frontend/styles/_back_link.scss
@@ -1,8 +1,3 @@
-.app-back-link {
-  @extend .govuk-back-link;
-  @extend .govuk-\!-display-none-print;
-}
-
 .app-back-link--no-js {
   display: none;
 }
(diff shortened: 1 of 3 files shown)


app/frontend/styles/_back_link.scss
```diff
@@ -1,8 +1,3 @@
-.app-back-link {
-  @extend .govuk-back-link;
-  @extend .govuk-\!-display-none-print;
```
duncanjbrown, line 3: Goodbye hack!

duncanjbrown (approved):
Tidy!

PR #4374, opened 2021-03-24
Use GovukComponent::Breadcrumbs

paulrobertlloyd:
## Context

Uses `GovukComponent::Breadcrumbs` to render breadcrumb component.

Most changes relate to `GovukComponent::Breadcrumb` take a `breadcrumbs` hash instead of an `items` array. Annoyingly, we need to keep our own `BreadcrumbsComponent` so that we can add the `govuk-!-display-none-print` class to all breadcrumbs – better to do this than attempt the same via Sass `@extends` in CSS.

## Guidance to review

Possibly a few places where I’ve updated the code to work with hashes which can be written better – was scrabbling around a little on doing these conversions.

## Things to check


app/helpers/view_helper.rb
```diff
@@ -30,7 +30,18 @@ def govuk_back_link_to(url = :back, body = 'Back')
...
+      href: url,
+      classes: classes,
+    )
+  end
+
+  def govuk_breadcrumbs(breadcrumbs)
+    render GovukComponent::Breadcrumbs.new(
+      breadcrumbs: breadcrumbs,
+      hide_in_print: true,
+    )
```
paulrobertlloyd, line 44: Added a helper so that we can pass common options into GovukComponent::Breadcrumbs.

stevehook: I think govuk-components already defines this helper (https://github.com/DFE-Digital/govuk-components/blob/master/app/helpers/govuk_components_helper.rb). I think this is OK but it could cause confusion because it's overriding the helper from the gem.

app/views/support_interface/application_forms/_application_navigation.html.erb
```diff
@@ -1,18 +1,9 @@
...
-      path: support_interface_candidate_path(@application_form.candidate),
-    },
-    {
-      text: @application_form.support_reference,
-      path: support_interface_application_form_path(@application_form),
-    },
-  ].concat(title != 'Details' ? [{ text: title }] : [])) %>
+  <%= govuk_breadcrumbs({
+    'Candidates': support_interface_applications_path,
+    "#{@application_form.candidate.email_address}": support_interface_candidate_path(@application_form.candidate),
```
stevehook, line 4: I think `@application_form.candidate.email_address => support_interface_candidate_path(@application_form.candidate)` is preferred style here (though Rubocop doesn't seem to be bothered either way).

app/views/support_interface/course/_course_navigation.html.erb
```diff
@@ -1,24 +1,12 @@
...
-      text: @course.name,
-      href: support_interface_course_path(@course),
-    },
-  ].concat(title != 'Details' ? [{ text: title }] : [])) %>
+  <%= govuk_breadcrumbs({
+    'Providers': support_interface_providers_path,
+    "#{@course.provider.name}": support_interface_provider_path(@course.provider),
+    'Courses': support_interface_provider_courses_path(@course.provider),
+    "#{@course.name}": support_interface_course_path(@course),
+  }.merge(title != 'Details' ? { "#{title}": nil } : {})) %>
```
stevehook, line 9:
```suggestion
  }.merge(title != 'Details' ? { title: nil } : {})) %>
```

app/views/support_interface/data_exports/_data_export_navigation.html.erb
```diff
@@ -2,20 +2,11 @@
...
-    {
-      text: params[:data_export_type].humanize,
-      path: support_interface_view_export_information_path,
-    },
-  ].concat(title != 'Details' ? [{ text: title }] : [])) %>
+  <%= govuk_breadcrumbs({
+    'Performance': support_interface_performance_path,
+    'Data directory': support_interface_data_directory_path,
+    "#{params[:data_export_type].humanize}": support_interface_view_export_information_path,
+  }.merge(title != 'Details' ? { "#{title}": nil } : {})) %>
```
stevehook, line 9:
```suggestion
  }.merge(title != 'Details' ? { title: nil } : {})) %>
```

app/views/support_interface/ucas_matches/_ucas_match_navigation.html.erb
```diff
@@ -2,20 +2,11 @@
...
-    {
-      text: @match.candidate.email_address,
-      path: support_interface_ucas_match_path(@match),
-    },
-  ].concat(title != 'Details' ? [{ text: title }] : [])) %>
+  <%= govuk_breadcrumbs({
+    'Candidates': support_interface_candidates_path,
+    'UCAS matches': support_interface_ucas_matches_path,
+    "#{@match.candidate.email_address}": support_interface_ucas_match_path(@match),
+  }.merge(title != 'Details' ? { "#{title}": nil } : {})) %>
```
stevehook, line 9:
```suggestion
  }.merge(title != 'Details' ? { title: nil } : {})) %>
```

app/views/support_interface/providers/_provider_navigation.html.erb
```diff
@@ -1,16 +1,10 @@
...
-    },
-    {
-      text: @provider.name,
-      path: support_interface_provider_path(@provider),
-    },
-  ].concat(title != 'Details' ? [{ text: title }] : [])) %>
+  <%= govuk_breadcrumbs({
+    'Providers': support_interface_providers_path(onboarding_stages: %w[synced dsa_signed]),
+    "#{@provider.name}": support_interface_provider_path(@provider),
+  }.merge(title != 'Details' ? { "#{title}": nil } : {})) %>
```
stevehook, line 7:
```suggestion
  }.merge(title != 'Details' ? { title: nil } : {})) %>
```

app/views/support_interface/provider_users/_provider_user_navigation.html.erb
```diff
@@ -1,20 +1,11 @@
...
-    {
-      text: provider_user.display_name,
-      path: support_interface_provider_user_path(provider_user),
-    },
-  ].concat(title != 'Details' ? [{ text: title }] : [])) %>
+  <%= govuk_breadcrumbs({
+    'Providers': support_interface_providers_path,
+    'Provider users': support_interface_provider_users_path,
+    "#{provider_user.display_name}": support_interface_provider_user_path(provider_user),
+  }.merge(title != 'Details' ? { "#{title}": nil } : {})) %>
```
stevehook, line 8:
```suggestion
  }.merge(title != 'Details' ? { title: nil } : {})) %>
```

stevehook (commented):
I think this is great. I've added a few nits inline (sorry) about the hashes.

app/helpers/view_helper.rb
paulrobertlloyd: The helper only aliases the component, so using the same name and passing default options options seems fine to me. Would a comment help?

paulrobertlloyd: Oh wait, we’re doing a bit more than that, as we’re taking `breadcrumbs` value directly. Hmmm.

paulrobertlloyd: Updated name of helper to `breadcrumbs`.

app/views/support_interface/application_forms/_application_navigation.html.erb
paulrobertlloyd: Updated to use preferred style across all breadcrumbs.

stevehook (approved):

PR #4380, opened 2021-03-25
Add Register API with /applications endpoint

AgaDufrat:
## Context

The Register team need to consume data from Apply. They've already carried out a spike https://github.com/DFE-Digital/register-trainee-teachers/pull/566 into consuming data in the form provided by the Vendor API. They're ready to start consuming more of it.

Whilst they've tested against the Vendor API, it's not suitable for their use case, so we need to make them a new one.

See the commits

## Guidance to review

Register team questions:
- are we missing any attributes?
- are there any attributes you don't need and we should remove them?

ProVendor team questions:
- is the recruitment_cycle_year_param validation ok? Should it be a bit more relaxed?
- do we need to be checking if the application is in `recruited`, `withdrawn` or `offer_deferred` states explicitly or is filtering `where.not(recruited_at: nil)` enough to get the relevant ones?

## Link to Trello card

## Things to check


spec/system/register_api/register_receives_application_spec.rb
```diff
@@ -0,0 +1,201 @@
...
+          english_main_language: true,
+          other_languages: nil,
+          english_language_qualifications: nil,
+          disability_disclosure: 'I have difficulty climbing stairs',
+          gender: 'male',
+          disabilities: %w[learning],
+          ethnic_group: 'Asian or Asian British',
+          ethnic_background: 'Asian or Asian British',
+        },
+        course: {
```
defong, line 84: Do we need qualification of the course? @gpeng

app/presenters/register_api/single_application_presenter.rb
```diff
@@ -0,0 +1,245 @@
...
+
+    def provisionally_eligible_for_gov_funding?
+      return true if (PROVISIONALLY_ELIGIBLE_FOR_GOV_FUNDING_COUNTRY_CODES & application_choice.nationalities).any?
+
+      (EU_EEA_SWISS_COUNTRY_CODES & application_choice.nationalities).any? &&
+        application_form.right_to_work_or_study_yes? &&
+        application_form.uk?
+    end
+
+    def course_info_for(course_option)
```
duncanjbrown, line 91: Sorry to come late to this! Does anyone else feel a bit funny about Apply passing on data from TTAPI? This will now be a third copy of the data. Do we risk course data going stale in register — what if the provider made a mistake in the subject codes and updated it in Publish? Or the course name changed? Or if these details changed in Apply, would this application be marked as changed in the Register API? (That might be fiddly). Maybe a snapshot at the point of application is desirable, though? cc @gpeng @AbigailMcP

duncanjbrown:
Looking good so far! I think this would be a good time to add an OpenAPI spec for the request specs (and next, docs) a la https://github.com/DFE-Digital/apply-for-teacher-training/commit/c6f6ecbafd5329b03fb55e8be059f2305153a5a3 (PR shortly)

app/presenters/register_api/single_application_presenter.rb
```diff
@@ -0,0 +1,245 @@
...
+            gender: application_form.equality_and_diversity.try(:[], 'sex'),
+            disabilities: application_form.equality_and_diversity.try(:[], 'disabilities'),
+            ethnic_group: application_form.equality_and_diversity.try(:[], 'ethnic_group'),
+            ethnic_background: application_form.equality_and_diversity.try(:[], 'ethnic_background'),
+          },
+          contact_details: contact_details,
+          course: course_info_for(application_choice.offered_option),
+          provider: provider_info_for(application_choice.offered_option.course.provider),
+          qualifications: qualifications,
+          hesa_itt_data: hesa_itt_data,
```
cwrw, line 46: should we keep some of the logic from the vendor api around setting this value, as it doesn't set it if the application is withdrawn/declined?

app/presenters/register_api/single_application_presenter.rb
```diff
@@ -0,0 +1,245 @@
...
+    def as_json
+      {
+        id: application_choice.id.to_s,
+        type: 'application',
+        attributes: {
+          support_reference: application_form.support_reference,
+          status: status, # keep to show offer withdrawn and deferred
+          updated_at: application_choice.updated_at.iso8601,
+          submitted_at: application_form.submitted_at.iso8601,
+          recruited_at: application_choice.recruited_at.iso8601,
```
cwrw, line 22: good catch adding it here! I reckon we have to fix this in the vendor API as well

app/presenters/register_api/single_application_presenter.rb
```diff
@@ -0,0 +1,245 @@
...
+            domicile: application_form.domicile,
+            uk_residency_status: uk_residency_status,
+            uk_residency_status_code: uk_residency_status_code,
+            fee_payer: provisional_fee_payer_status,
+            english_main_language: application_form.english_main_language,
+            english_language_qualifications: application_form.english_language_qualification_details,
+            other_languages: application_form.other_language_details,
+            disability_disclosure: application_form.disability_disclosure, # do we need this?
+            gender: application_form.equality_and_diversity.try(:[], 'sex'),
+            disabilities: application_form.equality_and_diversity.try(:[], 'disabilities'),
```
cwrw, line 38: should we default to an empty array here if there are no values present, as the values here are an array?

app/presenters/register_api/single_application_presenter.rb
```diff
@@ -0,0 +1,245 @@
...
+          address_line4: application_form.address_line4,
+          postcode: application_form.postcode,
+          country: application_form.country,
+          email: application_form.candidate.email_address,
+        }
+      end
+    end
+
+    def hesa_itt_data
+      equality_and_diversity_data = application_form&.equality_and_diversity
```
cwrw, line 234: we seem to be using `equality_and_diversity` here and for the fields above, but here we aren't protecting against it being `nil`, should we extract it to a method as so and reuse everywhere it occurs instead?
```
def equality_and_diversity
  application_form.equality_and_diversity || {}
end
```

app/presenters/register_api/single_application_presenter.rb
```diff
@@ -0,0 +1,245 @@
...
+    def course_info_for(course_option)
+      {
+        recruitment_cycle_year: course_option.course.recruitment_cycle_year,
+        course_code: course_option.course.code,
+        course_name: course_option.course.name,
+        level: course_option.course.level,
+        subject_codes: course_option.course.subject_codes,
+        program_type: course_option.course.program_type,
+        start_date: course_option.course.start_date.strftime('%Y-%m'),
+        # Register would like end_date but we can't calculate it becuase
```
cwrw, line 100: ✂️ becuase/because

app/presenters/register_api/single_application_presenter.rb
```diff
@@ -0,0 +1,245 @@
...
+        # course_length is a string. Typical values are 'OneYear', 'Two years'
+        # and some free text inputs
```
cwrw, line 102: should we return the `end_date` anyway even if it's a string, or would it be of no use in that format?

app/controllers/register_api/applications_controller.rb
```diff
@@ -2,6 +2,29 @@ module RegisterAPI
...
+
+    def get_recruited_applications(recruitment_cycle_year:)
+      ApplicationChoice.includes([:application_form])
+        .where(application_forms: { recruitment_cycle_year: recruitment_cycle_year }).references(:application_forms)
+        .where(status: ApplicationStateChange::STATES_VISIBLE_TO_REGISTER)
+        .where.not(recruited_at: nil)
+    end
+
+    def recruitment_cycle_year_param
+      year = params.fetch(:recruitment_cycle_year)
```
cwrw, line 22: We've seen in the vendor API today someone passing in a parameter that does not exist, which then throws a `KeyError` and a 500 https://sentry.io/organizations/dfe-bat/issues/2296340364/?project=1765973&referrer=slack
Can we instead handle this gracefully and return a 422 instead?

app/controllers/register_api/applications_controller.rb
```diff
@@ -2,6 +2,29 @@ module RegisterAPI
...
+      application_choices = get_recruited_applications(
+        recruitment_cycle_year: recruitment_cycle_year_param,
+      )
+      render json: { data: MultipleApplicationsPresenter.new(application_choices, api: RegisterAPI).as_json }
+    end
+
+  private
+
+    def get_recruited_applications(recruitment_cycle_year:)
+      ApplicationChoice.includes([:application_form])
```
cwrw, line 15: When running the endpoint locally, this query is yielding n+1s due to further queries in the presenter. I think we might need to tweak it and the include list to ensure its more performant.

app/controllers/register_api/applications_controller.rb
```diff
@@ -2,6 +2,29 @@ module RegisterAPI
...
+        .where(application_forms: { recruitment_cycle_year: recruitment_cycle_year }).references(:application_forms)
+        .where(status: ApplicationStateChange::STATES_VISIBLE_TO_REGISTER)
```
cwrw, line 17: I had a look through the state changes and I don't think we need to search for the status, as the `recruited_at` is only unset if we `reinstate_pending_conditions`, but will be set on all the other states. @duncanjbrown can you confirm?

cwrw (commented):
Looking great 🙌  ! left a few comments and nits.

Since we aren't paginating the results, we might need to filter further by provider as well as recruitment year to avoid the performance repercussions in production. or maybe introduce pagination, what are people's thoughts. 

Also I was having a think around the duplication in the single presenter between the vendor/register API. Does not have to be in this iteration and can come later, but something to keep in mind around how we can extract the standalone shared methods between the two.

app/controllers/register_api/applications_controller.rb
```diff
@@ -2,6 +2,29 @@ module RegisterAPI
...
+        .where.not(recruited_at: nil)
+    end
+
+    def recruitment_cycle_year_param
+      year = params.fetch(:recruitment_cycle_year)
+      raise ParameterInvalid, 'Parameter is invalid: recruitment_cycle_year' unless year.in?(RecruitmentCycle.years_visible_to_providers.map(&:to_s))
+
+      year
+    rescue ArgumentError
```
cwrw, line 26: I don't think we'll get an `ArgumentError` in this instance as we aren't doing any parsing on this param, so maybe we can remove the `rescue` section.

cwrw:
After a discussion with @AbigailMcP @gpeng @duncanjbrown: we're going to add a new optional param `changed_since` which will help with incrementally pulling in applications. I'll add the changes shortly

app/presenters/register_api/single_application_presenter.rb
duncanjbrown: - How would it be if we cut all fields that weren't necessary for identifying the course option? (ie leave only `recruitment_cycle_year`, `course_code`, `provider_code`, `site_code` and `study_mode`). Register can collect the rest, including `accredited_body`, from the TTAPI anyway. 
- Should we call the `provider_code` `training_provider_code` to be unambiguous?

app/presenters/register_api/single_application_presenter.rb
```diff
@@ -0,0 +1,247 @@
...
+            other_languages: application_form.other_language_details,
+            disability_disclosure: application_form.disability_disclosure, # do we need this?
+            gender: equality_and_diversity_data['sex'],
+            disabilities: equality_and_diversity_data['disabilities'].presence || [],
+            ethnic_group: equality_and_diversity_data['ethnic_group'],
+            ethnic_background: equality_and_diversity_data['ethnic_background'],
+          },
+          contact_details: contact_details,
+          course: course_info_for(application_choice.offered_option),
+          provider: provider_info_for(application_choice.offered_option.course.provider),
```
duncanjbrown, line 44: Can we dispose of this altogether? Register can know the Provider from the Course, and TTAPI is a better source of truth than we are

app/presenters/register_api/single_application_presenter.rb
```diff
@@ -0,0 +1,247 @@
...
+          qualifications: qualifications,
+          hesa_itt_data: hesa_itt_data.presence || {},
+        },
+      }
+    end
+
+  private
```
duncanjbrown, line 51: Maybe when it's time to refactor (soon!) we could push some of these methods up into the `Decorator` and share that? Feels a little bit more testable too.

spec/queries/get_recruited_application_choices_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+    )
+
+    recruited_application = create(
+      :application_choice,
+      :with_recruited,
+      application_form: build(:application_form, recruitment_cycle_year: '2021'),
+    )
+
+    application_choices = described_class.call(recruitment_cycle_year: '2021')
+    expect(application_choices).to contain_exactly(recruited_application)
```
duncanjbrown, line 20: `contain_exactly`!

spec/requests/register_api/get_multiple_applications_spec.rb
```diff
@@ -0,0 +1,52 @@
+require 'rails_helper'
+
+RSpec.describe 'GET /register-api/applications', type: :request do
```
duncanjbrown, line 3: Nice :+1:

duncanjbrown (commented):
Few more comments!

app/presenters/multiple_applications_presenter.rb
```diff
@@ -0,0 +1,14 @@
...
+  attr_reader :application_choices, :api
+
+  def initialize(application_choices, api:)
+    @application_choices = application_choices
+    @api = api
+  end
+
+  def as_json
+    application_choices.map do |application_choice|
+      api::SingleApplicationPresenter.new(application_choice).as_json
```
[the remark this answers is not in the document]
cwrw: done

app/presenters/register_api/single_application_presenter.rb
```diff
@@ -0,0 +1,247 @@
+module RegisterAPI
+  class SingleApplicationPresenter
+    UCAS_FEE_PAYER_CODES = {
+      'SLC,SAAS,NIBd,EU,Chl,IoM' => '02',
+      'Not Known' => '99',
+    }.freeze
+
+    def initialize(application_choice)
+      @application_choice = ApplicationChoiceExportDecorator.new(application_choice)
```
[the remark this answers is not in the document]
cwrw: There are a few decorators named incorrectly, I think we can rename those in the cleanup PR to follow shortly, as I think `ApplicationChoiceHesaExportDecorator` should also be renamed.

app/presenters/register_api/single_application_presenter.rb
cwrw: Yes, many of those are standalone and can be tested in isolation in a shared class for both APIs, in cleanup PR to follow this.

app/presenters/register_api/single_application_presenter.rb
cwrw: done

app/presenters/multiple_applications_presenter.rb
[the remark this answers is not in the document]
duncanjbrown: Thanks!

app/presenters/register_api/single_application_presenter.rb
duncanjbrown: 👌

PR #4381, opened 2021-03-25
Remove mimemagic

duncanjbrown:
We don't need this dependency and it has licensing issues: https://ukgovernmentdfe.slack.com/archives/CAGBHB4JV/p1616668662022200

Rationale for `rails` dep changes here: https://github.com/rails/rails/issues/41750#issuecomment-805913270
Relevant `govuk-component` changes: https://github.com/DFE-Digital/govuk-components/pull/124


db/schema.rb
```diff
@@ -10,54 +10,142 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_03_19_163050) do
+ActiveRecord::Schema.define(version: 2021_03_22_215908) do
```
malcolmbaig, line 14: Some schema changes have snuck in

duncanjbrown: oops! fixed

malcolmbaig (dismissed):

PR #4382, opened 2021-03-25
Change application qualifications HESA codes to strings

steventux:
## Context

API docs for `ApplicationQualification` HESA codes specify the type to be `string` but the db column type is `bigint`.
We should be consistent with this and migrate to more flexible string columns.

Migrates column types for `qualification_type_hesa_code`, `subject_hesa_code`, `institution_hesa_code` and `grade_hesa_code` to `string`
Updates the config for hesa codes to strings for consistency.

## Guidance to review

I'm unsure whether updating the config is the correct approach but it seemed less intuitive to have a config using integers while we write strings to the db. This seemed the most comprehensive way to migrate these codes.

The migration and code changes are combined in this PR because the migration alone will break tests.

## Link to Trello card

https://trello.com/c/gBw5KJfm/3503-change-application-qualifications-hesa-codes-to-strings

## Things to check


richardpattinson (approved):
Looks great to me, lots of ''.

TobyRet (approved):
LGTM!

PR #4383, opened 2021-03-25
Refactor queries where multiple args are passed into where.not

davidgisbey:
## Context

This functionality is being deprecated in Rails 6.1 and needs to be fixed before we upgrade.

It also stops the deprecation spam we're getting when we run the test suite.

- Refactor `where.not` queries with multiple args to only pass in one argument

## Things to check


despo (approved):
Thanks for sorting this out. 

:shipit:

PR #4386, opened 2021-03-25
Decrease chances of duplicate emails in test data

benswannack:
## Context
There are higher volumes of test applications being added to sandbox, and the email addresses we generate aren't complicated enough to avoid duplicates

Append a number to test email addresses to prevent making duplicate emails

## Guidance to review
Hopefully simple

## Link to Trello card
https://trello.com/c/zS5hJ8CB/3495-error-when-generating-test-data-for-existing-candidate-in-sandbox-via-api

## Things to check

- [x]  This code does not rely on the addition/removal of Azure config environment variables in the same Pull Request

### DIFF
--- a/app/lib/test_applications.rb
+++ b/app/lib/test_applications.rb
@@ -116,7 +116,7 @@ private
       previous_application_form = nil
       candidate = candidate.presence || FactoryBot.create(
         :candidate,
-        email_address: "#{first_name.downcase}.#{last_name.downcase}@example.com",
+        email_address: "#{first_name.downcase}.#{last_name.downcase}#{rand(10000)}@example.com",
         created_at: time,
         last_signed_in_at: time,
       )


app/lib/test_applications.rb
```diff
@@ -116,7 +116,7 @@ def create_application_to_courses(
       previous_application_form = nil
       candidate = candidate.presence || FactoryBot.create(
         :candidate,
-        email_address: "#{first_name.downcase}.#{last_name.downcase}@example.com",
+        email_address: "#{first_name.downcase}.#{last_name.downcase}#{rand(10000)}@example.com",
```
despo, line 119: Is it not best to use [sequences](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#sequences) for this?

despo: Good point. It doesnt go into as much detail in the documentation but as this is not the intention of the gem it probably won't work in this scenario.

despo (approved):

cwrw (approved):

PR #4388, opened 2021-03-25
Unconditional Offers can't be submitted through Manage

richardpattinson:
Update service to cater for empty condition array, this ensures support for unconditional offers and enables use of the service to validate data. 

## Link to Trello card

https://trello.com/c/h4CaMKp2/3532-unconditional-offers-cant-be-submitted-through-manage%F0%9F%91%80

## Things to check

### DIFF
--- a/app/controllers/provider_interface/decisions_controller.rb
+++ b/app/controllers/provider_interface/decisions_controller.rb
@@ -84,7 +84,7 @@ module ProviderInterface
         actor: current_provider_user,
         application_choice: @application_choice,
         course_option: course_option,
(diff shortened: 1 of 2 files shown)


app/services/make_an_offer.rb
```diff
@@ -62,6 +62,7 @@ def save
   end
 
   def offer_conditions
+
```
cwrw, line 65: 🔥

cwrw (approved):
Looks good, just rubocop 👍 

I was wondering around doing a backfill but we can't really as we erroneously saved the offers so can't really know which conditions were deselected anyway

PR #4389, opened 2021-03-25
Account for KeyError being raised by ActiveSupport::TimeZone

steventux:
## Context

It's possible to supply a value which is parsed internally by `ActiveSupport::TimeZone` (by `Date._iso8601`) but doesn't contain the expected fields.
A `KeyError` is raised in this case, which is possibly due to over-optimistic expectations in the `ActiveSupport::TimeZone.iso8601` method or a bug in the underlying call to `Date._iso8601`.

I've also raised an issue [here](https://github.com/rails/rails/issues/41763) arguably this is a problem with the ruby `Date._iso8601` method but may be of interest to the rails maintainers.

This commit rescues from this type of error to gracefully handle the invalid parameter format.

## Guidance to review

Details of Sentry exception https://sentry.io/organizations/dfe-bat/issues/2296340364/?project=1765973&referrer=slack

## Link to Trello card

https://trello.com/c/Oy3lRIEA/3539-gracefully-handle-api-parameters-we-do-not-support

## Things to check

### DIFF
--- a/app/controllers/vendor_api/applications_controller.rb
+++ b/app/controllers/vendor_api/applications_controller.rb
@@ -35,12 +35,16 @@ module VendorAPI
     end
 
     def since_param
-      since = Time.zone.iso8601(params.fetch(:since))
-      raise ParameterInvalid, 'Parameter is invalid (date is nonsense): since' unless since.year.positive?
+      since_value = params.fetch(:since)
 
-      since
-    rescue ArgumentError
-      raise ParameterInvalid, 'Parameter is invalid (should be ISO8601): since'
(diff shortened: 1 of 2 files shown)


cwrw (approved):
I couldn't have gotten the explanation for this error more wrong earlier today! Excellent investigation into the [iso8601](https://github.com/rails/rails/blob/5aaaa1630ae9a71b3c3ecc4dc46074d678c08d67/activesupport/lib/active_support/values/time_zone.rb#L390) method, and thanks for raising the issue as well! 👍

PR #4390, opened 2021-03-26
Only retrieve study modes of course options with vacancies

despo:
Resolves sentry error caused by candidate ending up on location selection page with no locations rendered due to the selected study mode not having any sites with availability.

https://sentry.io/organizations/dfe-bat/issues/2297226890/?project=1765973&referrer=slack

A new method is introduced as the existing `available_study_mode_from_options` is used on Make Offer, on the provider side, where availability should not be taken into account.

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -85,7 +85,7 @@ class Course < ApplicationRecord
   end
 
   def currently_has_both_study_modes_available?
-    available_study_modes_from_options.count == 2
+    available_study_modes_with_vacancies.count == 2
   end
 
   def supports_study_mode?(mode)
@@ -94,6 +94,10 @@ class Course < ApplicationRecord
 
   def available_study_modes_from_options
     course_options.select(&:site_still_valid).pluck(:study_mode).uniq
+  end
+
+  def available_study_modes_with_vacancies
+    course_options.available.pluck(:study_mode).uniq
   end
 
   def full?
--- a/spec/models/course_spec.rb
+++ b/spec/models/course_spec.rb
@@ -9,22 +9,20 @@ RSpec.describe Course, type: :model do
   end
 
   describe '#currently_has_both_study_modes_available?' do
+    let(:course) { build(:course) }
+
     it 'is true when a course has full time and part time course options' do
-      course_option1 = build_stubbed(:course_option, :full_time)
-      course_option2 = build_stubbed(:course_option, :part_time)
-      course = build_stubbed(:course, course_options: [course_option1, course_option2])
+      create(:course_option, :full_time, course: course)
+      create(:course_option, :part_time, course: course)
 
       expect(course.currently_has_both_study_modes_available?).to be true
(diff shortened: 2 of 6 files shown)


spec/models/course_spec.rb
```diff
@@ -47,6 +45,17 @@
     end
   end
 
+  describe '#available_study_modes_with_vacancies' do
+    let(:course) { build(:course) }
+
+    it 'returns an array of unique study modes for course options with available vacancies' do
```
cwrw, line 51: can you add a spec for the `site_still_valid` option as well?

cwrw (approved):
Looks good 👍  Might be worth getting someone from the candidate side as well to confirm

stevehook (commented):
I think this change makes sense.

I think I found another variation on the original bug though when testing locally. I've reproduced on the review app by changing the availability of _Blackfriars Teaching School Alliance (1JA)_ part-time course option for _English and Communication (SEND) (V252)_ to `no_vacancies` and try to add that course. I end up getting redirected to the empty site selection page.

image

Pressing continue triggers a 500.

I think part of the problem here is that we are in a 'single-site' situation here but https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/controllers/candidate_interface/course_choices/course_selection_controller.rb#L50 doesn't work because there are multiple options (due to different study modes).

despo:
@stevehook thanks for pointing this out. Looking more into it I think all course option retrievals in the `pick_*` forms should be updated to ensure that they satisfy the conditions specified in `available_study_modes_with_vacancies`.

app/models/course.rb
```diff
@@ -96,6 +96,10 @@ def available_study_modes_from_options
     course_options.select(&:site_still_valid).pluck(:study_mode).uniq
   end
 
+  def available_study_modes_with_vacancies
+    course_options.available.selectable.pluck(:study_mode).uniq
```
malcolmbaig, line 100: We can just do `course_options.available`, as the `available` scope already uses `selectable`.

despo: Thanks for pointing out. Will have a look in a bit.

malcolmbaig:
> @stevehook thanks for pointing this out. Looking more into it I think all course option retrievals in the `pick_*` forms should be updated to ensure that they satisfy the conditions specified in `available_study_modes_with_vacancies`.

💯 — I think there are two spots where we need additional `available` scopes:

https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/forms/candidate_interface/pick_course_form.rb#L48
https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/forms/candidate_interface/pick_study_mode_form.rb#L9

despo:
@stevehook retested with the latest updates using the details you provided and things seem to work now.

stevehook (approved):
Re-tested review app and all works now. Nice fix! 💯

PR #4394, opened 2021-03-26
Display DBD information on application page

benswannack:
## Context
Providers find it useful to see how long candidates have to response to their offer, so they know if they should chase them.

Add to the application choice header component to display text about the application's DBD date.

Small refactor to the component to make it more readable.

## Guidance to review
Per commit to block out refactor noise

## Link to Trello card
https://trello.com/c/x9vTYttQ/3510-display-when-an-offer-will-be-automatically-declined

## Things to check

### DIFF
--- a/app/helpers/view_helper.rb
+++ b/app/helpers/view_helper.rb
@@ -122,6 +122,17 @@ module ViewHelper
     end
   end
 
+  def date_and_time_today_or_tomorrow(time)
+    unless time_is_today_or_tomorrow?(time)
+      raise "#{time} was expected to be today or tomorrow, but is not"
+    end
+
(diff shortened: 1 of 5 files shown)


app/helpers/view_helper.rb
```diff
@@ -122,6 +122,17 @@ def time_today_or_tomorrow(time)
     end
   end
 
+  def date_and_time_today_or_tomorrow(time)
```
benswannack, line 125: From the prototype, we will be moving towards this over the above `time_today_or_tomorrow` method in other places too

cwrw (approved):
👍  need to add a ticket for the make a decision part, will discuss with James tomorrow

PR #4398, opened 2021-03-28
Document the Data API and support retrieving single reports over it

duncanjbrown:
## Context

The TAD API is currently documented in the same way the data exports are, ie a list of fields.

This is a proper API and there's no reason we can't document it using an OpenAPI spec and our standard documentation templates.

This PR also adds two endpoints requested by TAD colleagues to support a use case where they aren't able to download a report on the day it was generated, due to holidays etc. (The existing `/latest` endpoint doesn’t support that).

- Turn the Data API into an OpenAPI spec, drawing parts of its definition from the definition of the CSV in support
- Reuse the documentation classes we already have (`APIDocs::*`) to automatically generate docs for it
- Enable those docs to render example CSV responses (we serve exports as CSV)
- Use the new tooling to add and validate the new endpoints along with documentation

## Guidance to review

Consider the above 4 bullet points as three separate but interdependent pieces of work — each one should be complete and should integrate properly with the others.

## Link to Trello card

https://trello.com/c/BYb4den5/3544-enable-tad-to-download-past-exports

## Things to check


app/controllers/data_api/tad_data_exports_controller.rb
```diff
@@ -2,10 +2,58 @@ module DataAPI
   class TADDataExportsController < ActionController::API
     include ServiceAPIUserAuthentication
 
+    def index
+      exports = DataAPI::TADExport.all
+
+      formatted_exports = exports.map do |e|
```
cwrw, line 8: nit: can we use `export` instead of `e`?

app/controllers/data_api/tad_data_exports_controller.rb
```diff
@@ -2,10 +2,58 @@ module DataAPI
...
+    end
+
+  private
+
+    def serve_export(export)
+      export.update!(audit_comment: "File downloaded via API using token ID #{@authenticating_token.id}")
+      send_data export.data, filename: export.filename
+    end
+
+    def verify_token!
```
cwrw, line 36: this can be removed as it's part of the `ServiceAPIUserAuthentication` module which is included in the controller

spec/support/data_api/shared_examples.rb
```diff
@@ -0,0 +1,7 @@
+RSpec.shared_examples 'a TAD API endpoint' do |path|
+  it 'verifies the API token' do
```
cwrw, line 2: I think the logic here is now deprecated: https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/spec/requests/data_api/tad_api_spec.rb as some of it is covered by the above specs. Maybe we can move the other authorisation specs for other API users etc. into here and remove the old spec

spec/support/data_api/data_api_spec_helper.rb
```diff
@@ -0,0 +1,23 @@
+module DataAPISpecHelper
+  def get_api_request(url, token:, options: {})
```
cwrw, line 2: Since this logic will be used in the register API, is it possible to move this to a more generic helper so we can accommodate both?

cwrw (commented):
Looks good, some minor comments. Could you add details and an example for the attributes missing those example:`course_level` and `nctl_subject ` in the API docs

spec/support/data_api/data_api_spec_helper.rb
duncanjbrown: Yep - if possible let's make it generic. I'll leave that to the Register PR.

app/controllers/data_api/tad_data_exports_controller.rb
duncanjbrown: Done

app/controllers/data_api/tad_data_exports_controller.rb
duncanjbrown: Good call, thank you

spec/support/data_api/shared_examples.rb
duncanjbrown: Good shout. I've removed the `tad_api_spec.rb` file. Let's look at refactoring post-Register?

duncanjbrown:
Thanks @cwrw — updated!

cwrw (approved):
👍  thanks for addressing

config/routes.rb
```diff
@@ -992,6 +993,11 @@
       get '/help' => 'pages#help', as: :help
       get '/spec.yml' => 'openapi#spec', as: :spec
     end
+
+    namespace :data_api_docs, path: '/data-api' do
```
AgaDufrat, line 997: Nothing blocking (as I need to rebase against it 😂) but I was wondering if it'd be good to include "docs" in the path to keep it consistent within this namespace
```suggestion
    namespace :data_api_docs, path: '/data-api-docs' do
```

duncanjbrown: Not sure — merging, can revisit later!

PR #4403, opened 2021-03-29
Fix validation errors on site select page

malcolmbaig:
We're missing an instance variable when re-rendering the form after a
validation error. Use a helper method to retrieve available sites
instead, so that both actions in SiteSelectionController have access to
this data.

## Things to check


despo (approved):
Looks good. Maybe consider whitelisting the parameters?

I think rubocop is going to complain about the `private` indendation..

PR #4405, opened 2021-03-29
Alert when an application is submitted with more than three course choices

JR-G:
## Context

We currently don't receive an alert if an application has been submitted with more than three course choices. This isn't something we allow so need to be aware if this state ever occurs.

Added a new check to the DataInvariant worker to alert if this happens

## Guidance to review

Are there any scenarios I've not considered?

## Link to Trello card

https://trello.com/c/etQENcy9/3198-data-check-alert-when-an-application-is-submitted-with-more-than-3-courses

## Things to check

### DIFF
--- a/spec/workers/detect_invariants_spec.rb
+++ b/spec/workers/detect_invariants_spec.rb
@@ -107,5 +107,23 @@ RSpec.describe DetectInvariants do
         ),
       )
     end
+
+    it 'detects submitted applications with more than three course choices' do
+      create(:completed_application_form, submitted_application_choices_count: 3)
+      bad_application_form = create(:completed_application_form, submitted_application_choices_count: 4)
+
+      DetectInvariants.new.perform
+
+      expect(Raven).to have_received(:capture_exception).once
+      expect(Raven).to have_received(:capture_exception).with(
+        DetectInvariants::SubmittedApplicationHasMoreThanThreeChoices.new(
+          <<~MSG,
+            The following application forms have been submitted with more than three course choices
+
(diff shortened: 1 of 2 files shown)


spec/workers/detect_invariants_spec.rb
```diff
@@ -107,5 +107,21 @@
         ),
       )
     end
+
+    it 'detects submitted applications with more than three course choices' do
+      application_form = create(:completed_application_form, submitted_application_choices_count: 4)
```
malcolmbaig, line 112: Suggestion: if we create another application form with 3 choices, we can be more certain that the expectation below is testing the behaviour we're after (rather than, eg—simply returning all application forms).

spec/workers/detect_invariants_spec.rb
```diff
@@ -107,5 +107,32 @@
...
+        DetectInvariants::SubmittedApplicationHasMoreThanThreeChoices.new(
+          <<~MSG,
+            The following application forms have been submitted with more than three course choices
+
+            http://localhost:3000/support/applications/#{bad_application_form.id}
+          MSG
+        ),
+      )
+
+      expect(Raven).not_to have_received(:capture_exception).with(
```
malcolmbaig, line 127: Nit: could express more or less the same idea (ie—"Raven doesn't receive anything related to the good_application") by replacing this second expectation with:

`expect(Raven).to have_received(:capture_exception).once`

malcolmbaig (approved):
👍🏼

PR #4407, opened 2021-03-29
Update OffersController to use new ChangeOffer service wrapper

despo:
## Context

Integrate new ChangeOffer service wrapper to new edit offer flow.

## Link to Trello card

https://trello.com/c/skeX4L9n/3546-integrate-changeoffer-wrapper-to-new-offer-flow

## Things to check

### DIFF
--- a/app/controllers/provider_interface/offers_controller.rb
+++ b/app/controllers/provider_interface/offers_controller.rb
@@ -49,10 +49,10 @@ module ProviderInterface
     def update
       @wizard = OfferWizard.new(offer_store)
       if @wizard.valid?(:save)
-        ChangeAnOffer.new(actor: current_provider_user,
+        ::ChangeOffer.new(actor: current_provider_user,
                           application_choice: @application_choice,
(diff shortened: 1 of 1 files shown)


app/controllers/provider_interface/offers_controller.rb
```diff
@@ -49,10 +49,10 @@ def show
     def update
       @wizard = OfferWizard.new(offer_store)
       if @wizard.valid?(:save)
-        ChangeAnOffer.new(actor: current_provider_user,
-                          application_choice: @application_choice,
-                          course_option: @wizard.course_option,
-                          offer_conditions: @wizard.conditions).save
+        ::ChangeOffer.new(actor: current_provider_user,
```
cwrw, line 52: should we move both those under the `app/services/provider_interface`?

despo: We decided to move away from the namespace, I think because a lot of the services are reused accross support and provider.

cwrw (approved):

PR #4408, opened 2021-03-29
Fix flaky ChangeOffer spec

mnacos:
## Context

It's flaky...

Fix it by moving a few things around e.g. creating the provider_user/provider first.

## Guidance to review

Run the tests several times on CI

## Link to Trello card

No card

### DIFF
--- a/spec/services/change_offer_spec.rb
+++ b/spec/services/change_offer_spec.rb
@@ -7,33 +7,38 @@ RSpec.describe ChangeOffer do
     ChangeOffer.new(actor: provider_user,
                     application_choice: application_choice,
                     course_option: course_option,
-                    conditions: conditions)
+                    conditions: new_conditions)
   end
 
   describe '#save!' do
-    let(:application_choice) { create(:application_choice, :with_offer) }
+    let(:provider_user) do
+      create(:provider_user,
+             :with_make_decisions,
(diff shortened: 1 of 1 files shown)


spec/services/change_offer_spec.rb
```diff
@@ -7,33 +7,38 @@
...
+      create(:provider_user,
+             :with_make_decisions,
+             providers: [application_choice.offered_option.provider])
+    end
     let(:course_option) { course_option_for_provider(provider: application_choice.course_option.provider) }
-    let(:provider_user) { create(:provider_user, providers: [build(:provider)]) }
-    let(:conditions) { [Faker::Lorem.sentence] }
+    let(:new_conditions) { [Faker::Lorem.sentence] }
+    let(:application_choice) do
+      create(:application_choice, :with_offer, offer: { 'conditions' => ['DBS check'] })
```
despo, line 22: The `:with_offer` scope already sets up some conditions, so the additional application choice setup is not necessary

https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/spec/factories/application_choice.rb#L155

mnacos: We're overriding the conditions set by the factory to support L41, in order to produce an identical offer.

SamJamCul (approved):

PR #4410, opened 2021-03-29
Fix flaky spec

despo:
- The issue is caused by the application choice status being set to`interviewing`. As that status is only available through `ApplicationStateChange.states_visible_to_provider` when the `interviews` flag is active, it’s causing GetApplicationChoiceForProvider to not use interviewing as one of the available statuses, and as a result not return the application choice, causing ProviderAuthentication to fail.

A more general fix can also be made to the factory, however, as to change an offer we should be in an offer state, given the state does not change, in this instance we should correctly setup the expected data.

### DIFF
--- a/spec/services/change_offer_spec.rb
+++ b/spec/services/change_offer_spec.rb
@@ -11,7 +11,7 @@ RSpec.describe ChangeOffer do
   end
 
   describe '#save!' do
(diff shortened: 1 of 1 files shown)


cwrw (approved):
👏

PR #4411, opened 2021-03-29
Return correct set of valid application states

despo:
This fix should resolve random spec failures caused by ProviderAuthorisation issues.

### Root cause
The issue is caused by the `application_choice` factory using a [random value from`valid_states`](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/spec/factories/application_choice.rb#L6) to set the application choice status.

### How?
As the [GetApplicationChoicesForProvider service](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/queries/get_application_choices_for_providers.rb#L19) is queried from the [ProviderAuthorisation](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/services/provider_authorisation.rb#L176) service in order for the application's visibility to be established, it sometimes returns false due to  the`interviewing` state [only being available when the `interview` flag is active](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/services/application_state_change.rb#L110).

### Fix
Only available statuses are exposed through `valid_states`

### DIFF
--- a/app/services/application_state_change.rb
+++ b/app/services/application_state_change.rb
@@ -104,7 +104,9 @@ class ApplicationStateChange
   end
 
   def self.valid_states
-    workflow_spec.states.keys
+    return workflow_spec.states.keys if FeatureFlag.active?(:interviews)
+
+    workflow_spec.states.keys - [:interviewing]
   end
 
   def self.states_visible_to_provider
--- a/spec/services/application_state_change_spec.rb
+++ b/spec/services/application_state_change_spec.rb
@@ -1,28 +1,30 @@
 require 'rails_helper'
 
 RSpec.describe ApplicationStateChange do
+  before do
+    FeatureFlag.deactivate(:interviews)
+  end
+
   describe '.valid_states' do
     it 'has human readable translations' do
       expect(ApplicationStateChange.valid_states)
-        .to match_array(I18n.t('application_states').keys)
+        .to match_array(I18n.t('application_states').keys - %i[interviewing])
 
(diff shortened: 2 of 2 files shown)


malcolmbaig (approved):

cwrw (approved):
can you add a test for when the feature flag is on just for posterity

despo:
> can you add a test for when the feature flag is on just for posterity

Done!

♻️

config/vendor-api-v1.yml
```diff
@@ -429,6 +429,7 @@ components:
             lifecycle diagram](/api-docs/lifecycle) for states and transitions.
           enum:
           - awaiting_provider_decision
+          - interviewing
```
duncanjbrown, line 432: We can't add this as it's a breaking change

despo: Why is that? Do we need to update the description then, as the lifecycle diagram does render the interviewing state (which is why I made the change here)?

duncanjbrown: SRSs won't know what to do with this state if they receive it over the API — obv we haven't started sending it here but it could still cause alarm for those with integrations already live or in late stages of development. We should probably tackle the lifecycle diagram. I think we can knock states out of it fairly easily.

despo: @duncanjbrown ok updated the PR. We can add a different card for the lifecycle diagram.

cwrw (approved):

duncanjbrown (approved):

PR #4413, opened 2021-03-30
Small fixes to the API docs

duncanjbrown:
## Context

@frankieroberto and @defong pointed out some deficiencies in the API docs.

https://ukgovernmentdfe.slack.com/archives/CGVNJ2Q2F/p1617027395198300
https://ukgovernmentdfe.slack.com/archives/CPH8J9G65/p1617008885212300

- always include a number > 13 in dates so it's clear which field is the "day"
- always format dates as ISO8601 (its necessary to quote them to prevent the OpenAPI parser turning them into `1 Jan 2021` — which seems a slightly weird thing to do?)
- render missing qualification grades as "Not entered" (we used to pass `nil`, which wasn't correct)
- remove misleading examples on deprecated fields

## Link to Trello card

N/A

## Things to check


defong:
Looking nice

config/vendor-api-v1.yml
```diff
@@ -1077,13 +1076,13 @@ components:
           type: string
           format: date
           description: The date the position began
-          example: 2020-09-10
+          example: "2020-09-13"
```
frankieroberto, line 1079: Suggest making this the first of the month, as it always will be, because we don't collect date.

```suggestion
          example: "2020-09-01"
```

config/vendor-api-v1.yml
```diff
@@ -1077,13 +1076,13 @@ components:
...
         end_date:
           type: string
           format: date
           nullable: true
           description: The date the position finished, if applicable
-          example: 2019-04-01
+          example: "2019-04-13"
```
frankieroberto, line 1085:
```suggestion
          example: "2019-04-01"
```

frankieroberto (approved):

PR #4414, opened 2021-03-30
Offer flow - back button error

despo:

When no step is specified default to the step before the last.

## Guidance to review

1. Make a new offer or edit an existing offer
2. On the conditions page, enter a condition that exceeds 255 characters
3. Press continue

## Link to Trello card

https://trello.com/c/l3VKxiDF/3576-offer-flow-back-button-error

## Things to check

### DIFF
--- a/app/services/wizard_path_history.rb
+++ b/app/services/wizard_path_history.rb
@@ -1,4 +1,6 @@
 class WizardPathHistory
+  PREVIOUS_STEP_INDEX = -2
+
   class NoSuchStepError < StandardError
     def message
       'Invalid wizard step'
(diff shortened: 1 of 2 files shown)


app/services/wizard_path_history.rb
```diff
@@ -22,6 +24,7 @@ def update
   end
 
   def previous_step
+    return path_history[PREVIOUS_STEP_INDEX] if step.nil?
```
benswannack, line 27: I feel like it would be clearer to specify the step in the controller. Is there a reason we shouldn't do that?

despo: Yes. Adding a step multiple times will also end up taking the user back to that step multiple times when pressing back, which is what I want to avoid.

benswannack (approved):
LGTM 👍

PR #4415, opened 2021-03-30
Minor course choice updates

malcolmbaig:
## Context

A small bugfix and the addition of a spec, identified during recent work on the course choices flow.

**Add spec for PickSiteForm.available_sites method**

**Add `available` scope to CourseChoicesReviewComponent**
Check for available course options when deciding whether or not to
display a 'Change' link for location. This mirrors how we only show the
site selection step of the form if there's more than one available
choice.

## Guidance to review
Small enough to just check the whole diff

## Link to Trello card
na

## Things to check


despo (approved):
👍🏼

PR #4416, opened 2021-03-30
Refactor conditions part of make offer

benswannack:
## Context
We are planning to allow providers to add and remove conditions boxes from the form. In its current state it will be difficult to do that.

Make the number of condition boxes the only dependency for showing and submitting conditions of an offer.

## Guidance to review
Is this approach the best?

It feels like a battle against rails to get this to work as intended.

## Trello
https://trello.com/c/AMcA4XK8/3514-spike-conditions-of-offer

### DIFF
--- a/app/forms/provider_interface/offer_wizard.rb
+++ b/app/forms/provider_interface/offer_wizard.rb
@@ -6,27 +6,26 @@ module ProviderInterface
               change_offer: %i[select_option providers courses study_modes locations conditions check] }.freeze
 
     attr_accessor :provider_id, :course_id, :course_option_id, :study_mode, :location_id,
-                  :standard_conditions, :further_condition_1, :further_condition_2,
-                  :further_condition_3, :further_condition_4, :current_step, :decision,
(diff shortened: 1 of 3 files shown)


app/forms/provider_interface/offer_wizard.rb
```diff
@@ -116,6 +123,19 @@ def state
...
     end
 
+    def setup_further_conditions(attrs)
+      return unless attrs['further_conditions']
+
+      attrs['further_conditions'].each do |index, value|
+        singleton_class.class_eval do
+          attr_accessor "further_conditions_#{index}"
+          alias :"further_conditions[#{index}]" :"further_conditions_#{index}"
+          alias :"further_conditions[#{index}]=" :"further_conditions_#{index}="
```
benswannack, line 133: The ideal here would be to have an attr_accessor for `further_conditions[index]`, but that isn't allowed because `[]` aren't allowed in variable names.

It's required to let the model populate values in the form, and to correctly link up the errors

benswannack:
Steve is going to look into a different approach for this, any comments on this in the meantime are still useful though!

PR #4418, opened 2021-03-30
Change the way a candidate's 'current application' is determined

TobyRet:
## Context

We've noticed some tricky bugs occur in code that relies on the `current_application` method in the Candidate model (see https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/models/candidate.rb#L26)

It's possible that `application_forms.last` may not always return what is thought to be the latest application form if prior to this, the application_forms  array has been mutated in some way.

Instead the safer option would be to order by created date 
 `application_forms.order(:created_at).last` 

## Replicating the bug

In the console find a candidate with multiple application forms (or create them)

```
# previously the '#current_application' returns the application form with the largest id (or creates and application form and returns it)

def current_application
  application_form = application_forms.last
  ...
end


# At this point the current_application id is as expected

candidate.current_application.id = x

# Somewhere else in the code, we try to determine a candidate's previous applications (as an array). Note how the 'current_application' is subtracted from all applications

previous_applications = candidate.application_forms - [ candidate.current_application ]

# But now the current_application id is  x - 1
candidate.current_application.id = x

```

- Determine the current application by sorting all application forms by created date and return the most recent application form
- The above change has impacted pre-filling of blank applications (in Sandbox). Blank applications are destroyed in order to make [the system test](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/spec/system/candidate_interface/signup_and_signin/candidate_signs_in_and_prefills_application_in_sandbox_spec.rb) green. See https://github.com/DFE-Digital/apply-for-teacher-training/pull/4418/commits/41d60d0ffd128c8d43449e25aa829b3740ae9a50 for context

## Link to Trello card

https://trello.com/c/tOE6GT5l/3247-change-the-way-a-candidates-current-application-is-determined

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/prefill_application_form_controller.rb
+++ b/app/controllers/candidate_interface/prefill_application_form_controller.rb
@@ -28,6 +28,9 @@ module CandidateInterface
         courses_to_apply_to: Course.current_cycle.open_on_apply.joins(:course_options).merge(CourseOption.available),
         candidate: current_candidate,
       )
+
+      destroy_blank_application
+
       example_application_form = example_application_choices.first.application_form
       current_candidate.application_forms << example_application_form
     end
@@ -36,6 +39,11 @@ module CandidateInterface
 
     def prefill_application_or_not_params
       params.fetch(:candidate_interface_prefill_application_or_not_form, {}).permit(:prefill)
+    end
+
+    def destroy_blank_application
+      application_form = current_candidate.application_forms.first
+      application_form.destroy if application_form.blank_application?
     end
   end
 end
--- a/app/models/candidate.rb
+++ b/app/models/candidate.rb
@@ -23,7 +23,7 @@ class Candidate < ApplicationRecord
   end
 
   def current_application
-    application_form = application_forms.last
+    application_form = application_forms.order(:created_at).last
     application_form ||= application_forms.create!
     application_form
   end
--- a/spec/mailers/candidate_mailer_ucas_match_initial_email_spec.rb
+++ b/spec/mailers/candidate_mailer_ucas_match_initial_email_spec.rb
@@ -9,13 +9,12 @@ RSpec.describe CandidateMailer, type: :mailer do
     end
   end
 
-  let(:application_form) { build_stubbed(:application_form, first_name: 'Jane', application_choices: [application_choice]) }
-  let(:provider) { build_stubbed(:provider, name: 'Coventry University') }
-  let(:course_option) { build_stubbed(:course_option, course: build_stubbed(:course, name: 'Physics', code: '3PH5', provider: provider)) }
-  let(:application_choice) { build_stubbed(:application_choice, course_option: course_option) }
-
   describe '.ucas_match_initial_email_duplicate_applications' do
     let(:email) { mailer.ucas_match_initial_email_duplicate_applications(application_form.application_choices.first) }
+    let(:application_form) { build_stubbed(:application_form, first_name: 'Jane', application_choices: [application_choice]) }
+    let(:provider) { build_stubbed(:provider, name: 'Coventry University') }
+    let(:course_option) { build_stubbed(:course_option, course: build_stubbed(:course, name: 'Physics', code: '3PH5', provider: provider)) }
+    let(:application_choice) { build_stubbed(:application_choice, course_option: course_option) }
 
     it_behaves_like(
       'a mail with subject and content',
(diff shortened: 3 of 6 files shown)


davidgisbey (approved):

TobyRet:
@stevehook - I'm getting a single failure on the `mailers_controller_spec` 

mailer_error

A 'candidate` is [passed into the method ](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/mailers/candidate_mailer.rb#L369) but upon inspecting the Candidate `COUNT` is 0.

stevehook:
> @stevehook - I'm getting a single failure on the `mailers_controller_spec`. Do you know what might be going on here?
> 
> mailer_error
> 
> A 'candidate' is [passed into the method](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/mailers/candidate_mailer.rb#L369) but upon inspecting the Candidate `COUNT` is 0.

So I think this is happening just because `candidate` in this test is created with `build_stubbed`. It's a fake - it has an `id` but isn't persisted. So when `current_application` is called here - https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/mailers/candidate_mailer.rb#L369, `Candidate#current_application` calls `application_forms.order(:created_at).last` and because of the `order` (and some internals to do with FactoryBot stubbing that I don't understand) it has to run a query (rather than just returning the `application_forms` in memory) and ends up trying to persist the application forms first.

TobyRet:
> So I think this is happening just because `candidate` in this test is created with `build_stubbed`. It's a fake - it has an `id` but isn't persisted. So when `current_application` is called here - https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/mailers/candidate_mailer.rb#L369, `Candidate#current_application` calls `application_forms.order(:created_at).last` and because of the `order` (and some internals to do with FactoryBot stubbing that I don't understand) it has to run a query (rather than just returning the `application_forms` in memory) and ends up trying to persist the application forms first.
> 
> `application_forms.sort_by(&:created_at).last` works because it's not hitting the DB but that doesn't seem like the right thing to do.

Yep looks like it is an issue with `build_stubbed` - where the specs are broken we need to persist the candidate and associations. Have updated specs and now looks ok. Thanks @stevehook

app/controllers/candidate_interface/prefill_application_form_controller.rb
```diff
@@ -37,5 +40,17 @@ def prefill_candidate_application_form
...
+      blank_application.destroy if blank_application?
+    end
+
+    def blank_application?
+      blank_application.created_at == blank_application.updated_at
+    end
+
+    def blank_application
+      @blank_application ||= current_candidate.application_forms.first
+    end
```
TobyRet, line 54: @benswannack - I've added a guard clause to ensure we only destroy applications that are actually blank. What do you think?

Not sure if this logic should sit here or somewhere else though ... Could potentially create a service but is that overkill? 🤔

TobyRet: Maybe `#blank_application?` could move to the model 🤔

duncanjbrown: There's already a `#blank_application?` method on `ApplicationForm`

duncanjbrown:
This should solve https://trello.com/c/fT2Fvq3O/3531-on-sandbox-we-ask-the-would-you-like-to-prefill-this-application-question-when-the-candidate-has-an-application-in-progress as well?

app/controllers/candidate_interface/prefill_application_form_controller.rb
TobyRet: Ah yes - there is! I'll fix

benswannack (approved):
This looks good to me! (and will fix the ticket Duncan has linked as a bonus :) )

TobyRet:
> This should solve https://trello.com/c/fT2Fvq3O/3531-on-sandbox-we-ask-the-would-you-like-to-prefill-this-application-question-when-the-candidate-has-an-application-in-progress as well?

@duncanjbrown - I was not able to replicate the bug you listed in the card on master.
I did the following ...
1.  signed up for an account
2.  selected 'pre-fill application'
3. signed out
4. was taken straight to my pre-filled application

.env.development
```diff
@@ -4,3 +4,4 @@ HOSTING_ENVIRONMENT_NAME=development
 STATE_CHANGE_SLACK_URL=
 GOOGLE_ANALYTICS_APPLY=UA-APPLY-XX
 GOOGLE_ANALYTICS_MANAGE=UA-MANAGE-XX
+SANDBOX=true
```
benswannack, line 7: Is this intentional?

TobyRet: Woops nope - have reverted

PR #4425, opened 2021-03-30
Use GovukComponent::Panel

paulrobertlloyd:
## Context

Uses `GovukComponent::Panel` to render panel component (via `govuk-components` `govuk_panel` helper method).

## Guidance to review

* Right now we have to set a `nil` value for `body` where no body content is shown. [This may change in a future version of the component](https://github.com/DFE-Digital/govuk-components/issues/135).
* I’ve used `html_safe` in a few places. Is that okay?
* Adjusted white space around headings and paragraphs on these pages to be consistent. It’s a losing battle, but I fight on.

## Things to check

### DIFF
--- a/app/views/provider_interface/provider_relationship_permissions_setup/success.html.erb
+++ b/app/views/provider_interface/provider_relationship_permissions_setup/success.html.erb
@@ -1,10 +1,6 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <div class="govuk-panel govuk-panel--confirmation">
-      <h1 class="govuk-panel__title">
-        Organisational permissions set
-      </h1>
-    </div>
+    <%= govuk_panel(title: 'Organisational permissions set', body: nil) %>
 
     <p class="govuk-body govuk-!-margin-bottom-4">
       You can now <%= govuk_link_to 'view applications', provider_interface_applications_path %><% if current_provider_user.authorisation.can_manage_users_for_at_least_one_provider? %> or <%= govuk_link_to 'invite users', provider_interface_provider_users_path %><% end %>.
--- a/app/views/referee_interface/reference/confirmation.html.erb
+++ b/app/views/referee_interface/reference/confirmation.html.erb
@@ -14,11 +14,7 @@
 <% else %>
   <% content_for :title, t('page_titles.referee.confirmation', full_name: @reference.application_form.full_name) %>
 
-  <div class="govuk-panel govuk-panel--confirmation govuk-!-margin-bottom-8">
-    <h1 class="govuk-panel__title">
-      <%= t('page_titles.referee.confirmation', full_name: @reference.application_form.full_name) %>
-    </h1>
-  </div>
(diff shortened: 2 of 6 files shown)


app/views/provider_interface/provider_relationship_permissions_setup/success.html.erb
```diff
@@ -1,17 +1,10 @@
...
-    </p>
-
-    <p class="govuk-body govuk-!-margin-bottom-4">
-      You can change organisational permissions in <%= govuk_link_to t('page_titles.provider.account').downcase, provider_interface_account_path %>.
-    </p>
+    <%= govuk_panel(
+      title: 'Organisational permissions set',
+      body: nil,
+    ) %>
+    <p class="govuk-body govuk-!-margin-bottom-4">You can now <%= govuk_link_to 'view applications', provider_interface_applications_path %><% if current_provider_user.authorisation.can_manage_users_for_at_least_one_provider? %> or <%= govuk_link_to 'invite users', provider_interface_provider_users_path %><% end %>.</p>
```
raamSoftwire, line 7: This is just a personal preference, but I find putting everything on one line harder to read and work with. Especially in this case with the extra `if` statement.

raamSoftwire (approved):
This looks good to me. I think the uses of `html_safe` here are OK because it is used for values that we generate rather than user input.

app/views/referee_interface/reference/confirmation.html.erb
```diff
@@ -14,11 +14,10 @@
...
   <% content_for :title, t('page_titles.referee.confirmation', full_name: @reference.application_form.full_name) %>
 
-  <div class="govuk-panel govuk-panel--confirmation govuk-!-margin-bottom-8">
-    <h1 class="govuk-panel__title">
-      <%= t('page_titles.referee.confirmation', full_name: @reference.application_form.full_name) %>
-    </h1>
-  </div>
+  <%= govuk_panel(
+    title: t('page_titles.referee.confirmation', full_name: @reference.application_form.full_name),
+    body: nil,
```
duncanjbrown, line 19: Tiny nit: the body element still takes up space! 

Screenshot 2021-03-31 at 09 34 56

Happy to merge this tho — we can fix upstream?

duncanjbrown (approved):
vg ✅  — agree with @raamSoftwire 's comment!

app/views/referee_interface/reference/confirmation.html.erb
paulrobertlloyd: Oh yes, bugger. 😕

app/views/provider_interface/provider_relationship_permissions_setup/success.html.erb
paulrobertlloyd: No worries, reverted. I just wish we had a settled set of conventions for markup… but it’s a small, tiny thing.

app/views/referee_interface/reference/confirmation.html.erb
paulrobertlloyd: Yeah, we can get a way with it… and when the component is updated we can go back and remove `body: nil`.

PR #4428, opened 2021-03-30
Use GovukComponent::Tag

paulrobertlloyd:
## Context

Use `GovukComponent::Tag` to render tag component (via `govuk-components` `govuk_tag` helper method).

* Replace instances of markup that uses `govuk-tag`
* Replace `TagComponent` with `govuk_tag` component from `govuk-components`.
* Where we’ve defined `type` to pass onto`TagComponent`, rename to `colour` to pass onto the respective param on `govuk_tag`
* Ensure tag text is correctly capitalised
* Remove `app-tag` class and override `govuk-tag` directly

## Guidance to review

I’ve removed instances of `app-tag`, which we were using everywhere (either via `TagComponent` or by adding manually). 

Rather than create a class, given such widespread usage, we should add the additional CSS properties to `govuk-tag` via our own application styles, i.e. use the cascade.

Possibly controversial, but given the additional styles we are adding (preventing wrapping and adjusting vertical alignment), and that we’re adding them to every tag, this seems like a far more pragmatic solution. I defer to @adamsilver (unless he disagrees with me 😉).

## Things to check


app/components/support_interface/ucas_match_action_component.rb
```diff
@@ -58,7 +58,7 @@ def button
   private
 
     def type_of_action
-      "<strong class='govuk-tag govuk-tag--yellow app-tag'>Action needed</strong> #{ACTIONS[@match.next_action][:description]}".html_safe
+      "#{govuk_tag(text: 'Action needed', colour: 'yellow')} #{ACTIONS[@match.next_action][:description]}".html_safe
```
stevehook, line 61: If you want to get rid of the string interpolation and `html_safe` you can do something like...
```
      capture do
        concat govuk_tag(text: 'Action needed', colour: 'yellow')
        concat ' '
        concat ACTIONS[@match.next_action][:description]
      end
```
Not sure that makes it any clearer. Totally optional.

app/components/task_list_item_component.html.erb
```diff
@@ -1,20 +1,12 @@
...
   <% elsif custom_status %>
-    <strong class="govuk-tag govuk-tag--purple app-tag" id="<%= tag_id %>">
-      <%= custom_status %>
-    </strong>
+    <%= govuk_tag(text: custom_status, colour: 'purple', html_attributes: { id: tag_id }) %>
   <% elsif completed %>
-    <strong class="govuk-tag govuk-tag--complete app-tag" id="<%= tag_id %>">
-      Completed
-    </strong>
+    <%= govuk_tag(text: 'Completed', html_attributes: { id: tag_id }) %>
```
stevehook, line 8: Does this one need a `colour`? Looks like `govuk-components` handles this and just skips the colour class.

stevehook (approved):
Looks good. Couple of questions inline but good to 🚢 as is

app/components/task_list_item_component.html.erb
paulrobertlloyd: Nah; defaults to the tag default, which is white on dark blue.

app/components/provider_interface/application_status_tag_component.html.erb
```diff
@@ -1 +1 @@
-<%= render TagComponent.new(text: text, type: type) %>
+<%= govuk_tag(text: text, colour: colour) %>
```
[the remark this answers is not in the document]
paulrobertlloyd: Yeah, we’ve a few places where intermediary components become kind of useless once we start introducing the components gem. Converting to helpers sounds like a good idea, one for another PR.

app/components/support_interface/ucas_match_action_component.rb
paulrobertlloyd: Nice! Have refactored the other content in this part of the UI, too: https://github.com/DFE-Digital/apply-for-teacher-training/pull/4428/commits/1f5598389fd6c3afeb32d065bd5fe8628af8d1be

PR #4432, opened 2021-03-31
Create service-key for database and redis connection

vigneshmsft:
## Context

On PaaS when redis is bound to an app, it populates the `VCAP_SERVICES` env variable and connection string
has to be parsed from that json. 
It is decided to keep code change minimal and populate the `REDIS_URL` as this is the default env variable most libraries use when connecting to redis.

BLAZER requires the database connection string in the `BLAZER_DATABASE_URL` variable.

Create service-keys for postgres and redis and populate
`BLAZER_DATABASE_URL` and `REDIS_URL` env vars respectively.

## Things to check


malcolmbaig (approved):
This approach seems fine to me. Would be good to get @duncanjbrown 's view on it as well.

duncanjbrown (approved):

saliceti (approved):

PR #4434, opened 2021-03-31
Refactor offer wizard conditions into separate model

benswannack:
## Context
We are planning on allowing any number of conditions on an offer, so in an attempt to make that change easy, we move away from a hard coded 4 conditions.

Add a model to manage conditions in the offer wizard
Use fields_for to correctly index the condition params and ensure that error links and field population works

Still to do:
- [x] Fix the tests
- [x] Make the error text consistent with what it used to be

## Link to Trello card
https://trello.com/c/AMcA4XK8/3514-spike-conditions-of-offer

### DIFF
--- a/app/controllers/provider_interface/offer/conditions_controller.rb
+++ b/app/controllers/provider_interface/offer/conditions_controller.rb
@@ -7,7 +7,7 @@ module ProviderInterface
       end
 
       def create
-        @wizard = OfferWizard.new(offer_store, conditions_params.to_h)
+        @wizard = OfferWizard.new(offer_store, conditions_attributes_for_wizard)
 
         if @wizard.valid_for_current_step?
           @wizard.save_state!
@@ -24,7 +24,7 @@ module ProviderInterface
       end
 
       def update
-        @wizard = OfferWizard.new(offer_store, conditions_params.to_h)
+        @wizard = OfferWizard.new(offer_store, conditions_attributes_for_wizard)
 
         if @wizard.valid_for_current_step?
           @wizard.save_state!
@@ -38,9 +38,18 @@ module ProviderInterface
     private
 
       def conditions_params
-        params.require(:provider_interface_offer_wizard).permit(:further_condition_1, :further_condition_2,
-                                                                :further_condition_3, :further_condition_4,
-                                                                standard_conditions: [])
+        params.require(:provider_interface_offer_wizard)
+              .permit(further_conditions: {}, standard_conditions: [])
+      end
+
+      def conditions_attributes_for_wizard
+        params = conditions_params.to_h
+
+        params['further_conditions'] = params['further_conditions'].values.map do |hash|
+          hash['text']
+        end
+
+        params
       end
     end
   end
--- a/app/controllers/provider_interface/offers_controller.rb
+++ b/app/controllers/provider_interface/offers_controller.rb
@@ -32,9 +32,13 @@ module ProviderInterface
     def edit; end
 
     def show
-      @wizard = OfferWizard.new(offer_store,
-                                offer_context_params(:change_offer).merge!(current_step: :offer))
-      @wizard.configure_additional_conditions(@application_choice.offer['conditions'] - MakeAnOffer::STANDARD_CONDITIONS)
+      @wizard = OfferWizard.build_from_application_choice(
+        offer_store,
+        @application_choice,
+        provider_user_id: current_provider_user.id,
+        current_step: :offer,
+        decision: :change_offer,
+      )
       @wizard.save_state!
 
       return unless provider_user_can_make_decisions
@@ -102,23 +106,6 @@ module ProviderInterface
         user: current_provider_user,
         current_course: @application_choice.offered_course,
       )
-    end
-
-    def offer_context_params(decision = :default)
-      course_option = @application_choice.offered_option
-      conditions = @application_choice.offer['conditions'] || MakeAnOffer::STANDARD_CONDITIONS
-
-      {
-        provider_user_id: current_provider_user.id,
-        application_choice_id: @application_choice.id,
-        course_id: course_option.course.id,
-        course_option_id: course_option.id,
-        provider_id: course_option.provider.id,
-        study_mode: course_option.study_mode,
-        location_id: course_option.site.id,
-        decision: decision,
-        standard_conditions: conditions,
-      }
     end
 
     helper_method :provider_user_can_make_decisions
--- a/app/forms/provider_interface/offer_wizard.rb
+++ b/app/forms/provider_interface/offer_wizard.rb
@@ -4,10 +4,10 @@ module ProviderInterface
 
     STEPS = { make_offer: %i[select_option conditions check],
               change_offer: %i[select_option providers courses study_modes locations conditions check] }.freeze
+    NUMBER_OF_FURTHER_CONDITIONS = 4
 
     attr_accessor :provider_id, :course_id, :course_option_id, :study_mode, :location_id,
-                  :standard_conditions, :further_condition_1, :further_condition_2,
-                  :further_condition_3, :further_condition_4, :current_step, :decision,
+                  :standard_conditions, :further_conditions, :current_step, :decision,
                   :action, :path_history, :wizard_path_history,
                   :provider_user_id, :application_choice_id
 
@@ -15,19 +15,37 @@ module ProviderInterface
     validates :course_option_id, presence: true, on: %i[locations save]
     validates :study_mode, presence: true, on: %i[study_modes save]
     validates :course_id, presence: true, on: %i[courses save]
-    validates :further_condition_1, :further_condition_2, :further_condition_3, :further_condition_4, length: { maximum: 255 }
+    validate :further_conditions_valid, on: %i[conditions]
     validate :course_option_details, if: :course_option_id, on: :save
 
     def initialize(state_store, attrs = {})
       @state_store = state_store
 
       super(last_saved_state.deep_merge(attrs))
+      setup_further_conditions
       update_path_history(attrs)
     end
 
+    def self.build_from_application_choice(state_store, application_choice, options = {})
+      course_option = application_choice.offered_option
+
+      attrs = {
+        application_choice_id: application_choice.id,
+        course_id: course_option.course.id,
+        course_option_id: course_option.id,
+        provider_id: course_option.provider.id,
+        study_mode: course_option.study_mode,
+        location_id: course_option.site.id,
+        decision: :default,
+        standard_conditions: standard_conditions_from(application_choice.offer),
+        further_conditions: further_conditions_from(application_choice.offer),
+      }.merge(options)
+
+      new(state_store, attrs)
+    end
+
     def conditions
-      @conditions = (standard_conditions + [further_condition_1, further_condition_2,
-                                            further_condition_3, further_condition_4]).reject!(&:blank?)
+      @conditions = (standard_conditions + further_conditions).reject(&:blank?)
     end
 
     def course_option
@@ -62,13 +80,31 @@ module ProviderInterface
 
     delegate :previous_step, to: :wizard_path_history
 
-    def configure_additional_conditions(conditions)
-      conditions.each_with_index do |condition, index|
-        instance_variable_set "@further_condition_#{index + 1}", condition
+    def condition_models
+      @_condition_models ||= further_conditions.map.with_index do |further_condition, index|
+        OfferConditionField.new(id: index, text: further_condition)
       end
     end
 
   private
+
+    def self.standard_conditions_from(offer)
+      return MakeAnOffer::STANDARD_CONDITIONS if offer.blank?
+
+      conditions = offer['conditions']
+      conditions & MakeAnOffer::STANDARD_CONDITIONS
+    end
+
+    private_class_method :standard_conditions_from
+
+    def self.further_conditions_from(offer)
+      return [] if offer.blank?
+
+      conditions = offer['conditions']
+      conditions - MakeAnOffer::STANDARD_CONDITIONS
+    end
+
+    private_class_method :further_conditions_from
 
     def course_option_details
       OfferedCourseOptionDetailsCheck.new(provider_id: provider_id,
@@ -137,7 +173,7 @@ module ProviderInterface
 
     def state
       as_json(
-        except: %w[state_store errors validation_context query_service wizard_path_history],
+        except: %w[state_store errors validation_context query_service wizard_path_history _condition_models],
       ).to_json
     end
 
@@ -147,6 +183,23 @@ module ProviderInterface
                                                    action: attrs[:action].presence)
       @wizard_path_history.update
       @path_history = @wizard_path_history.path_history
+    end
+
+    def setup_further_conditions
+      self.further_conditions = NUMBER_OF_FURTHER_CONDITIONS.times.map do |index|
+        existing_value = further_conditions && further_conditions[index]
+        existing_value.presence || ''
+      end
+    end
+
+    def further_conditions_valid
+      condition_models.map(&:valid?).all?
+
+      condition_models.each do |model|
+        model.errors.each do |key, message|
+          errors.add("further_conditions[#{model.id}][#{key}]", message)
+        end
+      end
     end
   end
 end
--- a/spec/forms/provider_interface/offer_wizard_spec.rb
+++ b/spec/forms/provider_interface/offer_wizard_spec.rb
@@ -9,10 +9,7 @@ RSpec.describe ProviderInterface::OfferWizard do
                         location_id: location_id,
                         application_choice_id: application_choice_id,
                         standard_conditions: standard_conditions,
-                        further_condition_1: further_condition_1,
-                        further_condition_2: further_condition_2,
-                        further_condition_3: further_condition_3,
-                        further_condition_4: further_condition_4,
+                        further_conditions: further_conditions,
                         current_step: current_step,
                         decision: decision)
   end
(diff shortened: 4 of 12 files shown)


app/controllers/provider_interface/offer/conditions_controller.rb
```diff
@@ -38,8 +38,7 @@ def update
     private
 
       def conditions_params
-        params.require(:provider_interface_offer_wizard).permit(:further_condition_1, :further_condition_2,
-                                                                :further_condition_3, :further_condition_4,
+        params.require(:provider_interface_offer_wizard).permit(further_conditions: {},
```
benswannack, line 41: Not sure if there's a way to be more strict about the params here, since we don't necessarily know the keys of the hash

app/forms/provider_interface/offer_condition_fields.rb
```diff
@@ -0,0 +1,9 @@
+module ProviderInterface
+  class OfferConditionFields
+    include ActiveModel::Model
+
+    attr_accessor :id, :text
+
+    validates :text, length: { maximum: 255 }
```
benswannack, line 7: Need to customise this error message to include the ID of the condition

app/forms/provider_interface/offer_wizard.rb
```diff
@@ -123,5 +122,21 @@ def update_path_history(attrs)
       @wizard_path_history.update
       @path_history = @wizard_path_history.path_history
     end
+
+    def setup_further_conditions(attrs)
+      self.further_conditions = 4.times.to_a.to_h do |id|
+        [id, attrs.dig('further_conditions', id.to_s).presence || { 'text' => '' }]
+      end
+    end
```
benswannack, line 130: This is a bit messy. The intention is to parameterize how many boxes we want, and then to populate the first boxes with values that we get from params

cwrw: I was wondering if we can move the determination of how many conditions are rendered to the views, and as far as the wizard is concerned it just validates what it gets. This makes the logic here simpler. And its a smaller step when we integrate the add another feature which will be determining that as well?

benswannack: I'm not sure we can keep it just to the view, since the view needs a number of already created OfferConditions models in order to render the fields_for

app/controllers/provider_interface/offer/conditions_controller.rb
cwrw: I think its ok to accept that we have dynamic params in `further_conditions`

app/forms/provider_interface/offer_wizard.rb
benswannack: I think the problem here is that it's trying to do more than it needs to at the moment. Simpler version to come

app/controllers/provider_interface/offers_controller.rb
```diff
@@ -101,6 +100,7 @@ def available_course_options(course_id, study_mode)
     end
 
     def offer_context_params(course_option, conditions = MakeAnOffer::STANDARD_CONDITIONS, decision = :default)
+      further_conditions_params = (conditions - MakeAnOffer::STANDARD_CONDITIONS).map.with_index.to_h { |c, index| [index.to_s, { 'text' => c }] }
```
benswannack, line 103: I'm going to move this into a `build_from_application_choice` method inside the wizard, since the controller will be cleaner if it just passes params to the wizard

app/forms/provider_interface/offer_wizard.rb
```diff
@@ -6,28 +6,51 @@ class OfferWizard
...
     def initialize(state_store, attrs = {})
       @state_store = state_store
 
       super(last_saved_state.deep_merge(attrs))
+      setup_further_conditions
       update_path_history(attrs)
     end
 
+    def self.build_from_application_choice(state_store, application_choice, options = {})
+      course_option = application_choice.offered_course_option || application_choice.course_option
```
despo, line 29: This should always retrieve the offered_course_option

app/forms/provider_interface/offer_wizard.rb
```diff
@@ -6,28 +6,51 @@ class OfferWizard
...
+
+      conditions = (application_choice.offer || {})['conditions'].presence || MakeAnOffer::STANDARD_CONDITIONS
```
despo, line 31: I find this really confusing and hard to understand. Was there a problem with how conditions were retrieved on the first place?

app/forms/provider_interface/offer_wizard.rb
```diff
@@ -6,28 +6,51 @@ class OfferWizard
...
+
+      further_conditions = (conditions - MakeAnOffer::STANDARD_CONDITIONS).map.with_index.to_h do |c, index|
```
despo, line 33: Can this be extracted into its own method and more meaninful variable names used, e.g. condition instead of c.

app/forms/provider_interface/offer_wizard.rb
```diff
@@ -6,28 +6,51 @@ class OfferWizard
...
+        decision: :default,
+      }.merge(options)
+
+      new(state_store, attrs)
+    end
+
     def conditions
-      @conditions = (standard_conditions + [further_condition_1, further_condition_2,
-                                            further_condition_3, further_condition_4]).reject!(&:blank?)
+      @conditions = (standard_conditions + further_conditions.values.map { |c| c['text'] }).reject(&:blank?)
```
despo, line 53:
```suggestion
      @conditions = (standard_conditions + further_conditions.values.map { |condition| condition['text'] }).reject(&:blank?)
```

app/forms/provider_interface/offer_wizard.rb
```diff
@@ -62,9 +85,9 @@ def next_step(step = current_step)
 
     delegate :previous_step, to: :wizard_path_history
 
-    def configure_additional_conditions(conditions)
-      conditions.each_with_index do |condition, index|
-        instance_variable_set "@further_condition_#{index + 1}", condition
+    def conditions_models
```
despo, line 88: I find this method name confusing as this is not a model and also, why is it important?

app/forms/provider_interface/offer_wizard.rb
```diff
@@ -148,5 +171,22 @@ def update_path_history(attrs)
...
+        existing_value = further_conditions && further_conditions[id.to_s]
+        [id.to_s, existing_value.presence || { 'text' => '' }]
+      end
+    end
+
+    def further_conditions_valid
+      return if conditions_models.all?(&:valid?)
+
+      conditions_models.each do |model|
+        model.errors.map do |key, message|
```
despo, line 186:
```suggestion
        model.errors.each do |key, message|
```

app/forms/provider_interface/offer_wizard.rb
benswannack: I don't think that's the case. Previously we were building an OfferWizard from an application choice in two places: decisions_controller and offer_controller. In the former, the application won't have had an offer yet, so we used the course_option, and in the latter we need to use the offered_option.

I'll restructure this to make it clearer there is differing behaviour between already offered applications and ones with no offer yet

app/forms/provider_interface/offer_wizard.rb
benswannack: It is a list of models, which are passed to the fields_for in the view.

This is done because:
1. It commonises the validation of condition length
2. It allows the params coming from the form to be correctly indexed (which is why `id` is needed) which hooks up to formbuilder's error summary nicely

app/forms/provider_interface/offer_wizard.rb
cwrw: maybe we can use `offered_option` which does this check?

benswannack: I've done some restructuring so hopefully the intention of the method is clearer. Let me know if there are any more improvements you can think of!

spec/system/provider_interface/change_existing_offer_spec.rb
```diff
@@ -104,7 +104,7 @@ def and_the_default_conditions_are_checked
   end
 
   def when_i_add_further_conditions
-    fill_in('provider_interface_offer_wizard[further_condition_1]', with: 'A* on Maths A Level')
+    fill_in('provider_interface_offer_wizard[further_condition[0]]', with: 'A* on Maths A Level')
```
benswannack, line 107: This is wrong, but the method is unused, fixing...

app/controllers/provider_interface/offer/conditions_controller.rb
```diff
@@ -38,9 +38,15 @@ def update
     private
 
       def conditions_params
-        params.require(:provider_interface_offer_wizard).permit(:further_condition_1, :further_condition_2,
-                                                                :further_condition_3, :further_condition_4,
-                                                                standard_conditions: [])
+        conditions_params = params.require(:provider_interface_offer_wizard)
```
cwrw, line 41: this method is doing too much, can this be split up into multiple methods, one to permit params and one to calculate the `further_conditions`?

app/forms/provider_interface/offer_wizard.rb
```diff
@@ -6,28 +6,54 @@ class OfferWizard
...
 
     def initialize(state_store, attrs = {})
       @state_store = state_store
 
       super(last_saved_state.deep_merge(attrs))
+      setup_further_conditions
       update_path_history(attrs)
     end
 
+    def self.build_from_application_choice(state_store, application_choice, options = {})
```
cwrw, line 28: This method is doing too much, can we break it up into smaller methods, where setting standard/further conditions is set in separate methods?

spec/forms/provider_interface/offer_wizard_spec.rb
```diff
@@ -39,10 +44,6 @@
     it { is_expected.to validate_presence_of(:course_option_id).on(:locations).on(:save) }
     it { is_expected.to validate_presence_of(:study_mode).on(:study_modes).on(:save) }
     it { is_expected.to validate_presence_of(:course_id).on(:courses).on(:save) }
-    it { is_expected.to validate_length_of(:further_condition_1).is_at_most(255) }
```
cwrw, line 42: can we add a test that we surface the validations that come from the offer conditions object?

spec/system/provider_interface/make_offer_spec.rb
```diff
@@ -39,7 +39,10 @@
     then_the_conditions_page_is_loaded
     and_the_default_conditions_are_checked
 
-    when_i_add_further_conditions
+    when_i_add_invalid_further_conditions
```
cwrw, line 42: Maybe this should be at the offer wizard spec level, as usually we test for the happy path 🤔

cwrw (commented):
Looks good and tested locally, a few comments about large methods. I think my only further comment would be around the wizard knowing too much about what a standard vs further condition is, can we extract this into the offer conditions field object instead?

spec/forms/provider_interface/offer_wizard_spec.rb
benswannack: Yes, also found a bug with this!

app/forms/provider_interface/offer_wizard.rb
```diff
@@ -148,5 +181,22 @@ def update_path_history(attrs)
...
+
+    def setup_further_conditions
+      self.further_conditions = 4.times.map do |index|
+        existing_value = further_conditions && further_conditions[index]
+        existing_value.presence || ''
+      end
+    end
+
+    def further_conditions_valid
+      conditions_models.map(&:valid?).all?
```
benswannack, line 193: This, because we need to run `valid?` on all of the conditions, rather than returning at the first invalid one

app/forms/provider_interface/offer_wizard.rb
```diff
@@ -62,14 +79,30 @@ def next_step(step = current_step)
...
+      else
+        standard_conditions = MakeAnOffer::STANDARD_CONDITIONS
+        further_conditions = []
+      end
+
+      {
+        standard_conditions: standard_conditions,
+        further_conditions: further_conditions,
+      }
+    end
```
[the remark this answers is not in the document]
cwrw: hmm actually maybe we can shuffle the methods around as such:
```
def self.standard_conditions(offer)
  return  MakeAnOffer::STANDARD_CONDITIONS unless offer.present?

  conditions & MakeAnOffer::STANDARD_CONDITIONS
end

def self.further_conditions(offer)
  return [] unless offer.present?

  conditions - MakeAnOffer::STANDARD_CONDITIONS
end

private_class_method :standard_conditions, :further_conditions
```

then we can just set the conditions in the main attrs hash:
```
attrs = {
    application_choice_id: application_choice.id,
    course_id: course_option.course.id,
    course_option_id: course_option.id,
    provider_id: course_option.provider.id,
    study_mode: course_option.study_mode,
    location_id: course_option.site.id,
    decision: :default,
    standard_conditions: standard_conditions(application_choice.offer)
    further_conditions: further_conditions(application_choice.offer)
  }
```
what do you think?

cwrw (approved):
Looks good thanks for addressing comments. We can live with the conditions being in offer wizard for now until we move the offer/conditions into their own models.

Left a nit/suggestion around the methods otherwise 👍

cwrw (approved):

app/forms/provider_interface/offer_condition_fields.rb
```diff
@@ -0,0 +1,9 @@
+module ProviderInterface
+  class OfferConditionFields
+    include ActiveModel::Model
+
+    attr_accessor :id, :text
+
+    validates :text, length: { maximum: 255, too_long: ->(c, _) { "Condition #{c.id + 1} must be %{count} characters or fewer" } }
```
despo, line 7: Consider moving the validation error in the make offer locale file for consistency

app/forms/provider_interface/offer_wizard.rb
```diff
@@ -148,5 +183,22 @@ def update_path_history(attrs)
       @wizard_path_history.update
       @path_history = @wizard_path_history.path_history
     end
+
+    def setup_further_conditions
+      self.further_conditions = 4.times.map do |index|
```
despo, line 188: Can we extract `4` into a constant, so it has some context. Maybe `MAX_NUMBER_OF_CONDITIONS_ALLOWED` (or something shorter)?

spec/forms/provider_interface/offer_wizard_spec.rb
```diff
@@ -57,6 +70,59 @@
     end
   end
 
+  describe '.build_from_application_choice' do
+    let(:application_choice) do
+      create(
+        :application_choice,
+        course_option: create(:course_option),
+        offer: offer,
+      )
```
despo, line 79:
```suggestion
      create(
        :application_choice,
        offer: offer,
      )
```
The default application_choice factory creats a course_option, so I think that line can be removed.
https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/spec/factories/application_choice.rb#L3

spec/system/provider_interface/change_existing_offer_spec.rb
```diff
@@ -71,6 +75,35 @@ def and_i_sign_in_to_the_provider_interface
     provider_signs_in_using_dfe_sign_in
   end
 
+  def given_the_provider_user_can_offer_multiple_provider_courses
+    @selected_provider = create(:provider, :with_signed_agreement)
+    create(:provider_permissions, provider: @selected_provider, provider_user: provider_user, make_decisions: true)
+    courses = [create(:course, :open_on_apply, study_mode: :full_time_or_part_time, provider: @selected_provider, accredited_provider: ratifying_provider),
+               create(:course, :open_on_apply, study_mode: :full_time_or_part_time, provider: @selected_provider, accredited_provider: ratifying_provider)]
```
despo, line 82:
```suggestion
    courses = create_list(:course, 2, :open_on_apply, study_mode: :full_time_or_part_time, provider: @selected_provider, accredited_provider: ratifying_provider)
```

despo: Sorry I know this was not introduced in this PR!!

spec/forms/provider_interface/offer_wizard_spec.rb
```diff
@@ -57,6 +70,59 @@
...
+    end
+    let(:offer) { { 'conditions' => ['Fitness to train to teach check', 'Be cool'] } }
+
+    it 'correctly populates the wizard with offer conditions' do
+      wizard = described_class.build_from_application_choice(
+        store,
+        application_choice,
+      )
```
despo, line 87: Consider using a context around all examples and moving the wizard setup in a let block, to follow the pattern of the other examples in this spec.

app/forms/provider_interface/offer_wizard.rb
despo: Could it maybe be renamed to `condition_models` instead. I find the name to be a bit confusing and it prompts me to think of a many to many association (because of the rails table name expectation when you have many to many)

despo (approved):
Left some minor feedback but overall happy to approve :shipit:

app/forms/provider_interface/offer_wizard.rb
benswannack: Gone with `NUMBER_OF_FURTHER_CONDITIONS` because it is actually a fixed number of conditions at the moment. (To be changed later when we allow add another)

app/forms/provider_interface/offer_condition_fields.rb
benswannack: I had a look into this, but it doesn't seem to work.

It is possible to pass attributes through to locales by doing e.g. `validates :text, length: { maximum: 255,  attribute: 'whatever' }`, but we cannot refer to the ID of the condition in those, because adding validations happens at the class level, rather than at the instance level.

We rely on the too_long method passing through the instance so we can refer to the id in the string. We can revisit this at a later date

PR #4437, opened 2021-03-31
Add Register API docs

AgaDufrat:
## Context

We need to document the new shiny API ✨ we made for the Register team 💙 

Add OpenAPI spec

## Guidance to review

visit `/register-api`

Pay the most attention to the following new attributes:

- gender
- disabilities
- ethnic_group
- ethnic_background
- comparable_uk_degree

We are also missing `ParameterInvalid` response in the Vendor API docs for GET /applications.
I added this as a separate commit which can be dropped or fixed up.

Screenshot 2021-04-07 at 14 20 04

## Link to Trello card

## Things to check

### DIFF
--- a/app/presenters/register_api/single_application_presenter.rb
+++ b/app/presenters/register_api/single_application_presenter.rb
@@ -42,7 +42,7 @@ module RegisterAPI
           contact_details: contact_details,
           course: course_info_for(application_choice.offered_option),
           qualifications: qualifications,
-          hesa_itt_data: hesa_itt_data.presence || {},
+          hesa_itt_data: hesa_itt_data.presence,
         },
       }
     end
--- a/app/controllers/api_docs/register_api_docs/open_api_controller.rb
+++ b/app/controllers/api_docs/register_api_docs/open_api_controller.rb
@@ -1,0 +1,9 @@
+module APIDocs
+  module RegisterAPIDocs
+    class OpenAPIController < APIDocsController
+      def spec
+        render plain: RegisterAPISpecification.as_yaml, content_type: 'text/yaml'
+      end
+    end
+  end
+end
--- a/app/controllers/api_docs/register_api_docs/pages_controller.rb
+++ b/app/controllers/api_docs/register_api_docs/pages_controller.rb
@@ -1,0 +1,18 @@
+module APIDocs
+  module RegisterAPIDocs
+    class PagesController < APIDocsController
+      def release_notes
+        render_content_page :release_notes
+      end
+
+    private
+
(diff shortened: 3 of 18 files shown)


duncanjbrown:
For guaranteed usefulness, I think this PR should include adding schema checks to the request specs _a la_ https://github.com/DFE-Digital/apply-for-teacher-training/blob/4dd04dfaaecc4e1e796fa0d4f9b3e0f637e0d907/spec/requests/vendor_api/get_multiple_applications_spec.rb#L52

app/presenters/register_api/single_application_presenter.rb
```diff
@@ -34,15 +34,11 @@ def as_json
             english_language_qualifications: application_form.english_language_qualification_details,
             other_languages: application_form.other_language_details,
             disability_disclosure: application_form.disability_disclosure,
-            gender: equality_and_diversity_data['sex'],
```
cwrw, line 37: Didn't we decide to keep this or was there an update around just using HESA instead?

cwrw (commented):
Just the one comment, looks good otherwise, only thing I also noticed when having a look locally is that examples are not rendered as code blocks (example to look at `HTTP 200 - An array of applications` in register vs vendor API)

app/presenters/register_api/single_application_presenter.rb
```diff
@@ -34,15 +34,11 @@ def as_json
...
-            disabilities: equality_and_diversity_data['disabilities'].presence || [],
-            ethnic_group: equality_and_diversity_data['ethnic_group'],
-            ethnic_background: equality_and_diversity_data['ethnic_background'],
           },
           contact_details: contact_details,
           course: course_info_for(application_choice.offered_option),
           qualifications: qualifications,
-          hesa_itt_data: hesa_itt_data.presence || {},
```
cwrw, line 45: do we not want to default to a hash if this is empty?

duncanjbrown: I think it's better to return an explicit `null` than a hash which could have misleading semantics like `hash[:gender] == nil`, which could imply eg the user declined to give this specific piece of information

duncanjbrown:
> Just the one comment, looks good otherwise, only thing I also noticed when having a look locally is that examples are not rendered as code blocks (example to look at `HTTP 200 - An array of applications` in register vs vendor API)

This should be fixed by https://github.com/DFE-Digital/apply-for-teacher-training/pull/4437/commits/d3751849d8c0345a4b2f93e6f425808125aac01e

cwrw (approved):
👍

PR #4438, opened 2021-04-01
Integrate GetChangeOfferOptions into new offer flows

mnacos:
## Context

The new offer flows do not yet use `GetChangeOfferOptions` to generate and display the lists of providers, courses etc. shown in the wizard. This PR addresses this.

Change `OffersController`, `Offer::StudyModesController` and `OfferWizard` to use `GetChangeOfferOptions` when accessing `available_providers`, `available_courses` etc.

Add `application_choice_id` to the wizard's json blob at the start of offer flows, in `OffersController` and `DecisionsController`.

Fix use of `CollectionSelectComponent` for study modes.

Fix offer summary components when using `available_providers`.

## Guidance to review

`application_choice_id` is needed because of the extra logic in `GetChangeOfferOptions` around preserving the HEI/ratifying provider, based on the current course in the application choice.

Setting `application_choice_id` at the start of each flow and not for every step is ok because the key used when persisting wizard state in Redis is application_choice-specific.

Changing the provider is still possible in the context of self-ratified courses, but it's trickier to set up, as it requires the provider to be ratifying courses run by other organisations.

In the offer summary components, `available_providers.many?` fails because the underlying `available_providers.size` chooses to perform a `count` and this returns a hash. Forcing the code to use `length` addresses this.

## Link to Trello card

https://trello.com/c/we8DLCos

## Things to check

### DIFF
--- a/app/controllers/provider_interface/offers_controller.rb
+++ b/app/controllers/provider_interface/offers_controller.rb
@@ -85,13 +85,6 @@ module ProviderInterface
       'back' if !!params[:back]
     end
 
-    def query_service
-      @query_service ||= GetChangeOfferOptions.new(
-        user: current_provider_user,
-        current_course: @application_choice.offered_course,
-      )
-    end
-
     def available_providers
       query_service.available_providers
     end
@@ -102,6 +95,13 @@ module ProviderInterface
 
     def available_course_options(course_id, study_mode)
       query_service.available_course_options(course: Course.find(course_id), study_mode: study_mode)
+    end
+
+    def query_service
+      @query_service ||= GetChangeOfferOptions.new(
+        user: current_provider_user,
+        current_course: @application_choice.offered_course,
+      )
     end
 
     def offer_context_params(decision = :default)
--- a/app/controllers/provider_interface/offers_controller.rb
+++ b/app/controllers/provider_interface/offers_controller.rb
@@ -33,16 +33,14 @@ module ProviderInterface
 
     def show
       @wizard = OfferWizard.new(offer_store,
-                                offer_context_params(@application_choice.offered_course_option,
-                                                     @application_choice.offer['conditions'],
-                                                     :change_offer).merge!(current_step: :offer))
+                                offer_context_params(:change_offer).merge!(current_step: :offer))
       @wizard.configure_additional_conditions(@application_choice.offer['conditions'] - MakeAnOffer::STANDARD_CONDITIONS)
       @wizard.save_state!
 
       return unless provider_user_can_make_decisions
 
       @providers = available_providers
-      @courses = available_courses(@application_choice.offered_course.provider_id)
+      @courses = available_courses(@application_choice.offered_course.provider.id)
       @course_options = available_course_options(@application_choice.offered_course.id, @application_choice.offered_course_option.study_mode)
     end
 
@@ -87,22 +85,32 @@ module ProviderInterface
       'back' if !!params[:back]
     end
 
+    def query_service
+      @query_service ||= GetChangeOfferOptions.new(
+        user: current_provider_user,
+        current_course: @application_choice.offered_course,
+      )
+    end
+
     def available_providers
-      current_provider_user.providers
+      query_service.available_providers
     end
 
     def available_courses(provider_id)
-      Course.where(provider_id: provider_id)
+      query_service.available_courses(provider: Provider.find(provider_id))
     end
 
     def available_course_options(course_id, study_mode)
-      CourseOption.where(course_id: course_id, study_mode: study_mode)
-                  .includes(:site).order('sites.name')
+      query_service.available_course_options(course: Course.find(course_id), study_mode: study_mode)
     end
 
-    def offer_context_params(course_option, conditions = MakeAnOffer::STANDARD_CONDITIONS, decision = :default)
+    def offer_context_params(decision = :default)
+      course_option = @application_choice.offered_option
+      conditions = @application_choice.offer['conditions'] || MakeAnOffer::STANDARD_CONDITIONS
+
       {
         provider_user_id: current_provider_user.id,
+        application_choice_id: @application_choice.id,
         course_id: course_option.course.id,
         course_option_id: course_option.id,
         provider_id: course_option.provider.id,
--- a/app/forms/provider_interface/offer_wizard.rb
+++ b/app/forms/provider_interface/offer_wizard.rb
@@ -31,7 +31,7 @@ module ProviderInterface
     end
 
     def course_option
-      @course_option = CourseOption.find(course_option_id)
+      CourseOption.find(course_option_id)
     end
 
     def save_state!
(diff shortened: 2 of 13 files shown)


app/controllers/provider_interface/offers_controller.rb
```diff
@@ -87,22 +85,32 @@ def action
       'back' if !!params[:back]
     end
 
+    def query_service
+      @query_service ||= GetChangeOfferOptions.new(
+        user: current_provider_user,
+        current_course: @application_choice.offered_course,
+      )
+    end
+
```
despo, line 94: Can we move this further down so the methods accessed by the controller's public methods are more prominent in the controller.

app/forms/provider_interface/offer_wizard.rb
```diff
@@ -8,13 +8,15 @@ class OfferWizard
...
-                  :path_history, :action, :provider_user_id, :wizard_path_history
+                  :action, :path_history, :wizard_path_history,
+                  :provider_user_id, :application_choice_id
 
     validates :decision, presence: true, on: %i[select_option]
     validates :course_option_id, presence: true, on: %i[locations save]
     validates :study_mode, presence: true, on: %i[study_modes save]
     validates :course_id, presence: true, on: %i[courses save]
     validates :further_condition_1, :further_condition_2, :further_condition_3, :further_condition_4, length: { maximum: 255 }
+    validates :application_choice_id, presence: true
```
despo, line 19: We don't want to render an application choice error to the provider_user (as it's not something they can address), so we should not check for presence here. `find` on line 98 will cause a resource not found error which is what we want.

app/forms/provider_interface/offer_wizard.rb
```diff
@@ -90,21 +92,27 @@ def save_and_go_to_next_step(step)
       next_step(step)
     end
 
+    def query_service
+      @query_service ||= GetChangeOfferOptions.new(
+        user: ProviderUser.find(provider_user_id),
+        current_course: ApplicationChoice.find(application_choice_id).offered_course,
```
despo, line 98: Can we extract provider_user and offered_course to private reusable methods.

Also can this method definition be moved further down so the private methods accessed by the public offer wizard methods are more prominent.

app/forms/provider_interface/offer_wizard.rb
```diff
@@ -90,21 +92,27 @@ def save_and_go_to_next_step(step)
...
+      )
+    end
+
     def available_study_modes
-      Course.find(course_id).available_study_modes_from_options
+      query_service.available_study_modes(course: Course.find(course_id))
```
despo, line 103: Can we extract `course` to a private reusable method.

app/forms/provider_interface/offer_wizard.rb
```diff
@@ -90,21 +92,27 @@ def save_and_go_to_next_step(step)
...
     end
 
     def available_course_options
-      CourseOption.where(course_id: course_id, study_mode: study_mode)
+      query_service.available_course_options(course: Course.find(course_id), study_mode: study_mode)
```
despo, line 107: Similar to above, here we should be able to reuse a defined `course` reusable method.

app/forms/provider_interface/offer_wizard.rb
```diff
@@ -90,21 +92,27 @@ def save_and_go_to_next_step(step)
...
     end
 
     def available_courses
-      Course.where(provider_id: provider_id)
+      query_service.available_courses(provider: Provider.find(provider_id))
```
despo, line 111: Can we extract `provider` to a private reusable method.

spec/forms/provider_interface/offer_wizard_spec.rb
```diff
@@ -112,26 +119,18 @@
         let(:current_step) { :providers }
 
         context 'when there are multiple available courses' do
-          before do
-            allow(Course).to receive(:where).and_return(class_double(Course, one?: false))
-          end
-
           it 'returns :courses' do
+            allow(query_service).to receive(:available_courses).and_return([create(:course), create(:course)])
```
despo, line 123:
```suggestion
            allow(query_service).to receive(:available_courses).and_return(create_list(:course, 2))
```

spec/forms/provider_interface/offer_wizard_spec.rb
```diff
@@ -141,29 +140,18 @@
...
-
-            allow(course).to receive(:available_study_modes_from_options).and_return([:full_time])
-            allow(Course).to receive(:find).and_return(course)
-            allow(CourseOption).to receive(:where).and_return(course_option)
-            allow(store).to receive(:write)
-          end
-
           it 'returns :study_modes' do
+            allow(query_service).to receive(:available_study_modes).and_return(%w[part_time])
+            allow(query_service).to receive(:available_course_options).and_return([create(:course_option), create(:course_option)])
```
despo, line 153:
```suggestion
            allow(query_service).to receive(:available_course_options).and_return(create_list(:course_option, 2))
```

spec/forms/provider_interface/offer_wizard_spec.rb
```diff
@@ -173,26 +161,17 @@
...
 
         context 'when there are multiple locations available' do
-          before do
-            course_option = class_double(CourseOption, one?: false)
-
-            allow(CourseOption).to receive(:where).and_return(course_option)
-          end
-
           it 'returns :locations' do
+            allow(query_service).to receive(:available_course_options).and_return([create(:course_option), create(:course_option)])
```
despo, line 165:
```suggestion
            allow(query_service).to receive(:available_course_options).and_return(create_list(:course_option, 2))
```

spec/forms/provider_interface/offer_wizard_spec.rb
```diff
@@ -78,31 +81,35 @@
...
         let(:current_step) { :select_option }
 
         context 'when there are multiple available providers' do
-          before do
-            provider_user = instance_double(ProviderUser, providers: build_stubbed_list(:provider, 2))
-            allow(ProviderUser).to receive(:find).and_return(provider_user)
-          end
-
           it 'returns :providers' do
+            allow(query_service).to receive(:available_providers).and_return([create(:provider), create(:provider)])
```
despo, line 102:
```suggestion
            allow(query_service).to receive(:available_providers).and_return(create_list(:provider,2 ))
```

spec/forms/provider_interface/offer_wizard_spec.rb
```diff
@@ -78,31 +81,35 @@
...
-            courses = build_stubbed_list(:course, 2)
-            provider = instance_double(Provider, id: :stub_id, courses: courses)
-            provider_user = instance_double(ProviderUser, providers: [provider])
-            allow(ProviderUser).to receive(:find).and_return(provider_user)
-            allow(store).to receive(:write)
-          end
-
           it 'returns :courses' do
+            allow(query_service).to receive(:available_providers).and_return([create(:provider)])
+            allow(query_service).to receive(:available_courses).and_return([create(:course), create(:course)])
```
despo, line 111:
```suggestion
            allow(query_service).to receive(:available_courses).and_return(create_list(:course, 2))
```

spec/forms/provider_interface/offer_wizard_spec.rb
despo, line 165: Can we keep the existing pattern of the spec, and set any expectations in the before block rather than the example, as it makes it easier to distinguish the setup and the expectation of each example.

spec/forms/provider_interface/offer_wizard_spec.rb
SamJamCul: We did look into this - we ended up feeling like create_list made it a bit harder to read, as unless you're familiar with the method, you can't immediately see that it results in an array.

cwrw (commented):
looks great 👍 , discussed in person: we will need to change `.one?`in the offer wizard to use the `.length == 1` to accommodate the query and avoid showing steps for one of results

cwrw (approved):

PR #4439, opened 2021-04-01
Skip post-submission confirmation page

stevehook:
## Context

As part of the application dashboard refresh, this PR removes the post-submission confirmation page and redirect candidates straight to dashboard with notification banner instead.

- [x] Feedback form is included in the submit flow (no opt-in/out) but the inputs are all optional.

image

- [x] Redirect to the dashboard straight after submitting the feedback form

image

- [x] Extend `FlashMessageComponent` to handle a secondary as well as the primary message (the requirement is for a message consisting of a heading and sub-text).
- [x] Fix spec regressions

## Guidance to review

- I've implemented primary/secondary flash messages using string array hash messages, e.g. `['Primary message', 'Secondary message']`. With a single string flash message the `FlashMessageComponent` works as it does today. With an array it assumes that `flash[0]` is the primary and `flash[1]` is the secondary message. 
```
flash[:success] = ['Application successfully submitted', 'You will get an email when something changes.']
```

`FlashMessageComponent` now checks type of the flash message and if its an array passes both messages to the `NotificationMessageComponent` which already handles secondary messages. Is there a neater/simpler way? 

## Link to Trello card

https://trello.com/c/TQP8dhal/3192-remove-confirmation-page-and-redirect-candidates-to-dashboard-with-notification-banner-instead

## Things to check


paulrobertlloyd (approved):
LGTM!

PR #4446, opened 2021-04-06
[Spike] Integrate subject filter

steventux:
## Context

https://trello.com/c/anMxp90Q/3559-spike-add-a-subject-filter-to-applications

Feasibility spike, no real issues with integration. More details on Trello.

image

## Guidance to review

Not addressed in this PR: Filtering the list of applications in `FilterApplicationChoicesForProviders` as this will need to be done in conjunction with subject names being available in the database. This will behave like a standard checkbox filter so nothing new here.

## Link to Trello card

https://trello.com/c/anMxp90Q/3559-spike-add-a-subject-filter-to-applications

## Things to check

### DIFF
--- a/app/frontend/packs/components/checkbox_filter.js
+++ b/app/frontend/packs/components/checkbox_filter.js
@@ -1,0 +1,164 @@
+let CheckboxFilter = function(params) {
+  this.params = params
+  this.container = $(params.container)
+  this.container.addClass('app-checkbox-filter--enhanced')
+  this.checkboxes = this.container.find("input[type='checkbox']")
+  this.checkboxesContainer = this.container.find('.app-checkbox-filter__container')
+  this.checkboxesInnerContainer = this.checkboxesContainer.children('.app-checkbox-filter__container-inner')
+  this.legend = this.container.find('legend')
+  this.legend.addClass('govuk-visually-hidden')
+  this.setupStatusBox()
+  this.setupHeading()
+  this.setupTextBox()
+  this.setupHeight()
+}
+
+CheckboxFilter.prototype.setupHeading = function() {
+  this.heading = $('<p class="app-checkbox-filter__title" aria-hidden="true">' + this.legend.text() + '</p>')
+  this.container.prepend(this.heading)
+}
+
+CheckboxFilter.prototype.setupTextBox = function() {
+  var tagContainer = this.container.find('.app-checkbox-filter__selected')
+  if(tagContainer[0]) {
+    tagContainer.after(this.getTextBoxHtml())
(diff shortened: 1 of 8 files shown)


app/frontend/packs/components/checkbox_filter.js
```diff
@@ -0,0 +1,164 @@
+let CheckboxFilter = function(params) {
```
cwrw, line 1: I think we might need to think about renaming this in the real implementation to be representative of it including a search box and checkbox filters

app/frontend/styles/components/_checkbox_filter.scss
```diff
@@ -0,0 +1,135 @@
...
+    background-color: $govuk-focus-colour;
+  }
+
+  &:hover {
+    background-color: govuk-colour("black");
+    color: govuk-colour("white");
+  }
+
+  &:after {
+    background-image: url("/public/images/icon-tag-remove-cross.svg");
```
cwrw, line 121: could we reuse styles/svgs already existing for the filters for the subject filter?

app/frontend/packs/components/checkbox_filter.js
```diff
@@ -0,0 +1,164 @@
...
+  this.checkboxesInnerContainer = this.checkboxesContainer.children('.app-checkbox-filter__container-inner')
+  this.legend = this.container.find('legend')
+  this.legend.addClass('govuk-visually-hidden')
+  this.setupStatusBox()
+  this.setupHeading()
+  this.setupTextBox()
+  this.setupHeight()
+}
+
+CheckboxFilter.prototype.setupHeading = function() {
```
cwrw, line 16: The prototype JS seems to work well when tested in our app and for this spike, however I wonder if we can reuse some of the JS from the original filter and leverage existing logic for to autocompletes? 
We also might need to tidy up this JS so it's more inline with the JS syntax and style we have in the app in the real implementation.

cwrw (commented):
This looks good and works well with our app when tested. However I wonder if we can reuse some of the styles/JS existing currently in the app for the real implementation to avoid duplication and deviating from the established formats currently existing.

cwrw:
spoke in person: agreed most comments are for the real implementation. Re JS syntax @steventux  has suggested an approach to add tests before changing the syntax to ensure nothing breaks which can be done in the real implementation. 

Only thing left to check is reuse of existing styles/JS, and comments will be left in the implementation ticket. Happy for this spike to go to done otherwise, cheers @steventux

cwrw:
attached to trello implementation ticket, closing pr

PR #4447, opened 2021-04-06
Change offerable_courses to scope

SamJamCul:
## Context
A method in the query service missed out on some key information about a course as it only checked the `open_on_apply` field. There is a scope by the same name which is more appropriate, so this is what's changed! 

https://trello.com/c/pN3s3OPa/3588-offerable-courses-needs-to-use-openonapply-scope

Changed `#offerable_courses` in `get_change_offer_options.rb` to use the `open_on_apply` scope.

### DIFF
--- a/app/queries/get_change_offer_options.rb
+++ b/app/queries/get_change_offer_options.rb
@@ -44,11 +44,9 @@ class GetChangeOfferOptions
 
   def offerable_courses
     make_decisions_courses
-      .where(
-        open_on_apply: true,
-        recruitment_cycle_year: recruitment_cycle_year,
-      )
-      .where(ratifying_provider_is_preserved)
+    .open_on_apply
+    .where(recruitment_cycle_year: recruitment_cycle_year)
+    .where(ratifying_provider_is_preserved)
   end
 
   def make_decisions_courses
--- a/spec/queries/get_change_offer_options_spec.rb
+++ b/spec/queries/get_change_offer_options_spec.rb
@@ -82,9 +82,10 @@ RSpec.describe GetChangeOfferOptions do
   end
 
   describe '#offerable_courses' do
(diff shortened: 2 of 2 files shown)


app/queries/get_change_offer_options.rb
```diff
@@ -44,10 +44,8 @@ def available_sites(course:, study_mode:)
 
   def offerable_courses
     make_decisions_courses
-      .where(
-        open_on_apply: true,
-        recruitment_cycle_year: recruitment_cycle_year,
-      )
+      .where(recruitment_cycle_year: recruitment_cycle_year)
+      .open_on_apply
```
cwrw, line 48: nit: can we move this before the `.where`?
Also can you add a test to check that we only include courses which are both `open_on_apply` and exposed_in_find`?

mnacos: We already have a test checking whether `open_on_apply` is taken into account, maybe it's worth modifying this to manually set `open_on_apply: true` for the excluded course, as the `:open_on_apply` factory trait also sets `exposed_in_find`.

SamJamCul: Done, I've amended an existing test to make it better

cwrw (approved):
👍

app/queries/get_change_offer_options.rb
```diff
@@ -44,11 +44,9 @@ def available_sites(course:, study_mode:)
 
   def offerable_courses
     make_decisions_courses
-      .where(
-        open_on_apply: true,
-        recruitment_cycle_year: recruitment_cycle_year,
-      )
-      .where(ratifying_provider_is_preserved)
+    .open_on_apply
```
despo, line 47: 👍🏼

despo (approved):
💯

PR #4448, opened 2021-04-06
Update migration to catch old export name

raamSoftwire:
## Context

Only a small subset of export history records are being shown.

Only `data_export` records with an `export_type` are shown. This field is automatically filled for all new export downloads. We wrote a migration to backfill the old `data_export` records but this migration failed to run successfully because one of the exports was renamed (`Unexplained breaks in work history` to `Work history break`).

This PR updates the migration to catch this particular case. 

## Guidance to review

It would be good if you could test it locally please.

1) Run this migration on master. Open `backfill_export_type.rb` and change line 4 to `MANUAL_RUN = true`. Then run `rake data:migrate:manual[DataMigrations::BackfillExportType]`. It should fail with this error, (if you have a `data_export` with the name "Unexplained breaks in work history").

image

2. Take a look at the data in your `data_exports` table. The older records (>1 month) probably won't have the `export_type` filled in.
3. Then run this migration on this branch and it should succeed. To run this migration a second time you will have to remove the latest record in the `data_migrations` table
4. Take a look at the data in your `data_exports` table. The older records should now have the `export_type` filled in.

## Link to Trello card

https://trello.com/c/K5o3HiH8/3258-fix-export-history-bug

## Things to check

### DIFF
--- a/app/services/data_migrations/backfill_export_type.rb
+++ b/app/services/data_migrations/backfill_export_type.rb
@@ -7,7 +7,9 @@ module DataMigrations
       data_exports = DataExport.all.where(export_type: nil)
 
       data_exports.each do |export|
-        export.update!(export_type: export.name.parameterize.underscore)
+        export_type = export.name == 'Unexplained breaks in work history' ? 'work_history_break' : export.name.parameterize.underscore
+
+        export.update!(export_type: export_type)
       end
     end
   end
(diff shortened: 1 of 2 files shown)


stevehook (approved):
Makes sense to me. Only question I have is how this gets run a second time, is it done manually re-run or has it not yet run in production?

raamSoftwire:
I believe that it never ran in production. I've checked the `data_migrations` table in Blazer and it's not there. I think it tried to run, but then fell over and rolled back.

despo (approved):
Looks good. 👍🏼

PR #4450, opened 2021-04-07
Remove ApplyRedisConnection

malcolmbaig:
This was first introduced as a way of having fine-grained controlled
over how Redis is configured for different environments, motivated
primarily by the configuration of parallel tests.

Parallel test config is now handled by a few lines in spec_helper and
ApplyRedisConnection class is no longer used.

## Things to check


stevehook (approved):
:+1:

PR #4452, opened 2021-04-07
Separate mail templates from different mailers

raamSoftwire:
## Context

The "Candidate email send counts" export counts the number of emails sent, with one row per `mail_template`. However, some emails share the same `mail_template` name but are from different mailers, which means they have different content and different recipients. These emails are currently counted as one, but they need to be separated out.

Update the export by changing the way the groupings work and adding a `mailer` column to make it explicit as to which mailer the `mail_template` is from.

## Guidance to review

Have I missed anything? Was this done on purpose?

## Link to Trello card

https://trello.com/c/66V90gyP/3245-investigate-why-email-notifications-data-extract-has-a-different-list-of-emails-sent-compared-to-qa

## Things to check

### DIFF
--- a/app/exports/candidate_email_send_counts_export.yml
+++ b/app/exports/candidate_email_send_counts_export.yml
@@ -17,3 +17,13 @@ custom_columns:
   unique_recipients:
     type: integer
     description: The number of unique recipients of this email
+
+  mailer:
+    type: string
+    description: The number of unique recipients of this email
+    enum:
+      - Authentication mailer
+      - Candidate mailer
+      - Provider mailer
+      - Referee mailer
+      - Support mailer
--- a/app/exports/candidate_email_send_counts_export.yml
+++ b/app/exports/candidate_email_send_counts_export.yml
@@ -20,7 +20,7 @@ custom_columns:
 
   mailer:
     type: string
-    description: The number of unique recipients of this email
+    description: The Rails mailer which sent the email
     enum:
       - Authentication mailer
       - Candidate mailer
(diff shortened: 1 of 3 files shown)


app/services/support_interface/candidate_email_send_counts_export.rb
```diff
@@ -16,9 +17,9 @@ def data_for_export
     def emails_with_total_send_counts
       Email
         .select(
-          'mail_template, COUNT(mail_template) AS send_count, COUNT(DISTINCT "to") AS unique_recipients, MAX(created_at) AS last_sent_at',
+          'mail_template, COUNT(mail_template) AS send_count, COUNT(DISTINCT "to") AS unique_recipients, MAX(created_at) AS last_sent_at, mailer',
         )
-        .group('mail_template')
+        .group('mail_template, mailer')
```
raamSoftwire, line 22: I used this post for reference here - https://stackoverflow.com/questions/1241352/rails-group-by-multiple-columns

app/exports/candidate_email_send_counts_export.yml
```diff
@@ -17,3 +17,13 @@ custom_columns:
   unique_recipients:
     type: integer
     description: The number of unique recipients of this email
+
+  mailer:
+    type: string
+    description: The number of unique recipients of this email
```
duncanjbrown, line 23:
```suggestion
    description: The Rails mailer which sent the email
```

duncanjbrown (approved):
This looks good — one documentation nit!

app/exports/candidate_email_send_counts_export.yml
raamSoftwire: Yep, I missed that. Cheers @duncanjbrown!

PR #4454, opened 2021-04-07
Update audited gem

malcolmbaig:
## Context
Our fork of audited at https://github.com/DFE-Digital/audited/ has been
updated with commits from upstream, including changes to support Rails
6.1.

Update the gem so that the lock file is pointing to the appropriate SHA.

Ideally we'd use the original gem, but this will break the mailer previews index page until https://github.com/collectiveidea/audited/pull/532 is merged.

## Guidance to review
- Audits can be tested by changing a candidate's application form in the review app support interface and then checking the history.

## Link to Trello card
https://trello.com/c/EHl7AXlh

## Things to check

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,9 +1,9 @@
 GIT
   remote: https://github.com/DFE-Digital/audited
-  revision: 9b3c366573712801ae43b04e80f827c78397192e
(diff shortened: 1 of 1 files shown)


stevehook (approved):
Looks good and review app works as expected.

PR #4455, opened 2021-04-07
Use GovukLinkHelper helpers from govuk-components

paulrobertlloyd:
## Context

The `govuk-components` gem provides the following helpers:

* `govuk_link_to`
* `govuk_mail_to`
* `govuk_button_to`

We have our own helpers that replicate this functionality. In addition, we have a `govuk_button_link_to` helper which styles links as buttons. The same can be achieved by using the `govuk_link_to` helper from `govuk-components`, with the `button` parameter set to `true`.

* Remove `govuk_link_to` helper and tests
* Remove `govuk_button_to` helper and tests
* Refactor `bat_contact_mail_to` helper to call `govuk_mail_to`
* Refactor the feature flag button toggles into a component – this is because the `govuk_button_to` helper in `govuk-components` does not accept block content. Hopefully this is a bit drier as a result.

I also reviewed the usage of these link helpers:

* Ensure we are always using the `bat_contact_mail_to` helper
* Ensure we are always using the `govuk_link_to` helper
* Ensure we are always using the `govuk_mail_to` helper

This means that there are no places in the application where `.govuk-link` is manually set. Actually, that’s not true… it still appears in the phase banner text, but can review this when refactor the phase banner to use the respective component from `govuk-components`.

## Guidance to review

* Could the `FeatureToggleComponent` be simplified or improved?
* Should I add new tests for `FeatureToggleComponent`, or move existing tests for the view into component tests?
* Are we happy to use `govuk_link_to` with `button: true` as opposed to a dedicated helper?

## Things to check

### DIFF
--- a/app/views/provider_interface/provider_agreements/data_sharing_agreement.html.erb
+++ b/app/views/provider_interface/provider_agreements/data_sharing_agreement.html.erb
@@ -19,97 +19,97 @@
     <h2 class="govuk-heading-m">Contents</h2>
     <ol class="app-contents-list__list">
       <li class="app-contents-list__list-item app-contents-list__list-item--parent">
-        <a class="govuk-link app-contents-list__link" href="#section-1">Section 1: Introduction</a>
+        <%= govuk_link_to 'Section 1: Introduction', '#section-1', class: 'app-contents-list__link' %>
         <ol class="app-contents-list__nested-list">
           <li class="app-contents-list__list-item app-contents-list__list-item--dashed">
-            <a class="govuk-link app-contents-list__link" href="#section-1-1">Background</a>
+            <%= govuk_link_to 'Background', '#section-1-1', class: 'app-contents-list__link' %>
           </li>
         </ol>
       </li>
       <li class="app-contents-list__list-item app-contents-list__list-item--parent">
-        <a class="govuk-link app-contents-list__link" href="#section-2">Section 2: What personal data DfE and the provider process/share and why</a>
+        <%= govuk_link_to 'Section 2: What personal data DfE and the provider process/share and why', '#section-2', class: 'app-contents-list__link' %>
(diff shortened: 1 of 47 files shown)


app/views/provider_interface/provider_agreements/data_sharing_agreement.html.erb
```diff
@@ -19,97 +19,97 @@
     <h2 class="govuk-heading-m">Contents</h2>
     <ol class="app-contents-list__list">
       <li class="app-contents-list__list-item app-contents-list__list-item--parent">
-        <a class="govuk-link app-contents-list__link" href="#section-1">Section 1: Introduction</a>
+        <a class="app-contents-list__link" href="#section-1">Section 1: Introduction</a>
```
stevehook, line 22: Are these left as regular markup rather than converted to `govuk_link` because they are navigating to the same page?

stevehook (approved):
Looks good. Some questions inline but good to merge.

app/views/provider_interface/provider_agreements/data_sharing_agreement.html.erb
paulrobertlloyd: These could use the `govuk_link_to` helper, but would be far more verbose, i.e. instead of:

```html
<a class="govuk-link app-contents-list__link" href="#section-1">Section 1: Introduction</a>
```

we could write:

```erb
<%= govuk_link_to 'Section 1: Introduction', '#section-1', class: 'app-contents-list__link' %>
```

Hmmm, on reflection, maybe that’s fine? (Arguably `app-contents-list` would be a separate component, but beyond the scope of this PR)

paulrobertlloyd: Updated to use `govuk_link_to` helper, as used elsewhere for contents lists.

PR #4456, opened 2021-04-07
Use govuk_skip_link helper

paulrobertlloyd:
## Context

Use [`govuk_skip_link` helper](https://dfe-digital.github.io/govuk-components/#skip-link) from `govuk-components`.

### DIFF
--- a/app/views/layouts/base.html.erb
+++ b/app/views/layouts/base.html.erb
@@ -54,7 +54,7 @@
       <% end %>
     <% end %>
 
-    <a href="#main-content" class="govuk-skip-link">Skip to main content</a>
+    <%= govuk_skip_link %>
 
     <%= render 'layouts/header' %>
 


benswannack (approved):
Looks good, I can't see any change in behaviour in the review app 👍

app/views/layouts/base.html.erb
```diff
@@ -54,7 +54,7 @@
       <% end %>
     <% end %>
 
-    <a href="#main-content" class="govuk-skip-link">Skip to main content</a>
```
benswannack, line 57: Just wondering, should this be the first focussed thing when tabbing? At the moment, if the cookie banner is shown, then its buttons are focussed first

paulrobertlloyd: I would have thought so too, but [the guidance says](https://design-system.service.gov.uk/components/cookie-banner/#how-it-works):

> Position the cookie banner after the opening <body> tag and before the ‘skip to main content’ link.

PR #4459, opened 2021-04-08
Add subject_code to qualifications in API

benswannack:
## Context
Vendors have asked for this to make it easier to create sensible qualification records at their end

Add a subject_code field to qualifications in the API. This will currently only be populated for GCSEs and A-levels where a mapping is known (we don't do anything clever, we just lean on the autocomplete).

## Guidance to review
Is the documentation good?
Do we need to think about how to allow for future changes to be easier?

## Link to Trello card
https://trello.com/c/oUJWNqkb/3556-%F0%9F%8F%88-map-a-level-and-gcse-subjects-to-hecos-codes-in-the-api-where-possible

## Things to check


app/presenters/vendor_api/single_application_presenter.rb
```diff
@@ -255,6 +261,59 @@ def qualification_to_hash(qualification)
...
+      #   These are either maths, science or english GCSEs that are required by some courses
+      #   - Maths GCSEs are not structured, so we map to a single maths subject code
+      #   - Science GCSEs are structured, but we combine them into one in the API response, so we can only be as granular as 'science'
+      #   - English GCSEs are now structured, but they did not used to be. We map to a generic english code here for the unstructured ones,
+      #     and have a better mapping within the structured_gcse_to_hashes method for newer english GCSEs
+      #
+      # Qualification level: other
+      #   These are other qualifications that a candidate has that are not direct requirements for the course.
+      #   This includes GCSEs and A levels, and none of them are structured. If their subject has been chosen
+      #   directly from the autocomplete, then we can map to a code. Otherwise, we leave the code blank.
```
benswannack, line 277: Is this comment useful here?

I often find qualifications really difficult to deal with because they are so inconsistent with how they hold data. We could add better documentation in a separate place, but I don't think it would be very visible.

config/vendor-api-v1.yml
```diff
@@ -831,6 +832,11 @@ components:
           description: The subject studied
           maxLength: 256
           example: History and Politics
+        subject_code:
+          type: string
+          description: The HECoS code for the subject studied, if available
```
benswannack, line 837: Is the specificity of saying HECoS useful here? Are they going to be extensible to all types of qualification, or just limited to GCSEs and A levels?

app/controllers/vendor_api/reference_data_controller.rb
```diff
@@ -1,11 +1,11 @@
 module VendorAPI
   class ReferenceDataController < VendorAPIController
     def gcse_subjects
-      render json: { data: GCSE_SUBJECTS }
+      render json: { data: GCSE_SUBJECTS.map { |subject| subject[:name] } }
```
cwrw, line 4: Instead of mapping the `:name` keys of both `GCSE_SUBJECTS` and `A_AND_AS_LEVEL_SUBJECTS` constants, we can instead define new constants with the key mappings and define `GCSE_SUBJECTS` and `A_AND_AS_LEVEL_SUBJECTS` in their respective classes as so:
```
GCSE_SUBJECTS = GCSE_SUBJECTS_TO_CODES.map { |subject| subject[:name] }

A_AND_AS_LEVEL_SUBJECTS = A_AND_AS_LEVEL_SUBJECTS_TO_CODES..map { |subject| subject[:name] }
```
This way we wrap the old constant definitions and won't have to amend every instance of their occurrence, what do you think?

app/presenters/vendor_api/single_application_presenter.rb
```diff
@@ -255,6 +261,59 @@ def qualification_to_hash(qualification)
...
+
+      if qualification.gcse?
+        subject_code_for(qualification.subject)
+      elsif qualification.other?
+        subject_code_for_other_qualification(qualification)
+      end
+    end
+
+    def subject_code_for(subject)
+      case subject
```
cwrw, line 287: Im quite unsure about the different subjects and mappings here, let's have a chat in person. 

But my main concern is this logic should not live here, and be extracted into the `ApplicationQualification` and `A_AND_AS_LEVEL_SUBJECTS_TO_CODES` etc.

app/presenters/vendor_api/single_application_presenter.rb
```diff
@@ -255,6 +261,59 @@ def qualification_to_hash(qualification)
...
+      when 'english', 'english_single_award', 'english_double_award', 'english_studies_single_award', 'english_studies_double_award'
+        '100320'
+      when 'english_language'
+        '100318'
+      when 'english_literature'
+        '100319'
+      end
+    end
+
+    def subject_code_for_other_qualification(qualification)
```
cwrw, line 301: I think we can simplify this method further. Maybe if we defined the constants as the key being the name of the qualification and the code being the hecos code, maybe we can simplify the method as so:
```
if qualification.qualification_type == 'GCSE'
  GCSE_SUBJECTS[qualification.subject]
elsif ['A level', 'AS level'].include? qualification.qualification_type
  A_AND_AS_LEVEL_SUBJECTS[qualification.subject]
end
```
If not we can still avoid returning many times in the method

config/vendor-api-v1.yml
cwrw: Might be better to be explicit and add GCSE and A levels, but something to check with James I think to be sure

app/presenters/vendor_api/single_application_presenter.rb
cwrw: I would avoid large comments in the code. I think either have something in the commit message, as well as a smaller comment across the code relating to each subject, or maybe having a rethink of the code itself and if we can make it less complex/self documenting to help. But happy to have a chat around this specific case as Im not fully following the different cases

app/controllers/vendor_api/reference_data_controller.rb
benswannack: Ok this sounds better 👍

cwrw (approved):
Looks good, thanks for addressing comments. Could you remove the extra file `config/initializers/qualifications/gcse_subjects_to_codes.rb`?

PR #4462, opened 2021-04-08
Fix export history bug 2

raamSoftwire:
## Context

Only a small subset of export history records are being shown.

Only data_export records with an export_type are shown. This field is automatically filled for all new export downloads. We wrote a migration to backfill the old data_export records but this migration failed for a couple of reasons.

1. Some of the exports have been renamed or removed. Eg. `Unexplained breaks in work history` was renamed `Work history break`. `RejectedCandidatesExport` was removed.
2. The `export_type` is not equal to `export.name.parameterize.underscore` in all cases so that caused the migration to fail. Eg. 

```
    sites_export: {
      name: 'Sites',
      export_type: 'sites_export',
      description: 'A list of sites that are being synced from Find, along with distances to their respective providers.',
      class: SupportInterface::SitesExport,
    }
```

This follows on from [this PR](https://github.com/DFE-Digital/apply-for-teacher-training/pull/4448). The previous PR solved one instance of the first problem above but didn't catch them all.

For this PR, I started by getting all the export names from the `data_exports` table on prod by running `SELECT DISTINCT "name" FROM "data_exports";` in Blazer.

```
Active provider user permissions
Active provider users
Application references
Application timings
Applications for TAD
Candidate application feedback
Candidate autosuggest usage
Candidate course choice withdrawal survey
Candidate email send counts
Candidate feedback
Candidate journey tracking
Candidate survey
Daily export of applications for TAD
Daily export of notifications breakdown
Equality and diversity data
Interview changes
Locations
Locations export
Notes
Notifications
Offer conditions
Organisation permissions
Persona data
Provider Access Controls
Provider performance for TAD
Providers
Qualifications
Referee survey
RejectedCandidatesExport
Sites
Structured reasons for rejection
Submitted application choices
Unexplained   breaks in work history
User permissions
Work history break
```

I was then able to work through the git history to identify which exports still exist, but have been renamed and which exports no longer exist. I have added a `case` statement to handle these accordingly.

To solve the second problem, I updated the logic so that it looks up the `export_type` from `DataExport::EXPORT_TYPES` rather than assuming it can be calculated from the `name`.

In order to test that all the cases are handled I added some tests, but also manually added a record in to my local DB for each of the names seen on the prod db. I was then able to run the migration locally and check that all cases have been handled.

image

## Link to Trello card

https://trello.com/c/K5o3HiH8/3258-fix-export-history-bug

## Things to check

### DIFF
--- a/app/services/data_migrations/backfill_export_type.rb
+++ b/app/services/data_migrations/backfill_export_type.rb
@@ -24,7 +24,7 @@ module DataMigrations
       when 'Candidate survey', 'Daily export of applications for TAD', 'Daily export of notifications breakdown', 'RejectedCandidatesExport'
         nil
       else
-        export.name.parameterize.underscore
+        DataExport::EXPORT_TYPES.select { |_key, value| value[:name] == export.name }.values[0][:export_type]
       end
     end
   end
--- a/app/services/data_migrations/backfill_export_type.rb
+++ b/app/services/data_migrations/backfill_export_type.rb
@@ -6,10 +6,25 @@ module DataMigrations
     def change
       data_exports = DataExport.all.where(export_type: nil)
 
-      data_exports.each do |export|
-        export_type = export.name == 'Unexplained breaks in work history' ? 'work_history_break' : export.name.parameterize.underscore
+      data_exports.each { |export| export.update!(export_type: export_type(export)) }
+    end
 
-        export.update!(export_type: export_type)
+  private
+
+    def export_type(export)
+      case export.name
+      when 'Unexplained breaks in work history'
+        'work_history_break'
+      when 'Locations', 'Locations export'
+        'persona_export'
+      when 'Applications for TAD'
+        'tad_applications'
+      when 'Provider performance for TAD'
+        'tad_provider_performance'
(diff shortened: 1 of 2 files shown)


app/services/data_migrations/backfill_export_type.rb
```diff
@@ -6,10 +6,25 @@ class BackfillExportType
...
+      case export.name
+      when 'Unexplained breaks in work history'
+        'work_history_break'
+      when 'Locations', 'Locations export'
+        'persona_export'
+      when 'Applications for TAD'
+        'tad_applications'
+      when 'Provider performance for TAD'
+        'tad_provider_performance'
+      when 'Candidate survey', 'Daily export of applications for TAD', 'Daily export of notifications breakdown', 'RejectedCandidatesExport'
```
raamSoftwire, line 24: These exports no longer exist, so it is OK to set these as `nil`

despo (approved):
👍🏼

davidgisbey (approved):
Just had a good look through the code and it maps up nicely to your spreadsheet/blazer. When we merge this in let's double-check the data in QA, but it looks good to me

PR #4466, opened 2021-04-08
Refactor tests to use Time.zone.now or CycleTimetableHelper

davidgisbey:
## Context

A large number of our tests are frozen to specific dates, normally the 1/1/2020 when they don't need to be. Ideally, we should either be using a randomised time within a discreet phase of the cycle where appropriate (between the apply1 and apply2 deadlines etc.) or we should just be freezing at the current time. 

By setting the bulk of our tests that require Timecop to a specific date we may well be masking legitimate bugs in our system. 

There are a couple of downsides to the work in this PR. Firstly, time-related tests that freeze on the current time can tend to be a bit flakey. Secondly, they're definitely a little harder to parse. Although, I don't think many/any of the changes in this PR are particularly difficult to parse. It's definitely less clear than a sentence reads. 

For example, 

`expect(result.text).to include('Delete entry for break between Feb 2019 and Apr 2019')`

becomes 

` expect(result.text).to include("Delete entry for break between #{8.months.ago.to_s(:short_month_and_year)} and #{6.months.ago.to_s(:short_month_and_year)}")`

- Adds a helper which returns a randomised time for discreet phases of the cycle
- - Where possible, uses the current time when Timecop is used
- Uses the helper to set the time for the appropriate tests (largely tests related to the EOC

## Guidance to review

I'm definitely not the best at avoiding flaky tests for time-related tests. On other projects, I've largely used the to be_within matcher to avoid issues with nanosecond drift. 

If anyone has any tips/ wants to make any amends, please do feel free. 

Also, if people feel like this isn't the way to go that's fine too.

In terms of reviewing, i've tried to keep the tests for each UI in separate commits. It might be worth picking off the tests that you're most familiar to begin with.

Once i've got some feedback I'll apply it.

One thing i think we could potentially implement is a helper like this, which applies microsecond precision rather than nanoseconds 

image

## Link to Trello card

https://trello.com/c/4SPscuL1/3189-refactor-specs-to-be-cycle-aware

## Things to check


spec/services/support_interface/candidate_journey_tracker_spec.rb
```diff
@@ -1,7 +1,7 @@
 require 'rails_helper'
 
 RSpec.describe SupportInterface::CandidateJourneyTracker, with_audited: true do
-  let!(:now) { Time.zone.now }
+  let!(:now) { Time.zone.now.change(usec: 0) }
```
davidgisbey, line 4: These tests were failing by nanoseconds. This works, but probably isn't the best solution. Normally, i'd just use the to be_within(1.second).of(Time.zone.now) matcher, but i was interested to see what people thought.

spec/support/test_helpers/cycle_timetable_helper.rb
```diff
@@ -0,0 +1,39 @@
...
+  def after_find_closes
+    rand((current_end_of_cycle_timetable[:find_closes])..current_end_of_cycle_timetable[:find_reopens]).midday
+  end
+
+  def after_find_reopens
+    rand((current_end_of_cycle_timetable[:find_reopens])..current_end_of_cycle_timetable[:apply_reopens]).midday
+  end
+
+  def after_apply_reopens
+    rand((current_end_of_cycle_timetable[:find_reopens])..Date.new(EndOfCycleTimetable::CURRENT_YEAR_FOR_SCHEDULE, 12, 31)).midday
```
duncanjbrown, line 27:
```suggestion
    rand((current_end_of_cycle_timetable[:apply_reopens])..Date.new(EndOfCycleTimetable::CURRENT_YEAR_FOR_SCHEDULE, 12, 31)).midday
```

spec/system/candidate_interface/offers_and_withdrawals/candidate_receives_rejection_email_spec.rb
```diff
@@ -53,7 +52,7 @@ def when_i_have_a_single_offer
     @offer = create(:application_choice,
                     status: :offer,
                     application_form: @application_form,
-                    decline_by_default_at: 10.business_days.from_now,
+                    decline_by_default_at: 10.business_days.from_now.midday,
```
duncanjbrown, line 55: irl I think `TimeLimitCalculator` sets this to end_of_day?

duncanjbrown: The `:with_offer` trait ought to set this, but the DBD looks a bit dodgy

https://github.com/DFE-Digital/apply-for-teacher-training/blob/9a3cb721f5f9d6204259a9d4569ecf66d0ab0ee2/spec/factories/application_choice.rb#L155-L161

spec/services/carry_over_application_spec.rb
```diff
@@ -45,7 +46,7 @@ def original_application_form
 
   context 'when the application_form has references has an application_reference in the cancelled_at_end_of_cycle state' do
     around do |example|
-      Timecop.freeze(Time.zone.local(2020, 9, 19, 12, 0, 0)) do
+      Timecop.freeze(after_apply_2_deadline) do
```
duncanjbrown, line 49: 💎 this is 10 times better

spec/services/support_interface/candidate_journey_tracker_spec.rb
duncanjbrown: I like `be_within` — it makes it clear that time is still moving. But in this case (and the other `usec:` specifications) I don't think either is necessary; the tests still pass without the `.change` call? Or maybe there's something different between our environments

spec/system/candidate_interface/apply_again/candidate_carries_over_unsubmitted_application_between_cycles_spec.rb
```diff
@@ -2,16 +2,16 @@
...
   around do |example|
-    Timecop.freeze(Date.new(2020, 8, 1)) do
+    Timecop.freeze(mid_cycle) do
       example.run
     end
   end
 
   scenario 'Carry over application and remove all application choices when new cycle opens' do
     given_i_am_signed_in_as_a_candidate
-    and_i_am_in_the_2020_recruitment_cycle
```
duncanjbrown, line 14: 🚮

duncanjbrown (commented):
Props @davidgisbey — this is brilliant. I don't want to hit approve now as it's Friday, it's huge and other eyes should look over it. But there's so much improvement here.

I don't think the hit of not being able to specify literal strings for dates in specs is nearly as costly as the risk of not running specs in (relatively) real time. And seeing references to specific cycles go away from specs is really really great. 

I couldn't reproduce your `usec` woes locally — I found things worked fine without the call to `.change`. Anyone else have the same experience?

spec/services/support_interface/candidate_journey_tracker_spec.rb
davidgisbey: Yea they were passing locally it was failing on the builds.

spec/support/test_helpers/cycle_timetable_helper.rb
```diff
@@ -0,0 +1,39 @@
...
+private
+
+  def previous_end_of_cycle_timetable
+    EndOfCycleTimetable::CYCLE_DATES[EndOfCycleTimetable::CURRENT_YEAR_FOR_SCHEDULE - 1]
+  end
+
+  def current_end_of_cycle_timetable
+    EndOfCycleTimetable::CYCLE_DATES[EndOfCycleTimetable::CURRENT_YEAR_FOR_SCHEDULE]
+  end
+end
```
malcolmbaig, line 39: 👌🏼 this entire module

spec/services/support_interface/candidate_journey_tracker_spec.rb
malcolmbaig: Getting the same issue locally. It looks like it's caused by a difference in precision between in-memory time and database time.

I like the idea outlined in the PR description re: a `current_time` helper. Interested to hear other opinions.

spec/support/test_helpers/cycle_timetable_helper.rb
```diff
@@ -0,0 +1,39 @@
+module CycleTimetableHelper
```
stevehook, line 1: I think this is great. One question is whether it's safe to call any of these methods twice in the same test? I guess you'd get different (random) results? I wonder if it's useful to memoize the results per spec, i.e. memoize all these methods but reset the state `after` each.

stevehook (approved):
:100: I think this is great and will really help when it comes to the next end of cycle. I've got one query inline but I don't see any reason why we shouldn't merge subject to comments from others.

duncanjbrown (approved):
+1 for the `current_time` helper

malcolmbaig (approved):

spec/services/support_interface/candidate_journey_tracker_spec.rb
cwrw: I also like the `current_time` helper to mitigate this issue, which we can use across problematic specs. I'm curious how often this is happening, is it just this spec or across others (couldn't replicate locally)?

cwrw (approved):
This is great, and the approach very sensible ⭐  hoping to do another round of reviewing but as an initial walkthrough 👍

davidgisbey:
Thanks for all the feedback! I'm going to apply the changes and merge. Then pick up adding the helper in a separate PR.

Abeer, there were a few that were failing on CI and a couple of others on the build.

spec/services/support_interface/candidate_journey_tracker_spec.rb
davidgisbey: there were a few that were failing on CI and a couple of others on the build. They were failing 100% of the time too.

spec/forms/provider_interface/interview_wizard_spec.rb
```diff
@@ -252,7 +252,7 @@
       expect(wizard.send('date(1i)')).to eq(interview.date_and_time.year.to_s)
       expect(wizard.location).to eq(interview.location)
       expect(wizard.provider_id).to eq(interview.provider_id)
-      expect(wizard.time).to eq(interview.date_and_time.strftime('%l:%M%P'))
+      expect(wizard.time).to eq(interview.date_and_time.strftime('%l:%M%P').gsub(' ', ''))
```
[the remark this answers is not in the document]
davidgisbey: I wish i could tell you why i did it haha. I've removed it. I'm interested to see if it fails on CI.

spec/support/test_helpers/cycle_timetable_helper.rb
davidgisbey: That's a really good point Steve. I'll sort that now!

spec/forms/provider_interface/interview_wizard_spec.rb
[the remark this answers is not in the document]
benswannack: Might be a flaky one - we render the time as `strftime('%-l:%M%P')`, which gets rid of the empty space padding on single digit hours. We probably want to match that format in the spec :)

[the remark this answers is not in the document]
davidgisbey: Amazing! Thanks i've updated it.

spec/system/candidate_interface/offers_and_withdrawals/candidate_receives_rejection_email_spec.rb
davidgisbey: Yea good spot. This should be using business days. Will sort now 👍

PR #4467, opened 2021-04-08
Replace NotificationMessageComponent with GovukComponent::NotificationBanner

paulrobertlloyd:
## Context

Use `GovukComponent:: NotificationBanner` to render notification banners (via `govuk_notification_banner` helper method).

* Remove `NotificationMessageComponent` component, tests and preview
* Refactor `FlashMessageComponent` to use `GovukComponent::NotificationBanner`

## Guidance to review

The way the component from `govuk-components` works is significantly different from the component we are using. In most cases, the new component offers enough flexibility to achieve the same rendered markup.

However, the flash banner needed a little refactoring so that the data passed to it can be correctly passed onto the new component. Curious to know if the way I’ve achieved this is desirable.

## Things to check


spec/components/notification_message_component_spec.rb
```diff
@@ -1,111 +0,0 @@
-require 'rails_helper'
-
-RSpec.describe NotificationMessageComponent do
```
stevehook, line 3: Do any of these tests now belong in `FlashMessageComponent`? (There is some extra logic in `FlashMessageComponent` that we're not testing other than in system specs).

stevehook (commented):
Tested locally and all seems good. I just had one question about tests. Otherwise good to merge I reckon.

app/controllers/concerns/cookie_preference_concerns.rb
```diff
@@ -20,18 +20,4 @@ def cookie_preferences_consent
...
-
-    if previous_referer_uri && request_uri.eql?(cookie_page_path) && !previous_referer_uri.eql?(cookie_page_path)
-      flash[:success] = 'Your cookie preferences have been updated.'
-      flash[:link] = {
-        text: 'Go back to the page you were looking at',
-        url: session[:previous_referer],
-      }
-      session.delete(:previous_referer)
-    end
-  end
```
paulrobertlloyd, line 36: @TobyRet @stevehook @duncanjbrown Removing this code because, since shipping the new cookie banners, you’d only ever see the success banner should you change your cookie preferences on this page. Thus, ‘the page you were looking at’ is likely to be a.) wrong and b.) not relevant.

Given that this is the only place we then add a link to a flash message, I have removed this functionality from the `FlashMessageComponent`.

app/components/candidate_interface/apply_again_banner_component.html.erb
```diff
@@ -1,3 +1,4 @@
-<%= render NotificationMessageComponent.new(:info,
-                                            message: "#{all_withdrawn_copy} #{govuk_link_to('Do you want to apply again?', start_path, class: 'govuk-notification-banner__link')}",
-                                            secondary_message: deadline_copy) %>
+<%= govuk_notification_banner(title: t('notification_banner.info')) do |notification_banner| %>
+  <%= notification_banner.slot(:heading, text: heading) %>
+  <%= body %>
+<% end %>
```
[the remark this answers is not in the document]
paulrobertlloyd: This component wasn’t deleted, but it should have been, so have done so in this commit: 237657f7b25f92699bae2da8e83447b91a67530f

spec/components/notification_message_component_spec.rb
paulrobertlloyd: The only extra tests needed related to adding a link. But I’ve removed this functionality in 91b8d5be7c857a60ac240d4c8dcab98d2506ee96 because the one place we were using it, turns out we no longer need it.

app/controllers/concerns/cookie_preference_concerns.rb
TobyRet: I'm not sure if `the page you were looking at` is likely to be a.) wrong and b.) not relevant.

Seems to work ok for me locally. For example if I am in the references section of the application then update my cookies, the `the page you were looking at` link does redirect me back to the references section correction. 

Whether you want to have this link (along with the flash message) is, I guess, a design decision?

paulrobertlloyd: Yeah, I’m questioning the need for a back link at all, given you have to now explicitly visit this page and change the cookie settings for this flash banner to show.

paulrobertlloyd: Let me merge in master, now that the JS-powered banner has shipped. Might make more sense.

paulrobertlloyd: Try it now: https://apply-for-te-govuk-noti-rdrv6o.herokuapp.com/

duncanjbrown: OK, got it. Much better! 

I find this message bit odd when JS is disabled:

`To accept or reject cookies, turn on JavaScript in your browser settings or reload this page.` Reloading the page doesn't do anything!

duncanjbrown (approved):

app/controllers/concerns/cookie_preference_concerns.rb
paulrobertlloyd: https://design-system.service.gov.uk/components/cookie-banner/#if-the-user-s-device-is-not-running-javascript

I guess reloading the page may fix a banner not loading in some scenarios.

PR #4470, opened 2021-04-09
Add `iso8601` to timestamp in notes export to avoid flakey specs

cwrw:
## Context

We're getting spec failures due to date precision in the notes export

use `iso8601` to avoid any random failures

## Things to check

### DIFF
--- a/app/services/support_interface/notes_export.rb
+++ b/app/services/support_interface/notes_export.rb
@@ -25,7 +25,7 @@ module SupportInterface
         {
           note_subject: note.subject,
           note_message: note.message,
-          note_created_at: note.created_at,
+          note_created_at: note.created_at.iso8601,
           candidate_id: note.application_choice.application_form.candidate.id,
           provider_code: provider&.code,
           provider_name: provider&.name,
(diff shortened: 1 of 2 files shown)


stevehook (commented):
I saw this late yesterday and struggled to work out what the cause was. I'm not sure `contains_exactly` will give a different result (it still fails for me locally 1/4 times - in my not very scientific test). Is `contains_exactly` not just a synonym of `matches_array`? (https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/contain-exactly-matcher).

I was wondering whether it's a `created_at` date precision thing (but that's pure guesswork). It might be good to just use `iso8601` formatted date times for exports in any case.

cwrw:
@stevehook Good point, after I opened the PR I was looking at the failures and diffing them out, and wasn't also able to find why it was failing actually. The timestamps seem the same as well (in the diff) and in the implementation. Will do more investigation

cwrw:
update: using `iso8601` has fixed the issue. Tested with master and the branch:
`for i in {1..5}; do be rspec spec/services/support_interface/notes_export_spec.rb; done` and on master failed 3 out of 5 times vs none with the fix, cheers @stevehook

stevehook (approved):
:+1:

cwrw:
apparently this didn't get fixed however I've got a lead now, will open another pr shortly

PR #4474, opened 2021-04-09
Add Candidate ID in manage

duncanjbrown:
## Context

Providers with student record systems aren't able to cross-reference the Candidate ID we send them with the IDs they see in Manage. It could be that through duplicate matching they've attached that ID to an existing candidate, whether from a previous cycle or from a UCAS application.

Add the Candidate ID in Manage.

Screenshot 2021-04-09 at 15 26 49

## Link to Trello card

https://ukgovernmentdfe.slack.com/archives/C01EZFNUPLP/p1616678101044700?thread_ts=1616678080.044400&cid=C01EZFNUPLP

### DIFF
--- a/app/components/provider_interface/personal_details_component.rb
+++ b/app/components/provider_interface/personal_details_component.rb
@@ -32,6 +32,7 @@ module ProviderInterface
         phone_number_row,
         email_row,
         address_row,
+        candidate_id_row,
       ].compact
     end
 
@@ -101,6 +102,13 @@ module ProviderInterface
       {
         key: 'Address',
         value: application_form.full_address,
+      }
+    end
+
+    def candidate_id_row
+      {
+        key: 'Candidate ID',
+        value: candidate.public_id,
       }
     end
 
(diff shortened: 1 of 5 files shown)


app/presenters/vendor_api/single_application_presenter.rb
```diff
@@ -27,7 +27,7 @@ def as_json
           reject_by_default_at: application_choice.reject_by_default_at&.iso8601,
           recruited_at: application_choice.recruited_at,
           candidate: {
-            id: "C#{application_form.candidate.id}",
+            id: application_form.candidate.public_id,
```
cwrw, line 30: can you also mirror this in the register API (I'll add a ticket to refactor this so we can pick it up soon)

app/components/provider_interface/personal_details_component.rb
```diff
@@ -32,6 +32,7 @@ def rows
         phone_number_row,
         email_row,
         address_row,
+        candidate_id_row,
```
cwrw, line 35: not sure if we want to mirror this in the support application details component, although support does surface the candidate id elsewhere on that page 🤔

cwrw (approved):
Looks good, some minor comments

app/presenters/vendor_api/single_application_presenter.rb
duncanjbrown: Yes — done!

app/components/provider_interface/personal_details_component.rb
duncanjbrown: I think it's ok as it is — right up at the top of the page in Support

PR #4477, opened 2021-04-09
Present Register API responses in Support

duncanjbrown:
## Context

We serve applications over two APIs now — render both versions in Support

- refactor out a pointless component
- add the Register API view next to the existing Vendor API view

Screenshot 2021-04-09 at 17 26 43

Screenshot 2021-04-09 at 17 26 59


app/lib/allowed_cross_namespace_usage.rb
```diff
@@ -6,4 +6,5 @@
 # reuse.
 module AllowedCrossNamespaceUsage
```
cwrw, line 7: TIL

cwrw (approved):

PR #4478, opened 2021-04-11
Change how references are retrieved in CandidateJourneyTracker

malcolmbaig:
## Context
The "completed references" fields of this export are supposed to show
timestamps for the first and second reference received for a given
application. Currently the export retrieves these references ordered by
id, irrespective of when they were actually received.

Further, some references in the production database have blank
feedback_provided_at timestamps, leading to blank fields in the export.

Address both of these issues as follows:

- Update the reference retrieval logic so that all `feedback_provided`
  references are returned ordered by their `feedback_provided_at`
  timestamp
- If a `feedback_provided` reference has a missing
  `feedback_provided_at` timestamp, rather than show an incomplete and
  potentially confusing/misleading picture in the export, opt instead
  to not display any reference data for that application. This allows
  the application to be easily filtered out of any reference-related analysis.

## Guidance to review
Can just review the diff, although the spec changes are probably best reviewed in split mode.

## Link to Trello card
https://trello.com/c/ls2bG8KZ

## Things to check

### DIFF
--- a/app/services/support_interface/candidate_journey_tracker.rb
+++ b/app/services/support_interface/candidate_journey_tracker.rb
@@ -143,7 +143,16 @@ module SupportInterface
     end
 
     def received_references
-      @received_references ||= all_references.select(&:feedback_provided?)
+      @received_references ||=
+        begin
+          references = all_references.unscoped.feedback_provided
+
+          if references.any? { |ref| ref.feedback_provided_at.blank? } # an error state caused by bad data
+            [] # show nothing about references in the export so that the application can be easily ignored/filtered out
+          else
+            references.order(feedback_provided_at: :asc)
+          end
+        end
     end
 
     def earliest_update_audit_for(model, attributes)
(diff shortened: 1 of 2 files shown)


app/services/support_interface/candidate_journey_tracker.rb
```diff
@@ -143,7 +143,16 @@ def all_references
     end
 
     def received_references
-      @received_references ||= all_references.select(&:feedback_provided?)
+      @received_references ||=
+        begin
+          references = all_references.unscoped.feedback_provided
+
+          if references.any? { |ref| ref.feedback_provided_at.blank? } # an error state caused by bad data
+            [] # show nothing about references in the export so that the application can be easily ignored/filtered out
```
malcolmbaig, line 151: We should assess if it's possible to fix the data (I'll write up a card for this), but in the meantime this seems like the simplest thing to do. Milan is already filtering out applications with blank `feedback_provided_at`.

davidgisbey (approved):

PR #4483, opened 2021-04-12
Use GovukComponent::PhaseBanner

paulrobertlloyd:
## Context

Use `GovukComponent::PhaseBanner` to render phase banner (via `govuk-components` `govuk_phase_banner` helper method).

* Update `govuk-components` to latest version (v1.1.7) to enable this change
* Use `govuk_link_to` helper for phase banner links
* Change where the default value for `feedback_link` is provided (in default param, not in `govuk_link_to` helper)
* Change `HostingEnvironment.phase_colour` to return string values for colour name

## Guidance to review

* Is it better to set the `feedback_link` default in this way, or was the previous approach preferred?

## Things to check


app/components/phase_banner_component.rb
```diff
@@ -1,19 +1,19 @@
 class PhaseBannerComponent < ViewComponent::Base
   DEFAULT_FEEDBACK_LINK = 'mailto:becomingateacher@digital.education.gov.uk?subject=Feedback%20about%20Apply%20for%20teacher%20training'.freeze
 
-  def initialize(no_border: false, feedback_link: nil)
+  def initialize(no_border: false, feedback_link: DEFAULT_FEEDBACK_LINK)
```
duncanjbrown, line 4: This is better than the original code... which I wrote! 😭

duncanjbrown (approved):

app/components/phase_banner_component.rb
paulrobertlloyd: Don’t worry @duncanjbrown, I won’t tell anyone. Much.

app/components/phase_banner_component.html.erb
```diff
@@ -1,10 +1,11 @@
...
-    <div class="govuk-phase-banner govuk-!-display-none-print">
-      <p class="govuk-phase-banner__content">
-        <%= govuk_tag(text: HostingEnvironment.phase, classes: "govuk-phase-banner__content__tag #{phase_tag_class}") %>
-        <span class="govuk-phase-banner__text"><%= text %></span>
-      </p>
-    </div>
+    <%= govuk_phase_banner(
+      phase_tag: {
+        text: HostingEnvironment.phase,
+        colour: HostingEnvironment.phase_colour,
```
paulrobertlloyd, line 6: @duncanjbrown This is not going to return a value if the hosting environment is `production`. Will this break anything (and how can I test that?)

paulrobertlloyd: I’m guessing it will return `nil`? In which case, nothing will break.

duncanjbrown: `nil` is safe here — you can see the `phase_banner_component_spec` already runs itself in "Production" by setting an environment variable in an `around` block

paulrobertlloyd: Cool, will merge.

PR #4486, opened 2021-04-12
Upgrade to Rails 6.1

malcolmbaig:
## Context
Where possible we should be using the latest version of Rails.

- Fix errors found by our test suite
- Fix several (but not all) deprecation warnings

## Guidance to review
- Per-commit recommended
- A quick click around the review app to ensure things are working as expected would be helpful
- There are still a few deprecation warnings left, mostly to do with changes to the ActiveModel::Errors API. There's probably enough going on in this PR, will fix remaining warnings in a followup.
- Can anyone figure out why 4940f222f1ee61a59c11ee076df9cade0dada606 is necessary?

## Link to Trello card
https://trello.com/c/HCpYZuT0

## Things to check


config/application.rb
```diff
@@ -52,7 +52,7 @@ class Application < Rails::Application
     config.action_view.default_form_builder = GOVUKDesignSystemFormBuilder::FormBuilder
 
     config.i18n.exception_handler = Proc.new { |exception| raise exception.to_exception }
-    config.action_view.raise_on_missing_translations = true
+    config.i18n.raise_on_missing_translations = true
```
duncanjbrown, line 55: Does this make https://github.com/DFE-Digital/apply-for-teacher-training/pull/4220 redundant?

duncanjbrown (approved):
Thanks @malcolmbaig!  Nice commits too 💎. Review app seems ok to me.

config/application.rb
malcolmbaig: Unfortunately still need the custom exception handler to also raise errors in the case of activemodel errors that don't resolve correctly and don't have a default (eg — errors for custom validations).

malcolmbaig:
Updated the commit message for https://github.com/DFE-Digital/apply-for-teacher-training/pull/4486/commits/b3c11b0f240ea6726b90db45637b3051b6d97131

PR #4487, opened 2021-04-12
Fix bug where predicted_grade is not set when entering degree

benswannack:
## Context
From a support request, a candidate has entered a degree and from what they can see it is complete. However, they get a validation error when trying to mark it as complete. See https://becomingateacher.zendesk.com/agent/tickets/19044

This occurs because they filled out the degree until the "Have you completed this degree?" page and then left the flow. They then re-entered the flow to see a review page that looks as though that section has been filled in, and never filled it in.

Handle the case where `predicted_grade` is `nil` in the db in the form and review pages

## Guidance to review
Have I missed anything?

### DIFF
--- a/app/components/candidate_interface/degrees_review_component.rb
+++ b/app/components/candidate_interface/degrees_review_component.rb
@@ -161,6 +161,8 @@ module CandidateInterface
     end
 
     def formatted_completion_status(degree)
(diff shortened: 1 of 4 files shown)


malcolmbaig (approved):
👍🏼

PR #4488, opened 2021-04-12
Add google tag manager code to apply and manage

steventux:
## Context

We'd like to track various events and page interactions using Google Tag Manager.

Adds environment variable config and JS for Apply and Manage to enable Google Tag Manager scripts. 
These will be separate environment variables for candidate/Apply and provider/Manage, if they are not set, the tag manager JS is not included in the page.

I've tested this locally using the GTM configuration for the Test Manage account:

image

## Guidance to review

@davidgisbey I've included config for Apply, which is possibly out of scope, will remove if you don't think this necessary.
Environment variables are configured in the same PR, this should be OK as the GTM script will not be included without the var but is that assumption correct?
 

## Link to Trello card

https://trello.com/c/88FPUaKz/3515-add-google-tag-manager-to-manage

## Things to check


app/controllers/provider_interface/provider_interface_controller.rb
```diff
@@ -33,7 +33,10 @@ def current_provider_user
     end
 
     def check_cookie_preferences
-      @google_analytics_id = ENV.fetch('GOOGLE_ANALYTICS_MANAGE', '') if cookies['consented-to-manage-cookies'].eql?('yes')
+      if cookies['consented-to-manage-cookies'].eql?('yes')
+        @google_analytics_id = ENV.fetch('GOOGLE_ANALYTICS_MANAGE', '')
+        @google_tag_manager_id = ENV.fetch('GOOGLE_TAG_MANAGER_MANAGE', '')
```
cwrw, line 38: do we want to add this to the .env files as well, similar to `GOOGLE_ANALYTICS_MANAGE`?

cwrw: We seem to be surfacing the analytics id on the cookies page: https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/views/content/cookies.html.erb#L114
I wonder if we should do the same for the tag manager (not sure) 🤔

app/controllers/candidate_interface/candidate_interface_controller.rb
```diff
@@ -18,7 +18,10 @@ def add_identity_to_log(candidate_id = current_candidate&.id)
     end
 
     def check_cookie_preferences
-      @google_analytics_id = ENV.fetch('GOOGLE_ANALYTICS_APPLY', '') if cookies['consented-to-apply-cookies'].eql?('yes')
+      if cookies['consented-to-apply-cookies'].eql?('yes')
+        @google_analytics_id = ENV.fetch('GOOGLE_ANALYTICS_APPLY', '')
+        @google_tag_manager_id = ENV.fetch('GOOGLE_TAG_MANAGER_APPLY', '')
```
cwrw, line 23: same question here around .env files and showing this value on the cookies page

app/views/shared/_analytics.html.erb
```diff
@@ -6,3 +6,13 @@
   gtag('js', new Date());
   gtag('config', '<%= @google_analytics_id %>');
 </script>
+
+<% if @google_tag_manager_id.present? %>
+<!-- Google Tag Manager -->
+<script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
```
cwrw, line 12: nit: can we move the js onto a new line after the script tag and indent as so:
```
<script>
  (function(w,d,s,l,i)...
</script>
```

app/views/shared/_analytics.html.erb
```diff
@@ -6,3 +6,13 @@
   gtag('js', new Date());
   gtag('config', '<%= @google_analytics_id %>');
 </script>
+
+<% if @google_tag_manager_id.present? %>
```
cwrw, line 10: I wonder if we should move the tag manager into its own layout, as to get to analytics we need to check if we have a `google_analytics_id` present: https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/views/layouts/base.html.erb#L4
or maybe we move the check for `google_analytics_id` inline here, so we keep the analytics in the same layout. Let me know your thoughts

azure/pipelines/templates/deploy.yml
```diff
@@ -192,7 +192,9 @@ jobs:
                 -govukNotifyCallbackAPIKey "${{parameters.govukNotifyCallbackAPIKey}}"
                 -sandbox "${{parameters.sandbox}}"
                 -googleAnalyticsApply "${{parameters.googleAnalyticsApply}}"
-                -googleAnalyticsManage "${{parameters.googleAnalyticsManage}}"'
+                -googleAnalyticsManage "${{parameters.googleAnalyticsManage}}"
+                -googleTagManagerApply "${{parameters.googleTagManagerApply}}"
```
cwrw, line 196: we are setting `googleAnalyticsApply` and `googleAnalyticsManage` earlier in the deploy template: https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/azure/pipelines/templates/deploy.yml#L69 should we do the same for `googleTagManagerApply` and `googleTagManagerManage`?

cwrw: Should we also add those in the `release.yml` for prod? https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/azure/pipelines/release.yml#L207

cwrw (commented):
looking good! thanks for tackling the analytics beast :D just a few comments otherwise 👍

azure/pipelines/templates/deploy.yml
steventux: :+1: The docs might be a bit thin on detail here https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/docs/environment-variables.md but I'll do what has been done for analytics to be sure.

app/views/shared/_analytics.html.erb
steventux: :+1: Much neater https://github.com/DFE-Digital/apply-for-teacher-training/pull/4488/commits/e167ade180a9c9495b30de4b508f33cc2aaaf3cf

paulrobertlloyd:
@steventux @cwrw What is the performance impact of this change? Am **_super_** wary of adding JavaScript code (tracking code even more so). Is this only required for Manage? In which case, this code shouldn’t be exposed to the candidate interface.

paulrobertlloyd:
> if they are not set, the tag manager JS is not included in the page.

Just read the above - phew! (Still not a fan…)

app/controllers/candidate_interface/candidate_interface_controller.rb
steventux: @cwrw I can't see any additional cookies being set on my development environment when I include the test GTM code so I'm not sure there's anything to add here. Will re-check in QA.

steventux:
@paulrobertlloyd Like the GA script, the GTM script is loaded asynchronously. This will have a small performance impact though it is non-blocking.
The various events pushed to the GTM data layer are sent via XHR POST request to collect the data, this is will be true of any pages including the GTM script (and _I think_) regardless of what we intend to track, so perhaps we need a smarter way of deciding which pages we include GTM on.

cwrw (approved):
LGTM!

PR #4489, opened 2021-04-12
Sync course subject names from ttapi [1/3]

despo:
## Context

Introduces subejcts and associates with courses
PR that utilises new association to follow.

## Guidance to review
- Run ` TeacherTrainingPublicAPI::SyncSubjects.perform_async`
- Check that Subjects table is populated
- Run `TeacherTrainingPublicAPI::SyncAllProvidersAndCoursesWorker.perform_async`
- Courses should now be associated with subjects 

## Link to Trello card

https://trello.com/c/aFNWvHnY/3581-sync-course-subject-names-from-ttapi-1

## Things to check


config/clock.rb
```diff
@@ -44,6 +44,10 @@ class Clock
     TeacherTrainingPublicAPI::SyncAllProvidersAndCoursesWorker.perform_async
   end
 
+  every(1.week, 'TeacherTrainingPublicAPI::SyncSubjects') do
```
cwrw, line 47: Should the subject sync be part of the wider `SyncAllFromTeacherTrainingPublicAPI`? I think it might be better to sync all TTAPI resources together per hour, than having this run once per week, what do you think?

db/schema.rb
```diff
@@ -396,6 +397,7 @@
     t.jsonb "qualifications"
     t.string "program_type"
     t.boolean "withdrawn"
+    t.jsonb "subjects"
```
cwrw, line 411: 🔥

app/models/subject.rb
```diff
@@ -1,3 +1,5 @@
 class Subject < ApplicationRecord
+  has_many :courses, through: :course_subjects
+
```
cwrw, line 3: do we also need the association: `has_many :course_subjects` as well?

db/migrate/20210409114058_create_course_subjects.rb
```diff
@@ -0,0 +1,11 @@
+class CreateCourseSubjects < ActiveRecord::Migration[6.0]
+  def change
+    create_table :course_subjects do |t|
+      t.references :course, null: false, foreign_key: true
+      t.references :subject, null: false, foreign_key: true
+
+      t.timestamps
+    end
+    add_index :course_subjects, %i[course_id subject_id], unique: true
```
cwrw, line 9: can we add a unique validation on the model as well?

cwrw (commented):
looking good, some minor comments

db/schema.rb
despo: 🤦🏼‍♀️

app/models/subject.rb
despo: We do. I introduced it in the next PR.

app/workers/teacher_training_public_api/sync_all_providers_and_courses_worker.rb
```diff
@@ -4,6 +4,7 @@ class SyncAllProvidersAndCoursesWorker
     sidekiq_options retry: 3, queue: :low_priority
 
     def perform
+      SyncSubjects.perform_async
```
cwrw, line 7: I think this should no longer be async and we should change it to a normal service as such:
```
SyncSubjects.call
```
to allow this to be run sequentially before syncing providers and courses

cwrw: or maybe just call `.perform` actually

steventux (approved):
:+1: Looks good, pending @cwrw 's comment here https://github.com/DFE-Digital/apply-for-teacher-training/pull/4489/files#r613053549

app/workers/teacher_training_public_api/sync_all_providers_and_courses_worker.rb
despo: Addressed

cwrw (approved):
👍

PR #4490, opened 2021-04-12
Sync course subject names from ttapi [2/3]

despo:
## Context

Use new subjects association to map course subject codes to exports

## Link to Trello card

https://trello.com/c/nkoWKwIn/3582-sync-course-subject-names-from-ttapi-2

## Things to check

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -137,4 +137,8 @@ class Course < ApplicationRecord
       .where(application_choices: { course_options: { course: self } })
       .distinct
   end
+
+  def subject_codes
+    @subject_codes ||= subjects.includes(:course_subjects).map(&:code)
+  end
 end
--- a/app/models/subject.rb
+++ b/app/models/subject.rb
@@ -1,4 +1,5 @@
 class Subject < ApplicationRecord
+  has_many :course_subjects
   has_many :courses, through: :course_subjects
 
   validates :code, uniqueness: true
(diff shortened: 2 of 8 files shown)


app/services/data_api/tad_application_export.rb
```diff
@@ -59,7 +59,7 @@ def as_json
         programme_outcome: application_choice.course.description,
         course_name: application_choice.course.name,
         course_code: application_choice.course.code,
-        nctl_subject: concatenate(application_choice.course.subject_codes),
+        nctl_subject: concatenate(application_choice.course.subjects.map(&:code)),
```
cwrw, line 62: if we add a method to course called `subject_codes`, that maps those codes from the new `subject` object, we can avoid having to call those changes in every instance, what do you think?

app/models/subject.rb
```diff
@@ -1,4 +1,5 @@
 class Subject < ApplicationRecord
+  has_many :course_subjects
```
cwrw, line 2: 😅

app/services/data_api/tad_application_export.rb
despo: Agreed

app/models/course.rb
```diff
@@ -137,4 +137,8 @@ def application_forms
       .where(application_choices: { course_options: { course: self } })
       .distinct
   end
+
+  def subject_codes
+    @subject_codes ||= subjects.includes(%w[course_subjects]).map(&:code)
```
cwrw, line 142:
```suggestion
    @subject_codes ||= subjects.includes(:course_subjects).map(&:code)
```

cwrw (approved):

PR #4492, opened 2021-04-12
Use GovukComponent::InsetText

paulrobertlloyd:
## Context

Use `GovukComponent::InsetText` to render inset text (via `govuk-components` `govuk_inset_text` helper method).

* Essentially replaces HTML markup with equivalent method from the gem and its params
* Also fixes apparent bug with when ‘No admin users’ warning is shown in support

## Guidance to review

* Is this bug fix correct (see inline comment)? Am using the same test as used in the view that lists all provider users

## Things to check

### DIFF
--- a/app/components/support_interface/ucas_match_action_component.rb
+++ b/app/components/support_interface/ucas_match_action_component.rb
@@ -68,7 +68,7 @@ module SupportInterface
     def required_action_details
       capture do
         concat tag.p(ACTIONS[@match.next_action][:instructions])
-        concat tag.p("Please refer to #{govuk_link_to('Dual-running user support manual', 'https://docs.google.com/document/d/1XvZiD8_ng_aG_7nvDGuJ9JIdPu6pFdCO2ujfKeFDOk4')} for more information about the current process.".html_safe)
+        concat tag.p("Please refer to the #{govuk_link_to('dual-running user support manual', 'https://docs.google.com/document/d/1XvZiD8_ng_aG_7nvDGuJ9JIdPu6pFdCO2ujfKeFDOk4')} for more information about the current process.".html_safe)
       end
     end
 
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -65,9 +65,10 @@ class Provider < ApplicationRecord
     accredited_providers.all?(&:onboarded?)
   end
 
-  def no_admin_users?
-    !(provider_permissions.exists?(manage_users: true) &&
-      provider_permissions.exists?(manage_organisations: true))
+  def lacks_admin_users?
+    sync_courses &&
+      !(provider_permissions.exists?(manage_users: true) &&
+        provider_permissions.exists?(manage_organisations: true))
   end
 
   def not_accepting_appplications_on_ucas?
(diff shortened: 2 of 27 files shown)


app/views/support_interface/providers/_no_admin_users_warning.html.erb
```diff
@@ -1,6 +1,6 @@
-<% if @provider.sync_courses? && @provider.no_admin_users? %>
-  <div class="govuk-inset-text govuk-!-margin-bottom-4">
+<% if @provider.no_admin_users? %>
```
paulrobertlloyd, line 1: Changed to match same test as used here: https://github.com/DFE-Digital/apply-for-teacher-training/blob/7858f4c5ad0cf4c1207eb66be7295e1c98c5c66f/app/views/support_interface/providers/index.html.erb#L26

app/components/support_interface/ucas_match_action_component.rb
```diff
@@ -68,7 +68,7 @@ def type_of_action
     def required_action_details
       capture do
         concat tag.p(ACTIONS[@match.next_action][:instructions])
-        concat tag.p("Please refer to #{govuk_link_to('Dual-running user support manual', 'https://docs.google.com/document/d/1XvZiD8_ng_aG_7nvDGuJ9JIdPu6pFdCO2ujfKeFDOk4')} for more information about the current process.".html_safe)
+        concat tag.p("Please refer to the #{govuk_link_to('dual-running user support manual', 'https://docs.google.com/document/d/1XvZiD8_ng_aG_7nvDGuJ9JIdPu6pFdCO2ujfKeFDOk4')} for more information about the current process.".html_safe)
```
duncanjbrown, line 71: 📐

duncanjbrown (commented):
Great! Want to fix `no_admin_users?` while you're here...?

paulrobertlloyd:
> Great! Want to fix `no_admin_users?` while you're here...?

Can do. Why rename the method though?

paulrobertlloyd:
@duncanjbrown Making the following change:

```rb
def lacks_admin_users?
  !(provider_permissions.exists?(sync_courses: true) &&
    provider_permissions.exists?(manage_users: true) &&
    provider_permissions.exists?(manage_organisations: true))
end
```

Results in this error when running the tests:

`PG::UndefinedColumn: ERROR:  column provider_users_providers.sync_courses does not exist`

Is this the correct change?

app/models/provider.rb
```diff
@@ -65,8 +65,9 @@ def all_associated_accredited_providers_onboarded?
     accredited_providers.all?(&:onboarded?)
   end
 
-  def no_admin_users?
-    !(provider_permissions.exists?(manage_users: true) &&
+  def lacks_admin_users?
+    !(sync_courses &&
```
duncanjbrown, line 69: I think this should be `sync_courses && !(...` — at the moment the `sync_courses` condition is backwards which means everything gets the "No admin user" label

Screenshot 2021-04-15 at 17 03 32

duncanjbrown (approved):
📐📐📐

PR #4495, opened 2021-04-13
Remove duplicate references from the ApplicationReferences export

davidgisbey:
## Context

At the moment, duplicate references are exposed in the ApplicationReference export. This causes issues with automation of the export analysis purposes as specific columns are used to calculate average response time etc. which is skewed when you have duplicate values.

- Reject duplicate references from the ApplicationReference export

## Guidance to review

Does this seem reasonable?

There's definitely a trade-off as this is now only showing *new* references which could be slightly misleading if you don't know that is the behaviour.

## Link to Trello card

https://trello.com/c/hXZDSy12/3264-remove-duplicate-references-from-the-applicationreferenceexport

## Things to check

### DIFF
--- a/app/models/data_export.rb
+++ b/app/models/data_export.rb
@@ -15,7 +15,7 @@ class DataExport < ApplicationRecord
     application_references: {
       name: 'Application references',
       export_type: 'application_references',
-      description: 'A list of all application references which have been selected by candidates to date.',
+      description: 'A list of all application references which have been selected by candidates to date. All duplicate references caused by duplication of a candidates application form caused by Applying Again or carrying over their application between cycles have been removed.',
       class: SupportInterface::ApplicationReferencesExport,
     },
     application_timings: {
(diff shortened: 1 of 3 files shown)


malcolmbaig:
I think on the whole this achieves what we need from the export. Regarding this:

> There's definitely a trade-off as this is now only showing new references which could be slightly misleading if you don't know that is the behaviour.

I think we should do either one of these:
- Include the duplicates, but flag them as such via a new column in the export so that they can be filtered out if needs be
- Document their exclusion in the description of this export

davidgisbey:
Yea we had the show they're dupes vs hide them chat and Mike/Milan wanted them removed. You're right it should definitely be misleading so I'll document it now.

stevehook (approved):
I think this makes sense so long as whoever consumes the report understands that duplicates are excluded. It could be stated in the `description` in `app/models/data_export.rb`. In any case it's now consistent with the way the reference feature metrics work.

davidgisbey:
Yea I must admit that I do have a soft preference for marking them as dupes and filtering. There's also the question of is a feedback provided reference that gets cancelled still a dupe?

Bit of a can of worms though.

PR #4496, opened 2021-04-13
Add ContentSecurityPolicy header

stevehook:
## Context

This PR configures our application to set the ContentSecurityPolicy response header. This limits the kind of content that browsers will run in our service. For example, we only allow JavaScript to run from our own site or a short whitelist of trusted providers. This helps to mitigate the risk of XSS.

- Uncomment the default CSP config in `config/initializers/content_security_policy.rb`.
- We pull in a handful of scripts from external sites (for ZenDesk and Google Analytics). We need to whitelist these.
- We have a few inline JavaScripts. We can use a nonce to validate these.
- We have a few inline CSS `style` elements. There doesn't seem to be a way to use a nonce to validate these in the same way as JavaScript so we permit unsafe_inline for styles. See https://github.com/omniauth/omniauth/blob/a62d36b3f847e0e55b077790112e96950c35085a/lib/omniauth/form.rb
- ZenDesk integration also uses Web sockets (for live chat I guess) and this too needs to be whitelisted.

## Guidance to review

- Is anything missing?
- I would have liked to have feature-flagged this change because it potentially breaks significant pieces of functionality. But given that the config is loaded in an initialiser I don't think it would be an effective way to switch it off (you need to restart the application servers). Is there another way that we can de-risk this change?
- Is it reasonable to allow `unsafe-inline` CSS?

## Link to Trello card

https://trello.com/c/ApkZRgMC/3259-pentest-416-were-missing-application-security-headers-low-🚧

## Things to check

### DIFF
--- a/config/initializers/content_security_policy.rb
+++ b/config/initializers/content_security_policy.rb
@@ -4,19 +4,19 @@
 # For further information see the following documentation
 # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy
 
-# Rails.application.config.content_security_policy do |policy|
-#   policy.default_src :self, :https
-#   policy.font_src    :self, :https, :data
-#   policy.img_src     :self, :https, :data
-#   policy.object_src  :none
-#   policy.script_src  :self, :https
-#   policy.style_src   :self, :https
-#   # If you are using webpack-dev-server then specify webpack-dev-server host
-#   policy.connect_src :self, :https, "http://localhost:3035", "ws://localhost:3035" if Rails.env.development?
+Rails.application.config.content_security_policy do |policy|
+  policy.default_src :self, :https
+  policy.font_src    :self, :https, :data
+  policy.img_src     :self, :https, :data
+  policy.object_src  :none
+  policy.script_src  :self, :https
+  policy.style_src   :self, :https
+  # If you are using webpack-dev-server then specify webpack-dev-server host
+  policy.connect_src :self, :https, "http://localhost:3035", "ws://localhost:3035" if Rails.env.development?
 
-#   # Specify URI for violation reports
-#   # policy.report_uri "/csp-violation-report-endpoint"
-# end
+  # Specify URI for violation reports
+  # policy.report_uri "/csp-violation-report-endpoint"
+end
 
 # If you are using UJS then enable automatic nonce generation
 # Rails.application.config.content_security_policy_nonce_generator = -> request { SecureRandom.base64(16) }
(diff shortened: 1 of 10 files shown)


config/initializers/content_security_policy.rb
```diff
@@ -4,19 +4,19 @@
...
-#   policy.connect_src :self, :https, "http://localhost:3035", "ws://localhost:3035" if Rails.env.development?
+Rails.application.config.content_security_policy do |policy|
+  policy.default_src :self, :https
+  policy.font_src    :self, :https, :data
+  policy.img_src     :self, :https, :data
+  policy.object_src  :none
+  policy.script_src  :self, :https
+  policy.style_src   :self, :https
+  # If you are using webpack-dev-server then specify webpack-dev-server host
+  policy.connect_src :self, :https, "http://localhost:3035", "ws://localhost:3035" if Rails.env.development?
```
duncanjbrown: I'm not sure if anyone's using it:

https://ukgovernmentdfe.slack.com/archives/CN1MCQCHZ/p1618416694494700

config/initializers/content_security_policy.rb
```diff
@@ -4,25 +4,28 @@
...
-#   policy.default_src :self, :https
-#   policy.font_src    :self, :https, :data
-#   policy.img_src     :self, :https, :data
-#   policy.object_src  :none
-#   policy.script_src  :self, :https
-#   policy.style_src   :self, :https
-#   # If you are using webpack-dev-server then specify webpack-dev-server host
-#   policy.connect_src :self, :https, "http://localhost:3035", "ws://localhost:3035" if Rails.env.development?
+Rails.application.config.content_security_policy do |policy|
+  connect_src = ['wss://widget-mediator.zopim.com']
```
duncanjbrown, line 8: Maybe good to label this with `ZenDesk chat`?

duncanjbrown (commented):
I saw this on `/provider` on the review app:

Screenshot 2021-04-14 at 17 29 59

If we want to feature flag, it should be possible to replace the `ContentSecurityPolicy` middleware with an inherited version? https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/http/content_security_policy.rb

I don't think `unsafe-inline` CSS is meaningfully risky for us. Also, if I understand it correctly, this PR doesn't make it any less secure than it is now?

duncanjbrown (approved):
⛓️ ✅

stevehook:
Thanks @duncanjbrown for review and testing. I had a go at your suggestion for subclassing the middleware (despite being a bit unsure about adding a DB call per request in a middleware) and it seems to work pretty well. Do you think it's worth including to be on the safe side? https://github.com/DFE-Digital/apply-for-teacher-training/compare/3259-application-security-headers...3259-content-security-policy-feature-flag?expand=1

duncanjbrown:
Oh, that middleware is really neat. I think it is worth including — I know TVS managed to break something with CSP https://ukgovernmentdfe.slack.com/archives/CQ3UX00P6/p1604510505010700

PR #4497, opened 2021-04-13
Add view to handle unsubmitted previous applications

richardpattinson:
## Context

Controller didnt differentiate between unsubmitted and submitted applications on the review page. This was causing the page to look for a submitted date on an unsubmitted application; throwing an error. A new review_previous_application.html.erb page was created to handle unsubmitted previous applications

## Link to Sentry Error

https://sentry.io/organizations/dfe-bat/issues/2335598643/?project=1765973&referrer=slack#
## Things to check

### DIFF
--- a/app/views/candidate_interface/application_form/review_previous_application.html.erb
+++ b/app/views/candidate_interface/application_form/review_previous_application.html.erb
@@ -1,0 +1,9 @@
+<% content_for :title, submitted_at_date.present? ? t('page_titles.submitted_application') : t('page_titles.unsubmitted_previous_application') %>
+
+<% content_for :before_content, govuk_back_link_to(candidate_interface_application_form_path) %>
+
+<h1 class="govuk-heading-xl">
+  <%= submitted_at_date.present? ? t('page_titles.submitted_application') : t('page_titles.unsubmitted_previous_application') %>
+</h1>
+
+<%= render 'candidate_interface/application_form/review', application_form: @application_form, editable: false %>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -48,6 +48,7 @@ en:
     apply_from_find_with_ucas: Apply for this course with UCAS
     apply_to_course_on_ucas: You need to apply to this course on UCAS
     apply_to_provider_on_ucas: You need to apply to this provider on UCAS
+    unsubmitted_previous_application: Your previously unsubmitted application
     submitted_application: Your submitted application
     create_account_or_sign_in: Create an account or sign in
     sign_up: Create an account
--- a/app/controllers/candidate_interface/application_form_controller.rb
+++ b/app/controllers/candidate_interface/application_form_controller.rb
@@ -2,9 +2,6 @@ module CandidateInterface
   class ApplicationFormController < CandidateInterfaceController
     def review_previous_application
       @application_form = current_candidate.application_forms.find(params[:id])
-      @review_previous_application = true
-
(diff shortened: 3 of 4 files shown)


app/views/candidate_interface/application_form/review_previous_application.html.erb
```diff
@@ -0,0 +1,9 @@
+<% content_for :title, t('page_titles.submitted_application') %>
```
benswannack, line 1: This should match the h1

config/locales/en.yml
```diff
@@ -50,6 +50,7 @@ en:
     apply_from_find_with_ucas: Apply for this course with UCAS
     apply_to_course_on_ucas: You need to apply to this course on UCAS
     apply_to_provider_on_ucas: You need to apply to this provider on UCAS
+    unsubmitted_previous_application: Your unsubmitted application
```
benswannack, line 53: Should this be something like `Your previous application`? Now we have a separate view for the review_previous_application, but it is not necessarily unsubmitted

paulrobertlloyd: Possibly one for @Tomcowan1. For a previously *submitted* application we say ‘Your submitted application’, correct? (Can offer context if needed Tom)

benswannack: We can potentially go back to sharing the view and render submitted vs unsubmitted application depending on what its state is. Although "Your unsubmitted application" feels a bit ambiguous as it could refer to the current one.

richardpattinson: So 'Your previous unsubmitted aplication' would work? Or is it too verbose

Tomcowan1: Hi @richardpattinson Richard - can I just check the scenario - is an unsubmitted application only for people who carry over from one cycle to another?

richardpattinson: If the candidate has made a previous application and didnt submit it then they should be able to view this page. In practice that should only happen when they carry over from a previous year.

Tomcowan1: Thanks @richardpattinson. I'm happy to go with "Your previously unsubmitted application." We need to differentiate it from the current one (that's also unsubmitted).

app/views/candidate_interface/application_form/review_previous_application.html.erb
```diff
@@ -0,0 +1,9 @@
...
+
+<% content_for :before_content, govuk_back_link_to(candidate_interface_application_form_path) %>
+
+<h1 class="govuk-heading-xl">
+  <%= t('page_titles.unsubmitted_previous_application') %>
```
benswannack, line 6: Can we make sure this reflects whether the application is submitted or not. Suggest wrapping this in an `if submitted_at_date.present?`, and choosing between the submitted and unsubmitted titles in there. Similarly for the `content_for :title` above

richardpattinson: Sure thing, ill put it in now

app/views/candidate_interface/application_form/review_previous_application.html.erb
benswannack:
```suggestion
<% content_for :title, submitted_at_date.present? ? t('page_titles.submitted_application') : t('page_titles.unsubmitted_previous_application') %>
```

cwrw (approved):

benswannack (approved):
Thanks!

PR #4498, opened 2021-04-13
Sync course subject names from ttapi [3/3]

despo:
## Context

Removes `subject_codes` field from `courses` table

## Link to Trello card

https://trello.com/c/81C0pJrz/3583-sync-course-subject-names-from-ttapi-3

## Things to check

### DIFF
--- a/spec/system/find_sync/syncing_providers_spec.rb
+++ b/spec/system/find_sync/syncing_providers_spec.rb
@@ -8,7 +8,7 @@ RSpec.describe 'Syncing providers', sidekiq: true do
     and_there_is_a_provider_with_a_course_in_find
 
     when_the_sync_runs
-    then_it_updates_the_course_subject_codes
+    then_it_updates_the_course_subjects
     and_it_sets_the_last_synced_timestamp
   end
 
@@ -42,10 +42,10 @@ RSpec.describe 'Syncing providers', sidekiq: true do
     TeacherTrainingPublicAPI::SyncAllProvidersAndCoursesWorker.perform_async
   end
 
-  def then_it_updates_the_course_subject_codes
+  def then_it_updates_the_course_subjects
     course = Course.find_by(code: 'ABC1')
 
-    expect(course.subject_codes).to eq(%w[08])
+    expect(course.subjects.map(&:code)).to eq(%w[08])
   end
 
   def and_it_sets_the_last_synced_timestamp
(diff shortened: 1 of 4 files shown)


mnacos (commented):
Looks very good. I would approve if it wasn't a draft! It even covers exports via the `.subject_codes` helper method on course. This could return an empty list, of course, if a sync hasn't happened, but maybe we could trigger one as part of the deploy?

despo:
@mnacos As syncs happen hourly I expect one to have been done prior to this being merged (given it has already worked in QA and Staging, and part 2 still needs to be merged), so I don't think we need any manual intervention! 

I'm waiting for part 2 to be approved, merged and deployed though before opening the PR for review just to avoid any confusion or accidental merge!

SamJamCul:
Should we do the migration in a separate PR?

despo:
@SamJamCul every other dependency has already been satisfied through PRs #4489 and #4490

SamJamCul:
Would somewhere like [this](https://github.com/DFE-Digital/apply-for-teacher-training/blob/7ab6bdff99c91efe7e6b179401b192fbcfd73a62/app/services/data_api/tad_application_export.rb#L62) also need to have the `subject_codes` removed?

despo:
@SamJamCul No as I have introduced a method as part of https://github.com/DFE-Digital/apply-for-teacher-training/pull/4490 that maps subject.code to `subject_codes` for simplicity.

spec/system/find_sync/syncing_providers_spec.rb
```diff
@@ -42,10 +42,10 @@ def when_the_sync_runs
     TeacherTrainingPublicAPI::SyncAllProvidersAndCoursesWorker.perform_async
   end
 
-  def then_it_updates_the_course_subject_codes
+  def then_it_updates_the_course_subjects
     course = Course.find_by(code: 'ABC1')
 
-    expect(course.subject_codes).to eq(%w[08])
```
cwrw, line 48: can we not keep this as `subject_codes` as we defined the helper method to map those?

cwrw (approved):
Looks good, just the one comment

spec/system/find_sync/syncing_providers_spec.rb
despo: We could, however this reflects what the sync actually does and it doesn't require digging into the course code in order to understand why subjects is set and subject_codes is returned.

PR #4499, opened 2021-04-13
Start reading from current_course_option instead of offered_course_option

mnacos:
## Context

Finding the current course option and its associated models (course, provider, site) is hard because an application choice may be pointing to a new course option via `offered_course_option_id`, if it is in the 'offer' state or beyond. Furthermore, application choices already in the 'offer' state or beyond from longer than 6 months ago may also have a nil `offered_course_option_id`, if no course details changed as part of the offer.

All this makes joins hard and leads to `COALESCE` statements in our queries. This PR addresses some of the pain by starting to maintain `current_course_option_id`, introduced already by https://github.com/DFE-Digital/apply-for-teacher-training/pull/4519 as part of this work. From this PR onwards, `current_course_option_id` will be set to the same course option as the application choice from the start and will be updated during offer events. This means `current_course_option` will always point to the latest course option for this ApplicationChoice and it is guaranteed to be populated. The `course_option_id` attribute will be retained, as a record of the original candidate choice, but `offered_course_option_id` will be removed in a future PR.

This work will also unblock fixing the current bugs around filtering on current (vs original) providers and sites.

Add `current_course_option`, `current_course`, `current_site` associations to ApplicationChoice for querying convenience e.g. `ApplicationChoice.where(current_site: site)`

Add data migration to backfill `current_course_option_id` based on `offered_course_option_id` and `course_option_id`. This is done in SQL to avoid triggering any timestamp changes, audits and also for performance.

Update all code and tests to use `current_course_option` instead of `offered_option` or any of its siblings (`offered_course`). Remove `offered_option` and `offered_course`.

## Guidance to review

Run the specs, interact with the app. Everything should behave exactly the same.

~~Should the database migration introducing `current_course_option_id` be deployed first/separately? I wouldn't expect any user-facing exceptions, since the older `offered_course_option_id` field is still present and correct. I've included the db and data migrations in the same PR as the code changes, so that we have a distinct point in time when we `current_course_option_id` starts being used. Would be interested to hear any other views on this.~~ We've already introduced the `current_course_option_id` column in a separate PR.

Should we also add a foreign key constraint for `current_course_option_id` and/or an index?

I haven't removed any `COALESCE` statements yet, as they will prevent any errors in Manage if the data migration takes a long time. I propose we add an index for `current_course_option_id` and remove the `COALESCE` statements in the next PR, which will remove `offered_course_option_id`.

## Link to Trello card

https://trello.com/c/oYjeu4Hh

## Things to check

- [x] This code does not rely on migrations in the same Pull Request -- ~~it does but no user-facing exceptions are expected~~


app/components/provider_interface/activity_log_event_component.rb
```diff
@@ -58,17 +58,20 @@ def event_description
...
+      option_at_time = CourseOption.find_by(id: changed_option_id)
+
+      @course_option ||= if changes['offer_changed_at'].present? && option_at_time.present?
+                           option_at_time
                          elsif current_status == 'offer'
-                           # older offers won't have a course option id as part of this
-                           from_event = CourseOption.find_by(id: changes['offered_course_option_id']&.second)
-                           from_event || application_choice.offered_option
+                           # older offers may not have option_at_time available
+                           # FIXME: should not be possible after data migration
```
duncanjbrown: Could you explain the logic here a bit more?

If the audit log contains changes to either the `current_option_id` or `offered_course_option_id` fields, then `option_at_time` contains whichever course we changed to. If neither of those are present (ie because the choice is from before we started always setting `offered_course_option_id`) then the offer hasn't changed, and we simply show whatever option the `application_choice` has now along with the `offer` event? I'm missing something here — what's not possible after the data migration?

app/controllers/provider_interface/application_choices_controller.rb
```diff
@@ -21,7 +21,7 @@ def index
 
       # Eager load / prevent Bullet::Notification::UnoptimizedQueryError
       with_includes = ApplicationChoice.includes(
-        %i[application_form provider offered_course_option site accredited_provider],
+        %i[application_form provider current_option current_course site accredited_provider],
```
duncanjbrown, line 24: I wonder why this changed? Because of the move to `has_one` maybe?

app/models/application_choice.rb
```diff
@@ -4,13 +4,19 @@ class ApplicationChoice < ApplicationRecord
...
+  has_one :course, through: :course_option
   has_one :provider, through: :course
   has_one :accredited_provider, through: :course, class_name: 'Provider'
-  has_one :candidate, through: :application_form
+
+  belongs_to :current_option, class_name: 'CourseOption', optional: true
+  has_one :current_site, through: :current_option, source: :site
+  has_one :current_course, through: :current_option, source: :course
+  has_one :current_provider, through: :current_course, source: :provider
+  has_one :current_accredited_provider, through: :current_course, source: :accredited_provider
```
duncanjbrown, line 19: Nice to have this

app/services/provider_interface/application_data_export.rb
```diff
@@ -34,13 +34,13 @@ def self.call(application_choices:)
           'postcode' => application.application_form.postcode,
           'country' => application.application_form.country,
           'recruitment_cycle_year' => application.application_form.recruitment_cycle_year,
-          'provider_code' => application.offered_provider.code,
+          'provider_code' => application.current_course.provider.code,
```
duncanjbrown, line 37: `current_provider`?

spec/factories/application_choice.rb
```diff
@@ -3,6 +3,10 @@
     course_option
     application_form
 
+    after(:build) do |application_choice, _evaluator|
+      application_choice.current_option = application_choice.course_option
```
duncanjbrown: This is lying to us until we've done the migration, right?

spec/services/provider_interface/application_data_export_spec.rb
```diff
@@ -72,13 +72,13 @@ def expect_row_to_match_application_choice(row, application_choice)
...
         'country' => application_choice.application_form.country,
         'recruitment_cycle_year' => application_choice.application_form.recruitment_cycle_year&.to_s,
-        'provider_code' => application_choice.offered_provider.code,
-        'accredited_provider_name' => application_choice.accredited_provider&.name,
-        'accredited_provider_code' => application_choice.accredited_provider&.code,
-        'course_code' => application_choice.offered_course.code,
-        'site_code' => application_choice.offered_site.code,
-        'study_mode' => application_choice.offered_course.study_mode,
-        'start_date' => application_choice.offered_course.start_date&.to_s,
+        'provider_code' => application_choice.current_course.provider.code,
```
duncanjbrown, line 75: `current_provider`?

spec/services/provider_interface/application_data_export_spec.rb
```diff
@@ -72,13 +72,13 @@ def expect_row_to_match_application_choice(row, application_choice)
...
+        'accredited_provider_name' => application_choice.current_course.accredited_provider&.name,
```
duncanjbrown, line 76: `current_accredited_provider`?

spec/services/support_interface/offer_conditions_export_spec.rb
```diff
@@ -115,12 +115,12 @@
       choice = create(:application_choice, :with_modified_offer)
 
       offers = described_class.new.offers
-      expect(offers.first[:offered_provider_code]).to eq(choice.offered_course.provider.code)
-      expect(offers.first[:offered_provider_name]).to eq(choice.offered_course.provider.name)
-      expect(offers.first[:offered_course_name]).to eq(choice.offered_course.name)
-      expect(offers.first[:offered_course_code]).to eq(choice.offered_course.code)
-      expect(offers.first[:offered_course_location]).to eq(choice.offered_site.name)
-      expect(offers.first[:offered_course_study_mode]).to eq(choice.offered_option.study_mode)
+      expect(offers.first[:offered_provider_code]).to eq(choice.current_course.provider.code)
```
duncanjbrown, line 118: As above

app/forms/candidate_interface/pick_site_form.rb
```diff
@@ -20,13 +20,17 @@ def save
...
       )
     end
 
     def update(application_choice)
       return unless valid?
 
-      application_choice.update!(course_option: course_option)
+      application_choice.update!(
+        course_option: course_option,
+        current_option: course_option,
```
duncanjbrown, line 32: I could see a future developer wondering about redundancy here. Maybe give this procedure a name, like `ApplicationChoice#set_initial_course_choice!(course_option)`?

app/views/provider_mailer/unconditional_offer_accepted.text.erb
```diff
@@ -3,7 +3,7 @@ Dear <%= @provider_user.full_name || 'colleague' %>
 
 # Offer accepted
 
-<%= @application_choice.application_form.full_name %> has accepted your offer to study <%= @application_choice.current_course.name_and_code %>.
+<%= @application_choice.application_form.full_name %> has accepted your offer to study <%= @application_choice.current_option.course.name_and_code %>.
```
duncanjbrown, line 6: Why can't we have `current_course` here and above?

spec/mailers/candidate_mailer_offers_and_rejections_spec.rb
```diff
@@ -232,7 +232,7 @@
   describe '.reinstated_offer' do
     let(:email) { mailer.reinstated_offer(application_choices.first) }
     let(:application_choices) { [application_choice] }
-    let(:application_choice) { build_stubbed(:application_choice, :with_deferred_offer, course_option: other_option, offer_deferred_at: Time.zone.local(2019, 10, 3)) }
+    let(:application_choice) { build_stubbed(:application_choice, :with_deferred_offer, course_option: other_option, current_option: other_option, offer_deferred_at: Time.zone.local(2019, 10, 3)) }
```
duncanjbrown, line 235: Why doesn't the `after(:build)` hook catch this and assign `course_option` to `current_option`?

app/components/provider_interface/activity_log_event_component.rb
```diff
@@ -64,10 +64,8 @@ def course_option
 
       @course_option ||= if changes['offer_changed_at'].present? && option_at_time.present?
                            option_at_time
-                         elsif current_status == 'offer'
-                           # older offers may not have option_at_time available
-                           # FIXME: should not be possible after data migration
-                           application_choice.current_option
+                         elsif option_at_time.present?
+                           option_at_time
```
duncanjbrown, line 68: OK this makes more sense!

spec/services/data_migrations/backfill_current_option_id_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    expect(application_choice.reload.current_option_id).to eq(course_option.id)
+  end
+
+  it 'falls back to course_option_id, if offered_course_option_id is blank' do
+    application_choice = create(:application_choice)
+
+    described_class.new.change
+
+    expect(application_choice.reload.current_option_id).to eq(application_choice.course_option.id)
```
duncanjbrown, line 18: Would the factory's hooks have set this anyway?

duncanjbrown (commented):
🎉  Massive improvement in readability imo. Few questions inline!

db/migrate/20210406141658_add_current_option_to_application_choice.rb
```diff
@@ -0,0 +1,5 @@
+class AddCurrentOptionToApplicationChoice < ActiveRecord::Migration[6.0]
+  def change
+    add_column :application_choices, :current_option_id, :bigint
```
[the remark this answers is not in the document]
despo: Can the migration be moved to a separate PR so the column can be added before it's used, so we can avoid any potential error?

 We had some issues with doing something similar on course yesteday and ended up having to remove any calls to the field, deploy to production an then reintroduce the calls.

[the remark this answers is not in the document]
mnacos: Both are good suggestions, please see https://github.com/DFE-Digital/apply-for-teacher-training/pull/4519

app/views/provider_mailer/unconditional_offer_accepted.text.erb
mnacos: The main issue I've had with the mailers is the previews, which use `build_stubbed` and also reference `current_option` for other things. I could have also specified `current_course` for the stubbed models but it felt like setting current_option is safer, as it can also be used for other things. Specifying `current_course` and `current_option` should work, but my main priority was to get all tests to pass without spending too much time on single cases.

spec/services/data_migrations/backfill_current_option_id_spec.rb
mnacos: Good catch, adding an explicit `application_choice.update_columns(current_course_option_id: nil)`

app/components/provider_interface/location_presentation_component.rb
```diff
@@ -7,7 +7,7 @@ def initialize(application_choice:)
     end
 
     def course_option
-      application_choice.offered_course_option || application_choice.course_option
+      application_choice.current_option
```
cwrw, line 10: 👍

cwrw:
Thanks for this, as it will really help in moving things towards better modelling and more explicit querying!

I have a few questions/suggestions:
I just want to confirm that the only difference between the current `offered_course` and the new `current_course` is that we are writing to `current_course` when creating the application choice? This is aside from the `nil` historic values and adding the new associations to help queries.

In terms of rolling this out my preferred strategy would be as follows:
- Migrations (already done and merged)
- Model associations and starting to write to the new field
- Backfill the old values, this is preferably in a separate PR. Although the possibility is low, I think its safer to make sure backfill runs after we start writing the field to avoid anything happening between the migration running and the code deployed for writing becoming live
- Reading from the new column. We could wrap the current field name `current_course` to read from the `offered_course` to avoid renaming everything initially. But I'm in two minds on this as we will have to do it eventually
- Removing the old field and cleanup

Let me know your thoughts.

mnacos:
Yes, `current_course_option` is similar to `offered_course_option` but:
- it's guaranteed to be set (cannot be `nil` - we could even enforce this at the database level during cleanup)
- it's guaranteed to be current, so safe to use at any stage (before, during and after the 'offer' stage)
 
> * Migrations (already done and merged)
> * Model associations and starting to write to the new field
> * Backfill the old values, this is preferably in a separate PR. Although the possibility is low, I think its safer to make sure backfill runs after we start writing the field to avoid anything happening between the migration running and the code deployed for writing becoming live
> * Reading from the new column. We could wrap the current field name `current_course` to read from the `offered_course` to avoid renaming everything initially. But I'm in two minds on this as we will have to do it eventually
> * Removing the old field and cleanup
> 
> Let me know your thoughts.

How about a variation of this:

- Database migration for adding `current_course_option_id` (already done)
- Another PR with just a data migration with MANUAL=true setting values for the new field. This will populate `current_course_option` for all application choices until now.
- Third PR is this one, minus the two migrations in the previous steps, plus a new MANUAL=false data migration identical to the previous one. This PR reads from/writes to the new field, so `current_course_option` and friends will be automatically populated from there on. The new, identical but automatic data migration will populate any nil `current_course_option_id` values which arose between PRs and should take very little time.
- A fourth PR will remove the old field and cleanup.

cwrw:
My worry with in the suggestion is our attempt to read/write in the same PR, even though small there is still a window where we have a `nil` value in the new field, meaning our manual rerun of a backfill is time dependent. This also means we have to manually run the backfill on every environment as well. I think running the backfill once as a separate step will help mitigate that overhead, as well as avoid having a certain time where we have `nil` values in the new field. Happy to chat over a call as well

mnacos:
> My worry in the above suggestion is our attempt to read/write in the same PR, even though small there is still a window where we have a `nil` value in the new field, meaning our manual rerun of a backfill is time dependent. This also means we have to manually run the backfill on every environment as well. I think running the backfill once as a separate step will help mitigate that overhead, as well as avoid having a certain time where we have `nil` values in the new field. Happy to chat over a call as well

Based on our slack chat, I'll open another PR to be deployed before this one, which will make the app start writing to `current_course_option_id`, so that any new application choices are guaranteed to have it set. I'll also prepare a separate PR for the data migration with `MANUAL=true`, also to be deployed before this one. We'll then rebase this PR against master and deploy it, at which point there is zero chance of a nil `current_course_option_id`.

app/controllers/provider_interface/application_choices_controller.rb
mnacos: Probably, the new association declarations make Rails more aware of the relationships.

spec/factories/application_choice.rb
mnacos: The factory will set `current_course_option_id` only if it isn't set, in which case it'll use the original `course_option`.

spec/mailers/candidate_mailer_offers_and_rejections_spec.rb
mnacos: I think `after(:build)` is used if data is set up via `.build` or `.create`, but `.build_stubbed` triggers `after(:stub)` without triggering `after(:build)`.

spec/factories/application_choice.rb
```diff
@@ -40,7 +46,7 @@
 
       after(:build) do |application_choice, _evaluator|
         application_choice.status = :awaiting_provider_decision
-        application_choice.interviews << build(:interview, :cancelled, provider: application_choice.provider)
+        application_choice.interviews << build(:interview, :cancelled, provider: application_choice.current_course_option.course.provider)
```
cwrw, line 49: Is there we can't use `current_provider` and similar in the factory file?

app/services/support_interface/notes_export.rb
```diff
@@ -9,7 +9,7 @@ def data_for_export
 
       notes.map do |note|
         providers = note.provider_user.providers
-        course = note.application_choice.offered_course_option&.course || note.application_choice.course_option.course
+        course = note.application_choice.current_course_option.course
```
cwrw, line 12: `current_course`?

app/services/provider_interface/application_data_export.rb
```diff
@@ -34,13 +34,13 @@ def self.call(application_choices:)
...
-          'course_code' => application.offered_course.code,
-          'site_code' => application.offered_site.code,
-          'study_mode' => application.offered_course.study_mode,
-          'start_date' => application.offered_course.start_date,
+          'provider_code' => application.current_provider.code,
+          'accredited_provider_name' => application.current_accredited_provider&.name,
+          'accredited_provider_code' => application.current_accredited_provider&.code,
+          'course_code' => application.current_course.code,
+          'site_code' => application.current_site.code,
+          'study_mode' => application.current_course_option.study_mode,
```
cwrw, line 42: The old code is looking at `current_course`, is there a difference between study mode on a course vs an option?

app/services/provider_authorisation_analysis.rb
```diff
@@ -8,10 +8,10 @@ def initialize(permission:, auth:, application_choice:, course_option_id: nil)
     @auth = auth
     @application_choice = application_choice
     @course_option_id = course_option_id
-    @course_option_id ||= @application_choice.offered_option.id
+    @course_option_id ||= @application_choice.current_course_option.id
 
-    @training_provider = application_choice.offered_course.provider
-    @ratifying_provider = application_choice.offered_course.accredited_provider
+    @training_provider = application_choice.current_course_option.course.provider
```
cwrw, line 13: `current_provider`

app/queries/get_all_change_options_from_offered_option.rb
```diff
@@ -5,22 +5,22 @@ class GetAllChangeOptionsFromOfferedOption
   def initialize(application_choice:, available_providers: nil)
     @application_choice = application_choice
 
-    @available_providers = available_providers || [application_choice.offered_course.provider]
+    @available_providers = available_providers || [application_choice.current_course.provider]
```
cwrw, line 8: scope for using `current_provider` and other attributes in this file

db/schema.rb
```diff
@@ -11,7 +11,6 @@
 # It's strongly recommended that you check this file into your version control system.
 
 ActiveRecord::Schema.define(version: 2021_04_15_141658) do
-
```
cwrw, line 14: 🔥

spec/services/data_migrations/backfill_current_course_option_id_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'rails_helper'
```
cwrw, line 1: I guess now this has been merged we can remove from this commit with the data migration file?

cwrw (approved):
Looks great @mnacos! left some minor comments/questions. I think we can also remove the generated model binary change from this PR as well. I'll give this a quick whirl tomorrow, to make sure all the critical paths still work as expected before merging, but otherwise 👍

spec/factories/application_choice.rb
mnacos: Should be fine

app/services/provider_interface/application_data_export.rb
mnacos: Yes, `study_mode` on course can take the following values: `full_time`, `part_time`, `full_time_or_part_time` but `study_mode` on a course option can only be `full_time` or `part_time`. Considering the export deals with application choices, we should be using the course option one, as specific choices can only be full-time or part-time.

PR #4500, opened 2021-04-13
Fix incorrect error message in the add course Apply Again journey

davidgisbey:
## Context

At the moment, if a candidate has an apply_2 application form with an application choice, when they arrive from Find with course params then they are told that they cannot have more than 3 courses associated with their application form.

This commit checks to see if the application form is apply_1 or apply_2 before rendering the correct error message.

- Ensure that the `apply_again_course_already_chosen` error is rendered when a candidate already has a course choice on their application form and arrives from Find with course params in the query string.

|Before|After|
|------------|------------|
|image|image|

## Link to Trello card

https://trello.com/c/Z1ySPc8g/3292-bug-fix-incorrect-add-course-choice-error-message-for-apply-again-applications

## Things to check

### DIFF
--- a/app/controllers/candidate_interface/after_sign_in_controller.rb
+++ b/app/controllers/candidate_interface/after_sign_in_controller.rb
@@ -14,7 +14,9 @@ module CandidateInterface
         flash[:warning] = "You have already selected #{course.name_and_code}."
         redirect_to candidate_interface_course_choices_review_path
       elsif current_application.has_the_maximum_number_of_course_choices?
-        flash[:warning] = I18n.t('errors.messages.too_many_course_choices', course_name_and_code: course.name_and_code)
+        error_message_key = current_application.apply_1? ? 'errors.messages.too_many_course_choices' : 'errors.messages.apply_again_course_already_chosen'
+        flash[:warning] = I18n.t(error_message_key, course_name_and_code: course.name_and_code)
+
         redirect_to candidate_interface_course_choices_review_path
       else
         redirect_to candidate_interface_course_confirm_selection_path(course.id)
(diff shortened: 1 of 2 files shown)


app/controllers/candidate_interface/after_sign_in_controller.rb
```diff
@@ -14,7 +14,11 @@ def interstitial
         flash[:warning] = "You have already selected #{course.name_and_code}."
         redirect_to candidate_interface_course_choices_review_path
       elsif current_application.has_the_maximum_number_of_course_choices?
-        flash[:warning] = I18n.t('errors.messages.too_many_course_choices', course_name_and_code: course.name_and_code)
+        if current_application.apply_1?
+          flash[:warning] = I18n.t('errors.messages.too_many_course_choices', course_name_and_code: course.name_and_code)
+        else
+          flash[:warning] = I18n.t('errors.messages.apply_again_course_already_chosen', course_name_and_code: course.name_and_code)
+        end
```
benswannack, line 21: Could we do something like:
```suggestion
        error_message_key = current_application.apply_1? ? 'errors.messages.too_many_course_choices' : 'errors.messages.apply_again_course_already_chosen'
        flash[:warning] = I18n.t(error_message_key, course_name_and_code: course.name_and_code)
```

benswannack (approved):
Nice find - thanks :) Minor tweak suggested inline

app/controllers/candidate_interface/after_sign_in_controller.rb
davidgisbey: have updated 👍 Thanks for the review

PR #4503, opened 2021-04-14
Make unique courses per application invariant check more lenient

benswannack:
## Context
There are some applications which do not pass the check, even though they should really be valid. We want to allow multiple applications to the same course if the application choices have reached a non-success state.

Ignore application choices that have been withdrawn or rejected when checking if a candidate has applied to the same course multiple times

## Guidance to review
Withdrawn is actually an end state, it's not possible to go back without support help.
Rejected is not actually an end state, since it's possible to make an offer on a rejected application. I don't think we need to worry about this until it happens, especially since we'll be notified by the checks if they break.

## Slack thread
https://ukgovernmentdfe.slack.com/archives/CN1MCQCHZ/p1618063976394000

## Things to check

### DIFF
--- a/app/workers/detect_invariants.rb
+++ b/app/workers/detect_invariants.rb
@@ -122,6 +122,7 @@ class DetectInvariants
     applications_with_the_same_choice = ApplicationForm
       .joins(application_choices: [:course_option])
       .where.not(submitted_at: nil)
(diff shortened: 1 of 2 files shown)


davidgisbey (approved):
Looks good to me 👍

PR #4505, opened 2021-04-14
Add Feature-Policy response header

stevehook:
## Context

This PR configures the `Feature-Policy` HTTP response header to declare the set of browser features that the application is permitted (or not permitted in our case) to use. For example, Apply does not need to use the users computers camera.

This is a follow up to https://github.com/DFE-Digital/apply-for-teacher-training/pull/4496 (though it's independent).

- [x] Add the `Feature-Policy` header and disable browser features that we don't use

## Guidance to review

- Is there some feature on this list (that I've missed) that we do actually use?

There are a couple of other headers that came up in pen-testing that we were encouraged to consider:
- I've skipped `Referer-Policy` because we use redirect_to :back in our application that uses the `Referer` header.
- Also skipping `Expect-CT` as it's likely to be obsolete in a couple of months (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expect-CT)

## Link to Trello card

https://trello.com/c/ApkZRgMC/3259-pentest-416-were-missing-application-security-headers-low-🚧

## Things to check

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -69,5 +69,9 @@ module ApplyForPostgraduateTeacherTraining
     config.after_initialize do |app|
       app.routes.append { get '*path', to: 'errors#not_found' }
     end
+
+    config.action_dispatch.default_headers = {
+      'Feature-Policy' => "accelerometer 'none'; camera 'none'; geolocation 'none'; gyroscope 'none'; magnetometer 'none'; microphone 'none'; payment 'none'; usb 'none'",
+    }
   end
 end
--- a/config/application.rb
+++ b/config/application.rb
@@ -71,7 +71,7 @@ module ApplyForPostgraduateTeacherTraining
     end
 
     config.action_dispatch.default_headers = {
-      'Feature-Policy' => "accelerometer 'none'; camera 'none'; geolocation 'none'; gyroscope 'none'; magnetometer 'none'; microphone 'none'; payment 'none'; usb 'none'",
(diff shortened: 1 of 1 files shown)


config/application.rb
```diff
@@ -69,5 +69,9 @@ class Application < Rails::Application
     config.after_initialize do |app|
       app.routes.append { get '*path', to: 'errors#not_found' }
     end
+
+    config.action_dispatch.default_headers = {
+      'Feature-Policy' => "accelerometer 'none'; camera 'none'; geolocation 'none'; gyroscope 'none'; magnetometer 'none'; microphone 'none'; payment 'none'; usb 'none'",
```
duncanjbrown, line 74: Do we want to change to `Permissions-Policy`?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Feature_Policy

duncanjbrown: It might be worth giving the list of possible permissions a once-over in case they've added anything since the pen test: https://developer.mozilla.org/en-US/docs/Web/HTTP/Feature_Policy#granular_control_over_certain_features.

stevehook: Good questions. It looks like `Feature-Policy` is better supported than `Permissions-Policy`

https://caniuse.com/?search=Feature-policy - All the modern browsers support it
https://caniuse.com/?search=Permissions-policy - Just recent Chrome and Edge

Looks like there are a lot of _features_ listed in the docs (though support is pretty sketchy for a lot of them). I've added any that look like they could be relevant from a security point of view.

duncanjbrown (approved):
Looks good to me!

PR #4506, opened 2021-04-14
Tag GA sessions with metadata to indicate whether a user is an hei scitt or sd user

steventux:
## Context

We'd like to be able to slice GA data by provider type. 

- Adds a method to determine the primary provider type, this will be either `university` or `scitt` or `school_direct` based on the types of the providers the user belongs to.
- Sets the primary provider type on the custom GA dimension (index 2 because that's the next one available).

## Guidance to review

We have this dimension set on our Manage Test GA account, verify it's working as expected before further deployment.

## Link to Trello card

https://trello.com/c/QzpVg0M5/3511-tag-ga-sessions-with-metadata-to-indicate-whether-a-user-is-an-hei-scitt-or-sd-user

## Things to check


app/controllers/provider_interface/provider_interface_controller.rb
```diff
@@ -34,6 +34,7 @@ def current_provider_user
 
     def check_cookie_preferences
       @google_analytics_id = ENV.fetch('GOOGLE_ANALYTICS_MANAGE', '') if cookies['consented-to-manage-cookies'].eql?('yes')
+      @provider_type = current_provider_user && current_provider_user.primary_provider_type
```
cwrw, line 37: I think you might need to use `current_provider_user&.primary_provider_type` with the new rubocop rules (went in recently)

cwrw (approved):
👍  just the one comment to avoid failures after rebasing

steventux:
@cwrw I had to make a slight tweak here https://github.com/DFE-Digital/apply-for-teacher-training/pull/4506/commits/3ca96d7a29ae93119f329ec51626d5cf03264441#diff-5183cb85f5b7e77048cc9ae789796cfa9ae0378ac7743b6f7a8294a4bf6fa35fR37 as `current_provider_user` will return `false` if not present and this will raise `undefined method 'primary_provider_type' for false:FalseClass` if we use the `.&` safe navigation operator.

cwrw (approved):

PR #4507, opened 2021-04-14
Temporary removal of mapping subjects to a course

despo:
to address sentry error https://sentry.io/organizations/dfe-bat/issues/2338215141/?referrer=slack


cwrw (commented):
Can we also remove the subject sync code just to be on the safe side?

cwrw (approved):

steventux (approved):
:+1:

PR #4508, opened 2021-04-14
Fix DBD bug with interviewing state

benswannack:
## Context
If an application has an offer and the rest of the choices are interviewing, then DBD is mistakenly set.

Check for application choices in the decisions pending statuses, which include interviewing


mnacos (approved):
Good catch

PR #4509, opened 2021-04-14
Rework BackfillExportType data migration

malcolmbaig:
## Context
Seeing some issues with this in production. It appears to be related to
either memory usage or a SQL timeout.

Change as follows:

- Use a simple hash lookup for the export_type values
- Use a small batch size of 10
- Set statement_timeout to 0 before execution

## Guidance to review
- Review of the diff should be sufficient, including a onceover of the lookup hash.

## Things to check

### DIFF
--- a/app/services/data_migrations/backfill_export_type.rb
+++ b/app/services/data_migrations/backfill_export_type.rb
@@ -4,28 +4,56 @@ module DataMigrations
     MANUAL_RUN = false
 
     def change
-      data_exports = DataExport.all.where(export_type: nil)
+      data_exports = DataExport.select(:id, :name, :export_type).where(export_type: nil)
 
-      data_exports.each { |export| export.update!(export_type: export_type(export)) }
+      data_exports.find_in_batches(batch_size: 10).with_index do |batch, batch_number|
+        Rails.logger.info("Updating DataExport export_types, batch no. #{batch_number}...")
+        batch.each { |export| export.update!(export_type: export_type_lookup.fetch(export.name)) }
+      end
     end
 
   private
 
-    def export_type(export)
-      case export.name
-      when 'Unexplained breaks in work history'
-        'work_history_break'
-      when 'Locations', 'Locations export'
-        'persona_export'
-      when 'Applications for TAD'
-        'tad_applications'
-      when 'Provider performance for TAD'
-        'tad_provider_performance'
-      when 'Candidate survey', 'Daily export of applications for TAD', 'Daily export of notifications breakdown', 'RejectedCandidatesExport'
-        nil
-      else
-        DataExport::EXPORT_TYPES.select { |_key, value| value[:name] == export.name }.values[0][:export_type]
-      end
+    def export_type_lookup
+      @export_type_lookup ||= {
+        'Unexplained breaks in work history' => 'work_history_break',
+        'Locations' => 'persona_export',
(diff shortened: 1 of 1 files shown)


app/services/data_migrations/backfill_export_type.rb
```diff
@@ -4,28 +4,58 @@ class BackfillExportType
     MANUAL_RUN = false
 
     def change
-      data_exports = DataExport.all.where(export_type: nil)
+      ActiveRecord::Base.connection.execute('SET statement_timeout = 0')
+      data_exports = DataExport.where(export_type: nil)
```
vigneshmsft, line 8: is it possible to not pull in the `data` column?
would this work?
```suggestion
      data_exports = DataExport.select(:id, :name, :export_type).where(export_type: nil)
```

davidgisbey (approved):

app/services/data_migrations/backfill_export_type.rb
davidgisbey: Yea this works

vigneshmsft (approved):

app/services/data_migrations/backfill_export_type.rb
```diff
@@ -4,28 +4,57 @@ class BackfillExportType
...
 
     def change
-      data_exports = DataExport.all.where(export_type: nil)
+      ActiveRecord::Base.connection.execute('SET statement_timeout = 0')
+      data_exports = DataExport.select(:id, :name, :export_type).where(export_type: nil)
 
-      data_exports.each { |export| export.update!(export_type: export_type(export)) }
+      data_exports.find_in_batches(batch_size: 10).with_index do |batch, batch_number|
+        Rails.logger.info("Updating DataExport export_types, batch no. #{batch_number}...")
+        batch.each { |export| export.update!(export_type: export_type_lookup[export.name]) }
```
duncanjbrown, line 12: Worth an `export_type_lookup.fetch` here.... just in case?

app/services/data_migrations/backfill_export_type.rb
```diff
@@ -4,28 +4,57 @@ class BackfillExportType
     MANUAL_RUN = false
 
     def change
-      data_exports = DataExport.all.where(export_type: nil)
+      ActiveRecord::Base.connection.execute('SET statement_timeout = 0')
```
duncanjbrown, line 7: Why do we need this? The query doesn't look too heavy

app/services/data_migrations/backfill_export_type.rb
```diff
@@ -4,28 +4,57 @@ class BackfillExportType
...
+      data_exports = DataExport.select(:id, :name, :export_type).where(export_type: nil)
 
-      data_exports.each { |export| export.update!(export_type: export_type(export)) }
+      data_exports.find_in_batches(batch_size: 10).with_index do |batch, batch_number|
+        Rails.logger.info("Updating DataExport export_types, batch no. #{batch_number}...")
```
duncanjbrown, line 11: 💙

app/services/data_migrations/backfill_export_type.rb
malcolmbaig: Prompted by the root cause, but with the other changes we've made here (not retrieving the data column in particular) I don't think is needed. Will remove.

PR #4510, opened 2021-04-14
Fix bug around withdrawing from interviewing choices when accepting an offer

benswannack:
## Context
If a candidate accepts an offer, we want to withdraw from any in progress applications and decline any other offers. Since adding the interviewing state, we do not withdraw interviewing applications when an offer is accepted

Look for and withdraw interviewing application choices when offers are accepted


mnacos (approved):
Loving this, variations on a theme 🚀

PR #4515, opened 2021-04-14
Reverse study mode with locations on offer summary cards

despo:
## Context
Update components displaying requested of offered course details in the order displayed in the make offer flow.

Update offer summary, course summary and old make offer components to first display the study mode and then the location of a course.

## Guidance to review

Changes should be reflected on
1. Make offer decision page
2. Offer summary page (at the end of make or change offer flow)
3. Offer tab
4. Defer & withdraw offer (can be accessed via offer tab when application choice is in correct state)
4. Reconfirm offer (end of defer offer flow)

## Link to Trello card

https://trello.com/c/F9HCtsg2/3618-reverse-study-mode-with-locations-on-offer-summary-cards

## Things to check

### DIFF
--- a/app/components/provider_interface/course_summary_component.rb
+++ b/app/components/provider_interface/course_summary_component.rb
@@ -22,12 +22,12 @@ module ProviderInterface
           value: course_name_and_code,
         },
         {
-          key: 'Location',
-          value: location_name_and_address,
-        },
-        {
           key: 'Full time or part time',
           value: study_mode,
+        },
+        {
+          key: 'Location',
+          value: location_name_and_address,
         },
       ]
       return rows if course_option.course.accredited_provider.blank?
--- a/app/components/provider_interface/deferred_offer_details_component.rb
+++ b/app/components/provider_interface/deferred_offer_details_component.rb
@@ -21,12 +21,12 @@ module ProviderInterface
           value: course_option.course.name_and_code,
         },
         {
-          key: 'Location',
-          value: course_option.site.name_and_address,
-        },
-        {
           key: 'Full time or part time',
           value: course_option.study_mode.humanize,
+        },
+        {
+          key: 'Location',
+          value: course_option.site.name_and_address,
         },
       ]
     end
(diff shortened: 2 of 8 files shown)


app/components/provider_interface/completed_offer_summary_component.rb
```diff
@@ -12,14 +12,14 @@ def rows
...
           change_path: change_course_path },
-        { key: 'Full time or part time',
-          value: course_option.study_mode.humanize,
-          action: 'Change',
-          change_path: change_study_mode_path },
         { key: 'Location',
           value: course_option.site.name_and_address,
           action: 'Change',
           change_path: change_location_path },
+        { key: 'Full time or part time',
```
mnacos, line 19: I'm a little confused, the card says 'Full time/part time should come before locations in all summary displays'. Is this change required?

app/components/provider_interface/read_only_completed_offer_summary_component.rb
```diff
@@ -8,10 +8,10 @@ def rows
           value: course_option.provider.name_and_code },
         { key: 'Course',
           value: course_option.course.name_and_code },
-        { key: 'Full time or part time',
-          value: course_option.study_mode.humanize },
         { key: 'Location',
           value: course_option.site.name_and_address },
+        { key: 'Full time or part time',
```
mnacos, line 13: Same, the previous order seems correct.

spec/components/provider_interface/completed_offer_summary_component_spec.rb
```diff
@@ -21,8 +21,8 @@ def row_text_selector(row_name, render)
     rows = {
       provider: 0,
       course: 1,
-      full_or_part_time: 2,
-      location: 3,
+      location: 2,
+      full_or_part_time: 3,
```
mnacos, line 25: Same question

mnacos (commented):
Code-wise it's fine, but could we double-check the order is correct in all cases? If the order in all components is correct, maybe we need to update Trello.

app/components/provider_interface/completed_offer_summary_component.rb
despo: I think in my attempt to get through all the make offer components I missed that some had been implemented with the details in the correct order :)

Thanks for flagging. Sorted now.

cwrw (approved):
⏬  looks good, only question is do we want to also do this while we're at it? https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/components/provider_interface/deferred_offer_details_component.rb#L28

PR #4516, opened 2021-04-14
Force study_mode selection after course changes in the wizard

mnacos:
## Context

In the new offer flow, after changing a course the wizard pre-selects the same `study_mode` (full-time/part-time) as the last user choice, on the relevant step.

Change the course selection step to reset the `study_mode` value to `nil`, so that no study_mode is pre-selected when the user lands on the full-time/part-time step.

## Guidance to review

Note: The study mode step is skipped if the selected course is only available as full-time or part-time.

## Link to Trello card

https://trello.com/c/VUDzbeAw

## Things to check

### DIFF
--- a/app/controllers/provider_interface/offer/courses_controller.rb
+++ b/app/controllers/provider_interface/offer/courses_controller.rb
@@ -9,7 +9,8 @@ module ProviderInterface
       end
 
       def create
-        @wizard = OfferWizard.new(offer_store, course_params.to_h)
+        @wizard = OfferWizard.new(offer_store)
+        reset_study_mode_if_course_has_changed_and_update_course
 
         if @wizard.valid_for_current_step?
           @wizard.save_state!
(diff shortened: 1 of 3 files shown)


app/controllers/provider_interface/offer/courses_controller.rb
```diff
@@ -45,6 +47,12 @@ def update
 
     private
 
+      def reset_study_mode_if_course_has_changed_and_update_course
+        new_val = course_params['course_id']
+        @wizard.study_mode = nil if new_val != @wizard.course_id
+        @wizard.course_id = new_val
+      end
```
despo, line 54: Can we use a more descriptive variable name? As we are checking that it's not identical to the wizard's course, I think course_id is sufficient without the `new` prefix.
```suggestion
      def reset_study_mode_if_course_has_changed_and_update_course
        course_id = course_params['course_id']
        @wizard.study_mode = nil if course_id != @wizard.course_id
        @wizard.course_id = course_id
      end
```

despo (approved):
Left some minor feedback but other than that it looks good. 
Happy for it to be merged after the feedback is addressed!

PR #4517, opened 2021-04-14
Pentest 410 email all when new device log in

george-bit:
## Context

In order to comply with the results of a recent pentest, Candidates and Providers need to be notified by email of sign ins from new devices for security reasons. This is currently already implmented for support are made to the and candidate interface sign in controller. Changes for providers are made to the on the existing functionality for Support Users.

Emails have been created and logic implemented to that they first. Please test in the review app/qa.

image
image

## Guidance to review

Dev - Check test coverage and functionality, and respond to my

Content/design - Please review email content and provide revisions as necessary. this can be viewed in the support interface of the review app

## Link to Trello card

https://trello.com/c/tSwJMeMf/3197-pentest-410-email-candidates-and-providers-when-they-log-into-a-new-device-low-%F0%9F%9A%A7

## Things to check


app/views/candidate_mailer/confirm_sign_in.text.erb
```diff
@@ -0,0 +1,12 @@
+Dear <%= @candidate_user.email_address %>
```
george-bit, line 1: N.B. Due to magic link log in we  may only have access to candidate email at this stage, so I defaulted to this in case account does not yet hav a a name provided.

spec/system/candidate_interface/signup_and_signin/candidate_signs_in_from_new_device_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+    within 'header' do
+      expect(page).to have_content current_candidate.email_address
+    end
+
+    click_link 'Sign out'
+    expect(page).to have_current_path(candidate_interface_create_account_or_sign_in_path)
+  end
+
+  def when_i_sign_in_from_a_new_device
+    allow_any_instance_of(ActionDispatch::Request).to receive(:remote_ip).and_return('192.168.0.1')
```
george-bit, line 44: I'm having an issue with Rubocop not allowing me to use `allow_any_instance_of` I can't think of a different way to stub this? Can anyone help?

app/views/candidate_mailer/confirm_sign_in.text.erb
paulrobertlloyd: Pretty sure we can get rid of this line altogether. Whatcha think @Tomcowan1?

Tomcowan1: Hello - I'll take a look through the content and come back with edits. One thing that does come to mind is whether people will know their devices by IP addresses? Is it possible to base it on something like location instead?

george-bit: Thanks Tom, 

This email will only fire if there is a log in to an account from a device with different credentials. If a user logs in say from their phone it should be fairly obvious that this is the case as their IP address will stay the same (provided they're using the same internet) but the listed browser will change.

Admittedly, a location would make things slightly clearer. I have just had a look into whether we can access this from the same request or work it out using Geocoder but unfortunately my knowledge runs out here. @malcolmbaig would it be possible to pull through location here?

If it is possible we can amend this update this or update a future PR, depending on the complexity.

Tomcowan1:
Hi @george-bit have taken a look at the content - just a couple of tweaks really. I agree with @paulrobertlloyd Paul, I think we can just take out the "Dear email address" line:

**New sign in to Apply for teacher training**

We detected you have signed in to Apply for teacher training from a new device. 

**New device details**
IP address
Browser
Gecko

If this was you, you can ignore this email.

If this was not you, contact us immediately on the number below.

Get support

You can chat to a Get Into Teaching adviser online for help and advice:

https://getintoteaching.education.gov.uk/#talk-to-us

You can also call for free on 0800 389 2500, Monday to Friday 8:30am to 5pm (except public holidays).

app/views/candidate_mailer/confirm_sign_in.text.erb
paulrobertlloyd: You can get a location from IP address… chances that it’s right however, which could then make people even more concerned when they needn’t be.

Tomcowan1: OK. I was probably channelling the Apple ones where it's clear that you haven't just logged in because it's showing you a completely different country.

george-bit: Cool well we can park it for now and if we get some support calls about it we can always address it then :) Open to something being carded up too.

george-bit:
Thanks @Tomcowan1 - would the provider email need any altering? Would it be better to ping a Prove

george-bit:
Hi @Tomcowan1 amendments have been made as specified

image

Tomcowan1:
Thanks George - looks good. On the version I had on email there's an errand hash tag before "New device details" is that just me?

Screenshot here: https://drive.google.com/drive/folders/1zZgOXX-kzC04y6v5CX8B_OERJxw1y3ds

Tomcowan1:
Email

george-bit:
@Tomcowan1 - Yeqh that was me uploading the wrong screenshot mea cupla 🤦

Tomcowan1 (approved):

george-bit:
@davidgisbey are we okay to merge? Should we test in QA after?

app/mailers/candidate_mailer.rb
```diff
@@ -1,6 +1,16 @@
 class CandidateMailer < ApplicationMailer
   layout 'candidate_email_with_support_footer'
 
+  def confirm_sign_in(candidate_user, device:)
```
davidgisbey, line 4: i think this can just be `candidate` wdyt?

app/controllers/candidate_interface/sign_in_controller.rb
```diff
@@ -96,5 +101,28 @@ def create_from_expired_token
     def candidate_params
       params.require(:candidate).permit(:email_address)
     end
+
+    def send_candidate_sign_in_confirmation_email
+      return if cookies.signed[:sign_in_confirmation] == @local_user.id
+
+      cookies.signed[:sign_in_confirmation] = {
+        value: @local_user.id,
+        expires: 20.years.from_now,
```
davidgisbey, line 110: I think this should be 6 months right? https://www.apply-for-teacher-training.service.gov.uk/candidate/cookies

app/mailers/candidate_mailer.rb
george-bit: Yeah definitely can be, makes sense to be consistent with the code base 👍

app/controllers/candidate_interface/sign_in_controller.rb
george-bit: I think so, am a little uncertain... it would need to be the same for the providers too in that case.

Surely if we set this cookie to expire in six months, this will trigger a new sign in email on a log in after this date is exceeded for both providers and candidates... is this what we want to happen?

app/controllers/dfe_sign_in_controller.rb
```diff
@@ -62,6 +64,25 @@ def send_support_sign_in_confirmation_email
     ).deliver_later
   end
 
+  def send_provider_sign_in_confirmation_email
+    return if cookies.signed[:sign_in_confirmation] == @local_user.id
+
+    cookies.signed[:sign_in_confirmation] = {
+      value: @local_user.id,
+      expires: 20.years.from_now,
```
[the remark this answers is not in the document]
george-bit: Ah 1 step ahead of me 👍

app/mailers/candidate_mailer.rb
george-bit: Should also probably extract the subject lines into the yml file

app/controllers/candidate_interface/sign_in_controller.rb
```diff
@@ -46,11 +47,15 @@ def authenticate
...
         candidate = authentication_token.user
-        flash[:success] = t('apply_from_find.account_created_message') if candidate.last_signed_in_at.nil?
+        first_sign_in = candidate.last_signed_in_at.nil?
+        flash[:success] = t('apply_from_find.account_created_message') if first_sign_in
         sign_in(candidate, scope: :candidate)
         add_identity_to_log(candidate.id)
         authentication_token.use!
 
+        @local_user = get_local_user
+        send_candidate_sign_in_confirmation_email unless first_sign_in
```
duncanjbrown, line 57: Would it be a bit clearer to pass `local_user` to this function explicitly rather than set it (relatively) globally for that function to pick up?

app/views/candidate_mailer/confirm_sign_in.text.erb
```diff
@@ -0,0 +1,11 @@
+# New sign in to Apply for teacher training
+
+We detected you have signed in to Apply for teacher training from a new device.
```
duncanjbrown, line 3:
```suggestion
We detected you signed in to Apply for teacher training from a new device.
```

app/views/candidate_mailer/confirm_sign_in.text.erb
```diff
@@ -0,0 +1,11 @@
...
+
+# New device details
+IP address: <%= @device[:ip_address] %>
+Browser: <%= @device[:user_agent] %>
+
+If this was you, you can ignore this email.
+
+If this was not you, contact us immediately on the number below.
```
duncanjbrown, line 11: for consistency?

```suggestion
If this was not you, please contact us immediately.
```

app/views/provider_mailer/confirm_sign_in.text.erb
```diff
@@ -0,0 +1,12 @@
+Dear <%= @provider_user.display_name %>
+
+# New sign in to Apply for teacher training
```
duncanjbrown, line 3:
```suggestion
# New sign in to Manage teacher training applications
```

app/views/provider_mailer/confirm_sign_in.text.erb
```diff
@@ -0,0 +1,12 @@
...
+
+We detected you signed into Apply for teacher training on a new device.
```
duncanjbrown, line 5:
```suggestion
We detected you signed into Manage teacher training applications on a new device.
```

spec/mailers/previews/candidate_mailer_preview.rb
```diff
@@ -1,4 +1,14 @@
 class CandidateMailerPreview < ActionMailer::Preview
+  def confirm_sign_in
```
duncanjbrown, line 2: Previews ⭐

spec/system/candidate_interface/signup_and_signin/candidate_signs_in_from_new_device_spec.rb
duncanjbrown: I think you could probably set headers on Capybara?

spec/system/provider_interface/provider_signs_in_from_new_device_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+
+    when_i_visit_the_provider_interface_applications_path
+    then_i_am_redirected_to_the_provider_sign_in_path
+
+    when_i_sign_in_via_dfe_sign_in
+    then_i_am_redirected_to_the_provider_interface_applications_path
+
+    when_i_click_sign_out
+    and_sign_in_from_a_different_device
+    then_i_receive_a_confirmation_email_with_correct_details
```
duncanjbrown, line 20: I'd give it one more turn of the handle here — if you then come back and sign in with the same device, do you get another email?

duncanjbrown (commented):
Looking great — couple of qs inline!

app/views/candidate_mailer/confirm_sign_in.text.erb
george-bit: Hey Duncan,

@Tomcowan1 has made some revisions to the content on the candidate email. Should the provider and candiate emails be the same?

Happy to make the suggested changes to the provider content for the time being

app/controllers/candidate_interface/sign_in_controller.rb
george-bit: Hmm just had a crack at this. The issue I'm gaving is that the `@local_user` variable is also being called within the `get_local_user` method itself. By replacing the global variable with an instance variable it breaks the method.

Wecould just leave things as is, unless you can think of a better way?

duncanjbrown: Looking at this again, maybe we don't even need `local_user`, given that we already have a `current_candidate` facility and we only call the `send_candidate_sign_in_confirmation_email` once they're signed in. Maybe I'd add a `raise` in `send_candidate_sign_in_confirmation_email` tho?

```ruby
def send_candidate_sign_in_confirmation_email
  raise 'Tried to send a confirmation email to a nonexistent candidate' unless current_candidate

  return if cookies.signed[:sign_in_confirmation] == current_candidate.id
  # etc...
end
```

app/views/candidate_mailer/confirm_sign_in.text.erb
duncanjbrown: Ha — I'm not a content designer. But at least if we're saying "please" to providers we can probably say it to candidates too?

PR #4518, opened 2021-04-15
Fix Rejection by default applications in the API

cwrw:
## Context

If an application RDB'd, we are not returning the rejection object in the API correctly as we only check for the presence of rejection reasons or withdrawal reasons .

Return the rejection date and a default reason if an application was rejected by default, but no feedback has been supplied yet.

## Guidance to review

anything I missed?
## Link to Trello card

trello ticket: https://trello.com/c/qhj5L5Mm/3595-rbdd-applications-dont-always-have-a-rejection-object-in-the-api

## Things to check

### DIFF
--- a/app/presenters/vendor_api/single_application_presenter.rb
+++ b/app/presenters/vendor_api/single_application_presenter.rb
@@ -95,6 +95,11 @@ module VendorAPI
           reason: application_choice.offer_withdrawal_reason,
           date: application_choice.offer_withdrawn_at.iso8601,
         }
+      elsif application_choice.rejected_by_default?
+        {
+          reason: 'Not entered',
+          date: application_choice.rejected_at.iso8601,
+        }
       end
     end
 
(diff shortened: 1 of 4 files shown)


app/presenters/vendor_api/single_application_presenter.rb
```diff
@@ -95,6 +95,11 @@ def get_rejection
           reason: application_choice.offer_withdrawal_reason,
           date: application_choice.offer_withdrawn_at.iso8601,
         }
+      elsif application_choice.rejected_by_default?
```
duncanjbrown, line 98: I wonder if there are any older applications without rejection reasons and which aren't RBD?

duncanjbrown (approved):
Looks good — one idle question!

app/presenters/vendor_api/single_application_presenter.rb
duncanjbrown: great ⭐

PR #4522, opened 2021-04-15
Fix bugs relating to not accounting for the interviewing state

benswannack:
## Context
There are a few issues around not accounting for the interviewing state.

Main fix is to send candidates the correct email when they get an offer. At the moment, we send a 'single_offer' email, where we should send a 'decisions_pending' email.

## Guidance to review
I've put some specs in that failed before the changes. Do we want to cover any other cases here too?

### DIFF
--- a/app/components/candidate_interface/application_dashboard_course_choices_component.rb
+++ b/app/components/candidate_interface/application_dashboard_course_choices_component.rb
@@ -99,7 +99,7 @@ module CandidateInterface
 
     def interview_row(application_choice)
       return unless application_choice.interviews.kept.any? ||
-        application_choice.awaiting_provider_decision?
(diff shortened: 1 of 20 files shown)


app/components/candidate_interface/application_dashboard_course_choices_component.rb
```diff
@@ -99,7 +99,7 @@ def title_for(application_choice)
 
     def interview_row(application_choice)
       return unless application_choice.interviews.kept.any? ||
-        application_choice.awaiting_provider_decision?
+        application_choice.decision_pending?
```
benswannack, line 102: Not really necessary, but there's no harm using this method here

despo (approved):
👍🏼

PR #4523, opened 2021-04-15
Whitelist host machine's IP address for HostAuthorization and use http health check for web app

vigneshmsft:
## Context

cf http healthcheck runs on each instance of the app and the hostname used is the local IP address of the instance.
So using a http health check while using HostAuthorization on the web app fails without whitelisting this IP address of the app instance.

Add the IP address of the host instance to `config.hosts`

## Things to check

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -116,6 +116,11 @@ Rails.application.configure do
     config.hosts << host
   end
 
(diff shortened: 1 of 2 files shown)


vigneshmsft:
Closing this, see #4529

PR #4526, opened 2021-04-16
Fix `govuk_button` not rendering correctly

cwrw:
## Context

The current button isn't rendering correctly  and without any styles
Screenshot 2021-04-16 at 11 42 20

Add a method which ensures the correct styles are rendered

## Guidance to review

I've checked other occurrences and this is the only one.

## Link to Trello card

trello-ticket: https://trello.com/c/t0D1R5m9/3624-change-link-for-interview-with-one-organisation

## Things to check

### DIFF
--- a/app/views/provider_interface/interviews/check.html.erb
+++ b/app/views/provider_interface/interviews/check.html.erb
@@ -13,7 +13,7 @@
     <% if @interview.present? %>
       <%= govuk_button_to t('.update.confirm'), { action: 'update', id: @interview.id }, method: :put %>
     <% else %>
-      <%= govuk_button_to t('.confirm'), action: 'commit' %>
+      <%= govuk_button_to t('.confirm'), { action: 'commit' }, method: :post %>
     <% end %>
 
     <p class="govuk-body">


app/views/provider_interface/interviews/check.html.erb
```diff
@@ -13,7 +13,7 @@
     <% if @interview.present? %>
       <%= govuk_button_to t('.update.confirm'), { action: 'update', id: @interview.id }, method: :put %>
     <% else %>
-      <%= govuk_button_to t('.confirm'), action: 'commit' %>
+      <%= govuk_button_to t('.confirm'), { action: 'commit' }, method: :post %>
```
paulrobertlloyd, line 16: @peteryates This seems surprising to me. Do HTML attributes always need to be in `{ }`, and why not `method`?

despo: I think it has to do with the button_to implementation https://apidock.com/rails/ActionView/Helpers/UrlHelper/button_to

despo (approved):

app/views/provider_interface/interviews/check.html.erb
peteryates: Ah, `#govuk_button_to` is a pretty much a wrapper for [`#button_to`](https://api.rubyonrails.org/v6.1.0/classes/ActionView/Helpers/UrlHelper.html#method-i-button_to), which takes its `options` argument as the second arg and `html_options` as the third (which here is `method: :post`).

The only thing the govuk variant does is forces in a `govuk-button` class.

PR #4527, opened 2021-04-16
Apply the discussed fixes to the application dashboard

davidgisbey:
## Context

After reviewing the recent changes to the dashboard there were a few minor bugs that needed fixing.

- Remove `Before your offer is confirmed, there are some final conditions to meet.` for deferred offers
- `Two of your training providers have made a decision on your application.` should be `2 of your training providers have made a decision on your application.`
- Remove the full stop from `You’ve declined your offer. `

## Link to Trello card

https://trello.com/c/xpNnuh0R/3302-dashboard-walkthrough-changes

## Things to check

### DIFF
--- a/app/components/candidate_interface/application_dashboard_guidance_component.html.erb
+++ b/app/components/candidate_interface/application_dashboard_guidance_component.html.erb
@@ -10,7 +10,7 @@
 
 <% elsif has_multiple_offers_but_awaiting_decisions? %>
 
-  <p class="govuk-body"><%= number_of_offers_in_words %> of your training providers have made a decision on your application.</p>
+  <p class="govuk-body">2 of your training providers have made a decision on your application.</p>
   <p class="govuk-body">Our <%= govuk_link_to('teacher training advisers', 'https://beta-adviser-getintoteaching.education.gov.uk/') %> can help you prepare for any interviews.</p>
 
 <% elsif has_single_offer_but_awaiting_decisions? %>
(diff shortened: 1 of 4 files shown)


app/components/candidate_interface/application_dashboard_guidance_component.html.erb
```diff
@@ -10,7 +10,7 @@
 
 <% elsif has_multiple_offers_but_awaiting_decisions? %>
 
-  <p class="govuk-body"><%= number_of_offers_in_words %> of your training providers have made a decision on your application.</p>
+  <p class="govuk-body">2 of your training providers have made a decision on your application.</p>
```
paulrobertlloyd, line 13: Could this number not also be 3 (in which case it should possibly say ‘All’). Or is there a separate state for that?

davidgisbey: You can't be awaiting a decision and have 3 offers right? We will have to revisit this once a candidate can apply to 5 choices though.

stevehook: Wouldn't it just make sense to use `<%= number_of_offers %>` so that we don't have to revisit?

stevehook (approved):
One question inline but for me it's good to merge as is.

app/components/candidate_interface/application_dashboard_guidance_component.html.erb
davidgisbey: Yea tbh i didn't really think about that at the time. I'll revert now

davidgisbey: So i think it would end up being something like

`  <p class="govuk-body"><%= @application_form.application_choices.offer.count %> of your training providers have made a decision on your application.</p> 
`

right? Since it should be numbers rather than words.

PR #4528, opened 2021-04-16
Add data export history view to support interface

stevehook:
## Context

Support users need to see a complete history of data exports that have been requested, their status, when they ran and by whom.

- [x] Add a new route, controller action and view to implement the new page
image

- [x] Add a link to the data directory index page to access the history
image

- [x] Update the data directory system spec
- [x] Extract a `DataExportHistoryComponent` and applied this to the existing single-type history page as well as the new one. On the single-type page this component doesn't need to display the _Name_ column because they are all the same and it's shown in the page title already.

image

## Guidance to review

- UX and content just represents my best guess at this stage. Do we need to change the information displayed and layout on the new page? Is there a better place for the _History_ link on the index page?
- Are the tests sufficient?
- Is there a better name for the route?

## Link to Trello card

https://trello.com/c/4nRf96iG/3268-add-export-history-page

## Things to check

### DIFF
--- a/app/components/support_interface/data_export_history_component.html.erb
+++ b/app/components/support_interface/data_export_history_component.html.erb
@@ -1,0 +1,48 @@
+<% if data_exports.empty? %>
+  <p class="govuk-body">No exports have been generated</p>
+<% else %>
+  <table class="govuk-table">
+    <thead class="govuk-table__head">
+      <tr class="govuk-table__row">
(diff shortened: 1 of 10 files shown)


davidgisbey (approved):
Tested locally and it's working beatifully. I've left one comment about potentially adding a component, but tbh i think it's fine to go as is

davidgisbey:
The component is looking great 🙌

PR #4535, opened 2021-04-19
Add custom colors to TaskListItemComponent

malcolmbaig:
## Context

We're adding 'Review' badges to certain sections of the application when a candidate applied again. Just as we currently allow for custom statuses to be rendered by this component, we also need to ability to specify colors that aren't supported by default.

Add a `custom_color` argument to the component.

With `custom_color: green` —

Screenshot_20210419_083612

## Guidance to review
Per commit would be useful if only because there's also a small cleanup commit that removes an unused argument.

## Things to check


davidgisbey (approved):
Very nice!

PR #4536, opened 2021-04-19
Fix RBD date calculation in specs

cwrw:
## Context

RBD is always set to EOD, the current spec is setting it to midday, skewing the calculation of when the RBD date should be displayed

Change the RBD to end of day and fix the specs accordingly.
## Things to check

### DIFF
--- a/spec/components/provider_interface/application_card_component_spec.rb
+++ b/spec/components/provider_interface/application_card_component_spec.rb
@@ -157,13 +157,13 @@ RSpec.describe ProviderInterface::ApplicationCardComponent do
     end
 
     context 'when 1 day is left to respond' do
-      let(:rbd) { 2.days.from_now.midday }
+      let(:rbd) { 1.day.from_now.end_of_day }
 
       it { is_expected.to eq('1 day to make decision') }
     end
 
     context 'when 2 days are left to respond' do
-      let(:rbd) { 3.days.from_now.midday }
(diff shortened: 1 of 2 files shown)


spec/components/provider_interface/application_card_component_spec.rb
```diff
@@ -157,13 +157,13 @@
     end
 
     context 'when 1 day is left to respond' do
-      let(:rbd) { 2.days.from_now.midday }
+      let(:rbd) { 1.days.from_now.end_of_day }
```
malcolmbaig, line 160:
```suggestion
      let(:rbd) { 1.day.from_now.end_of_day }
```

malcolmbaig (approved):

davidgisbey (approved):
that was me sorry!

spec/services/set_decline_by_default_spec.rb
```diff
@@ -33,11 +33,11 @@ def expect_all_relevant_decline_by_default_at_values_to_be(expected)
 
     context 'when all the application choices have an offer' do
       it 'the DBD is set to 10 business days from the date of the most recent offer' do
-        choices[0].update(status: :offer, offered_at: 1.business_days.before(now).end_of_day)
+        most_recent_offer_date = 1.business_days.before(now).end_of_day
```
malcolmbaig, line 36: 👌🏼

malcolmbaig (approved):
LGTM

PR #4538, opened 2021-04-19
Add trailing text to indicate approximate dates for work experiences

richardpattinson:
## Context

Update manage interface to indicate if work experience start/end dates have been flagged as 'approximate' by candidate. 

## Link to Trello card

https://trello.com/c/p5bxyt7C/3567-work-history-mark-approximate-dates-in-manage

## Things to check

### DIFF
--- a/app/components/work_history_item_component.rb
+++ b/app/components/work_history_item_component.rb
@@ -42,11 +42,19 @@ private
   attr_accessor :item
 
   def formatted_start_date
+    if item.is_a?(ApplicationWorkExperience) && item.start_date_unknown
+      return "#{item.start_date.to_s(:month_and_year)} (approximate)"
+    end
+
     item.start_date.to_s(:month_and_year)
   end
 
   def formatted_end_date
     return 'Present' if item.end_date.nil?
+
+    if item.is_a?(ApplicationWorkExperience) && item.end_date_unknown
+      return "#{item.end_date.to_s(:month_and_year)} (approximate)"
+    end
 
     item.end_date.to_s(:month_and_year)
   end
--- a/spec/components/work_history_component_spec.rb
+++ b/spec/components/work_history_component_spec.rb
@@ -191,4 +191,53 @@ RSpec.describe WorkHistoryComponent do
       expect(rendered.text).to include 'Pig herder - Part time'
     end
   end
+
+  context 'with work experiences with approximate start and end dates' do
+    it 'renders work experience details, explained break and approximate dates' do
+      application_form = instance_double(ApplicationForm, submitted_at: Time.zone.local(2020, 2, 1))
+      breaks = [
+        build(
+          :application_work_history_break,
+          start_date: Date.new(2018, 2, 1),
+          end_date: Date.new(2019, 12, 1),
+          reason: 'I found sheep farming very stressful and needed to take time off work',
+        ),
+      ]
+      experiences = [
+        build(
+          :application_work_experience,
+          start_date: Date.new(2014, 10, 1),
+          end_date: Date.new(2018, 2, 1),
+          start_date_unknown: true,
+          end_date_unknown: true,
+          role: 'Sheep herder',
+          commitment: 'full_time',
+          working_pattern: '',
+          organisation: 'Bobs Farm',
+          details: 'Livestock management',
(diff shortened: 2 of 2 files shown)


app/components/work_history_item_component.rb
```diff
@@ -38,13 +38,29 @@ def organisation
...
-    item.start_date.to_s(:month_and_year)
+    if FeatureFlag.active?(:restructured_work_history)
+      if item.is_a? ApplicationWorkExperience
+        item.start_date_unknown ? item.start_date.to_s(:month_and_year) + ' (approximate)' : item.start_date.to_s(:month_and_year)
+      else
+        item.start_date.to_s(:month_and_year)
+      end
+    else
+      item.start_date.to_s(:month_and_year)
+    end
```
despo, line 49: Consider simplifying the condition and also use interpolaration to construct the string.

```suggestion
    if FeatureFlag.active?(:restructured_work_history) 
      if item.is_a?(ApplicationWorkExperience) && item.start_date_unknown 
         return "#{item.start_date.to_s(:month_and_year)} (approximate)" 
       end
    end
    
    item.start_date.to_s(:month_and_year)
```

app/components/work_history_item_component.rb
```diff
@@ -38,13 +38,29 @@ def organisation
...
-    item.end_date.to_s(:month_and_year)
+    if FeatureFlag.active?(:restructured_work_history)
+      if item.is_a? ApplicationWorkExperience
+        item.end_date_unknown ? item.end_date.to_s(:month_and_year) + ' (approximate)' : item.end_date.to_s(:month_and_year)
+      else
+        item.end_date.to_s(:month_and_year)
+      end
+    else
+      item.end_date.to_s(:month_and_year)
+    end
```
despo, line 63: Similar to above.


```suggestion
    if FeatureFlag.active?(:restructured_work_history) 
      if item.is_a?(ApplicationWorkExperience) && item.end_date_unknown 
         return "#{item.end_date.to_s(:month_and_year)} (approximate)" 
       end
    end
    
    item.end_date.to_s(:month_and_year)
```

despo:
Given the duplication can this code also be extracted to a reusable private method.

app/components/work_history_item_component.rb
```diff
@@ -38,12 +38,24 @@ def organisation
   attr_accessor :item
 
   def formatted_start_date
+    if FeatureFlag.active?(:restructured_work_history)
```
duncanjbrown, line 41: Is it right to feature flag this? We'll have to handle both kinds of work history (old and new structures) in production. I think `start_date_unknown` will only be set for "restructured" entries.

spec/components/work_history_component_spec.rb
```diff
@@ -167,4 +167,54 @@
...
+        ),
+      ]
+      allow(application_form).to receive(:application_work_experiences).and_return(experiences)
+      allow(application_form).to receive(:application_work_history_breaks).and_return(breaks)
+
+      rendered = render_inline(described_class.new(application_form: application_form))
+      expect(rendered.text).to include 'October 2014 (approximate) - February 2018 (approximate)'
+      expect(rendered.text).to include 'Sheep herder - Full time'
+      expect(rendered.text).to include 'Break (1 year and 10 months)'
+      expect(rendered.text).to include 'I found pig farming very stressful and needed to take time off work'
```
duncanjbrown, line 215:
```suggestion
      expect(rendered.text).to include 'I found sheep farming very stressful and needed to take time off work'
```

duncanjbrown:
Minor comments inline, otherwise looking good!

app/components/work_history_item_component.rb
richardpattinson: My understanding was that the candidate side changes (the approximate flag) was behind a feature flag. I assumed therefore that things that depend on this flag should be too. However it looks like the DB changes have already been made so I can remove the flag.

spec/components/work_history_component_spec.rb
richardpattinson: Good spot wouldnt want anyone to be confused out of their mind.

cwrw (approved):

PR #4539, opened 2021-04-19
 Support browser back button in Make offer flow

despo:
## Context
Currently when a user goes to the previous Make offer step using the browser back button, it causes the flow to break, as it only has knowledge of what the current step is through the `new` and `edit` offer controller actions.

I tried adding some system specs but I was getting some unrelated errors when enabling javascript. Any suggestions on how to approach this? Do we need to test this?

Ensure that current step is also set as part of submit actions (`create` and `update`). This should enable the wizard to always return the correct next step and avoid any errors.

## Link to Trello card
https://trello.com/c/MjEArmnE/3617-investigate-supporting-browser-back-button-in-make-offer-flow

## Things to check

### DIFF
--- a/app/controllers/provider_interface/offer/conditions_controller.rb
+++ b/app/controllers/provider_interface/offer/conditions_controller.rb
@@ -7,7 +7,7 @@ module ProviderInterface
       end
 
       def create
-        @wizard = OfferWizard.new(offer_store, conditions_attributes_for_wizard)
+        @wizard = OfferWizard.new(offer_store, conditions_attributes_for_wizard.merge!(current_step: 'conditions'))
 
         if @wizard.valid_for_current_step?
           @wizard.save_state!
@@ -24,7 +24,7 @@ module ProviderInterface
       end
 
       def update
-        @wizard = OfferWizard.new(offer_store, conditions_attributes_for_wizard)
+        @wizard = OfferWizard.new(offer_store, conditions_attributes_for_wizard.merge!(current_step: 'conditions'))
 
         if @wizard.valid_for_current_step?
           @wizard.save_state!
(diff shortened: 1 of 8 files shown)


app/controllers/provider_interface/offer/conditions_controller.rb
```diff
@@ -7,7 +7,7 @@ def new
       end
 
       def create
-        @wizard = OfferWizard.new(offer_store, conditions_attributes_for_wizard)
+        @wizard = OfferWizard.new(offer_store, conditions_attributes_for_wizard.merge!(current_step: 'conditions'))
```
benswannack, line 10: Is it the case that the current step will be the same for all actions in each of the controllers? Could we move the `merge` to somewhere common?

benswannack:
Can L27 of WizardPathHistory now be removed? Maybe we can enforce in there that a step is passed in?

cwrw (approved):
I had a look at the back button and implementations in specs, as Im wary of enabling JS for this as well. I found a method from Capybara but I couldn't get it to work locally: `go_back` as I think you need to register a driver for that to work, it doesn't with the current one, maybe something to look into briefly? But otherwise happy to keep as is also.

I tested locally and works perfectly! Really happy with the simple solution ⭐  
Past the suggestion around drying up merging the current step if possible (maybe in the params of the controller eg. `conditions_attributes_for_wizard`) looks good to me 👍 

I had a question around `OfferWizard.new` in the `offers_controller`, just wanted to double check we don't need to set anything there as well?

despo:
@cwrw we shouldn't need to set anything up in the OffersController as in that instance the validation scope we use will always be 'save' (to ensure that all validations take place before submission).

benswannack (approved):
Nice!

PR #4542, opened 2021-04-19
Remove international degrees from the autocomplete export

davidgisbey:
## Context

We use two different mechanisms to identify international qualifications. One of which is `qualification_type: 'non_uk'` for GCSEs and OtherQuals and `international: true` for degrees.

We should be using one mechanism going forward. I'll card something up, but for now let's exclude them from the export 

- Remove international degrees from the autofill export 

## Things to check


TobyRet (approved):
LGTM!

PR #4545, opened 2021-04-19
Indicate this candidate has used relevant skills for work experience

steventux:
## Context

If the candidate has indicated that work experience has been relevant to their application by answering a new question as part of the application then prefer this over the older 'working with children' flag.

If the restructured work history feature flag is active and the candidate has indicated their work experience used relevant skills then rendering this.

image

## Guidance to review

Enable the `restructured_work_history` feature flag and update a work experience item so that `relevant_skills: true`.

## Link to Trello card

https://trello.com/c/oxeICHx0/3571-work-history-marking-roles-that-are-relevant-to-teaching-on-a-candidates-application-in-manage

## Things to check

### DIFF
--- a/app/components/work_history_item_component.html.erb
+++ b/app/components/work_history_item_component.html.erb
@@ -10,7 +10,14 @@
   </p>
   <p class="govuk-body">
     <%= details %>
-    <% if working_with_children? %>
+    <% if relevant_skills? %>
+      <p class="govuk-body govuk-!-font-size-16">
+        <svg class="app-icon govuk-!-margin-right-1" xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 200 200" focusable="false" fill="currentColor" aria-hidden="true">
+          <path d="M100 200a100 100 0 1 1 0-200 100 100 0 0 1 0 200zm-60-85l40 40 80-80-20-20-60 60-20-20-20 20z"></path>
+        </svg>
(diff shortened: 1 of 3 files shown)


app/components/work_history_item_component.html.erb
```diff
@@ -10,7 +10,14 @@
   </p>
   <p class="govuk-body">
     <%= details %>
-    <% if working_with_children? %>
+    <% if FeatureFlag.active?(:restructured_work_history) && relevant_skills? %>
```
duncanjbrown, line 13: Do we need the feature flag here? Once we turn it on we'll still have to maintain backwards compatibility

app/components/work_history_item_component.rb
```diff
@@ -29,6 +29,10 @@ def working_with_children?
     item.respond_to?(:working_with_children?) && item.working_with_children?
   end
 
+  def relevant_skills?
+    item.respond_to?(:relevant_skills?) && item.relevant_skills?
```
duncanjbrown, line 33: `item.try(:relevant_skills?)`?

app/components/work_history_item_component.html.erb
steventux: :+1: No I don't think it does, these conditional methods are guarded enough to fall through to the old flag.

duncanjbrown (approved):
:+1::+1::+1:

PR #4546, opened 2021-04-19
Start writing current_course_option_id

mnacos:
## Context

To reduce the risk of nil `current_course_option_id` values when we deploy https://github.com/DFE-Digital/apply-for-teacher-training/pull/4499 here is some temporary code to start populating this field when new course choices are made (candidate interface) or when course details change as part of making an offer (provider interface).

Start setting `current_course_option_id`.

## Guidance to review

We're setting `current_course_option_id` instead of `current_course_option` in the create/update statements because no ActiveRecord associations have been defined yet.

## Link to Trello card

https://trello.com/c/oYjeu4Hh

## Things to check

### DIFF
--- a/app/forms/candidate_interface/pick_site_form.rb
+++ b/app/forms/candidate_interface/pick_site_form.rb
@@ -20,13 +20,17 @@ module CandidateInterface
 
       application_form.application_choices.create!(
         course_option: course_option,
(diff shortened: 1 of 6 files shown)


cwrw (approved):
Looks good! Could you add specs otherwise 👍

PR #4549, opened 2021-04-20
WIP - Use radio buttons to complete individual sections instead of checkboxes

TobyRet:
## Context

```
As a... candidate completing my application
I need to... mark sections as complete
So that... I know which parts of my application still need to be completed
```

Rather than use a check box we want to now use radio buttons to indicate whether a section is complete

https://apply-beta-prototype.herokuapp.com/application/12345

Seeking some early feedback on this!

I have tried to create a generic form object that be used everywhere. The challenge however is that the state of each section that is completed is stored as a separate column in the db. I've had to use some potentially horrible meta programming to create the corresponding `attr_accessor` and validator in the form object.

What do people think?

## Link to Trello card

https://trello.com/c/MgPkClPi/3303-use-radio-buttons-to-complete-individual-sections-instead-of-checkboxes

## Things to check

### DIFF
--- a/app/components/candidate_interface/complete_section_component.html.erb
+++ b/app/components/candidate_interface/complete_section_component.html.erb
@@ -1,8 +1,12 @@
-<%= form_with model: @application_form, url: @path, method: @request_method do |f| %>
-  <div class='govuk-form-group'>
-    <%= f.hidden_field @field_name, value: false %>
-    <%= f.govuk_check_box @field_name, true, multiple: false, label: { text: t('application_form.completed_checkbox') } %>
-  </div>
+<%= form_with model: completion_form, url: path, method: request_method do |f| %>
+  <%= f.govuk_error_summary %>
+
+  <%= render(summary) %>
+
+  <%= f.govuk_radio_buttons_fieldset field_name, legend: { text: 'Have you completed this section?', size: 'm' } do %>
+    <%= f.govuk_radio_button field_name, true, label: { text: t('application_form.completed_radio') }, link_errors: true%>
+    <%= f.govuk_radio_button field_name, false, label: { text: t('application_form.incomplete_radio') } %>
+  <% end %>
 
   <%= f.govuk_submit t('continue') %>
 <% end %>
--- a/app/forms/candidate_interface/completion_form.rb
+++ b/app/forms/candidate_interface/completion_form.rb
@@ -1,0 +1,32 @@
+module CandidateInterface
+  class CompletionForm
+    include ActiveModel::Model
+
+    def self.build_from_application(attr_hash)
+      attr = attr_hash.keys.first
+
+      create_attr_accessor_from(attr)
+      create_validation_from(attr)
+
+      new(attr_hash)
+    end
+
+    def self.create_attr_accessor_from(attr)
+      instance_eval do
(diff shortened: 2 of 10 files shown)


app/components/candidate_interface/complete_section_component.html.erb
```diff
@@ -1,8 +1,12 @@
...
-  <div class='govuk-form-group'>
-    <%= f.hidden_field @field_name, value: false %>
-    <%= f.govuk_check_box @field_name, true, multiple: false, label: { text: t('application_form.completed_checkbox') } %>
-  </div>
+<%= form_with model: completion_form, url: path, method: request_method do |f| %>
+  <%= f.govuk_error_summary %>
+
+  <%= render(summary) %>
+
+  <%= f.govuk_radio_buttons_fieldset field_name, legend: { text: 'Have you completed this section?', size: 'm' } do %>
```
TobyRet, line 6: @davidgisbey rightly pointed out that the CompleteForm can be simplified further if we use a standardised field (eg `:complete`)

app/forms/candidate_interface/completion_form.rb
```diff
@@ -0,0 +1,32 @@
+module CandidateInterface
+  class CompletionForm
+    include ActiveModel::Model
+
+    def self.build_from_application(attr_hash)
+      attr = attr_hash.keys.first
+
+      create_attr_accessor_from(attr)
+      create_validation_from(attr)
```
TobyRet, line 9: This will not be required if we have a standardised field name
https://github.com/DFE-Digital/apply-for-teacher-training/pull/4549/files#r616524868

app/forms/candidate_interface/completion_form.rb
```diff
@@ -0,0 +1,32 @@
...
+
+      new(attr_hash)
+    end
+
+    def self.create_attr_accessor_from(attr)
+      instance_eval do
+        attr_accessor attr
+      end
```
stevehook, line 17: I'm not sure this `instance_eval` is doing what you expect(?). It adds the attribute to the class rather than the instance (and the relevant instance hasn't been created yet).

```
[1] pry(main)> f = CandidateInterface::CompletionForm.build_from_application(foo: 'foo')
=> #<CandidateInterface::CompletionForm:0x00007fa0994a4bb0 @foo="foo">
[2] pry(main)> f.foo
=> "foo"
[3] pry(main)> b = CandidateInterface::CompletionForm.build_from_application(bar: 'bar')
=> #<CandidateInterface::CompletionForm:0x00007fa099567d18 @bar="bar">
[4] pry(main)> b.foo
=> nil
[5] pry(main)> b.respond_to?(:foo)
=> true
[6] pry(main)> f.respond_to?(:bar)
=> true
[7] pry(main)>
```

I guess this works anyway because the `CandidateInterface::CompletionForm` will be given the attributes it needs before it needs them and it doesn't matter that it accumulates them all over time because they won't interfere.

But I'm not sure about the validation, as soon as you add a second attribute the form will never be valid...

```
[8] pry(main)> f.valid?
=> false
from (pry):8:in `<main>'
[9] pry(main)> f.errors.full_messages
=> ["Bar can't be blank"]
```

PR #4551, opened 2021-04-20
Add review functionality to personal statement section

malcolmbaig:
## Context

As a.... candidate
In order to... improve my chance of success when applying again
I want to... review the sections of my application that providers said needed improvement.

We want to encourage candidates to review feedback from past,
unsuccessful applications. This commit adds a review workflow to the
'Why do you want to teach' section of the personal statement.

This new workflow is implemented as a set of cosmetic changes on top of
the existing 'completed' flags we have for each section. This means we
can use the existing database column and preserve the same validation
behaviour while changing only the messaging the candidate sees.

Broadly, this change does the following:

- Mark the 'becoming_a_teacher' section as incomplete when applying
  again if there is structured feedback for this section in the
  previous application form.
- Show a 'Review' badge for this section on the application form page.
- Add a RejectionReasonsHistory class that handles retrieval of past
  rejection reasons, providing methods to return the entire rejection
  history of a section or just the most recent one.
- Summarize the rejection history of the section when reviewing it.

Finally, we also add updated error messaging for this section on app submit.

**Review badge on application form page**:

Screenshot_20210420_122710

**Inline rejection reasons for the section**:

Screenshot_20210420_122627

**Error on app submit**:

Screenshot_20210420_122427-1

## Guidance to review
- Per commit highly recommended. There are a couple of precursor commits and the system spec has been added on its own to outline the expected behaviour.
- Manual testing will need an unsuccessful application with structured feedback in `quality_of_application_personal_statement_what_to_improve`. The review workflow can then be tested by applying again.

## Link to Trello card

https://trello.com/c/BkHTZAJE
## Things to check

### DIFF
--- a/app/models/application_form.rb
+++ b/app/models/application_form.rb
@@ -356,6 +356,28 @@ class ApplicationForm < ApplicationRecord
     end
   end
 
+  def mark_sections_incomplete_if_review_needed!
+    if becoming_a_teacher_reviewable?
+      update!(becoming_a_teacher_completed: false)
+    end
+  end
+
+  def becoming_a_teacher_rejection_reasons
+    CandidateInterface::RejectionReasonsHistory.all_previous_applications(self, :becoming_a_teacher)
+  end
+
+  def becoming_a_teacher_most_recent_rejection_reason
+    CandidateInterface::RejectionReasonsHistory.most_recent(self, :becoming_a_teacher)
+  end
+
+  def becoming_a_teacher_review_pending?
+    !becoming_a_teacher_completed? && becoming_a_teacher_reviewable?
+  end
+
+  def becoming_a_teacher_reviewable?
+    apply_2? && becoming_a_teacher_most_recent_rejection_reason.present?
+  end
+
 private
 
   def geocode_address_if_required
--- a/app/models/application_form.rb
+++ b/app/models/application_form.rb
@@ -366,8 +366,8 @@ class ApplicationForm < ApplicationRecord
     CandidateInterface::RejectionReasonsHistory.all_previous_applications(self, :becoming_a_teacher)
   end
 
-  def becoming_a_teacher_most_recent_rejection_reason
-    CandidateInterface::RejectionReasonsHistory.most_recent(self, :becoming_a_teacher)
+  def becoming_a_teacher_previous_application_rejection_reason
+    CandidateInterface::RejectionReasonsHistory.previous_application(self, :becoming_a_teacher)
   end
 
   def becoming_a_teacher_review_pending?
@@ -375,7 +375,7 @@ class ApplicationForm < ApplicationRecord
   end
 
   def becoming_a_teacher_reviewable?
-    apply_2? && becoming_a_teacher_most_recent_rejection_reason.present?
+    apply_2? && becoming_a_teacher_previous_application_rejection_reason.present?
   end
 
 private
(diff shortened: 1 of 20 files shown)


stevehook (commented):
Code looks very clear and correct. I found an edge case issue when I was testing. Steps to reproduce:
1. Try to submit application with review outstanding. Missing review is flagged correctly. 
2. Follow the _Why do you want to teach_ link.
3. Hit _Change_ the personal statement and hit continue without updating it.
4. Then select the _I have reviewed this section_ checkbox and hit continue again. The review label is still displayed because (I guess) the selected checkbox wasn't persisted to the DB.

I like the idea of reusing the existing DB column and making this a 'cosmetic' change. I only wonder whether we might be asked to analyse the effectiveness of this feature (on the feature metrics dashboard) and what that would entail. I guess these changes are going to be seen by all apply again candidates after a certain date so I guess what matters is the proportion of candidates making changes to their personal statement before and after.

app/models/application_form.rb
```diff
@@ -356,6 +356,28 @@ def domicile
...
+      update!(becoming_a_teacher_completed: false)
+    end
+  end
+
+  def becoming_a_teacher_rejection_reasons
+    CandidateInterface::RejectionReasonsHistory.all_previous_applications(self, :becoming_a_teacher)
+  end
+
+  def becoming_a_teacher_most_recent_rejection_reason
+    CandidateInterface::RejectionReasonsHistory.most_recent(self, :becoming_a_teacher)
```
malcolmbaig, line 370: Just realised these method names are a bit misleading — it's the most recent application, not the most recent rejection reason (which may be from the application before last, for example).

Going to rename these to:
`becoming_a_teacher_previous_application_rejection_reason`
`RejectionResaonsHistory.previous_application`

malcolmbaig:
> I only wonder whether we might be asked to analyse the effectiveness of this feature (on the feature metrics dashboard) and what that would entail.

Yeah that's a good point. I think the potentially useful observations are:

- How many candidates were prompted to review?
- Did they change anything?

In both cases we're probably covered by:

- For any given apply_2 application, did the previous one have the relevant rejection reason?
- As you pointed out, is there a difference in personal statement between current application and previous?

stevehook (approved):

app/views/candidate_interface/personal_statement/show.html.erb
```diff
@@ -5,13 +5,26 @@
...
-</ul>
-<p class="govuk-body govuk-!-font-weight-bold">You’ll be able to review it again before you submit your application.</p>
+<% if @application_form.becoming_a_teacher_review_pending? %>
+  <div class='govuk-inset-text govuk-!-width-two-thirds govuk-!-margin-top-0'>
+    <h2 class="govuk-heading-m">Feedback from previous applications</h2>
+
+    <% @application_form.becoming_a_teacher_rejection_reasons.each do |rejection_reason| %>
+      <h3 class="govuk-heading-s govuk-!-margin-bottom-2"><%= rejection_reason.provider_name %></h3>
+      <blockquote class="govuk-!-margin-left-0 govuk-!-margin-top-0">
+        <p class="govuk-body">“<%= rejection_reason.feedback %>”</p>
```
frankieroberto, line 15: Can I check whether newlines get converted into line breaks here? (I don't think we need to worry about any other mark up though, like markdown lists)

malcolmbaig: Good question - they don't, no. Will add to https://github.com/DFE-Digital/apply-for-teacher-training/pull/4569

PR #4552, opened 2021-04-20
Set current_course_option_id for existing application choices

mnacos:
## Context

After deploying https://github.com/DFE-Digital/apply-for-teacher-training/pull/4546, which will start setting `current_course_option_id` for all new applications, we need to deal with existing application choices and populate `current_course_option_id` for them, before starting to read from this field (see https://github.com/DFE-Digital/apply-for-teacher-training/pull/4499).

Extract the data migration from https://github.com/DFE-Digital/apply-for-teacher-training/pull/4499 and deploy this first, to avoid any race conditions.

## Guidance to review

See the spec.

## Link to Trello card

https://trello.com/c/oYjeu4Hh

## Things to check

### DIFF
--- a/app/services/data_migrations/backfill_current_course_option_id.rb
+++ b/app/services/data_migrations/backfill_current_course_option_id.rb
@@ -1,0 +1,17 @@
+module DataMigrations
+  class BackfillCurrentCourseOptionId
+    TIMESTAMP = 20210419154319
+    MANUAL_RUN = false
+
+    def change
+      sql = <<~POPULATE_CURRENT_OPTION_ID_SQL.squish
+        UPDATE application_choices
+           SET current_course_option_id = COALESCE(offered_course_option_id, course_option_id)
+      POPULATE_CURRENT_OPTION_ID_SQL
+
+      ActiveRecord::Base.connection.execute(sql)
+
+      raise 'setting current_course_option_id failed' if ApplicationChoice.where(current_course_option_id: nil).any?
+    end
+  end
+end
--- a/lib/tasks/data.rake
+++ b/lib/tasks/data.rake
@@ -1,5 +1,6 @@
 DATA_MIGRATION_SERVICES = [
   # do not delete or edit this line - services added below by generator
+  'DataMigrations::BackfillCurrentCourseOptionId',
   'DataMigrations::BackfillExportType',
   'DataMigrations::FixLatLongFlipFlops',
 ].freeze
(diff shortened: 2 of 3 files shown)


app/services/data_migrations/backfill_current_course_option_id.rb
```diff
@@ -0,0 +1,17 @@
+module DataMigrations
+  class BackfillCurrentCourseOptionId
+    TIMESTAMP = 20210419154319
+    MANUAL_RUN = false
+
+    def change
+      sql = <<~POPULATE_CURRENT_OPTION_ID_SQL.squish
+        UPDATE application_choices
+           SET current_course_option_id = COALESCE(current_course_option_id, course_option_id)
```
duncanjbrown, line 9: Should this also consider `offered_course_option_id`?

mnacos: That's what it should say! Fixing...

app/services/data_migrations/backfill_current_course_option_id.rb
```diff
@@ -0,0 +1,17 @@
+module DataMigrations
+  class BackfillCurrentCourseOptionId
+    TIMESTAMP = 20210419154319
+    MANUAL_RUN = false
+
+    def change
+      sql = <<~POPULATE_CURRENT_OPTION_ID_SQL.squish
+        UPDATE application_choices
+           SET current_course_option_id = COALESCE(offered_course_option_id, course_option_id)
```
duncanjbrown, line 9: I don't _think_ we need `WHERE current_course_option_id IS NULL` as this will always give us the correct response (ie we've always populating both fields since https://github.com/DFE-Digital/apply-for-teacher-training/pull/4546/files)?

mnacos: I extracted this from https://github.com/DFE-Digital/apply-for-teacher-training/pull/4499 which stops writing to `offered_course_option_id`, where this `WHERE` made more sense.

mnacos: Having said that, I can't actually see this `WHERE` clause in the migration!

duncanjbrown: Haha sorry, I was just imagining it for argument's sake!

duncanjbrown (approved):

PR #4556, opened 2021-04-20
Set application form created date in the past

richardpattinson:
## Context

Spec used to flake when occasionally created_at field != updated_at field. This issue has been fixed by specifying that created_at != updated_at under all circumstances. This is more reflective of the real behaviour of an application (which is created and then references are added later).
## Things to check


duncanjbrown (approved):
⭐  Nice solution!

PR #4558, opened 2021-04-21
Prompt candidates with GCSE fail grades about how they are going to achieve passes

stevehook:
## Context

As a candidate who does not have a pass grade for English or maths (or science) GCSE I need to say how I plan to meet the requirement so that I can improve my chances of having a successful application.

- [x] If a candidate does not enter a pass grade, we show an additional page in the flow: ‘You need a [subject] GCSE grade 4 (C) or above, or equivalent’, which gives candidates an opportunity to say how the expect to meet this requirement.
image

- [x] Show the response to the above question on the review page if applicable
image

- [x] Only for Maths, English and Science
- [x] Fix tests
- [x] Handle multi-grade awards

## Guidance to review

- I've not added a feature flag for this change. Is one needed?
- Can you see any issues with reusing the existing `ApplicationQualification#missing_explanation` attribute to store the new value? (Up to now it's only been used in cases where there is no qualification or grade).
- I've tried to handle single/multiple grade qualifications by taking the `grade` or extracting the individual grades from `constituent_grades`. Does this make sense? A pass is defined as a qualification with at least one pass grade.

## Link to Trello card

https://trello.com/c/ONl9m9RD/3307-ask-additional-question-if-candidate-does-not-enter-a-gcse-pass-grade

## Things to check

### DIFF
--- a/app/components/candidate_interface/gcse_qualification_review_component.rb
+++ b/app/components/candidate_interface/gcse_qualification_review_component.rb
@@ -21,7 +21,7 @@ module CandidateInterface
           enic_reference_row,
           comparable_uk_qualification_row,
           grade_row,
-          grade_explanation_row,
+          failing_grade_explanation_row,
           award_year_row,
         ].compact
       end
@@ -76,7 +76,7 @@ module CandidateInterface
       }
     end
 
-    def grade_explanation_row
+    def failing_grade_explanation_row
       return nil unless application_qualification.failed_required_gcse? && application_qualification.missing_explanation.present?
 
       {
--- a/app/components/candidate_interface/gcse_qualification_review_component.rb
+++ b/app/components/candidate_interface/gcse_qualification_review_component.rb
@@ -21,6 +21,7 @@ module CandidateInterface
           enic_reference_row,
           comparable_uk_qualification_row,
           grade_row,
+          grade_explanation_row,
           award_year_row,
         ].compact
       end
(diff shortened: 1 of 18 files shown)


app/models/application_qualification.rb
```diff
@@ -144,6 +154,29 @@ def composite_equivalency_details
     details.strip if details.present?
   end
 
+  GCSE_PASS_GRADES = %w[A* A B C A*A* A*A AA AB BB BC CC CD 9 8 7 6 5 4 99 98 88 87 77 76 66 65 55 54 44 43].freeze
```
stevehook, line 157: Taken directly from https://trello.com/c/ONl9m9RD/3307-ask-additional-question-if-candidate-does-not-enter-a-gcse-pass-grade

app/components/candidate_interface/gcse_qualification_review_component.rb
```diff
@@ -75,6 +76,17 @@ def grade_row
       }
     end
 
+    def grade_explanation_row
```
davidgisbey, line 79: super nitty but maybe `failing_grade_explanation_row`

app/forms/candidate_interface/gcse_grade_explanation_form.rb
```diff
@@ -0,0 +1,19 @@
+module CandidateInterface
+  class GcseGradeExplanationForm
+    include ActiveModel::Model
+
+    attr_accessor :missing_explanation
```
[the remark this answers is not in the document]
davidgisbey: At the moment if you select you are missing the qual then add a gcse with a failing grade, it prepopulates the missing box on the failing grade explanation page
image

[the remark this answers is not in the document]
davidgisbey: I actually think this is fine as we specify C or above in the title, so storing their response on a shared attr seems totally reasonable.

app/models/application_qualification.rb
```diff
@@ -144,6 +154,29 @@ def composite_equivalency_details
...
+  def failed_required_gcse?
+    return true if required_gcse? &&
+      all_grades.present? &&
```
davidgisbey, line 160: Super nitty, but i don't think we need to check for presence here

davidgisbey (approved):
Looks really great. Gave it a thorough test and it's working nicely. One super super nitty comment.

app/models/application_qualification.rb
stevehook: I had a test failure without this, can't remember the test, but it came down to a qualification that didn't (yet) have a grade at all being flagged as failed. That doesn't make a lot of sense - you can't be flagged as failed if you've not entered a grade yet.

PR #4559, opened 2021-04-21
Surface Vendor API validation errors in the Support Interface

cwrw:
## Context

We would like to surface Vendor API validation errors to ensure we aren't introducing any breaking changes by rolling out the new permissions from the Make/Change offer flow.

We have an existing Validation errors summary page on the support interface, and we could reuse it to also show Vendor API errors.

- Change structure of support interface validation errors to list the different types of errors and link to them separately
- Add Vendor API validation errors to the support interface and the relevant methods to help show those errors similar to the validation error structure
- Tidy up candidate interface errors and add some missing specs.

## Guidance to review

It would be easier to commit by commit.
I attempted to extract partials for the different views, however the differences between both the Vendor API and Validation errors prompted me to revert that and add different erb files for each. Happy to hear thoughts around how we can DRY those up. I also am keen to leave that till adding Provider interface validation errors where we can revisit this as well.

## Link to Trello card

https://trello.com/b/5IiPW0Ok/provendor-team-board

## Things to check

- [ x This code does not rely on the addition/removal of Azure config environment variables in the same Pull Request


spec/models/vendor_api_request_spec.rb
```diff
@@ -0,0 +1,61 @@
+require 'rails_helper'
+
+RSpec.describe VendorAPIRequest, type: :model do
+  let(:response_body) do
+    {
+      'errors' => [
+        {
```
cwrw, line 7: add test for multiple errors

app/views/support_interface/validation_errors/_list.html.erb
```diff
@@ -0,0 +1,47 @@
+<table class="govuk-table">
```
cwrw, line 1: move partials back into respective folders as they turned out to be quite different

app/controllers/support_interface/validation_errors/vendor_api_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+          .page(params[:page] || 1)
+      end
+
+      def summary
+        sort_param = params.permit(:sortby)[:sortby]
+
+        @validation_error_summary = ::VendorAPIRequestSummaryQuery.new(sort_param).call
+      end
+
+      class ValidationErrorSearch
```
cwrw, line 23: move out and test

config/routes.rb
```diff
@@ -899,6 +899,10 @@
         get 'candidate' => 'candidate#index', as: :candidate
         get 'candidate/search' => 'candidate#search', as: :candidate_search
         get 'candidate/summary' => 'candidate#summary', as: :candidate_summary
+
```
cwrw, line 902: tidy up routes

app/queries/vendor_api_request_summary_query.rb
```diff
@@ -0,0 +1,78 @@
+class VendorAPIRequestSummaryQuery
```
cwrw, line 1: add specs

app/views/support_interface/validation_errors/vendor_api/summary.html.erb
```diff
@@ -0,0 +1,99 @@
+<% content_for :title, 'Validation error summary' %>
```
cwrw, line 1: add system spec

app/controllers/support_interface/validation_errors_controller.rb
```diff
@@ -1,33 +1,5 @@
...
-        .page(params[:page] || 1)
-    end
-
-    def summary
-      sort_param = params.permit(:sortby)[:sortby]
-
-      @validation_error_summary = ::ValidationErrorSummaryQuery.new(sort_param).call
-    end
-
-    class ValidationErrorSearch
```
cwrw, line 22: I moved this to the `ValidationError` class as felt suitable to live there as well as test

app/queries/vendor_api_request_summary_query.rb
```diff
@@ -0,0 +1,78 @@
...
+    ).to_a
+  end
+
+private
+
+  def count_sql
+    "WITH vendor_api_request_counts AS (
+      SELECT
+        request_path,
+        jsonb_extract_path_text(jsonb_array_elements(response_body->'errors'), 'error') AS attribute,
```
cwrw, line 24: this will extract text from the `error` field nested in an array of hashes under the `errors` section in the response body. Let me know if there is a better way to query this.

app/models/vendor_api_request.rb
```diff
@@ -1,3 +1,32 @@
 class VendorAPIRequest < ApplicationRecord
   belongs_to :provider, optional: true
+  scope :validation_errors, -> { where(status_code: 422) }
+
+  def self.list_of_distinct_errors_with_count
+    distinct_errors = validation_errors.flat_map do |request|
```
benswannack, line 6: Can this be called `error_messages` or similar? It looks like they aren't distinct until we tally them up

app/views/support_interface/validation_errors/vendor_api/summary.html.erb
```diff
@@ -0,0 +1,97 @@
...
+      <td class="govuk-table__cell govuk-table__cell--numeric">
+        <%= row['incidents_last_week'] %>
+      </td>
+      <td class="govuk-table__cell govuk-table__cell--numeric">
+        <%= row['unique_providers_last_week'] %>
+      </td>
+    </tr>
+  <% end %>
+  </tbody>
+</table>
```
despo, line 97: Can this be moved in a reusable component (so it can be shared between vendor api and candidate)? 
Consider passing the main body (lines 62 to 95 here) as I can see how it may be difficult to use a shared component for that.

app/views/support_interface/validation_errors/candidate/index.html.erb
```diff
@@ -0,0 +1,59 @@
...
+
+<table class="govuk-table">
+  <thead class="govuk-table__head">
+    <tr class="govuk-table__row">
+      <th class="govuk-table__header govuk-!-width-three-quarters">Attribute error</th>
+      <th class="govuk-table__header govuk-table__header--numeric govuk-!-width-one-quarter">Error count</th>
+    </tr>
+  </thead>
+  <tbody class="govuk-table__body">
+  <% @grouped_column_error_counts.each do |(form_object, attribute, message), count| %>
```
despo, line 21: Consider renaming this to error for clarify and consistency between this and the vendor api view.

app/views/support_interface/validation_errors/vendor_api/index.html.erb
```diff
@@ -0,0 +1,59 @@
...
+      <td class="govuk-table__cell">
+        <h2 class="govuk-heading-s govuk-!-margin-bottom-0">
+          <%= govuk_link_to request, support_interface_validation_errors_vendor_api_search_path(request_path: request) %>
+        </h2>
+      </td>
+      <td class="govuk-table__cell govuk-table__cell--numeric">
+        <%= count %>
+      </td>
+    </tr>
+  <% end %>
```
despo, line 57: Excluding this bit of code, the rest can potentially be extracted to an ErrorsListComponent that can be reshared between candidate and vendor API. A content block can be specified in the component (similar to [here](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/components/collection_select_component.html.erb#L15) to enable us to leave this code in the view.

What do you think? I'm not sure if i missed anything else that makes that more complicated.

app/views/support_interface/validation_errors/vendor_api/search.html.erb
```diff
@@ -0,0 +1,60 @@
...
+                <span class="govuk-!-font-weight-regular"><%= govuk_link_to vendor_api_request.request_path, support_interface_validation_errors_vendor_api_search_path(request_path: vendor_api_request.request_path) %>:</span>
+                <%= govuk_link_to validation_error['error'].humanize, support_interface_validation_errors_vendor_api_search_path(request_path: vendor_api_request.request_path, attribute: validation_error['error']) %>
+              </h3>
+              <p class="govuk-body govuk-!-margin-bottom-0"><%= validation_error['message'] %></p>
+            </td>
+            <td class="govuk-table__cell">
+              <% if vendor_api_request.request_body.blank? %>
+                <%= govuk_tag(text: 'Empty', colour: 'grey') %>
+              <% else %>
+                <code><%= JSON.pretty_generate(vendor_api_request.request_body) %></code>
```
benswannack, line 50: Will this show the correct request body and query parameters for POST requests? E.g. Do we show all the data if I POST to `/endpoint?my_param=ben` with a body of `{"do_something":"false"}`

app/views/support_interface/validation_errors/vendor_api/index.html.erb
```diff
@@ -0,0 +1,59 @@
...
+      <th class="govuk-table__header govuk-!-width-three-quarters">Attribute error</th>
+      <th class="govuk-table__header govuk-table__header--numeric govuk-!-width-one-quarter">Error count</th>
+    </tr>
+  </thead>
+  <tbody class="govuk-table__body">
+  <% @grouped_column_error_counts.each do |(request_path, error, message), count| %>
+    <tr class="govuk-table__row">
+      <td class="govuk-table__cell">
+        <h2 class="govuk-heading-s govuk-!-margin-bottom-0">
+          <span class="govuk-!-font-weight-regular"><%= govuk_link_to request_path, support_interface_validation_errors_vendor_api_search_path(request_path: request_path) %>:</span>
```
benswannack, line 25: (As discussed, probably one for a future iteration)

Grouping by exact request path might not be as useful as grouping by request path _up to ids_. It may be nicer to have them grouped by e.g. `/application/{id}/offer` (and capture the id in details somewhere), so we can see that the offer endpoint gets lots of validation errors, rather than by `/application/420/offer`, where we would have to do some maths to work out how many offers resulted in a validation error.

app/models/vendor_api_request.rb
```diff
@@ -1,3 +1,32 @@
 class VendorAPIRequest < ApplicationRecord
   belongs_to :provider, optional: true
+  scope :validation_errors, -> { where(status_code: 422) }
```
despo, line 3: This implies that we are only getting back the errors so consider renaming, maybe `validation_error_requests`?

benswannack (commented):
I agree with Despo about trying to wrap some of the view bits into components, and passing anything special via the `content`. Otherwise the general approach is nice!

app/models/vendor_api_request.rb
benswannack: Potentially `with_validation_errors` since it's a scope on `requests` already?

app/queries/vendor_api_request_summary_query.rb
despo: I can't think of a better way to achieve this with the current data structure.

However, as we already know that we are interested in the errors occuring for unprocessable requests maybe we should consider updating the VendorAPIRequest model to include an errors field, and the relevant worker to store the errors in that field as they occur, to make them easier to access for both development but potentially also by the analysts via blazer.

app/models/vendor_api_request.rb
despo: I did initially think of that, but I wonder if it may cause some confusion when used on line 6. 

E.g. 
```ruby
distinct_errors = with_validation_errors.flat_map do |request|
```

app/queries/vendor_api_request_summary_query.rb
cwrw: That is a good idea. I think I'd like to tackle this in a further iterations as adding a new column now would mean we would need to backfill those values for existing fields, something that is beyond the scope of this PR.

app/models/vendor_api_request.rb
cwrw: I went through a few iterations of this but landed on `unprocessable_entities` as felt like the most suitable and maps to the response we return to users

app/views/support_interface/validation_errors/vendor_api/search.html.erb
cwrw: I double checked and it does not, I'll add a query string and request body to the search view 👍

app/views/support_interface/validation_errors/vendor_api/summary.html.erb
cwrw: done in 1dda5794495a138725549e3e11ff2fbcf7951e14

app/views/support_interface/validation_errors/vendor_api/index.html.erb
cwrw: done in 1dda5794495a138725549e3e11ff2fbcf7951e14

benswannack (approved):
Looks good, aside from a few minor points.

Tested in the review app, both vendor api and candidate validation errors are displayed correctly :)

app/controllers/support_interface/validation_errors/vendor_api_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+          .search_validation_errors(params)
+          .includes('provider')
+          .order('created_at DESC')
+          .page(params[:page] || 1)
+      end
+
+      def summary
+        sort_param = params.permit(:sortby)[:sortby]
+
+        @validation_error_summary = ::VendorAPIRequestSummaryQuery.new(sort_param).call
```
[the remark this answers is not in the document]
cwrw: We don't seem to have any namespaces in the query directory so keen on not adding any

config/routes.rb
```diff
@@ -887,16 +887,28 @@
       get '/course-options', to: 'performance#course_options', as: :course_options
       get '/unavailable-choices' => 'performance#unavailable_choices', as: :unavailable_choices
 
-      get '/validation-errors' => 'validation_errors#index', as: :validation_errors
-      get '/validation-errors/search' => 'validation_errors#search', as: :validation_error_search
-      get '/validation-errors/summary' => 'validation_errors#summary', as: :validation_error_summary
-
```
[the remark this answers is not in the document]
cwrw: I've double checked with the Candidate team, as well as sent a message on the channel previously to ensure that everyone is informed of this change

PR #4562, opened 2021-04-21
Create data export which shows who has initiated each export

davidgisbey:
## Context

We want to be able to identify who the users of our exports are so that we can assign the primary user ownership of the export.

This export provides the export_type, created_at and email address of the initiator.

- Add export 
- Add yml file for docs 

## Guidance to review

Can anyone think of a better name for this export?

## Link to Trello card

https://trello.com/c/loC6hQ4V/3320-dev-build-export-which-captures-each-exports-users

## Things to check

### DIFF
--- a/app/services/support_interface/who_ran_which_export_export.rb
+++ b/app/services/support_interface/who_ran_which_export_export.rb
@@ -1,0 +1,18 @@
+module SupportInterface
+  class WhoRanWhichExportExport
+    def data_for_export
+      DataExport
+      .where
+      .not(export_type: nil)
+      .order(%i[export_type created_at])
+      .find_each(batch_size: 100)
+      .map do |export|
+        {
+          export_type: export.export_type,
+          created_at: export.created_at,
+          initiated_by: export.initiator.email_address,
+        }
+      end
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


davidgisbey:
☝️ this can be ignored since it's a new export

app/services/support_interface/export_initiators_export.rb
```diff
@@ -0,0 +1,18 @@
+module SupportInterface
+  class ExportInitiatorsExport
```
davidgisbey, line 2: This name makes me sad. Anyone got a better suggestion?

duncanjbrown: How about... `WhoRanWhichExportExport`?!

app/services/support_interface/who_ran_which_export_export.rb
```diff
@@ -0,0 +1,18 @@
+module SupportInterface
+  class WhoRanWhichExportExport
+    def data_for_export
+      DataExport
+      .where
+      .not(export_type: nil)
```
duncanjbrown, line 6: Do we still have any?

duncanjbrown (approved):
Looks good to me!

app/services/support_interface/who_ran_which_export_export.rb
davidgisbey: Yes, for ones that have been deprecated.

PR #4563, opened 2021-04-21
Don't use blue-green deployment strategy for clock/web containers 

vigneshmsft:
## Context

When a deployment of a new version is underway, the jobs on the worker app are still running.
The worker container might still run a job on the old schema while the version is being updated and causing the job to error.

Use standard deployment strategy, so the worker container is updated and restarted.

## Things to check

### DIFF
--- a/terraform/modules/paas/main.tf
+++ b/terraform/modules/paas/main.tf
@@ -48,7 +48,6 @@ resource "cloudfoundry_app" "clock" {
   instances          = var.clock_app_instances
   memory             = var.clock_app_memory
   space              = data.cloudfoundry_space.space.id
-  strategy           = "blue-green-v2"
   timeout            = 180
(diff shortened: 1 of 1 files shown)


benswannack:
Interesting! Does the standard deployment stop all old instances first, and then start up the new ones once they've stopped?
Is there a chance we get errors because the containers aren't available?

vigneshmsft:
> Interesting! Does the standard deployment stop all old instances first, and then start up the new ones once they've stopped?
> Is there a chance we get errors because the containers aren't available?

@benswannack I the standard strategy just restarts the app container with the new docker image.
we do this only for the worker containers to check if it prevents errors during deployment, because blue-green might run an instance of the app which might run against an out of date db schema

saliceti (approved):

PR #4564, opened 2021-04-21
Include accredited providers as part of check for candidate masquerading

steventux:
## Context

If a provider user belongs to an org which ratifies a course a candidate has applied for, they should be able to masquerade as the candidate assuming they have the correct user and org permissions.

Widen the check to verify that the provider user may also belong to an accrediting org with the correct user and org permissions.

Paired with @benswannack on this

## Guidance to review

To test you need an application where the course is ratified by your provider user's org and the training provider for the course is not one of your orgs.
You also need to make sure the ratifying provider org has permissions to make decisions on the relationship between the course provider and the ratifying (accredited) provider.

Is the permissions check for make decisions necessary? We've erred on the side of being more restrictive but maybe this isn't necessary.

## Link to Trello card

https://trello.com/c/hYXKo0OV/3528-provider-is-unable-to-masquerade-as-candidate-in-sandbox

## Things to check

### DIFF
--- a/app/controllers/provider_interface/candidates_controller.rb
+++ b/app/controllers/provider_interface/candidates_controller.rb
@@ -21,11 +21,9 @@ module ProviderInterface
   private
 
     def verify_provider_association(candidate:, providers:)
-      provider_ids_from_candidate = candidate.application_forms
-                                              .map(&:application_choices).flatten
-                                              .map(&:provider).map(&:id).uniq
+      application_choices = candidate.application_forms.map(&:application_choices).flatten
 
-      providers.any? { |provider| provider_ids_from_candidate.include? provider.id }
+      (application_choices.map(&:associated_providers).flatten & providers).any?
     end
 
     def disable_on_production
--- a/spec/system/provider_interface/provider_acts_as_candidate_on_sandbox_spec.rb
+++ b/spec/system/provider_interface/provider_acts_as_candidate_on_sandbox_spec.rb
@@ -16,6 +16,14 @@ RSpec.describe 'A Provider can sign in as a candidate' do
     then_i_am_redirected_to_the_candidate_interface
     and_i_see_a_flash_message
     and_i_am_logged_in_as_that_candidate
+
+    when_my_organisation_ratifies_a_course_for_an_application
+    and_i_visit_the_ratified_application_in_the_provider_interface
+    and_i_click_on_the_sign_in_button
+
+    then_i_am_redirected_to_the_candidate_interface
+    and_i_see_a_flash_message
+    and_i_am_logged_in_as_the_ratified_application_candidate
   end
 
   def given_i_am_a_provider_user_with_dfe_sign_in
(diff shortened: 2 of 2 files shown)


app/controllers/provider_interface/candidates_controller.rb
```diff
@@ -21,11 +21,13 @@ def impersonate
...
-      provider_ids_from_candidate = candidate.application_forms
-                                              .map(&:application_choices).flatten
-                                              .map(&:provider).map(&:id).uniq
+      application_choices = candidate.application_forms.map(&:application_choices).flatten
+      candidate_providers = application_choices.map(&:provider) + application_choices.map(&:accredited_provider)
+      provider_ids_from_candidate = candidate_providers.compact.map(&:id).uniq
+      auth = ProviderAuthorisation.new(actor: current_provider_user)
 
-      providers.any? { |provider| provider_ids_from_candidate.include? provider.id }
+      providers.any? { |provider| provider_ids_from_candidate.include? provider.id } &&
```
benswannack, line 29: Just thought: if we're going with this permissions requirement, does the auth do the provider membership check for us?

steventux: Good point, it does call `GetApplicationChoicesForProvider` which seems to be doing the same thing.

cwrw (commented):
> Is the permissions check for make decisions necessary? We've erred on the side of being more restrictive but maybe this isn't necessary.

I wonder if it's safer to maintain the current behaviour regarding the provider on the course, and extend that to the ratified course instead of enforcing the make decisions. My worry is users will lose access since they don't have make decisions set up. What are the risks of this if we allow the ratifying course masquerading without the make decisions permission?

steventux:
@cwrw I think you're right, given the nature of how we seed courses in Sandbox the training and ratifying providers of a course, the organisations are effectively siloed anyway, so I don't think there's a danger of this allowing cross-vendor changes. I'll remove the permissions check.

steventux:
@cwrw amended here https://github.com/DFE-Digital/apply-for-teacher-training/pull/4564/commits/ba75af59d36e56c3d16a448fefcd0a17349fe580 with no permissions check

app/controllers/provider_interface/candidates_controller.rb
```diff
@@ -21,11 +21,10 @@ def impersonate
   private
 
     def verify_provider_association(candidate:, providers:)
-      provider_ids_from_candidate = candidate.application_forms
-                                              .map(&:application_choices).flatten
-                                              .map(&:provider).map(&:id).uniq
+      application_choices = candidate.application_forms.map(&:application_choices).flatten
+      providers_from_applications = application_choices.map(&:provider) + application_choices.map(&:accredited_provider)
```
cwrw, line 25: Could we utilise [associated_providers](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/models/application_choice.rb#L155)  which extracts both provider and accredited provider? 
```
providers_from_applications  = application_choices.flat_map(&:associated_providers).uniq
```

cwrw (approved):
Looks good, just one minor comment

PR #4568, opened 2021-04-22
Change interview date hint text and error message to include RBD

benswannack:
## Context
It previously wasn't clear that the RBD date of the application was part of the form validation

Add text about restrictions on interview date to do with RBD date, and update error message to match

## Guidance to review
Here's what it looks like now:
Screenshot 2021-04-22 at 09 07 46

## Link to Trello card
https://trello.com/c/w641x9Gp/3615-help-users-set-up-an-interview-before-the-automatic-rejection-date

## Things to check

### DIFF
--- a/app/components/provider_interface/interview_form_component.html.erb
+++ b/app/components/provider_interface/interview_form_component.html.erb
@@ -16,7 +16,16 @@
     </div>
   <% end %>
 
-  <%= f.govuk_date_field :date, hint: { text: t('helpers.hint.provider_interface_interview_wizard.date', example_date: example_date) } %>
+  <%= f.govuk_date_field(
+        :date,
+        hint: {
+          text: t(
+            'helpers.hint.provider_interface_interview_wizard.date',
+            rbd_date: application_choice.reject_by_default_at.to_s(:govuk_date),
+            example_date: example_date,
+          ),
+        },
+      ) %>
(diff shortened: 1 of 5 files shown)


app/components/provider_interface/interview_form_component.html.erb
```diff
@@ -16,7 +16,16 @@
     </div>
   <% end %>
 
-  <%= f.govuk_date_field :date, hint: { text: t('helpers.hint.provider_interface_interview_wizard.date', example_date: example_date) } %>
+  <%= f.govuk_date_field(
+        :date,
+        hint: {
+          text: t(
+            'helpers.hint.provider_interface_interview_wizard.date',
+            rbd_date: application_choice.reject_by_default_at.to_s(:govuk_date),
```
steventux, line 24: Is it ever possible for `reject_by_default_at` to be `nil` here?

steventux (approved):
One question about dates being `nil` and calling `to_s(:govuk_date)`.

app/components/provider_interface/interview_form_component.html.erb
benswannack: Theoretically it can be nil. In practice, applications need to have been submitted before interviews can be arranged, so the rbd date should have been set

PR #4574, opened 2021-04-22
Ensure R4R controller redirects unless the application choice can be rejected

steventux:
## Context

It's possible to access the reasons for rejection wizard flow by URL for an application which ultimately cannot transition to a rejected state.

Adds a before_action filter to prevent access to any R4R controller actions unless the application is rejectable.
 

## Link to Trello card

https://trello.com/c/yqm6ZWTd/3578-only-applications-in-the-awaiting-decision-or-interviewing-state-can-be-rejected

## Things to check

### DIFF
--- a/app/controllers/provider_interface/reasons_for_rejection_controller.rb
+++ b/app/controllers/provider_interface/reasons_for_rejection_controller.rb
@@ -2,6 +2,7 @@ module ProviderInterface
   class ReasonsForRejectionController < ProviderInterfaceController
     before_action :set_application_choice
     before_action :redirect_if_application_rejected_and_feedback_provided
+    before_action :check_application_is_rejectable
 
     def edit_initial_questions
       @wizard = ReasonsForRejectionWizard.new(store, current_step: 'initial_questions')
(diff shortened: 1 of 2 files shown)


app/controllers/provider_interface/reasons_for_rejection_controller.rb
```diff
@@ -134,5 +135,13 @@ def offer_store
       key = "offer_wizard_store_#{current_provider_user.id}_#{@application_choice.id}"
       WizardStateStores::RedisStore.new(key: key)
     end
+
+    def check_application_is_rejectable
+      return if ApplicationStateChange.new(@application_choice).can_reject?
+
+      flash[:warning] = 'This application cannot be rejected'
```
despo, line 142: I don't think we want to render a message (based on [Offers](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/controllers/provider_interface/offers_controller.rb#L76) and [Interview](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/controllers/provider_interface/interviews_controller.rb#L160)) but maybe verify with design (cc @adamsilver @johndoates).

despo (approved):
Happy to approve but as per comment please verify with design regarding displaying an error message.

app/controllers/provider_interface/reasons_for_rejection_controller.rb
adamsilver: I would suggest using the standard 404.

steventux: Thanks for raising this @despo , having spoken with @adamsilver we think a 404 is appropriate. Amended here https://github.com/DFE-Digital/apply-for-teacher-training/pull/4574/commits/f9b5acd1a67923e01c013bfce72d5e6b4c7f925c

PR #4575, opened 2021-04-22
Use nationality to determine international candidates in other qualifications

stevehook:
## Context

Currently the _A levels and other qualifications_ section uses `international?` (based on `address_type` enum) to determine if the candidate is an international student and if they are it will then render the section titles, and other pieces of copy, as 'Other qualifications' rather an 'A levels and other qualifications'. We want to use nationality instead so it's consistent with the EFL feature.

- [x] Add `_address` suffix to the `ApplicationForm#address_type` enum. So helpers like `ApplicationForm#uk?` now become `ApplicationForm#uk_address` etc.
- [x] Add an `ApplicationForm#international_applicant?` method that uses the same logic as the existing `efl_section_required?` method. Use this in the other qualifications front end to switch the content between UK (including term _A level_) and more generic international variants.
- [x] Drop the `efl_section_required?` method and just use `international_applicant?`
- [x] Rename `ApplicationForm#english_speaking_nationality?` to simply `british_or_irish?`
- [x] Update system specs

## Guidance to review

- Per commit is probably easier.
- The card says that we should change the enum values - I'm a bit reluctant to do that unless there is a strong need due to the effort needed to migrate the existing values in the DB. Does this make sense?

## Link to Trello card

https://trello.com/c/5AxBkwIz/3319-update-the-criteria-for-classifying-international-candidates-on-the-a-levels-and-other-qualifications-section

## Things to check

### DIFF
--- a/app/models/application_form.rb
+++ b/app/models/application_form.rb
@@ -28,7 +28,7 @@ class ApplicationForm < ApplicationRecord
   MINIMUM_COMPLETE_REFERENCES = 2
   MAXIMUM_REFERENCES = 10
   EQUALITY_AND_DIVERSITY_MINIMAL_ATTR = %w[sex disabilities ethnic_group].freeze
-  ENGLISH_SPEAKING_NATIONALITIES = %w[GB IE].freeze
+  BRITISH_OR_IRISH_NATIONALITIES = %w[GB IE].freeze
   MAXIMUM_PHASE_ONE_COURSE_CHOICES = 3
   MAXIMUM_PHASE_TWO_COURSE_CHOICES = 1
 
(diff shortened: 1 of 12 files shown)


app/models/application_form.rb
```diff
@@ -58,7 +58,7 @@ def equality_and_diversity_answers_provided?
   enum address_type: {
     uk: 'uk',
     international: 'international',
-  }
+  }, _suffix: :address
```
malcolmbaig, line 61: 👌🏼 nice, much better idea!

malcolmbaig (approved):

davidgisbey:
Beautiful stuff 👌

davidgisbey (approved):

PR #4576, opened 2021-04-22
Use radio buttons to complete individual sections instead of checkboxes

TobyRet:
## Context

```
As a... candidate completing my application
I need to... mark sections as complete
So that... I know which parts of my application still need to be completed
```

- The current ‘I have completed this section’ checkboxes on section review pages have been updated to use radio buttons. 
- Additionally, some sections show an additional contextual hint text, which encourage candidates to review certain aspects before marking a section as complete.

## Guidance to review

- There is little complexity here, just a lot of file changes here (sorry) . The `CompleteSection` component was updated to match the design. This uses a standard field (`:completed`) and form object (for validation). As a result all the controllers required updating along with specs. I've tried to break up into small / reviewable commits. 

complete-radio-buttons

## Link to Trello card

https://trello.com/c/MgPkClPi/3303-use-radio-buttons-to-complete-individual-sections-instead-of-checkboxes

## Things to check

### DIFF
--- a/app/components/candidate_interface/complete_section_component.html.erb
+++ b/app/components/candidate_interface/complete_section_component.html.erb
@@ -1,9 +1,7 @@
 <%= form_with model: section_complete_form, url: path, method: request_method do |f| %>
   <%= f.govuk_error_summary %>
 
-  <% if summary_component %>
-    <%= render(summary_component) %>
-  <% end %>
+  <%= render(review_component) %>
 
   <%= f.govuk_radio_buttons_fieldset :completed, classes: 'govuk-!-width-two-thirds ', legend: { text: t('application_form.completed_question'), size: 'm' } do %>
     <% if hint_text %>
--- a/app/components/candidate_interface/complete_section_component.html.erb
+++ b/app/components/candidate_interface/complete_section_component.html.erb
@@ -7,7 +7,7 @@
     <% if hint_text %>
       <div class="govuk-hint"><%= hint_text %></div>
     <% end %>
-    <%= f.govuk_radio_button :completed, true, label: { text: radio_button_label }, link_errors: true %>
+    <%= f.govuk_radio_button :completed, true, label: { text: complete_or_reviewed_radio_button_label }, link_errors: true %>
     <%= f.govuk_radio_button :completed, false, label: { text: t('application_form.incomplete_radio') } %>
   <% end %>
 
--- a/app/components/candidate_interface/complete_section_component.html.erb
+++ b/app/components/candidate_interface/complete_section_component.html.erb
@@ -1,8 +1,12 @@
-<%= form_with model: @application_form, url: @path, method: @request_method do |f| %>
-  <div class='govuk-form-group'>
-    <%= f.hidden_field @field_name, value: false %>
-    <%= f.govuk_check_box @field_name, true, multiple: false, label: { text: t(checkbox_label) } %>
-  </div>
+<%= form_with model: section_complete_form, url: path, method: request_method do |f| %>
+  <%= f.govuk_error_summary %>
+
+  <%= render(summary_component) %>
+
+  <%= f.govuk_radio_buttons_fieldset :completed, legend: { text: t('application_form.completed_question'), size: 'm' } do %>
+    <%= f.govuk_radio_button :completed, true, label: { text: radio_button_label }, link_errors: true%>
+    <%= f.govuk_radio_button :completed, false, label: { text: t('application_form.incomplete_radio') } %>
+  <% end %>
 
   <%= f.govuk_submit t('continue') %>
 <% end %>
--- a/app/components/candidate_interface/complete_section_component.html.erb
+++ b/app/components/candidate_interface/complete_section_component.html.erb
@@ -1,7 +1,9 @@
 <%= form_with model: section_complete_form, url: path, method: request_method do |f| %>
   <%= f.govuk_error_summary %>
 
-  <%= render(summary_component) %>
+  <% if summary_component %>
+    <%= render(summary_component) %>
+  <% end %>
 
   <%= f.govuk_radio_buttons_fieldset :completed, classes: 'govuk-!-width-two-thirds ', legend: { text: t('application_form.completed_question'), size: 'm' } do %>
     <% if complete_hint_text %>
--- a/app/components/candidate_interface/complete_section_component.html.erb
+++ b/app/components/candidate_interface/complete_section_component.html.erb
@@ -6,8 +6,8 @@
   <% end %>
 
   <%= f.govuk_radio_buttons_fieldset :completed, classes: 'govuk-!-width-two-thirds ', legend: { text: t('application_form.completed_question'), size: 'm' } do %>
-    <% if complete_hint_text %>
-      <div class="govuk-hint"><%= complete_hint_text %></div>
+    <% if hint_text %>
+      <div class="govuk-hint"><%= hint_text %></div>
     <% end %>
     <%= f.govuk_radio_button :completed, true, label: { text: radio_button_label }, link_errors: true %>
     <%= f.govuk_radio_button :completed, false, label: { text: t('application_form.incomplete_radio') } %>
--- a/app/components/candidate_interface/complete_section_component.html.erb
+++ b/app/components/candidate_interface/complete_section_component.html.erb
@@ -3,8 +3,11 @@
 
   <%= render(summary_component) %>
 
-  <%= f.govuk_radio_buttons_fieldset :completed, legend: { text: t('application_form.completed_question'), size: 'm' } do %>
-    <%= f.govuk_radio_button :completed, true, label: { text: radio_button_label }, link_errors: true%>
+  <%= f.govuk_radio_buttons_fieldset :completed, classes: 'govuk-!-width-two-thirds ', legend: { text: t('application_form.completed_question'), size: 'm' } do %>
+    <% if complete_hint_text %>
+      <div class="govuk-hint"><%= complete_hint_text %></div>
+    <% end %>
+    <%= f.govuk_radio_button :completed, true, label: { text: radio_button_label }, link_errors: true %>
     <%= f.govuk_radio_button :completed, false, label: { text: t('application_form.incomplete_radio') } %>
   <% end %>
 
--- a/app/components/candidate_interface/complete_section_component.html.erb
+++ b/app/components/candidate_interface/complete_section_component.html.erb
@@ -5,7 +5,13 @@
 
   <%= f.govuk_radio_buttons_fieldset :completed, classes: 'govuk-!-width-two-thirds ', legend: { text: t('application_form.completed_question'), size: 'm' } do %>
     <% if hint_text %>
-      <div class="govuk-hint"><%= hint_text %></div>
+      <% if hint_text.is_a?(Array) %>
+        <% hint_text.each do |paragraph| %>
+          <div class="govuk-hint"><%= paragraph %></div>
+        <% end %>
+      <% else %>
+        <div class="govuk-hint"><%= hint_text %></div>
+      <% end %>
     <% end %>
     <%= f.govuk_radio_button :completed, true, label: { text: complete_or_reviewed_radio_button_label }, link_errors: true %>
     <%= f.govuk_radio_button :completed, false, label: { text: t('application_form.incomplete_radio') } %>
--- a/app/components/candidate_interface/complete_section_component.rb
+++ b/app/components/candidate_interface/complete_section_component.rb
@@ -2,7 +2,7 @@ module CandidateInterface
   class CompleteSectionComponent < ViewComponent::Base
     attr_reader :section_complete_form, :path, :request_method, :summary_component, :complete_hint_text, :section_review
 
-    def initialize(section_complete_form:, path:, request_method:, summary_component:, section_review: false, complete_hint_text: false)
+    def initialize(section_complete_form:, path:, request_method:, summary_component: nil, section_review: false, complete_hint_text: false)
       @section_complete_form = section_complete_form
       @path = path
       @request_method = request_method
@@ -13,9 +13,9 @@ module CandidateInterface
 
     def radio_button_label
       if section_review
-        'application_form.reviewed_radio'
+        t('application_form.reviewed_radio')
       else
-        'application_form.completed_radio'
+        t('application_form.completed_radio')
       end
     end
   end
--- a/app/components/candidate_interface/complete_section_component.rb
+++ b/app/components/candidate_interface/complete_section_component.rb
@@ -11,7 +11,7 @@ module CandidateInterface
       @hint_text = hint_text
     end
 
-    def radio_button_label
+    def complete_or_reviewed_radio_button_label
       if section_review
         t('application_form.reviewed_radio')
       else
--- a/app/components/candidate_interface/complete_section_component.rb
+++ b/app/components/candidate_interface/complete_section_component.rb
@@ -1,18 +1,20 @@
 module CandidateInterface
   class CompleteSectionComponent < ViewComponent::Base
-    def initialize(application_form:, path:, request_method:, field_name:, section_review: false)
-      @application_form = application_form
+    attr_reader :section_complete_form, :path, :request_method, :summary_component, :section_review
+
+    def initialize(section_complete_form:, path:, request_method:, summary_component:, section_review: false)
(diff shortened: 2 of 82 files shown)


config/locales/candidate_interface/personal_statement.yml
```diff
@@ -4,10 +4,16 @@ en:
       becoming_a_teacher:
         change_action: why do you want to be a teacher?
         label: Why do you want to be a teacher?
+        complete_hint_text_html: |
+          Proofread this section carefully for spelling and grammar. Ask someone you trust for their opinion.
+          <p>Ensure this is all your own work as plagiarism will be penalised.</p>
```
TobyRet, line 9: Is this an acceptable way to deal with paragraphs?

app/views/candidate_interface/application_choices/review.html.erb
```diff
@@ -20,31 +20,20 @@
...
+        section_complete_form: @section_complete_form,
+        path: candidate_interface_course_choices_complete_path,
+        request_method: :patch,
+        summary_component: CandidateInterface::CourseChoicesReviewComponent.new(application_form: @application_form),
+        complete_hint_text: if @application_choices.count == 1
+                              t('application_form.courses.complete.hint_text_two_courses')
+                            elsif @application_choices.count == 2
+                              t('application_form.courses.complete.hint_text_one_course')
+                            end,
+      )) %>
```
TobyRet, line 36: This is the only case where the hint text depends on some sort of condition. I don't think this logic should sit inside the component as it is very specific, however it doesn't feel right leaving it here either. Would it be best placed in a controller instance var or helper?

TobyRet:
@paulrobertlloyd ✋ 

Just spotted this merged PR (https://github.com/DFE-Digital/apply-for-teacher-training/pull/4551) re: reviewing sections as opposed to completing sections. Can you provide a steer on this in relation to this ticket pls?

CC: @malcolmbaig

app/components/candidate_interface/complete_section_component.html.erb
```diff
@@ -1,8 +1,17 @@
...
-  <div class='govuk-form-group'>
-    <%= f.hidden_field @field_name, value: false %>
-    <%= f.govuk_check_box @field_name, true, multiple: false, label: { text: t(checkbox_label) } %>
-  </div>
+<%= form_with model: section_complete_form, url: path, method: request_method do |f| %>
+  <%= f.govuk_error_summary %>
+
+  <% if summary_component %>
+    <%= render(summary_component) %>
+  <% end %>
```
malcolmbaig, line 6: If the summary_component is optional, would it make sense to not make it a dependency of the CompleteSectionComponent (and instead render them both separately in the view template as we do currently)?

Related question: where are we rendering this without providing a summary_component?

app/components/candidate_interface/complete_section_component.rb
```diff
@@ -1,18 +1,21 @@
...
+    attr_reader :section_complete_form, :path, :request_method, :summary_component, :complete_hint_text, :section_review
+
+    def initialize(section_complete_form:, path:, request_method:, summary_component: nil, section_review: false, complete_hint_text: false)
+      @section_complete_form = section_complete_form
       @path = path
       @request_method = request_method
-      @field_name = field_name
+      @summary_component = summary_component
       @section_review = section_review
+      @complete_hint_text = complete_hint_text
```
malcolmbaig, line 11: Naming suggestion: just `hint_text`.

app/components/candidate_interface/complete_section_component.rb
```diff
@@ -1,18 +1,21 @@
...
-    def initialize(application_form:, path:, request_method:, field_name:, section_review: false)
-      @application_form = application_form
+    attr_reader :section_complete_form, :path, :request_method, :summary_component, :complete_hint_text, :section_review
+
+    def initialize(section_complete_form:, path:, request_method:, summary_component: nil, section_review: false, complete_hint_text: false)
+      @section_complete_form = section_complete_form
       @path = path
       @request_method = request_method
-      @field_name = field_name
+      @summary_component = summary_component
```
malcolmbaig, line 9: I've questioned above whether or not this is necessary, but I like the choice of naming here. Leaving this comment as a reminder—if we do decided to keep this dependency, we should either:

- Card up renaming the various `_ReviewComponent` objects that we pass in, making them `_SummaryComponent` classes instead.
- Change this argument to `review_component` for now to keep things consistent.

app/components/candidate_interface/complete_section_component.html.erb
TobyRet: Hey @malcolmbaig - the summary_component should not be optional - I'll fix

app/components/candidate_interface/complete_section_component.rb
TobyRet: 👍

app/components/candidate_interface/complete_section_component.rb
TobyRet: mmm ... good spot.

Yep we are only passing in a SummaryCardComponent once so I think it makes sense to rename the arg to `review_component`

config/locales/candidate_interface/personal_statement.yml
```diff
@@ -4,10 +4,16 @@ en:
       becoming_a_teacher:
         change_action: why do you want to be a teacher?
         label: Why do you want to be a teacher?
+        complete_hint_text_html: |
+          Proofread this section carefully for spelling and grammar. Ask someone you trust for their opinion.
```
paulrobertlloyd, line 8:
```suggestion
          <p>Proofread this section carefully for spelling and grammar. Ask someone you trust for their opinion.</p>
```

config/locales/candidate_interface/personal_statement.yml
paulrobertlloyd: Feels a bit icky.

config/locales/candidate_interface/personal_statement.yml
```diff
@@ -4,10 +4,16 @@ en:
...
+          <p>Ensure this is all your own work as plagiarism will be penalised.</p>
       subject_knowledge:
         key: Your knowledge about the subject you want to teach
         change_action: evidence of subject knowledge
         label: Tell us what you know about the subject you want to teach
+        complete_hint_text_html: |
+          Proofread this section carefully for spelling and grammar. Ask someone you trust for their opinion.
```
paulrobertlloyd, line 15:
```suggestion
          <p>Proofread this section carefully for spelling and grammar. Ask someone you trust for their opinion.</p>
```

paulrobertlloyd (changes requested):
This following inset text is no longer needed here, should just show the button.

Screenshot 2021-04-23 at 14 11 53

Also, the summary cards on this page should appear full width:

Screenshot 2021-04-23 at 14 14 36

config/locales/candidate_interface/personal_statement.yml
frankieroberto: Should we also add the `govuk-body` class to this paragraph?

app/views/candidate_interface/application_choices/review.html.erb
frankieroberto: @TobyRet I think you might be able to tidy this up by using the [pluralization feature](https://guides.rubyonrails.org/i18n.html#pluralization) from Rails’s I18n gem?

Something like:

```ruby
<%= t('application_form.courses.complete.hint_text', count: @application_choices.count)
```

And then in the locale file:

```yml
hint_text:
  one: You can add one more course
  other: You can add %{count} more courses
```

app/components/candidate_interface/complete_section_component.html.erb
```diff
@@ -1,8 +1,15 @@
...
+
+  <%= render(review_component) %>
+
+  <%= f.govuk_radio_buttons_fieldset :completed, classes: 'govuk-!-width-two-thirds ', legend: { text: t('application_form.completed_question'), size: 'm' } do %>
+    <% if hint_text %>
+      <div class="govuk-hint"><%= hint_text %></div>
+    <% end %>
+    <%= f.govuk_radio_button :completed, true, label: { text: radio_button_label }, link_errors: true %>
+    <%= f.govuk_radio_button :completed, false, label: { text: t('application_form.incomplete_radio') } %>
+  <% end %>
```
frankieroberto, line 12: Should we use the [`govuk_collection_radio_buttons`](https://govuk-form-builder.netlify.app/form-elements/radios/) method from the formbuilder gem here?

Would save having to hardcode the HTML for the hint text.

app/components/candidate_interface/complete_section_component.rb
```diff
@@ -1,18 +1,21 @@
...
-      @field_name = field_name
+      @review_component = review_component
       @section_review = section_review
+      @hint_text = hint_text
     end
 
-    def checkbox_label
-      if @section_review
-        'application_form.reviewed_checkbox'
+    def radio_button_label
```
frankieroberto, line 14: nitpick: would this be more descriptive as something like `completed_or_reviewed_radio_button_label` as there are two radio button labels? Bit longwinded perhaps, but might save ambiguity later?

app/views/candidate_interface/application_choices/review.html.erb
paulrobertlloyd: Ooooh, 👌

config/locales/candidate_interface/personal_statement.yml
paulrobertlloyd: No, because this text appears within <div class="govuk-hint">…</div>. Not sure I like that much, either, but not sure how we could improve?

config/locales/candidate_interface/section_complete.yml
```diff
@@ -0,0 +1,8 @@
+en:
+  activemodel:
+    errors:
+      models:
+        candidate_interface/section_complete_form:
+          attributes:
+            completed:
+              blank: Select whether you have completed this section
```
frankieroberto, line 8: I think we also need a version of this validation error for the "reviewed" state?

app/forms/candidate_interface/section_complete_form.rb
```diff
@@ -0,0 +1,14 @@
+module CandidateInterface
+  class SectionCompleteForm
+    include ActiveModel::Model
+
+    attr_accessor :completed
+    validates :completed, presence: true
```
frankieroberto, line 6: As this is essentially a boolean, should we use the `inclusion` validator rather than presence (as false is a valid option)?

```suggestion
    validates :completed, inclusion: [true, false]
```

frankieroberto (changes requested):
I couldn't see an easy way to test the validation in the review app, as all of the existing applications already have the completed radio pre-set. When I removed the `checked=checked` html from the selected radio (using web inspector), I get a 404 page... 🤔

config/locales/candidate_interface/personal_statement.yml
TobyRet: Could potentially pass in an array of hints (paragraphs) into the component and loop through, if that's preferred

app/components/candidate_interface/complete_section_component.rb
TobyRet: 👍

config/locales/candidate_interface/personal_statement.yml
paulrobertlloyd: That would work! That way you could have the following:

```
<p class="govuk-hint">One line hint</p>
```

or 

```
<p class="govuk-hint">Two line hint, first line</p>
<p class="govuk-hint">Two line hint, second line</p>
```

app/views/candidate_interface/application_choices/review.html.erb
TobyRet: Nice 👍

app/components/candidate_interface/complete_section_component.html.erb
TobyRet: Some additional logic has been added to handle multiple hints so I don't think we use `govuk_collection_radio_buttons` in this instance ...

https://github.com/DFE-Digital/apply-for-teacher-training/pull/4576/commits/71603914c7278042dcfd53d0b0aae883659bd9bc#diff-50d730970ac17db67fd2b404df73aa94495376e376aff3e845abd36972a6cec5R8-R14

app/forms/candidate_interface/section_complete_form.rb
TobyRet: 👍

app/components/candidate_interface/complete_section_component.html.erb
frankieroberto: Boo.. 😞  Could this be avoided by passing in the `hint` as pre-formatted HTML instead, I wonder?

config/locales/candidate_interface/section_complete.yml
TobyRet: @frankieroberto discussed this and agreed it is not necessary as a candidate will have previously 'completed' a given section  (i.e. indicate whether they have completed it or not). We also have logic to indicate whether a section review is required (this sets the value of `completed` to `false`) - https://github.com/DFE-Digital/apply-for-teacher-training/pull/4551/files#diff-22fdd3b57e65e63df6dd2ff571a3632cbb3a7a45f1ec11fbb3b338753fdb6c4bR361

TobyRet:
Closing and re-opening PR to spin up new review app

app/components/candidate_interface/complete_section_component.html.erb
TobyRet: Yep cool - if it's not a blocker then I could look into this further after this PR is merged. I think passing in html into the form builder also feels a bit yucky to be honest (ideally the form builder component should just accept an array of hints and the component should deal with the markup?). Perhaps I can post an issue on the repo and provide a fix.

TobyRet:
Another PR will follow to remove the defaults for all the `xxx_completed` columns in the ApplicationForm table. Once this happens the form validation will kick in. 

https://ukgovernmentdfe.slack.com/archives/CP18YJXPY/p1619447832374700

OK to merge @paulrobertlloyd & @frankieroberto ?

frankieroberto (approved):

frankieroberto:
@TobyRet yep, a follow-up PR makes sense to me.

app/components/candidate_interface/complete_section_component.html.erb
```diff
@@ -1,8 +1,21 @@
...
+<%= form_with model: section_complete_form, url: path, method: request_method do |f| %>
+  <%= f.govuk_error_summary %>
+
+  <%= render(review_component) %>
+
+  <%= f.govuk_radio_buttons_fieldset :completed, classes: 'govuk-!-width-two-thirds ', legend: { text: t('application_form.completed_question'), size: 'm' } do %>
+    <% if hint_text %>
+      <% if hint_text.is_a?(Array) %>
+        <% hint_text.each do |paragraph| %>
+          <div class="govuk-hint"><%= paragraph %></div>
```
paulrobertlloyd, line 10:
```suggestion
          <p class="govuk-hint"><%= paragraph %></p>
```

app/components/candidate_interface/complete_section_component.html.erb
```diff
@@ -1,8 +1,21 @@
...
+        <% end %>
+      <% else %>
+        <div class="govuk-hint"><%= hint_text %></div>
```
paulrobertlloyd, line 13:
```suggestion
        <p class="govuk-hint"><%= hint_text %></p>
```

paulrobertlloyd (changes requested):

app/components/candidate_interface/complete_section_component.html.erb
TobyRet: @paulrobertlloyd - the reason for using `<div>` was because an extra margin was being added by `<p>` - would you rather this be overridden?

frankieroberto: @paulrobertlloyd @TobyRet the Design System component uses `<div>`s for hints in radios, for some reason: https://design-system.service.gov.uk/components/radios/

TobyRet: @frankieroberto - @paulrobertlloyd has decided to have the multiple hints functionality removed. This will simplify things and allow us to use the form builder component. I'm going to do this in a followup PR

@paulrobertlloyd can you approve the PR pls?

paulrobertlloyd (approved):

PR #4577, opened 2021-04-22
Add course type parameters to test data generate endpoint

benswannack:
## Context
Providers have asked to be able to generate more realistic applications on Sandbox. The requirement is to be able to generate applications that apply to a mixture of courses: run by the org, ratified by the org, and run by a separate unrelated org.

Add a mixed param to the endpoint which forces the application that s generated to be to the required organisations.
We use the TEST provider that's available on Sandbox, to put the external applications in, to avoid clashing with other testing.

## Link to Trello card
https://trello.com/c/flCTHBtA/3560-support-generating-test-applications-to-both-the-accredited-body-and-their-sds-over-the-api

## Things to check

### DIFF
--- a/app/services/vendor_api/generate_test_applications_for_provider.rb
+++ b/app/services/vendor_api/generate_test_applications_for_provider.rb
@@ -1,27 +1,72 @@
 module VendorAPI
   class GenerateTestApplicationsForProvider
-    def call(provider:, courses_per_application:, count:, for_ratified_courses: false)
+    def initialize(provider:, courses_per_application:, count:, for_training_courses: false, for_ratified_courses: false, for_test_provider_courses: false)
+      @provider = provider
+      @courses_per_application = courses_per_application
+      @application_count = count
+
+      if [for_training_courses, for_ratified_courses, for_test_provider_courses].none?
+        for_training_courses = true
+      end
+      @for_training_courses = for_training_courses
+      @for_ratified_courses = for_ratified_courses
+      @for_test_provider_courses = for_test_provider_courses
+    end
+
+    def call
       raise ParameterInvalid, 'Parameter is invalid (cannot be zero): courses_per_application' if courses_per_application.zero?
 
-      course_ids = course_list_for_provider(provider, for_ratified_courses).pluck(:id)
+      application_count.times do
+        course_ids = random_course_ids_to_apply_for
 
-      raise ParameterInvalid, 'Parameter is invalid (cannot be greater than number of available courses): courses_per_application' if course_ids.count < courses_per_application
+        raise ParameterInvalid, 'Parameter is invalid (cannot be greater than number of available courses): courses_per_application' if course_ids.count < courses_per_application
 
-      GenerateTestApplicationsForCourses.perform_async(course_ids, courses_per_application, count)
+        GenerateTestApplicationsForCourses.perform_async(course_ids, courses_per_application)
+      end
     end
 
   private
 
-    def course_list_for_provider(provider, for_ratified_courses)
-      if for_ratified_courses
-        GetCoursesRatifiedByProvider.call(provider: provider)
-      else
-        Course.current_cycle
-              .open_on_apply
-              .joins(:course_options)
-              .distinct
-              .where(provider: provider)
+    attr_reader :provider, :courses_per_application, :application_count, :for_training_courses, :for_ratified_courses, :for_test_provider_courses
+
+    def random_course_ids_to_apply_for
+      even_split = even_split_for_number_of_course_types
+
+      courses = []
+
+      if for_training_courses
+        courses += courses_run_by_provider.sample(even_split.pop)
       end
+      if for_ratified_courses
+        courses += courses_ratified_by_provider.sample(even_split.pop)
+      end
+      if for_test_provider_courses
+        courses += courses_run_by_test_provider.sample(even_split.pop)
+      end
+
+      courses.pluck(:id)
+    end
+
+    def even_split_for_number_of_course_types
+      course_types_count = [for_training_courses, for_ratified_courses, for_test_provider_courses].count(true)
+
+      3.times.to_a.in_groups(course_types_count, false).map(&:count).shuffle
+    end
+
+    def courses_ratified_by_provider
+      @_courses_ratified_by_provider ||= GetCoursesRatifiedByProvider.call(provider: provider)
+    end
+
+    def courses_run_by_provider
+      @_courses_run_by_provider ||= Course.current_cycle
+                                          .open_on_apply
+                                          .joins(:course_options)
+                                          .distinct
+                                          .where(provider: provider)
+    end
+
+    def courses_run_by_test_provider
+      @_courses_run_by_test_provider ||= TestProvider.training_courses
     end
   end
 end
--- a/app/views/api_docs/vendor_api_docs/pages/release_notes.md
+++ b/app/views/api_docs/vendor_api_docs/pages/release_notes.md
@@ -1,3 +1,14 @@
+## 5th May
+
+The following experimental/sandbox endpoint has been updated:
+
+`/test-data/generate` now accepts optional `for_training_courses` and `for_test_provider_courses` query params.
+
+Supplying `for_training_courses=true` will ensure that applications are generated for courses run by the organisation.
+Supplying `for_test_provider_courses=true` will ensure that applications are generated for courses run by a separate, sandbox only, test provider.
+Supplying none of `for_ratified_courses`, `for_training_courses` or `for_test_provider_courses` as `true`, will result in applications being generated to courses run by the organisation (the same effect as just `for_training_courses=true`)
+
+
 ## 26th April
 
 Add [documentation](/api-docs#how-candidates-and-applications-are-identified) about application and candidate IDs
--- a/app/workers/generate_test_applications_for_courses.rb
+++ b/app/workers/generate_test_applications_for_courses.rb
@@ -1,7 +1,28 @@
 class GenerateTestApplicationsForCourses
   include Sidekiq::Worker
 
-  def perform(course_ids, courses_per_application, count)
+  def perform(course_ids, courses_per_application, count = nil)
+    if count.present?
+      generate_multiple(course_ids, courses_per_application, count)
+    else
+      generate_single(course_ids, courses_per_application)
+    end
+  end
+
+private
+
+  def generate_single(course_ids, courses_per_application)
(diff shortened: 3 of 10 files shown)


app/services/vendor_api/generate_test_applications_for_provider.rb
```diff
@@ -12,16 +12,45 @@ def call(provider:, courses_per_application:, count:, for_ratified_courses: fals
...
+    def course_list_for_provider(provider, for_ratified_courses, mixed_application)
       if for_ratified_courses
-        GetCoursesRatifiedByProvider.call(provider: provider)
+        courses_ratified_by_provider(provider)
+      elsif mixed_application
+        [
+          courses_run_by_provider(provider).first,
+          courses_ratified_by_provider(provider).first,
+          course_run_by_test_provider,
+        ].compact
```
benswannack, line 23: Currently rethinking this...

Issue at the moment is that this won't work if the provider doesn't accredit anywhere. Maybe we can fall back to another course_run_by_provider in this case?

benswannack: One question is, do we want to make sure that all of the applications are "mixed", or just pick randomly from a mixed set of courses? The reason I didn't go for the latter is because asking for "mixed" suggests that you will definitely get a mixture. The former is slightly less nice to look at codewise though

app/services/vendor_api/generate_test_applications_for_provider.rb
```diff
@@ -1,9 +1,9 @@
 module VendorAPI
   class GenerateTestApplicationsForProvider
-    def call(provider:, courses_per_application:, count:, for_ratified_courses: false)
+    def call(provider:, courses_per_application:, count:, for_ratified_courses: false, mixed_application: false)
       raise ParameterInvalid, 'Parameter is invalid (cannot be zero): courses_per_application' if courses_per_application.zero?
 
-      course_ids = course_list_for_provider(provider, for_ratified_courses).pluck(:id)
+      course_ids = course_list_for_provider(provider, for_ratified_courses, mixed_application).pluck(:id)
```
cwrw, line 6: Im unsure around allowing both those values to be set, shouldn't it be either or, as it becomes confusing what takes precedence 🤔

app/services/vendor_api/generate_test_applications_for_provider.rb
```diff
@@ -12,16 +12,45 @@ def call(provider:, courses_per_application:, count:, for_ratified_courses: fals
...
+    def courses_run_by_provider(provider)
+      Course.current_cycle
+            .open_on_apply
+            .joins(:course_options)
+            .distinct
+            .where(provider: provider)
+    end
+
+    def course_run_by_test_provider
+      test_provider = Provider.default_scoped.find_or_create_by(name: 'Test Provider', code: 'TEST')
```
cwrw, line 42: I thought we had a test provider for everyone, I prefer going with one test provider per provider, but wanted to double check after our conversation. Can we reuse `GenerateFakeProvider` to return the existing one with the courses required?

app/services/vendor_api/generate_test_applications_for_provider.rb
cwrw: Agree that mixed should ensure returning at least one of each to avoid confusion

app/services/vendor_api/generate_test_applications_for_provider.rb
benswannack: As discussed, we will add parameters to additively specify which courses we should choose from:
- for_ratified_courses
- for_training_courses
- for_test_provider_courses

Open to suggestions for names of these

app/services/vendor_api/generate_test_applications_for_provider.rb
```diff
@@ -12,16 +12,42 @@ def call(provider:, courses_per_application:, count:, for_ratified_courses: fals
...
-              .where(provider: provider)
+        course_lists << courses_ratified_by_provider(provider)
+      end
+      if for_test_provider_courses
+        course_lists << courses_run_by_test_provider
+      end
+
+      3.times.map do |index|
+        course_lists[index.modulo(course_lists.count)].sample
       end
```
benswannack, line 34: This logic needs to change. We pretty much want to get courses from the lists we have specified with the booleans, and then fill up the array (to length 3) with random courses from all available lists

app/services/vendor_api/generate_test_applications_for_provider.rb
```diff
@@ -1,9 +1,9 @@
 module VendorAPI
   class GenerateTestApplicationsForProvider
-    def call(provider:, courses_per_application:, count:, for_ratified_courses: false)
+    def call(provider:, courses_per_application:, count:, for_training_courses: false, for_ratified_courses: false, for_test_provider_courses: false)
       raise ParameterInvalid, 'Parameter is invalid (cannot be zero): courses_per_application' if courses_per_application.zero?
 
-      course_ids = course_list_for_provider(provider, for_ratified_courses).pluck(:id)
+      course_ids = course_list_for_provider(provider, for_training_courses, for_ratified_courses, for_test_provider_courses).pluck(:id)
```
benswannack, line 6: At the moment, this will generate `count` application forms, all with the same course choices.

To make it a bit more realistic, we can run this `count` times and enqueue a Sidekiq job for each. I think that's the only was to do this without a bigger refactor.

app/services/vendor_api/generate_test_applications_for_provider.rb
```diff
@@ -1,27 +1,68 @@
...
 
-      course_ids = course_list_for_provider(provider, for_ratified_courses).pluck(:id)
+      application_count.times do
+        course_ids = random_course_ids_to_apply_for
 
-      raise ParameterInvalid, 'Parameter is invalid (cannot be greater than number of available courses): courses_per_application' if course_ids.count < courses_per_application
+        raise ParameterInvalid, 'Parameter is invalid (cannot be greater than number of available courses): courses_per_application' if course_ids.count < courses_per_application
 
-      GenerateTestApplicationsForCourses.perform_async(course_ids, courses_per_application, count)
+        GenerateTestApplicationsForCourses.perform_async(course_ids, courses_per_application)
```
cwrw, line 24: I've been having a think around this: What if we call the test generator as many times as flags, example:
I've set all 3 flags as true. I grab the courses for each of those, and call `GenerateTestApplicationsForCourses.perform_async(course_ids, courses_per_application, count)` three time, each time with the specific course (ratified courses or training courses). This way we don't have to worry about the random sampling later on. Only thing we need to pass in then is the count which we can spread across the options, for instance the count requested is 30, I can pass 10 to each job. Happy to chat through this and hear your thoughts

benswannack: I don't think this will meet the need for mixed applications :( 
We need this to generate `count` application _forms_, with `courses_per_application` application choices, where the course choices (per application form) are spread between the course types specified.

For example, if we do `?count=2&courses_per_application=3&for_ratified_courses=true&for_training_courses=true&for_test_provider_courses=true`. Then I'd expect to get 2 application forms, each with 3 application choices. On each form, there should be an application choice applying to each of: a course run by the provider, a course ratified by the provider, and a course run by the test provider.

benswannack:
I still need to add API docs for this

app/services/vendor_api/generate_test_applications_for_provider.rb
cwrw: bah ok makes sense, forgot about the mixed bit!

app/views/api_docs/vendor_api_docs/pages/release_notes.md
```diff
@@ -1,3 +1,14 @@
+## 29th April
+
+The following experimental/sandbox endpoint has been updated:
+
+- `/test-data/generate` now accepts optional `for_training_courses` and `for_test_provider_courses` query params.
```
cwrw, line 5:
```suggestion
`/test-data/generate` now accepts optional `for_training_courses` and `for_test_provider_courses` query params.
```

spec/services/vendor_api/generate_test_applications_for_provider_spec.rb
```diff
@@ -1,70 +1,139 @@
...
+      count: application_count,
+      for_training_courses: for_training_courses,
+      for_ratified_courses: for_ratified_courses,
+      for_test_provider_courses: for_test_provider_courses,
+    }
+  end
+
+  before do
+    create(:course_option)
+    # rubocop:disable FactoryBot/CreateList
```
cwrw, line 23: why did we need to disable rubocop and not use `create_list`?

app/services/vendor_api/generate_test_applications_for_provider.rb
```diff
@@ -1,27 +1,68 @@
...
+      end
     end
 
   private
 
-    def course_list_for_provider(provider, for_ratified_courses)
+    attr_reader :provider, :courses_per_application, :application_count, :for_training_courses, :for_ratified_courses, :for_test_provider_courses
+
+    def random_course_ids_to_apply_for
+      course_types_count = [for_training_courses, for_ratified_courses, for_test_provider_courses].count(true)
```
cwrw, line 33: can the even_split logic move to its own method?

cwrw (approved):
This is a good compromise for what we are working with, well done!  some minor comments, going to give this a test locally as well but otherwise 👍

spec/services/vendor_api/generate_test_applications_for_provider_spec.rb
benswannack: Because I'm creating a new course each time too. If I use a create_list, then it makes 3 course options all with the same course

app/workers/generate_test_applications_for_courses.rb
```diff
@@ -1,15 +1,13 @@
 class GenerateTestApplicationsForCourses
   include Sidekiq::Worker
 
-  def perform(course_ids, courses_per_application, count)
+  def perform(course_ids, courses_per_application)
```
cwrw, line 4: while testing locally old jobs errored due to the wrong number of arguments being passed in, we need to watch out for this when deploying this and any existing jobs that could be enqueued

cwrw: Or we can make this backward compatible for now, and remove later

benswannack: That's a really good spot, backwards compatibility sounds sensible! We can leave in the old method verbatim and have a separate PR removing it later

PR #4578, opened 2021-04-22
Report and ignore UCASMatch records with no applications

mnacos:
## Context

We are getting many sidekiq failures on QA related to bad UCASMatch data. Many tasks still enqueued in Retries, which will never succeed because they have `nil` arguments.

image

Obviously, we could clear the Retries queue, but the moment we switch on ucas matching again, other similar tasks will probably be enqueued. I haven't got to the bottom of the data issue, but the failures are caused by UCASMatch records with next_action of initial_emails_sent which return no `application_for_the_same_course_in_progress_on_both_services?` records. These UCASMatch records are not actionable, but will always be considered during ucas matching, which is not ideal. Maybe we need an additional enum/action_taken along the lines of :no_application_choice which is set so that the process forgets about them. But we should also raise an error, lest this masks data issues on production. 

Introduce a `no_application_choice` enum/action_taken, and include it in the `resolved?` states.
Set this as `action_taken` for UCASMatch records that return no application choices.
Raise a meaningful error so that we hear about such cases from Sentry.

Add a counter for problematic UCAS matches to the relevant dashboard:

image

Individual UCAS matches with `no_application_choice` are also visible in the usual list:

image

## Guidance to review

Should be easy. Does the approach make sense?

## Link to Trello card

No Trello card

## Things to check

### DIFF
--- a/app/models/ucas_match.rb
+++ b/app/models/ucas_match.rb
@@ -10,6 +10,7 @@ class UCASMatch < ApplicationRecord
     resolved_on_apply: 'resolved_on_apply',
     resolved_on_ucas: 'resolved_on_ucas',
     manually_resolved: 'manually_resolved',
+    no_application_choice: 'no_application_choice',
   }
 
   def ready_to_resolve?
@@ -17,7 +18,7 @@ class UCASMatch < ApplicationRecord
   end
 
   def resolved?
-    %w[resolved_on_apply resolved_on_ucas manually_resolved].include?(action_taken)
+    %w[resolved_on_apply resolved_on_ucas manually_resolved no_application_choice].include?(action_taken)
   end
 
   def trackable_applicant_key
--- a/app/services/ucas_matches/send_ucas_match_initial_emails_duplicate_applications.rb
+++ b/app/services/ucas_matches/send_ucas_match_initial_emails_duplicate_applications.rb
@@ -10,13 +10,18 @@ module UCASMatches
       raise "Initial emails for UCAS match ##{ucas_match.id} were already sent" if ucas_match.initial_emails_sent?
 
       application_choice = ucas_match.application_choices_for_same_course_on_both_services.first
-      raise "No application choices found for UCAS match ##{ucas_match.id}" unless application_choice
+      report_bad_data!(ucas_match) if application_choice.blank?
 
       CandidateMailer.ucas_match_initial_email_duplicate_applications(application_choice).deliver_later
 
       application_choice.provider.provider_users.each do |provider_user|
         ProviderMailer.ucas_match_initial_email_duplicate_applications(provider_user, application_choice).deliver_later
       end
+    end
+
+    def report_bad_data!(ucas_match)
+      ucas_match.update!(action_taken: :no_application_choice)
+      raise "No application choices found for UCAS match ##{ucas_match.id}"
     end
   end
 end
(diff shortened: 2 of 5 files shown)


mnacos:
> **Database-level enum changes detected**
> 
> Please include a data migration for these attributes and values:

No data migration needed as we are adding a new value which has never been set before.

app/models/ucas_match.rb
```diff
@@ -10,14 +10,15 @@ class UCASMatch < ApplicationRecord
...
+    no_application_choice: 'no_application_choice',
   }
 
   def ready_to_resolve?
     action_taken.present? && !action_needed? && !resolved?
   end
 
   def resolved?
-    %w[resolved_on_apply resolved_on_ucas manually_resolved].include?(action_taken)
+    %w[resolved_on_apply resolved_on_ucas manually_resolved no_application_choice].include?(action_taken)
```
cwrw, line 21: can we add a spec for this in `resolved`?

app/models/ucas_match.rb
```diff
@@ -10,14 +10,15 @@ class UCASMatch < ApplicationRecord
     resolved_on_apply: 'resolved_on_apply',
     resolved_on_ucas: 'resolved_on_ucas',
     manually_resolved: 'manually_resolved',
+    no_application_choice: 'no_application_choice',
```
cwrw, line 13: I think we might need to surface this new value in the UCAS views and action list

app/services/ucas_matches/send_ucas_match_initial_emails_duplicate_applications.rb
```diff
@@ -10,12 +10,18 @@ def call
...
       CandidateMailer.ucas_match_initial_email_duplicate_applications(application_choice).deliver_later
 
       application_choice.provider.provider_users.each do |provider_user|
         ProviderMailer.ucas_match_initial_email_duplicate_applications(provider_user, application_choice).deliver_later
       end
     end
+
+    def report_bad_data!(ucas_match)
+      ucas_match.update!(action_taken: :no_application_choice)
+      raise "No application choices found for UCAS match ##{ucas_match.id}"
```
cwrw, line 24: What will this mean for the sidekiq failures, will we still see a failure for the job which will then retry? However we won't see those again on a subsequent import as they will pass the resolved stage. 

What is the plan to revisit those matches once the UCAS issue is resolved, will we need to reset those values?

cwrw:
@mnacos Hello, any update/blockers on this PR?

app/models/ucas_match.rb
mnacos: Added a 'bad data' section to the dashboard. Do you think we need a new label?

app/services/ucas_matches/send_ucas_match_initial_emails_duplicate_applications.rb
mnacos: The idea is we still get a Sentry notification but we automatically resolve such cases, because there's nothing we can do with them. For the pending retries on QA, this means we'll still get one last notification for each pending message, before they are automatically resolved. The Retries queue will be cleared when this is deployed.

app/models/ucas_match.rb
mnacos: 'No application choice' UCAS matches already show up, can be filtered etc.

cwrw (approved):
looks good, thanks for addressing comments 👍

PR #4580, opened 2021-04-23
Fix the WhoRanWhichExportExport

davidgisbey:
## Context

I forgot to update the export type enum when I refactored the name of the export

- Add the correct name to the export_type enum and alphabetise it

## Things to check


davidgisbey:
this can be ignored as one hasn't been created in the db yet ☝️

JR-G (approved):
Checked on the review app and can now view the history tab 👍

PR #4581, opened 2021-04-23
Do not display Accept Cookies prompt when printing applications

TobyRet:
## Context

When the Cookie banners were updated to use JS we did not add the necessary class to ensure they do not appear when printing applications

- Add `govuk-!-display-none-print` class to banners

## Link to Trello card

https://trello.com/c/eHPVAjGu/3328-do-not-display-accept-cookies-prompt-when-you-download-a-pdf-of-an-application

## Things to check

### DIFF
--- a/app/components/cookie_banners_component.html.erb
+++ b/app/components/cookie_banners_component.html.erb
@@ -1,60 +1,65 @@
-<%= govuk_cookie_banner(
-  title: "Cookies on #{service_name}",
-  html_attributes: {
-    hidden: false,
-    aria: {
-      label: "Cookies on #{service_name}",
-    },
-    data: {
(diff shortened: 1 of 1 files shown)


paulrobertlloyd:
`govuk-!-display-none-print` is currently replacing when it should be in addition to `govuk-cookie-banner`. Not sure if using the `classes` attribute overwrites the default class too? Either way, containing div should have both of these classes on them.

duncanjbrown:
Is the `class:` option overriding some built in styling?

### Before

Screenshot 2021-04-23 at 14 47 54

### After

Screenshot 2021-04-23 at 14 48 07

davidgisbey (approved):

TobyRet:
All fixed now

printable_application

PR #4582, opened 2021-04-23
Remove candidate sign-in confirmation email

duncanjbrown:
As discussed with @cwrw @davidgisbey and @george-bit, it's only possible to sign in via a magic link, which is delivered to the candidate via email. If an attacker controls the candidate's email account, they're also going to see the sign-in confirmation email and be able to delete it. Therefore it's not useful to send sign in confirmations to candidates, so revert a bit of #4517.


george-bit (approved):
LGTM! Punch it chewy 🚀

PR #4584, opened 2021-04-23
Add strong_migrations gem

malcolmbaig:
## Context
This automatically detects potentially unsafe schema migrations and
provides guidance around how to complete these operations with zero
downtime and/or better performance.

Any migration code that triggers an error can be reviewed and then
marked as safe by wrapping it in a block, eg -

```
safety_assured {
  # migration code
}
```
Benefits:
- Detecting unsafe schema migrations becomes an automated process rather than an entirely manual one driven by code reviews.
- Encourages good migration practice via its error messages and documentation.
- Easily removed/disabled if it starts causing any issues with migrations.

Add the gem and some default config.

### Example

**Operation:**
```ruby
add_column :application_forms, :testing_strong_migrations, :boolean, default: true
```
**Output:**
```ruby
Migrating to TestStrongMigrationsGem (20210423113024)
== 20210423113024 TestStrongMigrationsGem: migrating ==========================
  TRANSACTION (0.1ms)  BEGIN
   (0.1ms)  SET statement_timeout TO 3600000
   (0.1ms)  SET lock_timeout TO 10000
   (0.1ms)  SHOW server_version_num
  TRANSACTION (0.1ms)  ROLLBACK
   (0.1ms)  SELECT pg_advisory_unlock(1548636341943550335)
An error has occurred, this and all later migrations canceled:


=== Dangerous operation detected #strong_migrations ===

Adding a column with a non-null default blocks reads and writes while the entire table is rewritten.
Instead, add the column without a default value, then change the default.

class TestStrongMigrationsGem < ActiveRecord::Migration[6.1]
  def up
    add_column :application_forms, :testing_strong_migrations, :boolean
    change_column_default :application_forms, :testing_strong_migrations, true
  end

  def down
    remove_column :application_forms, :testing_strong_migrations
  end
end

Then backfill the existing rows in the Rails console or a separate migration with disable_ddl_transaction!.

class BackfillTestStrongMigrationsGem < ActiveRecord::Migration[6.1]
  disable_ddl_transaction!

  def up
    ApplicationForm.unscoped.in_batches do |relation|
      relation.update_all testing_strong_migrations: true
      sleep(0.01)
    end
  end
end
```

## Things to check

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -106,6 +106,9 @@ gem 'geocoder'
 
 gem 'strip_attributes'
 
+# Automate checks for potentially unsafe migrations
+gem 'strong_migrations'
+
 group :development do
   gem 'web-console', '>= 3.3.0'
   gem 'listen', '>= 3.0.5', '< 3.6'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -477,6 +477,8 @@ GEM
       sprockets (>= 3.0.0)
     strip_attributes (1.11.0)
       activemodel (>= 3.0, < 7.0)
+    strong_migrations (0.7.6)
+      activerecord (>= 5)
     super_diff (0.6.2)
       attr_extras (>= 6.2.4)
       diff-lcs
@@ -611,6 +613,7 @@ DEPENDENCIES
   skylight
   sprockets-rails
   strip_attributes
+  strong_migrations
   super_diff
   timecop
   tzinfo-data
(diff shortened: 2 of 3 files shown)


duncanjbrown (approved):
Let's give this a go?

mnacos:
> Let's give this a go?

Should we also fix all db migrations so far so that a `rake db:migrate` succeeds for a brand new database/setup?

PR #4587, opened 2021-04-26
Add the candidate ID to application choices export

cwrw:
## Context

We do not surface the candidate ID in the application choice export on the applications list page.

Add a new column for candidate ID as requested by providers.

## Guidance to review

Is this a breaking change for providers consuming the export?

## Things to check


duncanjbrown (approved):

duncanjbrown:
I don't think we should support consumers using offsets to pick columns when they've already got a header row in the CSV

PR #4589, opened 2021-04-26
Add export for feedback provided on Find

davidgisbey:
## Context

On the course and results page on Find, we give users the option to provide feedback. This is persisted to the Apply db, but we currently aren't exposing it via an export.

- Add an export to for feedback provided on Find
- Add docs 

## Guidance to review

Does linking to the page they were on seem reasonable? I could just expose the path/find_controller

## Link to Trello card

https://trello.com/c/aIl3Z8AB/3329-add-an-export-for-findfeedback

## Things to check

### DIFF
--- a/app/exports/find_feedback_export.yml
+++ b/app/exports/find_feedback_export.yml
@@ -1,0 +1,16 @@
+common_columns:
+  - email
+
+custom_columns:
+  feedback_provided_at:
+    type: string
+    format: date-time
+    example: 2020-11-01T00:00:00+00:00
+  find_url:
+    type: string
+    format: link
+    description: A link to the page that the candidate was on when they chose to provide feedback.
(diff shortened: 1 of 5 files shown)


davidgisbey:
This ☝️ is a new enum value so the warning can be ignored

app/exports/find_feedback_export.yml
```diff
@@ -0,0 +1,16 @@
...
+  - email
+
+custom_columns:
+  feedback_provided_at:
+    type: string
+    format: date-time
+    example: 2020-11-01T00:00:00+00:00
+  find_url:
+    type: string
+    format: link
```
davidgisbey, line 11: This might be wrong. I'm just trying to get across that it links directly to the page on Find. That's covered by description though so it might be redundant. WDYT

JR-G (approved):
Tested on the review app and it works ✅

PR #4592, opened 2021-04-26
SPIKE - Trim trail/leading whitespace autosuggest inputs

stevehook:
## Context

Currently when you enter a leading or trailing space in any of the auto-completes it filters out any entries that don't match the pattern of characters and spaces input. We want to ignore trailing and leading spaces for auto-complete purposes as well as trim them from values that end up in the database.

This can be done by providing a custom `source` option.

## Guidance to review

- We probably would want to make this an option for our own wrapper. So this would need to be generalised. Ideas on how to do that?
- Also would want to avoid the `JSON.parse` every time we suggest.

## Link to Trello card

https://trello.com/c/f2yaj7CF/3318-strip-away-white-space-degree-autofill

## Things to check

### DIFF
--- a/app/frontend/packs/autosuggests/init-autosuggest.js
+++ b/app/frontend/packs/autosuggests/init-autosuggest.js
@@ -16,7 +16,12 @@ export const initAutosuggest = ({ inputIds, containerId, templates = {}, styles 
           templates: {
             inputValue: templates.inputTemplate,
             suggestion: templates.suggestionTemplate
-          }
+          },
+          source: (query, populateResults) => {
+            const source = JSON.parse(container.dataset.source)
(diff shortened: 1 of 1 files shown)


malcolmbaig:
Approach looks sound 👍🏼 . 

Could the server handle the lower casing and stripping of whitespace, and then render it in the page alongside the user-facing dataset for use by the function?

stevehook:
> Approach looks sound 👍🏼 .
> 
> Could the server handle the lower casing and stripping of whitespace, and then render it in the page alongside the user-facing dataset for use by the function?

Thats a good question, I think the server could lower case the list of available values, though we would need to have the mixed case list too for display purposes. I think we would still need to lower case and trim the user input in JS though.

stevehook:
Closing this one in favour of https://github.com/DFE-Digital/apply-for-teacher-training/pull/4628

PR #4594, opened 2021-04-26
Trim degree type values before they are saved

stevehook:
## Context

Candidates can enter degree types as free text (as well as by selecting from the auto-suggest). This sometimes leads to values for the same type differing only by trailing or leading whitespace. We want to remove this so that degree qualifications are easier to analyse.

This PR attempts to stop values with trailing or leading whitespace getting into the database in future by trimming them before save.
A separate migration will be needed to fix up any values that are already in the database and should be deployed separately after this one.

## Guidance to review

- Are there other ways to input this attribute that I've missed?

## Link to Trello card

https://trello.com/c/rvoXuX31/3317-strip-away-trailing-space-degree-autofills

## Things to check

### DIFF
--- a/app/forms/candidate_interface/degree_type_form.rb
+++ b/app/forms/candidate_interface/degree_type_form.rb
@@ -14,6 +14,7 @@ module CandidateInterface
     validates :international_type_description, length: { maximum: 255 }
 
     def save
+      sanitize!
       return false unless valid?
       return false unless application_form_present?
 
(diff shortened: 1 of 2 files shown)


davidgisbey (commented):
Sorry Steve, i just thought, i think this is being done in the controller now right?

image

davidgisbey:
We def need the data migration though!

stevehook:
> Sorry Steve, i just thought, i think this is being done in the controller now right?
> 
> image

Ah, my search missed that. OK I'm going to close this one and make the migration ready for review...

stevehook:
Closing because this is already covered

PR #4601, opened 2021-04-27
Send request data to Bigquery

steventux:
## Context

The Data infrastructure team are moving some of Apply's reporting to BigQuery and they. As part of this piece of work they would like to see what they can glean from request-level data — ie which path users are requesting, if possible accompanied by user context like ID.

Emits a common request data event via an `ActiveSupport::Concern` which is queued for processing with Sidekiq.
The Sidekiq worker sends this request event data to the specified BigQuery project and dataset, these dependencies are configured outside of this PR.

## Guidance to review

To run locally you will need the BiqQuery service keyfile for the test account which can be configured as the ENV var `BIG_QUERY_API_JSON_KEY`. (I can supply this).
It's possible then to make requests to your local development server (be sure to have sidekiq processing jobs) and see the events emitted and saved in BigQuery with the following console code:

```
bq = Google::Cloud::Bigquery.new(project: ENV.fetch('BIG_QUERY_PROJECT_ID'))
dataset = bq.dataset(ENV.fetch('BIG_QUERY_DATASET'), skip_lookup: true)
bq_table = dataset.table('events', skip_lookup: true)
bq_table.data max: 50
```

## Link to Trello card

https://trello.com/c/CHjbpUUY/3526-spike-sending-request-data-to-bigquery

## Things to check


Gemfile.lock
```diff
@@ -163,6 +164,37 @@ GEM
     geocoder (1.6.7)
     globalid (0.4.2)
       activesupport (>= 4.2.0)
+    google-apis-bigquery_v2 (0.9.0)
```
cwrw, line 167: wow that gem added a huge amount of dependencies 🤔

app/services/data_migrations/define_bq_apply_request_events_schema.rb
```diff
@@ -0,0 +1,19 @@
+module DataMigrations
+  class DefineBqApplyRequestEventsSchema
+    TIMESTAMP = 20210427185904
+    MANUAL_RUN = false
+
+    def change
```
misaka, line 6: This created a table in the US data center, can that be sorted here?

And anyway, do we want to create the table from the app? If we were manipulating the data further with the app then maybe, but I'm not sure about the responsibility for creating the table belonging to the app when we're likely to have a fair bit of machinery built in BQ that relies on this data ... the app is just something to stream data into it.

🤔

config/initializers/bigquery.rb
```diff
@@ -0,0 +1,12 @@
+require 'google/cloud/bigquery'
+
+BIG_QUERY_API_JSON_KEY = ENV['BIG_QUERY_API_JSON_KEY']
+
+if BIG_QUERY_API_JSON_KEY.present?
+  Google::Cloud::Bigquery.configure do |config|
+    config.project_id  = ENV['BIG_QUERY_PROJECT_ID']
+    config.credentials = "config/bq.json"
```
misaka, line 8: The docs on that appear to say that this can be a JSON string of the file contents, it'll probably make sense to use an env var in deployed envs.

app/lib/events/event.rb
```diff
@@ -0,0 +1,32 @@
...
+      )
+
+      self
+    end
+
+    def with_user_and_namespace(user, namespace)
+      @event_hash.merge!(
+        namespace: namespace,
+        user_id: user&.id,
+      )
```
misaka, line 27: Do we have `support_ref` at this point? That's a key identifier used in the existing reports, could be useful here.

duncanjbrown: only for candidates — what's here is the lowest common denominator between namespaces. Would it makes sense to get the machinery in & ship this structure, then have another conversation about the schema?

misaka: Ya, I think one step at a time works.

config/initializers/bigquery.rb
steventux: :+1: Using an ENV var for JSON key file contents is the pattern used in [teaching-vacancies](https://github.com/DFE-Digital/teaching-vacancies/blob/e81f594426d731d3c88139a23b59f5cea7f48e1a/config/initializers/big_query.rb#L8)

app/services/data_migrations/define_bq_apply_request_events_schema.rb
steventux: It looks like when we create the dataset we should specify the location https://github.com/googleapis/google-cloud-ruby/blob/HEAD/google-cloud-bigquery/samples/snippets/create_dataset.rb#L21
I may have rails-consoled this test US dataset into life, can we blow it away easily?

I'm not sold on doing application migrations to create this dataset and table either, happy to let that be an admin task outside the application.

config/initializers/bigquery.rb
```diff
@@ -0,0 +1,12 @@
+require 'google/cloud/bigquery'
+
+BIG_QUERY_API_JSON_KEY = ENV['BIG_QUERY_API_JSON_KEY']
+
+if BIG_QUERY_API_JSON_KEY.present?
+  Google::Cloud::Bigquery.configure do |config|
+    config.project_id  = ENV['BIG_QUERY_PROJECT_ID']
+    config.credentials = JSON.parse(BIG_QUERY_API_JSON_KEY)
```
misaka, line 8: Just curious about the failure mode here. In the initialiser here if the required ENV vars aren't set, then the code continues to try to configure `Google::Cloud::Bigquery` (it doesn't use `ENV.fetch`). In the Sidekiq worker it then requires the ENV vars be set by using `ENV.fetch`.

Are we doing it this way to allow the app to run even if the ENV vars aren't set? This means the Sidekiq workers could die in certain circumstances, would it make more sense to have the initialiser ensure that BigQuery has been setup, and use an actual feature flag to control it?

steventux: Good point, I think a feature flag to control this would be useful. I suppose we need to know that the configuration is good at application startup and use the same method of configuration in the worker. I'll add as part of this spike PR. I've added a A/C of feature flag to the implementation card.

app/workers/send_request_events_to_bigquery.rb
```diff
@@ -0,0 +1,19 @@
+class SendRequestEventsToBigquery
+  include Sidekiq::Worker
+
+  sidekiq_options retry: 3, queue: :low_priority
+
+  TABLE_NAME = 'apply_request_events'.freeze
+
+  def perform(request_event_json)
+    big_query_project_id = ENV['BIG_QUERY_PROJECT_ID']
+    big_query_dataset = ENV['BIG_QUERY_DATASET']
```
duncanjbrown, line 10: Should these use `fetch`? Good to fail early in case we misconfigured

config/initializers/bigquery.rb
```diff
@@ -0,0 +1,14 @@
...
+if FeatureFlag.active?(:send_request_data_to_bigquery)
+  BIG_QUERY_API_JSON_KEY = ENV['BIG_QUERY_API_JSON_KEY']
+
+  if BIG_QUERY_API_JSON_KEY.present?
+    Google::Cloud::Bigquery.configure do |config|
+      config.project_id  = ENV.fetch('BIG_QUERY_PROJECT_ID')
+      config.credentials = JSON.parse(BIG_QUERY_API_JSON_KEY)
+    end
+
+    Google::Cloud::Bigquery.new(project: ENV.fetch('BIG_QUERY_PROJECT_ID'))
```
duncanjbrown, line 12: Haven't we just configured this at line 8?

spec/support/bigquery.rb
```diff
@@ -0,0 +1,12 @@
+RSpec.configure do |config|
```
duncanjbrown, line 1: I guess this is so we don't try to push every request of every system spec to BQ? Would it be better to stub the worker's `perform` in that case? (and maybe add metadata to say "stub it unless example.metadata[:with_bigquery].present?" or something?)

duncanjbrown (commented):
Looking very good, last minor qs inline

config/initializers/bigquery.rb
steventux: :+1: https://googleapis.dev/ruby/google-cloud-bigquery/latest/Google/Cloud/Bigquery.html#new-class_method
```
Parameters:

    project_id (String) (defaults to: nil) —

    Identifier for a BigQuery project. If not present, the default project for the credentials is used.
```

so no need to pass `project_id` twice here

spec/support/bigquery.rb
steventux: Yeah I really haven't taken care of the test suite comprehensively in this PR but it would be good to have a default mechanism to stub the BQ API calls and still test the worker class.

app/services/data_migrations/define_bq_apply_request_events_schema.rb
steventux: Resolving this as we've removed the table creating data migration from this PR - will be done in Cloud Studio.

app/lib/events/event.rb
```diff
@@ -0,0 +1,32 @@
+module Events
+  class Event
+    def initialize
+      @event_hash = {
+        environment: HostingEnvironment.environment_name,
+      }
+    end
+
+    def as_json
+      @event_hash.as_json
```
misaka, line 10: Just noticed that we don't appear to be getting a timestamp in the BigQuery table. I'll recreate the table with that field and let you know it's name so we can add it here.

spec/support/bigquery.rb
steventux: @duncanjbrown I've amended with your suggestion here https://github.com/DFE-Digital/apply-for-teacher-training/pull/4601/commits/8a37543823cc721871e4069c7270d37db7e0f77b

app/lib/events/event.rb
misaka: I've just:

- added the optional field `timestamp`. I'm considering making this required, will see what Ravi thinks
- changed `user_id` to be a `BIGNUMERIC`
- renamed the table to `request_events`, just to reduce the duplication of `apply` in the name

config/initializers/bigquery.rb
steventux: https://github.com/DFE-Digital/apply-for-teacher-training/pull/4601/files#diff-5265f727a65a1861aa971a561450e9d062a12e37cd8b21531a1ab4721bb85f71R11

app/workers/send_request_events_to_bigquery.rb
```diff
@@ -0,0 +1,14 @@
+class SendRequestEventsToBigquery
+  include Sidekiq::Worker
+
+  sidekiq_options retry: 3, queue: :low_priority
+
+  TABLE_NAME = 'apply_request_events'.freeze
```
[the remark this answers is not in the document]
steventux: As discussed table name is now `events`

config/initializers/bigquery.rb
steventux: Added feature flag here https://github.com/DFE-Digital/apply-for-teacher-training/pull/4601/commits/9f79d635de040ffecc98bec00456cde054f5f998

duncanjbrown (approved):

duncanjbrown:
Thanks for changes @steventux!

.env.development
```diff
@@ -6,3 +6,6 @@ GOOGLE_ANALYTICS_APPLY=UA-APPLY-XX
 GOOGLE_ANALYTICS_MANAGE=UA-MANAGE-XX
 GOOGLE_TAG_MANAGER_APPLY=GTM-APPLY-XX
 GOOGLE_TAG_MANAGER_MANAGE=GTM-MANAGE-XX
+BIG_QUERY_API_JSON_KEY={}
+BIG_QUERY_PROJECT_ID=apply-bigquery-dataset
```
misaka, line 10: I guess this is just a placeholder, but it's not right is it? Will we add documentation for how to configure this? Can be a separate PR.

misaka (approved):
🚀

.env.development
steventux: I'll add some info on configuration for BQ in docs and see if we even need these placeholders, will merge this one and update in next PR.

PR #4603, opened 2021-04-27
Introduce ValidationException and update provider authorisation

despo:
## Context
Introduce a ValidationException and update ProviderAuthorisation to raise error if no course_option or course_option.id is provided. This enables getting more meaninful errors and being able to check if the user is authorised to perform certain actions before running validations and potentially disclosing information on data that they don't have the right to amend, for example in [MakeAnOffer](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/services/make_an_offer.rb#L37).

- Introduce ValidationException
- Update ProviderAuthorisation to allow for course_option as well as course_option_id and validate presence of course_option prior to authorising user.

## Link to Trello card
Related to: 3620-add-make-offer-validations-to-the-vendor-api

## Things to check

### DIFF
--- a/app/errors/validation_exception.rb
+++ b/app/errors/validation_exception.rb
@@ -1,0 +1,22 @@
+class ValidationException < StandardError
+  attr_accessor :error_messages
+
+  def initialize(error_messages)
+    @error_messages = error_messages
+  end
+
+  def as_json
+    errors = error_messages.map do |error_message|
+      {
+        error: 'ValidationError',
+        message: error_message,
+      }
+    end
+
+    { errors: errors }
+  end
+
+  def message
+    error_messages.join(', ')
+  end
+end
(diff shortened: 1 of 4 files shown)


app/errors/validation_exception.rb
```diff
@@ -0,0 +1,22 @@
...
+  def as_json
+    errors = error_messages.map do |error_message|
+      {
+        error: 'ValidationError',
+        message: error_message,
+      }
+    end
+
+    { errors: errors }
+  end
```
despo, line 17: Enables easy convertion of the errors in the format required by the VendorAPI.
https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/controllers/vendor_api/vendor_api_controller.rb#L35

cwrw (approved):

PR #4604, opened 2021-04-27
Remove 'multiple hints' functionality from Section Completed component

TobyRet:
## Context

Originally the design intent was to be able to pass in multiple hints into the `SectionCompletedComponent` and display as separate paragraphs. However for simplicity we will now pass in a single string and display as a single paragraph. This will also allow us to use the `govuk_collection_radio_buttons` from Govuk Design System Formbuilder gem.

- `SectionCompleteComponent` now displays a single hint only
- Migrate to `govuk_collection_radio_buttons` in the `SectionCompleteComponent`

## Guidance to review

- Go to any review section in the application to view the SectionCompleteComponent

section_complete_single_hint

## Link to Trello card

https://trello.com/c/Fu5pfwQi/3333-remove-multiple-hints-functionality-from-section-completed-component

## Things to check

### DIFF
--- a/app/components/candidate_interface/complete_section_component.html.erb
+++ b/app/components/candidate_interface/complete_section_component.html.erb
@@ -5,13 +5,7 @@
 
   <%= f.govuk_radio_buttons_fieldset :completed, classes: 'govuk-!-width-two-thirds ', legend: { text: t('application_form.completed_question'), size: 'm' } do %>
     <% if hint_text %>
-      <% if hint_text.is_a?(Array) %>
-        <% hint_text.each do |paragraph| %>
-          <div class="govuk-hint"><%= paragraph %></div>
-        <% end %>
-      <% else %>
-        <div class="govuk-hint"><%= hint_text %></div>
-      <% end %>
+      <div class="govuk-hint"><%= hint_text %></div>
     <% end %>
     <%= f.govuk_radio_button :completed, true, label: { text: complete_or_reviewed_radio_button_label }, link_errors: true %>
     <%= f.govuk_radio_button :completed, false, label: { text: t('application_form.incomplete_radio') } %>
--- a/app/components/candidate_interface/complete_section_component.html.erb
+++ b/app/components/candidate_interface/complete_section_component.html.erb
@@ -3,12 +3,17 @@
 
   <%= render(review_component) %>
 
-  <%= f.govuk_radio_buttons_fieldset :completed, classes: 'govuk-!-width-two-thirds ', legend: { text: t('application_form.completed_question'), size: 'm' } do %>
-    <% if hint_text %>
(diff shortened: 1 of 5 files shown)


app/components/candidate_interface/complete_section_component.html.erb
```diff
@@ -3,19 +3,14 @@
...
-          <div class="govuk-hint"><%= paragraph %></div>
-        <% end %>
-      <% else %>
-        <div class="govuk-hint"><%= hint_text %></div>
-      <% end %>
-    <% end %>
-    <%= f.govuk_radio_button :completed, true, label: { text: complete_or_reviewed_radio_button_label }, link_errors: true %>
-    <%= f.govuk_radio_button :completed, false, label: { text: t('application_form.incomplete_radio') } %>
-  <% end %>
+  <div class="govuk-!-width-two-thirds">
```
TobyRet, line 6: The `govuk_collection_radio_buttons` accepts classes but unfortunately, these styles are not applied to any hints. Would wrapping in a `<div>` suffice?

paulrobertlloyd: Wonder if this is an issue that needs to be filed upstream? @peteryates

paulrobertlloyd: For now, a `div` around the `govuk_collection_radio_buttons` with the width override class will work.

paulrobertlloyd (approved):

app/components/candidate_interface/complete_section_component.html.erb
TobyRet: > Wonder if this is an issue that needs to be filed upstream? @peteryates

Something like `hint: { text: hint_text, classes: 'foo bar boz' } ` could be nice? 

Happy to raise a fix if it's a goer @peteryates

paulrobertlloyd: I was wondering more if classes added should apply to `govuk-form-group` and not `govuk-radios`? I think the GDS Nunjucks templates allow you to do both.

peteryates: So the form builder has allowed custom hint content for some time via a proc. When a proc is provided it'll use a `div` instead of a `span`. Example:

```erb
<%= form.govuk_text_field :name, hint: -> do %>
  <section style="outline: 2px dashed palevioletred; padding: .5em 1em">
    <h4>Look!</h4>
    <p>Multi-paragraph</p>
  </section>
<% end %>
```

Renders:

Screenshot from 2021-04-27 18-00-43


I did notice that the guide's a little out of date, I'll update it tomorrow.

TobyRet: Great thanks for that @peteryates!

PR #4606, opened 2021-04-27
Add offer validations to the vendor api

despo:
## Context

Introduces an OfferValidation service_model that can be used by both the `MakeOffer` and `ChangeOffer` services to check for data validity and updates `VendorAPI::DecisionsController` to use new services

1. Add OfferValidation model
2. Update `MakeOffer` and `ChangeOffer` to use `OfferValidation`
3. Update `VendorAPI::DecisionsController` to use new Offer services when the `updated_offer_flow` flag is active.

## Link to Trello card

https://trello.com/c/TaJgQjeP/3620-add-make-offer-validations-to-the-vendor-api

## Things to check

### DIFF
--- a/app/controllers/vendor_api/decisions_controller.rb
+++ b/app/controllers/vendor_api/decisions_controller.rb
@@ -17,7 +17,15 @@ module VendorAPI
                         application_choice.course_option
                       end
 
-      if application_choice.offer?
+      if FeatureFlag.active?(:updated_offer_flow)
+        offer_service = if application_choice.offer?
+                          ChangeOffer.new(offer_params(application_choice, course_option))
+                        else
+                          MakeOffer.new(offer_params(application_choice, course_option))
+                        end
+
+        respond_to_decision(offer_service)
+      elsif application_choice.offer?
         change_offer = ChangeAnOffer.new(
           actor: audit_user,
           application_choice: application_choice,
@@ -100,11 +108,25 @@ module VendorAPI
     end
 
     def respond_to_decision(decision)
-      if decision.save
+      if FeatureFlag.active?(:updated_offer_flow) && [MakeOffer, ChangeOffer].include?(decision.class)
+        decision.save!
+        render_application
+      elsif decision.save
         render_application
       else
         render_validation_errors(decision.errors)
       end
+    rescue IdenticalOfferError
+      render_application
+    rescue ValidationException => e
+      render status: :unprocessable_entity, json: e.as_json
+    rescue Workflow::NoTransitionAllowed
+      render status: :unprocessable_entity, json: {
+        errors: [
+          error: 'StateTransitionError',
+          message: 'The application is not ready for that action',
+        ],
+      }
     rescue ProviderAuthorisation::NotAuthorisedError => e
       render status: :unprocessable_entity, json: {
         errors: [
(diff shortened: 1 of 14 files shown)


spec/service_models/offer_validations_spec.rb
```diff
@@ -1,6 +1,33 @@
...
+        let(:course_option) { nil }
+
+        it 'does not add a :not_open_on_apply error' do
+          offer.valid?
+
+          expect(offer.errors[:course_option]).not_to contain_exactly('is not open for applications via the Apply service')
+        end
+      end
+
+      context 'when open on apply' do
```
cwrw, line 24: when not open on apply

cwrw (commented):
looks great! will do another round of testing/reviewing tomorrow but generally love the approach 👍

app/services/change_offer.rb
```diff
@@ -12,21 +14,37 @@ def initialize(actor:,
...
+        ActiveRecord::Base.transaction do
+          ApplicationStateChange.new(application_choice).make_offer!
+
+          application_choice.current_course_option = course_option
+          application_choice.offer = { 'conditions' => conditions }
+          application_choice.offer_changed_at = Time.zone.now
+          application_choice.save!
+        end
+
+        SetDeclineByDefault.new(application_form: application_choice.application_form).call
```
despo, line 30: I think this should probably be wrapped within the transaction as well.

app/services/make_offer.rb
```diff
@@ -12,17 +14,35 @@ def initialize(actor:,
...
   end
+
+private
+
+  def auth
+    @auth ||= ProviderAuthorisation.new(actor: actor)
+  end
+
+  def offer
+    @offer ||= OfferValidations.new(course_option: course_option,
```
cwrw, line 45: I'm getting `uninitialized constant MakeOffer::OfferValidations` whilst testing locally, we need to ensure service models is being loaded, or move the validation to the models

cwrw: or maybe services 🤔

despo: Autoload paths automatically pick any custom directories under app, so we shouldn't. (https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#autoload-paths)

Did you restart your server?

despo: Maybe change it to `::OfferValidations` to ensure that it's not attempting to use local scope?

cwrw: bah yes I restarted and worked! nvm then :D

app/services/change_offer.rb
cwrw: Yea as long as this doesn't trigger any jobs

despo: It only updates the decline_by_default date and number of dates, so it makes sense to rollback the entire transaction if it fails.

app/service_models/offer_validations.rb
```diff
@@ -0,0 +1,36 @@
+class OfferValidations
```
cwrw, line 1: Discussed in person but for posterity: we're going to move this to the services folder pending a wider discussion on where service models will live (ticket added to Dev team board https://trello.com/c/9jCgx5Lv/461-discuss-introducing-new-directory-to-seperate-validation-specific-to-services)

cwrw (commented):
looks good 👍 minor questions and the folder change. The only thing left I believe is left is the ratified provider validation see (https://github.com/DFE-Digital/apply-for-teacher-training/pull/4349/files#diff-60f0d21b0b3a72c1cc33078bfb79a1f47aded3a76d9ab58942e6a754d9fb104dR41). I think we might need to add this, let me know your thoughts

app/controllers/vendor_api/decisions_controller.rb
```diff
@@ -100,11 +108,25 @@ def render_application
     end
 
     def respond_to_decision(decision)
-      if decision.save
+      if FeatureFlag.active?(:updated_offer_flow) && [MakeOffer, ChangeOffer].include?(decision.class)
```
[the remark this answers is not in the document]
despo: `ConfirmOfferConditions`, `ConditionsNotMet`, `WithdrawOffer` and `RejectApplication` are also possible decisions of this controller.

despo:
@cwrw Ratidying provider validation introduced in last commit. I completely missed this so thanks for flagging!

cwrw (approved):
thanks @despo!

PR #4607, opened 2021-04-27
Bugfix - set current_course_option in AddCourseChoiceAfterSubmission

malcolmbaig:
## Context
This is a service object invoked via the support UI when a candidate requests a course choice added to their application after they've already submitted.

Use the `set_initial_course_choice!` method to set both `course_option` and `current_course_option`

Not doing so causes the application dashboard to error out as it expects
the `current_course_option` to be set.

## Guidance to review
- Is there a better way to approach this? Alternatives considered:
  - Fallback to course_option if current_course_option blank
  - Define a single method in `ApplicationChoice` for creating new ones, and use that wherever we're adding a choice to an application form.

## Link to Trello card
na

## Things to check

### DIFF
--- a/app/services/support_interface/add_course_choice_after_submission.rb
+++ b/app/services/support_interface/add_course_choice_after_submission.rb
@@ -8,11 +8,11 @@ module SupportInterface
     end
 
     def call
-      application_choice = ApplicationChoice.create!(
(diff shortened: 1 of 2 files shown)


malcolmbaig:
Closing — had a stale master, didn't realise @mnacos patched this earlier!

PR #4608, opened 2021-04-27
Clear test data for applications the current provider ratifies.

steventux:
## Context

The API endpoint `test-data/clear` clears applications for courses run by the current provider.
It also needs to clear applications data for courses accredited/ratified by the current provider. 

Adds an additional step in the test data controller `clear!` action which destroys any application data (Candidate, and dependents) where the current provider ratifies the course.
 

## Link to Trello card

https://trello.com/c/y1mz1pb3/3494-api-endpoint-for-test-data-clear-does-not-clear-data-for-courses-the-provider-ratifies

## Things to check

### DIFF
--- a/app/controllers/vendor_api/test_data_controller.rb
+++ b/app/controllers/vendor_api/test_data_controller.rb
@@ -27,6 +27,7 @@ module VendorAPI
 
     def clear!
       current_provider.application_choices.map(&:application_form).map(&:candidate).map(&:destroy!)
+      clear_accredited_application_data!
 
       render json: { data: { message: 'Applications cleared' } }
     end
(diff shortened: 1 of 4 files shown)


app/controllers/vendor_api/test_data_controller.rb
```diff
@@ -56,5 +57,12 @@ def courses_per_application_param
     def for_ratified_courses_param
       params[:for_ratified_courses] == 'true'
     end
+
+    def clear_accredited_application_data!
+      Candidate
```
cwrw, line 62: I'll attempt one query for both cases, will push the change

app/models/candidate.rb
```diff
@@ -22,6 +22,12 @@ def self.for_email(email)
     find_or_initialize_by(email_address: email.downcase) if email
   end
 
+  def self.clear_application_data!(provider)
+    scope = joins(application_forms: { application_choices: { course_option: :course } })
+    scope.where("courses.accredited_provider": provider).or(scope.where("courses.provider": provider))
+      .delete_all
+  end
+
```
benswannack, line 30: Could we move this out to a small service?

cwrw: done

benswannack (approved):
LGTM!

PR #4615, opened 2021-04-28
Always open new courses on Sandbox

duncanjbrown:
If we learn about a new course in Sandbox this means somebody has
deliberately created it in the Publish Sandbox, so it saves time to
open it automatically.

## Context

We have to manually open courses for providers/vendors right now and this is distracting and unnecessary.


cwrw (approved):
I just added a ticket for this :D in the backlog, will remove

PR #4616, opened 2021-04-28
Resolve API encoding error

despo:
## Context

Resolve API encoding error by forcing UTF-8 encoding.

## Guidance to review
You should be able to replicate the error by removing encording and running spec/requests/vendor_api/post_make_an_offer_spec.rb spec

## Link to Trello card

https://trello.com/c/pglY7nw1/3586-encoding-error-when-posting-an-offer-through-the-api

## Things to check

### DIFF
--- a/spec/requests/vendor_api/post_make_an_offer_spec.rb
+++ b/spec/requests/vendor_api/post_make_an_offer_spec.rb
@@ -65,6 +65,39 @@ RSpec.describe 'Vendor API - POST /api/v1/applications/:application_id/offer', t
     end
   end
 
+  describe 'making an offer with conditions with non UTF-8 characters' do
+    it 'returns the updated application' do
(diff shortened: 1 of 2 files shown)


cwrw (approved):

spec/requests/vendor_api/post_make_an_offer_spec.rb
```diff
@@ -65,6 +65,39 @@
     end
   end
 
+  describe 'making an offer with conditions with non UTF-8 characters' do
```
malcolmbaig, line 68: Should this read 'with non-ASCII characters'?

From the error message, it looks like somewhere along the way the interpreter treats the string as if it were ASCII_8BIT,   calls something like `encode(Encoding::UTF_8)` on it, and then errors because it can't convert the non-ASCII character.

PR #4619, opened 2021-04-28
Make non visible text in work history links visible.

george-bit:
## Context

As part of a design review designers have concluded that his text should be visible.

image

## Guidance to review

@paulrobertlloyd - does this look right?

## Link to Trello card

https://trello.com/c/m84YjtGn/3313-add-accessible-text-to-work-history-break-action-links

## Things to check


paulrobertlloyd:
Ah, so this is from the old work history design (which we should probably remove and enable the new one by default). The changes made here should be reverted, as this code is correct.

This ticket refers to this component:

Screenshot 2021-04-28 at 15 59 37

The links should remain visible the same, but include hidden text denoting the dates - much like the component that’s been edited here.

PR #4622, opened 2021-04-28
Add Hidden Accessibility text to Restructured Work History view

george-bit:
## Context

As part of a review of the resructured work history feature, designers noticed that the old accessibility text is missing.
This PR adds this back in.

Please check mark up as intended in review app.

## Link to Trello card

https://trello.com/c/m84YjtGn/3313-add-accessible-text-to-work-history-break-action-links

## Things to check

### DIFF
--- a/app/components/candidate_interface/restructured_work_history/gap_component.html.erb
+++ b/app/components/candidate_interface/restructured_work_history/gap_component.html.erb
@@ -1,6 +1,10 @@
 <%= govuk_inset_text(classes: 'app-inset-text--important govuk-!-margin-top-0 govuk-!-margin-bottom-0') do %>
   You have a break in your work history (<%= format_months_to_years_and_months(@break_period.length) %>)
   <br>
-  <%= govuk_link_to 'Add another job', candidate_interface_new_restructured_work_history_path %> or
-  <%= govuk_link_to 'add a reason for this break', candidate_interface_new_restructured_work_history_break_path(start_date: @break_period.start_date, end_date: @break_period.end_date) %>
+  <%= govuk_link_to candidate_interface_new_restructured_work_history_path do %>
+    Add another job<span class="govuk-visually-hidden"> <%= between_formatted_dates %></span><% end %>
+    or
+  <%= govuk_link_to candidate_interface_new_restructured_work_history_break_path(start_date: @break_period.start_date, end_date: @break_period.end_date) do %>
+    add a reason for this break<span class="govuk-visually-hidden"> <%= between_formatted_dates %></span>
+  <% end %>
 <% end %>
--- a/app/components/candidate_interface/restructured_work_history/gap_component.rb
+++ b/app/components/candidate_interface/restructured_work_history/gap_component.rb
@@ -6,6 +6,10 @@ module CandidateInterface
       def initialize(break_period:)
         @break_period = break_period
       end
+
+      def between_formatted_dates
+        "between #{@break_period.start_date.to_s(:month_and_year)} and #{@break_period.end_date.to_s(:month_and_year)}"
+      end
     end
   end
 end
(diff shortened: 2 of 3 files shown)


app/components/candidate_interface/restructured_work_history/gap_component.html.erb
```diff
@@ -1,6 +1,10 @@
 <%= govuk_inset_text(classes: 'app-inset-text--important govuk-!-margin-top-0 govuk-!-margin-bottom-0') do %>
   You have a break in your work history (<%= format_months_to_years_and_months(@break_period.length) %>)
   <br>
-  <%= govuk_link_to 'Add another job', candidate_interface_new_restructured_work_history_path %> or
-  <%= govuk_link_to 'add a reason for this break', candidate_interface_new_restructured_work_history_break_path(start_date: @break_period.start_date, end_date: @break_period.end_date) %>
+  <%= govuk_link_to candidate_interface_new_restructured_work_history_path do %>
+    Add another job <span class="govuk-visually-hidden"><%= between_formatted_dates %></span>
```
paulrobertlloyd, line 5: This should hopefully fix this whitespace issue with the underline:

Screenshot 2021-04-28 at 18 55 32

```suggestion
    Add another job<span class="govuk-visually-hidden"> <%= between_formatted_dates %></span>
```

app/components/candidate_interface/restructured_work_history/gap_component.html.erb
```diff
@@ -1,6 +1,10 @@
...
+  <% end %> or
+  <%= govuk_link_to candidate_interface_new_restructured_work_history_break_path(start_date: @break_period.start_date, end_date: @break_period.end_date) do %>
+    add a reason for this break <span class="govuk-visually-hidden"><%= between_formatted_dates %></span>
```
paulrobertlloyd, line 8: Same:

```suggestion
    add a reason for this break<span class="govuk-visually-hidden"> <%= between_formatted_dates %></span>
```

paulrobertlloyd (changes requested):

app/components/candidate_interface/restructured_work_history/gap_component.html.erb
george-bit: Okay so a bit of a weird one... i made the changes you suggested and the whitespece remained there 🤔 

It seems that it's being rendered after the span but only for the first link.

After playing around for a bit I managed to get rid of it by pulling the <% end %> up onto the same line. 

The front end looks right but the html.erb markup is a bit ugly, is this okay? 

Happy to go with a different solution if you can think of one :)

app/components/candidate_interface/restructured_work_history/gap_component.html.erb
```diff
@@ -1,6 +1,10 @@
 <%= govuk_inset_text(classes: 'app-inset-text--important govuk-!-margin-top-0 govuk-!-margin-bottom-0') do %>
   You have a break in your work history (<%= format_months_to_years_and_months(@break_period.length) %>)
   <br>
-  <%= govuk_link_to 'Add another job', candidate_interface_new_restructured_work_history_path %> or
-  <%= govuk_link_to 'add a reason for this break', candidate_interface_new_restructured_work_history_break_path(start_date: @break_period.start_date, end_date: @break_period.end_date) %>
+  <%= govuk_link_to candidate_interface_new_restructured_work_history_path do %>
+    Add another job<span class="govuk-visually-hidden"><%= between_formatted_dates %></span><% end %>
```
paulrobertlloyd, line 5: Need to ensure there’s a space between words in the sentence when read out loud.

```suggestion
    Add another job<span class="govuk-visually-hidden"> <%= between_formatted_dates %></span><% end %>
```

paulrobertlloyd (changes requested):
Apart from one nit, looks good to me.

paulrobertlloyd (approved):

app/components/candidate_interface/restructured_work_history/gap_component.html.erb
george-bit: Sorry about that, it didn't make any difference on the front end so missed it! Should be g2g now 👍

paulrobertlloyd: I think it may have done actually… need to view raw page source and not inspect the HTML - no space between the words.

PR #4623, opened 2021-04-28
Remove application form '_completed' field defaults

TobyRet:
## Context

Following on from [Use radio buttons to complete individual sections instead of checkboxes](https://trello.com/c/MgPkClPi/3303-use-radio-buttons-to-complete-individual-sections-instead-of-checkboxes) we need to remove the database defaults for all `_completed` columns in the `application_forms` table

Once this is done, form validation will kick and the candidate will need to select an answer to _'Have you completed this section?'_

- Add migration to remove null constraint on `_completed` columns in `application_forms` table
- Add migration to remove default (`false`) on `_completed` columns in `application_forms` table
- Update controllers to assign `nil` to `_completed` fields instead of false, in order to reset the field so that the candidate will be forced to answer the question 'Have you completed this section?'

select_option

## Guidance to review

- Code: recommend reviewing by commit. Note that backfilling is not required here as these changes only apply to new applications moving forwards. I think it is safe to deploy the migrations and code changes together without affecting users, but let me know if you can think of a reason why that may not be the case.
- Review app - start with a blank application. You should see that the completed section no longer defaults to false. Validation should kick in if you do not provide an answer. In addition, when a job, volunteering role or qualification is created, updated or destroyed, the completed section should 'reset' and the candidate should be forced to submit their answer again.

## Link to Trello card

https://trello.com/c/Gd3jwIMQ/3332-remove-database-defaults-for-completed-columns

## Things to check


paulrobertlloyd:
Just looked at this on one section, but assuming the same on all… error summaries should appear before the `h1`.

Screenshot 2021-04-29 at 14 11 14

TobyRet:
> Just looked at this on one section, but assuming the same on all… error summaries should appear before the `h1`.

😭

malcolmbaig:
Tested a few sections, works as expected and code LGTM 👍🏼 . Ready to ship after @paulrobertlloyd 's comment addressed.

Would be a good idea to squash all `false -> nil` commits into a single one with a generalised commit message to keep the history tidy.

TobyRet:
> Tested a few sections, works as expected and code LGTM 👍🏼 . Ready to ship after @paulrobertlloyd 's comment addressed.
> 
> Would be a good idea (just prior to merge) to squash all `false -> nil` commits into a single one with a generalised commit message to keep the history tidy.

@malcolmbaig - The CompleteSectionComponent takes the review component as one of it's parameters. Given that we will now need to pass in the page title, I'm wondering if the name of this component needs to change. What do you think?

```
<%= render(CandidateInterface::CompleteSectionComponent.new(
  section_complete_form: @section_complete_form,
  page_title: t('page_titles.personal_information'),
  path: candidate_interface_personal_details_complete_path,
  request_method: :patch,
  review_component: SummaryCardComponent.new(rows: @personal_details_review.rows),
)) %>
```

paulrobertlloyd:
@TobyRet Would inheriting a layout be a better option, rather than use a component?

TobyRet:
> @TobyRet Would inheriting a layout be a better option, rather than use a component?

@paulrobertlloyd - I think the component is a now not an option because the pages that the component sits on may have any number of things being rendered before it (e.g 'add another xxx' buttons, rejection reasons)

Example - https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/views/candidate_interface/personal_statement/show.html.erb#L8

Would layout inheritance be suitable here or should we render the complete radio buttons markup inline on each of these pages?

frankieroberto:
@TobyRet another small bug: when editing the details for a section, the "completed" flag reverts back to unspecified, rather than retaining its existing completed / not yet completed status. I think it should retain its previous state.

Looks like this is caused by
```ruby
  current_application.update!(contact_details_completed: nil)
```
in the controllers?

TobyRet:
> @TobyRet another small bug: when editing the details for a section, the "completed" flag reverts back to unspecified, rather than retaining its existing completed / not yet completed status. I think it should retain its previous state.
> 
> Looks like this is caused by
> 
> ```ruby
>   current_application.update!(contact_details_completed: nil)
> ```
> 
> in the controllers?

Hey @frankieroberto - yes I thought that was the intention, but perhaps I assumed incorrectly. My understanding was that if a candidate edits a section they are forced to indicate whether they have completed that section - these changes are all in this commit - https://github.com/DFE-Digital/apply-for-teacher-training/pull/4623/commits/42a8db9ef08c6dc13f266da56b6e488e666d15c6 - can drop it if this was done in error. 

What say you @paulrobertlloyd?

app/components/candidate_interface/complete_section_component.html.erb
```diff
@@ -1,20 +1,8 @@
...
-      <% if hint_text %>
-        <div class="govuk-!-width-two-thirds">
-          <%= hint_text %>
-        </div>
-      <% end %>
-  <% end %>
-
-  <%= f.govuk_submit t('continue') %>
-<% end %>
+<div class="govuk-!-width-two-thirds">
```
TobyRet, line 1: Passing the custom styling for the hint as a Proc now caused a rendering issue and I have no idea why 😱  For now I have reverted back to wrapping the radio buttons in a div.

hnit_proc_madness

frankieroberto:
@TobyRet ha, my understanding/assumption was that it'd only ask people to select the radio once, on initial completion for a blank form. After that, it'd always retain the previous answer and all a candidate could do is flip it between completed or uncompleted.

When applying again, I think it also makes sense that all of the sections would retain the previous "completed" status, except for "review" ones which would be uncompleted/need review (rather than unspecified) – BUT with the exception of "course choice" which would revert back to unspecified rather than uncompleted?

Keen to hear what @paulrobertlloyd’s assumptions were though!

app/components/candidate_interface/complete_section_component.html.erb
peteryates: Hmm, not sure. What happens if you move the hint text to a helper and call it like this? If it's the same I'll take a closer look.

```erb
  <%= form.govuk_collection_radio_buttons :completed,
      [[true, complete_or_reviewed_radio_button_label], [false, t('application_form.incomplete_radio')]],
      :first,
      :last,
      legend: { text: t('application_form.completed_question'), size: 'm' },
      hint: -> { some_hint_text_helper } %>
```

paulrobertlloyd:
@TobyRet Hmmm, not sure I have a strong opinion – maybe it would get annoying if you were to make changes and have to keep checking if the section is complete or not. Happy for the chosen option to persist, as @frankieroberto suggests.

As to what happens when you apply again, given it’s a different question (or at least is presented as one), I agree with @frankieroberto that it should revert to being unspecified as you would have answered that specific question before.

Does that make sense?

frankieroberto:
> As to what happens when you apply again, given it’s a different question (or at least is presented as one), I agree with @frankieroberto that it should revert to being unspecified as you would have answered that specific question before.

Ha, I thought I'd suggested the opposite, that the "review" sections would display this, rather than having neither selected:

> **Have you reviewed this section?**
> ( ) Yes, I have reviewed this section
> (*) No, I'll come back to it later

– but I don't feel too strongly on this.

app/components/candidate_interface/complete_section_component.html.erb
TobyRet: ### Does not work

```

```

### Almost works

```
 <%= form.govuk_collection_radio_buttons :completed,
      [[true, complete_or_reviewed_radio_button_label], [false, t('application_form.incomplete_radio')]],
      :first,
      :last,
      legend: { text: t('application_form.completed_question'), size: 'm' },
      hint: -> do %>
        <%= "<div class=\"govuk-!-width-two-thirds govuk-hint\">#{hint_text}</div>".html_safe %>
  <% end %>
```

almost_works


> > As to what happens when you apply again, given it’s a different question (or at least is presented as one), I agree with @frankieroberto that it should revert to being unspecified as you would have answered that specific question before.
> 
> Ha, I thought I'd suggested the opposite, that the "review" sections would display this, rather than having neither selected:
> 
> > **Have you reviewed this section?**
> > ( ) Yes, I have reviewed this section
> > (*) No, I'll come back to it later
> 
> – but I don't feel too strongly on this.

@paulrobertlloyd @frankieroberto let's re-group on Tuesday to discuss and confirm requirements 👍

TobyRet:
> @TobyRet Hmmm, not sure I have a strong opinion – maybe it would get annoying if you were to make changes and have to keep checking if the section is complete or not. Happy for the chosen option to persist, as @frankieroberto suggests.
> 
> As to what happens when you apply again, given it’s a different question (or at least is presented as one), I agree with @frankieroberto that it should revert to being unspecified as you would have answered that specific question before.
> 
> Does that make sense?

@paulrobertlloyd @frankieroberto 

Previously, for some sections, such as degrees and work history we were not always displaying the chosen answer, [and instead, updating the chosen answer to `false` (aka "I'll complete later")](https://github.com/DFE-Digital/apply-for-teacher-training/pull/4623/commits/42a8db9ef08c6dc13f266da56b6e488e666d15c6 ) whenever a degree or work history entry was updated. Are you saying you want to revert back to this behaviour remove these updates completely (displaying the candidate's chosen answer). 

Shall we have a chat about this post-standup?

TobyRet:
> > @TobyRet Hmmm, not sure I have a strong opinion – maybe it would get annoying if you were to make changes and have to keep checking if the section is complete or not. Happy for the chosen option to persist, as @frankieroberto suggests.
> > As to what happens when you apply again, given it’s a different question (or at least is presented as one), I agree with @frankieroberto that it should revert to being unspecified as you would have answered that specific question before.
> > Does that make sense?
> 
> @paulrobertlloyd @frankieroberto
> 
> Previously, for some sections, such as degrees and work history we were not always displaying the chosen answer, [and instead, updating the chosen answer to `false` (aka "I'll complete later")](https://github.com/DFE-Digital/apply-for-teacher-training/pull/4623/commits/42a8db9ef08c6dc13f266da56b6e488e666d15c6) whenever a degree or work history entry was updated. Are you saying you want to revert back to this behaviour remove these updates completely (displaying the candidate's chosen answer).
> 
> Shall we have a chat about this post-standup?

As discussed with @paulrobertlloyd I dropped [this commit ](https://github.com/DFE-Digital/apply-for-teacher-training/pull/4623/commits/42a8db9ef08c6dc13f266da56b6e488e666d15c6) and reverted back to the original behaviour (i.e updating some section completes to `false`)

paulrobertlloyd (approved):

app/components/candidate_interface/complete_section_component.html.erb
peteryates: Odd, locally this worked for me. Without the `safe_join` it outputs an empty div, maybe try `text.html_safe`?

### `_form.html.erb`

```erb
  <%= form_with(model: order) do |form| %>
  <%= form.govuk_error_summary %>

  <%= form.govuk_radio_buttons_fieldset(
    :extra_topping_id,
    legend: { text: "Select your toppings" },
    hint: -> { topping_hint }) do
  %>

    <%= form.govuk_radio_button(:extra_topping_id, "ham", label: { text: "Ham" }) %>
    <%= form.govuk_radio_button(:extra_topping_id, "pineapple", label: { text: "Pineapple 🍍" }) %>
    <%= form.govuk_radio_button(:extra_topping_id, "sweetcorn", label: { text: "Sweetcorn" }) %>
    <%= form.govuk_radio_button(:extra_topping_id, "caviar", label: { text: "Caviar" }) %>
    <%= form.govuk_radio_button(:extra_topping_id, "saffron", label: { text: "Saffron" }) %>
  <% end %>

  <%= form.govuk_submit("Place order") %>
<% end %>                                                                                     
```

### `orders_helper.rb`

```ruby
module OrdersHelper
  def topping_hint
    tag.div do
      safe_join(
        [
          tag.p("You will be charged extra for"),
          tag.ul do
            safe_join(
              [
                tag.li("Caviar"),
                tag.li("Saffron"),
              ],
            )
          end,
        ],
      )
    end
  end
end
```

### Result

Screenshot from 2021-05-04 15-25-35

PR #4628, opened 2021-04-29
Trim trail/leading whitespace in autosuggest inputs

stevehook:
## Context

If a candidate enters spaces before or after an otherwise matching string into the degree type auto-suggest the potential matches are filtered out of the auto-suggestion list. We want to ignore trailing and leading whitespace for auto-suggest matching purposes.

- [x] Add a `stripWhitespace` option to our `initAutosuggest` wrapper that injects a custom `source` function that strips off any trailing and leading whitespace before finding a match.
- [x] Add a test to exercise this behaviour.

## Guidance to review

- Does adding a `stripWhitespace` option make sense here?
- Can we extend this implementation to support synonym mappings? (another imminent requirement for auto-suggest)
- Is it sensible to make `stripWhitespace` on by default?
- How can I test the behaviour?

I wanted to write a test that exercises the autosuggest behaviour like this:
```
  it('should suggest matching options', () => {
    const autosuggest = document.querySelector('#autosuggest-containerId')
    const input = document.querySelector('#autosuggest-containerId input')
    const keyVal = 65
    input.value = ' A'
    input.dispatchEvent(new KeyboardEvent(
      'keydown',
      {
        keyCode: keyVal,
        which: keyVal,
        charCode: keyVal,
      })
    )
    # TODO: Assert that the matching options are rendered
  })
```
It doesn't work because the event isn't triggering the desired behaviour in Jest/JSDOM. I ended up writing unit tests to cover the filtering but would really like to know how to make this kind of test work.

## Link to Trello card

https://trello.com/c/f2yaj7CF/3318-strip-away-white-space-degree-autofill

## Things to check

### DIFF
--- a/app/frontend/packs/autosuggests/init-autosuggest.js
+++ b/app/frontend/packs/autosuggests/init-autosuggest.js
@@ -35,8 +35,8 @@ export const initAutosuggest = ({ inputIds, containerId, templates = {}, styles 
   }
 }
 
-initAutosuggest.stripWhitespaceFilter = (_source) => {
-  const source = _source
+initAutosuggest.stripWhitespaceFilter = (s) => {
+  const source = s
   return (query, populateResults) => {
     const matches = source.filter(r => r.toLowerCase().indexOf(query.toLowerCase().trim()) !== -1)
     populateResults(matches)
--- a/app/frontend/packs/autosuggests/init-autosuggest.js
+++ b/app/frontend/packs/autosuggests/init-autosuggest.js
@@ -1,6 +1,6 @@
 import { accessibleAutosuggestFromSource } from './helpers'
 
-export const initAutosuggest = ({ inputIds, containerId, templates = {}, styles = () => {} }) => {
+export const initAutosuggest = ({ inputIds, containerId, templates = {}, styles = () => {}, stripWhitespace = true }) => {
   try {
     inputIds.forEach(inputId => {
       const input = document.getElementById(inputId)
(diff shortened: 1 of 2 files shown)


TobyRet (commented):
Hi @stevehook - I ran this and it looks good - works as expected.

Regarding the spec I quickly tried to get the `dispatchEvent` version to work and ran into the same issue. I'm wondering if it's worth trying InputEvent as per the guidance here (https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent) or having a look at the autosuggest library specs for inspiration?

app/frontend/packs/autosuggests/init-autosuggest.js
```diff
@@ -26,3 +34,11 @@ export const initAutosuggest = ({ inputIds, containerId, templates = {}, styles
     console.error(`Could not enhance ${containerId}:`, err)
   }
 }
+
+initAutosuggest.stripWhitespaceFilter = (_source) => {
```
TobyRet, line 38: Should `_source` have a leading underscore given that it is not an ignored param?

stevehook:
> Hi @stevehook - I ran this and it looks good - works as expected.
> 
> Regarding the spec I quickly tried to get the `dispatchEvent` version to work and ran into the same issue. I'm wondering if it's worth trying InputEvent as per the guidance here (https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent) or having a look at the autosuggest library specs for inspiration?

I had a go with `InputEvent` but that doesn't seem to work either. I may dispatching the event to the wrong DOM element. Either that or JSDOM just isn't processing the even the way I expected. Will have another look when I get chance...

The accessible-autocomplete library seems to just test this sort of thing at the React component level (that's what it looks like to me anyway) so it's calling an event handler directly.

app/frontend/packs/autosuggests/init-autosuggest.js
stevehook: Good point, I think I did that to avoid `source = source` below - this method only serves to create a closure for `source` so that it's accessible in the callback below. There may well be a better way to do this...

davidgisbey (approved):
i tested this locally yesterday and it was working as expected. Was just waiting to see if anyone came up with anything for testing this. Sorry I'm not much use when it comes to testing JS.

PR #4631, opened 2021-04-29
add uuid field to sync api method

richardpattinson:
## Context

In order to have the sync Api method add UUID fields to the existing items in the database, it needs to be added to the relevant method. This corresponds to point 2 of the attached Trello card. 

## Link to Trello card

https://trello.com/c/MFIi8aj1/3628-ensure-courses-we-get-from-ttapi-are-unique

### DIFF
--- a/spec/examples/teacher_training_api/course_list_response.json
+++ b/spec/examples/teacher_training_api/course_list_response.json
@@ -17,6 +17,7 @@
         ],
         "changed_at": "2019-06-13T10:44:31Z",
         "code": "3GTY",
+        "uuid": "906c6f3c-b2d6-46e1-8bf7-3fbd13d3ea06",
         "course_length": "OneYear",
         "created_at": "2019-06-13T10:44:31Z",
         "fee_details": "For those wishing to top up their qualification to the full PGCE, a further £1800 will be payable.",
--- a/app/services/teacher_training_public_api/sync_courses.rb
+++ b/app/services/teacher_training_public_api/sync_courses.rb
@@ -52,6 +52,7 @@ module TeacherTrainingPublicAPI
     end
 
     def assign_course_attributes(course, course_from_api, recruitment_cycle_year)
+      course.uuid = course_from_api.uuid
       course.name = course_from_api.name
       course.level = course_from_api.level
       course.study_mode = study_mode(course_from_api)
--- a/spec/examples/teacher_training_api/course_single_response.json
+++ b/spec/examples/teacher_training_api/course_single_response.json
@@ -16,6 +16,7 @@
       ],
       "changed_at": "2019-06-13T10:44:31Z",
       "code": "3GTY",
+      "uuid": "906c6f3c-b2d6-46e1-8bf7-3fbd13d3ea06",
       "course_length": "OneYear",
       "created_at": "2019-06-13T10:44:31Z",
       "fee_details": "For those wishing to top up their qualification to the full PGCE, a further £1800 will be payable.",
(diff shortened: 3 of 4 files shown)


spec/support/test_helpers/teacher_training_public_api_helper.rb
```diff
@@ -27,20 +27,20 @@ def stub_teacher_training_api_provider(provider_code:, recruitment_cycle_year: R
...
-    response_body = build_response_body('course_single_response.json', specified_attributes.merge(code: course_code))
+  def stub_teacher_training_api_course(recruitment_cycle_year: RecruitmentCycle.current_year, provider_code:, course_code:, specified_attributes: [], course_uuid: SecureRandom.uuid)
+    response_body = build_response_body('course_single_response.json', specified_attributes.merge(code: course_code, uuid: course_uuid))
     stub_teacher_training_single_api_request("#{ENV.fetch('TEACHER_TRAINING_API_BASE_URL')}recruitment_cycles/#{recruitment_cycle_year}/providers/#{provider_code}/courses/#{course_code}", response_body)
   end
 
-  def stub_teacher_training_api_courses(recruitment_cycle_year: RecruitmentCycle.current_year, provider_code:, specified_attributes: [])
-    response_body = build_response_body('course_list_response.json', specified_attributes)
+  def stub_teacher_training_api_courses(recruitment_cycle_year: RecruitmentCycle.current_year, provider_code:, specified_attributes: [], course_uuid: SecureRandom.uuid)
+    response_body = build_response_body('course_list_response.json', specified_attributes.first.merge(uuid: course_uuid) )
```
richardpattinson, line 43: I need advice on how to restructure this method. It will break if we add multiple courses as part of a single stub.

spec/support/test_helpers/teacher_training_public_api_helper.rb
```diff
@@ -27,20 +27,20 @@ def stub_teacher_training_api_provider(provider_code:, recruitment_cycle_year: R
     )
   end
 
-  def stub_teacher_training_api_course_with_site(recruitment_cycle_year: RecruitmentCycle.current_year, provider_code:, course_code:, site_code:, vacancy_status: 'full_time_vacancies', course_attributes: [], site_attributes: [])
-    course_attributes = course_attributes.any? ? [course_attributes.first.merge(code: course_code)] : [{ code: course_code }]
+  def stub_teacher_training_api_course_with_site(recruitment_cycle_year: RecruitmentCycle.current_year, provider_code:, course_code:, site_code:, vacancy_status: 'full_time_vacancies', course_attributes: [], site_attributes: [], course_uuid: SecureRandom.uuid)
+    course_attributes = course_attributes.any? ? [course_attributes.first.merge(code: course_code, uuid: course_uuid)] : [{ code: course_code, uuid: course_uuid }]
```
richardpattinson, line 31: Would it be better practice to modify the fixtures file rather than amending the stubbing method?

despo: Yes. The fixture should reflect what we receive from the TTAPI so it should be updated.

richardpattinson: Ok great, will implement

spec/examples/teacher_training_api/course_list_response.json
```diff
@@ -17,6 +17,7 @@
         ],
         "changed_at": "2019-06-13T10:44:31Z",
         "code": "3GTY",
+        "uuid": "903c6f3c-b2d6-46e1-8bf7-3fbd13d3ea09",
```
cwrw, line 20: ah actually the course code is equal on both those fixtures so best keep the uuid the same

richardpattinson: Done =)

cwrw (approved):
🙌

PR #4632, opened 2021-04-29
Add DeleteApplication service for data deletion requests

mnacos:
## Context

Candidates who have started applications but decided not to submit their applications have requested the deletion of their data from our service. This is restricted to unsubmitted applications in this PR, but could be extended to service GDPR-style data deletion requests for any application. Once an application has been sent to providers, we may need to take additional steps.

Overwrite ApplicationForm fields, drop most associations, rewrite the audit log.

## Guidance to review

We are erasing the audit log because it contains a lot of personal information. This means the application's history will be empty, except for a comment documenting the deletion event.

image

We are keeping the ApplicationForm object, its unique support reference and any application choices so that we are still able to include it in our stats. Does this make sense?

Have I missed any fields/associations?

## Link to Trello card

Support-driven https://trello.com/c/3QATwvwN

## Things to check

### DIFF
--- a/app/services/delete_application.rb
+++ b/app/services/delete_application.rb
@@ -1,0 +1,92 @@
+class DeleteApplication
+  include ImpersonationAuditHelper
+
+  attr_reader :actor, :application_form, :zendesk_url
+
+  APPLICATION_FORM_FIELDS_TO_REDACT = %i[
+    first_name
+    last_name
+    first_nationality
+    second_nationality
+    english_main_language
+    english_language_details
+    other_language_details
+    date_of_birth
+    further_information
+    phone_number
+    address_line1
+    address_line2
+    address_line3
+    address_line4
+    country
+    postcode
+    disability_disclosure
+    uk_residency_status
+    work_history_explanation
+    becoming_a_teacher
+    subject_knowledge
+    interview_preferences
+    work_history_breaks
+    volunteering_experience
+    equality_and_diversity
+    safeguarding_issues
+    international_address
+    right_to_work_or_study
+    right_to_work_or_study_details
+    third_nationality
+    fourth_nationality
+    fifth_nationality
+    feedback_satisfaction_level
+    feedback_suggestions
+    work_history_status
+  ].freeze
+
+  ASSOCIATIONS_TO_DESTROY = %i[
+    application_work_experiences
+    application_volunteering_experiences
+    application_qualifications
+    application_references
+    application_work_history_breaks
+    application_feedback
+  ].freeze
+
+  def initialize(actor:, application_form:, zendesk_url:)
+    @actor = actor
+    @application_form = application_form
+    @zendesk_url = zendesk_url
+  end
+
+  def call!
+    raise 'Application has been sent to providers' \
+      unless application_form.application_choices.all?(&:unsubmitted?)
+
+    audit(actor) do
+      ActiveRecord::Base.transaction do
+        ASSOCIATIONS_TO_DESTROY.each { |assoc| application_form.send(assoc)&.destroy_all }
+        APPLICATION_FORM_FIELDS_TO_REDACT.each { |attr| application_form.send("#{attr}=", nil) }
+        application_form.save!
+
+        reference = application_form.support_reference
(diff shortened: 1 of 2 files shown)


app/services/delete_application.rb
```diff
@@ -0,0 +1,90 @@
...
+    application_qualifications
+    application_references
+    application_work_history_breaks
+    application_feedback
+  ].freeze
+
+  def initialize(actor:, application_form:, ticket:)
+    @actor = actor
+    @application_form = application_form
+    @ticket = ticket
```
malcolmbaig, line 56: To remove ambiguity, would it make sense for this to be `zendesk_url:` or similar?

app/services/delete_application.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def call!
+    raise 'Application has been sent to providers' \
+      unless application_form.application_choices.all?(&:unsubmitted?)
+
+    audit(actor) do
```
malcolmbaig, line 63: Does the following happen in a transaction? If not, should it?

malcolmbaig (approved):
Looks really good 👍🏼 

Couple of minor comments left inline.

app/services/delete_application.rb
```diff
@@ -0,0 +1,90 @@
...
+    international_address
+    right_to_work_or_study
+    right_to_work_or_study_details
+    third_nationality
+    fourth_nationality
+    fifth_nationality
+    feedback_satisfaction_level
+    feedback_suggestions
+    work_history_status
+  ].freeze
```
despo, line 42: Considering that the application is anonymised, have we considered keeping non sensitive data for statistical reasons? E.g. safeguarding issues, right to work or study, satisfaction level, disability etc?

app/services/delete_application.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  ASSOCIATIONS_TO_DESTROY = %i[
+    application_work_experiences
+    application_volunteering_experiences
+    application_qualifications
+    application_references
+    application_work_history_breaks
+    application_feedback
```
despo, line 50: Wouldn't it be useful to keep the feedback, so we may be able to get a better inside as to why deletion of the application was requested?

app/services/delete_application.rb
despo: ➕

despo:
Left some questions that I think should be confirmed by UR about the fields to be deleted, but after the call is wrapped in a transaction and the questions answered I think it's good to go!

app/services/delete_application.rb
mnacos: Left it vague because sometimes we have a Trello reference without the Zendesk one, but it should really be Zendesk. Will update.

app/services/delete_application.rb
mnacos: It probably should :)

app/services/delete_application.rb
mnacos: The problem with free-text fields is you can't really know what's in them and whether they include information that could be used to identify the candidate. Probably best to err on the side of caution here. Impact on stats should be minimal, as we've only ever received a few deletion requests. We retain their application choices (the courses they've tried to apply to), their support reference and timestamps so should still be able to produce some basic metrics. Things like safeguarding information, disability etc. sound problematic to retain, especially as we store the Zendesk link to the deletion request.

app/services/delete_application.rb
mnacos: Probably less problematic than safeguarding/diversity/disability but taking the cautious approach, as requests tend to say 'delete all my data'. We should still be able to know this from the Zendesk tickets.

mnacos:
Thanks @malcolmbaig and @despo for reviewing. I've made the suggested code changes. I'll just wait for comments on specific fields, otherwise I think this is safe to merge.

despo:
@mnacos we had a discussion with Mike and James after the standup this morning, and as the applications have not been submitted yet their data is not and will not be used anywhere so it's safe to go ahead with merging this PR as there is no value in keeping any information.

duncanjbrown:
Merging as Michael is off

PR #4636, opened 2021-04-29
Add Maintenance page

cwrw:
## Context

To switch to PaaS we need to stop anyone using the service.

Add a maintenance page which will show on both candidate/provider urls. The exceptions are the monitoring and checks urls
Screenshot 2021-05-04 at 11 52 54
 
As for the API the returned request value will be a 503 but with application/json format.

## Guidance to review

Try testing will all cases 

## Link to Trello card

https://trello.com/c/HpKuxGvH/3657-provider-information-banners-and-maintenance-pages-for-service-downtime-due-to-paas-migration

## Things to check

### DIFF
--- a/app/middlewares/service_unavailable_middleware.rb
+++ b/app/middlewares/service_unavailable_middleware.rb
@@ -1,0 +1,44 @@
+class ServiceUnavailableMiddleware
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    @request = Rack::Request.new(env)
+    if FeatureFlag.active?(:service_unavailable_page) && !monitoring_paths?
+      [503, { 'Content-Type' => content_type }, [body]]
+    else
+      @app.call(env)
+    end
+  end
+
+private
+
+  def content_type
+    return 'application/json' if api_path?
+
+    'text/html'
+  end
+
+  def body
+    return 'Service Unavailable' if api_path?
+
+    action_view.render(template: 'errors/service_unavailable', layout: 'layouts/error')
+  end
+
+  def lookup_context
+    @lookup_context ||= ActionView::LookupContext.new(ActionController::Base.view_paths)
+  end
+
+  def action_view
+    @action_view ||= ActionView::Base.with_empty_template_cache.new(lookup_context, {}, ActionController::Base.new)
+  end
+
+  def api_path?
+    @request.path =~ /^\/api\/.*$/
+  end
+
+  def monitoring_paths?
+    @request.path =~ /^\/check$/ || @request.path =~ /^\/integrations\/monitoring\/.*$/
+  end
+end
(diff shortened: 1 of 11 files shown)


cwrw (commented):
@duncanjbrown got this in a rough state but seems to work, needed to add an error layout as well.

app/middlewares/service_unavailable_middleware.rb
```diff
@@ -0,0 +1,16 @@
...
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    if FeatureFlag.active?(:service_unavailable_page)
+      lookup_context = ActionView::LookupContext.new(ActionController::Base.view_paths)
+      view = ActionView::Base.with_empty_template_cache.new(lookup_context, {}, ActionController::Base.new)
+      body = view.render(template: 'errors/service_unavailable', layout: 'layouts/error')
+      [503, {'Content-Type' => 'text/html'}, [body]]
```
vigneshmsft, line 11: is it possible to return 200 statuscode for `/check` and `/integrations/` endpoints?
if not just 200 instead of 503?

cwrw: Good point, will try and exclude those endpoints from the 503

spec/components/previews/service_unavailable_component_preview.rb
```diff
@@ -0,0 +1,5 @@
+class ServiceUnavailableComponentPreview < ViewComponent::Preview
+  def service_unavailable_oage
```
duncanjbrown, line 2:
```suggestion
  def service_unavailable_page
```

vigneshmsft (approved):

despo (approved):
LGTM 👍🏼

config/locales/service_unavailable.yml
```diff
@@ -0,0 +1,7 @@
+en:
+  service_unavailable:
+    sandbox_title: Sorry, the sandbox is unavailable
+    sandbox_downtime: You will be able to use the sandbox from 7pm on Tuesday 4 May.
+    title: Sorry, this service is unavailable
+    downtime: You will be able to use this service from 9am on Thursday 6 May.
+    body: If you reached this page after submitting information then it has not been saved. You will need to enter it again when the service is available.
```
despo, line 7:
```suggestion
    body: If you reached this page after submitting information then it has not been saved, you will need to enter it again when the service is available.
```

spec/components/service_unavailable_component_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'rails_helper'
+
+RSpec.describe ServiceUnavailableComponent do
+  subject(:result) { render_inline(ServiceUnavailableComponent.new) }
+
+  it 'renders the page title' do
+    expect(result.text).to include('Sorry, the service is unavailable')
```
despo, line 7:
```suggestion
    expect(result.text).to include('Sorry, this service is unavailable')
```

PR #4638, opened 2021-05-04
Service information banner

duncanjbrown:
## Context

We need to notify Providers and Candidates of coming downtime

Turn the `ProviderInterface::NotificationBannerComponent` into a `ServiceInformationBanner`. We can set special text for the sandbox.

### For providers

The banner appears on the landing page

Screenshot 2021-05-04 at 09 47 47

And on the applications index

Screenshot 2021-05-04 at 09 47 42

### For candidates

The banner appears on the sign in page

Screenshot 2021-05-04 at 09 46 53

On applications in progress

Screenshot 2021-05-04 at 09 47 17

And on the dashboard

Screenshot 2021-05-04 at 09 47 31

## Guidance to review

Is the banner in the right places? Is the wording correct?

## Things to check


app/views/provider_interface/start_page/show.html.erb
```diff
@@ -24,10 +24,13 @@
   </div>
 </div>
 
-<%= render ProviderInterface::InformationBannerComponent.new %>
-
 <div class="govuk-width-container">
   <section class="app-section">
+    <div class="govuk-grid-row">
```
cwrw: do we need those wrappers as they exist in the component, the original looked like this: https://github.com/DFE-Digital/apply-for-teacher-training/pull/4635/files#diff-b8517c38a7d309ffcbdf847eb27a7451e081f4b1270d800954e334b4c89b640dL28

cwrw: nvm removed in next commit 👍

cwrw (approved):

PR #4640, opened 2021-05-04
Make it easier for providers to spot gaps in work history in manage

despo:

Display voluntary work as part of work history in manage.

As the existing components are reused by the candidate side I decided to introduce a new WorkHistoryAndUnpaidExperience component to avoid accidentally changing the candidate view.

## Guidance to review

- Enable restructred work history flag
- Navigate to a candidate application

## Link to Trello card

https://trello.com/c/Gg9XKGw2/3573-make-it-easier-for-providers-to-spot-gaps-in-work-history-in-manage

## Things to check

### DIFF
--- a/app/components/work_history_and_unpaid_experience_item_component.html.erb
+++ b/app/components/work_history_and_unpaid_experience_item_component.html.erb
@@ -1,0 +1,39 @@
+<section>
+  <% if break? %>
+    <div class="govuk-inset-text">
+  <% end %>
+
+  <h3 class="govuk-heading-s govuk-!-margin-bottom-0">
+    <%= title %>
+  </h3>
+  <p class="govuk-!-margin-bottom-0 govuk-body">
+    <%= organisation %>
+  </p>
+  <p class="govuk-caption-m govuk-!-font-size-16 govuk-!-margin-top-0">
+    <%= dates %>
+  </p>
+  <p class="govuk-body">
+    <%= details %>
+    <% if relevant_skills? %>
+      <p class="govuk-body govuk-!-font-size-16">
+        <svg class="app-icon govuk-!-margin-right-1" xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 200 200" focusable="false" fill="currentColor" aria-hidden="true">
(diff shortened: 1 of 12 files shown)


app/components/work_history_and_unpaid_experience_item_component.html.erb
```diff
@@ -0,0 +1,39 @@
...
+      <p class="govuk-body govuk-!-font-size-16">
+        <svg class="app-icon govuk-!-margin-right-1" xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 200 200" focusable="false" fill="currentColor" aria-hidden="true">
+          <path d="M100 200a100 100 0 1 1 0-200 100 100 0 0 1 0 200zm-60-85l40 40 80-80-20-20-60 60-20-20-20 20z"></path>
+        </svg>
+        This role used skills relevant to teaching
+      </p>
+    <% elsif working_with_children? %>
+      <p class="govuk-body govuk-!-font-size-16">
+        <svg class="app-icon govuk-!-margin-right-1" xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 200 200" focusable="false" fill="currentColor" aria-hidden="true">
+          <path d="M100 200a100 100 0 1 1 0-200 100 100 0 0 1 0 200zm-60-85l40 40 80-80-20-20-60 60-20-20-20 20z"></path>
```
richardpattinson, line 27: What is this path? Why is it so unreadable?

despo: It's a scalable vector path so I don't think it can be made more readable. I reused the code from the original template from https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/components/work_history_item_component.html.erb as I didn't want to modify the original one and condition using the feature flag, in case it affects any other areas it's used (as mentioned [here](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/components/work_history_component.rb#L1))

cwrw (commented):
looks good and all works as expected. Only point left to address is: The work history section of the application form is moved above qualifications, otherwise 👍

richardpattinson (approved):
Looks good to me despot =)

PR #4649, opened 2021-05-05
Don't show withdrawal reasons if choice is not 'offer_withdrawn'

malcolmbaig:
## Context
In a couple of places in the app, we have components which summarise a
candidate's course choices. In the scenario where an 'offer_withdrawn'
application choice has another offer made against it, we still show the
withdrawal reasons in the summary.

Hide the reasons in this scenario by checking the state of the
application choice.

## Guidance to review
- Can be reviewed manually by starting with an application choice that has an offer against it.
  - Withdraw the offer as a provider
  - Review the app dashboard for this application and note that the withdrawal reasons are rendered.
  - Use the console (or the vendor API?) to make the state of the application choice `offer`.
  - Review the app dashboard for this application and note that we no longer display the withdrawal reason.
- To answer a question posed in the card - rejected choices already make a similar state check before displaying.

## Link to Trello card
https://trello.com/c/GtPA69fk

## Things to check

### DIFF
--- a/app/components/candidate_interface/course_choices_review_component.rb
+++ b/app/components/candidate_interface/course_choices_review_component.rb
@@ -186,6 +186,8 @@ module CandidateInterface
     end
 
     def offer_withdrawal_reason_row(application_choice)
+      return nil unless application_choice.offer_withdrawn?
+
       if application_choice.offer_withdrawal_reason.present?
         {
           key: 'Reason for offer withdrawal',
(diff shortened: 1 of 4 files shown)


app/components/candidate_interface/course_choices_review_component.rb
```diff
@@ -186,6 +186,8 @@ def status_row(application_choice)
     end
 
     def offer_withdrawal_reason_row(application_choice)
+      return nil unless application_choice.offer_withdrawn?
```
stevehook, line 189: Maybe simpler if this guard clause was combined with the `if` below it? e.g.
```
      return nil unless application_choice.offer_withdrawn? && application_choice.offer_withdrawal_reason.present?
```
Not sure.

stevehook (approved):
Tested review app and all works nicely.

PR #4650, opened 2021-05-05
Do not update `_section_completed' fields to false when adding to, deleting from or updating sections

TobyRet:
## Context

This follows on from switching from checkboxes to radio buttons when asking a candidate if they have completed a section.

When a candidate changes sections of the application form, either through creation, editing or deleting (e.g. course choices,  volunteering, degrees, work experience) we are updating the corresponding `_section_completed` field to `false`. 

We do not want to retain this behaviour anymore, and rather simply refer to what ever choice they saved previously.

- remove code that updates `_section_completed` to `false`
- Fix specs

## Guidance to review

In the prototype, start with a blank application. When you add to a section, delete an item from a section or edit, you should see that your submitted answer to 'Have you completed this section?' should always be displayed when you go to review that section

## Link to Trello card

https://trello.com/c/dxkMUjHV/3358-do-not-update-sectioncompleted-fields-to-false-when-updating-application-sections

## Things to check


TobyRet:
Tested out and looks ok apart from one issue re: course choices

It's now possible to get in a weird state. If you mark course choices as complete and subsequently delete course choices, when you go back to the application page, the link to 'choose your courses' no longer points to `/candidate/application/courses` and instead points to `/candidate/application/courses/review`. On this page the `add a course` button and no longer displays because [application choices are not present](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/views/candidate_interface/application_choices/review.html.erb#L17) and the section complete component no longer display because application choices are not > 1

One option would be to retain the original behaviour for this section only, or agree the behaviour / apply fixes

db/migrate/20210506101257_remove_more_defaults_from_application_forms_table.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveMoreDefaultsFromApplicationFormsTable < ActiveRecord::Migration[6.1]
```
TobyRet, line 1: Note there is a separate PR for this migration - commit was cherry picked - https://github.com/DFE-Digital/apply-for-teacher-training/pull/4671

paulrobertlloyd:
Few things…

* Change `No, I'll come back to it later` to `No, I’ll come back to it later` (use smart quote)
* The completed state for **other qualifications** persists even if you’ve deleted the last qualification (should probably be reset to `nil`)
* The completed state for other **work history** persists even if you’ve deleted the last qualification (should probably be reset to `nil`)
* The completed state for **unpaid experience** persists even if you’ve deleted the last role (should probably be reset to `nil`). Additionally, the answer for ‘Do you have any relevant unpaid experience?’ defaults to ‘No’ when in fact this should also be reset to `nil` and (you’ll be asked this question again)
* The completed state for **degree** persists even if you’ve deleted the last degree (should probably be reset to `nil`). Additionally, clicking on the link in the application menu takes you to section review page when it should in fact take you to page asking for first degree.

app/controllers/candidate_interface/volunteering/destroy_controller.rb
```diff
@@ -12,10 +12,9 @@ def destroy
         .destroy!
 
       if current_application.application_volunteering_experiences.blank?
-        current_application.update!(volunteering_completed: false, volunteering_experience: false)
+        current_application.update!(volunteering_experience: false)
```
davidgisbey, line 15: Should this come out too?

TobyRet: Yep - this will address the issue @paulrobertlloyd mentioned above re: unpaid experience

davidgisbey:
I might be missing something, but doesn't this mean someone could add a job, complete the section, delete the job so they have a blank work history and the section would remain completed which would allow they to submit their application?

TobyRet:
> I might be missing something, but doesn't this mean someone could add a job, complete the section, delete the job so they have a blank work history and the section would remain completed which would allow they to submit their application?

Hey @davidgisbey 

Yes that is correct, we will need to fix the logic on the review page. You can see in the screenshot that I have marked degrees section as complete but have gone back and deleted my degree. @paulrobertlloyd what copy should go here, given the candidate has marked the section as complete in this case?


review_application

TobyRet:
As discussed @paulrobertlloyd @davidgisbey - we will now automatically mark sections as incomplete when a candidate has deleted all entries (for jobs, quals etc)

paulrobertlloyd (approved):
Perfect!

PR #4651, opened 2021-05-05
Remove duplicate accrediting providers

steventux:
## Context

There are 9 provider relationships in production where the training and ratifying provider have the same name. 
There are 457 courses for these providers which are provided/ratified by the two sides of the seemingly unnecessary relationships. 
This seems to be bad data as a self-ratified course should have a nil accredited provider and the duplication of provider records and relationship records are undesirable.

This data migration finds these relationships and iterates over courses concerned nillifying the accredited_provider thus making them self-ratified. 
The relationship and duplicate provider records are then destroyed.

## Guidance to review

Are my assumptions about self-ratified courses still correct? (I may be working on old info)
What are the side effects of updating these courses and relationships? Perhaps there should also be a check that the ratifying provider is never also a training provider for other courses?

## Link to Trello card

From support query https://trello.com/c/IEHp1BE0/446-united-teaching-national-scitt-codes

## Things to check

### DIFF
--- a/app/services/data_migrations/remove_duplicate_provider.rb
+++ b/app/services/data_migrations/remove_duplicate_provider.rb
@@ -10,10 +10,11 @@ module DataMigrations
         .where('tp.name = rp.name')
 
       if permissions.any?
-        ActiveRecord::Base.transaction do
-          permissions.each do |permission|
+        permissions.each do |permission|
+          ActiveRecord::Base.transaction do
             courses = permission.training_provider.courses.where(accredited_provider: permission.ratifying_provider)
-            next if courses.empty?
+
+            next if permission.ratifying_provider.courses.any?
 
             courses.each do |course|
               course.update!(accredited_provider: nil, audit_comment: 'Deduplicating accredited provider, this course is self-ratified')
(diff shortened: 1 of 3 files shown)


app/services/data_migrations/remove_duplicate_provider.rb
```diff
@@ -0,0 +1,30 @@
...
+        .joins('INNER JOIN providers tp ON training_provider_id = tp.id')
+        .joins('INNER JOIN providers rp ON ratifying_provider_id = rp.id')
+        .where('tp.name = rp.name')
+
+      if permissions.any?
+        permissions.each do |permission|
+          ActiveRecord::Base.transaction do
+            courses = permission.training_provider.courses.where(accredited_provider: permission.ratifying_provider)
+
+            next if permission.ratifying_provider.courses.any?
```
benswannack, line 17: This check can move to the top of the transaction.

Do we also need to check that the provider we are about to delete is not part of other relationships and whether it has any `accredited_courses`? We should check we aren't going to orphan any provider users too.

cwrw:
Just an update after discussing in person and slack: If we remove those providers they will come back on the next sync. The Find and Publish team has added a card on their side to look into those providers and merge them: https://trello.com/c/LANwfOEG/1665-fix-duplicate-providers and slack thread: https://ukgovernmentdfe.slack.com/archives/CN1MCQCHZ/p1620139919238400 until then this work is blocked

steventux:
Closing for now as this migration can't be run until data upstream in the TTAPI is also free of these duplicate providers.

PR #4654, opened 2021-05-05
[1/3] Remove the subject field from notes: stop writing to subject field

despo:

1. Remove field from add note page
2. Remove field from notes show and index page
3. The value of the subject field of existing notes is prefixed at the start of the notes description field (https://ukgovernmentdfe.slack.com/archives/CPH8J9G65/p1617886095337000?thread_ts=1617872257.329800&cid=CPH8J9G65)
4. No update was require in the Notes export to support a note without a subject, as nothing will be rendered when there isn't one at this point

## Link to Trello card
https://trello.com/c/a0IxokKH/3641-1-remove-the-subject-field-from-notes-stop-writing-to-subject-field

## Things to check

### DIFF
--- a/app/controllers/provider_interface/notes_controller.rb
+++ b/app/controllers/provider_interface/notes_controller.rb
@@ -33,9 +33,8 @@ module ProviderInterface
   private
 
     def note_params
-      params.require(:provider_interface_new_note_form).permit(:subject, :message).merge \
-        application_choice: @application_choice,
-        provider_user: current_provider_user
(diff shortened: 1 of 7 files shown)


mnacos (approved):
Looks fine code-wise. I think the notes index looks a little funny with the subject + text, maybe using just the first line for the listing would be better, now you can click through to #show. But this is the design, so 🤷🏼 

image

despo:
@mnacos I agree. I guess design can always revise after this is released as it would probably make sense to only show an extract of the text in the index.

PR #4658, opened 2021-05-05
[2-1/3] Remove the subject field from notes: stop reading from subject field

despo:
## Context
1. Add data migration to prefix existing subject to note message
2. Only read message in notes show, index and export
3. Remove subject from note factory

## Link to Trello card

https://trello.com/c/KtL53Ipv/3642-2-remove-the-subject-field-from-notes-stop-reading-from-subject-field

## Things to check

### DIFF
--- a/app/exports/notes_export.yml
+++ b/app/exports/notes_export.yml
@@ -5,10 +5,6 @@ common_columns:
 - provider_user_id
 
 custom_columns:
-  note_subject:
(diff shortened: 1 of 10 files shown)


richardpattinson (approved):
Not a lot to comment on really, seems like a straightforward deprecation of the subject field.
Looks great to me =)

PR #4660, opened 2021-05-05
Remove Course qualification flip flops from audits table, keep recent ones

mnacos:
## Context

Our audits table on production has about 18 million records, and `Course`-related changes to `qualifications` account for ~9 million of these. There is probably a bug in our sync process there somewhere, but this PR focuses on removing most of these audit records to reclaim space.

Add automatic data migration `DataMigrations::PruneQualificationFlipFlopsFromCourseAudits`, which removes all Course-related events by '(Automated process)' setting and unsetting `qualifications` except the most recent ones for each type of change.

## Guidance to review

Run the spec.

## Link to Trello card

https://trello.com/c/6UMC4zRm

## Things to check

### DIFF
--- a/app/services/data_migrations/prune_qualification_flip_flops_from_course_audits.rb
+++ b/app/services/data_migrations/prune_qualification_flip_flops_from_course_audits.rb
@@ -1,0 +1,33 @@
+module DataMigrations
+  class PruneQualificationFlipFlopsFromCourseAudits
+    TIMESTAMP = 20210505154302
+    MANUAL_RUN = false
+
+    def change
+      sql = <<~REMOVE_DUPLICATE_QUALIFICATION_CHANGES_FROM_COURSE_AUDITS.squish
+        DELETE FROM audits
+        WHERE id IN (
+          SELECT id FROM (
+            SELECT
+              id,
+              ROW_NUMBER() OVER(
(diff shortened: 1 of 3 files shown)


cwrw:
I traced back the bug to January, and found that it has been fixed: https://github.com/DFE-Digital/apply-for-teacher-training/pull/3844

app/services/data_migrations/prune_qualification_flip_flops_from_course_audits.rb
```diff
@@ -0,0 +1,33 @@
+module DataMigrations
+  class PruneQualificationFlipFlopsFromCourseAudits
+    TIMESTAMP = 20210505154302
+    MANUAL_RUN = false
+
+    def change
```
cwrw, line 6: disable timeout

cwrw:
After having a chat, we are thinking of removing *all* course audits as they aren't used, after we start capturing the opened on apply timestamp. We have opened this question to the wider dev team to make sure we aren't missing any usage of those audits, after that we can amend this to remove all Course audits.

mnacos:
Closing, as we've gone with deleting all course audits: https://github.com/DFE-Digital/apply-for-teacher-training/pull/4720

PR #4663, opened 2021-05-05
Add Satisfaction Survey Response rate to the Feature Metric Dashboard

george-bit:
## Context
Support Users currently cannot view the percentage rate of responses givento the post submission satisfaction survey. 

This PR adds new section to the Feature Metrics Dashboard in order to display this info.

image

## Guidance to review

Test in review app. I have left a couple of comments RE: testing.

## Link to Trello card

https://trello.com/c/EYeqCBJb/3335-add-satisfaction-survey-response-rate-to-the-feature-metric-dashboard

## Things to check

### DIFF
--- a/app/models/satisfaction_survey_feature_metrics.rb
+++ b/app/models/satisfaction_survey_feature_metrics.rb
@@ -5,15 +5,8 @@ class SatisfactionSurveyFeatureMetrics
     start_time,
     end_time = Time.zone.now.end_of_day
   )
-    success_count = 0.0
-    fail_count = 0.0
-    application_forms(start_time, end_time).find_each do |application_form|
-      if application_form.feedback_satisfaction_level? || application_form.feedback_suggestions?
-        success_count += 1
-      else
-        fail_count += 1
-      end
-    end
+    success_count = application_forms_with_feedback(start_time, end_time).count.to_f
+    fail_count = application_forms_with_no_feedback(start_time, end_time).count.to_f
     return nil if (fail_count + success_count).zero?
 
     success_count / (fail_count + success_count)
@@ -39,12 +32,24 @@ private
     "#{percentage}%"
   end
 
-  def application_forms(start_time, end_time)
+  def application_forms_with_feedback(start_time, end_time)
     ApplicationForm
       .where(
         recruitment_cycle_year: RecruitmentCycle.current_year,
       )
       .where('application_forms.submitted_at BETWEEN ? AND ?', start_time, end_time)
+      .where.not(feedback_satisfaction_level: '').where.not(feedback_suggestions: '')
+      .where(recruitment_cycle_year: RecruitmentCycle.current_year)
+  end
+
+  def application_forms_with_no_feedback(start_time, end_time)
+    ApplicationForm
+      .where(
+        recruitment_cycle_year: RecruitmentCycle.current_year,
+      )
+      .where('application_forms.submitted_at BETWEEN ? AND ?', start_time, end_time)
+      .where(feedback_satisfaction_level: nil).or(ApplicationForm.where(feedback_satisfaction_level: ''))
+      .where(feedback_suggestions: nil).or(ApplicationForm.where(feedback_suggestions: ''))
       .where(recruitment_cycle_year: RecruitmentCycle.current_year)
   end
 end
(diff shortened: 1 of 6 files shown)


spec/models/satisfaction_survey_feature_metrics_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+        expect(feature_metrics.formatted_response_rate(1.month.ago)).to eq('0%')
+      end
+
+      it 'returns the right percentages when applications with feedback exist' do
+        create_application
+        2.times { create_application_with_feeback }
+        expect(feature_metrics.formatted_response_rate(1.month.ago)).to eq('66.7%')
+      end
+
+      it 'returns the right percentages over a range of dates' do
```
george-bit, line 35: This test block is inspired by some adjacent objects used by the dashboard. The aim is to cover some of the functionality missing from the integration test. It would be more expensive to replicate this in that spec. 

However, I may have missed the point here so open to any thoughts.

app/models/satisfaction_survey_feature_metrics.rb
```diff
@@ -0,0 +1,50 @@
...
+  include ActionView::Helpers::NumberHelper
+
+  def response_rate(
+    start_time,
+    end_time = Time.zone.now.end_of_day
+  )
+    success_count = 0.0
+    fail_count = 0.0
+    application_forms(start_time, end_time).find_each do |application_form|
+      if application_form.feedback_satisfaction_level? || application_form.feedback_suggestions?
```
stevehook, line 11: You could probably implement these conditions in the DB query - e.g. `.where.not(feedback_satisfaction_level: '').where.not(feedback_suggestions: '')`. That way you wouldn't have to load so many model objects into Ruby though you might end up with two queries - one to count the with feedback and one without.

:+1: for using `find_each` versus `each`.

spec/models/satisfaction_survey_feature_metrics_spec.rb
stevehook: This looks good to me.

I think the intention here was just to cover some of the query logic for time intervals. 

Using Timecop is a bit clumsy. I couldn't think of a better solution for some of the other metrics mainly because there were conditions on things in the audit trail that the conditions depended on and faking those out would have been worse. In this case the query condition only depends on `ApplicationForm#submitted_at` so you could explicitly set values for each application form that you create. That might be tidier but it's consistent the way it is so don't mind either way.

stevehook (commented):
:+1: Looks good and works well on the review app - I faked some feedback data and it picked it up nicely.

A couple of inline questions.

app/models/satisfaction_survey_feature_metrics.rb
george-bit: Will have a go at refactoring 👍

george-bit: managed to refactor as suggested, please let me know your thoughts :)

app/models/satisfaction_survey_feature_metrics.rb
```diff
@@ -0,0 +1,54 @@
...
+    "#{percentage}%"
+  end
+
+  def application_forms_with_feedback(start_time, end_time)
+    ApplicationForm
+      .where(
+        recruitment_cycle_year: RecruitmentCycle.current_year,
+      )
+      .where('application_forms.submitted_at BETWEEN ? AND ?', start_time, end_time)
+      .where.not(feedback_satisfaction_level: nil).where.not(feedback_suggestions: nil)
```
stevehook, line 41:
```suggestion
      .where.not(feedback_satisfaction_level: '').where.not(feedback_suggestions: '')
```
There are quite a few records in the prod DB with empty string values for these fields. `.where.not(feedback_suggestions: '')` will filter those out as well as `null` values (because `null` isn't equal to or not equal to any value).

stevehook (commented):
Looks good. One last suggestion inline.

app/models/satisfaction_survey_feature_metrics.rb
george-bit: Ah okay 👍, that's my fault for forgetting about null!

stevehook (approved):

PR #4664, opened 2021-05-06
Allow up to 20 further conditions of offer

benswannack:
## Context
We want to allow providers to give more or fewer than four further conditions of offer to candidates.

Add a shared component for the conditions form

Add "add another" and "remove" buttons to the form

## Guidance to review
This should be able to support the JS enhancements that are forthcoming. (i.e. should be easy to drop some JS in)

All the buttons POST to the same endpoint and we handle the button detection in the action. It doesn't seem like there's a better way to do this. It probably doesn't work exactly like this when we switch on JS though.

We have tried to match the behaviour of the prototype - look out for places that we differ!

Test the back/forward buttons in the browser. Do they do what you would expect?

## Link to Trello card
https://trello.com/c/9MQAO3J2/3612-implement-non-js-add-another-offer-conditions

## Things to check


app/components/provider_interface/conditions_form_component.html.erb
```diff
@@ -0,0 +1,51 @@
...
+                                         standard_conditions_checkboxes,
+                                         :id,
+                                         :name,
+                                         legend: { size: 'm' }) %>
+
+      <%= f.govuk_fieldset(
+        legend: {
+          text: t('.further_conditions'),
+          class: 'govuk-heading-m govuk-!-margin-bottom-1 app-add-another__heading',
+          tabindex: '-1',
```
benswannack, line 21: This makes it focusable, as per the prototype.

I'm also going to add an id for this element and focus it when the remove button is clicked (also as per prototype)

app/controllers/provider_interface/offer/conditions_controller.rb
```diff
@@ -24,31 +16,76 @@ def edit
...
-          @wizard.save_state!
+    private
 
-          redirect_to [:edit, :provider_interface, @application_choice, :offer, @wizard.next_step]
+      def handle_form_input(action:)
+        @wizard = OfferWizard.new(offer_store, attributes_for_wizard)
+
+        if add_another_condition?
+          add_empty_condition
+          redirect_to action: action, anchor: anchor_for_further_condition
```
benswannack, line 29: This follows a POST-redirect-GET pattern so that refreshing the page does not re-submit the form.

It's difficult to get the back button to behave nicely with this (and even with just POSTs), but we've agreed that we can live with this for now & it won't be an issue with the JS implementation

app/controllers/provider_interface/offer/conditions_controller.rb
```diff
@@ -24,31 +16,75 @@ def edit
...
+        if @wizard.valid_for_current_step?
+          @wizard.save_state!
+
+          redirect_to [action, :provider_interface, @application_choice, :offer, @wizard.next_step]
+        else
+          render action
+        end
+      end
+
+      def anchor_for_further_condition
```
cwrw, line 81: can we move this logic outside the controller?

app/controllers/provider_interface/offer/conditions_controller.rb
```diff
@@ -24,31 +16,75 @@ def edit
...
 
         attributes.merge!(further_conditions: further_conditions, current_step: 'conditions')
       end
+
+      def conditions_params
+        params.require(:provider_interface_offer_wizard)
+              .permit(further_conditions: {}, standard_conditions: [])
+      end
+
+      def add_empty_condition
```
cwrw, line 59: this logic feels more suited in the wizard, which should decide what we do next?

cwrw (commented):
looks good, a few comments. Whilst testing locally, when adding extra conditions submitting the form and then going back to amend the conditions I am still seeing the empty conditions, we should make sure to clear the empty conditions, happy to walk through this over a call

app/forms/provider_interface/offer_wizard.rb
```diff
@@ -4,7 +4,7 @@ class OfferWizard
 
     STEPS = { make_offer: %i[select_option conditions check],
               change_offer: %i[select_option providers courses study_modes locations conditions check] }.freeze
-    NUMBER_OF_FURTHER_CONDITIONS = 4
+    MAXIMUM_NUMBER_OF_FURTHER_CONDITIONS = 20
```
benswannack, line 7: Rethinking whether this is true: Is the actual limit just on the number of conditions (standard + further)?

benswannack: Agreed to limit to 18 further conditions in Manage

cwrw (approved):

PR #4666, opened 2021-05-06
Let support users change conditions on accepted applications

stevehook:
## Context

Providers regularly ask support to change the conditions for accepted applications. This should only be done after the candidate has consented to the change and we should link the change (via the audit trail) to a ZenDesk ticket that records the request.

- Show the current conditions for applications in the `offer` and `pending_conditions` statuses:
image
- Add a _Change conditions_ link that leads to a new form for editing conditions. This mirrors the form in the provider interface for adding conditions (before the offer is made) but includes a warning notification and an additional field to prompt for an audit comment
image

- Support more than 4 further conditions (to match upcoming changes in the provider interface). To allow more than 4 conditions we display at least one empty text box at the end of the list of further conditions up to a maximum of 18 (because there is still a limit of 20 on the total number of conditions).
- If the support user removes all the conditions then the offer will be made unconditional and the application choice is moved straight to the `recruited` status. To ensure that this is intentional the user will see an extra confirmation page in this case:
image

## Guidance to review

- The 'design' for the changes to the dashboard and the new change conditions for are up for debate.
- Is the content for these new pages OK?
- Should we be sending email notifications to candidates/providers?
- Can we extract and reuse some of the code that's in the provider interface for editing conditions?

## Link to Trello card

https://trello.com/c/TisFW3VV/3337-enable-conditions-to-be-changed-after-offer-acceptance
## Things to check

### DIFF
--- a/app/components/support_interface/application_choice_component.rb
+++ b/app/components/support_interface/application_choice_component.rb
@@ -107,8 +107,8 @@ module SupportInterface
       {
         key: 'Conditions',
         value: render(SupportInterface::ConditionsComponent.new(conditions: conditions)),
-        action: 'Change conditions',
-        action_path: support_interface_edit_application_choice_conditions_path(application_choice_id: @application_choice.id),
+        action: 'conditions',
+        change_path: support_interface_edit_application_choice_conditions_path(application_choice_id: @application_choice.id),
       }
     end
 
--- a/app/components/support_interface/application_choice_component.rb
+++ b/app/components/support_interface/application_choice_component.rb
@@ -17,6 +17,7 @@ module SupportInterface
         course_candidate_applied_for_row,
         course_offered_by_provider_row,
         course_row,
+        offer_conditions_row,
         rejected_at_or_by_default_at_row,
         rejection_reason_row,
         sent_to_provider_at_row,
@@ -95,6 +96,20 @@ module SupportInterface
       return unless rejection_reasons_text
 
       { key: 'Rejection reason', value: rejection_reasons_text }
+    end
+
+    def offer_conditions_row
+      return unless application_choice.pending_conditions? || application_choice.offer?
+
+      conditions = application_choice.offer&.[]('conditions')
+      return if conditions.blank?
+
+      {
+        key: 'Conditions',
+        value: render(SupportInterface::ConditionsComponent.new(conditions: conditions)),
+        action: 'Change conditions',
+        action_path: support_interface_edit_application_choice_conditions_path(application_choice_id: @application_choice.id),
+      }
     end
 
     def sent_to_provider_at_row
--- a/app/forms/support_interface/conditions_form.rb
+++ b/app/forms/support_interface/conditions_form.rb
@@ -5,14 +5,15 @@ module SupportInterface
 
       attr_accessor :id, :text
 
-      validates :text, length: { maximum: 255, too_long: ->(c, _) { "Condition #{c.id + 1} must be %{count} characters or fewer" } }
+      validates :text, length: { maximum: MakeAnOffer::MAX_CONDITION_LENGTH, too_long: ->(c, _) { "Condition #{c.id + 1} must be %{count} characters or fewer" } }
     end
 
     include ActiveModel::Model
 
     attr_accessor :application_choice, :standard_conditions, :further_conditions, :audit_comment
 
-    NUMBER_OF_FURTHER_CONDITIONS = 4
+    MIN_FURTHER_CONDITIONS = 4
+    MAX_FURTHER_CONDITIONS = MakeAnOffer::MAX_CONDITIONS_COUNT - MakeAnOffer::STANDARD_CONDITIONS.length
 
     validates :application_choice, presence: true
     validates :audit_comment, presence: true
@@ -54,7 +55,13 @@ module SupportInterface
     end
 
     def add_slots_for_new_conditions
-      self.further_conditions = NUMBER_OF_FURTHER_CONDITIONS.times.map do |index|
+      further_condition_count = (further_conditions&.reject(&:blank?)&.count || 0)
+      number_of_conditions_to_display = [
+        [MIN_FURTHER_CONDITIONS, further_condition_count + 1].max,
+        MAX_FURTHER_CONDITIONS,
+      ].min
+
+      self.further_conditions = number_of_conditions_to_display.times.map do |index|
         existing_value = further_conditions && further_conditions[index]
         existing_value.presence || ''
       end
--- a/app/forms/support_interface/conditions_form.rb
+++ b/app/forms/support_interface/conditions_form.rb
@@ -13,8 +13,17 @@ module SupportInterface
         standard_conditions: standard_conditions_from(application_choice.offer),
         further_conditions: further_conditions_from(application_choice.offer),
       }
-
       new(attrs).tap do |form|
+        form.setup_further_conditions
+      end
+    end
+
+    def self.build_from_params(application_choice, params)
+      form = build_from_application_choice(application_choice)
+
+      form.tap do |form|
+        form.standard_conditions = params['standard_conditions']
+        form.further_conditions = params['further_conditions'].values.map { |v| v['text'] }
         form.setup_further_conditions
       end
     end
@@ -47,7 +56,9 @@ module SupportInterface
     end
 
     def save
-      raise 'TODO'
+      application_choice.update(
+        offer: { conditions: (standard_conditions + further_conditions).reject(&:blank?) }
+      )
     end
   end
 end
--- a/app/forms/support_interface/conditions_form.rb
+++ b/app/forms/support_interface/conditions_form.rb
@@ -17,6 +17,7 @@ module SupportInterface
     validates :audit_comment_ticket, presence: true
     validates :audit_comment_ticket, format: { with: /\A((http|https):\/\/)?(www.)?becomingateacher.zendesk.com\/agent\/tickets\// }
     validate :further_conditions_length
+    validates :confirm_make_unconditional, presence: true, if: :conditions_empty?
 
     def self.build_from_application_choice(application_choice, attrs = {})
       attrs = {
--- a/app/forms/support_interface/conditions_form.rb
+++ b/app/forms/support_interface/conditions_form.rb
@@ -20,12 +20,13 @@ module SupportInterface
 
     attr_accessor :application_choice, :standard_conditions, :further_conditions, :audit_comment_ticket
 
-    MAX_FURTHER_CONDITIONS = OfferValidations::MAX_CONDITIONS_COUNT - MakeAnOffer::STANDARD_CONDITIONS.length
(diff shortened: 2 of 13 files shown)


spec/forms/support_interface/conditions_form_spec.rb
```diff
@@ -0,0 +1,127 @@
+require 'rails_helper'
+
+RSpec.describe SupportInterface::ConditionsForm do
```
stevehook, line 3: This form duplicates some of the logic from `ProviderInterface::OfferWizard`. Can we extract the shared logic into something generic or is that not practical/desirable? (I struggled to see an obvious thing to extract initially but plan to revisit).

app/forms/support_interface/conditions_form.rb
```diff
@@ -0,0 +1,78 @@
...
+  class ConditionsForm
+    class OfferConditionField
+      include ActiveModel::Model
+
+      attr_accessor :id, :text
+
+      validates :text, length: { maximum: 255, too_long: ->(c, _) { "Condition #{c.id + 1} must be %{count} characters or fewer" } }
+    end
+
+    include ActiveModel::Model
```
TobyRet, line 11: Dup!

spec/forms/support_interface/conditions_form_spec.rb
```diff
@@ -0,0 +1,127 @@
...
+  describe '#save' do
+    it 'returns false with a validation error if audit_comment is missing' do
```
TobyRet, line 5: Should there be an assertion re: the validation error message in this spec?

app/forms/support_interface/conditions_form.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    attr_accessor :application_choice, :standard_conditions, :further_conditions, :audit_comment
+
+    NUMBER_OF_FURTHER_CONDITIONS = 4
```
TobyRet, line 15: Is there any relationship between `NUMBER_OF_FURTHER_CONDTIONS` and `MAX_CONDITIONS_COUNT`? https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/services/make_an_offer.rb#L3

app/forms/support_interface/conditions_form.rb
```diff
@@ -0,0 +1,78 @@
...
+    validates :audit_comment, presence: true
+
+    def self.build_from_application_choice(application_choice, attrs = {})
+      attrs = {
+        application_choice: application_choice,
+        standard_conditions: standard_conditions_from(application_choice.offer),
+        further_conditions: further_conditions_from(application_choice.offer),
+      }.merge(attrs)
+
+      new(attrs).tap(&:setup_further_conditions)
```
TobyRet, line 27: I don't get why `setup_further_conditions` is invoked here given we have assigned further conditions on line 24 🤔

TobyRet (approved):
Tested and works like a charm. Some small comments but nothing major ...

app/forms/support_interface/conditions_form.rb
stevehook: Maybe there should be, I'm not sure why `MAX_CONDITIONS_COUNT` is as many as 20. This has changed recently, might need to rethink this a bit.

app/forms/support_interface/conditions_form.rb
stevehook: I think this one is needed because the earlier `include` is for the nested `OfferConditionField` class

app/forms/support_interface/conditions_form.rb
stevehook: It's there to fill out the data structure that drives the list of conditions on the form. It's not well named. I gonna rename it `add_slots_for_new_conditions` to try and make it clearer for now. However I think this needs to be revisited given that we want to up the limit to 20 conditions (and rejig the UI in the process).

TobyRet: Ok that makes sense now 👍

app/forms/support_interface/conditions_form.rb
```diff
@@ -0,0 +1,85 @@
+module SupportInterface
+  class ConditionsForm
+    class OfferConditionField
+      include ActiveModel::Model
+
+      attr_accessor :id, :text
+
+      validates :text, length: { maximum: MakeAnOffer::MAX_CONDITION_LENGTH, too_long: ->(c, _) { "Condition #{c.id + 1} must be %{count} characters or fewer" } }
```
despo, line 8: If the [OfferValidations](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/services/offer_validations.rb#L4) service can't be used to check the conditions length, can we reference the `MAX_CONDITIONS_LENGTH` constant from there, as we will be deprecating MakeAnOffer soon.

app/forms/support_interface/conditions_form.rb
```diff
@@ -0,0 +1,85 @@
...
+    end
+
+    include ActiveModel::Model
+
+    attr_accessor :application_choice, :standard_conditions, :further_conditions, :audit_comment
+
+    MIN_FURTHER_CONDITIONS = 4
```
benswannack, line 15: With the new changes in manage, there is no minimum on the number of conditions set.

app/views/support_interface/application_choice_conditions/edit.html.erb
```diff
@@ -0,0 +1,51 @@
+<% content_for :browser_title, title_with_error_prefix('Update conditions', @form.errors.any?) %>
+<% content_for :before_content, govuk_back_link_to(support_interface_application_form_path(@form.application_choice.application_form_id)) %>
+
+<h1 class="govuk-heading-l">
+  Change offer conditions
+</h1>
+
+<%= form_with model: @form, url: support_interface_update_application_choice_conditions_path(@form.application_choice.id), method: :post do |f| %>
```
benswannack, line 8: Maybe `put` rather than `post`

app/forms/support_interface/conditions_form.rb
cwrw: Yes, the extended limit is on the further conditions, as some vendor integrations can't separate conditions so they send them as a single condition, thus the leniency with condition 1

app/views/support_interface/application_choice_conditions/edit.html.erb
```diff
@@ -0,0 +1,51 @@
...
+
+<%= form_with model: @form, url: support_interface_update_application_choice_conditions_path(@form.application_choice.id), method: :put do |f| %>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <%= govuk_notification_banner(
+        title: t('notification_banner.warning'),
+        classes: 'govuk-notification-banner--warning',
+      ) do |_notification_banner| %>
+        <p class="govuk-body">Offer conditions should only be changed on accepted applications following a request from a provider that has already informed the candidate about the change</p>
+      <% end %>
```
paulrobertlloyd, line 16: Let’s use the warning text component here.

```suggestion
      <%= govuk_warning(text: 'Offer conditions should only be changed on accepted applications following a request from a provider that has already informed the candidate about the change' %>
```

paulrobertlloyd:
Here we are asking for an audit comment (in a textarea), but in other places we ask only for a Zendesk URL (in a text input). Should we be asking for the same info in the same format for all of these support tasks?

stevehook:
> Here we are asking for an audit comment (in a textarea), but in other places we ask only for a Zendesk URL (in a text input). Should we be asking for the same info in the same format for all of these support tasks?

I must admit I didn't see the Zendesk URL prompt until yesterday - but I think it's a better solution so I reckon I should change it...

app/forms/support_interface/conditions_form.rb
stevehook: Support users can now add up to 20 conditions (18 'further' conditions) though they can only add one new one at a time. We should revisit this once the JavaScript enhanced version is done in the provider interface.

app/components/support_interface/application_choice_component.rb
```diff
@@ -97,6 +98,20 @@ def rejection_reason_row
...
+      return unless application_choice.pending_conditions? || application_choice.offer?
+
+      conditions = application_choice.offer&.[]('conditions')
+      return unless conditions.present?
+
+      {
+        key: 'Conditions',
+        value: render(SupportInterface::ConditionsComponent.new(conditions: conditions)),
+        action: 'Change conditions',
+        action_path: support_interface_edit_application_choice_conditions_path(application_choice_id: @application_choice.id),
```
paulrobertlloyd, line 111: Use ‘Change’ link with hidden text specifying ‘conditions’:

```suggestion
        action: 'conditions',
        change_path: support_interface_edit_application_choice_conditions_path(application_choice_id: @application_choice.id),
```

paulrobertlloyd:
Looks good to me. My only question is do we need to ask for the Zendesk ticket URL again, and have the checkbox on the confirmation page if making unconditional? I was imagining a confirmation similar to what we use elsewhere (albeit with a green, not red button):

Screenshot 2021-05-17 at 19 27 17

stevehook:
> Looks good to me. My only question is do we need to ask for the Zendesk ticket URL again, and have the checkbox on the confirmation page if making unconditional? I was imagining a confirmation similar to what we use elsewhere (albeit with a green, not red button):
> 
> Screenshot 2021-05-17 at 19 27 17

Yes that would be simpler I guess. I have to pass the Zendesk URL with this form somehow (because nothing has been saved to the database at this point) but it can be a hidden field.

cwrw:
looks good when was testing locally, however I couldn't see the validation errors?

stevehook:
> looks good when was testing locally, however I couldn't see the validation errors?

I think the only validation we have now that the forms have been simplified a bit is on the Zendesk URL:
image

But I just tested it and that one isn't being enforced when the offer is made unconditional, so I need to fix that!

Is there any other validation missing?

cwrw:
I tried the condition going over 255

cwrw (approved):
👍  thanks for addressing!

There is an edge case here around conditions set by the API vs Manage.
In the API we allow 20 conditions to be set, and they don't have to include standard conditions so could all be further conditions. However in Manage we allow 18 further conditions and 2 standard conditions. If a support user were to attempt to edit an offer from the API they won't be able to see or edit the extra 2 conditions in support. I wonder if we can show up to 20 further conditions just incase if they came from the API but still throw a validation error if someone attempts to add the extra standard conditions. Happy to chat through this

stevehook:
> There is an edge case here around conditions set by the API vs Manage.
> In the API we allow 20 conditions to be set, and they don't have to include standard conditions so could all be further conditions. However in Manage we allow 18 further conditions and 2 standard conditions. If a support user were to attempt to edit an offer from the API they won't be able to see or edit the extra 2 conditions in support. I wonder if we can show up to 20 further conditions just incase if they came from the API but still throw a validation error if someone attempts to add the extra standard conditions. Happy to chat through this

That makes sense, just pushed a change to the validations. Thanks @cwrw !

paulrobertlloyd:
🎉 Sorry this one took a while to figure out, @stevehook.

PR #4670, opened 2021-05-06
Static check page

duncanjbrown:
## Context

If the /check endpoint is slower than 1s PaaS considers our instance unhealthy. To avoid this make it return a static file. We would like this to return a SHA in due course but this should prevent the issue in the short term.


public/check
```diff
@@ -0,0 +1 @@
+OK
```
cwrw, line 1: can we drop the SHA here instead (the command you dropped re docker)

vigneshmsft (approved):

public/check
vigneshmsft: will do that as a separate PR, 
we also have `/integrations/monitoring/version` for the version check for the time being

cwrw (approved):

PR #4671, opened 2021-05-06
Remove default false for 'safe_guarding_issues_completed' and 'interview_preferences_completed'

TobyRet:
## Context

The corresponding sections in the application form now use radio buttons and we want candidates to see a validation error if they do not submit an answer to 'Have you completed this section?'

I missed two columns when [creating the last migration](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/db/migrate/20210429084354_remove_defaults_from_application_forms_table.rb).

- Create db migration to remove default `false`  for 'safe_guarding_issues_completed' and 'interview_preferences_completed' columns

## Link to Trello card

https://trello.com/c/dxkMUjHV/3358-do-not-update-sectioncompleted-fields-to-false-when-updating-application-sections

## Things to check

### DIFF
--- a/db/migrate/20210506101257_remove_more_defaults_from_application_forms_table.rb
+++ b/db/migrate/20210506101257_remove_more_defaults_from_application_forms_table.rb
@@ -1,0 +1,6 @@
+class RemoveMoreDefaultsFromApplicationFormsTable < ActiveRecord::Migration[6.1]
+  def change
+    change_column_default :application_forms, :safeguarding_issues_completed, from: false, to: nil
+    change_column_default :application_forms, :interview_preferences_completed, from: false, to: nil
+  end
+end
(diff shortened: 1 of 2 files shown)


stevehook (approved):
Makes sense for consistency. Though I'm not a fan of 3-valued logic for booleans - because you have to be careful with things like `ApplicationForm.where(safeguarding_issues_completed: false)` because it won't return all the `nil` values and that's probably not what you want. I guess I prefer my booleans to be `NOT NULL`.

TobyRet:
> Makes sense for consistency. Though I'm not a fan of 3-valued logic for booleans - because you have to be careful with things like `ApplicationForm.where(safeguarding_issues_completed: false)` because it won't return all the `nil` values and that's probably not what you want. I guess I prefer my booleans to be `NOT NULL`.

Yeah that's a very valid point @stevehook.

The section complete buttons now use radio buttons instead of checkboxes and the design intent is for them to not be pre-selected as before so that form validation kicks in. I'm struggling to think of an alternative to 3-valued logic for booleans in this instance though. 😞

PR #4674, opened 2021-05-06
Only enable rails_semantic_logger in production

duncanjbrown:
This keeps dev and test logs clean and readable. I tested the logging config still worked in prod by running in prod RAILS_ENV locally — example prod-style log line:

```json
{"host":"maple.local","application":"Semantic Logger","environment":"qa","time":"2021-05-06T12:40:36.987+01:00","level":"info","level_index":2,"pid":71457,"thread":"puma threadpool 003","duration_ms":4.019000101834536,"duration":"4.019ms","tags":["[\"1a3d372a-e967-4003-91d6-683d649e26d0\"]"],"name":"CandidateInterface::SubmittedApplicationFormController","message":"Completed #complete","payload":{"controller":"CandidateInterface::SubmittedApplicationFormController","action":"complete","format":"HTML","method":"GET","path":"/candidate/application/complete","view_runtime":0.0,"db_runtime":0.0,"status":401,"allocations":1216,"status_message":"Unauthorized"},"domain":null,"hosting_environment":"qa","service":"web"}
```

### Before
Screenshot 2021-05-06 at 10 13 51

Screenshot 2021-05-06 at 12 44 13

### After

Screenshot 2021-05-06 at 12 43 05

Screenshot 2021-05-06 at 12 44 38

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -72,7 +72,6 @@ gem 'okcomputer'
 gem 'skylight'
 
 # Logging
-gem 'rails_semantic_logger'
 gem 'request_store_rails'
 gem 'request_store-sidekiq'
 
@@ -108,6 +107,10 @@ gem 'strong_migrations'
 
 # Rails console colours
 gem 'colorize'
+
+group :production do
+  gem 'rails_semantic_logger'
+end
 
 group :development do
   gem 'web-console', '>= 3.3.0'
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ RUN apk -U upgrade && \
     bundler -v && \
     bundle config set no-cache 'true' && \
     bundle config set no-binstubs 'true' && \
-    bundle --retry=5 --jobs=4 --without=development && \
+    bundle --retry=5 --jobs=4 --without=development --with=production && \
     yarn install --check-files && \
     apk del .gem-installdeps && \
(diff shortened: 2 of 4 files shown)


Gemfile
```diff
@@ -109,6 +108,10 @@ gem 'strong_migrations'
 # Rails console colours
 gem 'colorize'
 
+group :production do
```
vigneshmsft, line 111: should anything change on the build setup for bundle to include this group?

duncanjbrown: I don't think so, it looks at RAILS_ENV. From [docs](https://bundler.io/v2.2/guides/groups.html):

```
By default, a Rails generated app calls Bundler.require(:default, Rails.env) in your application.rb, which links the groups in your Gemfile to the Rails environment.
```

duncanjbrown: It worked locally with RAILS_ENV=production

vigneshmsft: we'll have to set RAILS_ENV before bundle install stage ?
https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/Dockerfile#L39

duncanjbrown: or pass `--with production` — I'll add

duncanjbrown: added

malcolmbaig (approved):

PR #4675, opened 2021-05-06
Add condition 1 max length on offers

cwrw:
## Context

Providers using SITS require a longer condition length as they cannot be separated in the API

The API requires the condition 1 to be a max length of 2000.

## Guidance to review

I've included this in both manage/API validation for now 

## Link to Trello card

https://trello.com/c/1rO3K8vw/3660-sandbox-condition-1-validation

## Things to check

### DIFF
--- a/app/services/offer_validations.rb
+++ b/app/services/offer_validations.rb
@@ -2,6 +2,8 @@ class OfferValidations
   include ActiveModel::Model
 
   MAX_CONDITIONS_COUNT = 20
+  # This is required for the API integrations which send conditions together
+  MAX_CONDITION_1_LENGTH = 2000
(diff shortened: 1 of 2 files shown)


despo:
Is there a way to identify if the provider is using SITS so we can do some processing on our side?

Also how does this impact marking individual conditions as met, as well as the email send to the candidate where all conditions will be merged into one?

despo (approved):
👍🏼

cwrw:
Discussed with @despo but for posterity: this will only be for the API, as we cover Manage validations in forms. We cannot distinguish SITS unfortunately. In terms of grouping conditions this is currently seen as one condition, but work is being discussed around handling individual conditions in both the API and vendor integrations.

PR #4677, opened 2021-05-06
Sentry - show application support url in tags

malcolmbaig:
## Context
Right now we're displaying this as extra info. Move it into the tags
context for better visibility and convenience.

See diff

## Guidance to review
I think we need a development environment set up in Sentry to test this locally. This might be one to merge and test on QA.

## Link to Trello card
https://trello.com/c/gIUsCtQa

## Things to check


duncanjbrown (approved):
+1 to a Sentry sandbox

PR #4680, opened 2021-05-06
Replace 'onchange' with a JS script in the validation error summary

JR-G:
## Context

When viewing the validation error summary table, the order of the errors doesn't change when using 'last week' or 'last month' sorting drop down option (unless you press `enter` afterwards).

The console would display the following message, after attempting to change the 'sort by' value:

image

Replace `onchange: 'this.form.submit()'` with a separate JS script that has an event listener for this change.

## Guidance to review

This works locally now. My JS script knowledge is quite rusty though, so I'm interested to hear if there is a more efficient/better way of writing this script.

## Link to Trello card

https://trello.com/c/LSaBKlO6/3359-validation-error-table-order-doesnt-change-when-using-last-week-or-last-month-sorting

## Things to check

### DIFF
--- a/app/components/support_interface/validation_errors_summary_component.html.erb
+++ b/app/components/support_interface/validation_errors_summary_component.html.erb
@@ -17,8 +17,10 @@
                 },
                 class: 'sortedby-label',
                 selected: 2,
-                onchange: 'this.form.submit()',
                 role: 'listbox',
+                data: {
+                  module: 'sort-by-filter-option',
+                },
               ) %>
         </div>
         <%= f.govuk_submit('Update') %>
--- a/app/frontend/packs/sort-by-filter.js
+++ b/app/frontend/packs/sort-by-filter.js
@@ -1,0 +1,10 @@
+const sortByFilter = () => {
+  const filter = document.querySelector('[data-module="sort-by-filter-option"]')
+  if (filter) {
+    filter.addEventListener('change', (event) => {
+      event.target.form.submit()
+    })
+  };
+}
+
+export default sortByFilter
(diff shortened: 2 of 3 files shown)


app/components/support_interface/validation_errors_summary_component.html.erb
```diff
@@ -15,10 +15,9 @@
                 options: {
                   selected: params['sortby'] || 'all_time',
                 },
-                class: 'sortedby-label',
-                selected: 2,
-                onchange: 'this.form.submit()',
-                role: 'listbox',
```
stevehook, line 21: Why do we need to get rid of `class`, `selected` and `role` here?

app/frontend/packs/sort-by-filter.js
```diff
@@ -0,0 +1,10 @@
+const sortByFilter = () => {
+    const filter = document.querySelector('[data-module="sort-by-filter-option"]')
+    if(filter){
+        filter.addEventListener('change', (event) => {
+            event.target.form.submit();
+        });
+    };
+};
```
stevehook, line 8: This makes sense to me. The only question I've got is whether we need to add the `data-sort-by-filter-option`? Can we use `#sortby-field` as the selector? @paulrobertlloyd might have a view on a preferred way?

stevehook (commented):
👍 Review works now and code looks good.

A couple of questions inline.

app/frontend/packs/sort-by-filter.js
paulrobertlloyd: I’d usually select a field via a data attribute, or a `js-` prefixed class.

app/components/support_interface/validation_errors_summary_component.html.erb
JR-G: I've now added these back in

stevehook (approved):

davidgisbey (approved):
lgtm

PR #4689, opened 2021-05-07
Fix JS back links

duncanjbrown:
## Context

The Content Security Policy broke our back links because they had `javascript:` URLs in the `href`

~Instead of using the `href`, use an `onclick` handler to trigger `history.back()`~

Remove the JavaScript and always try to use `referer` when no explicit link is passed. And if we have no `referer`, fall back to the root path for the current namespace.

## Link to Trello card

https://trello.com/c/Ilb9lDg8/3365-javascript-back-links-are-not-working


duncanjbrown:
This needs a bit more work — currently JS clobbers a valid referer from the server and onclick is not optimal

app/frontend/packs/app-back-link.js
```diff
@@ -4,7 +4,8 @@ const initBackLinks = () => {
   const backLinks = document.querySelectorAll('.app-back-link--fallback')
```
benswannack, line 4: We currently only `initBackLinks` in the candidate js file. Can it be moved to the main js file?

duncanjbrown: Yes, we're doing this

frankieroberto (changes requested):
I had a quick test, and whilst it mostly works, if you get a validation error on a page with a back link, the back link then links back to the page that you're already on, and you get completely stuck...  (possible temporary fix: check if the referrer matches the path of the current page, and if so, then default to the root path which would at least get you back to the application overview page?)

A harder problem to quickly solve is that if you press the back link twice (eg within the Add a degree flow), you end up stuck in an infinite loop bouncing between two pages...

frankieroberto (approved):
Approved as we'll start fixing this properly in separate PRs.

PR #4696, opened 2021-05-10
Offer another course for pending conditions choices

davidgisbey:
## Context

At the moment, the thing that we're asked to do most via support is changing an offered course for a pending conditions application.

This PR adds in this functionality to the support UI. It's currently behind a feature flag, but if we're happy with it i'll remove it pretty quickly.

The PR largely adds 3 new pages:  

- A search page which piggy backs from the functionality Toby previously built in the 
`SupportInterface::ApplicationForms::PickCourseForm`. 
- A course option selection page where a support user can pick a course option to replace the current option with
- A confirm selection page where a support user is provided guidance and asked to confirm they've read it.  They are also expected to input the Zendesk ticket where it was requested

Application form show page on the support UI 
- show `Change offered course` link for pending conditions application choices
image

Search page 

image

Choose course option page 

image

Confirmation page 

image

## Guidance to review

Probably best per commit, but i do break the controllers out later so bear that in mind.

Probs best to test it locally.

I've kept the validations pretty light touch. For example you can offer the same course etc. Happy to add stuff in, but as it's done via the support UI i'm not convinced it's needed.

## Link to Trello card

https://trello.com/c/lFU7kZwN/3336-enable-course-to-be-changed-after-offer-acceptance

## Things to check


JR-G:
Just an open question - will/should there be any checks in place for what course can be offered in place of the existing one?
image

app/views/support_interface/application_forms/application_choices/change_offered_course/change_offered_course_search.html.erb
```diff
@@ -0,0 +1,35 @@
...
+    <p class="govuk-body">You can find the course code by searching on
+      <%= govuk_link_to(
+        'Find postgraduate teacher training',
+        'https://www.find-postgraduate-teacher-training.service.gov.uk/',
+        target: '_blank',
+        rel: 'nofollow',
+      ) %>.
+    </p>
+    <p class="govuk-body">The code you need is in the title of the course.</p>
+    <p class="govuk-body">For example, the course code for York St John University Primary (Q859), is Q859.</p>
```
paulrobertlloyd, line 20:
```suggestion
    <p class="govuk-body">A candidate can only accept one of the courses offered by any of the providers they have submitted an application to.</p>
    <p class="govuk-body">The support team can change the offered course accepted by a candidate if a request is made before the end of the cycle.</p>
    <p class="govuk-body">Candidates should decline any other offers they have received.</p>
    <p class="govuk-body">You can find the course code by searching on
      <%= govuk_link_to(
        'Find postgraduate teacher training (opens in a new tab)',
        'https://www.find-postgraduate-teacher-training.service.gov.uk/',
        target: '_blank',
        rel: 'nofollow',
      ) %>.
    </p>
    <p class="govuk-body">The code you need is in the title of the course. For example, the course code for York St John University Primary (Q859), is Q859.</p>
```

app/views/support_interface/application_forms/application_choices/change_offered_course/confirm_offered_course_option.html.erb
```diff
@@ -0,0 +1,52 @@
...
+    <div class="govuk-grid-column-two-thirds">
+      <%= f.govuk_text_field(
+              :audit_comment,
+              label: {
+                text: 'Zendesk ticket URL',
+                size: 'm',
+              },
+              rows: 1,
+              hint: { text: 'For example https://becomingateacher.zendesk.com/agent/tickets/12345' },
+            ) %>
```
paulrobertlloyd, line 43: 😵‍💫

```suggestion
      <%= f.govuk_text_field(
        :audit_comment,
        label: {
          text: 'Zendesk ticket URL',
          size: 'm',
        },
        rows: 1,
        hint: { text: 'For example https://becomingateacher.zendesk.com/agent/tickets/12345' },
      ) %>
```

app/views/support_interface/application_forms/application_choices/change_offered_course/offered_course_options.html.erb
```diff
@@ -0,0 +1,31 @@
+<% content_for :title, title_with_error_prefix("Choose a course to replace #{@application_choice.current_course_option.course.name_and_code}", @pick_course.errors.any?) %>
+<% content_for :before_content, govuk_back_link_to(support_interface_application_form_application_choice_change_offered_course_search_path(course_code: @pick_course.course_code)) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+
+    <% if @pick_course.unavailable_courses.present? %>
+      <h2 class='govuk-heading-m'>Courses with no vacancies</h2>
```
paulrobertlloyd, line 8:
```suggestion
      <h2 class="govuk-heading-m">Courses with no vacancies</h2>
```

malcolmbaig:
`SupportInterface::PickCourseForm` doesn't appear to be scoping by provider. Is that acceptable this cycle? Historically I don't think course codes have been unique, just unique per provider.

davidgisbey:
Yea this is an interesting one. Some courses do have the same code. a while ago i added the provider to the radio buttons to mitigate this issue as two courses have the same code and name. 

If we want to make this error prone, we could do something like 

`@pick_course.course_options_for_provider(provider_id: @application_choice.provider.id)`

then add a method which is similar to the course_options method, but scopes the courses to something like 

`Course.where(provider_id: provider.id).or(Course.where(accredited_provider_id: provider.id) `

malcolmbaig:
@davidgisbey That sounds like a good approach to me. Choosing a course from a different provider is something we actively don't want in this scenario, so having the form code match that business rule seems sensible.

app/forms/support_interface/application_forms/pick_course_form.rb
```diff
@@ -82,6 +96,18 @@ def courses
           .where(code: sanitize(course_code))
       end
 
+      def courses_for_provider(provider)
+        return [] if provider.blank?
+
+        @courses ||= Course
```
malcolmbaig, line 102: Can remove the memoization here - only calling the method once, and if we did want to memoize we'd use a different pattern so that different provider values would return the correct results.

malcolmbaig (approved):
One further comment left inline, but otherwise LG 👍🏼

app/forms/support_interface/application_forms/pick_course_form.rb
davidgisbey: Yea good shout

PR #4698, opened 2021-05-10
Use GovukComponent::Header

paulrobertlloyd:
## Context

Uses `GovukComponent::Header` to render headers (via `govuk-components` `govuk_header` helper method).

Required the following changes to `govuk-components`:

- [x] [Can’t set product name value](https://github.com/DFE-Digital/govuk-components/issues/133)
- [x] [Product name and phase tag are not linked](https://github.com/DFE-Digital/govuk-components/issues/140)
- [x] [No menu button to reveal navigation items on mobile](https://github.com/DFE-Digital/govuk-components/issues/132)
- [x] [Can’t set classes on the navigation list](https://github.com/DFE-Digital/govuk-components/issues/141)
- [x] [Can’t pass anything other than a link in a navigation item](https://github.com/DFE-Digital/govuk-components/issues/142)
- [x] [Fix outstanding header issues](https://github.com/DFE-Digital/govuk-components/pull/151)

## Guidance to review

* Are there too many params for the `HeaderComponent`?
* Should `HeaderComponent` be a component, or a helper?

## Things to check


duncanjbrown (commented):
Very happy to see this! Small comments:

- all the headers link to `/` apart from the `candidate_interface` one... which is odd, as they're still using `service_link`?
- There's a stray `ProductHeaderComponent` in `layouts/error.html.erb`
- Email addresses in the header are now lighter-weight — is this right?
Screenshot 2021-05-11 at 21 51 54

paulrobertlloyd:
> all the headers link to / apart from the candidate_interface one... which is odd, as they're still using service_link?

Ah, turns out to link the product name you need to provide `logo_href`; refactored the component to always take `service_link`, and populate `logo_href` and `service_name_href` params with this value.

> There's a stray ProductHeaderComponent in layouts/error.html.erb

Fixed!

> Email addresses in the header are now lighter-weight — is this right?

It’s different from what we had previously, but I think it makes sense (now bold == linked).

duncanjbrown (approved):

PR #4701, opened 2021-05-10
Create local dev provider users in a consistent way

duncanjbrown:
We weren't creating notification prefs for provider users in `setup_local_dev_data` so the app crashed when trying to view one of these users through support.

Move all user creation into the `CreateExampleProviderUsersWithPermissions` class and use a method which depends on `SaveProviderUser` to do the saving.


mnacos (approved):
LGTM

PR #4702, opened 2021-05-10
Start writing opened_on_apply_at for courses

mnacos:
## Context

Following on from https://github.com/DFE-Digital/apply-for-teacher-training/pull/4687

Populate `opened_on_apply_at` any time `open_on_apply` changes. If the course is being opened, record the timestamp. If the course is being closed (e.g. reverting an accidental opening), reset timestamp to `nil`.

## Guidance to review

Should be straightforward. Inspect the tests and the review app.

## Link to Trello card

https://trello.com/c/DG1rdUbK

## Things to check

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -145,4 +145,13 @@ class Course < ApplicationRecord
   def ratifying_provider
     accredited_provider || provider
   end
+
+  def open!
+    return if persisted? && open_on_apply
+
+    update!(
+      open_on_apply: true,
+      opened_on_apply_at: Time.zone.now,
+    )
+  end
 end
--- a/app/controllers/support_interface/course_controller.rb
+++ b/app/controllers/support_interface/course_controller.rb
@@ -16,12 +16,26 @@ module SupportInterface
     def update
       @course = Course.find(params[:course_id])
 
-      if @course.update(params.require(:course).permit(:open_on_apply))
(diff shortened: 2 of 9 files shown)


app/models/course.rb
```diff
@@ -9,6 +9,8 @@ class Course < ApplicationRecord
 
   audited associated_with: :provider
 
+  before_save :set_opened_on_apply_at
```
duncanjbrown, line 12: Historically we haven't handled this kind of thing with ActiveRecord callbacks and I think that's generally been good. Would it clearer to set the date in `OpenProviderCourses` — or at least to have an `open!` method on each course?

And do we need to set back to `nil`? It's still fact that the course was opened on apply at that date.

cwrw: +1 for `#open!`
Regarding `nil`,  in our last conversation as @mnacos pointed out, we were hoping to only have it set if the value is true. Since there seems to precedence with timestamps such as `rejected_at` never being unset maybe we can keep it if a course is reverted back to false, but always update it if a value moves from false to true.

mnacos: I've added an `#open!` method. May result in more saves during a sync, but we're planning to drop audits for courses anyway.

cwrw (approved):
could you add the new timestamp to the `open_on_apply` trait in the factory, otherwise 👍

PR #4704, opened 2021-05-10
Display vacancy status on course selection

JR-G:
## Context

At the moment, when a candidate chooses a course, there's no indication that it is full in the drop down. As more courses close towards the end of the cycle a candidate could end up selecting many full courses in a row

As a candidate I want to know whether a course has vacancies or is only available via UCAS at the point that I select it (rather than finding out two screens further into the journey).

|Before|After|
|---|---|
|image|image|
|image|image|

## Link to Trello card

https://trello.com/c/rgEO3WLn/2133-display-vacancy-status-and-ucas-only-when-prompting-candidates-for-course

## Things to check

### DIFF
--- a/app/forms/candidate_interface/pick_course_form.rb
+++ b/app/forms/candidate_interface/pick_course_form.rb
@@ -11,13 +11,13 @@ module CandidateInterface
 
     def radio_available_courses
       @radio_available_courses ||= begin
-        courses_for_current_cycle.exposed_in_find.order(:name)
+        courses_for_current_cycle.exposed_in_find.order(:name).includes(:course_options)
       end
     end
 
     def dropdown_available_courses
       @dropdown_available_courses ||= begin
-        courses = courses_for_current_cycle.exposed_in_find.includes(:accredited_provider)
+        courses = courses_for_current_cycle.exposed_in_find.includes(:accredited_provider, :course_options)
 
         courses_with_names = courses.map(&:name).map(&:downcase)
         courses_with_descriptions = courses.map(&:name_and_description).map(&:downcase)
@@ -36,6 +36,14 @@ module CandidateInterface
                  else
                    course.name_and_code
                  end
+
+          if !course.open_on_apply?
+            name += ' – Only on UCAS'
+          end
+
+          if course.course_options.available.blank?
+            name += ' – No vacancies'
+          end
 
           DropdownOption.new(course.id, name)
         end
--- a/app/views/candidate_interface/course_choices/course_selection/new.html.erb
+++ b/app/views/candidate_interface/course_choices/course_selection/new.html.erb
@@ -22,7 +22,13 @@
       <% else %>
         <%= f.govuk_radio_buttons_fieldset :course_id, legend: { size: 'xl', text: t('page_titles.which_course'), tag: 'h1' } do %>
           <% @pick_course.radio_available_courses.each_with_index do |course, i| %>
-            <%= f.govuk_radio_button :course_id, course.id, label: { text: "#{course.name} (#{course.code})" }, hint: { text: course.description }, link_errors: i.zero? %>
+            <% if !course.open_on_apply? %>
+              <%= f.govuk_radio_button :course_id, course.id, label: { text: "#{course.name} (#{course.code}) – Only on UCAS" }, hint: { text: course.description }, link_errors: i.zero? %>
+            <% elsif course.course_options.available.blank? %>
+              <%= f.govuk_radio_button :course_id, course.id, label: { text: "#{course.name} (#{course.code}) – No vacancies" }, hint: { text: course.description }, link_errors: i.zero? %>
+            <% else %>
+              <%= f.govuk_radio_button :course_id, course.id, label: { text: "#{course.name} (#{course.code})" }, hint: { text: course.description }, link_errors: i.zero? %>
+            <% end %>
           <% end %>
         <% end %>
       <% end %>
--- a/spec/forms/candidate_interface/pick_course_form_spec.rb
+++ b/spec/forms/candidate_interface/pick_course_form_spec.rb
@@ -18,10 +18,36 @@ RSpec.describe CandidateInterface::PickCourseForm do
   end
 
   describe '#dropdown_available_courses' do
+    it 'displays the vacancy status' do
+      provider = create(:provider)
+      course = create(:course, open_on_apply: true, exposed_in_find: true, name: 'Maths', code: '123', provider: provider)
+      create(:course, open_on_apply: true, exposed_in_find: true, name: 'English', code: '456', description: 'PGCE with QTS full time', provider: provider)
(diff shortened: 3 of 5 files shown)


app/forms/candidate_interface/pick_course_form.rb
```diff
@@ -17,7 +17,7 @@ def radio_available_courses
 
     def dropdown_available_courses
       @dropdown_available_courses ||= begin
-        courses = courses_for_current_cycle.exposed_in_find.includes(:accredited_provider)
+        courses = courses_for_current_cycle.exposed_in_find.includes(:accredited_provider, :course_options)
```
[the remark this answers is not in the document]
stevehook: The alternative to using `includes` is to move the logic into the DB query with a join and condition on the joined table:
`courses_for_current_cycle.joins(:course_options).where('course_options.available': true).distinct`
the problem is that that doesn't return the ones that aren't available, so you would need two queries. But if performance is a still a problem that's probably the next thing to try.

stevehook (commented):
Looks good - just a few comments related to the perf thing we were discussing on Slack.

app/forms/candidate_interface/pick_course_form.rb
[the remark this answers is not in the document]
JR-G: I've updated `radio_available_courses` and it seems to be performing better on the review app

paulrobertlloyd:
Wondering if we can display courses as follows:

* Without age range:
  `Music (X123) – No vacancies`
* With age range:
  `Music, 11 to 16 (X124) – No vacancies`

JR-G:
> Wondering if we can display courses as follows:
> 
> * Without age range:
>   `Music (X123) – No vacancies`
> * With age range:
>   `Music, 11 to 16 (X124) – No vacancies`

@paulrobertlloyd There are some courses that will display an additional description: `English (456) – PGCE with QTS full time - No vacancies`. Shall I follow a similar pattern in reducing the number of hyphens?

paulrobertlloyd:
How about this…

Select option (only shows qualification when disambiguating):

```
Music, 11 to 16 (X124) – PGCE with QTS full time – No vacancies
```

Radio option (shows qualification as hint text):

```
* Music, 11 to 16 (X124) – No vacancies
  PGCE with QTS full time
```

frankieroberto (approved):

app/forms/candidate_interface/pick_course_form.rb
```diff
@@ -37,6 +37,14 @@ def dropdown_available_courses
                    course.name_and_code
                  end
 
+          if !course.open_on_apply?
+            name += ' - Only on UCAS'
```
paulrobertlloyd, line 41:
```suggestion
            name += ' – Only on UCAS'
```

app/forms/candidate_interface/pick_course_form.rb
```diff
@@ -37,6 +37,14 @@ def dropdown_available_courses
...
+          end
+
+          if course.course_options.available.blank?
+            name += ' - No vacancies'
```
paulrobertlloyd, line 45:
```suggestion
            name += ' – No vacancies'
```

app/views/candidate_interface/course_choices/course_selection/new.html.erb
```diff
@@ -22,7 +22,13 @@
       <% else %>
         <%= f.govuk_radio_buttons_fieldset :course_id, legend: { size: 'xl', text: t('page_titles.which_course'), tag: 'h1' } do %>
           <% @pick_course.radio_available_courses.each_with_index do |course, i| %>
-            <%= f.govuk_radio_button :course_id, course.id, label: { text: "#{course.name} (#{course.code})" }, hint: { text: course.description }, link_errors: i.zero? %>
+            <% if !course.open_on_apply? %>
+              <%= f.govuk_radio_button :course_id, course.id, label: { text: "#{course.name} (#{course.code}) - Only on UCAS" }, hint: { text: course.description }, link_errors: i.zero? %>
```
paulrobertlloyd, line 26:
```suggestion
              <%= f.govuk_radio_button :course_id, course.id, label: { text: "#{course.name} (#{course.code}) – Only on UCAS" }, hint: { text: course.description }, link_errors: i.zero? %>
```

app/views/candidate_interface/course_choices/course_selection/new.html.erb
```diff
@@ -22,7 +22,13 @@
...
+            <% elsif course.course_options.available.blank? %>
+              <%= f.govuk_radio_button :course_id, course.id, label: { text: "#{course.name} (#{course.code}) - No vacancies" }, hint: { text: course.description }, link_errors: i.zero? %>
```
paulrobertlloyd, line 28:
```suggestion
              <%= f.govuk_radio_button :course_id, course.id, label: { text: "#{course.name} (#{course.code}) – No vacancies" }, hint: { text: course.description }, link_errors: i.zero? %>
```

spec/forms/candidate_interface/pick_course_form_spec.rb
```diff
@@ -18,11 +18,37 @@
...
+    it 'displays the vacancy status' do
+      provider = create(:provider)
+      course = create(:course, open_on_apply: true, exposed_in_find: true, name: 'Maths', code: '123', provider: provider)
+      create(:course, open_on_apply: true, exposed_in_find: true, name: 'English', code: '456', description: 'PGCE with QTS full time', provider: provider)
+      create(:course, open_on_apply: true, exposed_in_find: true, name: 'English', code: '789', description: 'PGCE full time', provider: provider)
+      create(:course_option, course: course)
+
+      form = described_class.new(provider_id: provider.id)
+
+      expect(form.dropdown_available_courses.map(&:name)).to eql(['English (456) – PGCE with QTS full time - No vacancies', 'English (789) – PGCE full time - No vacancies', 'Maths (123)'])
```
paulrobertlloyd, line 30:
```suggestion
      expect(form.dropdown_available_courses.map(&:name)).to eql(['English (456) – PGCE with QTS full time – No vacancies', 'English (789) – PGCE full time – No vacancies', 'Maths (123)'])
```

spec/system/candidate_interface/course_selection/candidate_selecting_a_course_that_is_full_spec.rb
```diff
@@ -32,6 +32,8 @@ def when_i_select_the_full_course
     select @course.provider.name
     click_button t('continue')
 
+    expect(page).to have_text("#{@course.name_and_code} - No vacancies")
```
paulrobertlloyd, line 35:
```suggestion
    expect(page).to have_text("#{@course.name_and_code} – No vacancies")
```

paulrobertlloyd (changes requested):
Just need to use an [en-dash](https://en.wikipedia.org/wiki/Dash#En_dash) not a hyphen for separating words.

davidgisbey (approved):
lgtm once the change have been amde

paulrobertlloyd (approved):

PR #4705, opened 2021-05-11
Update flow and forms so that references are created after name step

george-bit:
## Context

When candidates  leave the flow part of the way through adding a reference, incomplete references are displayed on the review page.

image

Candidates now see 'Not entered' in place of whitespace on the review component.

Candidates now only can create an incomplete reference by leaving the flow after submitting the referee name. submitting Referee Type now passes a param to the name form which creates the reference ehich created the db record.

## Guidance to review

Test in the review app and let me know RE: test coverage 

## Link to Trello card

https://trello.com/c/T3fMitzm/3261-candidates-can-remove-information-from-references-by-accident

## Things to check

### DIFF
--- a/app/components/candidate_interface/references_review_component.rb
+++ b/app/components/candidate_interface/references_review_component.rb
@@ -80,7 +80,7 @@ module CandidateInterface
     def name_row(reference)
       {
         key: 'Name',
-        value: reference.name,
+        value: reference.name || 'Not entered',
         action: "name for #{reference.name}",
         change_path: candidate_interface_references_edit_name_path(
           reference.id, return_to: :review
@@ -91,7 +91,7 @@ module CandidateInterface
     def email_row(reference)
       {
         key: 'Email address',
-        value: reference.email_address,
+        value: reference.email_address || 'Not entered',
         action: "email address for #{reference.name}",
         change_path: candidate_interface_references_edit_email_address_path(
           reference.id, return_to: :review
@@ -102,7 +102,7 @@ module CandidateInterface
     def relationship_row(reference)
       {
         key: 'Relationship to referee',
-        value: reference.relationship,
+        value: reference.relationship || 'Not entered',
         action: "relationship for #{reference.name}",
         change_path: candidate_interface_references_edit_relationship_path(
           reference.id, return_to: :review
--- a/app/components/candidate_interface/references_review_component.rb
+++ b/app/components/candidate_interface/references_review_component.rb
@@ -80,7 +80,7 @@ module CandidateInterface
     def name_row(reference)
       {
         key: 'Name',
-        value: reference.name || 'Not entered',
+        value: reference.name,
         action: "name for #{reference.name}",
         change_path: candidate_interface_references_edit_name_path(
           reference.id, return_to: :review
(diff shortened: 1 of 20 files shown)


spec/system/candidate_interface/references/candidate_adding_referees_in_sandbox_spec.rb
```diff
@@ -57,4 +60,21 @@ def then_i_see_that_references_are_given
...
+    click_button t('save_and_continue')
+  end
+
+  def then_i_see_that_the_incomplete_reference_rendered
+    visit candidate_interface_references_review_path
+    within all('.app-summary-card')[2] do
+      expect(all('.govuk-summary-list__value')[0].text).to have_content('Not entered')
+      expect(all('.govuk-summary-list__value')[1].text).to have_content('Not entered')
+      expect(all('.govuk-summary-list__value')[2].text).to have_content('Academic')
+      expect(all('.govuk-summary-list__value')[3].text).to have_content('Not entered')
```
malcolmbaig, line 79: Can use the `within_summary_row` [helper](https://github.com/DFE-Digital/apply-for-teacher-training/blob/3261-references-save-at-end-of-flow/spec/support/test_helpers/candidate_helper.rb#L449) to make this slightly clearer.

eg -
`within_summary_row('Reference type') { expect(page.text).to have_content('Academic') }`

spec/system/candidate_interface/references/candidate_adding_referees_in_sandbox_spec.rb
```diff
@@ -57,4 +60,21 @@ def then_i_see_that_references_are_given
       expect(all('.govuk-summary-list__value')[4].text).to have_content('Reference given')
     end
   end
+
```
malcolmbaig, line 63: I think these steps warrant their own system spec file, leaving this one to focus on the sandbox behaviour.

app/components/candidate_interface/references_review_component.rb
```diff
@@ -91,7 +91,7 @@ def name_row(reference)
     def email_row(reference)
       {
         key: 'Email address',
-        value: reference.email_address,
+        value: reference.email_address || 'Not entered',
```
frankieroberto, line 94: In theory this isn't needed, as there will never be a blank/nil name, is that right? Probably doesn't do any harm to leave it here though.

config/routes.rb
```diff
@@ -443,8 +443,8 @@
         get '/type/edit/:id' => 'references/type#edit', as: :references_edit_type
         patch '/type/edit/:id' => 'references/type#update'
 
-        get '/name/:id' => 'references/name#new', as: :references_name
-        patch '/name/:id' => 'references/name#create'
+        get '/name/:referee_type' => 'references/name#new', as: :references_name
+        patch '/name/:referee_type' => 'references/name#create'
```
frankieroberto, line 447: Might be worth adding some constraints on the `:referee_type` here?

I managed to cause an exception by visiting `/candidate/application/references/name/invalid` and then filling out the form with a name and pressing save. Unlikely to happen in production though admittedly! Alternatively, could use a query string param and validate it?

```suggestion
        constraints: { referee_type: /(academic|professional|school-based|character)/ } do 
          get '/name/:referee_type' => 'references/name#new', as: :references_name
          patch '/name/:referee_type' => 'references/name#create'
        end
```

frankieroberto (approved):
Looks good, thanks! Tested in the review app and it works well.

Possible design changes for the future:

* Change "Save and Continue" to just "Continue" (not sure there’s any logic behind using one of these vs the other, and it might be simpler to use the same everywhere?)
* Change the "Change" link to "Enter [thing]" if it’s not yet been added. Possibly in place of the of the "Not entered" text?
* Possibly the order of rows in the summary should match the order of questions - ie move the the "Reference type" to the top? That way it might be clearer that you'd not yet answered the last 2 questions.
* Routing: after entering an email, if there is also no "relationship to referee" entered, it might make sense to redirect to that next question, rather than back to the review page again.

config/routes.rb
malcolmbaig: A route constraint's a good idea, and probably sufficient  —  no need for graceful error handling and validation messages in this scenario IMHO

app/components/candidate_interface/references_review_component.rb
george-bit: Yeah I did think this but it felt kinda wrong to exclude this row aesthetically will leave be

davidgisbey:
So at the moment if do this 
image

when i submit i get this 

image


I think we need to make it fail gracefully. We could either just redirect them to the type page or add an inclusion validation. I'd lean towards just redirecting them to the type page

config/routes.rb
george-bit: @davidgisbey ☝️ I've implemented this above and it prevents the problem below. Will push up shortly.

frankieroberto:
@davidgisbey would the route constraint fix this? see https://github.com/DFE-Digital/apply-for-teacher-training/pull/4705#discussion_r630848774

app/components/candidate_interface/references_review_component.rb
george-bit: Hey @frankieroberto, I managed to update the continue button and the row ordering. 

The "change" is a bit more difficult as it appears to be part of the ViewComponent class. Igt woul be possible to remove it and perhaps put a hard link in the place of not entered to redirect in the same way. I think should be handled in a different PR though as it will likely mean the associated tests will have to be changed to accomodate this.

I'm not to sure as to your point about routing? There is a validation on this page to prevent a blank entry?

config/routes.rb
davidgisbey: Yea this would work 👍

davidgisbey:
Yea that would work 👍

george-bit:
Morning @frankieroberto, I had another look at this last night and managed to implement a solution for the remaining points.  

Please could you re-review? :)

frankieroberto (changes requested):
@george-bit thanks, this is looking good. Couple more minor things:

1) can we have a lowercase `e` and `r` in this link text please?

Screenshot 2021-05-13 at 09 01 12

2) Minor bug. If you go through the flow to add a new referee, enter a name, then press the "Back" link from the email address page, the previous screen is shown but without the name pre-filled, and if you enter a name and continue the flow, you’ll create a duplicate referee. To fix this, I think we’d need to link back to the name field page using the ID of the newly-created referee?  You might still get a different version of this issue if using the browser back button though – it'll reload the previous page from a cache and will prefill the name, but will again create a duplicate if you continue through the flow. Can’t think of an easy way around that though so it might be acceptable?

frankieroberto (approved):

malcolmbaig (approved):

PR #4706, opened 2021-05-11
Fix for flakey time-dependent test

stevehook:
## Context

This test failed late in the evening with an off by one 1 day error when
asserting a date output. I think this is because the test takes a
finite amount of time to complete it's steps and I got unlucky with the
assertion happening the day after the code under test ran. So a bit of
an edge case. 

I think the only reliable way to fix this is with `Timecop.freeze` rather than `travel`. Open to any other ideas.

## Guidance to review

Is there a better way to fix this that avoids `Timecop.freeze`?

## Link to Trello card

n/a

## Things to check

### DIFF
--- a/spec/system/candidate_interface/references/candidate_views_reference_history_spec.rb
+++ b/spec/system/candidate_interface/references/candidate_views_reference_history_spec.rb
@@ -3,6 +3,10 @@ require 'rails_helper'
 RSpec.feature 'Reference history on review page' do
   include CandidateHelper
 
+  around do |example|
+    Timecop.freeze { example.run }
+  end
+
   scenario 'candidate views reference history', with_audited: true do
     given_i_am_signed_in
     and_i_add_a_reference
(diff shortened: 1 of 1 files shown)


spec/system/candidate_interface/references/candidate_views_reference_history_spec.rb
```diff
@@ -29,22 +33,22 @@ def and_i_add_a_reference
...
-    Timecop.travel(@reference.requested_at + 1.day) do
+    Timecop.freeze(@reference.requested_at + 1.day) do
       click_link 'Send a reminder to this referee'
       click_button 'Yes I’m sure - send a reminder'
     end
   end
 
   def and_the_system_sends_an_automated_reminder
-    Timecop.travel(@reference.requested_at + TimeLimitConfig.chase_referee_by.days) do
+    Timecop.freeze(@reference.requested_at + TimeLimitConfig.chase_referee_by.days + 1.minute) do
```
stevehook, line 51: The plus one minute is needed here because `ChaseReferences` checks for time now being greater than (not greater than or equal) to the time limit.

davidgisbey (approved):

PR #4707, opened 2021-05-11
Append custom messages to Semantic Logger

cwrw:
## Context

In the previous logging methods, we could set and get `RequestLocal` store values. However due to the way the new logger is initialised the `RequestLocal` current store id isn't being set in the custom logger, all values retrieved are `nil`

- Utilise [`append_infor_for_payload`](https://github.com/reidmorrison/semantic_logger/blob/6a7be665a8c6eaf5b9441d8aabc16ec47b3b0231/docs/rails.md#adding-custom-data-to-the-rails-completed-log-message) to add custom messages to append to our logs.

- Merge `params` as well, as we set those in `LogQueryParams`
- Remove methods to do this in our custom logger

## Guidance to review

Ideally this would be in the application controller, however we aren't inheriting from those, so I've kept the current pattern.
## Things to check

### DIFF
--- a/app/controllers/api_docs/api_docs_controller.rb
+++ b/app/controllers/api_docs/api_docs_controller.rb
@@ -1,6 +1,6 @@
 module APIDocs
   class APIDocsController < ActionController::Base
-    include LogQueryParams
+    include RequestQueryParams
     layout 'application'
 
   private
@@ -8,7 +8,7 @@ module APIDocs
     def append_info_to_payload(payload)
       super
 
-      payload.merge!(log_query_params)
+      payload.merge!(request_query_params)
     end
   end
 end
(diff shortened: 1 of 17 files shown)


vigneshmsft:
will the params appear as top level fields? or under `params.{id}` ?

cwrw:
@vigneshmsft they will appear as top level fields in the payload, example of `application_form_id` and and `support_user_id`:
```
{"application":"Semantic Logger","environment":"development","time":"2021-05-14T11:28:35.736+01:00","level":"info","level_index":2,"pid":85411,"thread":"puma threadpool 001","duration_ms":8552.392999990843,"duration":"8.552s","name":"SupportInterface::ApplicationFormsController","message":"Completed #show","payload":{"controller":"support_interface/application_forms","action":"show","params":{"application_form_id":"1297"},"format":"HTML","method":"GET","path":"/support/applications/1297","status":200,"view_runtime":625.27,"db_runtime":94.9,"application_form_id":"1297","support_user_id":1,"allocations":468938,"status_message":"OK"},"domain":"127.0.0.1","hosting_environment":"development","service":null}
```

vigneshmsft (approved):

PR #4708, opened 2021-05-11
Backfill `opened_on_apply_at` for all courses using audits

mnacos:
## Context

Following on from: https://github.com/DFE-Digital/apply-for-teacher-training/pull/4702

Add a data migration updating `opened_on_apply_at` for all courses that have had an 'open' event in the audits. Last such event wins in terms of timestamp.

## Guidance to review

The `open_on_apply` field is boolean, not nullable and with a default value of `false`. Therefore, there are two types of open event within the audits table:
- Course was created with `open_on_apply` set to `false` and was later opened, which results in  `"open_on_apply":[false, true]` within audited_changes
- Course was created with `open_on_apply` set to `true` (e.g. a test course). This will have `"open_on_apply":true` within audited_changes

To avoid timeouts on production, considering the size of the audits table and the lack of appropriate indexes, I have modified the query to be run in batches. We'll run this migration manually to monitor and retry, if needed.

Does 2020-03-26 make sense as a value for `opened_on_apply_at` for courses that were created before auditing was added to courses? See comments.

## Link to Trello card

https://trello.com/c/DG1rdUbK

## Things to check

### DIFF
--- a/app/services/data_migrations/backfill_opened_on_apply_at_from_audits.rb
+++ b/app/services/data_migrations/backfill_opened_on_apply_at_from_audits.rb
@@ -1,0 +1,24 @@
+module DataMigrations
+  class BackfillOpenedOnApplyAtFromAudits
+    TIMESTAMP = 20210511084457
+    MANUAL_RUN = false
+
+    def change
+      sql = <<~SET_OPENED_ON_APPLY_AT_FROM_AUDITS.squish
+        UPDATE courses c
+        SET opened_on_apply_at = courses_with_open_events.last_opened_at
+        FROM (
+          SELECT auditable_id AS course_id, max(created_at) AS last_opened_at
+          FROM audits
+          WHERE
+            auditable_type = 'Course'
+            AND audited_changes->>'open_on_apply' IN ('true', '[false, true]')
+          GROUP BY auditable_id
+        ) AS courses_with_open_events
+        WHERE c.id = courses_with_open_events.course_id
+      SET_OPENED_ON_APPLY_AT_FROM_AUDITS
+
+      ActiveRecord::Base.connection.execute(sql)
+    end
+  end
+end
--- a/app/services/data_migrations/backfill_opened_on_apply_at_from_audits.rb
+++ b/app/services/data_migrations/backfill_opened_on_apply_at_from_audits.rb
@@ -1,24 +1,40 @@
 module DataMigrations
   class BackfillOpenedOnApplyAtFromAudits
     TIMESTAMP = 20210511084457
-    MANUAL_RUN = false
+    MANUAL_RUN = true
 
     def change
-      sql = <<~SET_OPENED_ON_APPLY_AT_FROM_AUDITS.squish
+      Course
+        .where(open_on_apply: true)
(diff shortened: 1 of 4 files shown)


cwrw (commented):
Looks good 👍 Do you know how long this will take to run on prod? as worried it might time out

mnacos:
> Looks good  Do you know how long this will take to run on prod? as worried it might time out

I tested the aggregate query on production and it times out, as I should have expected. I have prepared a new version that performs the same query in batches and seems to complete (based on a modified version minus the UPDATE). I'll change this migration to be a manual one, so we can monitor and retry if needed.

Another thing to note is when the `open_on_apply` column was introduced the default value was true, so existing courses acquired `open_on_apply` without relevant audits to extract timestamps from. For these courses (created before `open_on_apply` was a field), maybe we could hard-code `opened_on_apply_at` to 
2019-11-18.

mnacos:
Looking closer at the history of courses, we only started auditing changes to courses on 2020-03-26, so about 1500  courses will not have audits for `open_on_apply` events. We have about 7500 courses open on Apply, so it's worth proceeding, I think.

app/services/data_migrations/backfill_opened_on_apply_at_from_audits.rb
```diff
@@ -0,0 +1,40 @@
+module DataMigrations
+  class BackfillOpenedOnApplyAtFromAudits
+    TIMESTAMP = 20210511084457
+    MANUAL_RUN = true
+
+    def change
+      Course
+        .where(open_on_apply: true)
+        .where('created_at < \'2020-04-01\'')
+        .update_all('opened_on_apply_at = \'2020-03-26\'')
```
cwrw, line 10: Why did we go with `created_at` less than `2020-04-01` instead of `2020-03-27`? 
I was also having a think, maybe going with when we added the `open_on_apply` field, to set the timestamp might be more accurate as you mentioned `2019-11-18` rather than when we started auditing?

mnacos: 2020-03-26 is the timestamp from the commit and is approximate in that it does not tell us when the PR was merged and the code deployed. Another approach would be to set this timestamp for all open courses first, then overwrite it with information from audits, if available. The reason I haven't done this is I'd like to see if there are any records not covered by the current approach.

I think something around 2020-03-26 is more accurate than when the field was first introduced. This is because the default value for `open_on_apply` in the database was changed soon afterwards to `false` in another migration and because without the `audited` statement, introduced around 2020-03-26, we don't know when these courses were actually opened. 

I guess we could have a tiered approach, in which courses created before 2019-11-18 are set to 2019-11-18, courses before 2020-04-01 are set to 2020-03-26 and information from audits trumps all, if available. What do you think? Considering these courses are from a previous cycle, I don't think it matters that much what we choose.

cwrw: Thanks for the explanation, let's keep as is I reckon, best not overcomplicate it as like you said this data is old anyway.

cwrw (approved):

PR #4709, opened 2021-05-11
Refactor ApplicationChoiceComponent

george-bit:
## Context

Tech debt

This PR refactors the rows method on this view component to draw from several private methods for greater efficiency and
readability.

## Guidance to review

Could this be refactored any further?

## Link to Trello card

https://trello.com/c/9FLsTXEg/3362-refactor-the-supportinterfaceapplicationchoicecomponent

## Things to check

### DIFF
--- a/app/components/support_interface/application_choice_component.rb
+++ b/app/components/support_interface/application_choice_component.rb
@@ -10,77 +10,21 @@ module SupportInterface
     end
 
     def rows
-      rows = if FeatureFlag.active?(:support_user_reinstate_offer) && application_choice.declined? && !application_choice.declined_by_default
-               [
-                 { key: 'Status',
-                   value: render(SupportInterface::ApplicationStatusTagComponent.new(status: application_choice.status)),
-                   action: 'Reinstate offer',
-                   action_path: support_interface_application_form_reinstate_offer_path(application_form_id: @application_choice.application_form.id, application_choice_id: @application_choice.id) },
-               ]
-             else
-               [{ key: 'Status', value: render(SupportInterface::ApplicationStatusTagComponent.new(status: application_choice.status)) }]
-             end
-
-      if application_choice.offer?
-        rows << { key: 'Offer made at', value: application_choice.offered_at.to_s(:govuk_date_and_time) }
-        rows << { key: 'Decline by default at', value: application_choice.decline_by_default_at.to_s(:govuk_date_and_time) } if application_choice.decline_by_default_at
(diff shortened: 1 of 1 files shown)


app/components/support_interface/application_choice_component.rb
```diff
@@ -10,92 +10,139 @@ def initialize(application_choice)
...
-      rows << { key: 'Sent to provider at', value: application_choice.sent_to_provider_at.to_s(:govuk_date_and_time) } if application_choice.sent_to_provider_at
-      rows << { key: 'Reject by default at', value: application_choice.reject_by_default_at.to_s(:govuk_date_and_time) } if application_choice.reject_by_default_at && application_choice.decision_pending?
-      rows << { key: 'Decline by default at', value: application_choice.decline_by_default_at.to_s(:govuk_date_and_time) } if application_choice.decline_by_default_at && application_choice.offer?
+    def rejection_reason_row
+      return unless rejection_reasons_text
 
+      { key: 'Rejection reason', value: rejection_reasons_text }
+    end
+
+    def sent_to_provider_at_row
```
davidgisbey, line 100: super nitty but 

`return unless application_choice.sent_to_provider_at`

davidgisbey (approved):
Looks good to me George. This is way easier to parse now

app/components/support_interface/application_choice_component.rb
```diff
@@ -10,92 +10,139 @@ def initialize(application_choice)
...
+      { key: 'Sent to provider at', value: application_choice.sent_to_provider_at.to_s(:govuk_date_and_time) } if application_choice.sent_to_provider_at
+    end
+
+    def reject_by_default_at_row
```
davidgisbey, line 104: I think this is covered on line 84 right?

george-bit: Not quite... as I understand `reject_by_default_at` is a date in the future that is generated when app choice is decision pending.

Line 84 displays `rejected_at` which is a date generic to any rejected application choice. The logic around this draws from `rejected_by_default` which is a boolean set to true when an application choice is rejected by default.

TL;DR - This is present/future tense theother is past tense and they pull dates from different db columns.

davidgisbey: Right, sorry about that i missed the tense 👍  Good to go then

JR-G (approved):
Nice one George! Much cleaner 🙌

PR #4712, opened 2021-05-12
Allow Providers and Courses to be synced incrementally

TobyRet:
## Context
At the moment the course sync is quite taxing on our infrastructure and it takes several minutes to run.

The new public TTAPI supports fetching only providers and courses which have changed over the API, which means we don't need to iterate over every single provider (and their courses) every time. We can also potentially sync more frequently. 

The TTAPI will be updated so that [Providers are touched whenever, a Course is updated](https://github.com/DFE-Digital/teacher-training-api/pull/1910).

- Add the ability to sync providers and courses incrementally
- By default `incremental_sync` is set to true however, this can be set to `false` if a full sync is required.

## Guidance to review

- Fire up the TTAPI locally (pointing to this PR branch if not merged - https://github.com/DFE-Digital/teacher-training-api/pull/1910) and in the TTAPI database update an attribute on a course the current recruitment cycle (this will also `touch` the associated Provider). In your Apply `.env.development` set `TEACHER_TRAINING_API_BASE_URL=http://127.0.0.1:3001/api/public/v1`
- run `TeacherTrainingPublicAPI::IncrementalSyncAllProvidersAndCourses.call` from the console or go to the tasks dashboard in the support UI and run the sync that way
- You should see in the logs that only this Provider and Course are synced, and that the course has been updated with the new attribute value in the Apply DB

## Link to Trello card

https://trello.com/c/RsbjA4sq/3299-incrementally-sync-courses-from-the-ttapi

## Things to check

### DIFF
--- a/app/models/teacher_training_public_api/sync_check.rb
+++ b/app/models/teacher_training_public_api/sync_check.rb
@@ -14,6 +14,10 @@ module TeacherTrainingPublicAPI
       Redis.current.get(LAST_SUCCESSFUL_SYNC)
     end
 
+    def self.updated_since
+      (Time.zone.parse(last_sync) - 2.hours)
+    end
+
     def self.check
       if last_sync.nil?
         false
--- a/app/services/teacher_training_public_api/sync_all_providers_and_courses.rb
+++ b/app/services/teacher_training_public_api/sync_all_providers_and_courses.rb
@@ -1,15 +1,19 @@
 module TeacherTrainingPublicAPI
   class SyncAllProvidersAndCourses
-    def self.call(recruitment_cycle_year: ::RecruitmentCycle.current_year)
+    def self.call(recruitment_cycle_year: ::RecruitmentCycle.current_year, incremental_sync: true)
       is_last_page = false
       page_number = 0
       until is_last_page
         page_number += 1
-        response = TeacherTrainingPublicAPI::Provider
+
+        scope = TeacherTrainingPublicAPI::Provider
           .where(year: recruitment_cycle_year)
           .paginate(page: page_number, per_page: 500)
-          .all
-        sync_providers(response, recruitment_cycle_year)
+        scope = scope.where(updated_since: TeacherTrainingPublicAPI::SyncCheck.updated_since) if incremental_sync
+        response = scope.all
+
+        sync_providers(response, recruitment_cycle_year, incremental_sync: incremental_sync)
+
         is_last_page = true if response.links.links['next'].nil?
       end
 
@@ -18,12 +22,12 @@ module TeacherTrainingPublicAPI
       raise TeacherTrainingPublicAPI::SyncError
     end
 
-    def self.sync_providers(providers_from_api, recruitment_cycle_year)
+    def self.sync_providers(providers_from_api, recruitment_cycle_year, incremental_sync: true)
       providers_from_api.each do |provider_from_api|
         TeacherTrainingPublicAPI::SyncProvider.new(
           provider_from_api: provider_from_api,
           recruitment_cycle_year: recruitment_cycle_year,
-        ).call
+        ).call(incremental_sync: incremental_sync)
       end
     end
 
--- a/app/services/teacher_training_public_api/sync_provider.rb
+++ b/app/services/teacher_training_public_api/sync_provider.rb
@@ -5,7 +5,7 @@ module TeacherTrainingPublicAPI
       @recruitment_cycle_year = recruitment_cycle_year
     end
 
-    def call(run_in_background: true, force_sync_courses: false)
+    def call(run_in_background: true, force_sync_courses: false, incremental_sync: true)
       @force_sync_courses = force_sync_courses
 
       provider_attrs = if existing_provider
@@ -17,15 +17,15 @@ module TeacherTrainingPublicAPI
                        end
 
       provider = create_or_update_provider(provider_attrs)
-      sync_courses(run_in_background, provider)
+      sync_courses(run_in_background, provider, incremental_sync: incremental_sync)
     end
 
-    def sync_courses(run_in_background, provider)
+    def sync_courses(run_in_background, provider, incremental_sync: true)
       if sync_courses?
         if run_in_background
-          TeacherTrainingPublicAPI::SyncCourses.perform_async(provider.id, @recruitment_cycle_year)
+          TeacherTrainingPublicAPI::SyncCourses.perform_async(provider.id, @recruitment_cycle_year, incremental_sync)
         else
-          TeacherTrainingPublicAPI::SyncCourses.new.perform(provider.id, @recruitment_cycle_year, run_in_background: false)
+          TeacherTrainingPublicAPI::SyncCourses.new.perform(provider.id, @recruitment_cycle_year, incremental_sync, run_in_background: false)
         end
       end
     end
--- a/config/clock.rb
+++ b/config/clock.rb
@@ -9,7 +9,7 @@ class Clock
   error_handler { |error| Raven.capture_exception(error) if defined? Raven }
 
   # Hourly jobs
-  every(1.hour, 'SyncAllFromTeacherTrainingPublicAPI', at: '**:00') { TeacherTrainingPublicAPI::SyncAllProvidersAndCoursesWorker.perform_async }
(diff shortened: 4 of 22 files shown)


app/models/teacher_training_public_api/sync_check.rb
```diff
@@ -14,6 +14,10 @@ def self.last_sync
       Redis.current.get(LAST_SUCCESSFUL_SYNC)
     end
 
+    def self.updated_since
+      (Time.zone.parse(last_sync) - 2.hours)
```
TobyRet, line 18: We want to offset by the length of time it takes to complete the sync. Is 2 hours enough?

app/services/teacher_training_public_api/sync_all_providers_and_courses.rb
```diff
@@ -1,15 +1,19 @@
 module TeacherTrainingPublicAPI
   class SyncAllProvidersAndCourses
-    def self.call(recruitment_cycle_year: ::RecruitmentCycle.current_year)
+    def self.call(recruitment_cycle_year: ::RecruitmentCycle.current_year, incremental_sync: true)
```
TobyRet, line 3: How do we want to toggle the incremental_sync? Via the support interface? `support/settings/tasks`

app/models/teacher_training_public_api/sync_check.rb
davidgisbey: Yea easily. It normally takes about 10-15 atm without incremental sync

app/services/teacher_training_public_api/sync_all_providers_and_courses.rb
davidgisbey: That's an interesting question. I definitely think you should be able to run if them the tasks. 

I wonder too if we should update the worker to occasionally run a full sync too. At the moment it runs every hour. We could potentially create a new attr like`fully_syned_at` and pass in `incramental_sync: false` if 3 days have passed and it's between 12pm & 7am or something like that.

davidgisbey:
This is looking really good 👍

app/services/teacher_training_public_api/sync_all_providers_and_courses.rb
TobyRet: Yeah sounds like a good idea 👍 
Will have a think about the UI - sounds like we could simply replace the existing sync task button with one for full sync and another for partial sync. We may want to display a warning if full sync is activated during daytime?

app/services/teacher_training_public_api/sync_provider.rb
```diff
@@ -17,15 +17,15 @@ def call(run_in_background: true, force_sync_courses: false)
...
-      sync_courses(run_in_background, provider)
+      sync_courses(run_in_background, provider, incremental_sync: incremental_sync)
     end
 
-    def sync_courses(run_in_background, provider)
+    def sync_courses(run_in_background, provider, incremental_sync: true)
       if sync_courses?
         if run_in_background
-          TeacherTrainingPublicAPI::SyncCourses.perform_async(provider.id, @recruitment_cycle_year)
+          TeacherTrainingPublicAPI::SyncCourses.perform_async(provider.id, @recruitment_cycle_year, incremental_sync)
```
TobyRet, line 26: I am really not liking this. Sidekiq's `perform_async` method does not accept keyword arguments. is there are better way?

spec/services/teacher_training_public_api/sync_courses_spec.rb
```diff
@@ -4,6 +4,8 @@
   include TeacherTrainingPublicAPIHelper
 
   describe '.call' do
+    let(:incremental_sync_false) { false }
```
TobyRet, line 7: Not good. Ideas?
https://github.com/DFE-Digital/apply-for-teacher-training/pull/4712/files#r631177409

app/services/teacher_training_public_api/sync_provider.rb
TobyRet: I was wondering if a FeatureFlag may be an alternative but then this causes an issue when we want to run either a full sync or incremental sync via the support UI 🤔

davidgisbey: Yea, thats really annoying that it doesn't accept keyword args 😠 I'm not sure there's a much better way. Maybe someone else will. I don't think it's the end of the world though.

app/services/teacher_training_public_api/sync_all_providers_and_courses.rb
TobyRet: Apart from updating the provider sync task description we have agreed to leave the UI as is
https://ukgovernmentdfe.slack.com/archives/CN1MCQCHZ/p1620894860454000

TobyRet: https://github.com/DFE-Digital/apply-for-teacher-training/pull/4712/commits/72eb011ee51ee4c615854c2eb547c650d3e1fb17

incrementally_sync_providers

TobyRet: https://github.com/DFE-Digital/apply-for-teacher-training/pull/4712/commits/c32c58c13890fadf4a876961b602c9d6961e0857

incrementally_sync_providers

app/services/teacher_training_public_api/sync_all_providers_and_courses.rb
```diff
@@ -18,12 +22,12 @@ def self.call(recruitment_cycle_year: ::RecruitmentCycle.current_year)
       raise TeacherTrainingPublicAPI::SyncError
     end
 
-    def self.sync_providers(providers_from_api, recruitment_cycle_year)
+    def self.sync_providers(providers_from_api, recruitment_cycle_year, incremental_sync: true)
```
duncanjbrown, line 25: Why do we need to pass this down here — won't it be the case that only providers that have been touched will be included in `providers_from_api`?

TobyRet: The flag gets passed down to `TeacherTrainingPublicAPI::SyncCourses` - I figured given a Provider can be updated independently of its courses we may as well optimise course sync also?

duncanjbrown: I think 90% of the benefit will come from reducing the number of providers whose courses we have to look at. But if we've done the work, great!

app/services/teacher_training_public_api/sync_all_providers_and_courses.rb
davidgisbey: This is great 👌

config/clock.rb
```diff
@@ -47,4 +47,6 @@ class Clock
   end
 
   every(1.day, 'Generate export for TAD', at: '23:59') { DataAPI::TADExport.run_daily }
+
+  every(3.days, 'FullSyncAllFromTeacherTrainingPublicAPI', at: '00:59') { TeacherTrainingPublicAPI::FullSyncAllProvidersAndCoursesWorker.perform_async }
```
davidgisbey, line 51: This is great 🙌

app/services/teacher_training_public_api/sync_all_providers_and_courses.rb
```diff
@@ -1,15 +1,19 @@
...
       until is_last_page
         page_number += 1
-        response = TeacherTrainingPublicAPI::Provider
+
+        scope = TeacherTrainingPublicAPI::Provider
           .where(year: recruitment_cycle_year)
           .paginate(page: page_number, per_page: 500)
-          .all
-        sync_providers(response, recruitment_cycle_year)
+        scope = scope.where(updated_since: TeacherTrainingPublicAPI::SyncCheck.updated_since) if incremental_sync
```
davidgisbey, line 12: So i have one question about this. This basically means that we get all the records updated in the last 2ish hours right?

Would it be better if we used this?

image


I'm just thinking that at some point the sync might fail a few times and we would lose data until a full sync happens 3 days later.

TobyRet: Yeah my thoughts exactly. We may need to take into account that the incremental sync fails so `TeacherTrainingPublicAPI::SyncCheck.updated_since` offsets `last_sync` by two hours. Can increase it if you like though to be safe?

`(Time.zone.parse(last_sync) - 2.hours)`

https://github.com/DFE-Digital/apply-for-teacher-training/pull/4712/files#diff-86a2f62cd0d36bb071a44016c9f545566bce203ff84184ddbb07031f374e6dfbR17

davidgisbey (approved):

PR #4713, opened 2021-05-12
Configure seconds value for BusinessTime end of workday

steventux:
## Context

The end of the business day was configured as `23:59` and the `end_of_day` date helper uses the time `23:59:59.999999999`.
These tests were failing when the revised dbd date fell at the end of the week or weekend ie. after the close of business. 
This meant depending on which day the test was run we'd see a failure because the rails helper calculates the end of day just outside the business day and any further business day calculations make the equality comparison off by one day.

Amend the `BusinessTime::Config.end_of_workday` value to include seconds. (There's no millisecond value stored against `BusinessTime::ParsedTime`). This fixed the `SetDeclineByDefault` tests which were relying a modified dbd date.

## Guidance to review

It's still possible for us to introduce tests which will fail in a similar way if we compare date time instances including millisecond values to instances of `BusinessTime::ParsedTime`. 
We usually call `end_of_day` in conjunction with `business_days` which means we get a millisecond value which works fine eg. https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/services/time_limit_calculator.rb#L23 or as in `spec/services/set_decline_by_default_spec.rb` [we disregard the millisecond value when making date time comparisons](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/spec/services/set_decline_by_default_spec.rb#L17).

## Link to Trello card

https://trello.com/c/1ATaBJJn/3656-investigate-causes-of-flakey-tests-due-to-time

## Things to check

### DIFF
--- a/config/initializers/business_time.rb
+++ b/config/initializers/business_time.rb
@@ -13,4 +13,4 @@ UCAS_HOLIDAYS.each do |date|
 end
 
 BusinessTime::Config.beginning_of_workday = '0:00 am'
-BusinessTime::Config.end_of_workday = '11:59 pm'
+BusinessTime::Config.end_of_workday = '11:59:59 pm'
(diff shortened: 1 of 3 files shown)


SamJamCul (approved):
Looks good - thanks for going through the millisecond exception with me, it looks like we'll just need to remember if it ever shows up

PR #4714, opened 2021-05-12
JS enhancements for "Add another" further conditions

benswannack:
## Context
We added the ability to add and remove further conditions previously. This required refreshing the page each time though. This PR adds progressive JS enhancements for that feature to avoid page reloads.

Add Rails UJS
Use remote form to submit using JS

## Guidance to review
Submit does not work well with validations

## Link to Trello card
https://trello.com/c/u0fZBMtg/3613-implement-js-progressive-enhancements-to-infinite-offer-conditions

## Things to check

### DIFF
--- a/app/controllers/provider_interface/offer/conditions_controller.rb
+++ b/app/controllers/provider_interface/offer/conditions_controller.rb
@@ -27,16 +27,19 @@ module ProviderInterface
         if add_another_condition?
           @wizard.add_empty_condition
 
+          @focus = anchor_for_further_condition
           respond_to do |format|
             format.js { render :add_another }
-            format.html { redirect_to action: action, anchor: anchor_for_further_condition }
+            format.html { redirect_to action: action, anchor: @focus }
           end
         elsif remove_condition_param.present?
           @wizard.remove_condition(remove_condition_param)
+
+          @focus = anchor_for_further_condition
           respond_to do |format|
             @condition_id = remove_condition_param
             format.js { render :remove }
-            format.html { redirect_to action: action, anchor: anchor_for_further_condition }
+            format.html { redirect_to action: action, anchor: @focus }
           end
         else
           submit_form(action: action)
--- a/app/controllers/provider_interface/offer/conditions_controller.rb
+++ b/app/controllers/provider_interface/offer/conditions_controller.rb
@@ -26,7 +26,11 @@ module ProviderInterface
 
         if add_another_condition?
           @wizard.add_empty_condition
-          redirect_to action: action, anchor: anchor_for_further_condition
+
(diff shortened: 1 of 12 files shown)


app/helpers/add_another_button_helper.rb
```diff
@@ -0,0 +1,16 @@
+module AddAnotherButtonHelper
```
benswannack, line 1: Add some specs for this

app/views/provider_interface/offer/conditions/add_another.js.erb
```diff
@@ -0,0 +1,13 @@
+<%= render 'provider_interface/offer/conditions/set_add_button_visibility' %>
```
benswannack, line 1: Add a system spec to cover the JS behaviour

app/controllers/provider_interface/offer/conditions_controller.rb
```diff
@@ -61,9 +72,15 @@ def submit_form(action:)
           @wizard.remove_empty_conditions!
           @wizard.save_state!
 
-          redirect_to [action, :provider_interface, @application_choice, :offer, @wizard.next_step]
+          respond_to do |format|
+            format.js { render js: "window.location.replace('#{polymorphic_url([action, :provider_interface, @application_choice, :offer, @wizard.next_step])}')" }
+            format.html { redirect_to [action, :provider_interface, @application_choice, :offer, @wizard.next_step] }
```
benswannack, line 77: Commonize the redirect URL

app/controllers/provider_interface/offer/conditions_controller.rb
```diff
@@ -61,9 +72,15 @@ def submit_form(action:)
...
+          end
         else
-          render action
+          respond_to do |format|
+            format.js { render js: 'window.location.reload()' }
```
benswannack, line 81: This doesn't work:
- The HTML method relies on running the validations on the wizard, and then rendering the view with the already validated wizard object
- This just redirects to the new action, which makes a new (unvalidated) wizard.

paulrobertlloyd:
Didn't we add UJS before and then remove it for some reason? Faint memories… @duncanjbrown

benswannack:
> Didn't we add UJS before and then remove it for some reason? Faint memories… @duncanjbrown

When we added it before, it made all forms `remote: true` by default. That was since changed ~fixed~ in rails, so the other forms will still be submitted normally.

paulrobertlloyd:
@benswannack 😅 – I’m not going mad then!

benswannack:
Closing this for now, because there seems to be a simpler way that solves all the problems.

#4721

PR #4716, opened 2021-05-12
Add an expanded Qualifications export

malcolmbaig:
## Context
The Qualifications export currently only contains a subset of each
candidate's qualifications. It needs to cover all qualifications in the
system so that a more complete analysis can take place.

Add a new export class that:
- Contains details about every qualification for any given candidate
- Maintains the relationship between application form and course choices
  that was present in the previous export as well. Note that this means
  that each set of qualification data per candidate is repeated in the
  data table for every course choice.
## Guidance to review

- Switch on `expanded_quals_export` feature flag
- Download Qualifications export from support interface
- Review export

This PR will be followed by a separate documentation update for the export once Milan/Mike have had a chance to review and any changes are agreed.

## Link to Trello card

https://trello.com/c/JxXFJ5KG
## Things to check

### DIFF
--- a/app/services/support_interface/expanded_qualifications_export.rb
+++ b/app/services/support_interface/expanded_qualifications_export.rb
@@ -2,7 +2,7 @@ module SupportInterface
   class ExpandedQualificationsExport
     def data_for_export
       application_choices = ApplicationChoice
-        .select(:id, :application_form_id, :status, :course_option_id, :rejection_reason, :structured_rejection_reasons)
+        .select(:id, :application_form_id, :status, :course_option_id)
         .includes(:course_option, :course, :provider, application_form: [:application_qualifications])
         .order(:application_form_id)
 
@@ -22,7 +22,6 @@ module SupportInterface
             recruitment_cycle_year: application_form.recruitment_cycle_year,
 
             choice_status: application_choice.status,
-            rejection_reason: application_choice.structured_rejection_reasons || application_choice.rejection_reason,
             course_code: course.code,
             provider_code: course.provider.code,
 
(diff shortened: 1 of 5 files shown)


app/services/support_interface/expanded_qualifications_export.rb
```diff
@@ -0,0 +1,48 @@
...
+            subject: qualification.subject,
+            predicted_grade: qualification.predicted_grade,
+            grade: qualification.grade,
+            institution_name: qualification.institution_name,
+            institution_country: qualification.institution_country,
+            international: qualification.international,
+            comparable_uk_degree: qualification.comparable_uk_degree,
+            non_uk_qualification_type: qualification.non_uk_qualification_type,
+            comparable_uk_qualification: qualification.comparable_uk_qualification,
+            constituent_grades: qualification.constituent_grades,
```
davidgisbey, line 42: wondering if these could be grouped slightly more intuitively. Maybe `constituent_grades` should be on line 36? and we could have `non_uk_qualification_type ` on line 37 followed by all the related attrs like `institution_country ` and `comparable_uk_degree` etc.

davidgisbey:
I won't approve yet as i think it would be good for Mike and Milan to get eyes on it, but it looks great to me 👍

app/services/support_interface/expanded_qualifications_export.rb
malcolmbaig: yep, definitely - have pushed a change

davidgisbey:
Great. Looks to me 👍

app/services/support_interface/expanded_qualifications_export.rb
davidgisbey: Great. Looks to me 👍

malcolmbaig:
Made some changes following a discussion with Milan, going to merge this and see how it looks with production data.

davidgisbey (approved):

PR #4720, opened 2021-05-12
Delete all course audits to reduce table size

mnacos:
## Context

We have decided to drop all course-related audits in order to reduce the size of the `audits` table by about 20 million rows.

Delete all course-related audits.

## Guidance to review

Batching mechanism from https://github.com/DFE-Digital/apply-for-teacher-training/pull/4717 seems to have worked well.

## Link to Trello card

https://trello.com/c/6UMC4zRm

## Things to check


cwrw (approved):
👍 🔥

cwrw (commented):
ah actually can we stop auditing courses, before we run this, because they'll come back

PR #4721, opened 2021-05-12
JS enhancements for "Add another" further conditions

benswannack:
## Context
We added the ability to add and remove further conditions previously. This required refreshing the page each time though. This PR adds progressive JS enhancements for that feature to avoid page reloads.

Add Rails UJS
Use remote form to submit using JS
Add JS system spec for the feature

## Guidance to review
This is pretty simple codewise. We render the whole form again serverside, and inject it into the page with js. Are there any limitations to this approach?

To test: Go though the flow in the review app. Try this with JS enabled and disabled and ensure the behaviour is the same

## Link to Trello card
https://trello.com/c/u0fZBMtg/3613-implement-js-progressive-enhancements-to-infinite-offer-conditions

## Things to check

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -135,7 +135,7 @@ jobs:
         run: bundle exec rake db:setup
 
       - name: ${{ matrix.tests }} tests with feature flags ${{ matrix.feature-flags }}
-        run: bundle exec --verbose rspec --exclude-pattern ${EXCLUDE_PATTERN} --pattern ${INCLUDE_PATTERN} --format documentation
+        run: apk add chromium chromium-chromedriver && bundle exec --verbose rspec --exclude-pattern ${EXCLUDE_PATTERN} --pattern ${INCLUDE_PATTERN} --format documentation
         env:
           INCLUDE_PATTERN: ${{ matrix.include-pattern }}
           EXCLUDE_PATTERN: ${{ matrix.exclude-pattern }}
--- a/app/views/provider_interface/offer/conditions/_form.html.erb
+++ b/app/views/provider_interface/offer/conditions/_form.html.erb
@@ -1,4 +1,4 @@
-<%= form_with model: form_object, url: provider_interface_application_choice_offer_conditions_path(application_choice), method: form_method do |f| %>
+<%= form_with model: form_object, url: provider_interface_application_choice_offer_conditions_path(application_choice), method: form_method, local: false do |f| %>
   <%= f.govuk_error_summary %>
 
   <h1 class="govuk-heading-l">
--- a/spec/support/capybara.rb
+++ b/spec/support/capybara.rb
@@ -5,8 +5,26 @@ if ENV['TEST_ENV_NUMBER']
   Capybara.server_port = 9887 + ENV['TEST_ENV_NUMBER'].to_i
 end
 
+Capybara.register_driver :chrome_headless do |app|
+  options = ::Selenium::WebDriver::Chrome::Options.new
+
+  options.add_argument('--headless')
+  options.add_argument('--no-sandbox')
+  options.add_argument('--disable-dev-shm-usage')
+  options.add_argument('--disable-gpu')
+  options.add_argument('--window-size=1400,1400')
+
+  Capybara::Selenium::Driver.new(app, browser: :chrome, options: options)
+end
+
+Capybara.javascript_driver = :chrome_headless
+
 RSpec.configure do |config|
   config.before(:each, type: :system) do
-    driven_by(:rack_test)
+    if self.class.metadata[:js] == true
+      driven_by(:chrome_headless)
+    else
+      driven_by(:rack_test)
+    end
   end
 end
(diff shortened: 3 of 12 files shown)


.github/workflows/build.yml
```diff
@@ -135,7 +135,7 @@ jobs:
         run: bundle exec rake db:setup
 
       - name: ${{ matrix.tests }} tests with feature flags ${{ matrix.feature-flags }}
-        run: bundle exec --verbose rspec --exclude-pattern ${EXCLUDE_PATTERN} --pattern ${INCLUDE_PATTERN} --format documentation
+        run: apk add chromium chromium-chromedriver && bundle exec --verbose rspec --exclude-pattern ${EXCLUDE_PATTERN} --pattern ${INCLUDE_PATTERN} --format documentation
```
benswannack, line 138: There may be a better place to put this, but we can't put it in the Dockerfile, as non-runtime dependencies are removed.

spec/support/capybara.rb
```diff
@@ -5,8 +5,27 @@
   Capybara.server_port = 9887 + ENV['TEST_ENV_NUMBER'].to_i
 end
 
+Capybara.register_driver :chrome_headless do |app|
```
benswannack, line 8: The default selenium chrome headless driver doesn't work in Docker. See https://stackoverflow.com/questions/50610316/capybara-headless-chrome-in-docker-returns-devtoolsactiveport-file-doesnt-exist

spec/support/capybara.rb
```diff
@@ -5,8 +5,27 @@
...
+
+Capybara.javascript_driver = :chrome_headless
+
 RSpec.configure do |config|
   config.before(:each, type: :system) do
     driven_by(:rack_test)
   end
+
+  config.before(:each, type: :system, js: true) do
+    WebMock.disable_net_connect!(allow_localhost: true)
```
benswannack, line 28: Needed for the chrome headless driver. Do we need to set this back to what it was before the example runs?

spec/system/provider_interface/make_offer_js_spec.rb
```diff
@@ -0,0 +1,133 @@
...
+
+  let(:provider_user) { create(:provider_user, :with_dfe_sign_in) }
+  let(:provider) { provider_user.providers.first }
+  let(:ratifying_provider) { create(:provider) }
+
+  let(:application_form) { build(:application_form, :minimum_info) }
+  let(:course) { build(:course, :open_on_apply, provider: provider, accredited_provider: ratifying_provider) }
+  let(:course_option) { build(:course_option, course: course) }
+
+  scenario 'Setting offer conditions' do
```
benswannack, line 15: None of the behaviour I've implemented is any different from the non-JS behaviour, so it's hard to prove that JS is running and working.

It might be nice to have this same spec run in JS and non-JS modes

spec/support/capybara.rb
benswannack: I tried this in an `around` block that resets back to disabling everything. It looks like something sends a request to clear the session after the end of the around hook (and the `after` one too), which fails if we block requests again after.

Can we just set this in `spec/support/webmock.rb`?

spec/system/provider_interface/set_offer_conditions_js_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+
+  def and_the_default_conditions_are_checked
+    expect(find("input[value='Fitness to train to teach check']", visible: false)).to be_checked
+    expect(find("input[value='Disclosure and Barring Service (DBS) check']", visible: false)).to be_checked
+  end
+
+  def when_i_add_a_further_condition(text:, index:)
+    click_on 'Add another condition'
+    fill_in("provider_interface_offer_wizard[further_conditions][#{index}][text]", with: text)
+    expect(page.current_url).not_to include("#provider-interface-offer-wizard-further-conditions-#{index}-text-field")
```
benswannack, line 92: This is the only line that fails if the test runs in non-JS mode, but that should be good enough!

app/views/provider_interface/offer/conditions/_form.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= form_with model: form_object, url: provider_interface_application_choice_offer_conditions_path(application_choice), method: form_method do |f| %>
+<%= form_with model: form_object, url: provider_interface_application_choice_offer_conditions_path(application_choice), method: form_method, local: false do |f| %>
```
benswannack, line 1: Submitting the form using AJAX has accessibility implications. We're going to redo this without.

benswannack:
Closing because we are taking a non-ajax approach. See -> #4754

app/views/provider_interface/offer/conditions/_form.html.erb
benswannack: More details on this:
1. The idea of this feature is to speed up the data entry process in this form. Having to go back to the server for each button press results in no gain. (All it really achieves is no page reload).
2. We don't communicate that the form is loading very well (if at all), which is not accessible

PR #4722, opened 2021-05-12
Link back to personal details review page when editing

frankieroberto:
This changes the back link so that it no longer relies upon the HTTP referer header when editing an answer to a question.

This fixes a bug where if you edit a question, and then get a validation error and get redirected back to the same page, the back link links to the page you're already on.

### DIFF
--- a/app/presenters/candidate_interface/personal_details_review_presenter.rb
+++ b/app/presenters/candidate_interface/personal_details_review_presenter.rb
@@ -63,7 +63,7 @@ module CandidateInterface
         key: I18n.t('application_form.personal_details.english_main_language.label'),
         value: @languages_form.english_main_language&.titleize,
         action: ('if English is your main language' if @editable),
-        change_path: candidate_interface_languages_path,
+        change_path: candidate_interface_edit_languages_path,
       }
     end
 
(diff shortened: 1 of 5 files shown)


paulrobertlloyd:
@frankieroberto For completeness sake, think we need to make the same change for `languages/edit.erb.html` (it’s a deprecated view, but still available for older applications, I think)

frankieroberto:
@paulrobertlloyd good point – I didn’t know about that extra page..! Fixed in 1ee4c98. Also spotted a separate bug which was that the "Change" link for languages was going to the path for "new" flow rather than the "edit" flow, which is fixed in 380503b.

george-bit (changes requested):
Looks good just 1 comment

app/presenters/candidate_interface/personal_details_review_presenter.rb
```diff
@@ -63,7 +63,7 @@ def english_main_language_row
         key: I18n.t('application_form.personal_details.english_main_language.label'),
         value: @languages_form.english_main_language&.titleize,
         action: ('if English is your main language' if @editable),
-        change_path: candidate_interface_languages_path,
+        change_path: candidate_interface_edit_languages_path,
       }
     end
```
george-bit, line 69: I think that the change path on `other_language_details_row` needs to be updated too 👇

frankieroberto: @george-bit thanks, changed in 8285b75. Although I confess that I have no idea when/how these are used...!

george-bit (approved):
👍

PR #4572, opened 2021-04-22
Create an 'undo' action for support agents to reinstate a declined offer

JR-G:
## Context

There has been a reoccurring request for developers, who are on support, to reinstate an offer that has been accidentally declined by the candidate. This requires the dev to go into the support console, find the application and update it there.

This PR adds functionality for support agents to be able to reinstate the offer from the support UI, without the need for a developer to be involved.

## Changes proposed in this pull request

- Add a feature flag that is then used on the `ApplicationChoiceComponent` to conditionally render a 'Reinstate offer' link
- Add a new view that the user is directed to after clicking the link, that will contain guidance on when this feature should/can be used, a mandatory text box for the Zendesk request link and a confirmation checkbox
- A new form and controller to handle this logic and validate that there is a Zendesk ticket - in the correct format - and that the support user has confirmed that they've read the guidance
- A new `ReinstateDeclinedOffer` service that is used to update the course choice status, reset the DBD days (currently 10) and also add an audit comment

## Guidance to review

- The reinstate offer page is still quite rough and needs a bit of cleaning up. Even though the support user has clicked through from the application choice, would it be helpful to also list the application choice they are about to update on this page as well - just for another visual sense check? Or rather than having the title as a generic _"Are you sure you want to reinstate the offer for this course choice?"_, be specific about which course choice is going to be reinstated?
- **Policy guidance still needs to be confirmed**
- Is there enough "friction" to reinstating the offer?
- How are the tests - what can I do to improve these and is there anything I haven't covered?

## Link to Trello card

https://trello.com/c/1k17OycL/3265-create-undo-actions-for-applicationchoice-states

## Things to check

- [ ] This code does not rely on migrations in the same Pull Request
- [ ] If this code includes a migration adding or changing columns, it also backfills existing records for consistency
- [ ] API release notes have been updated if necessary
- [ ] This code does not rely on the addition/removal of Azure config environment variables in the same Pull Request
- [ ] New environment variables have been [added to the Azure config](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/docs/environment-variables.md#azure-hosting-devops-pipeline)

PR #4724, opened 2021-05-13
Support interface revert a rejection

stevehook:
## Context

This PR will allow support users to revert accidental rejections by a provider. To revert a rejection a support user must enter a support ticket URL.

The design used in this PR is based on https://github.com/DFE-Digital/apply-for-teacher-training/pull/4572.

- [x] Add a new _Revert rejection_ link to the status line on application page in the support interface.
image

- [x] The above link is to a new confirmation page:
image

- [x] Implement a form and service backing the UI above.
- [x] New system spec

## Guidance to review

- The content for the confirmation is adapted from the reinstate offer feature. It needs some work, in particular I don't know what the guidance should be notifications for candidate and provider.

## Link to Trello card

https://trello.com/c/LPKMNuUN/3361-add-functionality-to-the-support-ui-to-revert-a-rejection

## Things to check

### DIFF
--- a/app/views/support_interface/application_forms/application_choices/confirm_revert_rejection.html.erb
+++ b/app/views/support_interface/application_forms/application_choices/confirm_revert_rejection.html.erb
@@ -1,0 +1,46 @@
+<% content_for :browser_title, title_with_error_prefix('Revert rejection', @application_choice.errors.any?) %>
+<% content_for :before_content, govuk_back_link_to(support_interface_application_form_path(application_form_id: params[:application_form_id]), 'Back to application') %>
+
+<%= form_with(
+    model: @form,
+    url: support_interface_application_form_revert_rejection_path(application_form_id: params[:application_form_id], application_choice_id: params[:application_choice_id]),
+    method: :patch
+  ) do |f| %>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <%= f.govuk_error_summary %>
+      <h1 class="govuk-heading-l">
+        Are you sure you want to revert this rejection?
+      </h1>
+
+      <p class="govuk-body">A rejection can only be reverted if:</p>
+      <ul class="govuk-list govuk-list--bullet">
+        <li>it was rejected within 5 working days of the request to revert and,</li>
+        <li>the candidate has not accepted any other offers</li>
+      </ul>
+      <p class="govuk-body">In order to revert the rejection you must first contact the provider to confirm that they agree to this.</p>
+      <p class="govuk-body">Once the rejection has been reverted, please email the candidate using the macro.</p>
+      <p class="govuk-body">There are separate macros if the request was made after 5 working days or if the candidate has already accepted another offer.</p>
+    </div>
+  </div>
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <%= f.govuk_text_field(
+              :audit_comment_ticket,
+              label: {
+                text: 'Zendesk ticket URL',
+                size: 'm',
+              },
+              rows: 1,
+              hint: { text: 'For example https://becomingateacher.zendesk.com/agent/tickets/12345' },
+            ) %>
+    </div>
+  </div>
+
+  <%= f.govuk_check_boxes_fieldset :accept_guidance, legend: nil do %>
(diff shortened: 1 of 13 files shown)


spec/system/support_interface/revert_rejection_spec.rb
```diff
@@ -0,0 +1,81 @@
+require 'rails_helper'
+
+RSpec.feature 'Revert an accidental rejection' do
+  include DfESignInHelpers
+
+  scenario 'View structured reasons for rejection', with_audited: true do
```
malcolmbaig: nit: worth correcting the scenario here so the commit makes sense in isolation (can see it's corrected later on)

spec/components/support_interface/application_choice_component_spec.rb
```diff
@@ -39,6 +39,32 @@
...
+      expect(result.css('.govuk-summary-list__actions a')[0].attr('href')).to include(
+        Rails.application.routes.url_helpers.support_interface_application_form_revert_rejection_path(
+          application_form_id: rejected_application_choice.application_form.id,
+          application_choice_id: rejected_application_choice.id,
+        ),
+      )
+      expect(result.css('.govuk-summary-list__actions').text.strip).to include('Revert rejection')
+    end
+
+    it 'Does not renders a link to the revert rejection page when application was rejected by default' do
```
malcolmbaig, line 57: typo: renders

spec/services/support_interface/revert_rejection_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+      described_class.new(
+        application_choice: application_choice,
+        zendesk_ticket: zendesk_ticket,
+      ).save!
+
+      expect(application_choice).to be_awaiting_provider_decision
+      expect(application_choice.audits.last.comment).to include(zendesk_ticket)
+    end
+  end
+end
```
malcolmbaig, line 18: Quite liked the testing pattern used [here](https://github.com/DFE-Digital/apply-for-teacher-training/pull/4572/files#diff-6e3f92ab4664c6b3c021cb2511b292d4a484c9c4acec2bde45f3ba7f1dfda083) where DeclineOffer is used to make the state change and we expect equality between the original choice and the updated one. Would it make sense to do similar here?

malcolmbaig (approved):
Code looks good, and works as expected in review app. A few minor comments left inline. Maybe needs a designer to take a look as well.

app/views/support_interface/application_forms/application_choices/confirm_revert_rejection.html.erb
```diff
@@ -0,0 +1,45 @@
...
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <%= f.govuk_error_summary %>
+      <h1 class="govuk-heading-l">
+        Are you sure you want to revert this rejection?
+      </h1>
+
+      <p class="govuk-body">A rejection can only be reverted if:</p>
+      <ul class="govuk-list govuk-list--bullet">
+        <li>it was rejected within 5 working days of the request to revert and,</li>
```
paulrobertlloyd, line 18:
```suggestion
        <li>it was rejected within 5 working days of the request to revert</li>
```

app/views/support_interface/application_forms/application_choices/confirm_revert_rejection.html.erb
```diff
@@ -0,0 +1,45 @@
...
+      <p class="govuk-body">Once the rejection has been reverted, please email the candidate to let them know.</p>
+    </div>
+  </div>
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <%= f.govuk_text_field(
+              :audit_comment_ticket,
+              label: {
+                text: 'Zendesk ticket URL',
```
paulrobertlloyd, line 31: Given we are using the same label (and hint) value for auditing changes like this, should we localise these strings to me this a bit DRYer?

paulrobertlloyd (changes requested):
Just a few small things.

stevehook:
> Oh, I was thinking all occurrences use one set of strings? Is that a good idea?

I wasn't sure about this. There is a school of thought that says that duplication in i18n files is ok because different things in different contexts may need different content. But I don't think that applies here so I've moved them all to more a generic key.

paulrobertlloyd (approved):

PR #4725, opened 2021-05-13
Correct documentation about qualification IDs in the API

duncanjbrown:
## Context

A provider pointed out that we claim these IDs aren't guaranteed to be unique, but they are since #4017 

Fix the docs to say the IDs are unique


benswannack (approved):
Whoops, thanks for addressing!

PR #4728, opened 2021-05-13
Install nodejs & yarn for local development in docker container instance

vigneshmsft:

install nodejs & yarn on the web container (this is to enable running local development/running js tests)
remove ci. prefix in test commands,

## Things to check

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,48 +20,41 @@ build: ## Create a new image
 .PHONY: setup
 setup: build ## Set up a clean database and node_modules folder for running the app or the specs in docker
 	docker-compose down -v
-	docker-compose run --rm web bundle exec rake db:setup
+	docker-compose up -d -V --no-build
+	docker-compose exec web bundle exec rake db:setup
+	docker-compose exec web apk add nodejs yarn
 
-.PHONY: az_setup
-az_setup: ## Set up a clean database and node_modules folder for running the app or the specs in docker
-	touch .env ## Create an empty .env file if it does not exist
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -20,48 +20,41 @@ build: ## Create a new image
...
-	touch .env ## Create an empty .env file if it does not exist
-	docker-compose run --rm web bundle exec rake db:setup
+.PHONY: stop
+stop: ## bring down the containers
+	docker-compose down -v
 
 .PHONY: test
 test: ## Run the linters and specs
-	docker-compose run --rm web /bin/sh -c "bundle exec rake"
+	docker-compose exec web /bin/sh -c "yarn install && bundle exec rake"
```
benswannack, line 33: The default `rake` doesn't run in docker for me locally due to parallel_rspec failing - potentially it can't run in a container(?). I think this would be fine if it just ran `rspec`, but if we can solve then that's better!

vigneshmsft: @benswannack we discussed this is due to RAILS_ENV and the gem groupings?

benswannack: Looks like that is part of the issue. We also need to set up the database to work with parallel rspec, which requires `bundle exec rake parallel:create && bundle exec rake parallel:prepare` to run. Once that's done, each test fails because of `Error connecting to Redis on localhost:6379 (Errno::EADDRNOTAVAIL)`

benswannack (approved):
This is generally an improvement, so approving. We can revisit the `bundle exec rake` that doesn't work in the future

PR #4730, opened 2021-05-13
Add export type to notifications exports

mnacos:
## Context

These exports run but are not accessible because `export_type` has not been set when they were triggered.

Set it when scheduling it. Backfill `export_type` for all previous notifications exports.

Do the same for daily TAD applications exports.

## Guidance to review

Straightforward

## Link to Trello card

https://trello.com/c/82W7yz61

## Things to check

### DIFF
--- a/app/services/data_api/tad_export.rb
+++ b/app/services/data_api/tad_export.rb
@@ -1,15 +1,16 @@
 module DataAPI
   class TADExport
-    EXPORT_NAME = 'Daily export of applications for TAD'.freeze
-
     def self.run_daily
-      data_export = DataExport.create!(name: EXPORT_NAME)
+      data_export = DataExport.create!(
+        name: 'Daily export of applications for TAD',
+        export_type: :tad_applications,
+      )
       DataExporter.perform_async(DataAPI::TADExport, data_export.id)
     end
 
     def self.all
       DataExport
-        .where(name: EXPORT_NAME)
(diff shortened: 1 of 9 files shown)


app/services/data_api/tad_export.rb
```diff
@@ -3,7 +3,10 @@ class TADExport
     EXPORT_NAME = 'Daily export of applications for TAD'.freeze
 
     def self.run_daily
-      data_export = DataExport.create!(name: EXPORT_NAME)
+      data_export = DataExport.create!(
+        name: EXPORT_NAME,
+        export_type: :tad_applications,
+      )
```
duncanjbrown, line 9: I think we can start using `export_type` in `self.all` in this class too now 🎉

duncanjbrown (approved):
Lovely — one small improvement we can make to `TADExport`!

app/services/data_api/tad_export.rb
```diff
@@ -3,13 +3,16 @@ class TADExport
...
       DataExporter.perform_async(DataAPI::TADExport, data_export.id)
     end
 
     def self.all
       DataExport
-        .where(name: EXPORT_NAME)
+        .where(name: EXPORT_NAME, export_type: :tad_applications)
```
duncanjbrown, line 15: I think we can lose `EXPORT_NAME` as well?

PR #4731, opened 2021-05-13
Fix typo in EthnicBackgroundHelper and update affected ethnic_group and hesa_ethnicity values

steventux:
## Context

A typo in a view helper which supplies the ethnicity background for the _Black, African, Black British or Caribbean_  ethnic group meant that the code for `hesa_ethnicity` was not being set.

This PR fixes the typo changing _Carribean_ to _Caribbean_ in the view helper and adds a data migration which finds application forms with the ethnic group typo value updates them with the correct the ethnic group value and HESA ethnicity code.

## Link to Trello card

https://trello.com/c/z0tg8C7o/3654-hesaethnicity-sometimes-null-or-incorrect-in-api-response-depending-on-ethnicity-chosen

## Things to check

### DIFF
--- a/spec/services/data_migrations/fix_misspelling_of_caribbean_ethnic_group_and_set_hesa_codes_spec.rb
+++ b/spec/services/data_migrations/fix_misspelling_of_caribbean_ethnic_group_and_set_hesa_codes_spec.rb
@@ -1,0 +1,24 @@
+require 'rails_helper'
+
+RSpec.describe DataMigrations::FixMisspellingOfCaribbeanEthnicGroupAndSetHesaCodes do
+  it 'fixes the misspelled Carribean ethnicity background value' do
+    equality_and_diversity = {
+      'sex' => 'female',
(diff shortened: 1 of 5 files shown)


spec/services/data_migrations/fix_misspelling_of_caribbean_ethnic_group_and_set_hesa_codes_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'rails_helper'
+
+RSpec.describe DataMigrations::FixMisspellingOfCaribbeanEthnicGroupAndSetHesaCodes do
```
benswannack, line 3: Could we pad this spec out with tests for `hesa_ethnicity` already being set and different recruitment cycle years (not sure if that actually makes a difference)

benswannack (approved):
Nice fix :D 

We can potentially add more test coverage, otherwise 👍

PR #4732, opened 2021-05-13
Remove useless bigquery config in .env files

steventux:
## Context

Introduced in https://github.com/DFE-Digital/apply-for-teacher-training/pull/4601
These were intended as an example but prevent the app from starting in development without the correct credentials.

Removes useless env placeholders for BigQuery config.

## Things to check


mnacos (approved):
:)

PR #4740, opened 2021-05-14
Change qualification card components to ENIC/NARIC

SamJamCul:
## Context

The body that makes comparative decisions about qualifications changed their name from NARIC to ENIC. On the candidate side, we've updated the interface so that they know they're entering ENIC details - however, we can't track which details existing candidates input, which means at the moment, there may be ENIC or NARIC versions of information showing up provider-side. 

That means we want to change the qualification card components to account for this, and indicate that the ENIC reference may also have been entered in as a NARIC reference when the candidate was applying. 

Old: 

image

New: 

image

I've added the additional locale, rather than renaming an existing one, as the existing `short_name` is used in other parts of the app.

## Guidance to review

You'll need to either find or modify an application form's `application_qualification` so that one of them has the `enic_reference` and `comparable_uk_degree` filled in. This will trigger the additional part of the component that sets off the `ENIC or NARIC` text.

## Link to Trello card

https://trello.com/c/QIkk8wCr/3644-naric-to-enic-organisation

## Things to check


cwrw (approved):
👍

PR #4742, opened 2021-05-14
Fix query timeouts when refreshing feature metrics

stevehook:
## Context

The feature metrics dashboard is updated hourly by a background job that runs a number of DB queries in order to populate the dashboard. One (at least one) such query is frequently timing out - it's just too slow. This PR attempts to optimise that query.

- [x] Add a GIN index to the `audits.audited_changes` column because this is the column on which we need to define the most important query condition.
- [x] Add a redundant condition to the query that checks for the existence the `work_history_completed` key in the `audited_changes` column. This should take advantage of the index added above. There is already a condition that checks the _value_ of the `work_history_completed` key but it doesn't look like it's possible to use an index for this.

Before index:

```
bat_apply_development=# explain analyze SELECT "audits"."auditable_id"
FROM "audits"
WHERE "audits"."auditable_type" = 'ApplicationForm'
  AND (audited_changes ? 'work_history_completed')
  AND (audited_changes#>>'{work_history_completed, 1}' = 'true')
  AND (created_at BETWEEN '2021-3-31' AND '2021-4-30');
┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                                                QUERY PLAN                                                                 │
├───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Bitmap Heap Scan on audits  (cost=9.14..204.28 rows=1 width=4) (actual time=0.849..1.897 rows=25 loops=1)                                 │
│   Recheck Cond: ((auditable_type)::text = 'ApplicationForm'::text)                                                                        │
│   Filter: ((audited_changes ? 'work_history_completed'::text) AND (created_at >= '2021-03-31 00:00:00'::timestamp without time zone) AND …│
│…(created_at <= '2021-04-30 00:00:00'::timestamp without time zone) AND ((audited_changes #>> '{work_history_completed,1}'::text[]) = 'tru…│
│…e'::text))                                                                                                                                │
│   Rows Removed by Filter: 90                                                                                                              │
│   Heap Blocks: exact=80                                                                                                                   │
│   ->  Bitmap Index Scan on index_audits_on_auditable_type_and_id  (cost=0.00..9.13 rows=114 width=0) (actual time=0.095..0.095 rows=115 l…│
│…oops=1)                                                                                                                                   │
│         Index Cond: ((auditable_type)::text = 'ApplicationForm'::text)                                                                    │
│ Planning Time: 0.738 ms                                                                                                                   │
│ Execution Time: 2.039 ms                                                                                                                  │
└───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
(9 rows)
```

After index:

```
bat_apply_development=# explain analyze SELECT "audits"."auditable_id"
FROM "audits"
WHERE "audits"."auditable_type" = 'ApplicationForm'
  AND (audited_changes ? 'work_history_completed')
  AND (audited_changes#>>'{work_history_completed, 1}' = 'true')
  AND (created_at BETWEEN '2021-3-31' AND '2021-4-30');
┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                                                QUERY PLAN                                                                 │
├───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Bitmap Heap Scan on audits  (cost=25.67..29.70 rows=1 width=4) (actual time=0.504..1.645 rows=25 loops=1)                                 │
│   Recheck Cond: (((auditable_type)::text = 'ApplicationForm'::text) AND (audited_changes ? 'work_history_completed'::text))               │
│   Filter: ((created_at >= '2021-03-31 00:00:00'::timestamp without time zone) AND (created_at <= '2021-04-30 00:00:00'::timestamp without…│
│… time zone) AND ((audited_changes #>> '{work_history_completed,1}'::text[]) = 'true'::text))                                              │
│   Rows Removed by Filter: 53                                                                                                              │
│   Heap Blocks: exact=70                                                                                                                   │
│   ->  BitmapAnd  (cost=25.67..25.67 rows=1 width=0) (actual time=0.095..0.098 rows=0 loops=1)                                             │
│         ->  Bitmap Index Scan on index_audits_on_auditable_type_and_id  (cost=0.00..9.14 rows=115 width=0) (actual time=0.050..0.050 rows…│
│…=115 loops=1)                                                                                                                             │
│               Index Cond: ((auditable_type)::text = 'ApplicationForm'::text)                                                              │
│         ->  Bitmap Index Scan on index_audits_on_audited_changes  (cost=0.00..16.28 rows=37 width=0) (actual time=0.039..0.040 rows=78 lo…│
│…ops=1)                                                                                                                                    │
│               Index Cond: (audited_changes ? 'work_history_completed'::text)                                                              │
│ Planning Time: 2.284 ms                                                                                                                   │
│ Execution Time: 1.683 ms                                                                                                                  │
└───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
(12 rows)
```

The amount of data in the test here isn't anywhere near enough to measure performance but we can see that the index does get utilised by the query - so it _should_ be faster.

## Guidance to review

- Does the approach make sense?
- How expensive is the index going to be to maintain? Especially wrt to the impact on write performance.
- I'm also concerned about the time it will take to create the index. (how long the migration will take).
- How can we test this before deployment to `production`?
- Note that there is a code change and migration in the same PR here. I _think_ that's reasonable here because the migration only adds an index - so the code change doesn't depend on it to run, though the (theoretical) performance gain does depend on the migration and code change being in place.

## Link to Trello card

https://trello.com/c/70B0qwOa/3395-investigate-query-timeouts-when-updating-feature-metrics

## Things to check


db/schema.rb
```diff
@@ -10,55 +10,103 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_05_07_131855) do
+ActiveRecord::Schema.define(version: 2021_05_14_135324) do
+
   create_sequence "application_choices_id_seq"
+  create_sequence "application_choices_id_seq1"
```
cwrw, line 16: can we remove the extra sequencing from being committed?

db/migrate/20210514135324_add_audits_audited_changes_index.rb
```diff
@@ -0,0 +1,7 @@
+class AddAuditsAuditedChangesIndex < ActiveRecord::Migration[6.1]
+  disable_ddl_transaction!
+
+  def change
+    add_index :audits, :audited_changes, using: :gin, algorithm: :concurrently
```
cwrw, line 5: will this be ok with our `statement_timeout`, I think it might take longer than 20 seconds?

cwrw: we mitigated this last time with: https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/db/migrate/20210319163050_add_auditable_type_id_index_to_audits.rb#L7

cwrw:
> How expensive is the index going to be to maintain? Especially wrt to the impact on write performance.
I think this might be an issue due to how different each audit change is between objects, and different models/field, but this is only theory not sure in practice

I'm also concerned about the time it will take to create the index. (how long the migration will take).
> How can we test this before deployment to production?
Vignesh created a db dump for us from prod last time, which we could test against

Although this seems ok, my concern still remains around the change we saw between Azure/PaaS where now we even have a significantly smaller table size but we are not utilising any optimised queries through the planner. I ran some index queries to check if we are hitting indexes as we should, and it doesn't seem we are (for audits). I wonder if we are missing something, a setting/autovacuum/reindex or just the query planner itself not optimised between the migration that we could help the issue. Happy to discuss suggestions

db/migrate/20210514135324_add_audits_audited_changes_index.rb
stevehook: makes, will do the same...

stevehook:
I've done some testing on this with a more realistic data set (local database that had `generate_test_data` for a few hours leading to about 600K audit records).

The results for the index in this PR are not great. It appeared to make a significant and consistent different when tested on a small dataset reducing query times from ~90ms to ~55ms. However with the larger dataset the timings were around 500-600ms with and without the index.

There are a lot of different kinds of index we can create on `jsonb` column. I tried a couple of others, this one makes the most difference reducing query times by an order of magnitude:

```
CREATE INDEX hash_test ON audits USING HASH ((audited_changes->'work_history_completed'));
```

This only works with the `->` and `=` operators. So the query changes to:

```
SELECT count("audits"."auditable_id")
FROM "audits"
WHERE (audited_changes->'work_history_completed' = '[false, true]')
```

leading to this query plan:

```
┌────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                                           QUERY PLAN                                                           │
├────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Aggregate  (cost=10549.81..10549.82 rows=1 width=8) (actual time=33.224..33.226 rows=1 loops=1)                                │
│   ->  Bitmap Heap Scan on audits  (cost=81.27..10541.66 rows=3260 width=4) (actual time=9.320..31.731 rows=17215 loops=1)      │
│         Recheck Cond: ((audited_changes -> 'work_history_completed'::text) = '[false, true]'::jsonb)                           │
│         Heap Blocks: exact=16223                                                                                               │
│         ->  Bitmap Index Scan on hash_test  (cost=0.00..80.45 rows=3260 width=0) (actual time=6.736..6.736 rows=17215 loops=1) │
│               Index Cond: ((audited_changes -> 'work_history_completed'::text) = '[false, true]'::jsonb)                       │
│ Planning Time: 0.121 ms                                                                                                        │
│ Execution Time: 33.270 ms                                                                                                      │
└────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
(8 rows)

Time: 33.668 ms
```

stevehook:
Parking this one for now following meeting because we are going to try some database config changes first.

PR #4743, opened 2021-05-14
Add equality and diversity response rate to the feature metric dashboard

george-bit:
## Context

So that Support users can easily view the rate of E&D survey completion, this value needs to be exposed on the Feature Metrics Dashboard.

This PR achieves this in much the same way as: https://github.com/DFE-Digital/apply-for-teacher-training/pull/4663

image

## Guidance to review
Please test in review app.

## Link to Trello card

https://trello.com/c/dSrr6jRd/3375-add-equality-and-diversity-response-rate-to-the-feature-metric-dashboard

## Things to check

### DIFF
--- a/app/models/equality_and_diversity_feature_metrics.rb
+++ b/app/models/equality_and_diversity_feature_metrics.rb
@@ -1,0 +1,54 @@
+class EqualityAndDiversityFeatureMetrics
+  include ActionView::Helpers::NumberHelper
+
+  def response_rate(
+    start_time,
+    end_time = Time.zone.now.end_of_day
+  )
+    success_count = application_forms_with_feedback(start_time, end_time).count.to_f
+    fail_count = application_forms_with_no_feedback(start_time, end_time).count.to_f
+    return nil if (fail_count + success_count).zero?
+
+    success_count / (fail_count + success_count)
+  end
+
+  def formatted_response_rate(
+    start_time,
+    end_time = Time.zone.now.end_of_day
+  )
+    format_as_percentage(response_rate(start_time, end_time))
+  end
+
+private
+
+  def format_as_percentage(ratio)
+    return 'n/a' if ratio.nil?
+
+    percentage = number_with_precision(
+      100.0 * ratio,
+      precision: 1,
+      strip_insignificant_zeros: true,
+    )
+    "#{percentage}%"
+  end
+
+  def application_forms_with_feedback(start_time, end_time)
+    ApplicationForm
+      .where(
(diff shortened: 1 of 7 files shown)


app/models/equality_and_diversity_feature_metrics.rb
```diff
@@ -0,0 +1,54 @@
...
+      precision: 1,
+      strip_insignificant_zeros: true,
+    )
+    "#{percentage}%"
+  end
+
+  def application_forms_with_feedback(start_time, end_time)
+    ApplicationForm
+      .where(
+        recruitment_cycle_year: RecruitmentCycle.current_year,
```
davidgisbey, line 38: Do we always scope the results on the feature metric dashboard to the current cycle? I'm just wondering if we should or not. I wouldn't have thought that we would as it means that we wouldn't really get anything out of it for the first 30 days of a new cycle.

Wdyt?

davidgisbey:
The code looks good George. I've just got one question on the scope of which application forms we include. I'm interested to hear what you and other people think.

app/models/equality_and_diversity_feature_metrics.rb
stevehook: @davidgisbey yes rightly or wrongly this is the way all the metrics are scoped right now. Something worth looking at across the board before we start the next cycle.

spec/models/equality_and_diversity_feature_metrics_spec.rb
```diff
@@ -0,0 +1,54 @@
+require 'rails_helper'
+
+RSpec.describe EqualityAndDiversityFeatureMetrics, with_audited: true do
+  subject(:feature_metrics) { described_class.new }
+
+  def create_application
+    application_form = create(:completed_application_form, submitted_at: Time.zone.now)
```
stevehook, line 7: Don't need the local variable here or in the next method.

app/models/equality_and_diversity_feature_metrics.rb
george-bit: Perhaps it's worth a discussion with  Support Users as to what would be most useful?

I guess current cycle makes sense from a perspective of keeping the info light and relevant. If we scoped it to across all cycles it would quickly beome not as useful.

Maybe there's scope for including `last_cycle` metrics as a comparison point? Appreciate the dashboard is pretty chunky atm and is continuing to grow.

stevehook (approved):
:+1: Code looks good and it's clearly working on the review app.

PR #4744, opened 2021-05-14
Update back links across Referee flow

george-bit:
## Context

As part of the work completed on `3261-references-save-at-end-of-flow`
this back link update should have been included. 

It will direct candidate back to edit form to prevent duplication of referee.

## Guidance to review

Test in review app.

## Link to Trello card

https://trello.com/c/GuiJ30Jg/3397-catch-all-give-back-links-explicit-targets

## Things to check


frankieroberto:
@george-bit this looks good, but there's a couple of other back links which need fixing, it think:

1) If you go to "Add a referee" > "Continue" > choose one, "Continue" > "Back" > "Back" > "Back" > "Back" – then you get stuck in a loop between the Reference type and name pages.

2) If you do the same as above but get as far as the "How do you know this referee..." page, and then use the Back links, you end up in a loop between the name page and that page.

The way I've implemented this in the past is to only add the `?return_to=summary` query string on the "Change" links on the summary pages, and then if that's not present, always link back to the previous page in the flow.

davidgisbey (approved):

george-bit:
@frankieroberto Think I've fixed it  🤞 

Please have a go in the review app and let me know

frankieroberto:
@george-bit it's getting closer!

Remaining issues:

1) In the Add referee flow, if you do "Add another referee" -> "Continue" -> "Back" then you should be back on the interstitial "Choose your referees" page, but instead end up on the review page.

2) When you do "Add another referee" the interstitial page should have a "Back" link to the review page, at the moment it always has "Back to application". It should still be "Back to the application" if you arrive at this page straight from the application page though via the "Add your references" link (ie when adding the very first reference).

3) From the "Check your answers before sending your request" page, if you click "Change" on the name, email address or relationship rows, and then click "Back", you end up on the References review page instead of the "Check your answers" page. Clicking the Change link for Reference type and then clicking "Back" works correctly though. Do we need to add a `?return_to=review_unsubmitted` query string for the others?

4) Where adding a new referee, if you do "Add referee" -> "Continue" -> select Academic, press "Continue" -> leave name field blank, press "Continue" then you get a "Back" link to the "What is the referees name?" page, rather than the reference type page. If you also keep pressing "Back" you just remain on the same page.

Sorry! This is quite a complicated flow! Give me a shout if you can't repeat any of these, or if we need to take a call on which ones to fix immediately and which to leave.

app/views/candidate_interface/references/candidate_name/edit.html.erb
```diff
@@ -0,0 +1,27 @@
+<% content_for :title, title_with_error_prefix(t('page_titles.references_candidate_name'), @reference_candidate_name_form.errors.any?) %>
+<% content_for :before_content, govuk_back_link_to(candidate_interface_references_review_path) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with(
+      model: @reference_candidate_name_form,
+      url: candidate_interface_references_update_candidate_name_path(@reference.id, return_to: params[:return_to]),
```
davidgisbey, line 8: I don't think we need this view as you are only able to edit this via the personal details section. This endpoint isn't in the routes either.

george-bit: My bad I think I added this in error thinking it was the 'name' attribute 🤦

davidgisbey:
There are probably more elegant solutions, bit i've done the work to fix all the backlinks. It definitely needs a review though. I've given it a decent manual test, but it would be good if someone else could try and break it too

george-bit:
Looking good David, much better to separate the two flows out,

Tested a manually and is now working really well. Just two minor comments inline:

1. When you go back from the name page in the 'new' flow the parameter for the referee type is not retained. I appreciate the difficulty in passing back the param in this specific circumstance only though.
2. When travelling to the edit pages, submitting the value returns the candidate to the review page if successful. Previously i had some logic implemented here that would redirect the candidate first to any other field which was not filled in and then subsequently to the unsubmitted review page. Appreciate it's simpler in it's current form though.

I don't think either of these area huge problem though and think we could probably merge as is 👍

davidgisbey:
Hey George

1. So i think it does for everything, but school-based as it passes it back as an underscore. Good spot!
2. I might be wrong, but i think generally if you click to edit a field you should always be returned to the page you chose to edit from rather than being directed back into a flow

Wdyt Frankie?

frankieroberto (approved):
Awesome work - all the back links work as expected, in all contexts, even when there are error messages shown.

The only remaining bug I can find is the one @george-bit mentioned:

> When you go back from the name page in the 'new' flow the parameter for the referee type is not retained.

Is this an easyish fix? I note that the type is in the URL path already?

george-bit:
👍

davidgisbey:
Yea all good. It was just a tiny fix

PR #4752, opened 2021-05-17
Fix queries running multiple times in application choices index

steventux:
## Context

Obtaining the count of applications and then checking whether we had applications to render was causing duplicate queries to run in the `/provider/applications` action and views.

Use Kaminari's `#total_count` method after `#load` to reduce the number of queries. [Kaminari will check to see if records have been loaded when determining the total count of records](https://github.com/kaminari/kaminari/blob/00657f252244413b512349f4de20c63ee478baa5/kaminari-activerecord/lib/kaminari/activerecord/active_record_relation_methods.rb#L25), this allows us to make one query for the page of applications _if_ the applications count is less than the page size. This won't always be the case and inevitably we'll see a query with limit and offset (fetching a page of results) _and_ a count query for total record count. 
This seemed to be the best compromise as it removes the EXISTS query being issued by the call to `any?` in the view. In some cases, (eg. pages with refined results from filtering) it will only issue one query for the results as the results are contained in one page. In most cases there will be a paginating query and a count query for applications.
 

## Guidance to review

Currently we perform 3 `ApplicationChoice` db queries when rendering the applications index (count, exists? and load). These queries can be seen in the development log.
This PR should reduce this so that for 1 page of results there should only be 1 `ApplicationChoice` db query and for more than 1 page of results there should be 2 queries (pagination and count).

## Link to Trello card

https://trello.com/c/d1lsWdCD/3681-fix-queries-running-multiple-times-in-applicationchoicesindex

## Things to check

### DIFF
--- a/app/queries/get_application_choices_for_providers.rb
+++ b/app/queries/get_application_choices_for_providers.rb
@@ -4,9 +4,11 @@ class GetApplicationChoicesForProviders
     :current_course_option,
     :provider,
     :site,
-    application_form: %i[candidate application_references application_work_experiences application_volunteering_experiences application_qualifications application_work_history_breaks],
-    course: %i[provider accredited_provider],
-    course_option: [{ course: %i[provider] }, :site],
+    {
+      application_form: %i[candidate application_references application_work_experiences application_volunteering_experiences application_qualifications application_work_history_breaks],
+      course: %i[provider accredited_provider],
+      course_option: [{ course: %i[provider] }, :site],
+    },
   ].freeze
 
   DEFAULT_RECRUITMENT_CYCLE_YEAR = RecruitmentCycle.years_visible_to_providers
--- a/app/controllers/provider_interface/application_choices_controller.rb
+++ b/app/controllers/provider_interface/application_choices_controller.rb
@@ -30,8 +30,7 @@ module ProviderInterface
         application_choices: with_includes.where(id: application_choices),
       )
 
-      @application_choices_count = application_choices.count
-      @application_choices = application_choices.page(params[:page] || 1).per(30)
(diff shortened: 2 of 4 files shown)


app/queries/get_application_choices_for_providers.rb
```diff
@@ -4,9 +4,9 @@ class GetApplicationChoicesForProviders
     :current_course_option,
     :provider,
     :site,
-    application_form: %i[candidate application_references application_work_experiences application_volunteering_experiences application_qualifications application_work_history_breaks],
-    course: %i[provider accredited_provider],
-    course_option: [{ course: %i[provider] }, :site],
+    { application_form: %i[candidate application_references application_work_experiences application_volunteering_experiences application_qualifications application_work_history_breaks] },
```
benswannack, line 7: What do the extra brackets do here?

steventux: I got a Rubocop warning https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/HashAsLastArrayItem so I did what I was told.

benswannack: Ah, can we just have one set of brackets around all three lines then?

benswannack (approved):
This all looks sensible 👍

app/queries/get_application_choices_for_providers.rb
steventux: :+1: https://github.com/DFE-Digital/apply-for-teacher-training/pull/4752/commits/b77fb872384727967d77f65057f9ae7f1e8f63bd

cwrw (approved):
looks good, now down to 2 queries. Couldn't find a trivial way for Kaminari total count

